PR #2067, opened 2023-10-11
Cache playwright browsers.

MitjaBezensek:
Adds caching for playwright browsers, should make the e2e tests run much faster.

CleanShot 2023-10-11 at 13 06 48

Here's how it looks like when there's a cache hit (added an empty commit to this PR to test it out)
CleanShot 2023-10-11 at 13 19 30

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know


ds300 (approved):
heck yeah! 📉

PR #2069, opened 2023-10-13
[fix] Don't select locked shapes on pointer up

steveruizok:
This PR fixes a bug where a locked shape could be selected when clicking the shape's label.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Place a no-fill geo shape
2. Lock the shape
3. Click on the middle of the shape (its label)
4. It should not be selected (previously it would be)

1. Place a no-fill shape in a frame
2. Lock the shape
3. Click on the middle of the shape or the edge of the shape
4. It should not be selected (previously it would be)

- [x] Unit Tests

### Release Notes

- Fix bug where locked shape could be selected by clicking on its label


Taha-Hassan-Git (approved):
🔐👍

PR #2070, opened 2023-10-13
[fix] reparenting locked shapes

steveruizok:
This PR allows locked shapes to be reparented. This fixes a bug where grouped and locked shapes would disappear when their parent group would be ungrouped.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Create two shapes.
2. Group the shapes.
3. Lock one shape.
4. Ungroup the group

- [x] Unit Tests

### Release Notes

- Fix a bug where grouped locked shapes would be deleted when ungrouped.


Taha-Hassan-Git (approved):
👨‍👦‍👦👍

PR #2072, opened 2023-10-13
Compact children when updating parents to children.

steveruizok:
This PR fixes (or rather buries) a bug that could occur when the parents to children array includes shapes that no longer exist.

### Change Type

- [x] `patch` — Bug fix


ds300 (approved):
fine by me. weird that this happens at all though. any deleted shapes should be mentioned in the diff 🤔

PR #2073, opened 2023-10-13
[fix] locked shape of opacity problem with eraser.pointing

momenthana:
locked shape of opacity problem with the pointing state of the eraser tool
this is only in the pointing (clicking) state.
I don't know if this is intended or a bug, but it's true that it's a little uncomfortable. 

blinks
blinks when the user starts to remove the handwriting on locked content.

Defs https://github.com/tldraw/tldraw/pull/2061

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Make a locked shape
2. Click erase over the shape
3. The shape shouldn't change opacity

### Release Notes

- locked shape of opacity problem with eraser.pointing
Before/after:
A B


steveruizok:
Hi @momenthana! That's a good catch on the bug, I'll review.

steveruizok (approved):

steveruizok:
Looks good! Thanks for the PR. 🙇

PR #2109, opened 2023-10-20
Remove (optional) from jsdocs

TodePond:
This PR removes all mentions of "(optional)" from our jsdocs. This is for:

* consistency — many of our jsdocs don't mention "(optional)" for optional parameters. The developer is expected to use the type definition to find this out. But it's a bit unclear because we use "(optional)" in many places too.
* docs site — on our docs site, we use type definitions to figure out what is optional, and what isn't. We use that info to denote optional parameters. It looks funny having two "(optional)"s on a page. We *could* strip them, but it's probably better to just remove them at the source.

image

### Change Type

### Release Notes

- dev: Removed duplicate/inconsistent `(optional)`s from docs

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -164,15 +164,15 @@ export interface TLEditorOptions {
 	 */
 	getContainer: () => HTMLElement
 	/**
-	 * (optional) A user defined externally to replace the default user.
+	 * A user defined externally to replace the default user.
 	 */
 	user?: TLUser
 	/**
-	 * (optional) The editor's initial active tool (or other state node id).
+	 * The editor's initial active tool (or other state node id).
 	 */
 	initialState?: string
 	/**
-	 * (optional) Whether to infer dark mode from the user's system preferences. Defaults to false.
+	 * Whether to infer dark mode from the user's system preferences. Defaults to false.
 	 */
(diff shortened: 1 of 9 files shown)


steveruizok:
I'd actually rather strip these from the parsing of the JSDocs rather from the JSDocs themselves!

ds300:
> I'd actually rather strip these from the parsing of the JSDocs rather from the JSDocs themselves!

why's that? 

Also worth noting that these are `TSDoc` comments which is a different thing from `JSDoc` and has quite different syntax.

I'm fine with this change. It should still be plenty clear that the params are optional for developers via intellisense and whatnot.

ds300 (approved):

steveruizok (approved):

steveruizok:
> > I'd actually rather strip these from the parsing of the JSDocs rather from the JSDocs themselves!
> 
> why's that?
> 
> Also worth noting that these are `TSDoc` comments which is a different thing from `JSDoc` and has quite different syntax.
> 
> I'm fine with this change. It should still be plenty clear that the params are optional (via intellisense and whatnot).

My logic here is that the VS Code inline documentation (the view of the tsdocs in vscode) don't indicate which are optional. It's true though that these docs are fairly fragile, so the less information we have in them the better. But I'm convinced, let's land this!

PR #2114, opened 2023-10-23
Fix an issue with `addEventListener` in old Safari (pre v14)

MitjaBezensek:
Seems Safari only added `MediaQueryList: change event` in version 14. This is a workaround for older versions.
https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Fixes an issue with `addEventListener` on MediaQueryList object in old versions of Safari.

### DIFF
--- a/packages/editor/src/lib/hooks/useDocumentEvents.ts
+++ b/packages/editor/src/lib/hooks/useDocumentEvents.ts
@@ -23,7 +23,11 @@ export function useDocumentEvents() {
 			const media = matchMedia(mqString)
 			// Safari only started supporting `addEventListener('change',...) in version 14
 			// https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event
-			const safariCb = (_ev: any) => updatePixelRatio
+			const safariCb = (ev: any) => {
+				if (ev.type === 'change') {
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/hooks/useDocumentEvents.ts
```diff
@@ -21,9 +21,24 @@ export function useDocumentEvents() {
 			}
 			const mqString = `(resolution: ${window.devicePixelRatio}dppx)`
 			const media = matchMedia(mqString)
-			media.addEventListener('change', updatePixelRatio)
+			// Safari only started supporting `addEventListener('change',...) in version 14
+			// https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event
+			const safariCb = (ev: any) => {
+				if (ev.type === 'change') {
```
MitjaBezensek, line 27: Not sure if this is needed, I think we might only have a `change` event. Example from react native
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/AccessibilityInfo/index.js#L30

PR #2119, opened 2023-10-26
[hot take] Improve auto-size text

TodePond:
This is a hot-take PR that improves auto-sizing text in two ways.

## Context

There are currently two ways to make a text shape.
* Clicking
* Dragging

2023-10-26 at 11 43 19 - Rose Cat

Clicking makes an auto-sizing text shape.
Dragging makes a fixed-size text shape.

* You can switch from auto to fixed by dragging the left/right edge of a text shape.
* You can switch from fixed to auto by double-clicking the left/right edge of a text shape.

2023-10-26 at 11 48 08 - Peach Armadillo

> We already have this difference in place. It's a bit of extra complexity, but it's quite easy to learn, and we do a good job at guessing what the user wants.

## Others

In Excalidraw, text works quite differently. When you click, you immediately start making an 'auto-sizing' shape. You can't make a fixed size text shape.

2023-10-26 at 11 50 11 - Copper Toucan

Ignoring other issues, it feels like you're typing straight onto the canvas. I've seen streamers use it for this purpose. eg: Theo, ThePrimeagen. They basically write prose straight onto the canvas. 

There's no indicator around the text, so it feels like you're typing straight onto the screen. You're not dealing with the bounds of the shape. You can write more freely.

image

Also, it defaults to left-aligned, so you can write more freely. Your text doesn't "move" around. You (and your viewers) can see the text you've already written without it moving around.

2023-10-26 at 11 56 34 - Azure Toad

## Proposal

Let's do the best of both worlds!

* Excalidraw style for auto-size text. (ie: when you click)
* tldraw style for fixed text. (ie: when you drag)

2023-10-26 at 12 02 47 - Maroon Seahorse

We *already* have a difference between these two interactions. Let's make the most of it! And let's make it *even clearer* to the user when they're in one mode VS another. Let's let them learn it quicker.

2023-10-26 at 12 04 28 - Magenta Bandicoot

> Also this might make a good twitter thread. (Maybe worth changing the text in some of those gifs though).

## Compromise

At the very least, I think we should do the indicator change.

(Even if we don't do the alignment change).

### Change Type

I think it's probably major because some people might be relying on the automatic 'center-align' of clicking.

### Test Plan

1. Make a text shape by clicking.
2. While editing, you shouldn't be able to see the indicator.
3. It should also be left-aligned regardless of your current setting.
4. Make a text shape by dragging.
5. It should follow the style panel's setting.
6. And the indicator should be visible while editing.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Changed the behaviour of auto-size text shapes.

### DIFF
--- a/packages/editor/api/api.json
+++ b/packages/editor/api/api.json
@@ -6827,7 +6827,7 @@
             {
               "kind": "Method",
               "canonicalReference": "@tldraw/editor!Editor#animateShape:member(1)",
-              "docComment": "/**\n * Animate a shape.\n *\n * @param partial - The shape partial to update.\n *\n * @param options - (optional) The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 })\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 }, { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
+              "docComment": "/**\n * Animate a shape.\n *\n * @param partial - The shape partial to update.\n *\n * @param options - The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 })\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 }, { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
               "excerptTokens": [
                 {
                   "kind": "Content",
@@ -6901,7 +6901,7 @@
             {
               "kind": "Method",
               "canonicalReference": "@tldraw/editor!Editor#animateShapes:member(1)",
-              "docComment": "/**\n * Animate shapes.\n *\n * @param partials - The shape partials to update.\n *\n * @param options - (optional) The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShapes([{ id: 'box1', type: 'box', x: 100, y: 100 }])\n * editor.animateShapes([{ id: 'box1', type: 'box', x: 100, y: 100 }], { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
+              "docComment": "/**\n * Animate shapes.\n *\n * @param partials - The shape partials to update.\n *\n * @param options - The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShapes([{ id: 'box1', type: 'box', x: 100, y: 100 }])\n * editor.animateShapes([{ id: 'box1', type: 'box', x: 100, y: 100 }], { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
               "excerptTokens": [
                 {
                   "kind": "Content",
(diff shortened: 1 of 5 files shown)


packages/editor/api/api.json
```diff
@@ -6827,7 +6827,7 @@
             {
               "kind": "Method",
               "canonicalReference": "@tldraw/editor!Editor#animateShape:member(1)",
-              "docComment": "/**\n * Animate a shape.\n *\n * @param partial - The shape partial to update.\n *\n * @param options - (optional) The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 })\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 }, { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
+              "docComment": "/**\n * Animate a shape.\n *\n * @param partial - The shape partial to update.\n *\n * @param options - The animation's options.\n *\n * @example\n * ```ts\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 })\n * editor.animateShape({ id: 'box1', type: 'box', x: 100, y: 100 }, { duration: 100, ease: t => t*t })\n * ```\n *\n * @public\n */\n",
```
TodePond, line 6830: These files really wanted to gen. Maybe they got missed off that previous PR.

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -83,6 +83,7 @@ export class Pointing extends StateNode {
 					props: {
 						text: '',
 						autoSize: true,
+						align: 'start',
```
TodePond, line 86: Always left align when clicking

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -141,6 +142,8 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
 
 	indicator(shape: TLTextShape) {
 		const bounds = this.editor.getShapeGeometry(shape).bounds
+		const editor = useEditor()
+		if (shape.props.autoSize && editor.editingShapeId === shape.id) return null
```
TodePond, line 146: Don't show indicator when editing an auto-size text shape.
We still show it when you're not editing.

ds300:
+1 on indicator hiding
+1 on left-align by default, but maybe not like this? I think I'd prefer it to just be the default alignment option for the text tool so the behavior is consistent and it always does what's shown in the style panel? Not sure it makes especially more sense for fixed-width to be center-aligned anyway.

steveruizok:
A few first impressions:

- I think dropping the indicator for text as you're editing it is a good one, at least when it's autosizing.
- I think resetting the text to left-aligned on each new shape is a miss; even if we defaulted to left aligned, we wouldn't want to replace a prior user choice for alignment
- setting the initial alignment to left could work... but alignment is also shared with geo shape labels
  - so we would want to split these out into different styles for text alignment and label alignment, which would also mean displaying both alignments when a user has a text shape and a geo shape selected; 
  - or else we insist that all labels default to center, which would also feel strange; 
  - or else we do proper tool-based settings so that the user's choice for text is preserved for only text shapes and geo shapes is preserved for only geo shape, but this would be a fundamental change.

TodePond:
Thanks for your points! I see the complications with alignment.

It's worth bearing in mind that alignment actually does nothing for autosize shapes. See gif 👇
2023-10-26 at 13 40 40 - Salmon Junglefowl

Maybe we can completely remove alignment from the picture somehow for autosize shapes. 🤔
Maybe treat all alignments the same way, somehow...

TodePond:
I've made a new PR which only contains the indicator change, which there's clearer support for: #2120

steveruizok:
Closing this one—it's produced a good side PR, though we'll need to do other systems work before we come back to this feature.

PR #2150, opened 2023-11-06
[feature] Things on the canvas

steveruizok:
This PR adds two new component overrides to the editor's `components` slot. They are: 

- `<OnTheCanvas/>`, which renders inside of the html layer that scales and translates with the camera
- `<InFrontOfTheCanvas/>`, which renders in front of the canvas but behind any UI elements, and which does not scale / pan with the camera.

Kapture 2023-11-06 at 12 19 15

### Change Type

- [x] `minor` — New feature

### Test Plan

1. See the "on the canvas" example.

### Release Notes

- [editor] Adds two new components, `OnTheCanvas` and `InFrontOfTheCanvas`.

### DIFF
--- a/packages/editor/src/lib/components/default-components/DefaultInFrontOfTheCanvas.tsx
+++ b/packages/editor/src/lib/components/default-components/DefaultInFrontOfTheCanvas.tsx
@@ -1,0 +1,9 @@
+import { ComponentType } from 'react'
+
+/** @public */
+export type TLInFrontOfTheCanvas = ComponentType<object>
+
+/** @public */
+export const DefaultInFrontOfTheCanvas: TLInFrontOfTheCanvas = () => {
+	return <div className="tl-things-in-front-of-the-canvas" />
+}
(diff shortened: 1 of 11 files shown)


packages/editor/src/lib/components/default-components/DefaultInFrontOfTheCanvas.tsx
```diff
@@ -0,0 +1,9 @@
+import { ComponentType } from 'react'
+
+/** @public */
+export type TLInFrontOfTheCanvas = ComponentType<object>
+
+/** @public */
+export const DefaultInFrontOfTheCanvas: TLInFrontOfTheCanvas = () => {
+	return <div className="tl-things-in-front-of-the-canvas" />
+}
```
TodePond, line 9: I believe this is unused, so I think we can remove it

packages/editor/src/lib/components/default-components/DefaultOnTheCanvas.tsx
```diff
@@ -0,0 +1,13 @@
...
+/** @public */
+export type TLOnTheCanvas = ComponentType<object>
+
+/** @public */
+export const DefaultOnTheCanvas: TLOnTheCanvas = () => {
+	const rContainer = useRef<HTMLDivElement>(null)
+	useTransform(rContainer, 0, 0)
+
+	return <div className="tl-things-on-the-canvas" ref={rContainer} />
+}
```
TodePond, line 13: I think this is also unused.

TodePond (approved):
Just some nits, looks useful!

Issue #2156, opened 2023-11-07
Text in geo shapes is not centred


ds300:
1. Create a small-ish geo shape.
2. Add a text label.
3. Start typing.

If it doesn't happen, try making the shape smaller, or type more!

Some examples here
https://staging.tldraw.com/r/deploy-2023-11-07?viewport=0,0,853,1302&page=page:bv1533422fvZVpezYNPAC

![Image](https://github.com/tldraw/tldraw/assets/2523721/6d997e8a-2c03-466f-bb45-c4faf0d9399b)

It's even worse with mono font

![Image](https://github.com/tldraw/tldraw/assets/2523721/fd4e5ad6-433e-4396-a31f-e39b3d90391f)

TodePond:
![Image](https://github.com/tldraw/tldraw/assets/15892272/31b88ea9-c5a1-4a43-a70d-6e79bc5d5571)

PR #2168, opened 2023-11-07
Fix the problem with text not being correctly aligned in small geo shapes.

MitjaBezensek:
Seems like the `HTMLContainer` didn't have the correct size, so text label was positioned a bit further up than it should be.

Fixes [#2156](https://github.com/tldraw/tldraw/issues/2156)

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Create a small geo shape
2. Add some text to it. Test it with text that wraps to new lines.
3. Make sure text is centred.

### Release Notes

- Fixes position of Text labels in geo shapes.


TodePond (approved):
Yep I caused this one! ✋

PR #2172, opened 2023-11-07
[fix] Frame label not following staying aligned correctly on rotation

### LINKED ISSUE #2157: frame labels slow to update pos on rotate
the label should jump to the top as you're dragging

Image

Taha-Hassan-Git:
closes #2157 
@ds300 fixed this

The Frame shape component now re-renders on rotation, ensuring that the label position is always correct.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Make a frame shape
2. rotate it 90 degrees
before: it would stay where it was until the props of the shape changed i.e on resize
now: it updates immediately

### Release Notes

- Frame labels immediately update their position on rotation.

### DIFF
--- a/packages/tldraw/api/api.json
+++ b/packages/tldraw/api/api.json
@@ -14583,7 +14583,7 @@
               "text": "export interface TLUiContextMenuProps "
             }
           ],
-          "fileUrlPath": "packages/tldraw/src/lib/ui/components/ContextMenu.tsx",
+          "fileUrlPath": "packages/tldraw/.tsbuild-api/lib/ui/components/ContextMenu.d.ts",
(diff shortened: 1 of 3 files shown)


steveruizok:
Hey, I pushed a change that improves this. The bug here was that the canonicalized rotation was not updating when the shape updated. This was a regression that likely occurred when I moved the component from `track` to `useValue`. We don't want to have the shape render every time the shape changes—but we DO want its heading (and only its heading) to render every time its rotation changes.

steveruizok (approved):

PR #2175, opened 2023-11-07
[fix] Prevent navigating back with the trackpad

### LINKED ISSUE #2155: Trackpad swiping with two fingers left/right navigates back/forwards
this is probably a focus management issue. we were also seeing pinch/zoom bugs probably related to that.

TodePond:
This PR prevents navigating forward/back by swiping two fingers on the trackpad.

I have no idea why this was only happening on dotcom. It doesn't happen in bublic. This feels like a strange fix, but maybe it's more effective in general anyway!

I think we should do some cross-browser testing on this one, before hitting the release button.

fixes #2155 

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Navigate to dotcom from another page (so that your browser has something in the history) (ie: not a fresh tab).
2. Hover over the UI with the cursor (eg: the toolbar).
3. With two fingers, swipe left on the trackpad.
4. The page should not go back to the previous page.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixed a bug where you could accidentally go back by swiping with two fingers.

### DIFF
--- a/packages/tldraw/api/api.json
+++ b/packages/tldraw/api/api.json
@@ -14583,7 +14583,7 @@
               "text": "export interface TLUiContextMenuProps "
             }
           ],
-          "fileUrlPath": "packages/tldraw/src/lib/ui/components/ContextMenu.tsx",
+          "fileUrlPath": "packages/tldraw/.tsbuild-api/lib/ui/components/ContextMenu.d.ts",
           "releaseTag": "Public",
           "name": "TLUiContextMenuProps",
           "preserveMemberOrder": false,
@@ -14606,6 +14606,7 @@
                   "text": ";"
                 }
               ],
+              "fileUrlPath": "packages/tldraw/src/lib/ui/components/ContextMenu.tsx",
               "isReadonly": false,
               "isOptional": false,
               "releaseTag": "Public",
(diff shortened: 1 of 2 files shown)


packages/tldraw/api/api.json
```diff
@@ -14583,7 +14583,7 @@
               "text": "export interface TLUiContextMenuProps "
             }
           ],
-          "fileUrlPath": "packages/tldraw/src/lib/ui/components/ContextMenu.tsx",
+          "fileUrlPath": "packages/tldraw/.tsbuild-api/lib/ui/components/ContextMenu.d.ts",
```
TodePond, line 14586: It wanted to do some api changes?

ds300: it does the same on my machine, no idea why but i don't think we use these paths anywhere so it should be ok to let api-extractor be weird here.

ds300: err looks like I was wrong and we use these paths in the docs so we should find a way to fix these .tsbuild-api paths. But that's obvs outside the scope of this PR so it's still fine to leave these as is.

ds300:
I'm gonna leave it to @steveruizok to review this one. This doesn't seem to happen most of the time so I'm worried the root cause might be a different issue. maybe an html meta tag or the focus thing. I don't know.

ds300:
replaced by https://github.com/tldraw/brivate/pull/3186

image

Issue #2147, opened 2023-11-06
[Bug]: PDF Export Frame Broke


calebeby:
### What happened?

This was working as of a few weeks ago but now it doesn't work correctly.

Here is my document: https://www.tldraw.com/r/hkRFOW-UWjju0xg8r3L4j?viewport=594,2922,3331,1893&page=page:OS7K6JByaiADITjPQu0bO

Previously, I could have nothing selected and do ctrl+p to do a PDF export. It would fit everything into the PDF viewport and it worked great! Now it exports like this:

![image](https://github.com/tldraw/tldraw/assets/13206945/f9386632-9f83-48d5-bf52-82eeda9333eb)

It looks like the TLDraw UI is rendered on top of the part that is intended to be exported (the "this is an example" text.

Additionally, previously I was able to select a single frame to export just the content in that frame. This workflow used to work really well for me to set up documents/worksheets for my teaching. Now, it has a similar ui-rendering-on-top issue (if I select frame 1 and then do ctrl+p):

![image](https://github.com/tldraw/tldraw/assets/13206945/95c833b1-f5c0-4fa9-a0f4-fa86811af6de)

I'd appreciate however you are able to fix this! I assume it was some sort of small CSS change that affected the print output that wasn't tested.

Thanks for this great product I use it all the time!

- Caleb :)

### How can we reproduce the bug?

https://www.tldraw.com/r/hkRFOW-UWjju0xg8r3L4j?viewport=594,2922,3331,1893&page=page:OS7K6JByaiADITjPQu0bO

Select "frame 1"

Ctrl+p or cmd+p

### What browsers are you seeing the problem on?

Firefox, Chrome, Microsoft Edge

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Jatin-Ahirwar:
i am getting this same issue this is my code 
import React from 'react';
import { FaRegFilePdf } from 'react-icons/fa';

const PrintPdf = () => {
    const handleClick = () => {
        window.print(); 
    };

    return (
        <button
            onClick={handleClick}
            className="absolute top-0 left-[22%] z-[999] bg-[#EDF0F2] text-black font-bold py-3 px-4 rounded-br hover:bg-[#E3E6E8] transition duration-200"
        >
            <FaRegFilePdf />
        </button>
    );
};

export default PrintPdf;

import React from 'react';
import { Tldraw } from 'tldraw';
[...]

PR #2177, opened 2023-11-08
Fix printing.

MitjaBezensek:
Fixes printing. The issue is that SPA added an additional `root` div to the body and all our elements now live inside of that. The following css stopped working because of that. 	
```css
body > * {
   display: none;
}
```

For now I just added a rule to hide the `tldraw__editor`, which should be a relatively safe thing to do.

We should think about improving this down the line:
1. We should allow the user to tell us about the root of where the tldraw lives and then target elements based on that (or allow users to customize the `usePrint` and other similar features)
2. We probably should not append stuff directly to the `body` as we currently do.

Fixes https://github.com/tldraw/tldraw/issues/2147

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Add some shapes to the doc.
2. Use print and check that we only show the shapes / frames in the print preview. There should be no UI visible.

### Release Notes

- Fixes printing of shapes.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/usePrint.ts
+++ b/packages/tldraw/src/lib/ui/hooks/usePrint.ts
@@ -59,6 +59,10 @@ export function usePrint() {
 					display: none;
 				}
 
+				.tldraw__editor {
+					display: none;
+				}
+
 				.${className} {
 					display: block !important;
 					background: white;


ds300 (approved):
seems to work, nice one

calebeby:
Thank you!

MitjaBezensek:
> Thank you!

Thanks for reporting. I expect this to ship by the end of the week.

PR #2178, opened 2023-11-08
[Don't merge] Custom ui with getting editor from on mount

MitjaBezensek:


MitjaBezensek:
Just used this for a demo.

PR #2180, opened 2023-11-08
fix scroll event coords

ds300:
Follow up to #2149 to make sure it works when tldraw is not mounted at 0,0 in document space. Try it out in the 'multiple' examples

### Change Type

- [x] `patch` — Bug fix


MitjaBezensek (approved):
Nice catch!

Issue #2085, opened 2023-10-17
Edit link dialog prevents mouse events from working correctly


MitjaBezensek:
Could not select shapes, changing tools did change the tool on the toolbar, but I could not use the tools. 

Steps to reproduce:
1. Insert a shape
2. Right click the shape and Edit link
3. Dismiss the dialog
4. Range selection stops working. You can still single click select shapes.

Seems like other tools still work if you use them on top of other shapes, so it sees like it's just the canvas is not receiving pointer events?

https://github.com/tldraw/tldraw/assets/2523721/34ee05ab-4ae6-4d95-88df-ccac0f3df62f

steveruizok:
Should be closed by https://github.com/tldraw/tldraw/pull/2086

MitjaBezensek:
> Should be closed by #2086

Still seems to be an issue.

PR #2184, opened 2023-11-09
Fix an issue with edit link.

MitjaBezensek:
After the dialog is closed the `body` still has `pointer-events: none` applied to it. Looks like a radix issue in an older version: https://github.com/radix-ui/primitives/issues/1241

To test this in our private repo you also need to run `yarn dedupe` to make this work. Otherwise some old dependencies are kept and make this not work.

Fixes https://github.com/tldraw/tldraw/issues/2085

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Add a geo shape, right click it, Edit link, add a link or cancel the dialog.
2. You should be able to deselect the shape after the dialog is closed.

### Release Notes

- Fixes an issue with using the Edit link dialog.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/ActionsMenu.tsx
+++ b/packages/tldraw/src/lib/ui/components/ActionsMenu.tsx
@@ -57,7 +57,7 @@ export const ActionsMenu = memo(function ActionsMenu() {
 					smallIcon
 				/>
 			</PopoverTrigger>
-			<PopoverPrimitive.Portal dir="ltr" container={container}>
+			<PopoverPrimitive.Portal container={container}>
 				<PopoverPrimitive.Content
 					className="tlui-popover__content"
 					side="bottom"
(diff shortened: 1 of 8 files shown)


packages/tldraw/src/lib/ui/components/ActionsMenu.tsx
```diff
@@ -57,7 +57,7 @@ export const ActionsMenu = memo(function ActionsMenu() {
 					smallIcon
 				/>
 			</PopoverTrigger>
-			<PopoverPrimitive.Portal dir="ltr" container={container}>
```
MitjaBezensek, line 60: Looks like radix removed `div` props from portal https://github.com/radix-ui/primitives/issues/2410#issuecomment-1736936484

TodePond (approved):
That fixed it! Tested locally, thanks for the heads up Mitja :)

packages/tldraw/src/lib/ui/components/ActionsMenu.tsx
steveruizok: good catch!

PR #2189, opened 2023-11-09
No impure getters pt 1

steveruizok:
This PR replaces the `.value` getter for the atom with `.get()`

### Change Type


steveruizok:
This one turned out to be pretty straightforward @ds300.

- used the IDE to rename the `value` symbol to `ATOM_VALUE`
- find and replace `ATOM_VALUE` with `get()`
- fix one issue where the symbol rename was missed (in `useValue`)
- few little changes where we were passing `.value` as a dependency
- bobs your uncle

ds300:
We need to keep .value around for a while and mark it ***@***.***`
otherwise we'll break everyone who is using tldraw in nontrivial ways.

The tricky part of supporting clickup's requirements however is refactoring
all other getters in the app, ones that use @computed and ones that don't,
to make sure they are pure. We should also deprecate @computed

ds300:
@steveruizok i'm gonna hijack this PR

PR #2211, opened 2023-11-14
Allow users to extend the base Editor class

anglinb:
This allows you to Subclass Editor. You'd want to do this to centralize logic for deeper integrations and extensions of the the base editing capabilities. 

I've also added a trivial example showing how to use the new API

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

1. You can open the `/custom-editor` example
2. It should alert "hello!" after 2 seconds

- [x] Unit Tests
- [x] End to end tests

### Release Notes

- Adds `createEditor` to the `TldrawEditorProps` to allow users to instantiate a custom Editor subclass


anglinb:

steveruizok:
Hey Brian @anglinb, thanks for the PR! I don't think we can merge this yet, though there's a separate design that I think is compatible here. What I'd like to have (and what I think would be compatible with your intention here) would be to have the editor be declared outside of the component and then passed into it or provided to it.

```tsx
function Example() {
  const editor = useCreateEditor({...myConfiguration})
  return <Tldraw editor={editor}/>
}
```

or

```tsx
function Example() {
  const editor = useCreateEditor({...myConfiguration})
  return (
    <EditorProvider editor={editor}>
      <Tldraw/>
      <OtherThingsThatRelyOnTheEditor/>
    </EditorProvider
}
```

or

```tsx
const editor = new Editor({...myConfiguration})

function Example() {
  return (
    <EditorProvider editor={editor}>
      <Tldraw/>
      <OtherThingsThatRelyOnTheEditor/>
    </EditorProvider
}
```

Curious though, what would be your use-case / reason for extending the editor?

kiastorm:
@steveruizok I just submitted a similar request. https://github.com/tldraw/tldraw/issues/2278

I have a "Fretboard" shape which is essentially a composition of other shape utils.

The fretboard is similar to a group, where the items inside of it (strings, fret areas etc) are not hit/hoverable/selectable unless the fretboard is selected.

I think the logic that allows this behaviour for the group shape is:

```
getOutermostSelectableShape(
shape: TLShape | TLShapeId,
filter?: (shape: TLShape) => boolean
): TLShape {
const id = typeof shape === 'string' ? shape : shape.id
const freshShape = this.getShape(id)!
let match = freshShape
let node = freshShape as TLShape | undefined

const focusedGroup = this.getFocusedGroup()

while (node) {
	if (
		this.isShapeOfType<TLGroupShape>(node, 'group') &&
		focusedGroup?.id !== node.id &&
		!this.hasAncestor(focusedGroup, node.id) &&
		(filter?.(node) ?? true)
	) {
		match = node
	} else if (focusedGroup?.id === node.id) {
		break
	}
	node = this.getShapeParent(node)
}

return match
}
```

I'd like to override this method (and others) so that the fretboard behaves similarly to the group.

image

image

image

- Fretboard has both a Util and a Tool
- Creating a Fretboard also automatically creates Frets and Strings, but there is not explicit tool.
- The Frets and Strings can be hovered/selected only when the parent Fretboard is already selected. Similarly to how shapes within a group aren't hit unless their parent group is selected.

steveruizok:
Hey, we'll give this some more attention in the new year. It's always surprising where customization is needed in the editor, thanks sharing @kiastorm !

steveruizok:
Closing this one, we'll come back to this one. If you'd like to make a second issue requesting that customization, it would be helpful. Cheers!

PR #2214, opened 2023-11-14
Stop bookmark shape jumping position when content is fetched

TodePond:
This PR stops the bookmark shape changing position when content is fetched. (When you paste it and move it before it loads).

### Change Type

- [x] `patch` — Bug fix
### Test Plan

1. Paste a link.
2. Before the content fetches, move the shape.
3. Make sure it doesn't move back to the original paste position.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixed a bug where pasted bookmark shapes could change position when their content loads.


ds300:
sorry i didn't realise you were making this when I did the same fix @TodePond

PR #2219, opened 2023-11-14
Removing frames and adding elements to frames

MitjaBezensek:
- Add simple frame removing - it just drops the frame and parent children to frames parent.
- Select children after removing the frame.
- Add children to the frame if we resize the frame so that it encloses them.

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


steveruizok:
We'll want to make this only snap up children that are completely contained by the frame rather than partially contained.

steveruizok:
Ok, I've made a few changes here:

- rather than absorbing children on resize end (which would make it occur when after a normal resize), we absorb "on create", a new callback that I've written into resizing. There's a bunch more that could be done here, replacing `onInteractionEnd` in certain places.
- we now hide the frame while it's being created. This seems like it could be a general thing to have in the editor (i.e. a "creating shape" similar to "editing shape") because there are a few tools that casually switch based on whether a shape is the creating shape or not

And a rider that seemed to be appropriate
- I've added functionality to preserve the location of the frame's children while holding command while resizing

steveruizok:
Still needs tests!

packages/tldraw/src/lib/shapes/frame/FrameShapeTool.test.ts
```diff
@@ -15,7 +15,7 @@ describe(FrameShapeTool, () => {
 		expect(editor.getCurrentPageShapes().length).toBe(0)
 
 		editor.setCurrentTool('frame')
-		editor.pointerDown(50, 50)
```
MitjaBezensek, line 18: Not sure if this change is intended or not?

packages/tldraw/src/test/frames.test.ts
```diff
@@ -110,6 +116,22 @@ describe('creating frames', () => {
 		})
 	})
 
+	it('parents a shape when drag-creating a frame over it', () => {
+		const rectId: TLShapeId = createRect()
```
MitjaBezensek, line 120: In general I'm all for pulling out logic, but on this case it makes it a bit harder to understand the test? I need to know the values of the default params to understand the code. Maybe we can always require the arguments so that you don't have to jump to `createRect` (and `dragCreateFrame`) to understand the test.

packages/tldraw/src/test/frames.test.ts
```diff
@@ -234,6 +256,58 @@ describe('frame shapes', () => {
...
+		editor.pointerDown(100, 50, { target: 'selection', handle: 'right' })
+		editor.pointerMove(50, 50)
+		editor.pointerUp(50, 50)
+		const parent = editor.getShape(rectId)?.parentId
+		expect(parent).toBe('page:page')
+	})
+
+	it('doesnt unparent a shape that is only partially out of bounds', () => {
+		const rectId: TLShapeId = createRect({ pos: [70, 10] })
+		const frameId = dragCreateFrame()
```
MitjaBezensek, line 272: I'd probably add an expect below this line which makes sure that the rect is parented to the frame. Makes it a bit more explicit what change we expect to occur.

packages/tldraw/src/test/frames.test.ts
```diff
@@ -234,6 +256,58 @@ describe('frame shapes', () => {
...
+		dragCreateFrame({ move: [60, 100], up: [60, 100] })
+		// resize the frame so the shape is totally covered
+		editor.pointerDown(60, 50, { target: 'selection', handle: 'right' })
+		editor.pointerMove(100, 50)
+		editor.pointerUp(100, 50)
+		const parent = editor.getShape(rectId)?.parentId
+		expect(parent).toBe('page:page')
+	})
+
+	it('Moves children when resizing a parent frame', () => {
```
MitjaBezensek, line 293: I'd use the same casing for all the tests, probably lowercase (looks like we aren't consistent here).

packages/tldraw/src/test/frames.test.ts
```diff
@@ -234,6 +256,58 @@ describe('frame shapes', () => {
...
+		const rectId: TLShapeId = createRect()
+		dragCreateFrame()
+		editor.pointerDown(0, 0, { target: 'selection', handle: 'top_left' })
```
MitjaBezensek, line 296: Would probably also do an expect on the `y` position. I generally like to make sure that things (that I will expect later on) are in the correct state before the action.

MitjaBezensek (commented):
Looking good, left a few comments.

I might also add a `does not parent a shape when drag-creating a frame over part of the shape` test.

PR #2198, opened 2023-11-12
Custom Tools DX + screenshot example

steveruizok:
This PR adds a custom tool example, the `Screenshot Tool`.

It demonstrates how a user can create a custom tool together with custom tool UI. 

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Use the screenshot example

### Release Notes

- adds ScreenshotTool custom tool example
- improvements and new exports related to copying and exporting images / files
- loosens up types around icons and translations
- moving `StateNode.isActive` into an atom
- adding `Editor.path`

PR #2228, opened 2023-11-15
[DX] Separately export export as, copy as

steveruizok:
This PR extracts part of #2198 that relate to exporting images. It shares the `exportAs` and `copyAs` as separate exports.

### Change Type

- [x] `minor`


steveruizok:
Eh, turns out this wasn't as clean to extract as I thought. Closing in favor of #2198

PR #2258, opened 2023-11-27
improvement(draw): improve ink elbow rendering and reduce SVG size

ds300:
Our ink has kinda homely elbows sometimes. This PR believes elbows can and should be beautiful.

The way this is achieved is by partitioning the points fed into perfect-freehand at elbow points, and then rendering each partition separately.

Doing this naively ballooned the size of the SVG path data so I also did a tiny bit of refactoring to allow us to use the SVG arc command. At the same time we are able to easily omit some of the points around the beginnings and ends of segments thanks to the nature of the corners.

All of this results in an average 13% reduction in SVG path data size over the current version.

Kapture 2023-11-27 at 14 39 57

### Change type

- [x] `improvement`

### Test plan

1. Draw shapes with sharp corners/elbows and verify rendering quality.
2. Inspect SVG output to verify path data reduction.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improved the visual quality of corners (elbows) on ink shapes and reduced SVG path data size by ~13%.

### DIFF
--- a/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
@@ -109,44 +109,51 @@ export class DrawShapeUtil extends ShapeUtil<TLDrawShape> {
 		}
 
 		const options = getFreehandOptions(shape.props, sw, showAsComplete, forceSolid)
-		const strokePoints = getStrokePoints(allPointsFromSegments, options)
 
-		const solidStrokePath =
-			strokePoints.length > 1
-				? getSvgPathFromStrokePoints(strokePoints, shape.props.isClosed)
-				: getDot(allPointsFromSegments[0], sw)
-
-		if ((!forceSolid && shape.props.dash === 'draw') || strokePoints.length < 2) {
-			const drawStyleFill = svgInk(allPointsFromSegments, options)
-
+		if (!forceSolid && shape.props.dash === 'draw') {
 			return (
 				<SVGContainer id={shape.id}>
-					<ShapeFill
-						theme={theme}
-						fill={shape.props.isClosed ? shape.props.fill : 'none'}
-						color={shape.props.color}
-						d={solidStrokePath}
+					{shape.props.isClosed && shape.props.fill && allPointsFromSegments.length > 1 ? (
+						<ShapeFill
+							theme={theme}
+							fill={shape.props.isClosed ? shape.props.fill : 'none'}
+							color={shape.props.color}
+							d={getSvgPathFromStrokePoints(
+								getStrokePoints(allPointsFromSegments, options),
+								shape.props.isClosed
+							)}
+						/>
+					) : null}
+					<path
+						d={svgInk(allPointsFromSegments, options)}
+						strokeLinecap="round"
+						fill={theme[shape.props.color].solid}
 					/>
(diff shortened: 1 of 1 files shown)


steveruizok:
Hey, reviewing now. In general this is an improvement: the only place where I've noticed a regression in quality is that sharp corners tend to be reduced in size. 

image

IIRC in our current freehand we actually push these corners out by a certain distance in order to capture detail. I'm not sure what the motivation for that was—it may have been a "feels good" feature, I vaguely recall adding that after watching videos of me drawing on the ipad.

steveruizok:
Seems to handle short scribbles very well.

ds300:
ok i think i fixed those things, take another look when you have a moment.

packages/tldraw/src/lib/shapes/shared/freehand/getStrokeOutlinePoints.ts
```diff
@@ -7,26 +7,17 @@ const { PI } = Math
...
-export function getStrokeOutlinePoints(
+export function getStrokeOutlineTracks(
 	strokePoints: StrokePoint[],
 	options: StrokeOptions = {}
-): Vec2d[] {
-	const { size = 16, smoothing = 0.5, start = {}, end = {}, last: isComplete = false } = options
-
-	const { cap: capStart = true } = start
-	const { cap: capEnd = true } = end
+): { left: Vec2d[]; right: Vec2d[] } {
```
ds300, line 15: To allow me to draw the caps with SVG arcs, I extracted out the part of this that gets the left and right tracks so I could use it on its own.

ds300:
i think if we can get rid of these guys in a follow up it will actually be pretty close to perfect
image

packages/tldraw/src/lib/shapes/shared/freehand/getStrokeOutlinePoints.ts
```diff
@@ -185,6 +162,65 @@ export function getStrokeOutlinePoints(
...
+			: start.taper === true
+			? Math.max(size, totalLength)
+			: (start.taper as number)
+
+	const taperEnd =
+		end.taper === false
+			? 0
+			: end.taper === true
+			? Math.max(size, totalLength)
+			: (end.taper as number)
```
KryptonBD, line 218:
```suggestion
	const taperStart = start.taper === true ? Math.max(size, totalLength) : (start.taper as number) || 0
	const taperEnd = end.taper === true ? Math.max(size, totalLength) : (end.taper as number) || 0
```

steveruizok:
Kapture 2023-11-29 at 10 54 14

steveruizok:
Comparing the results, a few things I've noticed:

Knees are pointier due to the algorithm identifying a "sharp curve" as a "sharp corner". This is really difficult to distinguish; I remember trying to track the rate of directional change using acceleration in the past but likely just picked a "good enough" dot product change. 

For example, the W and E here: 

image

The settings you've used here is often an improvement, however, so I don't think it's a problem. For example, see the much improved corners around the W, A, and the twist to the D.

image

In general this is a big improvement in the detail we're able to pull from the same inputs. Well done!

steveruizok:
I'm pushing a few small changes to the rendering for draw shapes.

- now that we're not passing in stroke points, we don't need to calculate stroke points unless a draw-style shape has a fill
- dots are now correctly scaled

Issue #2254, opened 2023-11-25
[Bug]: Cropped images are positioned incorrectly when exported


myandrienko:
### What happened?

When page contains a cropped image, it's positioned incorrectly when exported.

Here's an example page containing two images, one of them cropped:

![page](https://github.com/tldraw/tldraw/assets/975978/dddd4bda-41b9-441a-b8d4-9d5b52114ed6)

This is an expected export result:

![expected](https://github.com/tldraw/tldraw/assets/975978/890e561b-3162-4783-bf10-cdee55a3a9e9)

This is an actual export result:

<img width="1064" alt="actual" src="https://github.com/tldraw/tldraw/assets/975978/df1fabb5-d33c-423d-acc7-15af618710af">

### How can we reproduce the bug?

Reproduction project: https://www.tldraw.com/r/khidz2Q97fC4Q0uN8waQY?viewport=-236%2C-162%2C1652%2C1111&page=page%3Apage

1. Add an image
2. Crop it
3. Export it (as PNG in my case)

### What browsers are you seeing the problem on?

Chrome, Microsoft Edge

### Contact Details

m.y.andrienko@outlook.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MitjaBezensek:
Thanks for reporting this, opened a PR for this, will probably land with the next update.

PR #2268, opened 2023-11-30
Fix exporting of cropped images.

MitjaBezensek:
Open this PR in different browsers, and you will see that the svg below will render differently in different browsers. The svg was exported from our staging.

Seems like Chrome handles `clip-path` when set via the style differently than Safari and Firefox. I've reworked the logic so that it now uses a `clip-path` definition and applies that to the image.

shape_xPSLLIG9yQkqAACrv1OxE

Also fixes a bunch of issues when copy pasting via the menu. It seems like we can't store the write function to a variable:
image

Fixes https://github.com/tldraw/tldraw/issues/2254

### Before
CleanShot 2023-11-30 at 09 30 24

### After
CleanShot 2023-11-30 at 09 30 55

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Add an image, apply crop to it (best if you crop from all sides, just to make sure)
2. Copy as png and make sure the image is correctly cropped in the created image.

### Release Notes

- Fix exporting of cropped images.


MitjaBezensek:
Oh, nice catch. I'll take a look. Looks like it might be an issue with how safari handles `clipPath`.

Issue #2284, opened 2023-12-05
Cropping preview not working until changed


MitjaBezensek:
Looks like this only happens locally and it might be due to strict mode?

### What happened?

For me it happens every time when:
1. I have a cropped image.
2. I refresh the page.

As soon as I interact with the image it starts to show the cropped area correctly. Happens in all browsers.

### How can we reproduce the bug?

https://github.com/tldraw/tldraw/assets/2523721/1c39fddc-71d2-466c-9157-4d7d1fb6c7b4

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2295, opened 2023-12-06
Fix an issue with a stale editor reference in shape utils

MitjaBezensek:
Fixes an issue where the editor reference in shape utils was not up to date with the editor returned from `useEditor`. Actually, the whole util was the incorrect one and was holding a reference to the previous instantiation of the editor.

This only occurred in dev mode, but could also happen in other cases where editor is created multiple times. To see the kinds of issues this causes in dev mode you can do the following:
1. Create an image, crop it.
2. Refresh the page.
3. Select the image, then double click it to enter crop mode.
4. You will not see the cropped area of the image. You need to change the crop slightly and then it suddenly appears. This is because this changes props, which reruns the memoized function.

Fixes https://github.com/tldraw/tldraw/issues/2284

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Fix an issue where the shape utils could have a stale reference to the editor.


ds300 (approved):
Well done on finding this one!

PR #2299, opened 2023-12-06
Start scrolling if we are dragging close to the window edges.

MitjaBezensek:
Start scrolling when we get close to the edges of the window. This works for brush selecting, translating, and resizing.

https://github.com/tldraw/tldraw/assets/2523721/4a5effc8-5445-411b-b317-36097233d36c

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Select a shape.
2. Move it towards the edge of the window. The camera position should change.
3. Also try resizing, brush selecting.

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds the logic to change the camera position when you get close to the edges of the window. This allows you to drag, resize, brush select past the edges of the current viewport.


trinhtuanvy:
Love you team support

trinhtuanvy:
Love you team support

steveruizok:
Hey @MitjaBezensek, I spent some time with this one and had a few design changes:

1. Looking at the enter/exit events that used `on('tick', handler)` made me realize we could add an `onTick` event to the state node itself which we set and unset when a node becomes active or inactive. I've added this for the places where we would call the pointer moves. I think we could refactor several other state nodes to use this as well.
2. I moved the `moveCameraWhenCloseToEdge` into an external helper function that we export.
3. I squished your nice `moveCameraWhenCloseToEdge` helper.
4. I moved the `edgeScrollingSpeed` into user preferences and added a UI preference for this.
5. I tweaked some of the distances / speeds.
6. I updated the tests to account for the adjusted speeds.

steveruizok (approved):

Issue #2305, opened 2023-12-09
[Bug]: Arrows without target are not rendered in the PNG export


angaz:
### What happened?

The arrow should be rendered in the PNG export.

### How can we reproduce the bug?

1. Draw a rectangle
2. Draw a frame around it
3. Draw an arrow coming out of the box, but without a target
4. Export the frame as PNG

The arrow is missing

### What browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

angaz:
It looks like the arrow is not part of the frame, because if I duplicate the frame, the arrow is not part of it.

Steffo99:
I think I stumbled on this bug too with the attached frame.

[Example JSON Export](https://github.com/tldraw/tldraw/files/13626110/Grafo.iniziale.1.json)

Exporting as PNG is fine on Chromium:

<img width="1082" alt="Lines are rendered perfectly" src="https://github.com/tldraw/tldraw/assets/1540885/5a2e50ac-916d-43e8-8878-97280e110d4e">

But on Firefox 120 it exports as follows:

<img width="1082" alt="Lines are rendered only partially" src="https://github.com/tldraw/tldraw/assets/1540885/13c93e2c-a528-49b4-8623-dcda66740e3c">

Seems related to how Firefox renders SVGs; when exporting to SVG from any browser, Chrome renders them as in the former image, while Firefox renders them as in the latter.

MitjaBezensek:
@angaz would you expect the arrow to point to nothing in this case? Or also the to the element it is attached to? 

The logic now is to only show the elements that are inside the frame (without showing the frame itself). One thing you an do is to select both the frame and the arrow and then do the export. But that will show the frame as well, not sure if you'd like that.

@Steffo99 Could you send me the `.tldr` file of this? Just go to `File -> Save a copy` to create it.

angaz:
@MitjaBezensek pointing at nothing. I want to indicate a request leaving the network, just have the arrow point towards the edge of the frame to indicate that.

I tried selecting both the frame and the arrow and it didn't make a difference, at least, on Firefox. Should I try on Chrome?

angaz:
Ok, this is weird, it's working now, even in firefox, so I don't know what I was doing wrong before because I'm 100% sure I tried this, and it didn't work. Makes no sense.

[...]

PR #2319, opened 2023-12-13
Update arrow parenting

Ch1sKey:
Issue #2305
Shapes are assigned to frame as children on creation or once they are placed inside one.
Arrow are kinda special. They act the same but also "reparents" each time we change them.

### Root Cause
Arrows are not assigned to frame as children properly 

### Solution
On arrow reparenting:
 * Check if both arrow's bindings are the same object. Handles the case when we have "unattached" lonely arrow inside frame
 * Also update `findCommonAncestor` function to start searching not from the shape's parent but from the shape itself. Handles the case where arrow is bound to something inside frame (also if I did my CS homework well, it is usually considered that tree's node is in fact ancestor to itself so it seems consistent)

### Results
| Before        | After          |
| ------------- |:-------------:|
| before-gif | after-gif |

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Check if the issue is fixes
2. Should check all the tricky cases for arrows. I'm new in this repository. Hope mature maintainers could help to find some test cases 

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes
- Fix bug when arrows are not visible inside frame after PNG export


Ch1sKey:

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -290,6 +290,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 			if (startShape && endShape) {
 				// if arrow has two bindings, always parent arrow to closest common ancestor of the bindings
 				nextParentId = this.findCommonAncestor([startShape, endShape]) ?? parentPageId
+
+				// if arrow has two bindings to the same shape, parrent arrow to it
+				if (startShape.id === endShape.id) {
+					nextParentId = startShape.id
```
steveruizok, line 296: This would parent the arrow to any shape, not just frames. I agree with the basic concept of this PR (and am surprised that I haven't noticed it before!) but we'd have to achieve it in a slightly different way.

PR #2322, opened 2023-12-13
Use a global singleton for tlstate

ds300:
One minor issue with signia is that it uses global state for bookkeeping, so it is potentially disastrous if there is more than one version of it included in a bundle.

To prevent that being an issue before we had a warning that would trigger if signia detects multiple initializations.

> Multiple versions of @tldraw/state detected. This will cause unexpected behavior. Please add "resolutions" (yarn/pnpm) or "overrides" (npm) in your package.json to ensure only one version of @tldraw/state is loaded.

Alas I think this warning triggers too often in development environments, e.g. during HMR or janky bundlers.

Something that can prevent the need for this particular warning is having a global singleton version of signia that we only instantiate once, and then re-use that one on subsequent module initializations. We didn't do this before because it has a few downsides:

- breaks HMR if you are working on signia itself, since updated modules won't be used and you'll need to do a full refresh.
- introduces the possibility of breakage if we remove or even add APIs to signia. We can't rely on having the latest version of signia be the first to instantiate, and we can't allow later instantiations to take precedence since atoms n stuff may have already been created with the prior version. To mitigate this I've introduced a `apiVersion` const that we can increment when we make any kind of additions or removals. If there is a mismatch between the `apiVersion` in the global singleton vs the currently-initializing module, then it throws.

Ultimately i think the pros outweigh the cons here, i.e. far fewer people will see and have to deal with the error message shown above, and fewer people should encounter a situation where the editor appears to load but nothing changes when you interact with it.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Make a global singleton for tlstate.

### DIFF
--- a/packages/state/api-report.md
+++ b/packages/state/api-report.md
@@ -14,10 +14,26 @@ export interface Atom<Value, Diff = unknown> extends Signal<Value, Diff> {
 }
 
 // @public
-export function atom<Value, Diff = unknown>(
-name: string,
-initialValue: Value,
-options?: AtomOptions<Value, Diff>): Atom<Value, Diff>;
+export const
+/**
+* Creates a new [[Atom]].
+*
+* An Atom is a signal that can be updated directly by calling [[Atom.set]] or [[Atom.update]].
+*
+* @example
+* ```ts
+* const name = atom('name', 'John')
+*
+* name.get() // 'John'
+*
+* name.set('Jane')
+*
+* name.get() // 'Jane'
+* ```
+*
+* @public
+*/
+atom: typeof atom_2;
 
 // @public
 export interface AtomOptions<Value, Diff> {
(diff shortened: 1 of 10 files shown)


ds300:
gonna merge this since it's blocking a user, would still appreciate a sanity check review @SomeHats if you have time.

SomeHats (commented):
This looks good, but seems like a regression overall for docs etc? some of the changes there look unexpected. i wonder if instead of 

```ts
const impl = globalImplementationSingleton()

export const {
    /** docs go here */
   atom,
} = impl
```

doing things more like:
```ts
const impl = globalImplementationSingleton()

/** docs here */
export function track(...fullSignatureHere) {
    return impl.track(...args)
}
```

Issue #2320, opened 2023-12-13
Enormous snapshot size for little sized assets


irg1008:
### What happened?

When uploading small sized assets (around 50kB images) and then downloading snapshot for later restore, the generated snapshot with encoded assets has a x10-15 increase in size. 

For example a bare bone example with one 45kB image results in a 1.9MB snapshot. 

When trying to persist and store this in a custom backend storage, this shortly becomes a problem when handling hundreds of store snapshots.

Is there a better way to store assets in editor store or maybe provide a way to allow for custom asset storage or asset shape url manipulation

Anyway, thank you for this amazing library!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

irg1008:
For example when adding this image to the editor:
<img src="https://github.com/tldraw/tldraw/assets/47559427/84d6fd7a-6b2f-4251-a2a6-3db2198fae4d" alt="drawing" width="200"/>


The jpg size is: `42.3KB`

When encoding using https://base64.guru/converter/encode/image the size is `56.4KB`

But when exporting the editor snapshots the base64 for said assets takes up to whopping `1.9MB`

This is probably because we are converting a canvas element to base64 instead of the original image:

```typescript
const getStateFrame = async (url: string) => {
	const image = await loadImage(url)

	const canvas = document.createElement('canvas')
	canvas.width = image.width
	canvas.height = image.height

	const ctx = canvas.getContext('2d')
	if (!ctx) return

	ctx.drawImage(image, 0, 0)
	return canvas.toDataURL()
}
```

Is this happening to allow for asset cropping? If it is, is there any optimization option here?

Thank you

irg1008:
In the next snippet of [ImageShapeUtil](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx):
```typescript
async function getDataURIFromURL(url: string): Promise<string> {
	const response = await fetch(url)
	const blob = await response.blob()
	return new Promise((resolve, reject) => {
		const reader = new FileReader()
		reader.onloadend = () => resolve(reader.result as string)
		reader.onerror = reject
		reader.readAsDataURL(blob)
	})
}
```

Could replacing readAsDataURL with createObjectURL fix the asset oversize? Does this break something else?

irg1008:
It is definitely the downscale library: Check the next example:

[...]

PR #2325, opened 2023-12-15
Fix downscaling

MitjaBezensek:
Fixes an issue where uploaded images could significantly increase the document size. The problem was with calling `downscale` library, which expect the type to be one of: `jpg`, `png`, `webp`. We were sending the type with the `image/` prefix. Thanks to @irg1008 for pointing that out.

Test image:
testImage

The difference I saw in testing:
CleanShot 2023-12-15 at 10 34 56

@steveruizok why are we[ doubling the size here](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/utils/assets/assets.ts#L61-L62)? For retina quality?

Fixes https://github.com/tldraw/tldraw/issues/2320

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Use the supplied test image.
2. Add it to a tldraw document.
3. Save the file and check the file size. It should be significantly lower than doing the same thing on tldraw.com

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Decrease the size of uploaded assets.

### DIFF
--- a/packages/tldraw/src/lib/utils/assets/assets.ts
+++ b/packages/tldraw/src/lib/utils/assets/assets.ts
@@ -82,7 +82,11 @@ export async function getResizedImageDataUrl(
 		desiredHeight *= ratio
 	}
 
-	return await downscale(dataURLForImage, desiredWidth, desiredHeight, { imageType: type, quality })
+	return await downscale(dataURLForImage, desiredWidth, desiredHeight, {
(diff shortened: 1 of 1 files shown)


steveruizok (approved):

sac10nikam:
Would like to increase the size of the loaded image dynamically?

Issue #2329, opened 2023-12-16
[Bug] Matrix2d.Scale


zfedoran:
The matrix class seems to be using this representation:
<img width="129" alt="image" src="https://github.com/tldraw/tldraw/assets/623790/b2e3bf68-6144-408c-a36a-63fd942f9e3e">

When considering how Identity and Translate are defined:
```
	static Identity() {
		return new Matrix2d(1.0, 0.0, 0.0, 1.0, 0.0, 0.0)
	}

	static Translate(x: number, y: number) {
		return new Matrix2d(1.0, 0.0, 0.0, 1.0, x, y)
	}
```

## Issue

The `y` scale value seems to be in the wrong place inside the `Matrix2d.Scale` function ([reference](https://github.com/tldraw/tldraw/blame/0171d1498d432750a0fe2c2046565c2f4ed39c37/packages/editor/src/lib/primitives/Matrix2d.ts#L176)):

`const scaleMatrix2d = new Matrix2d(x, 0, 0, 0, y, 0)`

## Bug Fix
The fix would be to move the `y` over by one:
```const scaleMatrix2d = new Matrix2d(x, 0, 0, y, 0, 0)```

## Reproduce Issue
`Matrix2d.Compose(Matrix2d.Scale(1, 1), Matrix2d.Translate(12, 34))`

Gives an incorrect result of:
`{"a":1,"b":0,"c":0,"d":0,"e":124,"f":0}`

trinhtuanvy:
Thanks help

steveruizok:
Closed by #2330

PR #2330, opened 2023-12-16
[bug] Fix for issue #2329

zfedoran:
This PR fixes the issue outlined in issue #2329

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Fix for `Matrix2d.Scale` function


zfedoran:

trinhtuanvy:
Help support

MitjaBezensek:
Looks good. Maybe we would want to add a test for this? There's a `Matrix2d.test.ts` file in the same folder where we could add it.

steveruizok:
I'd like to see some tests as well. IIRC we don't use scale anywhere in the code base, which is perhaps why this hasn't been caught, but I can think of some places where we may in the future!

PR #2373, opened 2023-12-21
Avoid importing `editor.css` twice

MitjaBezensek:
This PR prevents importing `editor.css` twice. Looks like we currently import all the css files for all the routes in the examples app. This means that we were importing both `tldraw.css` (which includes `editor.css`) as well as `editor.css` directly (this last one we imported from `OnlyEditor.tsx`).

This caused all the classes to be defined twice and it made it harder to debug things - you might have disabled the styles in one of the definitions but you didn't see any changes because the other definitions still applied.

image

I don't think importing `editor.css` is needed as we do import the full `tldraw.css` anyway.

image

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Prevent importing `editor.css` twice which should help when debugging the styles via developer console.

### DIFF
--- a/apps/examples/src/examples/only-editor/OnlyEditor.tsx
+++ b/apps/examples/src/examples/only-editor/OnlyEditor.tsx
@@ -1,7 +1,6 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
 import { Editor, PositionedOnCanvas, TldrawEditor, createShapeId, track } from '@tldraw/editor'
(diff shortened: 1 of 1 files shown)


steveruizok (approved):

steveruizok:
LGTM.

I think we also import tailwind styles on root?

PR #2375, opened 2023-12-22
add descriptions to examples

SomeHats:
Adds descriptions to examples.

Kapture 2023-12-22 at 17 08 32

They show as a list on the index page, and on individual examples they show in a three-js style sidebar. For now, this is disabled completely on mobile. Examples can still be opened in 'standalone' mode to get rid of the sidebar.

Note: the 'view code' link won't work until after these changes are merged.

There's a small impact on authoring examples: each one needs to live in a folder with a README.md. At a minimum, the readme needs to look like this:
```md
---
title: My Example
component: ./MyExample.tsx
---

Here is a 1-liner about my example
```

Optionally, you can:
- Add `hide: true` to the frontmatter to remove the example from the list (you can skip the description this way)
- Add `order: 3` to control the order in which the example appears. They're alphabetical otherwise
- Add some more description or links to docs below a `---`. This won't show in the listing, but will be visible on GitHub and on the example page itself.

As a follow-up, I'd like to add an 'Open in CodeSandbox' link to each example. These won't work until we've made a release with these examples (as our special examples codesandbox is tied to our release process) but the code is there & ready to go!

Have a play, let me know what you think!

### Change Type


steveruizok:
I did a few tweaks on the CSS for this:
- the "ordered" items are separated from the rest of the items by a <hr>
- the list items are only one column for easier scanning
- the sidebar items don't show descriptions

Overall a huge improvement, very happy to see it go in!

We sort of punted on mobile, which is probably fine; but I think a little tab / nibblydoo UI would really work here:

image

PR #2394, opened 2024-01-02
Refactor speech bubble

steveruizok:
This PR is a refactor of the speech bubble geometry code. It does not change the output but removes some parts that I found confusing while working through the logic.

### Change Type

- [x] `patch` — Bug fix

### DIFF
--- a/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
+++ b/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
@@ -18,13 +18,15 @@ import {
 	resizeBox,
 	structuredClone,
 } from '@tldraw/tldraw'
-import { STROKE_SIZES } from '@tldraw/tldraw/src/lib/shapes/shared/default-shape-constants'
-import {
-	getHandleIntersectionPoint,
-	getHandleinShapeSpace,
-	getHandlesInHandleSpace,
-	getSpeechBubbleGeometry,
-} from './helpers'
+import { getHandleIntersectionPoint, getSpeechBubbleVertices } from './helpers'
+
+// Copied from tldraw/tldraw
+export const STROKE_SIZES = {
+	s: 2,
+	m: 3.5,
+	l: 5,
+	xl: 10,
+}
 
 // There's a guide at the bottom of this file!
 
@@ -87,8 +89,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 	}
 
 	getGeometry(shape: SpeechBubbleShape): Geometry2d {
-		const newShape = getHandleinShapeSpace(shape)
-		const speechBubbleGeometry = getSpeechBubbleGeometry(newShape)
+		const speechBubbleGeometry = getSpeechBubbleVertices(shape)
 		const body = new Polygon2d({
 			points: speechBubbleGeometry,
 			isFilled: true,
@@ -97,10 +98,21 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 	}
 
 	override getHandles(shape: SpeechBubbleShape) {
-		const handles = getHandleinShapeSpace(shape).props.handles
-		const handlesArray = Object.values(handles)
+		const {
+			handles: { handle },
+			w,
+			h,
+		} = shape.props
 
-		return handlesArray
+		return [
+			{
+				...handle,
+				// props.handles.handle coordinates are normalized
+				// but here we need them in shape space
+				x: handle.x * w,
+				y: handle.y * h,
+			},
+		]
 	}
 
 	// [4]
@@ -108,49 +120,46 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 		_: SpeechBubbleShape,
 		next: SpeechBubbleShape
 	) => {
-		const shape = getHandleinShapeSpace(next)
+		const { w, h, handles } = next.props
 
-		const { originalIntersection: intersection, insideShape } = getHandleIntersectionPoint({
-			w: shape.props.w,
-			h: shape.props.h,
-			handle: shape.props.handles.handle,
-		})
+		const { segmentsIntersection, insideShape } = getHandleIntersectionPoint(next)
 
-		if (!intersection) throw new Error('No intersection')
+		const slantedLength = Math.hypot(w, h)
+		const MIN_DISTANCE = slantedLength / 5
+		const MAX_DISTANCE = slantedLength / 1.5
 
-		const intersectionVector = new Vec2d(intersection.x, intersection.y)
-		const handleVector = new Vec2d(shape.props.handles.handle.x, shape.props.handles.handle.y)
+		const handleInShapeSpace = new Vec2d(handles.handle.x * w, handles.handle.y * h)
 
-		const topLeft = new Vec2d(0, 0)
-		const bottomRight = new Vec2d(shape.props.w, shape.props.h)
-		const center = new Vec2d(shape.props.w / 2, shape.props.h / 2)
-		const MIN_DISTANCE = topLeft.dist(bottomRight) / 5
+		const distanceToIntersection = handleInShapeSpace.dist(segmentsIntersection)
+		const center = new Vec2d(w / 2, h / 2)
+		const vHandle = Vec2d.Sub(handleInShapeSpace, center).uni()
 
-		const MAX_DISTANCE = topLeft.dist(bottomRight) / 1.5
-
-		const distanceToIntersection = handleVector.dist(intersectionVector)
-		const angle = Math.atan2(handleVector.y - center.y, handleVector.x - center.x)
-		let newPoint = handleVector
+		let newPoint = handleInShapeSpace
 
 		if (insideShape) {
-			const direction = Vec2d.FromAngle(angle, MIN_DISTANCE)
-			newPoint = intersectionVector.add(direction)
-			shape.props.handles.handle.x = newPoint.x
-			shape.props.handles.handle.y = newPoint.y
-			return getHandlesInHandleSpace(shape)
-		}
-		if (distanceToIntersection <= MIN_DISTANCE) {
-			const direction = Vec2d.FromAngle(angle, MIN_DISTANCE)
-			newPoint = intersectionVector.add(direction)
-		}
-		if (distanceToIntersection >= MAX_DISTANCE) {
-			const direction = Vec2d.FromAngle(angle, MAX_DISTANCE)
-			newPoint = intersectionVector.add(direction)
+			newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MIN_DISTANCE))
+		} else {
+			if (distanceToIntersection <= MIN_DISTANCE) {
+				newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MIN_DISTANCE))
+			} else if (distanceToIntersection >= MAX_DISTANCE) {
+				newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MAX_DISTANCE))
+			}
 		}
 
-		shape.props.handles.handle.x = newPoint.x
-		shape.props.handles.handle.y = newPoint.y
-		return getHandlesInHandleSpace(shape)
+		return {
+			...next,
+			props: {
+				...next.props,
+				handles: {
+					handle: {
+						...next.props.handles.handle,
+						// handle coordinates in normalized space
+						x: newPoint.x / w,
+						y: newPoint.y / h,
+					},
+				},
+			},
+		}
 	}
 
 	override onHandleChange: TLOnHandleChangeHandler<SpeechBubbleShape> = (
@@ -161,11 +170,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 		newHandle.x = newHandle.x / initial!.props.w
 		newHandle.y = newHandle.y / initial!.props.h
 		const next = deepCopy(initial!)
-		next.props.handles['handle'] = {
-			...next.props.handles['handle'],
-			x: newHandle.x,
-			y: newHandle.y,
-		}
+		next.props.handles.handle = newHandle
 
 		return next
 	}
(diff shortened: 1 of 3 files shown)


apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -18,13 +18,15 @@ import {
 	resizeBox,
 	structuredClone,
 } from '@tldraw/tldraw'
-import { STROKE_SIZES } from '@tldraw/tldraw/src/lib/shapes/shared/default-shape-constants'
```
steveruizok, line 21: We don't want to do this type of deep importing. It would not work outside of our monorepo, plus (IMO) we'd rather have folks copy this stuff out.

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -87,8 +89,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 	}
 
 	getGeometry(shape: SpeechBubbleShape): Geometry2d {
-		const newShape = getHandleinShapeSpace(shape)
```
steveruizok, line 90: These two helpers were a big source of confusion for me. I get the need to switch the handles from normalized space to shape space and back, but it's easier to pull these out into a variable like "handleInShapeSpace" rather than having the concept of a shape with a handle that's sometimes normalized and sometimes not

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -97,60 +98,68 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
...
+
+		return [
+			{
+				...handle,
+				// props.handles.handle coordinates are normalized
+				// but here we need them in shape space
+				x: handle.x * w,
+				y: handle.y * h,
+			},
+		]
```
steveruizok, line 115: Not much is happening here, let's just show it!

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -97,60 +98,68 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
...
+		const { w, h, handles } = next.props
 
-		const { originalIntersection: intersection, insideShape } = getHandleIntersectionPoint({
-			w: shape.props.w,
-			h: shape.props.h,
-			handle: shape.props.handles.handle,
-		})
-
-		if (!intersection) throw new Error('No intersection')
+		const { segmentsIntersection, insideShape } = getHandleIntersectionPoint(next)
```
steveruizok, line 125: Easier to pass the shape in as props, not that we don't have to worry about what coordinate space its handles are in. It's cheap to compute the handle in shape space so we'll just do it where we need it.

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -97,60 +98,68 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
...
-		const handleVector = new Vec2d(shape.props.handles.handle.x, shape.props.handles.handle.y)
+		const slantedLength = Math.hypot(w, h)
+		const MIN_DISTANCE = slantedLength / 5
+		const MAX_DISTANCE = slantedLength / 1.5
 
-		const topLeft = new Vec2d(0, 0)
-		const bottomRight = new Vec2d(shape.props.w, shape.props.h)
-		const center = new Vec2d(shape.props.w / 2, shape.props.h / 2)
-		const MIN_DISTANCE = topLeft.dist(bottomRight) / 5
+		const handleInShapeSpace = new Vec2d(handles.handle.x * w, handles.handle.y * h)
```
steveruizok, line 131: ...like here!

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -97,60 +98,68 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
...
 
-		const MAX_DISTANCE = topLeft.dist(bottomRight) / 1.5
+		const distanceToIntersection = handleInShapeSpace.dist(segmentsIntersection)
+		const center = new Vec2d(w / 2, h / 2)
+		const vHandle = Vec2d.Sub(handleInShapeSpace, center).uni()
 
-		const distanceToIntersection = handleVector.dist(intersectionVector)
-		const angle = Math.atan2(handleVector.y - center.y, handleVector.x - center.x)
-		let newPoint = handleVector
```
steveruizok, line 133: It's confusing because we use Vec2d for both points and vectors, but when naming stuff remember that "vector" is used to indicate a "direction" or 2d distance, while a "point" is used to indicate a "place" (technically a 2d distance from the coordinate space's origin to a point).

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -97,60 +98,68 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
...
-		if (distanceToIntersection >= MAX_DISTANCE) {
-			const direction = Vec2d.FromAngle(angle, MAX_DISTANCE)
-			newPoint = intersectionVector.add(direction)
+			newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MIN_DISTANCE))
+		} else {
+			if (distanceToIntersection <= MIN_DISTANCE) {
+				newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MIN_DISTANCE))
+			} else if (distanceToIntersection >= MAX_DISTANCE) {
+				newPoint = Vec2d.Add(segmentsIntersection, vHandle.mul(MAX_DISTANCE))
+			}
```
steveruizok, line 146: slorp

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -161,11 +170,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 		newHandle.x = newHandle.x / initial!.props.w
 		newHandle.y = newHandle.y / initial!.props.h
 		const next = deepCopy(initial!)
-		next.props.handles['handle'] = {
-			...next.props.handles['handle'],
-			x: newHandle.x,
-			y: newHandle.y,
-		}
+		next.props.handles.handle = newHandle
```
steveruizok, line 164: Since we're already copying and mutating, we can stick the handle on here

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -1,211 +1,158 @@
...
+			// left
+			vertices = [
+				tl,
+				tr,
+				br,
+				bl,
+				new Vec2d(adjustedIntersection.x, adjustedIntersection.y + offsetV),
+				new Vec2d(handleInShapeSpace.x, handleInShapeSpace.y),
+				new Vec2d(adjustedIntersection.x, adjustedIntersection.y - offsetV),
+			]
```
steveruizok, line 66: This was another place where I had to squint pretty hard to get what was going on. This is IMO a good spend of repetition in order to communicate (visually, even!) what's going on.

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -1,211 +1,158 @@
...
 			break
 		default:
 			// eslint-disable-next-line no-console
-			console.log('default')
+			throw Error("no intersection found, this shouldn't happen")
```
steveruizok, line 70: Exhaustive switch statements are great

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -1,211 +1,158 @@
...
-	h,
-	handle,
-}: {
-	w: number
-	h: number
-	handle: TLHandle
-}) {
-	const offset = { horizontal: w / 10, vertical: h / 10 }
-	const handleVec = new Vec2d(handle.x, handle.y)
+export function getHandleIntersectionPoint(shape: SpeechBubbleShape) {
```
steveruizok, line 76: I had a hard time reading through this function. My hunch was that by "flattening out" the helper functions, I could tell the story a bit easier. You'll hear other opinions on this kind of thing, but IMO a flatter function is more useful than splitting subroutines out into their own smaller named functions just for the sake of organization, like you had with `getAdjustedIntersectionPoint` and `checkIntersection`. (Sidebar, this is the opposite of how I feel about React components—split those fuckers out as much as you can!—but that's because unlike stuff like this, bits and pieces of UI tend to develop complexity over time and it's nice to have that complexity contained).

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -1,211 +1,158 @@
...
-	const lineCoordinates = {
-		0: { start: new Vec2d(0, 0), end: new Vec2d(w, 0) },
-		1: { start: new Vec2d(w, 0), end: new Vec2d(w, h) },
-		2: { start: new Vec2d(0, h), end: new Vec2d(w, h) },
-		3: { start: new Vec2d(0, 0), end: new Vec2d(0, h) },
-	}
+	// We want to be sure we get an intersection, so if the point is
+	// inside the shape, push it away from the center by a big distance
+	const pointToCheck = insideShape
+		? Vec2d.Add(handleInShapeSpace, Vec2d.Sub(handleInShapeSpace, center).uni().mul(1000000))
```
steveruizok, line 98: You'll see this pattern a lot: `Vec2d.Add(start, Vec2d.Sub(end, start).uni().mul(distance)`.

It's saying, "find the normalized vector from start to end, then move in that direction by some distance". The resulting point will be `distance` away from `start` in the direction of `end`. Kind of like what you had with the angle, but IMO easier to read and recognize.

In this case, we want to handle that inside case by shoving that other point really far away.

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -1,211 +1,158 @@
...
+
 	// a normalised vector from start to end, so this can work in any direction
-	const unit = Vec2d.Sub(end, start).norm()
+	const unit = Vec2d.Sub(end, start).uni()
 
 	// Where is the intersection relative to the start?
-	const totalDistance = start.dist(end)
-	const distance = intersectionVec.dist(start)
+	const totalDistance = Vec2d.Dist(start, end)
+	const distance = Vec2d.Dist(segmentsIntersection, start)
```
steveruizok, line 128: It's my fault for having both ways of doing stuff in the Vec2d library, but generally Vec2d.Dist is preferred because it works for Vec2d instances or Vec2dModel objects, save some headaches if things ever change.

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -234,28 +181,13 @@ export function intersectLineSegmentLineSegment(
...
-	return newShape
-}
-
-export const getHandlesInHandleSpace = (shape: SpeechBubbleShape): SpeechBubbleShape => {
-	const newShape = deepCopy(shape)
-	newShape.props.handles.handle.x = newShape.props.handles.handle.x / newShape.props.w
-	newShape.props.handles.handle.y = newShape.props.handles.handle.y / newShape.props.h
-
-	return newShape
-}
```
steveruizok, line 261: Wrong direction here. The names are misleading for both functions (returning a shape and not handles), and can be easily confused between them. As mentioned above, it's also confusing to switch between normalized and not normalized values within the same object (shape).

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
```diff
@@ -234,28 +181,13 @@ export function intersectLineSegmentLineSegment(
...
-export function invLerp(a: number, b: number, v: number) {
+function invLerp(a: number, b: number, v: number) {
 	return (v - a) / (b - a)
 }
 /**
  * Maps a value from one range to another.
  * e.g. mapRange(10, 20, 50, 100, 15) => 75
  */
-export function mapRange(a1: number, b1: number, a2: number, b2: number, s: number) {
+function mapRange(a1: number, b1: number, a2: number, b2: number, s: number) {
```
steveruizok, line 191: If they're not used elsewhere, don't export them; it makes it hard to find dead code later on since exports are always treated as "used variables" by your IDE

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
Taha-Hassan-Git: This makes a lot of sense, thanks

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
Taha-Hassan-Git: Thanks for the tip! I definitely heard the opposite opinion a lot when starting out. Clarity should always be the goal though, and this way is clearer for sure.

apps/examples/src/examples/speech-bubble/SpeechBubble/helpers.tsx
Taha-Hassan-Git: Before, the distance to move the point was calculated by using the slanted length of the shape, from top left to bottom right. Is using a really big number preferable?

Taha-Hassan-Git:
I'm really grateful for the time you took to go through this. A lot to learn from here.

PR #2401, opened 2024-01-03
[improvement] account for coarse pointers / insets in edge scrolling

steveruizok:
This PR:
- shrinks the distance for edge scrolling and insets the distance for coarse pointers
- adds edge inset tracking

## Scroll distances

Rather than increasing the distance, we move the "zero" in from the edges, so that the middle of a honkin' fat finger would be at "zero" when the edge of the finger is touching the edge of the screen. This is a bit more reliable than looking at just the component size.

## Inset tracking

We now track whether a shape's edges are identical to the edges of the document body. When an edge is inset, we extend the edge scrolling distance outside of the component, so that dragging PAST the edge of the component will scroll. When an edge is NOT inset, we bring that distance into the component's bounds, so that dragging NEAR TO the edge will begin to scroll.

image

image

image

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Use edge scrolling on mobile
2. Use edge scrolling on desktop
3. Use edge scrolling in the "scrolling example"

- [x] Unit Tests

### Release Notes

- Add `instanceState.insets` to track which edges of the component are inset from the edges of the document body.
- Improve behavior around edge scrolling

### DIFF
--- a/packages/editor/src/lib/constants.ts
+++ b/packages/editor/src/lib/constants.ts
@@ -97,4 +97,7 @@ export const HIT_TEST_MARGIN = 8
 export const EDGE_SCROLL_SPEED = 20
 
 /** @internal */
-export const EDGE_SCROLL_DISTANCE = 32
+export const EDGE_SCROLL_DISTANCE = 8
+
+/** @internal */
+export const COARSE_POINTER_WIDTH = 12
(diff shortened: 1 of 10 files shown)


MitjaBezensek (approved):
Works 👍 

TBH I preferred the old logic. It was consistent and felt like an easier mental model to understand. If you try the basic example on desktop right now it feels quite awkward that the left edge scrolls when you go outside and all the others scroll on the inside 🤷‍♂️

steveruizok:
We could switch alternatively to a boolean, i.e. `isInset`, rather than tracking each edge independently? I had that originally but had the opposite feeling with the example edges.

PR #2405, opened 2024-01-04
Annotate custom styles example

Taha-Hassan-Git:
Adds annotation to the custom styles example

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Add annotation to the custom styles example

### DIFF
--- a/apps/examples/src/examples/custom-styles/FilterStyleUi.tsx
+++ b/apps/examples/src/examples/custom-styles/FilterStyleUi.tsx
@@ -1,11 +1,13 @@
 import { track, useEditor } from '@tldraw/tldraw'
 import { MyFilterStyle } from './CardShape'
 
+// There's a guide at the bottom of this file!
+
+//[1]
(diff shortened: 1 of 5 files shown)


apps/examples/src/examples/custom-styles/FilterStyleUi.tsx
```diff
@@ -33,3 +48,30 @@ export const FilterStyleUi = track(function FilterStyleUi() {
 		</div>
 	)
 })
+
+/* 
+Introduction:
+This is a very simple example of how to create a custom ui for your custom style. We want
```
steveruizok, line 54: Beware of using the word "simple" in docs. It's common that a user might be having a hard time with an example and docs that are overly enthusiastic ("it's easy! simple! my dog could do this!") can be aggravating.

```suggestion
This is a an example of how to create a custom ui for your custom style. We want
```

Issue #2387, opened 2023-12-28
ClipMask issue with nested masking shapes


mohdzaryab:
### What happened?

As soon as you nest two shapes with clipping capability such as frames, and if the width of both the frame becomes same, there is a weird diagonal mask appearing on child frame.

https://github.com/tldraw/tldraw/assets/45558484/d1a52af8-7527-417f-99ce-cd116c889b33

https://github.com/tldraw/tldraw/assets/45558484/c8ef6028-bd4f-4a3e-acc9-5df45238d057

### How can we reproduce the bug?

1. Create a frame F1. 
2. Create another frame F2 and drag it into F1 so that it becomes child of F1.
3. Now, resize F2 so that it has width equal to that of F1. 
4. You will see a mask which is diagonal on F2. To verify, add an image or stroke in F2

### What browsers are you seeing the problem on?

Chrome

### Contact Details

mohammadzaryab0@gmail.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2406, opened 2024-01-04
Fix issues with clip paths for frames

MitjaBezensek:
Fixes an issue with frame clipping paths. In fact, this should also solve other issues we might have with intersect. Seems like our `pointInPolygon` did not correctly detect that points in the corners or on the edges of the polygon are in fact part of the polygon. 

When calculating the intersection of two regular, intersecting rectangles the`intersectPolygonPolygon` was returning a polygon with 2, 3, or sometimes even 0 points, which also could result in an error when dragging one frame out of another frame.

It seems that for all intents and purposes the `pointInPolygon` function should also consider corners and edges, but maybe we might want to rename it?

Before:

https://github.com/tldraw/tldraw/assets/2523721/155d351d-8ceb-47c3-a263-024cab487d03

After:

https://github.com/tldraw/tldraw/assets/2523721/338b923a-f902-4dc4-a1b7-e954f906fb8d

Fixes https://github.com/tldraw/tldraw/issues/2387

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/src/lib/primitives/intersect.ts
+++ b/packages/editor/src/lib/primitives/intersect.ts
@@ -243,26 +243,32 @@ export function intersectPolygonPolygon(
 	polygonA: VecLike[],
 	polygonB: VecLike[]
 ): VecLike[] | null {
-	// Create an empty polygon as P
-	const result: VecLike[] = []
+	// Create an empty polygon as result
+	const result: Map<string, VecLike> = new Map()
 	let a: VecLike, b: VecLike, c: VecLike, d: VecLike
 
-	// Add all corners of PolygonA that is inside PolygonB to P
+	// Add all corners of PolygonA that is inside PolygonB to result
 	for (let i = 0, n = polygonA.length; i < n; i++) {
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/primitives/intersect.ts
```diff
@@ -243,26 +243,32 @@ export function intersectPolygonPolygon(
 	polygonA: VecLike[],
 	polygonB: VecLike[]
 ): VecLike[] | null {
-	// Create an empty polygon as P
-	const result: VecLike[] = []
+	// Create an empty polygon as result
+	const result: Map<string, VecLike> = new Map()
```
MitjaBezensek, line 247: Had to add deduping to intersect since polygons that have same points might add these points twice.

steveruizok (approved):
Nice!

PR #2409, opened 2024-01-05
Positional keyboard shortcuts for toolbar

SomeHats:
Adds positional keyboard shortcuts to the toolbar. Use the 1, 2, 3, 4 etc keys to activate the corresponding tool on the toolbar.

Kapture 2024-01-05 at 11 52 30

### Change Type
- [x] `minor` — New feature

### Test Plan

1. Use the number keys to activate toolbar items.
- [x] End to end tests

### Release Notes

- You can now use the number keys to select the corresponding tool from the toolbar


steveruizok:
This one might need some time for me to think through it. IIRC I avoided these shortcuts because of the design question—should they be based on positions? Should we force this as part of our UI, or make it part of the `tool` definition so that it could be more easily customized?

SomeHats:
i would love for the "main"/"direct" keyboard shortcuts (e.g. 'd' for 'draw') to be customisable.

I think these extra ones make sense being strictly positional though (maybe with an option to disable if they conflict with your own app?) - there's no reason for 3 to be draw other than it being the 3rd item, but it is a very quick way to navigate the toolbar without having to learn a bunch of shortcuts or reach for your mouse. there's a lot of prior art for these in other apps that have a horizontal toolbar like ours

steveruizok:
Ok, let's go!

steveruizok (approved):

mimecuvalo:
neat!

PR #2413, opened 2024-01-05
Annotate exploded example

Taha-Hassan-Git:
Alters the exploded example so that it imports all the default components and annotates with an explanation.

I felt like the purpose of the exploded example was to show everything so that users could see what was available for them to customise. I think this is a step further in that direction.

I found it tricky to work around the double meaning of components here, I think that needs looking at. Components is a prop on the TldrawEditor, but it also means something in React. Makes explaining certain things difficult, e.g : 

"Here we've passed the default components object to the TldrawEditor component. Along
with default tools and shapeutils, You could input your own custom shapes/tools here. 
For help creating your own shapes/tools check out the custom config example.

We also set the initial state to 'select' and render the UI, context menu and canvas
components. You could add your own custom components here, omit these ones, and/or 
change the initial state of the application to whatever you want. "

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds annotation to exploded example
- imports all default components instead of just a few


steveruizok:
Hm, not sure I agree with adding the default components here. We have a few that we override within on the tldraw library, and that are included here to show that they're necessary (to avoid the more minimal defaults), which was a common stumbling block for people. If we don't have a separate example for the component prop overrides, we should make one.

steveruizok (approved):

PR #2416, opened 2024-01-07
Update README.md to reflect it is now source available rather than open source.

mrcodechef:
Update README.md to reflect it is now source available rather than open source after license change.


steveruizok:
Appreciate the note. We'll consider the language around open source.

PR #2424, opened 2024-01-08
Maintain bindings whilst translating arrows

SomeHats:
This diff tries to maintain bindings whilst translating arrows. It looks at where the terminal of the arrow ends up, and if it's still over the same shape, it updates the binding to a precise one at that location rather than removing the binding entirely.

Kapture 2024-01-08 at 18 22 12

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Create an arrow with bindings
2. Move the arrow (translation, stacking, nudging, distribution, etc)
3. Make sure that the end point of the arrow remains bound if appropriate

- [x] Unit Tests

### Release Notes

- You can now move arrows without them becoming unattached the shapes they're pointing to

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5062,6 +5062,35 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this
 	}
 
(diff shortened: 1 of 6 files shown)


steveruizok:
Hey, just did one superficial change (moving a method out of the class). Love this PR!

PR #2428, opened 2024-01-09
Mitja/validate urls

MitjaBezensek:
Adds validation for urls we use for our shapes and assets. This PR includes a migration so we should check that existing rooms still load correctly. There might be some that won't, but that means that they had invalid url set.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan
1. Existing rooms should still load correctly (there should be no validation errors).
2. Adding new images and videos should also work (test both local and multiplayer rooms as they handle assets differently).

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Add validation to urls.


SomeHats (approved):
lgtm!

futugyou:
Have you ever thought that I might use local resources, like './image/tldraw.svg'.
You can visit https://examples-git-mitja-validate-urls-tldraw.vercel.app/local-images, it throw an error.

MitjaBezensek:
Thanks for the heads up, will fix this in a follow up PR 🙌

PR #2440, opened 2024-01-10
Fix the publish script

MitjaBezensek:
Follow up to https://github.com/tldraw/tldraw/pull/2439. Should fix the publishing script so that the next time we run it we don't run in the the same issue.

We could also move the file directly to the `src` dir, but it seems since only `ui` uses the version it was nested under the `ui` dir 🤷

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### DIFF
--- a/scripts/publish-new.ts
+++ b/scripts/publish-new.ts
@@ -62,8 +62,7 @@ async function main() {
 		'add',
 		'lerna.json',
 		...packageJsonFilesToAdd,
-		BUBLIC_ROOT + '/packages/*/src/version.ts',
-		BUBLIC_ROOT + '/packages/*/src/lib/ui/version.ts',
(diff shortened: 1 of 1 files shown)


scripts/publish-new.ts
```diff
@@ -63,6 +63,7 @@ async function main() {
 		'lerna.json',
 		...packageJsonFilesToAdd,
 		BUBLIC_ROOT + '/packages/*/src/version.ts',
+		BUBLIC_ROOT + '/packages/*/src/lib/ui/version.ts',
```
ds300, line 66: I think it should be safe to do this, just in case we move those files around at some point.

```suggestion
		BUBLIC_ROOT + '/packages/*/src/**/version.ts',
```

ds300 (approved):
happy with this change, or my suggestion 👍🏼

PR #2442, opened 2024-01-10
Don't bother measuring canvas max size for small images

### LINKED ISSUE #2433: [Android] First 'copy as' is slow
The first time you 'copy as' has a delay. Following 'copy as' are quick, even if you copy different shapes, etc.

1. Refresh the page.
2. 'Copy as' some shapes.
3. 'Copy as' some shapes (can be the same shapes, or different shapes).

Expected: Both happen quickly.
Actual: First is slow, second is fast.

https://github.com/tldraw/tldraw/assets/15892272/5bba7b3e-f4be-49d6-ba18-6e259877dc64

TodePond:
This PR makes it so we don't bother measuring the max size of your browser canvas - if you're manipulating an image that is relatively small (ie: less than 8K resolution).

Closes #2433 

@SomeHats I reckon this has some conflicts with your PR, #2441 

## Why?

Calculating the max canvas size... takes a long time on some devices, eg: Android Chrome.

Most of the time, you don't need such a big canvas anyway. When you're exporting/importing a small image, we can assume that the max size will be big enough (based on some testing [here](https://github.com/jhildenbiddle/canvas-size?tab=readme-ov-file#test-results)). So we don't need to bother in those cases - speeding up the operation.

If you're manipulating a larger image, we measure the max canvas size - as before.

## Anything else?

This PR also moves the 'clamping' code to one place. Previously, it was repeated in three places, and it was annoying having to update them all at the same time.

## What's next?

Measuring for large images is still slow. We should look at how to make that experience better. eg: Visual feedback that it's happening. eg: Try not to block the main UI thread. eg: Can we do anything in parallel?

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Draw some shapes.
2. Select them.
3. Copy/export as png.
4. Repeat with various sizes of shapes. eg: Select shapes over a large area, so that your exported image is larger than 8K resolution.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Android: Sped up exporting and importing images.


steveruizok (approved):

SomeHats (approved):
love it, lets do this and i'll update my changes to use yours

TodePond:
@SomeHats this PR has now merged in your PR's changes! ready to review again

SomeHats (approved):
lets goooo

PR #2447, opened 2024-01-11
Fix validation for local files.

MitjaBezensek:
Allow urls for local files. This addresses the comment from [here](https://github.com/tldraw/tldraw/pull/2428#issuecomment-1886221841).

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Local images example should now work. We use images from the public folder there.

### DIFF
--- a/packages/validate/src/lib/validation.ts
+++ b/packages/validate/src/lib/validation.ts
@@ -617,6 +617,13 @@ function parseUrl(str: string) {
 	try {
 		return new URL(str)
 	} catch (error) {
+		if (str.startsWith('/') || str.startsWith('./')) {
+			try {
+				return new URL(str, 'http://example.com')
+			} catch (error) {
+				throw new ValidationError(`Expected a valid url, got ${JSON.stringify(str)}`)
+			}
+		}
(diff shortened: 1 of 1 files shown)


packages/validate/src/lib/validation.ts
```diff
@@ -617,6 +617,13 @@ function parseUrl(str: string) {
 	try {
 		return new URL(str)
 	} catch (error) {
+		if (str.startsWith('/') || str.startsWith('./')) {
+			try {
+				return new URL(str, 'http://example.com')
```
steveruizok, line 622: Is including `https://example.com` correct here? Not sure I follow

MitjaBezensek: We first check if we have a fully qualified url (first try catch). If we don't we then check if we have a local image like in our `LocalImage` example: `/tldraw.png`. URL constructor only works with fully qualified urls, so we could either use the current base url or just a dummy one like `example.com` 🤷‍♂️

steveruizok: Sounds good, let's land it!

Issue #2264, opened 2023-11-29
[Feature]: ⌘ D to repeat last duplication direction instead of always defaulting to the right


ddramone:
### What's the feature?

**Current Behavior:**

In the current version using the '⌘ + D' shortcut duplicates an object directly to its right. This is a straightforward duplication method but lacks flexibility for complex designs.

**Requested Feature:**

I suggest to remember the last duplication action and apply the same directional pattern for subsequent duplications. 

For example, if I use 'Alt + Drag' to move an object downwards and then press 'Command + D', the new duplication should occur in the same direction (downwards in this case), rather than defaulting to the right. 

This would greatly enhance the efficiency of the workflow, especially in designing complex layouts.

- - -

I would love to contribute If someone could point me towards the relevant sections of the codebase.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ddramone:
For the reference, this is how it works in Figma:


https://github.com/tldraw/tldraw/assets/2891206/025b7bcd-868a-4508-ae61-7d09c4cf98d1

PR #2480, opened 2024-01-16
Improved duplication

MitjaBezensek:
Improves duplication. It will now remember the offset and shapes of your `alt + drag` duplication and use that if you use the duplicate action immediately after this.

duplicate

Fixes https://github.com/tldraw/tldraw/issues/2264

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -37,6 +37,9 @@ export class Idle extends StateNode {
 	override onPointerDown: TLEventHandlers['onPointerDown'] = (info) => {
 		if (this.editor.getIsMenuOpen()) return
 
+		// We want to clear the duplicate props when we deselect the current shapes or when we select other shapes
```
steveruizok, line 40: There are a few ways to deselect shapes, so this won't cover all of them. If we really wanted to be sure, we would need to have the side effect manager clear this part of the state, or have a reactor in the select tool constructor that clears it when the state changes.

In fact, maybe this state should belong to the select tool rather than the global instance state. (There's a bunch of stuff in the instance state that feels kind of tooly rather than instancey, but this could be a first try at a pattern where the state stays with the tool?)

MitjaBezensek: Sounds good, will give it a try 👍

steveruizok:
Looks good!

steveruizok (approved):

PR #2484, opened 2024-01-17
Fix the first run of dev script.

MitjaBezensek:
When you first clone the repo and run the `dev` script for the first time you get the following error:
image

This is because the `dev` script in examples runs before `dev` in `tldraw` (which creates the `tldraw.css` file). We now have a `predev` step that does it.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Clone the repo from start (or delete `packages/tldraw/tldraw.css` file)
2. Run `yarn dev`
3. You should not see the above error.

### Release Notes

- Fix first `yarn dev` experience.

### DIFF
--- a/packages/tldraw/package.json
+++ b/packages/tldraw/package.json
@@ -31,7 +31,8 @@
 	"scripts": {
 		"test": "lazy inherit",
 		"test-coverage": "lazy inherit",
-		"dev": "chokidar '../{editor/editor,tldraw/src/lib/ui}.css' -c 'node ./scripts/copy-css-files.mjs' --initial",
+		"predev": "node ./scripts/copy-css-files.mjs",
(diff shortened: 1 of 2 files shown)


packages/tldraw/package.json
```diff
@@ -31,7 +31,8 @@
 	"scripts": {
 		"test": "lazy inherit",
 		"test-coverage": "lazy inherit",
-		"dev": "chokidar '../{editor/editor,tldraw/src/lib/ui}.css' -c 'node ./scripts/copy-css-files.mjs' --initial",
```
MitjaBezensek, line 34: I guess we don't need `--initial` any longer, since `predev` will now do it.

ds300 (approved):

PR #2498, opened 2024-01-17
auto-it/typescript fails the release, patch the problem away while mods are asleep

si14:
Currently main can't be released: https://github.com/tldraw/tldraw/actions/runs/7557438453/job/20576664393

The issue was bisected down to https://github.com/tldraw/tldraw/commit/29044867dd2e49a3711e95c547fa9352e66720b9 and can be "fixed" by forcing TypeScript to the previous resolution (5.3.3->4.9.5).

It looks like Auto runs node-ts on everything, and Typescript picks up a module/moduleResolution mismatch somewhere. It is likely that the mismatch is not in our code, since blind replacement of all instances of `"moduleResolution"` to `"NodeNext"` in the project doesn't help.

It is likely that the relevant TypeScript change is this one: https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-moduleresolution-compiler-option

Further investigation is pending, but this PR should fix the release process.

### Change Type


ds300 (approved):

ds300:
@si14 added an upstream PR for this https://github.com/intuit/auto/pull/2420

PR #2499, opened 2024-01-17
Introduce a Cloudflare health worker

si14:
This PR introduces a new Cloudflare worker for health checks.

At the moment the worker only translates Updown webhooks into Discord webhooks. In the future we can teach this worker to check more things.

### Change Type


scripts/deploy.ts
```diff
@@ -100,16 +103,19 @@ async function main() {
...
-	await discordStep('[4/6] deploying asset uploader to cloudflare', async () => {
+	await discordStep('[4/7] deploying asset uploader to cloudflare', async () => {
 		await deployAssetUploadWorker({ dryRun: false })
 	})
-	await discordStep('[5/6] deploying multiplayer worker to cloudflare', async () => {
+	await discordStep('[5/7] deploying multiplayer worker to cloudflare', async () => {
 		await deployTlsyncWorker({ dryRun: false })
 	})
+	await discordStep('[6/7] deploying health worker to cloudflare', async () => {
+		await deployHealthWorker({ dryRun: false })
```
ds300, line 113: you should add a `dryRun: true` version up above in the 'cloudflare deploy dry run' zone

apps/health-worker/src/discord.ts
```diff
@@ -0,0 +1,96 @@
+import { exhaustiveSwitchError } from '@tldraw/utils'
+import { type APIEmbed } from 'discord-api-types/v10'
+import { Event as UpdownEvent } from './updown_types'
+
+// discord wants decimal colours
+const GREEN = 4243543
+const RED = 14692657
+const ORANGE = 16213767
```
ds300, line 8: weird, never seen colors passed around like this.

ds300 (approved):
Nice one! 🙏🏼

apps/health-worker/src/discord.ts
si14: yeah, it's a weird API

scripts/deploy.ts
si14: fixed

PR #2520, opened 2024-01-18
Allow snapping of shapes to the frame when dragging inside the frame.

MitjaBezensek:
Allows you to snap to frames when dragging inside them.

https://github.com/tldraw/tldraw/assets/2523721/41816b9b-5969-416d-af15-77b8f102ad21

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Create a frame.
2. Add some shape inside.
3. Drag the shapes while holding `cmd` or turning on always snap. You should be able to snap to the edges and the centre of the frame.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds snapping to frames when dragging shapes inside a frame.


ds300 (approved):
yeah this seems sensible. the edge snapping will presumably stop working as intended when the frame is rotated at non-90° increments, but what kind of monster rotates frames anyway.

Issue #2486, opened 2024-01-17
Need to do a hard reload to get the new styles


SomeHats:
When opening staging I didn't not see the new dark mode colors. I had to refresh the page to get them. Happened in different browsers and also on two different devices.

Probably a service worker caching issue.

PR #2552, opened 2024-01-18
[dotcom] Delete service worker, cache tldraw assets

ds300:
A few things happening here

- Delete our service worker. Turns out that a couple of years back browsers decided that a service worker is no longer required for a PWA so you can just have the manifest and still install on the user's device.
- Cache tldraw's assets as part of the dotcom vite asset pipeline. This allows them to participate in the asset coalescing (preserving old versions of asset files so old clients don't stop working when you deploy new versions of things, see https://github.com/tldraw/brivate/pull/3132 for more context).
- Add a new 'imports.vite.js' file to the assets package, because we import a bunch of json translation files, and vite imports .json files as parsed json objects instead of string urls, and there's no good way to tell it not to. Even if there was we wouldn't want to impose that config on our users. So another way to tell vite to load any asset as a url string is to append `?url` to the end of the import path. That's what this file does. 

closes [#2486](https://github.com/tldraw/tldraw/issues/2486)

### Change Type

- [x] `minor` — New feature

### Release Notes

- Fix 'could not load assets' error that we often see on tldraw.com after a deploy


apps/dotcom/src/utils/assetUrls.ts
```diff
@@ -1,21 +1,7 @@
...
+import { getAssetUrlsByImport } from '@tldraw/assets/imports.vite'
 
-export const assetUrls: AssetUrls = isProductionEnv
-	? _assetUrls
-	: // let's try out shantell sans 'informal' style for a bit on staging/dev
-		{
-			..._assetUrls,
-			fonts: {
-				..._assetUrls.fonts,
-				draw: '/Shantell_Sans-Tldrawish.woff2',
```
ds300, line 16: our custom font is shipped everywhere now, so this can be removed

MitjaBezensek (approved):

mimecuvalo:
yesss exciting

PR #2558, opened 2024-01-19
debug: add FPS counter

mimecuvalo:
Adds an FPS counter to detect when there's a UI slowdown.
(btw, drive-by typo fix for a file)

https://github.com/tldraw/tldraw/assets/469604/b83d4b10-35d9-4584-af46-c63b5cc107ac

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

1. 

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds FPS counter to debug panel.


packages/tldraw/src/lib/ui/components/DebugPanel.tsx
```diff
@@ -68,6 +82,39 @@ const CurrentState = track(function CurrentState() {
 	return <div className="tlui-debug-panel__current-state">{editor.getPath()}</div>
 })
 
+const times: number[] = []
+let isFPSCounterRunning = false
+let fps = 0
+function refreshFpsLoop() {
+	window.requestAnimationFrame(function () {
```
ds300, line 89: we have a batched requestAnimationFrame called `throttledRaf` that we use for performing store changes. I think you'd get more accurate results by using that. With this code as-is I believe it's actually testing how many extra animation frames it can fit between the frames we actually paint on?

ds300 (approved):
nice! great idea.

packages/tldraw/src/lib/ui/components/DebugPanel.tsx
ds300: actually forget my advice, I think it'll work as intended.

packages/tldraw/src/lib/ui/components/DebugPanel.tsx
```diff
@@ -68,6 +82,39 @@ const CurrentState = track(function CurrentState() {
...
+	return (
+		<div
+			className={classNames('tlui-debug-panel__fps', {
+				'tlui-debug-panel__fps__slow ': isLow,
+			})}
+		>
+			FPS {fps}
+		</div>
+	)
+}
```
steveruizok, line 116: My instinct here would be to put the frame-by-frame stuff outside of the react lifecycle (i.e. with a little canvas). We'd be adding noise to the system by pushing frame-to-frame updates through React.

steveruizok: ie the test should be as separate as possible from the system it is measuring

mimecuvalo: talked about this offline, just repeating here for context:
but yeah, totally. right now `refreshFpsLoop` is measuring outside of the React lifecycle and this render loop inside the component is just to "grab" that measurement number from time-to-time. We could make it grab it once a second if we wanted to it instead of every 33ms

steveruizok: A lot of our performance comes from how (and how often) the React tree updates. By pushing a state change through React every frame, we are adding stress to that system and will add noise to other debugging tools. My suggestion is to avoid react state when updating the count: i.e. run the loop inside of an effect and imperatively paint a canvas or update an element's innerText, just to avoid that per-frame react state update.

si14 (commented):
Couple comments

mimecuvalo:
thanks all for your feedback! addressed all the concerns 👍

PR #2559, opened 2024-01-19
Incremental snap shapes calculation.

MitjaBezensek:
Compute the snappable shapes incrementally.

There's a bit of a [gotcha with selections](https://github.com/tldraw/tldraw/pull/2559/files#diff-89d2d9565f6feb3808fb9e0d7a055bc0d327e2e8a4a50df5cd23aa02feff8bfeR356-R367). Here's the scenario.
- We have shape A and shape B.
- User selects shape A and starts dragging it around.
- At this point snappable shapes consists of shape B (we don't allow snapping to selected shapes, so shape A does not apply).
- User now selects shape B.
- We have to remove shape B from the list of snappabable shapes, but we also have to add shape A to the list (it was previously selected so it was not considered, but since it is now no longer selected it should be).

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Try various snapping options: see if snapping shapes flat on the page works, then drag shapes inside a frame and see if it correctly starts snapping to shapes within the frame (and the frame itself). Also try snapping to grouped shapes (we don't snap to to group itself, but it should snap to individual shapes within it).


MitjaBezensek:
On @ds300's suggestion we decided to postpone this. There's a slight hiccup with it - it seems to freshly calculate things when you move the camera. 

Closing this for now, we can always pick it up later.

PR #2560, opened 2024-01-19
debug: start adding more tooling for debugging when interacting with shapes

mimecuvalo:
I borrowed `useTick` from `GeometryDebuggingView.tsx` — didn't see a good place to have that in a shared file somewhere but lemme know if there is ¯\_(ツ)_/¯ 

https://github.com/tldraw/tldraw/assets/469604/89ae4bb1-0680-4275-8e34-ead7c0cd091e

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds more information in the debug view about what shape is selected and coordinates.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/DebugPanel.tsx
+++ b/packages/tldraw/src/lib/ui/components/DebugPanel.tsx
@@ -63,9 +63,34 @@ export const DebugPanel = React.memo(function DebugPanel({
 	)
 })
 
-const CurrentState = track(function CurrentState() {
+function useTick(isEnabled = true) {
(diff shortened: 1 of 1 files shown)


ds300 (approved):
It's unfortunate that the `editor.inputs` fields are not reactive 🤔 . But this `useTick` seems like a sensible workaround for now.

This extra info seems useful. I could definitely use pointer coords in page space too, I have wanted that on multiple occasions when debugging rendering/geometry stuff. Feel free to merge as-is though.

Issue #2550, opened 2024-01-18
[Feature]: Trackpad Enhancement


Zenkylo:
### What's the feature?

Using trackpads such as the ones built into laptops or external ones (Apple Magic Trackpad) can sometimes be tricky to use with drawing tools such as tldraw. They can be difficult because because to initiate a tool such as _Draw_ or _Rectangle_ pressure needs to be applied and maintained to the trackpad _while_ you move your finger to create your drawing or shape. 
It would be helpful if with your non-trackpad hand you could hold down a key, `ctrl` or example, that would emulate the mousedown event. While the `ctrl` key is in its down state you could use your trackpad hand to create your drawing or shape without having to apply pressure to the trackpad. To end your drawing or shape you'd release the `ctrl` key.

Hope this is explained well enough. Let me know if more clarity is needed.

### Contact Details

k4nderson@gmail.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

steveruizok:
Hey, we do this already with the comma key! But it wasn't working the way I expected it to anymore, so I've made a quick fix.

Zenkylo:
Amazing. Thank you Steve. Works very well. I see the list of keyboard shortcuts in the app via the __?__ menu in the bottom right. However, the `,` behavior isn't listed. Perhaps because it's not technically a keyboard shortcut. Is this behavior described somewhere? I may have had trouble finding it.

Taha-Hassan-Git:
Thanks for the suggestion, I've opened an issue for that here: https://github.com/tldraw/tldraw/issues/3531

Closing this now.

PR #2568, opened 2024-01-21
[improvement] better comma control for pointer

steveruizok:
This PR fixes a few bugs with the "comma as pointer" feature.

In tldraw, the `,` key can be used as a replacement for "pointer down" and "pointer up". This is most useful on laptops with trackpads that make dragging inconvenient. (See https://github.com/tldraw/tldraw/issues/2550).

Previously, the canvas had to be focused in order for the comma key to work. If you clicked on a menu item and then pressed comma, it would not product a pointer event until you first clicked on the canvas. This is now fixed by moving the listener out of the `useDocumentEvents` and into `useKeyboardShortcuts`.

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Click the canvas.
2. Use the comma key to control pointer down / up.
3. Click a shape tool on the toolbar.
4. Move your mouse over the canvas.
5. Press the comma key. It should produce a dot / shape / etc

### Release Notes

- Improve comma key as a replacement for pointer down / pointer up.

### DIFF
--- a/packages/editor/src/lib/hooks/useDocumentEvents.ts
+++ b/packages/editor/src/lib/hooks/useDocumentEvents.ts
@@ -1,6 +1,6 @@
 import { useValue } from '@tldraw/state'
 import { useEffect } from 'react'
-import { TLKeyboardEventInfo, TLPointerEventInfo } from '../editor/types/event-types'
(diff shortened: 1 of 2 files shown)


steveruizok:
Closes #2550

PR #2585, opened 2024-01-22
Grouping examples into categories

Taha-Hassan-Git:
This PR adds collapsible groups to the examples app.

it's not finished, but I'd like a review before continuing as I've made a few decisions I'd like feedback on. I'd like to make a separate issue for abstracting the accordion component, as I wasn't sure how I would do it and I thought it would be best to prioritise the functionality first. Especially considering there are more pressing issues to be getting on with.

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Add collapsible categories to the examples app


apps/examples/src/ExamplePage.tsx
```diff
@@ -39,41 +40,69 @@ export function ExamplePage({
...
+						</Accordion.Item>
+					))}
+				</Accordion.Root>
 			</div>
 			<div className="example__content">{children}</div>
 		</div>
 	)
 }
+
+const Chevron = () => {
```
mimecuvalo, line 89: feels like this isn't the best place to put this - but i'm not familiar yet with a better place in the codebase to put this? i guess it's fine for now ¯\_(ツ)_/¯

mimecuvalo, line 89: feels like this isn't the best place to put this - but i'm not familiar yet with a better place in the codebase to put this? i guess it's fine for now ¯\\_(ツ)_/¯

apps/examples/src/ExamplePage.tsx
```diff
@@ -39,41 +40,69 @@ export function ExamplePage({
...
+								<div className="examples__list__item accordion__trigger__container">
+									<h3 className="accordion__trigger__heading">{c}</h3>
+									<Chevron />
+								</div>
+							</Accordion.Trigger>
+							<Accordion.Content className="accordion__content">
+								<span className="accordion__content__separator"></span>
+								<div className="accordion__content__examples">
+									{examples
+										.find((e) => e.id === c)
```
mimecuvalo, line 69: nit: i know this was like this before but it'd be better to avoid a short name like `e` because it can better to avoid shortnames to improve readability. so maybe something like `sidebarExample`? (open to other ideas)

apps/examples/src/ExamplePage.tsx
```diff
@@ -39,41 +40,69 @@ export function ExamplePage({
...
-							/>
-						))}
-				</ul>
+				<Accordion.Root
+					type="multiple"
+					defaultValue={categories}
+					className="example__info__list scroll-light"
+					ref={scrollElRef}
+				>
+					{categories.map((c) => (
```
mimecuvalo, line 57: nit: (similar to below), it'd be better to have `category` vs `c` - easier to read the code and also easier to grep through when looking for where the variable is used

apps/examples/src/HomePage.tsx
```diff
@@ -1,5 +1,5 @@
 import ExamplesTldrawLogo from './components/ExamplesTldrawLogo'
```
mimecuvalo, line 1: i think we should just kill this page but that's neither here nor there :)

apps/examples/src/HomePage.tsx
```diff
@@ -14,22 +14,11 @@ export function HomePage() {
 				</p>
 			</div>
 			<ul className="examples__list">
-				{examples
-					.filter((example) => !example.hide)
```
mimecuvalo, line 18: this logic doesn't apply anymore?

apps/examples/src/components/SpanLink.tsx
```diff
@@ -6,7 +6,7 @@ import { useMergedRefs } from '../hooks/useMegedRefs'
 import { StandaloneIcon } from './Icons'
 import { Markdown } from './Markdown'
 
-export const ListLink = forwardRef(function ListLink(
+export const SpanLink = forwardRef(function ListLink(
```
mimecuvalo, line 9: may i suggest actually just calling this Link?
in the future it might not be a `span` (but, let's say a div) and we wouldn't want to rename this again :)

apps/examples/src/examples.tsx
```diff
@@ -7,22 +7,99 @@ export type Example = {
 	path: string
 	codeUrl: string
 	hide: boolean
+	category: 'basic' | 'editor' | 'ui' | 'collaboration' | 'data/assets' | 'shapes/tools' | null
```
mimecuvalo, line 10: i don't think null should be a category option? is there a good reason for that?

apps/examples/src/examples.tsx
```diff
@@ -7,22 +7,99 @@ export type Example = {
...
 	order: number | null
 	componentFile: string
 	loadComponent: () => Promise<ComponentType>
 }
 
-export const examples = (
+const basicExamples = (
```
mimecuvalo, line 16: so, for this logic below, what would be better is to consolidate it (make it so when adding a category/removing a category, you don't have to think of another place where you have to add a codeblock/maintain extra code to make something work)

we should make it more like:
```
enum CATEGORIES = {
  BASIC = 'basic',
  EDITOR = 'editor'
  etc...
}

const CATEGORY_TO_TITLE = {
  BASIC = 'Getting Started',
  EDITOR = 'Editor API'
  etc...
}

export const examples = CATEGORIES.map(category => {
  Object.values(import.meta.glob('./examples/*/README.md', { eager: true })) as Example[]
)
	.filter((e) => e.category === category)
  ...

  return { id: CATEGORY_TO_TITLE[category], values: ...}
})
```
(also, above in the type, you can do something like: `category: CATEGORIES` instead of listing it out again)

apps/examples/src/examples.tsx
```diff
@@ -7,22 +7,99 @@ export type Example = {
...
+		} else if (b.order === null) {
+			return -1
+		} else {
+			return a.order - b.order
+		}
+	})
+
+export const examples = [
+	{ id: 'Getting Started', array: basicExamples },
+	{ id: 'UI/Theming', array: uiExamples },
```
mimecuvalo, line 100: this is touched upon above but `array` is too specific on what basicExamples is. it might be that basicExamples becomes an object in the future. a better wording here would be `values` because it might change.

apps/examples/src/styles.css
```diff
@@ -88,12 +86,65 @@ html,
 	}
 }
 
+.accordion__trigger{
```
mimecuvalo, line 89: nit: (does prettier take care of this?) should be a space before `{` (and below)

apps/examples/src/styles.css
```diff
@@ -88,12 +86,65 @@ html,
...
+	font-size: 10px;
+}
+
+.accordion__trigger__chevron{
+	color:rgba(0, 0, 0, 0.3);
+	transition: transform 300ms;
+	-webkit-transition: transform 300ms;
+	-moz-transition: transform 300ms;
+	-ms-transition: transform 300ms;
+	-o-transition: transform 300ms;
```
mimecuvalo, line 104: you don't need to do the browser prefixes anymore
(i'm assuming that we're using postcss or something as the processor here, i could be wrong) - but you just need to do `transition`

apps/examples/src/styles.css
```diff
@@ -88,12 +86,65 @@ html,
...
+	margin-top: 0.25rem;
+	margin-bottom: 0.25rem;
+}
+
+.accordion__content{
+	display: flex;
+	flex-direction: row;
+}
+.accordion__content__separator{
+	border-right: 1px solid #e8e8e8;
```
mimecuvalo, line 129: would love to see these colors in a consolidated variable somewhere but i won't press you on this if it's not really strictly done in this app

apps/examples/src/examples/api/README.md
```diff
@@ -1,7 +1,8 @@
 ---
 title: Editor API
 component: ./APIExample.tsx
-order: 2
+category: editor
+order: 1
```
mimecuvalo, line 5: did i miss where these `order`s are being used? maybe you pulled it out later?

in any case, it's a moot point: we shouldn't have `order` - if you pull an example or add a new one, you would have to change all the numbers in several files - no fun :P

we can talk about different approaches on this but, either we have maybe a `priority` option where some things bubble up higher or lower in the list, or, i guess, alphabetize.

mimecuvalo (commented):
great start @Taha-Hassan-Git ! i think the accordion feels great :)

couple comments — lemme know if you have any questions!

apps/examples/src/HomePage.tsx
Taha-Hassan-Git: There are one or two hidden examples, not sure why they've been hidden though.

apps/examples/src/styles.css
Taha-Hassan-Git: Weirdly, it doesn't. Might have to ask someone about this.

apps/examples/src/examples/api/README.md
Taha-Hassan-Git: I forgot to sort them! I was going to order them all according to their level of complexity, but you raise a good point about maintainability. I think I'll go for priority, a number between 1 and 3 that communicates how complex or difficult the example is.

apps/examples/src/HomePage.tsx
mimecuvalo: Oh, ok - then we should keep this logic around?

.github/ISSUE_TEMPLATE/request_example.yml
```diff
@@ -0,0 +1,33 @@
+name: Example Request
```
mimecuvalo, line 1: i think this might have leaked out from your other PR ( https://github.com/tldraw/tldraw/pull/2597 ) :) let's remove

apps/examples/vite.config.ts
```diff
@@ -82,9 +83,14 @@ function parseFrontMatter(data: unknown, fileName: string) {
 		throw new Error(`Frontmatter key 'component' must be string in ${fileName}`)
 	}
 
-	const order = 'order' in data ? data.order : null
-	if (order !== null && typeof order !== 'number') {
-		throw new Error(`Frontmatter key 'order' must be number in ${fileName}`)
+	const priority = 'priority' in data ? data.priority : null
+	if (priority !== null && typeof priority !== 'number') {
```
mimecuvalo, line 87: minor: we should get rid of `priority !== null`
and prbly same goes for category check below too

apps/examples/src/styles.css
```diff
@@ -1,5 +1,12 @@
 @import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;700&display=swap');
 
+:root {
+	--gray-light: #f5f5f5;
```
mimecuvalo, line 4: niiiice

apps/examples/src/examples/canvas-events/CanvasEventsExample.tsx
```diff
@@ -5,12 +5,28 @@ import { useCallback, useState } from 'react'
...
 
 export default function CanvasEventsExample() {
+	// [1]
 	const [events, setEvents] = useState<string[]>([])
-
+	// [2]
 	const handleEvent = useCallback((data: TLEventInfo) => {
-		setEvents((events) => [JSON.stringify(data, null, 2), ...events.slice(0, 100)])
+		setEvents((currentEvents) => {
+			const parsedEvents = currentEvents.map((e) => JSON.parse(e))
```
mimecuvalo, line 13: this is a great start! already better than what we had - let's split this change into a separate PR - i think it still needs a little more work and I don't wanna hold up this larger change :)

mimecuvalo (approved):
pre-approving modulo the latest comments before you land please!

apps/examples/src/examples/canvas-events/CanvasEventsExample.tsx
Taha-Hassan-Git: This leaked as well!

PR #2594, opened 2024-01-23
Update README and examples copy in the docs

Taha-Hassan-Git:
This PR should make the examples page copy clearer:

1. Moved the examples app copy above the framework specific and yjs copy, as it's more likely to be what the user is looking for.
2. Moved the codesandbox below the copy, user no longer needs to scroll to see most of the text.
3. Gives more hints about what can be found in the examples app and also highlights that running it locally is the better experience.
4. Added some clarification on the relationship between the framework-specific examples/yjs example and the examples app
5. Updates the tldraw repo README accordingly.
6. Removes (development) from the Basic example.

Next steps:
- The codesandbox
   - quite noisy, would be good to hide the navigation
   - doesn't reliably show the code and implementation at the same time
   - Should stack vertically on small screens

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Update examples copy and tldraw README

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ Welcome to the public monorepo for [tldraw](https://github.com/tldraw/tldraw). t
 
 ## Local development
 
-The local development server will run our examples app. The basic example will show any changes you've made to the codebase. To run the local development server, first clone this repo.
+The local development server will run our examples app. The basic example will show any changes you've made to the codebase.
+
+To run the local development server, first clone this repo.
 
 Install dependencies:
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Welcome to the public monorepo for [tldraw](https://github.com/tldraw/tldraw). t
 
 ## Local development
 
-To run the local development server, first clone this repo.
+The local development server will run our examples app. The basic example will show any changes you've made to the codebase. To run the local development server, first clone this repo.
 
 Install dependencies:
 
--- a/apps/docs/content/getting-started/examples.mdx
+++ b/apps/docs/content/getting-started/examples.mdx
@@ -10,6 +10,6 @@ Here's a [very simple sandbox](https://codesandbox.io/s/tldraw-example-canary-2l
 
 <Embed src="https://codesandbox.io/embed/2lrzmy?view=Editor+%2B+Preview&module=%2Fsrc%2Fapp.tsx" />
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ Welcome to the public monorepo for [tldraw](https://github.com/tldraw/tldraw). t
 
 ## Local development
 
-To run the local development server, first clone this repo.
+The local development server will run our examples app. The basic example will show any changes you've made to the codebase. To run the local development server, first clone this repo.
```
mimecuvalo, line 17: as long as you're touching this file, let's fix the broken images above please!

apps/docs/content/getting-started/examples.mdx
```diff
@@ -6,10 +6,11 @@ date: 10/11/2023
 order: 3
 ---
 
-Here's a [very simple sandbox](https://codesandbox.io/s/tldraw-example-canary-2lrzmy) showing the tldraw component.
+Below is a [simple sandbox](https://codesandbox.io/s/tldraw-example-canary-2lrzmy) showing the tldraw component.
```
mimecuvalo, line 9: Let's cut this paragraph - I think it's self-evident that it's the sandbox is below (which should go away anyway soon)

apps/docs/content/getting-started/examples.mdx
```diff
@@ -6,10 +6,11 @@ date: 10/11/2023
...
 
-<Embed src="https://codesandbox.io/embed/2lrzmy?view=Editor+%2B+Preview&module=%2Fsrc%2Fapp.tsx" />
+The [tldraw examples app](https://github.com/tldraw/tldraw/blob/main/apps/examples/src) contains a lot of helpful examples showing how to customise the UI, use the editor API, handle assets, create custom shapes/tools and much more. However, the examples app is best experienced via cloning and running the [tldraw repo](https://github.com/tldraw/tldraw) locally.
```
mimecuvalo, line 11: Grab me offline and maybe let's talk about this in person: I think we should rethink that there's copy here in general, really. We want more UI around this.

There should be a big main call-to-action - [See our examples] which is a link to https://examples.tldraw.com/

After that big CTA, there should just be an organized list:
- Framework examples: Next.js, Vite, CRA
- Examples source: source code for examples listed above
- Multiplayer implementation: adds collaboration via Yjs

And _really_ we want this list to go away and be more consolidated in general with the examples app in the end.

Again, let's talk more in person!

mimecuvalo (approved):

steveruizok (approved):

PR #2597, opened 2024-01-23
Add button to Examples to request an example

Taha-Hassan-Git:
This PR adds an issue template to request an example on the repo. And adds a button to the examples app to request an example.

I've added a blue background to the view code button to distinguish it from the request example. I think it makes the most sense as our primary button for the page.

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Add a button to request an example to the examples app


.github/ISSUE_TEMPLATE/example_request.yml
```diff
@@ -0,0 +1,33 @@
...
+assignees: []
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Need an example of how to do something with tldraw?
+  - type: textarea
+    id: description
+    attributes:
+      label: What's example?
```
mimecuvalo, line 14: typo: What's the example?

.github/ISSUE_TEMPLATE/example_request.yml
```diff
@@ -0,0 +1,33 @@
...
+      description: Describe the example.
+    validations:
+      required: true
+  - type: input
+    id: contact
```
mimecuvalo, line 19: i don't think we need contact - it'll be through the issue, right?

mimecuvalo (commented):
I think this is a good direction — I think, realistically speaking, requesting an example (or thinking of one) is a heavier lift for our team to implement (and, well, for the person to think of in the first place).

I think _above_ this button you'd want something like: [Join us on discord]
Basically, we can reduce the friction, and make it less of a dead end if you don't find what you're looking for. It also makes it less formal in most cases. I worry that Requesting an Example, if people use it, might have a high noise-to-signal — meaning that people might ask for very specific examples that are very custom to their organization or something. But I might be wrong!

(If we find people aren't using this btw, I think we probably would want to nix it after a while if it's not getting much use.)

Taha-Hassan-Git:
> I think this is a good direction — I think, realistically speaking, requesting an example (or thinking of one) is a heavier lift for our team to implement (and, well, for the person to think of in the first place).
> 
> I think _above_ this button you'd want something like: [Join us on discord] Basically, we can reduce the friction, and make it less of a dead end if you don't find what you're looking for. It also makes it less formal in most cases. I worry that Requesting an Example, if people use it, might have a high noise-to-signal — meaning that people might ask for very specific examples that are very custom to their organization or something. But I might be wrong!
> 
> (If we find people aren't using this btw, I think we probably would want to nix it after a while if it's not getting much use.)

This makes sense, do you think not having a button to request an example makes more sense?  They could request an example in the forum after joining the discord.

steveruizok:
I'd like to land this one today. Adding some quick social links to the sidebar. We also wanted to get rid of the regular landing page, right? I'll skip adding the links to that page.

PR #2621, opened 2024-01-24
arrows: separate out handle behavior from labels

mimecuvalo:
This is a followup on the arrows work.
- allow labels to go to the ends if no arrowhead is present
- avoid using / overloading TLHandle and use a new PointingLabel state to specifically address label movement
- removes the feature flag to launch this feature!

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Arrow labels: provide more polish on label placement


mimecuvalo:
@ds300 reassigned this to you for code review to stave off any boredom ;)

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -376,6 +323,41 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
...
+		const info = this.editor.getArrowInfo(shape)!
+
+		const geometry = this.editor.getShapeGeometry<Group2d>(shape)
+		const lineGeometry = geometry.children[0] as Geometry2d
+		const pointInShapeSpace = this.editor.getPointInShapeSpace(
+			shape,
+			this.editor.inputs.currentPagePoint
+		)
+		const nearestPoint = lineGeometry.nearestPoint(
+			Vec.Add(pointInShapeSpace, this._labelDragOffset)
```
ds300, line 337: this is more like feedback for the previous PR, but I'm not crazy about keeping interaction state on the util class (i.e. this `_labelDragOffset` field). It's not a pattern I've seen elsewhere, and I can imagine it breaking if people invoke these callbacks ad hoc, which I have seen happen.

I'm not saying this needs to be changed in this PR or even a follow up any time soon. I think it's a more general problem with our api design that we've talked about as part of the bindings work. Just something to keep in mind.

ds300 (approved):
This feels really nice compared to last Friday. Great work. Just one thing to add (undo/redo boundaries) and then it's good to go from my side 👍🏼

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
steveruizok: Same here! I think we could move this to the tool, right? In many interactions, the second argument is an info object with where this kind of thing would be passed into the handler.

steveruizok: You might not even need the offset here, you could also pass in the already-offset point, similar to how we pass in the handle with an offset applied.

The hard part here is that, because no other shape has a draggable label, we don't exactly know what needs to be part of a generic API (expressed through callbacks like `onLabelDrag`) and what can be part of the tool itself.

I'd suggest expanding the tool ("PointingArrowLabel" / "DraggingArrowLabel") rather than expanding the shape util APIs, especially since we only enter the "pointing handle" state if we are pointing an arrow, that's what I'd suggest. It doesn't feel like we've worked out the platonic qualities of label dragging, which is okay! Most of our APIs started with one-off features.

packages/editor/editor.css
```diff
@@ -1153,6 +1143,10 @@ input,
 	align-items: center;
 }
 
+.tl-arrow-label__inner p {
+	cursor: grab;
```
TodePond, line 1147:
```suggestion
	cursor: var(--tl-cursor-grab);
```

We can use our custom cursor here

TodePond (commented):
Looking forward to this landing!

I left a suggested change for using our custom cursor.

And I have a few UX questions (which we could also address in future PRs):

- When I 'click' an arrow label like this, I expect the arrow to become selected. Has that been considered/tried? 

2024-01-30 at 10 26 00 - Maroon Hawk

- Another thing I find a bit hard is: I want to move an arrow by grabbing and dragging its handle. I wonder if there's a way we can allow this? Might be too complicated for the user to understand? Could depend on the direction of the dragging? Not sure.

2024-01-30 at 10 41 31 - Rose Flea

- I want to be able to move a label by dragging the middle handle. I guess this is related to the above point as well - can the behaviour depend on the direction of my dragging?

2024-01-30 at 10 44 23 - Emerald Puffin

- I want to be able to reset the arrow label position back to the center. I'm not sure if there's a way of doing this? I wonder if we could make it snap similar to how we snap a curved arrow back to a straight arrow?

2024-01-30 at 10 49 30 - Azure Rhinoceros

mimecuvalo:
latest changes add:
- undo/redo
- snapping
- select shape on click of label
- fix cursor nit

thx for the feedback @TodePond and @ds300 !

still going to talk with @steveruizok about the api ideas and jam on that

steveruizok:
- Fixed a bug related to finding the nearest point on an arc
- Moved logic from shape util to the tool state
- Renamed tool state to PointingArrowLabel (kinda un-generalized the tool, we can make it generalized it later as we have other shapes with labels like this)

mimecuvalo:
@steveruizok nice! does the `getPointInArcT` change need a test - that logic seems pretty fiddly, like somebody might run into a problem there in the future 🤔

PR #2624, opened 2024-01-25
menus: address several little big things about menu styling

mimecuvalo:
- menu: fix accidental stacking of menus https://linear.app/tldraw/issue/TLD-2077/radix-menus-double-submenu which is a followup to https://github.com/tldraw/tldraw/pull/2101
   - cc @steveruizok I might be missing context here but looking at the original PR I don't see why this is needed anymore? can you show me in person?
- menu: rework checkboxes https://linear.app/tldraw/issue/TLD-2071/improve-appearance-of-menu-checkboxes
- menu: nix the text-shadow on menus
- menu: fix the kbd being off by a couple pixels

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Fixes nits on styling on our Radix menus.


MitjaBezensek (approved):
Looks good! 💅 

Tried to repro the double submenu issue but wasn't able to.

Should we maybe make the checkbox a bit darker, like similar color that the check mark has? Feels a bit weird that the two icons have different colors, almost feels like the checkbox is disabled since it's lighter.

image

mimecuvalo:
> Should we maybe make the checkbox a bit darker, like similar color that the check mark has? Feels a bit weird that the two icons have different colors, almost feels like the checkbox is disabled since it's lighter.

good call! will do

PR #2626, opened 2024-01-25
freehand: add readme for future notes

mimecuvalo:
### Change Type

### Release Notes

- Add dev notes on perfect freehand.


mimecuvalo:
sounds like this is out of date potentially!

PR #2627, opened 2024-01-25
dev: add test-dev command for easier testing of packages

mimecuvalo:
@si14 you might know a better way to wire this up! lemme know if there's something more clever here.

### Change Type

### Release Notes

- Adds easier testing command for individual packages.

### DIFF
--- a/apps/dotcom-asset-upload/package.json
+++ b/apps/dotcom-asset-upload/package.json
@@ -11,6 +11,7 @@
 	"scripts": {
 		"dev": "cross-env NODE_ENV=development wrangler dev --log-level info --persist-to tmp-assets",
 		"test": "lazy inherit --passWithNoTests",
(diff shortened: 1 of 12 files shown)


steveruizok:
@ds300 is this the true lazy way?

ds300:
I think the intention would not be to use this as a lazyrepo task, but rather as something to run while you're in a particular package directory?

Definitely running a bunch of `jest --watch` instances for the whole repo is not ideal and I don't know why anyone would want to do that anyway.

steveruizok:
So not the repo root script? I do `npx jest --watch` a lot whenever I want to kick off a test session in a package.

ds300:
yeah I assume that's what these commands are for. I think this is useful since it is definitely a question that gets asked by new folks.

PR #2634, opened 2024-01-25
Fix nudge bug

steveruizok:
This PR fixes a bug in the nudge code. The offset was previously mutated in a loop by a `Vec.Cast`.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Select some shapes
2. nudge em

- [x] Unit Tests

### Release Notes

- Fixes a bug with keyboard nudging.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5133,7 +5133,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		for (const id of ids) {
 			const shape = this.getShape(id)!
-			const localDelta = Vec.Cast(offset)
+			const localDelta = Vec.From(offset)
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5133,7 +5133,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		for (const id of ids) {
 			const shape = this.getShape(id)!
-			const localDelta = Vec.Cast(offset)
+			const localDelta = Vec.From(offset)
```
mimecuvalo, line 5136: i wonder if `Cast` should be consistent on whether it returns a new object or the same object in memory. maybe `Cast` should always return a new `A` to avoid this pitfall? not sure

PR #2663, opened 2024-01-27
[Fix] Overlapping non-adjacent handles

steveruizok:
In a previous PR, we improved logic for skipping virtual handles if a handle's neighbor is too close to it. This PR improves this logic further by testing against all other vertex handles, not just the neighbors; and skipping tests against handles that are not vertices (such as arrow text labels).

This PR also:
- removes some static cursors from `useCursors`
- adds a tiny bit of hover to text labels to differentiate them when hovered

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Create an arrow with a text label
2. Select the arrow
3. The arrow's middle handle should be selectable and in front of the text label

### Release Notes

- Fixed a bug with virtual / create handle visibility.


steveruizok:
I'm yanking a few SVG cursors here that aren't being used programmatically. These never rotate, etc. Rather than serializing these in our CSS, we should consider adding these to our assets and using them in both the editor layer AND the UI layer, where some (like grabbing) are used.

PR #2673, opened 2024-01-28
[dx] Use Biome instead of Prettier.

steveruizok:
This PR switches our formatting solution from Prettier to Biome.

| Formatter | Time |
|--------|--------|
| Prettier (first run) | 30.30s |
| Prettier (after cache) | 2.30s |
| Biome  | 1.23s | 

There are a few superficial differences in formatting but they're all whitespace and minor. In general it's the same.

# File checking

We generate a bunch of files during the build when we refresh our assets. Previously, we were using prettier to do this imperatively on the files that we made, however that no longer makes sense as they'll be separately formatted by Biome. While the generated files will be formatted later, there's a before/after check that will fail if the after is not matched to the before; and in CI, the "before" will be formatted and the "after" won't yet be formatted.

I've disabled this check for now. If it's critical, we'll have to find a better way of doing it.

# VS Code

For DX, you would want to switch VS Code's default formatter to Biome. I've added the extension as a recommendation.

### Change Type

- [x] `patch` — Formatting change


steveruizok:
Pausing on this as there's something up with biome not working correctly in the Vercel environment. (See examples log.)

si14:
TIL that mixing tabs and spaces is OK for Prettier and by extension for Biome https://github.com/biomejs/biome/issues/1661

si14:
okay, figured this out: our Vercel builder has a ridiculously old Linux version, Biome is built on a newer 

There are two ways forward on this:
- update node, apparently [newer node builder has a newer linux](https://vercel.com/guides/serverless-function-contains-invalid-runtime-error)
- rebuild Biome and vendor it, it shouldn't be too complicated

ematipico:
Hey @si14 , what's the minimum version of linux that you require?

si14:
@ematipico seems that Vercel runs node 18 build workers [with glibc 2.26](https://vercel.com/tldraw/examples/2MoCgCvQ4aF2jDeG2KtfxHFemNMs) from 2017

si14:
@ematipico you might be interested: biome builds just fine on vercel's node 20 workers!

lazy.config.ts
```diff
@@ -80,7 +80,7 @@ const config = {
 		},
 		'refresh-assets': {
 			execution: 'top-level',
-			baseCommand: `tsx <rootDir>/scripts/refresh-assets.ts`,
+			baseCommand: `tsx <rootDir>/scripts/refresh-assets.ts && biome format . --write`,
```
SomeHats, line 83: why do we need this here? I think this could affect lazy's caching as it might touch files outside of what lazy expects

si14:
Superseded by https://github.com/tldraw/tldraw/pull/2776 saga

PR #2683, opened 2024-01-29
make CI check for yarn install warnings and fix the peer deps ones we have

si14:

### DIFF
--- a/apps/examples/package.json
+++ b/apps/examples/package.json
@@ -33,6 +33,7 @@
 		"e2e-ui": "playwright test --ui -c ./e2e/playwright.config.ts"
 	},
 	"dependencies": {
+		"@babel/core": "^7.23.7",
 		"@babel/plugin-proposal-decorators": "^7.21.0",
 		"@playwright/test": "^1.38.1",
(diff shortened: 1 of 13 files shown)


si14:
@MitjaBezensek this touches vscode extension, let's chat tomorrow to make sure I got it right (or if I didn't I want to understand those deps better)

apps/examples/package.json
```diff
@@ -33,6 +33,7 @@
 		"e2e-ui": "playwright test --ui -c ./e2e/playwright.config.ts"
 	},
 	"dependencies": {
+		"@babel/core": "^7.23.7",
 		"@babel/plugin-proposal-decorators": "^7.21.0",
```
ds300, line 37: I'm 99% sure we can get rid of these. I don't believe we use babel anywhere now.
```suggestion
```

ds300 (approved):

apps/examples/package.json
si14: yeah, I wasn't sure it's not used, thanks for confirming it!

PR #2685, opened 2024-01-30
Allow users to set document name and use it for exporting / saving

MitjaBezensek:
Adds the ability to change document names in the top center part of the UI. This mostly brings back the functionality we already had in the past.

This is basically a port of what @SomeHats did a while back. I changed the dropdown options and removed some of the things (we are not dealing with network requests directly so some of that logic did not apply any longer). We did have autosave back then, not sure if we want to bring that back?

Changes the `exportAs` api, thus braking.

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Top center should now show a new UI element. It has a dropdown with a few actions.
2. Double clicking the name should also start editing it.
3. The name should also be respected when exporting things. Not if you select some shapes or a frame. In that case we still use the old names. But if you don't have anything selected and then export / save a project it should have the document name.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Allow users to name their documents.


ds300:
some quick UI feedback

- the focused input outline feels too strong to me, was it always like that? 
  image

- Seems like the horizontal padding is inconsistent (bigger on the left)
- The minimum input size feels too small to me.
- The input size jumps around as you toggle between empty and filled
  Kapture 2024-01-30 at 14 02 37

ds300:
Another thought: might be nice if 'fork this project' altered the title like copying a file e.g. appending `(1)`? I think we have the logic for this somewhere already for pages

packages/tlschema/src/records/TLDocument.ts
```diff
@@ -2,6 +2,9 @@ import { BaseRecord, createRecordType, defineMigrations, RecordId } from '@tldra
 import { JsonObject } from '@tldraw/utils'
 import { T } from '@tldraw/validate'
 
+/** @public */
+export const DEFAULT_DOCUMENT_NAME = 'Untitled'
+
```
ds300, line 7: Not sure we should have this.
I think in the places where we render/use the document name we should probably be using a localized version of 'Untitiled' if the .name property is the empty string.

MitjaBezensek:
Good suggestions! Yeah, there's some more ux polish to do. I ported all the existing styles, but we might have had a different "reset" when we had this rename previously.

ds300 (commented):
Looking good! I forgot how complicated that document name field is 😮

packages/tldraw/src/lib/utils/export/exportAs.ts
```diff
@@ -18,15 +19,19 @@ export async function exportAs(
 	editor: Editor,
 	ids: TLShapeId[],
 	format: TLExportType = 'png',
+	name: string | undefined,
 	opts = {} as Partial<TLSvgOptions>
```
ds300, line 23: (general observation, not feedback on this PR) if we standardised on the single-object style function args we would be able to make changes like this without breakage. I wish I had pushed for that more often.

packages/tlschema/src/records/TLDocument.ts
MitjaBezensek: I'll take a look. Think the problem was that our current translations always expect us to get translations via hooks. But I think in one of the cases that wasn't really feasible here.

That said we can definitely remove this default here and always check if it's an empty string where we show it.

packages/tldraw/src/lib/utils/export/exportAs.ts
MitjaBezensek: Clojure all the way 😄 

It's also confusing that we have opts, but then on top of that we also have a bunch of args 🤷

ds300: it's not too late. There is a migration path for functions/methods with multiple args to a single arg, using overloads. something like this (rough sketch)

```ts
function myFunction(the: number, old: number, params: number): Blah
function myFunction(args: {the: number, old: number, params: number, andTheNewOne?: number}): Blah
function myFunction() {
  let the, old, params, andTheNewOne
  if (arguments.length === 1) {
     ;({the, old, params, andTheNewOne}) = arguments[0]
  } else {
     the = arguments[0]
     old = arguments[1]
     params = arguments[2]
     andTheNewOne = undefined
  }

  ....
}
```

ds300: at the same time you could add a console.warn for the old signature to make people switch over, and then eventually delete the old version.

steveruizok:
There are some style changes that we need to land here:
- the title should be centered on the board; we had some complicated logic around this in pro but it's important so that the tile is not offset by the uneven left and right content
- the title input should have a placeholder and a min width that fits the placeholder
- the title input should not have a border, just an outline

Kapture 2024-02-02 at 09 05 33

apps/dotcom/src/utils/useFileSystem.tsx
```diff
@@ -128,7 +133,12 @@ export function getSaveFileCopyAction(
 		kbd: '$s',
 		async onSelect(source) {
 			handleUiEvent('save-project-to-file', { source })
-			const defaultName = saveFileNames.get(editor.store) || `Untitled${TLDRAW_FILE_EXTENSION}`
+			const documentName =
+				editor.getDocumentSettings().name === ''
```
mimecuvalo, line 137: q: is `editor.getDocumentSettings().name` always available? should this check for `!editor.getDocumentSettings().name`?

apps/dotcom/styles/z-board.css
```diff
@@ -317,3 +317,70 @@
 .tlui-layout[data-breakpoint='0'] .tlui-offline-indicator {
 	margin-top: 4px;
 }
+
+/* Document Name */
+
+.tlui-document-name__inner {
+	border-radius: calc(var(--radius-2) + var(--space-2));
+	background-color: var(--color-background);
+	padding: 0px 0px 0px 0px;
```
mimecuvalo, line 326: nit: padding: 0

apps/dotcom/styles/z-board.css
```diff
@@ -317,3 +317,70 @@
...
+
+.tlui-document-name__input,
+.tlui-document-name__text {
+	padding: var(--space-2) var(--space-3);
+	white-space: pre;
+	line-height: 24px;
+}
+
+.tlui-document-name__input {
+	position: absolute;
```
mimecuvalo, line 352: styling thing: feels like the input field should have a min-width of some sort. when typing in the first time it starts out as 60px but then goes to the single width of the character which is slightly jarring

packages/tldraw/src/lib/ui/hooks/useActions.tsx
```diff
@@ -71,6 +72,17 @@ function makeActions(actions: TLUiActionItem[]) {
 	return Object.fromEntries(actions.map((action) => [action.id, action])) as TLUiActionsContextType
 }
 
+function getExportName(editor: Editor, defaultName: string) {
+	const selectedShapes = editor.getSelectedShapes()
+	// When we don't have any shapes selected, we want to use the document name
+	if (selectedShapes.length === 0) {
+		const documentName = editor.getDocumentSettings().name
+		if (documentName === '') return defaultName
+		documentName
```
mimecuvalo, line 81: `return documentName` :)

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+export const DocumentTopZone = track(function DocumentTopZone({
+	isOffline,
+}: {
+	isOffline: boolean
+}) {
+	const isDocumentNameVisible = useBreakpoint() >= 4
+
+	return (
+		<DocumentTopZoneContainer>
```
mimecuvalo, line 43: philosophical q: why have this be separated from the `tlui-menu-zone`? why not have it next to the Page menu area?

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+	const actions = useActions()
+	const forkAction = actions[FORK_PROJECT_ACTION]
+	const saveFileAction = actions[SAVE_FILE_COPY_ACTION]
+	const editor = useEditor()
+	const msg = useTranslation()
+
+	return (
+		<div className="tlui-document-name__inner">
+			<DocumentNameEditor state={state} setState={setState} />
+			<DropdownMenu.Root id="document name">
```
mimecuvalo, line 61: `document name` might be a typo? did you mean to add a '-'? or, is this needed anyway?

apps/dotcom/src/components/DocumentName/DocumentName.tsx
mimecuvalo: it might end up making this a bit simpler b/c then it wouldn't duplicate the other actions (save & fork)

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+				>
+					{addRealSpaceForWhitespace(name) || ' '}
+				</div>
+			)}
+		</div>
+	)
+})
+
+function addRealSpaceForWhitespace(string: string) {
+	if (string === '') string = ' '
```
mimecuvalo, line 293: `!string` ?
a comment here might be useful too - hard to say what this is trying to address. also, it throws me off that above the logic is `addRealSpaceForWhitespace(name) || ' '` (using a regular space, i think for the `|| ' '`) 🤔

mimecuvalo: well, probably the `|| ' '` isn't necessary anyway ¯\_(ツ)_/¯

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+			const trimmed = state.name.trim()
+			if (!trimmed || trimmed === documentSettings.name.trim()) {
+				if (!state.isEditing) setState((prev) => ({ ...prev, name: null }))
+				return
+			}
+
+			editor.updateDocumentSettings({ name: trimmed })
+		}
+
+		if (state.isEditing === false) {
```
mimecuvalo, line 221: nit: just `!state.isEditing` since it's a bool

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+	return (
+		<div ref={ref} className="tlui-top-zone__container">
+			{children}
+		</div>
+	)
+}
+
+function getName(state: NameState, document: TLDocument, defaultName: string) {
+	if (state.name) return state.name
```
mimecuvalo, line 183: fwiw, this could be `state.name || document.name || defaultName`

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+		const layoutTop = element.parentElement!.parentElement!
+		const leftPanel = layoutTop.querySelector('.tlui-layout__top__left')! as HTMLElement
+		const rightPanel = layoutTop.querySelector('.tlui-layout__top__right')! as HTMLElement
+
+		const totalWidth = layoutTop.offsetWidth
+		const leftWidth = leftPanel.offsetWidth
+		const rightWidth = rightPanel.offsetWidth
+		// ignore the width of the button:
+		const selfWidth = element.offsetWidth - 44
```
mimecuvalo, line 128: the 44, 12, and 16 numbers are a bit magical - can you name this as const's that might help explain what they're accounting for?

apps/dotcom/src/utils/useFileSystem.tsx
MitjaBezensek: It should be.  The signature of `getDocumentSettings` says it always returns `TLDocument` and that one always has a `name` prop.

packages/tldraw/src/lib/ui/hooks/useActions.tsx
MitjaBezensek: Nice catch. I guess my mind is still in Clojure where everything is an expression 😂

apps/dotcom/src/components/DocumentName/DocumentName.tsx
MitjaBezensek: Think I mostly just replaced `board` with `document` from the original [PR that added board titles in tldraw pro](https://github.com/tldraw/brivate/pull/2972/files#diff-d825ed9304ac39ce1e768f560737448b1af61616404664d42547ed631cd397fcR54). Seems like we used a space delimited name there as well. I'm fine with both 🤷

apps/dotcom/src/components/DocumentName/DocumentName.tsx
MitjaBezensek: Here I also just [copied the existing code](https://github.com/tldraw/brivate/pull/2972/files#diff-d825ed9304ac39ce1e768f560737448b1af61616404664d42547ed631cd397fcR304-R307), but happy to add some comment.

apps/dotcom/src/components/DocumentName/DocumentName.tsx
mimecuvalo: the space weirds me out heh - feels like it shouldn't work :P

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,287 @@
...
+		const selfWidth = element.offsetWidth - BUTTON_WIDTH
+
+		let xCoordIfCentered = (totalWidth - selfWidth) / 2
+
+		// Prevent subpixel bullsh
+		if (totalWidth % 2 !== 0) {
+			xCoordIfCentered -= 0.5
+		}
+
+		const xCoordIfLeftAligned = leftWidth + 12
```
mimecuvalo, line 137: can you make this 12 a const too? 🙏

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+		const left = element.offsetLeft
+		const xCoord = Math.max(xCoordIfCentered, xCoordIfLeftAligned) - left
+		const maxWidth = Math.min(totalWidth - rightWidth - leftWidth - 16, MAX_TITLE_WIDTH_PX)
```
mimecuvalo, line 141: and this 16, too :)

apps/dotcom/src/components/DocumentName/DocumentName.tsx
MitjaBezensek: Looking at this again it seems like the `addRealSpaceForWhitespace(name) || ' '` it was probably meant as a guard so that we wouldn't show `undefined / null` in the ui. With use having a `defaultName` fallback now we can probably remove this.

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+	const name = state.name || documentSettings.name || defaultDocumentName
+
+	return (
+		<div className="tlui-document-name__input__wrapper">
+			{state.isEditing && (
+				<input
+					ref={inputRef}
+					className="tlui-document-name__input"
+					value={state.name ?? documentSettings.name}
```
mimecuvalo, line 255: you're comparing to `state.name` to `null` in other places so i think this logic here won't work how you want it too. prbly you want `||` instead of `??`

apps/dotcom/styles/z-board.css
```diff
@@ -317,3 +317,71 @@
...
+	display: flex;
+	flex: auto;
+}
+
+.tlui-document-name__input,
+.tlui-document-name__text {
+	padding: var(--space-2) var(--space-3);
+	white-space: pre;
+	line-height: 24px;
+	min-width: 60px;
```
mimecuvalo, line 349: tiniest nit: for "Untitled" to fit it should be `62px` - thx!

mimecuvalo (approved):
LGTM pending a couple new comments.

there's also maybe a bug, not sure if intended: once you add a title, you can't remove it to a blank title? i think that _might_ be ok but it feels strange. prbly you want to allow blank titles

apps/dotcom/src/components/DocumentName/DocumentName.tsx
MitjaBezensek: Looks like this one is correct, with an `||` it would not allow us to clear the input:

https://github.com/tldraw/tldraw/assets/2523721/1b5fe3e8-93ed-4114-81fa-8dc0f30942e6

PR #2688, opened 2024-01-30
Don't require wrapping the tldraw component

TodePond:
This PR simplifies the default behaviour of the `Tldraw` component.

- Changed the default size of the component to 720x480. Previously, it grew to fill its parent.
- Added a new `TldrawWrapper` component for use inside the `Tldraw` component. It seems to be important for performance in Safari. It can also be used by library consumers, as shown in the 'exploded' example.
- Added a `style` prop to the `Tldraw` component. You can use it to override the default dimensions (or other styles).
- Added better support for the `className` prop on the `Tldraw` component. It gets applies to `TldrawWrapper`.

I made some other changes to reflect the new behaviour.
- Updated our examples. Replaced the scrolling example with a fullscreen example.
- Updated our guides.
- Updated our tests.
- Changed the layout of the examples iframe on the docs site - to match the new default dimensions. Future work: Change the iframe to a 'native' `Tldraw` component.
- Changed the default development mode example from basic to fullscreen, so that our development experience doesn't change.

Comparison and pros and cons below! See code comments for more detail

## Comparison

image

image

image

image

image

image

## Pros

- It reduces the number of concepts we need to explain in our getting started guide.
- It makes our examples more concise.
- Placing down a single `Tldraw` component makes something appear now.
- You don't need to manually wrap each `Tldraw` component in a wrapper anymore.
- We can now ensure that the `Tldraw` component always uses a wrapper, which prevents a Safari performance issue. Previously, we relied on the library consumer doing the right thing.
- I think these changes make the `Tldraw` component behave in a more expected way. It works more like a regular canvas element, or an image.
- I think these changes help to highlight that "tldraw is just react". You don't need to change how you use it.

## Cons

- This is a big breaking change. Most people will need to change some code. Are we ready to communicate this kind of change?
- It might be annoying to have to override the default height and width, if you just want to use tldraw in fullscreen.
- If you set the component's inset to 0, it might not behave as expected. You also need to set height and width to auto.
- Our examples app looks weird now for most examples. On the docs site, we probably want our examples to have the default dimensions. But in the examples app, we probably want them to be fullscreen. How should we approach this?

### Change Type

### Test Plan

1. 

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- SDK: You don't need to wrap the tldraw component anymore


apps/docs/components/Header.tsx
```diff
@@ -70,7 +70,7 @@ export function SectionLinks({ sectionId }: { sectionId?: string | null }) {
 				Reference
 			</a>
 			<a
-				href="/examples/basic/develop"
+				href="/examples/basic/basic"
```
TodePond, line 73: renamed this because we probably want to use the fullscreen example for development now

apps/examples/src/examples/exploded/ExplodedExample.tsx
```diff
@@ -29,7 +30,7 @@ const defaultComponents = {
 //[2]
 export default function ExplodedExample() {
 	return (
-		<div className="tldraw__editor">
+		<TldrawWrapper style={{ width: '100%', height: '100%' }}>
```
TodePond, line 33: I used a style prop here to indicate to the reader that styles go on the wrapper component, not the editor component.

apps/examples/src/examples/floaty-window/FloatyExample.tsx
```diff
@@ -4,11 +4,9 @@ import { useEffect } from 'react'
...
 export default function FloatyExample() {
 	return (
-		<div className="tldraw__editor">
-			<Tldraw persistenceKey="tldraw_floaty_example">
-				<SneakyFloatyHook />
-			</Tldraw>
-		</div>
+		<Tldraw persistenceKey="tldraw_floaty_example" style={{ width: '100%', height: '100%' }}>
+			<SneakyFloatyHook />
+		</Tldraw>
```
TodePond, line 9: I made this example fullscreen because it requires it

apps/examples/src/examples/only-editor/OnlyEditor.tsx
```diff
@@ -13,29 +13,27 @@ const myShapeUtils = [MiniBoxShapeUtil]
...
+							type: 'box',
+							x: 100,
+							y: 100,
+						},
+					])
+			}}
+			components={{
+				Background: BackgroundComponent,
+			}}
+		/>
```
TodePond, line 36: I'm letting this one go unwrapped. The editor component fills its parent. Is that the right thing to do? Could wrap it still 🤷

apps/examples/src/examples/persistence/PersistenceExample.tsx
```diff
@@ -50,26 +50,22 @@ export default function PersistenceExample() {
...
 
 	if (loadingState.status === 'error') {
 		return (
-			<div className="tldraw__editor">
+			<div style={{ padding: 32 }}>
 				<h2>Error!</h2>
 				<p>{loadingState.error}</p>
 			</div>
 		)
 	}
```
TodePond, line 66: scout cleaning: These loading and error states looked dodgy as they were, so I replaced the classname with some super simple padding.

apps/examples/src/index.tsx
```diff
@@ -43,7 +44,11 @@ const router = createBrowserRouter([
 	},
 	{
 		path: 'end-to-end',
-		element: <EndToEnd />,
+		lazy: async () => ({ element: <EndToEnd /> }),
+	},
+	{
+		path: 'hello-world',
+		lazy: async () => ({ element: <HelloWorld /> }),
```
TodePond, line 51: I've added another 'hidden' example, just for our own use at `/hello-world` in the examples app. It was useful to check that my changes still worked in a completely vanilla setup. Our examples app now has some more layout going on, which meant it was hard to fully debug a 'hello world' scenario.

apps/examples/src/index.tsx
```diff
@@ -43,7 +44,11 @@ const router = createBrowserRouter([
 	},
 	{
 		path: 'end-to-end',
-		element: <EndToEnd />,
+		lazy: async () => ({ element: <EndToEnd /> }),
```
TodePond, line 47: fix: this wasn't doing lazy loading like the other examples. this was causing some styling issues

apps/examples/src/index.tsx
```diff
@@ -31,10 +32,10 @@ const router = createBrowserRouter([
 	{
 		path: '/',
 		lazy: async () => {
-			const Component = await basicExample.loadComponent()
+			const Component = await fullscreenExample.loadComponent()
```
TodePond, line 35: default to the fullscreen example for development

packages/editor/editor.css
```diff
@@ -1,8 +1,6 @@
 /* @tldraw/editor */
 
 .tl-container {
-	width: 100%;
-	height: 100%;
```
TodePond, line 5: tidy: these are duplicated lower down

packages/tldraw/src/lib/ui.css
```diff
@@ -1,5 +1,10 @@
 /* @tldraw/ui */
 
+:where(.tl-wrapper) {
+	height: 480px;
+	width: 720px;
+}
+
```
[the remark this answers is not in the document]
mimecuvalo: this is definitely worth a comment in the code for future intrepid explorers :)

mimecuvalo:
@TodePond have you all discussed perhaps having a `fullscreen` boolean prop on the `<Tldraw />` component? i think that might encapsulate nicely what some people are looking for. i can't say whether it should be true by default, not sure...

packages/tldraw/src/lib/ui.css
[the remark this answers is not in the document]
TodePond: good idea! done!

TodePond:
@mimecuvalo
> have you all discussed perhaps having a `fullscreen` boolean prop on the `<Tldraw />` component?

I've thought about it but we haven't discussed it yet. Not sure what the write name for that prop should be, as it's not really "fullscreen" - it's more like "fills parent". It may or may not be fullscreen depending on the parent.

Some moodboarding...

```tsx
<Tldraw inline />
<Tldraw inline={false} />
```

And here are some suggestions from chatgpt:
- `fullSize`
- `fillContainer`
- `expandToParent`
- `parentFullSize`
- `useParentDimensions`
- `full`
- `stretch`
- `autoSize`
- `fit`
- `expand`

Taha-Hassan-Git:
> * Our examples app looks weird now for most examples. On the docs site, we probably want our examples to have the default dimensions. But in the examples app, we probably want them to be fullscreen. How should we approach this?

Coming from writing the quick start guide, this feels like a step in the right direction. It's closer to how you expect a React component to behave out of the box, and avoids having to explain a couple of fiddly steps. I don't have much to say about the other cons, but I think the extra space in the examples app is actually a great opportunity to surface more context and information around the examples.

SomeHats:
Could we preserve the existing "fill space by default" approach whilst solving the 0-width, 0-height issue and not over-complicating the css with extra properties to override by using [`minmax`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax)? ie
```ts
:where(.tl-wrapper) {
	height: minmax(480px, 100%);
	width: minmax(720px, 100%);
}
```

TodePond:
> Could we preserve the existing "fill space by default" approach whilst solving the 0-width, 0-height issue and not over-complicating the css with extra properties to override by using [`minmax`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax)?

I don't think `minmax` can be used in height and width? MDN says it's just for grids.

image

SomeHats:
> I don't think `minmax` can be used in height and width? MDN says it's just for grids.

This is true! from steve, maybe just the normal `min` function would be better here: https://developer.mozilla.org/en-US/docs/Web/CSS/min

TodePond:
We decided to not do this. Considerations were:

**Reasons against a default height and width:**

- Even if you're using the component inline, you still probably don't want our arbitrarily decided dimensions. So it doesn't help much.

**Reasons against an unwrapped component:**
- "Not needing to wrap the component" is not a valuable-enough improvement.
- The passed down `style` prop can be deceptive. It doesn't set the styles of *everything* in the component, just the top-level.

**A couple of things I've noticed:**
- We do have a `className` prop on the component (and the editor component). I think it has the same downsides as the `style` prop. It think we should consider removing or deprecating it, so that we fully commit to the *wrapped tldraw* approach.
- In our examples app, we add extra properties like `overscroll-behavior: none`, `touch-action: none` and `overflow-hidden`. I guess that's why we style the wrapper in a separate css file - it would become cluttered if we did all of these inline. Are these needed in all cases? (ie: fullscreen as well as inline) Added a linear ticket to test this.

PR #2707, opened 2024-02-02
fix: defineEnum and literalEnum typescript

thatsjonsense:
When trying to use tldraw in my app, I get a bunch of Typesccript errors coming out of schema and validation

image

These all stem from the same pattern of using "const" when defining a generic type, which I don't think is valid Typescript? Please correct me if I'm missing something here, but it looks like a typo

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan
There is no functionality change here, just types fix

### Release Notes

- Fix enum types

### DIFF
--- a/packages/tlschema/src/styles/StyleProp.ts
+++ b/packages/tlschema/src/styles/StyleProp.ts
@@ -72,7 +72,7 @@ export class StyleProp<Type> implements T.Validatable<Type> {
 	 * })
 	 * ```
 	 */
-	static defineEnum<const Values extends readonly unknown[]>(
+	static defineEnum<Values extends readonly unknown[]>(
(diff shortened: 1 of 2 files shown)


thatsjonsense:

thatsjonsense:
Got it, thanks for the context @SomeHats . TS5 upgrade is not a possibility for us yet because some other packages we rely on have their own conflicts with it, so it's a catch-22. I'll look into patch-package, but it might help for your adoption to hold off on TS5-only features until it's been more widely adopted

steveruizok:
Closing this one, hopefully folks can find this if they run into a bug with `const` TypeScript types.

PR #2709, opened 2024-02-02
E2e tests for the toolbar

Taha-Hassan-Git:
This PR adds some e2e tests for the toolbar.

Fixtures have been set up for the toolbar and style panel, and are fairly barebones at the moment. Eventually each menu should have a fixture associated with it, and all tests will use the class defined in the fixtures file. 

### Change Type

### Release Notes

- Add e2e tests for the toolbar


apps/examples/e2e/tests/test-shapes.spec.ts
```diff
@@ -73,17 +73,17 @@ test.describe('Shape Tools', () => {
 		for (const { tool } of otherTools) {
 			// Find and click the button
 			if (!(await page.getByTestId(`tools.${tool}`).isVisible())) {
-				if (!(await page.getByTestId(`tools.more`).isVisible())) {
+				if (!(await page.getByTestId('tools.more.button').isVisible())) {
```
MitjaBezensek, line 76: We usually limited the nesting to two, kind of like we do for translations. Maybe `tools.more-button`? This locator also seems to be used a lot, maybe we can extract it into a variable?

apps/examples/e2e/tests/test-ui.spec.ts
```diff
@@ -1,10 +1,109 @@
...
 import { setup } from '../shared-e2e'
 
-test.describe('ui', () => {
+test.describe('mobile ui', () => {
 	test.beforeEach(setup)
-
-	test('mobile style panel opens and closes when tapped', async ({ isMobile }) => {
+	test('mobile style panel opens and closes when tapped', async ({ isMobile, page }) => {
+		test.skip(!isMobile, 'only run on mobile')
+		expect(await page.isVisible('.tlui-style-panel')).toBe(false)
```
MitjaBezensek, line 8: I'd extract most of these hard coded strings into variables / consts. Unless we only use it once it will make maintaining this code much easier as it will mean we only have to rename it once. It also makes the code a bit more nicer to read (`isVisible(StylePanel)` feels a bit better than `isVisible(.tlui-style-panel)`)

I think it's also worthwile to avoid using css selectors in most cases. It's not what user sees. So in most cases I'd use things like text selectors, role selectors, and if none of those are available I'd probably go for a data test id.

MitjaBezensek: Could also change this to:
```
await expect(page.getByTestId(StylePanel)).toBeVisible()
```

These kinds of assertions feel better to read to me.

apps/examples/e2e/tests/test-ui.spec.ts
```diff
@@ -1,10 +1,109 @@
...
+		await page.getByTestId('mobile-styles.button').click()
+		expect(await page.isVisible('.tlui-style-panel')).toBe(true)
+		await page.getByTestId('mobile-styles.button').click()
+		expect(await page.isVisible('.tlui-style-panel')).toBe(false)
+	})
+	test('mobile style panel closes when tapping off the menu', async ({ isMobile, page }) => {
 		test.skip(!isMobile, 'only run on mobile')
+		await page.getByTestId('mobile-styles.button').click()
+		expect(await page.isVisible('.tlui-style-panel')).toBe(true)
+		await page.getByTestId('tools.select').click()
```
MitjaBezensek, line 18: I'd also extract all the tools locators. Maybe we can use something like
```
const TOOLS = {
    select: 'tools.select',
    draw: `tools.draw,` 
} as const
```

We can then use them in tests like so
```
const selectTool = page.getByTestId(TOOLS.select)

```

apps/examples/e2e/tests/test-ui.spec.ts
```diff
@@ -1,10 +1,109 @@
...
+
+test.describe('when selecting a tool from the toolbar', () => {
+	test.beforeEach(setup)
+	test('selecting a tool changes the button color', async ({ page }) => {
+		const selectTool = page.getByTestId('tools.select')
+		const drawTool = page.getByTestId('tools.draw')
+		await selectTool.click()
+		// We can't check the style of pseudo elements, as the browser doesn't expose them to the test
+		// So we use text color as a proxy
+		expect(selectTool).toHaveCSS('color', 'rgb(255, 255, 255)')
```
MitjaBezensek, line 38: Maybe we can introduce helper functions like `toBeSelected` which do this assertion? Would make it more obvious what the exact intention is.

apps/examples/e2e/tests/test-ui.spec.ts
```diff
@@ -1,10 +1,109 @@
-import test from '@playwright/test'
+import test, { expect } from '@playwright/test'
 import { setup } from '../shared-e2e'
 
-test.describe('ui', () => {
+test.describe('mobile ui', () => {
 	test.beforeEach(setup)
-
-	test('mobile style panel opens and closes when tapped', async ({ isMobile }) => {
+	test('mobile style panel opens and closes when tapped', async ({ isMobile, page }) => {
```
MitjaBezensek, line 6: We might not needto repeat mobile here as the parent descript already mentions it?

apps/examples/e2e/tests/test-ui.spec.ts
MitjaBezensek, line 1: Maybe we'd want to change the name of the file. `ui` feels so generic? We could maybe pull the mobile tests into `mobile-style-panel` file and the rest seem more like `toolbar` tests?

MitjaBezensek (commented):
I think that test are great, just left a few comments that deal more with legibility and maintainability of the tests.

I really like e2e tests that read like talking to a person so I like to pull out selectors into more human friendly names and might also introduce functions with clear names to name what I'm trying to check (like `toolIsSelected` vs `this locator has some color applied to it`)

MitjaBezensek:
> Is there a way to avoid doing the set up for each test, and have it all be part of the same flow? For example, I want to check if the tool is highlighted correctly, and I also want to make sure the correct styles are exposed. I could check both in one test to avoid the setup, but then it wouldn't be clear which one had failed. So instead I have to do the set up all over again.

I don't think running setup multiple times is a problem. Sure, if the test start be slow, we might start looking into optimizing things. But until then I wouldn't worry about it.

You can group those two together under one umbrella test like "make sure switching tools work correctly" and then the highlight part and the styles part fall under that. You would see which of the parts failed since that assert would fail. You could also combine all the things that each of these two parts do into two steps by using `test.step`, which would make it even more obvious which of the two steps failed.

apps/examples/e2e/tests/test-mobile-style-panel.spec.ts
```diff
@@ -0,0 +1,24 @@
+import test, { expect } from '@playwright/test'
+import { setup } from '../shared-e2e'
+
+test.describe('mobile ui', () => {
+	test.beforeEach(setup)
+	test('style panel opens and closes as expected', async ({ isMobile, page }) => {
+		test.skip(!isMobile, 'only run on mobile')
+		const mobileStylesButton = page.getByTestId('mobile-styles.button')
```
MitjaBezensek, line 8: I guess we could get this from the toolbar?

apps/examples/e2e/tests/fixtures/menus/toolbar.ts
```diff
@@ -0,0 +1,36 @@
...
+	readonly toolNames: { [key: string]: Locator }
+	readonly popoverToolNames: { [key: string]: Locator }
+
+	constructor(public readonly page: Page) {
+		this.page = page
+		this.toolLock = this.page.locator('[data-testid="tool-lock"]')
+		this.moreToolsButton = this.page.locator('[data-testid="tools.more-button"]')
+		this.moreToolsPopover = this.page.locator('[data-testid="tools.more-content"]')
+		this.mobileStylesButton = this.page.locator('[data-testid="mobile-styles.button"]')
+		this.toolNames = {
```
MitjaBezensek, line 17: Maybe just call these `tools` instead of `toolNames`?

apps/examples/e2e/tests/fixtures/menus/stylepanel.ts
```diff
@@ -0,0 +1,18 @@
+import { Page } from '@playwright/test'
+
+export class StylePanel {
+	readonly stylesArray: string[]
+
+	constructor(public readonly page: Page) {
+		this.page = page
+		this.stylesArray = [
```
MitjaBezensek, line 8: It would be best if we could get this straight from code. Then the code and the tests could not diverge. But it doesn't seem like we keep this organized in some sort of an array. So it's going to be hard to know that we have to update our tests if we add an additional style option to the styles panel. But I guess it's fine for now.

apps/examples/e2e/tests/test-mobile-style-panel.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+		const stylePanel = page.getByTestId('style.panel')
```
MitjaBezensek, line 9: How about we introduce `getElement` on the `stylePanel` fixture (and `toolbar` if needed), which just returns `page.getByTestId('style.panel')`?

apps/examples/e2e/tests/test-mobile-style-panel.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+		await expect(stylePanel).toBeHidden()
+		await mobileStylesButton.click()
+		await expect(stylePanel).toBeVisible()
+		// clicking off the style panel should close it
+		page.mouse.click(200, 200)
+		await expect(stylePanel).toBeHidden()
+	})
+	test('style menu button is disabled for the eraser tool', async ({ isMobile, page }) => {
+		test.skip(!isMobile, 'only run on mobile')
+		const eraserTool = page.getByTestId('tools.eraser')
```
MitjaBezensek, line 19: Should this also be part of `toolbar`?

apps/examples/e2e/tests/test-shapes.spec.ts
```diff
@@ -73,17 +73,17 @@ test.describe('Shape Tools', () => {
 		for (const { tool } of otherTools) {
 			// Find and click the button
 			if (!(await page.getByTestId(`tools.${tool}`).isVisible())) {
-				if (!(await page.getByTestId(`tools.more`).isVisible())) {
+				if (!(await page.getByTestId('tools.more-button').isVisible())) {
```
MitjaBezensek, line 76: Use `toolbar` fixture to get this?

apps/examples/e2e/tests/test-toolbar.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+		await test.step('selecting a tool changes the button color', async () => {
+			await select.click()
+			await toolbar.isSelected(select, true)
+			await toolbar.isSelected(draw, false)
+			await draw.click()
+			await toolbar.isSelected(select, false)
+			await toolbar.isSelected(draw, true)
+		})
+
+		await test.step('selecting certain tools exposes the tool-lock button', async () => {
```
MitjaBezensek, line 21: Great test, we broke this a couple of times in the past!

Taha-Hassan-Git:
Had to get the shape tests to use the page from the fixtures, which meant changing the setup for those tests. They now work, but for some reason that seemed to break the text-measurement tests, I can't figure out why.

MitjaBezensek:
> Had to get the shape tests to use the page from the fixtures, which meant changing the setup for those tests. They now work, but for some reason that seemed to break the text-measurement tests, I can't figure out why.

Looks like they passed now, could just be flaky?

apps/examples/e2e/tests/fixtures/menus/toolbar.ts
```diff
@@ -0,0 +1,37 @@
...
+		this.moreToolsPopover = this.page.locator('[data-testid="tools.more-content"]')
+		this.mobileStylesButton = this.page.locator('[data-testid="mobile-styles.button"]')
+		this.tools = {
+			select: this.page.locator('[data-testid="tools.select"]'),
+			draw: this.page.locator('[data-testid="tools.draw"]'),
+			arrow: this.page.locator('[data-testid="tools.arrow"]'),
+			cloud: this.page.locator('[data-testid="tools.cloud"]'),
+			eraser: this.page.locator('[data-testid="tools.eraser"]'),
+		}
+		this.popoverToolNames = {
```
MitjaBezensek, line 24: Could maybe also rename this to `popoverTools`, just as we did with `toolNames` -> `tools`.

apps/examples/e2e/tests/fixtures/menus/stylepanel.ts
MitjaBezensek, line 1: Maybe rename this file to `StylePanel` and `toolbar` to `Toolbar`?

apps/examples/e2e/tests/test-shapes.spec.ts
```diff
@@ -111,9 +106,9 @@ test.describe('Shape Tools', () => {
 		for (const { tool, shape } of clickableShapeCreators) {
 			// Find and click the button
 			if (!(await page.getByTestId(`tools.${tool}`).isVisible())) {
-				await page.getByTestId('tools.more').click()
+				await page.getByTestId('tools.more-button').click()
```
MitjaBezensek, line 109: Maybe use `toolbar.moreToolsButton` here as well (there's a few more further down the file).

MitjaBezensek (approved):
API feels quite good now, love how readable the tests are. Left a few small comments, but feels ready to merge after those.

steveruizok:
Heads up, this branch may have a bunch of conflicts with #2796.

PR #2712, opened 2024-02-03
[dotcom] `TLSyncRoom` tidy

steveruizok:
I went through the sync room code today while thinking about syncing. As I worked through it, I added some comments etc for my own readability. There is no change in how the code works.

There are a few functions declared in the body of `handlePushRequest`. We might want to re-arrange things so that we're not re-declaring those on each push (there are a LOT of pushes, basically every mouse move from every client). I did try that but reverted that change; it feels like a proper solution would be bigger than what I wanted to do here.

### Change Type

### DIFF
--- a/packages/tlsync/src/lib/TLSyncRoom.ts
+++ b/packages/tlsync/src/lib/TLSyncRoom.ts
@@ -333,11 +333,11 @@ export class TLSyncRoom<R extends UnknownRecord> {
 		})
 	}
 
-	private getDocument(id: string) {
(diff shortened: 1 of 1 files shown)


ds300:
> We might want to re-arrange things so that we're not re-declaring those on each push

I wouldn't worry about this. It only happens once per request and the cost of handling json (de)serialization alone would massively dwarf the cost of allocating a couple of closures, even if v8 isn't able to optimize them away.

PR #2722, opened 2024-02-05
Fix some problem under Windows OS development enviroment

Rokixy:
1. language.ts will be changed in every commit when commiting with a 'zh-cn' OS language setting. Fix it by specifying the locale 'en' when sorting languages with localeCompare().

2. The parameter of glob.sync() must in posix path form. In Win32 system the path separator '\\' considered as a escape char by node-glob, Fix it by replace '\\' with '/'.

3. build-api with child_process.execFile() will cause "Error: spawn yarn ENOENT" in Win32. Replace it with child_process.exec().
### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- stablize language.ts when running under different OS language.
- add isWin32() and posixPath() to format the parameter of glob.sync().
- use child_process.exec() instead of child_process.execFile() for win32 platform.

### DIFF
--- a/scripts/lib/exec.ts
+++ b/scripts/lib/exec.ts
@@ -1,4 +1,4 @@
-import { execFile } from 'child_process'
+import { exec as execute } from 'child_process'
 import { nicelog } from './nicelog'
 
 type ExecOpts = {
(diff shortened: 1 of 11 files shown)


scripts/lib/exec.ts
```diff
@@ -27,50 +27,29 @@ export async function exec(
 ): Promise<string> {
 	nicelog(`> $ ${command} ${args.join(' ')} (in ${pwd}))`)
 	return new Promise((resolve, reject) => {
-		const data: string[] = []
-
-		const childProcess = execFile(
-			command,
-			args.filter((arg): arg is string => !!arg),
+		const childProcess = execute(
+			`${command} ${args.filter((arg): arg is string => !!arg).join(' ')}`,
```
SomeHats, line 31: `execFile` was used here because it sanitizes the arguments passed to it. Can we restore sanitization here?

scripts/lib/exec.ts
```diff
@@ -27,50 +27,29 @@ export async function exec(
...
 			{ cwd: pwd, env: { ...process.env, ...env } },
-			(err) => {
-				if (err) reject(err)
-				else resolve(data.join(''))
+			(err, stdout, stderr) => {
+				if (err) {
+					reject(err)
+				} else {
+					const combinedOutput = `${stdout}${stderr}`
+					resolve(combinedOutput.trim())
```
SomeHats, line 38: We want `stdout` and `stderr` to be interleaved rather than one after the other. Why this change?

scripts/lib/exec.ts
```diff
@@ -27,50 +27,29 @@ export async function exec(
...
+			childProcess.stdout.on('data', (data) => {
+				processStdoutLine(data)
+			})
+		}
+
+		if (childProcess.stderr) {
+			childProcess.stderr.on('data', (data) => {
+				processStderrLine(data)
+			})
+		}
```
SomeHats, line 53: afaict, these `processLine` callbacks now might be dealing with larger chunks of data instead of single lines if the child process outputs either incomplete lines or multiple lines at once.

What's the intention behind removing the line processing?

SomeHats (changes requested):
Thanks @Rokixy!

I'm a little hesitant for us to take all of these changes right now. In general, although we support the tldraw library on windows, we don't particularly aim to support developing this repo on win32. This repo does run within the [WSL](https://learn.microsoft.com/en-us/windows/wsl/) though which might help.

The language.ts stability thing seems generally useful - I would very happily land this on its own.

The other windows-specific changes i'm a little unsure about as they're likely to break in future as we don't develop for windows and don't have any tests. If you wanted to accompany this with a windows CI action that tests some of these scripts then that would be cool!

For the `exec` changes specifically, I think there are some regressions in the functionality of that helper that this introduces. I'd want to see those resolved before landing this. One option we could consider there is switching from plain `child_process` to https://github.com/sindresorhus/execa.

Sorry to be so resistant to this change! I just want to make sure we can support anything we do here going forward. In general, I would definitely suggest just working with this repo under the WSL instead.

Rokixy:
Thank you for your review @SomeHats .

Everything works fine in WSL. I'm not sure about windows-specific changes and `exec` changes is needed. So I revert  those two commits.

SomeHats (approved):
Thanks! This is a great change

PR #2723, opened 2024-02-05
docs: rework search UI

mimecuvalo:
Reworks search to not be a page and instead to be inline dropdown.

Screenshot 2024-02-05 at 13 22 58

- rework search completely
- rm Search Results css
- uses Ariakit and add appropriate hooks / styling
  - I couldn't use Radix unfortunately since they're still working on adding a Combox: https://github.com/radix-ui/primitives/issues/1342
  - I'm open to other suggestions but Ariakit plays nicely with Radix and keeps things open to migrate to Radix in the future
- fixes bug with not scrolling to right place when having a direct link
- adds categories in the search results - examples / reference / learn
- and adds category icons. Let me know if there's a better policy for adding new SVG icons cc @steveruizok 

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Test searches using normal method for each type (examples, docs, refs)
2. Test searches using AI for each type (ditto)

### Release Notes

- Docs: rework the search to be an inline dropdown.


Taha-Hassan-Git (approved):
Looks great! I think this is so much nicer than going to a page to then select the thing you want. I just have a couple of questions.

Re: The dropdown, I wonder if it makes sense to use up this space to the right a bit more. When there are lots of results it can take a while to scroll through. I think the user is mostly going to want something from examples or articles, could we put them side by side and have reference stuff underneath?
shapes at 24-02-05 13 18 33

Do you think we might need a different solution for smaller screens? Things are getting quite squished below 750px, and not really useable on mobile.

Also, I've not been able to test the AI search properly. Keep getting a 500 in the console.  Is there something I need to be doing?

mimecuvalo:
> Re: The dropdown, I wonder if it makes sense to use up this space to the right a bit more. When there are lots of results it can take a while to scroll through. I think the user is mostly going to want something from examples or articles, could we put them side by side and have reference stuff underneath?

Good feedback — I've gone ahead and made it not match the width for now. And in the future maybe we can explore maybe some side-by-side UI — that might be cool.

> Do you think we might need a different solution for smaller screens? Things are getting quite squished below 750px, and not really useable on mobile.

Nice catch! Fixed up!

> Also, I've not been able to test the AI search properly. Keep getting a 500 in the console. Is there something I need to be doing?

Yeah, the OpenAI key has to be in your .env.

PR #2730, opened 2024-02-05
Fixed actions menu opening in wrong direction on mobile (and add an inline layout example)

TodePond:
This PR changes the direction of the actions menu popover when it's at the bottom of the screen.
It's now consistent with all other menu dropdowns (or dropups?).

This PR also adds an example that demonstrates the Tldraw component at various different size points. It was helpful when trying out this change. And I'm using it to demonstrate more incoming changes.

image

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Open the inline example.
2. Click the actions overflow button.
3. Make sure it appears above the button, instead of below.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Dev: Fixed the actions menu opening in the wrong direction.


ds300 (approved):
⬆️ ⬇️ ✅

ds300:
nonblocking: the focus state on the inline example is maybe slightly too subtle?

TodePond:
> nonblocking: the focus state on the inline example is maybe slightly too subtle?

I agree - we also see this on the 'multiple editors' example. It's something I'm looking at with this inline component work

PR #2738, opened 2024-02-06
error reporting: rm ids from msgs for better Sentry grouping

mimecuvalo:
This removes the ids from shape paths so that they can be grouped on our error reporting tool.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Error reporting: improve grouping for Sentry.


packages/validate/src/lib/validation.ts
```diff
@@ -24,6 +25,10 @@ function formatPath(path: ReadonlyArray<number | string>): string | null {
 			formattedPath += `.${item}`
 		}
 	}
+
+	// N.B. We don't want id's in the path because they make grouping in Sentry tough.
+	formattedPath = formattedPath.replace(/id = [^,]+, /, '').replace(/id = [^)]+/, '')
```
SomeHats, line 30: is it possible to do this in the sentry reporter rather than in the validation library? these are often useful to find in the console when debugging

Alternatively, we could remove the `T.model` type from validators. All it does is add this ID information, so if we're going to strip it out later we might as well remove that type entirely.

mimecuvalo: I've been looking at the Sentry grouping mechanisms and it doesn't have, like, a regex system to ignore this bit and group on the other parts unfortunately :-/

I definitely could see this being useful in development — maybe it's something that is not stripped when `NODE_ENV === 'development'`?

Hm, the `T.Model` afaict has more than just id - it has things like `type = 'arrow'` and that sort of thing.

ds300 (approved):
this is fine by me. the ids are mainly useful when debugging locally anyhow, and you can always figure them out ad hoc if needed.

PR #2742, opened 2024-02-06
arrows: update cursor only when in Select mode

mimecuvalo:
The cursor was updating for the arrow label even when in other tools (e.g. Draw) and it should only be updating when in Select mode. That addresses this issue: https://github.com/tldraw/tldraw/issues/2750

Also, there was a problem with arrows and zero length arrows and labels. The problem was actually in `Vec.ts` where we were dividing by zero. Addresses this bug https://github.com/tldraw/tldraw/issues/2749

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Make sure that the cursor is only applicable to the Select tool.

### Release Notes

- Cursor tweak for arrow labels.

### DIFF
--- a/packages/editor/src/lib/primitives/Vec.ts
+++ b/packages/editor/src/lib/primitives/Vec.ts
@@ -362,7 +362,10 @@ export class Vec {
 	 * Get the unit vector of A.
 	 */
 	static Uni(A: VecLike) {
-		return Vec.Div(A, Vec.Len(A))
+		const len = Vec.Len(A)
+		if (len === 0) return Vec.From(A)
+
+		return Vec.Div(A, len)
 	}
 
 	static Tan(A: VecLike, B: VecLike): Vec {
--- a/packages/editor/src/lib/primitives/Vec.ts
+++ b/packages/editor/src/lib/primitives/Vec.ts
@@ -362,10 +362,7 @@ export class Vec {
 	 * Get the unit vector of A.
 	 */
 	static Uni(A: VecLike) {
-		const len = Vec.Len(A)
-		if (len === 0) return Vec.From(A)
-
-		return Vec.Div(A, len)
+		return Vec.Div(A, Vec.Len(A))
 	}
 
 	static Tan(A: VecLike, B: VecLike): Vec {
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -35,26 +35,11 @@ export class Idle extends StateNode {
 	override onPointerMove: TLEventHandlers['onPointerMove'] = () => {
 		updateHoveredId(this.editor)
 
-		const hitShape = this.editor.getShapeAtPoint(this.editor.inputs.currentPagePoint, {
-			hitInside: false,
-			hitLabels: false,
-			margin: HIT_TEST_MARGIN / this.editor.getZoomLevel(),
-			renderingOnly: true,
-		})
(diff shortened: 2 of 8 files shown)


packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -34,6 +34,28 @@ export class Idle extends StateNode {
 
 	override onPointerMove: TLEventHandlers['onPointerMove'] = () => {
 		updateHoveredId(this.editor)
+
+		const hitShape = this.editor.getShapeAtPoint(this.editor.inputs.currentPagePoint, {
+			hitInside: false,
+			hitLabels: false,
+			margin: HIT_TEST_MARGIN / this.editor.getZoomLevel(),
+			renderingOnly: true,
+		})
```
ds300, line 43: I think you could use this?
```suggestion
		const hitShape = editor.getHoveredShape()
```

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -499,6 +510,28 @@ export class Idle extends StateNode {
...
+
+		// todo: Extract into general hit test for arrows
+		if (this.editor.isShapeOfType<TLArrowShape>(shape, 'arrow')) {
+			// How should we handle multiple labels? Do shapes ever have multiple labels?
+			const labelGeometry = this.editor.getShapeGeometry<Group2d>(shape).children[1]
+			// Knowing what we know about arrows... if the shape has no text in its label,
+			// then the label geometry should not be there.
+			if (labelGeometry && pointInPolygon(pointInShapeSpace, labelGeometry.vertices)) {
+				return true
+			}
```
ds300, line 529: nonblocking: I know this was in a previous PR, but it feels brittle to me, i wonder whether we could put the logic on the ArrowShapeUtil class, something like `arrowUtil.isPointWithinArrowLabelBounds(arrow, pointInShapeSpace)`?

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -34,6 +34,28 @@ export class Idle extends StateNode {
...
+			margin: HIT_TEST_MARGIN / this.editor.getZoomLevel(),
+			renderingOnly: true,
+		})
+		if (this.isOverArrowLabelTest(hitShape)) {
+			this.editor.updateInstanceState({
+				cursor: {
+					type: 'pointer',
+					rotation: 0,
+				},
+			})
```
SomeHats, line 50: this is fine, but you could also use `this.editor.setCursor` for these if you wanted

packages/editor/src/lib/primitives/Vec.ts
```diff
@@ -362,7 +362,10 @@ export class Vec {
 	 * Get the unit vector of A.
 	 */
 	static Uni(A: VecLike) {
-		return Vec.Div(A, Vec.Len(A))
+		const len = Vec.Len(A)
+		if (len === 0) return Vec.From(A)
+
+		return Vec.Div(A, len)
```
SomeHats, line 368: discussed offline - moving this out of `Vec` to the callsite instead

SomeHats (approved):

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
mimecuvalo: sounds good - will followup later 👍

PR #2747, opened 2024-02-06
Split snap manager into ShapeBoundsSnaps and HandleSnaps

SomeHats:
Taking the opportunity for some last-minute low-consequence breaking changes before 2.0, this diff does some prep work for adding a new snapping API by making the distinction between the two types of snapping a bit clearer and cleaning up some naming.

- `SnapManager` has had most of the actual snapping logic moved into two properties: `shapeBounds` (for snapping shape bounds on translate and resize) and `handles` (for snapping handles)
- `SnapLine`s are renamed to `SnapIndicator`s. The 'line' name was a bit confusing because not all of these indicators are lines (the new vertex snap type will be a single point)

I'm not too worried about this being a breaking change as it touches an area of the API that I'd be very surprised if more than a couple of people were using.

### Change Type

### Test Plan

- No user-facing changes.

### Release Notes

- `SnapLine`s are now called `SnapIndicator`s
- Snapping methods moved from `editor.snaps` to `editor.snaps.shapeBounds` and `editor.snaps.handles` depending on the type of snapping you're trying to do.


ds300 (approved):
makes sense!

PR #2754, opened 2024-02-06
feat(minimap): make minimap public for import

judicaelandria:
Make `Minimap` component public and expose it to `@tldraw/tldraw` package for import

### Change Type

- [x] `minor` — New feature

### Test Plan

You should be able to import `Minimap` from `@tldraw/tldraw` package

### Release Notes

- Make `Minimap` component public so we can import it from `@tldraw/tldraw` package


judicaelandria:
I don't think an example in necessary here since we can just import it like
```js
import { Minimap } from "@tldraw/tldraw"
```

PR #2758, opened 2024-02-06
Style UI based on component size instead of window size

TodePond:
Previously, our UI changed appearance based on your browser's **window size**.
This PR makes it change appearance based on tldraw's **component size** instead.

It makes the UI behave as we intend in small inline components.
See the **Inline component** example for a comparison of different screen sizes.
See code comments for more info.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Open the inline component example.
2. Gradually change the width of the browser window.
3. The UI of the components shouldn't change when you do this. 

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Dev: Fixed the default tldraw UI not matching the size of the component.


apps/examples/src/examples/inline/InlineExample.tsx
```diff
@@ -23,6 +23,7 @@ export default function InlineExample() {
 					gap: 12,
 				}}
 			>
+				<InlineEditor width={900} height={600} />
```
TodePond, line 26: Added another size to this example to highlight these sorts of changes more clearly.

packages/tldraw/src/lib/ui.css
```diff
@@ -264,6 +264,16 @@
...
+.tlui-layout__mobile .tlui-button__tool {
+	height: 48px;
+	width: 44px;
+}
+
+.tlui-layout__mobile .tlui-button__tool > .tlui-icon {
+	height: 16px;
+	width: 16px;
+}
+
```
TodePond, line 276: We already dynamically apply the `tlui-layout__mobile` class, so let's use that instead of a media query.

packages/tldraw/src/lib/ui.css
```diff
@@ -1496,13 +1484,15 @@
 	touch-action: auto;
 }
 
+/* TODO: refactor shortcuts dialog to make it work in small inline components */
```
TodePond, line 1487: The keyboard shortcut dialog has various issues at smaller inline-component sizes. I think there's no use in porting those over, so will address separately.

packages/tldraw/src/lib/ui/components/primitives/Kbd.tsx
```diff
@@ -7,6 +8,8 @@ export interface KbdProps {
 
 /** @internal */
 export function Kbd({ children }: KbdProps) {
+	const breakpoint = useBreakpoint()
+	if (breakpoint < 4) return null
```
TodePond, line 12: Feels slightly weird putting this breakpoint guard here, but it's what our CSS is currently doing. I'm happy to keep the same behaviour, but I'd also understand if we'd prefer to move this behaviour up to a higher level.

PR #2761, opened 2024-02-07
[docs] url changes, hide breadcrumb on desktop

steveruizok:
Small changes to docs site

### Change Type

### DIFF
--- a/apps/docs/components/Header.tsx
+++ b/apps/docs/components/Header.tsx
@@ -69,7 +69,7 @@ export function SectionLinks({ sectionId }: { sectionId?: string | null }) {
 				data-active={sectionId === 'reference'}
 				className="layout_header__section"
 			>
-				Reference
+				API
 			</a>
 			<a
 				href="/examples/basic/basic"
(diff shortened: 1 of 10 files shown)


apps/docs/components/Header.tsx
```diff
@@ -69,7 +69,7 @@ export function SectionLinks({ sectionId }: { sectionId?: string | null }) {
 				data-active={sectionId === 'reference'}
 				className="layout_header__section"
 			>
-				Reference
+				API
```
mimecuvalo, line 72: this should still say 'Reference' i think - it makes it consistent with other doc sites, and besides, "API" can feel overloaded/vague a bit.

apps/examples/src/examples.tsx
```diff
@@ -13,7 +13,7 @@ export type Example = {
 	loadComponent: () => Promise<ComponentType>
 }
 
-type Category = 'basic' | 'editor-api' | 'ui' | 'collaboration' | 'data/assets' | 'shapes/tools'
+type Category = 'basic' | 'editor-api' | 'ui' | 'collaboration' | 'data' | 'shapes-tools'
```
mimecuvalo, line 16: this is a bit annoying but you'll need to update `generateExamplesContent` to match these categories

mimecuvalo (changes requested):
flagging that the `generateExamplesContent` set needs updating

PR #2773, opened 2024-02-07
dev: swap yarn test and test-dev for better dx

mimecuvalo:
As discussed offline, just making `yarn test` do what we expect it to.

### Change Type


SomeHats (approved):

mimecuvalo:
@si14 made the good point that `--watch` shouldn't be the default option on `yarn test` — fixed up to have `--watch` be something you have to proactively add to `yarn test --watch` if you want to go into that mode.

PR #2781, opened 2024-02-07
Use github actions to mirror templates from monorepo to appropriate repos

si14:
The idea here is to mirror project templates from our monorepo to the corresponding "top level" repos in our organisation. The main benefits are:

- being able to easily synchronise template versions to the last release on release
- being able to test those templates together with the rest of the project
- being able to deploy the unreleased version of the templates 

However, this PR will only do the synchronisation part. The main roadblock for the latter at the moment is running `yarn install` on packages in the current `main` that are still unreleased. There are ways around that, but this PR is deliberately minimalistic to land it sooner.

### Change Type


Taha-Hassan-Git:
Added the templates but got stuck trying to generate an ssh deploy key, I don't think I have the necessary permissions anyway.

mimecuvalo:
(this is so cool btw)

si14:
@Taha-Hassan-Git added the keys & edited the workflow so you can sleep well on the weekend 😄

mimecuvalo:
@si14 just curious - why does the `.yarn/cache` need to be checked in? what does that help with?

si14:
@mimecuvalo it shouldn't, it's a bug in the template. Mea culpa, I didn't check the templates themselves

PR #2783, opened 2024-02-07
menu: ensure Export menu is under File menu

mimecuvalo:
It's currently missing under the File menu and it really should be there for intuitive placement.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Menu: show Export under the File menu.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useMenuSchema.tsx
+++ b/packages/tldraw/src/lib/ui/hooks/useMenuSchema.tsx
@@ -116,12 +116,28 @@ export function TLUiMenuSchemaProvider({ overrides, children }: TLUiMenuSchemaPr
 	)
 
 	const menuSchema = useMemo<TLUiMenuSchema>(() => {
+		const exportSubmenu = menuSubmenu(
+			'export-as',
+			'menu.export-as',
+			menuGroup(
+				'export-as-group',
+				menuItem(actions['export-as-svg'], { disabled: emptyPage }),
+				menuItem(actions['export-as-png'], { disabled: emptyPage }),
+				menuItem(actions['export-as-json'], { disabled: emptyPage })
+			),
+			menuGroup(
+				'export-bg',
+				menuItem(actions['toggle-transparent'], { checked: !exportBackground })
+			)
+		)
+
 		const menuSchema: TLUiMenuSchema = compactMenuItems([
 			menuGroup(
 				'menu',
 				menuSubmenu(
 					'file',
 					'menu.file',
+					exportSubmenu,
 					menuGroup('print', menuItem(actions['print'], { disabled: emptyPage }))
 				),
 				menuSubmenu(
@@ -154,20 +170,7 @@ export function TLUiMenuSchemaProvider({ overrides, children }: TLUiMenuSchemaPr
 								menuItem(actions['toggle-transparent'], { checked: !exportBackground })
 							)
 						),
-						menuSubmenu(
-							'export-as',
-							'menu.export-as',
-							menuGroup(
-								'export-as-group',
-								menuItem(actions['export-as-svg'], { disabled: emptyPage }),
-								menuItem(actions['export-as-png'], { disabled: emptyPage }),
-								menuItem(actions['export-as-json'], { disabled: emptyPage })
-							),
-							menuGroup(
-								'export-bg',
-								menuItem(actions['toggle-transparent'], { checked: !exportBackground })
-							)
-						)
+						exportSubmenu
 					),
 					menuGroup(
 						'set-selection-group',


mimecuvalo:
@SomeHats thx for the context — totally makes sense. I think related to https://github.com/tldraw/tldraw/pull/2782 it might help alleviate the tension if we create an Object menu. Export might fit more naturally in there, too.

Actually, maybe what makes more sense is to just create a Save button in File, which might have a dialog of the Export options. I mean, the impetus for putting this in File is b/c folks wanna save stuff. That way, you could provide:
- export as SVG, PNG, etc.
- but also, provide a Save to .tldraw file which would be great

SomeHats:
Save to .tldr is supported! It's the 'Save a copy' option - which is under file. We don't have a general "save" because browser limitations mean that we can't make a proper 'save' option where we overwrite the same file. 

I think we could merge this as-is, or make the versions under file export the entire page instead of the selection (maybe with a copy tweak to reflect that)

mimecuvalo:
(going to wait until https://github.com/tldraw/tldraw/pull/2796 lands and then i'll refactor)

mimecuvalo:
@steveruizok take a look at this when you get the chance - i'm changing a bunch of things here
@SomeHats PTAL because this has changed since i first sent it to you.

SomeHats (commented):
Weird bug here with multiplayer rooms:
Screenshot 2024-02-20 at 17 31 22

So, the reason that all this file stuff lived in dotcom rather than in the app is because in general, these file based things only really make sense within dotcom where we can make sensible decisions about what to do when people open new files etc. The logic in these menus is a bit weird, very specific to what we want on .com, and different between the local scratchpad and shared multiplayer rooms.

If I'm a library user and I'm integrating tldraw as e.g. an embedded whiteboard in my document editing app, I don't want users to be able to open a file in that whiteboard - that's not an operation that makes sense in that context.

mimecuvalo:
> Weird bug here with multiplayer rooms:

hah! nice catch. fixed.

> So, the reason that all this file stuff lived in dotcom rather than in the app is because in general, these file based things only really make sense within dotcom where we can make sensible decisions about what to do when people open new files etc. The logic in these menus is a bit weird, very specific to what we want on .com, and different between the local scratchpad and shared multiplayer rooms.

I disagree but let's discuss! Because it's a fundamental question of batteries included vs. not. The New/Open/Save functionality isn't really dotcom behavior since it doesn't _actually_ create a cloud or some equivalent real file behavior. The New behavior, for example, as it currently is implement, merely clears the board.

I _could_ concede that there's an argument that the .tldr format is _only_ for dotcom but I would argue that users would want a batteries-included format that should generally work for non-custom rolled tldraw instances. Furthermore, the fact that `parseAndLoadDocument` is in the tldraw package is in favor of making open part of tldraw. If `parseAndLoadDocument` was _outside_ of tldraw and just part dotcom then it would be less arguable.

SomeHats:
> The New/Open/Save functionality isn't really dotcom behavior since it doesn't actually create a cloud or some equivalent real file behavior. The New behavior, for example, as it currently is implement, merely clears the board.

I disagree - "New" clearing the document is completely unexpected (and almost certainly not what you want) in any scenario _other_ than our .com which has this very specific "local scratchpad" functionality. If you have any other sort of "project" notion, that's not what you want. 

The file parsing stuff is in the library to give people an easy way to add whatever file support works well for their specific use-case. If you have a product where file reading/writing is important, you can take that library function and put together a file menu fairly easily.

`parseAndLoadDocument` itself is actually marked `@internal` (ie hidden from the public API) because it contains a lot of extra us-specific code around e.g. tldraw v1 docs. It's there because it's used by our vscode extension (which is a good example of where the specific .com file menu doesn't make sense, but file parsing itself is useful)

mimecuvalo:
ok! we discussed this offline. thx for hashing this out with me @SomeHats

packages/tldraw/src/lib/ui/components/MainMenu/DefaultMainMenuContent.tsx
```diff
@@ -37,15 +36,54 @@ import { TldrawUiMenuSubmenu } from '../primitives/menus/TldrawUiMenuSubmenu'
 export function DefaultMainMenuContent() {
 	return (
 		<>
+			<FileSubmenu />
```
SomeHats, line 39: I still don't think thing belongs as a default in the library. I don't think wanting this extra place for export justifies a menu that in a lot of contexts where people are using tldraw is just going to feel a bit weird and out of place. Really, unless you're using tldraw as a full-screen drop-in whiteboard, "File" isn't going to make sense. If you _are_, you probably want to override this and do your own thing anyway.

If we thought it was important to add separate options for exporting the entire tldraw document - maybe all pages as PNGs or something - then I could maybe see file being more justified there. As it is though, we're just doing the same "export the current selection" option which is already accessible in the edit & context menus.

packages/tldraw/src/lib/ui/components/menu-items.tsx
```diff
@@ -446,9 +461,13 @@ export function EmbedsGroup() {
 
 	return (
 		<TldrawUiMenuGroup id="embeds">
-			{oneEmbedSelected && <TldrawUiMenuItem {...actions['edit-embed']} />}
-			{oneEmbedSelected && <TldrawUiMenuItem {...actions['convert-to-bookmark']} />}
-			{oneEmbeddableBookmarkSelected && <TldrawUiMenuItem {...actions['convert-to-embed']} />}
+			{/* XXX this doesn't exist?? */}
+			{/* <TldrawUiMenuItem {...actions['edit-embed']} disabled={!oneEmbedSelected} /> */}
```
SomeHats, line 465: ooft the UI around embeds is kinda broken at the moment huh...

SomeHats (commented):
* _re-orders to be the normative New / Open / Save order — we shouldn't be messing with this conventional ordering_ - love it!
* _removes the "Don't ask again" from New/Open dialogs because they're non-undoable and not what anybody should ever select. we shouldn't offer users a loaded footgun! :P_ - makes sense!
* _makes File menu be part of the default menu — it's presence is glaringly missing for regular development_ - I think its right that it's missing. More justification in the notes. I don't think it feeling missing in regular development is enough justification to add something that doesn't make a tonne of sense in most actual use-cases - especially when it's relatively easy to add if you want it.
* _along with that, make the pieces of that menu available as lego pieces to use - it can't just be DefaultMainMenuContent, all or nothing, forcing downstream users to import everything from scratch_- love it!
* _finally, adds the Export menu as initially intended by this PR!_ -😅

packages/tldraw/src/lib/ui/components/MainMenu/DefaultMainMenuContent.tsx
mimecuvalo: alright, sounds good — I see your position. I'll pull this part out and we can revisit later maybe 👍

PR #2786, opened 2024-02-08
edit link: add little transition to make the external link button a little more noticeable

mimecuvalo:
https://github.com/tldraw/tldraw/assets/469604/9b9078c6-2025-452e-a08c-2ecbf701b863

@steveruizok go ahead and play around with the css if you like

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- UI: adds a tiny animation when adding a link

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -956,7 +956,8 @@ input,
 	cursor: var(--tl-cursor-pointer);
 	border: none;
 	outline: none;
-	pointer-events: all;
+	pointer-events: none;
+	opacity: 0;
 }
 
 .tl-hyperlink-button::after {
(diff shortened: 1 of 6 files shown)


mimecuvalo:
whatchya think @steveruizok ? feel free to merge/close this one — this was a fun idea to play around with.
i want to push on the discoverability of the edit link icon

right now it's hard to see what happened when you finish editing a link. you have to _know_ that that little icon will show up and it's not obvious at the moment — currently it looks like nothing happened.

SomeHats:
one issue with this as it's currently implemented is that the animation plays whenever the component mounts, not just when it's added. for example, when loading the project, when switching pages, or when scrolling the shape into view if it has been culled.

mimecuvalo:
> one issue with this as it's currently implemented is that the animation plays whenever the component mounts, not just when it's added. for example, when loading the project, when switching pages, or when scrolling the shape into view if it has been culled.

good point! i can fix that up before landing if we decide we want this in the first place 👍

PR #2787, opened 2024-02-08
[experiment] paste: show little puff when pasting to denote something happened

mimecuvalo:
@steveruizok go ahead and take a look at the animation css and tweak as you like.
@SomeHats and @steveruizok this is an experiment in that this is prbly the wrong way to approach it? But I'd be curious to learn if there was a more proper route here 

https://github.com/tldraw/tldraw/assets/469604/40a7029c-f4e8-4f2a-914e-8e6f264be4c7

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- UI: add a little 'puff' when something is pasted to tell that something has happened.


steveruizok:
Hey, I put up a PR here that swapped the effect here with our hinted shapes. I'd prefer if we used that system; it's a bit more subtle but still provides some visual feedback. https://github.com/tldraw/tldraw/pull/2852

mimecuvalo:
> Hey, I put up a PR here that swapped the effect here with our hinted shapes. I'd prefer if we used that system; it's a bit more subtle but still provides some visual feedback. https://github.com/tldraw/tldraw/pull/2852

yaaaasss.

PR #2789, opened 2024-02-08
arrows: followup to clean up some hit test logic

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/2742#discussion_r1479816541
@ds300 is this what you had in mind?

### Change Type

- [x] `patch` — Bug fix


packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -505,13 +505,8 @@ export class Idle extends StateNode {
...
 		if (this.editor.isShapeOfType<TLArrowShape>(shape, 'arrow')) {
-			// How should we handle multiple labels? Do shapes ever have multiple labels?
-			const labelGeometry = this.editor.getShapeGeometry<Group2d>(shape).children[1]
-			// Knowing what we know about arrows... if the shape has no text in its label,
-			// then the label geometry should not be there.
-			if (labelGeometry && pointInPolygon(pointInShapeSpace, labelGeometry.vertices)) {
-				return true
-			}
+			const arrowShapeUtil = this.editor.getShapeUtil(shape) as ArrowShapeUtil
+			return arrowShapeUtil.isOverArrowLabelTest(shape, pointInShapeSpace)
```
steveruizok, line 509: Given the choice, I'd prefer the original implementation. It's a little fragile but acceptable, and better IMO than extending the public interface of a shapeutil.

mimecuvalo: i agree with David that we should keep the arrow logic together with other arrow logic - otherwise it might start getting out of sync.

steveruizok:
closing this one!

mimecuvalo:
re-opening b/c i'll rework this to add a `isLabelHovered` function

SomeHats:
I think we should have an offline conversation about this:
> Sorry, I have a strong feeling here about not adding one-off methods to the shape utils. Its important to keep that public surface area as generic as possible. 

This feels like it comes up a lot, and it's not a position I've ever really been able to get my head around if i'm honest. To me, adding a one-off to a specific shape where we need it is much better than adding something very specific to our abstraction for _all shapes_ when it's only needed for one.

steveruizok:
Apologies for the early close if there was still work to do here.

> This feels like it comes up a lot, and it's not a position I've ever really been able to get my head around if i'm honest. To me, adding a one-off to a specific shape where we need it is much better than adding something very specific to our abstraction for all shapes when it's only needed for one.

I agree that we don't need to create general abstractions if it's logic that's only needed in one place, but the disagreement is where we put that logic.

The purpose of shape utils is to answer questions that our editor / tools are unable to answer because the answer changes depending on the type of shape. An example is resizing: different shapes resize in different ways, and so shape util methods are a glorified way of inserting cases into all the switch statements that would otherwise exist throughout the app.

As a comparison, shape-specific logic is scattered throughout the excalidraw codebase; switches like [this one](https://github.com/excalidraw/excalidraw/blob/47f87f4ecbc62058b6e2b6b7d953952bf6f3ecaf/packages/excalidraw/renderer/renderScene.ts#L1300) are common, and it makes extending Excalidraw virtually impossible. In tldraw, I wanted extension to be a core feature, and so shape utils were a way of avoiding shape-specific code inside of tools / behavior.

In this case (and in cases like this) there really isn't a "hole" that is being filled by a shape util method. The tool is already running logic specific to a certain shape because we haven't identified a situation where the code would apply to multiple shapes, and each shape would behave differently. That's fine! It's really just a question of where we put that logic.

We could put that logic as a unique method on the shape util, or we could put the logic in the tool. It's mostly arbitrary, but I'd prefer to leave it in the tool or extracting it into a little internal function. This way we don't add to the public interface of the shape util or get into cases where we start having to cast utils in order to pick up those one-off methods.

mimecuvalo:
@steveruizok this was reworked a bit to start being more generic. does this fit what you were thinking? lemme know.

mimecuvalo:
ping ping @steveruizok

steveruizok:
Sorry, this would now fall under expanding the general APIs for a one-off (see Alex and my discussion above). The question wasn't "should label hit tests be generic" but rather "where should we put the one-off code related to a single arrow label hit test", and I had argued for placing that code in the tool for the reasons above.

mimecuvalo:
👍 no worries - i didn't feel strongly about this in the first place, was a prompt from @ds300 — so maybe @ds300 feels more strongly (if so, please weigh in David)

closing in the meantime!

PR #2792, opened 2024-02-08
Managers for better property caching for InstanceState & InstancePageState

SomeHats:
Big diff, sorry!

This diff deprecates `getInstanceState`, `getCurrentPageState`, and their related updaters in favour of managers that handle caching on the properties of those records.

These records are both bags of largely unrelated data, some of which updates very frequently. These frequent updates cause calls to the deprecated methods to trigger a lot of unneeded signia updates & react renders. By encouraging reads of those values to instead go through `@computed`, we cut down drastically on those extra updates/renders:

Kapture 2024-02-08 at 15 52 37

Running these two through managers mirrors `editor.userPreferences`, which has per-property getters, and a single top-level updater command. This diff mostly follows that pattern, with one deviation: accessors that already existed on `editor` are still there rather than in the manager (e.g. we do `editor.getSelectedShapeIds()`, not `editor.currentPageState.getSelectedShapeIds()`).

This has resulted in things being a bit inconsistent. We could also deprecate the editor versions of those getters and move them all to the manager, but i think this is probably fine for now.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Largely relying on existing unit & end to end tests for this one :/

### Release Notes

- `editor.getInstanceState()` is deprecated in favour of getters on `editor.instanceState` e.g. `editor.instanceState.getIsCoarsePointer()`
- `editor.getCurrentPageState()` is deprecated in favour of getters on `editor.currentPageState` e.g. `editor.currentPageState.getCroppingShapeId()`
- `editor.updateInstanceState` is deprecated in favour of `editor.instanceState.update`
- `editor.updateCurrentPageState` is deprecated in favour of `editor.currentPageState.update`

### DIFF
--- a/apps/docs/content/docs/editor.mdx
+++ b/apps/docs/content/docs/editor.mdx
@@ -273,7 +273,7 @@ You can get a shape using the shape's `id` or the shape record itself.
 You can use the [Editor#updateInstanceState](?) method to turn on read only mode.
 
 ```ts
-editor.updateInstanceState({ isReadonly: true })
+editor.instanceState.update({ isReadonly: true })
 ```
 
 ### Move the camera
@@ -289,7 +289,7 @@ editor.setCamera(0, 0, 1)
 You can prevent the user from changing the camera using the [Editor#updateInstanceState](?) method.
 
 ```ts
-editor.updateInstanceState({ canMoveCamera: false })
+editor.instanceState.update({ canMoveCamera: false })
 ```
 
 ### Turn on dark mode
--- a/apps/docs/content/docs/editor.mdx
+++ b/apps/docs/content/docs/editor.mdx
@@ -270,7 +270,7 @@ You can get a shape using the shape's `id` or the shape record itself.
 
 ### Turn on read only mode
 
-You can use the [Editor#updateInstanceState](?) method to turn on read only mode.
+You can use the [InstanceStateManager#update](?) method to turn on read only mode.
 
 ```ts
 editor.instanceState.update({ isReadonly: true })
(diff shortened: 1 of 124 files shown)


ds300:
I think we should get @steveruizok's input on this API pattern before moving ahead with this.

> before

Ah i can see we are updating the cursor on every frame here. That shouldn't really be happening in this case, but it does need to happen during rotations. I wonder whether we could split the cursor into it's own record type and achieve the same benefit with a much smaller diff. We did this for the camera back in the day for the same reason.

h2thien:
Hey 👋 Not sure if this is related, but while investigating a performance regression on our app, we found that compare to Beta 2, the latest Canary has a different behaviour that may be unexpected:

**TL;DR: When pointer moves, there's a redundant `instance` event.**

Specifically, when the user moves their pointer, without triggering any meaningful hover (on any shapes):
- Before (Beta 2): Most of the time, there's only "pointer" events in the "changes" array.
- After (Canary): Most of the time, there's not only "pointer", but also "instance" event.

We logged a diff on changes and found that the "instance" event does not have a meaningful change at all. The only reported field is "cursor", which for some reason changes its identity on every move:

image

Also, maybe this is an X vs Y problem, because what we ultimately want is just a way to watch for tool changes 😅 

image

steveruizok:
Thanks for the info @h2thien. This may be due to a change I'd made a few weeks ago, I'll look into it!

h2thien:
Thanks @steveruizok 🙏 We are no longer blocked by this thank to https://github.com/tldraw/tldraw/pull/2798 (this PR enables us to us TLDraw as a provider at root, eliminating all our needs for manual event listening and filtering).

That being said, if you think it's useful (for other folks in the future?), I'd be more than happy to extract the comment to an issue. Sorry everyone here for my off-topic comment 🙇

SomeHats:
I think this needs a little more thought - I'm gonna close for now in favour of just fixing the updates being to frequent here: https://github.com/tldraw/tldraw/pull/2820

PR #2796, opened 2024-02-10
Composable custom UI 

steveruizok:
This PR refactors our menu systems and provides an interface to hide or replace individual user interface elements.

# Background

Previously, we've had two types of overrides:
- "schema" overrides that would allow insertion or replacement of items in the different menus
- "component" overrides that would replace components in the editor's user interface

This PR is an attempt to unify the two and to provide for additional cases where the "schema-based" user interface had begun to break down. 

# Approach

This PR makes no attempt to change the `actions` or `tools` overrides—the current system seems to be correct for those because they are not reactive. The challenge with the other ui schemas is that they _are_ reactive, and thus the overrides both need to a) be fed in from outside of the editor as props, and b) react to changes from the editor, which is an impossible situation.

The new approach is to use React to declare menu items. (Surprise!) 

```tsx
function CustomHelpMenuContent() {
	return (
		<>
			<DefaultHelpMenuContent />
			<TldrawUiMenuGroup id="custom stuff">
				<TldrawUiMenuItem
					id="about"
					label="Like my posts"
					icon="external-link"
					readonlyOk
					onSelect={() => {
						window.open('https://x.com/tldraw', '_blank')
					}}
				/>
			</TldrawUiMenuGroup>
		</>
	)
}

const components: TLComponents = {
	HelpMenuContent: CustomHelpMenuContent,
}

export default function CustomHelpMenuContentExample() {
	return (
			<Tldraw components={components} />
		</div>
	)
}
```

We use a `components` prop with the combined editor and ui components.

- [ ] Create a "layout" component?
- [ ] Make UI components more isolated? If possible, they shouldn't depend on styles outside of themselves, so that they can be used in other layouts. Maybe we wait on this because I'm feeling a slippery slope toward presumptions about configurability.
- [ ] OTOH maybe we go hard and consider these things as separate components, even packages, with their own interfaces for customizability / configurability, just go all the way with it, and see what that looks like.

# Pros

Top line: you can customize tldraw's user interface in a MUCH more granular / powerful way than before.

It solves a case where menu items could not be made stateful from outside of the editor context, and provides the option to do things in the menus that we couldn't allow previously with the "schema-based" approach.

It also may (who knows) be more performant because we can locate the state inside of the components for individual buttons and groups, instead of all at the top level above the "schema". Because items / groups decide their own state, we don't have to have big checks on how many items are selected, or whether we have a flippable state. Items and groups themselves are allowed to re-build as part of the regular React lifecycle. Menus aren't constantly being rebuilt, if that were ever an issue. 

Menu items can be shared between different menu types. We'll are sometimes able to re-use items between, for example, the menu and the context menu and the actions menu.

Our overrides no longer mutate anything, so there's less weird searching and finding.

# Cons

This approach can make customization menu contents significantly more complex, as an end user would need to re-declare most of a menu in order to make any change to it. Luckily a user can add things to the top or bottom of the context menu fairly easily. (And who knows, folks may actually want to do deep customization, and this allows for it.)

It's more code. We are shipping more react components, basically one for each menu item / group.

Currently this PR does not export the subcomponents, i.e. menu items. If we do want to export these, then heaven help us, it's going to be a _lot_ of exports.

# Progress 

- [x] Context menu
- [x] Main menu
- [x] Zoom menu
- [x] Help menu
- [x] Actions menu
- [x] Keyboard shortcuts menu
- [x] Quick actions in main menu? (new)
- [x] Helper buttons? (new)
- [x] Debug Menu

And potentially
- [x] Toolbar
- [x] Style menu
- [ ] Share zone
- [x] Navigation zone
- [ ] Other zones

### Change Type

### Test Plan

1. use the context menu
2. use the custom context menu example
3. use cursor chat in the context menu

- [x] Unit Tests
- [ ] End to end tests

### Release Notes


steveruizok:
Turns out the toolbar is going to be much more complicated, so I'd say this is now ready for review. There's some QA to do but after adding all of our examples, this is probably ready to ship.

packages/tldraw/src/lib/ui/components/Toolbar/DefaultToolbarContent.tsx
```diff
@@ -0,0 +1,52 @@
+import { useReadonly } from '../../hooks/useReadonly'
+import { useTools } from '../../hooks/useTools'
+import { TldrawUiMenuItem } from '../menus/TldrawUiMenuItem'
+
+/** @public */
+export function DefaultToolbarContent() {
```
SomeHats, line 6: Where does this get used?

packages/tldraw/src/lib/ui/components/PenModeToggle.tsx
```diff
@@ -16,7 +16,7 @@ export const ExitPenMode = track(function ExitPenMode() {
 	return (
 		<Button
 			type="normal"
-			label={action.label}
+			label={action.label as string}
```
SomeHats, line 19: should this use your unwrap helper?

packages/tldraw/src/lib/ui/hooks/useTldrawUiComponents.tsx
```diff
@@ -0,0 +1,104 @@
...
+	overrides?: TLUiComponents
+	children: any
+}
+
+/** @public */
+export function TldrawUiComponentsProvider({
+	overrides,
+	children,
+}: ComponentsContextProviderProps) {
+	return (
```
SomeHats, line 73: Can we stick a `overrides = useShallowObjectIdentity(overrides)` in here? That way if people want to write `<Tldraw overrides={{ Thing: MyThing }} />` we won't have re-render loops etc.

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -30,9 +30,9 @@ export type TLExternalContentProps = {
 	// The maximum size (in bytes) of an asset. Assets larger than this will be rejected. Defaults to 10mb (10 * 1024 * 1024).
 	maxAssetSize: number
 	// The mime types of images that are allowed to be handled. Defaults to ['image/jpeg', 'image/png', 'image/gif', 'image/svg+xml'].
-	acceptedImageMimeTypes: string[]
+	acceptedImageMimeTypes: string[] | readonly string[]
 	// The mime types of videos that are allowed to be handled. Defaults to ['video/mp4', 'video/webm', 'video/quicktime'].
-	acceptedVideoMimeTypes: string[]
+	acceptedVideoMimeTypes: string[] | readonly string[]
```
SomeHats, line 35: u can probably simplify this by just saying `readonly string[]`, you don't need both - one's a strict subset of the other

packages/tldraw/src/test/ui/ContextMenu.test.tsx
```diff
@@ -26,49 +21,78 @@ it('opens on right-click', async () => {
...
+	const elm = await screen.findByTestId('abc123')
+	expect(elm).toBeDefined()
+})
 
-	await act(() => editor.createShape({ id: createShapeId(), type: 'geo' }).selectAll())
+// The tunneling / overriding should address this
 
-	// open the context menu:
-	fireEvent.contextMenu(await screen.findByTestId('canvas'))
+// it.failing('updates overrides reactively', async () => {
```
SomeHats, line 53: we can probably just delete this now?

package.json
```diff
@@ -114,6 +114,7 @@
 		"@yarnpkg/types": "^4.0.0",
 		"cross-env": "^7.0.3",
 		"purgecss": "^5.0.0",
-		"svgo": "^3.0.2"
+		"svgo": "^3.0.2",
+		"tunnel-rat": "^0.1.2"
```
SomeHats, line 118: afaict this is unused - can we remove it?

packages/validate/src/lib/validation.d.ts.map
```diff
@@ -0,0 +1 @@
...G;AACH,eAAO,MAAM,SAAS,sBAMpB,CAAA;AAEF;;;;GAIG;AACH,wBAAgB,QAAQ,IAAI,aAAa,CAAC,MAAM,EAAE,SAAS,CAAC,CAE3D;AAED;;;;GAIG;AACH,wBAAgB,IAAI,CAAC,GAAG,SAAS,MAAM,EAAE,KAAK,EAC7C,YAAY,EAAE,WAAW,CAAC,GAAG,CAAC,EAC9B,cAAc,EAAE,WAAW,CAAC,KAAK,CAAC,GAChC,aAAa,CAAC,GAAG,EAAE,KAAK,CAAC,CAE3B;AAED;;;;;;;;;;;;;GAaG;AACH,wBAAgB,KAAK,CAAC,GAAG,SAAS,MAAM,EAAE,MAAM,SAAS,oBAAoB,CAAC,GAAG,EAAE,MAAM,CAAC,EACzF,GAAG,EAAE,GAAG,EACR,MAAM,EAAE,MAAM,GACZ,cAAc,CAAC,GAAG,EAAE,MAAM,CAAC,CAS7B;AAED;;;;;GAKG;AACH,wBAAgB,KAAK,CAAC,CAAC,SAAS;IAAE,QAAQ,CAAC,EAAE,EAAE,MAAM,CAAA;CAAE,EACtD,IAAI,EAAE,MAAM,EACZ,SAAS,EAAE,WAAW,CAAC,CAAC,CAAC,GACvB,SAAS,CAAC,CAAC,CAAC,CASd;AAED,cAAc;AACd,wBAAgB,OAAO,CAAC,CAAC,EAAE,MAAM,EAAE,WAAW,CAAC,CAAC,CAAC,GAAG,SAAS,CAAC,CAAC,CAAC,CAQ/D;AAED,cAAc;AACd,wBAAgB,QAAQ,CAAC,CAAC,EAAE,SAAS,EAAE,WAAW,CAAC,CAAC,CAAC,GAAG,SAAS,CAAC,CAAC,GAAG,SAAS,CAAC,CAK/E;AAED,cAAc;AACd,wBAAgB,QAAQ,CAAC,CAAC,EAAE,SAAS,EAAE,WAAW,CAAC,CAAC,CAAC,GAAG,SAAS,CAAC,CAAC,GAAG,IAAI,CAAC,CAK1E;AAED,cAAc;AACd,wBAAgB,WAAW,CAAC,KAAK,CAAC,MAAM,SAAS,SAAS,OAAO,EAAE,EAClE,GAAG,MAAM,EAAE,MAAM,GACf,SAAS,CAAC,MAAM,CAAC,MAAM,CAAC,CAAC,CAE3B;AAmBD;;;;GAIG;AACH,eAAO,MAAM,OAAO,mBASlB,CAAA;AAIF;;;;GAIG;AACH,eAAO,MAAM,MAAM,mBASjB,CAAA"}
```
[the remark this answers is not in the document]
steveruizok: nope!

packages/validate/tsconfig.json
```diff
@@ -4,7 +4,9 @@
 	"exclude": ["node_modules", "dist", ".tsbuild*"],
 	"compilerOptions": {
 		"outDir": "./.tsbuild",
-		"rootDir": "src"
+		"rootDir": "src",
+		"declaration": true,
+		"emitDeclarationOnly": true
```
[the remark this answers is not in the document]
steveruizok: tangential, I'll revert those

packages/tldraw/package.json
```diff
@@ -57,7 +57,8 @@
 		"canvas-size": "^1.2.6",
 		"classnames": "^2.3.2",
 		"hotkeys-js": "^3.11.2",
-		"lz-string": "^1.4.4"
+		"lz-string": "^1.4.4",
+		"zustand": "^4.5.0"
```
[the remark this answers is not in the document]
steveruizok: yep, that was used in the tunnel rat stuff, good catch!

apps/dotcom/src/utils/sharing.ts
```diff
@@ -188,24 +185,24 @@ export function useSharing({ isMultiplayer }: { isMultiplayer: boolean }): TLUiO
...
+			// 				'share',
+			// 				menuItem(actions[FORK_PROJECT_ACTION]),
+			// 				menuItem(actions[LEAVE_SHARED_PROJECT_ACTION])
+			// 			)!
+			// 		)
+			// 	} else {
+			// 		fileMenu.children.unshift(menuGroup('share', menuItem(actions[SHARE_PROJECT_ACTION]))!)
+			// 	}
+			// 	return menu
+			// },
```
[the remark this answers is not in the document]
steveruizok: yep

packages/tldraw/src/lib/ui/components/Toolbar/DefaultToolbarContent.tsx
steveruizok: Not used yet. I'll bin it until we get to the toolbar.

packages/tldraw/src/lib/ui/hooks/useTldrawUiComponents.tsx
steveruizok: I think that happens higher up? but yes, can do

steveruizok: I'll add it to the editor components too

packages/tldraw/src/test/ui/ContextMenu.test.tsx
steveruizok: got it

package.json
steveruizok: yep

packages/tldraw/src/lib/ui/components/PenModeToggle.tsx
steveruizok: Yep, there was another one too.

packages/tldraw/package.json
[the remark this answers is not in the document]
mimecuvalo: haha TIL what a tunnel rat is :P

apps/examples/src/examples/custom-actions-menu/README.md
```diff
@@ -0,0 +1,11 @@
...
+title: Custom actions menu
+component: ./CustomActionsMenuExample.tsx
+category: ui
+---
+
+You can customize tldraw's actions menu.
+
+---
+
+The actions menu can be customized by providing a `ActionsMenu` component to the `Tldraw` component's `uiComponents` prop. If you provide `null`, then that component will be hidden.
```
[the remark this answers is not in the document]
steveruizok: The examples copy here is pretty rough, will do a pass before landing.

apps/vscode/editor/src/Links.tsx
```diff
@@ -0,0 +1,45 @@
...
+					openUrl('https://discord.gg/SBBEVCA4PG')
+				}}
+			/>
+			<TldrawUiMenuItem
+				id="about"
+				label="help-menu.about"
+				icon="external-link"
+				readonlyOk
+				onSelect={() => {
+					openUrl('https://www.tldraw.dev')
```
mimecuvalo, line 40: ditto on prev. comment here
(also feels like these links should be in a constants file somewhere)

packages/tldraw/src/lib/ui/hooks/useUnlockedSelectedShapesCount.ts
```diff
@@ -0,0 +1,34 @@
+import { useEditor, useValue } from '@tldraw/editor'
+
+/**
+ * Returns true if the number of UNLOCKED selected shapes is at least min or at most max.
+ */
+export function useUnlockedSelectedShapesCount(min?: number, max?: number) {
```
mimecuvalo, line 6: this name is throwing me off a bit - should it be more like `useIsUnlockedSelectedShapesClamped` since it's a boolean, not a number being returned?

packages/tldraw/src/lib/ui/hooks/useUnlockedSelectedShapesCount.ts
```diff
@@ -0,0 +1,34 @@
...
+	const editor = useEditor()
+	return useValue(
+		'selectedShapes',
+		() => {
+			const len = editor
+				.getSelectedShapes()
+				.filter((s) => !editor.isShapeOrAncestorLocked(s)).length
+			if (min === undefined) {
```
mimecuvalo, line 14: would this help simplify this?
`len >= (min || 0) || len <= (max || Infinity)`

packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
```diff
@@ -105,7 +105,7 @@ export const TranslationProvider = track(function TranslationProvider({
 export function useTranslation() {
 	const translation = useCurrentTranslation()
 	return React.useCallback(
-		function msg(id: Exclude<string, TLUiTranslationKey> | string) {
+		function msg(id?: Exclude<string, TLUiTranslationKey> | string) {
```
mimecuvalo, line 108: what does a msg with no `id` mean? shouldn't that error?

packages/tldraw/src/lib/ui/hooks/useAnySelectedShapesCount.ts
```diff
@@ -0,0 +1,32 @@
+import { useEditor, useValue } from '@tldraw/editor'
+
+/**
+ * Returns true if the number of LOCKED OR UNLOCKED selected shapes is at least min or at most max.
+ */
+export function useAnySelectedShapesCount(min?: number, max?: number) {
```
mimecuvalo, line 6: same comments as `useUnlockedSelectedShapesCount` w.r.t. to naming and simplifying logic

packages/tldraw/src/lib/ui/hooks/useActions.tsx
```diff
@@ -1211,3 +1255,14 @@ export function useActions() {
 function asActions<T extends Record<string, TLUiActionItem>>(actions: T) {
 	return actions as Record<keyof typeof actions, TLUiActionItem>
 }
+
+/** @public */
+export function unwrapLabel(label?: TLUiActionItem['label'], menuType?: string) {
```
mimecuvalo, line 1260: too hard to read :)
(i overlove ternaries too so sometimes i have the same problem :P)

packages/tldraw/src/lib/ui/hooks/useActions.tsx
```diff
@@ -184,9 +185,11 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 			},
 			{
 				id: 'export-as-json',
-				label: 'action.export-as-json',
-				menuLabel: 'action.export-as-json.short',
-				contextMenuLabel: 'action.export-as-json.short',
+				label: {
+					default: 'action.export-as-json',
+					menu: 'action.export-as-json.short',
+					['context-menu']: 'action.export-as-json.short',
```
mimecuvalo, line 191: reading all of these labels makes me think i wish it was:
```
  default: 'action.export-as-json',
  short: 'action.export-as-json.short',
```

and `short` is used where appropriate, if present. thoughts?

packages/tldraw/src/lib/ui/components/menus/TldrawUiMenuItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+}: {
+	icon?: IconType
+	id: string
+	kbd?: string
+	title?: string
+	label?: TransationKey | { [key: string]: TransationKey }
+	readonlyOk: boolean
+	onSelect: (source: TLUiEventSource) => Promise<void> | void
+	disabled?: boolean
+	noClose?: boolean
```
mimecuvalo, line 37: this should be renamed :)
`noClose` → `shouldPreventClose`

packages/tldraw/src/index.ts
```diff
@@ -170,6 +133,96 @@ export {
...
+} from './lib/ui/components/StylePanel/DefaultStylePanelContent'
+
+// Toolbar
+export { DefaultToolbar } from './lib/ui/components/Toolbar/DefaultToolbar'
+
+// Actions menu
+export {
+	DefaultActionsMenu,
+	type TLUiActionsMenuProps,
+} from './lib/ui/components/menus/ActionsMenu/DefaultActionsMenu'
```
mimecuvalo, line 171: feels like there shoud be a `/lib/ui/components/index.tsx` or `/lib/ui/components/menus/index.tsx` export list and this should all re-export from that. thoughts?

packages/tldraw/src/lib/ui/components/menus/TldrawUiMenuCheckboxItem.tsx
```diff
@@ -0,0 +1,98 @@
...
+	onSelect,
+	disabled = false,
+	checked = false,
+}: {
+	icon?: IconType
+	id: string
+	kbd?: string
+	title?: string
+	label?: TransationKey | { [key: string]: TransationKey }
+	readonlyOk: boolean
```
mimecuvalo, line 31: trying to parse `if (isReadOnly && !readonlyOk)` hurts my brain :P
can you rename this into something like `isReadonlySomethingSomething` or `shouldReadonly...`?

packages/tldraw/src/lib/ui/components/menus/TldrawUiMenuItem.tsx
```diff
@@ -0,0 +1,133 @@
...
+	readonlyOk,
+	onSelect,
+	noClose,
+}: {
+	icon?: IconType
+	id: string
+	kbd?: string
+	title?: string
+	label?: TransationKey | { [key: string]: TransationKey }
+	readonlyOk: boolean
```
mimecuvalo, line 34: ditto on prev. comment on renaming `readonlyOk`

packages/tldraw/src/lib/ui/hooks/useActions.tsx
TodePond: I think we tried something like this at one point, but it seemed simpler to make it specific to specific UI components. It avoids unclear things like `short`, `shorter`, etc

packages/tldraw/src/lib/ui/components/menus/TldrawUiMenuCheckboxItem.tsx
steveruizok: Hm, rather than renaming the variable, I'll make `isReadOnly` -> `isReadOnlyMode`, to differentiate between the editor's mode and the property of the action item.

packages/tldraw/src/index.ts
steveruizok: generally we avoid barrel files, some bundlers have tripped over them in the past

apps/dotcom/src/components/ExportMenu.tsx
```diff
@@ -36,7 +43,7 @@ export const ExportMenu = React.memo(function ExportMenu() {
 					<div className="tlui-menu__group">
 						<Button
 							type="menu"
-							label={shareProject.label}
+							label={unwrapLabel(shareProject.label)}
```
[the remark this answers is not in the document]
steveruizok: I haven't done a customizable menu for the "share menu" yet but we could refactor these to use the TldrawMenuItem stuff instead, which will take care of that?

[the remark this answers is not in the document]
mimecuvalo: yeah, there's some kind of refactor here that would help this. it can be for a future PR but wanted to flag it

packages/tldraw/src/index.ts
mimecuvalo: fair point — yeah that's true.

packages/tldraw/src/lib/ui/components/menus/TldrawUiMenuCheckboxItem.tsx
mimecuvalo: ah i see: my suggestion to say `isReadonlySomethingSomething`/`shouldReadonly` was actually off-base b/c i think i was confused by the naming heh, i was misreading the intent — i thought it was trying to convey that it should "go into read-only mode" vs. "being allowed to work in readonly mode". to that end, i think what would be better is to say `readonlyAllowed` or `allowedInReadonly` to be clearer.

i think it's maybe that "ok" implies a 'soft yes' and that's confusing when combined next to the word 'only' which implies strictness. like, for example, you wouldn't have a variable saying `downloadOnlyOk` because when it's false, that I guess would mean `uploadNeverMaybe`? 🙃 

(this is me getting pedantic on you, i know 😆 )

apps/dotcom/src/components/FileMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+	const actions = useActions()
+
+	return (
+		<TldrawUiMenuSubmenu id="file" label="menu.file">
+			<TldrawUiMenuGroup id="file-actions">
+				<TldrawUiMenuItem {...actions[SAVE_FILE_COPY_ACTION]} />
+				<TldrawUiMenuItem {...actions[OPEN_FILE_ACTION]} />
+				<TldrawUiMenuItem {...actions[NEW_PROJECT_ACTION]} />
+			</TldrawUiMenuGroup>
+			<TldrawUiMenuGroup id="share">
```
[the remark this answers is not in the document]
steveruizok: done!

PR #2798, opened 2024-02-10
Use canvas bounds for viewport bounds

steveruizok:
This PR changes the way that viewport bounds are calculated by using the canvas element as the source of truth, rather than the container. This allows for cases where the canvas is not the same dimensions as the component. (Given the way our UI and context works, there are cases where this is desired, i.e. toolbars and other items overlaid on top of the canvas area).

The editor's `getContainer` is now only used for the text measurement. It would be good to get that out somehow.

# Pros

We can inset the canvas

# Cons

We can no longer imperatively call `updateScreenBounds`, as we need to provide those bounds externally.

### Change Type

### Test Plan

1. Use the examples, including the new inset canvas example.

- [x] Unit Tests

### Release Notes

- Changes the source of truth for the viewport page bounds to be the canvas instead.


packages/editor/src/lib/hooks/useScreenBounds.ts
```diff
@@ -1,22 +1,35 @@
...
+					Math.max(rect.height, 1)
+				)
+			)
+		}
+
+		const updateBounds = throttle(updateScreenBounds, 200, {
+			trailing: true,
+		})
+
+		updateScreenBounds()
```
SomeHats, line 32: maybe out of scope of this change, but as we're doing more work on allowing the tldraw component to be used outside of full-screen contexts, we could think about using a `ResizeObserver` here too

SomeHats (approved):
Love this!

packages/editor/src/lib/hooks/useScreenBounds.ts
steveruizok: That's a good shout! I thought we did have one here. I'll add.

PR #2804, opened 2024-02-12
Add component for viewing an image of a snapshot 

TodePond:
This PR adds the `TldrawImage` component that displays a tldraw snapshot as an SVG image.

2024-02-15 at 12 29 52 - Coral Cod

## Why

We've seen requests for this kind of thing from users. eg: GitBook, and on discord:

image

The component provides a way to do that.
This PR also untangles various bits of editor state from image exporting, which makes it easier for library users to export images more agnostically. (ie: they can now export any shapes on any page in any theme. previously, they had to change the user's state to do that).

## What else

- This PR also adds an **Image snapshot** example to demonstrate the new component.
- We now pass an `isDarkMode` property to the `toSvg` method (inside the `ctx` argument). This means that `toSvg` doesn't have to rely on editor state anymore. I updated all our `toSvg` methods to use it.
- See code comments for more info.

## Any issues?

When you toggle to editing mode in the new example, text measurements are initially wrong (until you edit the size of a text shape). Click on the text shape to see how its indicator is wrong. Not sure why this is, or if it's even related. Does it ring a bell with anyone? If not, I'll take a closer look. (fixed, see comments --steve)

## Future work

Now that we've untangled image exporting from editor state, we could expose some more helpful helpers for making this easier.

Fixes tld-2122

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Open the **Image snapshot** example.
2. Try editing the image, saving the image, and making sure the image updates.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Dev: Added the `TldrawImage` component.


steveruizok:
viewport / page?

TodePond:
> viewport / page?

as props? yes that would be good. im planning to add some props for things like page, background, etc

apps/examples/src/examples/inline/README.md
```diff
@@ -5,8 +5,6 @@ category: ui
 priority: 3
 ---
 
-...
```
TodePond, line 8: unrelated cleanup

apps/examples/src/examples/snapshot-image/SnapshotImageExample.tsx
```diff
@@ -0,0 +1,74 @@
...
+[1] Pass your snapshot to the `snapshot` prop of the `TldrawImage` component.
+
+[2] You can specify which page to display by using the `pageId` prop. By
+    default, the first page is shown.
+	
+[3] You can customize the appearance of the image by passing options to the
+    `opts` prop. Try changing the `background` and `darkMode` options to see
+    their effects.
+
+ */
```
TodePond, line 74: Is this too granular? or is it ok?
cc @Taha-Hassan-Git

packages/editor/api/api.json
```diff
@@ -11134,7 +11134,7 @@
             {
               "kind": "Method",
               "canonicalReference": "@tldraw/editor!Editor#getPageShapeIds:member(1)",
-              "docComment": "/**\n * Get the ids of shapes on a page.\n *\n * @param page - The page (or page id) to get.\n *\n * @example\n * ```ts\n * const idsOnPage1 = editor.getCurrentPageShapeIds('page1')\n * const idsOnPage2 = editor.getCurrentPageShapeIds(myPage2)\n * ```\n *\n * @public\n */\n",
+              "docComment": "/**\n * Get the ids of shapes on a page.\n *\n * @param page - The page (or page id) to get.\n *\n * @example\n * ```ts\n * const idsOnPage1 = editor.getPageShapeIds('page1')\n * const idsOnPage2 = editor.getPageShapeIds(myPage2)\n * ```\n *\n * @public\n */\n",
```
TodePond, line 11137: Unrelated cleanup correction

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3296,8 +3301,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @example
 	 * ```ts
-	 * const idsOnPage1 = editor.getCurrentPageShapeIds('page1')
-	 * const idsOnPage2 = editor.getCurrentPageShapeIds(myPage2)
+	 * const idsOnPage1 = editor.getPageShapeIds('page1')
+	 * const idsOnPage2 = editor.getPageShapeIds(myPage2)
```
TodePond, line 3305: Unrelated cleanup correction

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3141,8 +3141,13 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 
-		for (const childId of this.getSortedChildIdsForParent(this.getCurrentPageId())) {
-			addShapeById(childId, 1, false)
+		// If we're using editor state, then we're only interested in on-screen shapes.
+		// If we're not using the editor state, then we're interested in ALL shapes, even those from other pages.
+		const pages = useEditorState ? [this.getCurrentPage()] : this.getPages()
+		for (const page of pages) {
+			for (const childId of this.getSortedChildIdsForParent(page.id)) {
+				addShapeById(childId, 1, false)
+			}
```
TodePond, line 3150: We're already passing the argument `useEditorState` to state whether we want to depend on editor state. But it was always dependant on the editor's state for your current page. 
This PR fixes that.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8049,8 +8054,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 			preserveAspectRatio = false,
 		} = opts
 
-		// todo: we shouldn't depend on the public theme here
-		const theme = getDefaultColorTheme({ isDarkMode: this.user.getIsDarkMode() })
+		const isDarkMode = opts.darkMode ?? this.user.getIsDarkMode()
+		const theme = getDefaultColorTheme({ isDarkMode })
```
TodePond, line 8058: We can finally remove this todo

packages/editor/src/lib/hooks/useEditor.ts
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import { Editor } from '../editor/Editor'
 
+/** @internal */
 export const EditorContext = React.createContext({} as Editor)
```
TodePond, line 5: Need this for `TldrawImage`

packages/tldraw/src/lib/TldrawImage.tsx
```diff
@@ -0,0 +1,138 @@
...
+					{editor && (
+						<EditorContext.Provider value={editor}>
+							<TldrawImageSvg pageId={props.pageId} opts={props.opts} />
+						</EditorContext.Provider>
+					)}
+				</ContainerProvider>
+			)}
+		</div>
+	)
+}
```
TodePond, line 114: too messy?
could split out into different components

packages/tldraw/src/lib/TldrawImage.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+function TldrawImageSvg({ pageId, opts }: { pageId?: TLPageId; opts?: Partial<TLSvgOptions> }) {
+	const editor = useEditor()
+	const [url, setUrl] = useState<string | null>(null)
+
+	useEffect(() => {
+		getImageUrl(editor, pageId, opts).then((url) => {
+			setUrl(url)
+		})
+	}, [editor, opts, pageId])
```
TodePond, line 124: is this ok?
whats best practice for async useMemo?
or is there a better approach im missing?

SomeHats: this is pretty much it! the one thing to add here is a way of stopping the action if/when the thing fails:
```ts
	useEffect(() => {
		let isCancelled = false
		getImageUrl(editor, pageId, opts).then((url) => {
			if (isCancelled) return
			setUrl(url)
		})
		return () => {
			isCancelled = true
		}
	}, [editor, opts, pageId])

TodePond: thanks, done!

apps/examples/src/examples/snapshot-image/SnapshotImageExample.tsx
Taha-Hassan-Git: Granularity is ok! Something I've been doing with these annotations is when a concept comes up that isn't directly relevant to the example, I've given a quick overview and pointed them to docs or a more relevant example so they can explore it in detail. You're using quite a cool trick to access the editor from outside of tldraw, and while it shouldn't be the focus, it's def worth mentioning imo.

You could:

- Find a way to show this concept without the extra complexity, and save this for an editor api example.
- Annotate and explain the remaining code step by step. (in which case I'd bundle these 3 comments together into one)
- Use a single annotation to give a brief overview of the remaining code.
- Have something in the introduction to explain where most of the relevant code is, and what the rest of it does.

steveruizok:
Riding in some changes here Lu:
- rather than opts, use the SVG options as individual props
- rather than keeping the editor in state, create the editor only in order to get the svg, then dispose it right away
- make a slightly more complex snapshot example
- have the example also keep track of background / dark mode

steveruizok:
There seems to be a problem with text measurement, not sure if this is particular to this project or to tldraw in general.

steveruizok:
Kind of an unusual add here, but we didn't have the ability to zoom to a specific bounds without also adding an inset. I've combined the inset property with the animations object on `zoomToBounds`.

steveruizok:
Added format option for SVG or PNG.

steveruizok:
Text measurements are wrong because the container is not getting tldraw's styles mounted to it.

TodePond:
nice one, thanks for the changes Steve. Looks good

PR #2813, opened 2024-02-13
Per shape styles for next shape

MitjaBezensek:

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -761,7 +761,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     getSortedChildIdsForParent(parent: TLPage | TLParentId | TLShape): TLShapeId[];
     getStateDescendant<T extends StateNode>(path: string): T | undefined;
     // @internal (undocumented)
-    getStyleForNextShape<T>(style: StyleProp<T>): T;
+    getStyleForNextShape<T>(style: StyleProp<T>, shapeType: string): T;
     getSvg(shapes: TLShape[] | TLShapeId[], opts?: Partial<TLSvgOptions>): Promise<SVGSVGElement | undefined>;
     getViewportPageBounds(): Box;
     getViewportPageCenter(): Vec;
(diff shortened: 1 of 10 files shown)


ds300:
In terms of UX I quite like this. The toolbar hiding feels fine. I would be sad that the tldraw UI would become less colorful in its default state (on desktop), but I think most people are gonna click a shape tool pretty quickly after landing on tldraw for the first time, and they are maybe even gonna get a little dopamine hit from the colorful toolbar popping up as a result of something they did? Either way, after a few minutes of playing around trying to draw diagrams, it certainly feels that per-shape styles work a lot better, i.e. they involve less futzing with styling.

The button active states weren't working for me, and that might change how it feels? I expect it would feel even better.

MitjaBezensek:
The active button states is a bug that was fix in https://github.com/tldraw/tldraw/pull/2862. Let me updated this branch.

ds300:
I think the main situation where this is worse than how it was before is if you are inclined to switch color before you switch tool. e.g. the thought "i want a red arrow" translating into "pick red" -> "pick arrow". We could probably retain that feature on top of this without ruffling too many feathers though?

e.g. something like

- when you change a style prop we mark it as being 'dirty'
- when you create a new shape that uses the style prop it is set to 'clean'
- when you switch tool, any 'dirty' style props are carried over into the new tool.

MitjaBezensek:
Yeah. If we go with per shape styles I kinda see two options:
1. Hide the style panel for select tool if no shapes are selected. In general I really like this UX, but it does come with two drawbacks:
- You can't first apply style changes and then select the tool (your red arrow example).
- We loose the style panel being visible when you open tldraw. This is worse for branding. Also a bit worse UX for the first run experience. But I think people switch to some shape tool the first thing they do and then then the style panel appears so probably not that big of a deal 🤷‍♂️ .
2. We still show the style panel but we go with the dirty state option, which only gets applied on the next shape creation. 
Like I mentioned I don't like that you can't pick all the style options with this one. Like if I want a red arrow that's fine. But what if I want a red arrow with different arrow endings? Then I either adjust everything after creating the arrow or I adjust the color before creating it and then arrow ends after creating it.

ds300:
Closing due to inactivity

Issue #2720, opened 2024-02-04
Read only mode tweaks


Taha-Hassan-Git:
### What's the feature?

I was using tldraw over the weekend to share resources for a meetup I organised. I had put a bunch of links on a project and shared the read-only version of it, but no one could actually visit the links. Also people were having trouble navigating around the board because the select tool was active. So if they tried to scroll over a shape, the blue indicator would flicker and the screen wouldn't move. 

I think you should be able to follow links in read only mode and it should default to the hand tool.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2819, opened 2024-02-13
Read Only Mode tweaks

Taha-Hassan-Git:
#2720 
I was using tldraw for a meetup last week. I used read only mode to share a bunch of links and QR codes. People found it difficult to navigate around the page because it defaulted to the select tool, and had trouble following links.

This PR makes readonly mode default to the hand tool, and makes the entire bookmark shape a link. 

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Make a bookmark shape
2. Open the project in Read only mode
3. It should default to the hand tool 
4. The entire bookmark shape should be a link

### Release Notes

- Read only mode defaults to the hand tool
- The entire bookmark shape is a link


steveruizok:
Hand tool is a good default for multiplayer rooms! Let's make this PR only that change. We can't make the bookmark shape all a link but maybe we can improve the linkiness of it.

Taha-Hassan-Git:
> Hand tool is a good default for multiplayer rooms! Let's make this PR only that change. We can't make the bookmark shape all a link but maybe we can improve the linkiness of it.

Yeah I only realised how challenging making it a link was after I put the PR up and saw how buggy it was, spent a while poking around in the different tool states, but I got stumped.

PR #2820, opened 2024-02-13
fix cursor update frequency

SomeHats:
Previously, no-op changes to `setCursor` would still trigger full signia updates and network diffs. This would happen on every pointer move event even though the change wasn't actually doing anything. This is because we were using `{...original, ...update}` style updates, which produce a fresh object regardless of whether anything actually changed. This diff introduces an `applyPartial` helper which does the same thing as `{...original, ...update}`, but if the result is the same as `original` (according to `Object.is` on each key in the partial), we return the unchanged `original` instead.

See https://github.com/tldraw/tldraw/pull/2792#issuecomment-1936327594 for details

### Change Type

- [x] `patch` — Bug fix

### Test Plan

- [x] Unit Tests

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -41,6 +41,7 @@ import {
 import {
 	JsonObject,
 	annotateError,
+	applyPartial,
 	assert,
 	compact,
 	dedupe,
@@ -1179,7 +1180,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 **/
 	updateDocumentSettings(settings: Partial<TLDocument>): this {
-		this.store.put([{ ...this.getDocumentSettings(), ...settings }])
+		this.store.put([applyPartial(this.getDocumentSettings(), settings)])
 		return this
 	}
 
@@ -1229,7 +1230,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 			historyOptions?: TLCommandHistoryOptions
 		) => {
 			const prev = this.store.get(this.getInstanceState().id)!
-			const next = { ...prev, ...partial }
+			const next = applyPartial<TLInstance>(prev, partial)
 
 			return {
 				data: { prev, next },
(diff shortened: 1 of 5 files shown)


steveruizok:
Actually let's pull that logic entirely, it's not the right behavior anyway (pointer on hover?).

steveruizok:
lgtm, separate PR here to remove this behavior: https://github.com/tldraw/tldraw/pull/2824

ds300 (approved):
This is a good safety valve for sure, and I'm happy to merge. 

Still, our cursor does need to change on every frame during rotation interactions. So to alleviate pressure on other things that reference the instance state, we can move it to it's own record type at some point. Probably no rush though.

steveruizok:
> We don't make these types of "did anything actually change" checks / guards for other methods.

@ds300 thoughts on ^^?

ds300:
> @ds300 thoughts on ^^?

I'm actually wanting to do it in the store.put method. In the early early days of tldraw-lite I kinda wanted to do it then too but refrained because it adds a tiny bit of overhead. But then we added validators for everything which crawl everything anyway so there's plenty of overhead now.

There's an interesting possibility to actually reduce the current level of validations overhead while also adding this equalty-checking behavior similar to alex's thing here. We can do it by crawling the new records using the previous (validated) version as a comparison.

Here's a really dumb pseudocode version of what i mean

```ts
// 'invalid' is not an option because validators just throw if it's invalid
type ValidationResult = 'unchanged' | 'valid'

class ObjectValidator implements Validator {
	validators: Record<string, Validator>
	validateAndCheck(prevVersion, newVersion): ValidationResult {
		let didChange = false
		for (const [key, val] of Object.entries(newVersion)) {
			// if the val is same as previously, we don't need to validate it
			if (Object.is(prevVersion[key], val)) {
				continue
			}
			const validationResult = this.validators[key].validateAndCheck(prevVersion[key], val)

			if (validationResult !== 'unchanged') {
				didChange = true
			}
		}

		if (didChange) {
			return 'valid'
		}
		return 'unchanged'
	}
}

```

steveruizok:
If it can be handled in a single place, then I'd prefer that!

SomeHats:
Seems like there's not much appetite for this specific change, closing for now

PR #2824, opened 2024-02-13
Remove pointer check for arrow labels

steveruizok:
This PR removes a check for pointer changes over arrow labels. We don't really have any other uses of the `pointer` in the canvas except for buttons, so I'd rather keep the default cursor for arrow labels.

### Change Type

- [x] `minor` — New feature


mimecuvalo:
¯\_(ツ)_/¯ okie dokie

PR #2825, opened 2024-02-13
textfields [3 of 3]: add custom key triggers to textlabels; emojis as first test case! 🧑‍🎨 🎨 ✏️

si14:
This PR started as an emoji and has been reworked to follow the composable UI pattern of providing components.
Now we add an insertion point for certain hooks to be called upon within our codebase. The first of these is the `useTextTriggerCharacter` so that we can enable emojis on the dotcom codebase.

🎵 this is dangerous, I walk through textfields so watch your head rock 🎵

Original original description:
This is a copy of the accidentally merged #2814 by @mimecuvalo. Original description:

everyone ❤️'s emojis: https://dropbox.tech/application/dropbox-paper-emojis-and-exformation

https://github.com/tldraw/tldraw/assets/469604/8f99f485-de98-44d1-93cb-6eb9c2d87d99

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Test adding lots of emojis!

### Release Notes

- Adds hook into hooks to allow logic insertion points into the editor.
- Adds emoji picker to text fields.


packages/tldraw/src/lib/shapes/shared/useEditableText.ts
```diff
@@ -207,3 +236,46 @@ export function useEditableText<T extends Extract<TLShape, { props: { text: stri
 		isEmpty,
 	}
 }
+
+function getCaretPosition(editor: Editor, inputEl: HTMLTextAreaElement | null) {
```
SomeHats, line 240: instead of all the manual geometry & text measurement here, we could use some browser-native APIs:
```ts
// browser doesn't let us measure ranges inputs for some reason, but we have
// the `text-content` el which we can use. probs safer to pass this as a ref tho
const measureEl = inputEl.previousSibling

// there should only be one child - the text node 
const measureNode = measureEl.childNodes[0]

// the range API is awkward but lets u do cool stuff:
const range = document.createRange()
range.setStart(measureNode, inputEl.selectionStart)
range.setEnd(measureNode, inputEl.selectionEnd)
const rangeBounds = range.getBoundingClientRect()

// lets say cursor position is bottom-center
const cursorScreenPos = new Vec(rangeBounds.bottom, rangeBounds.left + rangeBounds.width / 2)
const cursorViewportPos = Vec.Add(editor.getViewportScreenBounds(), cursorScreenPos)

return cursorViewportPos

File: packages/tldraw/src/lib/shapes/shared/useEmojis.ts [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+		if (!inputEl) return
+
+		const searchText = EmojiDialogSingleton?.component.refs.searchInput.current.value
+		inputEl.focus()
+		inputEl.setSelectionRange(
+			inputEl.selectionStart - searchText.length - 1,
+			inputEl.selectionStart
+		)
+		inputEl.setRangeText(emoji.native)
+		inputEl.setSelectionRange(inputEl.selectionStart + 1, inputEl.selectionStart + 1)
```
SomeHats, line 27: you could `inputEl.setRangeText(emoji.native, undefined, undefined, 'end')` if u wanted

packages/tldraw/src/lib/shapes/shared/useEmojis.ts
```diff
@@ -0,0 +1,125 @@
...
+						},
+					})
+					e.preventDefault()
+					return true
+				}
+
+				return false
+			}
+
+			case 'Backspace':
```
SomeHats, line 96: I don't love us effectively implementing manual text editing here... there's a lot missing (copy-paste, delete-key, etc).

In particular, I'd be curious about how this works with non-keyboard input methods (dictation, mobile pinyin keyboards, etc)

SomeHats: An alternative here would be to mark the point at the start of the input and pull the latest range from there to the current caret? might have some issues with concurrent editing, but we don't officially support that anyway so

SomeHats (commented):
I'm a bit unsure about this change from a product perspective. I think it's nice for tldraw.com, but I don't think it should be a part of the core library right now.

I also think that the way we're trying to implement this at the moment is a bit fragile/hacky. One thing we've been talking about for a while is a proper rich-text editor, built on a real text-editing framework like prosemirror. Frameworks like that have much better solutions to some of the things we're trying to do here (e.g. @-mention style inline popups in a way that works with all input methods & concurrent edits).

For me, this feature is one I'd like, but I don't know if hacking it onto our plain text labels is the right way forward. I'd rather we tackle rich text editing as a proper project, and add this as part of that.

Alternatively, if there's a way to add this as part of dotcom that doesn't change the core library, then I'd be much happier shipping this sort of approach

mimecuvalo:
> Alternatively, if there's a way to add this as part of dotcom that doesn't change the core library, then I'd be happy to ship that

This sounds good to me! I'll rework this to make it a dotcom thing, or at least something that can be plugged in somehow — will have to think about this 🤔 

> For me, this feature is one I'd like, but I don't know if hacking it onto our plain text labels is the right way forward. I'd rather we tackle rich text editing as a proper project, and add this as part of that.

I hear that but definitely wouldn't want to make rich-text editing a prerequisite for this. I would liken this to how you can do `:` inside a GitHub comment — for that reason, I would disagree that it's fragile/hacky to have this in a plain text field — it's being used as such in the very thing we're communicating in currently! :D So, rich-text editing is orthogonal to this feature. 

I'd be curious, separately, to talk about the rich-text editing stuff — I'm not sure how much of that we would desire longterm in the product — it'd be a potentially heavy thing to ship with the product vs. something more lightweight like a Markdown-based quick render thingy.

SomeHats:
> I would disagree that it's fragile/hacky to have this in a plain text field
I agree! The fragility/hackiness was more about the way we're attempting to sync the search with a special hidden field by re-implementing part of text editing. I meant we should find a more robust, that also we know won't e.g. crash on concurrent edits. These sorts of cases tend to be handled by text editing frameworks in a way we wouldn't need to worry about so much.

packages/editor/src/lib/primitives/geometry/Group2d.ts
```diff
@@ -70,6 +70,10 @@ export class Group2d extends Geometry2d {
 		return this.children[0].area
 	}
 
+	getLabel() {
+		return this.children.filter((c) => c.isLabel)[0]
+	}
+
```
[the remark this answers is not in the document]
mimecuvalo: Ah, good point — I changed this to be `findLabels` and then the caller can figure out the best option for now.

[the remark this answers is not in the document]
mimecuvalo: But I'd imagine a better signature down the line, like, `getActiveLabel`

packages/tldraw/src/lib/shapes/shared/useEmojis.ts
```diff
@@ -0,0 +1,125 @@
+import { useState } from 'react'
+import { useDialogs } from '../../ui/hooks/useDialogsProvider'
+import EmojiDialog from './emojis'
+import { EmojiDialogSingleton } from './emojis/EmojiDialog'
+
+export function useEmojis(inputEl: HTMLTextAreaElement | null, onComplete: (text: string) => void) {
+	const { addDialog, removeDialog } = useDialogs()
```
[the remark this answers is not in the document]
mimecuvalo: Yeah, given our offline conversation about separating out text labels to be swappable in general, I think that when I rework this PR to start doing that, we can avoid using the internal dialog management altogether.

packages/tldraw/src/lib/shapes/shared/emojis/EmojiDialog.tsx
```diff
@@ -0,0 +1,68 @@
...
+	const ref = useRef(null)
+	const instance = useRef<any>(null)
+
+	useEffect(() => {
+		const eventListener = (event: TLEventInfo) => {
+			if (event.name === 'pointer_down') {
+				onClickOutside()
+			}
+		}
+		editor.on('event', eventListener)
```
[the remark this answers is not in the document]
mimecuvalo: hmm, I'm doing disposable right (line 45) below unless I'm missing something else? I checked that it's unmounting in the scenarios you describe and it seems to be operating as expected.

packages/tldraw/src/lib/shapes/shared/emojis/EmojiDialog.tsx
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}, [editor, theme.id, onEmojiSelect, onClickOutside])
+
+	return (
+		<div
+			ref={ref}
+			style={{
+				position: 'absolute',
+				top,
+				left,
```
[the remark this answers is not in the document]
mimecuvalo: yeah, definitely this first pass doesn't do the best job of making that dialog not be in weird positions (and mobile makes it worse). we can definitely account for this (with things like the floating-ui library you mention above)

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
mimecuvalo: haha, that's funny - that _was_ my first version of how I wrote this (i'm very familiar with the range API from before heh). i was trying to think ahead to this review and thought you might all ask me to not rewrite logic that existed in `editor.textMeasure.measureText` 😛 

again, especially after offline conversation of making the text labels more swappable in general, the more general logic fits that paradigm much better — i'll rework to use the general range API.

packages/tldraw/src/lib/shapes/shared/useEmojis.ts
mimecuvalo: hm, i see what you're going for but that overloaded sig. doesn't allow undefined and even if i rework it with the correct values, i think splitting it out is a little clearer on the intent.

packages/tldraw/src/lib/shapes/shared/useEmojis.ts
mimecuvalo: agree to disagree here. i think you're overthinking it :)
i'm obviously borrowing the (exact) mechanics of Dropbox Paper's emoji picker which "steals" the keyboard events whilst being open.

i'm definitely aware of complexities around concurrent and IME (Input Method Editors) but that's out-of-scope for this feature set to be honest. those input methods are just separate beasts altogether :)

mimecuvalo:
>  The fragility/hackiness was more about the way we're attempting to sync the search with a special hidden field by re-implementing part of text editing.

meh, i mean we'll prbly re-implement our own picker eventually (it's sort of a pain, having done it before, and we'd have to maintain it to be up-to-date with Unicode updates).
to the second half of your point — i talked about that above as far as re-implementing text editing and we'll have to agree to disagree, i mentioned that that's overthinking it a bit.

> I meant we should find a more robust, that also we know won't e.g. crash on concurrent edits. 
to be fair, the concurrent case is overthinking this a bit much, for where this PR is trying to just having a starting ground. we don't support concurrent text editing anyway, and the concurrent case of the shape being unmounted or something like that will just be handled by closing the dialog which is the expected behavior.

thanks for the review! i'll be reworking this as an entrypoint into start looking into plumbing swappable text fields in general.

SomeHats:
Before i dive in to the code properly, i wanted to ask about the approach here - from #3051, I expected us to use the text label override system to implement this by replacing the text labels with ones that use the emojis. That seems a lot more flexible & open-ended than this hooks based approach. What's the thinking here? Do we need both these approaches?

mimecuvalo:
> Before i dive in to the code properly, i wanted to ask about the approach here - from https://github.com/tldraw/tldraw/pull/3051, I expected us to use the text label override system to implement this by replacing the text labels with ones that use the emojis. That seems a lot more flexible & open-ended than this hooks based approach. What's the thinking here? Do we need both these approaches?

Yeah, let's talk more offline, but briefly, I think that we need both. There's custom text rendering but, also, there's just being able to say: "tldraw's text editor is good enough, I just want to be able to hook into it sometimes". I don't if you necessarily want to swap out the `DefaultTextLabel` necessarily because then you also have to reimplement `useEditableText`. We could just force downstream developers to reimplement everything any time they 'eject' and go custom, but I think there are some tooling we can provide to help them avoid doing that (and then worrying that by ejecting they will more easily get broken by us).  

I'll put time on your cal to talk more - or we'll chat tomorrow on the 🚋

apps/dotcom/sentry.client.config.ts
```diff
@@ -22,7 +22,7 @@ Sentry.init({
 	tracesSampleRate: 1.0,
 	release: sentryReleaseName,
 	environment: env,
-	integrations: [new ExtraErrorData({ depth: 10 }) as any],
+	integrations: [extraErrorDataIntegration({ depth: 10, captureErrorCause: true })],
```
mimecuvalo, line 25: fyi, this is b/c i had to run `yarn dedupe` and then this necessitated this followup change

mimecuvalo:
superseded by https://github.com/tldraw/tldraw/pull/5124

PR #2827, opened 2024-02-13
[Snapping 1/5] Validation & strict types for fractional indexes 

SomeHats:
Currently, we type our fractional index keys as `string` and don't have any validation for them. I'm touching some of this code for my work on line handles and wanted to change that:
- fractional indexes are now `IndexKey`s, not `string`s. `IndexKey`s have a brand property so can't be used interchangeably with strings (like our IDs)
- There's a new `T.indexKey` validator which we can use in our validations to make sure we don't end up with nonsense keys.

1. #2827 (you are here)
2. #2831
3. #2793
4. #2841
5. #2845

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Mostly relying on unit & end to end tests here - no user facing changes.

- [x] Unit Tests


steveruizok:
I like moving these into the utils package. I don't think I've ever seen an error related to bad strings as indexes, did you run into a case where that was an issue?

packages/tlschema/src/TLStore.ts
```diff
@@ -81,7 +81,12 @@ export const onValidationFailure: StoreSchemaOptions<
 
 function getDefaultPages() {
 	return [
-		PageRecordType.create({ id: 'page:page' as TLPageId, name: 'Page 1', index: 'a1', meta: {} }),
+		PageRecordType.create({
+			id: 'page:page' as TLPageId,
+			name: 'Page 1',
+			index: 'a1' as IndexKey,
```
ds300, line 87: nonblocking nitpick
```suggestion
			index: ZERO_INDEX_KEY,
```

packages/tldraw/src/test/commands/setCurrentPage.test.ts
```diff
@@ -39,15 +39,15 @@ describe('setCurrentPage', () => {
...
+		const page = PageRecordType.create({ name: 'test', index: 'a4' as IndexKey })
 		expect(editor.getPageStates().find((p) => p.pageId === page.id)).toBeUndefined()
 		editor.store.put([page])
 		expect(editor.getPageStates().find((p) => p.pageId === page.id)).not.toBeUndefined()
 	})
 
 	it('squashes', () => {
 		const page2Id = PageRecordType.createId('page2')
-		editor.createPage({ name: 'New Page 2', index: page2Id })
+		editor.createPage({ name: 'New Page 2', id: page2Id })
```
ds300, line 50: 👏🏼

ds300 (approved):

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4885,12 +4884,12 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 	 * @public
 	 */
-	getHighestIndexForParent(parent: TLParentId | TLPage | TLShape): string {
+	getHighestIndexForParent(parent: TLParentId | TLPage | TLShape): IndexKey {
 		const parentId = typeof parent === 'string' ? parent : parent.id
 		const children = this._parentIdsToChildIds.get()[parentId]
 
 		if (!children || children.length === 0) {
-			return 'a1'
+			return 'a1' as IndexKey
```
mimecuvalo, line 4892: can this be `ZERO_INDEX_KEY` too? 🤔

packages/tlsync/src/lib/TLSyncRoom.ts
```diff
@@ -234,7 +235,7 @@ export class TLSyncRoom<R extends UnknownRecord> {
 						lastChangedClock: 0,
 					},
 					{
-						state: PageRecordType.create({ name: 'Page 1', index: 'a1' }),
+						state: PageRecordType.create({ name: 'Page 1', index: 'a1' as IndexKey }),
```
mimecuvalo, line 238: similar to David's comment `TLStore` - can this be `ZERO_INDEX_KEY`? if not, i'd love to understand the difference.

SomeHats:
re: replacing `'a1'` with `ZERO_INDEX_KEY`, and earlier version of this PR did that, but it produced some test failures. I figured I'd leave them as-is for now because although they're probably all fine, I didn't want to spend the time to fix them all and leaving as-is seemed low risk. 

im probably gonna leave this as-is for now, but I'd like to clean these up as a follow-up when i have more time to properly understand those test failures.

PR #2680, opened 2024-01-29
New migrations

ds300:
This PR aims to implement the ideas discussed in #2359

At the time of opening this is still very wip and needs a good amount of testing, backwards-compat work, and documentation.

I'm gonna try super hard to avoid making any truly breaking changes, and to have a clear upgrade path. Might not be totally doable though. At the very least I'll try to make any breakage very easy to remedy.

TODO
- [x] Implement new migrations system and API
- [x] Get dev env and tests working
- [x] Audit usages of schema.migrate*
- [x] Test against old documents to make sure everything migrates correctly
- [ ] Memoize getMigrationsSince?
- [x] Write more unit tests for new bits
- [x] Make sure user-facing errors are helpful
- [ ] ~Add snapshot test for schema to make sure migrations are added~ - too yak shavey it turns out
- [x] Write up user-facing documentation
- [ ] Update tsdoc comments to point to docs
- [ ] Clean up TODOs
- [ ] Communicate changes with tldraw team

### Change Type

- [x] `major` — Breaking change

[^1]: publishes a `patch` release, for devDependencies use `internal`
[^2]: will not publish a new version

### Test Plan

1. Add a step-by-step description of how to test your PR here.
2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Add a brief release note for your PR here.

PR #2828, opened 2024-02-13
[docs] Fix missing Persistence page

### LINKED ISSUE #2826: [docs]: Persistence page is not showing
we have a bug in our addContent strategy that will overwrite a page if another page is added later with the same id. This is currently affecting our `persistence.mdx` page which is being overwritten by the `persistence/` example.

ds300:
Fixes #2826, extracted from #2680 

The problem is that we had two different articles whose ids were being derived as `persistence`, the `persistence.mdx` file and the `persistence/` example.

I've

1. Made it an error for two articles to have the same id
2. Renamed the `persistence/` article to `local-storage` since that's the API it's using.

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

3.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


apps/docs/utils/addContent.ts
```diff
@@ -8,19 +8,19 @@ export async function addContentToDb(
 	content: GeneratedContent
 ) {
 	const sectionInsert = await db.prepare(
-		`REPLACE INTO sections (id, idx, title, description, path, sidebar_behavior) VALUES (?, ?, ?, ?, ?, ?)`
+		`INSERT INTO sections (id, idx, title, description, path, sidebar_behavior) VALUES (?, ?, ?, ?, ?, ?)`
```
steveruizok, line 11: Will these changes mean the db will have an error if the id is already there?

steveruizok (approved):

apps/docs/utils/addContent.ts
ds300: yeah, but it clears the db before running so if there's something there already it's overwriting a different bit of content and is therefore an error we want to throw.

steveruizok:
I vaguely remember choosing REPLACE instead of INSERT in order to avoid duplicated items in the database when editing files, but I can't reproduce any of those things, so we might as well land this one and then maybe keep an eye out?

PR #2831, opened 2024-02-13
[Snapping 2/5] Fix line-handle mid-point snapping

SomeHats:
Currently, only the end handles of the line tool snap. It should be all of them.

Line handles work kind of weirdly at the moment: instead of just storing the positions, we store full `TLHandle` objects complete with IDs, `canSnap`/`canBind` properties, etc. Currently, all the handles get written to the store with `canSnap: false`, when really it should be up to the shape util to decide which handles are snappable.

This diff replaces the current handles map (from arbitrary ID to `TLHandle`) with just the data we need: a map from index to x/y. The extra information that the `Editor` needs for `TLHandle` is hydrated at runtime (with `canSnap` set to `true` this time!)

Fixes TLD-2200

1. #2827 
2. #2831 (you are here)
3. #2793
4. #2841
5. #2845

### Change Type

### Test Plan

1. Create a funky line shape on tldraw.com
2. Paste it into staging and make sure it comes across ok
3. Make some funky line shape in staging - make sure you use dragging, mid-point creation, and shift-clicking

- [x] Unit Tests

### Release Notes

- Simplify the contents of `TLLineShape.props.handles`

### DIFF
--- a/packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
@@ -1,7 +1,6 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 import {
 	CubicSpline2d,
-	IndexKey,
 	Polyline2d,
 	SVGContainer,
 	ShapeUtil,
@@ -14,8 +13,10 @@ import {
 	deepCopy,
 	getDefaultColorTheme,
 	getIndexBetween,
+	getIndices,
 	lineShapeMigrations,
 	lineShapeProps,
+	objectMapEntries,
 	sortByIndex,
 } from '@tldraw/editor'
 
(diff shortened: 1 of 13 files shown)


ds300 (commented):
🎉 so much simpler, nice one!

ds300 (approved):

mimecuvalo (approved):
👍

steveruizok:
Great PR. Having that extra / redundant data in the handle prop was definitely an accident of evolution.

packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
```diff
@@ -45,27 +46,18 @@ export class LineShapeUtil extends ShapeUtil<TLLineShape> {
...
 			color: 'black',
 			spline: 'line',
 			handles: {
-				start: {
-					id: 'start',
-					type: 'vertex',
-					canBind: false,
-					canSnap: true,
-					index: 'a1' as IndexKey,
+				[startIndex]: {
```
steveruizok, line 56: Why is it necessary to use indices here at keys? I would expect the handles to just be an array of VecModels. We use indices specifically to avoid cases where changing one record's place in order would mean changing other records, too. (If a shape's index was an integer, then moving a shape to the back of its parent's children would mean also incrementing all of its siblings' indices). That doesn't matter for records because we update "whole records", so the diff would be the same.

SomeHats: just because doing all that array/index juggling is kinda fiddly tbh? fractional indexes are just a bit nicer for these types of things imo. i could change it but it would make how we update this a fair bit more complicated so i'd rather not

PR #2833, opened 2024-02-14
Readonly defaults to the hand tool

Taha-Hassan-Git:
#2720 
This PR makes it so that the editor defaults to the hand tool in read only mode

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Open the editor in readonly mode
2. It should default to the hand tool

### Release Notes

- Shared projects in  read only mode now default to the hand tool


steveruizok (approved):

steveruizok:
Should apply also to snapshots in a separate PR

PR #2841, opened 2024-02-14
[Snapping 4/5] Add handle-point snapping

SomeHats:
Currently, when dragging line handles they'll snap to the outlines of other shapes, but not to their vertices. This can make it hard to snap precisely to certain key places, like the handles of other lines, or the corners of `geo` shapes.

This diff adds a new snap type for handles - snapping to points:

Kapture 2024-02-14 at 16 30 41

This adds to the new snapping API so the snapping points can very easily be customised on a shape-by-shape basis. Closes TLD-2198

1. #2827 
2. #2831
3. #2793 
4. #2841 (you are here)
5. #2845

### Change Type

- [x] `minor` — New feature

### Test Plan

1. create a line shape
2. drag its handles whilst holding command
3. it should snap to the outlines of other shapes, vertices of other line shapes, and the bounding box corners/center of most 'boxy' shapes (geo, embed, etc)

- [x] Unit Tests

### Release Notes

- Line handles


packages/editor/src/lib/editor/managers/SnapManager/HandleSnaps.ts
```diff
@@ -52,23 +61,45 @@ export class HandleSnaps {
...
+					const snapPointInPageSpace = shapePageTransform.applyToPoint(snapPointInShapeSpace)
+					const distance = Vec.Dist(handlePoint, snapPointInPageSpace)
+
+					if (distance < minDistanceForSnapPoint) {
+						minDistanceForSnapPoint = distance
+						nearestSnapPoint = snapPointInPageSpace
+					}
+				}
 			}
 		}
```
ds300, line 104: Nonblocking nitpick:

Snap point hits always take precedence over outline hits, regardless of which is closer. So i think you could do the snap point checking first, then if a hit was found you can avoid doing the outline checking (including the `additionalSegments` bit below).

ds300 (approved):
More great work 🥇

packages/editor/src/lib/editor/managers/SnapManager/HandleSnaps.ts
mimecuvalo: yeah, to make this clearer, what do you think of something like this?
```ts
const pointsToTest = [...snapGeometry.points ?? []]
for (const shapeId of this.manager.getSnappableShapes()) {
	...
	if (snapGeometry.outline) {
		const nearestShapePointInShapeSpace = snapGeometry.outline.nearestPoint(pointInShapeSpace)
		pointsToTest.push(nearestShapePointInShapeSpace)
	}
}
for (const segment of additionalSegments) {
	const nearestOnSegment = Vec.NearestPointOnLineSegment(segment[0], segment[1], handlePoint)
	pointsToTest.push(nearestOnSegment)
}
for (const point of pointsToTest) {
	const snapPointInPageSpace = shapePageTransform.applyToPoint(point)
	const distance = Vec.Dist(handlePoint, snapPointInPageSpace)

	if (distance < minDistanceForSnapPoint) {
		minDistanceForSnapPoint = distance
		nearestSnapPoint = snapPointInPageSpace
	}
}
```

related to this, just checking: does the additionalSegments portion need to also do `shapePageTransform.applyToPoint`?

SomeHats: good idea! reworked this a bit

> related to this, just checking: does the additionalSegments portion need to also do shapePageTransform.applyToPoint?
nope, these come in in page space

PR #2844, opened 2024-02-14
errors: improve msg in dialog when error happens

mimecuvalo:
Right now it's hard to parse the stack trace to see what the problem is. This just raises the msg up to a first-class position to get the gist of the problem.

Screenshot 2024-02-14 at 17 09 55

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Improves error dialog messaging.


MitjaBezensek (approved):
Loving these small improvements you keep making 👍

PR #2845, opened 2024-02-14
[Snapping 5/5] Better handle snapping for geo shapes

SomeHats:
Currently, geo shapes have slightly janky handle-snapping: they snap to label geometry (even though its invisible) and because they extend from `BaseBoxShapeUtil` they snap to the corners of their bounding box (even if that's not where the actual shape is).

With this PR, we no longer snap to labels, and we snap to the actual vertices of the geo shape rather than its bounding points.

1. #2827
2. #2831
3. #2793
4. #2841
5. #2845 (you are here)

### Change Type

- [x] `minor` — New feature

### Test Plan
- [x] Unit Tests

### Release Notes

- You can now snap the handles of lines to the corners of rectangles, stars, triangles, etc.

### DIFF
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
@@ -7,6 +7,7 @@ import {
 	Group2d,
 	HALF_PI,
 	HTMLContainer,
+	HandleSnapGeometry,
 	PI2,
 	Polygon2d,
 	Polyline2d,
@@ -21,6 +22,7 @@ import {
 	TLShapeUtilCanvasSvgDef,
 	Vec,
 	VecLike,
+	exhaustiveSwitchError,
 	geoShapeMigrations,
 	geoShapeProps,
 	getDefaultColorTheme,
@@ -89,7 +91,7 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 		}
 	}
 
-	override getGeometry(shape: TLGeoShape): Geometry2d {
+	override getGeometry(shape: TLGeoShape) {
 		const w = Math.max(1, shape.props.w)
 		const h = Math.max(1, shape.props.h + shape.props.growY)
 		const cx = w / 2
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -339,6 +341,39 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
...
+			case 'diamond':
+			case 'hexagon':
+			case 'octagon':
+			case 'pentagon':
+			case 'rectangle':
+			case 'rhombus':
+			case 'rhombus-2':
+			case 'star':
+			case 'trapezoid':
+			case 'triangle':
```
ds300, line 363: extremely nonblocking:

triangle might want a special case for its center point since it's bounding box is not center aligned with the triangle itself

Kapture 2024-02-15 at 11 41 03

ds300 (approved):

ds300:
feels really good 💆🏼

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -339,6 +341,39 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 		})
 	}
 
+	override getHandleSnapGeometry(shape: TLGeoShape): HandleSnapGeometry {
+		const geometry = this.getGeometry(shape)
+		// we only want to snap handles to the outline of the shape - not to its label etc.
+		const outline = geometry.children[0]
+		switch (shape.props.geo) {
```
mimecuvalo, line 348: non-blocking, for a future refactor:
i'm thinking we should better type this because right now it feels like a runtime error if you forget to add a new shape here (and in `getGeometry` in this file)

something like:
```
enum GeoShapes {
	cloud = 'cloud',
	rectangle = 'rectangle',
	ellipse = 'ellipse',
	...
}

const actions: {[key in GeoShapes]: () => void} = {
	cloud: () => {
		...
	},
	triangle: () => {
		...
	},
	...
}
```

this way TS will give an error saying the object is non-exhaustive and will fall at compile time

mimecuvalo: er, i meant `actions` for `getGeometry` - prbly something like `geometryForShape` instead in this instance

ds300: @mimecuvalo the switch statement has exhaustive checking already. it's a common TS pattern where you add a `default` clause that tries to use a `never` type and fails if the type is not `never`. see `exhaustiveSwitchError`

SomeHats: it would be cool to centralise the logic for each of these shape types a bit - they're very spread out and hard to track down right now. but yeah, this will already give a compile time error in the case you describe

mimecuvalo: @ds300 ahhh i hadn't noticed that — great!
@SomeHats good point as well — it is hard to track down usage of things via string though

PR #2848, opened 2024-02-15
Faster validations + record reference stability at the same time

ds300:
This PR adds a validation mode whereby previous known-to-be-valid values can be used to speed up the validation process itself. At the same time it enables us to do fine-grained equality checking on records much more quickly than by using something like lodash isEqual, and using that we can prevent triggering effects for record updates that don't actually alter any values in the store.

Here's some preliminary perf testing of average time spent in `store.put()` during some common interactions

| task | before (ms) | after (ms) |
| ---- | ---- | ---- |
| drawing lines | 0.0403 | 0.0214 |
| drawing boxes | 0.0408 | 0.0348 |
| translating lines | 0.0352 | 0.0042 |
| translating boxes | 0.0051 | 0.0032 |
| rotating lines | 0.0312 | 0.0065 |
| rotating boxes | 0.0053 | 0.0035 |
| brush selecting boxes | 0.0200 | 0.0232 |
| traversal with shapes | 0.0130 | 0.0108 |
| traversal without shapes | 0.0201 | 0.0173 |

**traversal** means moving the camera and pointer around the canvas

#### Discussion

At the scale of hundredths of a millisecond these .put operations are so fast that even if they became literally instantaneous the change would not be human perceptible. That said, there is an overall marked improvement here. Especially for dealing with draw shapes.

These figures are also mostly in line with expectations, aside from a couple of things:

- I don't understand why the `brush selecting boxes` task got slower after the change. 
- I don't understand why the `traversal` tasks are slower than the `translating boxes` task, both before and after. I would expect that .putting shape records would be much slower than .putting pointer/camera records (since the latter have fewer and simpler properties)

### Change Type

- [x] `patch` — Bug fix

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5,6 +5,7 @@ import {
 	InstancePageStateRecordType,
 	PageRecordType,
 	StyleProp,
+	StylePropValue,
 	TLArrowShape,
 	TLAsset,
 	TLAssetId,
@@ -737,7 +738,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	shapeUtils: { readonly [K in string]?: ShapeUtil<TLUnknownShape> }
 
-	styleProps: { [key: string]: Map<StyleProp<unknown>, string> }
+	styleProps: { [key: string]: Map<StyleProp<any>, string> }
 
 	/**
 	 * Get a shape util from a shape itself.
@@ -7385,9 +7386,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	setStyleForSelectedShapes<T>(
-		style: StyleProp<T>,
-		value: T,
+	setStyleForSelectedShapes<S extends StyleProp<any>>(
+		style: S,
+		value: StylePropValue<S>,
 		historyOptions?: TLCommandHistoryOptions
 	): this {
 		const selectedShapes = this.getSelectedShapes()
--- a/packages/validate/src/lib/validation.ts
+++ b/packages/validate/src/lib/validation.ts
@@ -670,7 +670,9 @@ type ExtractRequiredKeys<T extends object> = {
 	[K in keyof T]: undefined extends T[K] ? never : K
 }[keyof T]
 
-type MakeUndefinedOptional<T extends object> = Pick<T, ExtractRequiredKeys<T>> & Partial<T>
+type ExtractOptionalKeys<T extends object> = {
+	[K in keyof T]: undefined extends T[K] ? K : never
+}[keyof T]
 
 /**
  * Validate an object has a particular shape.
@@ -679,7 +681,9 @@ type MakeUndefinedOptional<T extends object> = Pick<T, ExtractRequiredKeys<T>> &
  */
 export function object<Shape extends object>(config: {
 	readonly [K in keyof Shape]: Validatable<Shape[K]>
-}): ObjectValidator<MakeUndefinedOptional<Shape>> {
+}): ObjectValidator<
+	{ [P in ExtractRequiredKeys<Shape>]: Shape[P] } & { [P in ExtractOptionalKeys<Shape>]?: Shape[P] }
+> {
 	return new ObjectValidator(config) as any
 }
 
--- a/packages/validate/src/lib/validation.ts
+++ b/packages/validate/src/lib/validation.ts
@@ -841,17 +841,12 @@ export function model<T extends { readonly id: string }>(
 	name: string,
 	validator: Validatable<T>
 ): Validator<T> {
-	const getPrefix = (value: unknown) => {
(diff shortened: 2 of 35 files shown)


steveruizok:
Nice! I like the idea of validating from a previous "known good version".

packages/validate/src/lib/validation.ts
```diff
@@ -141,9 +173,17 @@ export class Validator<T> implements Validatable<T> {
 	 * if the value can't be converted to the new type, or return the new type otherwise.
 	 */
 	refine<U>(otherValidationFn: (value: T) => U): Validator<U> {
-		return new Validator((value) => {
-			return otherValidationFn(this.validate(value))
-		})
+		// TODO: is this OK? that it doesn't use the knownGoodVersion function?
```
SomeHats, line 176: this function originally had two purposes:
1. allow type refinements that don't change the value (e.g. `T.string.refine(validateShapeId)` becomes a `Validator<ShapeId>`)
2. allow different types at runtime than at rest (e.g. `T.object({x: ..., y: ...}).refine(Vec.From)` becomes a `Validator<Vec>`

2 went away early on in the process because we decided that validator function shouldn't operate as parsers in this way - I think there are checks that calls to `.validate` return identical values. So this should be fine!

packages/validate/src/lib/validation.ts
```diff
@@ -477,15 +666,21 @@ export const unknownObject = new Validator<Record<string, unknown>>((value) => {
 	return value as Record<string, unknown>
 })
 
+type ExtractRequiredKeys<T extends object> = {
+	[K in keyof T]: undefined extends T[K] ? never : K
+}[keyof T]
+
+type MakeUndefinedOptional<T extends object> = Pick<T, ExtractRequiredKeys<T>> & Partial<T>
+
```
SomeHats, line 674: why was this needed? it seems to have made the hover-completions and generated types from validators a lot clunkier

SomeHats: Might be worth wrapping it or its use in an `Expand` and seeing if that helps at all?

packages/validate/src/lib/validation.ts
```diff
@@ -581,14 +776,25 @@ export function model<T extends { readonly id: string }>(
...
-				? `${name}(id = ${value.id})`
-				: name
-
-		return prefixError(prefix, () => validator.validate(value))
-	})
+	const getPrefix = (value: unknown) => {
+		return value && typeof value === 'object' && 'id' in value && typeof value.id === 'string'
+			? `${name}(id = ${value.id})`
+			: name
+	}
```
SomeHats, line 783: we strip IDs so we can probably remove this

packages/validate/src/lib/validation.ts
```diff
@@ -9,9 +9,26 @@ import {
 
 /** @public */
 export type ValidatorFn<T> = (value: unknown) => T
+/** @public */
+export type ValidatorUsingKnownGoodVersionFn<In, Out = In> = (
+	knownGoodValue: In,
+	value: unknown
+) => Out
```
SomeHats, line 16: is it worth adding tests to the validator library for this change?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7389,9 +7390,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	setStyleForSelectedShapes<T>(
-		style: StyleProp<T>,
-		value: T,
+	setStyleForSelectedShapes<S extends StyleProp<any>>(
+		style: S,
+		value: StylePropValue<S>,
```
SomeHats, line 7395: why were these changes needed here? seems like we've added a lot of `any`s

SomeHats (changes requested):
Overall I really like this change, but I'm not sure about some of the surrounding type changes to style props etc. that have come with this - it seems like we've got a lot more `any`s going on before, and I'm not super clear on why they're needed. 

Where have these come from? Is there a way we can implement this change without losing the type safety & inference we had before?

packages/validate/src/lib/validation.ts
ds300: working on tests now

SomeHats (approved):

packages/editor/src/lib/editor/Editor.ts
ds300: We discussed this offline. It's some kind of variance thing. I couldn't figure it out but also I don't think it matters since the `any`s don't leak into values, and I don't think the generic type gives us much (if any?) extra safety for these style props.

packages/validate/src/lib/validation.ts
ds300: cool thanks. I will remove that TODO and make sure I test the (1) behavior.

PR #2856, opened 2024-02-17
[handles] Line shape handles -> points

steveruizok:
This PR replaces the line shape's `handles` prop with `points`, an array of `VecModel`s.

### Change Type

- [x] `minor` — New feature

### Test Plan

- [x] Unit Tests
- [ ] End to end tests


steveruizok:
Gosh, handles is such a mess. This `DraggingHandle` is full of "if we're a line, do this" and "if we're an arrow, do this".

steveruizok:
I'm very very tempted to just split up DraggingHandle into two different states. I wouldn't expect anyone to use these handle APIs and be able to do much with them. If I could time travel back to 2022, I'd tell past me to maybe keep handles as a UI abstraction but add it as a component instead, i.e. `handles: (shape: TLShape) => handle components`.

packages/tlschema/src/shapes/TLLineShape.ts
```diff
@@ -8,6 +8,16 @@ import { DefaultDashStyle } from '../styles/TLDashStyle'
...
 
+const handleModelValidator = T.object({
+	id: T.string,
+	index: T.indexKey,
+	x: T.number,
+	y: T.number,
+})
+
+/** @public */
+export type TLLineShapeHandle = T.TypeOf<typeof handleModelValidator>
```
SomeHats, line 19: can these be deleted now?

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -604,6 +604,16 @@ export type TLOnBindingChangeHandler<T extends TLShape> = (shape: T) => TLShapeP
...
+/** @public */
+export type TLOnHandleCreateHandler<T extends TLShape> = (
+	shape: T,
+	info: {
+		handle: TLHandle
+		isPrecise: boolean
+		initial?: T | undefined
+	}
+) => TLShapePartial<T> | void
+
```
SomeHats, line 616: I don't see this used anywhere - do we still need it?

SomeHats (approved):
I much prefer this to the ID version! Let's go for it

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
steveruizok: ah yes, we can lose this

packages/tlschema/src/shapes/TLLineShape.ts
steveruizok: think so

PR #2869, opened 2024-02-19
[Snapping 6/6] Self-snapping API

SomeHats:
This diff adds a self-snapping API for handles. Self-snapping is used when a shape's handles want to snap to the shape itself. By default, this isn't allowed because moving the handle might move the snap point, which creates a janky user experience.

Now, shapes can return customised versions of their normal handle snapping geometry in these cases. As a bonus, line shapes now snap to other handles on their own line!

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Line handles should snap to other handles on the same line when holding command

- [x] Unit Tests

### Release Notes

- Line handles now snap to other handles on the same line when holding command


steveruizok:
Which shapes self-snap? Is it just the line shape?

SomeHats:
just line shapes - they're the only ones that have handle-snapping at all at the moment. the main unlock here is snapping for custom shapes

steveruizok (approved):

PR #2870, opened 2024-02-19
Fix some incorrect translation keys

TodePond:
This PR fixes some cases where translation keys became wrong. They slipped in with the composable UI changes.

(side-note: the styling of those titles looks wrong right? that wasnt an intentional change right?)

image

image

To validate for yourself, you can:
- Change the typing of `msg` in `useTranslation` to accept only translation keys. (no strings)
- Run `yarn lint`.
- Look for any errors where we're passing in a string literal. There should be none now!

---

In the future, I'd love to add a tiny eslint rule that enforces this for string literals. We accept strings in `msg` because it makes it easier for consumers (and us) to add custom labels, etc.

Or... if @mimecuvalo does some bigger work on translations, we might have a better solution.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

To validate for yourself, you can:
- Change the typing of `msg` in `useTranslation` to accept only translation keys. (no strings)
- Run `yarn lint`.
- Look for any errors where we're passing in a string literal. There should be none now!

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Unreleased issue. Fixed some translation keys being wrong.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
+++ b/packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
@@ -1,4 +1,3 @@
-import { useTranslation } from '../../hooks/useTranslation/useTranslation'
 import { TldrawUiButton } from '../primitives/Button/TldrawUiButton'
 import { TldrawUiButtonIcon } from '../primitives/Button/TldrawUiButtonIcon'
 import {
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
```diff
@@ -16,13 +15,12 @@ export type TLUiDebugMenuProps = {
...
 /** @public */
 export function DefaultDebugMenu({ children }: TLUiDebugMenuProps) {
-	const msg = useTranslation()
 	const content = children ?? <DefaultDebugMenuContent />
 
 	return (
 		<TldrawUiDropdownMenuRoot id="debug">
 			<TldrawUiDropdownMenuTrigger>
-				<TldrawUiButton type="icon" title={msg('debug-menu.title')}>
+				<TldrawUiButton type="icon" title="Debug menu">
```
TodePond, line 23: We typically hardcode all debug menu copy.

steveruizok:
lgtm!

PR #2871, opened 2024-02-19
Update Hungarian and Korean

TodePond:
This PR pulls over updated translations of Hungarian and Korean from Lokalise.


assets/translations/hr.json
```diff
@@ -22,7 +22,6 @@
 	"action.copy": "Kopiraj",
```
TodePond, line 22: it's hard to see the changes in this file's diff because it does some ordering changes due to some inconsistencies with how we're ordering these files in various places. I will fix that in a follow-up PR. 

To see the changes to Hungarian, please see this commit: f9368143a00e827cd9b44a0ee08121f88d8b43dd

PR #2874, opened 2024-02-19
[maybe] split handle states for lines arrows

steveruizok:
This PR is a quick spike to see what it would look like to split handles into different states for arrows and lines. Currently there is a "generic" handles state that is full of special casing for the two different shapes, making the states difficult to understand and not very generally useful.

### Change Type

### Test Plan

- [x] Unit Tests

### Release Notes

- Removes `PointingHandle`
- Removes `DraggingHandle`
- Adds `PointingArrowHandle` and `DraggingArrowHandle`
- Adds `PointingLineHandle` and `DraggingLineHandle`


steveruizok:
bleh, not sure

PR #2878, opened 2024-02-20
E2E Style Panel Tests

Taha-Hassan-Git:
This PR adds E2E tests for the style panel.

It checks that: 
- the style panel opens and closes as expected on mobile
- the style panel button is disabled for the eraser tool on mobile
- selecting a style hints the button
- changing a style changes the appearance of the shape
- It also moves a test from the toolbar tests that checks the correct styles are exposed for the right tools

fixes tld-2222

### Release Notes

- Add style panel E2E tests


apps/examples/e2e/tests/fixtures/menus/StylePanel.ts
```diff
@@ -14,6 +20,64 @@ export class StylePanel {
...
+		}
+
+		this.align = {
+			start: this.page.getByTestId('style.align.start'),
+			middle: this.page.getByTestId('style.align.middle'),
+			end: this.page.getByTestId('style.align.end'),
+		}
+	}
+	async isHinted(style: Locator) {
+		const getAfterElementStyle = (element: Element, property: string) =>
```
MitjaBezensek, line 69: This one could be pulled out since it's the same as in the `isNotHinted`. In fact, we might create a new function called something like `getAfterElementComputedStyle`, which can be called with the property name?

`getAfterElementComputedStyle('background-color')`

MitjaBezensek (approved):

PR #2884, opened 2024-02-20
Improve dialog appearance on small components

TodePond:
This PR fixes and improves the appearance on dialogs on small tldraw components, eg: Inline components.

image

image
image

image

Marking as minor instead of patch because it adds a new prop to `TldrawUiKbd`.

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Open the "Inset editor" example.
2. Open the keyboard shortcuts dialog.
3. Shrink the window down.
4. Make sure the dialog remains visible at all window sizes.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Dev: Made default dialogs work better when used in small components.

### DIFF
--- a/apps/examples/src/examples/inline/InlineExample.tsx
+++ b/apps/examples/src/examples/inline/InlineExample.tsx
@@ -23,10 +23,10 @@ export default function InlineExample() {
 					gap: 12,
 				}}
 			>
-				<InlineEditor width={900} height={600} />
-				<InlineEditor width={700} height={500} />
-				<InlineEditor width={600} height={400} />
 				<InlineEditor width={500} height={300} />
+				<InlineEditor width={600} height={400} />
+				<InlineEditor width={700} height={500} />
+				<InlineEditor width={900} height={600} />
 			</div>
 		</FocusedEditorContext.Provider>
 	)
--- a/packages/tldraw/src/lib/ui.css
+++ b/packages/tldraw/src/lib/ui.css
@@ -1482,26 +1482,18 @@
 
 .tlui-shortcuts-dialog__body {
 	position: relative;
-	columns: 1;
+	columns: 3;
 	column-gap: var(--space-9);
 	pointer-events: all;
 	touch-action: auto;
 }
 
-/* TODO: refactor shortcuts dialog to make it work in small inline components */
-@media (min-width: 475px) {
-	.tlui-shortcuts-dialog__body {
-		columns: 2;
-		column-gap: var(--space-9);
-	}
+.tlui-shortcuts-dialog__body__tablet {
+	columns: 2;
 }
 
-/* TODO: refactor shortcuts dialog to make it work in small inline components */
-@media (min-width: 960px) {
-	.tlui-shortcuts-dialog__body {
-		columns: 3;
-		column-gap: var(--space-9);
-	}
+.tlui-shortcuts-dialog__body__mobile {
+	columns: 1;
 }
 
 .tlui-shortcuts-dialog__group {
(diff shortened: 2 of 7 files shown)


apps/examples/src/examples/inline/InlineExample.tsx
```diff
@@ -23,10 +23,10 @@ export default function InlineExample() {
 					gap: 12,
 				}}
 			>
-				<InlineEditor width={900} height={600} />
-				<InlineEditor width={700} height={500} />
-				<InlineEditor width={600} height={400} />
 				<InlineEditor width={500} height={300} />
+				<InlineEditor width={600} height={400} />
+				<InlineEditor width={700} height={500} />
+				<InlineEditor width={900} height={600} />
```
TodePond, line 29: re-ordered these because the small ones are most useful to have at the top, at-hand

packages/tldraw/src/lib/ui.css
```diff
@@ -748,8 +748,8 @@
 	font-size: 12px;
 	overflow: hidden;
 	min-width: 300px;
-	max-width: 80vw;
-	max-height: 80vh;
+	max-width: 80%;
+	max-height: 80%;
```
TodePond, line 752: this was the main problem with dialogs.
no more vh/vw allowed

packages/tldraw/src/lib/ui.css
```diff
@@ -1482,26 +1482,18 @@
...
 }
 
-/* TODO: refactor shortcuts dialog to make it work in small inline components */
-@media (min-width: 960px) {
-	.tlui-shortcuts-dialog__body {
-		columns: 3;
-		column-gap: var(--space-9);
-	}
+.tlui-shortcuts-dialog__body__mobile {
+	columns: 1;
```
TodePond, line 1496: this is a similar pattern to what we've done elsewhere for a while.
this is a dialog though, so it's *outside* the layout component that we usually use (`.tl-layout__mobile`)

packages/tldraw/src/lib/ui/components/primitives/TldrawUiKbd.tsx
```diff
@@ -5,12 +5,13 @@ import { kbd } from '../../kbd-utils'
...
 	children: string
+	visibleOnMobileLayout?: boolean
 }
 
 /** @public */
-export function TldrawUiKbd({ children }: TLUiKbdProps) {
+export function TldrawUiKbd({ children, visibleOnMobileLayout = false }: TLUiKbdProps) {
 	const breakpoint = useBreakpoint()
-	if (breakpoint < PORTRAIT_BREAKPOINT.MOBILE) return null
+	if (!visibleOnMobileLayout && breakpoint < PORTRAIT_BREAKPOINT.MOBILE) return null
```
TodePond, line 14: Normally we hide keyboard shortcuts on small screens (ie: mobile screens).

But if you're looking at the keyboard shortcuts menu, we should *always* show the keyboard shortcut. So I added a prop that lets you override the auto-hiding. This is a pretty common case you can get to by shrinking your browser window when you have many windows open.

I could understand the viewpoint that this should decision should come from higher up the tree, but I'm also happy to keep behaviour how it currently is - seems fine to me.

mimecuvalo: hah, yeah this feels funny. i think it's prbly better to leave this one be? feels like it's the wrong abstraction for this and hard to get into this case anyway.

mimecuvalo (approved):
LGTM

PR #2890, opened 2024-02-20
Add line IDs & fractional indexes

SomeHats:
In #2856, we moved changed line handles into an array of points. This introduced an issue where some concurrent operations wouldn't work because they array indexes change. We need some sort of stable way of referring to these points. Our existing fractional indexing system is a good fit.

In this version, instead of making the points be a map from index to x/y, we make the points be a map from id (the index) to x/y/index/id(also index). This is "kinda silly" (steve's words) but might be more familiar to devs who are expecting maps to be keyed on IDs rather than anything else.

### Change Type


steveruizok (approved):
LSBGTM!

PR #2891, opened 2024-02-20
Check tsconfig "references" arrays

### LINKED ISSUE #2800: [dx]: Check TS project references
We should add a precommit check to make sure that any project references that need to present are. i.e. for any pair of workspaces in our monorepo where `@tldraw/A` depends on `@tldraw/B` in either `package.json#dependencies` or `package.json#devDependencies`, `A`'s `tsconfig.json` file should include a project reference pointing to `B`.

We miiight be able to use the yarn constraints thing for this? in yarn.config.cjs. Or there might be some other way to inject code into the yarn checks phase.

ds300:
Closes #2800

This PR makes it so that `check-scripts` will error out if you forget to add a "references" entry to a tsconfig file when adding an internal dependency in our monorepo.

If these project references are missed it can prevent TS from building/rebuilding things when they need to be built/rebuilt.

### Change Type

### DIFF
--- a/scripts/check-scripts.ts
+++ b/scripts/check-scripts.ts
@@ -1,8 +1,6 @@
-import { existsSync, readFileSync, writeFileSync } from 'fs'
-import { parse } from 'json5'
 import kleur from 'kleur'
 import path from 'path'
-import { REPO_ROOT, writeJsonFile } from './lib/file'
+import { REPO_ROOT, readJsonIfExists, writeJsonFile } from './lib/file'
 import { nicelog } from './lib/nicelog'
 import { Package, getAllWorkspacePackages } from './lib/workspace'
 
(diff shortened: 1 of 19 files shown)


scripts/check-scripts.ts
```diff
@@ -53,10 +55,70 @@ const perPackageExceptions: Record<string, Record<string, () => string | undefin
...
+				!tsconfig.references?.some(({ path }: { path: string }) => path === tsconfigReferencePath)
+			) {
+				fixedDeps.push({ path: tsconfigReferencePath })
+				missingRefs.push(dep)
+			}
+		}
+		if (missingRefs.length) {
+			if (fix) {
+				tsconfig.references = fixedDeps.sort((a: any, b: any) => a.path.localeCompare(b.path))
+				writeFileSync(tsconfigPath, JSON.stringify(tsconfig, null, '\t'), 'utf-8')
```
SomeHats, line 96: Using our `writeJsonFile` helper here will make sure that the generated files match our prettier formatting

scripts/check-scripts.ts
```diff
@@ -53,10 +55,70 @@ const perPackageExceptions: Record<string, Record<string, () => string | undefin
...
+		const tsconfigPath = path.join(workspace.path, 'tsconfig.json')
+		if (packagesWithoutTSConfigs.has(workspace.name)) {
+			continue
+		}
+
+		if (!existsSync(tsconfigPath)) {
+			throw new Error('No tsconfig.json found at ' + tsconfigPath)
+		}
+
+		const tsconfig = parse(readFileSync(tsconfigPath, 'utf-8'))
```
SomeHats, line 71: we have a `readJsonIfExists` if it'd be helpful

SomeHats (approved):
Love this! Should we rename `check-scripts` to something like `check-package-config` or something?

PR #2902, opened 2024-02-21
menu fixes: add company links in general; add tracking to lang menu

mimecuvalo:
- the company links appear back in the burger menu. they could be selectively shown if mobile but i'd argue they should just always be there.
- add the `track` to LanguageMenu to make the menu update. however, i'm a little annoyed that i don't understand why the Help menu already works without this :-/

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Add company menu links back in and make sure the Language menu is updated on change.

### DIFF
--- a/apps/dotcom/src/components/LocalEditor.tsx
+++ b/apps/dotcom/src/components/LocalEditor.tsx
@@ -45,6 +45,7 @@ const components: TLComponents = {
 		<DefaultMainMenu>
 			<LocalFileMenu />
 			<DefaultMainMenuContent />
+			<Links />
 		</DefaultMainMenu>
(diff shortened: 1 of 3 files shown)


apps/dotcom/src/components/LocalEditor.tsx
```diff
@@ -45,6 +45,7 @@ const components: TLComponents = {
 		<DefaultMainMenu>
 			<LocalFileMenu />
 			<DefaultMainMenuContent />
+			<Links />
```
SomeHats, line 48: should this be in `MultiplayerEditor` too?

PR #2913, opened 2024-02-21
Make Vercel URL rewrites precise

si14:
### The problem

Right now we use a catchall path in Vercel routing config to rewrite all requests that don't match existing assets to `/index.html`, which is needed for client side routing to work. This, however, messes up 404 errors for truly non-existing files which won't be handled by the SPA, because they get redirected to index.html.

Even worse, this interacts very poorly with caching. Normally if we request a non-existent file, then put the file in place, and request the file again, we'll get 404 the first time and the actual file the second time. However, in our case we instead return `/index.html` after the first attempt and cache that response, making it impossible to correct a missing file without cache flush.

### The solution

One way to fix this is to make the regex in Vercel config precise, so that they only match our SPA routes. However, it can be dangerous, because this means we'll need to manually update the config with new SPA routes every time we add any. This PR tests that the regex we're using in Vercel matches all routes that we set in the SPA router.

### Change Type

### Test Plan

1. Might need a light smoke test after deployment to dotcom.

- [x] End to end tests

### DIFF
--- a/apps/dotcom/spaRouteFilter.ts
+++ b/apps/dotcom/spaRouteFilter.ts
@@ -1,0 +1,19 @@
+// This value is used to configure Vercel routing to rewrite dotcom SPA routes to /index.html.
+// It is also tested in routes.test.ts to make sure it matches all React Router routes.
+//
+// It is a string and not a regular expression as it's substituted
+// into a string in Vercel's "build output spec" in scripts/build.ts.
+//
+// Make sure it's not overly broad, because otherwise we won't give correct 404 responses.
+export const SPA_ROUTE_FILTER =
+	'^\\/new.*' +
+	'|' +
+	'^\\/r' +
+	'|' +
+	'^\\/r\\/.*' +
+	'|' +
+	'^\\/s\\/.*' +
+	'|' +
+	'^\\/v\\/.*' +
+	'|' +
+	'^\\/$'
--- a/apps/dotcom/src/routes.test.ts
+++ b/apps/dotcom/src/routes.test.ts
@@ -1,0 +1,26 @@
+import { RouteObject } from 'react-router-dom'
+import { SPA_ROUTE_FILTER } from '../spaRouteFilter'
+import { router } from './routes'
+
+test('SPA_ROUTE_FILTER matches all React routes', () => {
+	const paths: string[] = []
+
+	function walk(routeObject: RouteObject) {
+		if (routeObject.path && routeObject.path !== '*') {
+			paths.push(routeObject.path)
+		}
+		if (routeObject.children) {
+			routeObject.children.forEach(walk)
+		}
+	}
+
+	router.forEach(walk)
+
+	console.log(paths)
+
+	const regex = new RegExp(SPA_ROUTE_FILTER)
+	console.log(regex)
+	paths.forEach((path) => {
+		expect(path).toMatch(regex)
+	})
+})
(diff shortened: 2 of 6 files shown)


si14:
I'm not sure how to test it on Vercel though 🤔

ds300 (changes requested):
Nice, this is a decent solution! 

One change we need to make with this is to add a 404 fallback to index.html so we can get our own 404 page instead of the vercel one 
image

should look like 

image

Also, and this is nonblocking, I think I'd prefer to have seen the filter be derived from the routing config, rather than configured separately, so people don't need to manually update a big regex source when adding routes. But this works okay for now.

apps/dotcom/spaRouteFilter.ts
```diff
@@ -0,0 +1,19 @@
...
+	'|' +
+	'^\\/r' +
+	'|' +
+	'^\\/r\\/.*' +
+	'|' +
+	'^\\/s\\/.*' +
+	'|' +
+	'^\\/v\\/.*' +
+	'|' +
+	'^\\/$'
```
mimecuvalo, line 19: nit: this might be easier to read if you use the backtick string templates?
or, actually, could this be simplified to be: `^/new.*|^/r|^/[rsv]/.*|^/$`

actually, aren't these equivalent: 
'^\\/r'
and 
'^\\/r\\/.*'

mimecuvalo: small bug: ^/new.* would capture /newblahblah which shouldn't be captured. should be: ^/new$

si14:
@ds300 thank you!

Re: 404: will do

Re: autoderivation: I thought about that, but it seems a bit more involved and would require React Routing during build. We don't add those routes often, so I think failing a build should be a decent compromise between complexity and value?

apps/dotcom/src/routes.test.ts
```diff
@@ -0,0 +1,25 @@
+import { RouteObject } from 'react-router-dom'
+import { SPA_ROUTE_FILTER } from '../spaRouteFilter'
+import { router } from './routes'
+
+test('SPA_ROUTE_FILTER matches all React routes', () => {
```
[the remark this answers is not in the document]
si14: good point

ds300:
> Re: autoderivation: I thought about that, but it seems a bit more involved and would require React Routing during build. We don't add those routes often, so I think failing a build should be a decent compromise between complexity and value?

I think it'd be quick and simple to implement, happy to pair on it if you like. The build script is in the dotcom package, so it should be able to use the router definition like you did in the test.

And once we do start working on a logged in experience again the routes will pile up quickly and change often so I think it's worth doing now to avoid friction later.

But like I said it's nonblocking so feel free to merge without. I might take a look at it one day in the future.

apps/dotcom/spaRouteFilter.ts
si14: good points!

re: `\\/`: you're absolutely right, I I was experimenting with the standard regex syntax and I didn't realise I don't need to escape slashes in strings. My bad 😅 

I also simplified the regex in a different way, please check out the last commit!

apps/dotcom/src/routes.test.ts
```diff
@@ -0,0 +1,58 @@
...
+	} else {
+		return {
+			message: () =>
+				`expected ${this.utils.printReceived(actual)} to match at least one of the regexes ${this.utils.printExpected(regexes)}`,
+			pass: false,
+		}
+	}
+}
+
+expect.extend({ toMatchAny })
```
ds300, line 32: adding a custom jest matcher is a neat way to make errors more readable for these tests, nice idea.

ds300 (approved):
thanks for adding the 404 fallback 🙏🏼 this looks good to go

mimecuvalo:
thx @si14 !!! this was much more work than expected just to add a 404!

PR #2922, opened 2024-02-22
[docs] docs design shuffle

steveruizok:
This PR makes some changes to the docs site:
- adds introduction
- removes quick start
- restores usage
- updates copy in usage

### Change Type


mimecuvalo (changes requested):
let's talk about this first! i think this is a step backwards but obviously we have different opinions on this, so let's see how we can maybe meet in the middle.

steveruizok:
Some more of my rationale here:
- it's critical that we have positioning information (including on license) on the first page of our developer sdk website; this information was removed when we added the quick start pr
- our quick start guide was essentially "install the library and add the react component" and so seemed redundant vs regular installation steps.
- our examples fill the same needs the QuickStart but better 
- our installation page was inaccurately large and complex 
- I don't want to include the interactive component on the first page, it's a subpar introduction vs a link

mimecuvalo:
@steveruizok so for this PR - am i waiting on this other one https://github.com/tldraw/tldraw/pull/2951 ?

Issue #2892, opened 2024-02-20
[Bug]: Shape indexes are not maintained when creating frame dragging over shapes


ricardo-crespo:
### What happened?

It seems that when we drag a frame over shapes to include them, the drawing order changes and doesn't respect any manipulation we've made to put some shapes in front of others

https://github.com/tldraw/tldraw/assets/9846113/90ca7e64-a200-416b-a5db-8b0b95a34137

### How can we reproduce the bug?

1. Create some shapes and change their original indexes positioning some on top of others. 
2. Select frame tool and drag to include the shapes to the newly created frame

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2928, opened 2024-02-23
Fix frames not preserving shape order

MitjaBezensek:
It looks like enclosing shapes with a new frame did not preserve the order of the shapes. Also makes framing inside of frames work.

Solves https://github.com/tldraw/tldraw/issues/2892

Before:

https://github.com/tldraw/tldraw/assets/2523721/90da4fc0-92a1-49fe-b658-73842c4ef4c2

After:

https://github.com/tldraw/tldraw/assets/2523721/0558d22e-8216-4d84-8a89-7dd049c37974

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Add a few shapes.
2. Make sure to change some of their order (example in the gif puts the last created shape to the back)
3. Create a new frame that encloses these shapes.
4. The order of the shapes should be preseved.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fix an issue when framing shapes did not preserve the original order of the shapes.

### DIFF
--- a/packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts
+++ b/packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts
@@ -12,22 +12,21 @@ export class FrameShapeTool extends BaseBoxShapeTool {
 		const bounds = this.editor.getShapePageBounds(shape)!
 		const shapesToAddToFrame: TLShapeId[] = []
 		const ancestorIds = this.editor.getShapeAncestors(shape).map((shape) => shape.id)
-		const pageId = this.editor.getCurrentPageId()
 
-		this.editor.getSortedChildIdsForParent(pageId).map((pageShapeId) => {
-			const pageShape = this.editor.getShape(pageShapeId)
-			if (!pageShape) return
+		this.editor.getSortedChildIdsForParent(shape.parentId).map((siblingShapeId) => {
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts
```diff
@@ -12,8 +12,11 @@ export class FrameShapeTool extends BaseBoxShapeTool {
 		const bounds = this.editor.getShapePageBounds(shape)!
 		const shapesToAddToFrame: TLShapeId[] = []
 		const ancestorIds = this.editor.getShapeAncestors(shape).map((shape) => shape.id)
+		const pageId = this.editor.getCurrentPageId()
 
-		this.editor.getCurrentPageShapes().map((pageShape) => {
+		this.editor.getSortedChildIdsForParent(pageId).map((pageShapeId) => {
+			const pageShape = this.editor.getShape(pageShapeId)
+			if (!pageShape) return
```
ds300, line 19: `getSortedChildIdsForParent` only returns the parent's immediate children, and I think that's preventing the frame capture thing from working within frames themselves

Kapture 2024-02-23 at 10 49 47

I think you could fix this by doing `this.editor.getSortedChildIdsForParent(shape.parentId)` instead?

ds300 (approved):

Issue #2930, opened 2024-02-23
Editor focus lost after undo


ds300:
### What happened?

This happens in at least Safari and Chrome on MacOS. The steps to reproduce seem very specific but I worry that the actual underlying problem is much broader.

### How can we reproduce the bug?

1. Create a geo box
2. Change the geo type to cloud
3. Hit undo on the keyboard (the bug will not manifest if you click the undo button in the UI)
4. All keyboard shortcuts stop working

### What browsers are you seeing the problem on?

Chrome, Safari

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mimecuvalo:
Anecdotally this seems to be happening as well — I'll try to watch out for it as well. I'll be hitting undo and then it reverts to a browser shortcuts instead of the editor shortcuts.

ds300:
this was fixed by #2936

PR #2936, opened 2024-02-23
Fix keyboard shortcuts bugs

steveruizok:
This PR moves the focus 

### Change Type

- [x] `minor` 

### Test Plan

1. Select an element.
2. Press the delete quick action menu button.
3. Undo the delete with a keyboard shortcut.

1. Create a geo shape
2. Use the style panel to change the geo type
3. Undo so that it deletes
4. Try to redo

### Release Notes

- [Fix] Keyboard shortcut focus bug


steveruizok:
Closes #2930

PR #2937, opened 2024-02-23
[dx] Derive vercel routes from react-router config

ds300:
I had some free time at the end of the week so I investigated the idea of deriving the vercel routing config from the react-router config, then storing the derived vercel route info in a jest snapshot, and then loading the jest snapshot during the build script. 

Seems to work well!

### Change Type


si14 (approved):
This PR is incredible and humbling 😆  A few nits and suggestions, but nothing major

ds300:
thanks @si14 , great feedback

PR #2938, opened 2024-02-23
[feature] wrap mode

steveruizok:
By default, tldraw's brushing mode will select when the box intersects an shape's geometry. A user can hold Command / Ctrl to require that the selection box fully contain a shape's bounds instead.

Some people really prefer the opposite. Three years! Three years I've been saying "no no no".

This PR adds a user preference to flip the logic. When `isWrapMode` is true, selection requires that the box completely contain a shape before it's added to the list of selecting shapes; and ctrl flips back to intersection instead.

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Turn on wrap mode in the user preferences menu.
2. Select stuff.
3. Use the ctrl key to except the behavior back to intersection.

- [x] Unit Tests

### Release Notes

- Added `isWrapMode` to user preferences.
- Added Wrap Mode toggle to user preferences menu.


h2thien:
This is super useful for us, thanks! Fun fact: in some softwares, this is called "direction-based selection", because in environments with nothing but a mouse, this is implemented based on whether you drag the mouse from left to right (select if touched) or right to left (select if filled). Glad to see this is supported, with better interaction IMO.

steveruizok:
> 
This is super useful for us, thanks! Fun fact: in some softwares, this is called "direction-based selection", because in environments with nothing but a mouse, this is implemented based on whether you drag the mouse from left to right (select if touched) or right to left (select if filled). Glad to see this is supported, with better interaction IMO.

Ha, I'd done the "direction based selection" before in v1 (old.tldraw.com) under a preference for CAD-style selection. No one really asked for it yet, but seems like selection can get much more complicated than I'm used to. Rare case where a user preference / mode feels like the right move.

PR #2942, opened 2024-02-23
[spike] babybrain migrations

steveruizok:
kee-rist this stuff is tricky

This PR is an attempt at allowing a user to specify when store migrations need to run before shape migrations.

We do this by allowing a record's migrators to include references to store migrations.

```ts
const shapeTypeMigrations = defineMigrations({
	currentVersion: ShapeVersion.C,
	migrators: {
		// regular migration
		[ShapeVersion.A]: { up, down },
		// a store migration that needs to run before the following migrations can run
		[ShapeVersion.B]: StoreVersions.X,
		// another regular migration, will run only after StoreVersions.X
		[ShapeVersion.C]: { up, down },
})
```

The logic is:

For each snapshot/store migration:
- migrate the store
- migrate all records, ignoring migrations that:
  - have already ran; or
  - require a later version

(This PR's a mess but that's the idea).

## We always migrate stores, never records in isolation

Once we have dependencies between store migrations and record migrations, we need to always migrate a record in the context of a store. We can't say run `schema.migratePersistedRecord` in isolation. If we want to migrate a record, we need to create a temporary store where that record is the only record, then migrate that store.

Given the fact that some store migrations might be comparing multiple records, for example arrows and shapes that are bound to arrows, migrating records alone (but in their own store) may not be enough—we may have to include all of those dependencies at the same time.

## Version compatibility in sync

For this reason, it may be impossible to run migrations as we have been on individual records in our sync logic. Simply put, we can never adequately handle a situation where the client's version of a record is older than the server's, because upgrade it would require access to their entire store. It would be easier and more reliable to force the client to upgrade and migrate their entire store.

### Change Type

- [x] `minor` — New feature


packages/editor/src/lib/config/TLSessionStateSnapshot.ts
```diff
@@ -133,7 +133,7 @@ function migrateAndValidateSessionStateSnapshot(state: unknown): TLSessionStateS
 		console.warn('No version in instance state')
 		return null
 	}
-	const result = migrate<TLSessionStateSnapshot>({
+	const result = migrateArbitraryValue<TLSessionStateSnapshot>({
```
steveruizok, line 136: changed temporarily to avoid confusion, we can revert this later

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7721,7 +7722,38 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+				case 'asset': {
+					assets.push(record)
+					break
+				}
+				case 'shape': {
+					shapes.push(record)
+					break
+				}
+			}
+		}
```
steveruizok, line 7754: When putting content into the page, we used to migrate each record individually. Here we migrate them together as a single store, then pull the values out later. It would be better to have `getContent` generate a store snapshot (it's basically doing that already).

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7865,18 +7897,6 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-				const result = this.store.schema.migratePersistedRecord(asset, content.schema)
-				if (result.type === 'success') {
-					assets[i] = result.value as TLAsset
-				} else {
-					throw Error(
-						`Could not put content:\ncould not migrate content for asset:\n${asset.type}\nreason:${result.reason}`
-					)
-				}
-			}
-
```
steveruizok, line 7879: We don't need these per-record migrations anymore.

PR #2945, opened 2024-02-24
fix structured clone reference in drawing

steveruizok:
This PR fixes a rogue structuredClone reference in the drawing state.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Fixes a reference to structuredClone that caused a crash on older browsers.

### DIFF
--- a/packages/utils/src/lib/value.ts
+++ b/packages/utils/src/lib/value.ts
@@ -30,7 +30,11 @@ export function isNonNullish<T>(
 	return value !== null && value !== undefined
 }
 
-/** @public */
+/**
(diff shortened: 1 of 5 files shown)


packages/utils/src/lib/value.ts
```diff
@@ -30,7 +30,11 @@ export function isNonNullish<T>(
 	return value !== null && value !== undefined
 }
 
-/** @public */
+/**
+ * Create a deep copy of a value. Uses the structuredClone API if available, otherwise uses JSON.parse(JSON.stringify()).
+ *
+ * @param i - The value to clone.
+ * @public */
```
mimecuvalo, line 37: @steveruizok is it worth using core-js polyfill here?

https://github.com/zloirock/core-js#structuredclone

PR #2947, opened 2024-02-24
Fix iframe warning

steveruizok:
This PR fixes a check on whether the multiplayer editor has been loaded in an iframe.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Load me in an iframe

### DIFF
--- a/apps/dotcom/src/components/MultiplayerEditor.tsx
+++ b/apps/dotcom/src/components/MultiplayerEditor.tsx
@@ -190,9 +190,7 @@ export function UrlStateSync() {
 
 function useIsEmbedded(slug: string) {
 	const isEmbedded =
-		typeof window !== 'undefined' &&
-		window.self !== window.top &&
-		window.location.host !== 'www.tldraw.com'
+		typeof window !== 'undefined' && (window !== window.top || window.self !== window.parent)
 
 	useEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/dotcom/styles/core.css
```diff
@@ -1,8 +1,17 @@
 @import url('https://fonts.googleapis.com/css2?family=Inter:wght@500;600;800&display=swap');
 
 :root {
-	font-family: Inter, -apple-system, 'system-ui', 'Segoe UI', 'Noto Sans', Helvetica, Arial,
-		sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji';
+	font-family:
+		Inter,
+		-apple-system,
```
mimecuvalo, line 6: drive-by nit: `system-ui` is prefered over `-apple-system` these days:
https://css-tricks.com/snippets/css/system-font-stack/

besides, this will help fix that serif problem we were talking about @steveruizok with the apple-system giving a serif font (yuck)

mimecuvalo:
@steveruizok high-level comment on this PR - do you explicitly want UI for this or should we utilize CSP to ensure compliance? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

PR #2953, opened 2024-02-26
[experiment] stickies: new look and feel

mimecuvalo:
https://github.com/tldraw/tldraw/assets/469604/498430d8-5be4-4ee9-a4d4-7c97411828fc

This is behind a feature flag and I'll land this as is and we can iterate on it / live with it for a while / develop it further!
- [x] tool: update iconography which didn't suggest a sticky note
- [x] mouse: new sticky 'create' mouse cursor
- [x] size: slightly bigger default notes
- [x] indicator: hide ShapeIndicator when editing text. it's distracting
- [x] styling: clean rectangle, no border radius, add box-shadow to suggest bent paper
- [x] sizing: maintain square aspect ratio feels better than the rectangular one
- [x] text: start position at top-left
- [x] style panel: lock down size, alignment and see if being locked in bothers us
- [x] rotation: place sticky slightly rotated randomly
- [x] multiple notes on creation; can scrub back-and-forth to change amount easily
- [x] text: should input on keystroke, not wait for double-click
- [x] keyboard shortcuts - tab & cmd-enter create new sticky next to them
- [x] text: don’t auto-resize font-size, but auto-resize note with aspect ratio, min-size
- [x] svg check
- [-] nixed for now: handle/control to create duplicate - this feels not right at the moment. since this is mimicking real life shapes i don't go to a real life post-it note to create a duplicate. i go to my pad of post-it notes, i.e. the tool. the ability to create multiple using the tool using scrubbing i think solves for this case more elegantly and intuitively.
    none of our other shapes have auto-duplicate handles and it feels like this shouldn't either. 
- [-] nixed for now: handle/control for resizing, for now. i'm not sure if we need it yet.

For future PRs:
- [ ] sizing: maybe snapped sizes only?
- [ ] sizing: sticky tabs?
- [ ] collab: be able to do reacji
- [ ] collab: add attribution via auto-color multiplayer?
- [ ] collab: add attribution via name

### Change Type

- [x] `minor` — New feature

### Test Plan

1. This is under a feature flag and will be dogfooded for a while.

- [x] Dogfood!

### Release Notes

- Refreshes the sticky look, feel, and mechanics.

### DIFF
--- a/assets/icons/icon/tool-note-new.svg
+++ b/assets/icons/icon/tool-note-new.svg
@@ -1,0 +1,4 @@
+<!-- MIT License: Icons8 https://github.com/icons8/windows-10-icons?ref=svgrepo.com -->
+<svg fill="black" viewBox="2 2 28 28" xmlns="http://www.w3.org/2000/svg">
+  <path d="M 5 5 L 5 27 L 20.40625 27 L 20.71875 26.71875 L 26.71875 20.71875 L 27 20.40625 L 27 5 Z M 7 7 L 25 7 L 25 19 L 19 19 L 19 25 L 7 25 Z M 21 21 L 23.5625 21 L 21 23.5625 Z" />
+</svg>
--- a/packages/assets/imports.js
+++ b/packages/assets/imports.js
@@ -171,6 +171,7 @@ import iconsToolHighlight from './icons/icon/tool-highlight.svg'
 import iconsToolLaser from './icons/icon/tool-laser.svg'
 import iconsToolLine from './icons/icon/tool-line.svg'
 import iconsToolMedia from './icons/icon/tool-media.svg'
+import iconsToolNoteNew from './icons/icon/tool-note-new.svg'
 import iconsToolNote from './icons/icon/tool-note.svg'
 import iconsToolPencil from './icons/icon/tool-pencil.svg'
 import iconsToolPointer from './icons/icon/tool-pointer.svg'
@@ -387,6 +388,7 @@ export function getAssetUrlsByImport(opts) {
 			'tool-laser': formatAssetUrl(iconsToolLaser, opts),
 			'tool-line': formatAssetUrl(iconsToolLine, opts),
 			'tool-media': formatAssetUrl(iconsToolMedia, opts),
+			'tool-note-new': formatAssetUrl(iconsToolNoteNew, opts),
 			'tool-note': formatAssetUrl(iconsToolNote, opts),
 			'tool-pencil': formatAssetUrl(iconsToolPencil, opts),
 			'tool-pointer': formatAssetUrl(iconsToolPointer, opts),
--- a/packages/assets/imports.vite.js
+++ b/packages/assets/imports.vite.js
@@ -171,6 +171,7 @@ import iconsToolHighlight from './icons/icon/tool-highlight.svg?url'
 import iconsToolLaser from './icons/icon/tool-laser.svg?url'
 import iconsToolLine from './icons/icon/tool-line.svg?url'
 import iconsToolMedia from './icons/icon/tool-media.svg?url'
+import iconsToolNoteNew from './icons/icon/tool-note-new.svg?url'
 import iconsToolNote from './icons/icon/tool-note.svg?url'
 import iconsToolPencil from './icons/icon/tool-pencil.svg?url'
 import iconsToolPointer from './icons/icon/tool-pointer.svg?url'
@@ -387,6 +388,7 @@ export function getAssetUrlsByImport(opts) {
 			'tool-laser': formatAssetUrl(iconsToolLaser, opts),
 			'tool-line': formatAssetUrl(iconsToolLine, opts),
 			'tool-media': formatAssetUrl(iconsToolMedia, opts),
+			'tool-note-new': formatAssetUrl(iconsToolNoteNew, opts),
 			'tool-note': formatAssetUrl(iconsToolNote, opts),
 			'tool-pencil': formatAssetUrl(iconsToolPencil, opts),
 			'tool-pointer': formatAssetUrl(iconsToolPointer, opts),
--- a/packages/assets/selfHosted.js
+++ b/packages/assets/selfHosted.js
@@ -166,6 +166,7 @@ export function getAssetUrls(opts) {
 			'tool-laser': formatAssetUrl('./icons/icon/tool-laser.svg', opts),
 			'tool-line': formatAssetUrl('./icons/icon/tool-line.svg', opts),
 			'tool-media': formatAssetUrl('./icons/icon/tool-media.svg', opts),
+			'tool-note-new': formatAssetUrl('./icons/icon/tool-note-new.svg', opts),
 			'tool-note': formatAssetUrl('./icons/icon/tool-note.svg', opts),
 			'tool-pencil': formatAssetUrl('./icons/icon/tool-pencil.svg', opts),
 			'tool-pointer': formatAssetUrl('./icons/icon/tool-pointer.svg', opts),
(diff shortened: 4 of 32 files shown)


mimecuvalo:
Going to land this so people can play around with it locally!
Let's grab a time to talk more offline @steveruizok about next steps.

ds300:
- I love the icon
- Love the intent behind the new cursor. But visually, on macos chrome at least, the builtin 'copy' cursor is kinda dorky and intense looking, so I'd advocate for trying out something custom.
- Love the skeuomorphic shadows and the random wobble
- Didn't love the auto sizing, although I do think it's overall better than the current approach of growing vertically. And it's a heckuva lot better than Miro's shrinking text. 
  - This might be a dreadful idea, but I'd be curious to see it tend towards 16:9 rather than staying at 1:1, since in my experience that's roughly the apsect ratio of larger stickies. At least that's what the office managers have purchased in the places I've worked 😇

mimecuvalo:
thx for the feedback!

>  But visually, on macos chrome at least, the builtin 'copy' cursor is kinda dorky and intense looking, so I'd advocate for trying out something custom.

hah, yeah - we need our own custom icon :P Agreed that it's dorky! I think the other tools should also have their own icon.

> Didn't love the auto sizing, although I do think it's overall better than the current approach of growing vertically. And it's a heckuva lot better than Miro's shrinking text.

yeah i'm not sure about it either. but i wanna sit with it for a while and see how it feels over the course of a week or more. i think what we can do is create 5 (or more!) different sticky note experiment/feature flags and, that way, we can play around with the different things for what feels right for auto-sizing and any other aspect of the sticky note.

> This might be a dreadful idea, but I'd be curious to see it tend towards 16:9 rather than staying at 1:1, since in my experience that's roughly the apsect ratio of larger stickies. At least that's what the office managers have purchased in the places I've worked 😇

hah, yeah, this is sort of implied in the future PR section of 'snapping sizes' for stickies. Miro does this when resizing and maybe we should too? we could allow a subset of different shapes that match the real-life ones? but i'm not sure if that leans too hard into their real world counterparts.

steveruizok:
### Skeumorphism
I think I like the design? High contrast from the rest of the diagram / app is a good thing. Dark mode would need some thinking—the shadows won't be strong there, so these will be rectangles.

### Edit on type

Great feature. It does mean that our tool-based shortcuts don't work when a sticky note is selected, but I think this is a fair trade. IMO worth landing immediately!

### Tab to create new sticky

Miss for me. I can't find anyone else that does this and tab should behave as tab does when editing text. I'd like a better shortcut here that doesn't conflict with text editing, but that's hard to find as most modifiers are taken already. Are there conventions here?

### Randomized rotation

I think the randomized rotation is a miss. Most of the time sticky notes are used in grids and columns, and having each sticky randomized is going to be hard to square with that.

### Auto-sizing

Autosizing also feels like a miss for the same reason. In Miro / Freeform, sticky notes don't grow at all, and the text shrinks to fit the sticky. In Figma, notes grow on their y axis as text gets bigger.

### Multiple on drag

Most apps do resize on drag and I'm really not a fan of that. IMO the current drag to position is more useful (Figjam style). I haven't looked at the implementation here but the interactions are kind of unpredictable? I do like the idea of creating multiple sticky notes at the same time. Could we have a different way of creating, ie a grid, maybe while holding shift?

### Icon

Good improvement! Can be adjusted to better match the other icons that we have (size, maybe radius).

### Styles

I'll have spend some time with this, seems like a big bet on the defaults.

mimecuvalo:
> I think I like the design? High contrast from the rest of the diagram / app is a good thing. Dark mode would need some thinking—the shadows won't be strong there, so these will be rectangles.

yeah, i'm not _entirely_ sold that we should go in this direction vs. a more platonic rectangle. but it's interesting to play with leaning into this for now. agreed on dark mode, ugh, halp — i could use some design help there!

> Great feature. It does mean that our tool-based shortcuts don't work when a sticky note is selected, but I think this is a fair trade. IMO worth landing immediately!

👍 🎉 

> Miss for me. I can't find anyone else that does this and tab should behave as tab does when editing text. I'd like a better shortcut here that doesn't conflict with text editing, but that's hard to find as most modifiers are taken already. Are there conventions here?

Hah, you might have not seen the other places that do this: Miro, Mural, and Zoom Whiteboard all do new sticky on Tab.
Plus, FigJam does new sticky on Cmd-Enter.
The problem is that we don't have rich-text to support lists. If that was the case, this might be rethought, but it isn't at the time.
Also, we _do_ need an option to be able to stay in typing mode without lifting fingers off the keyboard if possible.
Let's talk more offline but we're not forging new territory here — if anything, we're behind the "latest trends". :P

> I think the randomized rotation is a miss. Most of the time sticky notes are used in grids and columns, and having each sticky randomized is going to be hard to square with that.

yeah, this is part of leaning into the skeuomorphism bit and i'm not sold on it either to be fair — it's an idea to play with. i thought, well, either we're playing into the real-life equivalent or we should go more into the platonic ideal. it can possibly feel like a miss if we're middling about it and not picking one or the other maybe? another thing to consider is that we have the `draw` playful font as the default which is meant to convey comfort to the user/less formality and encourage flow of thoughts vs. fiddling with alignment.

i'll push back a bit on the assertion that "most of the time sticky notes are used in grids" — I think certainly there's a convention in other canvas apps to do so but i don't know if necessarily that's how people expect them to behave? (i think to our retros, for example). the first principle that i would like to convey is that informality/ephemerality/frictionless/flow-of-thought/dont-worry-about-styling-just-worry-about-writing. so, this is one of those things that would be fun to add as an experiment toggle so that folks can live with it, maybe sit with the discomfort of it for a bit and see if it helps give a certain feeling to stickies. again though, i'll stress that i'm not married to this idea, just something to consider and make us think about first principles for stickies.

> Autosizing also feels like a miss for the same reason. In Miro / Freeform, sticky notes don't grow at all, and the text shrinks to fit the sticky. In Figma, notes grow on their y axis as text gets bigger.

same reason as rotation? i'm not sure i follow - can you rephrase? but anyway, i think this is something that _is_ new and i think we can do better than other apps do. Miro / Freeform are rubbish for the text shrinking — i think that's a terrible way to do it (for a11y, aesthetics, everything). you mention Miro and they do do snapped sizing if you resize horizontally directly (16:9 vs. 1:1 aspect ratios). looking at prior art here (Miro, Figma, etc.) i think might be a trap — i don't think anybody has nailed this yet and we should make this work better than other people make it work.

the problem is that if the sticky grows longer then it is no longer a Sticky, at least as no one knows it in the real world paper equivalent. if someone were to visit our canvas and see those randomly sized boxes for the first time, they would not convey their Sticky-ness - they would just be boxes. this goes back to the first part of skeuomorphism. if we diverge from the real-life equivalent then it starts defeating the purpose of emulating/referencing that real-life paper equivalent.

> Most apps do resize on drag and I'm really not a fan of that. IMO the current drag to position is more useful (Figjam style). I haven't looked at the implementation here but the interactions are kind of unpredictable? I do like the idea of creating multiple sticky notes at the same time. Could we have a different way of creating, ie a grid, maybe while holding shift?

hm, maybe this is a misunderstanding? this drag action is only upon creation. the current behavior is _translation_ on creation+drag and i think that's a miss, or at least it doesn't feel useful/intuitive as it currently stands. (i.e. nothing else behaves like that afaik). let's talk more offline about this — we might not be talking about the same thing here possibly.

> Good improvement! Can be adjusted to better match the other icons that we have (size, maybe radius).

i'm not an icon-smith :) this is just from that icon8 library — can you help with this bit?

> I'll have spend some time with this, seems like a big bet on the defaults.

yeah, i think i'd like to live with this for a bit. i'm drawing on my Paper experience and going minimal to start with (again if this is an experimental direction to start with). allowing more options as defaults means more rope to hang yourself with and hard to retract it later if we change our minds. if we start more minimal we can see if we feel pain in not being able to change the font size / alignment; or, maybe we'll find out that we actually don't miss it and we never really changed it in the first place. but it would be good to "feel that pain" as it were, and see which UI bits actually are essential vs. not.

PR #2957, opened 2024-02-26
handles prototype 1: no gods no handles

SomeHats:
(i've not actually removed the handles code in this pr - just take a look at the speech bubble example for what a new API might look like)

in this version, I've added a couple of very light-weight additions to the library:
- A couple of utilities for mutating the state chart (`.find`, `.addChild`)
- A `<HandleControl position onPointerDown />` component designed to be rendered with `InFrontOfTheCanvas`

Then, in the example i've implemented what is effectively a handle on top of that:
- the `<SpeechBubbleHandle />` component adds its node to the state chart on mount
- it renders the handle when a speech bubble (and only a speech bubble) is selected
- when someone interacts with the speech bubble, we block the event from reaching tldraw's normal state chart, and instead add our own transition into the new state node

This works pretty well! There are some big drawbacks though:
- blocking the event from reaching our state chart means we don't get things like pointer capture, updates to `inputs`, gesture recognition, etc
- requiring users to pass this in to `OnTopOfCanvas` is a pain, and wouldn't scale to doing this for every single shape
- the state tree is conditional on react - so we can't test easily
- imo our state tree is pretty cumbersome to work with - it's one of the oldest parts of this codebase, has a bunch of weird quirks where the abstraction doesn't quite fit the use case (e.g. gesture recognition and tool masks), and I think could do with the same treatment that selection & ui composition have received

### DIFF
--- a/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleHandle.tsx
+++ b/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleHandle.tsx
@@ -1,0 +1,103 @@
+import { HandleControl, StateNode, Vec, track, useEditor } from '@tldraw/tldraw'
+import { useEffect } from 'react'
+import { SpeechBubbleShape } from './SpeechBubbleUtil'
+
+export const SpeechBubbleHandle = track(function SpeechBubbleHandle() {
+	const editor = useEditor()
+
+	useEffect(() => {
+		editor.root.find('select')!.addChild(DraggingSpeechBubble)
+	}, [editor])
+
+	if (!editor.isInAny('select.idle')) {
+		return null
+	}
+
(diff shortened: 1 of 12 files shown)


apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleHandle.tsx
```diff
@@ -0,0 +1,103 @@
...
+				e.stopPropagation()
+
+				editor.root.transition('select.draggingSpeechBubble')
+			}}
+		/>
+	)
+})
+
+// missing from here:
+// - tool masks (what even is this?)
```
steveruizok, line 41: Which tool should be shown as selected when this state is active? For example, creating an arrow technically goes to the select.dragging_handle, but we display the arrow tool as active.

PR #2969, opened 2024-02-27
[experiment] Accoutrements

SomeHats:
In #2957, we implemented some of the functionality of the speech bubble shape as a separate react component that needs to be passed in as a top-level `OnTopOfCanvas` component. This works well, but it means that if you're using this shape (or e.g. our own shapes were built in this way) you'd need to bring in both the shape and the `OnTopOfCanvas` component. It makes it very hard to distribute something like a custom shape as its own npm package.

One solution here would be a way to bundle together related customisations that make up a coherent feature into a single package that could then be plugged-in to the editor. These plug-in-able things are called "accoutrements".

Accoutrements are a purely additive change - they work with our existing extension methods, but let you bundle things up into a single object if they don't make sense on their own. They're fully backwards-compatible with our existing extension methods. You could still pull out individual parts if you wanted e.g. the shape definition, but wanted to supply your own interactions on top of it. 

(the exact API of defining and using these is up for debate! i just did the most minimal thing i could for this diff to illustrate the concept)

### DIFF
--- a/apps/examples/src/examples/speech-bubble/CustomShapeWithHandles.tsx
+++ b/apps/examples/src/examples/speech-bubble/CustomShapeWithHandles.tsx
@@ -1,23 +1,17 @@
 import { Tldraw } from '@tldraw/tldraw'
 import '@tldraw/tldraw/tldraw.css'
-import { SpeechBubbleTool } from './SpeechBubble/SpeechBubbleTool'
-import { SpeechBubbleUtil } from './SpeechBubble/SpeechBubbleUtil'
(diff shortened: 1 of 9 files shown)


mimecuvalo:
this is pretty cool. i could see this being useful in the plugin world. fwiw, probably the more tech-y name for accoutrements would be "mods" :)

SomeHats:
mods, DLC, extensions, plugins, addons, expansion packs, patches, tapas, skins, widgets, side dishes,..... uhhhh......... soup de jour

PR #2970, opened 2024-02-27
fix document name overlapping people menu

Taha-Hassan-Git:
When the people menu grew too large it would [overlap the document name](https://github.com/orgs/tldraw/projects/38/views/1?pane=issue&itemId=54609134)

This PR checks if the right layout panel has grown beyond the style panel width (plus the width of the button) and includes the button width in the calculation if so.

- [x] `patch` — Bug fix

### Release Notes

- Fix people menu overlapping with document name when it grew too large.

### DIFF
--- a/apps/dotcom/src/components/DocumentName/DocumentName.tsx
+++ b/apps/dotcom/src/components/DocumentName/DocumentName.tsx
@@ -37,6 +37,7 @@ type NameState = {
 
 const MAX_TITLE_WIDTH_PX = 420
 const BUTTON_WIDTH = 44
+const STYLE_PANEL_WIDTH = 148
 const MARGIN_BETWEEN_ZONES = 12
 
 export const DocumentTopZone = track(function DocumentTopZone({
@@ -135,8 +136,13 @@ function DocumentTopZoneContainer({ children }: { children: ReactNode }) {
 		const totalWidth = layoutTop.offsetWidth
 		const leftWidth = leftPanel.offsetWidth
 		const rightWidth = rightPanel.offsetWidth
-		// ignore the width of the button:
-		const selfWidth = element.offsetWidth - BUTTON_WIDTH
+		let selfWidth = 0
+
+		// We want to ignore the width of the button to make the title centered
+		// unless the right panel has grown beyond the style panel width, in which
+		// case we want to include the button width in the calculation
+		if (rightWidth > STYLE_PANEL_WIDTH + BUTTON_WIDTH) selfWidth = element.offsetWidth
+		else selfWidth = element.offsetWidth - BUTTON_WIDTH
 
 		let xCoordIfCentered = (totalWidth - selfWidth) / 2
 


TodePond (approved):
nice, happy to merge!

we could probably squeeze it even further to the left if we want, so that its bounds don't overlap

image

Taha-Hassan-Git:
This is better, I think. No more overlapping.

TodePond (approved):
oh nice! thats a load simpler

MitjaBezensek:
This PR changes the default position of the document name. I guess it's now in the middle between the panels where as before it was in the middle of the board. Not sure if that's a desired change? I know [Steve had some thoughts](https://github.com/tldraw/tldraw/pull/2685#issuecomment-1923376555) on this.

TodePond:
> I guess it's now in the middle between the panels where as before it was in the middle of the board.

Ah yeah good point, we should probably keep the position as the center. It just needs to not expand to overflow the presence indicators.

Taha-Hassan-Git:
> This PR changes the default position of the document name. I guess it's now in the middle between the panels where as before it was in the middle of the board. Not sure if that's a desired change? I know [Steve had some thoughts](https://github.com/tldraw/tldraw/pull/2685#issuecomment-1923376555) on this.

Hmm, I wasn't thinking about that. I won't have time to take another run at this today, though. Someone else could take over or I'll take a look tomorrow.

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -148,14 +150,19 @@ function DocumentTopZoneContainer({ children }: { children: ReactNode }) {
...
 		const left = element.offsetLeft
-		const xCoord = Math.max(xCoordIfCentered, xCoordIfLeftAligned) - left
 		const maxWidth = Math.min(
 			totalWidth - rightWidth - leftWidth - 2 * MARGIN_BETWEEN_ZONES,
 			MAX_TITLE_WIDTH_PX
 		)
+		const xCoord = Math.max(xCoordIfCentered, xCoordIfLeftAligned) - left
 
+		// Squeeze the title if the right panel is too wide on small screen
+		if (rightPanel.offsetWidth > STYLE_PANEL_WIDTH && totalWidth < 841) {
```
mimecuvalo, line 160: can you make these numbers into constants? 841 is very specific and we won't know what it means later on :)
same for the 60 below

Taha-Hassan-Git: Done, I had to go and figure out why it was that 60 worked and ended up changing it to 52.

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -39,6 +39,9 @@ const MAX_TITLE_WIDTH_PX = 420
 const BUTTON_WIDTH = 44
 const STYLE_PANEL_WIDTH = 148
 const MARGIN_BETWEEN_ZONES = 12
+const SMALL_SCREEN_WIDTH = 841
```
mimecuvalo, line 42: thx @Taha-Hassan-Git!
actually, i might ask you to make this change so that this number doesn't get of sync with our canonical constant:
`PORTRAIT_BREAKPOINTS[PORTRAIT_BREAKPOINT.TABLET]`

apps/dotcom/src/components/DocumentName/DocumentName.tsx
```diff
@@ -157,8 +160,8 @@ function DocumentTopZoneContainer({ children }: { children: ReactNode }) {
 		const xCoord = Math.max(xCoordIfCentered, xCoordIfLeftAligned) - left
 
 		// Squeeze the title if the right panel is too wide on small screen
-		if (rightPanel.offsetWidth > STYLE_PANEL_WIDTH && totalWidth < 841) {
-			element.style.setProperty('max-width', maxWidth - 60 + 'px')
+		if (rightPanel.offsetWidth > STYLE_PANEL_WIDTH && totalWidth < SMALL_SCREEN_WIDTH) {
```
mimecuvalo, line 163: and then down below here it would be `<=` instead of `<`

apps/dotcom/src/components/DocumentName/DocumentName.tsx
Taha-Hassan-Git: I can't do that because of a lint rule around importing from our own package. Having to use the useBreakpoint hook.

mimecuvalo (approved):
ty!

PR #2971, opened 2024-02-27
migrate shapes / assets as a store on `putContent`

steveruizok:
This PR updates the `putContentOntoCurrentPage` so that it migrates shapes / records as a complete store.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Copy and paste, ideally between versions.


ds300 (approved):
👍🏼 I remembered I had done exactly this in my migrations PR. Thanks for pulling it out into a separate PR.

PR #2973, opened 2024-02-27
export default ui items

Taha-Hassan-Git:
This PR exports all the components within each of the default menu content components. Should make it easier to customise the default UI.

- [x] `minor` — New feature

### Release Notes

- Components within default menu content components are now exported.


SomeHats (approved):
Thanks Taha!

mimecuvalo:
yassss thanks!
it does make me think of whether we should create a `@tldraw/ui` package?

PR #2975, opened 2024-02-27
speech bubble handle -> tail

SomeHats:
Handle's days are numbered, and in the line shape we've moved away from storing `TLHandle` directly. This diff updates the speech bubble shape to rename the 'handle' prop to 'tail' and make it just be the coordinates. The handle props are derived at runtime.

### Change Type


Taha-Hassan-Git (approved):
My shape! 🥹

PR #2977, opened 2024-02-27
Performance improvements

MitjaBezensek:
This PR does a few things to help with performance:
1. Instead of doing changes on raf we now do them 60 times per second. This limits the number of updates on high refresh rate screens like the iPad. With the current code this only applied to the history updates (so when you subscribed to the updates), but the next point takes this a bit futher.
2. We now trigger react updates 60 times per second. This is a change in `useValue` and `useStateTracking` hooks.
3. We now throttle the inputs (like the `pointerMove`) in state nodes. This means we batch multiple inputs and only apply them at most 60 times per second.

We had to adjust our own tests to pass after this change so I marked this as major as it might require the users of the library to do the same.

Few observations:
- The browser calls the raf callbacks when it can. If it gets overwhelmed it will call them further and further apart. As things call down it will start calling them more frequently again. You can clearly see this in the drawing example. When fps gets to a certain level we start to get fewer updates, then fps can recover a bit. This makes the experience quite janky. The updates can be kinda ok one second (dropping frames, but consistently) and then they can completely stop and you have to let go of the mouse to make them happen again. With the new logic it seems everything is a lot more consistent.
- We might look into variable refresh rates to prevent this overtaxing of the browser. Like when we see that the times between our updates are getting higher we could make the updates less frequent. If we then see that they are happening more often we could ramp them back up. I had an [experiment for this here](https://github.com/tldraw/tldraw/pull/2977/commits/48348639669e556798296eee82fc53ca8ef444f2#diff-318e71563d7c47173f89ec084ca44417cf70fc72faac85b96f48b856a8aec466L30-L35).

Few tests below. Used 6x slowdown for these.

# Resizing

### Before

https://github.com/tldraw/tldraw/assets/2523721/798a033f-5dfa-419e-9a2d-fd8908272ba0

### After

https://github.com/tldraw/tldraw/assets/2523721/45870a0c-c310-4be0-b63c-6c92c20ca037

# Drawing 
Comparison is not 100% fair, we don't store the intermediate inputs right now. That said, tick should still only produce once update so I do think we can get a sense of the differences.

### Before

https://github.com/tldraw/tldraw/assets/2523721/2e8ac8c5-bbdf-484b-bb0c-70c967f4541c

### After

https://github.com/tldraw/tldraw/assets/2523721/8f54b7a8-9a0e-4a39-b168-482caceb0149

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Release Notes

- Improves the performance of rendering.


packages/utils/src/lib/raf.ts
```diff
@@ -6,26 +6,101 @@ const isTest = () =>
...
 
-function raf() {
+function sixtyFps() {
 	if (frame) {
 		return
 	}
+	const now = Date.now()
+	const elapsed = now - last
 
+	// Variable frame rate
```
mimecuvalo, line 66: i'd be curious about this commented out code for variable frame rate — what did you learn about this technique? helped/didn't help?

packages/utils/src/lib/raf.ts
```diff
@@ -6,26 +6,101 @@ const isTest = () =>
...
+	console.log('time since last tick', timeSinceLastTick, fps)
+	lastFrameTime = now
+
 	const queue = rafQueue.splice(0, rafQueue.length)
 	for (const fn of queue) {
 		fn()
 	}
 }
 
+const fps = 60
```
mimecuvalo, line 52: i see down below that the you account for the frame rates falling below the target frame rate.

does starting the target at 60 hurt any machines that can't get that fast? if a machine has more a frame rate like 30, they basically go through the fallback path of lines 82-87, is that right? any downside?

mimecuvalo (commented):
this is sick @MitjaBezensek ! steve was trying it out on the iPad and it's working like butter. couple comments for my curiousity!

packages/utils/src/lib/raf.ts
MitjaBezensek: I think 60 is a good starting point. This affects all updates, even stuff like dragging the mouse cursor around, which probably can be handled even by slow devices. So I'd say we want that as default.

For those more expensive operations:
- Like mentioned in the PR description the browser starts throttling the calls to RAF callbacks. This happens at around 10-20 fps. So when that happens our code also gets throttled.
- We can also look into the mentioned variable frame rate, which should adjust how much we tax the browser.

That said it's a bit hard to test all of this. Like how does the variable frame rate interact with browser throttling? My guess is that if we adjust the refresh rate we should be taxing the browser less and thus hit the throttling later or maybe not hit it at all in some cases. We might also land in the sweet spot of what the device is able to do - this is probably also what happens due to browser throttling, but unsure 🤷‍♂️ 😂

packages/utils/src/lib/raf.ts
MitjaBezensek: Hard to tell still, just added this a bit before the standup. Will have to play more with it. My feeling is that it probably should help a bit. That said it might be a bit hard to hit the numbers right. Like we don't want to slow down the updates just because the last frame tool 1ms longer. Then also when do we start increasing the refresh rate? 

Think we should definitely explore it more.

apps/examples/package.json
```diff
@@ -29,8 +29,8 @@
 		"dev": "vite --host",
 		"build": "vite build",
 		"lint": "yarn run -T tsx ../../scripts/lint.ts",
-		"e2e": "playwright test -c ./e2e/playwright.config.ts",
-		"e2e-ui": "playwright test --ui -c ./e2e/playwright.config.ts"
+		"e2e": "NODE_ENV=test && playwright test -c ./e2e/playwright.config.ts",
```
MitjaBezensek, line 32: Fixes [some tests that started failing](https://github.com/tldraw/tldraw/actions/runs/8158802282/job/22301557609) as a result of these changes. This [makes stuff run sync](https://github.com/tldraw/tldraw/blob/mitja%2Fraf/packages/utils/src/lib/throttle.ts#L3) instead of on the next frame.

steveruizok:
Anything left on this one?

MitjaBezensek:
Yes, it's now ready for review 😄

packages/utils/src/lib/throttle.ts
```diff
@@ -0,0 +1,96 @@
...
+let last = 0
+
+const tick = () => {
+	const queue = fpsQueue.splice(0, fpsQueue.length)
+	for (const fn of queue) {
+		fn()
+	}
+}
+
+function fps() {
```
ds300, line 21: extremely nonblocking nitpick:

I found the naming here a little odd. I'd personally change `tick` to `flush` and change `fps` to `tick`. For me a `tick` is some kind of regularly scheduled event for managing asynchronous execution of other things, which is kinda what your `fps` function is doing. 

I dunno, could just be me though 🤷🏼

packages/utils/src/lib/throttle.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Calls the function on the next frame. The target frame rate is 60fps.
+ * If the same fn is passed again before the next frame, it will still be called only once.
+ * @param fn - the fun to call on the next frame
+ * @returns
+ * @internal
+ */
+export function throttleToNextFrame(fn: () => void) {
```
ds300, line 80: yes!

`throttleToNextFrame` and `fpsThrottle` is a much less confusing naming scheme than what we had before 🙌🏼

packages/tldraw/src/lib/tools/SelectTool/childStates/ScribbleBrushing.ts
```diff
@@ -51,8 +53,15 @@ export class ScribbleBrushing extends StateNode {
 		this.editor.scribbles.stop(this.scribbleId)
 	}
 
+	override onTick = () => {
+		if (this.isDirty) {
+			this.isDirty = false
+			this.updateScribbleSelection(true)
+		}
```
ds300, line 60: I wonder whether we should run these in `onComplete` too? To get the absolute final positions. Probably not gonna be noticeable unless the frame rate gets quite low tbh.

ds300 (approved):
Looks great! I wonder if we can simplify the state chart bits by supporting an `onThrottledMouseMove` callback method that does the dirty checking and uses the tick under the hood, to simplify this kind of workflow that you've had to reproduce in a bunch of places? Might be more trouble than it's worth and can definitely be done separately from this PR so not blocking in any way.

packages/utils/src/lib/throttle.ts
MitjaBezensek: Proposed names definitely make more sense to me, thanks for the suggestion.

PR #2980, opened 2024-02-27
Configure dependabot.

MitjaBezensek:
Adds dependabot config. 

Seems like monthly is the least frequent update you can choose right now (there are feature request for quarterly and for chron like syntax). 

For now I used the daily interval though. I want to make sure it works, will then switch to monthly. The main thing I'd like to see is that the main `yarn.lock` file gets updated. That didn't happen in one of the previous dependabot PRs (seems like [others faced the same issue](https://github.com/dependabot/dependabot-core/issues/6346)). So 🤞  this solves it.

Also added a dedupe workflow, that should only run on dependabot branches (prefixed with `depandabot/`). Otherwise we would need to manually do it, as [seen on this PR](https://github.com/tldraw/tldraw/pull/2982) ([failure](https://github.com/tldraw/tldraw/actions/runs/8070271847/job/22047204003)). 

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,5 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/'
     schedule:
-      interval: 'monthly'
-    # Always increase the version requirement
-    # to match the new version.
-    versioning-strategy: increase
+      interval: 'daily'
+    versioning-strategy: auto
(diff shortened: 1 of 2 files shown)


mimecuvalo:
neat!

SomeHats (approved):
looks good to me!

PR #2984, opened 2024-02-28
textfields: make them consistent

mimecuvalo:
as i look into textfields the places where we are using them are out-of-sync at the moment. this brings them in line with each other for further refactoring later.

### Change Type

- [x] `patch` — Bug fix


SomeHats (approved):
very excited for whatever this is in preparation for 👀

PR #2989, opened 2024-02-28
[experiment] Controls

SomeHats:
Controls are a sort of combined virtual shape / mini state node that render on top of the canvas? IDK. I was into this when i first thought about it but now i'm not so sure.

The idea: call `addControls` with a reactive function that returns some number of Control objects. You can think of this function as being similar to a react component, e.g.
```ts
editor.addControls(() => {
    if (!editor.isIn('select.idle')) return null
    return new MyControl()
})
```
is conceptually similar to
```ts
function MyControlWrapper() {
    if (!editor.isIn('select.idle')) return null
    return <MyControl />
}
```

The control object itself has geometry, a component, and can store any information it needs too.

When a pointer event comes in, if we're hovering a control, then the event is sent there first. From there, we can use normal state chart modifications or do whatever else we want. I had to add a `.preventDefault()` to get this to play nice. idk about it. very buggy and rough, but check the speech bubble example for a demo of how this might work


mimecuvalo:
neat exploration! i'd be curious to talk more about this option with you @SomeHats and your hesitation now that you've implemented it. i think initial sketch of it seems compelling to me as a first impression. i don't like how it has to be wired up via `onMount` but i could be convinced?

PR #2990, opened 2024-02-28
Show a broken image for files without assets

steveruizok:
This PR shows a broken state for images or video shapes without assets. It deletes assets when pasted content fails to generate a new asset for the shape. It includes some mild refactoring to the image shape. Previously, shapes that had no corresponding assets would be transparent. This PR preserves the transparent state for shapes with assets but without source data (ie loading assets).

After:
image
image

Before:
image

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Create an image / video
2. Delete its asset

### Release Notes

- Better handling of broken images / videos.


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7890,54 +7891,57 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 				)
 
+				// Get a new asset for the file
 				const newAsset = await this.getAssetForExternalContent({ type: 'file', file })
 
 				if (!newAsset) {
-					return null
+					// If we don't have a new asset, delete the old asset.
+					// The shapes that reference this asset should break.
+					this.deleteAssets([asset.id])
```
steveruizok, line 7937: This is what's new, btw

PR #2995, opened 2024-02-29
Don't add editor / app to window.

steveruizok:
This PR removes code that would add a reference to the editor to the window. This is a feature that we added very early on during testing, but which we should have moved out of the library earlier. Adding it here as one of our last PRs before release.

If you've relied on this, you'll need to update your use of the library to do it manually:

```ts
<Tldraw onMount={(editor) => {
  ;(window as any).app = editor
}}/>
```

### Change Type

### Release Notes

- Remove `window.editor` and `window.app` references to editor.


mimecuvalo (approved):
i think i agree on this change. i wonder if anybody's been relying on this mechanism ¯\\_(ツ)_/¯

PR #3011, opened 2024-02-29
Update readmes / docs for 2.0

steveruizok:
This PR updates the readmes for our 2.0 release.

### Change Type


README.md
```diff
@@ -21,22 +37,22 @@ To run the local development server, first clone this repo.
 Install dependencies:
 
 ```bash
-yarn
+yarn i
```
mimecuvalo, line 40: @steveruizok we use `yarn` here but use `npm` above - we should prbly be consistent - otherwise it'll be confusing - which one should we go with?

mimecuvalo: actually, we should specify `yarn` - a lot of our tooling/docs assumes yarn

mimecuvalo:
( @steveruizok ping: just to make sure you see my comment on this )

PR #3017, opened 2024-03-01
Example of using tldraw styles

Taha-Hassan-Git:
Adds an example of how to use tldraw styles in a custom shape

### Release Notes

- shape with tldraw styles example


steveruizok:
This was good! I did some copy changes, added a second example for using custom props, then removed the old custom styles example.

PR #3018, opened 2024-03-02
fix refresh-assets executing errors on Windows

cscxj:
# refresh-assets script execution error on Windows

`await import(join(REPO_ROOT, 'packages', 'tldraw', 'package.json'))` cannot execute correctly on Windows, resulting in project startup failure.

The received error message is:
`Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'`


cscxj:

steveruizok:
Will this have any effect on Mac / Linux builds? Seems to work fine on mine but maybe @si14 you've got an idea here?

cscxj:
Verified: This change operates smoothly on both MacOS and Ubuntu 22.04.

scripts/refresh-assets.ts
```diff
@@ -402,7 +402,7 @@ async function writeAssetDeclarationDTSFile() {
 }
 
 async function copyVersionToDotCom() {
-	const packageVersion = await import(join(REPO_ROOT, 'packages', 'tldraw', 'package.json')).then(
+	const packageVersion = await import('file://' + join(REPO_ROOT, 'packages', 'tldraw', 'package.json')).then(
 		(pkg) => pkg.version
 	)
```
SomeHats, line 407: Rather than using `import` to load a JSON file like this, lets use our `readJsonIfExists` helper:

```suggestion
	const packageVersion = (await readJsonIfExists(join(REPO_ROOT, 'packages', 'tldraw', 'package.json'))).version
```

SomeHats (changes requested):
should be fine on mac! using `import` to read json is kinda weird though, im not sure why we're doing that. We can avoid messing with the file protocol by just reading the file normally instead

si14:
Hey @cscxj! Thanks a ton for taking your time to send us a fix. I just applied the fix @SomeHats proposed in a separate PR #3116, which will be merged in a few minutes.

Again, thank you for the PR!

PR #3020, opened 2024-03-02
[experiment] Use DOM parser for SVG exports

steveruizok:
This PR gradually moves our shapes' `toSvg` methods to use DOM parsing instead of manually creating elements.

Pros:
- less code
- more readable code
- general refactoring

Cons:
- theoretically slower, though everything I've read suggests the differences only occur at high volumes

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Compare SVG exports before / after

### Release Notes

- Fixes a bug with SVG export on vertical arrows with labels


steveruizok:
Regarding the warnings above: 
- Do we care about XSS for SVG exports?
- Is our validation robust enough to permit this?

steveruizok:
Weirdly this requires writing namespaces by hand?

steveruizok:
It looks like we'll have to add the namespace ourselves if we want this to work.

steveruizok:
Another thing we could do is make the `toSvg` return a string, and combine the strings ourselves in the editor's `getSvg` method. At the moment we're constructive an svg element, then converting it to a string anyway before doing anything else with it.

PR #3038, opened 2024-03-02
[fix] Input tags

steveruizok:
This PR fixes the input tags, which were set to `false` rather than `off`, as they should be.

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Test autocomplete
2. Test autocapitalize
3. Test autocorrect
4. Make sure that password managers don't show up

### Release Notes

- Fixed autocomplete, autocapitalize, and autocorrect tags on text inputs.

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/components/ArrowTextLabel.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/components/ArrowTextLabel.tsx
@@ -61,10 +61,10 @@ export const ArrowTextLabel = React.memo(function ArrowTextLabel({
 						className="tl-text tl-text-input"
 						name="text"
 						tabIndex={-1}
-						autoComplete="false"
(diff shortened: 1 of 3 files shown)


mimecuvalo:
i venture to guess that this was `false` because some older browsers behaved not-as-expected when it was `off`
https://stackoverflow.com/questions/30053167/autocomplete-off-vs-false

so, whilst `off` is more correct, there was a tendency in the community to use false because of browser quirks

SomeHats (approved):
looks good!

mimecuvalo:
(yeah, to be clear, i think this is fine, i'm just saying that prbly the historical context here was that quirk that the original code was trying to deal with. i agree that that browser quirk has since gone way)

steveruizok:
I didn't know the history, appreciated (it will come up, I guarantee it)

PR #3042, opened 2024-03-02
Fix cursor chat bubble position.

steveruizok:
This PR fixes the position of the cursor chat bubble when the canvas is not positioned at the top left.

### Change Type

- [x] `internal`

### Test Plan

1. Using CSS, add a margin left to the tldraw component on a multiplayer route.
2. Use cursor chat.

### Release Notes

- Fixed a bug where cursor chat bubble position could be wrong when a sidebar was open.

### DIFF
--- a/apps/dotcom/src/components/CursorChatBubble.tsx
+++ b/apps/dotcom/src/components/CursorChatBubble.tsx
@@ -188,12 +188,9 @@ const CursorChatInput = track(function CursorChatInput({
 		[stopChatting, setValue]
 	)
 
-	const handlePaste = useCallback(
-		(e: ClipboardEvent) => {
-			e.stopPropagation()
-		},
-		[setValue]
-	)
+	const handlePaste = useCallback((e: ClipboardEvent) => {
+		e.stopPropagation()
+	}, [])
 
 	return (
 		<input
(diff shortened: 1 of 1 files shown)


apps/dotcom/src/components/CursorChatBubble.tsx
```diff
@@ -184,11 +188,14 @@ const CursorChatInput = track(function CursorChatInput({
...
-		e.stopPropagation()
-	}, [])
+	const handlePaste = useCallback(
+		(e: ClipboardEvent) => {
+			const text = e.clipboardData.getData('text/plain')
+			setValue(text)
+			e.stopPropagation()
+		},
+		[setValue]
+	)
```
TodePond, line 198: This is triggering the paste twice now, and overriding the entire chat bubble with your paste.

I think we added the max message length since this todo was placed here, so I reckon we can just remove this paste handler, and allow pastes to happen as normal now

```suggestion
	const handlePaste = useCallback(
		(e: ClipboardEvent) => {
			e.stopPropagation()
		},
		[setValue]
	)
```

steveruizok: Paste wasn't happening at all when I tried it, I thought maybe we had punted on paste? I'll roll back this, we can address it separately.

steveruizok:
Thanks @TodePond !

PR #3043, opened 2024-03-02
Protect local storage calls

steveruizok:
This PR provides some safe wrappers for local storage calls. Local storage is not available in all environments (for example, a React Native web view). The PR also adds an eslint rule preventing direct calls to local / session storage.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Fixes a bug that could cause crashes in React Native webviews.

### DIFF
--- a/packages/utils/src/lib/local-storage.ts
+++ b/packages/utils/src/lib/local-storage.ts
@@ -1,0 +1,29 @@
+const hasLocalStorage = typeof window === 'undefined' || !('localStorage' in window)
+
+/**
+ * Get a value from local storage.
+ *
+ * @param key - The key to get.
+ * @param defaultValue - The default value to return if the key is not found or if local storage is not available.
+ *
+ * @public
+ */
+export function getFromLocalStorage(key: string, defaultValue = null) {
+	if (!hasLocalStorage) return defaultValue
+	const value = localStorage.getItem(key)
+	if (value === null) return defaultValue
+	return JSON.parse(value)
+}
+
+/**
+ * Set a value in local storage. Will not throw an error if localStorage is not available.
+ *
+ * @param key - The key to set.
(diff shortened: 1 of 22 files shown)


SomeHats (approved):
looks good!

SomeHats:
For the test issues, rather using the `if` guard i'd wrap the whole thing in a try/catch and return the default value from the catch branch

packages/utils/src/lib/local-storage.ts
```diff
@@ -0,0 +1,55 @@
+const hasLocalStorage = typeof window === 'undefined' || !('localStorage' in window)
```
si14, line 1: isn't this the wrong way around?

packages/utils/src/lib/session-storage.ts
```diff
@@ -0,0 +1,55 @@
+const hasSessionStorage = typeof window === 'undefined' || !('sessionStorage' in window)
```
si14, line 1: same here

packages/utils/src/lib/local-storage.ts
steveruizok: ha yeah, renaming the variable

packages/utils/src/lib/session-storage.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Get a value from session storage.
+ *
+ * @param key - The key to get.
+ * @param defaultValue - The default value to return if the key is not found or if session storage is not available.
+ *
+ * @public
+ */
+export function getFromSessionStorage(key: string, defaultValue = null) {
```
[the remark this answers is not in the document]
steveruizok: Not sure we need to?

steveruizok:
Thanks @alex @si14, updated

PR #3050, opened 2024-03-04
textfields [1 of 3]: add text into speech bubble; also add rich text example

mimecuvalo:
This is the first of three textfield changes. This starts with making the speech bubble actually have text. Also, it creates a TipTap example and how that would be wired up.

🎵 this is dangerous, I walk through textfields so watch your head rock 🎵 

### Change Type

- [x] `minor` — New feature

### Release Notes

- Refactor textfields be composable/swappable.


SomeHats:
If we're going to add these examples, I'd like them to demonstrate a slightly more complete idea of what adding text to a shape might involve and the type of bugs you might run into e.g.

How should the shape change as the text grows?
Kapture 2024-03-04 at 12 24 42

Focus management is hard
Kapture 2024-03-04 at 12 29 57

packages/tldraw/src/lib/shapes/text/TextArea.tsx
```diff
@@ -0,0 +1,49 @@
...
+	rInput,
+	text,
+	handleFocus,
+	handleChange,
+	handleKeyDown,
+	handleBlur,
+	handleInputPointerDown,
+	handleDoubleClick,
+}: {
+	rInput: React.RefObject<HTMLTextAreaElement>
```
steveruizok, line 13: forward ref instead here?

steveruizok: (I mean they're taking it out in a few months anyway, but for now!)

steveruizok:
Not sure what the priorities are set to, but could we push the more complex speech bubble example down?

image

mimecuvalo:
- fixed up growing for shapes. the rich text one isn't 100% perfect for every edge case but it's meant to be an example so I figure it doesn't have to reach 100% there, it's meant to be a demo.
- fixed up focus management, good flag @SomeHats 
- fixed up the forwardRef, doh - thx @steveruizok 
- changed priority of examples

packages/tldraw/src/lib/shapes/text/TextArea.tsx
mimecuvalo: hah doh, not sure what i was smoking here - reworked!

mimecuvalo:
The refactor work in this PR would help support https://github.com/tldraw/tldraw/pull/3132 (another reason I was working on textfields)
We should land this first PR at least, please! @steveruizok it's waiting your approval ✅ - thx!

steveruizok:
Let's split this up!

Let's land:
- the changes that turn our <textareas> into a TextArea component
- the simple example

Let's not land:
- the changes to TextEditor
- the rich text example

steveruizok (approved):

PR #3057, opened 2024-03-04
Remove dependabot config since it only controls version updates?

MitjaBezensek:
My last PR added dependabot config. But it seems it only controls the version updates, which we probably don't want for now. So I'm removing this config for now. I guess security update frequency can't really be configured since they are meant as urgent and should be merged asap?

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: 'npm'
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
hm, gotcha - that's too bad (also confusing?)

PR #3061, opened 2024-03-04
children: any -> children: ReactNode

SomeHats:
We use `children: any` in a bunch of places, but the proper type for these is `ReactNode`. This diff fixes those.

### Change Type

- [x] `patch` — Bug fix

### DIFF
--- a/apps/docs/app/providers.tsx
+++ b/apps/docs/app/providers.tsx
@@ -1,7 +1,8 @@
 'use client'
 
 import { ThemeProvider } from 'next-themes'
+import { ReactNode } from 'react'
 
-export function Providers({ children }: { children: any }) {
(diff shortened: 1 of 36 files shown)


steveruizok:
The only place where this still breaks types is render props, though this is pretty rare:

image

steveruizok (approved):

steveruizok:
I'll have a hell of a time not typing children as any, deep muscle memory.

SomeHats:
> The only place where this still breaks types is render props, though this is pretty rare:

you want them to error on render props - react doesn't know what to do with them by default, so you'll get an error. If you want to use a render prop, you can do `children: () => ReactNode` to tell react you're expecting to be given a function as a child

mimecuvalo:
nice one @SomeHats

PR #3064, opened 2024-03-04
Add custom tool examples

Taha-Hassan-Git:
Two examples:

One in the UI section that shows how to add a tool to the toolbar along with an icon

One in the shapes and tools section that shows a simple sticker tool with no child states

I'll go over the copy again before it's merged, but don't want to spend too long on it right now in case the feeling is that these should both be a single example.

Next: The [minimal example](https://tldraw.dev/examples/editor-api/only-editor) is currently the best example we have of a tool with child states. I think this should be adapted and copied/moved over to the custom shapes and tools category.

closes tld-2266

### Release Notes

- Adds a simple custom tool example


steveruizok:
Looks good!

steveruizok (approved):

PR #3065, opened 2024-03-04
Fix validation errors for `duplicateProps`

MitjaBezensek:
Should fix `At instance.duplicateProps.offset.x: Expected a number, got NaN` validation errors.

Wasn't able to reproduce. We only assign the offset here, so `Vec.Averge` is the most likely offender here and for that to happen I guess `movingShapes` might not contain any shapes.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
@@ -213,17 +213,19 @@ export class Translating extends StateNode {
 	protected handleEnd() {
 		const { movingShapes } = this.snapshot
 
-		if (this.isCloning) {
+		if (this.isCloning && movingShapes.length > 0) {
 			const currentAveragePagePoint = Vec.Average(
 				movingShapes.map((s) => this.editor.getShapePageTransform(s.id)!.point())
 			)
 			const offset = Vec.Sub(currentAveragePagePoint, this.selectionSnapshot.averagePagePoint)
-			this.editor.updateInstanceState({
-				duplicateProps: {
-					shapeIds: movingShapes.map((s) => s.id),
-					offset: { x: offset.x, y: offset.y },
-				},
-			})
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
```diff
@@ -338,7 +340,7 @@ function getTranslatingSnapshot(editor: Editor) {
 	}
 
 	return {
-		averagePagePoint: Vec.Average(pagePoints),
+		averagePagePoint: pagePoints.length ? Vec.Average(pagePoints) : new Vec(0, 0),
```
steveruizok, line 343: The Vec.Average should handle this case

ds300 (approved):

packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
MitjaBezensek: [Now it does](https://github.com/tldraw/tldraw/pull/3065/files#diff-e8cd5987588f3ef008d49f455970555cd1a231028a771f1be7251deaf3638cdcR534-R536), before it called `div` with 0.

packages/editor/src/lib/primitives/Vec.ts
```diff
@@ -531,6 +531,9 @@ export class Vec {
 	static Average(arr: VecLike[]) {
 		const len = arr.length
 		const avg = new Vec(0, 0)
+		if (len === 0) {
```
mimecuvalo, line 534: @MitjaBezensek can you add a test for this in `Vec.test.ts`? :)

cc @SomeHats because we ran into a similar issue on this with the `Uni` function if you recall

PR #3066, opened 2024-03-04
Wrap local/session storage calls in try/catch (take 2)

SomeHats:
Steve tried this in #3043, but we reverted it in #3063. Steve's version added `JSON.parse`/`JSON.stringify` to the helpers without checking for where we were already `JSON.parse`ing (or not). In some places we just store strings directly rather than wanting them jsonified, so in this version we leave the jsonification to the callers - the helpers just do the reading/writing and return the string values.

### Change Type

- [x] `patch` — Bug fix


apps/examples/src/examples/local-storage/LocalStorageExample.tsx
```diff
@@ -45,7 +38,7 @@ export default function PersistenceExample() {
 		const cleanupFn = store.listen(
 			throttle(() => {
 				const snapshot = store.getSnapshot()
-				setInLocalStorage(PERSISTENCE_KEY, JSON.stringify(snapshot))
+				localStorage.setItem(PERSISTENCE_KEY, JSON.stringify(snapshot))
```
steveruizok, line 41: Are we keeping these calls?

SomeHats: i figured we keep them in the examples app? we don't need to provide localstorage wrappers for consumers imo

steveruizok (approved):

apps/examples/src/examples/local-storage/LocalStorageExample.tsx
SomeHats: oh yea i was gonna make them all `@internal` as localstorage helpers don't seem like things we need to provide. but if u want us to have them as API surface lmk

PR #3082, opened 2024-03-06
Allow blob src urls

thatsjonsense:
Allows blob URLs for image sources. This enables a workflow where you can immediately render an asset on paste using a temporary blob URL, then replace it with an uploaded one.

data URLs are already supported, but they can be quite large. Blob URLs are very short, which allows for using tldraw in cases where the overall data size needs t obe small.

### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Enable blob URLs for image sources

### DIFF
--- a/packages/validate/src/lib/validation.ts
+++ b/packages/validate/src/lib/validation.ts
@@ -948,7 +948,7 @@ export const linkUrl = string.check((value) => {
 	}
 })
 
-const validSrcProtocols = new Set(['http:', 'https:', 'data:'])
+const validSrcProtocols = new Set(['http:', 'https:', 'data:', 'blob:'])
 
 /**
  * Validates that a valid is a url safe to load as an asset.


SomeHats:
I don't want to land this because it would allow blob URLs to get synced across multiplayer sessions which won't work for other clients.

> This enables a workflow where you can immediately render an asset on paste using a temporary blob URL, then replace it with an uploaded one.

This is actually a much harder one to get right than it looks unfortunately! The trouble all comes from the undo/redo stack: you want adding an image to be a single action regardless of how long the upload takes, but with this technique you end up in a situation where you can e.g. upload the image, delete it before the upload completes, then undo the deletion and end up with a permanently blobified asset.

This is something that's on our minds to support better, but it's really hard to get right with tldraw's current architecture. i don't think this approach is right though. sorry!

thatsjonsense:
I see, thanks for the context. In our case we're using our own multiplayer where size constraints matter a lot, so persisting a data URL is a non starter. I ended up taking a different approach of extending the image shape and adding a tempUrl that can take a blob, which works ok for us though.

SomeHats:
nice! we don't actually store data urls in our multiplayer system either, we upload them to an external store and save the URLs to them: https://github.com/tldraw/tldraw/blob/83544a9ea8c04f8f42d025089fffdbace1f519c4/apps/dotcom/src/components/MultiplayerEditor.tsx#L148-L149

the restriction with this is that we have to wait until they're uploaded before we place the image on the canvas, which isn't the nicest UX in the workd

PR #3090, opened 2024-03-09
Arrow snapping behavior redux

steveruizok:
This PR makes a few small changes to arrow binding / center snapping. It incorporates some of the logic from https://github.com/tldraw/tldraw/pull/2924

It:
- makes better use of pointer velocity when deciding whether to trigger an imprecise binding
- incorporates the approximate "pointing-at-the-centerness" of the arrow when deciding whether to trigger an imprecise binding

### Change Type

- [x] `minor` — New feature, though really this shouldn't break anything that isn't relying on quirks in our shapes / states

### Test Plan

1. Use the arrow tool.

- [x] Unit Tests

### Release Notes

- Improves the arrow shape's center-snapping logic


steveruizok:
A key change here is making the "velocity at time of first binding" more decisive in whether to start in an imprecise or precise state. It's difficult to measure the "towards the centerness" of the arrow, but speed feels intentional (even on mobile). I also feel less in control of "time spent in the same binding". If we want to capture a "pause" intent, then we should start a timer whenever the velocity hits zeroish; and then I think we can reduce the amount of time it takes to trigger a pause.

Some of the "easy" actions are:
- enter a shape
- exit a shape
- pause

The "harder" actions are:
- point exactly at the center
- point toward the center
- control the velocity over time (ie keep the velocity slow while dragging inside of a bound shape)

By combining the harder intents with the easier actions, ie the velocity (or center-pointedness) at the time of entry, we can preserve the mode after that fact, and give a quick way of fixing the problem if in the wrong mode by leaving and re-entering.

steveruizok:
Bringing this out of experiment hell, now that a few other things have landed. (And want to be sure to get it considered before we redo arrow bindings).

ds300:
Closing due to inactivity

PR #3093, opened 2024-03-10
[fix] Rotated crop handle

steveruizok:
This PR fixes a bug where local rotation was used in cropping handles rather than absolute rotation.
Kapture 2024-03-10 at 18 21 51

### Change Type

- [x] `patch` — Bug fix

### Test Plan

1. Crop an image
2. Place the image into a rotated parent
3. Crop the image
4. Rotate the image
5. Crop the image

The handles should be accurately rotated in all cases.

### Release Notes

- Fixed a bug that could cause rotated cropping images to have incorrectly rotated handles.


MitjaBezensek (approved):
Nice catch!

PR #3097, opened 2024-03-11
stickies: update appearance

mimecuvalo:
Updates the appearance of the stickies to be more skeuomorphic. It also removes the selection outline when editing for better appearance as well.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Sticky notes: update appearance.


mimecuvalo:
@steveruizok you reworked the look of the stickies in your other PR - there's one tweak in this PR that i think we should keep which is disabling the indicator when in `editing_shape`. let's discuss in writer's room?

steveruizok:
Coming back to this, just wanted to say really good solution for the slightly bent shadow. I'm suddenly nervous about perf of shadows that require extra (pseudo) elements after watching my nice 3D shadows explode when more than 100 were on the screen at the same time. But if we feel confident that the app can handle it, I'd like to see this double-element approach for the shadows in the rc!

mimecuvalo:
i'll close this up again @steveruizok !

PR #3099, opened 2024-03-11
stickies: edit on type

mimecuvalo:
Auto-edit a shape upon keystroke. Related to this change is that we shouldn't do the selection of all text when editing (and it would conflict currently with this behavior, or at least we'd have to special case). In any case, I would argue that it's odd and doesn't feel like the right move — this PR updates it so that it puts the cursor at the end of the text which to me feels better or more expected.

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Sticky note: auto-edit on keystroke.


SomeHats:
when i was testing this one out, i got tripped up when trying to re-order stickies with the [ ] keys. if we do want this, i wonder if we should limit it to only apply to alpha keys instead?

mimecuvalo:
> when i was testing this one out, i got tripped up when trying to re-order stickies with the [ ] keys. if we do want this, i wonder if we should limit it to only apply to alpha keys instead?

_maaaybe_ :) a couple things:
- alpha keys won't cover i18n keys. we'd have to write a comprehensive non-symbol regex for 'alpha' letters (or, i think it would prbly be easier to do something like `[^~!@#$%^&*()_+-={}|[]\:";'<>?,./]`)?)
- there are other shortcuts that aren't using symbols, so I could see a counter-argument that it might be sort of strange skew that some keyboard shortcuts work (i.e. `[` and `]` where as `e` for eraser doesn't)
- finally, you're more of a power user, so you know all the keyboard shortcuts :)

SomeHats:
looking at this:
Screenshot 2024-03-12 at 16 56 35

i think that actually the rule might be that everything goes into edit, unless:
- cmd is held
- or alt is held
- or key is backspace, ], or [

the other shortcuts are tool switchers rather than actions, and for some reason it feels a lot less important to me that those work. I think it's because I would never select the shape specifically to use those keyboard shortcuts, whereas with [ ] if i want to move the shape, i have select it first.

mimecuvalo:
> cmd is held, or alt is held, or key is backspace, ], or [

yeah the logic right now excludes alt/ctrl and backspace. i just didn't add [ or ] yet :)

mimecuvalo:
it's a good push to add symbols in general though @SomeHats !

mimecuvalo:
latest tweaks as discussed:
- replace all when pressing a key
- Enter, however, selects all instead of replacing
- kbd shortcuts [ and ] are now excluded

PR #3100, opened 2024-03-11
Stickies: additive cursor icon

mimecuvalo:

I could very easily see an argument that this should just be a custom icon. I'm not married to this particular icon. @steveruizok feel free to add one!

### Change Type

- [x] `patch` — Bug fix

### Release Notes

- Sticky note: update insertion mouse icon.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -64,6 +64,7 @@
 	/* These cursor values get programmatically overridden */
 	/* They're just here to help your editor autocomplete */
 	--tl-cursor: var(--tl-cursor-default);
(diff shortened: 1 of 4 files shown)


steveruizok:
We don't currently do custom cursors for shapes, though that's been discussed. I'm not sure if I'd do a one-off for the sticky note shape.

However, I'm open to having _no cursor_ if we go with a ghost image for the sticky. The note shape is the only shape where we don't allow resizing, so this kind of "preview" would be useful. Same idea as the shapes we create with duplication.

image

PR #3105, opened 2024-03-11
[dx] Allow vscode to search inside md files by default

ds300:
Before this PR all .md files were targeted by the `.ignore` file, which has bitten me on a number of occasions since .md files often contain valuable information (e.g. the vscode extensions docs). This PR unignores .md files while still ignoring _generated_ .md files like our changelogs, the api-report files, and the generated docs sections.

Additionally, the `yarn format` and `yarn lint` commands were configured slightly differently, which was confusing, so I've unified those and simplified the lint.ts script at the same time.

### Change Type

- [ ] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know


mimecuvalo:
this has also bitten me - thank you!

PR #3111, opened 2024-03-11
docs: fix missing API entries

mimecuvalo:
following up on https://discord.com/channels/859816885297741824/1162726738774720574/1211715924613275681

several things here:
- `docs/api/.*json` were out-of-date — seems like fetch-api-source should run automatically? shouldn't `build-api` also override this directory? in particular, tldraw.api.json still had a ton of references to the old @tldraw/tldraw package
- the main problem was that `generateApiContent` was failing silently. we were relying on Promises and this broke silently because we never handled exceptions. i got rid of the Promise as it was unnecessary and made the exceptions bubble up
- two things were broken in the docs and those are fixed, so now the missing entries will resurface

### Change Type


steveruizok:
> docs/api/.*json were out-of-date — seems like fetch-api-source should run automatically? shouldn't build-api also override this directory? in particular, tldraw.api.json still had a ton of references to the old @tldraw/tldraw package

We need to include docs as part of our release process. The docs are pulled from the latest release SHA. That keeps the generated API docs pinned to the last release that we've made (rather than what's not been released yet). But bumping that SHA in the environment variables and doing a new release is not automated yet.

steveruizok:
I think I see the issues. I've bumped the SHA on the docs site to the latest release and will make some updates in the code as well (to log errors but not crash when encountering issues).

steveruizok:
We can't ship the new API json until we do a release, so we have to roll back those changes.

apps/docs/scripts/functions/fetchApiSource.ts
```diff
@@ -4,7 +4,7 @@ import path from 'path'
 import { TLDRAW_PACKAGES_TO_INCLUDE_IN_DOCS } from './package-list'
 
 const octokit = new Octokit({
-	auth: process.env.GITHUB_ACCESS_TOKEN,
+	auth: process.env.ACCESS_TOKEN,
```
steveruizok, line 7: Not sure when this changed, but the environment variable is named `ACCESS_TOKEN`, not `GITHUB_ACCESS_TOKEN`.

packages/tldraw/src/lib/ui/TldrawUi.tsx
```diff
@@ -54,7 +54,7 @@ export interface TldrawUiBaseProps {
 }
 
 /**
- * Props for the {@link @tldraw/tldraw#Tldraw} and {@link TldrawUi} components.
```
steveruizok, line 57: True fix here

packages/tldraw/src/lib/TldrawImage.tsx
```diff
@@ -51,7 +51,7 @@ export type TldrawImageProps = Expand<
  *
  * @example
  * ```tsx
- * <TldrawImage snapshot={snapshot} />
```
steveruizok, line 54: True fix here

steveruizok:
release label is cursed

steveruizok (approved):

PR #3118, opened 2024-03-11
squish sync data events before sending them out

si14:
Recently (https://github.com/tldraw/tldraw/pull/3012), we started aggregating data messages before sending them out. However, local testing shows that we generate *many* redundant messages (see the test file for an example of a real buffer captured during local testing with just two users). This PR adds a function to squish those updates together, reducing the amount of data we need to transfer and load on the client that won't need to process those redundant messages.

The function is checked with [fast-check](https://fast-check.dev/), a JS property test framework, to make sure that squished deltas result in exactly the same state as the original ones.

### Change Type

- [x] `minor` — New feature

### Test Plan

1. Needs a group smoke test

- [x] End to end tests


packages/tlsync/src/lib/squish.ts
```diff
@@ -0,0 +1,174 @@
...
+		case RecordOpType.Patch: {
+			// both are patches, merge them
+			const bailed = patchThePatch(lastRecordOp[1], newPatch)
+			if (bailed) {
+				return true
+			}
+			break
+		}
+		case RecordOpType.Remove:
+			// we're trying to patch an object that was removed, just disregard the update
```
ds300, line 87: this should never happen, might be worth adding an assert

packages/tlsync/src/lib/squish.ts
```diff
@@ -0,0 +1,174 @@
...
+					if (lastOp[0] === ValueOpType.Append) {
+						const lastValues = lastOp[1]
+						const lastOffset = lastOp[2]
+						const newValues = newOp[1]
+						const newOffset = newOp[2]
+						if (newOffset === lastOffset + lastValues.length) {
+							lastValues.push(...newValues)
+						} else {
+							// something weird is going on, bail out
+							return true
```
ds300, line 40: this should also never happen, I'd prefer to see an assert.

packages/tlsync/src/lib/squish.ts
```diff
@@ -0,0 +1,174 @@
...
+						}
+					} else {
+						// bail out, it's too hard
+						return true
+					}
```
ds300, line 45: if the last op is a PUT it should be ok to just append the values, right?

And also if the last op was a DELETE it should be an error rather than a bail.

packages/tlsync/src/lib/squish.ts
```diff
@@ -0,0 +1,174 @@
...
+				}
+				break
+			case ValueOpType.Patch:
+				if (lastPatch[newKey] === undefined) {
+					lastPatch[newKey] = newOp
+				} else {
+					// bail out, recursive patching is too hard
+					return true
```
ds300, line 53: again, if the last patch was a DELETe it should be an error.

ds300 (commented):
Left a few comments. In general the code looks OK but I'd need to see some generative testing to feel good about it (i have written code like this before that looked fine but had very sneaky bugs hidden).

also i'm curious to know

1. how regularly this would be used
2. how much time it would save in busy rooms.

I assume the time saved by serializing and sending a smaller json blob would far outweigh any extra time spent doing smooshing, but would it actually be a significant win?

packages/tlsync/src/lib/squish.ts
```diff
@@ -0,0 +1,174 @@
...
+				}
+				state.squished.push(e)
+				break
+			case 'patch':
+				if (state.lastPatch !== null) {
+					const bailed = squishInto(state.lastPatch.diff, e.diff)
+					if (bailed) {
+						// this is unfortunate, but some patches were too hard to patch, give up
+						// and return the original list
+						return dataEvents
```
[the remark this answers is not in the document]
si14: problem is, each event can have more than one operation, and `lastPatch` is being mutated, so I can't be sure which part of the current event was successfully applied. I could try to slice some arrays, but I don't think it's worth it, bailing out should be fairly rare.

si14:
+1 on generative testing (although our current fuzzing does catch bugs in this, that's how I found a few).

One of the benefits here is less work for the client. All of those deltas would've been applied separately, with change detection, callbacks, and whatnot.

si14:
@ds300 added a generative test and tested the test itself by injecting a number of non-trivial bugs (which were successfully found). Also fixed a really annoying mutability issue that it uncovered, there was indeed a sneaky bug 😆 

Re: assertions: after thinking about this long and hard, I think it's something best postponed. I agree it would be good to do extra validation on this step, but as things stand now our code is super chill about nonsensical diffs (for example, `applyObjectDiff` just silently ignores weird diffs), so this PR can retain that approach. Plus it'd make the property test substantially more annoying, so if you don't mind let's land this as is

packages/tlsync/src/lib/squish.ts
si14: commented below re: assertions

packages/tlsync/src/lib/squish.ts
si14: same, commented below re: assertions

packages/tlsync/src/lib/squish.ts
si14: good idea, added a clause to handle that

see below re: assertions

packages/tlsync/src/lib/squish.ts
si14: same, see below re: assertions

ds300 (approved):
Looking good. Let's wait until we have analytics in place for perf baselines until we merge this.

packages/tlsync/src/test/squish.test.ts
```diff
@@ -0,0 +1,546 @@
...
+		}))
+		const squishedDiffs = squishDataEvents(dataEvents).map((e) => {
+			assert(e.type === 'patch')
+			return e.diff
+		})
+
+		const baseStore = this.getStoreWithDiffs(this.diffs)
+		const squishedStore = this.getStoreWithDiffs(squishedDiffs)
+
+		expect(squishedStore.serialize()).toStrictEqual(baseStore.serialize())
```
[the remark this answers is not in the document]
si14: fixed

packages/tlsync/src/test/squish.test.ts
```diff
@@ -0,0 +1,546 @@
...
+	fc.nat(10).map((idx) => new RecordRemove(idx)),
+	fc.tuple(fc.nat(), ObjectDiffArb).map(([idx, diff]) => new RecordPatch(idx, diff)),
+]
+
+const initialStoreContentArb: Arbitrary<TestRecord[]> = fc.uniqueArray(TestRecordArb, {
+	selector: (r) => r.id,
+	maxLength: 3,
+})
+
+test('fast-checking squish', () => {
```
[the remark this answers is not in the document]
si14: done

packages/tlsync/src/test/squish.test.ts
```diff
@@ -0,0 +1,546 @@
...
+						model: new Model(initialStoreContent),
+						real: 'whatever',
+					}),
+					cmds
+				)
+			}
+		),
+		{
+			verbose: 1,
+			numRuns: 10_000,
```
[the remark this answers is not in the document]
si14: done

steveruizok:
@ds300 I thought this one landed, did we still want this?

PR #3127, opened 2024-03-12
Fix the issue with shape position getting out of sync.

MitjaBezensek:
Fixes an issue where the perf updates PR caused the shape position to be out of sync. We now also throttle things in the `useReactor` and we switch to using that one instead of `useQuickReactor` in a couple of places.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Create some shapes.
2. Resize them
3. It should not be janky. The effect is even more pronounced in multiplayer.

- [ ] Unit Tests
- [ ] End to end tests


MitjaBezensek:
Closing in favor of #3160

PR #3129, opened 2024-03-12
Revert throttling of useValue and useStateTracking.

MitjaBezensek:
Reverts the changes to the `state` package that were made in #2977. Should fix the issue with shape jittering discovered during QA.

### Change Type

- [x] `patch` — Bug fix
- [ ] `minor` — New feature
- [ ] I don't know

### Test Plan

1. Create some shapes (draw shapes work well).
2. Open the same room in a second browser.
3. Resize shapes (using option / alt makes it more obvious).
4. The shapes should not jitter in any of the screens.

### DIFF
--- a/packages/state/package.json
+++ b/packages/state/package.json
@@ -52,9 +52,6 @@
 			"node_modules/(?!(nanoid)/)"
 		]
 	},
-	"dependencies": {
-		"@tldraw/utils": "workspace:*"
-	},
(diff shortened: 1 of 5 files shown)


ds300:
> Looks good here! Is this removing the gains from the performance work though?

partially yes. We still have the perf gains for the tool interactions.

PR #3132, opened 2024-03-12
stickies: edit on type, part 2: place cursor at intended position in general

mimecuvalo:
This is what we discussed this morning! Places the cursor at the intended position. This is doing some magic with opacity to still capture the cursor position and we show it when we're actually editing. 

The discussion offline was that select all, being the current behavior implies that editing = 'replace all text' as the default behavior, requiring a third-click to deselect the entirely selected text. The previous 'edit on type' PR that this is stacked on top of start shifting that paradigm to at least choose append to text instead of replace - but! We can still do better. This PR goes the next logical step and places the cursor where the user would probably expect it to go.

(This obviously also ties in with the textfields PR work in general as well.)

https://github.com/tldraw/tldraw/assets/469604/7f747296-0801-4c01-baa8-25e33b06882c

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/tldraw/src/lib/shapes/arrow/components/ArrowTextLabel.tsx
```diff
@@ -54,34 +61,34 @@ export const ArrowTextLabel = React.memo(function ArrowTextLabel({
...
-						onContextMenu={stopEventPropagation}
-						onPointerDown={handleInputPointerDown}
-						onDoubleClick={handleDoubleClick}
-					/>
-				)}
+				<textarea
+					ref={rInput}
+					// We need to add the initial value as the key here because we need this component to
+					// 'reset' when this state changes and grab the latest defaultValue.
+					key={initialText}
```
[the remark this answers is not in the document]
SomeHats: (ditto for other instances of this pattern)

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
```diff
@@ -20,52 +20,22 @@ export function useEditableText<T extends Extract<TLShape, { props: { text: stri
...
-			if (shape) {
-				elm.value = shape.props.text
-				if (elm.value.length && !skipSelect) {
-					elm.select()
-				}
-			}
-		})
-	}, [editor, id, doesShapeAutoEditOnKeystroke])
+		/* do nothing */
+	}, [])
```
SomeHats, line 38: can this be deleted entirely? if it's a required prop somewhere, you could pass in `noop`

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -266,6 +267,16 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
...
 
+	/**
+	 * Retrieves the shape's label, if it has one.
+	 *
+	 * @param shape - The shape.
+	 * @public
+	 */
+	getLabel(shape: Shape): Rectangle2d | null {
+		return null
+	}
```
SomeHats, line 278: i'm not sure about adding this to the core API - imo this is quite an our-shapes-specific concept. Other shapes might have multiple labels, for example. 

Couple options here:
1. mark this as `@internal` for now so it's not a public API (although this means others can't get at this behaviour too)
2. `Geometry2d` already has an `isLabel` property. you can use this to e.g. have a group with a body shape, and a rectangle shape marked as a label. instead of `getLabel`, we could have a generic helper that iterates through a geometry and returns any parts marked with `isLabel` (i prefer this option)
3. it's probably fine we could just ship this anyway

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingArrowLabel.ts
```diff
@@ -100,14 +102,24 @@ export class PointingArrowLabel extends StateNode {
 			nextLabelPosition = 0.5
 		}
 
+		this.didDrag = true
```
SomeHats, line 105: 1. you might be able to use `editor.inputs.isDragging` here
2. I'm not sure i understand how this works. we check `didDrag` in `onPointerUp`, but we set it in `onExit`. I would expect `onPointerUp` to be called before `onExit` though, because after on exit we're no longer in this state node and it shouldn't be receiving events. any ideas what's going on here?

SomeHats (commented):
would this be easier if we already had #3050? if yes, we should maybe work on just getting that landed as it doesn't really touch user facing stuff

also, i would loooove to see some tests for this stuff

mimecuvalo:
> would this be easier if we already had https://github.com/tldraw/tldraw/pull/3050? if yes, we should maybe work on just getting that landed as it doesn't really touch user facing stuff

haha, yes, i had just said the same thing on that PR as the latest comment :) https://github.com/tldraw/tldraw/pull/3050

> also, i would loooove to see some tests for this stuff

definitely, the base PR that this is stacked upon has tests added. i didn't add new tests yet to this PR. in general, because there are so many sticky PR's and I don't know which ones we actually want to go forward with, i was waiting for the go ahead and then start adding tests to PRs we want to take past the 'prototype' phase in the Linear sticky notes project.

packages/tldraw/src/lib/shapes/arrow/components/ArrowTextLabel.tsx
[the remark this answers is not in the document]
mimecuvalo: that's not a bad idea/point on maybe just using the textarea in general — i like it, it's bold!
we can perhaps maybe lean into the `readonly` property of textareas, if need be.

i'll give it a go 👍

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
mimecuvalo: sorry, this hasn't been synced in a while with its base PR (it's a stacked PR) — this section hasn't been deleted.

i'll actually unstack this, because it's confusing now.

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
mimecuvalo: ah, i hadn't seen that Geometry2d had an `isLabel` property - that would be useful, and I agree about a generic helper to iterate through. i'll give it a go 👍 

i do think though that, inherently shapes probably are paired with text, i.e. the label concept. i think how Geometry2d is aware of labels is maybe indicative that it's just a useful pattern to have in our library. i guess i don't know if we want to be _too_ abstract about this.

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
mimecuvalo: actually, i retract that comment - i _did_ need this to be removed for all this to work, i started losing track of all of this 🙃 

`noop` sounds good!

packages/tldraw/src/lib/shapes/arrow/components/ArrowTextLabel.tsx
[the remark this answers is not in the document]
mimecuvalo: I tried just using the textarea and it seems untenable at the moment to remove the elements — it'd require a bigger rethink of things if we went that route.

actually the issue you raised @SomeHats seems to already been in production? it might be a slight variation but it's the same issue at the moment really. so i don't _think_ it's quite a regression for the multiplayer quirk.

PR #3147, opened 2024-03-13
Add image annotator example

SomeHats:
This diff mostly adds an image annotator example, but also has a couple of drive-by changes:
- Added a 'use-cases' category to the examples app for this style of mini-app
- Add `editor.pageToViewport`, which is like `editor.pageToScreen` but works with viewport coordinates (better for `InFrontOfTheCanvas` stuff)
- Prevent the chrome side-swipe-to-go-back thing in the examples app

Some cool features of the image annotator:
- The image cannot be unlocked, and cannot have shapes places behind it
  - I still need to work out a way of removing the context menu though
- Anything you place outside the bounds of the image (and therefore outside the bounds of the export) will be greyed out
- You can't change pages
  - unless you find the "move to page" action... need to fix that
- The camera is constrained! It'll keep the image roughly centered on the screen. If you pick a very long thin image, you can only scroll vertically. If you pick a very big one, it'll default it to a reasonable size.

### Change Type

- [x] `feature` — New feature


mimecuvalo:
(lemme know if you need an extra review here to move things along as an example, happy to help)

SomeHats:
gonna just land this for now as it's only an example, we can iterate on it more later

PR #3162, opened 2024-03-15
stickies: attribution experiment

mimecuvalo:
If we were to finish this experiment out, it would need:
- a migration
- probably when a user changes the name, it needs to actually modify the relevant shapes rather than just the ephemeral change that is being done right now
- the text shadow needs work!

https://github.com/tldraw/tldraw/assets/469604/b17ae1de-0949-43eb-bae9-7d96f24136f7

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/tlschema/api-report.md
+++ b/packages/tlschema/api-report.md
@@ -707,6 +707,8 @@ export const noteShapeProps: {
     growY: T.Validator<number>;
     url: T.Validator<string>;
     text: T.Validator<string>;
+    userName: T.Validator<string | undefined>;
+    userId: T.Validator<string | undefined>;
 };
 
 // @internal (undocumented)
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -81,7 +81,7 @@
 	--tl-font-sans: 'tldraw_sans', sans-serif;
 	--tl-font-serif: 'tldraw_serif', serif;
 	--tl-font-mono: 'tldraw_mono', monospace;
-	/* fext outline */
+	/* text outline */
 	--a: calc(min(0.5, 1 / var(--tl-zoom)) * 2px);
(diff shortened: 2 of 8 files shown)


packages/tlschema/api-report.md
```diff
@@ -707,6 +707,8 @@ export const noteShapeProps: {
     growY: T.Validator<number>;
     url: T.Validator<string>;
     text: T.Validator<string>;
+    userName: T.Validator<string | undefined>;
+    userId: T.Validator<string | undefined>;
```
steveruizok, line 711: Ideally we'd wait for this (and all other attribution) features until we have user accounts, both as a reason to create an account and because I don't think we can fully deliver the feature without those. But it's worth thinking about how "information about users" can become part of the document, i.e. for offline / downloads.

Perhaps the document can keep track of "known users", or users who have interacted with / created content in the document. That would give us a way to preserve the relationship between userIds and userNames. ie when a user joins or updates their name / color, we also update the document's list of "known users" and preserve that data there. So rather than keeping the name on the shape here, we just use the userId, and if that id is known to the document, then we pull the name from there. I don't know how often userIds are changed (i.e. if I reload my page, do I have a new user id? @ds300) so this may also make the feature not great. We would also want to add this "known users" to copy and paste data, similar to how we do assets, so that it can be brought across between documents.

ds300: > (i.e. if I reload my page, do I have a new user id? @ds300)

no the user ids are stable per browser context, unless you click 'reset data' at some point.

mimecuvalo:
Where did we land on this one @steveruizok ? nix for now I assume? or we shall we do a "good enough" version?

PR #3180, opened 2024-03-17
[experiment] Fix camera / process events on tick

steveruizok:
This PR implements a queue-and-flush system for dispatched events, giving us a better "update on tick" implementation.

Kapture 2024-03-17 at 16 47 26

## How it works

image

- When events are sent to the editor (via `editor.dispatch`) they are placed in a queue
- On each tick, the queue is emptied and processed within a single transaction
  - Each event is processed, updating the inputs, running any other immediately-processed events, etc.
  - If not otherwise prevented, each event is sent to the state chart and handled there
  - Finally, a `tick` event is processed through the state chart 

This allows us to handle all events on tick without having to design our handlers around this fact. Each frame's events are handled in a single transaction, including events that occur as the result of the initiating event (i.e. the pointer move events that occur while wheel-panning).

Pros:
- we have REALLY good control over how often we update
- we fix some camera-related cases where a frame would have multiple transactions

Cons:
- there are places where we do more per frame
- tests need to trigger frames more often
- there's some extra complexity around events that trigger camera changes

## Future improvements

An even-better implementation would be to collect input events between ticks and then have a state node (or whatever) process these all at the same time. For example if a resizing state received two pointer moves that occurred in the same tick, it could ignore the first move and only process the second.

This would need thinking through. If one of the events caused the the handler to need to transition / hand off the events to a different state, then it would need to also provide the remaining / unprocessed events to that new state.

```ts
class ExampleState {
	onTick = (eventsThisTick: TLEventInfo[]) => {
		for (let i = 0; i < eventsThisTick.length; i++) {
			const event = eventsThisTick[i]
			switch (event.type) {
				case 'pointer': {
					switch (event.name) {
						case 'pointer_move': {
							// We can handle events if we need to
							this.doSomethingOnPointerMove()
							break
						}
						case 'pointer_up': {
							// If we transition to a new state, we'll need that state's
							// on tick to be called with the remaining events
							this.parent.transition('idle', eventsThisTick.slice(i))
							break
						}
					}
					break
				}
			}
		}
	}
}
```

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Select a shape, begin dragging, trackpad pan to move the camera

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Improve performance using the editor's tick.

### DIFF
--- a/apps/examples/e2e/tests/test-canvas-events.spec.ts
+++ b/apps/examples/e2e/tests/test-canvas-events.spec.ts
@@ -1,6 +1,6 @@
 import test, { expect, Page } from '@playwright/test'
 import { Editor } from 'tldraw'
-import { setupPage } from '../shared-e2e'
+import { setupPage, sleep } from '../shared-e2e'
 
 declare const __tldraw_editor_events: any[]
 
@@ -21,6 +21,7 @@ test.describe('Canvas events', () => {
 			await page.mouse.move(200, 200) // to kill any double clicks
 			await page.mouse.move(100, 100)
 			await page.mouse.down()
+			await sleep(20)
 			expect(await page.evaluate(() => __tldraw_editor_events.at(-1))).toMatchObject({
 				target: 'canvas',
 				type: 'pointer',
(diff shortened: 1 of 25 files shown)


steveruizok:
FWIW I can't tell whether the performance here is better than in production. See https://github.com/tldraw/tldraw/pull/3181 for a quicker fix.

PR #3181, opened 2024-03-17
[fix] Batch tick events

steveruizok:
This PR fixes an issue where events happening on tick were not batched. 

Kapture 2024-03-17 at 22 49 52

We were listening to the `tick` event directly from the state node, rather than passing the event into the state chart at the top. This meant that it was bypassing the regular state chart rules, which was what got me looking at this; but then I noticed that we also weren't batching the changes, either. This causes computed stuff to re-compute after each atom is updated within the `onTick` handler, which can be a LOT.

Before:
image

After:
image

It's not game breaking but it's important enough to hotfix at least in the dot com.

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Select many shapes.
2. Resize them.

### Release Notes

- Fix a performance issue effecting resizing multiple shapes.

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -1782,7 +1782,7 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
     // (undocumented)
     onRightClick?: TLEventHandlers['onRightClick'];
     // (undocumented)
-    onTick?: TLTickEventHandler;
+    onTick?: TLEventHandlers['onTick'];
     // (undocumented)
     onTripleClick?: TLEventHandlers['onTripleClick'];
     // (undocumented)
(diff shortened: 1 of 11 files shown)


packages/editor/api-report.md
steveruizok, line 1: We make the `tick` a regular event info object so that we can pass it to the Editor.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8380,6 +8381,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 		])
 	}
 
+	/** @internal */
+	private tick = (elapsed = 0) => {
+		this.dispatch({ type: 'misc', name: 'tick', elapsed })
+		this.scribbles.tick(elapsed)
+	}
```
steveruizok, line 8388: We also update the scribbles after the tick event gets handled.

packages/editor/src/lib/editor/managers/ScribbleManager.ts
steveruizok, line 1: We can clean up some state here in the scribble manager.

MitjaBezensek (approved):
This helps a lot when interacting with multiple shapes 🤯 👏

ds300 (approved):
Good catch!

PR #3182, opened 2024-03-17
Don't double squash

steveruizok:
This PR changes the way `Store.squashHistoryEntries` works. Previously, the function would iterate through every entry and squash it against the current entry (using `squashRecordDiffs`) to get the new current entry. However, `squashRecordDiffs` does basically the same pattern, iterating through the properties of every diff. As a result, each diff would be iterated through twice: once as itself, and once again in the next current.

This PR tweaks the function to operate on as many diffs as possible at once.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Minor improvement when modifying multiple shapes at once.

### DIFF
--- a/packages/store/src/lib/Store.ts
+++ b/packages/store/src/lib/Store.ts
@@ -936,9 +936,9 @@ function squashHistoryEntries<T extends UnknownRecord>(
 	chunked.push(chunk)
 
 	return devFreeze(
-		chunked.map((entries) => ({
-			source: entries[0].source,
-			changes: squashRecordDiffs(entries.map((e) => e.changes)),
+		chunked.map((chunk) => ({
+			source: chunk[0].source,
+			changes: squashRecordDiffs(chunk.map((e) => e.changes)),
 		}))
 	)
 }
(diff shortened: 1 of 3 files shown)


packages/store/src/lib/Store.ts
```diff
@@ -916,28 +916,32 @@ export function squashRecordDiffs<T extends UnknownRecord>(
...
-	result.push(current)
+	toProcess.push(currentChunk)
 
-	return devFreeze(result)
+	return devFreeze(
+		toProcess.map((entries) => ({
+			source: entries[0].source,
+			changes: squashRecordDiffs(entries.map((e) => e.changes)),
+		}))
+	)
```
steveruizok, line 944: toProcess stores an array of all of the entries, chunked by their source

packages/store/src/lib/Store.ts
```diff
@@ -916,28 +916,32 @@ export function squashRecordDiffs<T extends UnknownRecord>(
...
 
+	let entry: HistoryEntry<T>
 	for (let i = 1, n = entries.length; i < n; i++) {
 		entry = entries[i]
 
 		if (current.source !== entry.source) {
-			result.push(current)
+			// result.push(current)
+			toProcess.push(currentChunk)
 			current = entry
```
ds300, line 930: i think you can delete this `current` let binding

ds300 (approved):
noice

PR #3192, opened 2024-03-18
Readonly / room creation omnibus

MitjaBezensek:
Reworks how the readonly urls work. Till now we just used a simple function that would scramble the slugs. Now we use a proper key value mapping between regular and readonly slugs:

- We use two KV stores. One is for going from a slug to a readonly slug and the other one for going the other way around. They are populated at the same time.
- We separate preview KV stores (dev, preview, staging) from production one. I've already created these on Cloudflare. [My understanding is ](https://developers.cloudflare.com/kv/reference/data-security/#encryption-at-rest)that ids [can be public](https://community.cloudflare.com/t/is-it-safe-to-keep-kv-ids-in-a-public-git-repo/517387/4) since we can only access KV from our worker. Happy to move them to env variables though.

- [x] Disable creating new rooms when tldraw is embedded inside iframes on other websites (we check the referrer and if it's not the same as the iframe's origin we don't allow it)
- [x] Fork a project when inside an iframe now opens the forked project on tldraw.com and not inside iframe.
- [x] We allow embeding of iframes, but we now track the where they are used via the referrer. We send this to Vercel analytics.
- [x] Improved UX of the share menu to make it less confusing. Toggle is gone.
- [x]  `/new` and `/r` routes not redirect to `/`.
- [x] This introduces a new `/ro` route for readonly rooms. Legacy rooms still live on `/v`.
- [x] Brought back `dotcom-shared` project to share code between BE and FE. Mostly types. 
- [x] Prevent creating of rooms by entering `/r/non-existing-slug`. 
- [x] Handle getting a readonly slug for old rooms. Added a comment about it [here](https://github.com/tldraw/tldraw/pull/3192/files#diff-c0954b3dc71bb7097c39656441175f3238ed60cf5cee64077c06e21da82182cbR17-R18).
- [x] We no longer expose editor on the window object for readonly rooms. Prevents the users disabling readonly rooms manually.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan
1. Make sure old readonly rooms still work.
2. Creating a readonly link from an existing room should still use `/v` path.
3. Newly created rooms should use `/ro` path for readonly rooms. Make sure these work as well.
4. `/r` room was disabled and redirects to `/`
5. `/new` should still work when not inside iframes.

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

1. This adds new functionality for readonly rooms:
- We have a new route `/ro` for newly created readonly rooms. These rooms no longer use the scrambling logic to create readonly slugs. Instead we now use KV storage from cloudflare to track the mapping for slugs -> readonly slug and readonly slug -> slug.
- The old route `/v` is preserved, so that the old room still work as they did before.
- For old rooms we will keep on generating the old readonly slugs, but for new rooms we'll start using the new logic.
2. We no longer prevent embedding of tldraw inside iframes. 
3. We do prevent generating new rooms from inside the iframes  though. `/r`, `/new`, `/r/non-existing-id` should not allow creation of new rooms inside iframes. Only `/new` still works when not inside iframes.
4. Forking a project from inside an iframe now opens it on tldraw.com
5. Slight copy change on the sharing menu. We no longer have a toggle between readonly and non-readonly links.
6. `editor` and `app` are no longer exposed on the window object for readonly rooms. Prevents users from using the `updateInstanceState` to escape readonly rooms.


apps/dotcom/src/__snapshots__/routes.test.tsx.snap
```diff
@@ -26,6 +22,10 @@ exports[`the_routes 1`] = `
     "reactRouterPattern": "/r/:roomId",
     "vercelRouterPattern": "^/r/[^/]*/?$",
   },
+  {
+    "reactRouterPattern": "/ro/:roomId",
```
MitjaBezensek, line 26: We probably need to add a test for this, similarly to how we removed it for `/new`

apps/dotcom-worker/src/lib/utils/roomOpenMode.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function getSlug(env: Environment, slug: string | null, roomOpenMode: RoomOpenMode) {
+	if (!slug) return null
+	switch (roomOpenMode) {
+		case ROOM_OPEN_MODE.READ_WRITE:
+			return slug
+		case ROOM_OPEN_MODE.READ_ONLY:
+			return await env.READONLY_SLUG_TO_SLUG.get(slug)
+		case ROOM_OPEN_MODE.READ_ONLY_LEGACY:
+			return lns(slug)
```
SomeHats, line 13: what are the drawbacks of doing something along the lines of `( || lns(slug)` and keeping the URL as `/v/`?

Off the top of my head, the main issue is maybe that that would mean lns URLs would work for new rooms (although they wouldn't be shared anywhere), but we could prevent that by doing something like:
```ts
const storedSlug = await env.READONLY_SLUG_TO_SLUG.get(slug)
if (storedSlug) return storedSlug

const legacySlug = lns(slug)
const isLegacyRoom = !(await env.SLUG_TO_READONLY_SLUG.get(legacySlug))
if (isLegacyRoom) return legacySlug
return null
```

apps/dotcom/src/components/ErrorPage/ErrorPage.tsx
```diff
@@ -19,9 +20,7 @@ export function ErrorPage({
 					<p>{messages.para1}</p>
 					{messages.para2 && <p>{messages.para2}</p>}
 				</div>
-				<Link to={'/'}>
-					<a>Take me home.</a>
-				</Link>
+				<Link to={'/'}>{isInIframe() ? 'Open tldraw.' : 'Back to tldraw.'}</Link>
```
SomeHats, line 23: this probably wants `target="_blank"` when `isInIframe` is true

apps/dotcom/src/components/IFrameProtector.tsx
```diff
@@ -76,24 +89,28 @@ export function IFrameProtector({
...
 			}
 		}
 
 		window.addEventListener('message', handleMessageEvent, false)
 
-		if (embeddedState === 'iframe-unknown') {
+		if (embeddedState === EMBEDDED_STATE.IFRAME_UNKNOWN) {
 			// We iframe embeddings on multiplayer or readonly
 			if (WHITELIST_CONTEXT.includes(context)) {
 				window.parent.postMessage(EXPECTED_QUESTION, '*') // todo: send to a specific origin?
```
SomeHats, line 102: feel free to call it out of scope of this diff, but how necessary is all this message-passing stuff at this stage? I feel like we could probably get by with just checking against the origin - i experimented with this here: https://github.com/tldraw/tldraw/pull/3214/files#diff-10fd47d203ff90d59bdb8ff3675076dfbb01942674b58450f529d2754b592554L58-L103

apps/dotcom/src/routes.tsx
```diff
@@ -30,28 +29,17 @@ export const router = createRoutesFromElements(
...
-					return redirect(`/r/${id}`)
-				}}
-			/>
-			<Route
-				path="/new"
-				loader={() => {
-					const id = 'v2' + nanoid()
-					return redirect(`/r/${id}`)
-				}}
-			/>
```
SomeHats, line 46: what's the thinking behind removing these vs keeping them and having them 404 when in an iframe? i think these are super useful for keeping e.g. a bookmark for "new tldraw room"

apps/dotcom/src/utils/iFrame.ts
```diff
@@ -0,0 +1,20 @@
+export const isInIframe = () => {
+	return typeof window !== 'undefined' && (window !== window.top || window.self !== window.parent)
+}
+
+export function getTopLevelOrigin() {
```
SomeHats, line 5: i think we should call this `getParentOrigin` and use `ancestorOrigins[0]` instead because:
- that is a bit more consistent with what `document.referrer` might do
- the parent is more truly the embedder that the top level document, I would guess? e.g. if i embed tldraw in notion then embed notion in my website, notion is still the main embedder i think?

what do you reckon?

SomeHats (commented):
looking good! few comments to think about

apps/dotcom/src/utils/iFrame.ts
MitjaBezensek: Definitely 👍

apps/dotcom/src/components/IFrameProtector.tsx
MitjaBezensek: @mimecuvalo we discussed allowing embedding of iFrames - just not creating new rooms from within them. Feels like we could completely remove these check? We'd still keep the `WHITELIST_CONTEXT` since we want to prevent things like local rooms from working in iframes?

apps/dotcom/src/routes.tsx
MitjaBezensek: I'd also love to keep this if we can.

It is another vector for how people that embed tldraw could generate new rooms. Spin up a browser, go to `/new` just to generate a new room, then use that new room in an iframe. That said, it's probably just as easy as using the api to get a new slug, so the benefit is quite small.

We did remove the creation of new rooms for slugs that don't exist yet, so we'd have to rework this to go through the `/new-room` api, but it's definitely doable.

apps/dotcom-worker/src/lib/utils/roomOpenMode.ts
MitjaBezensek: Now that we locked down room creation the only thing it gives us is a bit more traceability if something goes wrong. We could maybe reuse `/v` and loose quite a bit of complexity. WDYT @mimecuvalo?

apps/dotcom/src/components/ErrorPage/ErrorPage.tsx
mimecuvalo: nice catch - thx!

apps/dotcom-worker/src/lib/utils/roomOpenMode.ts
mimecuvalo: Yeah, I think a couple things:
- it opens up the path of eventually deprecating the older path, if we, say after a year (prbly longer), see that there's no traffic there. 
- As Mitja mentions, there's more traceability to narrowing down an issue with older paths vs. newer ones. If they start getting conflated it'll start being harder to debug/maintain.
- it felt icky in the sense of a threat of a collision being present. It's not possible actually to have a collision between these two, it's just that it felt like, eng-wise, if were migrating to, say, similar-but-different hashing functions we would have to account for this and either create a versioning scheme or different path. It's only a quirk that the lns logic is silly enough where it isn't possible. So I suppose I'm saying it feels more correct to separate these two.

I hear the argument for keeping parity, but the prefix of the url's aren't 'pretty' or memorable anyway ¯\\_(ツ)_/¯ 
Only `r` is arguably memorable, but the rest are perhaps incidental so it didn't feel too important to keep it the same.

apps/dotcom/src/components/IFrameProtector.tsx
mimecuvalo: I would agree that it's overkill. I think it's out-of-the-scope of this particular PR to refactor it though. (the iframe change here was just to add tracking, and I incidentally just made some constants).

Let's take a look in a separate PR to rework it!

apps/dotcom/src/routes.tsx
mimecuvalo: @steveruizok said in standup that he'll weigh on this to be the decider on this.

I am, personally, a little torn — yes, having a bookmark for this type of thing is definitely cool (we created the domain https://paper.new just to do this ;) 
However, I think the inherent difference with other apps that have this is that we still don't have auth :-/

So, in my mind, this is a temporary removal and once we have an auth story we can reintroduce this.

apps/dotcom/src/utils/iFrame.ts
```diff
@@ -2,19 +2,15 @@ export const isInIframe = () => {
...
 
-export function getTopLevelOrigin() {
-	let origin: string
+export function getParentOrigin() {
 	if (isInIframe()) {
 		const ancestorOrigins = window.location.ancestorOrigins
 		// ancestorOrigins is not supported in Firefox
 		if (ancestorOrigins && ancestorOrigins.length > 0) {
-			origin = ancestorOrigins[ancestorOrigins.length - 1]
+			return ancestorOrigins[0]
```
mimecuvalo, line 10: @MitjaBezensek why parent origin vs. top? i could see it being either way honestly - just curious what made you rethink it

MitjaBezensek: It was the [comment from Alex](https://github.com/tldraw/tldraw/pull/3192#discussion_r1576503766):

> the parent is more truly the embedder that the top level document, I would guess? e.g. if i embed tldraw in notion then embed notion in my website, notion is still the main embedder i think?

We are probably more interested in who does the initial embedding right? Makes it more actionable for us, we can reach out directly to them.

apps/dotcom-worker/src/lib/utils/roomOpenMode.ts
SomeHats: I feel like some of this could be mitigated with a migration to write the old lns IDs into the KV store.

if we do want to have two separate URLs, could we make the new one still be a single-character?

apps/dotcom/src/routes.tsx
SomeHats: I don't really see how auth comes into this but maybe I'm missing something. Let's see what steve says! Worth calling out that we can keep these but prevent them in an iframe though.

PR #3202, opened 2024-03-18
Stickies duplication- find an empty spot

Taha-Hassan-Git:
This PR adds to the duplication handle behaviour, now when creating a note shape in a particular direction, it will first check for an empty spot and prioritise a spot that is closest to that side without overlapping other shapes.

2024-03-18 at 16 52 50 - Blush Stork

- super inefficient at the moment. It regenerates all potential positions every time the handle is clicked. Would be better to either cache these or only generate as far as needed each time
- Hit-testing needs to be better, it shouldn't just check a single point, it should check the box that the new note shape will occupy
- Frames are kind of crazy, hit testing doesn't ignore frames!
- generated arrows should be sent to the back, so they don't overlap closer stickies, for some reason that's not working
- Handles should show while in editing state imo, and have their own styling to distinguish from other kinds of handles
- Would be nice to hint what the handles do on hover

Things to try next:

- making notes in a grid shape
- duplicating from a single handle

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -1216,8 +1216,16 @@ input,
 }
 
 .tl-note__duplicate-button{
+	border: none;
+	font-weight: 600;
+	background-color: var(--color-muted-2);
 	position: absolute;
 	pointer-events: all;
+	border-radius: 8px;
+}
+
+.tl-note__duplicate-button:hover{
+	background-color: var(--color-muted-1);
 }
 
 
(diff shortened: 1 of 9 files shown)


mimecuvalo:
so fun to play with!

PR #3208, opened 2024-03-19
Stickies: 3d stickies

TodePond:
This PR plays around with making it feel like you're 'picking up' and 'plonking down' stickies.
There are feature flags for the different aspects of this. Try toggling different ones on and off! 

The code for this PR is a complete mess - not representative of the code we'd write if we ship it.

2024-03-19 at 11 57 39 - Amaranth Reindeer
2024-03-20 at 16 38 12 - Rose Aardwolf

2024-03-20 at 17 06 45 - Indigo Shrimp

Closes TLD-2315

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [x] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Test Plan

1. Try placing some stickies and move them around.

- [ ] Unit Tests
- [ ] End to end tests

PR #3216, opened 2024-03-19
Stickies: 3d stickies combined with skeuomorphic

TodePond:
This combines #3097 and #3208 

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


steveruizok:
After spending some time with this, my thought is now that it's probably not the right UX. We don't have any other shapes that look different while translating. It makes accurate placement much harder, even with special rules around snapping. I could see this type of UX working if we were using "drop targets" rather than free placement on the canvas, because then the final position wasn't in the user's hands anyway; but since we're not doing this, we need to give users the ability to accurately and predictably place shapes.

TodePond:
Closing this because it's out of date from #3208

> we need to give users the ability to accurately and predictably place shapes.

agree with this, and I don't feel strongly about landing this 'lifting' change. I had a thought about how to have our cake and eat it too, might do a tiny mockup - but I'm prioritising the changes we're more sure on, like stickies as a surface

mimecuvalo:
Would the lifting change be helped if only the bottom part of the sticky lifted? I'm curious if this might feel better if the top of the shape was not transformed? 🤔

PR #3220, opened 2024-03-19
New migrations again

ds300:

### Change Type

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

#### BREAKING CHANGES

- The `Migrations` type is now called `LegacyMigrations`.
- The serialized schema format (e.g. returned by `StoreSchema.serialize()` and `Store.getSnapshot()`) has changed. You don't need to do anything about it unless you were reading data directly from the schema for some reason. In which case it'd be best to avoid that in the future! We have no plans to change the schema format again (this time was traumatic enough) but you never know.
- `compareRecordVersions` and the `RecordVersion` type have both disappeared. There is no replacement. These were public by mistake anyway, so hopefully nobody had been using it.
- `compareSchemas` is gone. Comparing the schemas directly is no longer really possible since we introduced some fuzziness. The best thing to do now to check compatibility is to call `schema.getMigraitonsSince(prevSchema)` and it will return an error if the schemas are not compatible, an empty array if there are no migrations to apply since the prev schema, and a nonempty array otherwise.

   Generally speaking, the best way to check schema compatibility now is to call `store.schema.getMigrationsSince(persistedSchema)`. This will throw an error if there is no upgrade path from the `persistedSchema` to the current version.

- `defineMigrations` has been deprecated and will be removed in a future release. For upgrade instructions see https://tldraw.dev/docs/persistence#Updating-legacy-shape-migrations-defineMigrations

- `migrate` has been removed. Nobody should have been using this but if you were you'll need to find an alternative. For migrating tldraw data, you should stick to using `schema.migrateStoreSnapshot` and, if you are building a nuanced sync engine that supports some amount of backwards compatibility, also feel free to use `schema.migratePersistedRecord`.
- the `Migration` type has changed. If you need the old one for some reason it has been renamed to `LegacyMigration`. It will be removed in a future release.
- the `Migrations` type has been renamed to `LegacyMigrations` and will be removed in a future release.
- the `SerializedSchema` type has been augmented. If you need the old version specifically you can use `SerializedSchemaV1`


mimecuvalo:
🙌 let us know @ds300 if you want some initial feedback at this stage, or if you're still hammering it out

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -74,6 +74,11 @@ export interface TldrawEditorBaseProps {
 	 */
 	shapeUtils?: readonly TLAnyShapeUtilConstructor[]
 
+	/**
+	 * An array of migrations to apply to the editor's schema.
+	 */
+	migrations?: readonly MigrationSequence[]
```
SomeHats, line 80: `TldrawEditorProps` has some separate props for whether we have a store or not. Could this go in the type-enforced "no store allowed" section?

packages/editor/src/lib/config/TLUserPreferences.ts
```diff
@@ -55,66 +54,28 @@ const Versions = {
...
-			down: ({ isWrapMode: _, ...user }: TLUserPreferences) => {
-				return user
-			},
-		},
-	},
-})
+const CURRENT_VERSION = Math.max(...Object.values(Versions))
+
+function migrateSnapshot(data: { version: number; user: any }) {
+	if (data.version < Versions.AddAnimationSpeed) {
```
SomeHats, line 60: classique 🏛️

packages/editor/src/lib/utils/sync/TLLocalSyncClient.test.ts
```diff
@@ -69,43 +69,7 @@ test('the client connects on instantiation, announcing its schema', async () =>
 	expect(channel.postMessage).toHaveBeenCalledTimes(1)
 	const [msg] = channel.postMessage.mock.calls[0]
 
-	expect(msg).toMatchObject({ type: 'announce', schema: { recordVersions: {} } })
-})
-
-test('when a client receives an announce with a newer schema version it reloads itself', async () => {
```
SomeHats, line 75: afaict this behaviour is still there - there's nothing removing it in `TLLocalSyncClient` at least. Why are we removing these?

packages/store/src/lib/StoreSchema.ts
```diff
@@ -62,16 +108,30 @@ export class StoreSchema<R extends UnknownRecord, P = unknown> {
...
+		this.sortedMigrations = sortMigrations(allMigrations)
 
-	// eslint-disable-next-line no-restricted-syntax
-	get currentStoreVersion(): number {
-		return this.options.snapshotMigrations?.currentVersion ?? 0
+		for (const migration of this.sortedMigrations) {
+			if (!migration.dependsOn?.length) continue
+			for (const dep of migration.dependsOn) {
+				const depMigration = allMigrations.find((m) => m.id === dep)
+				assert(depMigration, `Migration '${migration.id}' depends on missing migration '${dep}'`)
```
SomeHats, line 132: I later ran into an issue here with bindings. If I have a store migration that depends on a particular arrow migration but I'm using the editor without the arrows shape it blows up 😬

not sure what the right move here is. i think ideally i'd write the store migration that splits out bindings within the sequence for arrows themselves? not sure that's an option though...

packages/store/src/lib/test/dependsOn.test.ts
```diff
@@ -0,0 +1,82 @@
+import { StoreSchema } from '../StoreSchema'
+import { MigrationSequence } from '../migrate'
+
+describe('dependsOn', () => {
```
SomeHats, line 4: nit: could add a test for standalone `dependsOn` here

packages/store/src/lib/StoreSchema.ts
```diff
@@ -34,10 +43,47 @@ export interface SerializedSchema {
...
 }
 
+/** @public */
+export interface SerializedSchemaV2 {
+	schemaVersion: 2
+	sequences: {
+		[sequenceId: string]: {
+			version: number
+			retroactive: boolean
+		}
```
SomeHats, line 53: Why do we store `retroactive` here? afaict, we only ever read this property off of the migrations sequence in the code, not this stored one

packages/tlschema/src/records/TLShape.ts
```diff
@@ -200,16 +188,136 @@ export function getShapePropKeysByStyle(props: Record<string, T.Validatable<any>
...
+// We only really need them to smooth over the transition between versions, and some folks do keep
+// browser tabs open for months without refreshing, but at a certain point that kind of behavior is
+// on them. Plus anyway recently chrome has started to actually kill tabs that are open for too long rather
+// than just suspending them, so if other browsers follow suit maybe it's less of a concern.
+export const RETIRED_DOWN_MIGRATION = 'retired' as const
+
+/**
+ * @public
+ */
+export type TLShapePropsMigrations = {
```
SomeHats, line 202: in my record-ful bindings spike, i ended up pulling a lot of this stuff out into a file for generic "record with props" migrations.

having sat with it a while though, i actually think the thing i would maybe do here is maybe remove these shape-specific migration helpers & just have everything be record migration sequences... or maybe even normal sequences with record-level migrations being the default, and an optional default `recordFilter` or something so you don't have to filter everything.

one thing that would give us is a way to make the store migration to split bindings out of arrows as part of the sequence for arrows themselves, which avoids some of the awkward stuff around depending on a sequence that doesn't exist.

It's maybe pushing a little additional complexity onto users for that "writing migrations for a custom shape" use case, but also potentially simplifies the story a fair bit and unlocks some useful options. What do you think?

packages/tlschema/src/createTLSchema.ts
```diff
@@ -66,8 +76,11 @@ const defaultShapes: { [T in TLDefaultShape['type']]: SchemaShapeInfo } = {
  * @public */
 export function createTLSchema({
 	shapes = defaultShapes,
+	migrations,
+	// TODO: allow passing in custom migration sequences
```
SomeHats, line 80: has this not been addressed? it looked addressed

packages/store/src/lib/StoreSchema.ts
ds300: it is used in `compareSchemas` but I'm now considering removing that because it's not really reliable any more (see pr description for details). so I might be able to just have the version number here.

packages/store/src/lib/StoreSchema.ts
ds300: it's an option. I actually had an earlier commit where I made it so the `shapeUtil.migrations` property could be a standard `MigrationSequence` as an alternative to the special props migration sequence. I dropped it to avoid polluting the API but sounds like we already have a good use case for it.

packages/editor/src/lib/utils/sync/TLLocalSyncClient.test.ts
ds300: i brought these back

packages/store/src/lib/StoreSchema.ts
ds300: removed

packages/tlschema/src/records/TLShape.ts
ds300: I've made it so that folks can use a normal migration sequence on the shapeUtil.migrations property but I'm not sure I like the idea of removing the props helpers stuff because it will indeed add a lot of complexity for the most common use case: folks who just wanna migrate their shape props will have to know about filters and namespaces and scopes, and doing everything inside the .props object manually.

I'm going to go with the current thing for now. There's still time to rethink before our next sdk release.

steveruizok (approved):

PR #3204, opened 2024-03-18
Stickies: sticky notes as a surface

TodePond:
This PR allows the user to draw on sticky notes, and place things onto them.

As an experimental prototype, it lets you place *anything* on sticky notes, but we could filter down the list of allowed shapes if we wanted.

This PR breaks frames behaviour slightly in order to get the prototype to work. If we decide to ship this, we could fix frames. (However, hot take: Part of me wants to change the behaviour of frames to what this PR does anyway. That's another thing though).

More work to do:
- Place shapes on a sticky even if your cursor doesn't overlap the sticky.
- As a sticky grows longer/shorter, move its children with it.
- Some visual feedback that something has attached? 

https://github.com/tldraw/tldraw/assets/15892272/f5e84fe2-09bc-4cbe-8443-31d9a4df5b52

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [x] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Test Plan

1. Create a sticky note.
2. Try to draw on it.
3. Move the sticky. The drawing should move with it.
4. Drag another shape on top of the sticky note shape.
5. Move the sticky. It should move the dragged shape with it.
6. Remove the attached shape(s) from the sticky.
7. Move the sticky. Make sure the removed shape doesn't move with it.
8. With the sticky note selected, drag a shape that's on top of it.
9. The dragged shape should move - not the sticky.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Shapes can now be attached to sticky notes.

PR #3225, opened 2024-03-20
Stickies: bring to front

steveruizok:
This PR makes pointing a sticky note bring it to the front. Extracted from https://github.com/tldraw/tldraw/pull/3208/files.

We could consider changing the logic to "move to front of any overlapping shapes" rather than move to front of all shapes in the page / parent.

Kapture 2024-03-20 at 12 41 08

If we haven't landed #3204, then this would mean clicking a note would bring it in front of anything that has been placed on top of the note. Even if we did land that PR, this would only improve things going forward: any existing notes would not have the parent / child relationship with shapes that had been previously placed on top of it.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Place a sticky slightly behind another shape
2. Click on the sticky behind
3. It should be in front

- [ ] Unit Tests

### Release Notes

- When clicking a sticky, bring it to front

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
@@ -5,6 +5,7 @@ import {
 	TLArrowShape,
 	TLEventHandlers,
 	TLGeoShape,
+	TLNoteShape,
 	TLPointerEventInfo,
 	TLShape,
 } from '@tldraw/editor'
@@ -27,6 +28,13 @@ export class PointingShape extends StateNode {
 
 		this.hitShape = info.shape
 		const outermostSelectingShape = this.editor.getOutermostSelectableShape(info.shape)
+
+		// Bring sticky notes to front on pointer down;
+		// consider changing the logic to "move to front of any overlapping shapes"
+		// rather than move to front of all shapes in the page / parent
+		if (this.editor.isShapeOfType<TLNoteShape>(info.shape, 'note')) {
+			this.editor.bringToFront([info.shape.id])
+		}
 
 		if (
 			// If the shape has an onClick handler


mimecuvalo:
so good

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
```diff
@@ -28,6 +29,13 @@ export class PointingShape extends StateNode {
 		this.hitShape = info.shape
 		const outermostSelectingShape = this.editor.getOutermostSelectableShape(info.shape)
 
+		// Bring sticky notes to front on pointer down;
+		// consider changing the logic to "move to front of any overlapping shapes"
+		// rather than move to front of all shapes in the page / parent
+		if (this.editor.isShapeOfType<TLNoteShape>(info.shape, 'note')) {
```
mimecuvalo, line 35: let's make this an API property, not specific to `note`, something like `shouldBringToFrontOnSelect`

mimecuvalo (approved):
we should add this to the RC @steveruizok !

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
steveruizok: I'd prefer we keep these hardcoded for now!

steveruizok:
As discussed, we'll need to land the "stickies as a surface" first if this is going to land, or else things in front of the sticky would be dropped behind it. There's still a pretty good change that this happens (as not everything is going to be stuck to the sticky) but I think it's okay to assume that stuff overlapping stickies is rare enough.

PR #3226, opened 2024-03-20
Move camera APIs from `editor` to `editor.camera`

SomeHats:
This diff moves all camera controls stuff into a new `CameraManager` class. The old methods are still there, I've just deprecated them. Tagged as galaxy brain for "big refactor", but not actually a backwards compatibility breaking change due to just deprecating the old method.

As part of this, I've added some very basic support for deprecated items to our API reference docs generation.

### Change Type


ds300 (approved):
I'm into this. Namespaces are good sometimes and this feels like one of those cases. 👍🏼

steveruizok:
This doesn't seem necessary to me. The design for tldraw is intentional in the editor is a "big flat fat god object", rather than using namespaces for different functionality. (We have one exception, the `user`, which is a sneaky bastard that be eventually refactored out). Even if we did have a camera manager, I would prefer that methods like `zoomToBounds` be on the surface of the editor.

steveruizok (changes requested):
We can extract the improvements to the API markdown / deprecation. Let's talk today about the namespace change.

mimecuvalo:
Namespace change → exposure discussion?
Would be good to chat on it and get on the same page, there's arguments for both sides. There's prbly a lot we can borrow from other engines (i.e. game engines) and how they do/don't namespace things.

steveruizok:
I'm closing this one. There was a miscommunication on my part around extending the camera API. I don't see any strong reason at the moment to change API design direction, so let's put this one to rest.

PR #3142, opened 2024-03-13
stickies: try out auto-sizing text instead of auto-sizing the shape

mimecuvalo:
I don't think we should go this route after playing with it for a while, but maybe combined with @Taha-Hassan-Git 's exploration on "see more" this could play more nicely when it gets too small. Btw, @Taha-Hassan-Git if you wanna base your work off of this PR, it might be interesting, but it's up to you!

https://github.com/tldraw/tldraw/assets/469604/ca01a20e-31a8-46fa-84a5-6641d911e6bc

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [ ] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Test Plan

1. Add a step-by-step description of how to test your PR here.
2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Stickies: auto-size font on stickies when reaching the limits of the height.

PR #3242, opened 2024-03-22
stickies: resize font-size _only_ if we hit long words

mimecuvalo:
The gif shows it all. This is a further iteration on https://github.com/tldraw/tldraw/pull/3142/files
This doesn't mean that we necessarily want to keep the `growY` property. It just shows that we can sometimes rely on the CSS `overflow-wrap` property and sometimes we can try to actively avoid where it makes sense.

I put a lower size limit of 12 but I think I'm open to having a higher one, maybe 14? Hard to say.

- [x] SVG export
- [x] Check previous text wrapping PRs for a text torture file
- [x] Tests

https://github.com/tldraw/tldraw/assets/469604/352b2721-b2e7-40ec-bb19-46e0bd607e1c

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/apps/examples/e2e/tests/test-text.spec.ts
+++ b/apps/examples/e2e/tests/test-text.spec.ts
@@ -1,5 +1,5 @@
 import test, { Page, expect } from '@playwright/test'
-import { BoxModel, Editor } from 'tldraw'
+import { BoxModel, Editor, TLNoteShape, TLShapeId } from 'tldraw'
 import { setupPage } from '../shared-e2e'
 
 export function sleep(ms: number) {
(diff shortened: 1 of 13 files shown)


steveruizok:
Ah this is super slick. I really like this.

## GrowY

Kapture 2024-03-22 at 20 12 33

I noticed a bug with growY—not sure what's going on, but it doesn't happen on main, with the text "amsamster dam is a place where we like to be; it's a perfect city and I lov".

## Migrations

I'm not sure how it would handle migrations. There may be stickies which both 1) contain wrapping words and 2) have a growY, i.e are tall stickies. In this PR, it may be that the wrapping adjusts the height of the height of the sticky to reduce the growY; but in a migration, I'm not sure if we would be able to re-calculate the growY, so we'd end up with sticky notes that retain their growY despite having smaller text.

image

In the above example, the first sticky note (on main) has a word that breaks mid-word. When I paste the shape into this PR, it creates the second note shape, which shrinks the text to avoid the break and calculates out a zero growY. However, if we were to just migrate the shape, we'd end up with the third note shape, which would preserve the growY. If a user edited it, then the growY would disappear.

This is probably ok though because most people don't have sticky notes with words that break more than once, so the "jump" would be maybe one or two lines, max.

image

steveruizok:
For a separate PR, but I think this will combine really well with some of the other stuff we'd talked about with growY / multi-sticky towers. Either way this is a great balance of preserving text without turning shrinky text into a runaway feature.

mimecuvalo:
I'm taking a look at the growY bug - good catch!

Re: migrations. Ya know, I was wondering if maybe they shouldn't be migrated? Because the bounding size of the note might change, it might affect placement of things that people intended. I think I would be ok if the new behavior only kicked in once typing into an existing shape. 

So, to be clear, the size of the text _doesn't_ currently get changed unless you explicitly edit the older notes. (nor does growY)

steveruizok (approved):

PR #3243, opened 2024-03-22
[experiment] improve note+highlighter stickiness behavior

mimecuvalo:
This updates the highlighter + note drag&drop behavior. 
- adds an `isSticky` property to ShapeUtil's to demarcate tools as having sticky (parenting) behavior
- improves the drop (parenting) behavior to not rely on the pointer position but the bounding box collision behavior (for non-frames)
- also improves the unparenting behavior to not rely on the pointer as well being within the bounding box

https://github.com/tldraw/tldraw/assets/469604/e066bce3-85dd-40de-bbf2-7b1aab538afb

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -1599,7 +1599,6 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
     canBind: <K>(_shape: Shape, _otherShape?: K | undefined) => boolean;
     canCrop: TLShapeUtilFlag<Shape>;
     canDropShapes(shape: Shape, shapes: TLShape[]): boolean;
-    canDropShapesOnlyWithinMaskedBounds: TLShapeUtilFlag<Shape>;
     canEdit: TLShapeUtilFlag<Shape>;
     canEditInReadOnly: TLShapeUtilFlag<Shape>;
     canReceiveNewChildrenOfType(shape: Shape, type: TLShape['type']): boolean;
@@ -1624,6 +1623,7 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
     hideSelectionBoundsFg: TLShapeUtilFlag<Shape>;
     abstract indicator(shape: Shape): any;
     isAspectRatioLocked: TLShapeUtilFlag<Shape>;
+    isFrame: TLShapeUtilFlag<Shape>;
     isSticky: TLShapeUtilFlag<Shape>;
     // (undocumented)
     static migrations?: Migrations;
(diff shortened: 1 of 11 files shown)


steveruizok:
Good to have it as a point of comparison, but I think it's better to have the content decide whether it can parent a shape rather than the shape decide.

We may be here over-indexing here on the adhesive side of the sticky note. Despite the name, I don't think sticky notes should be "stuck to" diagrammatic shapes. 

My reasoning would be that diagrammatic shapes (drawings, geo shapes, text) usually compose up to a thing, like a flow chat, and so sticking a note onto an individual member of that flow chart feels inappropriate. We do have the ability to group shapes into a higher-order thing, but even those often get composed up... Better IMO to use the regular selection for that.

On the other hand, sticky notes (and embeds, and _maybe_ images) feel like things that _don't_ compose up to other things. If I add a sticky note to the page, it's unlikely or unexpected for them to dissolve into a larger diagram or structure, so they feel thingy enough to accept other content as children. 

It feels strange here that I can highlight onto a note shape but not draw onto it. I'm not sure how that feeling aligns with the above. I think it's again that I'm thinking of the thing I'm drawing onto rather than the tool I'm using.

I'm very interested in seeing what we can do about highlighted text... but I haven't seen any solution yet that really addresses it in a good way. The way we have it now is also sub-par, but as discussed I'd rather leave these kinds of less-good solutions up to the user as a folk practice rather than "own" them in our behaviors.

TodePond:
One thing I do like from this PR is the fix for this behaviour:

### Before
https://github.com/tldraw/tldraw/assets/15892272/857ea178-0704-449d-bcc4-29513fe9bd9f

### After
https://github.com/tldraw/tldraw/assets/15892272/8aef1d68-8f7f-4389-97c7-fed4845420f5

mimecuvalo:
Fair points @steveruizok! I agree that sticky notes stuck to other objects are probably cosplaying as comments (which maybe speakis to the fact that we would maybe need a commenting system, not a sticky note system that does commenting badly)

I do wonder if you think in a world where we have stickers one day, would that line of thinking change? I _do_ think highlighters are different, semantic-wise vs. drawing. Semantically, they act as annotators instead of a regular drawing shape - it's the same nuance that Figjam brings to highlighters. So I don't think it's strange that drawing and highlighting have different behavior - it's maybe something we can talk about more.

steveruizok:
I'd like to close this one for now—it's not out of the question that highlighers should stick to shapes, but at the moment the regular selection ux is more reliable than what I'm seeing here. In order to make this great, we would first need a **constraints API** that describes a child's position relative to its parent, so that we can do the type of normalization position / stretching that would make sense. We've never really gone down that rabbithole before but its very common to have in a canvas app so this could be a place to implement it.

mimecuvalo:
Sounds good, sounds good 👍

PR #3251, opened 2024-03-25
Stickies: expanded snap bounds

steveruizok:
This PR expands the bounds snap points for the note shape.

Kapture 2024-03-25 at 08 21 05

First thoughts are that this doesn't feel like it should be using the snap feature, it should sort of "just snap" when its center is near to where the center would be if the note was adjacent / padded out from another sticky note.

### Change Type

- [x] `feature` — New feature

### Test Plan

1. Drag a sticky note and snap it to other shapes / notes.

### Release Notes

- Expands the snapping bounds of sticky notes for easier grids.


steveruizok:
Closing this one.

PR #3256, opened 2024-03-25
Stickies: drop zone

Taha-Hassan-Git:
Drop zone experiment. I've continued to use shapes in place of UI for hinted zones, because it's something I can work with quickly. as well as the behaviour described in Steve's diagram below I've also added hinted zones when dragging a single note, as suggested by Mime.

Known issues:

- Should hit-test the entire box to check if it's empty before generating hinted shape, not just a single point
- Maximum update depth exceeded error on grid duplication. I think this comes from trying to create and delete shapes every frame. Doing this via UI would solve that.
- "Cannot read properties of undefined (reading 'bounds')" on grid duplication. Not sure what this is coming from, but again wouldn't be an issue with the actual implementation

Next steps: 
- experiment with handles and arrows
- think about interactions on mobile devices

image

2024-03-26 at 9 31 16 - Violet Antelope

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

### DIFF
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -848,7 +848,7 @@ export function GroupOrUngroupMenuItem(): JSX_2.Element;
 // @public (undocumented)
 export class HandTool extends StateNode {
     // (undocumented)
-    static children: () => (typeof Dragging | typeof Idle_8 | typeof Pointing_7)[];
+    static children: () => (typeof Dragging_2 | typeof Idle_8 | typeof Pointing_7)[];
     // (undocumented)
     static id: string;
     // (undocumented)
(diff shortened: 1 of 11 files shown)


Taha-Hassan-Git:
> I love the look of it!
> 
> It does feel like I want the drop zones to show up when it's being dragged around? I feel like, personally, I write write write, and then I organize things. (vs. the other way around) But I could see people maybe getting confused when it shows up only in one scenario.

That's a good point, I'll add it to the list!

mimecuvalo:
omg that latest version 🤩

steveruizok:
This should not be settled by the sticky pits / #3315 PR. If we get the "new shape preview" while the note shape is active, then the "drop zone" will be adjusting the preview to the pit location. We'll likely skip the grid-creation UX, though I'm glad it was explored here.

steveruizok:
Closing! We went with pits.

PR #3257, opened 2024-03-25
Make shapes leave frames when dragged completely outside

TodePond:
This PR adds an extra way for shapes to leave their frame. If a shape is dragged completely outside of a frame (so that it doesn't overlap it anymore), then it leaves the frame.

I added this so that you don't accidentally get a shape stuck 'behind the canvas' so easily. **This behaviour will also make 'stickies as a surface' work better. I think that it's a good change for frames, but if we don't want to do that, we can keep frames the way they are, and bring this behaviour only to stickies.**

2024-03-26 at 14 37 04 - Teal Buzzard

It also deals with multiple frames.

2024-03-26 at 14 45 43 - Violet Peafowl

Shapes only leave their frames if *all* dragged shapes are fully outside.

2024-03-26 at 14 48 28 - Olive Firefly

It works with nested frames too.

2024-03-26 at 14 52 23 - Amaranth Sailfish

It works with groups too

2024-03-26 at 14 50 19 - Black Egret

And it keeps grouped children together.

2024-03-26 at 16 20 34 - Black Butterfly

This decision tree shows the behaviour before VS after:

image

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Make a frame.
2. Put a shape in the frame.
3. Drag the shape so it gets partially masked by the frame.
4. Click the masked part of the shape, and drag it completely outside of the frame, so that the shape doesn't overlap with it anymore.
5. The shape should leave the frame.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixed/improved how shapes can escape frames.

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
@@ -46,22 +46,21 @@ export class DragAndDropManager {
 
 		// is the next dropping shape id same as the last one?
 		if (nextDroppingShapeId === this.prevDroppingShapeId) {
-			// Check if all shapes are outside their parents
-			const areAllShapesOutsideTheirParents = movingShapes.every((shape) => {
-				const parent = this.editor.getShape(shape.parentId)
-				if (!parent) return false
-				const parentBounds = this.editor.getShapePageBounds(parent)
-				const shapeBounds = this.editor.getShapePageBounds(shape)
-				if (!parentBounds || !shapeBounds) return false
-				return !parentBounds.includes(shapeBounds)
-			})
-
-			// If some shapes are within their parents still, don't do anything
-			if (!areAllShapesOutsideTheirParents) {
+			// if some shapes are still within their parent, do nothing
+			if (
+				movingShapes.some((shape) => {
+					const parent = this.editor.getShape(shape.parentId)
+					if (!parent) return true
+					const parentBounds = this.editor.getShapePageBounds(parent)
+					const shapeBounds = this.editor.getShapePageBounds(shape)
+					if (!parentBounds || !shapeBounds) return true
+					return parentBounds.includes(shapeBounds)
+				})
+			) {
 				return
 			}
 
-			// If all shapes are outside their parents, we'll reparent them to the next dropping shape
+			// if all shapes are outside their parents, reparent each one
 			for (const shape of movingShapes) {
 				const parent = this.editor.getShape(shape.parentId)
 				if (!parent) continue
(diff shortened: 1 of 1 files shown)


TodePond:
@mimecuvalo pointed out this issue with the PR 🙏

2024-03-25 at 15 22 22 - Teal Flea

TodePond:
Figma does this slightly weirdly too!
https://twitter.com/steveruizok/status/1772292841051468033

TodePond:
@steveruizok and @mimecuvalo I've reworked this, and it's ready to review. Shapes now leave frames when you *drag them completely outside* (even if your cursor doesn't overlap the frame edge). All other behaviour is unchanged. You can still 'adjust' a shape's position by clicking and dragging outside the frame, without causing it to leaving the frame.

packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
```diff
@@ -44,8 +44,40 @@ export class DragAndDropManager {
...
+				if (prevDroppingShape) {
+					this.editor
+						.getShapeUtil(prevDroppingShape)
+						.onDragShapesOut?.(prevDroppingShape, movingShapes)
+				}
+
+				for (const shape of movingShapes) {
+					this.editor.reparentShapes(
+						[shape.id],
+						nextDroppingShapeId ?? this.editor.getCurrentPageId()
```
TodePond, line 72: hmm, I guess this shouldn't be hardcoded here really. it should probably call `onDragShapesOut` instead.

packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
```diff
@@ -44,9 +44,30 @@ export class DragAndDropManager {
...
+				})
+			) {
+				return
+			}
+
+			// if all shapes are outside their parents, reparent each one
+			for (const shape of movingShapes) {
+				const parent = this.editor.getShape(shape.parentId)
+				if (!parent) continue
+				this.editor.getShapeUtil(parent).onDragShapesOut?.(parent, [shape])
```
TodePond, line 67: this shouldn't repeatedly call for each child

steveruizok:
Weirdly it feels like if I drag it out of the frame in this way, then dragging it back over the frame should reparent it.
Kapture 2024-03-26 at 18 36 46

mimecuvalo (commented):
love how deep you went on this @TodePond !

TodePond:
> Weirdly it feels like if I drag it out of the frame in this way, then dragging it back over the frame should reparent it.

@steveruizok I see what you mean. You can get a similar effect on tldraw.com right now, where returning the cursor to its original position results in a different parenting - so I don't think it's the end of the world that this happens

2024-03-27 at 9 50 39 - Plum Wildfowl

I wonder if it would feel better if we 'remembered' the initial parenting, and then let the user return to that in both these cases. I'm thinking about stickies with this as well - I'll set up a PR to try it with stickies

TodePond:
will be continued in #3297

PR #3258, opened 2024-03-25
side effects reference docs & examples

SomeHats:
Adds reference docs, guide in the "Editor" article, and examples for the side effects manager. 

There are 4 new examples:
1. Before create/update shape - constrains shapes to be places within a circle
2. Before delete shape - prevent red shapes from being deleted
3. After create/update shape - make sure there's only ever one red shape on the page at a time
4. After delete shape - delete frames after their last child is deleted

As these examples all require fairly specific configurations of shapes (or are hard to understand without some visual hinting in the case of placing shapes within a circle), I've included a `createDemoShapes` function in each of these which makes sure the examples start with shapes that will quickly show you the side effects in action. I've kept these separate from the main code (in a function at the bottom), so hopefully that won't be a source of confusion to anyone working from these examples.

### Change Type


steveruizok:
Could we add docs in the main docs repo as well? Similar I think to the Events section. At the moment the IA there is pretty flat. We could chuck it into the Editor article for now and then (as with other parts of that article) split it into a separate article once we have a better idea of what needs splitting.

SomeHats:
> Could we add docs in the main docs repo as well

added

steveruizok (approved):

PR #3263, opened 2024-03-26
fix document name overflow

SomeHats:
Fix the document name getting truncated as I forgot to update a measurement in the CSS. Also fixes an issue where if you had a long title which you then cleared, the input width wouldn't update until you entered your first character of the new name.

### Change Type
- [x] `bugfix` — Bug fix


MitjaBezensek:
👏

PR #3272, opened 2024-03-26
Improve performance of culling

steveruizok:
This PR tweaks the logic of _when_ we update the viewport screen bounds. Previously, we updated every one second in order to capture any changes to the viewport's screen position. In this PR, we _check_ every one second and update the screen bounds if the viewport's screen position has actually changed. Since we also update the rendering shapes when this happens, it would cause the rendering / culling shapes to update while the camera was moving.

I've also removed the "maximum time before we start culling shapes", as this wasn't very useful and could also cause frames to start dropping without recovering.

https://github.com/tldraw/tldraw/assets/23072548/9f474481-30c9-49b4-a009-66775ca6a0c1

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Zoom and pan around
2. Culled shapes should only update when you stop moving the camera.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Improve performance of the canvas when many shapes are present.

### DIFF
--- a/packages/editor/src/lib/constants.ts
+++ b/packages/editor/src/lib/constants.ts
@@ -87,9 +87,6 @@ export const INTERNAL_POINTER_IDS = {
 /** @internal */
 export const CAMERA_MOVING_TIMEOUT = 64
 
-/** @internal */
-export const CAMERA_MAX_RENDERING_INTERVAL = 620
-
 /** @public */
 export const HIT_TEST_MARGIN = 8
 
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -66,7 +66,6 @@ import { TLUser, createTLUser } from '../config/createTLUser'
 import { checkShapesAndAddCore } from '../config/defaultShapes'
 import {
 	ANIMATION_MEDIUM_MS,
-	CAMERA_MAX_RENDERING_INTERVAL,
 	CAMERA_MOVING_TIMEOUT,
 	CAMERA_SLIDE_FRICTION,
 	COARSE_DRAG_DISTANCE,
@@ -3085,10 +3084,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 			this._lastUpdateRenderingBoundsTimestamp = now // don't render right away
 			this._cameraState.set('moving')
 			this.on('tick', this._decayCameraStateTimeout)
-		} else {
-			if (now - this._lastUpdateRenderingBoundsTimestamp > CAMERA_MAX_RENDERING_INTERVAL) {
-				this.updateRenderingBounds()
-			}
 		}
 	}
 
--- a/packages/editor/src/lib/hooks/useScreenBounds.ts
+++ b/packages/editor/src/lib/hooks/useScreenBounds.ts
@@ -7,20 +7,24 @@ export function useScreenBounds(ref: React.RefObject<HTMLElement>) {
 	const editor = useEditor()
 
 	useLayoutEffect(() => {
+		let prevBounds = new Box()
+
 		function updateScreenBounds() {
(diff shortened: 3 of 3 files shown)


MitjaBezensek:
This definitely helps, but I'm still not convinced we should use this. There are some interactions where it definitely feels much slower.

Try out the following:
- create a bunch of shapes (think I have 2000 in the examples below)
- zoom in
- immediately zoom back out (I'm using cmd + mouse wheel)
- zoom in
- zoom back out
- ....

This should be smooth, continuous. But here is what it looks like on this branch:

https://github.com/tldraw/tldraw/assets/2523721/bf0bdcc1-b80e-4823-a567-4578cb557735

Here it is with culling removed ([PR here](https://github.com/tldraw/tldraw/pull/3276) so you can try it on your own):

https://github.com/tldraw/tldraw/assets/2523721/bdfa367c-a7fb-4912-a393-81e762914fe0


Here's panning aggressively with the hand tool. I'm trying to cull / uncull as many shapes as possible in each pan to make the effect more obvious.

https://github.com/tldraw/tldraw/assets/2523721/d47c5299-0d28-429e-97c7-4d1c7461bf16

Without culling

https://github.com/tldraw/tldraw/assets/2523721/dd74e5a7-f20a-4606-af3f-b460042ae0f9

Horizontal scrolling (shift + mouse wheel)

https://github.com/tldraw/tldraw/assets/2523721/96194b53-03d8-4cf9-b578-406a5341a336

Without culling

https://github.com/tldraw/tldraw/assets/2523721/0b99997f-5874-4ad3-a66a-57b87e9ba72d

steveruizok:
I'd be happy to turn off culling and see if it makes a difference.

There are two places where I've seen culling matter:
1. Performance when on an "empty" part a busy canvas
2. The amount of a frame used by hit testing

To measure the first:

1. Create many shapes
2. Pan to an empty part of the canvas
3. Use the draw tool to write "hey" or something

When shapes aren't culled, the frame rate while drawing is lower.
When shapes are culled, the frame rate while drawing is higher.

To measure the second:

1. Create many shapes
2. Pan to an empty part of the canvas
3. Make a few shapes.
4. Move the mouse to hover different shapes.
 
The hovering may feel choppy. Look at the length of the `hit test` event in performance tab.

steveruizok:
(Even if we don't use culling, IMO we should still land the changes in this PR as each update means busting our cache for rendering shapes and a few other things).

MitjaBezensek:
Yeah, can definitely merge this as it is an improvement 👍 

To me it kinda boils down to the question: is the number of culled shapes changing? 

When the answer is yes we are adding additional work to the browser: we have to calculate which shapes to cull and which not to cull, we also render different dom elements when going from one to the other. And that seems to be adding more work than we remove with culling.

When the answer is no (like with first paint and drawing in an empty part of the canvas) then it's a straight up win. We have fewer dom elements, which improves perf.

TBH it's quite hard to test and it's also somewhat down to preference / types of actions users perform. With culling it works well if you just pan once. With multiple pans in a row the experience gets worse: shapes popping in, fps drops, the experience is janky. This last one is the one that bother me the most: everything seems to work, then bam we need to render the shapes and fps drops by quite a bit. We don't get consistent rendering.

Might be worth exploring the following: 1s after the camera stops moving cull the shapes, as soon as we move again we stop culling. Or something along those lines.

MitjaBezensek (approved):

steveruizok:
> Might be worth exploring the following: 1s after the camera stops moving cull the shapes, as soon as we move again we stop culling. Or something along those lines.

The hard part is that if we scroll to formerly culled shapes, then we have to show those as soon as possible without impacting the interaction (ie chopping up their scroll).

PR #3274, opened 2024-03-26
log message size in worker analytics

SomeHats:
Adds logging of message size in worker analytics.

This also adds the environment to worker analytics as `blob2`. We need this because previously, all the analytics from all environments were going to the same place with no ability to tell them apart, which means we can't easily compare analytics on e.g. a particular PR.

This means that all the other blobs get shifted along one, so we won't be able to query across the boundary of when this gets released for those properties. I think this is fine though - it's things like `roomId` that I don't think we were querying on anyway.

You can query the analytics through grafana - [docs here](https://www.notion.so/tldraw/How-to-11fce2ed0be5480bb8e711c7ff1a0488?pvs=4#a66fae7bfcfe4ffe9d5348504598c6a0)

### Change Type


ds300 (approved):
looks great

PR #3275, opened 2024-03-26
Add long press event

steveruizok:
This PR adds a "long press" event that fires when pointing for more than 500ms. This event is used in the same way that dragging is used (e.g. to transition to from pointing_selection to translating) but only on desktop. On mobile, long presses are used to open the context menu.

Kapture 2024-03-26 at 18 57 15

## Background

This idea came out of @TodePond's #3208 PR. We use a "dead zone" to avoid accidentally moving / rotating things when clicking on them, which is especially common on mobile if a dead zone feature isn't implemented. However, this makes it difficult to make "fine adjustments" because you need to drag out of the dead zone (to start translating) and then drag back to where you want to go.

Kapture 2024-03-26 at 19 00 38

With this change, you can long press on desktop to get to that translating state. It's a micro UX optimization but especially nice if apps want to display different UI for "dragging" shapes before the user leaves the dead zone.

Kapture 2024-03-26 at 19 02 59

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Long press shapes, selections, resize handles, rotate handles, crop handles.
2. You should enter the corresponding states, just as you would have with a drag.

- [ ] Unit Tests TODO

### Release Notes

- Add support for long pressing on desktop.


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8345,6 +8346,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	/** @internal */
 	private _selectedShapeIdsAtPointerDown: TLShapeId[] = []
 
+	/** @internal */
+	private _longPressTimeout = -1 as any
```
mimecuvalo, line 8350: you can type this as `NodeJS.Timeout | null`
`private _longPressTimeout: NodeJS.Timeout | null = null`

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8531,6 +8536,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 								(this.getInstanceState().isCoarsePointer ? COARSE_DRAG_DISTANCE : DRAG_DISTANCE) /
 									this.getZoomLevel()
 						) {
+							clearTimeout(this._longPressTimeout)
```
mimecuvalo, line 8539: (as an aside: i still am thinking about how we prbly want a timeout tracker - otherwise you'll sometimes have some uncollected timeout's if we dispose the editor)

packages/editor/src/lib/editor/types/event-types.ts
```diff
@@ -16,6 +16,7 @@ export type TLPointerEventTarget =
 export type TLPointerEventName =
 	| 'pointer_down'
 	| 'pointer_move'
+	| 'long_press'
```
mimecuvalo, line 19: naming question: because now we'd have:
- pointer
- click
- press

`long_press` does sound the most intuitive and i honestly don't have a better alternative ¯\\_(ツ)_/¯ 
`pointer_long_down`?? meehhh

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingHandle.ts
```diff
@@ -37,10 +37,19 @@ export class PointingHandle extends StateNode {
...
-			this.parent.transition('dragging_handle', this.info)
+			this.kickoff()
 		}
 	}
 
+	override onLongPress: TLEventHandlers['onLongPress'] = () => {
+		this.kickoff()
+	}
+
+	private kickoff() {
```
mimecuvalo, line 48: let's rename `kickoff` - it's hard to know what it's kicking off :P 
for example, we might be kicking off a drag or a translate or different things depending on the current inputs in this state tool.

(here and elsewhere in this PR)

packages/editor/src/lib/editor/Editor.ts
steveruizok: I've seen this different ways (i.e. ReturnType<typeof setTimeout>) as it's a little different in different contexts. Maybe we should have a utility type for this?

packages/editor/src/lib/constants.ts
```diff
@@ -107,3 +107,6 @@ export const COARSE_HANDLE_RADIUS = 20
 
 /** @internal */
 export const HANDLE_RADIUS = 12
+
+/** @internal */
+export const LONG_PRESS_DURATION = 500
```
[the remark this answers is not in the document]
steveruizok: This long press feature isn't enabled on mobile. We have actual long presses on mobile that would conflict, so until we have our own gesture recognizer, we're just turning it off for mobile.

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingHandle.ts
steveruizok: Renamed to `startTranslating`, `startRotating`, etc.

packages/editor/src/lib/editor/types/event-types.ts
steveruizok: ehh long_press is pretty standard

packages/editor/src/lib/editor/Editor.ts
steveruizok: That's a good idea!

PR #3281, opened 2024-03-27
capture latency drift between client and worker

SomeHats:
We don't have a good way to measure the speed of our multiplayer, but we might be able to measure when the client gets locked up with messages - it'll cause larger drift between how far apart the server sent two messages, and how far apart the client thinks it received them. This diff captures that drift on the client, then sends a histogram of the drift back to the server every ping when there've been more than 100 messages received in the mean time. 

Not clear on whether this will be useful or not - spikey network latency etc. is probably a pretty big factor in this.

### Change Type


ds300:
we decided offline this is probably a little premature for perf optimization

PR #3291, opened 2024-03-28
Display none for culled shapes

MitjaBezensek:
Comparing different culling optimizations:

https://github.com/tldraw/tldraw/assets/2523721/0b3b8b42-ed70-45b7-bf83-41023c36a563

I think we should go with the `display: none` + showing the skeleteon. The way it works is:
- We now add a sibling to the shape wrapper div which serves as the skeleton for the culled shapes.
- Only one of the two divs (shape wrapper and skeleton div) is displayed. The other one is using `display: none` to improve performance.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

- Improve performance of culled shapes by using `display: none`.


steveruizok:
Is there any way for us to still show the culled shape but use display none for the rest of the shape?

MitjaBezensek:
> Is there any way for us to still show the culled shape but use display none for the rest of the shape?

Yeah, we can do something like [this](https://github.com/tldraw/tldraw/commit/805279a5ff9314e2cc8f01782ec10450c0a692f6) (just move the display none one level lower and make the wrapper show the culled shape).

I also tested using portals in [this PR](https://github.com/tldraw/tldraw/pull/3299). Will share some comparisons next week.

packages/editor/src/lib/components/Shape.tsx
```diff
@@ -71,7 +69,13 @@ export const Shape = memo(function Shape({
 			}
 
 			// Page transform
-			const transform = Mat.toCssString(editor.getShapePageTransform(id))
+			const pageTransform = editor.getShapePageTransform(shape).clone()
+			const bounds = editor.getShapeGeometry(shape).bounds
+			if (isCulled) {
```
MitjaBezensek, line 74: We have to transform the position according to the bounds, just like we did in the `CullingShape` below

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -65,7 +65,7 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 	 * @param shape - The shape.
 	 * @public
 	 */
-	abstract component(shape: Shape): any
+	abstract component(shape: Shape, isCulled: boolean): any
```
MitjaBezensek, line 68: Not sure I 100% like this. We have to do the `display: none` logic for culled shapes in all the shapes. OTOH it does the users the ability whether they want to cull shapes or not. Could make the perf slower for custom shapes if the users are not aware of this.

I guess we would need another div if we wanted to avoid this, which is also not the best?

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -89,13 +89,6 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 	 */
 	canScroll: TLShapeUtilFlag<Shape> = () => false
 
-	/**
-	 * Whether the shape should unmount when not visible in the editor. Consider keeping this to false if the shape's `component` has local state.
-	 *
-	 * @public
-	 */
-	canUnmount: TLShapeUtilFlag<Shape> = () => true
```
MitjaBezensek, line 97: We only used this to determine whether we should cull shapes or not. The old logic unmounted the culled shapes, which we didn't want for some embeds since they had local state. Since we just do `display: none` with the new logic we can probably remove `canUnmount` completely.

ds300 (commented):
For me, requiring shapes to handle `isCulled` by default is a little too much to ask. I think we'd either need to make it opt-in for custom shapes (if they need the extra perf, seems unlikely), or just work around the need for it somehow (ideas discussed in one of my comments). 

I'd prefer to work around it if possible, because even for our shapes it feels a bit too much.

MitjaBezensek:
Yeah, I agree. Will explore it some more.

ds300:
actually doing it with webgl might kinda kick ass since we could use the geometry outline rather than just the page bounds. i.e. What if the culled clouds looked like actual clouds!! We could even use isFilled to draw outlines only.

MitjaBezensek:
Updated the PR to use a separate div for showing the culled shape. It is added as a sibling to the shape wrapper div and only one of the two is ever displayed (shape when it is not culled and the culled shape div when it is). Feels like we don't loose much in performance by using this.

MitjaBezensek:
Final benchmark. Going from around 300ms to 130ms.

https://github.com/tldraw/tldraw/assets/2523721/cde4fb74-e6c0-44eb-9cc5-c922af50eb78

packages/editor/src/lib/components/Shape.tsx
```diff
@@ -132,23 +143,30 @@ export const Shape = memo(function Shape({
...
+			>
+				<OptionalErrorBoundary fallback={ShapeErrorFallback as any} onError={annotateError}>
+					<InnerShape shape={shape} util={util} />
+				</OptionalErrorBoundary>
 			</div>
+			<div
+				ref={culledContainerRef}
+				className="tl-shape__culled"
+				draggable={false}
+				style={{ display: isCulled ? undefined : 'none' }}
```
ds300, line 168: nonblocking: I imagine we could further reduce the culling/unculling overhead by setting the display properties sneakily in our useQuickReactor effect, thus avoiding the react render cycle.

ds300 (approved):
fantastic! let's gooo

packages/editor/src/lib/components/Shape.tsx
ds300: couldn't help myself, POC here https://github.com/tldraw/tldraw/pull/3340 seems fast!

steveruizok:
Moved the display property into a layout effect instead.

PR #3300, opened 2024-03-28
Don't update hovered id when the camera is moving

MitjaBezensek:
This improves performance of interactions like zooming out / panning around. The hit testing logic is quite slow when we have a lot of shapes. Doesn't feel like we loose anything if we remove this logic when the camera is moving. 

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3061,7 +3061,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 	// box just for rendering, and we only update after the camera stops moving.
 
 	private _cameraStateTimeoutRemaining = 0
-	private _lastUpdateRenderingBoundsTimestamp = Date.now()
```
MitjaBezensek, line 3064: Wasn't being used any more.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1915,6 +1915,43 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this
 	}
 
+	/**
+	 * Calculate and update the currently hovered shape based on the current page point.
+	 *
+	 * @public
+	 */
+	updateHoveredId() {
```
MitjaBezensek, line 1923: Felt strange that this was in the `tldraw` package since it needs so much from the editor? Might be better if it lives here?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3070,18 +3106,16 @@ export class Editor extends EventEmitter<TLEventMap> {
 			this.off('tick', this._decayCameraStateTimeout)
 			this._cameraState.set('idle')
 			this.updateRenderingBounds()
+			this.updateHoveredId()
```
MitjaBezensek, line 3109: We now need to update he hovered id when we are done with the camera movements. Otherwise interactions like scrolling might land us in a state where the cursor is on top of a shape, but the shape is not hovered.

steveruizok: We would only need to do this when the select tool is active, right? We could also emit an event here that says "hey, we updated the camera bounds" and pick it up from the select tool. We did try pretty hard to split out the editor from the rest of tldraw (which includes our select tool, hence the update hovering shape id being in the tldraw package), but at this point I think it's just adding unnecessary complexity.

steveruizok:
I bet we can achieve this another way, which is to only do the "pointer move" event when the pointer actually moves? Then it wouldnt trigger the update for hovered shapes on zoom, but it would on pan (which is ok?)

MitjaBezensek:
> I bet we can achieve this another way, which is to only do the "pointer move" event when the pointer actually moves? Then it wouldnt trigger the update for hovered shapes on zoom, but it would on pan (which is ok?)

Sure. That said when I mouse scroll I often also move my mouse, it's really hard to keep it completely still 🤷‍♂️

steveruizok:
Those "get shape under point" methods (which get hovered shape relies on) are pretty complex and also pretty hot, so we should consider optimizing them at some point. I wrote them as part of the selection refactor, which was my Grand Project during the tldraw pro era, so it didn't benefit from much feedback / improvement from the team. Could be a lot better!

MitjaBezensek:
Check out this commit. On my machine it makes the `getShapeAtPoint` go from 3-6ms to 0 ms. Had 2000 shapes in the document.

https://github.com/tldraw/tldraw/commit/6d6c1e48e967289393373d17da52209996bd26c0

I experimented with [using rbush](https://github.com/tldraw/tldraw/pull/3307) for most of the collision detection stuff. Mainly for culling, but once the index is established it can be used for a lot of other things like brushing, erasing, even get shape at point.

PR #3315, opened 2024-03-30
Stickies: grid placement helpers.

steveruizok:
This PR adds automatic placement helpers for sticky notes when placed almost next to other sticky notes.

Kapture 2024-03-30 at 20 06 27

Kapture 2024-03-30 at 20 08 12

image

image

### Change Type

- [x] `feature` — New feature

### Test Plan

1. Drag sticky notes near to other sticky notes.
2. Try dragging rotated sticky notes near other sticky notes. They should not fall into a grid unless they have the same rotation.

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Adds sticky note placement helpers.


steveruizok:
Thanks to @Taha-Hassan-Git for the UX work on #3286. I here decided against the "if you've come from a pit, don't fall into one again" logic, because it felt too... slippery? Hard to put my finger on but I kept wanting to come back to the pit from whence i came

steveruizok:
And we're in! Tests written, landing

PR #3316, opened 2024-03-31
Fix: li touch not working on mobile

linonetwo:
Autocomplete of textarea (or input by default, but here only allows textarea) created using lib like https://github.com/downshift-js/downshift will use li for dropdown item.

This PR allows them to be clicked.

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. I tested by patch it in pnpm and tested on my site locally
1. or create custom shape with a li inside, with a onTouch callback

### Release Notes

- Allow touch li element on mobile

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -107,6 +107,7 @@ export function useCanvasEvents() {
 				;(e as any).isKilled = true
 				if (
 					(e.target as HTMLElement).tagName !== 'A' &&
+					(e.target as HTMLElement).tagName !== 'LI' &&
 					(e.target as HTMLElement).tagName !== 'TEXTAREA'
 				) {
 					preventDefault(e)


linonetwo:

steveruizok:
Does the li have content editable enabled?

linonetwo:
No, it is a normal non-editable li, used as dropdown menu.

I debug down here and find only whitelisted elements are receiving ontouch callback.

steveruizok:
Ah, so you want keyboard shortcuts to be turned off when the menu is open, correct? If so, we have a different way of handling this!

The menu keeps track of "open menus" and turns off keyboard shortcuts when menus are open. If you can, I'd set event handlers like:

 ```ts
onCustomMenuOpen(() => editor.addOpenMenu("my-menu"))
```

and

```ts
onCustomMenuClose(() => editor.deleteOpenMenu("my-menu")
```

linonetwo:
@steveruizok sorry for confusion, it wasn't about the keyboard or tldraw's menu.

I was creating a autocomplete input that is inside a custom note shape (not on the tldraw's menu), that will show autocomplete options in a list (with ul and li element). And user can click on an li element to fill-in the input.

This li's onClick callback works on Desktop, but onTouch callback not work on Mobile, so I have to add li to whitelist here.

steveruizok:
Could you set a listener higher up that calls stopPropagation on that event? Sorry for being cautious about something small like this, but I'm trying to understand whether this change is our only option, as I could see it causing unexpected behavior for other users later.

linonetwo:
I finally got time for opensource...

Can you guide me where to add touch and click event handler? See also https://github.com/tldraw/tldraw/discussions/3537

linonetwo:
@steveruizok You are right, I figure it out thanks to your hint!

PR #3320, opened 2024-04-01
Stickies: clone handles

steveruizok:
This PR adds "cloning handles" for note shapes.

Kapture 2024-04-01 at 17 13 57

image

### Change Type

- [x] `feature` — New feature

### Test Plan

1. Use the clone handles to create new notes by clicking.
2. Use the clone handles to create new notes by dragging.

- [x] Unit Tests

### Release Notes

- Adds clone handles to note shapes.


steveruizok:
Some initial thoughts:
- We should display these while editing a note
- We should also display the selection outline while editing a note (looks weird with handles but no outline)

PR #3326, opened 2024-04-02
Bindings

SomeHats:
First draft of the new bindings API. We'll follow this up with some API refinements, tests, documentation, and examples.

Bindings are a new record type for establishing relationships between two shapes so they can update at the same time.

### Change Type

- [x] `feature` — New feature

#### Breaking changes
- The `start` and `end` properties on `TLArrowShape` no longer have `type: point | binding`. Instead, they're always a point, which may be out of date if a binding exists. To check for & retrieve arrow bindings, use `getArrowBindings(editor, shape)` instead.
- `getArrowTerminalsInArrowSpace` must be passed a `TLArrowBindings` as a third argument: `getArrowTerminalsInArrowSpace(editor, shape, getArrowBindings(editor, shape))`
- The following types have been renamed:
    - `ShapeProps` -> `RecordProps`
    - `ShapePropsType` -> `RecordPropsType`
    - `TLShapePropsMigrations` -> `TLPropsMigrations`
    - `SchemaShapeInfo` -> `SchemaPropsInfo`


packages/tlschema/src/shapes/TLArrowShape.ts
```diff
@@ -123,16 +115,88 @@ export const arrowShapeMigrations = createShapePropsMigrationSequence({
...
+								terminal: 'start',
+								normalizedAnchor: start.normalizedAnchor,
+								isExact: start.isExact,
+								isPrecise: start.isPrecise,
+							},
+						}
+
+						oldStore[id] = binding
+						arrow.props.start = { x: 0, y: 0 }
+					}
```
ds300, line 177: should this have an else clause that deletes the `type` property on the terminal?
```suggestion
					} else {
					  delete arrow.props.start.type
					}
```

same for the end binding below?

packages/tldraw/src/lib/utils/tldr/buildFromV1Document.ts
```diff
@@ -564,18 +561,19 @@ export function buildFromV1Document(editor: Editor, document: LegacyTldrawDocume
...
+									// 	if (terminal.type === 'binding') {
+									// 		terminal.isExact = binding.distance === 0
+
+									// 		if (terminal.boundShapeId !== targetId) {
+									// 			console.warn('Hit the wrong shape!')
+									// 			terminal.boundShapeId = targetId
+									// 			terminal.normalizedAnchor = { x: 0.5, y: 0.5 }
+									// 		}
+									// 	}
+									// }
```
ds300, line 576: looks like there is some logic to add here for v1 backwards compatibility but it will look a bit different because we need to update the newly-created binding record rather than the 'change' returned from the shape util callback.

apps/examples/src/examples/custom-config/CardShape/card-shape-props.ts
```diff
@@ -1,8 +1,8 @@
-import { DefaultColorStyle, ShapeProps, T } from 'tldraw'
+import { DefaultColorStyle, RecordProps, T } from 'tldraw'
```
ds300, line 1: `ShapeProps` seems like a good and easy target for graceful deprecation: lots of people probably using it and zero complexity to avoid breaking. we can keep examples using RecordProps, but let's just export an alias of it somewhere with a `@deprecated` tag?

ds300: ah ignore me, there are a few other types in this kinda Shape* namespace, might as well just rip off the band aid.

packages/editor/src/lib/config/createTLStore.ts
```diff
@@ -57,9 +57,9 @@ export function createTLStore({
 	})
 }
 
-function currentPageShapesToShapeMap(shapeUtils: TLShapeUtilConstructor<TLUnknownShape>[]) {
```
ds300, line 60: lol why was this ever even called `currentPageShapesToShapeMap`

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5630,11 +5649,11 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-					if (this.isShapeOfType<TLArrowShape>(shape, 'arrow')) {
-						if (shape.props.start.type === 'binding' || shape.props.end.type === 'binding') {
-							return false
-						}
-					}
+					// if (this.isShapeOfType<TLArrowShape>(shape, 'arrow')) {
+					// 	if (shape.props.start.type === 'binding' || shape.props.end.type === 'binding') {
+					// 		return false
+					// 	}
+					// }
```
ds300, line 5656: hm we probably should still be ignoring arrows if they are bound to other shapes. Would a new util flag make sense here? `canBeLaidOut`

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8761,6 +8710,41 @@ function applyPartialToShape<T extends TLShape>(prev: T, partial?: TLShapePartia
 	return next
 }
 
+function applyPartialToBinding<T extends TLBinding>(prev: T, partial?: TLBindingPartial<T>): T {
```
ds300, line 8713: looks like this is identical to `applyPartialToShape` can we consolidate?

packages/editor/src/lib/editor/shapes/shared/arrow/shared.ts
```diff
@@ -81,40 +87,113 @@ function getArrowTerminalInArrowSpace(
...
+					boundShapeRelationships === 'end-contains-start'
+			)
+		: Vec.From(shape.props.end)
 
 	return { start, end }
 }
 
+/**
+ * Create or update the arrow binding for a particular arrow terminal. Will clear up if needed.
+ * TODO(alex): #bindings find a better name for this
```
ds300, line 143: `doTheArrowThingsThnx`

packages/store/src/lib/Store.ts
```diff
@@ -870,7 +870,8 @@ export class Store<R extends UnknownRecord = UnknownRecord, Props = unknown> {
 
 			updateDepth++
 			if (updateDepth > 100) {
-				throw new Error('Maximum store update depth exceeded, bailing out')
+				console.warn('Maximum store update depth exceeded, bailing out')
+				return
```
ds300, line 874: Why not throw here? It's definitely an error and we should abort any transaction it happens in I think?

ds300 (approved):
Heck yeah! Amazing work!

packages/editor/src/lib/editor/shapes/shared/arrow/shared.ts
ds300: but seriously after reading through the arrow util and stuff, i'd suggest `createOrUpdateArrowBinding` and `unbindArrowTerminal`

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -132,6 +132,13 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 	 */
 	canCrop: TLShapeUtilFlag<Shape> = () => false
 
+	/**
+	 * Whether the stack participates in stacking, aligning, and distributing.
```
ds300, line 136:
```suggestion
	 * Whether the shape participates in stacking, aligning, and distributing.
```

PR #3283, opened 2024-03-27
Decrease the number of rendered dom nodes for geo shape and arrows

MitjaBezensek:
When geo shape has no url or text we don't show the html container containing the label and link. This results in 3 fewer dom nodes per empty geo shape (going from 7 to 4). Similarly for an arrow without the text label we go from 13 to 10. 

First paint experience with 2000 empty rectangle shapes
Before: 1.5-1.6s
After: 1.2-1.3s

2000 rectangles shapes with text is similar between the two, around 3.6s.

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [ ] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Release Notes

- Reduce the number of rendered dom nodes for geo shapes and arrows without text.

PR #3328, opened 2024-04-02
textareas: place cursor at intended position

mimecuvalo:
This is take 2 on this original PR: https://github.com/tldraw/tldraw/pull/3132

It goes further to try to address these issues that came up:
- [x] fixes up problem with being able to click "through" an occluded shape because of z-layer issues. This is done by only "raising" the textareas when a shape is selected, priming them for easy selection.
- [x] add a `data-isediting` property which lets us stay in editing mode and keep the z-indices of the textareas raised when in editing mode.
- [x] Initiating a drag from within a textarea was losing pointer events when going over the UI. Would love feedback on this piece - had to disable pointer events to the UI while dragging which I think is the right move anyway. But wiring that up properly could use work since it relies on cursor changes, heh.

Btw, @MitjaBezensek I had to undo your PR https://github.com/tldraw/tldraw/pull/3283 because that would conflict with this sticky work where we want text editing to be more fluid.

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


packages/editor/editor.css
```diff
@@ -1120,24 +1147,12 @@ input,
...
 	padding: 4px;
-	height: 100%;
-	width: 100%;
-	position: absolute;
-	resize: none;
-	border: 0px;
-	user-select: all;
-	-webkit-user-select: text;
-	caret-color: var(--color-text);
-	border-image: none;
```
mimecuvalo, line 1140: the rules that were removed here were all redundant - they're covered by the new css structure anyway

packages/tldraw/src/lib/shapes/frame/FrameShapeUtil.tsx
```diff
@@ -110,18 +110,18 @@ export class FrameShapeUtil extends BaseBoxShapeUtil<TLFrameShape> {
 
 		let labelTranslate: string
 		switch (labelSide) {
-			case 0:
```
mimecuvalo, line 113: drive-by cleanup

steveruizok:
> * [x]  Initiating a drag from within a textarea was losing pointer events when going over the UI. Would love feedback on this piece - had to disable pointer events to the UI while dragging which I think is the right move anyway. But wiring that up properly could use work since it relies on cursor changes, heh.

We normally call pointer capture on the canvas. Maybe we could call pointer capture on the element instead?

steveruizok:
There are a few things about the text changes that I haven't been able to follow. Some may be from this PR and others may be from the previous PR. It may be valuable to revert the last PR, merge this one on top of it, and then consider them all as a single whole.

All of the problems I'm seeing seem to come from the fact that we now have the textareas in the DOM even when we're not editing. Previously, we would only show the regular text element but bring in the text area when editing. When we had "click to other elements to edit directly" we also showed textareas on other elements while they're selected.

Having textareas in the DOM all the time seems to enable:
- clicking into a selected text element to place the caret at the clicked location
- clicking into a different text element while editing to place the caret at the clicked location

However, it seems to complicate:
- pointer down / pointer capture when not editing a shape
- accidentally selecting text when dragging it
- not being able to double click an element to select all text
- transformations for the text shape (?)

mimecuvalo:
> We normally call pointer capture on the canvas. Maybe we could call pointer capture on the element instead?

Let's look at this offline - I want to make sure I understand where you you intend the pointer capture would change.

> There are a few things about the text changes that I haven't been able to follow. Some may be from this PR and others may be from the previous PR. It may be valuable to revert the last PR, merge this one on top of it, and then consider them all as a single whole.

We talked offline and yeah, there's a couple different PRs that maybe got conflated. The particular thing we're addressing here is this PR and we'll revert this piece and rework it into this PR: https://github.com/tldraw/tldraw/pull/3132

mimecuvalo:
> Selection dragging

can you show me this one offline? I don't think this is a regression? It's hard to tell from that gif though.

> Double clicking to select

ah we did talk about adding this behavior - i haven't added it explicitly though yet. i'll work on doing that. right now, i think what's happening in your gif is that it's selecting the shape first and then just going to the cursor position on the next click.

> Scaling text

yeah, that one's addressed already - it was a regression from a separate PR/refactor bug of textfields.

—-

let's sync more tomorrow offline!

mimecuvalo:
ok, i think we're getting there - let's look at this PR tomorrow and bang on it some more. feeling more solid!

steveruizok:
> Btw, @MitjaBezensek I had to undo your PR https://github.com/tldraw/tldraw/pull/3283 because that would conflict with this sticky work where we want text editing to be more fluid.

Is this still the case? It still seems like we only need to show the text areas:
- for the selected shape
- for all shapes when editing

PR #3329, opened 2024-04-02
stickies: translucency, sometimes

mimecuvalo:
Slight translucency, when not zoomed out. I think maybe the performance hit is worth it — if we're concerned about it, I think we can use a higher breakpoint instead of `editor.getZoomLevel() < 0.35` for these kinds of nicities/LOD. Thoughts?

(argh, I don't know why that api-report keeps changing at the moment - it's just the `NoteShapeUtil.tsx` change)


steveruizok:
i fucked this PR up

steveruizok:
but I'll land the changes in the stickies rc

mimecuvalo:
woo!

mimecuvalo:
Had to revert - caused blurring on text editing https://linear.app/tldraw/issue/TLD-2380/blurred-text-when-editing

PR #3331, opened 2024-04-03
revert: "squish sync data events before sending them out"

ds300:
Reverts tldraw/tldraw#3118

### Change type

- [x] `bugfix` 

### Test plan

- [x] Unit tests (if present)

### Release notes

- Reverted a change that squished sync data events.


ds300:
sadly this broke the draw shape updates in multiplayer rooms, will try to bring it back soon.

PR #3342, opened 2024-04-03
Fix weird text complete

steveruizok:
This PR fixes a weird issue where editing text notes would call blur / complete, preventing certain interactions.

Before:
1. Select a sticky note
2. Start editing the note
3. pointer down on the canvas
4. Instead of being in pointing_canvas, you'll be in idle

After:
4. You'll be in pointing_canvas

### Change Type

- [x] `bugfix` — Bug fix

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -795,6 +795,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	undo(): this {
+		this._flushEventsForTick(0)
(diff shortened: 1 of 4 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -795,6 +795,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	undo(): this {
+		this._flushEventsForTick(0)
```
mimecuvalo, line 798: i think this change was meant to be done on `main`?

PR #3351, opened 2024-04-04
Text fixes pass

steveruizok:
This is a pass on the text fields improvements that are part of stickies.

### Change Type

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -22,15 +22,18 @@
 	--radius-3: 9px;
 	--radius-4: 11px;
 	/* Z Index */
-	--layer-canvas-shape: 1;
-	--layer-canvas-content: 3;
-	--layer-canvas-editor: 4;
 	--layer-background: 100;
 	--layer-grid: 150;
 	--layer-canvas: 200;
 	--layer-shapes: 300;
 	--layer-overlays: 400;
 	--layer-following-indicator: 1000;
+
+	/* z index for text editors */
+	--layer-text-container: 1;
+	--layer-text-content: 3;
+	--layer-text-editor: 4;
+
 	/* Misc */
 	--tl-zoom: 1;
 
@@ -1043,7 +1046,7 @@ input,
 
 .tl-text-wrapper .tl-text-content {
 	pointer-events: all;
-	z-index: var(--layer-canvas-content);
+	z-index: var(--layer-text-content);
 }
 
 .tl-text-label__inner > .tl-text-content {
@@ -1066,25 +1069,25 @@ input,
 }
 
 .tl-text-wrapper[data-isselected='true'] .tl-text-input {
-	z-index: var(--layer-canvas-editor);
+	z-index: var(--layer-text-editor);
 	pointer-events: all;
 }
 
 /* These three rules help preserve clicking into specific points in text areas *while*
  * already in edit mode when jumping from shape to shape. */
-.tl-canvas[data-isediting='true'] .tl-text-wrapper:hover .tl-text-input {
-	z-index: var(--layer-canvas-editor);
+.tl-canvas[data-iseditinganything='true'] .tl-text-wrapper:hover .tl-text-input {
+	z-index: var(--layer-text-editor);
 	pointer-events: all;
 }
 /* This part of the rule helps preserve the occlusion rules for the shapes so we
  * don't click on shapes that are behind other shapes.
  * One extra nuance is arrows which have weird geometry and just gets in the way.
  */
-.tl-canvas[data-isediting='true'] .tl-shape:not([data-shape-type='arrow']) {
+.tl-canvas[data-iseditinganything='true'] .tl-shape:not([data-shape-type='arrow']) {
 	pointer-events: all;
 }
 /* But, re-disable the pointer-events rule for the svg container. */
-.tl-canvas[data-isediting='true'] .tl-shape .tl-svg-container {
+.tl-canvas[data-iseditinganything='true'] .tl-shape .tl-svg-container {
 	pointer-events: none;
 }
 
@@ -1166,7 +1169,7 @@ input,
 	height: 100%;
 	pointer-events: all;
 	opacity: 1;
-	z-index: var(--layer-canvas-shape);
+	z-index: var(--layer-text-container);
 	border-radius: 1px;
 }
 
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -1599,10 +1599,21 @@ export class Editor extends EventEmitter<TLEventMap> {
 	}
 
 	/**
+	 * The id of the app's only selected shape.
+	 *
+	 * @returns Null if there is no shape or more than one selected shape, otherwise the selected shape's id.
+	 *
+	 * @public
+	 * @readonly
+	 */
+	@computed getOnlySelectedShapeId(): TLShapeId | null {
+		return this.getOnlySelectedShape()?.id ?? null
+	}
+
+	/**
 	 * The app's only selected shape.
 	 *
-	 * @returns Null if there is no shape or more than one selected shape, otherwise the selected
-	 *   shape.
+	 * @returns Null if there is no shape or more than one selected shape, otherwise the selected shape.
 	 *
 	 * @public
 	 * @readonly
--- a/packages/tldraw/src/lib/shapes/frame/FrameShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/frame/FrameShapeUtil.tsx
@@ -66,7 +66,7 @@ export class FrameShapeUtil extends BaseBoxShapeUtil<TLFrameShape> {
 				const info = (resizingState as typeof resizingState & { info: { isCreating: boolean } })
 					?.info
 				if (!info) return false
-				return info.isCreating && this.editor.getOnlySelectedShape()?.id === shape.id
+				return info.isCreating && this.editor.getOnlySelectedShapeId() === shape.id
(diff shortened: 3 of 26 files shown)


packages/tldraw/src/lib/shapes/shared/TextLabel.tsx
```diff
@@ -26,7 +23,7 @@ type TextLabelProps = {
 	verticalAlign: TLDefaultVerticalAlignStyle
 	wrap?: boolean
 	text: string
-	labelColor: TLDefaultColorStyle
+	labelColor: string
```
steveruizok, line 26: First change here is superficial: let's change `labelColor` from taking a style prop to taking a regular color instead. We're generally moving in this direction anyway; but here it lets us remove the `isNote` prop that I'd added in a previous PR.

packages/tldraw/src/lib/shapes/shared/TextLabel.tsx
```diff
@@ -49,16 +46,18 @@ export const TextLabel = React.memo(function TextLabel({
 	align,
 	verticalAlign,
 	wrap,
-	bounds,
```
steveruizok, line 52: Also removing bounds here because it's not used (cloud shape uses it but it has no detectable effect).

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1598,11 +1598,22 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 *
+	 * @returns Null if there is no shape or more than one selected shape, otherwise the selected shape's id.
+	 *
+	 * @public
+	 * @readonly
+	 */
+	@computed getOnlySelectedShapeId(): TLShapeId | null {
+		return this.getOnlySelectedShape()?.id ?? null
+	}
+
```
steveruizok, line 1612: There are a lot of places in the app where we check whether a shape is the selected shape. Using the regular `getOnlySelectedShape` works but that computed fires every time the selected shape changes, which means all of those components re-render when the selected shape changes, ie when you're typing in a label.

steveruizok: This computed only tells us _which_ shape is editing, which is useful for "am i selected" checks.

packages/tldraw/src/lib/shapes/frame/FrameShapeUtil.tsx
```diff
@@ -66,7 +66,7 @@ export class FrameShapeUtil extends BaseBoxShapeUtil<TLFrameShape> {
 				const info = (resizingState as typeof resizingState & { info: { isCreating: boolean } })
 					?.info
 				if (!info) return false
-				return info.isCreating && this.editor.getOnlySelectedShape()?.id === shape.id
+				return info.isCreating && this.editor.getOnlySelectedShapeId() === shape.id
```
steveruizok, line 69: There are going to be a bunch of these, it was a find and replace for `getOnlySelectedShape()?.id`

packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -164,50 +164,53 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
...
+						isNote
+						isSelected={isSelected}
+						labelColor={theme[color].note.text}
+						wrap
+						onKeyDown={handleKeyDown}
+					/>
+				</div>
 				{'url' in shape.props && shape.props.url && (
-					<HyperlinkButton url={shape.props.url} zoomLevel={this.editor.getZoomLevel()} />
+					<HyperlinkButton url={shape.props.url} zoomLevel={zoom} />
```
steveruizok, line 211: The link can be outside of the note_container element.

packages/tldraw/src/lib/shapes/shared/TextLabel.tsx
```diff
@@ -110,25 +100,27 @@ export const TextLabel = React.memo(function TextLabel({
...
+						ref={rInput}
+						// We need to add the initial value as the key here because we need this component to
+						// 'reset' when this state changes and grab the latest defaultValue.
+						key={initialText}
+						text={text}
+						isEditing={isEditing}
+						{...editableTextRest}
+						handleKeyDown={handleKeyDownCustom ?? editableTextRest.handleKeyDown}
+					/>
+				)}
```
steveruizok, line 123: Our logic for showing the text area is:
- are we editing anything? -> yes
- are we selected? -> yes
- otherwise -> no

This lets us keep those elements out of the DOM until we need them.

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
```diff
@@ -161,14 +167,17 @@ export function useEditableText(id: TLShapeId, type: string, text: string) {
 	return {
 		rInput,
 		isEditing,
-		handleFocus: () => {
-			/* noop */
-		},
+		handleFocus: noop,
```
steveruizok, line 170: just moving this to a static function

packages/tldraw/src/lib/shapes/text/TextArea.tsx
```diff
@@ -52,9 +52,7 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(function
 			onKeyDown={handleKeyDown}
 			onBlur={handleBlur}
 			onTouchEnd={stopEventPropagation}
-			onContextMenu={(e) => {
-				isEditing && stopEventPropagation(e)
-			}}
+			onContextMenu={isEditing ? stopEventPropagation : undefined}
```
steveruizok, line 55: static function here too

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -70,31 +70,30 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
...
+				labelColor={theme[color].solid}
+				isSelected={isSelected}
+				textWidth={width}
+				textHeight={height}
+				style={{
+					transform: `scale(${scale})`,
+					transformOrigin: 'top left',
+				}}
+				wrap
+			/>
```
steveruizok, line 96: No need for the extra element here (I don't think)

packages/tlschema/api-report.md
```diff
@@ -750,6 +750,11 @@ export type ShapeProps<Shape extends TLBaseShape<any, any>> = {
     [K in keyof Shape['props']]: T.Validatable<Shape['props'][K]>;
 };
 
+// @public (undocumented)
+export type ShapePropsType<Config extends Record<string, T.Validatable<any>>> = Expand<{
+    [K in keyof Config]: T.TypeOf<Config[K]>;
+}>;
+
```
steveruizok, line 757: Girlscout: this wasn't exported but it was used in one of our examples.

packages/editor/editor.css
```diff
@@ -22,15 +22,18 @@
...
 	--layer-grid: 150;
 	--layer-canvas: 200;
 	--layer-shapes: 300;
 	--layer-overlays: 400;
 	--layer-following-indicator: 1000;
+
+	/* z index for text editors */
+	--layer-text-container: 1;
+	--layer-text-content: 3;
+	--layer-text-editor: 4;
```
steveruizok, line 35: Renamed these

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
mimecuvalo: cc @SomeHats I'm not sure - this is pretty new and I think it might be necessary for SVG export - is that right @SomeHats ?

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: killer - makes sense!

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -404,14 +407,13 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
...
-						labelColor={labelColor}
+						labelColor={theme[props.labelColor].solid}
 						wrap
-						bounds={props.geo === 'cloud' ? this.getGeometry(shape).bounds : undefined}
 					/>
-					{shape.props.url && (
-						<HyperlinkButton url={shape.props.url} zoomLevel={this.editor.getZoomLevel()} />
-					)}
 				</HTMLContainer>
+				{shape.props.url && (
```
mimecuvalo, line 414: cc @SomeHats does this need to be in the HTMLContainer for svg export?

packages/tldraw/src/lib/shapes/shared/TextLabel.tsx
```diff
@@ -110,25 +100,27 @@ export const TextLabel = React.memo(function TextLabel({
 					lineHeight: fontSize * lineHeight + 'px',
 					minHeight: lineHeight + 32,
 					minWidth: textWidth || 0,
-					color: isNote ? theme[labelColor].note.text : theme[labelColor].solid,
+					color: labelColor,
```
mimecuvalo, line 103: 🎉

packages/tldraw/src/lib/shapes/shared/TextLabel.tsx
mimecuvalo: Makes sense to me, good call 👍

mimecuvalo (approved):
thx for the tweaks! make sense to me - let's just check on that SVG export stuff

PR #3361, opened 2024-04-04
stickies: update geometry to differentiate between shape and text label

mimecuvalo:
Separates out the geometry - we want clicking in the negative space to not have the text editing mechanics. This definitely feels better without the sticky note textfield assuming it takes up the whole note.

Before:
Screenshot 2024-04-04 at 15 35 58
After:
Screenshot 2024-04-04 at 15 35 42

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
@@ -39,8 +39,8 @@ import { getFontDefForExport } from '../shared/defaultStyleDefs'
 import { useForceSolid } from '../shared/useForceSolid'
 import {
 	ADJACENT_NOTE_MARGIN,
-	CENTER_OFFSET,
 	CLONE_HANDLE_MARGIN,
+	NOTE_CENTER_OFFSET,
 	NOTE_SIZE,
 	getNoteShapeForAdjacentPosition,
 	startEditingNoteShape,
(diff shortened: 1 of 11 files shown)


packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -296,36 +321,76 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
 }
 
 function getGrowY(editor: Editor, shape: TLNoteShape, prevGrowY = 0) {
+	const unadjustedFontSize = LABEL_FONT_SIZES[shape.props.size]
+	const { labelHeight, fontSizeAdjustment } = getLabelSize(editor, shape)
```
steveruizok, line 325: The geometry is cached, so we could ask the editor (ie `editor.getShapeGeometry(shape).children[`]) rather than recalculate this?

steveruizok:
Nice one! I think this is the right way to do it. Suggestion above for avoiding double measurements (I think).

steveruizok:
Since the label measurement is kinda expensive / DOMmy, and since we do it in two places on every change (the `onAfterChange` and the `getGeometry`), I've added a weak map cache to prevent double work.

PR #3362, opened 2024-04-04
textfields: bring shape into view when editing

mimecuvalo:
This actually is a problem in production anyway, for any shape, not particular to new auto-editing or stickies.
Case in point:
- shape is selected
- move the shape offscreen
- hit Enter
- you'll be editing the shape but it won't be visible to you.

This change consolidates some of the duplicate logic in `Idle.ts` and in `noteHelpers.ts`.

Two questions
- `Idle.ts` didn't have the `select()` call but `noteHelpers.ts` did - is this really important?
- `noteHelpers` didn't have the `mark()` call but `Idle.ts` did - seems like it was missing in noteHelpers, but lemme know if that was intended to be left out.

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

- Textfields: bring shape to view that's being edited.

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/TextHelpers.ts
+++ b/packages/tldraw/src/lib/shapes/shared/TextHelpers.ts
@@ -4,6 +4,8 @@
  * Copyright (c) Federico Brigante <opensource@bfred.it> (bfred.it)
  */
 
+import { ANIMATION_MEDIUM_MS, Editor, TLShape } from '@tldraw/editor'
+
 // TODO: Most of this file can be moved into a DOM utils library.
 
 /** @internal */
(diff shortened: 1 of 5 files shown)


steveruizok:
Great idea. I updated the positioning logic so that rather than zooming the whole camera over to the new shape, we just bring it into view at the current zoom level.

packages/tldraw/src/lib/shapes/shared/TextHelpers.ts
```diff
@@ -290,3 +292,50 @@ function getCaretIndex(element: HTMLElement) {
...
+		// then use the difference between the centers to calculate the offset
+		const nextBounds = viewportPageBounds.clone().translate({
+			x: (eb.center.x - viewportPageBounds.center.x) * 2,
+			y: (eb.center.y - viewportPageBounds.center.y) * 2,
+		})
+		editor.zoomToBounds(nextBounds, {
+			duration: ANIMATION_MEDIUM_MS,
+			inset: 0,
+		})
+	}
```
steveruizok, line 340: This is that positioning edit.

steveruizok (approved):

PR #3368, opened 2024-04-05
[draft] Perf testing...

Taha-Hassan-Git:
### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/apps/examples/e2e/tests/test-performance.spec.ts
+++ b/apps/examples/e2e/tests/test-performance.spec.ts
@@ -1,0 +1,35 @@
+import { setup } from '../shared-e2e'
+import test from './fixtures/fixtures'
+
+test.describe('Performance test', () => {
+	test.beforeEach(setup)
+	test.only('Get panning performance metrics', async ({ page, isMobile }) => {
+		test.skip(isMobile, 'no debug menu on mobile')
+		const debugMenuButton = page.getByTitle('Debug menu')
+		const panningTestMenuItem = page.getByTitle('Panning test')
+		await debugMenuButton.click()
+		const session = await page.context().newCDPSession(page)
+		await session.send('Performance.enable')
+		// Wait for the panning test to complete
+		await panningTestMenuItem.click()
+		await page.waitForTimeout(5000)
+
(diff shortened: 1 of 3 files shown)


steveruizok:
Some thoughts on perf here:

We should be using a library like [tinybench](https://github.com/tinylibs/tinybench) for measurements.

Our code can be measured in three ways:
1. How long the operations in the editor work
2. How long it takes to update the React tree / DOM as a result of the operations
3. How long it takes for those operations to actually get rendered and painted in the browser

The first result should be pretty much the same between any node environments.
The second result should also be pretty much the same between any node environment.
The third is going to vary a lot between browsers.

1 is easier to set up that 2, which is easier to set up than 3.

All three will produce very different results depending on which machine runs it. It still may be helpful to have this as something that runs locally (i.e. run against main, then run against your branch). There are services that help mitigate the difference between outputs between tests when run in CI (I linked [Codspeed](https://codspeed.io/) elsewhere).

Taha-Hassan-Git:
> Some thoughts on perf here:
> 
> We should be using a library like [tinybench](https://github.com/tinylibs/tinybench) for measurements.
> 
> Our code can be measured in three ways:
> 
> 1. How long the operations in the editor work
> 2. How long it takes to update the React tree / DOM as a result of the operations
> 3. How long it takes for those operations to actually get rendered and painted in the browser
> 
> The first result should be pretty much the same between any node environments. The second result should also be pretty much the same between any node environment. The third is going to vary a lot between browsers.
> 
> 1 is easier to set up that 2, which is easier to set up than 3.
> 
> All three will produce very different results depending on which machine runs it. It still may be helpful to have this as something that runs locally (i.e. run against main, then run against your branch). There are services that help mitigate the difference between outputs between tests when run in CI (I linked [Codspeed](https://codspeed.io/) elsewhere).

Let me know if you have time to talk about this today, I'd appreciate some guidance.

PR #3307, opened 2024-03-29
[culling] Use rbush to calculate shapes that are in view.

MitjaBezensek:
Investigation into using rbush for calculating the culled shapes.

### Change Type

- [ ] `sdk` — Changes the tldraw SDK
- [ ] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Test Plan

1. Add a step-by-step description of how to test your PR here.
2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Add a brief release note for your PR here.

PR #3376, opened 2024-04-05
[culling] Improve setting of display none.

MitjaBezensek:
Small improvement for culling shapes. We now use reactor to do it. .

Before:
image

After:
image

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/packages/editor/src/lib/components/Shape.tsx
+++ b/packages/editor/src/lib/components/Shape.tsx
@@ -118,30 +118,15 @@ export const Shape = memo(function Shape({
 		[opacity, index, backgroundIndex]
 	)
 
-	// This stuff changes pretty infrequently, so we can change them together
 	useQuickReactor(
 		'set display',
 		() => {
 			const shape = editor.getShape(id)
 			if (!shape) return // probably the shape was just deleted
 
-			// If renderingBoundsMargin is set to Infinity, then we won't cull offscreen shapes
-			const isCullingOffScreenShapes = Number.isFinite(editor.renderingBoundsMargin)
-			const selectedShapeIds = editor.getSelectedShapeIds()
-			const renderingBoundsExpanded = editor.getRenderingBoundsExpanded()
-			const maskedPageBounds = editor.getShapeMaskedPageBounds(id)
-			const isCulled =
-				isCullingOffScreenShapes &&
-				// never cull editing shapes
-				editor.getEditingShapeId() !== id &&
-				// if the shape is fully outside of its parent's clipping bounds...
-				(maskedPageBounds === undefined ||
-					// ...or if the shape is outside of the expanded viewport bounds...
-					(!renderingBoundsExpanded.includes(maskedPageBounds) &&
-						// ...and if it's not selected... then cull it
-						!selectedShapeIds.includes(id)))
+			// console.log(shape.id)
+			const isCulled = editor.isShapeCulled(shape)
 			setStyleProperty(containerRef.current, 'display', isCulled ? 'none' : 'block')
-			setStyleProperty(culledContainerRef.current, 'display', isCulled ? 'block' : 'none')
 			setStyleProperty(bgContainerRef.current, 'display', isCulled ? 'none' : 'block')
 		},
 		[editor]
--- a/packages/editor/src/lib/components/Shape.tsx
+++ b/packages/editor/src/lib/components/Shape.tsx
@@ -1,6 +1,6 @@
 import { useQuickReactor, useStateTracking } from '@tldraw/state'
 import { TLShape, TLShapeId } from '@tldraw/tlschema'
-import { memo, useCallback, useLayoutEffect, useRef } from 'react'
+import { memo, useCallback, useRef } from 'react'
 import { ShapeUtil } from '../editor/shapes/ShapeUtil'
 import { useEditor } from '../hooks/useEditor'
 import { useEditorComponents } from '../hooks/useEditorComponents'
@@ -26,7 +26,6 @@ export const Shape = memo(function Shape({
 	index,
 	backgroundIndex,
 	opacity,
-	isCulled,
 	dprMultiple,
 }: {
 	id: TLShapeId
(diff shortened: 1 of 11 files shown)


packages/editor/src/lib/components/Shape.tsx
```diff
@@ -129,14 +127,34 @@ export const Shape = memo(function Shape({
...
-	useLayoutEffect(() => {
-		const container = containerRef.current
-		const bgContainer = bgContainerRef.current
-		const culledContainer = culledContainerRef.current
-		setStyleProperty(container, 'display', isCulled ? 'none' : 'block')
-		setStyleProperty(bgContainer, 'display', isCulled ? 'none' : 'block')
-		setStyleProperty(culledContainer, 'display', isCulled ? 'block' : 'none')
-	}, [isCulled])
+	// This stuff changes pretty infrequently, so we can change them together
+	useQuickReactor(
```
MitjaBezensek, line 131: Moved it to its own reactor. We don't want panning to trigger other ones.

steveruizok:
The layout effect should only change when a shape switches from being culled to not-culled, right?

This feels like a half-measure. What's the intended design? To avoid calculating `isCulled` in the rendering shapes array, or to avoid it coming in as a prop, or...?

MitjaBezensek:
> The layout effect should only change when a shape switches from being culled to not-culled, right?

Yeah, but the performance difference is quite big as you can see. And you pay it on every pan, zoom, click on the minimap.

> This feels like a half-measure. What's the intended design? To avoid calculating `isCulled` in the rendering shapes array, or to avoid it coming in as a prop, or...?

Yeah, I'd like to rework this:
- The `Shape.tsx` would not be aware of culling, only the reactor so that it can set the display none.
- We have separate apis for the rendering of shapes (which `ShapesToDisplay` uses) and for hit testing (use for brushing erasing, hit point testing). Right now we `getRenderingShapes` for both cases, but it feels like it has out lived the purpose.

One suggestion [is here with](https://github.com/tldraw/tldraw/pull/3344/files#diff-af729319507f9150354b0c977c8133793a5ceeb736b99cad1c26aa8ba195886c) `isShapeCulled`, which you can call on demand and we'd remove it from the rendering shapes. But that also doesn't feel completely right, so I'm investigating how to change the apis and use rbush for the hit testing part.

steveruizok:
I'm sorry, I still don't understand.

With the layout effect, if a shape's culling status has not changed, then the shape would not receive a new prop, and so it would not render, and the layout effect would not fire.

In this PR, it seems that the reactor would fire in more circumstances.

steveruizok:
It seems that the bigger savings is in preventing the `getRenderingShapes` from running on every change. Perhaps removing the culling logic from that hook is the big improvement, because it would cause it to run less often? But we would still need to do work whenever shapes were reordered, changed, or the camera moved; so the reasoning was always to calculate it in one place rather than many.

ds300:
@steveruizok the big speedup simply comes from circumventing the react render lifecycle when the culling status of shapes change. It adds perceptible overhead when there are thousands of shapes in play.

> isShapeCulled

@MitjaBezensek I was imagining this PR containing that helper, while removing `isCulled` from the rendering array. We could make the `isShapeCulled` helper use createComputedCache under the hood and then we'd minimize the number of invocations of the quick reactor in Shape.tsx.

steveruizok:
Makes sense @ds300. If we're having every shape call a helper like `isShapeCulled`, which seems like it would be doing a lot of work for each call, would it be better to have a `getCulledShapes` computed that returns a set or something? Similar to `useRenderingShapes` but just for that fact?

MitjaBezensek:
> Makes sense @ds300. If we're having every shape call a helper like `isShapeCulled`, which seems like it would be doing a lot of work for each call, would it be better to have a `getCulledShapes` computed that returns a set or something? Similar to `useRenderingShapes` but just for that fact?


That's where rbush comes in. The ideas there is:
- We create what I called a spatial index (just a rbush tree), which we incrementally update. We can do this in a quite granular way: only shape positions affect it, nothing else. Not selected shapes, not even the viewport bounds. It's just a representation of shapes in space.
- We can then very quickly answer not just which shapes are culled, but also things like which shapes are close to the mouse pointer,... For that purpose I [created `getShapesInsideBounds` function](https://github.com/tldraw/tldraw/pull/3307/files#diff-af729319507f9150354b0c977c8133793a5ceeb736b99cad1c26aa8ba195886cR3111-R3114).

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4637,8 +4658,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	@computed getCurrentPageRenderingShapesSorted(): TLShape[] {
-		return this.getRenderingShapes()
-			.filter(({ isCulled }) => !isCulled)
+		return this.getUnorderedRenderingShapes(true)
```
MitjaBezensek, line 4661: We can use the unordered one since we sort below and loose the order anyway.

MitjaBezensek:
Added cache for culling info. Don't see much of a change compared to the solution without the cache🤷‍♂️  Might have missed something, or maybe it's just that everything is cached anyway (maskedPagebounds,..) so there's not much to gain.

steveruizok:
I think that cache is going to bust really often, ie for every shape any time the camera moves, or selection changes, or editing shape changes.

MitjaBezensek:
> I think that cache is going to bust really often, ie for every shape any time the camera moves, or selection changes, or editing shape changes.

True. We do always request the same info twice though (from webgl culled shape rendering and from the shape reactor) so we always have at least 1 hit 🤷‍♂️ Then also some hits from the hit testing.

ds300 (approved):
🙇🏼

ds300:
> I think that cache is going to bust really often, ie for every shape any time the camera moves, or selection changes, or editing shape changes.

It's not every time the camera moves at the moment, it's just when the viewport culling bounds updates.

PR #3377, opened 2024-04-05
[culling] minimal culled diff with webgl

steveruizok:
This PR extracts the #3344 changes to a smaller diff against main. It does not include the changes to how / where culled shapes are calculated, though I understand this could be much more efficiently done!

### Change Type

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -24,6 +24,7 @@
 	/* Z Index */
 	--layer-background: 100;
 	--layer-grid: 150;
+	--layer-culled-shapes: 175;
 	--layer-canvas: 200;
 	--layer-shapes: 300;
 	--layer-overlays: 400;
@@ -234,6 +235,20 @@ input,
 	content-visibility: auto;
 	touch-action: none;
 	contain: strict;
+}
+
+.tl-culled-shapes {
+	width: 100%;
+	height: 100%;
+	z-index: var(--layer-culled-shapes);
+	position: absolute;
+	pointer-events: none;
+	contain: size layout;
+}
+
+.tl-culled-shapes__canvas {
+	width: 100%;
+	height: 100%;
 }
 
 .tl-shapes {
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -285,12 +285,15 @@ input,
 /* ------------------- Background ------------------- */
 
 .tl-background {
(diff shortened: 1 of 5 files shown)


MitjaBezensek:
This creates a bunch of layers. I pulled `Background` one level up, as well as the `CulledShapes` to avoid that.

image

steveruizok:
Ah right, that makes sense!

packages/editor/editor.css
```diff
@@ -269,13 +284,16 @@ input,
 
 /* ------------------- Background ------------------- */
 
+.tl-background {
```
MitjaBezensek, line 287:
```suggestion
.tl-background__wrapper {
```

packages/editor/src/lib/components/CulledShapes.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+			const verticesArray = shapeVertices.get()
+
+			context.clear(context.COLOR_BUFFER_BIT | context.DEPTH_BUFFER_BIT)
+
+			if (verticesArray.length > 0) {
+				const viewport = editor.getViewportPageBounds() // when the viewport changes...
+				context.uniform2f(viewportStartUniformLocation, viewport.minX, viewport.minY)
+				context.uniform2f(viewportEndUniformLocation, viewport.maxX, viewport.maxY)
+				const triangleGeoCpuBuffer = new Float32Array(verticesArray)
```
TodePond, line 159: @MitjaBezensek is this happening whenever the culled shapes change? we could probably pull this out, and re-use the same array, so we don't need to allocate so much memory each time

MitjaBezensek: Yes, this is happening all the time. I guess we would then need to be checking if the number of shapes changed and only create a new array in that case?

That said you can completely remove the culled shapes logic and it's hard to even notice any perf improvements since the whole thing is so fast.

TodePond: > I guess we would then need to be checking if the number of shapes changed and only create a new array in that case?

No, we can have the array be oversized, and always leave it the same size. Then in `drawArrays` we specify how many triangles there are to draw, so that it doesn't draw all of them

MitjaBezensek: Ah, I see 👍

PR #3379, opened 2024-04-05
[experiment] Scarily blazingly faster resizing by manually creating a history entry

steveruizok:
This PR improves the speed of resizing many shapes. 

Kapture 2024-04-05 at 18 11 59

Previously, the `updateShapes` command would be called multiple times each frame for each resizing shape. As part of this work, changes would be squashed into the history entry. This squashing was very slow. In this PR, we manage the history entry ourselves in the resizing interaction.

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [x] `improvement` — Improving existing features

### Test Plan

1. Resize lots of shapes.
2. Ensure that undo / redo / cancelling works as expected.

### Release Notes

- Improves performance of resizing.

PR #3382, opened 2024-04-06
[perf] Improve resizing performance (better big updates)

steveruizok:
This PR improves the speed of resizing many shapes. See #3379 for original.

Kapture 2024-04-05 at 18 11 59

Previously, the `updateShapes` command would be called multiple times each frame for each resizing shape. As part of this work, changes would be squashed into the history entry. This squashing was very slow. In this PR, we manage the history entry ourselves in the resizing interaction.

### Change Type

### Test Plan

1. Resize lots of shapes.
2. Ensure that undo / redo / cancelling works as expected.

### Release Notes

- Improves performance of resizing.


packages/editor/src/lib/editor/resize.ts
```diff
@@ -0,0 +1,243 @@
...
+import { Editor, TLResizeShapeOptions } from './Editor'
+
+/**
+ * Get the partial for a resized shape.
+ *
+ * @param id - The id of the shape to resize.
+ * @param scale - The scale factor to apply to the shape.
+ * @param options - Additional options.
+ *
+ * @public
```
ds300, line 16: this feels like an implementation detail to me, should we keep it internal for now?

ds300 (commented):
This is a clever optimization and a big UX win. But I'd say let's hold off on merging for a day or two until Alex has had a chance to flesh out her work on undo/redo, which I suspect will both make this optimization unnecessary and allow other commands to benefit from the same kinds of perf wins.

steveruizok:
Closing for now, I'll use this as reference after Alex's PR is done

PR #3385, opened 2024-04-06
Fix typo in Store.ts

OrionReed:
An immense contribution, I know.

### Change Type

- [ x ] `docs` — Changes to the documentation, examples, or templates.

- [ x ] `chore` — Updating dependencies, other boring stuff

PR #3395, opened 2024-04-08
Allocate memory only once

MitjaBezensek:
Allocate the memory only once. This does mean we use more memory on average, but we only allocate it once instead of doing it on every change to shape / camera ....

A bit [more info](https://github.com/tldraw/tldraw/pull/3377#discussion_r1555513083).

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/packages/store/api/api.json
+++ b/packages/store/api/api.json
@@ -4231,7 +4231,7 @@
             {
               "kind": "Property",
               "canonicalReference": "@tldraw/store!Store#onBeforeChange:member",
-              "docComment": "/**\n * A callback before after each record's change.\n *\n * @param prev - The previous value, if any.\n *\n * @param next - The next value.\n */\n",
+              "docComment": "/**\n * A callback fired before each record's change.\n *\n * @param prev - The previous value, if any.\n *\n * @param next - The next value.\n */\n",
               "excerptTokens": [
                 {
                   "kind": "Content",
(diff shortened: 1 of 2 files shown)


packages/store/api/api.json
MitjaBezensek, line 1: Seems to be coming from another PR https://github.com/tldraw/tldraw/pull/3385

TodePond:
I think this is a case where it be very useful to do an 'evaluation' measurement, to compare before/after :)
I think that this approach is better, as long as we're ok with the higher-on-average memory cost, as speed seems to be our priority at the moment

MitjaBezensek:
I don't see any difference between the two for 2k shapes. Will close this one as it does use up more memory.

Creating the array once and setting it when things change:

image

Creating a new array each time:

image

TodePond:
It's possible the browser is abstracting both into the same thing, though one thing I'd say is - I find that single-line measurement like this isn't very reliable with javascript. I'd prefer to measure an entire interaction, eg: panning the camera, and measuring time to render, or something like that. As the engine can 'move' the actual memory allocation somewhere else, meaning we're not measuring the right thing

mimecuvalo:
if you revisit this, you could have it grow as needed by powers of 2 (and not shrink). 
even if you just prime it with a minimum of 64 shapes (for 3kb of memory) and then grow it from there, that feels like it'd be a little win. the max of 2000 shapes would be 100kb so it does feel like allocating/deallocating that on that fly constantly might be a slight perf hit.

PR #3401, opened 2024-04-08
Perf: (slightly) faster min dist checks

steveruizok:
This PR improves a bunch of places where we do "minimum distance checks". Previously, we were using `Vec.Dist`, which uses `Math.hypot` to find the actual distance, but we can just as well use the squared distance. So this PR makes a small improvement to `Vec.Dist2` and then switches to that method when checking minimum distances.

### Change Type

### Test Plan

- [x] Unit Tests

### Release Notes

- Performance: small improvements to hit testing.


packages/editor/src/lib/primitives/Vec.ts
```diff
@@ -308,19 +308,20 @@ export class Vec {
...
+	// Get the distance between two points.
 	static Dist(A: VecLike, B: VecLike): number {
 		return Math.hypot(A.y - B.y, A.x - B.x)
 	}
 
+	// Get the squared distance between two points. This is faster to calculate (no square root) so useful for "minimum distance" checks where the actual measurement does not matter.
+	static Dist2(A: VecLike, B: VecLike): number {
+		return (A.x - B.x) * (A.x - B.x) + (A.y - B.y) * (A.y - B.y)
+	}
+
```
steveruizok, line 324: Remove the new Vector created by Vec.Sub

PR #3403, opened 2024-04-08
Fix some tests

steveruizok:
This PR fixes some jest test.

- We skip the culling shapes in test environments.
- We skip rendering patterns in test environments.

### Change Type

- [x] `tests` — Changes to any test code


steveruizok:
Not sure if this is the best way but it's a way!

PR #3405, opened 2024-04-08
Fix kickout helper for all rotated frames 

TodePond:
This PR fixes the kickout helper to make it work for rotated frames.

https://github.com/tldraw/tldraw/assets/15892272/2ec6f74f-9713-401c-a748-7c9d5c700fa8

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
@@ -10,7 +10,7 @@ import {
 /** @internal */
 export function kickoutOccludedShapes(editor: Editor, shapeIds: TLShapeId[]) {
 	const shapes = shapeIds.map((id) => editor.getShape(id)).filter((s) => s) as TLShape[]
-	const effectedParents = shapes
+	const affectedParents = shapes
 		.map((shape) => {
 			const parent = editor.getShape(shape.parentId)
 			if (!parent) return shape
@@ -19,7 +19,7 @@ export function kickoutOccludedShapes(editor: Editor, shapeIds: TLShapeId[]) {
 		.filter((shape) => shape.type === 'frame')
 
 	const kickedOutChildren: TLShapeId[] = []
-	for (const parent of effectedParents) {
+	for (const parent of affectedParents) {
 		const childIds = editor.getSortedChildIdsForParent(parent.id)
 
 		// Get the bounds of the parent shape
--- a/packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
@@ -2,6 +2,7 @@ import {
 	Editor,
 	TLShape,
 	TLShapeId,
+	pointInPolygon,
 	polygonIntersectsPolyline,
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
```diff
@@ -2,20 +2,21 @@ import {
...
+	pointInPolygon,
 	polygonIntersectsPolyline,
 	polygonsIntersect,
 } from '@tldraw/editor'
 
 /** @internal */
 export function kickoutOccludedShapes(editor: Editor, shapeIds: TLShapeId[]) {
 	const shapes = shapeIds.map((id) => editor.getShape(id)).filter((s) => s) as TLShape[]
-	const effectedParents: TLShape[] = shapes
+	const effectedParents = shapes
```
mimecuvalo, line 13: nit: `affectedParents` :)

packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
```diff
@@ -34,38 +35,41 @@ export function kickoutOccludedShapes(editor: Editor, shapeIds: TLShapeId[]) {
...
-	if (!occluderPageBounds) return false
-
 	const shapePageBounds = editor.getShapePageBounds(shape)
-	if (!shapePageBounds) return true
-
-	// If the shape's bounds are completely inside the occluder, it's not occluded
-	if (occluderPageBounds.contains(shapePageBounds)) {
-		return false
-	}
+	if (!occluderPageBounds) return true
```
mimecuvalo, line 45: sorry, just to make sure i understand this part of the change:
now, the !occluderPageBounds returns true instead of false, and !shapePageBounds returns false instead of true. why? was that related to this fix, or unrelated?

mimecuvalo (approved):
pre-approved - couple q's though!

packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
TodePond: It's another think I tweaked in this PR!

It's a weird niche case: "What should happen when either shape has a size of zero?"
I think that, if a parent has zero size, then technically the child is *always* occluded, right?
And if a child has zero size, then it can't possibly be occluded, as it....... doesn't exist??? Weird one. I don't think this would every happen on tldraw.com

packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
TodePond: whoops fixed :)

packages/tldraw/src/lib/tools/SelectTool/selectHelpers.ts
mimecuvalo: haha, ok gotcha - i think that makes sense to me! agreed that it's sorta weird either way you slice it

PR #3411, opened 2024-04-09
Perf: Incremental culled shapes calculation.

MitjaBezensek:
Reworks our culling logic:
- No longer show the gray rectangles for culled shapes. 
- Don't use `renderingBoundExpanded`, instead we now use `viewportPageBounds`. I've removed `renderingBoundsExpanded`, but we might want to deprecate it?
- There's now a incremental computation of non visible shapes, which are shapes outside of `viewportPageBounds` and shapes that outside of their parents' clipping bounds.
- There's also a new `getCulledShapes` function in `Editor`, which uses the non visible shapes computation as a part of the culled shape computation.
- Also moved some of the `getRenderingShapes` tests to newly created `getCullingShapes` tests.

Feels much better on my old, 2017 ipad (first tab is this PR, second is current prod, third is staging).

https://github.com/tldraw/tldraw/assets/2523721/327a7313-9273-4350-89a0-617a30fc01a2

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Regular culling shapes tests. Pan / zoom around. Use minimap. Change pages.

- [x] Unit Tests
- [ ] End to end tests


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3224,19 +3225,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 	/** @internal */
 	private readonly _renderingBounds = atom('rendering viewport', new Box())
 
-	/**
-	 * The current rendering bounds in the current page space, expanded slightly. Used for determining which shapes
-	 * to render and which to "cull".
-	 *
-	 * @public
-	 */
-	getRenderingBoundsExpanded() {
```
MitjaBezensek, line 3233: Should we deprecate it?

steveruizok:
Tempted to switch from "not visible" / "culled" shapes to "visible shapes" or "shapesInViewport".

packages/editor/src/lib/editor/Editor.ts
steveruizok: Probably ok to just yank, it should have been internal.

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -0,0 +1,95 @@
...
+		const viewportPageBounds = editor.getViewportPageBounds()
+		prevViewportPageBounds = viewportPageBounds.clone()
+		const notVisibleShapes = new Set<TLShapeId>()
+		shapes.forEach((id) => {
+			if (isShapeNotVisible(editor, id, viewportPageBounds)) {
+				notVisibleShapes.add(id)
+			}
+		})
+		return notVisibleShapes
+	}
```
ds300, line 40: are we still planning to do an incremental rbush to make this step faster?

MitjaBezensek:
> Tempted to switch from "not visible" / "culled" shapes to "visible shapes" or "shapesInViewport" or "renderingShapes"

Yeah, also not fan of `notVisibleShapes`.

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -0,0 +1,95 @@
...
+		const currentPageId = editor.getCurrentPageId()
+		if (lastPageId !== currentPageId) {
+			return fromScratch(editor)
+		}
+		const viewportPageBounds = editor.getViewportPageBounds()
+		if (!prevViewportPageBounds || !viewportPageBounds.equals(prevViewportPageBounds)) {
+			return fromScratch(editor)
+		}
+		const nextValue = new Set(prevValue)
+		let isDirty = false
```
ds300, line 62: I prefer to lazily init the next value if possible
```suggestion
		let nextValue = null as null | Set<TLShapeId>
    const add = (id: TLShapeId) => {
      if (!nextValue) nextValue = new Set(prevValue)
      nextValue.add(id)
    }
    const delete = (id: TLShapeId) => {
      if (!nextValue) nextValue = new Set(prevValue)
      nextValue.delete(id)
    }
```
And then at the end you just do

```diff
return nextValue ?? prevValue
```

We have a helper for doing this kind of thing called `IncrementalSetConstructor` but it also keeps track of the diff between versions which we don't need to do here yet.

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		const checkShapeCullingInfo = (id: TLShapeId) => {
+			if (!prevValue.has(id) && isShapeNotVisible(editor, id, viewportPageBounds)) {
+				nextValue.add(id)
+				isDirty = true
+			}
```
ds300, line 68: I think for `updated` records you need an else clause here for the case when the `id` was previously in the set and it should no longer be.

packages/tldraw/src/test/getCulledShapes.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+	expect(editor.getCulledShapes()).toStrictEqual(new Set([ids.A, ids.B, ids.C, ids.D]))
+
+	editor.select(ids.B)
+	// We don't cull selected shapes
+	expect(editor.getCulledShapes()).toStrictEqual(new Set([ids.A, ids.C, ids.D]))
+
+	editor.setEditingShape(ids.C)
+	// or shapes being edited
+	expect(editor.getCulledShapes()).toStrictEqual(new Set([ids.A, ids.D]))
+})
```
ds300, line 48: the incremental logic is not really being tested here. I'd recommend doing some light fuzzing to test that the `fromScratch` implementation always returns the same thing as the incremental logic while moving shapes on/off-screen, adding shapes, and deleting shapes. I recommend using the `--coverage` option of jest to make sure it's covering all the branches.

packages/editor/src/lib/components/Shape.tsx
```diff
@@ -124,7 +124,8 @@ export const Shape = memo(function Shape({
 			const shape = editor.getShape(id)
 			if (!shape) return // probably the shape was just deleted
 
-			const isCulled = editor.isShapeCulled(shape)
+			const culledShapes = editor.getCulledShapes()
+			const isCulled = culledShapes.has(id)
 			setStyleProperty(containerRef.current, 'display', isCulled ? 'none' : 'block')
 			setStyleProperty(bgContainerRef.current, 'display', isCulled ? 'none' : 'block')
```
ds300, line 130: i did some perf testing. We run these `setStyleProperty` on every frame for every shape while panning, even when the isCulled value hasn't changed, and it seems to take up quite a lot of time. You could fix that by using a useRef to store the previous isCulled value and returning early if it didn't change.

steveruizok: I can add this, I did this once before

steveruizok: Done!

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
MitjaBezensek: Nice! Also moved some checks into these two new functions, it nicely simplified code.

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		let nextValue = null as null | Set<TLShapeId>
+		const addId = (id: TLShapeId) => {
+			// Already added
+			if (prevValue.has(id)) return
+			if (!nextValue) nextValue = new Set(prevValue)
+			nextValue.add(id)
+		}
+		const deleteId = (id: TLShapeId) => {
```
MitjaBezensek, line 69: Added `Id` at the end of both functions due to delete being reserved.

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
steveruizok: To me it seems plenty fast enough right now!

MitjaBezensek: Tested it [here](https://github.com/tldraw/tldraw/pull/3420), seems the current from scratch is quite fast. I switched between two areas, one with 1999 shapes and the other one with 1 shape.

This branch:
CleanShot 2024-04-09 at 16 45 46

Rbush for from scratch:

image

I guess the main benefit would be in the incremental part, but that one seems fast enough.

ds300 (approved):
let's go! thanks for adding the fuzz tests

PR #3412, opened 2024-04-09
Stickies: Emit `text-select-all` event

steveruizok:
This PR adds a `select-all-text` event that the editor emits when it intends for the editing shape's text to be selected. This replaces some temporary code that would query the shape's textarea element directly.

### Change Type

### Test Plan

For geo shapes / arrows / note shapes (any shape with a label):
1. Press enter to select all of its text.
2. Double click the shape to select all of its text
3. Select a text shape and press enter

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -2190,6 +2190,10 @@ export interface TLEventMap {
         count: number;
     }];
     // (undocumented)
+    'select-all-text': [{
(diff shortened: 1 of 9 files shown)


mimecuvalo (approved):
niiiiice. thank you for cleaning this up!

you could get rid of `id={`text-input-${id}`}` on TextArea/TextLabel if you want but not necessary

PR #3414, opened 2024-04-09
[experiment] Don't show culling shapes

steveruizok:


ds300:
I'm surprised how performant this is without the rbush! Still, while profiling you can see it does spend a huge amount of time on each frame checking which shapes are culled. And indeed, if you test this against staging with 4x cpu throttling, you'll see it's way slower for panning, and I'm certain it'd feel quite rough on lower end devices. So we should still stick to the original plan and wait for the incremental rbush before assessing the viability of this culling-on-every-frame business. Seems promising though!

PR #3417, opened 2024-04-09
textfields: fix Safari cursor rendering bug, take 2

mimecuvalo:
This is another take on https://github.com/tldraw/tldraw/pull/3373 and it _probably_ works better. However, I need to test with some hardware iOS devices because I think my simulator might be giving me some BS behavior. It might be that this PR breaks iOS editing. For Safari desktop it does seem to solve the cursor bug though.


steveruizok:
On iOS, this causes the keyboard to flash open and closed again. Maybe there's something we can do with selection instead?

mimecuvalo:
@steveruizok I'm curious now that your PR https://github.com/tldraw/tldraw/pull/3423 is merged into this one. Can you check again? I wanna make sure that the input buffering factor isn't affecting what you saw on your iOS device.

steveruizok:
Ok, tested on iPhone:
- creating text shapes works as expected (they didn't work before)
- editing text shapes / notes / labels fails as shown


https://github.com/tldraw/tldraw/assets/23072548/8db81dad-411e-472b-9349-b496912ee96e

PR #3418, opened 2024-04-09
Perf: block hit tests while moving camera

steveruizok:
This PR uses an element that prevents hit tests on shapes while the camera is moving.

https://github.com/tldraw/tldraw/assets/23072548/9905f3d4-ba64-4e4d-ae99-194f513eaac8

### Change Type

### Test Plan

1. Move the camera.
2. Interact with the canvas.
3. Zoom in and out.

### Release Notes

- Improves performance of canvas while the camera is moving.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -29,6 +29,7 @@
 	--layer-shapes: 300;
 	--layer-overlays: 400;
 	--layer-following-indicator: 1000;
+	--layer-blocker: 10000;
 	/* Misc */
 	--tl-zoom: 1;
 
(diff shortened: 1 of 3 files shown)


mimecuvalo:
so good

ds300:
@steveruizok instead of this could we just put the following line at the top of getShapeAtPoint ?

```
if (this.getCameraState() !== 'idle') return
```

MitjaBezensek (approved):
Nice!

Not sure if the feature flag changes are leftovers from testing or if you wanted to commit them?

PR #3419, opened 2024-04-09
Perf: throttle `updateHoveredId`

steveruizok:
This PR throttles the `updateHoveredId` call so that it happens ever 30ms.

### Change Type

### Release Notes

- Improves canvas performance by throttling the update to the editor's hovered id.


packages/tldraw/src/lib/tools/selection-logic/updateHoveredId.ts
```diff
@@ -30,3 +30,6 @@ export function updateHoveredId(editor: Editor) {
 
 	return editor.setHoveredShape(shapeToHover.id)
 }
+
+export const updateHoveredId =
+	process.env.NODE_ENV === 'test' ? _updateHoveredId : throttle(_updateHoveredId, 32)
```
mimecuvalo, line 35: why 32? seems oddly specific? why not the typical 33, for example?

PR #3420, opened 2024-04-09
[test] rbush for from scratch part

MitjaBezensek:
Using rbush for the from scratch part.


packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -26,17 +50,24 @@ export const notVisibleShapes = (editor: Editor) => {
 	let prevViewportPageBounds: Box
 
 	function fromScratch(editor: Editor): Set<TLShapeId> {
```
steveruizok, line 52: This `fromScratch` also gets called whenever the viewport changes. The shapes' masked page bounds won't change in this case, so we'd be re-populating the data structure with the same data as before. IIRC this is fast as shit and shouldn't matter, but perhaps there's a meaningful difference between `fromScratch` (ie when pages change) and `fromViewportChange`.

PR #3422, opened 2024-04-09
Stickies: restore tab in text shape

steveruizok:
This PR restores the tab key for the text shape.


mimecuvalo:
hah, ok - but feels like it should be more flipped logic, i.e. `enableTab` for the specific one you want

PR #1910, opened 2023-09-18
Snapping and selection roundup

steveruizok:
This PR:
- adds `canSnap` as a property to handle and ignores snapping when dragging a handle that does not have `canSnap` set to true. Arrows no longer snap.
- adds `isLabel` to Geometry2d
- fixes selection on empty text labels
- fixes vertices / snapping for empty text labels

### Change Type

- [x] `minor` — New feature

### Test Plan

- [x] Unit Tests

PR #3424, opened 2024-04-09
textfields: on mobile edit→edit, allow going to empty geo

mimecuvalo:
It doesn't quite feel like this is the right fix but it does solve the issue. I was trying to see if `getShapeAtPoint` needed more work but the further I went in that rabbit hole it seemed like I shouldn't touch that code without causing a bunch of disruption at the moment.

Specifically, the code that does `Check labels first` in Editor.ts is a little obscure (lines 4384-4397). It only checks a couple specifics shapes (with certain combinations, i.e. a geo with "none" fill) _and_ it doesn't check `hitLabels` which also maybe feels wrong? I tried unraveling it but there's a lot of code relying on it at the moment to mess with it in the stickies work.
(I was looking at https://github.com/tldraw/tldraw/pull/1910 and https://github.com/tldraw/tldraw/pull/1806 for historical context fwiw)

Before:

https://github.com/tldraw/tldraw/assets/469604/b263a192-2085-4ffb-9e47-6e9c32abe1f9

After:

https://github.com/tldraw/tldraw/assets/469604/5b0b422b-dd5c-4593-9ac5-dec595923ea6


mimecuvalo:
this PR got messed up - see https://github.com/tldraw/tldraw/pull/3469

PR #3429, opened 2024-04-10
Perf: Improve text outline performance

steveruizok:
We use text shadows to create "outlines" around text shapes. These shadows are rendered on the GPU. In Chrome (and on computers with a capable GPU) text shadows work pretty well, however on Safari—and in particular on iOS—they cause massive frame drops.

https://github.com/tldraw/tldraw/assets/23072548/b65cbcaa-6cc3-46f3-b54d-1f9cc07fc499

This PR:
- adds an LOD to text shadows, removing them at < 35% zoom
- removes text shadows entirely on Safari

If we had a "high performance" or "low-end device" mode, then shadows / text shadows would be the first to go.

### Change Type

### Test Plan

1. Use text shapes on iOS.
2. Use text shapes on Safari.
3. Use text shapes on Chrome.

### Release Notes

- Improves performance of text shapes on iOS / Safari.


steveruizok:
we must free the gpu

PR #3447, opened 2024-04-11
Performance measurement tool (for unit tests)

steveruizok:
This PR adds a micro benchmarking utility. We can use it in our jest tests or in random scripts, though given the other requirements of our library, benchmarking.

Screenshot 2024-04-11 at 2 44 23 PM

## What this isn't

This is not benchmarking. The speeds etc are based on your machine.

## What this is

This is a tool for measuring / comparing different implementations etc. Some things run much faster than others.

### Change Type

- [x] `sdk`
- [x] `internal`


MitjaBezensek (approved):
Nice!

PR #3450, opened 2024-04-11
No batches?

SomeHats:
megamind.jpg

`editor.batch` is an alias for `history.batch`. `history.batch` does almost nothing: it runs some code, and if it's the top-level batch, it calls an `onBatchComplete` callback. This was apparently not always the case - the documentation for `editor.batch` claims that it groups a bunch of operations together so that they all happen in one undo/redo operation. This hasn't been true for a very very long time though, even before #3364.

Given that `editor.batch` does almost nothing apart from scheduling a callback, what does it look like to remove? That callback is pretty useful for running some cleanup code whenever _something_ happens - for example, we use it to collect any invalidations of parents with per-record `after` callbacks, then deal with those invalidations once we've collected them all. We can't do that with just record-level callbacks, so we can add an overall store "after something happened" callback to replace it.

The result is some much simpler code (no need to call `editor.batch` ever) at the expense of a callback running slightly more often. The callback doesn't do much though, so I think it's fine! For optimisation reasons we can always wrap things in `editor.store.atomic` to defer callbacks until the end of a large set of operations.

## Breaking change
This is a breaking change - the `editor.batch` function is no more. It's a straight-forward migration path though: just delete it and run the code you were batching directly. The function was doing almost nothing anyway.

### Change Type

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -592,7 +592,6 @@ export class Editor extends EventEmitter<TLEventMap> {
     }): this;
     bail(): this;
     bailToMark(id: string): this;
-    batch(fn: () => void, opts?: TLHistoryBatchOptions): this;
     bringForward(shapes: TLShape[] | TLShapeId[]): this;
     bringToFront(shapes: TLShape[] | TLShapeId[]): this;
     cancel(): this;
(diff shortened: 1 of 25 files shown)


SomeHats:
wait no it also does `transact` and that guy is important ok dw

PR #3453, opened 2024-04-12
Perf: Improve perf of `getCurrentPageShapesSorted`

MitjaBezensek:
This significantly improves performance. Here's a comparison with 2k shapes. Top is the new logic, bottom the old one.

image

One place where this does make a significant difference is when you have a lot of shapes on the page and you start [creating a new arrow](https://github.com/orgs/tldraw/projects/40?pane=issue&itemId=59296136):

Before:
image
image

After:
image
image

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4573,31 +4573,38 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-		// todo: consider making a derivation or something, or merging with rendering shapes
 		const shapes = new Set(this.getCurrentPageShapes().sort(sortByIndex))
-
-		const results: TLShape[] = []
-
-		function pushShapeWithDescendants(shape: TLShape): void {
-			results.push(shape)
-			shapes.delete(shape)
-
-			shapes.forEach((otherShape) => {
```
MitjaBezensek, line 4586: This was the problematic part. For each shape we added we went through all the other shapes to find it's children. Now we have a map of parent child relationships which gets generated on the first pass through the shapes.

steveruizok:
That's an error from the 1 password extension btw

packages/editor/src/lib/editor/Editor.ts
steveruizok: damn, good one!

steveruizok:
did a few little hot path tricks that we've used elsewhere. It's setting around .35ms

PR #3454, opened 2024-04-12
Faster selection / erasing

steveruizok:
This PR makes a small improvement to the way we measure distances. (Often we measure distances multiple times per frame per shape on the screen). In many cases, we compare a minimum distance. This makes those checks faster by avoiding a square root.

### Change Type

### Release Notes

- Improve performance of minimum distance checks.


steveruizok:
I've also made some improvements to the speed of erasing and made margins work for hit segment hit testing, which were previously not working.

steveruizok:
Turns out with lots of shapes on the screen, showing and hiding indicators becomes slow; we can speed things up by leaving the indicators in the DOM and switching their display properties instead.

Issue #3273, opened 2024-03-26
[Bug]: Error when using the text shape after reaching 2000 shapes.


steveruizok:
### What happened?

After creating 2000 shapes, creating a text shape causes a crash (in editing state without an editing shape).

### How can we reproduce the bug?

1. Create 2000 shapes.
2. Create a text shape.

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mimecuvalo:
this looks fixed

PR #3458, opened 2024-04-12
Avoid the error when creating shape after we have reached max shapes.

MitjaBezensek:
Even though we had a check in `createShapes` we don't return anything useful from it, so the `StateNotes` didn't know something went wrong. Not sure if I like this though 🤷‍♂️ 

Fixes [#3273](https://github.com/tldraw/tldraw/issues/3273)

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Add 2000 shapes.
2. Try to add another shape. It should not bomb out. This happened with many shapes: notes, text, geo, line, frame,...

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Prevent an error when trying to add new shapes after we have already reached max shapes per page.

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -77,6 +77,9 @@ import { useValue } from '@tldraw/state';
 import { VecModel } from '@tldraw/tlschema';
 import { whyAmIRunning } from '@tldraw/state';
 
+// @public (undocumented)
+export function alertMaxShapes(editor: Editor, pageId?: TLPageId): void;
+
 // @public
 export function angleDistance(fromAngle: number, toAngle: number, direction: number): number;
 
@@ -830,6 +833,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     // (undocumented)
     isShapeOrAncestorLocked(id?: TLShapeId): boolean;
     mark(markId?: string, onUndo?: boolean, onRedo?: boolean): this;
+    maxShapesReached(): boolean;
     moveShapesToPage(shapes: TLShape[] | TLShapeId[], pageId: TLPageId): this;
     nudgeShapes(shapes: TLShape[] | TLShapeId[], offset: VecLike, historyOptions?: TLCommandHistoryOptions): this;
     packShapes(shapes: TLShape[] | TLShapeId[], gap: number): this;
--- a/packages/editor/api/api.json
+++ b/packages/editor/api/api.json
@@ -174,6 +174,69 @@
       "members": [
         {
           "kind": "Function",
+          "canonicalReference": "@tldraw/editor!alertMaxShapes:function(1)",
+          "docComment": "/**\n * @public\n */\n",
+          "excerptTokens": [
+            {
+              "kind": "Content",
+              "text": "export declare function alertMaxShapes(editor: "
+            },
+            {
+              "kind": "Reference",
+              "text": "Editor",
+              "canonicalReference": "@tldraw/editor!Editor:class"
+            },
+            {
+              "kind": "Content",
+              "text": ", pageId?: "
+            },
+            {
+              "kind": "Reference",
+              "text": "TLPageId",
+              "canonicalReference": "@tldraw/tlschema!TLPageId:type"
+            },
+            {
+              "kind": "Content",
+              "text": "): "
+            },
+            {
+              "kind": "Content",
+              "text": "void"
+            },
+            {
+              "kind": "Content",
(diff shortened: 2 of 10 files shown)


steveruizok:
We could check whether the shape was created, and if not bail back to select.idle? I'm trying to think of a way to "solve this in one place" or make it easier for other states in the future but I can't think of any.

MitjaBezensek:
> We could check whether the shape was created, and if not bail back to select.idle? I'm trying to think of a way to "solve this in one place" or make it easier for other states in the future but I can't think of any.

Yeah, it's quite tricky. 

I'd probably want an api that returns then created shape, then I'd decided the next steps based on that. But we seem to be going for the chaining approach of always returning `this` from the editor, which makes this harder.

It's also interesting how all these calls expected the creation to succeed. Maybe we should make it more obvious that `createShapes` can fail.

mimecuvalo:
> We could check whether the shape was created, and if not bail back to select.idle? I'm trying to think of a way to "solve this in one place" or make it easier for other states in the future but I can't think of any.

This makes me think of @SomeHats 's PR https://github.com/tldraw/tldraw/pull/3364 and whether there would be something listening more centrally and either a) cause an undo after the fact (if a constraint was violated like max shapes) or b) before something is committed to the history/store it checks constraints and prevents the commit from happening. Specifically, I'm wondering whether `_updateShapes` would handle these contraints violations.

MitjaBezensek:
> > We could check whether the shape was created, and if not bail back to select.idle? I'm trying to think of a way to "solve this in one place" or make it easier for other states in the future but I can't think of any.
> 
> This makes me think of @SomeHats 's PR #3364 and whether there would be something listening more centrally and either a) cause an undo after the fact (if a constraint was violated like max shapes) or b) before something is committed to the history/store it checks constraints and prevents the commit from happening. Specifically, I'm wondering whether `_updateShapes` would handle these contraints violations.

We [do handle this case](https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/editor/Editor.ts#L6486-L6490) in `_createShapes`. The problem is that the callers of `createShapes` does not check if the shape was successfully created before trying to interact with it.

mimecuvalo:
> We do handle this case in _createShapes.

I see, I see. With that in mind, it feels like this might not be the right fix (you say so yourself in the PR description that you don't like the fix :P) 

Like, when I look in Editor.ts at `duplicateShapes` or `putContentOntoCurrentPage` (for pasting) we handle the max shapes case and the `createShapes` call happens within a `history.batch`. I would expect `createShapes` to either throw or return false if it hadn't succeeded, for whatever reason.

I think the thing that's weird about this is that ends up being a leaky abstraction and we're forcing callers of `createShapes` to now what's a safe call of `createShapes`. But, say, we could have some weird constraint of "no circles are allowed next to triangles" and that would violate some other constraint.

So to reiterate my previous point or to rephrase it: what I mean to say is that it feels like in the `batch` function (effectively the undo/redo stack), it would handle a failure or constraint violation (max shapes, or otherwise), and alert + fail gracefully.

steveruizok:
Maybe the undo/redo changes from #3364 could matter here, i.e. if there's an error we could bail back to the state when the batch etc started?

The place where I'm most concerned is, for example, pasting shapes with arrows and having that hit the shape limit, and as a result have some of the arrows end up pointing to shapes that weren't created.

MitjaBezensek:
There are two things at play here. 
1. Preventing the editor reaching an invalid state. Think that part is handled well - at least in this case.
2. Communicating that certain operations succeed / fail. This part is not handled (at all?) well.

### Communicating operation results
In most case we just swallow any issues that might have occurred. Check `moveShapesToPage` as an example. If you try to move shapes to a page that doesn't exist we just do an early return. As a consumer I might expect everything went fine and then delete my current page. I'd also delete all the shapes with it, since my previous operation was a noop.

Sometimes we at least write to console, like in the `setCurrentPage` so developers might catch it while developing, but that might still cause issues in production.

I think we should do one of two things:
- API returns statuses. So the `createShapes` returns some sort of status whether it succeeded or not. Can be the ids of created shapes or just a bool flag.
- API throws when it fails.

And we should do this for the whole API, so that this is the expected behaviour and developers can depend on it.

MitjaBezensek:
Closing this for now. Makes sense to spend some time exploring if we can better handle errors in our API calls.

PR #3462, opened 2024-04-14
Cancel pointer velocity while pinching

steveruizok:
There was a bug that could occur if you pinched while using the hand tool, where on pinch end the hand tool would slide the camera based on the pinching velocity. The fix is to cancel out any velocity while pinching.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

On mobile...

1. Select the hand tool.
2. Begin a pinch
3. Stop the pinch
4. The camera should stay where it is

### Release Notes

- Fixed a bug that could occur while pinching with the hand tool selected.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -8168,11 +8168,16 @@ export class Editor extends EventEmitter<TLEventMap> {
 	private _updateInputsFromEvent(
 		info: TLPointerEventInfo | TLPinchEventInfo | TLWheelEventInfo
 	): void {
-		const { previousScreenPoint, previousPagePoint, currentScreenPoint, currentPagePoint } =
-			this.inputs
+		const {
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8190,24 +8195,25 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		this.inputs.isPen = info.type === 'pointer' && info.isPen
 
-		// Reset velocity on pointer down
-		if (info.name === 'pointer_down') {
-			this.inputs.pointerVelocity.set(0, 0)
+		// Reset velocity on pointer down, or when a pinch starts or ends
+		if (info.name === 'pointer_down' || this.inputs.isPinching) {
+			pointerVelocity.set(0, 0)
```
steveruizok, line 8200: This is the meat, the rest is potatoes

PR #3464, opened 2024-04-14
Perf: minor drawing speedup

steveruizok:
Tiny changes as I walk through freehand code. These would only really make a difference on pages with many freehand shapes.

### Change Type

### Release Notes

- Improve performance of draw shapes.


packages/tldraw/src/lib/shapes/shared/freehand/svgInk.ts
```diff
@@ -115,13 +115,14 @@ function cleanUpPartition(partition: StrokePoint[]) {
 	if (partition.length > 1) {
 		partition[0] = {
 			...partition[0],
-			vector: Vec.FromAngle(Vec.Angle(partition[1].point, partition[0].point)),
```
steveruizok, line 118: Generally working with angles is slower than working with vectors. I made a few changes like this in this PR, the rest are mostly drivebys.

packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
```diff
@@ -281,7 +282,8 @@ export class Drawing extends StateNode {
 		this.initialShape = this.editor.getShape<DrawableShape>(id)
 	}
 
-	private updateShapes() {
+	@measureAverageDuration
```
steveruizok, line 285:
```suggestion
```

steveruizok:
Took some snapshots on main and moved them here, seems to produce identical results.

PR #3249, opened 2024-03-24
Stickies: release candidate

steveruizok:
This PR is the target for the stickies PRs that are moving forward. It should collect changes.

- [x] New icon
- [x] Improved shadows
- [x] Shadow LOD
- [x] New colors / theme options
- [x] Shrink text size to avoid word breaks on the x axis
- [x] Hide indicator whilst typing (reverted)
- [x] Adjacent note positions
  - [x] buttons / clone handles
  - [x] position helpers for creating / translating (pits)
  - [x] keyboard shortcuts: (Tab, Shift+tab (RTL aware), Cmd-Enter, Shift+Cmd+enter)
  - [x] multiple shape translating 
- [x] Text editing
  - [x] Edit on type (feature flagged)
  - [x] click goes in correct place
- [x] Notes as parents (reverted)
- [x] Update colors
- [x] Update SVG appearance

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [x] `feature` — New feature

### Test Plan

Todo: fold in test plans for child PRs

### Unit tests:

- [ ] Shrink text size to avoid word breaks on the x axis
- [x] Adjacent notes
  - [x] buttons (clone handles)
  - [x] position helpers (pits)
  - [x] keyboard shortcuts: (Tab, Shift+tab (RTL aware), Cmd-Enter, Shift+Cmd+enter)
- [ ] Text editing
  - [ ] Edit on type
  - [ ] click goes in correct place

### Release Notes

- Improves sticky notes (see list)

PR #3469, opened 2024-04-15
textfields: on mobile edit->edit, allow going to empty geo

mimecuvalo:
(this is a PR redo of https://github.com/tldraw/tldraw/pull/3424 which got messed up a bit)

It doesn't quite feel like this is the right fix but it does solve the issue. I was trying to see if `getShapeAtPoint` needed more work but the further I went in that rabbit hole it seemed like I shouldn't touch that code without causing a bunch of disruption at the moment.

Specifically, the code that does `Check labels first` in Editor.ts is a little obscure (lines 4384-4397). It only checks a couple specifics shapes (with certain combinations, i.e. a geo with "none" fill) _and_ it doesn't check `hitLabels` which also maybe feels wrong? I tried unraveling it but there's a lot of code relying on it at the moment to mess with it in the stickies work.
(I was looking at https://github.com/tldraw/tldraw/pull/1910 and https://github.com/tldraw/tldraw/pull/1806 for historical context fwiw)

Before:

https://github.com/tldraw/tldraw/assets/469604/b263a192-2085-4ffb-9e47-6e9c32abe1f9

After:

https://github.com/tldraw/tldraw/assets/469604/5b0b422b-dd5c-4593-9ac5-dec595923ea6

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


mimecuvalo:
actually @steveruizok i just noticed that this commit in the RC while I was away https://github.com/tldraw/tldraw/pull/3249/commits/63c0223e4fb8f8a55b309935c8eafe172d20bb04#diff-084fecc8a9a937bd5f1e7bd40f96f2fcf961c1abae23242a396714a9570a497d messed up this PR.

I added a commit to revert the isCoarsePointer change in EditingShape but lemme know if you had a reason to keep that in. 🤔 I think we should still stay in editing mode, even on mobile.

PR #3471, opened 2024-04-15
[perf] faster signia capture

ds300:
This PR uses an additional ArraySet to make capturing parent relationships faster for computeds with more than a handful of parents. Seems to result in an overall ~20% speedup of the `maybeCaptureParent` function in normal usage.

### Change Type

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Slight performance improvement to reactivity bookkeeping.


steveruizok (approved):
Awesome! this one gets hot on big projects, so an improvement here will save frames.

PR #3472, opened 2024-04-15
Measure action durations and fps for our interactions

MitjaBezensek:
Adds a feature flag `Measure performance` that allows us to:
- Measure the performance of all the actions (it wraps them into `measureCbDuration`). 
- Measure the frame rate of certain interactions like resizing, erasing,....

Example of how it looks like:

CleanShot 2024-04-17 at 18 04 05

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know
=


apps/examples/src/hooks/usePerformance.ts
```diff
@@ -0,0 +1,23 @@
+import { TLUiEventSource, TLUiOverrides, debugFlags, measureCbDuration, useValue } from 'tldraw'
+
+export function usePerformance(): TLUiOverrides {
+	const measurePerformance = useValue(
+		'measurePerformance',
+		() => debugFlags.measurePerformance.get(),
```
mimecuvalo, line 6: memory, context (how mayn shapes are currently present), selecting which ones we want to use
console.debug
make it easier to parse FPS - prbly just floor it

packages/utils/src/lib/FpsTracker.ts
```diff
@@ -0,0 +1,36 @@
...
+		this.startTime = performance.now()
+	}
+
+	stop() {
+		this.started = false
+		if (this.frame !== null) cancelAnimationFrame(this.frame)
+		const duration = (performance.now() - this.startTime) / 1000
+		const fps = duration === 0 ? 0 : this.frames / duration
+		// eslint-disable-next-line no-console
+		console.log(`${this.name} FPS:`, fps)
```
mimecuvalo, line 30: to be a little fancier :)
add bold/color, and put it in `console.debug`

```suggestion
		console.debug('%c${this.name} FPS%c: ${fps}', 'font-weight: bold; color: lightblue', 'font-weight: normal');
```

packages/utils/src/lib/FpsTracker.ts
```diff
@@ -0,0 +1,36 @@
...
+		if (this.frame !== null) cancelAnimationFrame(this.frame)
+		this.frame = requestAnimationFrame(this.recordFrame)
+		this.startTime = performance.now()
+	}
+
+	stop() {
+		this.started = false
+		if (this.frame !== null) cancelAnimationFrame(this.frame)
+		const duration = (performance.now() - this.startTime) / 1000
+		const fps = duration === 0 ? 0 : this.frames / duration
```
mimecuvalo, line 28: let's floor it, hard to visually parse through all the floating point and the precision doesn't really matter

```suggestion
		const fps = duration === 0 ? 0 : Math.floor(this.frames / duration)
```

apps/examples/src/hooks/usePerformance.ts
```diff
@@ -0,0 +1,23 @@
...
+		[debugFlags]
+	)
+	if (!measurePerformance) return {}
+	return {
+		actions(_editor, actions) {
+			Object.keys(actions).forEach((key) => {
+				const action = actions[key]
+				const cb = action.onSelect
+				action.onSelect = (source: TLUiEventSource) => {
+					return measureCbDuration(`Action ${key}`, () => cb(source))
```
mimecuvalo, line 16: in measureCbDuration, let's also make it a little prettier!

```
  console.debug('%c${name} took%c: ${performance.now() - now}ms', 'font-weight: bold; color: orange', 'font-weight: normal');
```

(different color here to make it easier to distinguish)

packages/editor/src/lib/editor/tools/StateNode.ts
```diff
@@ -10,6 +12,17 @@ import {
 } from '../types/event-types'
 
 type TLStateNodeType = 'branch' | 'leaf' | 'root'
+const STATE_NODES_TO_MEASURE = [
```
mimecuvalo, line 15: feels like we might forget to keep this up-to-date when we create new state nodes. are there ones we want to ignore that are noisy? maybe let's just track all, might be easier to maintain. i'd suggest maybe just ignoring `idle`?

packages/editor/src/lib/editor/tools/StateNode.ts
```diff
@@ -159,6 +174,10 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
 
 	// todo: move this logic into transition
 	enter = (info: any, from: string) => {
+		if (debugFlags.measurePerformance.get() && STATE_NODES_TO_MEASURE.includes(this.id)) {
+			this.fpsTracker.start(this.id)
```
mimecuvalo, line 178: doesn't have to be for this PR but for you to think about:
it'd be interesting to consider things like, not just FPS, but say in the future we want to track something else like memory or some other metric like raw time (instead of FPS)
i guess what i'm saying is maybe we have a `perfTracker` (instead of only an `fpsTracker`) that does FPS, but also might do other things as well.

mimecuvalo (approved):
love it! just a couple comments but nothing blocking, per se. pre-approved :)

steveruizok:
Let's wait on this a bit, I'd like to learn more

packages/utils/src/lib/FpsTracker.ts
MitjaBezensek: Nice. Change it to blue, since it was hard to see on light background. But blue might be hard to see on dark backgrounds? 😂 

CleanShot 2024-04-17 at 13 42 46

packages/editor/src/lib/editor/tools/StateNode.ts
MitjaBezensek: There are a bunch that we probably don't want. Pretty much all `Idle`, `Pointing*`, probably others.
CleanShot 2024-04-17 at 13 48 42

We could go with a blacklist approach though. Then we don't miss any new ones and if the new ones are kinda pointless to track they will annoy someone and they will add it to the blacklist?

packages/editor/src/lib/editor/tools/StateNode.ts
MitjaBezensek: Yeah, makes sense. Even with this one it's easy to add duration and frame count to the printout, but we might improve it in the future.

It could even track perf for a random subset of prod users.

MitjaBezensek:
@steveruizok any thoughts on this one?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8318,7 +8328,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		if (info.ctrlKey) {
 			clearInterval(this._ctrlKeyTimeout)
 			this._ctrlKeyTimeout = -1
-			inputs.ctrlKey = true /** @internal */ /** @internal */ /** @internal */
```
MitjaBezensek, line 8321: Not sure what this was 😂

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8552,7 +8564,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 						// If the user is in pen mode, but the pointer is not a pen, stop here.
 						if (!isPen && isPenMode) return
 
-						// If we've started panning, then clear any long press timeout
```
MitjaBezensek, line 8555: We used to have `clearTimeout(this._longPressTimeout)` here in the past, lookse like a stale comment.

PR #3439, opened 2024-04-10
RBush again?

MitjaBezensek:
Adds RBush to handle spatial querying. We use it for:
- Culling. Helps a lot with panning as we don't have to compute the culled shapes from scratch. Instead we just query rbush again. It makes culling quite granular: spatial index updates when shapes change (additions, removals, changes to bounds), visible shapes depends on that, but also updates when the viewport page bound change, culled shapes then depend on that but also change with selections changes. The api stayed the same, which is great since the fuzz tests can stay as they are.
- Brushing 
- Erasing
- Scribble brushing
- Getting shapes at point (for example, when updating the hover id)

This improves performance of all of those operations. I might have missed some places where this might also be useful.

### Erasing before (Test on my old ipad)

https://github.com/tldraw/tldraw/assets/2523721/edb9c004-a44a-4779-b2d0-98617b057314

### Erasing after

https://github.com/tldraw/tldraw/assets/2523721/8f8367fd-fa8e-4963-ba13-720c5f0c2da5

### Creating an arrow before

https://github.com/tldraw/tldraw/assets/2523721/4068f8b7-f7b8-4826-83f2-083b1f3783bc

### After (much better, but still bad)

https://github.com/tldraw/tldraw/assets/2523721/11af6be6-01d8-4740-bf15-896e2dd31dd6

### Change Type

- [ ] `sdk` — Changes the tldraw SDK
- [ ] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [x] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

PR #3502, opened 2024-04-16
Rbush again again

MitjaBezensek:
Not sure if the perf benefits are that big any longer. Since [we merged my pr](https://github.com/tldraw/tldraw/pull/3453) for `getCurrentPageShapesSorted` it seems that we are quite close. Used [my other PR to measure FPS stats](https://github.com/tldraw/tldraw/pull/3472). That said this logic seems to work correctly, whereas the current logic on main also has an issue with not correctly updating shapes in some cases.

Before:
image

After:
image

Continuation of https://github.com/tldraw/tldraw/pull/3439. Solves the issue where we might have missed some bound changes. This happened for:
- Groups. If you moved a group we only updated the group's bounds, but not the bounds of its children.
- Frames. Like above.
- Arrows. If you bound arrows to shapes, then moved to shapes, we only updated the bounds of the moved shapes, but not of the arrows.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test culling.
2. Other interactions: erasing, brushing, scribble brushing, creating arrows.

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Improves performance of many interactions like brush selecting, creating arrows, panning, translating.


packages/editor/src/lib/editor/derivations/spatialIndex.ts
```diff
@@ -0,0 +1,165 @@
...
+						element.minY !== newBounds.minY ||
+						element.maxX !== newBounds.maxX ||
+						element.maxY !== newBounds.maxY
+					) {
+						this.shapesInTree.delete(id)
+						this.rBush.remove(element)
+						this.addElement(id, elementsToAdd, newBounds)
+						isDirty = true
+					}
+				})
```
MitjaBezensek, line 107: This is the main change. Instead of using the changes we go through all the existing shapes. 

We remove the `removed` first, so that we have fewer shapes to iterate through, and we only add `added` after this so we don't check those twice.

MitjaBezensek: Seems to be much faster to do it once per diff, even if there's only 1 change in diff 🤔

mimecuvalo:
Wow, yeah those FPS measurements are quite close now. It is worth re-evaluating then whether the added complexity is still needed. 🤔

MitjaBezensek:
Closing for now. We can pick it up at some later point if we feel we still need it. If we had a better way of tracking dependencies for our signals it would still be an improvement for things like erasing, brushing,...

mimecuvalo:
This is super sick tech, I'd be surprised if we didn't need this in the future!

PR #3516, opened 2024-04-17
[signia] Smart dirty checking of active computeds

ds300:
This is a huge perf win, and it came to me while procrastinating on making dinner.

The idea is that we can skip checking the parents of a computed value if

- it is being dereferenced during a reaction cycle
- the computed value was not traversed during the current reaction cycle

This more than doubles the speed of the webgl minimap render on my machine (from 2ms down to like 0.8ms).

This will make the biggest difference for anything that derives a value from a large collection of other computed values where typically only a small amount of them change at one time (e.g. iterating over all the shape page bounds to compile an RBush)

Most code paths where we see a big chunk of `haveParentsChanged` in flame graphs should be much faster after this.

### Change Type


steveruizok:
:o

MitjaBezensek:
Just did a quick check on the rbush branch, the perf gains are massive.

Before:
CleanShot 2024-04-17 at 21 46 49@2x

After:
image

MitjaBezensek:
And this if for the current culled calculation:

Before:
image

After: 
image

MitjaBezensek (approved):
👏 🎉

PR #3510, opened 2024-04-17
WebGL Minimap

ds300:
This PR replaces our current minimap implementation with one that uses WebGL

### Change Type

- [x] `sdk` — Changes the tldraw SDK
- [ ] `dotcom` — Changes the tldraw.com web app
- [ ] `docs` — Changes to the documentation, examples, or templates.
- [ ] `vs code` — Changes to the vscode plugin
- [ ] `internal` — Does not affect user-facing stuff

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [x] `improvement` — Improving existing features
- [ ] `chore` — Updating dependencies, other boring stuff
- [ ] `galaxy brain` — Architectural changes
- [ ] `tests` — Changes to any test code
- [ ] `tools` — Changes to infrastructure, CI, internal scripts, debugging tools, etc.
- [ ] `dunno` — I don't know

### Test Plan

1. Add a step-by-step description of how to test your PR here.
2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Add a brief release note for your PR here.

PR #3528, opened 2024-04-18
Webgl minimap (rects)

steveruizok:
This PR adapts #3510 to work with regular squares rather than shape geometry.

Cons:
- doesn't use the shape's actual geometry

Pros:
- fast as hell (2000 shapes in half a ms)
- all the benefits of David's work
- none of the complexity of using actual geometry


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8219,7 +8219,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 		// it will be 0,0 when its actual screen position is equal
 		// to screenBounds.point. This is confusing!
 		currentScreenPoint.set(sx, sy)
-		currentPagePoint.set(sx / cz - cx, sy / cz - cy, sz)
+		const nx = sx / cz - cx
+		const ny = sy / cz - cy
+		if (isFinite(nx) && isFinite(ny)) {
```
steveruizok, line 8224: This was that infinity bug that I ran into, and ran into again here when cx/cy is zero

steveruizok:
This seems to be very slightly slower than regular canvas minimap on higher shape counts (tested on fast and slow hardware).

mimecuvalo:
feels like the webgl win here is more about the UX improvement than any speed improvement, yeah? i think rects-only could be interesting but doesn't get us much for the added complexity?

steveruizok:
If we're going to be doing more with webgl in the minimap, I feel like this PR would be something we could land now and follow up with David's changes. It's microscopically slower than what we have now but barely registers anyway.

packages/tldraw/src/lib/ui/components/Minimap/triangulateGeometry.ts
```diff
@@ -1,59 +0,0 @@
-import { Geometry2d, Mat } from '@tldraw/editor'
-import earcut from 'earcut'
-// @ts-expect-error
-import extrudePolyline from 'extrude-polyline'
```
ds300, line 4: you should also remove these deps

steveruizok:
Ok, I'm going to merge into main then, and have yours as a follower. It'll have a smaller diff against main once this lands.

ds300:
I don't think that's possible? This diff doesn't make sense against main

steveruizok:
bleh, of course. ok

ds300:
just merge it into my PR branch and then merge that

steveruizok:
Sorry suffering from friday afternoon hard rn. Since it's branched off of yours, if I target main then it'll include your changes too, right?

PR #3532, opened 2024-04-19
Improve back to content

steveruizok:
This PR improves the "back to content" behavior using the new culled shapes APIs.

### Change Type

- [x] `improvement` 

### Test Plan

1. Create some shapes, then move the camera to an empty part of the canvas.
2. Check that the back to content button appears.
3. Ensure that the back to content button does not appear when the canvas is empty.


MitjaBezensek (approved):
Nice! 

The PR description is slightly out of date, seems you took another route.

PR #3533, opened 2024-04-19
Webgl minimap squares

ds300:
reopening #3528 

@steveruizok plz start writing actual commit messages again 🙈 this was a bit harder to untangle because they were all 'ok'


packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
```diff
@@ -245,22 +245,20 @@ export class MinimapManager {
...
+				if (selectedShapes.has(shapeId)) {
+					appendVertices(this.gl.selectedShapes, selectedShapeOffset, verticesForShape)
+					selectedShapeOffset += 12
+				} else {
+					appendVertices(this.gl.unselectedShapes, unselectedShapeOffset, verticesForShape)
+					unselectedShapeOffset += 12
+				}
 			}
-		}
+		})
```
ds300, line 261: I'm gonna revert this. The ordering is important for perf.

PR #3534, opened 2024-04-19
iframe protector: track referrer and refactor a bit for maintability

mimecuvalo:
- adds `document.referrer` to tracking to get a sense of who's embedding us
- refactors the hardcoded strings for maintainbility


apps/dotcom/src/components/IFrameProtector.tsx
```diff
@@ -26,8 +26,28 @@ and we should show an annoying messsage.
 If we're not in an iframe, we don't need to do anything.
 */
 
+export enum ROOM_CONTEXT {
```
MitjaBezensek, line 29: We hardly ever use enums in the repo, we seem to prefer unions. Feels kinda weird to see it 😄

MitjaBezensek (approved):

apps/dotcom/src/components/IFrameProtector.tsx
```diff
@@ -26,8 +26,28 @@ and we should show an annoying messsage.
...
+	HISTORY = 'history',
+	LOCAL = 'local',
+}
+
+enum EMBEDDED_STATE {
+	IFRAME_UNKNOWN = 'iframe-unknown',
+	IFRAME_NOT_ALLOWED = 'iframe-not-allowed',
+	NOT_IFRAME = 'not-iframe',
+	IFRAME_OK = 'iframe-ok',
+}
```
steveruizok, line 43:
```suggestion
const EMBEDDED_STATE = {
	IFRAME_UNKNOWN: 'iframe-unknown',
	IFRAME_NOT_ALLOWED: 'iframe-not-allowed',
	NOT_IFRAME: 'not-iframe',
	IFRAME_OK: 'iframe-ok',
} as const
```

Let's use this format for the enums. Enums are kind of a cursed TypeScript feature and so we've settled on this instead.

apps/dotcom/src/components/IFrameProtector.tsx
```diff
@@ -72,24 +86,25 @@ export function IFrameProtector({
...
-		if (embeddedState === 'iframe-unknown') {
+		if (embeddedState === EMBEDDED_STATE.IFRAME_UNKNOWN) {
 			// We iframe embeddings on multiplayer or readonly
 			if (WHITELIST_CONTEXT.includes(context)) {
 				window.parent.postMessage(EXPECTED_QUESTION, '*') // todo: send to a specific origin?
 				timeout = setTimeout(() => {
-					setEmbeddedState('iframe-not-allowed')
-					trackAnalyticsEvent('connect_to_room_in_iframe', { slug, context })
+					setEmbeddedState(EMBEDDED_STATE.IFRAME_NOT_ALLOWED)
+					const referrer = document.referrer
```
SomeHats, line 102: `ancestorOrigins` is a more reliable way to checking who's embedding us, referrer might be different if e.g. tldraw is linked to within an iframe: https://developer.mozilla.org/en-US/docs/Web/API/Location/ancestorOrigins

MitjaBezensek: Unfortunately [not supported in Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Location/ancestorOrigins#browser_compatibility). We could fall back to referrer just for Firefox though.

apps/dotcom/src/components/IFrameProtector.tsx
mimecuvalo: yeah, unions are super hard to grep for in the codebase. especially with generic words like, say, `'room'` this makes it hard to track down where things are being used. it's all about refactorability!

apps/dotcom/src/components/IFrameProtector.tsx
mimecuvalo: ok, added this in (with fallback)

PR #3551, opened 2024-04-22
sync: don't auto-create missing rooms; instead show a 'not found' error

mimecuvalo:
This PR removes the auto-creation of new rooms via just hitting a URL endpoint on dotcom. We now serve up (effectively) a 404. The tricky bit of this is that the 404 status is passed through the websocket. As such, there's quite a bit of plumbing to be able to bubble up a not-found event from the server→ client.

I'll copy the comment that I put in the code on how it passes through the code:
1. From here (`getInitialRoomState`), we send back a `room_not_found` to `TLDrawDurableObject`.
2. In `TLDrawDurableObject`, we accept and
  and then immediately close the client. This lets us send a `TLCloseEventCode.NOT_FOUND`
  closeCode down to the client.
3. `joinExistingRoom` which handles the websocket upgrade is not affected.
  Again, we accept the connection, it's just that we immediately close right after.
4. In `ClientWebSocketAdapter`, `ws.onclose` is called, and that calls `_handleDisconnect`.
5. `_handleDisconnect` sets the status to 'error' and calls the `onStatusChange` callback.
6. On dotcom in `useRemoteSyncClient`, we have `socket.onStatusChange` callback
  where we set `TLIncompatibilityReason.RoomNotFound` and close the client + socket.
7. Finally on the dotcom app we use `StoreErrorScreen` to display an appropriate msg.

Phew!

Screenshot 2024-04-23 at 14 13 17

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test `/r/random-slug` doesn't create a new room, nor does `/r`
2. Make sure `/r/existing-slug` still works.

### Release Notes

- Rooms: `/r/:roomId` no longer creates a new room. Instead, if the room doesn't exist we bubble up a `room_not_found` error.


packages/tlsync/src/lib/TLServer.ts
```diff
@@ -78,14 +79,24 @@ export abstract class TLServer {
...
+				// 3.) joinExistingRoom which handles the websocket upgrade is not affected.
+				//   Again, we accept the connection, it's just that we immediately close right after.
+				// 4.) In ClientWebSocketAdapter, ws.onclose is called, and that calls _handleDisconnect.
+				// 5.) _handleDisconnect sets the status to 'error' and calls the onStatusChange callback.
+				// 6.) On the dotcom app in useRemoteSyncClient, we have socket.onStatusChange callback
+				//   where we set TLIncompatibilityReason.RoomNotFound and close the client + socket.
+				// 7.) Finally on the dotcom app we use StoreErrorScreen to display an appropriate msg.
+				//
+				// Phew!
+				throw new Error('room_not_found')
```
MitjaBezensek: Would probably just return something here instead of throwing. It's expected that this will happen and not something out of the ordinary.

apps/dotcom/src/components/StoreErrorScreen.tsx
```diff
@@ -26,6 +26,11 @@ export function StoreErrorScreen({ error }: { error: Error }) {
 					'Your changes were rejected by the server. Please reload the page. If the problem persists contact the system administrator.'
 				break
 			}
+			case TLIncompatibilityReason.RoomNotFound: {
```
MitjaBezensek, line 29: Should we maybe reuse our page `ErrorPage` component for this case? Feels more related to page not found (like [this one](https://www.tldraw.com/asdfasdf)) than to a store error?

packages/tlsync/src/lib/TLServer.ts
MitjaBezensek: Seems you already updated this, resolving 😄

packages/tlsync/src/lib/TLSyncClient.ts
```diff
@@ -24,6 +24,17 @@ import './requestAnimationFrame.polyfill'
...
 
+/**
+ * These are our private codes to be sent from server->client.
+ * They are in the private range of the websocket code range.
+ * See: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code
+ *
+ * @public
+ */
+export enum TLCloseEventCode {
+	NOT_FOUND = 4004,
```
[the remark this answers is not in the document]
mimecuvalo: Yeah, it's confusing maybe :P 
We have to use the private range of the code for the CloseEvent, I say this in the comment above.

~~I'll make a better comment to say how it's intentionally _not_404.~~

Actually, I'll just change this to be 4099 to be less confusing.

apps/dotcom/src/components/StoreErrorScreen.tsx
mimecuvalo: I wouldn't disagree - it's just that the `<ErrorPage>` isn't really wire up in the path here. 

It's really in `MultiplayerEditor` (which has a reference to StoreErrorScreen and the store is what gets the error). I think this is ok actually. One could see our client, say, having a list of bookmarks of rooms to connect to and maybe clicking on some don't work but some do. So I think the _client_ doesn't 404, it's more that the _room_ has the 404. In theory, the client could keep going if it wanted to.

MitjaBezensek: > I wouldn't disagree - it's just that the `<ErrorPage>` isn't really wired up in the path here.
>
> It's really in `MultiplayerEditor` (which has a reference to StoreErrorScreen and the store is what gets the error).

Yeah, we could handle that case in `MultiplayerEditor` 🤷‍♂️ I guess I'm still thinking of this more in the traditional rest routes `/r/id` points to a certain room, but that room does not exist so I'd expect a 404 error. Still ok with this is you wish to keep it as is.

apps/dotcom/src/components/StoreErrorScreen.tsx
```diff
@@ -26,6 +26,11 @@ export function StoreErrorScreen({ error }: { error: Error }) {
...
+				message =
+					'The room you are trying to connect to does not exist. Please check the URL and try again. If the problem persists contact the system administrator.'
```
SomeHats, line 31: Let's have this message just read "The room you are trying to connect to does not exist.":

- "Please check the URL and try again." probably isn't actionable for most users - the URL is probably a random ID
- "If the problem persists contact the system administrator." is definitely not actionable imo

I'd also maybe suggest adding a "Back to tldraw" link. Maybe we can say if we're in an iframe, we make it read "Open tldraw" with `target="_blank"`

SomeHats (approved):
Looks good to me! Just some copy stuff I'd change

apps/dotcom/src/components/StoreErrorScreen.tsx
mimecuvalo: good call, i ended up consolidating this with ErrorPage so that they have some consistency between them.

Screenshot 2024-04-23 at 14 13 17

PR #3572, opened 2024-04-23
Make coarse pointer check dynamic

Taha-Hassan-Git:
When a device has both a touch screen and a mouse, pointer: coarse evaluates to false and then doesn't change even if the user begins to use the touch screen.

In this PR pointer: coarse is replaced with any-pointer: coarse, which checks if a touchscreen input is available. Event listeners for touchstart and mousemove update isCoursePointer dynamically. So if the user begins to move the mouse it changes to pointer: fine, if they then touch the screen the pointer is returned to coarse.

https://github.com/tldraw/tldraw/assets/98838967/fb86bb44-ec11-4161-bb2f-0e8c3ee83eb6

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Load tldraw on a device with both coarse and fine pointer inputs avaiable (e.g. an ipad with a keyboard and trackpad)
2. Switch between using the mouse and touch screen.
3. Handles on shapes should update dynamically. 

### Release Notes

### DIFF
--- a/packages/editor/src/lib/hooks/useCoarsePointer.ts
+++ b/packages/editor/src/lib/hooks/useCoarsePointer.ts
@@ -20,10 +20,8 @@ export function useCoarsePointer() {
 			const mql = window.matchMedia('(any-pointer: coarse)')
 
 			const handler = (coarse?: boolean) => {
-				editor.updateInstanceState({ isCoarsePointer: !!mql.matches })
-				if (coarse !== undefined) {
-					editor.updateInstanceState({ isCoarsePointer: coarse })
-				}
+				editor.updateInstanceState({ isCoarsePointer: coarse ?? !!mql.matches })```
+
 			}
 
 			const touchStartHandler = () => handler(true)
--- a/packages/editor/src/lib/hooks/useCoarsePointer.ts
+++ b/packages/editor/src/lib/hooks/useCoarsePointer.ts
@@ -20,8 +20,7 @@ export function useCoarsePointer() {
 			const mql = window.matchMedia('(any-pointer: coarse)')
 
 			const handler = (coarse?: boolean) => {
-				editor.updateInstanceState({ isCoarsePointer: coarse ?? !!mql.matches })```
-
+				editor.updateInstanceState({ isCoarsePointer: coarse ?? !!mql.matches })
 			}
 
 			const touchStartHandler = () => handler(true)
--- a/packages/editor/src/lib/hooks/useCoarsePointer.ts
+++ b/packages/editor/src/lib/hooks/useCoarsePointer.ts
@@ -1,9 +1,8 @@
-import { useEffect, useRef } from 'react'
+import { useEffect } from 'react'
 import { useEditor } from './useEditor'
 
 /** @internal */
 export function useCoarsePointer() {
-	const previousCursor = useRef<boolean | undefined>(undefined)
 	const editor = useEditor()
 
 	useEffect(() => {
--- a/packages/editor/src/lib/hooks/useCoarsePointer.ts
+++ b/packages/editor/src/lib/hooks/useCoarsePointer.ts
@@ -1,8 +1,9 @@
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { useEditor } from './useEditor'
 
 /** @internal */
 export function useCoarsePointer() {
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/hooks/useCoarsePointer.ts
```diff
@@ -17,14 +17,28 @@ export function useCoarsePointer() {
 			return
 		}
 		if (window.matchMedia) {
-			const mql = window.matchMedia('(pointer: coarse)')
-			const handler = () => {
+			const mql = window.matchMedia('(any-pointer: coarse)')
```
mimecuvalo, line 20: is `any-pointer: coarse` still a necessary change then? can it just be `pointer: coarse` as before?

packages/editor/src/lib/hooks/useCoarsePointer.ts
```diff
@@ -17,14 +17,28 @@ export function useCoarsePointer() {
...
+
+			const handler = (coarse?: boolean) => {
 				editor.updateInstanceState({ isCoarsePointer: !!mql.matches })
+				if (coarse !== undefined) {
+					editor.updateInstanceState({ isCoarsePointer: coarse })
+				}
```
mimecuvalo, line 26: i think this should go before the above and it should be an if / else type deal.
or, you could go simpler, as in:

```suggestion
				editor.updateInstanceState({ isCoarsePointer: coarse ?? !!mql.matches })```

**Review by mimecuvalo — COMMENTED** [review]
excited about this! thx for digging into it 🙌

File: packages/editor/src/lib/hooks/useCoarsePointer.ts (hunk 1 above)
Comment (Taha-Hassan-Git) [reply packages/editor/src/lib/hooks/useCoarsePointer.ts:20]: I think it is necessary, it's a better to default to the coarse pointer for multiple inputs imo. Imagine using an ipad with keyboard attached, you have no intention of touching the mouse and go to manipulate a shape, but the right handles aren't exposed for you to rotate it on the first touch, it's only on the second touch that it gets exposed for you. If we expose coarse pointer handles by default and you go to use the track pad, the handles are changed as some as the cursor moves. Does that make sense?

Comment (steveruizok) [reply packages/editor/src/lib/hooks/useCoarsePointer.ts:20]: So the logic is:
- if any of the pointers are coarse, default to coarse mode
- if in coarse mode and a non-coarse pointer event occurs, exit coarse mode
- if in not-coarse mode and a coarse pointer event occurs, enter coarse mode

?

File: packages/editor/src/lib/hooks/useCoarsePointer.ts [hunk 3]
```diff
@@ -17,14 +17,25 @@ export function useCoarsePointer() {
 			return
 		}
 		if (window.matchMedia) {
-			const mql = window.matchMedia('(pointer: coarse)')
-			const handler = () => {
-				editor.updateInstanceState({ isCoarsePointer: !!mql.matches })
+			const mql = window.matchMedia('(any-pointer: coarse)')
+
+			const handler = (coarse?: boolean) => {
+				editor.updateInstanceState({ isCoarsePointer: coarse ?? !!mql.matches })
```
steveruizok, line 23: We wouldn't want to update instance state on every pointer move / mouse move. I think we have logic now in our store that prevents us from updating state with its identical previous value, but I'd rather not even run that path.

packages/editor/src/lib/hooks/useCoarsePointer.ts
Taha-Hassan-Git: Yes, exactly

packages/editor/src/lib/hooks/useCoarsePointer.ts
Taha-Hassan-Git: That makes sense, I'll work on that!

packages/editor/src/lib/hooks/useCoarsePointer.ts
mimecuvalo: Gotcha gotcha, thanks for the explanation — so it's to help for that default state, makes sense.

Definitely worth a comment! Can you add that context and/or steve's list as a comment in this file somewhere? It would be helpful for future intrepid explorers spelunking in these parts ;)

PR #3605, opened 2024-04-25
Bring back `/new` for creating new rooms when not inside iframes.

MitjaBezensek:
This brings back the `/new` route for creating new rooms. But just when we are not inside an iframe.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/apps/dotcom/src/pages/new.tsx
+++ b/apps/dotcom/src/pages/new.tsx
@@ -1,0 +1,41 @@
+import { Snapshot } from '@tldraw/dotcom-shared'
+import { schema } from '@tldraw/tlsync'
+import { Navigate } from 'react-router-dom'
+import '../../styles/globals.css'
+import { ErrorPage } from '../components/ErrorPage/ErrorPage'
+import { defineLoader } from '../utils/defineLoader'
+import { isInIframe } from '../utils/iFrame'
+import { getNewRoomResponse } from '../utils/sharing'
+
+const { loader, useData } = defineLoader(async (_args) => {
+	if (isInIframe()) return null
+
+	const res = await getNewRoomResponse({
+		schema: schema.serialize(),
+		snapshot: {},
+	} satisfies Snapshot)
+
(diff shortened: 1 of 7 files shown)


apps/dotcom/src/routes.tsx
```diff
@@ -30,7 +30,7 @@ export const router = createRoutesFromElements(
 		<Route errorElement={<DefaultErrorFallback />}>
 			<Route path="/" lazy={() => import('./pages/root')} />
 			<Route path="/r" element={<Navigate to="/" />} />
```
mimecuvalo, line 32: so `/r` not coming back? should it redirect to `/new` spiritually?

apps/dotcom/src/pages/new.tsx
```diff
@@ -0,0 +1,41 @@
...
+	const data = useData()
+	if (!data)
+		return (
+			<ErrorPage
+				icon
+				messages={{
+					header: 'Page not found',
+					para1: 'The page you are looking does not exist or has been moved.',
+				}}
+				redirectTo="/"
```
mimecuvalo, line 37: we should get rid of `redirectTo` and in ErrorPage - it doesn't do anything 🙃

mimecuvalo (approved):
LGTM — thx!

apps/dotcom/src/routes.tsx
MitjaBezensek: I think we said we just keep one of the and that it should be `/new` 🤷 😄

apps/dotcom/src/pages/new.tsx
MitjaBezensek: Just removed it.

apps/dotcom/src/routes.tsx
SomeHats: I thought we'd said we'd keep both, but might be mistaken

PR #3617, opened 2024-04-25
Fix an issue with the minimap

MitjaBezensek:
In some cases (when you clicked close to the current viewport area) the minimap would not change the camera position. This is because the `clampedPoint` calculation would offset the clicked position slightly and then the `clampedPoint` would land inside the currently viewport.

With this change we will treat every click outside of the viewport (but still within allowed bounds) as a change that requires use to change the camera position.

Before

https://github.com/tldraw/tldraw/assets/2523721/ded202d1-94d7-4b40-ba77-45d283e4fa60

After

https://github.com/tldraw/tldraw/assets/2523721/2ccee77d-ec0e-4997-b51d-c14676657a94

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Use minimap in various ways to make sure it still works correctly. The change targets the `pointerDown` event, but I did also refactor one method that is also used by double click so it makes sense to test that as well. Make sure you are not double clicking since we don't use the same logic there.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixes clicking on the minimap when we clicked just slightly outside of the current viewport.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
+++ b/packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
@@ -168,13 +168,12 @@ export class MinimapManager {
 		clampToBounds = false
 	) => {
 		const { editor } = this
-		const viewportPageBounds = editor.getViewportPageBounds()
+		const vpPageBounds = editor.getViewportPageBounds()
 
 		let { x: px, y: py } = this.getMinimapPagePoint(x, y)
 
 		if (clampToBounds) {
 			const shapesPageBounds = this.editor.getCurrentPageBounds() ?? new Box()
-			const vpPageBounds = viewportPageBounds
 
 			const minX = shapesPageBounds.minX - vpPageBounds.width / 2
 			const maxX = shapesPageBounds.maxX + vpPageBounds.width / 2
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
```diff
@@ -186,22 +186,8 @@ export class MinimapManager {
...
-			} else if (qr) {
-				px -= qr / 2
-			}
-
-			if (qt && qt > qb) {
-				py += qt / 2
-			} else if (qb) {
-				py -= qb / 2
-			}
+			px += (lx - rx) / 2
```
MitjaBezensek, line 189: Refactored this a bit, feels much simpler and should result in the same logic.

packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
```diff
@@ -168,13 +168,12 @@ export class MinimapManager {
...
 	) => {
 		const { editor } = this
-		const viewportPageBounds = editor.getViewportPageBounds()
+		const vpPageBounds = editor.getViewportPageBounds()
 
 		let { x: px, y: py } = this.getMinimapPagePoint(x, y)
 
 		if (clampToBounds) {
 			const shapesPageBounds = this.editor.getCurrentPageBounds() ?? new Box()
-			const vpPageBounds = viewportPageBounds
```
MitjaBezensek, line 177: Just removed this reassignment as it served no purpose.

steveruizok:
Works for me!

steveruizok (approved):

PR #3621, opened 2024-04-26
Fix an issue with minimap.

MitjaBezensek:
Fixes an issue where minimap would sometimes completely bug out. The problem was when you resized the window height. We only had resize observer for the minimap's canvas element, but we were missing the observer for the window / container. This caused an issue with [calculating the minimap page point](https://github.com/tldraw/tldraw/blob/mitja%2Ffix-minimap-when-resizing-window/packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts#L153) via `getMinimapPagePoint`. They y coordinate would get incorrectly calculated.

Before

https://github.com/tldraw/tldraw/assets/2523721/ecf4c860-0aa2-46ac-8b90-964bff7f04b3

After

https://github.com/tldraw/tldraw/assets/2523721/38047877-7657-45cf-89b1-31ac0c47228c

Resolves https://github.com/tldraw/tldraw/issues/3589

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Insert some shapes.
2. Now open dev panel below or change the height of the window.
3. Use the minimap. It should no longer bug out.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixes an issue with the minimap bugging out after you change the window's height.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
+++ b/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
@@ -27,7 +27,7 @@ export function DefaultMinimap() {
 		const minimap = new MinimapManager(editor, rCanvas.current, container)
 		minimapRef.current = minimap
 		return minimapRef.current.close
-	}, [editor])
+	}, [editor, container])
 
 	const onDoubleClick = React.useCallback(
 		(e: React.MouseEvent<HTMLCanvasElement>) => {
--- a/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
+++ b/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
@@ -6,6 +6,7 @@ import {
 	normalizeWheel,
 	releasePointerCapture,
 	setPointerCapture,
+	useContainer,
 	useEditor,
 	useIsDarkMode,
 } from '@tldraw/editor'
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
```diff
@@ -15,17 +16,18 @@ import { MinimapManager } from './MinimapManager'
 /** @public */
 export function DefaultMinimap() {
 	const editor = useEditor()
+	const container = useContainer()
```
mimecuvalo, line 19: for my own knowledge: is it slightly easier/better to just do `editor.getContainer()` or does `useContainer()` provide some extra reactivity that i'm not aware of?

mimecuvalo (approved):
LGTM % small comment

packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
MitjaBezensek: Looks like it's basically the same thing? We avoid one anonymous function call by using the hook.
CleanShot 2024-04-26 at 10 47 03

mimecuvalo: gotcha - i guess it feels like `useContainer` is ever-so-slightly more heavyweight than just calling `getContainer` but, meh

PR #3623, opened 2024-04-26
copy/paste: fix pasting not working from Edit menu

mimecuvalo:
Looks like this has been broken for a while actually. I spelunked into the history of git blame around why we needed this but can't find when it was added easily. (maybe it was in brivate?)

I don't _think_ we need the menu check anymore but lemme know if there's something I'm missing here @steveruizok 

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

- Clipboard: fix pasting from the Edit menu.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -74,13 +74,12 @@ const INPUTS = ['input', 'select', 'textarea']
  * @param editor - The editor instance.
  * @internal
  */
-function disallowClipboardEvents(editor: Editor) {
+function disallowClipboardEvents() {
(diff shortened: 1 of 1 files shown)


steveruizok:
Hard to remember exactly. I believe it's mostly to do with dialogs.

For example on this PR's preview, if you:
1. copy some shapes
2. open the edit link dialog for some shape
3. blur the text input
4. paste with the dialog still open

...then you'll hit a broken state.

I actually think the fix would be to remove the `disallowClipboardEvents` check from `useMenuClipboardEvents` > `handlePaste`, because in that case we _do_ want it to work if the menu is open, because we'll be calling it from the menu. 

Could you try reverting the changes so far and then trying that?

PR #3624, opened 2024-04-26
make route prefixes have a single place where they are defined

mimecuvalo:
This is for maintainabilty of the paths. It's hard to track down all the places where a route is being referenced. This helps unify them so that it's easily searchable in the codebase. This came up during the readonly room refactor and being able to find the way a particular route was wired through the codebase.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/apps/dotcom/src/components/ShareMenu.tsx
+++ b/apps/dotcom/src/components/ShareMenu.tsx
@@ -38,8 +38,8 @@ type ShareState = {
 
 function isSharedReadonlyUrl(pathname: string) {
 	return (
-		pathname.startsWith(`${RoomOpenModeToPath[ROOM_OPEN_MODE.READ_ONLY]}/`) ||
-		pathname.startsWith(`${RoomOpenModeToPath[ROOM_OPEN_MODE.READ_ONLY_LEGACY]}/`)
+		pathname.startsWith(`/${RoomOpenModeToPath[ROOM_OPEN_MODE.READ_ONLY]}/`) ||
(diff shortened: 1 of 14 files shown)


MitjaBezensek:
Getting double slashes for readonly rooms when copying the link from the share menu:
image

We also don't seem to correctly detected that we are in a readonly room and show the wrong options in the share menu
image

We should be showing this:
image

apps/dotcom/src/components/ShareMenu.tsx
```diff
@@ -43,7 +44,7 @@ function isSharedReadonlyUrl(pathname: string) {
 }
 
 function isSharedReadWriteUrl(pathname: string) {
-	return pathname.startsWith('/r/')
+	return pathname.startsWith(`${ROOM_PREFIX}/`)
```
MitjaBezensek, line 47: Kinda undecided whether I like that we include the leading slash. This showcases why. We need to add the trailing slash, but not the leading one. It does save some typing, but makes me always think if I need to include it or not, so it might be slightly confusing. WDYT?

mimecuvalo:
whoops! fixed up the double-slash problem. thx for catching, very silly.

apps/dotcom/src/components/ShareMenu.tsx
mimecuvalo: I'm down to get rid of the slash - yeah, I see your point on this 👍 
Personally was torn as well so it's a fair point!

steveruizok (approved):

MitjaBezensek (approved):
Nice! So good to have this consistently used on BE and FE 👏

PR #3632, opened 2024-04-28
Fix text resizing with alt key

steveruizok:
This PR fixes a bug where alt-dragging the left or right handles of a text shape would not produce the correct outcome: the width would double but the center would change.

Kapture 2024-04-28 at 13 48 52

This is because the text shape is aspect ratio locked only when dragging handles other than the left or right, but we didn't have the ability to differentiate between that. We've had to add that optionality in, together with a hard-coded override of the normal behavior for text shapes.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

1. Resize text.
2. Resize text with the alt key held.

- [x] Unit Tests

### Release Notes

- Fixed a bug with resizing text shapes from the left and right while holding alt.

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -2505,6 +2505,7 @@ export type TLResizeShapeOptions = Partial<{
     initialBounds: Box;
     initialPageTransform: MatLike;
     initialShape: TLShape;
+    isAspectRatioLocked: boolean;
     mode: TLResizeMode;
     scaleAxisRotation: number;
     scaleOrigin: VecLike;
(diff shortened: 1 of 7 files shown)


steveruizok:
As a sidebar: resizing is very complicated right now. In addition to the regular math, we currently resize from the resizing interaction AND the editor's own interactions (like stretch and flip), and this adds a bunch of brainwork when working on resizing. It's also probably hiding some opportunities for simplification. This is improved somewhat by the #3092 PR, where any resizing means creating a new resize interaction, updating it, and completing it.

PR #3639, opened 2024-04-29
Improve pressure-detection logic in drawing

steveruizok:
This PR improves handling of pressure in the draw tool. While we differentiate between pen and not-pen in the pointer type, there are not-pen type inputs like styluses that present as "mouse" inputs. This PR makes the freehand options for these inputs more pen-like so that they benefit from the dialing-in of options for pressure-using devices.

### Change Type

### Test Plan

1. Use pen inputs
2. Use mouse inputs
3. Use stylus inputs

- [x] Unit Tests

### Release Notes

- Improves handling of mouse-type devices that support pressure, e.g. wacom tablets. They now use the same freehand options as true pen-type inputs.


packages/tldraw/src/lib/shapes/shared/polygon-helpers.ts
steveruizok, line 1: drive-by, remove duplicated code

Issue #3661, opened 2024-04-30
We fetch the readonly slug two times


SomeHats:
Refreshing a multiplayer room

![Image](https://uploads.linear.app/8ba8cf67-bb46-4faa-8a97-5eb61e0814c1/777af440-b583-4520-8194-3967f2858aa6/b4c62a11-477d-4971-b7da-653b33c5b62d?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzhiYThjZjY3LWJiNDYtNGZhYS04YTk3LTVlYjYxZTA4MTRjMS83NzdhZjQ0MC1iNTgzLTQ1MjAtODE5NC0zOTY3ZjI4NThhYTYvYjRjNjJhMTEtNDc3ZC00OTcxLWI3ZGEtNjUzYjMzYzViNjJkIiwiaWF0IjoxNzE0NDc2MjI0LCJleHAiOjMzMjg1MDM2MjI0fQ.FrcI9plMtdhr-jDhOdLOi3NRAQuoK-KI2HKmiqRbPDA)

Sharing a room is even worse

![Image](https://uploads.linear.app/8ba8cf67-bb46-4faa-8a97-5eb61e0814c1/c7934998-53c6-41b9-9275-aff8a0fbfa66/084380a8-e2f0-4e51-b480-d1f78c545f5b?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzhiYThjZjY3LWJiNDYtNGZhYS04YTk3LTVlYjYxZTA4MTRjMS9jNzkzNDk5OC01M2M2LTQxYjktOTI3NS1hZmY4YTBmYmZhNjYvMDg0MzgwYTgtZTJmMC00ZTUxLWI0ODAtZDFmNzhjNTQ1ZjViIiwiaWF0IjoxNzE0NDc2MjI0LCJleHAiOjMzMjg1MDM2MjI0fQ.Ibr8iZzfhYYNES5ItxcUzFYayOAIeJrR68DeKgLTbLQ)

PR #3663, opened 2024-04-30
Prevent unnecessary fetching of readonly slugs

MitjaBezensek:
Prevents unnecessary fetching of readonly slugs. We only need to fetch it if we don't have it yet. 

There was also a weird issue with `window.location.href` sometimes returning encoded search params and sometimes decoded ones:

image

This then caused an additional fetch in the `setInterval` since the [urls did not match](https://github.com/tldraw/tldraw/blob/main/apps/dotcom/src/components/ShareMenu.tsx#L140).

CleanShot 2024-04-30 at 14 37 12

Resolves https://github.com/tldraw/tldraw/issues/3661

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

Opening a multiplayer room

1. Open a multiplayer room. You should only see one fetch for the readonly slug.

Sharing a local room

1. Open a local room.
2. Share it. You should only see one fetch for the readonly slug.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Prevent unnecessary fetching of readonly slugs.

### DIFF
--- a/apps/dotcom/src/components/ShareMenu.tsx
+++ b/apps/dotcom/src/components/ShareMenu.tsx
@@ -135,6 +135,12 @@ export const ShareMenu = React.memo(function ShareMenu() {
 					})
 				}
 			})
+		} else if (!shareState.readonlyQrCodeDataUrl) {
+			createQRCodeImageDataString(shareState.readonlyUrl).then((dataUrl) => {
+				if (!cancelled) {
+					setShareState((s) => ({ ...s, readonlyQrCodeDataUrl: dataUrl }))
+				}
+			})
 		}
 
 		const interval = setInterval(() => {
--- a/apps/dotcom/src/components/ShareMenu.tsx
+++ b/apps/dotcom/src/components/ShareMenu.tsx
@@ -124,20 +124,22 @@ export const ShareMenu = React.memo(function ShareMenu() {
 			})
 		}
 
-		getReadonlyUrl().then((readonlyUrl) => {
-			if (readonlyUrl && !shareState.readonlyQrCodeDataUrl) {
-				// fetch the readonly QR code data URL
-				createQRCodeImageDataString(readonlyUrl).then((dataUrl) => {
-					if (!cancelled) {
-						setShareState((s) => ({ ...s, readonlyUrl, readonlyQrCodeDataUrl: dataUrl }))
-					}
-				})
-			}
-		})
+		if (!shareState.readonlyUrl) {
+			getReadonlyUrl().then((readonlyUrl) => {
+				if (readonlyUrl) {
+					// fetch the readonly QR code data URL
+					createQRCodeImageDataString(readonlyUrl).then((dataUrl) => {
+						if (!cancelled) {
+							setShareState((s) => ({ ...s, readonlyUrl, readonlyQrCodeDataUrl: dataUrl }))
+						}
+					})
+				}
+			})
+		}
 
 		const interval = setInterval(() => {
 			const url = window.location.href
-			if (shareState.url === url) return
+			if (decodeURIComponent(url) === decodeURIComponent(shareState.url)) return
 			setShareState(getFreshShareState())
 		}, 300)
 
--- a/apps/dotcom/src/components/ShareMenu.tsx
+++ b/apps/dotcom/src/components/ShareMenu.tsx
@@ -19,6 +19,7 @@ import {
 import { useShareMenuIsOpen } from '../hooks/useShareMenuOpen'
 import { createQRCodeImageDataString } from '../utils/qrcode'
 import { SHARE_PROJECT_ACTION, SHARE_SNAPSHOT_ACTION } from '../utils/sharing'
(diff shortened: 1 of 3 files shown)


apps/dotcom/src/components/ShareMenu.tsx
```diff
@@ -124,20 +124,22 @@ export const ShareMenu = React.memo(function ShareMenu() {
...
+						}
+					})
+				}
+			})
+		}
 
 		const interval = setInterval(() => {
 			const url = window.location.href
-			if (shareState.url === url) return
+			if (decodeURIComponent(url) === decodeURIComponent(shareState.url)) return
```
mimecuvalo, line 142: digging a little more into this - looks like we should fix more of an underlying problem here.

in sharing.ts, line 141: we should do:
```navigate(decodeURIComponent(pathname))```

the problem is that `new URLSearchParams` a couple lines before that encodes things unnecessarily (for us) and then when we navigate it's actually doing two navigations, to, for example:
1.) http://localhost:3000/r/3uc5l9JkT2dob3Q8DFr43?v=0%2C0%2C1460%2C679&p=page
2.) http://localhost:3000/r/3uc5l9JkT2dob3Q8DFr43?v=0,0,1460,679&p=page

you can see this if you put some breakpoints in as it navigates.

so anyway, instead of adding `decodeURIComponent` let's fix up sharing.ts!

**Review by mimecuvalo — CHANGES_REQUESTED** [review]

File: apps/dotcom/src/components/ShareMenu.tsx (hunk 1 above)
Comment (MitjaBezensek) [reply apps/dotcom/src/components/ShareMenu.tsx:142]: What if we do both? I can see us making the same kind of mistake again and it could take us a while to discover. Keeping the decoding where we do the check could prevent such future problems.

Comment (mimecuvalo) [reply apps/dotcom/src/components/ShareMenu.tsx:148]: My hesitation is that it really should be `decodeURI` (since it's looking at the whole URI and not just the pathname).
And then that actuall ywouldn't have fixed this issue :-/

File: apps/dotcom/src/utils/sharing.ts [hunk 2]
```diff
@@ -70,7 +70,9 @@ async function getSnapshotLink(
 		return ''
 	}
 	const paramsToUse = getViewportUrlQuery(editor)
-	const params = paramsToUse ? `?${new URLSearchParams(paramsToUse).toString()}` : ''
+	const params = paramsToUse
```
mimecuvalo, line 73: btw, we should add a comment (here and below) - this works but it's all a bit cursed. i think we're accommodating react-router to the fact that it doesn't encode commas, specifically.

_technically_, the commas _should_ be encoded in the URL so, because we're doing the decode here we're trying to undo what URLSearchParams is meant to do correctly, ugh. 🤦

MitjaBezensek: Yeah. I did a quick search and it seems react router has a bunch of encoding issues.

apps/dotcom/src/components/ShareMenu.tsx
MitjaBezensek: Added a helper function to  ccompare urls. Might come handy in the future. Can also drop the whole check, but it still feels worth it to me.

apps/dotcom/src/utils/url.ts
```diff
@@ -1,3 +1,26 @@
...
 	window.open(url, '_blank')
 }
+
+/**
+ * Compares two URLs to see if they are the same. This ignores encoding.
+ * @param first - The first URL to compare.
+ * @param second - The second URL to compare.
+ * @returns Whether the URLs are the same.
+ */
+export function sameUrls(first: string, second: string): boolean {
```
mimecuvalo, line 11: ach, I'm gonna be pedantic on this a bit :)

we shouldn't do this, because `decodeURIComponent` is meant to operate only on individual part of a search param.

so, this:
`decodeURIComponent(`?foo=a%3D1%26b%3D2%26c%3D3`)`
shouldn't actually be equate to:
`decodeURIComponent(`?foo=a=1&b=2&c=3`)`
because they're actually two different things.

The first is saying `foo` is equal to `a=1&b=2&c=3` (encoded)
and the second is saying `foo` = `a` and `b` = `2` and `c` = `3` (and yeah it's prbly a bit of a parser error there but it's a contrived example)

This is because you can have encoded things like this in URLs (security stuff aside here for a sec):
`https://example.com/?redirectTo=https%3A%2F%2Ffoo.com%3Fbaz%3Dbar`

Anyway, we shouldn't do this! It'll have subtle footguns in there!

MitjaBezensek: It's all cursed 😂 Removed it.

mimecuvalo: hehe agreed. thx :D

mimecuvalo (approved):
LG - thanks! i know it was a bit of an annoying one 🙃

PR #3673, opened 2024-05-01
Allow clients to gracefully handle rejection

ds300:
This PR fixes the issue where sync clients would get into a reconnect loop after being rejected by the sync server.

- Close the socket when in the error state (see useRemoteSyncClient)
- Show a 'plx refresh the page' screen that doesn't have a sad face on it
- If older clients who can't handle rejection well need to be rejected (e.g. due to a store migration being added) then we send them to a special purgatory where the canvas goes blank and it shows the offline indicator but the websocket connection stays open and it won't try to reconnect.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

1. Gonna manually test this one by doing sneaky deploys to a test PR

### DIFF
--- a/apps/dotcom/src/components/ErrorPage/ErrorPage.tsx
+++ b/apps/dotcom/src/components/ErrorPage/ErrorPage.tsx
@@ -1,28 +1,39 @@
+import { ReactNode } from 'react'
 import { Link } from 'react-router-dom'
 import { isInIframe } from '../../utils/iFrame'
 
-export function ErrorPage({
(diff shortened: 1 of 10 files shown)


ds300:
hm when testing this in preview deploys it didn't show the 'offline' indicator and it didn't clear the canvas so it looked like all was fine and could very easily lead to data loss, so I'm gonna investigate replacing the editor's content with a 'please refresh the page' notice of some kind.

PR #3684, opened 2024-05-02
LOD for images

steveruizok:
This PR adds support for "level of detail" for image shapes. When an image is created, we create multiple versions of that image (each smaller than the last) and display the smallest image for the current zoom level.

Some thoughts here:

- It works! and is quite nice.
- Very hard to notice the switching between sizes, even with big jumps. (1, 1/2, 1/4, etc.)
- This could make image-heavy rooms much more performant.
- This could prevent crashes on image-heavy rooms on iOS / memory-sensitive devices

## For shared rooms / hosted assets
- the performance benefits are definitely worth the extra storage space for us on the backend
- we could bump the maximum size for images on hosted rooms

## For offline rooms / serialized assets
- it's not clear that the performance benefits for offline are worth the extra storage space, because the extra image versions would be stored in the document JSON itself.
- Offline projects should probably continue to use a single downscaled images

## For the API
- it would be nice to make the TLImageAsset.props.src be either a string or an array of scale / strings, so that this isn't a breaking change for anyone.
- In this PR we have to compare the shape's width vs. the asset's width in order to determine how the image has been resized. The logic is a little complicated, maybe there's another way to do it, maybe not

### Change Type

- [x] `feature` — New feature


steveruizok:
Some thoughts here:

- It works! and is quite nice.
- Very hard to notice the switching between sizes, even with big jumps. (1, 1/2, 1/4, etc.)
- This could make image-heavy rooms much more performant.
- This could prevent crashes on image-heavy rooms on iOS / memory-sensitive devices

### For shared rooms / hosted assets
- the performance benefits are definitely worth the extra storage space for us on the backend
- we could bump the maximum size for images on hosted rooms

### For offline rooms / serialized assets
- it's not clear that the performance benefits for offline are worth the extra storage space, because the extra image versions would be stored in the document JSON itself.
- Offline projects should probably continue to use a single downscaled images

### For the API
- it would be nice to make the `TLImageAsset.props.src` be either a string or an array of scale / strings, so that this isn't a breaking change for anyone.
- In this PR we have to compare the shape's width vs. the asset's width in order to determine how the image has been resized. The logic is a little complicated, maybe there's another way to do it, maybe not

apps/dotcom/src/hooks/useMultiplayerAssets.ts
```diff
@@ -58,7 +43,68 @@ export function useMultiplayerAssets(assetUploaderUrl: string) {
...
+					assetId = AssetRecordType.createId(getHashForString(url))
+					if (file.type === 'image/gif') isAnimated = true
+					size = await MediaHelpers.getImageSize(file)
+				}
+
+				urls.push(url)
+				await fetch(url, {
+					method: 'POST',
+					body: file,
+				})
```
SomeHats, line 89: can we tweak this to perform uploads in parallel?

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -81,19 +82,63 @@ export function registerDefaultExternalContentHandlers(
...
+		if (isImageType) {
+			const sources: TLImageAsset['props']['sources'] = []
+
+			// Always rescale the image
+			if (file.type === 'image/jpeg' || file.type === 'image/png') {
+				sources.push(
+					{
+						scale: 1,
+						src: await FileHelpers.blobToDataUrl(
+							await downsizeImage(file, size.w, size.h, {
```
SomeHats, line 96: ditto re: doing some of this down-sizing in parallel

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -171,11 +180,15 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
 	}
 
 	override async toSvg(shape: TLImageShape) {
-		const asset = shape.props.assetId ? this.editor.getAsset(shape.props.assetId) : null
-
+		const zoomLevel = useValue('zoom level', () => this.editor.getZoomLevel(), [this.editor])
```
SomeHats, line 183: imo svg exports shouldn't depend on the editor's zoom level - we usually just assume zoom = 100% for exports

SomeHats:
When I tried a this on a multiplayer room preview i got an error when adding an image: `At asset(type = image).props.sources: Expected an array, got undefined`

apps/dotcom/src/hooks/useMultiplayerAssets.ts
mimecuvalo: if we did this in parallel it feels like that might be potentially harder on bandwidth, perhaps for certain users?

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
mimecuvalo: hah, i have a similar concern here on parallelization with less powerful machines and large images :)

apps/dotcom/src/hooks/useMultiplayerAssets.ts
SomeHats: fair! tbh for this multiplayer one we could always look at using https://developers.cloudflare.com/images/transform-images/ if we wanted to really optimize this - just upload the original, let the server handle resizing on-demand (plus optimizing for smaller file size in a way that the browser won't out of the box)

mimecuvalo: Hahah, Steve and I were indeed talking about this same thing this morning (albeit with Imgix, not Cloudflare's offering). I'll pull you aside and we can chat more offline on your thoughts on this.

mimecuvalo:
closing this one up! superseded by https://github.com/tldraw/tldraw/pull/3827

PR #3685, opened 2024-05-02
Incremental bindings index

ds300:


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4819,42 +4826,40 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 		shape: TLShape | TLShapeId,
-		type: Binding['type']
+		type?: Binding['type']
 	): Binding[] {
 		const id = typeof shape === 'string' ? shape : shape.id
-		return this.store.query.exec('binding', {
-			fromId: { eq: id },
-			type: { eq: type },
-		}) as Binding[]
+		return this.getBindingsInvolvingShape(id, type).filter((b) => b.fromId === id) as Binding[]
```
SomeHats, line 4846: this seems like it no longer does the `type` check. can you restore?

also, i think i would make type non-optional on this api. as these APIs are directional, i think they're fairly binding-type specific and more general binding queries are better answered by `getBindingsInvolvingShape`. I imagine that these two APIs are mostly going to be used by binding implementors for their own bindings

packages/editor/src/lib/editor/derivations/bindingsIndex.ts
```diff
@@ -0,0 +1,89 @@
+import { Computed, RESET_VALUE, computed, isUninitialized } from '@tldraw/state'
+import { TLBinding, TLShapeId } from '@tldraw/tlschema'
+import { objectMapValues } from '@tldraw/utils'
+import { Editor } from '../Editor'
+
+type TLBindingsIndex = Record<TLShapeId, undefined | TLBinding[]>
```
SomeHats, line 6: happy to call it out of scope of this diff as we use objects as maps _a lot_ in tldraw, but i believe that this pattern (using an object as a map instead of a `Map` or other ds) is a big perf foot-gun in JS as it forces V8/other engines to constantly create & throw out the hidden classes that back the objects. thoughts on using a normal map here?

packages/editor/src/lib/editor/Editor.ts
ds300: the type check happens in the underlying getBindingsInvolvingShape. I'll make it non-optional too, good call.

SomeHats: oh i missed that! thx

packages/editor/src/lib/editor/derivations/bindingsIndex.ts
ds300: i switched to a map. would be fun to do a perf comparison but I'm lazy.

yeah i'd be interested in an immutable map too. Immutable.js would be way too heavy a dependency but that red-black tree implementation looks cool, wasn't aware of it.

SomeHats: yeahhhh, i think there's a version of immutable with just the map? but not sure

packages/editor/src/lib/editor/Editor.ts
ds300: done

SomeHats (approved):
amazing, thanks david!

PR #3693, opened 2024-05-03
Move storing of snapshots to R2

MitjaBezensek:
Instead of storing them in supabse we will store them in r2.

I have already created `room-snapshots` and `room-snapshots-preview` buckets on cloudflare.

We could also migrate all the data from supabase, but it seems we haven't done so for the rooms, so I also didn't look into doing it for snapshots.

One slight drawback of moving to R2 is that it's harder to query data by parent slug. So answering questions like which room is the parent to the most snapshots is a bit harder to answer. Instead of just a simple query we'd need to do some custom logic to go through the bucket. Not sure if have ever needed this info though.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

Existing snapshots:

1. Load an existing snapshot. It should still load correctly. The best way to do that is probably to generate a few of them in advance.

New snapshots:
1. Create a new room.
2. Create a few snapshot links.
3. They should work.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Move storing of snapshots to cloudflare R2.

### DIFF
--- a/apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
+++ b/apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
@@ -338,7 +338,7 @@ export class TLDrawDurableObject extends TLServer {
 		this._roomState = undefined
 	}
 
-	// Load the room's drawing data from supabase
+	// Load the room's drawing data. First we check the bucket, then we fallback to supabase (legacy).
 	override async loadFromDatabase(persistenceKey: string): Promise<DBLoadResult> {
 		try {
 			const key = getR2KeyForRoom(persistenceKey)
--- a/apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
+++ b/apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
@@ -338,7 +338,7 @@ export class TLDrawDurableObject extends TLServer {
 		this._roomState = undefined
 	}
 
-	// Load the room's drawing data. First we check the bucket, then we fallback to supabase (legacy).
+	// Load the room's drawing data. First we check the R2 bucket, then we fallback to supabase (legacy).
 	override async loadFromDatabase(persistenceKey: string): Promise<DBLoadResult> {
 		try {
 			const key = getR2KeyForRoom(persistenceKey)
--- a/apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
+++ b/apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
@@ -1,10 +1,15 @@
 import { CreateSnapshotRequestBody } from '@tldraw/dotcom-shared'
+import { RoomSnapshot } from '@tldraw/tlsync'
 import { IRequest } from 'itty-router'
 import { nanoid } from 'nanoid'
+import { getR2KeyForRoom } from '../r2'
 import { Environment } from '../types'
-import { createSupabaseClient, noSupabaseSorry } from '../utils/createSupabaseClient'
-import { getSnapshotsTable } from '../utils/getSnapshotsTable'
 import { validateSnapshot } from '../utils/validateSnapshot'
+
+export type R2Snapshot = {
+	parent_slug: CreateSnapshotRequestBody['parent_slug']
+	drawing: RoomSnapshot
+}
 
 export async function createRoomSnapshot(request: IRequest, env: Environment): Promise<Response> {
 	const data = (await request.json()) as CreateSnapshotRequestBody
@@ -18,7 +23,6 @@ export async function createRoomSnapshot(request: IRequest, env: Environment): P
 
 	const persistedRoomSnapshot = {
 		parent_slug: data.parent_slug,
-		slug: roomId,
 		drawing: {
 			schema: data.schema,
 			clock: 0,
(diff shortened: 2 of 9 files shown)


apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
```diff
@@ -338,7 +338,7 @@ export class TLDrawDurableObject extends TLServer {
 		this._roomState = undefined
 	}
 
-	// Load the room's drawing data from supabase
+	// Load the room's drawing data. First we check the bucket, then we fallback to supabase (legacy).
```
MitjaBezensek, line 341: Just a cleanup, was out of date.

apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
```diff
@@ -28,13 +32,9 @@ export async function createRoomSnapshot(request: IRequest, env: Environment): P
...
 		},
-	}
-
-	const supabase = createSupabaseClient(env)
-	if (!supabase) return noSupabaseSorry()
+	} satisfies R2Snapshot
 
-	const supabaseTable = getSnapshotsTable(env)
-	await supabase.from(supabaseTable).insert(persistedRoomSnapshot)
+	await env.ROOM_SNAPSHOTS.put(getR2KeyForRoom(roomId), JSON.stringify(persistedRoomSnapshot))
```
MitjaBezensek, line 37: Also used `getR2KeyForRoom` helper. Felt best to be consistent with what we do for rooms.

apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
```diff
@@ -18,7 +23,6 @@ export async function createRoomSnapshot(request: IRequest, env: Environment): P
 
 	const persistedRoomSnapshot = {
 		parent_slug: data.parent_slug,
-		slug: roomId,
```
MitjaBezensek, line 21: This was used for supabase to allow us to query by the slug. It's not needed for R2 though.

apps/dotcom-worker/wrangler.toml
```diff
@@ -115,6 +115,27 @@ bucket_name = "rooms-history-ephemeral-preview"
 binding = "ROOMS_HISTORY_EPHEMERAL"
 bucket_name = "rooms-history-ephemeral"
 
+#################### Room snapshots bucket ####################
+# in dev, we write to the preview bucket and need a `preview_bucket_name`
+[[env.dev.r2_buckets]]
+binding = "ROOM_SNAPSHOTS"
+bucket_name = "room-snapshots-preview"
+preview_bucket_name = "room-snapshots-preview"
```
MitjaBezensek, line 123: Copied this from the `Rooms R2 bucket` definition, but wasn't 100% sure why we do this for dev?

apps/dotcom/src/pages/public-snapshot.tsx
```diff
@@ -7,20 +7,21 @@ import { defineLoader } from '../utils/defineLoader'
 const { loader, useData } = defineLoader(async (args) => {
 	const roomId = args.params.roomId
 	const result = await fetch(`/api/snapshot/${roomId}`)
```
MitjaBezensek, line 9: Here we just swallowed responses with `{error: true, mesage: ....}`. Made it slightly better.

apps/dotcom/src/utils/sharing.ts
```diff
@@ -96,7 +96,8 @@ export async function getNewRoomResponse(snapshot: Snapshot) {
 
 export function useSharing(): TLUiOverrides {
 	const navigate = useNavigate()
-	const id = useSearchParams()[0].get('id') ?? undefined
+	const params = useParams()
+	const roomId = params.roomId
```
MitjaBezensek, line 100: Looks like we lost the `parentSlug` logic for the snapshots when migrating to react router. We were looking at search params instead of regular params.

apps/dotcom-worker/src/lib/routes/getRoomSnapshot.ts
```diff
@@ -1,15 +1,42 @@
...
 	if (!roomId) return fourOhFour()
 
+	// Get the room snapshot from R2
+	const snapshot = await env.ROOM_SNAPSHOTS.get(getR2KeyForRoom(roomId))
+
+	if (snapshot) {
+		const data = ((await snapshot.json()) as R2Snapshot)?.drawing as RoomSnapshot
+		if (data) {
+			return generateReponse(roomId, data)
+		}
```
MitjaBezensek, line 36: I guess it's also safe to return `fourOhFour` here. If we have a snapshot then it must be a new snapshot, so there's no need to check supabase.

apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
```diff
@@ -18,7 +23,6 @@ export async function createRoomSnapshot(request: IRequest, env: Environment): P
```
ds300, line 23: maybe we could include the parent slug as a prefix for the room id

```ts
const roomId = `${data.parent_slug}/v2_c_${nanoid()}`
```

that way we can list/delete all snapshots associated with a particular room if we need to.

not 100% sure a non-url-safe char (`/`) will be ok when including the id in urls but 🤞🏼

ds300 (approved):
looks great. just one small thing I think we could look into to make our lives easier down the line.

apps/dotcom-worker/src/lib/routes/createRoomSnapshot.ts
MitjaBezensek: We probably wouldn't want to expose the parent slug though, otherwise people can get back to the original? Some rooms also don't have a parent slug - you can create a snapshot from your local room.

One thing we could do is store the parent slug in a KV, similar to what we do for readonly rooms. Then we could store snapshots in the R2 under the following paths:

* `public_rooms/parent_slug/snapshot_slug` for the snapshots that do have a parent
* `public_rooms/snapshot_slug` for the snapshots that don't have it

MitjaBezensek: Added the above suggestion.

ds300: ah cool. yeah I forgot about the privacy issue. kv works well!

PR #3695, opened 2024-05-03
Prevent wobble during viewport following

ds300:
This revives the old 'derived camera' idea to prevent cursor wobbling during viewport following.

Before this PR we updated the camera on a tick during viewport following, but the shapes and cursors were not moving on the same tick (we tried that during the perf work and it was all kinds of problematic). Frankly I've forgotten how we ever managed to eliminate wobble here in the first place?

Anyway after this PR we derive the camera based on whether or not we are following a user. When you follow a user it makes it so that your viewport contains their viewport. If your viewport is not already very close to their viewport it will animate the initial position, after which it will 'lock' in place and the derived value will be used from then on. 

This exposed a minor issue in our sync engine: the fact that we send presence updates in separate websocket messages from document updates. We get into situations like this

1. user A follows user B
2. user B deletes the current page they are on
3. user B's page deletion diff gets sent
4. user B's presence update gets sent with a new currentPageId
5. user A receives the page deletion
6. user A still thinks that user B is on the old page and doesn't know how to update the follow state.

So to fix this I made it so that we can (and do) send presence updates in the same websocket messages as document updates so the server can handle them atomically.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

8.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Fixes a bug that caused the cursor & shapes to wiggle around when following someone else's viewport

### DIFF
--- a/packages/tlsync/src/lib/TLSyncClient.ts
+++ b/packages/tlsync/src/lib/TLSyncClient.ts
@@ -431,26 +431,37 @@ export class TLSyncClient<R extends UnknownRecord, S extends Store<R> = Store<R>
 			// if we're offline, don't do anything
 			return
 		}
-		let req: TLPushRequest<R> | null = null
+
+		let presence: TLPushRequest<any>['presence'] = undefined
 		if (!this.lastPushedPresenceState && nextPresence) {
 			// we don't have a last presence state, so we need to push the full state
-			req = {
-				type: 'push',
-				presence: [RecordOpType.Put, nextPresence],
-				clientClock: this.clientClock++,
-			}
+			presence = [RecordOpType.Put, nextPresence]
 		} else if (this.lastPushedPresenceState && nextPresence) {
 			// we have a last presence state, so we need to push a diff if there is one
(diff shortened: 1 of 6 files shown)


ds300:
@TodePond I've heavily refactored the viewport following code in this PR, mind doing a little QA to make sure I haven't broken any features/behaviour I wasn't aware of?

packages/tlsync/src/lib/TLSyncClient.ts
```diff
@@ -423,30 +423,45 @@ export class TLSyncClient<R extends UnknownRecord, S extends Store<R> = Store<R>
 	lastPushedPresenceState: R | null = null
 
 	private pushPresence(nextPresence: R | null) {
+		// make sure we push any document changes first
+		// TODO: need to send presence changes in the same push request as document changes
```
ds300, line 427: ✂️

steveruizok:
Great stuff David, awesome that an issue like this revealed something so useful about how we do our syncing.

steveruizok (approved):

PR #3718, opened 2024-05-08
focus: rework and untangle existing focus management logic in the sdk

mimecuvalo:
Focus management is really scattered across the codebase. There's sort of a battle between different code paths to make the focus the correct desired state. It seemed to grow like a knot and once I started pulling on one thread to see if it was still needed you could see underneath that it was accounting for another thing underneath that perhaps wasn't needed.

The impetus for this PR came but especially during the text label rework, now that it's much more easy to jump around from textfield to textfield. It became apparent that we were playing whack-a-mole trying to preserve the right focus conditions (especially on iOS, ugh). 

This tries to remove as many hacks as possible, and bring together in place the focus logic (and in the darkness, bind them).

## Places affected
- [x] `useEditableText`: was able to remove a bunch of the focus logic here. In addition, it doesn't look like we need to save the selection range anymore.
  - lingering footgun that needed to be fixed anyway: if there are two labels in the same shape, because we were just checking `editingShapeId === id`, the two text labels would have just fought each other for control
- [x] `useFocusEvents`: nixed and refactored — we listen to the store in `FocusManager` and then take care of autoFocus there
- [x] `useSafariFocusOutFix`: nixed. not necessary anymore because we're not trying to refocus when blurring in `useEditableText`. original PR for reference: https://github.com/tldraw/brivate/pull/79
- [x] `defaultSideEffects`: moved logic to `FocusManager`
- [x] `PointingShape` focus for `startTranslating`, decided to leave this alone actually.
- [x] `TldrawUIButton`: it doesn't look like this focus bug fix is needed anymore, original PR for reference: https://github.com/tldraw/tldraw/pull/2630
- [x] `useDocumentEvents`: left alone its manual focus after the Escape key is hit
- [x] `FrameHeading`: double focus/select doesn't seem necessary anymore
- [x] `useCanvasEvents`: `onPointerDown` focus logic never happened b/c in `Editor.ts` we `clearedMenus` on pointer down
  - [x] `onTouchStart`: looks like `document.body.click()` is not necessary anymore

## Future Changes
- [ ] a11y: work on having an accessebility focus ring
- [ ] Page visibility API: (https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) events when tab is back in focus vs. background, different kind of focus
- [ ] Reexamine places we manually dispatch `pointer_down` events to see if they're necessary.
- [ ] Minor: get rid of `useContainer` maybe? Is it really necessary to have this hook? you can just do `useEditor` → `editor.getContainer()`, feels superfluous.

## Methodology
Looked for places where we do:
- `body.click()`
- places we do `container.focus()`
- places we do `container.blur()`
- places we do `editor.updateInstanceState({ isFocused })`
- places we do `autofocus`
- searched for `document.activeElement`

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

- [x] run test-focus.spec.ts
- [x] check MultipleExample
- [x] check EditorFocusExample
- [x] check autoFocus
- [x] check style panel usage and focus events in general
- [x] check text editing focus, lots of different devices, mobile/desktop

### Release Notes

- Focus: rework and untangle existing focus management logic in the SDK


apps/dotcom/src/components/CursorChatBubble.tsx
```diff
@@ -126,11 +124,9 @@ const CursorChatInput = track(function CursorChatInput({
 	}, [editor, value, placeholder])
 
 	useLayoutEffect(() => {
-		// Focus the editor
-		let raf = requestAnimationFrame(() => {
-			raf = requestAnimationFrame(() => {
-				ref.current?.focus()
-			})
+		// Focus the input
+		const raf = requestAnimationFrame(() => {
```
mimecuvalo, line 128: this double raf doesn't seem necessary... but maybe i'm missing something.

apps/dotcom/src/components/PeopleMenu/UserPresenceEditor.tsx
```diff
@@ -52,8 +52,8 @@ export function UserPresenceEditor() {
 					onCancel={toggleEditingName}
 					onBlur={handleBlur}
 					shouldManuallyMaintainScrollPositionWhenFocused
-					autofocus
-					autoselect
+					autoFocus
```
mimecuvalo, line 55: drive-by rename to make `autoFocus` across the SDK consistent

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -39,15 +39,6 @@ export function useCanvasEvents() {
 					name: 'pointer_down',
 					...getPointerInfo(e),
 				})
-
-				if (editor.getOpenMenus().length > 0) {
```
mimecuvalo, line 43: this logic path had bitrotted, as mentioned in the PR description

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -97,9 +88,6 @@ export function useCanvasEvents() {
 
 			function onTouchStart(e: React.TouchEvent) {
 				;(e as any).isKilled = true
-				// todo: investigate whether this effects keyboard shortcuts
```
mimecuvalo, line 100: doesn't look like this is necessary anymore...

packages/tldraw/src/lib/defaultSideEffects.ts
```diff
@@ -2,16 +2,6 @@ import { Editor } from '@tldraw/editor'
 
 export function registerDefaultSideEffects(editor: Editor) {
 	return [
-		editor.sideEffects.registerAfterChangeHandler('instance', (prev, next) => {
```
mimecuvalo, line 5: moved to FocusManager

packages/tldraw/src/lib/shapes/frame/components/FrameHeading.tsx
```diff
@@ -58,14 +58,6 @@ export const FrameHeading = function FrameHeading({
 			// On iOS, we must focus here
 			el.focus()
 			el.select()
-
-			requestAnimationFrame(() => {
```
mimecuvalo, line 62: this double raf doesn't seem necessary anymore...

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
```diff
@@ -44,75 +37,18 @@ export function useEditableText(id: TLShapeId, type: string, text: string) {
...
+			rInput.current?.blur()
+			rInput.current?.focus()
 		}
 	}, [editor, isEditing])
 
-	// 2. Restore the selection changes (and focus) if the element blurs
-	// When the label blurs, deselect all of the text and complete.
-	// This makes it so that the canvas does not have to be focused
-	// in order to exit the editing state and complete the editing state
-	const handleBlur = useCallback(() => {
```
mimecuvalo, line 89: we were doing a lot of hoops and somersaults to account for what `TldrawUiButtonPicker` should have been managing

packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
```diff
@@ -50,6 +51,15 @@ function _TldrawUiButtonPicker<T extends string>(props: TLUiButtonPickerProps<T>
 		const handlePointerUp = () => {
 			rPointing.current = false
 			window.removeEventListener('pointerup', handlePointerUp)
+
+			// This is fun little micro-optimization to make sure that the focus
+			// is retained on a text label. That way, you can continue typing
+			// after selecting a style.
+			const origActiveEl = rPointingOriginalActiveElement.current
```
mimecuvalo, line 58: The logic for what useEditableText was trying to accomplish was moved here

packages/tldraw/src/lib/ui/components/primitives/Button/TldrawUiButton.tsx
```diff
@@ -11,15 +10,6 @@ export interface TLUiButtonProps extends React.HTMLAttributes<HTMLButtonElement>
 /** @public */
 export const TldrawUiButton = React.forwardRef<HTMLButtonElement, TLUiButtonProps>(
 	function TldrawUiButton({ children, disabled, type, ...props }, ref) {
-		const editor = useEditor()
-
-		// If the button is getting disabled while it's focused, move focus to the editor
-		// so that the user can continue using keyboard shortcuts
```
mimecuvalo, line 17: As mentioned in the PR desc., it looks like this isn't needed anymore...

mimecuvalo:
@MitjaBezensek @SomeHats actually hold off on reviewing - there's something i have to fix first

mimecuvalo:
@MitjaBezensek @SomeHats ok! all good now

packages/tldraw/src/test/Editor.test.tsx
```diff
@@ -446,7 +446,9 @@ describe('isFocused', () => {
 		expect(editor.getInstanceState().isFocused).toBe(false)
 	})
 
-	it('becomes false when a child of the app container div receives a focusout event', () => {
+	it.skip('becomes false when a child of the app container div receives a focusout event', () => {
```
mimecuvalo, line 449: hm, i dunno why this test needed to be paused actually. i stared at it but not seeing the difference. @SomeHats if you have more historical context here maybe lemme know.

SomeHats: So in the past, if the tldraw editor was focused, and then focus left (e.g. went to some input somewhere else on the page) we would update the instance state appropriately. I honestly have no idea where we were/are doing this though, but the test still seems to make some conceptual sense to me?

SomeHats:
this is amazing work <3 would love to see tests for some of the places where we've removed workarounds if they don't already exist but also yikes this stuff is hard to test

apps/examples/e2e/tests/test-focus.spec.ts
```diff
@@ -139,7 +139,7 @@ test.describe('Focus', () => {
 		await page.waitForTimeout(100) // takes 30ms or so to focus
 		await page.keyboard.press('d')
 
-		expect(await EditorA.$('.tlui-button[data-testid="tools.draw"][aria-checked="true"]')).not.toBe(
+		expect(await EditorA.$('.tlui-button[data-testid="tools.draw"][aria-checked="true"]')).toBe(
```
MitjaBezensek, line 142: I'm not sure if understand why the draw tool should become unselected here?

packages/tldraw/src/test/Editor.test.tsx
MitjaBezensek: Searching for `focusout` it seems like we only used it in `useSafariFocusOutFix`, which has been completely removed in this PR. Not sure we even handle this event anywhere, so the test might just be obsolete?

mimecuvalo: yeah, but the weird thing is, this test was testing not just Safari 🤔

packages/editor/src/lib/editor/managers/FocusManager.ts
```diff
@@ -0,0 +1,50 @@
...
+ * @public
+ */
+export class FocusManager {
+	private disposeStoreListener?: () => void
+
+	constructor(
+		public editor: Editor,
+		autoFocus?: boolean
+	) {
+		this.disposeStoreListener = editor.store.listen(
```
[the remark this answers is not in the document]
mimecuvalo: ah, that's a good point - lemme double check. prbly better to not have a raf in there, just in case.

mimecuvalo:
> this is amazing work <3 would love to see tests for some of the places where we've removed workarounds if they don't already exist but also yikes this stuff is hard to test

good point + thx! and yeah, hah, hard to test - but yeah, worth it - lemme try to add some more 👍

apps/examples/e2e/tests/test-focus.spec.ts
mimecuvalo: yeah, it took me a little bit to figure out - it's a quirk of this test.
it's because the `autoFocus` prop is changing, so in `TldrawEditorWithReadyStore`→ `useLayoutEffect`, the changing prop causes the editor to be reset. you wouldn't _normally_ change the `autoFocus` prop to change focus, you'd want to programatically change the editor you already have.

mimecuvalo: that said, maybe i should update `MultipleExample` to not change props but actually update the instance state to more properly address this.

MitjaBezensek: > yeah, it took me a little bit to figure out - it's a quirk of this test. it's because the `autoFocus` prop is changing, so in `TldrawEditorWithReadyStore`→ `useLayoutEffect`, the changing prop causes the editor to be reset. you wouldn't _normally_ change the `autoFocus` prop to change focus, you'd want to programatically change the editor you already have.

ouch 😂 At least for me it definitely sounds like a property you set once and should not change, so the change in behaviour is fine.

Yeah, updating the example sounds good.

packages/tldraw/src/test/Editor.test.tsx
SomeHats: is this not generally desired behaviour though? if focus leaves the editor in the DOM, the internal state of the editor should reflect that, right?

apps/examples/e2e/tests/test-focus.spec.ts
mimecuvalo: i reworked MultipleExample to not changethe prop

SomeHats: i don't feel suuuuuper strongly here - happy to leave as is - but in general, I would look to the browser/react for how attributes mirroring built-in ones should behave.

In both of those, i believe setting this attribute on a pre-existing element does nothing, but it definitely doesn't clear the state of the items. in the browser, i think it's respected on page load, and in react it's respected when an element is first mounted.

packages/tldraw/src/test/Editor.test.tsx
mimecuvalo: Well, to be clear though, this is leaving focus from a child of the editor, not the editor directly. I think if it leaves focus from a child it would just go to a parent element possibly? It's sort of depends on where the person clicks next or tabs to next.

apps/examples/e2e/tests/test-focus.spec.ts
mimecuvalo: This is a good point, I reworked this little bit so that TldrawEditor only takes the initial auto focus value. Changing it shouldn't affect the editor.

mimecuvalo:
added a couple more tests (for edit→edit case and for the style picker+maintaining focus in textarea case). lemme know if there's another case to add but i think we should be good to go!

apps/dotcom/src/components/CursorChatBubble.tsx
ds300: yeah your version works in chrome, safari and firefox. cc @TodePond do you remember if this double raf was intentional?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8021,6 +8037,21 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * ```ts
+	 * editor.focus()
+	 * ```
+	 *
+	 * @public
+	 */
+	focus(): this {
+		this.focusManager.focus()
+		return this
+	}
```
ds300, line 8053: I wonder whether this should set the `instanceState.isFocused` to true if not already?

ds300: or really maybe it should be the job of the focus manager

packages/editor/src/lib/hooks/useCanvasEvents.ts
ds300: did you test this on iOS? seems fine on android

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
```diff
@@ -13,106 +13,43 @@ import { INDENT, TextHelpers } from './TextHelpers'
...
-						if (document.activeElement !== elm) {
-							elm.focus()
-						}
-						elm.select()
-					}
-				}
-			})
+			if (shapeId === id) {
+				rInput.current?.select()
+			}
```
ds300, line 25: you sure this is still working on iOS?

packages/tldraw/src/lib/ui/components/primitives/Button/TldrawUiButton.tsx
ds300: are you sure? the document.activeElement still reverts to the `body` when the button becomes inactive in my tests. admittedly that no longer seems to prevent keyboard shortcuts from working but maybe it could cause other issues?

ds300 (approved):
I don't feel super qualified to review this but things all seem to work on mac and android in my testing so 👍🏼 

The only thing I'm feeling pause on is the behaviour/api of the focus manager.

- should calling focusManager.focus update the instanceState.isFocused property, and same for the .blur method?
- should we keep the manager itself private and just have a .focus and .blur method on the editor that call the underlying manager methods? I know how steve feels about namespaces.

packages/editor/src/lib/hooks/useCanvasEvents.ts
mimecuvalo: yes, i've been testing on the iOS simulator. will definitely test again once it's on staging.

packages/tldraw/src/lib/ui/components/primitives/Button/TldrawUiButton.tsx
mimecuvalo: i think i'm specifically making sure that the original PR (https://github.com/tldraw/tldraw/pull/2630) is working. if it causes other issues, then i think i would rather that we know what those issues are vs. this covering them up potentially (as a fix for something unrelated)

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: i think possibly: that it's tricky. if there are multiple editors on the page, i could see an argument that you might want to temporarily translate/move a shape on one editor (Editor A), but the focus remains on a separate editor (Editor B). and that just because you focus Editor A, it doesn't mean that you want Editor A to take over control from Editor B. but perhaps that's overthinking it?

packages/tldraw/src/lib/shapes/shared/useEditableText.ts
mimecuvalo: i've been testing on the iOS sim - will definitely test on staging once it lands to double check!

mimecuvalo:
> should calling focusManager.focus update the instanceState.isFocused property, and same for the .blur method?

yeah, it's a good question - like i said above: i think possibly: that it's tricky. if there are multiple editors on the page, i could see an argument that you might want to temporarily translate/move a shape on one editor (Editor A), but the focus remains on a separate editor (Editor B). and that just because you focus Editor A, it doesn't mean that you want Editor A to take over control from Editor B. but perhaps that's overthinking it?

> should we keep the manager itself private and just have a .focus and .blur method on the editor that call the underlying manager methods? I know how steve feels about namespaces.

fair point - i think i can make the focus manager be private actually. (to be clear, no one is calling `blur()` explicitly at the moment, it's only used in FocusManager at the moment)

PR #3721, opened 2024-05-08
Move arrow helpers from editor to tldraw

SomeHats:
With the new work on bindings, we no longer need to keep any arrows stuff hard-coded in `editor`, so let's move it to `tldraw` with the rest of the shapes.

Couple other changes as part of this:
- We had two different types of `WeakMap` backed cache, but we now only have one
- There's a new free-standing version of `createComputedCache` that doesn't need access to the editor/store in order to create the cache. instead, it returns a `{get(editor, id)}` object and instantiates the cache on a per-editor basis for each call.
- Fixed a bug in `createSelectedComputedCache` where the selector derivation would get re-created on every call to `get`

### Change Type

#### Breaking changes
- `editor.getArrowInfo(shape)` has been replaced with `getArrowInfo(editor, shape)`
- `editor.getArrowsBoundTo(shape)` has been removed. Instead, use `editor.getBindingsToShape(shape, 'arrow')` and follow the `fromId` of each binding to the corresponding arrow shape
- These types have moved from `@tldraw/editor` to `tldraw`:
    - `TLArcInfo`
    - `TLArrowInfo`
    - `TLArrowPoint`
- `WeakMapCache` has been removed


mimecuvalo:
200w

packages/store/src/lib/Store.ts
```diff
@@ -989,3 +991,42 @@ class HistoryAccumulator<T extends UnknownRecord> {
...
+/**
+ * Free version of {@link Store.createComputedCache}.
+ *
+ * @example
+ * ```ts
+ * const myCache = createComputedCache('myCache', (editor: Editor, shape: TLShape) => {
+ *     return editor.getSomethingExpensive(shape)
+ * })
+ *
+ * myCache.get(editor, shape.id)
```
ds300, line 1008: cool stuff

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4900,7 +4843,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		const parentId = typeof parent === 'string' ? parent : parent.id
 		const ids = this._parentIdsToChildIds.get()[parentId]
 		if (!ids) return EMPTY_ARRAY
-		return this._childIdsCache.get(ids, () => ids)
```
ds300, line 4903: what was this even doing before 🤷🏼

packages/editor/api-report.md
```diff
@@ -2096,57 +2075,6 @@ export type TLAnyBindingUtilConstructor = TLBindingUtilConstructor<any>;
...
-// @public (undocumented)
-export interface TLArrowBindings {
-    // (undocumented)
-    end: TLArrowBinding | undefined;
-    // (undocumented)
-    start: TLArrowBinding | undefined;
-}
-
-// @public (undocumented)
-export type TLArrowInfo = {
```
ds300, line 2124: I think these types were supposed to remain public in the `'tldraw'` package but it looks like they're not being exported from the index there?

ds300 (approved):
Looks good. Just noticed one issue with exports I think

packages/editor/api-report.md
SomeHats: ahh good catch, thank you!

Issue #3728, opened 2024-05-09
Local images


steveruizok:
In local projects, could we put images separately into indexeddb, and then use the image src to point to it? Is there a more efficient way of doing local images?

danolekh:
This issue seems resolved, yet I cannot find anything related how to store blob separately in indexed db - as it was implemented in PR  #3836
I also didn't find any examples on how to do this. Was api removed in favor of something else? If so, how would implementing something that Steve asked would look like?

danolekh:
> This issue seems resolved, yet I cannot find anything related how to store blob separately in indexed db - as it was implemented in PR [#3836](https://github.com/tldraw/tldraw/pull/3836) I also didn't find any examples on how to do this. Was api removed in favor of something else? If so, how would implementing something that Steve asked would look like?

Okay, I see PR #4068. Exactly what I looked for

PR #3745, opened 2024-05-13
proof of concept: assets as references to blobs in indexedDB, not directly as base64 in store

mimecuvalo:
As I look at LOD holistically and whether we have multiple sources when working locally, I learned that our system used base64 encoding of assets directly. Issue https://github.com/tldraw/tldraw/issues/3728

assetstore

The motivations and benefits are:
- store size: not having a huge base64 blobs injected in room data
- perf on loading snapshot: this helps with loading the room data more quickly
- multiple sources: furthermore, if we do decide to have multiple sources locally (for each asset), then we won't get a multiplicative effect of even larger JSON blobs that have lots of base64 data in them
- encoding/decoding perf: this also saves the (slow) step of having to base64 encode/decode our assets, we can just strictly with work with blobs.

The benefit here is clear! I created an `AssetBlobObjectStore` which probably needs some thought (again, this is proof-of-concept), so if we decide to move forward, let's discuss how best to integrate that part.

Todo:
- [x] decodes video and images
- [x] make sure it syncs to other tabs
- [x] make sure it syncs to other multiplayer room
- [ ] fix tests
- [ ] have backup for existing editors instantiated without assetBlobStore

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test the shit out of uploading/downloading video/image assets, locally+multiplayer.

- [ ] Need to fix current tests and write new ones

### Release Notes

- Assets: store as reference to blob in indexedDB instead of storing directly as base64 in the snapshot.

### DIFF
--- a/apps/dotcom/src/utils/sharing.ts
+++ b/apps/dotcom/src/utils/sharing.ts
@@ -241,7 +241,7 @@ async function getRoomData(
 			// processed it
 			if (!asset) continue
 
-			data[asset.id] = await cloneAssetForShare(asset, uploadFileToAsset)
+			data[asset.id] = await cloneAssetForShare(asset, editor, uploadFileToAsset)
 			// remove the asset after processing so we don't clone it multiple times
 			assets.delete(asset.id)
 		}
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -54,6 +54,7 @@ import {
 	JsonObject,
 	PerformanceTracker,
 	Result,
+	WeakCache,
 	annotateError,
 	assert,
 	assertExists,
@@ -116,6 +117,7 @@ import { debugFlags } from '../utils/debug-flags'
 import { getIncrementedName } from '../utils/getIncrementedName'
 import { getReorderingShapesChanges } from '../utils/reorderShapes'
 import { applyRotationToSnapshotShapes, getRotationSnapshot } from '../utils/rotation'
+import { AssetBlobObjectStore } from '../utils/sync/AssetBlobObjectStore'
 import { uniqueId } from '../utils/uniqueId'
 import { BindingUtil, TLBindingUtilConstructor } from './bindings/BindingUtil'
 import { bindingsIndex } from './derivations/bindingsIndex'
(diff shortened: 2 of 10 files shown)


apps/dotcom/src/utils/sharing.ts
```diff
@@ -241,7 +241,7 @@ async function getRoomData(
 			// processed it
 			if (!asset) continue
 
-			data[asset.id] = await cloneAssetForShare(asset, uploadFileToAsset)
+			data[asset.id] = await cloneAssetForShare(asset, editor, uploadFileToAsset)
```
steveruizok, line 244: secret style guide nit: whenever editor is an argument to a function, let's make it the first argument

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3669,6 +3682,39 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * @example
+	 * ```ts
+	 * editor.getAssetBlobInObjectStore('asset1')
+	 * ```
+	 *
+	 * @public
+	 */
+	async getAssetBlobFromObjectStore(asset: TLAsset): Promise<Blob | undefined> {
+		return await this.assetBlobCache.get(asset, () =>
+			this.assetBlobStore.getAssetBlobFromObjectStore({ asset })
```
steveruizok, line 3700: TIL our weakmap caches can work with async stuff

SomeHats:
left some other thoughts on this here: https://github.com/tldraw/tldraw/pull/3764#issuecomment-2122516560

my take on this is that I like what we're doing with this, but i'm not sure it belongs quite so core as this makes it. i think blob storage is quite niche - it's really only relevant if you have an editor saving locally in the browser, whereas most people either have a multiplayer room or some sort of other saving mechanism. i certainly don't think it should be the default as this PR makes it.

instead, i would love to see a slightly more general system for making asset storage/retrieval more flexible. what would it look like for this to be build entirely in userland on your asset rewriting hook from #3764, for example?

mimecuvalo:
> my take on this is that I like what we're doing with this, but i'm not sure it belongs quite so core as this makes it. i think blob storage is quite niche - it's really only relevant if you have an editor saving locally in the browser, whereas most people either have a multiplayer room or some sort of other saving mechanism. i certainly don't think it should be the default as this PR makes it.

yeah, totally fair to not make it core, definitely would make sense more as a sidecar kind of option.

> instead, i would love to see a slightly more general system for making asset storage/retrieval more flexible. what would it look like for this to be build entirely in userland on your asset rewriting hook from https://github.com/tldraw/tldraw/pull/3764, for example?

i see the overlap for sure, but i think in my head they're quite different (as far as storage vs. dynamic retrieval). i'd avoid making a system that tries to address them both perhaps because they're trying to address two different problems. (it's maybe a coincidence that I'm working on both at the same time heh)

let's discuss more tomorrow!

mimecuvalo:
(closed in favor of https://github.com/tldraw/tldraw/pull/3836)

PR #3765, opened 2024-05-17
google meet: add hardware whiteboard integration

mimecuvalo:
pushing out my changes but this is a draft. we need to do look into this more:
- [ ] we might need to reach out to Google support and request to be put on their "start a whiteboard" whitelist/partner list? it's actually pretty unclear how to get on that list. I don't see any permissions/API scopes that is meant to enable that 🤔 
- [ ] but maybe this is something that you @steveruizok as Google Workspace admin see on your end? let's look together when you get back.

I initially tried doing a bundle (using esbuild (and i tried parcel/rollup too)) but it didn't feel like the right path, and also it didn't work when loading it in the Apps Scripts.
So then I went the route of just doing an iframe and I think that feels much better. This means though that we do want our iframe protector to let through this usecase. But also, we could maybe just redirect always to a new room? I'm not sure yet.

The build script helps either build the prod or staging version depending on what you want.

Once we do find that the staging version works, then we'll go through the process of:
- [x] applied for https://developers.google.com/workspace/preview (already did this to get access to new Google Meet APIs just in case)
- [x] added to google analytics (already done)
- [ ] turn off testing mode for oauth and submit for review
- [ ] continue publishing process to Create a store listing for our prod app: https://developers.google.com/workspace/marketplace/how-to-publish

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

- Google Meet: add hardware whiteboard integration


steveruizok:
Great progress! Would love to see steps to test this ourselves. Is there a sort of testflight for Google workspace apps if we wanted to launch it for our organization only, while we dial it in?

Once we have something ready, I'll do my song and dance to see if we can get it recommended. Most apps in this category use integrations as a way to generate users / accounts, which we don't really have at the moment, I would expect questions around "how does a user save this?" or "how can a user return to their board?". Maybe we could solve that with some UI, like an "email me a link" button.

mimecuvalo:
discussed some things offline with @steveruizok. i'm pushing this out to staging so i can test out the app properly!

@steveruizok and yeah, like i said offline, it's less about getting it recommended - it's more that it has to be in a special category, it looks like. but i'm not quite sure. i think if we start the support ticket, it's more about getting it approved as a special-whiteboard app.

PR #3783, opened 2024-05-20
Move many constants into configurable 'settings'

SomeHats:
In #3628, @steveruizok experimented with moving all our constants into a big object so people can override them. One problem with that is that as we were dereferencing those objects at module-init-time, we'd mostly ignore any overrides.

This diff takes an alternative approach: we change the constants that we want to be overridden to `let` bindings instead of `const` ones, and code-gen setters for them. Because esmodule imports import bindings rather than values, re-assigning those values updates them whenever they're references in future.

For example, to increase the max number of pages, you could do something like this:
```ts
import {updateTldrawSettings} from 'tldraw'

updateTldrawSettings({MAX_PAGES: 100})
```

My immediate motivation for this is allowing you to create shapes that can't have arrows bound to them using `SHAPES_WHICH_ARROWS_CANNOT_BIND_TO`.

### Change Type

- [x] `feature` — New feature

### Release Notes

- You can now update many of tldraw's internal settings using `updateTldrawSettings`


ds300:
hm i'm wary of this unless we can make sure we don't actually dereference the values at module init time. Even then, it's not great practice to have global state for config. e.g. in our examples app, wouldn't settings for one example bleed to all the others unless we force a page reload? Admittedly our users probably don't need multiple tldraw configs, but still I think I'd rather we solve this problem with a `<Tldraw constants={{ foo: bar }} />` prop, despite the extra faff it would require on our side.

SomeHats:
yup, i think i prefer a giant options bag for this too actually. the issue with that is we can't have different options for the editor and the tldraw packages

SomeHats:
Closing in favour of #3799

PR #3797, opened 2024-05-21
rework canBind callback

SomeHats:
This PR reworks the `canBind` callback to work with customizable bindings. It now accepts an object with a the shape, the other shape (optional - it may not exist yet), the direction, and the type of the binding. Devs can use this to create shapes that only participate in certain binding types, can have bindings from but not to them, etc.

If you're implementing a binding, you can see if binding two shapes is allowed using `editor.canBindShapes(fromShape, toShape, 'my binding type')`

### Change Type

#### Breaking changes
The `canBind` flag now accepts an options object instead of just the shape in question. If you're relying on its arguments, you need to change from `canBind(shape) {}` to `canBind({shape}) {}`.

### DIFF
--- a/apps/examples/src/examples/pin-bindings/PinExample.tsx
+++ b/apps/examples/src/examples/pin-bindings/PinExample.tsx
@@ -1,3 +1,4 @@
+import { TLShapeUtilCanBindOpts } from '@tldraw/editor/src/lib/editor/shapes/ShapeUtil'
 import {
 	BindingOnShapeChangeOptions,
 	BindingOnUnbindOptions,
@@ -44,7 +45,10 @@ class PinShapeUtil extends ShapeUtil<PinShape> {
 		return {}
 	}
 
-	override canBind = () => false
+	override canBind({ direction }: TLShapeUtilCanBindOpts<PinShape>) {
+		// bindings can go _from_ pins to other shapes, but not the other way round
+		return direction === 'from'
+	}
 	override canEdit = () => false
 	override canResize = () => false
 	override hideRotateHandle = () => true
@@ -93,7 +97,9 @@ class PinShapeUtil extends ShapeUtil<PinShape> {
 			.getShapesAtPoint(pageAnchor, { hitInside: true })
 			.filter(
 				(shape) =>
-					shape.type !== 'pin' && shape.parentId === pin.parentId && shape.index < pin.index
+					this.editor.canBindShapes(pin.id, shape.id, 'pin') &&
+					shape.parentId === pin.parentId &&
+					shape.index < pin.index
 			)
 
 		for (const target of targets) {
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -18,8 +18,9 @@ import {
 	TLAssetId,
 	TLAssetPartial,
 	TLBinding,
+	TLBindingCreate,
 	TLBindingId,
-	TLBindingPartial,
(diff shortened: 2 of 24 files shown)


packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -32,6 +32,22 @@ export interface TLShapeUtilConstructor<
...
+	/**
+	 * The direction:
+	 * - `"to"` means the other shape is being bound to this shape.
+	 * - `"from"` means this shape is being bound to the other shape.
+	 */
+	direction: 'to' | 'from'
+	/** The type of binding, e.g. `"arrow"`. */
+	type: string
+}
+
```
steveruizok, line 50: Something about shape / otherShape / direction had me confused at first, maybe its the framing of "this shape" vs "the other shape". `fromShape` / `toShape` seems more consistent with the binding format, though then it would be up to the user to figure out which shape (or botth) is the shape with the type corresponding to to this util. That might be ok.

If this function were a question, I think it would be "hey util, i'm trying to make a binding and at least one of these shapes is one of yours, does this look good to you?".

apps/examples/src/examples/pin-bindings/PinExample.tsx
```diff
@@ -93,7 +97,9 @@ class PinShapeUtil extends ShapeUtil<PinShape> {
 			.getShapesAtPoint(pageAnchor, { hitInside: true })
 			.filter(
 				(shape) =>
-					shape.type !== 'pin' && shape.parentId === pin.parentId && shape.index < pin.index
+					this.editor.canBindShapes(pin.id, shape.id, 'pin') &&
+					shape.parentId === pin.parentId &&
+					shape.index < pin.index
```
ds300, line 102: Did we lose the constraint that pins can't bind to other pins?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5197,6 +5210,33 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		const to = toShape ? this.getShape(toShape) : undefined
+
+		const canBindFrom = from
+			? this.getShapeUtil(from).canBind({ shape: from, otherShape: to, direction: 'from', type })
+			: undefined
+		const canBindTo = to
+			? this.getShapeUtil(to).canBind({ shape: to, otherShape: from, direction: 'to', type })
+			: undefined
+
+		return canBindFrom !== false && canBindTo !== false
```
ds300, line 5238: this feels quite heavy. I wonder whether we need to actually refer to the instances of the shapes when deciding if it can bind,  for the time being can we just use the shape `type` and then maybe add support for deciding based on the actual shape instances later if anybody needs it?

So we could have e.g. `editor.canBindShapes({fromShapeType, toShapeType, bindingType})` and then the shape utils can implement the same interface `canBind({ fromShapeType, toShapeType, bindingType })` too.

apps/examples/src/examples/pin-bindings/PinExample.tsx
SomeHats: nope, it's just covered by the `canBindShapes` check now

packages/editor/src/lib/editor/Editor.ts
SomeHats: we need the actual instances in some cases (or at least... we need to go from ID to type first). i've moved those to the callsites though - i think i prefer this the way it is now.

PR #3799, opened 2024-05-21
Move constants to options prop

SomeHats:
Another go at #3628 & #3783. This moves (most) constants into `editor.options`, configurable by the `options` prop on the tldraw component.

### Change Type

- [x] `feature` — New feature

### Release Notes

You can now override many options which were previously hard-coded constants. Pass an `options` prop into the tldraw component to change the maximum number of pages, grid steps, or other previously hard-coded values. See `TldrawOptions` for more


mimecuvalo:
I'll chime in hear to say that this feels like a much more natural way than the previous iterations. Great progress :)

I hadn't thought of this until @steveruizok mentioned it in the LOD PR (https://github.com/tldraw/tldraw/pull/3764) but it _is_ interesting to consider the thought of having per-shape options in parallel work with this, just a bit. To be clear, I _don't_ think it should influence this work but as we think about adding options holistically and developing a story we tell SDK users on how to configure things, there might be some similar patterns there perhaps.

Separate topic: I'll be a broken record about namespacing (sry Steve :P) but when I see a grab bag of options it's pretty hard to discern what they're used for without trying them out one-by-one (or, sort of arduously reading the manual, and let's be honest, I just wanna mess around and not read the manual :P).

So when I see an option that says `animationMediumMs`, _**I**_ know that it has to do only with the camera controls because I work at tldraw, but the average user won't necessarily know if changing that option will affect the camera, or shape resizing, some other UI, or globally affect any UI. And the opposite is true too, if they're looking for camera options, they won't know the full spectrum without combing through the options one-by-one :-/ That's my soapbox, folks!

packages/editor/src/lib/options.ts
```diff
@@ -0,0 +1,76 @@
...
+	readonly maxPages: number
+	readonly animationMediumMs: number
+	readonly followChaseViewportSnap: number
+	readonly doubleClickDurationMs: number
+	readonly multiClickDurationMs: number
+	readonly coarseDragDistanceSquared: number
+	readonly dragDistanceSquared: number
+	readonly defaultSvgPadding: number
+	readonly cameraSlideFriction: number
+	readonly gridSteps: readonly { min: number; mid: number; step: number }[]
```
ds300, line 29: i think if we wanna be careful any nested properties should be `readonly` too
```suggestion
	readonly gridSteps: readonly { 
	  readonly min: number
	  readonly mid: number
	  readonly step: number
	}[]
```

ds300 (approved):
love this

PR #3801, opened 2024-05-21
fix pattern fill lods

SomeHats:
Camera options broke pattern lods. This PR adapts the more flexible take of pattern LODs i did for my version of camera controls to the new version.

### Change Type

- [x] `bugfix` — Bug fix

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
+++ b/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
@@ -4,24 +4,17 @@ import {
 	DefaultFontStyle,
 	FileHelpers,
 	SvgExportDef,
+	TLDefaultColorTheme,
 	TLDefaultFillStyle,
 	TLDefaultFontStyle,
 	TLShapeUtilCanvasSvgDef,
 	debugFlags,
+	last,
 	useEditor,
+	useValue,
 } from '@tldraw/editor'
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
```diff
@@ -154,86 +147,111 @@ const canvasBlob = (size: [number, number], fn: (ctx: CanvasRenderingContext2D)
...
+	])
 }
 
 function usePattern() {
 	const editor = useEditor()
-	const dpr = editor.getInstanceState().devicePixelRatio
+	const dpr = useValue('devicePixelRatio', () => editor.getInstanceState().devicePixelRatio, [
+		editor,
+	])
+	const maxZoom = useValue('maxZoom', () => Math.ceil(last(editor.getCameraOptions().zoomSteps)!), [
```
SomeHats, line 201: @steveruizok is this the best way to get the max zoom level?

ds300 (approved):
This seems to work well in my tests!

packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
ds300: this seems right to me, let's go with it for now and @steveruizok can chime in later if it's wrong

PR #3810, opened 2024-05-22
security: add recommended OWASP settings; also Zoom apps require them

mimecuvalo:
As I was setting up the Zoom app, it turns out they're very strict about requiring OWASP headers for their apps. https://developers.zoom.us/docs/zoom-apps/security/owasp/

- `Strict-Transport-Security`: max-age is set to 2 years, and is suffixed with preload, which is necessary for inclusion in all major web browsers' HSTS preload lists, like Chromium, Edge, and Firefox.
- CSP: just set to the default, not blocking anything at the moment to avoid going down this rabbit hole.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


apps/dotcom/scripts/build.ts
```diff
@@ -27,6 +27,12 @@ function loadSpaRoutes() {
 		check: true,
 		src: route.vercelRouterPattern,
 		dest: '/index.html',
+		headers: {
+			'Strict-Transport-Security': 'max-age=63072000; includeSubDomains; preload',
+			'X-Content-Type-Options': 'nosniff',
+			'Referrer-Policy': 'no-referrer-when-downgrade',
+			'Content-Security-Policy': 'default-src: *',
+		},
```
ds300, line 35: would we not need to add any/all of these to our assets/api/index routes below?

mimecuvalo: good question.
- we don't need them for the API routes.
- for assets, the `X-Content-Type-Options` is useful to add - so i've added it.
- good point on the 404 page, i've added it there too 👍

mimecuvalo:
ugh, our SPA routes were wrong actually (so this was working on other paths but not on the / base path!). the SPA routes are generated from the mapping in `routes.test.tsx.snap` which i just realized was wrong for the index file 🤦  (it was set to `"^//?$"` instead of `"^/?$"`)

apps/dotcom/src/__snapshots__/routes.test.tsx.snap
```diff
@@ -4,7 +4,7 @@ exports[`the_routes 1`] = `
 [
   {
     "reactRouterPattern": "/",
-    "vercelRouterPattern": "^//?$",
+    "vercelRouterPattern": "^/?$",
```
ds300, line 7: wait why did this need to change? I can't imagine vercel was matching against the empty string for the root path.

mimecuvalo: yeah that's what i said above in the comment:
> ugh, our SPA routes were wrong actually (so this was working on other paths but not on the / base path!). the SPA routes are generated from the mapping in routes.test.tsx.snap which i just realized was wrong for the index file 🤦 (it was set to "^//?$" instead of "^/?$")

vercel was _not_ matching on this - this was broken :-/

mimecuvalo: double-forward slash isn't an escape sequence - `\\` double _back_slash would be though 🙃

ds300: it's not trying to escape anything, `"^//?$"` matches a '/' at the start of the string, followed by an optional '/', then nothing more. so it matches the paths `/` and `//`, so it should have been fine when you go to `tldraw.com` unless vercel was matching against the empty string which seems quite unlikely to me? i guess stranger things have been known to happen...

mimecuvalo: ach! right - totally read that regex wrong, my bad.

ok, actually something is still wrong here then. with my latest push, it's still not recognizing the main `/` path, only the other paths 🤔 it actually _might_ be the empty path?? i'm going to push up and try.

mimecuvalo: nope. that didn't work - going to keep trying other things - bit of a black box, not sure what it's seeing.
going to try `/index.html` next...

mimecuvalo: the latest change did it! thx for rubber ducking with me @ds300

ds300 (approved):
nice

PR #3811, opened 2024-05-22
Snapshots pit of success

ds300:
Lots of people are having a bad time with loading/restoring snapshots and there's a few reasons for that:

- It's not clear how to preserve UI state independently of document state.
- Loading a snapshot wipes the instance state, which means we almost always need to
  - update the viewport page bounds
  - refocus the editor
  - preserver some other sneaky properties of the `instance` record

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/editor/src/lib/hooks/useScreenBounds.ts
steveruizok, line 1: The previous code set the screen bounds to the canvas, which may not be identical to the container. (i.e. if a user is modifying the canvas to be inset within the editor, so that they can put UI elements to the left or right of it). Quite a niche use-case but some users were doing it.

See https://examples.tldraw.com/inset-canvas
vs. this PR's https://examples-htjbfcamq-tldraw.vercel.app/inset-canvas

packages/editor/src/lib/editor/Editor.ts
steveruizok, line 1: See other comment; if this is a drive-by then I'd prefer to keep the previous code, as we don't always call it with bounds that are identical to the container's bounds.

apps/examples/src/examples/snapshots/SnapshotExample.tsx
steveruizok, line 1: Comments go below, `[1]` etc, so that this code is easier to read / copy and paste for folks who are doing that storyline. Style guide sorry

steveruizok:
Good change! This should be much easier for users. Few changes in the comments above, not sure whether the screen bounds thing was critical to this or not, but it should use the bounds of the canvas rather than the container.

PR #3812, opened 2024-05-22
docs: add bindings documentation and update unbind reason

SomeHats:
Adds docs (reference material and a guide) for the bindings API. Also, the unbind reason enum is now a union of strings.

### Change type

- [x] `improvement`

### Test plan

1. Review the documentation changes in the docs app.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added comprehensive documentation for the bindings API and updated unbind reason to a string union.

### DIFF
--- a/apps/docs/content/docs/editor.mdx
+++ b/apps/docs/content/docs/editor.mdx
@@ -426,8 +426,7 @@ here's a binding record for one end of an arrow shape:
 Every binding contains some base information - its ID & the type of binding, as well as the ID of
 the shape that a binding comes _from_ and the ID of the shape that the binding goes _to_. These two
 properties work the same way, but it's often useful to have an explicit direction in a binding
-relationship. For example, an arrow binding always goes _from_ an arrow _to_ the shape it's bound
-to.
+relationship. For example, an arrow binding always goes _from_ an arrow _to_ another shape.
 
 Bindings contain their own type-specific information in the `props` object. Each type of binding can
 have different props.


apps/docs/content/docs/editor.mdx
```diff
@@ -350,6 +350,109 @@ The camera may be in two states, `idle` or `moving`.
...
+    },
+    "meta": {},
+}
+```
+
+Every binding contains some base information - its ID & the type of binding, as well as the ID of
+the shape that a binding comes _from_ and the ID of the shape that the binding goes _to_. These two
+properties work the same way, but it's often useful to have an explicit direction in a binding
+relationship. For example, an arrow binding always goes _from_ an arrow _to_ the shape it's bound
+to.
```
ds300, line 393:
```suggestion
relationship. For example, an arrow binding always goes _from_ an arrow _to_ another shape.
```

apps/docs/content/docs/editor.mdx
```diff
@@ -350,6 +350,109 @@ The camera may be in two states, `idle` or `moving`.
 
 You can get the current camera state with [Editor#getCameraState](?).
 
+# Bindings
```
ds300, line 353: Would this section be better positioned in the Shapes article?

ds300 (approved):
This is really good stuff!

apps/docs/content/docs/editor.mdx
SomeHats: I think it could even stand as an article of its own, but instructions from @steveruizok were to just put everything in the editor article

PR #3813, opened 2024-05-22
Fix markdown list rendering on docs site

SomeHats:
We write our API docs in markdown embedded in tsdocs comments. vscode's hover preview of these docs renders them as expected, but api-extract treats whitespace as insignificant and strips out most newlines, which breaks markdown lists. See https://github.com/microsoft/tsdoc/issues/178 for details.

This PR patches tsdoc's emitter so that it preserves newlines. That way, we can write markdown lists and have them picked up as expected by the docs site markdown parser. I don't expect this to introduce other issues with previously ignored line breaks and markdown is only sensitive to linebreaks in certain scenarios (like lists) anyway.

(Extracted from bindings docs - #3812)

Before:
Screenshot 2024-05-22 at 15 00 43

After:
Screenshot 2024-05-22 at 14 51 28

### Change Type

- [x] `bugfix` — Bug fix

### DIFF
--- a/.yarn/patches/@microsoft-tsdoc-npm-0.14.2-9988282153.patch
+++ b/.yarn/patches/@microsoft-tsdoc-npm-0.14.2-9988282153.patch
@@ -1,0 +1,91 @@
+diff --git a/lib-commonjs/emitters/TSDocEmitter.js b/lib-commonjs/emitters/TSDocEmitter.js
+index 36c607ac192810524ae28ce818d64652efedca98..41650ab83b6825d33b75fb86ec9b8eb0f0b5dfc7 100644
+--- a/lib-commonjs/emitters/TSDocEmitter.js
++++ b/lib-commonjs/emitters/TSDocEmitter.js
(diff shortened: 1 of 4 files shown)


steveruizok (approved):
Good catch!

mimecuvalo:
should we send a patch upstream @SomeHats ? or is this very particular to the way we do things?

SomeHats:
> should we send a patch upstream @SomeHats ? or is this very particular to the way we do things?

It looks like there's been some discussion about the right approach to this on their repo. I left our [patch in a comment](https://github.com/microsoft/tsdoc/issues/178#issuecomment-2124902734) on the open issue for lists support, but from the discussion in their issues on [markdown lists support](https://github.com/microsoft/tsdoc/issues/178) and [word wrapping](https://github.com/microsoft/tsdoc/issues/200#issuecomment-565770852) it sounds like they wouldn't accept this PR at the moment: they would want to add lists to the tsdoc AST, rather than simply preserve the formatting in what it understands as "plain text" to be interpreted later.

PR #3815, opened 2024-05-22
Force `interface` instead of `type` for better docs

SomeHats:
Typescript's type aliases (`type X = thing`) can refer to basically anything, which makes it hard to write an automatic document formatter for them. Interfaces on the other hand are only object, so they play much nicer with docs. Currently, object-flavoured type aliases don't really get expanded at all on our docs site, which means we have a bunch of docs content that's not shown on the site.

This diff introduces a lint rule that forces `interface X {foo: bar}`s instead of `type X = {foo: bar}` where possible, as it results in a much better documentation experience:

Before:
Screenshot 2024-05-22 at 15 24 13

After:
Screenshot 2024-05-22 at 15 33 01

### Change Type

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,6 +70,7 @@ module.exports = {
 			'error',
 			{ name: 'structuredClone', message: 'Use structuredClone from @tldraw/util instead' },
 		],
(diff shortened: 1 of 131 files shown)


ds300 (approved):
heck yeah good one

PR #3821, opened 2024-05-23
Fix broken files

steveruizok:
This PR fixes a bug that prevented files (which included an arrow) from opening correctly.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

1. Open a file that includes at least one arrow


SomeHats:
lgtm

PR #3827, opened 2024-05-24
assets: make option to transform urls dynamically / LOD

mimecuvalo:
this is take #2 of this PR https://github.com/tldraw/tldraw/pull/3764

This continues the idea kicked off in https://github.com/tldraw/tldraw/pull/3684 to explore LOD and takes it in a different direction.

Several things here to call out:
- our dotcom version would start to use Cloudflare's image transforms
- we don't rewrite non-image assets 
- we debounce zooming so that we're not swapping out images while zooming (it creates jank)
- we load different images based on steps of .25 (maybe we want to make this more, like 0.33). Feels like 0.5 might be a bit too much but we can play around with it.
- we take into account network connection speed. if you're on 3g, for example, we have the size of the image.
- dpr is taken into account - in our case, Cloudflare handles it. But if it wasn't Cloudflare, we could add it to our width equation.
- we use Cloudflare's `fit=scale-down` setting to never scale _up_ an image.
- we don't swap the image in until we've finished loading it programatically (to avoid a blank image while it loads)

TODO
- [x]  We need to enable Cloudflare's pricing on image transforms btw @steveruizok  😉  - this won't work quite yet until we do that.

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test images on staging, small, medium, large, mega
2. Test videos on staging

- [x] Unit Tests
- [ ] End to end tests

### Release Notes

- Assets: make option to transform urls dynamically to provide different sized images on demand.


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3858,6 +3869,20 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this.store.get(typeof asset === 'string' ? asset : asset.id) as TLAsset | undefined
 	}
 
+	async resolveAssetUrl(assetId: TLAssetId | null, context: { zoom: number }): Promise<string> {
+		if (!assetId) return ''
```
SomeHats, line 3873: same note here about maybe returning null

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3858,6 +3869,20 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		const asset = this.getAsset(assetId)
+		if (!asset) return ''
+
+		const networkEffectiveType: string | null =
+			'connection' in navigator ? (navigator as any).connection.effectiveType : null
+		const dpr = this.getInstanceState().devicePixelRatio
```
SomeHats, line 3879: both of these values could in theory be derived pretty easily by the developer - especially if we were to provide `editor` to `onResolveAsset`.

What would it look like to leave deriving these up to implementors? Would that be worth doing for a smaller API?

packages/editor/src/lib/editor/types/misc-types.ts
```diff
@@ -55,6 +55,18 @@ export interface TLCameraOptions {
 	constraints?: TLCameraConstraints
 }
 
+/** @public */
+export interface AssetContextProps {
+	zoom: number
```
SomeHats, line 60: instead of passing `zoom`, could we pass something like `screenScale`? This would be the page-space scale factor of the image (ie how much it has been resized from its original in page space) multiplied by the zoom level to produce how much it's resized from its original in screen space.

I think right now, we don't take into account the scaling of the image at all

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -0,0 +1,25 @@
...
+export function useAsset(assetId: TLAssetId | null) {
+	const editor = useEditor()
+	const [url, setUrl] = useState('')
+	const asset = assetId ? editor.getAsset(assetId) : null
+
+	// We debounce the zoom level to reduce the number of times we fetch a new image and,
+	// more importantly, to not cause zooming in and out to feel janky.
+	const [debouncedZoom, setDebouncedZoom] = useState(editor.getZoomLevel())
+	const zoomUpdater = useRef(debounce((zoom: number) => setDebouncedZoom(zoom), 500))
+	useValue('zoom level', () => zoomUpdater.current(editor.getZoomLevel()), [editor])
```
SomeHats, line 14: The way this is structured is a bit of a react anti-pattern:
- we create a new debounce function every render
- we don't cancel debounced calls after the function unmounts, resulting in set-state calls on unmounted components

im vaguely surprised we don't have a `useDebouncedValue` type utility hook already, but if you wanted to add one you could do something like:
```tsx
function useDebouncedValue<T>(value: T, ms: number): T {
    const [debouncedValue, setDebouncedValue] = useState(value)

    useEffect(() => {
        const timer = setTimeout(() => {
            setDebouncedValue(value)
        }, ms)
        return () => clearTimeout(timer)
    }, [value, ms])

    return value
}

// usage:
const rawZoomLevel = useValue(...)
const debouncedZoomLevel = useDebouncedValue(rawZoomLevel, 500)
```

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	useEffect(() => {
+		async function resolve() {
+			const resolvedUrl = await editor.resolveAssetUrl(assetId, { zoom: debouncedZoom })
```
SomeHats, line 18: what do you think about having this return a pre-multiplied (ie already transformed to the nearest LOD) zoom/scale value? 

I think there's a fair bit of nuance around selecting the right scale for zoom factor LODs - multiplicative vs linear scaling (ie powers of 2 instead of multiples of 0.25), rounding up vs rounding down, etc.

I wonder if we're possibly better placed than implementers to make those decisions & get that right whilst avoiding e.g. accidentally requesting a new image for every random fractional zoom level a user ends on. 

I think I'd be tempted here to pass the pre-rounded values to this function in the name of having devs fall into the pit of success. We can offer options for more fine-grained scaling down the line if we see a need for it. what do you think?

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -0,0 +1,33 @@
+import { AssetContextProps, TLAsset } from 'tldraw'
+import { ASSET_UPLOADER_URL } from './config'
+
+export async function resolveAsset(asset: TLAsset | null | undefined, context: AssetContextProps) {
+	if (!asset || !asset.props.src) return ''
```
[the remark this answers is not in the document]
mimecuvalo: good point!

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+	if (asset.type !== 'image') return ''
+
+	// Don't try to transform data: URLs, yikes.
+	if (!asset.props.src.startsWith('http:') && !asset.props.src.startsWith('https:'))
+		return asset.props.src
+
+	// N.B. navigator.connection is only available in certain browsers (mainly Blink-based browsers)
+	// 4g is as high the 'effectiveType' goes and we can pick a lower effective image quality for slower connections.
+	const networkCompensation =
+		!context.networkEffectiveType || context.networkEffectiveType === '4g' ? 1 : 0.5
```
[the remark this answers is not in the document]
mimecuvalo: yeah, it's interesting. it's based off this spec: https://wicg.github.io/netinfo/#networkinformation-interface
(see also: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType)

it definitely felt weird to me the way they designed this spec, but i think basically it's fine for now ¯\\_(ツ)_/¯ . basically, '4g' represents anything that is a 'great network'.

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	// We only look at the zoom level to the nearest 0.25
+	const zoomStepFunction = (zoom: number) => Math.floor(zoom * 4) / 4
+	const steppedZoom = Math.max(0.25, zoomStepFunction(context.zoom))
```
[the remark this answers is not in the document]
mimecuvalo: yes, good call out. definitely wasn't happy with the 0.25 setting (as mentioned in the PR description). made it be a step function based on powers of 2, much much better 🎉

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	const width = Math.ceil(asset.props.w * steppedZoom * networkCompensation)
```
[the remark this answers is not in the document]
mimecuvalo: no no, i think you misread: the check is for `!context.networkEffectiveType` - non-Blink browsers always get a multiplier of `1`.

so the _only_ case where that cames into play is: Blink browsers + 2g/3g

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: yeah, i was thinking that originally but 2 things:
- i originally (and erroneously was using just `window.devicePixelRatio` whereas i should have been using the instance state version which is more correct
- having these be present on the context give the developer ideas on what they should be thinking about when implementing (a DX reminder, if you will)
- plus it feels like better DX, sort of a nice/helpful butler of an API :)

pits of success as you mention below!

packages/editor/src/lib/editor/types/misc-types.ts
mimecuvalo: i'm gonna ask a silly question: how does one tell how much something has been resized? (original vs. current size)
i feel like i'm not seeing how to do that 🤔

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: > we create a new debounce function every render

correct me if i'm wrong here but the signature of useRef is: `useRef(initialValue)` - the argument is ignored after initial render... 🤔

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: definitely in agreement to create a pit of success and make it easier for implementers, that sounds fab.

i might be confused on where you're suggesting it should be pre-multiplied+returned. via `resolveAssetUrl`? or, are you saying some of the logic in `assetHandler` should be moved here? i definitely would be happier to have more of that logic live in the tldraw package! i guess i thought you wanted more of that stuff _out_ of the packages and less imposed upon the implementer :P

apps/dotcom/src/utils/assetHandler.ts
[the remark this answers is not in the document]
SomeHats: oh I see! that makes sense. I still think it could be worth using this with the quality setting instead though (the default is `85`, we set it to `92` for some reason)

packages/editor/src/lib/editor/Editor.ts
SomeHats: makes sense to me!

packages/editor/src/lib/editor/types/misc-types.ts
SomeHats: image shapes have a width/height property (how big is the shape) which is separate from the width/height of the asset (the natural pixel dimensions of the image). The image shape would need to tell the hook what size it's expecting to render on the page so it can combine it with the page transform

packages/tldraw/src/lib/shapes/shared/useAsset.ts
SomeHats: yup! it doesn't get used, but we do do the work of creating a new debounced function. that first bullet isn't at all a big deal other than creating garbage. i think i saw it has a bit of a smell that we were doing side-effect-y things in the render phase though (which is why we don't have cancelling rn)

packages/tldraw/src/lib/shapes/shared/useAsset.ts
SomeHats: i was suggesting that rather than just passing the zoom/scale/whatever to `onResolveAssetUrl`, we pass the version that's already been rounded to whatever scale we think makes sense

packages/editor/src/lib/editor/types/misc-types.ts
mimecuvalo: ah right right, ok - for some reason i was conflating original shape size and asset size.
i think the odd thing, slightly, is that the asset does get resized on initial load which, maybe we should rethink?

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: i see, i see, gotcha

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: ok, makes sense, gotcha. 
in that case, i wonder then if we should send two zooms : `steppedZoom` and `rawZoom`. i could see (possibly) some case where people need more fidelity then a stepped zoom.

apps/dotcom/src/utils/assetHandler.ts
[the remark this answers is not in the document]
mimecuvalo: yeah, you may be right on using `quality` - i mean, truth be told, i'm stabbing in the dark a bit at the moment.
what really needs to happen first is I need to test this on Cloudflare (after Steve enables it).

i think i was not changing `quality` to keep the surface area limited at the moment. but yeah, once i can actually test this work on Cloudflare then i'll see if `quality` is a better parameter to tweak vs. width as far as output. i'll circle back to this one!

side note: i picked `92` to keep parity with this: https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/defaultExternalContentHandlers.ts#L88

packages/editor/src/lib/editor/types/misc-types.ts
mimecuvalo: i think i wasn't clear in that last comment - what i mean to say is, maybe we nix this piece of code: https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/defaultExternalContentHandlers.ts#L84-L90

the clamping is fine - but i'm wondering if it should be set to 1.0 for quality instead of 0.92 🤔

SomeHats: So that quality param i think is a bit different to cloudflare's one. It's the format-quality for lossy formats like jpeg and webp: https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob#quality

i think the ideal thing to do here would be to check what the original quality was somehow and re-use that. for lossy formats, setting quality to 1 can cause a bit of an explosion in file size - for a random photo jpeg in my downloads, setting quality from whatever it is right now to 1 causes the file size to go from 2.3mb to 7.6mb

i do agree that we could arguably maybe nix that now though?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3885,6 +3895,24 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this.store.get(typeof asset === 'string' ? asset : asset.id) as TLAsset | undefined
 	}
 
+	async resolveAssetUrl(
+		assetId: TLAssetId | null,
+		context: { screenScale: number; steppedScreenScale: number }
```
SomeHats, line 3900: could we make the signature here just take `screenScale` and derive `steppedScreenScale` here instead?

packages/editor/src/lib/editor/types/misc-types.ts
```diff
@@ -55,6 +55,22 @@ export interface TLCameraOptions {
 	constraints?: TLCameraConstraints
 }
 
+/** @public */
+export interface AssetContextProps {
+	screenScale: number
+	steppedScreenScale: number
```
SomeHats, line 61: i keep going back and forth on whether `screen` in these names is needed/confusing..... hmmmm...........

packages/state/src/lib/react/useValueDebounced.ts
```diff
@@ -0,0 +1,33 @@
...
+import { useValue } from './useValue'
+
+/**
+ * Extracts the value from a signal and subscribes to it, debouncing the value by the given number of milliseconds.
+ *
+ * @see [[useValue]] for more information.
+ *
+ * @public
+ */
+export function useValueDebounced<Value>(
```
SomeHats, line 12: it seems like this would be simpler & more versatile if it took any value (ie from `useValue`) instead of calling `useValue` itself, like:
```ts
const rawValue = useValue(...)
const value = useDebounced(rawValue, 100)
```

what do you think?

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/** @internal */
+export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null) {
+	const editor = useEditor()
+	const [url, setUrl] = useState<string | null>(null)
+	const asset = assetId ? editor.getAsset(assetId) : null
+	const shape = editor.getShape(shapeId)
+
+	const shapeScale =
+		asset && shape && 'w' in shape.props && 'w' in asset.props ? shape.props.w / asset.props.w : 1
```
SomeHats, line 12: instead of trying to extract the scale from the shape like this (different shapes might scale assets in different ways), how to you feel about having the width / scale / whatever passed into the hook? like `useAsset(assetId, {width: 100})` or something

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -0,0 +1,38 @@
...
+	const debouncedScreenScale = useValueDebounced(
+		'zoom level',
+		() => editor.getZoomLevel() * shapeScale,
+		[editor, shapeScale],
+		500
+	)
+
+	// We only look at the zoom level at powers of 2.
+	const zoomStepFunction = (zoom: number) => Math.pow(2, Math.ceil(Math.log2(zoom)))
+	const steppedScreenScale = Math.max(0.25, zoomStepFunction(debouncedScreenScale))
```
SomeHats, line 24: why do we limit this to only being at a minimum 25% of the original size? if i upload a giant photo and then resize to like 100px would it not make sense to request it at 100px?

If anything, I would expect the other sort of limit here, so we dont request shapes over their largest size: `Math.min(1, ...)`

packages/editor/src/lib/editor/types/misc-types.ts
SomeHats: also i vaguely feel like these should be flipped (scale & rawScale, maybe?) for pit-of-success type reasons but don't feel strongly

packages/editor/src/lib/editor/types/misc-types.ts
mimecuvalo: nixed 👍

packages/editor/src/lib/editor/types/misc-types.ts
mimecuvalo: haha yeah, i feel ya on the tough choices on naming here. i could see it going either way myself, too. 🙃

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: > why do we limit this to only being at a minimum 25% of the original size? if i upload a giant photo and then resize to like 100px would it not make sense to request it at 100px?

yeah, that's fair - this bit was here before we added the screenscale (just purely zoom) and i needed to pick a lower bound (not 0) and 10% felt maybe too low. I think I'll set this to 0.125 b/c I think that's a reasonable lower bound.

> If anything, I would expect the other sort of limit here, so we dont request shapes over their largest size:

It's a good question about requesting over largest size: I _think_ that we should probably allow SDK users to do so, if they so desired. In our case, in assetHandler, this is prevented by this parameter to Cloudflare `fit=scale-down`. But I think technically, you could request a larger size and blow it up if you wanted ¯\\_(ツ)_/¯

mimecuvalo: but lemme know if you think 1 / 16 makes more sense to you, I could see an argument for that too

packages/state/src/lib/react/useValueDebounced.ts
mimecuvalo: The impetus here stemmed from the comment on the original PR from Steve (albeit in a little bit of a different form): https://github.com/tldraw/tldraw/pull/3764#discussion_r1605922639
It felt like maybe this would be a useful pattern to have for reactors.

Like, I don't disagree, per se, about splitting them up and then composing them. But I think it might be a useful primitive DX-wise to add alongside `useValue`

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: yeah, that's fair - i'm sorta ambivalent about it, will change up.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
SomeHats: why do we need a minimum size at all? For very large images I sort of want to be able to scale them down as low as I like if they're only ever going to be made tiny. I could maybe see an argument for a dynamic minimum size - like minimum = 100px largest dimension? but i think it's hard to pick an arbitrary multiplyer here as we don't know how big the images are or how they'll be used

packages/state/src/lib/react/useValueDebounced.ts
SomeHats: it almost sort of feels like the wrong level for a signia specific thing - like, i could imagine a function that debounces a signal being valuable as a signia primitive? but it feels a little weird to me to have this as a specific signia hook, especially as it forces quite an awkward signature compared to how most hooks-y code looks. happy to leave as is if you're sure tho!

apps/dotcom/src/utils/assetHandler.ts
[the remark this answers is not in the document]
mimecuvalo: now that i can play around with this it looks `quality` doesn't behave quite how we'd expect; for that reason, we should stick with tweaking `width` instead.

example image:
- 1000px width, quality 92, 604kb: https://assets.tldraw.xyz/cdn-cgi/image/width=1000,dpr=2,fit=scale-down,quality=92/https://assets.tldraw.xyz/uploads/ypeqp-SgVcRYsA18ccwMF-Screen-Shot-2021-08-16-at-8.48.36-AM.png
- 500px width, quality 92, 176kb: https://assets.tldraw.xyz/cdn-cgi/image/width=500,dpr=2,fit=scale-down,quality=92/https://assets.tldraw.xyz/uploads/ypeqp-SgVcRYsA18ccwMF-Screen-Shot-2021-08-16-at-8.48.36-AM.png

---

- 1000px width, quality 85, 604kb (same as quality 92!): https://assets.tldraw.xyz/cdn-cgi/image/width=1000,dpr=2,fit=scale-down,quality=85/https://assets.tldraw.xyz/uploads/ypeqp-SgVcRYsA18ccwMF-Screen-Shot-2021-08-16-at-8.48.36-AM.png
- 1000px width, quality 60, 369kb: https://assets.tldraw.xyz/cdn-cgi/image/width=1000,dpr=2,fit=scale-down,quality=60/https://assets.tldraw.xyz/uploads/ypeqp-SgVcRYsA18ccwMF-Screen-Shot-2021-08-16-at-8.48.36-AM.png
- 1000px width, quality 30, 229kb: https://assets.tldraw.xyz/cdn-cgi/image/width=1000,dpr=2,fit=scale-down,quality=30/https://assets.tldraw.xyz/uploads/ypeqp-SgVcRYsA18ccwMF-Screen-Shot-2021-08-16-at-8.48.36-AM.png

packages/state/src/lib/react/useValueDebounced.ts
mimecuvalo: forgot to respond to this: can't say i'm _entirely_ sure but i think there's probably room for both ideas! cc @ds300 
i think debouncing a signal as a signia primitive is something we should also add. but also, this composed hook feels like a useful additive thing.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: forgot to respond to this: i think let's see how this plays out as we test this on staging/real world. i think it's diminishing returns to make it go below a very low size. but let's test and assess!

packages/state/src/lib/react/useValueDebounced.ts
ds300: happy to merge this as `@internal` for now. Agree that a debounced thing makes sense to offer but i think it would make more sense to be an option for `useValue` and then be implemented by debouncing the effect scheduler, so that only the debounced updates actually trigger rerenders.

mimecuvalo: ah just seeing your comment - i'll mark this as `@internal` in a separate PR 👍

PR #3835, opened 2024-05-27
Dynamic size mode + fill fill

steveruizok:
This PR adds a user preference for "dynamic size mode" where the scale of shapes (text size, stroke width) is relative to the current zoom level. This means that the stroke width in screen pixels (or text size in screen pixels) is identical regardless of zoom level.

Kapture 2024-05-27 at 05 23 21

- [x] Draw shape
- [x] Text shape
- [x] Highlighter shape
- [x] Geo shape
- [x] Arrow shape
- [x] Note shape
- [x] Line shape

Embed shape?

### Change Type

- [x] `feature` — New feature

### Test Plan

1. Use the tools.
2. Change zoom

- [ ] Unit Tests

### Release Notes

- Adds a dynamic size user preferences.
- Removes double click to reset scale on text shapes.
- Removes double click to reset autosize on text shapes.


mimecuvalo:
i feel like i'm trippin' watchin that gif 🥴

steveruizok:
This was originally a customer request btw! Related to annotating very large / detailed diagrams. I originally balked because we would have to make significant changes to the way that our shapes rendered, and couldn't guarantee that other shapes also knew how to participate in this system. I'm now of the opinion however that it doesn't really matter, and that a "works for our shapes" solution is still very useful here.

packages/tldraw/src/lib/shapes/geo/components/GeoShapeBody.tsx
```diff
@@ -209,14 +216,19 @@ export function GeoShapeBody({ shape }: { shape: TLGeoShape }) {
...
 					</>
 				)
 			}
 		}
 		default: {
-			const geometry = editor.getShapeGeometry(shape)
+			let geometry = editor.getShapeGeometry(shape)
+			if (!shouldScale) {
+				// We don't want the cached geometry
+				geometry = editor.getShapeUtil(shape).getGeometry(shape)
```
Taha-Hassan-Git, line 241: I don't love this, but couldn't come up with a better way

packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
```diff
@@ -219,10 +231,10 @@ function LineShapeSvg({ shape }: { shape: TLLineShape }) {
 			const pathData = 'M' + outline[0] + 'L' + outline.slice(1)
 
 			return (
-				<>
+				<g transform={`scale(${scale})`}>
```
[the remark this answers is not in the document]
steveruizok: Hm, not sure why lines had shapefill elements at all. Moving the scale down to the regular paths where possible.

packages/tldraw/src/lib/shapes/geo/components/GeoShapeBody.tsx
steveruizok: Why would the cached geometry be wrong here?

PR #3836, opened 2024-05-28
assets: store in indexedDB, not as base64

mimecuvalo:
this is take #2 of this PR https://github.com/tldraw/tldraw/pull/3745

As I look at LOD holistically and whether we have multiple sources when working locally, I learned that our system used base64 encoding of assets directly. Issue https://github.com/tldraw/tldraw/issues/3728

assetstore

The motivations and benefits are:
- store size: not having a huge base64 blobs injected in room data
- perf on loading snapshot: this helps with loading the room data more quickly
- multiple sources: furthermore, if we do decide to have multiple sources locally (for each asset), then we won't get a multiplicative effect of even larger JSON blobs that have lots of base64 data in them
- encoding/decoding perf: this also saves the (slow) step of having to base64 encode/decode our assets, we can just strictly with work with blobs.

Todo:
- [x] decodes video and images
- [x] make sure it syncs to other tabs
- [x] make sure it syncs to other multiplayer room
- [x] fix tests

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test the shit out of uploading/downloading video/image assets, locally+multiplayer.

- [ ] Need to fix current tests and write new ones

### Release Notes

- Assets: store as reference to blob in indexedDB instead of storing directly as base64 in the snapshot.


SomeHats:
> I still feel strongly that it should be the default behavior

It sounds like this is the thing motivating a lot of the design decisions you're uncertain about here, so I think we should settle this one before we go any further.

my principle here is that with our defaults, we're optimising for a mix of "do the right thing" and "time to first prototype".

For "do the right thing", if you want to store tldraw content locally, then this is for sure the way to do it!  However: that's not our current default. Our current default is that tldraw is ephemeral. We store assets as base64 strings in memory. Not ideal for perf (we could do in-memory blobs i guess), but in that specific case i think the ephemerality is the most important property. 

**Opinion 1:** We should not store _anything_ in the user's browser when using the `<Tldraw />` component without any other storage/sync stuff specified.

If you want to store tldraw content locally, you can pass the `persistenceKey` prop to the tldraw component which switches on all our local storage stuff. This should probably use this new approach, and would be the appropriate place to add this to the library IMO.

For "time to first prototype", we're looking at how someone can get something together quickly that they can show stakeholders at their org so they turn into a tldraw customer. This is why our yjs stuff exists: it's not really good for production, but in <30mins you can take it and have a working multiplayer demo. This is a big deal! We're working on making this even faster/easier soon. One thing i wouldn't want: to force devs to set up external asset handling before they can get to that first initial prototype.

**Opinion 2:** By default, everything needed to render a tldraw document should live in the store for speed of prototyping.

tldr: imo this shouldn't be the default, but it should be part of the local sync system that gets enabled when you set a `persistenceKey` prop.

mimecuvalo:
Great points! I think that makes it clearer in my head about whether the persistence key is present or not. 

I totally agree about Option 1 as you listed out. I'll grab you when you're back in the office before I move forward (or maybe if you're around when traveling) and we can make sure we're on the same page before I tweak this further 👍

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -90,24 +92,49 @@ export function registerDefaultExternalContentHandlers(
...
+			return asset!
+		}
 
-		return asset
+		assetInfo.props.src = await FileHelpers.blobToDataUrl(file)
+		return AssetRecordType.create(assetInfo)
 	})
 
+	if (persistenceKey) {
+		editor.registerExternalAssetHandler('blob', async ({ assetInfo, blob }) => {
```
SomeHats, line 125: ok so i don't know if i can do a good job explaining this in a comment - half tempted to sketch what i mean in code, but also happy to try pairing.

in order to plumb this in, we've had to add a special new external asset handler, that's only ever called from one of _our_ external asset handlers. This is probably something I think we should avoid - I think these asset handlers are already pretty confusing, and adding another type to support this seems unnecesary.

The thing that I've been imagining here is that in the Tldraw component, we have some code that looks like this:
```ts
let onMount = props.onMount
if (props.persistenceKey) {
    onMount = (editor) => {
      editor.registerExternalAssetHandler('file', () => {
        // do the stuff from the default handler
        // but also do the blobby stuff
      })
      return props.onMount?.(editor)
    }
}

return <WhateverIsNext {...props} onMount={onMount} />
```

(i also imagined plumbing in the resolve asset stuff the same way - so rather than having `persistenceKey` go deep into these fns, it sort of stops at that component level)

I think this would help keep the APIs around these handlers a bit easier to conceptualise & understand, but I appreciate that this is something we've been plugging away at for ages now - if you wanted to land this as is then that's all good too

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -555,3 +582,28 @@ export function createEmptyBookmarkShape(
...
+
+		// Retrieve a local image from the DB.
+		if (persistenceKey && asset.props.src.startsWith('asset:')) {
+			const blob = await getAssetFromIndexedDb({
+				persistenceKey,
+				assetId: asset.id,
+			})
+			if (blob) {
+				return URL.createObjectURL(blob)
+			}
```
SomeHats, line 604: ditto here re: caching

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -555,3 +582,28 @@ export function createEmptyBookmarkShape(
 
 	return editor.getShape(partial.id) as TLBookmarkShape
 }
+
+export const defaultResolveAsset =
+	(persistenceKey?: string) => async (asset: TLAsset | null | undefined) => {
+		if (!asset || !asset.props.src) return ''
```
SomeHats, line 588: should these be `null`?

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -204,11 +204,16 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
...
-		const asset = shape.props.assetId ? this.editor.getAsset(shape.props.assetId) : null
+		if (!shape.props.assetId) return null
+
+		const asset = this.editor.getAsset(shape.props.assetId)
 
 		if (!asset) return null
 
-		let src = asset?.props.src || ''
+		let src = await this.editor.resolveAssetUrl(shape.props.assetId, {
+			shouldResolveToOriginalImage: true,
```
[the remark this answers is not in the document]
mimecuvalo: Mmm, kinda. I think the original image isn't necessarily at `scale: 1`. You can have a very large image and at scale: 1, it's only 1/4 of its size. I think when we export, we do want to force it return the original image.

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -1,29 +1,47 @@
...
+
+		// Retrieve a local image from the DB.
+		if (persistenceKey && asset.props.src.startsWith('asset:')) {
+			const blob = await getAssetFromIndexedDb({
+				assetId: asset.id,
+				persistenceKey,
+			})
+			if (blob) {
+				return URL.createObjectURL(blob)
+			}
```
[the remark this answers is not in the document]
mimecuvalo: nice one - yes, great idea 👍 

1. slight nuance to point #1 in your list that it's debounced so it's not _that_ often (it waits until the zooming settles) but your point is still taken about caching.

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
mimecuvalo: No worries and pushing this to get better :) Definitely wanna get this right. 

I _think_ I kinda understand what you mean but maybe I'm confused on how onMount comes into play here (because it feels like onMount might be too late for the initial load? or maybe I'm wrong).

Yeah, I think it might be worth a pairing. I'll grab you when you have time!

mimecuvalo: talked offline with Alex about this and simplified 👍

steveruizok:
Bug:
- 1. Add an image to the canvas on one project (i.e. develop)
- 2. Copy the image shapes
- 3. Visit a different tldraw project (i.e. the persistence key example)
- 4. Paste the images shapes

steveruizok:
It looks like there's work missing in `Editor.putContentOntoCurrentPage`.

When a user pastes new assets onto the page, we:
- get the image assets
- create new files from their source
- handle the files as new assets

The step that was missing seems to be creating a file from an `asset:` src.

mimecuvalo:
> Is this one ready to go?

Not ready to go in general because, as mentioned before, it's parented on top of the Cloudflare PR.
We need to enable Cloudflare's image transformation and then we can move forward with both PRs.

> When a user pastes new assets onto the page, we:

I'm looking into it! I thought this had worked in the first version but maybe it regressed with the rewrites that happened.

steveruizok:
> Not ready to go in general because, as mentioned before, it's parented on top of the Cloudflare PR. We need to enable Cloudflare's image transformation and then we can move forward with both PRs.

Right on, thought it was the other way around.

SomeHats:
re: clipboard, this feels very similar to https://discord.com/channels/859816885297741824/1242825747366936679/1242825747366936679

i wonder if there's a solution that would work for both?

also, could we unblock these PRs by updating the LOD one to add the APIs but not actually switch on cloudflare yet?

mimecuvalo:
Fixed pasting but this became a real red / black situation: https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/

Had to make a bunch of things with async/await.

mimecuvalo:
Talked to Alex offline and am going to rework the red/black kerfuffle here. We'll make copy/paste async but avoid making move/duplicate async.

mimecuvalo:
Er, red/blue, you know what I mean 🙃

mimecuvalo:
ok, one more time, with feeling. double checked these all:
- [x] copy
- [x] export as svg
- [x] export as png
- [x] export as json
- [x] tldr file

@SomeHats / @steveruizok lemme know if you can take another look!

SomeHats (approved):
looking good!

PR #3852, opened 2024-05-30
editor: register timeouts/intervals/rafs for disposal

mimecuvalo:
We have a lot of events that fire in the editor and, technically, they can fire after the Editor is long gone.
This adds a registry/manager to track those timeout/interval/raf IDs (and some eslint rules to enforce it).

Some other cleanups:
- `requestAnimationFrame.polyfill.ts` looks like it's unused now (it used to be used in a prev. revision)
- @ds300 I could use your feedback on the `EffectScheduler` tweak. in `useReactor` we do: `() => new EffectScheduler(name, reactFn, { scheduleEffect: (cb) => requestAnimationFrame(cb) }),` 
   and that looks like it doesn't currently get disposed of properly. thoughts? happy to do that separately from this PR if you think that's a trickier thing.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test async operations and make sure they don't fire after disposal.

### Release Notes

- Editor: add registry of timeouts/intervals/rafs

### DIFF
--- a/packages/tldraw/src/lib/utils/export/export.ts
+++ b/packages/tldraw/src/lib/utils/export/export.ts
@@ -13,7 +13,6 @@ import { TLExportType } from './exportAs'
 export async function getSvgAsImage(
 	editor: Editor,
 	svgString: string,
-	isSafari: boolean,
 	options: {
 		type: 'png' | 'jpeg' | 'webp'
 		quality: number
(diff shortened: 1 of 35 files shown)


packages/tldraw/src/lib/utils/export/export.ts
steveruizok, line 1: If we're going to change the arguments here, let's move isSafari into the object bag too

steveruizok (approved):

packages/tldraw/src/lib/utils/export/export.ts
steveruizok: nm, we can remove this argument

PR #3856, opened 2024-05-31
bookmark: fix up double request and rework extractor

mimecuvalo:
This code has started to bitrot a bit and this freshens it up a bit.
- there's a double request happening for every bookmark paste at the moment, yikes! One request originates from the paste logic, and the other originates from the `onBeforeCreate` in `BookmarkShapeUtil`. They both see that an asset is missing and race to make the request at the same time. It _seems_ like we don't need the `onBeforeCreate` anymore. But, if I'm mistaken on some edge case here lemme know and we can address this in a different way.
- the extractor is really crusty (the grabity code is from 5 yrs ago and hasn't been updated) and we don't have control over it. i've worked on unfurling stuff before with Paper and my other projects and this reworks things to use Cheerio, which is a more robust library.
- this adds `favicon` to the response request which should usually default to the apple-touch-icon. this helps with some better bookmark displays (e.g. like Wikipedia if an image is empty)

In general, this'll start to make this more maintainable and improvable on our end.

Double request:
Screenshot 2024-05-31 at 17 54 49

Before:
Screenshot 2024-05-31 at 17 55 02

After:
Screenshot 2024-05-31 at 17 55 44

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test pasting links in, and pasting again.

### Release Notes

- Bookmarks: fix up double request and rework extractor code.


steveruizok:
Maybe there's something better than the blank box? Dark grey globe icon or somethiing?

ds300:
the figjam one is the best of those IMO.

@mimecuvalo not sure if you figured this out already but due to laziness (or thriftiness, whatever you prefer) we don't have any deployment automation set up for the bookmark extractor app. It's just a case of making (backwards compatible!) changes and then doing a manual deploy using the vercel cli.

SomeHats:
(would it be worth moving bookmark extractor into `dotcom-worker` to simplify things a bit?)

ds300:
we investigated that when doing the SPA work, there were a couple of things working against it:

1. it would balloon the size of the worker script by like 500kb, which would make startup time slower (though by how much we didn't measure. the cloudflare docs seemed to imply it would be significant)
2. most of the off-the-shelf tools we found for extracting metadata from URLs didn't run in the worker runtime due to dependencies on node stuff

mimecuvalo:
reworked this with the feedback:

Screenshot 2024-06-05 at 16 36 41

steveruizok:
image

Handling a few more cases!

steveruizok:
@mimecuvalo we'll need to add a test for the new migration here

mimecuvalo:
> @mimecuvalo we'll need to add a test for the new migration here

sick, nice cleanup — thank you! good call on the `clamp` and fallback link icon.

i'll add the test 👍

mimecuvalo:
this is lookin' great - merging!

PR #3862, opened 2024-06-03
[bugfix] Preserve redo stack when selection changes

ds300:
This PR identifies a bunch of places where the redo stack was being blown when it shouldn't have been. I think this is going to be a constant footgun unless we switch back to `ignore-by-default` behavior for the store.

I also ignored changes to assets, which brings them back to how they were being handled before the undo/redo change (i.e. they lived outside of the undo/redo system). 

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -1871,7 +1871,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	setHoveredShape(shape: TLShapeId | TLShape | null): this {
 		const id = typeof shape === 'string' ? shape : shape?.id ?? null
 		if (id === this.getHoveredShapeId()) return this
-		this.updateCurrentPageState({ hoveredShapeId: id })
+		this.updateCurrentPageState({ hoveredShapeId: id }, { history: 'ignore' })
 		return this
 	}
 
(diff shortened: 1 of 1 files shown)


mimecuvalo:
> I think this is going to be a constant footgun unless we switch back to ignore-by-default behavior for the store.

mimecuvalo:
> I think this is going to be a constant footgun unless we switch back to ignore-by-default behavior for the store.

I would wonder how many places are doing ignore vs. not. My gut instinct would say ignore-by-default might not be what's expected by most consumers of the library. It's probably worth a discussion though.

I brought this up in some other thread on Discord or some PR but: there could be a design where there are high-level calls to the editor that do kind of what `batch` does currently. In Paper we called it the `execute` function and it would basically create a transaction and do the right thing for the history/store/etc. However, if you _didn't_ want to use the high-level function and wanted to compose a couple different custom operations (especially ones that wanted to avoid the history stack) then you could still access call the API outside of an `execute` "call stack", as it were, and put together what you wanted to happen.

What I mean to say is: I think it's less about what the right default for the history but more about how we present it to the consumer. They probably shouldn't even be _thinking_ about history by default at the high-level! We should be taking care of that for them.

PR #3871, opened 2024-06-04
Bindings onBeforeShapeIsolate?

ds300:
So we were kinda bending over backwards to capture the use case where we update the arrow's terminal x,y coords when unbinding, copy-pasting, and duplicating. 

- At first we abused the `onBeforeShapeDelete` callbacks, but that was footgunny.
- Then we created a `onBeforeUnbind` callback, which was less footgunny but still subtly footgunny.

This PR proposes reverting the `onBeforeUnbind` stuff, taking us back to having `onBeforeShapeDelete` stuff. But at the same time it adds `onBeforeShapeIsolate` callbacks which are triggered at the following times:

- When you delete the other shape in a bound shape pair
- When you copy/paste or duplicate one shape in a bound shape pair but not the other one
- When you opt-in while deleting bindings e.g. `deleteBindings([...], {isolateShapes: true})`

This PR also fixes the bound arrow drag interaction. We can probably extract that out to a separate PR if needed.

Kapture 2024-06-04 at 12 42 40

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/editor/src/lib/options.ts
```diff
@@ -34,7 +34,7 @@ export interface TldrawOptions {
 	readonly collaboratorInactiveTimeoutMs: number
 	readonly collaboratorIdleTimeoutMs: number
 	readonly collaboratorCheckIntervalMs: number
-	readonly cameraMovingTimoutMs: number
```
SomeHats, line 37: why this change? should we apply this to the others?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8943,36 +8949,36 @@ function pushShapeWithDescendants(editor: Editor, id: TLShapeId, result: TLShape
...
+						bindingsWithBoth.add(binding.id)
+						continue
+					}
+					if (!hasFrom || !hasTo) {
+						bindingsToRemove.add(binding.id)
+					}
+				}
+			}
+
+			editor.deleteBindings([...bindingsToRemove], { isolateShapes: true })
```
SomeHats, line 8981: thoughts on calling the onBeforeIsolate callbacks directly here rather than adding this as a special option to `deleteBindings`?

SomeHats (commented):
looking good!

packages/editor/src/lib/options.ts
ds300: just a driveby typo fix `Timout` -> `Timeout`

packages/editor/src/lib/editor/Editor.ts
ds300: it doesn't work, the bindings need to actually be deleted in between calls to onBeforeIsolate otherwise the arrow geometries are incorrect when calculating the new bend.

packages/editor/src/lib/options.ts
ds300: ah sorry seems i deleted the `Ms` bit by accident

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5247,13 +5242,24 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-		this.store.remove(ids)
+		if (isolateShapes) {
+			for (const id of ids) {
+				const binding = this.getBinding(id)
+				if (!binding) continue
+				const util = this.getBindingUtil(binding)
+				util.onBeforeIsolateFromShape?.({ binding, shape: this.getShape(binding.fromId)! })
+				util.onBeforeIsolateToShape?.({ binding, shape: this.getShape(binding.toId)! })
+				this.store.remove([id])
+			}
```
SomeHats, line 5255: let's wrap this is `this.store.atomic` so that the `after` callbacks happen in the same way they do even without setting `isolateShapes`

SomeHats (approved):
lgtm

packages/editor/src/lib/editor/Editor.ts
ds300: Hm that cause bugs if the binding's onAfterDelete does some important cleanup that the `onBeforeIsolate` callbacks need to function correctly. Is there any danger of something going wrong how it is?

SomeHats: If that's a source of bugs, then surely it would be a source of bugs for normal deletion too, right? When doing normal 1-sided deletion, these callbacks still get called, but the after-phase callbacks won't happen until everything has been deleted. I think you're more likely to end up with bugs as a result of two different calling patterns than a uniform one, considering that's how all the other after callbacks work.

ds300: huh yeah good point. alright I'll make this atomic and we can revisit the whole business if anybody complains.

PR #3875, opened 2024-06-04
Editor.blur method

ds300:
This PR adds a `editor.blur()` method to complement the `editor.focus()` method, and enhances both with an options param that allows to skip dispatching a focus/blur event on the container.

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
+++ b/apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
@@ -1,13 +1,31 @@
 import { useRef } from 'react'
 import { Editor, Tldraw } from 'tldraw'
 import 'tldraw/tldraw.css'
+import './editor-focus.css'
 
 export default function EditorFocusExample() {
+	return (
+		<div style={{ padding: 32 }}>
+			<ControlledFocusExample />
+			<p>
+				You should be able to type in this text input without worrying about triggering editor
+				shortcuts even when the editor is focused.
+			</p>
+			<input type="text" placeholder="Test me" />
+
+			<hr />
+			<FreeFocusExample />
+		</div>
+	)
+}
+
+function ControlledFocusExample() {
 	const editorRef = useRef<Editor | null>(null)
 
 	return (
-		<div style={{ padding: 32 }}>
+		<>
 			<div>
+				<h2>Controlled Focus</h2>
 				<div style={{ display: 'flex', gap: 4 }}>
 					<input
 						id="focus"
@@ -38,11 +56,36 @@ export default function EditorFocusExample() {
 					}}
 				/>
 			</div>
+		</>
+	)
+}
+
+function FreeFocusExample() {
+	const editorRef = useRef<Editor | null>(null)
+
+	return (
+		<div>
+			<h2>Free Focus</h2>
 			<p>
-				You should be able to type in this text input without worrying about triggering editor
-				shortcuts even when the editor is focused.
+				You can use `onBlur` and `onFocus` to control the editor's focus so that it behaves like a
+				native form input.
 			</p>
-			<input type="text" placeholder="Test me" />
+			<div
+				style={{ width: 800, maxWidth: '100%', height: 500 }}
+				onBlur={() => {
+					editorRef.current?.blur()
+				}}
+				onFocus={() => {
+					editorRef.current?.focus()
+				}}
+			>
+				<Tldraw
+					autoFocus={false}
+					onMount={(editor) => {
+						editorRef.current = editor
+					}}
+				/>
+			</div>
 		</div>
 	)
 }
--- a/apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
+++ b/apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
@@ -1,34 +1,23 @@
-import { useEffect, useRef, useState } from 'react'
+import { useRef } from 'react'
 import { Editor, Tldraw } from 'tldraw'
 import 'tldraw/tldraw.css'
 
 export default function EditorFocusExample() {
-	const [focused, setFocused] = useState(false)
-	const rEditorRef = useRef<Editor | null>(null)
-
-	useEffect(() => {
-		const editor = rEditorRef.current
-		if (!editor) return
-		editor.updateInstanceState({ isFocused: focused })
-	}, [focused])
+	const editorRef = useRef<Editor | null>(null)
 
 	return (
-		<div
-			style={{ padding: 32 }}
-			onPointerDown={() => {
-				const editor = rEditorRef.current
-				if (editor && editor.getInstanceState().isFocused) {
-					editor.updateInstanceState({ isFocused: false })
-				}
-			}}
-		>
+		<div style={{ padding: 32 }}>
 			<div>
 				<div style={{ display: 'flex', gap: 4 }}>
 					<input
 						id="focus"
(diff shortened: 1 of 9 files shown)


apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
```diff
@@ -1,34 +1,23 @@
...
 
 	return (
-		<div
-			style={{ padding: 32 }}
-			onPointerDown={() => {
-				const editor = rEditorRef.current
-				if (editor && editor.getInstanceState().isFocused) {
-					editor.updateInstanceState({ isFocused: false })
-				}
-			}}
```
ds300, line 23: I don't understand why this existed. It makes for some pretty damn confusing behavior in this example IMO.

ds300:
@steveruizok I added a couple of sneaky changes in here that I could use your 👍🏼 / 👎🏼  on.

1. The editor container was given a subtle outline on the css :focus-within selector. However, this was not kept in sync with the actual editor.instanceState.isFocused property, and in our focus example it was possible to get the outline while the editor was not actually focused. That felt wrong to me so I switched the strategy here to use a modifier class `.tl-container__focused` that gets added and removed by the focus manager instead.
2. Our hotkeys usage was a little bit problematic for pages with multiple tldraw editors. We were using hotkeys' "scopes" feature to group and later destroy our kbd shortcut callbacks. Unfortunately hotkeys only allows one scope to be active at a time and will not revert to the previous scope if you destroy a scope. This can leave editors in an unresponsive (to kbd) state if something like the following happens: 
   - editor A focuses
   - editor B focuses and then unfocuses
   - now editor A's keyboard shortcuts don't work but it's still in the focused state.
  To 'fix' this I just ditched the scope stuff and made sure we're calling `hotkeys.unbind` correctly. This also allows multiple editors on the same page to react to keyboard shortcuts simultaneously, for fun i guess.

apps/examples/src/examples/editor-focus/EditorFocusExample.tsx
mimecuvalo: hah yeah agreed 🙃

packages/editor/src/lib/editor/managers/FocusManager.ts
```diff
@@ -21,6 +21,7 @@ export class FocusManager {
 			(prev, next) => {
 				if (prev.isFocused !== next.isFocused) {
 					next.isFocused ? this.focus() : this.blur()
+					this.updateContainerClass()
```
mimecuvalo, line 24: feels like this should be called within `focus`/`blur`, especially because `focusContainer: false` would not focus the container in the Editor API?

packages/editor/src/lib/editor/managers/FocusManager.ts
```diff
@@ -29,6 +30,18 @@ export class FocusManager {
 		if (autoFocus !== currentFocusState) {
 			editor.updateInstanceState({ isFocused: !!autoFocus })
 		}
+		this.updateContainerClass()
```
mimecuvalo, line 33: same here

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8271,21 +8271,69 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 	 */
-	focus(): this {
-		this.focusManager.focus()
-		this.updateInstanceState({ isFocused: true }, { history: 'ignore' })
+	blur({ blurContainer = true } = {}): this {
+		if (!this.getIsFocused()) return this
+		if (blurContainer) {
+			this.focusManager.blur()
+		} else {
+			this.complete() // stop any interaction
```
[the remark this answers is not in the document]
ds300: `editor.complete()` is called by `focusManager.blur()`, which totally makes sense. And even if `blurContainer` is set to `false` it still totally makes sense to call `editor.complete()`, and it would be weird inconsistent behaviour not to IMO.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8271,21 +8271,69 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 *
+	 * @example
+	 * ```ts
+	 * editor.focus({ focusContainer: false })
+	 * ```
+	 *
+	 * @public
+	 */
+	focus({ focusContainer = true } = {}): this {
+		if (focusContainer) {
```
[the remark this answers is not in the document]
ds300: Yeah it's niche but you might be calling `.focus` and `.blur` from code paths that have already essentially handled the 'focus' and 'blur' logic, and where dispatching events may trigger other event handlers higher in the tree that you might not want to be triggered because it might duplicate or trample work you've already done.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8271,21 +8271,69 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-		this.focusManager.focus()
-		this.updateInstanceState({ isFocused: true }, { history: 'ignore' })
+	blur({ blurContainer = true } = {}): this {
+		if (!this.getIsFocused()) return this
+		if (blurContainer) {
+			this.focusManager.blur()
+		} else {
+			this.complete() // stop any interaction
+		}
+		this.updateInstanceState({ isFocused: false })
```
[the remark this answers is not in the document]
ds300: it's the default behavior for `updateInstanceState`

packages/editor/src/lib/editor/managers/FocusManager.ts
ds300: not sure I follow. this code is intended to keep the container classList in sync with the `instanceState.isFocused` property. It should be triggered whenever that property changes, which is what this guarantees.

packages/editor/src/lib/editor/managers/FocusManager.ts
ds300: see other comment. this line just makes sure the initial classList for the container is correct.

packages/editor/src/lib/editor/Editor.ts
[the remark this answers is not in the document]
mimecuvalo: wait, no, i'm confused 🙃 
in this PR: https://github.com/tldraw/tldraw/pull/3862/files
i was of the mind that history: ignore is _not_ the default?

packages/editor/src/lib/editor/Editor.ts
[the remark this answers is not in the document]
mimecuvalo: i guessss but i can't think of any other SDK/API that works like this. feels like a footgun in that i might get into a half-unfocused state 🤔

packages/editor/src/lib/editor/managers/FocusManager.ts
mimecuvalo: I think the confusion is sort of related to the `focusContainer` and how the container is where we set the classList.
so, if you manually do _not_ want the default behavior of `focusContainer: true` then, well, i guess the classList shouldn't change, because you're in theory explicitly not focusing the container, right?

mimecuvalo (approved):
talked offline and it makes more sense to me now. thanks @ds300 !

PR #3881, opened 2024-06-05
security: don't send referrer paths for images and bookmarks

mimecuvalo:
We're currently sending `referrer` with path for image/bookmark requests. We shouldn't do that as it exposes the rooms to other servers.

## ``
- `` tags have the right referrerpolicy to be `strict-origin-when-cross-origin`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#referrerpolicy
- _however_, because we use React, it looks like react creates a raw DOM node and adds properties one by one and it loses the default referrerpolicy it would otherwise get! So, in `BookmarkShapeUtil` we explicitly state the `referrerpolicy`
- `background-image` does the right thing 👍 
- _also_, I added this to places we do programmatic `new Image()`

## `fetch`
- _however_, fetch does not! wtf. https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

it's almost a footnote in this section of the docs (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options) that `no-referrer-when-downgrade` is the default.

## `new Image()`
ugh, but _also_ doing a programmatic `new Image()` doesn't do the right thing and we need to set the referrerpolicy here as well

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Test on staging that referrer with path isn't being sent anymore.

### Release Notes

- Security: fix referrer being sent for bookmarks and images.

### DIFF
--- a/apps/dotcom/index.html
+++ b/apps/dotcom/index.html
@@ -5,6 +5,7 @@
 
 		<link rel="manifest" href="/manifest.webmanifest" />
 
+		<meta name="referrer" content="strict-origin-when-cross-origin" />
 		<meta name="theme-color" content="#FFFFFF" data-rh="true" />
 		<meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
 		<link rel="icon" type="image/svg+xml" href="/favicon.svg" />
--- a/apps/dotcom/src/utils/assetUrls.ts
+++ b/apps/dotcom/src/utils/assetUrls.ts
@@ -17,6 +17,7 @@ function preloadIcon(url: string) {
 		;(image as any).fetchPriority = 'low'
 		image.onload = resolve
 		image.onerror = reject
+		image.referrerPolicy = 'strict-origin-when-cross-origin'
 		image.src = url
 	})
 }
--- a/apps/dotcom/src/utils/createAssetFromFile.ts
+++ b/apps/dotcom/src/utils/createAssetFromFile.ts
@@ -18,6 +18,7 @@ export async function createAssetFromFile({ file }: { type: 'file'; file: File }
 	await fetch(url, {
 		method: 'POST',
 		body: file,
+		referrerPolicy: 'strict-origin-when-cross-origin',
 	})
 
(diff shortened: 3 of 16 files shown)


ds300:
it looks like we can specify this in an html meta tag (and optionally an http response header), should we go with that instead?

→ `apps/dotcom/index.html`
image

SomeHats:
If we can do that meta tag approach that could be a good idea! I wonder if we need to still do something like this for library users though or need to suggest they include that meta tag themselves?

If we end up going with the approach here and you'd like to ban using these constructs without using a safer referrer policy, the pattern we usually employ here is to add a safe-by-default helper to `@tldraw/utils`, then add lint rules (see `no-restricted-globals` and `no-restricted-syntax` for examples) to disable access to the built-in, unsafe version. I can help out with that if needed!

Also: wow it's hard to find information about what actually goes on here. mdn is very lacking, the spec seems to imply that `strict-origin-when-cross-origin` is always the default... I'd guess that some cross-browser testing would be good here as it seems like maybe different browsers might be at different points with supporting the updated spec

mimecuvalo:
> If we can do that meta tag approach that could be a good idea! I wonder if we need to still do something like this for library users though or need to suggest they include that meta tag themselves?

yeah, good find @ds300 - i think we should do both. there's the dotcom usage but also, I _would_ worry that this is a security hole for SDK users as well as they won't have direct control over this potentially.

> If we end up going with the approach here and you'd like to ban using these constructs without using a safer referrer policy, the pattern we usually employ here is to add a safe-by-default helper to @tldraw/utils, then add lint rules (see no-restricted-globals and no-restricted-syntax for examples) to disable access to the built-in, unsafe version. I can help out with that if needed!

definitely. yeah, it wasn't as easy as doing the `no-restricted-globals` thing, we'll have to do something more custom to walk the AST. fuuuun. i think we should do that for a followup.

ds300 (approved):

PR #3884, opened 2024-06-05
security: enforce use of our fetch function and its default referrerpolicy

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/3881 to enforce this in the codebase

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -91,11 +91,11 @@ module.exports = {
 				'no-restricted-globals': [
 					'error',
 					{
-						name: 'setInterval',
+						name: 'setTimeout',
 						message: 'Use the timers from @tldraw/util instead.',
 					},
 					{
-						name: 'setTimeout',
+						name: 'setInterval',
 						message: 'Use the timers from @tldraw/util instead.',
 					},
 					{
@@ -119,6 +119,27 @@ module.exports = {
 						object: 'window',
 						property: 'requestAnimationFrame',
 						message: 'Use the timers from @tldraw/util instead.',
+					},
+				],
+			},
+		},
+
+		{
+			files: ['apps/dotcom/**/*', 'packages/editor/**/*', 'packages/tldraw/**/*'],
+			rules: {
+				'no-restricted-globals': [
+					'error',
+					{
+						name: 'fetch',
+						message: 'Use the fetch from @tldraw/util instead.',
+					},
+				],
+				'no-restricted-properties': [
+					'error',
+					{
+						object: 'window',
+						property: 'fetch',
+						message: 'Use the fetch from @tldraw/util instead.',
 					},
 				],
 			},
--- a/packages/utils/src/lib/network.ts
+++ b/packages/utils/src/lib/network.ts
@@ -1,0 +1,12 @@
+/**
+ * Just a wrapper around `window.fetch` that sets the `referrerPolicy` to `strict-origin-when-cross-origin`.
+ *
+ * @public
+ */
+export async function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+	return window.fetch(input, {
+		// We want to make sure that the referrer is not sent to other domains.
+		referrerPolicy: 'strict-origin-when-cross-origin',
+		...init,
+	})
+}
--- a/apps/dotcom/src/components/BoardHistorySnapshot/BoardHistorySnapshot.tsx
+++ b/apps/dotcom/src/components/BoardHistorySnapshot/BoardHistorySnapshot.tsx
@@ -1,7 +1,7 @@
 import { ROOM_PREFIX } from '@tldraw/dotcom-shared'
 import { RoomSnapshot } from '@tldraw/tlsync'
 import { useCallback, useState } from 'react'
-import { Tldraw } from 'tldraw'
+import { Tldraw, fetch } from 'tldraw'
 import '../../../styles/core.css'
 import { assetUrls } from '../../utils/assetUrls'
 import { useFileSystem } from '../../utils/useFileSystem'
--- a/apps/dotcom/src/components/ShareMenu.tsx
+++ b/apps/dotcom/src/components/ShareMenu.tsx
@@ -10,6 +10,7 @@ import {
 	TldrawUiMenuContextProvider,
 	TldrawUiMenuGroup,
 	TldrawUiMenuItem,
+	fetch,
 	unwrapLabel,
 	useActions,
 	useContainer,
--- a/apps/dotcom/src/hooks/useMultiplayerAssets.ts
+++ b/apps/dotcom/src/hooks/useMultiplayerAssets.ts
@@ -4,6 +4,7 @@ import {
 	MediaHelpers,
 	TLAsset,
 	TLAssetId,
+	fetch,
 	getHashForString,
 	uniqueId,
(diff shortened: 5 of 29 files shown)


packages/utils/src/lib/network.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Just a wrapper around `window.fetch` that sets the `referrerPolicy` to `strict-origin-when-cross-origin`.
+ *
+ * @public
```
SomeHats, line 4: let's mark this as `@internal` - i don't know if we want it as part of our public API

.eslintrc.js
```diff
@@ -91,11 +91,11 @@ module.exports = {
 				'no-restricted-globals': [
 					'error',
 					{
-						name: 'setInterval',
+						name: 'setTimeout',
 						message: 'Use the timers from @tldraw/util instead.',
 					},
```
SomeHats, line 96: feel free to call it out of scope but this made me curious and then i had a hard time tracking down what it was actually referring to & how to use it. maybe tweak this message to say "Use the timers from `editor.timers`" instead?

.eslintrc.js
```diff
@@ -123,6 +123,27 @@ module.exports = {
 				],
 			},
 		},
+
+		{
+			files: ['apps/dotcom/**/*', 'packages/editor/**/*', 'packages/tldraw/**/*'],
+			rules: {
+				'no-restricted-globals': [
```
SomeHats, line 130: just to confirm - how do these being specified here interact with the ones from above? can we double check that they merge rather than overwrite, and add a comment explaining that that's the case if they do?

SomeHats (commented):
nice! is it worth doing something similar for `new Image` and ``?

mimecuvalo:
> is it worth doing something similar for new Image and ?

yeah lemme take a look. for some reason, it feels like too much but i think you're right, it's probably the correct thing to do.

.eslintrc.js
mimecuvalo: hah! omg good catch. yes, i had this structured wrong 😬

mimecuvalo:
ok, incorporated the feedback — thanks @SomeHats!
PTAL

mimecuvalo:
As I went down this rabbit hole, this got really...complicated actually. I found some more places that weren't cleaning up their timers after disposal. Take your time to review, wanna get this right. 

cc @MitjaBezensek you might also wanna take a look at the `throttle.ts` changes around `fpsThrottle`. Lemme know what you think.

packages/store/src/lib/Store.ts
```diff
@@ -205,7 +214,7 @@ export class Store<R extends UnknownRecord = UnknownRecord, Props = unknown> {
 				// If we have accumulated history, flush it and update listeners
 				this._flushHistory()
 			},
-			{ scheduleEffect: (cb) => throttleToNextFrame(cb) }
+			{ scheduleEffect: (cb) => (this.cancelHistoryReactor = throttleToNextFrame(cb)) }
```
SomeHats, line 217: vaguely feels like this should be handled by something within signia but probably not worth changing here

packages/tldraw/package.json
```diff
@@ -58,6 +58,7 @@
 		"canvas-size": "^1.2.6",
 		"classnames": "^2.3.2",
 		"hotkeys-js": "^3.11.2",
+		"lodash.throttle": "^4.1.1",
```
SomeHats, line 61: Instead of removing `throttle` from `@tldraw/utils` and adding this in the `tldraw` package, could we add this dependency to `@tldraw/utils` and re-export it from there? I'd like there to be one sort of "blessed" way to do throttling and `@tldraw/utils` feels like a good way to signal that.

packages/utils/src/lib/network.ts
```diff
@@ -0,0 +1,26 @@
...
+	})
+}
+
+/**
+ * Just a wrapper around `new Image`, and yeah, it's a bit strange that it's in the network.ts file
+ * but the main concern here is the referrerPolicy and setting it correctly.
+ *
+ * @internal
+ */
+export const Image = (width?: number, height?: number) => {
```
SomeHats, line 21: I'd be tempted to make this something like `export function createImage(...) {...}`, and then update the lint rule to use `no-restricted-syntax` with `NewExpression[callee.name=Image]` as the selector. That way, people can still use `Image` as a type and in `instanceof` checks.

packages/utils/src/lib/throttle.ts
```diff
@@ -44,19 +48,25 @@ function tick() {
 
 let started = false
 
+interface PseudoThrottleFn {
```
SomeHats, line 51: maybe name this `CancellableFn` or `ThrottledFn` or something?

packages/tldraw/src/lib/TldrawImage.tsx
```diff
@@ -186,7 +186,13 @@ export const TldrawImage = memo(function TldrawImage(props: TldrawImageProps) {
 
 	return (
 		<div ref={setContainer} style={{ position: 'relative', width: '100%', height: '100%' }}>
-			{url && <img src={url} style={{ width: '100%', height: '100%' }} />}
+			{url && (
+				<img
```
SomeHats, line 190: we could ban `img` tags without a referrer policy with this `no-restricted-syntax` rule: `JSXOpeningElement[name.name=img]:not(:has(JSXAttribute[name.name=referrerPolicy]))`

SomeHats (approved):
looks good! probably out of scope for this diff, but i wonder if we should move all of our throttling and debouncing utils to the `Timers` class so the disposal can be managed centrally

packages/utils/src/lib/throttle.ts
```diff
@@ -68,6 +78,8 @@ export function fpsThrottle(fn: () => void) {
 		}
 		tick()
 	}
+	throttledFn.cancel = () => frame && cancelAnimationFrame(frame)
```
MitjaBezensek, line 81: Not sure we can cancel here. This would cancel all the functions in the `fpsQueue`, which we probably don't want?

What we can do though is filter out the function from the queue?

mimecuvalo:
>  but i wonder if we should move all of our throttling and debouncing utils to the Timers class so the disposal can be managed centrally

yeah not a bad call for a future PR 👍

packages/tldraw/src/lib/TldrawImage.tsx
mimecuvalo: yeah that was added in already :)

packages/utils/src/lib/network.ts
mimecuvalo: I think since it's `@internal` for now I feel kinda ¯\\_(ツ)_/¯ . But yeah, if it's `@public` one day, I think that'd make sense to do.

packages/utils/src/lib/throttle.ts
mimecuvalo: actually, this had a hard time exporting for some reason - gonna extract this out for now.

mimecuvalo: (exporting for the api docs, i mean)

PR #3891, opened 2024-06-06
Cropping undo/redo UX

ds300:
This PR aims to improve the UX around undo/redo and cropping. Before the PR if you do some cropping, then stop cropping, then hit `undo`, you will end up back in the cropping state and it will undo each of your resize/translate cropping operations individually. This is weird 🙅🏼 It should just undo the whole sequence of changes that happened during cropping.

To achieve that, this PR introduces a new history method called `squashToMark`, which strips out all the marks between the current head of the undo stack and the mark id you pass in.

This PR also makes the default history record mode of `updateCurrentPageState` to `ignore` like it already was for `updateInstanceState`. The fact that it was recording changes to the `croppingShapeId` was the reason that hitting undo would put you back into the cropping state. 

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -890,6 +890,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	undo(): this {
 		this._flushEventsForTick(0)
+		this.complete()
 		this.history.undo()
 		return this
 	}
```
ds300, line 896: @steveruizok can you think of any reason we wouldn't want to stop any in-progress interactions on regular undo/redo? It seems right to me, and it naturally produces correct behaviours with the cropping stuff.

steveruizok: No, this seems correct to me too.

SomeHats (approved):
lgtm!

PR #3898, opened 2024-06-07
Open share menu when navigating from create new shared project button

Taha-Hassan-Git:
This PR ensures that the share menu is open when navigating from the create new shared project button.

This could be achieved by calling ```editor.addOpenMenu('share-menu')``` after ```navigate('/new')``` , but that opens the menu on the local editor first so you get a flash of that before the multiplayer editor loads. 

Instead I've used the state prop on the Navigate component/function to pass a property called shouldOpenShareMenu. 

If the user presses the back button to return to the local editor instance then the open state of the share menu is preserved, which is actually something we don't want in this case. Not sure how to deal with that without adding a ton more complexity, is it worth it?

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. Click create new shared project from the file menu
2. Share menu should open after navigating to the new room

### Release Notes

- Open share menu when navigating from the create new shared project button.


Taha-Hassan-Git:
Hmm, just noticed the share menu doesn't close! Need to fix that.

steveruizok:
Looks good! I also added the logic to open the share menu if you go to /new or /r directly

PR #3905, opened 2024-06-10
Set up automatic VS Code publishing

MitjaBezensek:
When pushing to production branch we now also package and publish a new version of the VS Code extension. We get the last version from VS Code marketplace and update the package.json with that version. We don't commit that to the repo though (see the discussion below).

I added `VSCE_PAT` secret (my own personal access token from the dev.azure.com), which will expire in 1 year. This is used when running the publish command.

Some more info here:
- [Publishing from CI](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions)
- Publishing uses `VSCE_PAT` env variable
image

### Change Type

- [x] `vs code` — Changes to the vscode plugin

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

- Automate publishing of the VS Code extension.

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -73,13 +73,10 @@ async function main() {
 		env.TLDRAW_ENV === 'staging' || env.TLDRAW_ENV === 'production' || env.TLDRAW_ENV === 'preview',
 		'TLDRAW_ENV must be staging or production or preview'
 	)
-	let currentStep = 1
-	// Previews have an extra step of aliasing the deployment and production builds also release the VS Code extension
-	const totalSteps = previewId || env.TLDRAW_ENV === 'production' ? 8 : 7
 
 	await discordMessage(`--- **${env.TLDRAW_ENV} deploy pre-flight** ---`)
 
-	await discordStep(`[${currentStep++}/${totalSteps}] setting up deploy`, async () => {
+	await discordStep('[1/7] setting up deploy', async () => {
 		// make sure the tldraw .css files are built:
 		await exec('yarn', ['lazy', 'prebuild'])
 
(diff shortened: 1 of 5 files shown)


scripts/deploy.ts
```diff
@@ -73,10 +73,13 @@ async function main() {
...
 		'TLDRAW_ENV must be staging or production or preview'
 	)
+	let currentStep = 1
+	// Previews have an extra step of aliasing the deployment and production builds also release the VS Code extension
+	const totalSteps = previewId || env.TLDRAW_ENV === 'production' ? 8 : 7
 
 	await discordMessage(`--- **${env.TLDRAW_ENV} deploy pre-flight** ---`)
 
-	await discordStep('[1/7] setting up deploy', async () => {
+	await discordStep(`[${currentStep++}/${totalSteps}] setting up deploy`, async () => {
```
SomeHats, line 82: maybe worth pulling this logic into the `discordStep` function? i've generally tried to keep the top-level stuff in this function as basic as possible to make it clear how deploys work

scripts/deploy.ts
```diff
@@ -521,6 +542,15 @@ async function coalesceWithPreviousAssets(assetsDir: string) {
 	}
 }
 
+async function packageAndPublishVSCodeExtension() {
```
SomeHats, line 545: does this push the new commit anywhere? should it?

SomeHats (commented):
Love that we're automating this! 

I would be tempted to separate out vscode app deploys from the `deploy.ts` script which currently just does dotcom, but not sure. i could see us wanting to control deploys of each separately for hotfixing/versioning reasons

SomeHats:
another reason to separate this out: so that the token expiring doesn't affect dotcom deploys

scripts/deploy.ts
```diff
@@ -521,6 +542,15 @@ async function coalesceWithPreviousAssets(assetsDir: string) {
...
+	const EXTENSION_DIR = './apps/vscode/extension'
+	await exec('yarn', ['version', 'patch'], { pwd: EXTENSION_DIR })
```
ds300, line 547: As is, this version bump won't be persisted, and will end up being the same on subsequent deploys unless we manually bump it. Is that problematic for the marketplace? i.e. can we make multiple deploys with different code but with the same package.json version string?

ds300: If the markeplace isn't cool with that there are two options I see:

1. fetch the latest deployed version from the marketplace api and bump that
2. persist the package.json change by pushing the commit to both `main` and `production`

We do something like 2 for the sdk deploys but it's a pain to set up so I'd recommend trying out number 1 first.

MitjaBezensek:
Thanks for the feedback. I've pulled this into it's own action now. We get the version info directly from the marketplace using `vsce show` command. This means the version in the package.json will be outdated and we'll always get the most up to date  directly.

mimecuvalo:
excited about this :)

steveruizok:
Thanks for the initiative here, will be great to have this on rails.

ds300 (approved):
looks great.

For a followup I wonder whether we could set up an insiders version that deploys from main so we can use it for QA? Would be nice to have vscode be QA-able by anybody?

MitjaBezensek:
> looks great.
> 
> For a followup I wonder whether we could set up an insiders version that deploys from main so we can use it for QA? Would be nice to have vscode be QA-able by anybody?

Great idea. There's a concept of pre-releases with extensions, will create a separate PR to add it.

PR #3913, opened 2024-06-11
Make ArrowBindingUtil public

ds300:
I wanted to extend the arrow binding util to adapt holograph for the bindings api, but it wasn't being exported like all the shape utils were. I could have plucked it from the `defaultBindingUtils` array but that felt dirty. What do you think @SomeHats ?

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -9,6 +9,11 @@
 import { ArrayOfValidator } from '@tldraw/editor';
 import { BaseBoxShapeTool } from '@tldraw/editor';
 import { BaseBoxShapeUtil } from '@tldraw/editor';
+import { BindingOnChangeOptions } from '@tldraw/editor';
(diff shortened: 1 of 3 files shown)


SomeHats:
not sure about this. i don't know if this is a pattern we want to encourage right now - it's only really something you can do in very limited cases. buttt also it's probably fine??

ds300:
yeah i think it's probably OK. Certainly not less OK than making the shape utils public and implicitly allowing people to extend them with class inheritance. As long as we don't explicitly say "it's cool you can do this" i think nobody could sue us if their stuff breaks right?

steveruizok:
Landing! Acceptable risk

PR #3924, opened 2024-06-11
images: avoid double request for animated images

mimecuvalo:
right now, for animated images, we end up doing _two_ requests because we're trying to create a static frame if someone wants to pause the animation.

Screenshot 2024-06-11 at 16 26 28

the problem is that the two requests are slightly different:
1.) there's one request via a JS `Image` call that sets a `crossorigin="anonymous"`
2.) the other request is the basic image request via setting a background-image, but this doesn't specify a crossorigin, hence it causes a separate request.

this converts the image rendering to not use a div+background-image but to use a regular image tag and make the crossorigin consistent for animated images. you'll note that we _don't_ set crossorigin for non-animated and that's because for the new Cloudflare images the headers don't send back access-control headers (at the moment, until we want to set up workers).

drive-by cleanup to remove `strict-origin-when-cross-origin` that should have been removed in https://github.com/tldraw/tldraw/pull/3884

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Release Notes

- Images: avoid double request for animated images.

### DIFF
--- a/packages/state/api-report.md
+++ b/packages/state/api-report.md
@@ -207,7 +207,7 @@ export function useValue<Value>(value: Signal<Value>): Value;
 // @public (undocumented)
 export function useValue<Value>(name: string, fn: () => Value, deps: unknown[]): Value;
 
-// @internal
+// @public
 export function useValueDebounced<Value>(name: string, fn: () => Value, deps: unknown[], ms: number): Value;
 
 // @public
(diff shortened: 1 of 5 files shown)


SomeHats (approved):
looks good@

PR #3930, opened 2024-06-12
Better generated docs for react components

SomeHats:
Before:
Screenshot 2024-06-12 at 12 57 26

After:
Screenshot 2024-06-12 at 12 59 16

React components in docs now list their props, and appear under a new "Component" section instead of randomly under either `Function` or `Variable`. In order to have our docs generate this, a few criteria need to be met:
1. They need to be tagged with the `@react` tsdoc tag
2. Their props need to be a simple type alias, typically to an interface.

Both of these rules are enforced with a new lint rule - any component tagged as `@public` will have these rules enforced.

### Change Type


scripts/lib/eslint-plugin.ts
```diff
@@ -295,4 +295,137 @@ exports.rules = {
...
+
+			function checkComponentDeclaration(
+				services: utils.ParserServices,
+				node: TSESTree.VariableDeclarator | TSESTree.FunctionDeclaration,
+				propsType: ts.TypeNode | undefined
+			) {
+				const declaration = findTopLevelParent(node)
+				const comments = context.getSourceCode().getCommentsBefore(declaration)
+
+				// we only care about components tagged as public
```
mimecuvalo, line 312: actually, why only the `@public` ones? feels like eventually any item can eventually become `@public`. i don't feel strongly on it though and can see why we wouldn't need it for now.

scripts/lib/eslint-plugin.ts
```diff
@@ -295,4 +295,137 @@ exports.rules = {
...
+							callee.property.type === utils.AST_NODE_TYPES.Identifier &&
+							callee.property.name === 'memo')
+
+					const isForwardRefFn =
+						(callee.type === utils.AST_NODE_TYPES.Identifier && callee.name === 'forwardRef') ||
+						(callee.type === utils.AST_NODE_TYPES.MemberExpression &&
+							callee.property.type === utils.AST_NODE_TYPES.Identifier &&
+							callee.property.name === 'forwardRef')
+
+					const isComponenty =
```
mimecuvalo, line 386: heh

mimecuvalo (approved):
great stuff. 🎉 

small thing but it's funny, that example screenshot in the PR description has the sample code have a `;` at the beginning - what's up with that?

SomeHats:
> small thing but it's funny, that example screenshot in the PR description has the sample code have a `;` at the beginning - what's up with that?

yeah this is annoying... we format our examples with prettier, which for some reason adds this? I might see about getting rid of it

scripts/lib/eslint-plugin.ts
SomeHats: Mostly to avoid this creating churn and developer friction where it's not needed - if a component isn't public, it's not documented, so it doesn't need the `@react` tag. If it doesn't have the `@react` tag, it doesn't need to have its props specified in a very limited way

tsdoc.json
```diff
@@ -0,0 +1,15 @@
+{
+	"$schema": "https://developer.microsoft.com/json-schemas/tsdoc/v0/tsdoc.schema.json",
+	"extends": ["@microsoft/api-extractor/extends/tsdoc-base.json"],
+
+	"tagDefinitions": [
```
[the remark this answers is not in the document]
mimecuvalo: huh, i had tried using them before but i feel like the compiler threw up at me (at least at the time) ¯\\_(ツ)_/¯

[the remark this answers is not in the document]
SomeHats: we have them switched off because they're not integrated into the rest of our build/docs process, so they don't actually do anything at the moment

PR #3931, opened 2024-06-12
Inline documentation links in type excerpts

SomeHats:
Before: 
Screenshot 2024-06-12 at 15 54 38

After: 
Screenshot 2024-06-12 at 15 55 10

Previously, when items in our documentation referred to each other in code snippets, we'd put the links to their documentation pages in a separate "references" section at the bottom of the docs. Generally I find that this makes links harder to find (they're not in-context) and adds a fair bit of visual noise to our API documentation.

This diff moves those links inline by adding a post-processing step to our highlighted code. This is slightly more involved than I wanted it to be (see the comments in code.tsx for an explanation of why) but it gets the job done. I've added small link icons next to linked code items - i experimented with underlines and a 🔗 icon too, but this seemed to look the best.

### Change Type


apps/docs/components/mdx-components/code.tsx
```diff
@@ -2,10 +2,203 @@
...
+	links: Record<string, string>
+}) {
+	return <CodeLinksContext.Provider value={links}>{children}</CodeLinksContext.Provider>
+}
+
+export function Code({ children, ...props }: React.ComponentProps<'code'>) {
+	const codeLinks = useContext(CodeLinksContext)
+
+	const newChildren = useMemo(() => {
+		// to linkify code, we have to do quite a lot of work. we need to take the output of
```
mimecuvalo, line 35: hah whoa. a lot of work, indeed!

PR #3933, opened 2024-06-13
Flatten shapes to image(s)

steveruizok:
This PR adds some functionality for turning shapes into images.

Kapture 2024-06-13 at 12 51 00

It adds:
- the `flattenShapesToImages`
- the `useFlatten` hook
- a `flatten-shapes-to-images` action (shift + f)
- adds `flattenImageBoundsExpand` option
- adds `flattenImageBoundsPadding` option

## Flatten shapes to images

The `flattenShapesToImages` helper method will 1) create an image for the given shape ids, 2) add it to the canvas in the same location / size as the source shapes, and then 3) delete the original shapes. The new image will be placed correctly in the z index and in the correct rotation of the root-most ancestor of the given shape ids.

image

It has an argument, `flattenImageBoundsExpand`, which if provided will chunk the given shapes into images based on their overlapping (expanded) bounding boxes.

image

By default, the flatten action uses the editor's `options.flattenImageBoundsExpand`. The `flattenImageBoundsPadding` option is used as a value for how much larger the image should be than the source image bounds (to account for large strokes, for example).

### Change Type

- [x] `feature` — New feature

### Test Plan

1. Select shapes
2. Select context menu > edit > flatten

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Add Flatten, a new menu item to flatten shapes into images


packages/tldraw/src/lib/ui/hooks/useFlatten.ts
```diff
@@ -0,0 +1,191 @@
...
+				point.sub(new Vec(padding, padding).rot(-rotation))
+				x = point.x
+				y = point.y
+			} else {
+				// if the common ancestor is the page, then just adjust for the padding
+				x = bounds.x - padding
+				y = bounds.y - padding
+				rotation = 0
+			}
+
```
mimecuvalo, line 151: add a `mark` here to for undo/redo

packages/tldraw/src/lib/ui/hooks/useFlatten.ts
```diff
@@ -0,0 +1,191 @@
...
+					ids: [item.id],
+					bounds: item.bounds,
+				}
+				continue
+			}
+
+			let didLand = false
+
+			for (const group of groups) {
+				if (group.bounds.includes(item.bounds)) {
```
mimecuvalo, line 54: i'm wondering if we should skip over some shapes. if you flatten an embed right now, it's prbly not what you'd expect. makes me wonder if there should be exceptions, like embeds, or anything else.

mimecuvalo (approved):
lookin' great! pre-approving modulo couple comments

packages/tldraw/src/lib/ui/hooks/useFlatten.ts
steveruizok: Good call

packages/tldraw/src/lib/ui/hooks/useFlatten.ts
steveruizok: I've marked in the action, which feels like the right place for it

mimecuvalo: mm, something's not right about it - I can undo but I can't redo unless there's an explicit `mark`. Unless I'm missing something about how new history stuff works 🤔

JOHNKIM-KK:
Hello,

I'm trying to use the 'flatten-to-image' functionality that's available in the right-click context menu, but I want to trigger it programmatically from a custom UI button or another interface element.

Methods I've tried so far:
1. `editor.dispatch({ type: 'ui', name: 'flatten-to-image' })`
2. `editor.flattenShapes(shapeIds)`

However, neither of these approaches seems to work as expected.

Is there an official way to programmatically invoke the 'flatten-to-image' functionality without going through the right-click menu? Or is there a recommended API to implement equivalent functionality?

Thank you for your help!

PR #3934, opened 2024-06-13
image: follow-up fixes for LOD

mimecuvalo:
couple fixes and improvements for the LOD work.

- add `format=auto` for Cloudflare to send back more modern image formats
- fix the broken asset logic that regressed (should not have looked at `url`)
- fix stray parenthesis, omg
- rm the `useValueDebounced` function in lieu of just debouncing the resolver. the problem was that the initial load in a multiplayer room has a zoom of 1 but then the real zoom comes in (via the url) and so we would double load all images 😬. this switches the debouncing to the resolving stage, not making it tied to the zoom specifically.

### Change Type

- [x] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,31 +1,30 @@
...
 	const shapeScale = asset && 'w' in asset.props ? width / asset.props.w : 1
 	// We debounce the zoom level to reduce the number of times we fetch a new image and,
 	// more importantly, to not cause zooming in and out to feel janky.
-	const debouncedScreenScale = useValueDebounced(
-		'zoom level',
-		() => editor.getZoomLevel() * shapeScale,
-		[editor, shapeScale],
-		500
-	)
+	const screenScale = editor.getZoomLevel() * shapeScale
```
SomeHats, line 21: should this still be in a `useValue`?

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,31 +1,30 @@
...
-			const resolvedUrl = await editor.resolveAssetUrl(assetId, {
-				screenScale: debouncedScreenScale,
-			})
-			setUrl(resolvedUrl)
-		}
-		resolve()
-	}, [assetId, debouncedScreenScale, editor])
+		const debouncedFn = debouncedResolver.current
+		debouncedFn(screenScale)
+		return () => debouncedFn.cancel()
```
SomeHats, line 26: I think that this version with the `debounce` fn + `async` leaves a case where the component could unmount in the async gap after calling `resolveAssetUrl`. What if we move everything into the `useEffect`? (afaict, the timing of this is the same as a the debounce version?)

```suggestion
		let isCancelled = false
		const timer = setTimeout(() => {
			const resolvedUrl = await editor.resolveUrl(...)
			if (!isCancelled) setUrl(resolvedUrl)
		})
		return () => {
			clearTimeout(timer)
			isCancelled = true
		}
```

SomeHats (approved):
lgtm!

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: 🤦 oof - thx

packages/tldraw/src/lib/shapes/shared/useAsset.ts
mimecuvalo: sounds good to me! 👍

PR #3938, opened 2024-06-14
deploy: rm manual skew protection

mimecuvalo:
Noticed during our CDN/asset exploration that we're doing manual skew protection but we don't really need to do so anymore since Vercel introduced Skew Protection.
- Docs: https://vercel.com/docs/deployments/skew-protection
- Setting here (at the bottom): https://vercel.com/tldraw/tldraw/settings/advanced

This would simplify our build / avoid unnecessary gymnastics to deploy basically. And it would make builds a tiny bit faster.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know


MitjaBezensek:
Would this really work for us? Here's a few questions I have:
1. Seems like skew protection only works for maximum of 12h. Is that enough? Right now we kept several releases active, which means probably months of releases.
3. Another thing I noticed is that it works automatically for some frameworks (Next, SvelteKit,...) but I'm not sure if this would work for us? I guess it might automatically work for static assets?

SomeHats:
yeah @ds300 has more context on this but i believe we found that vercel's built-in skew protection wasn't appropriate for us for some reason? i can't remember what though, but possibly to do with the issues mitja is pointing out

mimecuvalo:
Sorry for the late response, it's been busy :)

> Seems like skew protection only works for maximum of 12h. Is that enough? Right now we kept several releases active, which means probably months of releases.

It should be. To be clear, this is only for dotcom, right? The original impetus is for people loading the site while a deploy is happening. We preload all our assets anyway. If we were delay-loading things until something triggered, then maybe it would be worth doing more, but I don't think so.

> Another thing I noticed is that it works automatically for some frameworks (Next, SvelteKit,...) but I'm not sure if this would work for us? I guess it might automatically work for static assets?

Ah, this is true actually. I think I forget sometimes that Next is used on the docs site but Vite is used on the dotcom site. I'll look into adding the manual setup (either a cookie or a flag) and giving that a go.

SomeHats:
the motivation isn't for people loading the site during deploy, it's for old clients that may (thru e.g. lazy loading or just opening different parts of the UI) be requesting assets (including JS bundles) from previous deploys. for example - if i have a tldraw tab open for a week and then i go to add an embed, we'll load all the embed icons at that moment. without longer-term skew protection, that would fail

ds300:
Sorry i missed this so far.

Alex is right, we can't get rid of this stuff without breaking older running clients. 12h is nowhere near long enough, it (unfortunately) has to be on the order of months.

Dan did suggest a better way to do it at some point though: just dump all the assets into a flat s3 bucket instead of uploading them via the normal vercel deploy. Since the filenames are content-hashed we can just keep pumping files into s3 forever and use url forwarding to serve them to users. From what I remember it wouldn't work with R2 for some reason, but maybe I was wrong about that.

mimecuvalo:
>  if i have a tldraw tab open for a week and then i go to add an embed, we'll load all the embed icons at that moment. without longer-term skew protection, that would fail

But they all would have been pre-loaded by then, right? We do that in `preloadIcons` in `assetUrls.ts` So they shouldn't fail.

Sounds like maybe we should discuss this offline. As far as I can see, we make sure all the assets are loaded in memory upon load and this feels like we're over-engineering it a bit on our end.

PR #3965, opened 2024-06-18
Fix asset positions

steveruizok:
This PR fixes the positions of assets created when multiple assets are created at once.

### Change Type

- [x] `bugfix` — Bug fix

### Test Plan

1. Drop / paste multiple images on the canvas.
2. The shapes should be top aligned and positioned next to eachother.

### Release Notes

- Fixes the position of multiple assets when pasted / dropped onto the canvas.


mimecuvalo (approved):
huh, did this ever work? did it regress from something else?

steveruizok:
No idea, but it works now!

PR #3980, opened 2024-06-18
bindings guards?

SomeHats:
@mimecuvalo found this spooky set of sentry bugs:
https://tldraw.sentry.io/issues/5396016166/events/?project=4504203639193600&statsPeriod=14d&stream_index=7

These seem to originate from a mix of undo/redo, multiplayer, parsing/loading files, and duplication/isolation stuff - ie, times where we're doing something funky with the store.

I don't know how to fix these. I couldn't find a way to repro them. This is maybe one idea, but i don't like it. @ds300, do you have any better ideas?

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -447,14 +447,14 @@ export class Editor extends EventEmitter<TLEventMap> {
 								for (const binding of this.getBindingsInvolvingShape(descendantShape)) {
 									invalidBindingTypes.add(binding.type)
 
-									if (binding.fromId === descendantShape.id) {
+									if (binding.fromId === descendantShape.id && this.ensureBindingValid(binding)) {
 										this.getBindingUtil(binding).onAfterChangeFromShape?.({
 											binding,
 											shapeBefore: descendantShape,
 											shapeAfter: descendantShape,
 										})
 									}
-									if (binding.toId === descendantShape.id) {
+									if (binding.toId === descendantShape.id && this.ensureBindingValid(binding)) {
 										this.getBindingUtil(binding).onAfterChangeToShape?.({
 											binding,
 											shapeBefore: descendantShape,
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -534,25 +538,33 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+							this.getBindingUtil(binding).onAfterCreate?.({ binding })
+						}
 					},
 					beforeChange: (bindingBefore, bindingAfter) => {
-						const updated = this.getBindingUtil(bindingAfter).onBeforeChange?.({
-							bindingBefore,
-							bindingAfter,
-						})
-						if (updated) return updated
+						if (this.ensureBindingValid(bindingAfter)) {
```
ds300, line 554: I don't think this should run on `beforeChange` or `beforeCreate`, only `afterChange` and `afterCreate`. Otherwise it makes updates rely on the order in which updates are applied during e.g. applyDiff

    applyDiff({ `add shape a`, `add binding between shape a and b`, `add shape b`})

if those diff operations were applied in the order I wrote them, the binding's `onBeforeCreate` would not be invoked.

    applyDiff({ `change binding toId to shape b`, `add shape b` })

and here the binding would end up being deleted instead of updated.

ds300:
hm this is rather worrying. I'm fine adding some safeguards as a stopgap, but I really think we should try hard to reproduce these errors and figure out what's going on. Especially for things like duplicateShapes I don't see how these errors are happening without there being a legit bug we need to address.

mimecuvalo:
I might suggest having a bridge PR where, when something is invalid, it is reported to Sentry with any relevant information/context that would be helpful in debugging where the issue is coming from.

PR #3987, opened 2024-06-19
[prototype] Server-side editor powered multiplayer

SomeHats:
a tiny multiplayer prototype based around the idea of running the `Editor` with its side-effects on the server as a sort of source of truth. I'd like to build on this with:
* adding network diffs instead of the current whole-record diffs
* not sending updates that come from local side effects
* looking at when we run side effects in response to multiplayer updates
* presence

this also explores DO storage for sync, but whatever.


ds300:
> not sending updates that come from local side effects

This is a really interesting point. If side effects are not idempotent they could error or cause bugs when running on the backend. I think most (all?) of ours are idempotent at the moment but who knows if that will stay true. Maybe we could have a react-style dev-mode feature that runs side effects twice and complains if they produce different results on the second run.

Also i guess there's an opportunity to reduce network traffic if we can prevent sending side effect diffs that would happen at the destination.

PR #4002, opened 2024-06-24
[bemo bae] Spike on tlsync public API improvement

ds300:
This PR replaces the extendable TLServer class with an instantiatable wrapper for the TLSyncRoom called TLSocketRoom.

The goal is to provide an API where you pretty much just

1. create a room from some (optional) snapshot
2. pass websockets into it when they connect

And then lifecycle stuff and persistence stuff is left to the consumer, since that all seems to be much more context dependent.

One thing remaining here is to work on observability. We had a slightly messy situation regarding logging and error handling and I want to clean that all up.

### Change Type

- [ ] `bugfix` — Bug fix
- [ ] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

4.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes


packages/tlsync/src/lib/TLSocketRoom.ts
```diff
@@ -0,0 +1,167 @@
...
+				ws: socket,
+				onBeforeSendMessage: this.opts.onBeforeSendMessage
+					? (msg, stringified) => this.opts.onBeforeSendMessage!(sessionId, msg, stringified)
+					: undefined,
+			})
+		)
+
+		socket.addEventListener('message', handleSocketMessage)
+		socket.addEventListener('close', handleSocketClose)
+		socket.addEventListener('error', handleSocketError)
```
ds300, line 85: these don't do anything if the socket was accepted with the hiberatation api

SomeHats:
one quirk of the this with the websocket hibernation API is that we can't rely on the `TLSocketRoom.sessions` set any more. If the DO hibernates,  all the in-memory state will be evicted. When it wakes up again, there won't appear to be any rooms connected. We should:
- use [`state.getWebSockets`](https://developers.cloudflare.com/durable-objects/api/websockets/#getwebsockets) instead of (or to populate) `sessions`
- maybe use [`socket.serializeAttachment`](https://developers.cloudflare.com/durable-objects/api/websockets/#serializeattachment) to stash data on web socket objects that will survive hibernation

ds300:
gonna do some manual testing for the following 

- [ ] room not found still works
- [ ] waking from sleep still works
- [ ] history + revert still works
- [ ] onSessionRemoved is invoked correctly
- [ ] onDataChange is invoked correctly
- [ ] onBeforeSendMessage is invoked correctly

then I think it's good to go

apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
```diff
@@ -376,48 +399,29 @@ export class TLDrawDurableObject extends TLServer {
...
+
+		const snapshot = JSON.stringify(room.getCurrentSnapshot())
+
+		const key = getR2KeyForRoom(slug)
+		await Promise.all([
+			this.r2.rooms.put(key, snapshot),
+			this.r2.versionCache.put(key + `/` + new Date().toISOString(), snapshot),
+		])
+		this._lastPersistedClock = clock
+	}, PERSIST_INTERVAL_MS / 2)
```
SomeHats, line 420: ooc, why does this get divided by two?

apps/dotcom-worker/src/lib/utils/createPersistQueue.test.ts
```diff
@@ -0,0 +1,58 @@
+import { createPersistQueue } from './createPersistQueue'
+
+const tick = () => Promise.resolve()
+
+const resolvable = () => {
```
SomeHats, line 5: `import {promiseWithResolve} from '@tldraw/utils'`

not identical, but it's a `Promise` with `resolve` and `reject` methods

packages/tlsync/src/lib/TLSocketRoom.ts
```diff
@@ -0,0 +1,182 @@
...
+			onSessionRemoved?: (
+				room: TLSocketRoom<R, SessionMeta>,
+				args: { sessionKey: string; numSessionsRemaining: number; meta: SessionMeta }
+			) => void
+			// a callback that is called whenever a message is sent
+			onBeforeSendMessage?: (
+				sessionId: string,
+				msg: TLSocketServerSentEvent<R>,
+				stringified: string
+			) => void
```
SomeHats, line 38: should we make this an options bag so its easier to change later?

SomeHats (approved):
love it!

apps/dotcom-worker/src/lib/TLDrawDurableObject.ts
ds300: the PERSIST_INTERVAL_MS const is kinda for an 'outer' loop, implemented with the alarm scheduler. This bit is kinda for an 'inner' loop that allows `await this.persistToDatabase()` to wait on the whole persist queue to drain rather than just one persist operation. (we do this when the room shuts down). I was worried that if the two loops use the same interval it might be possible for the outer loop to keep the queue backed up for as long as document changes are coming in.

now I'm writing this out it seems unlikely but 🤷🏼 can't hurt

ds300: added a comment explaining

PR #4021, opened 2024-06-26
sdk: wires up tldraw to have licensing mechanisms

mimecuvalo:
For non-commercial usage of tldraw, this adds a watermark in the corner, both for branding purposes and as an incentive for our enterprise customers to purchase a license.

For commercial usage of tldraw, you add a license to the `<Tldraw licenseKey={YOUR_LICENSE_KEY} />` component so that the watermark doesn't show.

The license is a signed key that has various bits of information in it, such as:
- license type
- hosts that the license is valid for
- whether it's an internal-only license
- expiry date

We check the license on load and show a watermark (or throw an error if internal-only) if the license is not valid in a production environment.

This is a @MitjaBezensek, @Taha-Hassan-Git, @mimecuvalo joint production! 🤜 🤛 

### Change Type

- [ ] `bugfix` — Bug fix
- [x] `feature` — New feature
- [ ] `dunno` — I don't know

### Test Plan

1. We will be dogfooding on staging.tldraw.com and tldraw.com itself before releasing this.

### Release Notes

- SDK: wires up tldraw to have licensing mechanisms.


packages/editor/modules.d.ts
```diff
@@ -2,3 +2,8 @@ declare module '*.png' {
 	const url: string
 	export default url
 }
+
+declare module '*.svg' {
```
MitjaBezensek, line 6: I think we could remove the declare module for the png above, I only added it for the temporary watermark.

PR #4023, opened 2024-06-26
Watermark improvements

MitjaBezensek:
A few improvements to the licensing:
* WatermarkManger is no longer created from the LicensingManager, felt better to separate the two.
* We now output some license information to the console when the license is invalid / has expired / does not apply to the current domain.
* Watermark now uses `tl-container` for appending the watermark. This class name is added as part of our code and should always be there.
* General code improvements.

### Change Type

- [ ] `feature` — New feature
- [ ] `api` — API change
- [ ] `bugfix` — Bug fix

### Release Notes

- Licensing and watermark improvements.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -135,6 +135,7 @@ import { SnapManager } from './managers/SnapManager/SnapManager'
 import { TextManager } from './managers/TextManager'
 import { TickManager } from './managers/TickManager'
 import { UserPreferencesManager } from './managers/UserPreferencesManager'
+import { WatermarkManager } from './managers/WatermarkManager'
 import { ShapeUtil, TLResizeMode, TLShapeUtilConstructor } from './shapes/ShapeUtil'
 import { RootState } from './tools/RootState'
 import { StateNode, TLStateNodeConstructor } from './tools/StateNode'
@@ -710,8 +711,10 @@ export class Editor extends EventEmitter<TLEventMap> {
 			this._tickManager.start()
 		})
 
-		this.licenseManager = new LicenseManager(this)
-		this.licenseManager.getLicenseFromKey(licenseKey)
+		this.licenseManager = new LicenseManager()
+		const license = this.licenseManager.getLicenseFromKey(licenseKey)
+		this.watermarkManager = new WatermarkManager(this)
+		this.watermarkManager.checkWatermark(license)
 
 		this.performanceTracker = new PerformanceTracker()
 	}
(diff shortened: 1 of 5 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -710,8 +711,10 @@ export class Editor extends EventEmitter<TLEventMap> {
 			this._tickManager.start()
 		})
 
-		this.licenseManager = new LicenseManager(this)
-		this.licenseManager.getLicenseFromKey(licenseKey)
+		this.licenseManager = new LicenseManager()
+		const license = this.licenseManager.getLicenseFromKey(licenseKey)
+		this.watermarkManager = new WatermarkManager(this)
```
MitjaBezensek, line 716: A bit unsure if we need these managers?

I can see us using the license manager in case we wanted it to hold the license information for displaying it in some place. Not sure about the watermark manager though. Still feels better than having it inside the license manager though.

packages/editor/src/lib/editor/managers/WatermarkManager.ts
```diff
@@ -1,19 +1,41 @@
...
+				return true
+			}
+		}
+		return false
+	}
+
+	checkWatermark(license: LicenseFromKeyResult) {
+		if (!this.shouldShowWatermark(license)) return
+		this.createWatermark()
+		this.editor.timers.setTimeout(() => {
```
MitjaBezensek, line 37: Now checking this only once instead of on an interval.

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -11,81 +9,107 @@ const licenseInfoValidator = T.object({
...
-			// console.log('License expired')
-			this.watermarkManager.createWatermark()
+	}
+
+	private outputMessage(message: string) {
+		this.outputMessages([message])
+	}
+
+	private outputMessages(messages: string[]) {
+		if (this.isTest) return
```
MitjaBezensek, line 102: Don't print out in tests. Maybe we need some other exceptions here as well?

packages/editor/src/lib/editor/Editor.ts
Taha-Hassan-Git: It was originally a React function component, I changed it based on Steve's suggestion that it should be created on mount. It doesn't need to be a class, or be called a manager, I just couldn't come up with another way of structuring it.

MitjaBezensek: Yeah, same here, don't have a better idea. I mean, we could just have a function since it doesn't really store any state atm.

Taha-Hassan-Git (approved):

PR #4036, opened 2024-06-28
Add publish dates, add parsing of flags

MitjaBezensek:
This PR:
* Adds parsing of flags embedded inside the licenses.
* Adds publish dates when we publish our packages.
* Adds additional license expiry check to account for perpetual and annual licenses.
* It also includes changes from #4051 

### Change Type

- [ ] `feature` — New feature
- [ ] `api` — API change
- [ ] `bugfix` — Bug fix

### Release Notes

- Add parsing of flags embedded in license keys. Add publish dates to the version.ts files when we publish.


packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -68,6 +93,33 @@ export class LicenseManager {
...
+		)
+	}
+
+	private isAnnualLicenseExpired(expiryDate: Date) {
+		const expiration = this.getExpirationDateWithGracePeriod(expiryDate)
+		return new Date() >= expiration
+	}
+
+	private isPerpetualLicenseExpired(expiryDate: Date) {
+		const expiration = this.getExpirationDateWithGracePeriod(expiryDate)
```
MitjaBezensek, line 110: Hm, not sure if we want to use the grace period for perpetual licenses. Feels strange since it's comparing to release dates. I'd probably remove it?

packages/tldraw/src/lib/ui/version.ts
```diff
@@ -2,3 +2,8 @@
 // Do not edit manually. Or do, I'm a comment, not a cop.
 
 export const version = '2.2.0'
+export const publishDates = {
+	major: '2024-06-28T10:56:07.893Z',
```
mimecuvalo, line 6: should these be GMT? i.e. '2024-06-28T00:00:00.000Z'
i'm just wondering about the day offset problem we were talking about

scripts/refresh-assets.ts
```diff
@@ -404,11 +406,43 @@ async function writeAssetDeclarationDTSFile() {
 	await writeCodeFile('scripts/refresh-assets.ts', 'typescript', assetDeclarationFilePath, dts)
 }
 
+function getNewPublishDates(packageVersion: string) {
+	const currentVersion = new SemVer(version)
+	const currentPacakgeVersion = new SemVer(packageVersion)
+	const now = new Date().toISOString()
+	if (currentPacakgeVersion.major > currentVersion.major) {
```
mimecuvalo, line 413: nit: typo `currentPacakgeVersion`

packages/editor/src/lib/editor/managers/WatermarkManager.ts
```diff
@@ -20,13 +20,14 @@ export class WatermarkManager {
...
-				return true
-			}
-			if (license.isLicenseExpired) {
-				return true
+		if (!license.isDomainValid) {
+			return true
+		}
+		if (license.isPerpetualLicenseExpired || license.isAnnualLicenseExpired) {
+			if (license.isInternalLicense) {
+				throw new Error('Internal license expired.')
```
mimecuvalo, line 28: i did this internal bit slightly differently (in Editor.ts) - it's not the biggest deal but it feels like it shouldn't be in the watermark logic maybe. but i also don't care too much :P

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -1,11 +1,22 @@
 import { T } from '@tldraw/validate'
 import nacl from 'tweetnacl'
 import util from 'tweetnacl-util'
+import { publishDates } from '../../../version'
+
+const GRACE_PERIOD_DAYS = 5
+
+const FLAGS = {
+	annualLicense: 1,
```
mimecuvalo, line 9: nit: let's use hex here (it'll be slightly easier to maintain).

e.g.
```

const FLAG_ANNUAL_LICENSE = 0x01
const FLAG_LIFETIME_LICENSE = 0x02
const FLAG_PERPETUAL_LICENSE = 0x04
const FLAG_DOMAIN_LOCKED = 0x08
const FLAG_STUFF = 0x10
const FLAG_OTHER = 0x20
```

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -68,6 +93,33 @@ export class LicenseManager {
 		}
 	}
 
+	private getExpirationDateWithGracePeriod(expiryDate: Date) {
```
mimecuvalo, line 96: killer - this is great

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -68,6 +93,33 @@ export class LicenseManager {
...
+		return new Date(
+			expiryDate.getFullYear(),
+			expiryDate.getMonth(),
+			expiryDate.getDate() + GRACE_PERIOD_DAYS + 1 // Add 1 day to include the expiration day
```
mimecuvalo, line 100: i think GRACE_PERIOD_DAYS will be per license but this is fine for now

mimecuvalo (approved):
pre-approving - thx!!

packages/editor/src/lib/editor/managers/LicenseManager.ts
MitjaBezensek: I think @steveruizok said we don't need that per license so I didn't include it. We can always add it if needed.

packages/editor/src/lib/editor/managers/WatermarkManager.ts
MitjaBezensek: Yeah, good call. We have a couple of PRs in flight, but I'll add an issue to our project to revisit this before it goes out.

packages/tldraw/src/lib/ui/version.ts
MitjaBezensek: Handled that by throwing away the hour / timezone part of the date and added an additional day to that.
https://github.com/tldraw/tldraw/pull/4036/files#diff-6e7aadaa37ccd9daf67f2589ea799f33bf1c67b5a47ace64d4096ee40f460358R100

packages/editor/src/lib/editor/managers/LicenseManager.ts
MitjaBezensek: I don't like the `FLAGS` prefix, feels so repetitive. Having an object also steers people into adding the flags there instead of (potentially) scattering them around in different places. Did change it to hex 👍

mimecuvalo: sounds good!

PR #4039, opened 2024-06-28
Unfurl bookmarks in worker

ds300:
This PR adds a `GET /api/unfurl?url=blahblah` endpoint to our worker.

I tried out the existing cheerio implementation but it added 300kb to our worker bundle in the end, due to transitive dependencies.

So I implemented the same logic with cloudflare's sanctioned streaming HTML parser `HTMLRewriter` and it seems to work fine.

I also made the vscode extension do it's fetching locally (from the node process so it's not bound by security policies), retaining the cheerio version for that. At the same time I fixed a bug in the RPC layer that was preventing unfurled metadata from loading correctly.

In a few months we can retire the bookmark-extractor app by just deleting it in the vercel dashboard.

### Change Type

- [ ] `feature` — New feature
- [ ] `api` — API change
- [ ] `bugfix` — Bug fix

### Test Plan

2.

- [ ] Unit Tests
- [ ] End to end tests

### Release Notes

- Do link unfurling on the same subdomain as all our other api endpoints.


apps/vscode/editor/src/utils/rpc.ts
```diff
@@ -36,7 +36,7 @@ export function rpc(
 		vscode.postMessage(inMessage)
 
 		const handler = ({ data: response }: MessageEvent<ResponseType | ErrorType>) => {
-			if (uuid === response.uuid) {
```
ds300, line 39: this fixes a bug that was preventing bookmark data from loading properly.

the problem was

- our custom rpc infra is set up to give each message a UUID
- after sending a message, it waits for a response
- it was ignoring messages with the same UUID, which makes sense
- but we were sending the response back with the same UUID as the request so it was being ignored
- and so each RPC promise was only being resolved after a _second_ request of the same type was submitted with a different UUID.
- i.e.
    1. await rpc('blah') // sends request A
    2. await rpc('blah') // sends request B, but the result resolves request A
    3. await rpc('blah') // sends request C, but the result resolves request B

and so on.

apps/vscode/extension/src/unfurl.ts
```diff
@@ -1,4 +1,4 @@
-import cheerio from 'cheerio'
+import { load } from 'cheerio'
```
ds300, line 1: moved this over for use in the vscode extension and it seemed to have different eslint rules so I had to a make a couple of minor syntax edits

packages/validate/src/lib/validation.ts
```diff
@@ -1003,6 +1003,22 @@ export const srcUrl = string.check((value) => {
 	}
 })
 
+/**
+ * Validates that a valid is a url safe to load as an asset.
+ *
+ * @public
+ */
+export const httpUrl = string.check((value) => {
```
mimecuvalo, line 1011: copy-pasta'd description ;)

apps/vscode/extension/src/unfurl.ts
mimecuvalo, line 1: ooc, does `load` prevent some of the the transitive packages to load? (less package size)?

apps/vscode/extension/src/WebViewMessageHandler.ts
```diff
@@ -74,24 +73,11 @@ export class WebViewMessageHandler {
...
-					.then((resp) => {
-						return resp.json()
-					})
-					.then((json: any) => {
+				await unfurl(url)
+					.then(async (json: any) => {
 						this.webviewPanel.webview.postMessage({
 							type: 'vscode:bookmark/response',
-							uuid: e.uuid,
+							uuid: e.uuid + '_response',
```
mimecuvalo, line 80: add a reference to the comment here, otherwise it's unclear why this `_response` is here :)

apps/vscode/editor/src/utils/rpc.ts
mimecuvalo: pls add this comment in the code - otherwise, we'll forget why :)

apps/dotcom-worker/src/lib/utils/unfurl.ts
```diff
@@ -0,0 +1,72 @@
...
+			this.appleIcon = element.getAttribute('href')!
+		}
+	}
+}
+
+export async function unfurl(url: string) {
+	const meta$ = new MetaExtractor()
+	const title$ = new TextExtractor()
+	const icon$ = new IconExtractor()
+	await new HTMLRewriter()
```
mimecuvalo, line 45: handy API! makes me a little nervous to tie ourselves to Cloudflare specific tech, but meh, it's fine i guess ¯\\_(ツ)_/¯ 

can you just a quick link to: https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/ ? it wasn't clear on first read (without seeing a PR description) that this isn't native web api.

mimecuvalo (approved):
looks great! some minor comments but let's go 🔜

SomeHats (approved):
nice!

apps/vscode/extension/src/unfurl.ts
ds300: i doubt it

PR #4042, opened 2024-06-29
Editor.run, locked shapes improvements

steveruizok:
This PR:
- creates `Editor.run` (previously `Editor.batch`)
- deprecates `Editor.batch`
- introduces a `ignoreShapeLock` option top the `Editor.run` method that allows the editor to update and delete locked shapes
- fixes a bug with `updateShapes` that allowed updating locked shapes
- fixes a bug with `ungroupShapes` that allowed ungrouping locked shapes
- makes `Editor.history` private
- adds `Editor.squashToMark`
- adds `Editor.clearHistory`
- removes `History.ignore`
- removes `History.onBatchComplete`
- makes `_updateCurrentPageState` private

```ts
editor.run(() => {
  editor.updateShape({ ...myLockedShape })
  editor.deleteShape(myLockedShape)
}, { ignoreShapeLock: true })
```

It also:

## How it works

Normally `updateShape`/`updateShapes` and `deleteShape`/`deleteShapes` do not effect locked shapes.

```ts
const myLockedShape = editor.getShape(myShapeId)!

// no change from update
editor.updateShape({ ...myLockedShape, x: 100 })
expect(editor.getShape(myShapeId)).toMatchObject(myLockedShape)

// no change from delete
editor.deleteShapes([myLockedShape])
expect(editor.getShape(myShapeId)).toMatchObject(myLockedShape)
```

The new `run` method adds the option to ignore shape lock.

```ts
const myLockedShape = editor.getShape(myShapeId)!

// update works
editor.run(() => { editor.updateShape({ ...myLockedShape, x: 100 }) }, { ignoreShapeLock: true })
expect(editor.getShape(myShapeId)).toMatchObject({ ...myLockedShape, x: 100 })

// delete works
editor.run(() => { editor.deleteShapes([myLockedShape]), { ignoreShapeLock: true })
expect(editor.getShape(myShapeId)).toBeUndefined()
```

## History changes

This is a related but not entirely related change in this PR.

Previously, we had a few ways to run code that ignored the history.

- `editor.history.ignore(() => { ... })`
- `editor.batch(() => { ... }, { history: "ignore" })`
- `editor.history.batch(() => { ... }, { history: "ignore" })`
- `editor.updateCurrentPageState(() => { ... }, { history: "ignore" })`

We now have one way to run code that ignores history:

- `editor.run(() => { ... }, { history: "ignore" })`

## Design notes

We want a user to be able to update or delete locked shapes programmatically.

### Callback vs. method options?

We could have added a `{ force: boolean }` property to the `updateShapes` / `deleteShapes` methods, however there are places where those methods are called from other methods (such as `distributeShapes`). If we wanted to make these work, we would have also had to provide a `force` option / bag to those methods.

Using a wrapper callback allows for "regular" tldraw editor code to work while allowing for updates and deletes.

### Interaction logic?

We don't want this change to effect any of our interaction logic.

A lot of our interaction logic depends on identifying which shapes are locked and which shapes aren't. For example, clicking on a locked shape will go to the `pointing_canvas` state rather than the `pointing_shape`. This PR has no effect on that part of the library.

It only effects the updateShapes and deleteShapes methods. As an example of this, when `_force` is set to true by default, the only tests that should fail are in `lockedShapes.test.ts`. The "user land" experience of locked shapes is identical to what it is now.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Create a shape
2. Lock it
3. From the console, update it
4. From the console, delete it

- [x] Unit tests

### Release notes

- SDK: Adds `Editor.force()` to permit updating / deleting locked shapes 
- Fixed a bug that would allow locked shapes to be updated programmatically
- Fixed a bug that would allow locked group shapes to be ungrouped programmatically


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5009,40 +5032,34 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 		const shapesToReparent = compact(ids.map((id) => this.getShape(id)))
 
-		// The user is allowed to re-parent locked shapes. Unintuitive? Yeah! But there are plenty of
-		// times when a locked shape's parent is deleted... and we need to put that shape somewhere!
-		const lockedShapes = shapesToReparent.filter((shape) => shape.isLocked)
-
-		if (lockedShapes.length) {
-			// If we have locked shapes, unlock them before we update them
-			this.updateShapes(lockedShapes.map(({ id, type }) => ({ id, type, isLocked: false })))
-		}
```
steveruizok, line 5019: Previously, we would need to unlock shapes in order to make these changes.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7251,8 +7276,21 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+					// also locked) then only allow an update that unlocks the shape.
+					if (!(Object.hasOwn(partial, 'isLocked') && !partial.isLocked)) {
+						continue
+					}
+				} else if (this.isShapeOrAncestorLocked(shape)) {
+					// If the shape itself is unlocked, and any of the shape's
+					// ancestors are locked then we'll skip the update
+					continue
+				}
+			}
```
steveruizok, line 7293: Previously, you could update a locked shape just by having the `isLocked` as part of the update. A call like `updateShape({...myLockedShape, x: 100 })` would go through because the `myLockedShape` would have the `isLocked` property on it. From here on, if you want to update a locked shape for any reason (apart from unlocking it) you'll need to use `force`.

Likewise (though less important) you could update a shape with a locked ancestor just by including `isLocked` in the partial, even if `isLocked` would produce no change.

mimecuvalo:
I think language-wise, there might be something to consider that there are other types of "locked" shapes:
- boards / items that are readonly
- things that, if deleted, would violate constraints.

`force` makes it sound like it might not be related to just lock shapes, per se. Putting myself in the shoes of a dev discovering that API for the first time, it might not be what I expect.

This sort of came up with Orion's bug and preventing certain operations with locked shapes: https://github.com/tldraw/tldraw/pull/4007

---

Language-stuff aside: I think, later down the line, in a much more ideal world, I would expect an API with the following shape:
```
editor.execute(() => {...doStuff...}, {variousOptions, ignoreLocked: true, })
```

Right now, we have the `editor.batch` function which is _kind of_ what this could evolve into eventually. 

I would hestitate in adding a specific top-level wrapper function that _just_ does force b/c then we'll start running into nesting of other wrapper functions if you want to combine different behaviours, e.g.:
```
editor.force(() => {
   editor.dontRecordHistory(() => {
     editor.localOnlyDontBroadcastToOtherUsers(() => {
        
     })
   })
})
```
you get the idea :)

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -4377,6 +4377,31 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * ```
+	 *
+	 * @param callback - The callback to run.
+	 *
+	 * @public
+	 */
+	force(callback: () => void) {
+		this._force = true
+		this.batch(callback)
+		this._force = false
```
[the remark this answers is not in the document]
steveruizok: Sounds good!

steveruizok:
Ok, @mimecuvalo had a good point about "we need a single way to run code with some context".

The `Editor.batch` method was leftover from a very very early version of tldraw v2 and didn't actually do anything except wrap the callback in a transaction. A few weeks ago @SomeHats added history options here, which make a lot of sense, and I was going to add `ignoreShapeLock` here as well. Seems like it's pretty useful, though the use isn't exactly in line with the word "batch". The comments were pretty clear though: "run a function... the function to run". So let's use `Editor.run`.

This feels like a good API that can grow as we add additional context.

packages/tldraw/src/test/TestEditor.ts
```diff
@@ -131,6 +131,10 @@ export class TestEditor extends Editor {
 		})
 	}
 
+	getHistory() {
+		return this.history
+	}
+
```
steveruizok, line 137: Now that Editor.history is protected, I've added a `getHistory` method for the test editor.

mimecuvalo:
this PR is becoming :chefs-kiss:

nice one @steveruizok !

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1451,39 +1537,27 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		>
 	): this {
-		this._updateCurrentPageState(partial, historyOptions)
+		this._updateCurrentPageState(partial)
 		return this
 	}
-	_updateCurrentPageState = (
-		partial: Partial<Omit<TLInstancePageState, 'selectedShapeIds'>>,
-		historyOptions?: TLHistoryBatchOptions
+	private _updateCurrentPageState = (
```
SomeHats, line 1554: Now that `updateCurrentPageStage` and `_updateCurrentPageShape` are the same, lets remove the `_` and just keep the main one. We don't need both.

Also, I'm nervous about this change. Having these defaults is pretty convenient, and this is silently introducing bugs for anyone updating to the next version unless they notice this change. We can call it out in the release notes, but i'm still not sure that would be enough...

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -676,16 +676,24 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-				hoveredShapeId: null,
-				erasingShapeIds: [],
-			})
-		})
+		this.run(
+			() => {
+				this.store.ensureStoreIsUsable()
+
+				// clear ephemeral state
+				this.run(
```
SomeHats, line 684: we don't need the nested `this.run` here, we're already in an ignored section

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -981,6 +994,23 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this
 	}
 
+	/**
+	 * Squash the history to the given mark id.
```
SomeHats, line 998: could you expand on what this means? I actually don't know what history squashing does 😅

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1011,16 +1041,69 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 	 */
-	batch(fn: () => void, opts?: TLHistoryBatchOptions): this {
-		this.history.batch(fn, opts)
+	run(
+		fn: () => void,
+		opts?: Partial<
+			TLHistoryBatchOptions & {
+				ignoreShapeLock: boolean
+			}
+		>
```
SomeHats, line 1077: In general, we should avoid `Partial` where its not needed because it produces much harder to read generated docs. Can we write this as `TLHistoryBatchOptions & {ignoreShapeLock?: boolean}` instead? Or even better:
```ts
interface TLBatchOptions extends TLHistoryBatchOptions {
    ignoreShapeLock?: boolean
}
```

SomeHats (commented):
In general I'm v into this change!

There are a alot of breaking changes in this diff though. Could you update the release notes in the diff description to include a breakdown of the breaking changes and a short guide for users around how they can update? In particular, calling out the page state change (or reverting it tbh) seems important.

I'm going to PR this PR with a little experiment in the argument order for editor.run

SomeHats (changes requested):

SomeHats:
PR on this PR: #4067

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: 3.0?

packages/editor/src/lib/editor/Editor.ts
ds300: essentially it removes any marks between the top of the undo stack and the given mark id. So that then next time you hit undo you'll stop at the given mark id instead of some intermediate mark.

this is currently only used when completing a cropping operation, that I'm aware of.

SomeHats: Ah makes sense! We should include this description in the comment

packages/editor/src/lib/editor/Editor.ts
ds300: yeah the page state changes should be ignored in all cases except updating selectedShapeIds, which there's a separate method for. I think we should revert this to include the historyOptions and use ignore by default.

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: actually, this reminds me - if you accidentally have a nested `run` - does this cause any problems? should it? (i don't think it should, but worth pondering over)

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1011,16 +1041,69 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	): this {
+		if (opts?.ignoreShapeLock) {
+			this._force = true
+		}
+		try {
+			this.history.batch(fn, opts)
+		} finally {
+			if (opts?.ignoreShapeLock) {
+				this._force = false
```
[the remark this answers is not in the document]
steveruizok: Fixed (by you) in #4067 !

PR #4045, opened 2024-06-29
Improve image asset loading

steveruizok:
This PR improves the loading cycle for image assets.
Kapture 2024-06-29 at 17 30 54

Showing the correctly debounced upscale after resize ends
https://github.com/tldraw/tldraw/assets/23072548/58ca3682-daa4-4bea-8c0a-fb28a482a657

Previously, creating a new image for any reason (copying and pasting, alt-dragging) would have a 500ms delay before showing the image, even if the image had already been loaded. I thought this might be a bug with the way we preloaded images, but it turned out to be part of the logic added in #3684 for debouncing new images.

In this PR, we take better advantage of the signals library (and other state in tldraw) to allow immediate fetching when the image first comes up, and then correctly debounced fetching when the user finishes resizing a shape or moving the camera.

We no longer update the shape if the resulting URL is the same as it was before.

### Design decisions

- Don't adjust the LOD while resizing, even if the user has "paused" during a resize interaction

### Change type

- [x] `bugfix`
- [x] `improvement`

### Test plan

1. Create an image shape
2. Duplicate it
3. The new image should immediately appear
4. Resize a shape
5. The LOD should change once you've finished resizing
4. Change the zoom level
5. The LOD should change once you've finished zooming

### Release notes

- Fixed a bug with new image shapes not immediately appearing.


packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,38 +1,104 @@
...
+								('isAnimated' in asset.props && asset.props.isAnimated),
+						})
+					})
+				return
+			}
+			// If the camera is still moving, don't get a new url
+			if (editor.getCameraState() !== 'idle') return
+
+			// If the editor is not in the idle state (ie if it is resizing) don't get a new url
+			if (!editor.isIn('select.idle')) return
```
mimecuvalo, line 66: mm, i don't we shouldn't check this, couple reasons:
- you might be in `hand.idle` and things become culled/unculled
- if you're in multiplayer, you might not be in a `select.idle` mode but someone else causes something that necessitates getting a new image

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,38 +1,104 @@
...
-		return () => {
-			clearTimeout(timer)
-			isCancelled = true
-		}
-	}, [assetId, asset?.props.src, isCulled, screenScale, editor])
-
-	return { asset, url }
+
+	// The resolved url of the asset that should be used by the shape (duplicated as a ref)
+	const [state, setState] = useState<{ url: string | null; isAnimated: boolean }>({
```
mimecuvalo, line 17: good idea on just putting `isAnimated` inside here (although it _might_ be a little image specific / unrelated to videos).

i don't think it needs to be part of `state`. it's only computed once and then just copy-pasted from then on. let's keep it simple and just have a state for the url and just return the `isAnimated`.

i.e.
```
const [url, setUrl] = useState(null)
const [isAnimated, setIsAnimated] = useState(() =>
  ('mimeType' in asset.props && MediaHelpers.isAnimatedImageType(asset?.props.mimeType)) ||
  ('isAnimated' in asset.props && asset.props.isAnimated))
```

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,38 +1,104 @@
...
+
+			// If the current screen scale is the same as it was last time we got a new url,
+			// then we don't need to ask for a new one.
+			if (rPrevScreenScale.current === screenScale) return
+			rPrevScreenScale.current = screenScale
+
+			// Clear any previous timers and set a new one. If we make it here again before the
+			// request is completed, we will clear this timer and set a new one. There's a chance
+			// that the timeout is cancelled after the timeout runs but before the promise resolves,
+			// but we'll just ignore that case (maybe a cancellable promise would be the solution).
```
mimecuvalo, line 76: hm, i don't know if we can ignore this. it'll get into potentially some weird concurrency bugs. cancellable promise sounds good or via some other mechanism.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,38 +1,104 @@
...
+			if (editor.getCulledShapes().has(shapeId)) return
+
+			const screenScale =
+				editor.getZoomLevel() * ('w' in asset.props ? shape.props.w / asset.props.w : 1)
+
+			// If the prev scale is zero (which is impossible), set it to the current
+			// screen scale and immediately request a new resolved url.
+			if (!rPrevScreenScale.current) {
+				rPrevScreenScale.current = screenScale
+				editor
```
mimecuvalo, line 46: let's avoid the duplication here and call a common function (that can be used below). otherwise, it hard to tell/maintain that this is the same function just meant to be the leadingEdge for the debounce.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -1,38 +1,104 @@
...
+			const shape = editor.getShape<TLImageShape | TLVideoShape>(shapeId)
+			if (!shape) return
+
+			const asset = assetId ? editor.getAsset(assetId) : null
+			if (!asset) return
+
+			if (editor.getCulledShapes().has(shapeId)) return
+
+			const screenScale =
+				editor.getZoomLevel() * ('w' in asset.props ? shape.props.w / asset.props.w : 1)
```
mimecuvalo, line 40: ugh, yeah. now i remember what was tickling my brain on why we needed to make this wait a tick on first page load :-/ (and why we were thinking of some kind of `onBeforeMount`).

the problem is that the board, when it loads, sets the zoom twice, in a multiplayer room. it first loads as a zoom of 1, but then, after reading the url it sets the desired zoom. so it would make _all_ the images in the room load twice at the moment if we didn't make this wait a tick. UGH.

thoughts?

mimecuvalo (commented):
nice catch on the delay for image loading, yikes. couple comments, the major one is the initial editor/room zoom and the timing around that. but also, we should avoid the concurrency thing with the promise/timeout stuff. lemme know if you wanna tag-team.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
steveruizok: Makes sense!

PR #4048, opened 2024-07-01
Ghost image tweaks

SomeHats:
There's a few things in here:
1. Animation changes
    - The shimmer is less sharp, we use a nicely eased gradient instead of a linear gradient with sharp edges
    - When the animation ends, the shimmer fades out instead of disappearing with a snap 
2. Loading bug fix
    - When an image first mounts, we currently wait 500ms before showing it. This is most obvious when duplicating an existing image
    - This is the same bug as fixed in #4045, although i think there's extra stuff in there? Not sure
3. React image preloading `key` trick
    - Instead of preloading in an effect with a temporary `Image`, you can preload with the actual `` you're planning on using with a `key`. 
    - When the loading is finished, the new `` will show and the old will be unmounted
    - Sometimes (depending on how caching etc. is configured), the separate `Image` trick can cause a flash whilst the new image is mounted and fetched from the cache. This doesn't happen with the react/key trick
4. Object URL resource management experiment
    - @mimecuvalo and I spent a while talking about different strategies for managing preview object URLs after the upload has finished. This is based on some ideas I had after that conversation
    - We use reference counting and a timeout to keep the object URL retained as long as the upload is happening, or anything is still making use of it.
    - imo this isn't worth the complexity. the 60s timeout (we could set it to 3-5 minutes if we're really worried) is probably good enough, even though there are times when it theoretically may result in temporarily broken images on very poor connections.

This PR has a bunch of ideas and they're all a bit muddled together, sorry! @mimecuvalo we should chat through what's in here. Roughly:
- I would like to land 1 and maybe 2 (depending on #4045). 
- I like 3, but the `Image` approach is probably good enough too. Could go either way!
- I don't think we should do 4, I don't think the gain is worth the complexity

### Change type

- [x] `improvement`

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -631,15 +631,58 @@ input,
 	align-items: center;
 }
 
-.tl-image-temporary {
-  mask: linear-gradient(-60deg, #000 30%, #0005, #000 70%) right / 400% 100%;
-  background-repeat: no-repeat;
-  animation: shimmer 1s ease-in-out infinite;
+.tl-image-container {
+	position: relative;
+	overflow: hidden;
+}
+.tl-image {
+	position: absolute;
+	inset: 0;
+}
+.tl-image-pending {
+	opacity: 0;
+	pointer-events: none;
+}
+.tl-image-container::after {
+	content: ' ';
+	display: block;
+	position: absolute;
+	left: -100%;
+	right: -100%;
+	top: 0;
+	bottom: 0;
+
+	background: linear-gradient(
+		-60deg,
+		color-mix(in srgb, transparent 100%, var(--color-background)) 30%,
+		color-mix(in srgb, transparent 27.8%, var(--color-background)) 41.3%,
+		color-mix(in srgb, transparent 4.2%, var(--color-background)) 47.22%,
+		color-mix(in srgb, transparent 0%, var(--color-background)) 50%,
+		color-mix(in srgb, transparent 4.2%, var(--color-background)) 52.78%,
+		color-mix(in srgb, transparent 27.8%, var(--color-background)) 58.7%,
+		color-mix(in srgb, transparent 100%, var(--color-background)) 70%
+	);
+
+	animation: shimmer 1s linear infinite;
+
+	opacity: 0;
+	display: none;
+	transition:
+		opacity 0.5s,
+		display 0.5s allow-discrete;
+}
+
+.tl-image-container.tl-image-container-loading::after {
+	opacity: 0.3;
+	display: block;
 }
 
 @keyframes shimmer {
+	0% {
+		transform: translateX(-100%);
+	}
 	100% {
-		mask-position: left;
+		transform: translateX(100%);
 	}
 }
 
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -58,6 +58,7 @@ import {
 	IndexKey,
 	JsonObject,
 	PerformanceTracker,
+	RC,
 	Result,
 	Timers,
 	annotateError,
@@ -3945,9 +3946,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 			shouldResolveToOriginalImage?: boolean
 		}
 	): Promise<string | null> {
-		if (!assetId) return ''
+		if (!assetId) return null
 		const asset = this.getAsset(assetId)
-		if (!asset) return ''
+		if (!asset) return null
 
 		const { screenScale, shouldResolveToOriginalImage } = context
 
@@ -7675,7 +7676,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	}
 
 	/** @internal */
-	private readonly temporaryAssetPreview: Record<TLAssetId, string> = {}
+	private readonly temporaryAssetPreview = new Map<TLAssetId, RC<string>>()
 
 	/**
 	 * Register an external content handler. This handler will be called when the editor receives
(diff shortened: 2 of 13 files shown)


packages/utils/src/lib/control.ts
```diff
@@ -64,3 +64,47 @@ export function promiseWithResolve<T>(): Promise<T> & {
 		reject: reject!,
 	})
 }
+
+/** @internal */
+export class RC<out T> {
```
mimecuvalo, line 69: can we call this `ReferenceCountWithFixedTimeout`? RC is too ambiguous :P

and also i don't want somebody to accidentally use this (and accidentally forget to set `timeout`) and it just immediately disposes things b/c the default is 0 😆

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -211,16 +162,40 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
...
-							draggable={false}
-						/>
+					<div
+						className={classNames(
+							'tl-image-container',
+							loaded?.isPlaceholder && 'tl-image-container-loading'
+						)}
+						style={containerStyle}
+					>
+						{/* We have two images: the currently loaded image, and the next image that
```
mimecuvalo, line 172: MUCH BETTER THANK YOU!

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -143,28 +112,12 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
...
-				</>
-			)
-		}
+		const nextUrl = url === loaded?.src ? null : url
+		const nextSrc = useRC(nextUrl)
+		const loadedSrc = useRC(!shape.props.playing || reduceMotion ? staticFrameSrc : loaded?.src)
 
 		// This is specifically `asset?.props.src` and not `url` because we're looking for broken assets.
-		if (!asset?.props.src) {
+		if (!url && !asset?.props.src) {
```
mimecuvalo, line 120: comment above should be updated if we need `!url` too

packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts
```diff
@@ -161,9 +161,6 @@ export class TLLocalSyncClient {
 		} catch (error: any) {
 			onLoadError(error)
 			showCantReadFromIndexDbAlert()
-			if (typeof window !== 'undefined') {
```
mimecuvalo, line 164: unintentional or drive-by?

packages/editor/editor.css
```diff
@@ -631,15 +631,58 @@ input,
...
+		color-mix(in srgb, transparent 100%, var(--color-background)) 70%
+	);
+
+	animation: shimmer 1s linear infinite;
+
+	opacity: 0;
+	display: none;
+	transition:
+		opacity 0.5s,
+		display 0.5s allow-discrete;
```
mimecuvalo, line 672: TIL about `allow-discrete` - neat!

mimecuvalo: looks like FF doesn't support it though? https://caniuse.com/?search=allow-discrete

does this still work as intended on FF?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7714,19 +7716,20 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-	clearTemporaryAssetPreview(objectUrl: string) {
-		URL.revokeObjectURL(objectUrl)
+		const urlString = URL.createObjectURL(file)
+		const urlResource = new RC(
+			urlString,
+			() => {
+				this.temporaryAssetPreview.delete(assetId)
+				URL.revokeObjectURL(urlString)
+			},
+			60_000
```
mimecuvalo, line 7729: nit: we should write as 60 * 1000 - i dunno, i think it's typically not written as the complete number for readability (in our codebase either)

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -17,22 +18,68 @@ export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null, width: n
...
+		if (didAlreadyResolve.current) {
+			const timer = editor.timers.setTimeout(resolve, 500)
+
+			return () => {
+				clearTimeout(timer)
+				isCancelled = true
+			}
+		} else {
+			// if not, resolve immediately:
+			resolve()
```
mimecuvalo, line 63: there is this issue (i mentioned on Steve's PR), maybe your PR has as well, where the first load of the page will have two network requests unfortunately (this was a discussion we had about adding an `onBeforeMount`, relatedly). we sort of need to at least wait a tick, i think, to avoid this dbl-request...

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -17,22 +18,68 @@ export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null, width: n
...
+		if (url) didAlreadyResolve.current = true
+	}, [url])
+
 	useEffect(() => {
 		if (isCulled) return
 
+		if (assetId && !asset?.props.src) {
+			const preview = editor.getTemporaryAssetPreview(assetId)
+			if (preview) {
+				setUrl(preview.retain())
```
mimecuvalo, line 33: a comment here would be good to describe the lifecycle. i think you and i know what's going on but reference counting is tricky and i think we should make this a little clearer to future devs on what is being mitigated here.

to be specific:
- that the `retain` here is to stop it being released from the uploader side of things (which initially kicked off the process). 
- that sometimes it's setting the RC and sometimes it's setting the string and how it eventually "settles"

mimecuvalo (commented):
this is so cool! 🤘🤘🤘🤘🤘🤘🤘🤘

PR #4049, opened 2024-07-01
Subtle crypto

Taha-Hassan-Git:
Adds subtle crypto

I've added a file to the utils folder called licensing, this contains some utils that are used by both the test file and the license manager. I wondered about keeping it close to the manager and test file, but this seemed better.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Adapts license manager to use subtle crypto

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -711,11 +711,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 			this._tickManager.start()
 		})
 
-		this.licenseManager = new LicenseManager()
-		this.watermarkManager = new WatermarkManager(this)
 		this.checkLicenseKey = async () => {
-			const license = await this.licenseManager.getLicenseFromKey(licenseKey)
-			this.watermarkManager.checkWatermark(license)
+			const licenseManager = new LicenseManager()
+			const watermarkManager = new WatermarkManager(this)
+			const license = await licenseManager.getLicenseFromKey(licenseKey)
+			watermarkManager.checkWatermark(license)
 		}
 		this.checkLicenseKey()
 
@@ -820,20 +820,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	private focusManager: FocusManager
 
 	/**
-	 * A manager for licensing.
-	 *
-	 * @internal
-	 */
-	private licenseManager: LicenseManager
-	/**
-	 * A manager for the watermark.
-	 *
-	 * @internal
-	 */
-	private watermarkManager: WatermarkManager
-
-	/**
-	 * A function that checks the watermark.
+	 * A function that instantiates the license manager and watermark manager, and
+	 * checks the license key. Showing a watermark if the license key is invalid.
 	 *
 	 */
 	private checkLicenseKey: () => void
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -713,14 +713,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		this.licenseManager = new LicenseManager()
 		this.watermarkManager = new WatermarkManager(this)
-		this.licenseManager
-			.getLicenseFromKey(licenseKey)
-			.then((license) => {
-				this.watermarkManager.checkWatermark(license)
-			})
-			.catch((error) => {
-				throw new Error(`Error getting license: ${error}`)
-			})
+		this.checkLicenseKey = async () => {
+			const license = await this.licenseManager.getLicenseFromKey(licenseKey)
+			this.watermarkManager.checkWatermark(license)
+		}
+		this.checkLicenseKey()
 
 		this.performanceTracker = new PerformanceTracker()
 	}
@@ -834,6 +831,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @internal
 	 */
 	private watermarkManager: WatermarkManager
+
+	/**
+	 * A function that checks the watermark.
+	 *
+	 */
+	private checkLicenseKey: () => void
 
 	/**
 	 * The current HTML element containing the editor.
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -712,9 +712,15 @@ export class Editor extends EventEmitter<TLEventMap> {
 		})
 
 		this.licenseManager = new LicenseManager()
-		const license = this.licenseManager.getLicenseFromKey(licenseKey)
 		this.watermarkManager = new WatermarkManager(this)
-		this.watermarkManager.checkWatermark(license)
+		this.licenseManager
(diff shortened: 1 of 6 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -712,9 +712,15 @@ export class Editor extends EventEmitter<TLEventMap> {
 		})
 
 		this.licenseManager = new LicenseManager()
-		const license = this.licenseManager.getLicenseFromKey(licenseKey)
 		this.watermarkManager = new WatermarkManager(this)
-		this.watermarkManager.checkWatermark(license)
+		this.licenseManager
```
Taha-Hassan-Git, line 716: Need to use Promise syntax here. Subtle crypto uses async functions.

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -26,38 +26,57 @@ interface ValidLicenseKeyResult {
...
-		if (!decoded) {
-			throw new Error('Invalid license key')
+		try {
+			await crypto.subtle.verify(
+				{
+					name: 'ECDSA',
+					hash: { name: 'SHA-384' },
+				},
+				publicCryptoKey,
+				new Uint8Array(str2ab(signature)),
```
Taha-Hassan-Git, line 52: It kept throwing an error saying that these and other arguments were not an Array Buffer. I checked them with ```instanceof``` , and saw that they definitely were. At some point I tried this and it worked. Not sure why it helps!

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -26,38 +26,57 @@ interface ValidLicenseKeyResult {
...
+			const licenseInfo = await this.extractLicenseKey(licenseKey)
+
 			const result: ValidLicenseKeyResult = {
 				license: licenseInfo,
 				isLicenseValid: true,
 				isDomainValid: licenseInfo.hosts.some(
 					(host) => host.toLowerCase() === window.location.hostname.toLowerCase()
 				),
-				isLicenseExpired: licenseInfo.expiry > Date.now(),
+				isLicenseExpired: licenseInfo.expiry < Date.now(),
```
Taha-Hassan-Git, line 79: This was the wrong way round

packages/editor/src/lib/utils/licensing.ts
```diff
@@ -0,0 +1,117 @@
+import crypto from 'crypto'
```
mimecuvalo, line 1: ah, this is just for Jest testing - we shouldn't import `crypto` :)

packages/editor/src/lib/utils/licensing.ts
```diff
@@ -0,0 +1,117 @@
...
+	const keyType = isPublic ? 'PUBLIC' : 'PRIVATE'
+	const exported = await crypto.subtle.exportKey(isPublic ? 'spki' : 'pkcs8', key)
+	const exportedAsBase64 = btoa(ab2str(exported))
+	return `-----BEGIN ${keyType} KEY-----\n${exportedAsBase64}\n-----END ${keyType} KEY-----`
+}
+
+/*
+  Generate a sign/verify key pair.
+*/
+export async function generateKeyPair() {
```
mimecuvalo, line 32: let's move the functions that are just needed for Jest testing into the Jest file. we shouldn't have the functions here that won't be used in the client (to keep the bundle smaller). so, to be specific: `generateKeyPair`, `importPrivateKey`, `generateLicenseKey`

packages/editor/src/lib/editor/managers/LicenseManager.test.ts
```diff
@@ -1,35 +1,48 @@
...
-		const result = licenseManager.getLicenseFromKey('')
+	it('Checks if a license key was provided', async () => {
+		const licenseManager = new LicenseManager()
+		const result = await licenseManager.getLicenseFromKey('')
 		expect(result).toMatchObject({ isLicenseValid: false, reason: 'no-key-provided' })
 	})
-	/* it('Validates the license key', () => {
-		const invalidLicenseKey = generateKey(
+	it('Validates the license key', async () => {
+		const keyPair = await generateKeyPair()
```
mimecuvalo, line 11: possible to just do `await generateKeyPair()` in a `before`? that way you can just generate it once and forget about it

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: let's use `await` syntax - makes it easier to read

packages/editor/src/lib/editor/managers/LicenseManager.ts
mimecuvalo: haha, yeah, it's kind of annoying - i think we can just do `as Uint8Array` instead of actually doing wrapping it a `new Uint8Array` but pull me aside if it's still giving you trouble

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -26,38 +26,57 @@ interface ValidLicenseKeyResult {
...
+				new Uint8Array(str2ab(encodedData))
+			)
+		} catch (e) {
+			console.error(e)
+			throw new Error('Invalid signature')
 		}
-		const licenseInfo = JSON.parse(util.encodeUTF8(decoded))
-		return licenseInfoValidator.validate(licenseInfo)
+
+		const decodedData = JSON.parse(atob(encodedData))
```
mimecuvalo, line 60: might need to put this in a try/catch too, in case the JSON parsing fails, quite possible

mimecuvalo (commented):
cool - thx! just a couple minor stuff

packages/editor/src/lib/editor/Editor.ts
Taha-Hassan-Git: I would have to make the constructor async, is that possible/reccomended?

packages/editor/src/lib/utils/licensing.ts
Taha-Hassan-Git: I get failed tests when I run them locally without importing crypto. Is that ok?

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: ah yeah, let's not make the constructor async. but we can make a private async function:
`private async licenseChecks()` or something like that

packages/editor/src/lib/utils/licensing.ts
mimecuvalo: yeah we just need to do `import crypto from 'crypto'` inside of `LicenseManager.test.ts` instead of in here.

packages/editor/src/lib/utils/licensing.ts
mimecuvalo: moving this stuff out will help solve the crypto import thing

PR #4050, opened 2024-07-01
Fixup staging worker deploys

SomeHats:
- Structure wrangler.toml for asset uploads the same way we structure it for other workers
- Extract worker name from wrangler output, not wrangler.toml - not all workers have explicit names for all environments
- Remove DNS settings for preview workers - `tldraw.workers.dev` is fine

### Change type

- [x] `bugfix`

### DIFF
--- a/scripts/deploy-bemo.ts
+++ b/scripts/deploy-bemo.ts
@@ -87,6 +87,7 @@ async function deployBemoWorker({ dryRun }: { dryRun: boolean }) {
 	if (previewId && !didUpdateBemoWorker) {
 		await setWranglerPreviewConfig(workerDir, {
 			name: workerId,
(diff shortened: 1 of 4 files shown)


scripts/deploy-bemo.ts
```diff
@@ -87,7 +87,6 @@ async function deployBemoWorker({ dryRun }: { dryRun: boolean }) {
 	if (previewId && !didUpdateBemoWorker) {
 		await setWranglerPreviewConfig(workerDir, {
 			name: workerId,
```
SomeHats, line 89:
```suggestion
			name: workerId,
			customDomain: `${previewId}-demo.tldraw.xyz`,
```

ds300 (approved):

PR #4051, opened 2024-07-01
Default to a local watermark when we are offline

MitjaBezensek:
Improves the watermarking:
- We now have a `watermark.png` asset in our assets folder. When releasing a new package this file now also gets uploaded to our cdn, just like our static assets.
- We also have a local `watermark.png` image, which we use when using our packages in a offline scenario. This file is now part of our package and it gets copied from the assets folder when `refresh-assets` runs.

This will allow us to update our watermark if we need to and always have it up to date for both scenarios (online and offline).

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Serve a local watermark in offline scenarios. Upload watermark to the cnd during packaging.


packages/editor/src/lib/editor/managers/WatermarkManager.ts
```diff
@@ -1,11 +1,19 @@
...
 import { Editor } from '../Editor'
 import { LicenseFromKeyResult } from './LicenseManager'
+const WATERMARK_FILENAME = 'watermark.png'
 
 export class WatermarkManager {
 	constructor(private editor: Editor) {}
 	private createWatermark() {
-		const watermark = document.createElement('a')
+		const watermark = document.createElement('img')
+		if (navigator.onLine) {
```
MitjaBezensek, line 12: How reliable is this? We could also react to changes to this status, but not sure if it's needed?

packages/tldraw/src/lib/utils/static-assets/assetUrls.ts
```diff
@@ -1,14 +1,5 @@
-import { RecursivePartial } from '@tldraw/editor'
+import { RecursivePartial, getDefaultCdnBaseUrl } from '@tldraw/editor'
 import { useMemo } from 'react'
-import { version } from '../../ui/version'
-
-/** @internal */
-const CDN_BASE_URL = 'https://cdn.tldraw.com'
-
-/** @public */
-export function getDefaultCdnBaseUrl() {
```
MitjaBezensek, line 9: Moved this to the editor package since it also needs it, didn't find a better spot for it.

packages/editor/src/lib/editor/managers/WatermarkManager.ts
mimecuvalo: i think it's good enough for now!

scripts/upload-static-assets.ts
```diff
@@ -56,6 +57,9 @@ export async function uploadStaticAssets(version: string) {
 				await uploadDirectory(`${version}/${folderName}`, `${ASSETS_FOLDER}/${folderName}`)
 			}
 		}
+		// Upload the watermark file
+		const fileName = 'watermark.png'
+		await uploadFile(`${version}/${fileName}`, `${ASSETS_FOLDER}/${fileName}`)
```
mimecuvalo: as mentioned offline, let's just create a folder for this

mimecuvalo (approved):
LG 👍 modulo the one comment

scripts/upload-static-assets.ts
mimecuvalo: which you just fixed 🙃

PR #4052, opened 2024-07-01
add workers-shared package and source-maps for dotcom-worker sentry reports

SomeHats:
Share the new sentry stuff that plays nice with releases/sourcemaps with dotcom-worker using a new `worker-shared` package.


scripts/deploy-bemo.ts
```diff
@@ -87,7 +87,6 @@ async function deployBemoWorker({ dryRun }: { dryRun: boolean }) {
 	if (previewId && !didUpdateBemoWorker) {
 		await setWranglerPreviewConfig(workerDir, {
 			name: workerId,
-			customDomain: `${previewId}-demo.tldraw.xyz`,
```
ds300: thought we were keeping this?

packages/worker-shared/src/sentry.ts
```diff
@@ -0,0 +1,34 @@
...
+export function createSentry(ctx: Context, env: SentryEnvironment, request?: Request) {
+	const { SENTRY_DSN, WORKER_NAME, CF_VERSION_METADATA } = requiredEnv(env, {
+		SENTRY_DSN: true,
+		WORKER_NAME: true,
+		CF_VERSION_METADATA: true,
+	})
+
+	return new Toucan({
+		dsn: SENTRY_DSN,
+		release: `${WORKER_NAME}.${CF_VERSION_METADATA.id}`,
```
ds300, line 25: i think this release name doesn't match with the one used when uploading the source maps now that it includes the full domain. I think we can set up the WORKER_ID in the deploy scripts and pass that through to `wranglerDeploy` to use when it creates the sentry release and uploads the source maps, so that it doesn't need to extract it from either the wrangler output or the wrangler.toml

ds300 (commented):
looks good except for the sentry release name is not synchronized between the worker runtime and the deploy script.

packages/worker-shared/src/sentry.ts
SomeHats: The release name that we use when uploading source maps is `${workerNameMatch[1]}.${versionMatch[1]}`, not the full domain i think?

ds300: ah sorry yeah you're right, i misread the code

ds300 (approved):

PR #4053, opened 2024-07-01
Allow license keys with additional properties

MitjaBezensek:
* Make sure old clients can handle licenses that might have some unknown properties.
* Notify the developers if we have additional properties or flags that we don't know about.
* Detect whether we are in a development environment and only output messages when we are.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Future proof the license key management by allowing unknown properties in the license keys. This will allow old clients to accept keys with some properties they aren't aware of. The same for new flags.


packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -56,7 +56,17 @@ export class LicenseManager {
 			throw new Error('Invalid license key')
 		}
 		const licenseInfo = JSON.parse(util.encodeUTF8(decoded))
-		return licenseInfoValidator.validate(licenseInfo)
+		try {
+			return licenseInfoValidator.validate(licenseInfo)
+		} catch (e: any) {
+			if (e.message.includes('Unexpected property')) {
```
SomeHats, line 62: It wouldn't give you the extra message (not sure how important that is) but in case u missed it you can do `T.object({ ... }).allowUnknownProperties()`

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -56,7 +56,17 @@ export class LicenseManager {
...
+				this.outputMessages([
+					'License key contains some uknown properties.',
```
SomeHats, line 64: typo `uknown`

packages/editor/src/lib/editor/managers/LicenseManager.ts
MitjaBezensek: Yeah, I kinda do that a [bit further down](https://github.com/tldraw/tldraw/pull/4053/files#diff-6e7aadaa37ccd9daf67f2589ea799f33bf1c67b5a47ace64d4096ee40f460358R69). This is just a hacky way to detect that we have some unexpected properties, which we can then communicate to the devs - they might not be aware that the licenses key they are using might unlock certain feature for them.

MitjaBezensek: There might be a better way to do that though? 😄

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -78,7 +94,17 @@ export class LicenseManager {
...
+		try {
+			return licenseInfoValidator.validate(decodedData)
+		} catch (e: any) {
+			if (e.message.includes('Unexpected property')) {
+				this.outputMessages([
+					'License key contains some unknown properties.',
+					'You may want to update tldraw packages to a newer version to get access to new functionality.',
+				])
+			}
+		}
```
mimecuvalo, line 106: we should log an error to sentry if the validation fails for other reasons (other than an unknown property)

MitjaBezensek: How should that work? We now only use sentry for our private stuff (dotcom,...). Did you think we'd set up a separate sentry project for it? How would we handle secrets?

packages/tldraw/api-report.md
```diff
@@ -438,7 +438,7 @@ export const DefaultQuickActions: NamedExoticComponent<TLUiQuickActionsProps>;
 export function DefaultQuickActionsContent(): JSX_2.Element | undefined;
 
 // @public (undocumented)
-export const defaultShapeTools: (typeof ArrowShapeTool)[];
```
MitjaBezensek, line 441: No sure why this is coming up? Rerean the api-check so it seems correct.

packages/editor/src/lib/editor/managers/LicenseManager.ts
mimecuvalo: Oh, right right. Nevermind, yes you're right.

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -213,15 +236,20 @@ export class LicenseManager {
 				`Please reach out to ${LICENSE_EMAIL} if you would like to use tldraw on other domains.`,
 			])
 		}
+		if (result.license.flags > HIGHEST_FLAG * 2) {
```
mimecuvalo, line 239: i might be being dense here 😆 
what is the * 2 for?

MitjaBezensek: Oh, I should have added the comment. If we add a new flag it will always be 2 times the current highest flag. And the sum of all the existing flags is `2 * HIGHEST_FLAG - 1`, so just lower than `HIGHEST_FLAG * 2`.

mimecuvalo: if the highest now is 4 and we do * 2, wouldn't the check fail?
8 > 4 * 2 would just be equal, eh?

i think actually, correct me if i'm wrong, we can just do `flags > HIGHEST_FLAG`, no?

so, 0x10 (a flag higher than expected) > 0x0a (highest)
or, in binary, 0b10000 > 0b01000

packages/editor/src/lib/editor/managers/LicenseManager.ts
MitjaBezensek: Move to using an array for representing the different properties so this part is obsolete now.

packages/editor/src/lib/editor/managers/LicenseManager.ts
mimecuvalo: oh i see, gotcha, now it makes sense 👍

PR #4055, opened 2024-07-01
Demo assets server

SomeHats:
Adds an assets server to the demo worker, and reworks the existing asset server to use the same code. There are a few simplifications to the code due to some DX improvements working with R2 and caches. I also removed the `HEAD` request from the assets server: i took a look at our logs and it's not actually used at all.

This also fixes an issue where users could overwrite the contents of the asset uploads bucket.

### Change type

- [x] `other`


apps/dotcom-asset-upload/src/index.ts
```diff
@@ -10,145 +11,27 @@ interface Env {
...
-				return cachedResponse
-			}
-		}
-
-		const ifNoneMatch = request.headers.get('if-none-match')
-		let hs = request.headers
-		if (ifNoneMatch?.startsWith('W/')) {
-			hs = new Headers(request.headers)
-			hs.set('if-none-match', ifNoneMatch.slice(2))
-		}
```
ds300, line 78: this was working around a bug https://github.com/cloudflare/workers-sdk/issues/3089 i'm not sure if they've fixed it so might be worth either checking or just keeping the code in case

ds300 (approved):

apps/dotcom-asset-upload/src/index.ts
SomeHats: oh thanks for pointing that out! I'll take a look

PR #4060, opened 2024-07-02
put sync stuff in bemo worker

ds300:
this PR puts sync stuff in the bemo worker, but before we can land this we need to synchronize our sdk deploys with our examples app deploys so that we don't ship bemo stuff to examples.tldraw.com. We should have done that a while ago tbh.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with...


apps/bemo-worker/src/BemoDO.ts
```diff
@@ -35,4 +52,150 @@ export class BemoDO extends DurableObject<Environment> {
 			})
 		}
 	}
+
+	async handleConnect(req: IRequest) {
+		// extract query params from request, should include instanceId
+		const url = new URL(req.url)
+		const params = Object.fromEntries(url.searchParams.entries())
+		const { sessionKey } = params
```
SomeHats, line 60:
```suggestion
		const { sessionKey } = req.query
```

Or use `parseRequestQuery` which comes with error handling

apps/dotcom-worker/tsconfig.json
```diff
@@ -1,6 +1,11 @@
 {
 	"extends": "../../config/tsconfig.base.json",
-	"include": ["src", "scripts"],
+	"include": [
+		"src",
+		"scripts",
+		"../../packages/worker-shared/src/createPersistQueue.test.ts",
+		"../../packages/worker-shared/src/createPersistQueue.ts"
```
SomeHats, line 7: typescript spam

SomeHats (approved):
looking good!

PR #4062, opened 2024-07-02
Demo server bookmark unfurl endpoint

SomeHats:
This adds the HTMLRewriter-based bookmark unfurler to the demo server. It moves the unfurler into worker-shared, and adds some better shared error handling across our workers.

I removed the fallback bookmark fetcher where we try and fetch websites locally. This will almost never work, as it requires sites to set public CORS.

### Change type
- [x] `other`

### DIFF
--- a/apps/bemo-worker/src/worker.ts
+++ b/apps/bemo-worker/src/worker.ts
@@ -2,10 +2,13 @@
 /// <reference types="@cloudflare/workers-types" />
 
 import {
-	createSentry,
+	getUrlMetadata,
+	handleApiRequest,
 	handleUserAssetGet,
 	handleUserAssetUpload,
 	notFound,
+	parseRequestQuery,
+	urlMetadataQueryValidator,
 } from '@tldraw/worker-shared'
 import { WorkerEntrypoint } from 'cloudflare:workers'
 import { Router, createCors } from 'itty-router'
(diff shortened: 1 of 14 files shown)


packages/worker-shared/src/getUrlMetadata.ts
```diff
@@ -38,17 +44,25 @@ class IconExtractor {
...
+
 	// we use cloudflare's special html parser https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/
 	await new HTMLRewriter()
 		.on('meta', meta$)
 		.on('title', title$)
 		.on('link', icon$)
-		.transform((await fetch(url)) as any)
-		.blob?.()
+		.transform(response as any)
+		.blob()
```
ds300, line 65: i wonder whether we should budge this up into the try block too, in case of malformed html or non-html responses

apps/bemo-worker/src/worker.ts
```diff
@@ -34,6 +37,10 @@ export default class Worker extends WorkerEntrypoint<Environment> {
 				context: this.ctx,
 			})
 		})
+		.get('/v1/bookmarks/unfurl', async (request) => {
```
ds300, line 40: i dunno if we need version prefixes since we're only guaranteeing the demo server will work with the latest tldraw release

ds300 (approved):

packages/worker-shared/src/getUrlMetadata.ts
SomeHats: good call

apps/bemo-worker/src/worker.ts
SomeHats: makes sense!

SomeHats: i kinda like them just cus they're free and give us room to change our minds about things down the line but i'e removed for now

PR #4067, opened 2024-07-03
Editor.run tweaks

SomeHats:
Couple of changes to editor.run, the main one being the order. Previously, we did `editor.run(fn, opts)`. This sort of makes sense, because options objects usually come last. This diff switched us to `editor.run(opts, fn)`, which I think makes more sense for a few reasons:
1. Callback-comes-last is IMO a stronger convention than options-come-last
2. When the callback is long, having the options all the way down at the end and possibly off the screen makes it really hard to actually read what is happening and understand it. We know this code is it a `run`, but we don't know if that run is affecting history/locking until the bottom. having the options first makes things much more readable imo
3. Better tooling support from e.g. formatters. This is related to 1, but the callback-comes-last convention is known by tools like prettier/biome and so they format it differently. For us, this means removing an unneeded layer of indentation, and keeping the `run` call and its options all on one line.

Other changes:
- fix a bug where forced-ness wouldn't be restored properly after a `run` function
- interface for run options for nicer docs
- clearer name for `._force`

### DIFF
--- a/apps/dotcom/src/hooks/useUrlState.ts
+++ b/apps/dotcom/src/hooks/useUrlState.ts
@@ -56,12 +56,9 @@ export function useUrlState(onChangeUrl: (params: UrlStateParams) => void) {
 				url.searchParams.get(PARAMS.page) ?? 'page:' + url.searchParams.get(PARAMS.p)
 			if (newPageId) {
 				if (editor.store.has(newPageId as TLPageId)) {
-					editor.run(
-						() => {
-							editor.setCurrentPage(newPageId as TLPageId)
-						},
-						{ history: 'ignore' }
-					)
+					editor.run({ history: 'ignore' }, () => {
+						editor.setCurrentPage(newPageId as TLPageId)
+					})
 				}
 			}
 		}
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -820,9 +820,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     bail(): this;
     bailToMark(id: string): this;
     // @deprecated (undocumented)
-    batch(fn: () => void, opts?: Partial<TLHistoryBatchOptions & {
-        ignoreShapeLock: boolean;
-    }>): this;
+    batch(fn: () => void, opts?: TLEditorRunOptions): this;
     bindingUtils: {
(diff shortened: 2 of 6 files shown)


mimecuvalo:
I'm not used to the options being in front, but seeing it coded up, it does make pause because it does seem more readable.

I think then my only concern is if the options is empty in the future, i.e. `editor.run({}, () => {...doStuff...)`
I could see this happening if we start moving towards the `run` function handling undo / redo more or other DX things like that. I would want wrapping my call in an `editor.run()` to:
- handle any errors that come back and rollback appropriately (we had seen these kinds of issues recently)
- handle undo/redo automagically for me
- maybe other things

So, with that preface being said, and to riff on your idea further, Alex, could the options parameter by, ahem, optional? We can have an overloaded interface where it's: `editor.run(optionsOrCallback: fn | {}, callback?: fn)`

SomeHats:
it is optional - there's a few examples in there that use plain `editor.run(() => ...)`

mimecuvalo:
> it is optional - there's a few examples in there that use plain editor.run(() => ...)

do you mean in Steve's PR? yeah, i get that.
i'm just saying with your PR, it implies that the options are always front-loaded and they don't look optional. :)

SomeHats:
no i mean here. it's already overloaded & optional: https://github.com/tldraw/tldraw/blob/104c0e43c412e4d47dae6b6f47d802e6d077e318/packages/editor/src/lib/editor/Editor.ts#L1071-L1082

SomeHats:
i'll add an explicit example of that to the doc comment but the options-less signature is already there

mimecuvalo:
> no i mean here. it's already overloaded & optional:

ohhh! i had missed that in my read-through. great then!

steveruizok:
After some deep soul searching and reflection, I'd prefer to keep options argument last. You've presented good arguments here for the change but here's my reasoning / taste call. 1) generally where optional options are present, we put them last, and I'm biased toward wider consistency even at the cost of minor local weirdness; and 2) there are times where this will be used without any arguments (ie as proxy for `transact`) and I never like the mouthfeel of overloaded argument juggling. Hope that makes sense.

I want the `run` to ship in the next SDK release, so I'm going to merge this into the other PR (because there are other improvements here) and then update the other PR to remove the argument swaps.

PR #4069, opened 2024-07-03
fix asset loading on staging

SomeHats:
Not 100% sure what's going on here or what changed, but in preview/staging deploys we were rewriting the src url to point to the wrong bucket URL which was breaking them

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/src/utils/assetHandler.ts
+++ b/apps/dotcom/src/utils/assetHandler.ts
@@ -6,7 +6,8 @@ import {
 	WeakCache,
 	getAssetFromIndexedDb,
 } from 'tldraw'
-import { ASSET_BUCKET_ORIGIN, ASSET_UPLOADER_URL } from './config'
+import { ASSET_BUCKET_ORIGIN } from './config'
+import { isDevelopmentEnv } from './env'
 
 const objectURLCache = new WeakCache<TLAsset, ReturnType<typeof getLocalAssetObjectURL>>()
 
@@ -55,13 +56,11 @@ export const resolveAsset =
 
 		const width = Math.ceil(asset.props.w * context.steppedScreenScale * networkCompensation)
 
-		if (process.env.NODE_ENV === 'development') {
+		if (isDevelopmentEnv) {
 			return asset.props.src
 		}
 
-		// On preview, builds the origin for the asset won't be the right one for the Cloudflare transform.
-		const src = asset.props.src.replace(ASSET_UPLOADER_URL, ASSET_BUCKET_ORIGIN)
-		return `${ASSET_BUCKET_ORIGIN}/cdn-cgi/image/format=auto,width=${width},dpr=${context.dpr},fit=scale-down,quality=92/${src}`
+		return `${ASSET_BUCKET_ORIGIN}/cdn-cgi/image/format=auto,width=${width},dpr=${context.dpr},fit=scale-down,quality=92/${asset.props.src}`
 	}
 
 async function getLocalAssetObjectURL(persistenceKey: string, assetId: TLAssetId) {


mimecuvalo:
hm, what is this fixing? on the dotcom-preview it still looks broken.

the problem it was trying to fix was for PRs we need to convert:
`https://assets.tldraw.xyz/cdn-cgi/image/format=auto,width=640,dpr=2,fit=scale-down,quality=92/https://pr-4069-tldraw-assets.tldraw.workers.dev/uploads/Hiq754UrA1S8StFXbLWLV-tumblr-n2e789Txe81qhttpto3-1280.jpg`
to:
`https://assets.tldraw.xyz/cdn-cgi/image/format=auto,width=640,dpr=2,fit=scale-down,quality=92/https://assets.tldraw.xyz/uploads/Hiq754UrA1S8StFXbLWLV-tumblr-n2e789Txe81qhttpto3-1280.jpg`

are assets in a different bucket now or something?

mimecuvalo:
to be clear: i'm seeing this broken with transformed images (over 1.5mb)
(sometimes it can look like it's working but it's just the uploaded img is under 1.5mb)

packages/worker-shared/src/userAssetUploads.ts
```diff
@@ -26,43 +29,93 @@ export async function handleUserAssetUpload({
 
 export async function handleUserAssetGet({ request, bucket, objectName, context }: UserAssetOpts) {
 	const cacheUrl = new URL(request.url)
+	const shouldOptimize =
+		request.query.tl_opt && !/image-resizing/.test(request.headers.get('via') ?? '')
+
+	let format = null
+	if (shouldOptimize) {
+		const accept = request.headers.get('Accept')
+		if (accept && /image\/avif/.test(accept)) {
```
mimecuvalo, line 38: quick initial comment: why do this manually? Cloudflare is handling the `format=auto` depending on the request headers to do the right thing. we should keep the behavior there (mostly it returns avif)

packages/worker-shared/src/userAssetUploads.ts
```diff
@@ -26,43 +29,93 @@ export async function handleUserAssetUpload({
...
+			format = 'avif' as const
+		} else if (accept && /image\/webp/.test(accept)) {
+			format = 'webp' as const
+		}
+	}
+
+	// if we have a format, we want to cache the asset with the format in the URL
+	if (format) cacheUrl.searchParams.set('format', format)
```
mimecuvalo, line 46: why does `format` matter for the cacheUrl? why not always cache?

SomeHats: `format=auto` isn't supported by workers transforms

packages/worker-shared/src/userAssetUploads.ts
SomeHats: `format=auto` isn't supported by these worker-based transforms, this is what's suggested in the cloudflare docs, see the note at the bottom of this `format` section: https://developers.cloudflare.com/images/transform-images/transform-via-workers/#format

packages/worker-shared/src/userAssetUploads.ts
SomeHats: two requests from different browsers might return differently formatted responses depending on what the browser `Accept` header contains. We can't cache an avif response and serve it to a browser that only supports jpg. This makes sure that we find and return a format in the cache that the requesting browser definitely supports

ds300 (approved):
seems to work well. I'd be a bit happier if it was set up to deploy alongside all our other workers just so folks don't need to be aware of the need to manually deploy. Also the fear of breaking production by pushing code that has only been locally tested. But I guess if we need to do any significant refactoring later we could fix the deployment situation then.

apps/dotcom/src/utils/assetHandler.ts
```diff
@@ -44,24 +45,35 @@ export const resolveAsset =
...
+		// we only transform images that are hosted on domains we control
+		const isTldrawImage = isDevelopmentEnv || /\.tldraw\.(?:com|svg|workers\.dev)$/.test(url.host)
+
+		if (!isTldrawImage) return asset.props.src
+
 		// Assets that are under a certain file size aren't worth transforming (and incurring cost).
-		if (asset.props.fileSize === -1 || asset.props.fileSize < 1024 * 1024 * 1.5 /* 1.5 MB */)
-			return asset.props.src
+		// We still send them through the image worker to get them optimized though.
+		const isWorthResizing = asset.props.fileSize !== -1 && asset.props.fileSize >= 1024 * 1024 * 1.5
```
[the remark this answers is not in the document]
mimecuvalo: changing to 85 is the right call.

for brief historical context on why it was 92, i mentioned this in the PR: https://github.com/tldraw/tldraw/pull/3827
that it was keeping parity with our original code that did 0.92 for whatever reason: https://github.com/tldraw/tldraw/blob/3adae06d9c1db0b047bf44d2dc216841bcbc6ce8/packages/tldraw/src/lib/defaultExternalContentHandlers.ts#L89

packages/worker-shared/src/userAssetUploads.ts
mimecuvalo: looks like this was updated in the end 👍

PR #4070, opened 2024-07-03
Fix licensing tests and adding missing ones

MitjaBezensek:

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix importing svgs inside of jest. Add missing tests.

### DIFF
--- a/packages/editor/src/importMeta.ts
+++ b/packages/editor/src/importMeta.ts
@@ -1,0 +1,3 @@
+const { MODE: IMPORT_META_ENV } = import.meta && (import.meta as any).env
+
+export { IMPORT_META_ENV }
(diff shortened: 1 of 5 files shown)


packages/editor/src/importMeta.ts
```diff
@@ -0,0 +1,4 @@
+// Used for getting around the issue of using import meta in jest. This is a workaround to allow us to mock the import.meta object in jest tests.
+// See LicenseManager.test.ts for an example of how to mock this.
+const { MODE: IMPORT_META_ENV } = import.meta && (import.meta as any).env
```
mimecuvalo, line 3: hm my Vite-fu isn't the best - i wonder if @SomeHats or @ds300 would have a better idea here 🤔

MitjaBezensek: I've simplified this a bit, but still using the same approach.
https://github.com/tldraw/tldraw/pull/4021/files#diff-15f06a16b6271f950e2caf09ed4d5f1a6df7becac09a69b4ea825637a3fabe2fR3

mimecuvalo: hm, this seems to break other tests...

Issue #4066, opened 2024-07-03
[Feature]: Paste at cursor by default


jorgen-braseth-adsk:
### What's the feature?

Today, the keyboard shortcut for pasting at cursor is shift+cmd+v
We find that our users would prefer this to be the default behaviour.

Feature request:
- Make paste at cursor the default
or
- Make the default configurable, so that we can configure our specific instance like this.

Discord discussion:
https://discord.com/channels/859816885297741824/1195314269030465586/threads/1257294262140403712

### Contact Details

jorgen.aars.braseth@autodesk.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4088, opened 2024-07-08
Add "paste at cursor" option, which toggles how `cmd + v` and `cmd + shift + v` work

MitjaBezensek:
Add an option to make paste at cursor the default. 

Not sure if we also want to expose this on tldraw.com? For now I did, but happy to remove if we'd want to keep the preferences simple.

We could also add this to the `TldrawOptions`, but it felt like some apps might actually allow this customization on a per user level.

Solves https://github.com/tldraw/tldraw/issues/4066

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Copy / pasting should still work as it works now: `⌘ + v` pastes on top of the shape, `⌘ + ⇧ + v` pastes at cursor.
2. There's now a new option under Preferences to paste at cursor. This just swaps the logic between the two shortcuts: `⌘  + v` then pastes at cursor and `⌘ + ⇧ + v` pastes on top of the shape.

### Release notes

- Allow users and sdk users to make pasting at the cursor a default instead of only being available with `⌘ + ⇧ + v`.


steveruizok (approved):

steveruizok (approved):

steveruizok:
two approvals!

PR #4094, opened 2024-07-08
[2/4] Rename sync hooks, add bookmarks to demo

SomeHats:
Adds a new `onEditorMount` callback to the store, allowing store creators to do things like registering bookmark handlers. We use this in the new demo hook. 

This also renames `useRemoteSyncClient` to `useMultiplayerSync`, and `useRemoteSyncDemo` to `useMultiplayerDemo`.

Closes TLD-2601

### Change type

- [x] `api`


apps/bemo-worker/wrangler.toml
```diff
@@ -1,5 +1,5 @@
 main = "src/worker.ts"
-compatibility_date = "2024-06-25"
```
ds300, line 2: why down?

ds300 (approved):

apps/bemo-worker/wrangler.toml
SomeHats: i was getting warnings that this was newer than the version of wrangler we had installed

PR #4114, opened 2024-07-09
Add a toast for file upload failures.

MitjaBezensek:
Adds toasts for cases when the file type is not allowed or the file is too big.

Resolves https://github.com/orgs/tldraw/projects/53?pane=issue&itemId=70298205

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Upload a file that is either too big (over 10mb) or of incorrect file type (pdf, docx,...).
2. You should see a toast explaining what went wrong.

### Release notes

- Show a toast when uploading an unsupported file type or a file that is too large (more than 10mb).


ds300 (approved):
Thanks for finally fixing this 🥳

PR #4117, opened 2024-07-09
Add a toast for missing clipboard permissions.

MitjaBezensek:
Shows a toast notifying the user that pasting failed due to clipboard permissions. Resolves https://github.com/orgs/tldraw/projects/53?pane=issue&itemId=70294454

image

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open a new private window.
2. Copy some shapes and paste them. Make sure you don't allow the clipboard access.
3. You should see the toast.

### Release notes

- Show a toast when pasting failed due to missing clipboard permissions.


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -944,13 +944,22 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+						?.read()
+						.then((clipboardItems) => {
+							paste(
+								clipboardItems,
+								source,
+								source === 'context-menu' ? editor.inputs.currentPagePoint : undefined
+							)
+						})
+						.catch(() => {
+							addToast({
```
MitjaBezensek, line 957: Looks like `read` only throws [one error type](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read#exceptions), so even though we catch everything it should only be the `NotAllowed` error.

PR #4119, opened 2024-07-09
[3/5] Automatically enable multiplayer UI when using demo sync

SomeHats:
Adds a new `multiplayerStatus` store prop. This can either be `null` (indicating this isn't a multiplayer store) or a signal containing `online` or `offline` indicating that it is. We move a bunch of previously dotcom specific UI into `tldraw` and use this new prop to turn it on or off by default.

closes TLD-2611

### Change type

- [x] `improvement`


packages/tldraw/src/lib/ui/components/SharePanel/DefaultSharePanel.tsx
```diff
@@ -0,0 +1,10 @@
+import { PeopleMenu } from './PeopleMenu'
+
+/** @public @react */
```
ds300, line 3: `@react` is that doing anything?

ds300 (approved):

packages/tldraw/src/lib/ui/components/SharePanel/DefaultSharePanel.tsx
SomeHats: that tags this export as a react component for our documentation system which:
- puts it in a category with other react components
- pulls the documentation for the props and renders it on the same page

PR #4124, opened 2024-07-09
toSvg method example

### LINKED ISSUE #4057: toSvg method
A custom shape with a toSvg method.

Taha-Hassan-Git:
An example of how to use the toSvg method on custom shape.

Resolves #4057 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- [Examples App] added an example for the toSvg method on a custom shape.


MitjaBezensek (approved):

apps/examples/src/examples/toSvg-method-example/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Shape toSvg method
+component: ./CustomShapeToSvgExample.tsx
+category: shapes/tools
+priority: 3
+keywords: [basic, svg, custom]
```
mimecuvalo, line 6: add "export" and "copy" as well

apps/examples/src/examples/toSvg-method-example/README.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+How to use the toSVG method of a shape util to determine how your custom shapes look when copied/exported as an image.
```
mimecuvalo, line 9: `toSVG`

mimecuvalo (commented):
The example doesn't maybe show this feature off too much. It's sorta hard to tell the difference between the `toSvg` function being present or not. (it just colors the square differently).

Can we make this more visually different so that the example presents itself more clearly?

Taha-Hassan-Git:
> The example doesn't maybe show this feature off too much. It's sorta hard to tell the difference between the `toSvg` function being present or not. (it just colors the square differently).
> 
> Can we make this more visually different so that the example presents itself more clearly?

Not sure what you mean by this. The component method renders the square using html and the toSvg method renders the same square using svg, they're not supposed to look different at all.

mimecuvalo:
Yeah, sorry, I wasn't being very clear, got muddled in my msg. I think what I'm saying is maybe a different shape, not a grey rectangle, might better show off this feature, say a pink star, I don't know. In my head, I guess it's just a touch hard to discern what is meant to be the shape vs. secondary things around the shape - and something that's gray/basic can recede into the bg.

Also, might be worth having `indicator` somehow share code, too.

Taha-Hassan-Git:
> Yeah, sorry, I wasn't being very clear, got muddled in my msg. I think what I'm saying is maybe a different shape, not a grey rectangle, might better show off this feature, say a pink star, I don't know. In my head, I guess it's just a touch hard to discern what is meant to be the shape vs. secondary things around the shape - and something that's gray/basic can recede into the bg.
> 
> Also, might be worth having `indicator` somehow share code, too.

I get the point about making the shape more visually distinctive, but I think doing something with complex geometry like a star would clutter the example with concepts that are best suited to the custom geometry example. I want this to communicate a single concept: shapes rendered in HTML will need to be recreated as svgs in the toSvg method in order to take advantage of copy/export features. 

I'll look at code-sharing and make the box pink. I could also make it a circle if you feel that would help. What do you think?

mimecuvalo:
> I get the point about making the shape more visually distinctive, but I think doing something with complex geometry like a star would clutter the example with concepts that are best suited to the custom geometry example. I want this to communicate a single concept: shapes rendered in HTML will need to be recreated as svgs in the toSvg method in order to take advantage of copy/export features.
> I'll look at code-sharing and make the box pink. I could also make it a circle if you feel that would help. What do you think?

Yeah, haha, pink star might be more extreme. :P But yeah, like a pink box/circle like you suggest is maybe my point. Just a little nit but a little better :)

PR #4128, opened 2024-07-10
Remove duplicate code

MitjaBezensek:
Looks like some leftovers since we are conditionally doing the same thing on the next like. I might be missing something though.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Remove some duplicate code which should make some of the exports a bit smaller.


SomeHats (approved):
oops, good catch!

PR #4133, opened 2024-07-10
Add dialog/toasts example

Taha-Hassan-Git:
Adds an example for how to use toasts and dialogs

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- [Examples] added a toasts/dialogs example

### DIFF
--- a/apps/examples/src/examples/toasts-and-dialogs/ToastsDialogsExample.tsx
+++ b/apps/examples/src/examples/toasts-and-dialogs/ToastsDialogsExample.tsx
@@ -1,0 +1,57 @@
+import { ExampleDialog, TLComponents, Tldraw, useDialogs, useToasts } from 'tldraw'
+import 'tldraw/tldraw.css'
+
+// There's a guide at the bottom of this file
+
+const CustomSharePanel = () => {
+	const { addToast } = useToasts()
+	const { addDialog } = useDialogs()
+
+	return (
+		<button
+			style={{ fontSize: 18, backgroundColor: 'thistle', pointerEvents: 'all' }}
+			onClick={() => {
+				addToast({ title: 'Hello world!', severity: 'success' })
+				addDialog({
+					component: ({ onClose }) => (
+						<ExampleDialog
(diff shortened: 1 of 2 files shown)


apps/examples/src/examples/toasts-and-dialogs/ToastsDialogsExample.tsx
```diff
@@ -0,0 +1,57 @@
+import { ExampleDialog, TLComponents, Tldraw, useDialogs, useToasts } from 'tldraw'
```
steveruizok, line 1: Kind of strange that we export the ExampleDialog and other debug menu items. Not sure if we should be doing this. In this example, let's create a new Dialog rather than use the ExampleDialog one, so that we can show what that looks like.

Taha-Hassan-Git: Will do!

Taha-Hassan-Git:
Getting this Lint error:
```6:34  error  Function property signature is forbidden. Use a method shorthand instead  @typescript-eslint/method-signature-style```
From this line:
```typescript
function MyDialog({ onClose }: { onClose: () => void }) {
```

Does it make sense to have the Lint rule not apply to files in the examples app, or am I actually committing a typescript sin, @MitjaBezensek ?

MitjaBezensek:
> Does it make sense to have the Lint rule not apply to files in the examples app, or am I actually committing a typescript sin, @MitjaBezensek ?

In this case I would change it to:
```ts
function MyDialog({ onClose }: { onClose(): void }) {
```

It's just a different way to say the same thing.

engin-can:
@Taha-Hassan-Git  any idea why I am getting`Error: useToasts must be used within a ToastsProvider` when I call `useToasts` in my component

PR #4135, opened 2024-07-10
[5/5] Move bemo from dotcom to examples

SomeHats:
Move our bemo playground from dotcom to the examples app. In preparation for more multiplayer examples, I built our a little bit of chrome around example room IDs: if you create an example with `multiplayer: true`, the examples app will render a little room ID picker above your example. The room IDs are scoped to each example, and each deploy of the examples app. By default people on the same example will be in the same room, but the default ID changes every hour.

As I was doing this, I noticed you could get an ugly situation where the docs site was in dark mode, tldraw was in dark mode, but the little bit of examples chrome was in light mode. To fix this I through together an extremely rough dark mode for the examples which switches on whenever the tldraw instance inside is in dark mode.

### Change type

- [x] `other`


ds300:
> To fix this I through together an extremely rough dark mode for the examples which switches on whenever the tldraw instance inside is in dark mode.

this does feel very rough, and there are nasty flashes when transitioning pages. Can this be enabled only for the example content pane rather than the site as a whole? The docs site always embeds things as full screen at the moment.

ds300:
> The room IDs are scoped to each example, and each deploy of the examples app. By default people on the same example will be in the same room, but the default ID changes every hour.

I like the room ID picker, but I'm scared of putting randos together in a room for fear of the ol' dick/swastika problem.

I think we should go for something more like:

- if the url has a `?roomId=blah` param in it, use that room ID
- otherwise load the roomId from localStorage (scoped to example name) or generate a random room id if none exists in local storage
- whenever the roomId changes, store it in local storage, so you always look at the most recent one.
- have a 'copy url' button and maybe a qr code button (with copy to say what they are for)

image

SomeHats:
Good shout on the dark mode thing! I've tweaked it so it only applies to the room picker.

I'm not too worried about the shared room thing for a few reasons:
1. I think _most_ people will actually be respectful, and finding a room with some doodles in it does a pretty good job of demonstrating how multiplayer works
2. The default room changes every hour, which significantly limits the impact of that sort of vandalism - it'll be gone within an hour tops
3. It's good enough for [yjs](https://demos.yjs.dev/prosemirror/prosemirror.html)

I have added a note explaining that the contents of the board are shared between visitors which I think would help too.

In general, I really want to keep the experience here super simple. I'd rather avoid copy/qr code buttons etc, they feel like overkill for what this needs to be. The experience of
1. Open demo
2. Open it in another browser/device
3. tada

Is really i think what i wanna aim for here. I definitely recognise vandalism as a risk, but I think it's a small enough one (and one that can be mitigated with some lightweight copy) that we don't need to design around it too much.

ds300:
> I think most people will actually be respectful

I personally don't want to take this risk. The negative cost of someone seeing very NSFW stuff while at work is much higher than the cost of having a couple extra clicks to test some multiplayer behaviour IMO. At the very least I'd want to gate the act of connecting to the room behind a disclaimer and a confirmation button. The slightly lower friction is just not enough to make me sleep well at night.

mimecuvalo:
> I personally don't want to take this risk. The negative cost of someone seeing very NSFW stuff while at work is much higher than the cost of having a couple extra clicks to test some multiplayer behaviour IMO. At the very least I'd want to gate the act of connecting to the room behind a disclaimer and a confirmation button. The slightly lower friction is just not enough to make me sleep well at night.

I'd personally, sadly, agree with @ds300 - I still believe people are good at heart, but the internet also brings out the demons in people.

ds300:
I'm gonna get this merged as-is and then follow up with a refactor to make things safer. Sorry @SomeHats but having a shared public room will legit give me anxiety, doesn't matter how short-lived the content is. Just one person seeing something unpleasant and complaining about it on elon musk's X the everything app would make us switch it off anyway, I'm pretty sure. Just wanna get ahead of that.

PR #4140, opened 2024-07-11
Use shape scale for geo shape min size

steveruizok:
This PR fixes a bug where a shape created with dynamic size would preserve the origin

### Change type

- [x] `bugfix`

### Test plan

1. Turn on dynamic size
2. Zoom in
3. Create a text shape
4. Give a label
5. Try to resize the shape to be narrow

### Release notes

- Fixed a bug with the minimum size on dynamically scaled text shapes


ds300:
did you test this? doesn't seem to work in the preview deploy

steveruizok:
@ds300 should be working now, it was a bit more complex than I thought. I'd mixed scaled and unscaled values when calculating the shape's size vs. its label's size. Should all be untangled now.

ds300 (approved):

PR #4143, opened 2024-07-11
Unify menus. Disable erroring.

MitjaBezensek:
Unify the VS Code extension menus with what we have on dot com. Prevent an error cycle.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Unify the VS Code extension menus (Help and Main menus) with what we have on tldraw.com
- Prevent an onerror cycle.

### DIFF
--- a/packages/editor/src/lib/editor/managers/WatermarkManager.ts
+++ b/packages/editor/src/lib/editor/managers/WatermarkManager.ts
@@ -117,6 +117,7 @@ export class WatermarkManager {
 			// we still fallback to the local watermark.
 			this.forceLocal = true
 			this.setWatermarkSrc(watermark)
+			watermark.onerror = null
 		}
 		watermark.onclick = () => {
 			window.open('https://tldraw.dev', '_blank', 'noopener noreferrer')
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/editor/managers/WatermarkManager.ts
```diff
@@ -117,6 +117,7 @@ export class WatermarkManager {
 			// we still fallback to the local watermark.
 			this.forceLocal = true
 			this.setWatermarkSrc(watermark)
+			watermark.onerror = null
```
MitjaBezensek, line 120: Without this we could end up in a cycle: 
1. We couldn't load the cd version.
2. Error handler gets called and changes the src.
3. If this one also failed we end up in a cycle of error handler getting called -> we replace the src -> it tries to refetch the image -> error again. 

Think it's safer to just do it once.

MitjaBezensek: image

mimecuvalo: ah good call!

mimecuvalo (approved):
lgtm

PR #4146, opened 2024-07-11
[bemo] add analytics to bemo worker

ds300:
Sets up worker analytics. No further grafana setup needed, just need to start querying once this is live.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/apps/bemo-worker/src/BemoDO.ts
+++ b/apps/bemo-worker/src/BemoDO.ts
@@ -18,6 +18,7 @@ interface AnalyticsEvent {
 	type: 'connect' | 'send_message' | 'receive_message'
 	origin: string
 	sessionKey: string
+	slug: string
 }
 
 export class BemoDO extends DurableObject<Environment> {
@@ -26,9 +27,9 @@ export class BemoDO extends DurableObject<Environment> {
 
 	analytics: AnalyticsEngineDataset
 
-	writeEvent({ type, origin, sessionKey }: AnalyticsEvent) {
+	writeEvent({ type, origin, sessionKey, slug }: AnalyticsEvent) {
 		this.analytics.writeDataPoint({
-			blobs: [type, origin, sessionKey],
+			blobs: [type, origin, slug, sessionKey],
 		})
 	}
 
@@ -100,6 +101,12 @@ export class BemoDO extends DurableObject<Environment> {
 
 			// all good
 			room.handleSocketConnect(sessionKey, serverWebSocket, { origin })
+			this.writeEvent({
+				type: 'connect',
+				origin,
+				sessionKey,
+				slug: this.getSlug(),
+			})
 			return new Response(null, { status: 101, webSocket: clientWebSocket })
 		} catch (e) {
 			if (e === ROOM_NOT_FOUND) {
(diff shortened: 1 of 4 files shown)


apps/bemo-worker/wrangler.toml
```diff
@@ -84,3 +84,26 @@ binding = "CF_VERSION_METADATA"
...
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS_PREVIEW"
+
+[[env.staging.analytics_engine_datasets]]
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS_STAGING"
+
+[[env.production.analytics_engine_datasets]]
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS_PRODUCTION"
```
ds300: so our dotcom worker only has one dataset called `MEASURE` and we add the worker name as part of the row data when sending events, so you can query individual preview deploys. But to me that doesn't seem useful and I think I'd rather just have different datasets for each env. WDYT?

apps/bemo-worker/wrangler.toml
```diff
@@ -84,3 +84,26 @@ binding = "CF_VERSION_METADATA"
...
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS_PREVIEW"
+
+[[env.staging.analytics_engine_datasets]]
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS_STAGING"
+
+[[env.production.analytics_engine_datasets]]
+binding = "BEMO_ANALYTICS"
+dataset = "BEMO_ANALYTICS"
```
ds300, line 109: so our dotcom worker only has one dataset called MEASURE and we add the worker name as part of the row data when sending events, so you can query individual preview deploys. But to me that doesn't seem useful and I think I'd rather just have different datasets for each env. WDYT?

apps/bemo-worker/src/BemoDO.ts
```diff
@@ -13,16 +14,31 @@ const connectRequestQuery = T.object({
...
+
 export class BemoDO extends DurableObject<Environment> {
 	r2: R2Bucket
 	_slug: string | null = null
 
+	analytics: AnalyticsEngineDataset
+
+	writeEvent({ type, origin, sessionKey }: AnalyticsEvent) {
+		this.analytics.writeDataPoint({
+			blobs: [type, origin, sessionKey],
```
SomeHats, line 31: what's the value in writing the session key? I could maybe see room ID, but i'm not sure where sessionkey would be actionable

apps/bemo-worker/wrangler.toml
SomeHats: this is way better i don't think i realised this existed

SomeHats (approved):

apps/bemo-worker/src/BemoDO.ts
ds300: i was thinking it would allow us to see how many different clients there independently of how many connect events there are. but also yes I added the room id, silly me for forgetting it.

mimecuvalo:
Any interest in using Datadog instead of Grafana? (since we have that setup for our assets analytics anyway)

PR #4147, opened 2024-07-11
State/store example

Taha-Hassan-Git:
An example of how to subscribe to values in the store and run side effects using reactors.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added an example that shows how to use track, useValue and useReactor

### DIFF
--- a/apps/examples/src/examples/state-store/StateStoreExample.tsx
+++ b/apps/examples/src/examples/state-store/StateStoreExample.tsx
@@ -1,0 +1,98 @@
+import { TLComponents, Tldraw, TLShape, track, useEditor, useQuickReactor } from 'tldraw'
+import 'tldraw/tldraw.css'
+
+// [1]
+const InfoPanel = track(() => {
+	const editor = useEditor()
+	const tool = editor.getCurrentToolId()
+	const zoom = editor.getZoomLevel().toFixed(2)
+	useQuickReactor(
+		'only red shapes',
+		() => {
+			const shapes = editor.getCurrentPageShapes()
+
+			editor.updateShapes(
+				shapes.map((shape: TLShape) => {
+					if ('color' in shape.props) {
+						return { ...shape, props: { ...shape.props, color: 'red' } }
+					} else {
+						return shape
+					}
+				})
+			)
+		},
+		[editor]
+	)
+	return (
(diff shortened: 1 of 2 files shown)


apps/examples/src/examples/state-store/StateStoreExample.tsx
```diff
@@ -0,0 +1,99 @@
...
+					if ('color' in shape.props) {
+						return { ...shape, props: { ...shape.props, color: 'red' } }
+					} else {
+						return shape
+					}
+				})
+			)
+		},
+		[editor]
+	)
```
ds300, line 25: this isn't a great example of `useQuickReactor`. This kind of thing would typically be done with the store sideEffects api.

I think a better option would be updating `document.title` to say how many shapes there are on the canvas. I also think the example should be for regular old `useReactor` instead of `useQuickReactor` which is only really useful for us to avoid tearing while things move around.

apps/examples/src/examples/state-store/StateStoreExample.tsx
```diff
@@ -0,0 +1,99 @@
...
+	return (
+		<div className="tldraw__editor">
+			<Tldraw components={components} />
+		</div>
+	)
+}
+
+/* 
+
+Tldraw uses a signals library called Signia to manage its state and store. You can subscribe to
```
ds300, line 79: this isn't really true any more
```suggestion
Tldraw uses a signals to manage its state and store. You can subscribe to
```

apps/examples/src/examples/state-store/StateStoreExample.tsx
```diff
@@ -0,0 +1,89 @@
...
+		<div style={{ pointerEvents: 'all', backgroundColor: 'thistle', fontSize: 14, padding: 8 }}>
+			<div>{tool}</div>
+			<div>{zoom}</div>
+		</div>
+	)
+}
+
+const components: TLComponents = {
+	SharePanel: InfoPanel,
+	// SharePanel: AlternativeInfoPanel,
```
ds300, line 56:
```suggestion
```

ds300 (approved):
🎉

PR #4150, opened 2024-07-11
perf: serve icons via a single merged .svg file

SomeHats:
We currently serve icons in the form of 141 separate svg files. This is pretty inefficient: it results in 141 http requests each with its own overheads and costs. Each one is generally less than 1kb of data.

This PR merges all of our icons into a single 44kb icon file. Each item in the svg has an ID, and we use the ID in the asset url to target the specific icon. This also fixes a lot of the icon preloading issues: because the icon file is already loaded, there's no extra request needed when panels open etc.

I was messing with this whilst killing time before a late meeting so made a shitty overengineered code generator, dont at me

### Change type

- [x] `improvement`

### Test plan

1. Open the app and verify icons load correctly.
2. Check network tab to ensure icons are served from a single sprite file.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Serve icons more efficiently, and make sure they're still available if tldraw goes offline.


mimecuvalo:
> I was messing with this whilst killing time before a late meeting so made a shitty overengineered code generator, dont at me

LOL 😆

apps/dotcom/src/utils/assetUrls.ts
```diff
@@ -7,8 +7,13 @@ let didPreloadIcons = false
 async function preloadIcons() {
 	if (didPreloadIcons) return
 	didPreloadIcons = true
-	const urlsToPreload = [...Object.values(assetUrls.icons), ...Object.values(assetUrls.embedIcons)]
-	await Promise.allSettled(urlsToPreload.map(preloadIcon))
+	const urlsToPreload = new Set(
+		[...Object.values(assetUrls.icons), ...Object.values(assetUrls.embedIcons)].map((url) =>
+			url.replace(/#.*$/, '')
```
mimecuvalo, line 12: comment here would be good for this regex

scripts/refresh-assets.ts
```diff
@@ -25,18 +25,21 @@ const FONT_MAPPING: Record<string, string> = {
...
+> = {
 	fonts: {},
 	icons: {},
 	translations: {},
 	embedIcons: {},
 }
 
+const mergedIconName = '0_merged.svg'
+const mergedIconHeader =
+	'<svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" fill="none">'
```
mimecuvalo, line 40: does the width/height/fill here do anything?

scripts/refresh-assets.ts
```diff
@@ -58,6 +61,29 @@ async function copyIcons() {
 		return { fileName: icon, data: svg.data }
 	})
 
+	// Merge svgs into a single file
+
+	const mergedSvgParts = [
+		mergedIconHeader,
+		'<style>',
+		'svg > [id] { display: none; }',
```
mimecuvalo, line 69: comment here would be good on what it's trying to achieve

scripts/refresh-assets.ts
```diff
@@ -58,6 +61,29 @@ async function copyIcons() {
...
+			nicelog('SVG does not match expected format for merging:', fileName)
+			process.exit(1)
+		}
+
+		const id = fileName.replace('.svg', '')
+		const svgContent = data.slice(mergedIconHeader.length, -mergedIconFooter.length)
+		mergedSvgParts.push(`<g id="${id}">${svgContent}</g>`)
+	}
+	mergedSvgParts.push(mergedIconFooter)
+	const mergedSvg = optimize(mergedSvgParts.join('\n')).data
```
mimecuvalo, line 84: this is prbly fine but just checking - `optimize` is a bit of a black box to me - now that the assets are merged, it doesn't over-eagerly optimize/weirdly combine two assets that are perhaps side-by-side?

scripts/refresh-assets.ts
```diff
@@ -480,6 +499,100 @@ async function copyVersionToDotCom() {
...
+		let name = this.imports.get(file)
+		if (!name) {
+			name = this.getName(file)
+			this.imports.set(file, name)
+		}
+		return name
+	}
+
+	getName(name: string) {
+		return `$_${name.replace(/[^a-zA-Z0-9_]+/g, '_')}_${this.nameId++}`
```
mimecuvalo, line 544: any reason not to just use `/\W/g`?

mimecuvalo (approved):
neeeeat! we've come full circle 20 years later to creating sprites for web pages. 😆 
couple minor comments, ship it yo!

ds300 (approved):
love this. proper icon preloading at last!

scripts/refresh-assets.ts
SomeHats: yes, they determine the size of everything in here and how fills work. this is what svgo outputs for all of our icons, and we need to make sure they all match. I've added a comment to clarify.

scripts/refresh-assets.ts
SomeHats: added

scripts/refresh-assets.ts
SomeHats: i can never remember what these single-letter shortcuts correspond to and have to look them up every time 😅 changed

PR #4170, opened 2024-07-15
Split @tldraw/state into @tldraw/state and @tldraw/state-react

ds300:
The backend code uses `@tldraw/state`, which is fine, but the package has a peer dependency on `react`, which is not fine to impose on backend consumers. So let's split this up again.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


SomeHats:
i don't think this is worth it. plain `react` without the dom stuff which isn't used in signia anyway is about 6kb, and works perfectly fine in all the server environments. what's the issue with including it?

SomeHats (approved):
(still approving cus like... its fine :P)

ds300:
> is it just the peer dep warning

Yeah. We shouldn't have to explain that to people.

PR #4173, opened 2024-07-15
Example node + bun server

ds300:
This PR adds an app with examples for building a tlsync backend in both node and bun.

Highlights
- simple vite frontend
- node backend using `fastify` (popular modern alternative to express, and much better api for websockets)
- bun backend using `itty-router` and `Bun.serve`. Interestingly Bun has a similar websockets api to the cloudflare hibernatable sockets api, which required some simplification of the socket argument on `TLSocketRoom`.
- updated itty-router and got rid of itty-cors. There were some minor breaking API changes to account for, and a new cors api that allows us to remove an ugly hack.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


apps/simple-server-example/package.json
```diff
@@ -0,0 +1,56 @@
...
+		]
+	},
+	"dependencies": {
+		"@fastify/cors": "^9.0.1",
+		"@fastify/websocket": "^10.0.1",
+		"@tldraw/sync": "workspace:*",
+		"@tldraw/sync-core": "workspace:*",
+		"@vitejs/plugin-react-swc": "^3.7.0",
+		"cheerio": "^1.0.0-rc.12",
+		"express": "^4.19.2",
```
SomeHats, line 46: do we need both express & fastify?

apps/simple-server-example/src/client/client.tsx
```diff
@@ -0,0 +1,92 @@
...
+	Route,
+	RouterProvider,
+	createBrowserRouter,
+	createRoutesFromElements,
+	useParams,
+} from 'react-router-dom'
+import { AssetRecordType, Editor, TLAsset, TLAssetStore, Tldraw, getHashForString } from 'tldraw'
+import 'tldraw/tldraw.css'
+
+const assets: TLAssetStore = {
```
SomeHats, line 14: i'd maybe consider moving `asset` + `unfurl` below `App` in this file. I feel like from a "reading this example for the first time" pov you want `App` up front to get the big picture of what the parts are and how they fit together, then continue reading to delve into the implementation details of assets/url unfurling

apps/simple-server-example/src/client/client.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+function App() {
+	const { roomId } = useParams()
+	const store = useMultiplayerSync({
+		uri: 'ws://localhost:5858/connect/' + roomId,
+		assets,
+		onEditorMount: useCallback((editor: Editor) => {
+			editor.registerExternalAssetHandler('url', unfurl)
+		}, []),
```
SomeHats, line 61: unless we have a good reason to, lets stick with the normal component `onMount` callback. it means we can skip the `useCallback` and will be more familiar wrt the other examples

apps/simple-server-example/src/client/client.tsx
```diff
@@ -0,0 +1,92 @@
...
+	})
+	return <Tldraw store={store} />
+}
+
+createRoot(document.getElementById('root')!).render(
+	<RouterProvider
+		router={createBrowserRouter(
+			createRoutesFromElements(
+				<>
+					<Route path="/:roomId" Component={() => <App />} />
```
SomeHats, line 71: in my workers starter, i just hard-coded a room ID so i could skip adding a router etc.

i think it helps the example focus on what it's meant to be demonstrating a little better. im not adverse to including a router, but what do you think?

apps/simple-server-example/src/server/rooms.ts
```diff
@@ -0,0 +1,93 @@
+import { RoomSnapshot, TLSocketRoom } from '@tldraw/sync-core'
+import { mkdir, readFile, writeFile } from 'fs/promises'
+import { join } from 'path'
+
+// eslint-disable-next-line no-console
```
SomeHats, line 5: let's add an override in eslintrc to disable no-console for this whole app, then we can write console.log normally

SomeHats (approved):
looking good!

apps/bemo-worker/src/worker.ts
```diff
@@ -11,16 +11,16 @@ import {
...
 
 export { BemoDO } from './BemoDO'
 
-const cors = createCors({ origins: ['*'] })
+const { preflight, corsify } = cors({ origin: '*' })
 
 export default class Worker extends WorkerEntrypoint<Environment> {
 	private readonly router = Router()
-		.all('*', cors.preflight)
+		.all('*', preflight)
```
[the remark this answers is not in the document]
ds300: it's the same behaviour i just couldn't be bothered to make the change to AutoRouter since it's not required and would have been a bigger diff. we can follow up after this PR

apps/dotcom-worker/src/worker.ts
```diff
@@ -84,12 +84,13 @@ export default class Worker extends WorkerEntrypoint<Environment> {
 }
 
 export function isAllowedOrigin(origin: string) {
-	if (origin === 'http://localhost:3000') return true
-	if (origin === 'http://localhost:5420') return true
-	if (origin === 'https://meet.google.com') return true
-	if (origin.endsWith('.tldraw.com')) return true
-	if (origin.endsWith('-tldraw.vercel.app')) return true
-	return false
+	if (!origin) return undefined
```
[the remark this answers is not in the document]
ds300: no that's handled separately I think, further down in this file, in `blockUnknownOrigins`.

from what I can tell reading the cors middleware this remains correct (no origin given in request means no access-control-allow-origin header is set on the response), but we should def test with the ipad

apps/simple-server-example/package.json
ds300: good catch

apps/simple-server-example/src/client/client.tsx
ds300: I think I'm just gonna copy your client code and hard code the urls

apps/simple-server-example/src/client/client.tsx
ds300: copied your client code

apps/simple-server-example/src/client/client.tsx
ds300: i copied your client code

ds300:
gonna merge this and follow up with docs + comments

PR #4174, opened 2024-07-15
bemo custom shape example

SomeHats:
Adds a custom shape bemo example. Instead of having to create a schema ahead of time, here i've added bindingUtils/shapeUtils props to our sync hooks to match the ones we have in the `<Tldraw />` component. Not 100% about this though, I could easily be convinced to go with just the schema prop.

### Change type

- [x] `other`


apps/examples/src/ExampleWrapper.tsx
```diff
@@ -92,14 +92,16 @@ function MultiplayerExampleWrapper({
...
 					onClick={() => {
 						// copy current url with roomId=roomId to clipboard
 						navigator.clipboard.writeText(window.location.href.split('?')[0] + `?roomId=${roomId}`)
 						confirmFn()
 					}}
 					aria-label="join"
 				>
-					{confirm ? 'Copied!' : 'Copy Share URL'}
+					Copy link
+					{confirm && <div className="MultiplayerExampleWrapper-copied">Copied!</div>}
```
ds300, line 104: nice 👍🏼

apps/examples/src/examples/multiplayer-custom-shape/ui.tsx
```diff
@@ -0,0 +1,38 @@
...
+			},
+		}
+		return tools
+	},
+}
+
+export const components: TLComponents = {
+	Toolbar: (props) => {
+		const tools = useTools()
+		const isCardSelected = useIsToolSelected(tools['counter'])
```
ds300, line 30:
```suggestion
		const isCounterSelected = useIsToolSelected(tools['counter'])
```

packages/editor/src/lib/config/createTLStore.ts
```diff
@@ -55,12 +54,40 @@ export const defaultAssetStore: TLAssetStore = {
...
 }
 
+/**
+ * A helper for creating a TLStore schema.
+ *
+ * @param opts - Options for creating the schema.
+ *
+ * @public
+ */
+export function createTLStoreSchema(
```
ds300, line 64: I'm a bit worried about the naming similarity to `createTLSchema` here. Can you think of a name that more clearly distinguishes the behaviour of this? (and the hook I guess)?

ds300 (approved):
i'm fine with shapeUtils over schema. I do think that whole aspect of our API is a mess and internally inconsistent, but at this point it would be weirder (or more unexpected) to only have a schema prop.

PR #4175, opened 2024-07-15
publish bemo canaries

SomeHats:
Switch on package publishing for sync libraries so we can start building templates on the canaries. 

### Change type

- [x] `other`


ds300 (commented):
seems good to me. Before we publish I'll do a pass of writing TSDoc comments, revising what's public/internal, and maybe massage some api things.

PR #4180, opened 2024-07-16
Fix watermark imports in published packages

SomeHats:
Different bundlers have different support for importing non-JS assets. We can't reliably import SVGs in our public packages and rely on that to do the right thing. When I tried installing a canary to work on a new cloudflare template, vite was erroring out due to those imports. I had to patch them out to get things running.

This diff replaces the file-based assets with inlined svg-in-js strings. I haven't been able to test this though because I'm not really sure how. @mimecuvalo could you take a look?

### Change type

- [x] `bugfix`


mimecuvalo (approved):
ughh, annoying but yeah this makes sense to me. thanks for fixing up! cc @MitjaBezensek as fyi

PR #4181, opened 2024-07-16
[Experiment] ui: rethink pager component

mimecuvalo:
messing around with how the pager menu could be different.
- provides easy create new page UI (with the right chevron)
- lets you go back-and-forth very easily
- adds thumbnails to the pager element to see what's inside each page, #burger-king style @SomeHats ;)

Screenshot 2024-07-16 at 10 08 25

Screenshot 2024-07-16 at 18 03 12

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- UI: redo the pager menu


MitjaBezensek:
Played around with this and here are some impressions:
* I really like how this cleans up the menu zone, it's much more tight.
* The pages menu feels a bit too prominent in the middle. Not sure if switching pages is used so often that it would justify that placement. (Wanted to check analytics but found out we weren't tracking many of the page actions).
* Switching between pages feels quite abrupt. The first time I used it I was like where did my content go? 😄 Generally not fan of animations but was wondering how a page curl / fade out animation would feel like. Might also fit with the pages icon theme you went with. Could just be that I'd need to get used to it, but I'm sure some users would have the same reaction.
* We'd probably also want to reconsider the pages menu. Feels like there are too many ways to do the same thing.
* The button to open the pages menu should be a bit more discoverable.
* Feels like we lose something by not showing the page name. I feel like indexes don't work well here.

mimecuvalo:
> Switching between pages feels quite abrupt. The first time I used it I was like where did my content go? 😄 Generally not fan of animations but was wondering how a page curl / fade out animation would feel like. Might also fit with the pages icon theme you went with. Could just be that I'd need to get used to it, but I'm sure some users would have the same reaction.

Very true - I think the difference between this and Jamboard is the animation and, yeah, it might help in this case!

> We'd probably also want to reconsider the pages menu. Feels like there are too many ways to do the same thing.

Definitely!

> The button to open the pages menu should be a bit more discoverable.

Yeah, the Jamboard equivalent has a down chevron - that would help.

> Feels like we lose something by not showing the page name. I feel like indexes don't work well here.

Eh, I would disagree on this maybe but it's a tradeoff. I think it could be argued either way potentially.

PR #4188, opened 2024-07-16
Interpolated line points

Taha-Hassan-Git:
Line shape points animate pretty nicely

2024-07-16 at 12 21 17 - Blush Limpet

I wanted to do the same for the spline, but I think it would require messing around with how the props are structured. My idea would be to turn spline into a number between 0 and 1, and have the SVG render lerped points between the Cubic Bezier and Line edges. 

It seemed like quite an intense change for a feature that would rarely be used so I decided to skip it. 

If there's a better way, let me know!

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


SomeHats:
whooaoaaaaahhhhh look at it go!!!!!

SomeHats:
can we add the sound effects from this https://www.youtube.com/watch?v=aUdCS1lLiYs

packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
```diff
@@ -185,6 +186,23 @@ export class LineShapeUtil extends ShapeUtil<TLLineShape> {
...
 	}
+	override getInterpolatedProps(
+		startShape: TLLineShape,
+		endShape: TLLineShape,
+		progress: number
+	): TLLineShape['props'] {
+		return {
+			...endShape.props,
+			points: {
+				...mapObjectMapValues(startShape.props.points, (_, { id, index, x, y }) => ({
```
SomeHats, line 197: I believe this will give spooky NaN values if there are less points in the end shape than in the start shape.

If you wanted to get the multi-point lines working, I think the way I would approach this would be to first say that if the spline is cubic and the start/end don't have the same number of point, we skip lerping entirely. cubic splines are spooky, and we don't have enough information in our props to interpolate between them well. 

I think it would be fine for us to just do the same thing for line interpolation, but if you wanted to support variable numbers of points for line segments, it would be possible to make that look good. You have to start by making sure that the two lines have the same number of points. The easiest way to do this is by repeating the last point of the shorter list until they're the same length. 

The hard way that produces better results is to insert points between existing ones on the shorter line in a way that attempts to keep the relative distances between points on the two lines similar. Almost certainly not worth it though! but also maybe fun

Taha-Hassan-Git: Hmm, thanks for the insight. Will need a bit more clarity to be able to implement what you've suggested, maybe we could have a brief call later? I'll convert this back to a draft in the meantime.

steveruizok:
Kapture 2024-07-18 at 14 04 06

I think I got it!

Taha-Hassan-Git:
> Kapture 2024-07-18 at 14 04 06 [ Kapture 2024-07-18 at 14 04 06 ](https://private-user-images.githubusercontent.com/23072548/349968606-6747b115-025f-4b9f-8655-4661963d7fcf.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjEzMTA5MDgsIm5iZiI6MTcyMTMxMDYwOCwicGF0aCI6Ii8yMzA3MjU0OC8zNDk5Njg2MDYtNjc0N2IxMTUtMDI1Zi00YjlmLTg2NTUtNDY2MTk2M2Q3ZmNmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzE4VDEzNTAwOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJmNTgwMDdhZDlmODM1NmY4NTY3MTQ1OGJkMDg4OGQ4NDM4ODZhYjRkODM0MWJiODI4ZDRiMmEyYjNjZWFjYWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.t5_PL5hg9XWOml1EHMpad7pIrH8rcnkdfhLQ-aIMwDU) [ ](https://private-user-images.githubusercontent.com/23072548/349968606-6747b115-025f-4b9f-8655-4661963d7fcf.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjEzMTA5MDgsIm5iZiI6MTcyMTMxMDYwOCwicGF0aCI6Ii8yMzA3MjU0OC8zNDk5Njg2MDYtNjc0N2IxMTUtMDI1Zi00YjlmLTg2NTUtNDY2MTk2M2Q3ZmNmLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzE4VDEzNTAwOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJmNTgwMDdhZDlmODM1NmY4NTY3MTQ1OGJkMDg4OGQ4NDM4ODZhYjRkODM0MWJiODI4ZDRiMmEyYjNjZWFjYWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.t5_PL5hg9XWOml1EHMpad7pIrH8rcnkdfhLQ-aIMwDU)
> 
> I think I got it!

Woaah! Looking at your solution I can see I was going about it the wrong way. Thanks!

Issue #3999, opened 2024-06-24
Mismatching types


merobal:
### What happened?

For `loadSnapshot` the fields of `TLEditorSnapshot` are optional, see `_snapshot: Partial<TLEditorSnapshot>`

but for `TldrawImage` and `useTLStore` (and at few other places) all fields of `TLEditorSnapshot` are required, see `snapshot: TLEditorSnapshot`

but the snapshot is loaded to `loadSnapshot`, so it seems it can be safely optional.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

EryouHao:
I encountered the same problem, have you solved it?

PR #4190, opened 2024-07-16
Relax the params

MitjaBezensek:
Feels like we can relax what we expect? It gets passed to `loadSnapshot` which allows for partials.

Resolves https://github.com/tldraw/tldraw/issues/3999

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Allow passing partial `TLEditorSnapshot` to `TldrawImage` and `useTLStore`.

### DIFF
--- a/packages/editor/src/lib/hooks/useTLStore.ts
+++ b/packages/editor/src/lib/hooks/useTLStore.ts
@@ -11,7 +11,7 @@ import {
 
 /** @public */
 type UseTLStoreOptions = TLStoreOptions & {
-	snapshot?: TLEditorSnapshot | TLStoreSnapshot
+	snapshot?: Partial<TLEditorSnapshot> | TLStoreSnapshot
 }
 
 function createStore(opts: UseTLStoreOptions) {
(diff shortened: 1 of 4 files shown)


packages/editor/src/lib/hooks/useTLStore.ts
```diff
@@ -24,7 +24,7 @@ function createStore(opts: UseTLStoreOptions) {
 
 /** @public */
 export function useTLStore(
-	opts: TLStoreOptions & { snapshot?: TLEditorSnapshot | TLStoreSnapshot }
+	opts: TLStoreOptions & { snapshot?: Partial<TLEditorSnapshot> | TLStoreSnapshot }
```
MitjaBezensek, line 27: Here we could even use `UseTLStoreOptions`, but maybe we want to expose the exploded type for better DX?

ds300: Would make a difference in this case? I'm fine either way.

ds300 (approved):

packages/editor/src/lib/hooks/useTLStore.ts
MitjaBezensek: No, not really 🤷‍♂️

PR #4196, opened 2024-07-16
Watermark II

steveruizok:
This PR is a second go at the watermark.

localhost_5420_develop (1)

It:
- updates the watermark icon
- removes the watermark on small devices
- makes the watermark a react component with inline styles
  - the classname for these styles is based on the current version
- improves the interactions around the watermark
  - the watermark requires a short delay before accepting events
  - events prior to that delay will be passed to the canvas
  - events after that delay will interact with the link to tldraw.dev
  - prevents interactions with the watermark while a menu is open
  - moves the watermark up when debug mode is active

It also:
- moves the "is unlicensed" logic into the license manager
- adds the license manager as a private member of the editor class
- removes the watermark manager

## Some thoughts

I couldn't get the interaction we wanted from the watermark manager itself. It's important that this just the right amount of disruptive, and accidental clicks seemed to be a step too far. After some thinking, I think an improved experience is worth a little less security.

Using a React component (with CSS styling) felt acceptable as long as we provided our own "inline" style sheet. My previous concern was that we'd designed a system where external CSS is acceptable, and so would require other users to provide our watermark CSS with any of their own CSS; but inline styles fix that.

### Change type

- [x] `other`

### Test plan

1. Same as the previous watermark tests.

- [x] Unit tests


packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -516,3 +524,118 @@ export function useOnMount(onMount?: TLOnMountHandler) {
...
+
+const WATERMARK_SRC = `url('data:image/svg+xml;utf8,${encodeURIComponent(watermarkDesktopSvg)}') center 100% / 100% no-repeat`
+
+const Watermark = React.memo(() => {
+	const events = useCanvasEvents()
+
+	const ref = useRef<HTMLAnchorElement>(null)
+	useLayoutEffect(() => {
+		// eslint-disable-next-line deprecation/deprecation
+		if (ref?.current) ref.current.style.webkitMask = WATERMARK_SRC
```
mimecuvalo, line 536: can't just do `webkitMask` - we have to also do `mask`

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -516,3 +524,118 @@ export function useOnMount(onMount?: TLOnMountHandler) {
...
+	}, [])
+
+	const editor = useEditor()
+	const showWatermark = useValue(
+		'show watermark',
+		() => {
+			const { width } = editor.getViewportScreenBounds()
+			if (width < 760) return false
```
mimecuvalo, line 544: why 760? we should keep parity with:
`export const PORTRAIT_BREAKPOINTS = [0, 390, 428, 468, 580, 640, 840, 1023]`

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -516,3 +524,118 @@ export function useOnMount(onMount?: TLOnMountHandler) {
...
+			if (!featureFlags.enableLicensing.get()) return false
+			return editor.getLicenseState() === 'unlicensed'
```
mimecuvalo, line 546: using strings like this is a bit brittle. let's use const, enum-like set of strings so this doesn't get messed up and can be type-checked better

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -300,4 +328,6 @@ export class LicenseManager {
 			`color: white; background: crimson; padding: 2px; border-radius: 3px;`
 		)
 	}
+
+	static classname = 'tl-watermark_SEE-LICENSE'
```
mimecuvalo, line 332: nit: might as well make this consistent with react-type casing to be `className`

packages/editor/src/lib/editor/managers/WatermarkManager.test.ts
```diff
@@ -1,149 +0,0 @@
-import { createTLStore } from '../../config/createTLStore'
```
mimecuvalo, line 1: one of the nice things about keeping this separate is that we were able to make this testable. we don't want to inadvertently make the watermark show up when you've actually paid (and vice-versa we want to make sure it shows up when it should), it could be very easy to mess this up!

recommendation: we pull out the <Watermark /> react component into a separate file so that we can easily add these tests back in and make sure we're doing the correct thing.

packages/editor/src/lib/editor/managers/LicenseManager.ts
```diff
@@ -50,16 +51,43 @@ export interface ValidLicenseKeyResult {
...
+		licenseKey: string | undefined,
+		testPublicKey?: string,
+		testEnvironment?: TestEnvironment
+	) {
 		this.isDevelopment = this.getIsDevelopment(testEnvironment)
 		this.publicKey = testPublicKey || this.publicKey
 		this.isCryptoAvailable = !!crypto.subtle
+
+		this.getLicenseFromKey(licenseKey).then((result) => {
+			function isEditorUnlicensed() {
```
mimecuvalo, line 72: we need to make sure this stuff is tested, it's pretty key

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -709,21 +708,33 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+						new Error(
+							'Watermark not found on unlicensed editor. Please see our license for more information.'
+						)
+					)
+				}
+			}
 		}
-		checkLicenseKey(licenseKey)
+		this.timers.setTimeout(youWouldntStealAWhiteboard, 5000)
+		this.timers.setInterval(youWouldntStealAWhiteboard, 5 * 60 * 1000)
```
mimecuvalo, line 725: just the interval is sufficient, don't need both

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -709,21 +708,33 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-			const licenseManager = new LicenseManager()
-			const watermarkManager = new WatermarkManager(this)
-			const license = await licenseManager.getLicenseFromKey(licenseKey)
-			watermarkManager.checkWatermark(license)
+		const youWouldntStealAWhiteboard = () => {
+			if (this.getLicenseState() === 'unlicensed') {
+				if (!this.getContainer().querySelector('.' + LicenseManager.classname)) {
+					this.crash(
+						new Error(
+							'Watermark not found on unlicensed editor. Please see our license for more information.'
```
mimecuvalo, line 718: translation needed or no?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -709,21 +708,33 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 
-		const checkLicenseKey = async (licenseKey: string | undefined) => {
-			if (!featureFlags.enableLicensing.get()) return
+		this.licenseManager = new LicenseManager(licenseKey)
 
-			const licenseManager = new LicenseManager()
-			const watermarkManager = new WatermarkManager(this)
-			const license = await licenseManager.getLicenseFromKey(licenseKey)
-			watermarkManager.checkWatermark(license)
+		const youWouldntStealAWhiteboard = () => {
```
mimecuvalo, line 713: just making sure: no more feature flag? official launch?

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -516,3 +524,118 @@ export function useOnMount(onMount?: TLOnMountHandler) {
 		if (editor) return onMountEvent?.(editor)
 	}, [editor, onMountEvent])
 }
+
+const WATERMARK_SRC = `url('data:image/svg+xml;utf8,${encodeURIComponent(watermarkDesktopSvg)}') center 100% / 100% no-repeat`
```
mimecuvalo, line 528: let's move this to another file, please for testability

apps/examples/src/misc/develop.tsx
```diff
@@ -8,7 +7,7 @@ export default function Develop() {
 	return (
 		<div className="tldraw__editor">
 			<Tldraw
-				licenseKey={getLicenseKey()}
+				// licenseKey={getLicenseKey()}
```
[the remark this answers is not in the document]
steveruizok: erp, dev stuff

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -516,3 +524,118 @@ export function useOnMount(onMount?: TLOnMountHandler) {
...
+		},
+		[editor]
+	)
+	const isDebugMode = useValue('debug mode', () => editor.getInstanceState().isDebugMode, [editor])
+	const isMenuOpen = useValue('is menu open', () => editor.getIsMenuOpen(), [editor])
+
+	if (!showWatermark) return null
+
+	const className = LicenseManager.classname
+	const imageClassName = LicenseManager.classname + lns(`watermark${version.replace(/\./g, '')}`)
```
mimecuvalo, line 556: `lns` would work here but you shouldn't put the `LicenseManager.classname` prefix on it, otherwise it's too easy to query select on. here's why:

you can just find that easily with an attribute selector (that looks for that string in the attribute:
`'[class^="tl-watermark_SEE-LICENSE"]'` or `'[class*="tl-watermark_SEE-LICENSE"]'`

packages/editor/src/lib/editor/Editor.ts
steveruizok: The logic here would be "once after five seconds, then every few minutes". Maybe the once after five seconds is enough actually?

packages/editor/src/lib/editor/managers/WatermarkManager.test.ts
steveruizok: That makes sense!

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: The breakpoints are at the UI layer. I picked a number semi-arbitrarily (that still fits with out UI).

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: Good call

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: ok, will remove this one

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -709,21 +708,19 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-		const checkLicenseKey = async (licenseKey: string | undefined) => {
-			if (!featureFlags.enableLicensing.get()) return
-
-			const licenseManager = new LicenseManager()
-			const watermarkManager = new WatermarkManager(this)
-			const license = await licenseManager.getLicenseFromKey(licenseKey)
-			watermarkManager.checkWatermark(license)
-		}
-		checkLicenseKey(licenseKey)
+		this.licenseManager = new LicenseManager(licenseKey)
```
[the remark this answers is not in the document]
steveruizok: This is a tricky one, as we'd need the watermark component to access it somehow. We could provide it with react context?

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: That whole thing feels unnecessary tbh, binning it

steveruizok:
Couple updates:

- moved Watermark into its own file
- created a test file for the Watermark (copied in the old tests so that we can compare)
- removed the license check after 5s / interval, since they'd always fail on mobile and doesn't seem necessary

steveruizok:
Pulled the help menu again, it's back in the corner now.

packages/editor/src/lib/editor/Editor.ts
steveruizok: My mistake (removed this bit now anyway)

packages/editor/src/lib/editor/managers/LicenseManager.ts
steveruizok: Extracted this to a separate function. I'll adapt the tests to make sure it produces the right output.

steveruizok:
Not sure if the watermark tests will work given that the license parsing requires some funky APIs that aren't present in Jest.

MitjaBezensek:
> Ok, thinking now about the CDN thing:
> 
> I'd prefer to keep the watermark out of the library if we can. Is there a reason why we wouldn't try to load this always from the CDN?

That's how we had it before. We had a check for online via `navigator.online` and we then only used the local one for offline cases.

packages/editor/src/lib/TldrawEditor.tsx
mimecuvalo: Ok, gotcha - I think that's worth a comment then... feels like this'll break at some point if/when we shift the UI and we'll forget about this.

packages/editor/src/lib/TldrawEditor.tsx
mimecuvalo: I see this is resolved - is it because the other place we do `webkitMask` in the code has this comment?
```

			// HACK: Fix for <https://linear.app/tldraw/issue/TLD-1700/dragging-around-with-the-handtool-makes-lots-of-requests-for-icons>
			// It seems that passing `WebkitMask` to react will cause a render on each call, no idea why... but this appears to be the fix.
			// @ts-ignore
			// eslint-disable-next-line deprecation/deprecation
```

prbly worth adding the comment here, too - since it seems this is trying to work around something.

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: yeah i think the once after 5 sec is enough. for the interval, i mean ¯\\_(ツ)_/¯ if people are gonna work around it, they'll just honestly fork the code

mimecuvalo:
> I'd prefer to keep the watermark out of the library if we can. Is there a reason why we wouldn't try to load this always from the CDN?

Yeah, we need both, one if you're offline (or, say, blocking that request to circumvent things) and if you're online.

> Not sure if the watermark tests will work given that the license parsing requires some funky APIs that aren't present in Jest. I've skipped those tests currently but I've added them in case anyone can figure them out. (The license manager's state gets stuck in pending)

I can look at the tests and fix up (the APIs are available in Jest, just have to include the `crypto` libraries though)

packages/tldraw/src/lib/shapes/shared/usePrefersReducedMotion.tsx
```diff
@@ -5,6 +5,7 @@ export function usePrefersReducedMotion() {
 	const [prefersReducedMotion, setPrefersReducedMotion] = useState(false)
 
 	useEffect(() => {
+		if (typeof window === undefined || !('matchMedia' in window)) return
```
mimecuvalo, line 8: looks like there's some copy-pasta bug here in the codebase - pushing up a fix to this PR.

should be: `typeof window === 'undefined'` not literally `undefined`

mimecuvalo:
a very special rant here that i f'ing hate `mock`/`spyOn` spent hours banging my head on it, and it's always some esoteric wiring up to get it working right UGH. `</rant>`

mimecuvalo (approved):
LGTM!

MitjaBezensek (approved):
Let's go! 🎉

steveruizok:
waaait one sec

PR #4198, opened 2024-07-16
asset lod: fix high-res images

mimecuvalo:
high-res images are busted currently, we weren't doing `Math.min` before https://github.com/tldraw/tldraw/pull/4069 and now it's limited by the natural width of the image which is actually at odds with what's been uploaded. 

## for separate PR
...actually, now that i think about it, this code isn't doing any resizing anymore - the server has the high-res images but we're only recording it as a smaller image:
```
if (isFinite(maxImageDimension)) {
	const resizedSize = containBoxSize(size, { w: maxImageDimension, h: maxImageDimension })
	if (size !== resizedSize && MediaHelpers.isStaticImageType(file.type)) {
		size = resizedSize
	}
}
```

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Images LOD: fix high-res images.

### DIFF
--- a/packages/tldraw/src/lib/utils/assets/assets.ts
+++ b/packages/tldraw/src/lib/utils/assets/assets.ts
@@ -44,7 +44,7 @@ export function containBoxSize(
  * ```ts
  * const image = await (await fetch('/image.jpg')).blob()
  * const size = await getImageSize(image)
- * const resizedImage = await downsizeImage(image, size.w / 2, size.h / 2, { type: "image/jpeg", quality: 0.92 })
+ * const resizedImage = await downsizeImage(image, size.w / 2, size.h / 2, { type: "image/jpeg", quality: 0.85 })
  * ```
  *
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/utils/assets/assets.ts
```diff
@@ -60,7 +60,7 @@ export async function downsizeImage(
 	opts = {} as { type?: string; quality?: number }
 ): Promise<Blob> {
 	const image = await MediaHelpers.usingObjectURL(blob, MediaHelpers.loadImage)
-	const { type = blob.type, quality = 0.92 } = opts
```
mimecuvalo, line 63: as long as i'm in here looking at this: this brings this more in line with the Cloudflare default of 85 (not that this code is actually used anywhere at the moment, to be clear)

PR #4206, opened 2024-07-17
Make asset.fileSize optional

steveruizok:
This PR makes the `fileSize` property of `TLImageAsset` and `TLVideoAsset` optional. I first noticed this when I was updating the Draw Fast repo, but there are a bunch of cases where we don't know the file size when we're creating an asset. Instead of setting it to -1 (which is sort of magic), we can leave it off if we didn't know it already.

### Change type

- [x] `api`

### Test plan

- [x] Unit tests

### Release notes

- Made the `fileSize` property of `TLImageAsset` and `TLVideoAsset` optional

### DIFF
--- a/packages/tlschema/src/assets/TLImageAsset.ts
+++ b/packages/tlschema/src/assets/TLImageAsset.ts
@@ -12,11 +12,11 @@ export type TLImageAsset = TLBaseAsset<
 	{
 		w: number
 		h: number
-		fileSize: number
 		name: string
 		isAnimated: boolean
 		mimeType: string | null
 		src: string | null
+		fileSize?: number
 	}
 >
 
@@ -26,11 +26,11 @@ export const imageAssetValidator: T.Validator<TLImageAsset> = createAssetValidat
 	T.object({
 		w: T.number,
 		h: T.number,
-		fileSize: T.number,
 		name: T.string,
 		isAnimated: T.boolean,
 		mimeType: T.string.nullable(),
 		src: T.srcUrl.nullable(),
+		fileSize: T.number.optional(),
 	})
 )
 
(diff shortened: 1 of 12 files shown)


packages/tlschema/src/assets/TLImageAsset.ts
```diff
@@ -26,11 +26,11 @@ export const imageAssetValidator: T.Validator<TLImageAsset> = createAssetValidat
 	T.object({
 		w: T.number,
 		h: T.number,
-		fileSize: T.number,
 		name: T.string,
 		isAnimated: T.boolean,
 		mimeType: T.string.nullable(),
 		src: T.srcUrl.nullable(),
+		fileSize: T.number.optional(),
```
SomeHats, line 33: whilst we're here, lets make this a `T.nonZeroInteger.optional()` maybe

steveruizok: Will it always be an integer? Or nonZeroNumber ok?

mimecuvalo (approved):
LGTM!

MitjaBezensek:
This will also help with pasting excalidraw content, saw some [sentry issues](https://tldraw.sentry.io/issues/5516778108/?project=4504203639193600&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+filesize&referrer=issue-stream&statsPeriod=14d&stream_index=0) about missing filesize 👍

packages/tlschema/src/assets/TLImageAsset.ts
SomeHats: it's measured in bytes, and you can't have fractions of a byte, so i think so?

steveruizok: on

ze

PR #4208, opened 2024-07-17
Allow custom tools to decide whether they can be lockable or not.

### LINKED ISSUE #2857: Not-lockable tools should extend via TLUiOverrides
https://github.com/tldraw/tldraw/blob/c4ffa05b1229214f6178e0680d4c613fa963112d/packages/tldraw/src/lib/ui/components/Toolbar/ToggleToolLockedButton.tsx#L11-L20

It's possible to add custom tools & modify built-in tools via `TLUiOverrides`, but the toggle lock behavior is filtered to a fixed set of tools. This behavior isn't compatible with the tool I was attempting to add, and hiding the toggle button feels shaky. Can this list be extendable?

I think I can get closer to providing PR than just a whine, but I'm fresh to working with tldraw and wanted to gut check the approach:

1.  adding an optional `TLUiOverrides` property where the default is as today, an unlocked toggle
2. a tool's lockability can be the default, locked upon tool selection, or never lockable (hides UI). On this point, the main question I have is whether the current list of not-lockable tools should remain strictly enforced (i.e. passing a `TLUiOverride` for a tool on the current list would be an error).

Apologies to linear bot and the noisy transcript. I fat-fingered the issue creation after typing 5 words.

MitjaBezensek:
Right now we have a hardcoded list of tools that are not lockable. This adds a field to the `StateNode`, which allows the users to customize whether their custom tools should be lockable or not.

Resolves #2857 

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Allows custom tools to control whether they can be lockable or not. By default they are lockable. You can opt out by overriding `StateNode`'s `isLockable` field:
```typescript
export class MyCustomTool extends StateNode {

   static override isLockable = false

}
```

### DIFF
--- a/packages/tldraw/src/lib/shapes/draw/DrawShapeTool.ts
+++ b/packages/tldraw/src/lib/shapes/draw/DrawShapeTool.ts
@@ -6,6 +6,7 @@ import { Idle } from './toolStates/Idle'
 export class DrawShapeTool extends StateNode {
 	static override id = 'draw'
 	static override initial = 'idle'
+	static override isLockable = false
 	static override children = (): TLStateNodeConstructor[] => [Idle, Drawing]
 
 	override shapeType = 'draw'
--- a/packages/tldraw/src/lib/ui/components/Toolbar/ToggleToolLockedButton.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/ToggleToolLockedButton.tsx
@@ -10,17 +10,6 @@ interface ToggleToolLockedButtonProps {
 	activeToolId?: string
 }
 
-const NOT_LOCKABLE_TOOLS = [
-	'select',
-	'hand',
-	'draw',
-	'eraser',
-	'text',
-	'zoom',
-	'laser',
-	'highlight',
-]
-
 export function ToggleToolLockedButton({ activeToolId }: ToggleToolLockedButtonProps) {
 	const editor = useEditor()
 	const breakpoint = useBreakpoint()
@@ -29,8 +18,9 @@ export function ToggleToolLockedButton({ activeToolId }: ToggleToolLockedButtonP
 	const isToolLocked = useValue('is tool locked', () => editor.getInstanceState().isToolLocked, [
 		editor,
 	])
+	const tool = useValue('current tool', () => editor.getCurrentTool(), [editor])
 
-	if (!activeToolId || NOT_LOCKABLE_TOOLS.includes(activeToolId)) return null
+	if (!activeToolId || !tool.isLockable) return null
 
 	return (
 		<TldrawUiButton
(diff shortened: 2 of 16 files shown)


packages/tldraw/src/lib/shapes/draw/DrawShapeTool.ts
```diff
@@ -6,6 +6,7 @@ import { Idle } from './toolStates/Idle'
 export class DrawShapeTool extends StateNode {
 	static override id = 'draw'
 	static override initial = 'idle'
+	static override isLockable = false
```
mimecuvalo, line 9: huh, didn't realize Draw and Text are not lockable...
actually, why is that? for my understanding

mimecuvalo (commented):
for a second, i almost thought an alternative would be specifying a field that was like `isManipulator` - something that manipulates shapes vs. creating them.

but then the fact that Draw+Text are also not lockable surprised me! why is that? 🤔

packages/tldraw/src/lib/shapes/draw/DrawShapeTool.ts
MitjaBezensek: Draw tool is a bit special since we don't exit it at all (even though it is not lockable). Makes sense since when you start drawing you probably want to draw multiple shapes. Not sure about text, has been like that since I can remember.

cc @steveruizok

steveruizok: Yep, the draw tool predates tool lock.

packages/tldraw/src/lib/ui/components/Toolbar/ToggleToolLockedButton.tsx
```diff
@@ -29,8 +18,9 @@ export function ToggleToolLockedButton({ activeToolId }: ToggleToolLockedButtonP
 	const isToolLocked = useValue('is tool locked', () => editor.getInstanceState().isToolLocked, [
```
steveruizok, line 18: Combine these two hooks into a single "show tool lock" value?

MitjaBezensek: We use `isToolLocked` to either show the `lock` or `unlock` icon, so we'd need this one anyway.

steveruizok (approved):

packages/tldraw/src/lib/shapes/draw/DrawShapeTool.ts
mimecuvalo: but whyyy though. why is Text not lockable? i don't think there's a good reason to not let it be lockable.

PR #4212, opened 2024-07-18
Finesse sync api

ds300:
- Make it so that for client-side stuff, only the api surface area of the two hooks are public.
- Make it so that for server-side stuff, only the TLSocketRoom api surface area is public.
- Rename `sessionKey` => `sessionId` for consistency
- Add tsdoc comments for public stuff

NEW!

- refactor `userPreferences` option quite heavily to make it simpler
  - rename the multiplayer option `userPreferences` -> `userInfo` and simplify the type since it's only a subset and don't want to pollute the api with confusing stuff.
  - `useTldrawUser()` api for easy-peasy user data integration + new example showing how it's used.
- make assets required, and make `resolve` optional.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


packages/sync/src/useMultiplayerSync.ts
```diff
@@ -36,7 +36,19 @@ export type RemoteTLStoreWithStatus = Exclude<
...
 >
 
-/** @public */
+/**
+ * useMultiplayerSync creates a store that is synced with a multiplayer server.
+ *
+ * The store can be passed directly into the `<Tldraw />` component to enable multiplayer features.
+ * It will handle loading states, and enable multiplayer UX like user cursors and following.
+ *
+ * To enable external blob storage, you should also pass in an `assets` object that implements the `TLAssetStore` interface.
```
SomeHats, line 45:
```suggestion
 * To enable external blob storage, you should also pass in an `assets` object that implements the {@link TLAssetStore} interface.
```

packages/sync/src/useMultiplayerSync.ts
```diff
@@ -155,13 +167,32 @@ export function useMultiplayerSync(
...
 }
 
-/** @public */
+/**
+ * Options for the {@link useMultiplayerSync} hook.
+ * @public
+ */
 export interface UseMultiplayerSyncOptions {
+	/**
+	 * The URI of the multiplayer server. This must include the protocol, e.g. `wss://server.example.com` or `http://localhost:5858`.
```
SomeHats, line 176: I feel like we should maybe make it clear that this should point not just to the multiplayer server, but to the actual room itself
```suggestion
	 * The URI of the multiplayer server. This must include the protocol, e.g. `wss://server.example.com/my-room` or `http://localhost:5858/my-room`.
```

I also wonder if having the `http` and `wss` examples here is confusing - is there a difference between those? maybe we should call out that although you can pass `http`/`https`, we always rewrite it to `ws`/`wss`

packages/sync/src/useMultiplayerSync.ts
```diff
@@ -155,13 +167,32 @@ export function useMultiplayerSync(
...
+	 */
 	uri: string
-	roomId?: string
+	/**
+	 * The user information. If not provided, a default implementation based on localStorage is used.
+	 * This should be synchronized with the configuration for the main `<Tldraw />` component.
+	 */
 	userPreferences?: Signal<TLUserPreferences>
```
SomeHats, line 183: mmmmm we should add an example around this

packages/sync/src/useMultiplayerSync.ts
```diff
@@ -155,13 +167,32 @@ export function useMultiplayerSync(
...
-	/* @internal */
-	trackAnalyticsEvent?(name: string, data: { [key: string]: any }): void
+	/**
+	 * The asset store for blob storage. See {@link tldraw#TLAssetStore}.
+	 * If not provided assets will be stored as base64 strings in the document, which can cause performance issues at serialization boundaries.
+	 */
 	assets?: Partial<TLAssetStore>
```
SomeHats, line 188: is it worth:
1. making this required & not partial
2. renaming `defaultAssetStore` to `inlineBase64AssetStore` or `slowBase64AssetStore` something
3. export it and say you can use it if you want for prototyping maybe

packages/sync/src/useMutltiplayerDemo.ts
```diff
@@ -18,7 +18,15 @@ import { RemoteTLStoreWithStatus, useMultiplayerSync } from './useMultiplayerSyn
 
 /** @public */
 export interface UseMultiplayerDemoOptions {
+	/**
+	 * The room ID to sync with.
```
SomeHats, line 22:
```suggestion
	 * The room ID to sync with. Make sure the room ID is unique. The namespace is shared by
	 * everyone using the demo server. Consider prefixing it with your company or project name.
```

packages/sync/src/useMutltiplayerDemo.ts
```diff
@@ -43,7 +51,18 @@ function getEnv(cb: () => string | undefined): string | undefined {
...
 const IMAGE_WORKER = getEnv(() => process.env.TLDRAW_IMAGE_URL) ?? 'https://images.tldraw.xyz'
 
-/** @public */
+/**
+ * Creates a tldraw store synced with a multiplayer room hosted on tldraw's demo server `https://demo.tldraw.xyz`.
+ *
+ * The store can be passed directly into the `<Tldraw />` component to enable multiplayer features.
+ * It will handle loading states, and enable multiplayer UX like user cursors and following.
+ *
+ * Any data you save will be publicly accessible to anyone who knows the room ID, and any stored data is wiped after a day or so.
```
SomeHats, line 60: maybe add a similar note about the room namespace/prefixing?

packages/sync/src/useMutltiplayerDemo.ts
```diff
@@ -43,7 +51,18 @@ function getEnv(cb: () => string | undefined): string | undefined {
...
+ *
```
SomeHats, line 61:
```suggestion
 *
 * @example
 * ```tsx
 * function MyApp() {
 *     const store = useMultiplayerDemo({roomId: 'my-app-test-room'})
 *     return <Tldraw store={store} />
 * }
 * ```
 *

File: packages/sync/src/useMultiplayerSync.ts [hunk 8]
```diff
@@ -155,13 +167,32 @@ export function useMultiplayerSync(
...
+	/**
+	 * Called when the editor is mounted. Can be used to configure the editor, e.g. registering external asset handlers.
+	 */
```
SomeHats, line 191: tempted to call this one @internal for now

packages/sync/src/useMultiplayerSync.ts
```diff
@@ -36,7 +36,19 @@ export type RemoteTLStoreWithStatus = Exclude<
...
-/** @public */
+/**
+ * useMultiplayerSync creates a store that is synced with a multiplayer server.
+ *
+ * The store can be passed directly into the `<Tldraw />` component to enable multiplayer features.
+ * It will handle loading states, and enable multiplayer UX like user cursors and following.
+ *
+ * To enable external blob storage, you should also pass in an `assets` object that implements the `TLAssetStore` interface.
+ * If you don't do this, adding large images and videos to rooms will cause performance issues at serialization boundaries.
+ *
```
SomeHats, line 47:
```suggestion
 *
 * @example
 * ```tsx
 * function MyApp() {
 *     const store = useMultiplayerSync({
 *         uri: 'wss://myapp.com/sync/my-test-room',
 *         assets: myAssetStore
 *     })
 *     return <Tldraw store={store} />
 * }
 *
 * ```

**Review by SomeHats — APPROVED** [review]

File: packages/sync/src/useMultiplayerSync.ts (hunk 4 above)
Comment (ds300) [reply packages/sync/src/useMultiplayerSync.ts:199]: Did this and in the process I made `resolve` optional, since it seems like that's an advanced-level feature. lemme know if I should revert.

File: packages/sync/src/useMultiplayerSync.ts (hunk 3 above)
Comment (ds300) [reply packages/sync/src/useMultiplayerSync.ts:194]: yes. i also want to make it optional to pass in a non-signal here. will give that a little go and see if it can be done well quickly.

File: packages/editor/src/lib/hooks/useIdentity.tsx [hunk 10]
```diff
@@ -19,3 +19,23 @@ export function useShallowArrayIdentity<T>(arr: readonly T[]): readonly T[] {
...
+		return true
+	}
+	if (!a || !b) {
+		return false
+	}
+	return areObjectsShallowEqual(a, b)
+}
+
+/** @internal */
+export function useNullableShallowObjectIdentity<T extends object | null | undefined>(obj: T): T {
```
SomeHats, line 39: i'd maybe think about replacing `useShallowObjectIdentity` with this

SomeHats (approved):
love the comments, love the new user preferences system - so much clearer! <3

PR #4214, opened 2024-07-18
Force flag should override isLocked in more cases

AlbertBrand:
After locking the camera movement programmatically, there are several editor functions that have an option to 'force' their behaviour even though the camera is locked. But some of these functions do not check correctly if the force flag is set. This PR changes their behaviour so you can force camera movement even when the camera is locked.

### Change type

- [x] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Set the camera options `isLocked` to `true`
2. Call `centerOnPoint`, `resetZoom`, `zoomIn`, `zoomOut`, `zoomToSelection` or `zoomToBounds` with the `force: true` option flag.
3. See that these calls are performed as if `isLocked` was `false`.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed the `force` flag not being taken in account after locking the camera and calling `centerOnPoint`, `resetZoom`, `zoomIn`, `zoomOut`, `zoomToSelection` or `zoomToBounds`.

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -1187,6 +1187,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     readonly sideEffects: StoreSideEffects<TLRecord>;
     slideCamera(opts?: {
         direction: VecLike;
+        force?: boolean | undefined;
         friction?: number | undefined;
         speed: number;
         speedThreshold?: number | undefined;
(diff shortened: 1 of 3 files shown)


AlbertBrand:

steveruizok:
Great catch @AlbertBrand!

PR #4215, opened 2024-07-18
camera: eagle vision 🦅👀

mimecuvalo:
Steve suggested we look into adding this kind of behaviour to the board. It's very handy!
I haven't played much Assassin's Creed but there's this as inspiration: https://assassinscreed.fandom.com/wiki/Eagle_Vision

https://github.com/user-attachments/assets/616973d7-c016-48cd-915e-d06b2edf6dd1

I started creating a separate state machine but then it made more sense to just use the existing `ZoomTool` and add a child state to it. I _did_ make this eagle vision the default for the `z` key and made the current behaviour under `Shift-z` which I think makes more sense.

General behaviour:
- toggling holding the `z` key goes in-and-out of the eagle mode
- because of animation and weird timing stuff, it tries to make sure it doesn't get into half zoomed-out states, some trickiness there.
- if you move your mouse over a certain threshold then we go to the intended area
- or, if you click on a particular area we'll also go to the intended area
- hitting Escape just goes back to your original spot.
- some drive-by fixes where `pointer<X>` didn't match the typing

For large boards this works quite well and feels awesome. For smaller boards, it's interesting because if there's only a shape or two, and you're already zoomed out then we zoom _in_ (because we're doing zoom to fit). I think this is the right behaviour but happy to discuss this nuance.

Screenshot 2024-07-18 at 14 28 11

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Test different boards of different sizes, see what feels right.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Camera: provide option to easily toggle zooming out to a bird's eye view of the board and then being able to zoom into a particular section, if desired.


steveruizok:
Awesome!

Some quick ideas:
- I think I'd like to see the current viewport immediately, rather than on mouse move. Would we have to animate that in though?
- z is parked for the zoom tool, and IMO we should keep that as it is
- we could try the toggle instead, i.e. shift+z, then a click to zoom in on a different viewport or escape/shift+z to return to the same viewport as before

mimecuvalo:
> I think I'd like to see the current viewport immediately, rather than on mouse move. Would we have to animate that in though?

I would disagree on this maybe because I don't want the viewport to move accidentally if I just wanted to zoom out quickly and accidentally tapped the trackpad. Plus, if you wanna just zoom out and see the landscape, that zoom brush can be a little distracting if you just wanted to see everything, not everything plus the extra "shape" that's maybe in the way. That's why there's a small threshold (much like a dragging intention). But let's talk more offline to refine! I could see the threshold maybe being a touch smaller.

> z is parked for the zoom tool, and IMO we should keep that as it is

Right, this PR's little gamble is that it's still Zoom but we swap it for Shift-Z for the _other_ zoom behavior. I think it's worth swapping the two because this is much more useful than the other zoom tool which honestly feels a touch awkward. Maybe let's look at stats and see if people are even using it.

> we could try the toggle instead, i.e. shift+z, then a click to zoom in on a different viewport or escape/shift+z to return to the same viewport as before

Mm, can you rephrase or let's talk offline? I don't think I follow this last bit.

steveruizok:
> I would disagree on this maybe because I don't want the viewport to move accidentally if I just wanted to zoom out quickly and accidentally tapped the trackpad. Plus, if you wanna just zoom out and see the landscape, that zoom brush can be a little distracting if you just wanted to see everything, not everything plus the extra "shape" that's maybe in the way. That's why there's a small threshold (much like a dragging intention). But let's talk more offline to refine! I could see the threshold maybe being a touch smaller.

This makes sense, the "where did I come from" is probably less important than "where will I go". I'll see if I can get a more detailed view of what I'm looking for.

> Right, this PR's little gamble is that it's still Zoom but we swap it for Shift-Z for the _other_ zoom behavior. I think it's worth swapping the two because this is much more useful than the other zoom tool which honestly feels a touch awkward. Maybe let's look at stats and see if people are even using it.

Zoom tool is pretty standard across tools, so we're stuck with that by convention.

mimecuvalo:
> Zoom tool is pretty standard across tools, so we're stuck with that by convention.

Bummer. Well, flipped it and reversed it.

mimecuvalo:
@steveruizok was there anything pending here?

steveruizok:
Mind if we experiment / draft this? I'm still not certain on the feature or interaction but haven't had a chance to dig into it.

mimecuvalo:
> Mind if we experiment / draft this? I'm still not certain on the feature or interaction but haven't had a chance to dig into it.

yeah! most def.

ds300:
Closing due to inactivity

mimecuvalo:
reopening this @steveruizok as part of doing zero backlog ;)

steveruizok:
Closing in favor of clean copy #7801

PR #4219, opened 2024-07-18
interpolated props: Experiment with discrete props

Taha-Hassan-Git:
Some ideas on how to handle props that can't be interpolated, like dash, size etc. I think it looks better when they change halfway through the animation. Animations will likely happen along with movement which helps smooth over the jolt of switching from one thing to another.

This is currently what it looks like:

2024-07-18 at 13 11 32 - Amaranth Primate

This is what it looks like when the props change halfway:

2024-07-18 at 13 12 40 - Teal Gerbil

The text usually changes at the halfway mark, but if there's no text to begin with, then any text in the end shape is streamed in:

2024-07-18 at 15 18 34 - Tan Catshark

What do you think?

A further option: when going between sizes we could animate to any intermediary sizes in between. e.g. going from 's' to 'xl' would be
up to 25%: s
25 - 50%: m
50-75%: lg
75-100%: xl

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
3.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -11,6 +11,7 @@ import {
 	SvgExportContext,
 	TLArrowBinding,
 	TLArrowShape,
+	TLArrowShapeProps,
 	TLHandle,
 	TLOnEditEndHandler,
 	TLOnHandleDragHandler,
@@ -26,6 +27,7 @@ import {
 	arrowShapeMigrations,
 	arrowShapeProps,
 	getDefaultColorTheme,
+	lerp,
 	mapObjectMapValues,
 	structuredClone,
 	toDomPrecision,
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -790,6 +792,56 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
...
+			'text',
+		]
+
+		const isOverHalfway = progress > 0.5
+		const sourceShape = isOverHalfway ? endShape : startShape
+
+		const discreteProps: Partial<TLArrowShapeProps> = {}
+
+		discretePropKeys.forEach((propKey) => {
+			;(discreteProps as any)[propKey] = (sourceShape.props as any)[propKey]
```
Taha-Hassan-Git, line 819: Not sure about this, but neither me nor @MitjaBezensek could come up with anything better.

SomeHats: the issue with this is if we add/remove props to the arrow shape, this won't give us errors. what about:

```tsx
function interpolateDiscreate<R extends { props: object }, P extends keyof T['props']>(start: R, end: R, prop: P, progress: number) {
    if (progress < 0.5) return start.props[prop]
    return end.props[prop]
}
```

and then listing out the things you want to interpolate this way:
```tsx
return {
    props: {
        color: interpolateDiscrete(start, end, 'color', p),
        // etc
    }
}
```

MitjaBezensek: I proposed 
```typescript
function copyKeys(sourceShape: TLArrowShape) {
	const discreteProps: Partial<TLArrowShapeProps> = {}
	const keysToCopy = ['labelColor' as const, 'color' as const]

	keysToCopy.forEach((key) => {
		discreteProps[key] = sourceShape.props[key]
	})
	return discreteProps
}
```

PR #4231, opened 2024-07-19
stickies: make ADJACENT_NOTE_MARGIN match adjacentShapeMargin (again)

mimecuvalo:
Regression from https://github.com/tldraw/tldraw/pull/4056
@steveruizok lemme know if there's something we wanna do here when dynamic size mode is on!
(thx @MitjaBezensek for catching)

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Stickies: fix adjacent margin to match shape margin


steveruizok:
Which one is the right one? As long as they're all the same then I'm good.

mimecuvalo:
> Which one is the right one? As long as they're all the same then I'm good.

I think the `adjacentShapeMargin` is the correct one and we should standardize on that.

PR #4236, opened 2024-07-22
Simplify unfurl in simple-server-example

ds300:
Use `unfurl.js` instead of `cheerio`.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/apps/simple-server-example/package.json
+++ b/apps/simple-server-example/package.json
@@ -48,6 +48,7 @@
 		"react-dom": "^18.2.0",
 		"react-router-dom": "^6.24.1",
 		"tldraw": "workspace:*",
+		"unfurl.js": "^6.4.0",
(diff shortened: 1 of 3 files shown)


apps/simple-server-example/package.json
```diff
@@ -48,6 +48,7 @@
 		"react-dom": "^18.2.0",
 		"react-router-dom": "^6.24.1",
 		"tldraw": "workspace:*",
+		"unfurl.js": "^6.4.0",
```
mimecuvalo, line 51: sounds like you can rm cheerio then :)

PR #4238, opened 2024-07-22
issues: setup bot to auto-close issues over 6 months

mimecuvalo:
### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Issues: bot to auto-close issues over 6 months


MitjaBezensek (approved):
Nice 👏 

We discussed maybe limiting this to just issues where we responded last, but doesn't seem like there's a setting for that?

PR #4240, opened 2024-07-22
db: add fallback for FF versions < 126

mimecuvalo:
FF < 126 doesn't support `window.indexedDB.databases()` so this adds our original logic as a fallback for those older browsers.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix issue with indexedDB in older versions of Firefox < 126.

### DIFF
--- a/packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
+++ b/packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
@@ -42,9 +42,12 @@ async function openLocalDb(persistenceKey: string) {
 }
 
 async function migrateLegacyAssetDbIfNeeded(persistenceKey: string) {
-	const databases = await window.indexedDB.databases()
+	let databases = (await window.indexedDB.databases?.())?.map((db) => db.name)
+	if (!databases) {
+		databases = getAllIndexDbNames()
+	}
 	const oldStoreId = LEGACY_ASSET_STORE_PREFIX + persistenceKey
-	const existing = databases.find((db) => db.name === oldStoreId)
+	const existing = databases.find((dbName) => dbName === oldStoreId)
 	if (!existing) return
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
```diff
@@ -42,9 +42,12 @@ async function openLocalDb(persistenceKey: string) {
 }
 
 async function migrateLegacyAssetDbIfNeeded(persistenceKey: string) {
-	const databases = await window.indexedDB.databases()
+	let databases = (await window.indexedDB.databases?.())?.map((db) => db.name)
+	if (!databases) {
```
MitjaBezensek, line 46: Don't think we'll ever end up in here. databases is `(string | undefined)[]`, which is always truthy.
```suggestion
	if (!databases.length) {
```

MitjaBezensek (changes requested):

packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
MitjaBezensek: Or maybe even add a check for databases
```typescript
if(!window.indexedDB.databases)
```

mimecuvalo: I think vscode is showing the wrong type info here actually:
`databases` is actually: `string[] | undefined`

but i'm happy to make this clearer!

MitjaBezensek: Yeah. My guess is that it can't understand that `databases` might not exist since it's checking the latest version of the api or something like that?

In any case I think the change you made is much easier to read and also removes any ambiguities.

MitjaBezensek (approved):

PR #4244, opened 2024-07-22
Add make real as an option.

MitjaBezensek:
Add make real as an option since we are also getting some bugs reports for it.

We could also enable issues on the make real repo instead? @TodePond @mimecuvalo 

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Add make real as one of the options.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -39,8 +39,8 @@ body:
       options:
         - The tldraw.com website
         - The developer package
-      validations:
-        required: true
+    validations:
+      required: true
   - type: input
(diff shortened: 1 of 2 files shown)


steveruizok (approved):

mimecuvalo:
> We could also enable issues on the make real repo instead? @TodePond @mimecuvalo

yes, i think we should to reduce noise on this repo - that's either @TodePond or @steveruizok to enable (i don't have perms on that particular repo)

steveruizok:
Good idea, I would have expected issues to be open there already. We can scoop these and transfer them there.

PR #4250, opened 2024-07-22
Deprecate editor.mark, fix cropping tests

ds300:
So it turns out `editor.mark(id)` is a bit problematic unless you always pass in unique id, because it's quite easy to create situations where you will call `bailToMark(id)` but the mark that you were _intending_ to bail to has already been popped off the stack due to another previous call to `bailToMark`.

I always suspected this might be the case (the original late 2022 history api was designed to avoid this, but it got changed at some point) and indeed I ran into this bug while investigating a cropping undo/redo test error.

To prevent issues for ourselves and our users, let's force people to use a randomly generated mark ID.

Also `editor.mark` is a bad name. `mark` could mean a million things, even in the context of `editor.history.mark` it's a pretty bad name. Let's help people out and make it more descriptive.

This PR deprecates the `editor.mark(id)` in favor of `id = editor.markHistoryStoppingPoint(name)`.

I converted a couple of usages of editor.mark over but there's a lot left to do so I only want to do it if you don't object @steveruizok 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

This deprecates `Editor.mark()` in favour of `Editor.markHistoryStoppingPoint()`.

This was done because calling `editor.mark(id)` is a potential footgun unless you always provide a random ID. So `editor.markHistoryStoppingPoint()` always returns a random id.

### DIFF
--- a/apps/docs/content/docs/editor.mdx
+++ b/apps/docs/content/docs/editor.mdx
@@ -94,21 +94,17 @@ For information about how to synchronize the store with other processes, i.e. ho
 
 ### Undo and redo
 
-The history stack in tldraw contains two types of data: "marks" and "commands". Commands have their own `undo` and `redo` methods that describe how the state should change when the command is undone or redone.
+The history stack in tldraw contains two types of data: 
 
-You can call [Editor#mark](?) to add a mark to the history stack with the given `id`.
+- "diffs" the changes you make to the store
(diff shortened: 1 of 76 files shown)


steveruizok:
I'm good with the long name!

PR #4254, opened 2024-07-22
Load watermark, and fall back to local if it fails or takes too long

SomeHats:
Instead of relying on isOffline, we actually attempt to load the watermark properly and if it fails, then we fall back to the local version. 

I also noticed that the watermark URL was incorrect whilst testing this.

### Change type

- [x] `improvement`


mimecuvalo (approved):
LGTM! thanks!

Issue #2894, opened 2024-02-20
Allow scribbles to record optional z value in addPoint()


rockinghelvetica:
https://github.com/tldraw/tldraw/blob/59285690d20dfbf28c530104a61da56d874be1c1/packages/editor/src/lib/editor/managers/ScribbleManager.ts#L85-L89

The scribble point is hard-coded to use `0.5` as its `z` value. Moving this to a default parameter (`z: number = 0.5`) would allow the point to optionally accept a pressure value for compatibility with drawing.

I'm currently experimenting with a drawing tool that has a secondary, temporal effect (using scribbles) and the parity would be desirable.

rockinghelvetica:
I could try to put this forward as a PR if someone hinted whether any tests, migrations, or a secret third thing might be part of such a change. I can speculate about docs & API reports, but the total surface area is fuzzy to me.

SomeHats:
Would love to have this as a PR! It's a small enough change that I don't think we'd be super fussed about tests, but they are a good way of stopping us from removing this because we don't realise people are depending on it. You could add a quick something in `ScribbleManager.test.tsx` perhaps? It'd look something like this:
```tsx
test('something', () => {
    const editor = new TestEditor()
    const scribble = new ScribbleManager(editor)

    // make sure that scribbles look good
})
```

No nned for migrations etc, docs & API reports should get automatically generated from your change when you do `git commit`.

PR #4260, opened 2024-07-23
Allow non default z value for scribble points.

MitjaBezensek:
Resolves https://github.com/tldraw/tldraw/issues/2894

Added a simple `TestEditor`, since the one we already have is defined in the `tldraw` package and the `editor` package should not depend on it.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Allow scribble points to have non default z values.


SomeHats (approved):
lgtm!

Issue #1995, opened 2023-10-03
[Chrome] Dotted draw shape rendering issue when zoomed out


MitjaBezensek:
When using the dotted option with the draw tool you might not see the shapes you draw if you are zoomed out.

Drawing not rendering on the left where I'm creating the shape, but it is shown on the right in a different browser.
https://github.com/tldraw/tldraw/assets/2523721/dd2e42ed-9646-4cd3-897d-ca843299fb87

mimecuvalo:
Looks like this is ok now, but lemme know if you're still seeing it @MitjaBezensek !

MitjaBezensek:
Still happening. Chrome, 10% zoom, draw tool, Dash - dotted option. Probably a Chrome rendering issue 🤷‍♂️ 

https://github.com/user-attachments/assets/b34aa1c7-88ee-4ae8-ae3f-2ec62546c212

mimecuvalo:
Huh, weird... it works on my machine (why I closed it). Let's figure out what's different on your machine!

mimecuvalo:
I had dynamic size on, is why it was different!

PR #4261, opened 2024-07-23
draw: fix dotted line rendering when zoomed out

mimecuvalo:
Fixes https://github.com/tldraw/tldraw/issues/1995

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Draw: fix dotted line shape rendering when zoomed out greatly.


packages/tldraw/src/lib/shapes/draw/getPath.ts
```diff
@@ -119,11 +119,17 @@ export function getPointsFromSegments(segments: TLDrawShapeSegment[]) {
...
+	strokeWidth: number,
+	zoomLevel: number
+) {
 	return {
 		draw: 'none',
 		solid: `none`,
-		dotted: `0.1 ${strokeWidth * 2}`,
+		// If we're zoomed way out (10%), then we need to make the dotted line go to 9 instead 0.1
+		// Chrome doesn't render anything otherwise.
+		dotted: `${zoomLevel < 0.2 ? 0 : 0.1} ${strokeWidth * 2}`,
```
MitjaBezensek, line 132: Seems like it also disappears at 21 and sometimes even 22%? At least from my tests 23% works.

https://github.com/user-attachments/assets/6ad40f86-a9fc-404a-aebc-69c41ead9666






```suggestion
		dotted: `${zoomLevel < 0.23 ? 0 : 0.1} ${strokeWidth * 2}`,
```

MitjaBezensek (approved):
Nice! 👏 

Found a small issue with some percentages around 20%, left a comment.

packages/tldraw/src/lib/shapes/draw/getPath.ts
steveruizok: Browser scale is going to effect this too :(

mimecuvalo: > Seems like it also disappears at 21 and sometimes even 22%? At least from my tests 23% works.

argh, that's worrisome then, that isn't consistent with when it's appearing/disappearing for me. fun fun fun.
@MitjaBezensek is that when it's on the non-retina screen maybe?

MitjaBezensek: Ah yes. On the laptop screen it seems at around 10% and on external, non-retina one its at 22% 🤷‍♂️

steveruizok:
If we can just improve it at 100% zoom then let's land this. Zoomers are beyond the pale

steveruizok:
Added fixes across all other shapes when using perfect dash props. Also deleted duplicated code for perfect dash props!

packages/tldraw/src/lib/shapes/draw/getPath.ts
steveruizok: Settling on 20%. we can dial this in

mimecuvalo: nice! but yeah, if it's still happening from 20 < x < 23, how can else we improve this then?

PR #4263, opened 2024-07-23
Sync docs, further refinements

adamwiggins:
More improvements to sync docs before announcement.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [X] `other`


apps/docs/content/docs/sync.mdx
```diff
@@ -130,10 +106,97 @@ editor [component overrides](/reference/tldraw/TLComponents):
...
+While tldraw sync is the easiest way to add sync capabilities to your tldraw canvas, you can
+also use other sync libraries or implement your own.
+
+### Yjs
+
+[tldraw-yjs example](https://github.com/tldraw/tldraw-yjs-example) illustrates a way of using the [yjs](https://yjs.dev) library with the tldraw SDK. Yjs is a CRDT and can resolve conflicts.
+
+### Partykit
+
+We have a [sockets example](https://github.com/tldraw/tldraw-sockets-example) that uses [PartyKit](https://www.partykit.io/) as a backend. Unlike the Yjs example, this example does not use any special data structures to handle conflicts. It should be a good starting point if you needed to write your own conflict-resolution logic.
```
SomeHats, line 120: i'm not sure about including this one, and less sure about calling it 'partykit'.

it's probably the least useful example/sync thingy that we have. it's very naive - it sort of shows the basic shape of how you might start implementing sync yourself, but there's no guarantee that two people using it will even end up with the same document.

Also, partykit isn't a sync system. it's a better way of working with durable objects. if we actually wanted a partykit example, we could have one - but it'd probably still use `@tldraw/sync` under the hood.

personally, I'd rather retire this example now that we have a proper sync solution. if anything, it could go under the "roll your own" heading? but even then it's not a great starting point imo.

PR #4264, opened 2024-07-23
Address some sync papercuts

SomeHats:
- Add a better error message if it looks like you're pointing sync at the wrong end point
- Split up the client in the workers template so we can refer users to specific parts
- Link to parts of the client from the sync docs
- Use less confusing URLs in examples
- Add instructions for how to use the workers template with an existing tldraw app

### Change type
- [x] `other`


packages/sync-core/src/lib/ClientWebSocketAdapter.ts
```diff
@@ -95,6 +99,12 @@ export class ClientWebSocketAdapter implements TLPersistentClientSocket<TLRecord
 				break
 		}
 
+		if (closeCode === 1006 && !didOpen) {
+			warnOnce(
+				"Could not open WebSocket connection. This might be because you're trying to load a URL that doesn't support websockets. Check the URL you're trying to connect to."
```
ds300, line 104: ❤️

packages/utils/src/lib/warn.ts
```diff
@@ -0,0 +1,18 @@
+const usedWarnings = new Set<string>()
+
+/** @internal */
+export function warnDeprecatedGetter(name: string) {
```
ds300, line 4: this isn't actually used or needed anymore, can just delete

ds300 (approved):

Issue #2901, opened 2024-02-21
Arrow label snapping shouldn't scale with arrow length


TodePond:
When an arrow is very long, the snap area gets larger. It should stay the same size, and be dependent on camera zoom.

![Image](https://uploads.linear.app/8ba8cf67-bb46-4faa-8a97-5eb61e0814c1/bee6b4b7-53a7-497c-82c9-01fc4c72653e/6af02a6f-451b-4890-bce7-1ff456656381?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzhiYThjZjY3LWJiNDYtNGZhYS04YTk3LTVlYjYxZTA4MTRjMS9iZWU2YjRiNy01M2E3LTQ5N2MtODJjOS0wMWZjNGM3MjY1M2UvNmFmMDJhNmYtNDUxYi00ODkwLWJjZTctMWZmNDU2NjU2MzgxIiwiaWF0IjoxNzA5NDA3OTg3LCJleHAiOjMzMjc5OTY3OTg3fQ.\_aiUex924-TmC_J_8h6WsSjoGmOyIAmFAx1pPdECgEg)

Taha-Hassan-Git:
This still happens, do we want it? @mimecuvalo

mimecuvalo:
Yeah, it's prbly worth tweaking, probably not too big of a change.

PR #4265, opened 2024-07-23
Improve arrow label snapping

Taha-Hassan-Git:
closes TLD-2242 , #2901

I tried scaling snapping behaviour with zoom, but then changing zoom levels changed the snap distance and caused some arrow labels to shift positions, which looked bad. Now the snap distance changes depending on the length of the arrow. Labels on longer arrows do still snap over longer distances, but the difference is less pronounced.

Labels do now snap to new positions sometimes when being resized.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a very long arrow
3. Label should now snap across a shorter distance

### Release notes

- Improved snapping distances on very long arrows

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
+++ b/packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
@@ -14,6 +14,7 @@ import {
 	getPointOnCircle,
 	intersectCirclePolygon,
 	intersectLineSegmentPolygon,
+	lerp,
 } from '@tldraw/editor'
 import {
 	ARROW_LABEL_FONT_SIZES,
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
```diff
@@ -309,6 +306,47 @@ export function getArrowLabelPosition(editor: Editor, shape: TLArrowShape) {
...
+				minSnapDistance,
+				maxSnapDistance,
+				(arrowLength - minLength) / (maxLength - minLength)
+			)
+		} else {
+			snapDistance = maxSnapDistance
+		}
+	}
+
+	const wasClamped = clampedPosition >= 0.5 - snapDistance && clampedPosition <= 0.5 + snapDistance
```
mimecuvalo, line 340: `wasClamped` → `shouldClamp`

packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
```diff
@@ -309,6 +306,47 @@ export function getArrowLabelPosition(editor: Editor, shape: TLArrowShape) {
...
+
+	if (wasClamped) {
+		clampedPosition = 0.5
+		editor.updateShape({ ...shape, props: { ...shape.props, labelPosition: clampedPosition } })
```
mimecuvalo, line 344: I think the big issue is that we shouldn't update the shape (that's why things will drift when changing the endpoints). `getClampedPosition` shouldn't have a side-effect like this basically. (if you `get` something it shouldn't _change_ the underlying thing)

packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
```diff
@@ -309,6 +306,47 @@ export function getArrowLabelPosition(editor: Editor, shape: TLArrowShape) {
...
+		shape.props.labelPosition,
+		hasStartArrowhead || hasStartBinding ? range.start : 0,
+		hasEndArrowhead || hasEndBinding ? range.end : 1
+	)
+	const minLength = 100
+	const maxLength = 1500
+	const minSnapDistance = 0.03
+	const maxSnapDistance = 0.005
+
+	let snapDistance = minSnapDistance
```
mimecuvalo, line 327: i think we can simplify this all. you should be able to do just this:
`const snapDistance = Math.min(0.02, 500 / arrowLength * 0.02)`

basically, we want it to make out at 0.02 but as the arrow grows longer we want it to go smaller and tighter to that midpoint

mimecuvalo (commented):
thx for tackling! couple comments

Issue #2799, opened 2024-02-10
[Bug]: Should prefer methods over function properties


h2thien:
### What happened?

Hey 👋 first of all I'm not really sure if this is a bug report or feature request or everything works as expected, so please feel free to close if necessary. Also I'll try to keep it brief here. Details to follow in the "reproduce" section below.

> What happened?

Some classes in TLDraw use function properties instead of methods, making it hard to extend the classes at user land.

> Also tell us, what did you expect to happen?

I think if there's no reason to use function properties, we should use methods by default, as it's much more extensible, which I believe is one of the best things about TLDraw.

### How can we reproduce the bug?

An example of a good case, where a class uses methods, I can easily remove the middle handle of a line, effectively disable the "drag from middle to create poly-line" behaviour, without the need for TLDraw to change anything:

```ts
export class LineShapeUtil extends TLLineShapeUtil {
  override getHandles(shape: TLLineShape): TLHandle[] {
    const prev = super.getHandles(shape)
    const next = prev.filter(handle => handle.type !== 'create')
    return next
  }
}
```

This is possible because getHandles is a **method** on TLLineShapeUtil.

Now an example of a not-so-good case, where a class uses function properties, so I cannot extend a method (meaning override but get the super value) but only replace it altogether (override and lost the super function):

```ts
export class ScaleTool extends LineShapeTool {
  static override id = 'scale'

  onEnter: TLEnterEventHandler = () => {
    super.onEnter()
  }
}
```

This is because onEnter is a function property on LineShapeTool.

References:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#using_super_in_method_definitions

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Contact Details

thien@h2corporation.jp, or h2thien on Discord

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SomeHats:
Thanks for raising! This is a good point - we should introduce a lint rule against this and fix existing use-cases. If you're interested in writing a PR for this I'd be happy to accept it.

h2thien:
Thanks! It's not currently blocking us so I'm not working on this very soon, but happy to if no one takes this yet!

mimecuvalo:
Marking this as still an issue to work on.

mimecuvalo:
Related PR being worked on https://github.com/tldraw/tldraw/pull/4288

PR #4266, opened 2024-07-23
Add function property warning.

MitjaBezensek:
Solves https://github.com/tldraw/tldraw/issues/2799

This isn't perfect. It's a bit hard to distinguish these two cases:

```ts
class StateNode {
   id: string
   onExit?: TLExitEventHandler
}
```
That said as soon as you write an `onExit` implementation you will see the warning. So I'm not sure if it's worth going further than this?

image

Number of warnings:
image
image

Happy to go through them if we decide to go with this.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Adds a lint rule to warn when using function properties. We prefer using methods, since they are more easily extensible.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,7 @@ module.exports = {
 		'local/no-export-star': 'error',
 		'local/no-internal-imports': 'error',
 		'local/tagged-components': 'error',
+		'local/prefer-class-methods': 'warn',
 		'no-only-tests/no-only-tests': 'error',
 		'no-restricted-syntax': [
 			'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,7 +57,6 @@ module.exports = {
 		'local/no-export-star': 'error',
 		'local/no-internal-imports': 'error',
 		'local/tagged-components': 'error',
-		'local/prefer-class-methods': 'warn',
 		'no-only-tests/no-only-tests': 'error',
 		'no-restricted-syntax': [
 			'error',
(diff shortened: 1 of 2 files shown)


SomeHats:
This has been on my mind for a little while. I have a couple thoughts here:

1. Most of our property-style method overrides should be converted to normal method ones. The user-ask in the issue is one reason, but they result in way nicer docs and a much nicer in-editor autocomplete experience for devs building on the SDK. The experience there right now is clunky but doesn't need to be.
2. The main legit reason for property-style methods is automatic `this` binding.
3. I find lint warnings are often not worth the noise they create over an always-actionable error. 

So I do want us to do this, but I was hoping that when we do we would:
- Fix the API definitions so they don't need property-style methods any more
- Add something like an `@bound` decorator to satisfy the `this`-binding usecase
- Enable a lint rule like this (plus https://typescript-eslint.io/rules/method-signature-style/) as an error and fix them all

Obviously thought that's massively blowing up the scope of this diff though so... idk. maybe not worth it. personally I'd rather not have warning-only lint rules though. Error-only with surpressions might be better?

MitjaBezensek:
Yeah, my thinking was to see how hard this rule would be to write, so I was thinking:
* Write the rule as a PoC, use it as a warning and see if the are any exceptions that are hard to catch.
* Blast through all the warnings and fix as many as I can and suppress the tricky ones.
* Turn it into an error.

Also makes the PRs smaller an easier to review. Happy to just blast through all of them tomorrow though 😂

PR #4267, opened 2024-07-24
Sync docs rework

ds300:
- More clearly split the path after 'try the demo' into two options: cloudflare template vs do your own thing
- Add a brief mention of things the cloudflare template _doesn't_ include
- Incorporate some of Geoffrey's feedback
- Clearer explanation of the backend anatomy, including bookmark unfurling (since although it's not part of tldraw sync it is very much part of hosting your own backend).
- Remove bit about presence, that's only gonna confuse people trying to use tldraw sync. It would definitely belong on a page specifically about rolling your own sync engine, but I think we can leave that for later.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Update sync.mdx

### DIFF
--- a/apps/docs/content/docs/sync.mdx
+++ b/apps/docs/content/docs/sync.mdx
@@ -33,19 +33,20 @@ tldraw sync is a library for fast, multiuser collaboration. It's purpose-built f
 	/>
 </div>
 
-### Try the demo server
+## Try the demo server
 
-The easiest way to start experimenting with multiplayer is with our demo server. Start by installing
+The easiest way to experiment with multiplayer is by using our demo server. To get started, install
 `@tldraw/sync`:
 
 ```bash
 npm install @tldraw/sync
 ```
 
-Then, in your app, call the `useSyncDemo` hook with a room ID. It'll return a
-[`store`](/docs/persistence#The-store-prop) that you can pass into the tldraw component:
+Then call the `useSyncDemo` React hook, passing in a room ID. It returns a
+[`store`](/docs/persistence#The-store-prop) that you can pass into the `<Tldraw />` component:
 
 ```tsx
+import { Tldraw } from 'tldraw'
 import { useSyncDemo } from '@tldraw/sync'
 
 function MyApp() {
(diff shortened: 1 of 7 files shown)


apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
 }
 ```
 
-Make sure the room ID is unique. Anyone who uses the same ID can access your room. Our sync demo
-server is a great way to prototype multiplayer in tldraw, but it shouldn't be used in production.
-Rooms will be deleted after 24 hours.
+> 💡 Make sure that `roomId` is unique. Anyone who uses the same ID can access your room.
```
SomeHats, line 58: You can use

```tsx
<Callout icon="💡">

Make sure that `roomId` is unique. Anyone who uses the same ID can access your room.

</Callout>
```

for some slightly nicer layout

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
-[`tldraw-sync-cloudflare`](https://github.com/tldraw/tldraw/blob/main/templates/sync-cloudflare/client/App.tsx)
-template.
+- Authentication and authorization.
+- Rate limiting and size limiting for asset uploads.
+- Storing snapshots of documents over time for long-term history.
+- Listing and searching for rooms.
 
-The sync client automatically enables some default collaborative UX which you can customize using
-editor [component overrides](/reference/tldraw/TLComponents):
+Make sure you also read the section below about [deployment concerns](#deployment_concerns).
```
SomeHats, line 87: Reading through this i sort of got to the end of this section expecting to find out how to actually get going with that template. Could we maybe add a link at the very bottom to the template as a sort of mini cta

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
+  [`tldraw-sync-cloudflare`](https://github.com/tldraw/tldraw/blob/main/templates/sync-cloudflare/client/multiplayerAssetStore.tsx)
+  template.
 
-// First you need to create a Signal containing the basic user details: id, name, and color
-const user = atom<{ id: string; color: string; name: string }>('user', {
-	id: myUser.id,
-	color: myUser.color,
-	name: myUser.name,
-})
+#### Unfurling service
```
SomeHats, line 174: I think we should avoid 'unfurl' in docs. I don't think it's very clear. What about "Bookmark preview service" or "URL metadata service" or something?

SomeHats (commented):
I really like the "two clear paths" thing here! Definitely feels like an improvement.

apps/docs/content/docs/sync.mdx
ds300: 'unfurl' seems to be pretty standard industry jargon? I'm fine with `URL metadata service` though. `Bookmark preview service` is a bit too specific or something.

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
 
-## Other sync systems
+The `@tldraw/sync-core` library can be used to integrate tldraw sync into any JavaScript server environment that supports WebSockets.
 
-While tldraw sync is the easiest way to add sync capabilities to your tldraw canvas, you can
-also use other sync libraries or implement your own.
+We have a [simple server example](https://github.com/tldraw/tldraw/tree/main/apps/simple-server-example), supporting both NodeJS and Bun, to use as a reference for how things should be stitched together.
 
-### Yjs
+#### Backend anatomy
```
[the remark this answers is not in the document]
ds300: Yeah, let's do it in a follow-up PR.

adamwiggins (commented):
Overall I think this is a good balance of making clear what you need to do to use tldraw sync (including implementing your own backend) while also offering the footnote that other sync layers are entirely possible to use.

In the future I can imagine we might want to make a more detailed page on custom sync layers, including third-party libraries and roll-your-own.

apps/docs/content/docs/sync.mdx
```diff
@@ -33,19 +33,20 @@ tldraw sync is a library for fast, multiuser collaboration. It's purpose-built f
 	/>
 </div>
 
-### Try the demo server
+## Try the demo server
 
-The easiest way to start experimenting with multiplayer is with our demo server. Start by installing
```
adamwiggins, line 38: I'd change this to:
> To get started, update your tldraw client to use our demo server. Install the package:

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
 }
 ```
 
-Make sure the room ID is unique. Anyone who uses the same ID can access your room. Our sync demo
-server is a great way to prototype multiplayer in tldraw, but it shouldn't be used in production.
-Rooms will be deleted after 24 hours.
+> 💡 Make sure that `roomId` is unique. Anyone who uses the same ID can access your room.
 
-### Self-host a server
+Data on the demo server only lasts for up to 24 hours. This is great for testing, but for production use you'll need to host your own server.
```
adamwiggins, line 60: "great for testing" -> "suitable for prototyping"

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
+	editor.registerExternalAssetHandler('url', async ({url}) => {
+		return await convertUrlToBookmarkAsset(url)
+	})
+}
+```
 
-Make sure there's only ever one `TLSocketRoom` for each room in your app. If there's
-more than one, users won't see each other and will overwrite others' changes. We use [Durable
-Objects](https://developers.cloudflare.com/durable-objects/) to achieve this on tldraw.com.
+And [here's an actual working example](https://github.com/tldraw/tldraw/blob/main/apps/simple-server-example/src/client/App.tsx).
```
adamwiggins, line 142: "actual working example" -> "full working example of the client-side code"

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
-tldraw has support for displaying the 'presence' of other users. Presence information consists of:
+- Store an authoritative in-memory copy of the document state
+- Transparently set up communication between multiple sync clients via WebSockets.
+- Provide hooks for persisting the document state when it changes.
 
-- The user's pointer position
-- The user's set of selected shapes
-- The user's viewport bounds (the part of the canvas they are currently viewing)
-- The user's name, id, and a color to represent them
+> :bulb: You should make sure that there's only ever one `TLSocketRoom` globally for each room in your app. If there's
```
adamwiggins, line 154: ":bulb:" should be "💡"

apps/docs/content/docs/sync.mdx
```diff
@@ -54,154 +55,153 @@ function MyApp() {
...
+Make sure you also read the section below about [deployment concerns](#deployment_concerns).
 
-- Collaborator cursors: [`TLComponents#CollaboratorCursor`](?)
-- Cursor chat: [`TLComponents#CursorChatBubble`](?)
-- Offline indicator: [`TLComponents#TopPanel`](?)
-- Collaborator list: [`TLComponents#SharePanel`](?)
+### Integrate tldraw sync into your own backend
 
-## Other sync systems
+The `@tldraw/sync-core` library can be used to integrate tldraw sync into any JavaScript server environment that supports WebSockets.
```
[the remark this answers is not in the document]
ds300: yeah only JS. We could theoretically make our actual websocket protocol public and allow people to do their own implementations in other languages, but that would be kinda heavy to think about and maintain.

adamwiggins:
Updated the table of contents for this page, since the "tldraw sync" header was out of date with the de-emphasizing of 3rd party and roll your own. It looks like this now:

CleanShot 2024-07-24 at 11 03 38@2x

Issue #4253, opened 2024-07-22
[Bug]: Inline example menus not opening


Taha-Hassan-Git:
### What happened?

When using the inline example, the quick actions menu, more tools button and style panel do not open.

### How can we reproduce the bug?

1. Open the inline behaviour example
2. Click the more tools menu, quick actions menu or style menu button
Expected: menu opens
Actual: Menu does not open

![2024-07-22 at 16 46 47 - Rose Dragon](https://github.com/user-attachments/assets/6c57f229-8ea2-414a-900d-788d5fa060e0)

### Where did this happen?

The tldraw.com website

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Taha-Hassan-Git:
Band-aid fix in the above PR, but the problem is not resolved. By sidestepping blurring and focusing the container we can get the menus to open again. The underlying problem is something to do with editor's container getting refocused before the dropdown menu has a chance to open. @ds300 has better context on this.

PR #4274, opened 2024-07-24
Fix dropdown menus not opening in focus examples

Taha-Hassan-Git:
Rel:  #4253 

Dropdown menus weren't opening in focus examples. By sidestepping blurring and focusing the container we can get them working again. However, this is a band-aid fix until the underlying problem can be solved. Something to do with editor's container getting refocused before the dropdown menu has a chance to open.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open the inline example and editor focus example
2. Click to open the style panel, more tools menu, pages menu or main menu
3. Menus should open as expected.

### Release notes

- Fixed a bug with dropdown menus not opening on editor focus examples


steveruizok:
Just a note here, following on our chat at today's standup: one of the reasons we introduced the `isFocused` property is because of the difficulty we had using browser focus / blur events and document.activeElement as an indicator for whether or not the editor should respond to keyboard shortcuts / scrolling / etc.

Radix's menus are a good example: because they're displayed outside of the editor's container, focus leaves the container whenever a menu opens. The same thing should happen with embeds, where the embedded document isn't considered a descendant of the element that contains the iframe.

steveruizok (changes requested):
We should remove the second example here.

There are other ways to reproduce this bug, for example open a menu, use Escape to close the menu; the editor will remain unfocused and keyboard shortcuts will not work.

One of the reasons we introduced the `isFocused` property is because of the difficulty we had using browser focus / blur events and document.activeElement as an indicator for whether or not the editor should respond to keyboard shortcuts / scrolling / etc.

Radix's menus are a good example: because they're displayed outside of the editor's container, focus leaves the container whenever a menu opens and doesn't necessarily return when the menu closes. The same thing should happen with embeds because the embedded document isn't considered a descendant of the element that contains the iframe.

It's too bad, because it means developers have to manage focus on their own. My worry is that the second example here is misleading about how the editor works. If anything we should be more explicit that this is _not_ how a developer should manage focus, given that it kinda should work that way but doesn't.

@ds300 that example was yours, is there any angle I'm missing?

steveruizok:
Ok, making those changes.

steveruizok (approved):

PR #4275, opened 2024-07-24
Steve's sync docs pass

steveruizok:
This PR is another bump on the docs for collaboration / sync. 

### Change type

- [x] `other`


ds300:
for @SomeHats and @adamwiggins benefit: steve hadn't seen the changes that went in this morning and was working from yesterday's version of the docs so this undoes a lot of good stuff in it's current state and I asked steve to re-apply the intent of this PR on top of the current version of the sync docs.

ds300:
@steveruizok I'm converting this to a draft for now. 

After comparing your version against yesterday's version, I'm not sure the juice is worth the squeeze here. So if you're not happy with the current version in main I'd recommend starting a fresh branch from main, and making minor edits to fix the things you don't like about it, rather than trying to merge main with your changes in this branch.

steveruizok:
Switched back to sync so that we have a proper diff. 👍

I'd like to push back here. IMO the changes that came in this morning complicate the story with depth changes mid-article, ie suggesting options for using tldraw in production, and then going deep on backend anatomy, before returning to the client code. Some of that depth is not necessary here. The deeper topics are not well served in this article either.

I think the original story is clear:

1. tldraw supports collaboration
2. you can use our tldraw sync library to do collaboration
3. here's how to use the tldraw sync demo 
4. here's how to use tldraw sync in production (host the server, connect the client; recommended option, links to examples)
5. here's how to use tldraw with other libraries (links out to other pages and examples)

ds300:
It feels like you're trying to write a blog post and we're trying to write technical documentation. This PR throws away a ton of extremely useful information for people who will be thinking about or trying to adopt `@tldraw/sync`.

Things I like about your version

- the mention of using an incognito window/other browser to test the multiplayerness
- front-loading the callout about making sure client and server deployment is synced
- inclusion of a codesandbox link

Other than that I think it's drastically less helpful than the current version. Is this not supposed to be documentation for `@tldraw/sync`? If it's just supposed to be light introduction can you move all the technical docs somewhere rather than just deleting them?

steveruizok:
Ok, caught up with David and here's where we landed:

As a core problem, I was thinking of this article as an overview of collaboration in tldraw more abstractly, while David was thinking of this as documentation for tldraw sync library. From my perspective, we were going into too much detail in the article; from David's perspective, I was being too superficial and leaving out important details.

Both valid takes! So let's make two articles.

- `collaboration.mdx` is a general introduction to collaboration in tldraw
- `sync.mdx` is documentation on using tldraw sync in production

One question is, where do we put the sync demo? In this PR I have it in the collaboration page. This lets the sync page carry more of the production detail. I think some users will come to the collaboration.mdx page while flirting with tldraw and evaluating the library, while sync.mdx is more of a long term investment of attention.

ds300 (approved):
This is a big improvement over the initial version of this PR. I think the sync page structure could be better, and some of the prose can be tightened up, but it's fine for now.

apps/docs/content/docs/collaboration.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+```tsx
+<Tldraw
+	store={store}
+	onMount={(editor) => {
+		editor.registerExternalAssetHandler('file', createAssetFromFile)
+		editor.registerExternalAssetHandler('url', createAssetFromUrl)
+	}}
+/>
+```
```
SomeHats, line 108: This is no longer really accurate. We should remove it or update it to match some of the stuff that was deleted from sync.mdx.

For files, overriding the external asset handler is now rarely needed - it's basically only if you want to customize how we figure out what sort of file it is and how we extract metadata. It doesn't have anything to do with collaboration.

For collaboration, they should have supplied a `TLAssetStore` with an `upload` method when they created their store. That method is how you change where assets code.

As for URLs, I don't see what this has to do with collaboration at all? It's also not really something you can customize per se - you just sort of have to implement it. I don't think it's worth calling out here, this makes it seem like too important a concept

apps/docs/content/docs/collaboration.mdx
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+In your browser, open an incognito window and visit your project's URL. You can also use a different browser or different device. The key is the room ID: any apps connecting to the same room ID will enter into a shared collaboration session.
+
+### Limitations of the demo
+
+The sync demo is great for prototyping but you should not use it in production. Data on the demo server only lasts for up to 24 hours, and anyone using the same room id will be able to edit the same room's project.
+
+### Using tlsync in production
```
SomeHats, line 76: should be just 'sync' or 'tldraw sync'

apps/docs/content/docs/collaboration.mdx
```diff
@@ -0,0 +1,133 @@
...
+See the [yjs-example](https://github.com/tldraw/tldraw-yjs-example/blob/45dfe7589ae44e4aaf9bb7359d0372f3a88bd597/src/useYjsStore.ts) for an example of how these records may be created, shared, and updated.
+
+### Collaboration UI
+
+The `tldraw` library includes several components specifically designed for collaboration.
+
+- Collaborator cursors: [`TLComponents#CollaboratorCursor`](?)
+- Cursor chat: [`TLComponents#CursorChatBubble`](?)
+- Offline indicator: [`TLComponents#TopPanel`](?)
+- Collaborator list: [`TLComponents#SharePanel`](?)
```
SomeHats, line 131: If you click these links, you'll see they don't go anywhere super useful - they just sort of list out the components you'd override to customize them. In my original version of this, I made sure to explain that the components prop is how these should be overridden before i listed these out. I think they're just sort of confusing otherwise. Can we move this below the line about TLComponents?

apps/docs/content/docs/sync.mdx
```diff
@@ -15,74 +15,36 @@ keywords:
...
-		alt="Multiplayer example using tldraw sync"
-		src="/images/multiplayer.gif"
-		style={{
-			// learn this one weird trick for transparent animated gifs. doctors hate her!!!
-			mask: 'url(/images/multiplayer_mask.png) luminance center 100% / 100% no-repeat',
-			WebkitMask: 'url(/images/multiplayer_mask.png) luminance center 100% / 100% no-repeat',
-		}}
-	/>
-</div>
+While we offer a [hosted demo](/collaboration#tldraw-sync-demo) of the tldraw sync library for prototyping, the demo is suitable only for prototyping. To use tldraw sync in production, you will need to self-host the tldraw sync server.
```
SomeHats, line 20: this feels clunky - double "for prototyping" in 1 sentence, double "tldraw sync" in the next.

apps/docs/content/docs/sync.mdx
```diff
@@ -141,67 +103,15 @@ function registerUrlHandler(editor: Editor) {
...
-	You should make sure that there's only ever one `TLSocketRoom` globally for each room in your app.
-	If there's more than one, users won't see each other and will overwrite others' changes. We use
-	[Durable Objects](https://developers.cloudflare.com/durable-objects/) to achieve this on
-	tldraw.com.
-</Callout>
-
-Read the reference docs for [`TLSocketRoom`](?), and see an example of how to use it in the [simple server example](https://github.com/tldraw/tldraw/blob/main/apps/simple-server-example/src/server/rooms.ts).
-
-#### Asset storage
+## Using tlsync
```
SomeHats, line 106: tlsync

apps/docs/content/docs/sync.mdx
```diff
@@ -141,67 +103,15 @@ function registerUrlHandler(editor: Editor) {
...
 
-As well as synchronizing the rapidly-changing document data, tldraw also needs a way to store and
-retrieve large binary assets like images or videos.
+## Custom shapes & bindings
```
SomeHats, line 108: it feels really weird to introduce custom shapes & bindings before we've even gone over the other server setup parts. We make reference to `TLSocketRoom` in this section, but don't introduce that until the very bottom of the page.

apps/docs/content/docs/sync.mdx
```diff
@@ -313,3 +206,22 @@ async function loadOrMakeRoom(roomId: string) {
 	return new TLSocketRoom()
 }
 ```
+
+## WebSocket server
```
SomeHats, line 210: it's weird to me that this is way at the bottom here. it's like the core part of our multiplayer server?

PR #4284, opened 2024-07-25
Pass assets prop to `useLocalStore`

steveruizok:
This PR fixes a bug where user asset code was not working.

### Change type

- [x] `bugfix`

### Release notes

- Asset stores created passed into tldraw were not being attached to the store instance.


SomeHats:
i made the same fix here with tests: https://github.com/tldraw/tldraw/pull/4283#issuecomment-2250053589

PR #4288, opened 2024-07-25
Move from function properties to methods

MitjaBezensek:
Things left to do
- [x] Update docs (things like the [tools page](https://tldraw-docs-fqnvru1os-tldraw.vercel.app/docs/tools), possibly more)
- [x] Write a list of breaking changes and how to upgrade.
- [x] Do another pass and check if we can update any lines that have `@typescript-eslint/method-signature-style` and `local/prefer-class-methods` disabled
- [x] Thinks about what to do with `TLEventHandlers`. Edit: Feels like keeping them is the best way to go.
- [x] Remove `override` keyword where it's not needed. Not sure if it's worth the effort. Edit: decided not to spend time here.
- [ ] What about possible detached / destructured uses?

Fixes https://github.com/tldraw/tldraw/issues/2799

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Adds eslint rules for enforcing the use of methods instead of function properties and fixes / disables all the resulting errors.

# Breaking changes

This change affects the syntax of how the event handlers for shape tools and utils are defined.

## Shape utils
**Before**
```ts
export class CustomShapeUtil extends ShapeUtil<CustomShape> {
   // Defining flags
   override canEdit = () => true

   // Defining event handlers
   override onResize: TLOnResizeHandler<CustomShape> = (shape, info) => {
      ...
   }
}
```

**After**
```ts
export class CustomShapeUtil extends ShapeUtil<CustomShape> {
   // Defining flags
   override canEdit() {
      return true
   }

   // Defining event handlers
   override onResize(shape: CustomShape, info: TLResizeInfo<CustomShape>) {
      ...
   }
}
```

## Tools

**Before**
```ts
export class CustomShapeTool extends StateNode {
   // Defining child states
   static override children = (): TLStateNodeConstructor[] => [Idle, Pointing]
   
   // Defining event handlers
   override onKeyDown: TLEventHandlers['onKeyDown'] = (info) => {
      ...
   }
}
```

**After**
```ts
export class CustomShapeTool extends StateNode {
   // Defining child states
   static override children(): TLStateNodeConstructor[] {
      return [Idle, Pointing]
   }
   
   // Defining event handlers
   override onKeyDown(info: TLKeyboardEventInfo) {
      ...
   }
}
```


steveruizok:
If we're going to land this, let's land it quick, since it'll touch a bunch of places in the app! @MitjaBezensek would you mind adding some comments on this PR that the most highlight significant changes?

apps/docs/content/docs/tools.mdx
```diff
@@ -85,17 +87,19 @@ Beginning at the root, each node will first handle the event and then pass the e
...
 	}
 }
 
 class MyTool extends StateNode {
 	static override id = 'my-tool'
 	static override initial = 'my-idle-state'
-	static override children = [MyIdleState]
+	static override children() {
+		return [MyIdleState]
+	}
```
steveruizok, line 100: Why is children a method now?

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -323,7 +334,7 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 		return update
 	}
 
-	override onTranslateStart: TLOnTranslateStartHandler<TLArrowShape> = (shape) => {
+	override onTranslateStart(shape: TLArrowShape) {
```
steveruizok, line 337: Losing the types here for the ShapeUtil methods is unfortunate, because it means all of the methods need to manually type the parameters. TypeScript doesn't make this very easy.

MitjaBezensek: Autocomplete is quite good in these cases, it can infer the types in the suggestions:

https://github.com/user-attachments/assets/57caf2c8-80b8-4fcc-8261-db4f46c2bdac

.eslintrc.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
 		'@typescript-eslint/no-non-null-assertion': 'off',
 		'@typescript-eslint/no-explicit-any': 'off',
 		'@typescript-eslint/ban-ts-comment': 'off',
+		'@typescript-eslint/method-signature-style': ['error', 'method'],
```
SomeHats, line 36: idk if you've seen this? i don't know if it's actually any good though ts is weird about `this`: https://typescript-eslint.io/rules/unbound-method/

apps/examples/src/examples/custom-config/CardShape/CardShapeTool.tsx
```diff
@@ -1,10 +1,10 @@
-import { BaseBoxShapeTool, TLClickEvent } from 'tldraw'
+import { BaseBoxShapeTool, TLClickEventInfo } from 'tldraw'
 export class CardShapeTool extends BaseBoxShapeTool {
 	static override id = 'card'
 	static override initial = 'idle'
 	override shapeType = 'card'
 
-	override onDoubleClick: TLClickEvent = (_info) => {
+	override onDoubleClick(_info: TLClickEventInfo) {
```
SomeHats, line 7: ty for adding release notes for these!

SomeHats (approved):
lets gooooo

steveruizok:
Some strange code required for binding:

```ts
const pointingHandleState =
	editor.getStateDescendant<PointingHandle>('select.pointing_handle')
if (!pointingHandleState) throw Error('SelectTool PointingHandle state not found')

// Needs to be bound to the pointing handle state
const ogPointerMove = pointingHandleState.onPointerMove.bind(pointingHandleState)

function onPointerMove(info: TLPointerEventInfo) {
	if (!info.shape) return

	if (info.shape.type === 'bezier-curve') {
		editor.updateInstanceState({ isToolLocked: true })
		editor.setCurrentTool('select.dragging_handle', {
			...info,
			onInteractionEnd: 'pen',
		})
		return
	}
	ogPointerMove(info)
}

pointingHandleState.onPointerMove = onPointerMove
```

Issue #3749, opened 2024-05-13
Add audio file import support


kranthivitta123:
### What's the feature?

Please include audio files support

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mimecuvalo:
We'll look into this in the future. 👍

mimecuvalo:
this is the PR https://github.com/tldraw/tldraw/pull/4290

PR #4290, opened 2024-07-25
audio: introduce new shape

mimecuvalo:
While looking through issues I ran across one that mentioned audio support that made me pause. (and then I hit play again on my Walkman) 🎧 Let's add audio shapes!

Screenshot 2024-08-02 at 14 25 23

For the custom controls:
- [x] play / pause
- [x] time stamp
- [x] seek
- [x] volume - just mute/unmute

Maybe we want to consider possibly adding these controls, but maybe it's fine without them?
- [ ] fine-grained volume, not just on/off
- [ ] download
- [ ] playback speed

To get the cover art/title tags we use `jsmediatags-web` which is fine (I had to patch it because it has a silly filename bug).
I tried thinking about incorporating directly into our repo but it's too much to do that kind of thing. 
I load it _dynamically_ since it's an extra ~12kb gzipped and felt like we don't need it maybe until you actually have an audio file to upload.

This doesn't do anything with customizing the controls yet, which I imagine we'd want to do with Radix icons and all that jazz so it's consistent across browsers. I will say that it does look half-decent just out of the box so ¯\\_(ツ)_/¯ but yeah, happy to add custom controls.

Fixes https://github.com/tldraw/tldraw/issues/3749

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Test lots of different types of audio files

### Release notes

- Shapes: introduce an Audio shape.


steveruizok:
Hey, taking a look at this now.

As a helper to others, here's a zip of an mp3 with cover art:

[winripped_music_and_art_example.zip](https://github.com/user-attachments/files/16389119/winripped_music_and_art_example.zip)

And here's an mp3 to download without cover art:
https://upload.wikimedia.org/wikipedia/commons/b/bb/Test_ogg_mp3_48kbps.wav

steveruizok:
First thought is about anonymous files. This is more of a general feeling of vulnerability about supporting assets on tldraw, where we support images and videos without any login. I'm hesitant to add more functionality in that direction, though I don't think audio is as much of a risk as images, so not necessarily a blocker.

Second thought is about UI. Audio controls vary by browser quite a bit, and since audio files themselves don't have a natural size, that difference is more noticeable.

Generally I think it's worth it to keep going here, especially if we can sync the audio experience. At the moment the sync functionality in videos is broken, let's remove that and then actually figure out how to do it. (This would be good for the timer too, if we get there).

mimecuvalo:
> First thought is about anonymous files. This is more of a general feeling of vulnerability about supporting assets on tldraw, where we support images and videos without any login. I'm hesitant to add more functionality in that direction, though I don't think audio is as much of a risk as images, so not necessarily a blocker.

Makes sense, yeah I'd worry less than the video stuff but I hear the concern.

> Second thought is about UI. Audio controls vary by browser quite a bit, and since audio files themselves don't have a natural size, that difference is more noticeable.

Yeah, we should definitely do our own custom UI with some radix icons to ensure consistency.

> Generally I think it's worth it to keep going here, especially if we can sync the audio experience. At the moment the sync functionality in videos is broken, let's remove that and then actually figure out how to do it. (This would be good for the timer too, if we get there).

👍

mimecuvalo:
@steveruizok custom controls UI is in. I assume you'll wanna tweak!

ds300:
Closing due to inactivity

PR #4293, opened 2024-07-26
text shape: dont make a fixed width unless more intentional drag

mimecuvalo:
Small thing I noticed when on the call with Jørgen. He created a narrow vertical column of text because he accidentally dragged just a touch and didn't know how to undo that state. This makes that drag state a little less accident-prone. (Maybe it means or dragging state is a touch too sensitive in general? I dunno)

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Text shape: dont make a fixed width unless more intentional drag


packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -14,13 +14,24 @@ export class Pointing extends StateNode {
...
 
 	override onPointerMove: TLEventHandlers['onPointerMove'] = (info) => {
-		if (this.editor.inputs.isDragging) {
+		// Don't create a fixed width shape unless the the drag is a little larger,
+		// otherwise you get a vertical column of single characters if you accidentally
+		// drag a bit unintentionally.
+		if (
+			this.editor.inputs.isDragging &&
+			!Vec.DistMin(this.editor.inputs.currentScreenPoint, this.originScreenPoint!, 32)
+		) {
```
steveruizok, line 34:
```suggestion
		// Don't create a fixed width shape unless the the drag is a little larger,
		// otherwise you get a vertical column of single characters if you accidentally
		// drag a bit unintentionally.
		const { editor } = this
		const { isPointing, originPagePoint, currentPagePoint } = editor.inputs
		if (
			isPointing &&
			Vec.Dist2(originPagePoint, currentPagePoint) >
				((editor.getInstanceState().isCoarsePointer
					? editor.options.coarseDragDistanceSquared
					: editor.options.dragDistanceSquared) *
					2) / // double the necessary drag distance for text shapes
					editor.getZoomLevel()
		) {
```

Let's copy the dragging code from the editor and add a multiplier (2x here). We shouldn't need to clone the origin screen point, that will be set just before entering and remain stable until pointer up.

steveruizok: Might be easier for me to push a commit that does this,since there are a few other places I'd change.

mimecuvalo:
thanks for making the suggestion @steveruizok

it still feels really sensitive to accidentally making a vertical text shape? i'm looking at the 100% zoom case. it doesn't take much a quick flick to get into the awkward state.

and yeah, if there's other places to change, lemme know, or feel free to push up.

steveruizok:
Want to try the dial in the modifier?

steveruizok:
Also, we should only be considering the x axis delta!

mimecuvalo:
was this ready to land? i didn't get a chance to do the dialing in / x-axis tweaks...
doesn't feel like this changed much?

PR #4294, opened 2024-07-26
Add option for max pasted / dropped files

steveruizok:
This PR adds an editor option for the maximum number of pasted or inserted files. The default is 100, a simple sanity check.

### Change type

- [x] `improvement`

### Test plan

1. paste 101 files
2. it should error

### Release notes

- We now have an editor option for the maximum number of files that a user can paste at once.


steveruizok:
We aren't really handling errors that come out of these methods, so there's no user feedback apart from the JavaScript console. As a follower, we could consider surfacing these as a "paste failed: here's why" type of toast.

SomeHats (approved):

PR #4297, opened 2024-07-26
why did we have this dpr constrained width/height stuff again?

ds300:
```js
// We round the shape width and height up to the nearest multiple of dprMultiple
// to avoid the browser making miscalculations when applying the transform.
```

this is what i wrote, and yet I can't seem to reproduce any errors at weird browser zoom levels if I remove this dpr stuff now. Maybe the bad transform stuff has been fixed in chrome. I should test a few other browsers.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


steveruizok:
I remember something about images? or HTML frames that accounted for the size of the container?

steveruizok:
Someone on our Discord was raising errors related that the DPR fix solved. I hope it was fixed on the browser level.

PR #4302, opened 2024-07-27
[Experiment] Crows-feet arrows, clever bars

steveruizok:
This PR adds information about the intersection segment to the arrow head info, which lets us do things like crows-feet arrows and smart bar arrowheads. We're not actually slicing the expanded outline of the shape (that's hard) but it mostly looks right!

Kapture 2024-07-27 at 12 34 09
Kapture 2024-07-27 at 12 32 03

Still need:
- [ ] Extra arrowheads
- [ ] Extra arrowhead icons
- [ ] Can we fix the spacing between the shape and the arrowhead?

### Change type

- [x] `feature`

### Test plan

1. Create a shape with a bar arrowhead
2. Bind it to a shape

### Release notes

- Introduces the crows-feet arrowhead, adds bar arrowheads that follow the curve of the shape they're bound to.

### DIFF
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -1548,9 +1548,14 @@ export interface TLArrowPoint {
     // (undocumented)
     arrowhead: TLArrowShapeArrowheadStyle;
     // (undocumented)
-    handle: VecLike;
+    handle: Vec;
     // (undocumented)
-    point: VecLike;
+    intersection?: {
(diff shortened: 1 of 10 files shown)


steveruizok:
We could do all of these if we wanted: 

image

mimecuvalo:
whoa, this is cool

🏹

ds300:
Closing due to inactivity

Issue #3495, opened 2024-04-16
[Feature]: Ability to search text


vickey-dev:
### What's the feature?

I have pretty big notes on the whiteboard, currently, the in-build browser search isn't helping to find the text which I'm looking for as the highlight is not very intuitive to get the area of attention where the text is on the whiteboard.

is there any way to find the text inside the tldraw, would be helpful if we could highlight the text in a different colour inside the editor to highlight the searched keyword & zoom or auto-focus on the searched keyword on the whiteboard.

![Screenshot 2024-04-16 at 07 20 57 PM](https://github.com/tldraw/tldraw/assets/16484526/010e46d1-feca-432b-ac6d-5946be105a3b)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

L-U-C-K-Y:
This would be a great feature, I have many pages and a search across pages would be great!

afomi:
anybody up to pair on this feature?

it should be possible to search through the js object text/attributes for matches,
then highlight matching elements

Issue #2141, opened 2023-11-02
[Feature]: Set text on lines to appear at each endpoint / connection point


erh-git:
### What's the feature?

The way text appears on the center of a line is great (A).  But I'd also love the option to have text appear at the end points (B):

![image](https://github.com/tldraw/tldraw/assets/8422319/2da1f65b-0bee-4cdd-ba44-40bbad963af0)

Perhaps the option to switch between the two options on a per line basis?  

There might be some use case to having text in all three areas on the same line, but that can be attained by simply having two lines (one with center-text, the other with endpoint-text).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

erh-git:
Checking in, are there any updates about this potential feature? Is it something likely to be implemented, or likely not?  Being able to add text at end points would revolutionize network topology diagrams (my primary use case).

Either way it would be nice to know.

mimecuvalo:
Hiya, this is semi-possible now if you connect them with an arrow. You can drag the text label to the left or right of the arrow. However, you can't add more than one label at the moment :-/

erh-git:
Oh, that's cool. I do like that feature.  That is implemented very well.

It doesn't allow for both end points though. I wonder if perhaps a separate issue of "multiple text boxes per line" would make more sense than keeping this old one open.

Issue #1780, opened 2023-07-30
[Feature]: Add table object type


maiconloure:
### What's the feature?

I would like to create a simple table with columns and rows. By default, it might start with one column and one row and the possibility to add more by clicking on the table borders.

### Contact Details

maiconloure@gmail.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

steveruizok:
Noted! It would be a big feature, but we could use this thread to start the research.

Are there any other apps that have implementations that you like?

maiconloure:
I usually use drawio when I need some tables; it can be a nice reference and starting point for implementation on tldraw, I'll leave some images of the tables I use in there. If this feature is able to be implemented in tldraw, I will try to make the implementation.
![112520](https://github.com/tldraw/tldraw/assets/59097467/d5ba3ddd-8218-4ca0-b3b1-7b62233d31ba)
![112450](https://github.com/tldraw/tldraw/assets/59097467/3c601671-a937-45ae-876c-58551d3426e0)

gustavotrott:
I usually use yEd for creating flowcharts because it lets me organize elements neatly using tables. However, I'm starting to switch to tldraw more often. It would be really helpful if tldraw could add a table feature!

mimecuvalo:
To be more realistic and clear of what could happen here _potentially_ in the future: there are tables (spreadsheets) and there are grids. So, I want to make clear here that a Grid is more realistic (simple boxes that are text-editable) where as a Table (which implies header rows, sorting, references, etc.) is huge project.

kiastorm:
If and when we do get a Table shape, I'd love it we could have "automatic detection of markdown tables when pasting into tldraw"

For example, LLM/ChatGPT markdown output -> tldraw shapes

Paste a markdown table -> get a tldraw table shape

[(Originally posted in Discord)](https://discord.com/channels/859816885297741824/859816885801713728/1372855387761541180)

cliftonc:
I came across this as I was trying to use TLDraw for a personal project (creating a ppt like experience in the education sector), and have made this work at a basic level via TipTap Table Extension.  Hope you don't mind the blog post but I have written about it here:  https://cliftonc.nl/blog/adding-tables-to-tldraw-with-tiptap/ and the source is here:  https://github.com/cliftonc/ppt-paste/tree/main/apps/worker/client

Feedback welcome, as it has a few rough edges, but for my use case is good enough.

[...]

PR #4306, opened 2024-07-29
feat(api): add getText to ShapeUtil and text search example

MitjaBezensek:
Motivated by #3495 

An example of how to add custom search via`⌘+f` keyboard shortcuts. This also adds `getText` method to `ShapeUtil`, which allows the users to query the shapes for text.

Down the line we could choose to add something like `getShapesWithText` to the editor and maybe even add some UI to dotcom. For now, let's just expose the API.

https://github.com/user-attachments/assets/5631301c-a02a-4b66-8dbf-09571e67a46c

### Change type

- [x] `feature` (or improvement, bugfix, api, other - pick ONE)

### Test plan

1. Open the text search example.
2. Use ⌘+f to search for text within shapes.

- [ ] Unit tests (if present)

### Release notes

- Adds `getText` to the `ShapeUtil` api so that we can allow searching for text in a nicely extensible way.
- Adds an example of how to add text search.


mimecuvalo:
i love this so much

apps/examples/src/examples/text-search/README.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Text search
+component: ./TextSearchExample.tsx
+category: editor-api
+priority: 3
+keywords:
+  [
+    annotation,
```
mimecuvalo, line 8: i think these keywords need updating maybe? the camera stuff doesn't seem related

apps/examples/src/examples/text-search/TextSearchExample.tsx
```diff
@@ -0,0 +1,39 @@
+import { TLComponents, TLUiOverrides, Tldraw, atom, track } from 'tldraw'
+import 'tldraw/tldraw.css'
+import { TextSearchPanel } from './TextSearchPanel'
+import './text-search.css'
+
+export const showSearch = atom('showSearch', false)
+
+const components: TLComponents = {
+	HelperButtons: TextSearchPanel,
+	Minimap: null,
```
mimecuvalo, line 10: does minimap need to be null?

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -373,6 +373,10 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 		return {}
 	}
 
+	getText(_shape: Shape): string | undefined {
```
mimecuvalo, line 376: since we're thinking of this new hook into the API, i think we should think of the general case where there is multiple text snippets, potentially, within a shape. so, i believe this should actually return:
`string[] | undefined` (and then the shape utils below should also return arrays).

_potentially_, that might mean it also returns more information? like id or position or something but i don't think we should get too complicated / over think this.

but i do think we should at least account for multiple textfields.

MitjaBezensek: I kinda like the lean approach. Until we see a use case I'd rather use the most simple api, which works for the current system. Stuff gets complicated enough already 😄

mimecuvalo: I feel ya on the lean approach. I just worry about the upgrade path here down the line. It feels like, yes, currently we only have one text field per shape. But this then doesn't become a usable API for other folks that have custom shapes (non-tldraw) that have multiple text fields. (there's feature requests that are asking for labels on both sides of an arrow for example). 

A different example would be how we have various checks in our geometry code to see if `isLabel` is true, because we don't want to bake in an assumption that there's only ever one label in the geometry. It keeps the optionality for multiple labels to exist open.

MitjaBezensek: I'm happy to change the api to return an array of strings instead of just a single one. As far as going further than that I don't feel confident in getting the API right without seeing more examples of how it could be used.

I don't think updating this in the future is that big of a problem. It's a fairly contained API and even if we do introduce breaking changes they should be quite limited in scope.

MitjaBezensek: Happy to let you take over if you understand the possible use cases better.

mimecuvalo: Yeah, some examples of how users could use it:
- tables: https://github.com/tldraw/tldraw/issues/1780
- arrows with multiple labels: https://github.com/tldraw/tldraw/issues/2141

And, again, we have to think about how we might have more primitive shapes (square, circle) on our board but our downstream users will have more complex shapes. So, another example could be, say, a slideshow creating tool. And there's a shape that let's you create title slides with Title, Subtitle, and a Date or something. Maybe they could compose that with separate text shapes individually each time but I could image that being annoying and having a combined Slide shape that encompasses multiple text labels like that.

But yeah, the table example above also shows a pretty good use case!

MitjaBezensek: It's still hypothetical. We don't have tables. Would you then expect the API to return an array of arrays for the tables?  Or just a flat array? Would it have some additional info associated with it (like positions)?
That's what I'm talking about when I say I don't understand the use case enough to be confident in coming up with an api.

Like I said I'm happy to change it from a single string to an array of strings if you feel that's better (I don't for now). But if you'd like more than that then you can either take this on or I can just close the PR if you feel it's not useful as is.

mimecuvalo: Hah, yeah, I know we don't have tables, I don't think we ever will in tldraw formally. But _other_ people will add tables to their usage of tldraw, and then they'll want to maybe use the search functionality.

> Would it have some additional info associated with it (like positions)?

Yes, I saw that above in my first comment: "potentially, that might mean it also returns more information? like id or position or something but i don't think we should get too complicated / over think this. but i do think we should at least account for multiple textfields."

Maybe we should talk offline at this point?  :)

mimecuvalo: Talked offline and we'll go ahead land as is!
We'll naturally revisit all of this work anyway because of things like rich text, community feedback, and other feature requests.

vickey-dev:
@MitjaBezensek thanks for introducing the search 🩷 , will it be part of [tldraw.com](https://www.tldraw.com/) because I couldn't find this on the tldraw website

vickey-dev:
@MitjaBezensek  thanks for introducing the search 🩷 , will it be part of [tldraw.com](https://www.tldraw.com/) because I couldn't find this on the tldraw website

MitjaBezensek:
> @MitjaBezensek thanks for introducing the search 🩷 , will it be part of [tldraw.com](https://www.tldraw.com/) because I couldn't find this on the tldraw website

Hey! We are considering adding it to tldraw.com, but don't have a short term plan of doing it. Would it be useful for you? would you expect it to work on a single page or across multiple pages? I guess you have complex documents with many shapes?

vickey-dev:
Yes definitely useful for me @MitjaBezensek. For my case, I'm fine with the single page, I could also see the request across multiple pages as well https://github.com/tldraw/tldraw/issues/3495#issuecomment-2404665622 


>  I guess you have complex documents with many shapes?
Actually no, I mostly use the text & lines alone, I prefer to use it as a digital whiteboard where I jot down all the points related to the work I'm doing & with some docs links & flow charts.

vickey-dev:
@MitjaBezensek  will this be part of any upcoming release to the tldraw.com

PR #4308, opened 2024-07-29
Add terms of use / privacy policy

steveruizok:
This PR adds a terms of use and privacy policy page to tldraw.dev (weirdly, for tldraw.com). It also updates the readme for the dotcom project.

Note that the because the documents are stored on GitHub, links to the documents will be broken until this PR lands.

### Change type

- [x] `other`


mimecuvalo:
gandalf

PR #4310, opened 2024-07-29
Add an eslint rule for unbound function use

MitjaBezensek:
Adds eslint rule to error on unbound function use and fix any existing issues. Moved back to arrow functions for some use cases. Felt better than using bind + disable / `this: void` combo.

Thanks @SomeHats  for pointing this out. Already discovered a few places where I missed this in the previous PR.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add an eslint rule to warn about unbound function use. Fix a few issues caused by #4288


apps/dotcom/src/utils/multiplayerAssetStore.test.ts
```diff
@@ -8,7 +8,7 @@ describe('multiplayerAssetStore.resolve', () => {
 	it('should return null if the asset has no src', async () => {
 		const asset = { type: 'image', props: { w: 100, fileSize: FILE_SIZE } }
 		expect(
-			await resolver(asset as TLAsset, {
+			resolver(asset as TLAsset, {
```
MitjaBezensek, line 11: Noticed this as well, seems like resolver is sync.

apps/dotcom/src/utils/defineLoader.tsx
```diff
@@ -1,8 +1,8 @@
 import { LoaderFunctionArgs, useLoaderData } from 'react-router-dom'
 
 export function defineLoader<T>(_loader: (args: LoaderFunctionArgs) => Promise<T>): {
-	loader(args: LoaderFunctionArgs): Promise<{ [specialSymbol]: T }>
-	useData(): Exclude<T, Response>
+	loader(this: void, args: LoaderFunctionArgs): Promise<{ [specialSymbol]: T }>
```
MitjaBezensek, line 4: Not sure I really like this, but it seems the way to handle these cases. We could also disable them on the usage side of things, but that's also not great.

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -341,7 +341,7 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 	): ReactElement | null | Promise<ReactElement | null>
 
 	/** @internal */
-	expandSelectionOutlinePx(shape: Shape): number {
+	expandSelectionOutlinePx(_shape: Shape): number {
```
MitjaBezensek, line 344: Flyby change.

packages/editor/src/lib/editor/tools/StateNode.ts
```diff
@@ -47,6 +47,7 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
 		public editor: Editor,
 		parent?: StateNode
 	) {
+		// eslint-disable-next-line @typescript-eslint/unbound-method
```
MitjaBezensek, line 50: This this one should be fine as it is done in the constructor so it should be bound to the correct thing.

packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
```diff
@@ -137,8 +137,8 @@ export class DragAndDropManager {
 		this.first = true
 	}
 
-	@bind
-	dispose() {
+	// eslint-disable-next-line local/prefer-class-methods
```
MitjaBezensek, line 140: It kinda made sense to move back to arrow functions for these cases. Otherwise we'd have to use bind + disable / `this: void`.

apps/dotcom/src/utils/multiplayerAssetStore.ts
```diff
@@ -3,7 +3,7 @@ import { ASSET_UPLOADER_URL, IMAGE_WORKER } from './config'
 import { isDevelopmentEnv } from './env'
 
 export const multiplayerAssetStore = {
-	upload: async (asset, file) => {
+	upload: async (_asset, file) => {
```
MitjaBezensek, line 6: Flyby change.

.eslintrc.js
```diff
@@ -34,6 +34,7 @@ module.exports = {
 		'@typescript-eslint/no-explicit-any': 'off',
 		'@typescript-eslint/ban-ts-comment': 'off',
 		'@typescript-eslint/method-signature-style': ['error', 'method'],
+		'@typescript-eslint/unbound-method': ['error', { ignoreStatic: true }],
```
MitjaBezensek, line 37: Ignoring static usage as it can't doesn't contain references to this.

SomeHats (approved):
so i feel sort of conflicted about this overall now seeing the result. I'm not a fan of the way that `this: void` creeps into the public API everywhere, and many of the things this find seem to be more like false positives.

How do you feel about it? Is this a good idea? Code looks good so i'm happy to land it if you like this change, but i think I could go either way on it after actually seeing how it works in practice.

MitjaBezensek:
After some discussion with @SomeHats I decided against this. 

Will close for now, but we can bring this back if we see we need a safety need for these kinds of issues. 

Will create a new PR to fix the issues with minimap on staging.

Issue #4210, opened 2024-07-18
Encountered this error while trying to ungroup a text element and a rectangle element: b0S >= b0S


rushabh-ss:
Hey, I ran into an error while using tldraw. I was trying to ungroup a rectangle element and text element using the keyboard shortcut `CTRL`+`SHIFT`+`G`:

```js
Error: b0S >= b0S
    at Ml (https://www.tldraw.com/assets/index-DKx_6mVa.js:105:5033)
    at qi (https://www.tldraw.com/assets/index-DKx_6mVa.js:105:5769)
    at Hc (https://www.tldraw.com/assets/index-DKx_6mVa.js:105:5876)
    at re.reparentShapes (https://www.tldraw.com/assets/index-DKx_6mVa.js:138:61672)
    at https://www.tldraw.com/assets/index-DKx_6mVa.js:138:82958
    at https://www.tldraw.com/assets/index-DKx_6mVa.js:134:38408
    at tq (https://www.tldraw.com/assets/index-DKx_6mVa.js:105:17740)
    at zi (https://www.tldraw.com/assets/index-DKx_6mVa.js:105:17919)
    at ete.batch (https://www.tldraw.com/assets/index-DKx_6mVa.js:134:38400)
    at re.batch (https://www.tldraw.com/assets/index-DKx_6mVa.js:138:27361)
    at re.ungroupShapes (https://www.tldraw.com/assets/index-DKx_6mVa.js:138:82808)
    at Object.onSelect (https://www.tldraw.com/assets/index-DKx_6mVa.js:329:7864)
    at Object.method (https://www.tldraw.com/assets/index-DKx_6mVa.js:357:50665)
    at Sk (https://www.tldraw.com/assets/index-DKx_6mVa.js:357:46663)
    at wk (https://www.tldraw.com/assets/index-DKx_6mVa.js:357:47979)
    at HTMLBodyElement.p (https://www.tldraw.com/assets/index-DKx_6mVa.js:357:48785)
    at HTMLBodyElement.r (https://www.tldraw.com/assets/index-DKx_6mVa.js:89:1426)
```

My browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36

https://github.com/user-attachments/assets/bb982d41-8bea-426e-964c-4b1cbf7f6224

MitjaBezensek:
[...]

Issue #4126, opened 2024-07-09
[Bug]: Crash when ungrouping elements


peilingjiang:
### What happened?

https://github.com/tldraw/tldraw/assets/25191575/b15383bf-d8c6-4d2e-aee1-309387fbd585

When I tried to ungroup elements that are connected with each other and some others outside of the group, the canvas crashed. Here's the detail:

```
Error: b16l >= b16l
    at Al (https://www.tldraw.com/assets/index-RwMyKwLM.js:105:5033)
    at Zi (https://www.tldraw.com/assets/index-RwMyKwLM.js:105:5769)
    at Kc (https://www.tldraw.com/assets/index-RwMyKwLM.js:105:5876)
    at ne.reparentShapes (https://www.tldraw.com/assets/index-RwMyKwLM.js:138:61668)
    at https://www.tldraw.com/assets/index-RwMyKwLM.js:138:82954
    at https://www.tldraw.com/assets/index-RwMyKwLM.js:134:38408
    at tq (https://www.tldraw.com/assets/index-RwMyKwLM.js:105:17740)
    at Bi (https://www.tldraw.com/assets/index-RwMyKwLM.js:105:17919)
    at ete.batch (https://www.tldraw.com/assets/index-RwMyKwLM.js:134:38400)
    at ne.batch (https://www.tldraw.com/assets/index-RwMyKwLM.js:138:27357)
    at ne.ungroupShapes (https://www.tldraw.com/assets/index-RwMyKwLM.js:138:82804)
    at onSelect (https://www.tldraw.com/assets/index-RwMyKwLM.js:329:7864)
    at onSelect (https://www.tldraw.com/assets/index-RwMyKwLM.js:329:80027)
    at d.addEventListener.once (https://www.tldraw.com/assets/index-RwMyKwLM.js:329:63757)
    at HTMLDivElement.r (https://www.tldraw.com/assets/index-RwMyKwLM.js:89:1426)
    at https://www.tldraw.com/assets/index-RwMyKwLM.js:206:10045
    at Ya (https://www.tldraw.com/assets/index-RwMyKwLM.js:43:37117)
    at Dr.flushSync (https://www.tldraw.com/assets/index-RwMyKwLM.js:43:56292)
    at r_ (https://www.tldraw.com/assets/index-RwMyKwLM.js:206:10029)
    at u (https://www.tldraw.com/assets/index-RwMyKwLM.js:329:63773)
[...]

Issue #3932, opened 2024-06-13
[Bug]: Ungrouping shapes causes crash


elinx:
### What happened?

hit ctrl-d on a group of elements, then the program crash:
```
Error: aUV >= aUV
    at Cl (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:4984)
    at Vi (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:5489)
    at cf (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:5900)
    at https://www.tldraw.com/assets/index-vZ7eszxo.js:138:65738
    at https://www.tldraw.com/assets/index-vZ7eszxo.js:139:9934
    at ja (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:17746)
    at h1.extractingChanges (https://www.tldraw.com/assets/index-vZ7eszxo.js:118:29238)
    at https://www.tldraw.com/assets/index-vZ7eszxo.js:139:9656
    at eee.batch (https://www.tldraw.com/assets/index-vZ7eszxo.js:134:32740)
    at eee.ignore (https://www.tldraw.com/assets/index-vZ7eszxo.js:134:35581)
    at UI (https://www.tldraw.com/assets/index-vZ7eszxo.js:139:9628)
    at https://www.tldraw.com/assets/index-vZ7eszxo.js:138:65257
    at https://www.tldraw.com/assets/index-vZ7eszxo.js:134:32780
    at WG (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:17571)
    at ja (https://www.tldraw.com/assets/index-vZ7eszxo.js:105:17750)
    at eee.batch (https://www.tldraw.com/assets/index-vZ7eszxo.js:134:32772)
    at re.duplicateShapes (https://www.tldraw.com/assets/index-vZ7eszxo.js:138:65015)
    at Object.onSelect (https://www.tldraw.com/assets/index-vZ7eszxo.js:329:7499)
    at Object.method (https://www.tldraw.com/assets/index-vZ7eszxo.js:357:50665)
    at rk (https://www.tldraw.com/assets/index-vZ7eszxo.js:357:46663)
    at ok (https://www.tldraw.com/assets/index-vZ7eszxo.js:357:47979)
    at HTMLBodyElement.p (https://www.tldraw.com/assets/index-vZ7eszxo.js:357:48785)
[...]

PR #4312, opened 2024-07-29
shape ordering: upgrade fractional indexing to use jitter, avoid conflicts

mimecuvalo:
We saw several duplicates related to ungrouping of shapes and fractional indexes having conflicts — it's hard to repro but the theory is that it has to do (mainly, but not always) with multiplayer conflicts (especially if you're offline for a while, you come back online and then you both have the same indices). However, we have seen this in single-player mode too.

### Here are some of the related bugs:
- https://github.com/tldraw/tldraw/issues/3932
- https://github.com/tldraw/tldraw/issues/4126
- https://github.com/tldraw/tldraw/issues/4210

As I looked into the issue, and looked at the original code it led me to:
- this issue on the repo: https://github.com/rocicorp/fractional-indexing/issues/14
- which led to this article: https://madebyevan.com/algos/crdt-fractional-indexing/
- and this repo implementing it: https://github.com/TMeerhof/fractional-indexing-jittered

So! This PR now switches tldraw to using the "jittered" version of fractional indexing to help with these conflicts. It still doesn't satisfy my curiosity on why we see these issues in single-player mode (when ungrouping sometimes). But, this should help alleviate this problem in general, whether single or multiplayer. As noted in the jitter repo:
> The default character set has a chance of roughly one in 47.000 to generate the same key for the same input at the cost of making the keys 3 characters longer on average.

Feels definitely like a good tradeoff! I know @SomeHats mentioned about just allowing for these conflicts but then it sounded like from reading other people running into this, we'd have to repair things when someone tried to insert yet another thing between the conflicting items.

### Reference:
- Original fractional index article by dgreensp: https://observablehq.com/@dgreensp/implementing-fractional-indexing

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Updated the unit tests, it uses the non-jitter version for testability

### Release notes

- Shape ordering: upgrade fractional indexing to use jitter, avoid conflicts


MitjaBezensek:
I managed to reproduce it:
1. Create three rectangles.
2. Now create an arrow that goes from the third rectangle to the second one.
3. Now group the first and second rectangle.
4. Now duplicate this group.
5. Now ungroup
6. 💣 

The order of things is important. I'll investigate further.

https://github.com/user-attachments/assets/9d34608a-72b1-49e2-bfb5-15a5b5e827e3

SomeHats:
Great change! Thanks mime. I would like us to preserve the `IndexKey` type though - it stops arbitrary keys and strings being used interchangably. Amazing work on the repro mitja!

Index generation is so easy to make mistakes about. I sort of want us to revisit our APIs around this - I feel like some more thoughtful API design would make it much harder to write this sort of bug. 

At AT, we designed our indexing APIs to work similar to text manipulation - mostly working around `getKey(s)After(shape)` and `getKey(s)Before(shape)` in a context. That prevented a lot of the footguns that you get with things like `getIndexAbove` (which we should honestly remove, it's almost never what you want) and `getIndexBetween` which you really have to be careful with to get them right

mimecuvalo:
@SomeHats added back the IndexKey type-thingy

packages/utils/src/lib/reordering/dgreensp/dgreensp.test.ts
MitjaBezensek, line 1: We could replace some of these tests with fuzz tests?

packages/utils/src/lib/reordering.ts
MitjaBezensek, line 1: Thanks for moving the file first, makes it easier to see the changes 👏

MitjaBezensek:
Feels good to have this, will definitely avoid many of these kinds of errors. Especially the multiplayer issues, since we don't really have a good alternative solution.

The only thing I don't like about it is that it will mask some of the real client side errors and probably make them even harder to reproduce.

packages/utils/src/lib/reordering/dgreensp/dgreensp.test.ts
mimecuvalo: Eh, that's the responsibility of the package maintainer. In this case, I'm not worried, there's a bunch of tests in the repo to ensure this stuff is correct, example: https://github.com/TMeerhof/fractional-indexing-jittered/blob/main/src/generateKeyBetween.spec.ts

We only had a test for this in our repo since there we were using the code directly instead of through npm

mimecuvalo:
> The only thing I don't like about it is that it will mask some of the real client side errors and probably make them even harder to reproduce. Still a good tradeoff though.

Yeah, agreed!

PR #4315, opened 2024-07-30
Gracefully handle deleted tools & actions

SomeHats:
Adding a new tool or action to our UI is relatively straightforward: you add it to the schema with overrides, then you override the component that makes use of it to append or prepend it. If you want to change the order of things, you have to copy-paste the content component into your app which is a pain.

Deleting tools or actions means modifying the schema with overrides, but then the app will crash - the UI will still be referencing what you deleted. You once again have to copy-paste the entire content of the all the components that reference it and remove the bits you don't need. Make sure you test the whole UI! The error will only appear when you open the relevant menu!

This diff tweaks the types for UI schemas to be able to include missing values, and the UI components to accept missing values. The spread approach that we currently use doesn't lend itself super well to this, so those components now accept a `tool` or an `action` prop instead of the spread tool data. The old API still works just fine too.

This has another bonus: because we know if a menu item is used for a tool, we can automatically pre-populate `isSelected`. People now don't need to call a special hook to find out if their tool is selected.

All in all, this change removes quite a bit of code and lets us deprecate like 30 components (the diff is green over all because of the added deprecation comments)

### Change type
- [x] `api`

### Release notes

- You no longer need to override components to remove tools & actions from our UI


apps/examples/src/examples/remove-tools-actions/RemoveToolsActionsExample.tsx
```diff
@@ -0,0 +1,32 @@
...
+		delete actions.delete
+		delete actions.duplicate
+		delete actions.undo
+		delete actions.redo
+
+		return actions
+	},
+	tools(editor, tools) {
+		// remove tools - see the toolbar
+		delete tools.draw
```
mimecuvalo, line 15: I hesitate that we should make this an example. Here's why:
I wouldn't want to encourage this pattern b/c it might lead the developer down a whack-a-mole path. Later when we add, say, a sticker tool, they'll have to keep coming back and pro-actively deleting new things as our surface area changes. 

Basically, I think we should encourage an example that shows the whitelist path (i.e. I want these 5 [hand, select, draw] tools).

apps/examples/src/hooks/usePerformance.ts
```diff
@@ -12,6 +12,7 @@ export function usePerformance(): TLUiOverrides {
 		actions(_editor, actions) {
 			Object.keys(actions).forEach((key) => {
 				const action = actions[key]
+				if (!action) return
```
mimecuvalo, line 15: nit: can this just be: `const action = actions[key]!` ?

packages/tldraw/src/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable deprecation/deprecation */
```
mimecuvalo, line 1: let's not do this for the entire file (b/c it might hide something useful down the line).

let's prefer doing a lint disable/enable around the lines that are relevant (then it's clearer later what it meant for).

packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenuContent.tsx
```diff
@@ -269,7 +269,6 @@ const DebugFlagToggle = track(function DebugFlagToggle({
 	return (
 		<TldrawUiMenuCheckboxItem
 			id={flag.name}
-			title={flag.name}
```
mimecuvalo, line 272: out of curiousity: why the removal of `title` here (and in the lang menu)?

packages/tldraw/src/lib/ui/components/Toolbar/DefaultToolbarContent.tsx
```diff
@@ -6,50 +6,56 @@ import { TldrawUiMenuItem } from '../primitives/menus/TldrawUiMenuItem'
...
-			<CheckBoxToolbarItem />
-			<ArrowLeftToolbarItem />
-			<ArrowUpToolbarItem />
-			<ArrowDownToolbarItem />
-			<ArrowRightToolbarItem />
-			<LineToolbarItem />
-			<HighlightToolbarItem />
-			<LaserToolbarItem />
-			<FrameToolbarItem />
+			<TldrawUiMenuItem tool="select" />
```
mimecuvalo, line 9: neat!

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1111,7 +1127,6 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					trackEvent('toggle-snap-mode', { source })
 					editor.user.updateUserPreferences({ isSnapMode: !editor.user.getIsSnapMode() })
 				},
-				checkbox: true,
```
mimecuvalo, line 1114: sorry, maybe i missed this in my read-through - this is captured somewhere else now?

packages/tldraw/src/lib/ui/components/NavigationPanel/DefaultNavigationPanel.tsx
```diff
@@ -49,23 +52,27 @@ export const DefaultNavigationPanel = memo(function DefaultNavigationPanel() {
...
 					<>
-						<TldrawUiButton
-							type="icon"
-							data-testid="minimap.zoom-out"
-							title={`${msg(unwrapLabel(actions['zoom-out'].label))} ${kbdStr(actions['zoom-out'].kbd!)}`}
-							onClick={() => actions['zoom-out'].onSelect('navigation-zone')}
-						>
-							<TldrawUiButtonIcon icon="minus" />
-						</TldrawUiButton>
+						{zoomOutAction && (
```
mimecuvalo, line 55: these can use `TldrawUiMenuItem`, right?

packages/tldraw/src/lib/ui/hooks/useActionProps.ts
```diff
@@ -0,0 +1,61 @@
...
+		},
+		[editor, tool, props.isSelected]
+	)
+
+	if ('tool' in props) {
+		if (tool) return { ...tool, isSelected }
+		return undefined
+	}
+
+	if ('action' in props) {
```
mimecuvalo, line 55: small nit here: i think we can put the action check above b/c most of this code is only relevant for the `tool` case

packages/tldraw/src/lib/ui/hooks/useActionProps.ts
```diff
@@ -0,0 +1,61 @@
...
+		if (typeof props.action === 'string') return actions[props.action]
+		return props.action
+	}
+
+	return props
```
mimecuvalo, line 60: this path here at the end, `return props` is for the deprecated path, is that right? if so, can you add a comment that this for the deprecated path (maybe temporarily?)

mimecuvalo (commented):
the DX around this is feeling much better 🙌 
just a couple minor comments really and then we should be able to 🚢 !

packages/tldraw/src/lib/ui/hooks/useActionProps.ts
steveruizok: I think that would call the isSelected hook conditionally?

steveruizok:
This is a good direction, though let's push on the design a little. Could we for example use a <TldrawUiToolMenuItem> and takes a `tool`, bails if there's no tool, otherwise returns a <TldrawUiMenuItem>?

steveruizok:
^^^ See my PR here for an implementation: https://github.com/tldraw/tldraw/pull/4345

apps/examples/src/hooks/usePerformance.ts
SomeHats: the key could be present, but the value could be `undefined`

packages/tldraw/src/index.ts
SomeHats: slightly wanna push back on this one. in general, I would agree that we don't want to disable lint rules for entire files. however, I think that this specific file and this specific rule are worth the exception. This file just lists everything in the library. It doesn't do anything, or make use of any of those exports. It should never contain anything other than the big long list of exports.

The `deprecation/deprecation` rule is to stop us from using deprecated things. But this file doesn't use anything. It just exports it - and it _has_ to export it. This is the one file where referencing deprecated things is actually legit and expected - if they're deprecated but not exported, then we might as well just delete them.

Let me know what you think - if you'd still rather do something more targeted lmk and i'll be happy to.

packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenuContent.tsx
SomeHats: This component doesn't actually do anything with that prop

packages/tldraw/src/lib/ui/context/actions.tsx
SomeHats: Nope - this was set here and then never used

packages/tldraw/src/lib/ui/hooks/useActionProps.ts
SomeHats: it's not deprecated! It's still completely valid to use these components without relying on a tool or an action

SomeHats:
What's the motivation for splitting these out into separate components? I thought about that here, but went with this approach because:
- it's easier to switch to/from using the `tool`/`action` prop to specifying the individual options
- there are less components that users need to know about - we can document the different modes of `<TldrawUiMenuItem />` more easily than we could with several other components

steveruizok:
We discussed in person but for posterity, my reasoning is that these new components really only need to be used by us—in end user apps, they don't need to worry about missing custom tools etc since they made them; and if folks are copying and pasting, then that's okay too. The menu item component should just render stuff, if we're adding more logic on top then I'd prefer other components!

PR #4316, opened 2024-07-30
Fix duplication of shapes.

MitjaBezensek:
This fixes some issues with indexes being incorrectly set when duplicating groups.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create three rectangles.
2. Create an arrow that goes from the third rectangle to the second one.
3. Group the first and third rectangle.
4. Now duplicate this group.
5. Now ungroup

Before this PR this would cause an error.

https://github.com/user-attachments/assets/1d61a74b-6763-4af1-9b1a-0ccbe67ab6fe

### Release notes

- Fix an issue with duplicating groups.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5731,12 +5731,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 				const parentId = originalShape.parentId
 				const siblings = this.getSortedChildIdsForParent(parentId)
 				const currentIndex = siblings.indexOf(originalShape.id)
-				const siblingAboveId = siblings[currentIndex + 1]
-				const siblingAbove = siblingAboveId ? this.getShape(siblingAboveId) : null
 
-				const index = siblingAbove
-					? getIndexBetween(originalShape.index, siblingAbove.index)
-					: getIndexAbove(originalShape.index)
+				const index = getIndexBetween(
+					originalShape.index,
+					this.getShape(siblings[currentIndex + 1])?.index
+				)
 				shape.index = index
 			})
 			const shapesToCreate = shapesToCreateWithOriginals.map(({ shape }) => shape)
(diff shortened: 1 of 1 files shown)


MitjaBezensek:
Ah, this seems to break bound arrow duplication.

https://github.com/user-attachments/assets/a9438122-e79b-43c1-81f7-00e0b3c6f6b6

mimecuvalo:
I defer to @SomeHats for review. This is still dark magic to me 🔮 

My only request is that you put a good comment in there :)

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5733,6 +5724,21 @@ export class Editor extends EventEmitter<TLEventMap> {
 				}
 			)
 
+			orderedShapeIds.forEach((originalId, i) => {
+				const originalShape = this.getShape(originalId)
+				if (!originalShape) return
```
steveruizok, line 5729: This should be an error, right? If there's no source shape then we shouldn't be duplicating it. As it is the code would leave all of these shapes with "a1" as their index.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5733,6 +5724,21 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+				const parentId = originalShape.parentId
+				const siblings = this.getSortedChildIdsForParent(parentId)
+				const currentIndex = siblings.indexOf(originalShape.id)
+				const siblingAboveId = siblings[currentIndex + 1]
+				const siblingAbove = siblingAboveId ? this.getShape(siblingAboveId) : null
+
+				const index = siblingAbove
+					? getIndexBetween(originalShape.index, siblingAbove.index)
```
steveruizok, line 5737: Its ok for the second parameter here to be undefined, so you could do 

`index = getIndexBetween(originalShape.index, this.getShape(siblings[currentIndex + 1]))`

packages/editor/src/lib/editor/Editor.ts
MitjaBezensek: I updated the code so we don't rely on the order of the elements any longer.

packages/editor/src/lib/editor/Editor.ts
MitjaBezensek: Updated this. Still had to do some checks so that we don't call `this.getShape` with `undefined`.

SomeHats (approved):
lgtm!

PR #4318, opened 2024-07-30
Fix an issue with minimap

MitjaBezensek:
Fixes an issue with minimap caused by #4288. Also makes the library use safer by handling some unbound calls in the library instead of relying of end users to bind things.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes
- Fixes a bug with the MiniMap not rendering. Makes the unbound uses safer.


apps/dotcom/src/utils/multiplayerAssetStore.test.ts
```diff
@@ -8,7 +8,7 @@ describe('multiplayerAssetStore.resolve', () => {
 	it('should return null if the asset has no src', async () => {
 		const asset = { type: 'image', props: { w: 100, fileSize: FILE_SIZE } }
 		expect(
-			await resolver(asset as TLAsset, {
```
MitjaBezensek, line 11: Flyby change.

PR #4319, opened 2024-07-30
Custom (non-asset) icons in UI

SomeHats:
The way we do icons (asset URLs that get loaded in) is pretty unusual for a react project. Many customers likely have their own icon systems, or want to use an icon font or a component-based icon library. We don't currently support that well.

This diff updates our tool and action schemas to accept react elements as well as icon key strings. This makes it possible to have anything as icons. in some examples, we now use emojis instead of displaying a broken or generic asset icon. 

I also simplified some of the types around icon and translation keys. I removed a bunch of generics that weren't really doing much, and tweaked the types. Now, they both allow plain strings (this was the case anyway) but show proper autocomplete.

### Change type
- [x] `api`

### Release notes

- You can now specify arbitrary react elements as the icons for tools & actions


packages/tldraw/src/lib/ui/components/primitives/TldrawUiIcon.tsx
```diff
@@ -41,27 +49,48 @@ export const TldrawUiIcon = memo(function TldrawUiIcon({
...
-		)
+		return <div className={classNames('tlui-icon', { 'tlui-icon__small': small }, className)} />
 	}
 
 	return (
 		<div
 			{...props}
 			ref={ref}
-			className={classNames('tlui-icon', { 'tlui-icon__small': small }, className)}
+			className={classNames('tlui-icon tlui-icon__asset', { 'tlui-icon__small': small }, className)}
```
mimecuvalo, line 59: ever so slight nit here (and below):
```suggestion
			className={classNames('tlui-icon', 'tlui-icon__asset', { 'tlui-icon__small': small }, className)}
```
just for readability and for prettier to do it's thing

mimecuvalo (approved):
great idea! i'm psyched about it

steveruizok (approved):

steveruizok:
LGTM! If this was tied back to an issue, would you mention it before you land?

PR #4321, opened 2024-07-31
Export some missing components that are used in the `DefaultToolbar`

MitjaBezensek:
This exports some components that are used in the `DefaultToolbar` components. This will allow the users to reuse them and customize the existing toolbar, while allowing them to reused these components.

[Discussion](https://discord.com/channels/859816885297741824/1266843619072016496/1268015153199190126).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Export `OverflowingToolbar` and `MobileStylePanel` panel components so they can be reused for building a custom `Toolbar`.


mimecuvalo (approved):
👍

PR #4324, opened 2024-07-31
text shape: dont make a fixed width unless more intentional drag, take 2

mimecuvalo:
follow-up to https://github.com/tldraw/tldraw/pull/4293

@steveruizok I think maybe we landed that PR a bit too soon. It still felt pretty easy to make accidental fixed-width text shapes. I know your technique was using pagePoint but I'm going for screenPoint b/c really it's not dependent on zoom level, we're trying to detect a flick of the finger, a fat-finger-fudge-factor, if you will — and that can happen at any zoom level.

This PR actually takes the first refinement and further adds a better starting width for a non-autoSized text shape. Right now, our starting width was 20 which made it a fixed column of 1 letter wide which...isn't really useful. The intention is, if you're gonna have a fixed width to at least have a couple letters.

I also noticed a bug with dynamic size + 800% zoom - the text shape would size in the wrong direction 🙃 

Let's talk tweak it some more!

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Text shape: dont make a fixed width unless more intentional drag

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/TextShapeTool.test.ts
+++ b/packages/tldraw/src/lib/shapes/text/TextShapeTool.test.ts
@@ -103,7 +103,7 @@ describe('When in the pointing state', () => {
 	it('transitions to select.resizing when dragging and edits on pointer up', () => {
 		editor.setCurrentTool('text')
 		editor.pointerDown(0, 0)
-		editor.pointerMove(24, 24)
+		editor.pointerMove(34, 34)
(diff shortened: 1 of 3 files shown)


steveruizok:
Folded into https://github.com/tldraw/tldraw/pull/4331 with some improvements!

PR #4325, opened 2024-07-31
Use press time to inform text shape creation intent

ds300:
@mimecuvalo looks like you and I had the same idea to fix this after the call last week. I updated it to use time-since-pointer-down as a threshold rather than relying on drag distance because if you're trying to create a fixed-width text shape you're gonna press and hold, rather than do a quick click. So I think it makes sense to treat any pointer up that happens very quickly after a pointer down as a normal click, even if the mouse moved quite a bit (which is possible to do by accident, even with the `* 4` threshold)

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
3.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Use time-since-pointerdown to decide whether to create a fixed-width text shape.

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
+++ b/packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
@@ -15,29 +15,38 @@ export class Pointing extends StateNode {
 
 	markId = ''
 
+	enterTime = 0
+	override onEnter(): void {
+		this.enterTime = Date.now()
+	}
+
 	override onExit() {
 		this.editor.setHintingShapes([])
 	}
 
 	override onPointerMove(info: TLPointerEventInfo) {
-		// Don't create a fixed width shape unless the the drag is a little larger,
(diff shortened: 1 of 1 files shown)


mimecuvalo:
great idea! i just pushed up a PR a tiny bit before you reworking the distance logic as well: https://github.com/tldraw/tldraw/pull/4324

(i can merge my stuff into yours or we just merge separately, either way)

cc @steveruizok

mimecuvalo:
I think if we have the time heuristic (prior to https://github.com/tldraw/tldraw/pull/4293) we can even just bring back the original logic of:
`if (isPointing && editor.inputs.isDragging) {`
might be sufficient to just use the timing

steveruizok:
Folded into https://github.com/tldraw/tldraw/pull/4331 !

PR #4326, opened 2024-07-31
Custom embeds API

MitjaBezensek:
Adds the ability to add your own custom embeds and also to disable some of the built in embeds. Also includes [an example](https://github.com/tldraw/tldraw/pull/4326/files#diff-3ff116d77a8ba1d316abbc94386266cf875ee34b4fc312e27f54d158849af572).

Few use cases and how you can achieve them:
* **Remove some (or all) of the built in embeds**: start with the `DEFAULT_EMBED_DEFINITIONS` and filter it down to just the ones you want. Then pass this to the `Tldraw` component.
* **Add a new custom embed on top of the existing ones**: create a new `CustomEmbedDefinition` (or an array of them), now create a new array of `TLEmbedDefinition` by concatenating `DEFAULT_EMBED_DEFINITIONS` and your custom defintions. Pass this to the `Tldraw` component. 
* **Reorder the embeds**: pick the different definitions from `DEFAULT_EMBED_DEFINTIONS` and add them to a new array in the order you wish the embeds to be displayed. Pass this to the `Tldraw`. 
* **Change the icon of an existing embed**: there's two options for this. 1. You can remove the embed from the `DEFAULT_EMBED_DEFINITIONS` then add a new `CustomEmbedDefition` that includes the `icon` field that points to the new icon. 2. You can override the asset url just for that embed.

For the exploded usage you can set the embed definitions after the editor is mounted. Something like 

```ts
onMount={(editor) => {
     const embedShapeUtil = editor.getShapeUtil('embed') as EmbedShapeUtil | undefined
     embedShapeUtil?.setEmbedDefinitions(embeds)
}}
```

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

Adds the ability to customize the embeds that are supported. You can now customize or reorder the existing embeds, as well as add completely new ones.

## Breaking changes
* `EMBED_DEFINITIONS` has been renamed to `DEFAULT_EMBED_DEFINTIONS`
* `matchEmbedUrl`, `matchUrl`, `getEmbedInfo`, `getEmbedInfoUnsafely` now also expect `readonly TLEmbedDefinition[]` as their first parameter. If you don't have any custom embeds you can pass in the `DEFAULT_EMBED_DEFINITIONS`.
* `registerExternalContentHandler` and `putExternalContent` have an additional generic argument for specifying the type of the embeds that the content handle supports. Example on how to update [here](https://github.com/tldraw/tldraw/pull/4326/files#diff-51f3e244dba5247299a2153f96348efdca84e0e8fb8fe27cff4443dd9d8d4161R196).


packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -38,7 +39,8 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
 		return true
 	}
 	override canResize(shape: TLEmbedShape) {
-		return !!getEmbedInfo(shape.props.url)?.definition?.doesResize
+		const definitions = useEmbedDefinitions()
+		return !!getEmbedInfo(definitions, shape.props.url)?.definition?.doesResize
```
steveruizok, line 43: If these are used together, it might be better to make a `useEmbedDefinition` hook?

packages/tlschema/src/shapes/TLEmbedShape.ts
```diff
@@ -15,7 +15,7 @@ const safeParseUrl = (url: string) => {
 }
 
 /** @public */
-export const EMBED_DEFINITIONS = [
+export const DEFAULT_EMBED_DEFINITIONS = [
```
steveruizok, line 18: Do we really really need our embed definitions in the schema? I never really liked this data being in the schema as opposed to being in the tldraw library. The consequences for having a bad name pass validation also seem pretty low, so long as we properly show "broken embeds" in the embed shape.

MitjaBezensek: I guess so. We don't actually do any validations on the type of the embed, since it is always resolved on the fly from the URL.

Schema does reference it once in a migration [here](https://github.com/tldraw/tldraw/blob/main/packages/tlschema/src/shapes/TLEmbedShape.ts#L686). And it seems that this is the main reason why these definitions live in the schema. 

One use case where it still might be useful to have access to the definitions is if one of the embeds changed their url. Then it could be easier to scope down the migration to use a new url. We could hard code it though 🤷‍♂️

packages/tldraw/src/lib/Tldraw.tsx
```diff
@@ -123,6 +126,9 @@ export function Tldraw(props: TldrawProps) {
 	if (preloadingError) {
 		return <ErrorScreen>Could not load assets. Please refresh the page.</ErrorScreen>
 	}
+
+	const embedsWithDefaults = embeds ?? DEFAULT_EMBED_DEFINITIONS
```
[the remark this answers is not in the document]
ds300: also this name `embedsWithDefaults` is not really accurate since we're not merging in the defaults like with the tools and shapes (this is good IMO, if a little inconsistent in terms of API design).


but yeah anyway I'd keep it named `embeds`

```ts
const embeds = useShallowArrayIdentity(rest.embeds ?? DEFAULT_EMBED_DEFINITIONS)
```

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -41,7 +40,7 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
 		return true
 	}
 	override canResize(shape: TLEmbedShape) {
-		return !!getEmbedInfo(shape.props.url)?.definition?.doesResize
+		return !!useEmbedDefinition(shape.props.url)?.definition?.doesResize
```
ds300, line 43:
```suggestion
		return !!useEmbedDefinition(shape.props.url)?.definition?.doesResize
```

Hmmm this is calling a hook from a non-react context, which could throw if we decide to use `canResize` conditionally, or outside of a react render.

I think probably you need to figure out a way to put all the embed definition info into the embed shape util before the editor mounts so you can access it directly rather than through react context

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -56,13 +55,13 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
 	}
 
 	override isAspectRatioLocked(shape: TLEmbedShape) {
-		const embedInfo = getEmbedInfo(shape.props.url)
+		const embedInfo = useEmbedDefinition(shape.props.url)
 		return embedInfo?.definition.isAspectRatioLocked ?? false
```
ds300, line 59: same problem here re: hooks/context

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -56,13 +55,13 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
...
 	}
 
 	override onResize(shape: TLEmbedShape, info: TLResizeInfo<TLEmbedShape>) {
 		const isAspectRatioLocked = this.isAspectRatioLocked(shape)
-		const embedInfo = getEmbedInfo(shape.props.url)
+		const embedInfo = useEmbedDefinition(shape.props.url)
```
ds300, line 64: same problem here re: hooks/context

packages/tlschema/src/shapes/TLEmbedShape.ts
ds300: i think we can just pull out the `fromEmbedUrl` logic for each embed type that was supported before that migration was written and inline it in the migration? From then on I don't think we'd ever need to migrate these URLs, even if the apps change their format, because we can just put that logic in the `toEmbedUrl` methods?

ds300: ah ignore me this might be more bother than it's worth.

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -86,7 +86,7 @@ export function registerDefaultExternalContentHandlers(
 			`File size too big: ${(file.size / 1024).toFixed()}kb > ${(maxAssetSize / 1024).toFixed()}kb`
 		)
 
-		const hash = await getHashForBuffer(await file.arrayBuffer())
```
MitjaBezensek, line 89: flyby change, `getHasForBuffer` is sync.

packages/editor/src/lib/editor/types/external-content.ts
```diff
@@ -24,7 +23,7 @@ export type TLExternalContentSource =
 	  }
 
 /** @public */
-export type TLExternalContent = {
+export type TLExternalContent<EmbedDefinition> = {
```
MitjaBezensek, line 26: Made this generic so `Editor.ts` does not need to know about the concrete `EmbedDefinition` type.

packages/tldraw/src/lib/Tldraw.tsx
```diff
@@ -118,6 +122,10 @@ export function Tldraw(props: TldrawProps) {
 		[_tools]
 	)
 
+	if (embeds) {
+		EmbedShapeUtil.setEmbedDefinitions(embeds)
```
ds300, line 126: This is updating global state, which we should avoid. Otherwise config might leak between tldraw instances.

I think a better way to do this would be to wait until after the Editor has been initialized, then find the instance of the EmbedShapeUtil by `editor.getShapeUtil('embed')` and calling `setEmbedDefinitions` on that, making it an instance method instead of a static method.

Would that work?

mimecuvalo:
so good! 👏

PR #4331, opened 2024-08-01
Combino fixed text

steveruizok:
This PR incorporates a few other PRs related to avoiding accidental fixed-width text shapes.

- the user must have been pointing for more that 150ms
- the user must have dragged 10 times more than usual for dragging (along the x axis)
- the initial shape's width is set using the user's drag distance

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a fixed size text shape while zoomed in / zoomed out
2. Click while moving the mouse with the text shape selected (should not create a fixed-width shape)

### Release notes

- Improved accidental fixed-width text shape creation.

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
+++ b/packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
@@ -35,30 +35,28 @@ export class Pointing extends StateNode {
 		if (Date.now() - this.enterTime < 150) return
 
 		const { editor } = this
-		const { isPointing, originPagePoint, currentPagePoint } = editor.inputs
+		const { isPointing } = editor.inputs
 
-		// only create a fixed width text shape if the user has dragged a certain distance
-		// along the x axis. the y axis doesn't matter, ignore that.
+		if (!isPointing) return
+
+		const { originPagePoint, currentPagePoint } = editor.inputs
+
 		const currentSquaredDragDist = Math.abs(originPagePoint.x - currentPagePoint.x) ** 2
 
-		// This is how far we want the user to drag before making a fixed width text shape
-		const minSquaredDragDistForFixedWidth = editor.getInstanceState().isCoarsePointer
+		const baseMinSquaredDragDistForFixedWidth = editor.getInstanceState().isCoarsePointer
 			? editor.options.coarseDragDistanceSquared
-			: editor.options.dragDistanceSquared * 4
+			: editor.options.dragDistanceSquared
 
-		// We want to multiply the zoom level when we're above 100%, but not below it
-		// The drag distance will be longer when zoomed in but never less than the
-		// minSquaredDragDistForFixedWidth distance.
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -84,7 +104,7 @@ export class Pointing extends StateNode {
 		this.editor.markHistoryStoppingPoint('creating text shape')
 		const id = createShapeId()
 		const { currentPagePoint } = this.editor.inputs
-		const shape = this.createTextShape(id, currentPagePoint, true)
+		const shape = this.createTextShape(id, currentPagePoint, true, 20)
```
steveruizok, line 107: Width doesn't really matter for auto-sized text shapes, but we'll set to 20

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -15,29 +15,50 @@ export class Pointing extends StateNode {
...
+		if (!isPointing) return
+
+		const { originPagePoint, currentPagePoint } = editor.inputs
+
+		const currentDragDist = Math.abs(originPagePoint.x - currentPagePoint.x)
+
+		const baseMinDragDistForFixedWidth = Math.sqrt(
+			editor.getInstanceState().isCoarsePointer
+				? editor.options.coarseDragDistanceSquared
+				: editor.options.dragDistanceSquared
```
steveruizok, line 49: This distance is squared by default (to skip sqrts in regular mouse move checks) so here we need to sqrt it back into regular units.

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -15,29 +15,50 @@ export class Pointing extends StateNode {
...
+		)
+
+		// Ten times the base drag distance for fixed width
+		const minSquaredDragDist = (baseMinDragDistForFixedWidth * 10) / editor.getZoomLevel()
```
ds300, line 53: this makes it take wayyyyy too long to get visual feedback on the drag IMO.

I have no issue with the fact that nobody would want a fixed-width text box smaller than this, but I'd rather have faster feedback while dragging the boxes than a loosely-enforced minimum size like this.

mimecuvalo: it feels better at other zoomed-in, but yeah at 100% it's prbly a touch too long to get the feedback

steveruizok: What feels right?

mimecuvalo: mm, personally, feels like 6 as the multiplier might suit better, perhaps 7

steveruizok:
Dropped it to 6!

PR #4333, opened 2024-08-01
Deep Links

ds300:
Deep Links are URLs which point to a specific part of a document. We provide a comprehensive set of tools to help you create and manage deep links in your application.

## The `deepLinks` prop

The highest-level API for managing deep links is the `deepLinks` prop on the `<Tldraw />` component. This prop is designed for manipulating `window.location` to add a search param which tldraw can use to navigate to a specific part of the document.

e.g. `https://my-app.com/document-name?d=v1234.-234.3.21`

If you set `deepLinks` to `true` e.g. `<Tldraw deepLinks />` the following default behavior will be enabled:

1. When the editor initializes, before the initial render, it will check the current `window.location` for a search param called `d`. If found, it will try to parse the value of this param as a deep link and navigate to that part of the document.
2. 500 milliseconds after every time the editor finishes navigating to a new part of the document, it will update `window.location` to add the latest version of the `d` param.

You can customize this behavior by passing a configuration object as the `deepLinks` prop. e.g.

```tsx
<Tldraw
	deepLinks={{
		// change the param name to `page`
		paramName: 'page',
		// only link to the current page
		getTarget(editor) {
			return { type: 'page', pageId: editor.getCurrentPageId() }
		},
		// log the new search params to the console instead of updating `window.location`
		onChange(url) {
			console.log('the new search params are', url.searchParams)
		},
		// set the debounce interval to 100ms instead of 500ms
		debounceMs: 100,
	}}
/>
```

For full options see the [`TLDeepLinkOptions`](?) API reference.

## Handling deep links manually

We expose the core functionality for managing deep links as a set of methods and utilities. This gives you more control e.g. if you prefer not to use search params in the URL.

### Creating a deep link

You can create an isolated deep link string using the [`createDeepLinkString`](?) helper which takes a [`TLDeepLink`](?) descriptor object.

```tsx
createDeepLinkString({ type: 'page', pageId: 'page:abc123' })
// => 'pabc123'
createDeepLinkString({ type: 'shapes', shapeIds: ['shape:foo', 'shape:bar'] })
// => 'sfoo.bar'
createDeepLinkString({
	type: 'viewport',
	pageId: 'page:abc123',
	bounds: {
		x: 0,
		y: 0,
		w: 1024,
		h: 768,
	},
})
// => 'v0.0.1024.768.abc123'
```

If you do prefer to put this in a URL as a query param, you can use the [`Editor#createDeepLink`](?) method.

```tsx
editor.createDeepLink({ to: { type: 'page', pageId: 'page:abc123' } })
// => 'https://my-app.com/document-name?d=pabc123'
```

### Handling a deep link

You can parse a deep link string with [`parseDeepLinkString`](?) which returns a [`TLDeepLink`](?) descriptor object.

You can then call [`Editor#handleDeepLink`](?) with this descriptor to navigate to the part of the document described by the deep link.

`Editor#handleDeepLink` also can take a plain URL if the deep link is encoded as a query param.

```tsx
editor.handleDeepLink(parseDeepLinkString('pabc123'))
// or pass in a url
editor.handleDeepLink({ url: 'https://my-app.com/document-name?d=pabc123' })
// or call without options to use the current `window.location`
editor.handleDeepLink()
```

### Listening for deep link changes

You can listen for deep link changes with the [`Editor#registerDeepLinkListener`](?) method, which takes the same options as the `deepLinks` prop.

```tsx
useEffect(() => {
	const unlisten = editor.registerDeepLinkListener({
		paramName: 'page',
		getTarget(editor) {
			return { type: 'page', pageId: editor.getCurrentPageId() }
		},
		onChange(url) {
			console.log('the new search params are', url.searchParams)
		},
		debounceMs: 100,
	})
	return () => {
		unlisten()
	}
}, [])
```

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Added support for managing deep links.


SomeHats:
so the main thing i'd want to think about here is some stuff around when this happens in the lifecycle of the editor. right now, it looks like this:

1. create editor
2. set initial viewport to the default
3. render everything, start loading images etc
4. call onMount
5. update state from url
6. re-render everything, start loading new images etc

Is there an API where we could remove the extra render and image fetch before onMount?

ds300:
I addressed @SomeHats feedback by adding a root prop to the `<Tldraw >` component (see PR comment), and doing the url state extraction before the canvas renders (in useLayoutEffect as soon as the editor is initialized). This involved updating the Editor to initialize the viewport bounds during the constructor too, but we can probably pull that out of the constructor and put it in the useLayoutEffect if we're worried about it being problematic for consumers. I think it should be fine as long as the container elem is available when the editor is instantiated.

@steveruizok curious to get your opinion on this API + implementation details

apps/dotcom/src/components/SnapshotsEditor.tsx
```diff
@@ -58,6 +58,9 @@ interface SnapshotEditorProps {
 }
 
 export function SnapshotsEditor({ schema, records }: SnapshotEditorProps) {
+	// make sure this runs before the editor is instantiated
+	useState(convertLegacyUrlParams)
```
SomeHats, line 62: should this maybe be a `useLayoutEffect`? Or just calling it directly? feels very odd to use `useState` to run an effect in this way.

alternatively, could we implement this via the `getUrl` API?

SomeHats (commented):
this is great! ty david. i like the viewport bounds change, much happier getting that without having to wait for render

apps/dotcom/src/components/SnapshotsEditor.tsx
ds300: switched this to a layoutEffect. good call

ds300:
I'm gonna look into collapsing page + viewport into one url param. The whole `paramNames` thing feels like a bit of a landmine for forwards compatibility.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8751,6 +8771,258 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this
 	}
 
+	private _zoomToFitPageContentAt100Percent() {
+		const bounds = this.getCurrentPageBounds()
+		const targetZoom = this.getBaseZoom()
+		if (!bounds) {
+			this.setCamera({ x: 0, y: 0, z: targetZoom }, { immediate: true })
```
ds300, line 8778: i should remove this, it's only valid on the initial navigation action, and I think this is the initial camera anyway, more or less.

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -377,6 +387,10 @@ function TldrawEditorWithReadyStore({
 		cameraOptions,
 	})
 
+	const deepLinks = useShallowObjectIdentity(_deepLinks === true ? {} : _deepLinks)
+	const deepLinksRef = useRef(deepLinks)
+	deepLinksRef.current = deepLinks
```
SomeHats, line 392: any reason we can't chuck this in `editorOptionsRef`? Ideally we wouldn't update refs in render as the render might be aborted with the ref change still taking effect.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8751,6 +8768,255 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * // disable page parameter and change viewport parameter to 'c'
+	 * editor.handleDeepLink({
+	 *   param: 'x',
+	 *   url: 'https://my-app.com/my-document?x=200.12.454.23.xyz123',
+	 * })
+	 * ```
+	 *
+	 * @param opts - Options for loading the state from the URL.
+	 */
+	handleDeepLink(opts?: TLDeepLink | { url?: string | URL; param?: string }): Editor {
```
SomeHats, line 8852: i don't feel super strongly about this, but i find `handle` to be a little ambiguous here - it's not clear what this does from the name. How do you feel about `navigateToDeepLink`?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8751,6 +8768,255 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+					type: 'viewport',
+					pageId: this.options.maxPages === 1 ? undefined : this.getCurrentPageId(),
+					bounds: this.getViewportPageBounds(),
+				}
+			)
+		)
+
+		if (opts?.updateAddressBar) {
+			window.history.replaceState({}, document.title, url.toString())
+		}
```
SomeHats, line 8940: some frameworks (next) don't deal well with navigation that doesn't go through them. should this use `onChange`? alternatively, I think it would be pretty reasonable to remove this option entirely and leave it up to users.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8751,6 +8768,255 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+			opts?.onChange ??
+			(() =>
+				this.createDeepLink({
+					param: opts?.param,
+					to: opts?.getTarget?.(this),
+					updateAddressBar: true,
+				}))
+
+		return react('update url on state change', () => announceChange(new URL(url$.get()), this), {
+			scheduleEffect: debounce((execute) => execute(), opts?.debounceMs ?? 500),
```
SomeHats, line 9016: do we need to worry about cleaning up this debounce when `editor.dispose` is called?

SomeHats (approved):
this is looking so good!

ds300:
great feedback @SomeHats thanks. I'll do a final manual test tomorrow and then get this merged.

PR #4337, opened 2024-08-02
Fix flipping of cropped images

MitjaBezensek:
Fixes an issue with using the Flip horizontal / vertical feature.

We need to change the crop positions based on the flip: if the crop is on the left and we then flip horizontally it now needs to be on the right.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Before

https://github.com/user-attachments/assets/489ad837-7d16-48a3-a250-a64f578c890e

### After

https://github.com/user-attachments/assets/e5c27cff-f872-444c-9573-c1c6e862d88e

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix flipping of cropped shapes. The crop was applied to the wrong part of the picture.


steveruizok:
Looks great!

steveruizok (approved):

PR #4338, opened 2024-08-02
video: rm sync that doesn't really work; fix fullscreen rendering

mimecuvalo:
Our syncing didn't really work, caused problems in multiplayer mode. We're ripping it out for now until we rethink this.

Drive-by fix: I noticed that in fullscreen, portrait videos were being cut off because we render in `cover` mode. This fixes that.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- video: rm sync that doesn't really work; fix fullscreen rendering

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -583,6 +583,11 @@ input,
 	height: 100%;
 }
 
+.tl-video.tl-video-is-fullscreen {
+	object-fit: contain;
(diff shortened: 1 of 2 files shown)


steveruizok:
Looks good!

PR #4344, opened 2024-08-04
Preserve focus search param

steveruizok:
This PR adds a kinda-secret "preserve focus" mode using the `preserveFocus` search param, along with the behavior for handling focus in that state.

In many of our examples, we want `autoFocus` to be true. However, in our docs site, we want each example's `autoFocus` to be false, and for the editor to be focused only when clicked on and blurred when clicking away.

One solution, which this PR implements, is to use a search parameter that says "if the URL has a `preserveFocus` search parameter, then even if `autoFocus` is true, start unfocused`. It also implements that "click to focus, click away to blur" behavior.

We could add support via a prop, however I'm not entirely sure about this feature so would prefer to keep it internal for now. Either way, we would want our docs examples to all use `preserveFocus` routes.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Visit the examples page in our docs.
2. The example should not have focus. Kbds should not work.
3. Click on the example to give it focus. Kbds should work.
4. Click away to blur the editor. Kbds should not work.

### DIFF
--- a/packages/editor/src/lib/TldrawEditor.tsx
+++ b/packages/editor/src/lib/TldrawEditor.tsx
@@ -5,6 +5,7 @@ import React, {
 	ReactNode,
 	memo,
 	useCallback,
+	useEffect,
 	useLayoutEffect,
 	useMemo,
 	useRef,
@@ -336,6 +337,8 @@ const TldrawEditorWithLoadingStore = memo(function TldrawEditorBeforeLoading({
 	return <TldrawEditorWithReadyStore {...rest} store={store.store} user={user} />
 })
 
+const noAutoFocus = () => document.location.search.includes('preserveFocus')
+
 function TldrawEditorWithReadyStore({
 	onMount,
 	children,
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -432,6 +437,38 @@ function TldrawEditorWithReadyStore({
...
+				editor.focus()
+			}
+
+			function handleBlurOnPointerDown() {
+				if (!editor) return
+				editor.blur()
+			}
+
+			if (autoFocus && noAutoFocus()) {
+				editor.getContainer().addEventListener('pointerdown', handleFocusOnPointerDown)
```
SomeHats, line 460: remind me why we didn't want something like this to be built-in? also, don't we already have something that does the blurring part of this?

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -336,6 +337,8 @@ const TldrawEditorWithLoadingStore = memo(function TldrawEditorBeforeLoading({
 	return <TldrawEditorWithReadyStore {...rest} store={store.store} user={user} />
 })
 
+const noAutoFocus = () => document.location.search.includes('preserveFocus')
```
SomeHats, line 340: i'm worried that a library consumer might include this for some other reason and then break things. can we call it something horrible like `tldraw_preserveFocus_Sq06lHBj`?

SomeHats (commented):
this makes sense to me but i kinda wish we had better focus behaviour out of the box

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: double checking

steveruizok: We have a test that seems to suggest that it's possible, though it doesn't work in practice. If I comment out these items then the editor never gains / loses focus.

steveruizok: Oh, there is a test that checks this, but it also sets up the events that make it work. It doesn't / shouldn't work normally! I wish it could but it can't.

PR #4345, opened 2024-08-04
Gracefully handle deleted tools & actions (remix)

steveruizok:
This PR is an attempt to achieve the same goals of https://github.com/tldraw/tldraw/pull/4315 but through a different means.

We introduce three new components: `TldrawUiMenuActionItem`, `TldrawUiMenuActionCheckboxItem`, and `TldrawUiMenuToolItem`. These accept `action` or `tool` ids as props and either return `null` if the associated action / tool is not found, or else return a `TldrawUiMenuItem`.

This PR is additive / non-breaking. All of the existing uses of `TldrawUiMenuItem` still work—you don't _have_ to use that component with an action or tool item—and I haven't changed its usage in places where it's not needed, such as in our examples.

One thing that stands out if we want to solve this problem is that there's probably still some logic in the menus about when to show groups etc that would still be unaware that the action is missing.

### Change type

- [x] `improvement`

### Test plan

1. Delete an action from the actions schema. It should disappear from the context menu / actions menu.
2. Delete a tool from the tools schema. It should disappear from the toolbar.

### Release notes

- When deleting actions or tools using overrides, menu items are automatically removed.


packages/tldraw/src/lib/ui/components/primitives/menus/TldrawUiMenuActionCheckboxItem.tsx
```diff
@@ -0,0 +1,21 @@
+import { useActions } from '../../../context/actions'
+import {
+	TldrawUiMenuCheckboxItem,
+	type TLUiMenuCheckboxItemProps,
+} from './TldrawUiMenuCheckboxItem'
+
+/** @public */
+export type TLUiMenuActionCheckboxItemProps = {
+	actionId?: string
+} & Pick<TLUiMenuCheckboxItemProps, 'disabled' | 'checked' | 'toggle'>
```
steveruizok, line 10: We could spread all of the props, though these are the only ones that are stateful. Maybe there's a better way of typing this to make that more apparent?

PR #4348, opened 2024-08-05
Disable Pages UI when maxPages === 1

ds300:
This PR makes it so that when you set `maxPages` to `1` it disables the UI related to pages

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Make the page management UI disappear when maxPages is 1

### DIFF
--- a/packages/tldraw/src/index.ts
+++ b/packages/tldraw/src/index.ts
@@ -228,10 +228,12 @@ export {
 	ConvertToEmbedMenuItem,
 	CopyAsMenuGroup,
 	CopyMenuItem,
+	CursorChatItem,
 	CutMenuItem,
 	DeleteMenuItem,
 	DuplicateMenuItem,
 	EditLinkMenuItem,
+	EditMenuSubmenu,
 	FitFrameToContentMenuItem,
 	GroupMenuItem,
 	MoveToPageMenu,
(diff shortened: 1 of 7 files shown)


packages/tldraw/src/index.ts
```diff
@@ -391,6 +393,7 @@ export { useCanRedo, useCanUndo } from './lib/ui/hooks/menu-hooks'
 export { useMenuClipboardEvents, useNativeClipboardEvents } from './lib/ui/hooks/useClipboardEvents'
 export { useCopyAs } from './lib/ui/hooks/useCopyAs'
 export { useExportAs } from './lib/ui/hooks/useExportAs'
+export { useIsMultiplayer, useMultiplayerStatus } from './lib/ui/hooks/useIsMultiplayer'
```
ds300, line 396: @SomeHats is it OK to make these public? They are used by the default context menu impl.

SomeHats: all good!

SomeHats (approved):
soooo good

steveruizok:
elegant, tasteful

packages/tldraw/src/index.ts
steveruizok: Should we rename that? Sync status or something? We don't use Multiplayer elsewhere anymore.

ds300: i don't think we made a decision on this in person so I'll just merge for now and we have until the next sdk release to think of alternatives.

steveruizok: PR here: https://github.com/tldraw/tldraw/pull/4349

PR #4349, opened 2024-08-05
Rename `StoreOptions.multiplayerStatus`

steveruizok:
This PR changes the name of `StoreOptions.multiplayerStatus` and a few related hooks.

These APIs have been so far private, so I doubt anyone will run into this as a problem. This only came up when I saw that we were exporting the `useMultiplayerStatus` hooks.

## Naming things (hard?)

We have a small naming crisis in our API. Originally, we'd used the term "multiplayer" for real-time collaboration, ie for the `MultiplayerEditor` component in the dot com app. We also used "collaboration" and "collaborators" in the SDK, e.g. `CollaboratorCursor`.

With the tldraw sync release, we've blurred this a little into the API names. For example, the `useSync` hook was called `useMultiplayer`, but we've come up with `sync` as a product name for our own solution, while documenting real-time collaboration as just "collaboration".

Thus, we have two names:

1. For _general_ real-time collaboration, we use `collaboration`.
2. For _tldraw sync_, we use `sync`.

...and we don't use "multiplayer" in our public APIs.

(We should probably scrub multiplayer from our whole code-base, to avoid some confusion later on, but that's not this PR).

## Store options

As part of tldraw sync, we decided to stash the remote connection status on the `Store`. We use this to decide whether to show the collaboration UI in the `Tldraw` component's UI, as well as to indicate the status of the connection. (This is useful for showing the offline indicator in a case where we _should_ be online but aren't).

Previously, the `StoreOptions.multiplayerStatus` property communicated three things:

1. if `null`, that the store wasn't supposed to be connected to a remote server
2. if a signal, that the store _was_ supposed to be connected to a remote server
3. if a signal, the `multiplayerStatus` was the current connection status of the socket

In this PR, I've added `collaboration` as an object on the store options. It currently has one property, which is `status`.

1. If left undefined, then the store is not supposed to be connected to a remote server.
2. If defined, then the store is supposed to be connected to a remote server
3. If defined, then `collaboration.status` has the signal that represents the connection status

### Change type

- [x] `api`

### Release notes

- Renames `StoreOptions.multiplayerStatus` to `StoreOptions.collaboration.status`.

### DIFF
--- a/packages/editor/src/lib/config/createTLStore.ts
+++ b/packages/editor/src/lib/config/createTLStore.ts
@@ -1,3 +1,4 @@
+import { Signal } from '@tldraw/state'
 import { HistoryEntry, MigrationSequence, SerializedStore, Store, StoreSchema } from '@tldraw/store'
 import {
 	SchemaPropsInfo,
@@ -44,9 +45,13 @@ export type TLStoreSchemaOptions =
 	  }
 
 /** @public */
-export type TLStoreOptions = TLStoreBaseOptions & { id?: string } & TLStoreSchemaOptions &
+export type TLStoreOptions = TLStoreBaseOptions & {
+	id?: string
 	/** Collaboration options for the store. */
-	Pick<TLStoreProps, 'collaboration'>
+	collaboration?: {
+		status: Signal<'online' | 'offline'> | null
+	}
+} & TLStoreSchemaOptions
 
 /** @public */
 export type TLStoreEventInfo = HistoryEntry<TLRecord>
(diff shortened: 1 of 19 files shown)


packages/editor/src/lib/config/createTLStore.ts
```diff
@@ -48,7 +44,9 @@ export type TLStoreSchemaOptions =
 	  }
 
 /** @public */
-export type TLStoreOptions = TLStoreBaseOptions & { id?: string } & TLStoreSchemaOptions
+export type TLStoreOptions = TLStoreBaseOptions & { id?: string } & TLStoreSchemaOptions &
+	/** Collaboration options for the store. */
+	Pick<TLStoreProps, 'collaboration'>
```
SomeHats, line 50: let's leave this specified in the base options, it results in much better reference docs than picking it off this way

SomeHats (approved):
looks good. we'll need to do a release with this before we announce

PR #4350, opened 2024-08-05
Fix issues with resizing cropped images

MitjaBezensek:
While working on #4337 I also discovered two other issues:
* Preview of the whole image (including the cropped out part) was incorrect when shapes were flipped. We weren't applying the scaling styles to the preview. 
* When resizing the shapes via the handles the same issue as in #4337 could occur. We also need to flip if resizing causes the image to flip.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create an image with a crop.
2. Use resizing handles to make it flip. The cropped part should stay the same.
3. Also check the preview of the cropped area when the image is flipped. It should also be flipped.

### Release notes

- Fix a bug with cropped and flipped images and their previews.


steveruizok:
Good catch!

PR #4361, opened 2024-08-06
Hotfix for index keys validation

ds300:
Our logic was bad, so as a stopgap let's just use the underlying library, which will throw an error if we pass it a bad key.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with the index key validation logic


mimecuvalo:
thank you for fixing!

PR #4369, opened 2024-08-08
feat: Able to pass disable to the Icon when UiMenuContext is in mode 'toolbar'

dimitriadamou:
Simple change - unable to pass disabled prop to an icon when the UIMenuContext type is set to 'toolbar'

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


ds300 (approved):
good catch, thanks! I'll add this for the `toolbar-overflow` buttons as well.

PR #4371, opened 2024-08-08
Timer

MitjaBezensek:
Adds a timer component.

Still a bit experimental. Few things that are not yet fully decided:
- UI: I roughly followed the wireframes that Steve showed me. He said he'll probably want to play around with things and update them in the end. I added it to quick actions since it felt the best to use it there. I also explored having it as a shape, but not sure if you'd really need multiple timers or that you'd want to move them around. We could also make it a singleton if we decided to go that route. I just pulled play and pause buttons from Mime's video controls PR, but we'd probably need to design some custom icons as well.
- UX: Do we need the ability to edit time? Feels like just plus and minus buttons might be enough.
- Where the timer should live. Right now it's almost like a plugin and you have to add it to the app. I didn't want investigate a plugin system, but it definitely feels like there should be a bit more structure for these kinds of things.
- We are using a NTP protocol to sync the time. This uses the server time of the worker to sync between clients. Not sure if two clients hitting different workers would have the same time. I'd expect they would, but didn't find any info about it. Cloudflare has something called [time services](https://www.cloudflare.com/time/), hopefully they use this internally to sync their servers.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests


mimecuvalo (commented):
exciting!

i forget (since i was away) if i talked to you or Alex about this but just to make sure you had seen this relevant logic:
- https://www.npmjs.com/package/timing-object
- https://github.com/chrisguttandin/timingsrc
- https://webtiming.github.io/timingsrc/doc/online_mediasync.html

ds300:
This is super promising!

The server offset stuff is good, and seems very accurate in my testing. In the past when I've implemented server offsets we've made them refresh every now and then to prevent drift or bad readings in poor network conditions, but I think we can ignore that for now.

The ux/ui design needs some tweaking.
- I think figjam's time editor is much better, with the directly editable mins/secs and the +1min button.
- This one is really hard to use for longer times (pretty common to have 20-30min timers in my experience)
- The icons don't match with the rest of our icons
- If there's a timer running it should be shown. People who didn't set the timer shouldn't have to click the clock to see it, and then click the clock again to hide it later. It should just appear and disappear for them.
- I like the idea of having a progress bar to show how much time is remaining, but the design feels weird IMO. I'd try something else or just take it out entirely.

Code-wise, its looking good but feels weird to me how it's distributed between dotcom and the tldraw package. 

If it's gonna be part of the tldraw package, we should also probably provide default UI and stuff for it in there too. We should also make the data a first-class part of the document type, rather than using the meta property. Just to keep things well-typed and in the pit of success for migrations etc. Also so far we've left the meta property entirely up to consumers to own. If we start putting things in there ourselves it might mess with their own bits.

But i'm also not convinced it should be part of the tldraw package yet. Like maybe we build it entirely in dotcom userland to let us and others use it for a while before moving it to the sdk?

I really feel like it needs an alarm sound too, and/or some completely unmissable visual feedback when the timer is up. Otherwise folks need to be keeping an eye on the timer and it will distract them.

mimecuvalo:
> I really feel like it needs an alarm sound too, and/or some completely unmissable visual feedback when the timer is up. Otherwise folks need to be keeping an eye on the timer and it will distract them.

We need a sound designer... @TodePond ? :)

ds300:
Closing due to inactivity

PR #4373, opened 2024-08-09
Remove the document.hasFocus check

steveruizok:
Unfortunately it doesn't work reliably, at least not in nextjs apps.

### Change type

- [x] `bugfix`


steveruizok:
@SomeHats let's come back to this

PR #4384, opened 2024-08-14
make arrow labels wrap dynamically

ds300:
Special feature request from @OrionReed 

Our arrow labels were using naive bad caching for the label width. This makes them use smart good caching so they rewrap while you're dragging stuff around, instead of being forced to click in an edit the label.

Before

Kapture 2024-08-14 at 14 02 18

After

Kapture 2024-08-14 at 14 00 41

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Make arrow labels reflow text dynamically as you move things around.


mimecuvalo:
nice! thank you!

PR #4388, opened 2024-08-15
feat(blog): Hook up newsletter form with SendGrid

lukaswiesehan:
### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Submit form with email address
2. Check the SendGrid list

### To do

- [ ] Add SendGrid list id as an environment variable


apps/docs/components/common/newsletter-signup.tsx
```diff
@@ -125,3 +128,13 @@ export const NewsletterSignup: FC<{ size?: 'small' | 'large' }> = ({ size = 'lar
 		</div>
 	)
 }
+
+// Sendgrid List
```
mimecuvalo, line 132: better to put this is the 'submit-newsletter-signup' file if you wanted to keep this reference around

mimecuvalo (approved):
Steve is out (who i think would otherwise review this code)

lgtm!

PR #4397, opened 2024-08-20
Fix an error with parsing embed urls.

MitjaBezensek:
Fix an issue with parsing embed urls. 

If you open the embed dialog, choose one of the embeds and start typing the url you'll see an error. After this change we handle the error correctly.

CleanShot 2024-08-20 at 16 23 56

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open embed dialog and choose one of the embeds.
2. Start typing an url.
3. You should no longer see an error.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix a bug with embed dialog throwing an error when entering an invalid url

### Breaking change
We have removed `getEmbedInfoUnsafely`. Use `getEmbedInfo` instead.

### DIFF
--- a/packages/tldraw/src/lib/utils/embeds/embeds.ts
+++ b/packages/tldraw/src/lib/utils/embeds/embeds.ts
@@ -75,20 +75,6 @@ export type TLEmbedResult =
 	| undefined
 
 /**
- * Tests whether an URL supports embedding and returns the result.
- *
- * @param inputUrl - The URL to match
- * @public
- */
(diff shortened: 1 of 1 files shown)


MitjaBezensek:
On second thought we seem to want to log the error?
https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/utils/embeds/embeds.ts#L101

PR #4398, opened 2024-08-20
Detect multiple installed versions of tldraw packages

SomeHats:
There are two semi-common failure modes around people using tldraw:
1. They upgrade 1 tldraw package, but not the other, resulting in having multiple versions of the same package running at once
2. Their bundler is misconfigured, and pulls in both the commonjs and esmodule versions of their packages.

This diff detects those cases and adds a warning message explaining the situation.

As part of this, I added another check on packages in the same vein as check-scripts and check-packages. Rather than add another script and step on github, I merged the three of them into one file.

### Change type
- [x] `api`

### Release notes

- We detect when there are multiple versions of tldraw installed and let you know, as this can cause bugs in your application


ds300 (approved):
good work, you already caught all the edge cases I could think of :)

PR #4403, opened 2024-08-21
add default <foreignObject> based export for shapes

SomeHats:
Custom shapes (and our own bookmark shapes) now support SVG exports by default! The default implementation isn't the most efficient and won't work in all SVG environments, but you can still write your own if needed. It's pretty reliable though!

Kapture 2024-08-27 at 17 29 31

This introduces a couple of new APIs for co-ordinating SVG exports. The main one is `useDelaySvgExport`. This is useful when your component might take a while to load, and you need to delay the export is until everything is ready & rendered.  You use it like this:
```tsx
function MyComponent() {
    const exportIsReady = useDelaySvgExport()
    const [dynamicData, setDynamicData] = useState(null)

    useEffect(() => {
        loadDynamicData.then((data) => {
            setDynamicData(data)
            exportIsReady()
        })
    })

    return <MyList data={data} />
}
```

This is a pretty low-level API that I wouldn't expect most people using these exports to need, but it does come in handy for some things.

### Change type

- [x] `improvement`

### Release notes

Custom shapes (and our own bookmark shapes) now render in image exports by default.


steveruizok:
Glad to see this come together Alex! I'd like to see more examples of what it will produce for HTML content, so that we can be sure that this is reliable enough to ship. My general principle is to only ship features in the SDK that we can really "solve", and which we feel confident in standing behind—and leave the rest up to the user. Converting HTML to SVG was one of those places where I couldn't find any solution that would reliably produce good results (except perhaps taking screenshots in Playwright). This PR seems much more advanced than anything I tried, so there's a good chance it's above that threshold for quality / consistency, but I'd like to be sure. Maybe we can work with some users who have created custom shapes?

packages/editor/src/lib/exports/html-to-image/clonePseudos.ts
```diff
@@ -0,0 +1,56 @@
...
+	style: CSSStyleDeclaration
+): Promise<Text> {
+	const selector = `.${className}:${pseudo}`
+	const cssText = await formatCSSProperties(style)
+
+	return document.createTextNode(`${selector}{${cssText}}`)
+}
+
+async function clonePseudoElement(node: HTMLElement, pseudo: Pseudo) {
+	const style = window.getComputedStyle(node, pseudo)
```
mimecuvalo, line 35: e.g. here it's just `:before` instead of `::before` - i'm sure it works just fine, but technically it should be `::`, eh?

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,67 @@
...
+	renderTarget.className = 'tldraw-svg-export'
+	renderTarget.inert = true
+	renderTarget.tabIndex = -1
+	Object.assign(renderTarget.style, {
+		position: 'absolute',
+		top: '0px',
+		left: '0px',
+		width: result.width + 'px',
+		height: result.height + 'px',
+		zIndex: '110000000',
```
mimecuvalo, line 26: very specific zIndex :P might as well make it 2147483647?

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+		return { svg, width: result.width, height: result.height }
+	} finally {
+		root.unmount()
+		container.removeChild(renderTarget)
+	}
+}
+
+async function applyChangesToForeignObjects(svg: SVGSVGElement) {
+	const foreignObjectChildren = svg.querySelectorAll('foreignObject.tl-shape-foreign-object > *')
```
mimecuvalo, line 50: this query selector could use a comment

apps/examples/src/misc/end-to-end.tsx
```diff
@@ -1,11 +1,50 @@
...
+		return <rect width={shape.props.w} height={shape.props.h} />
+	}
+}
+function HtmlCssShapeComponent({ shape }: { shape: HtmlCssShape }) {
+	// eslint-disable-next-line local/no-at-internal
+	const id = useSafeId()
+
+	useLayoutEffect(() => {
+		const style = document.createElement('style')
+		style.textContent = shape.props.css.replace(/#self/g, `#${id}`)
```
mimecuvalo, line 38: what is this `#self` doing?

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -0,0 +1,294 @@
...
+		// Expand by an extra 32 pixels
+		bbox.expandBy(padding)
+	}
+
+	// We want the svg image to be BIGGER THAN USUAL to account for image quality
+	const w = bbox.width * scale
+	const h = bbox.height * scale
+
+	try {
+		document.body.focus?.() // weird but necessary
```
mimecuvalo, line 75: why is it necessary?

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -0,0 +1,294 @@
...
+	const singleFrameShapeId =
+		ids.length === 1 && editor.isShapeOfType<TLFrameShape>(editor.getShape(ids[0])!, 'frame')
+			? ids[0]
+			: null
+	if (!singleFrameShapeId) {
+		// Expand by an extra 32 pixels
+		bbox.expandBy(padding)
+	}
+
+	// We want the svg image to be BIGGER THAN USUAL to account for image quality
```
mimecuvalo, line 70: bigger than usual, because of retina? or some other reason?

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -0,0 +1,294 @@
...
+							pageTransformString = `${pageTransformString} scale(${shape.props.scale}, ${shape.props.scale})`
+						}
+					}
+
+					// Create svg mask if shape has a frame as parent
+					const pageMask = editor.getShapeMask(shape.id)
+					const shapeMask = pageMask
+						? Mat.From(Mat.Inverse(pageTransform)).applyToPoints(pageMask)
+						: null
+					const shapeMaskId = `mask_${shape.id.replace(':', '_')}`
```
mimecuvalo, line 130: we gotta make this a fn in the codebase - we do this in several places 🙃

mimecuvalo (commented):
:phew: what a PR :)

i'll need another readthrough but thought i'd leave my initial comments on it!

packages/utils/src/lib/media/media.ts
```diff
@@ -60,6 +61,60 @@ export class MediaHelpers {
...
+				ctx.drawImage(video, 0, 0)
+				promise.resolve(canvas.toDataURL())
+			}
+		}
+		const onError = (e: Event) => {
+			console.error(e)
+			promise.reject(new Error('Could not get video frame'))
+		}
+
+		video.addEventListener('loadedmetadata', onReadyStateChanged)
```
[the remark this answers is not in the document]
SomeHats: there's no one ready-state change listener, you have to manually cover the cases that might cause the ready state to change

packages/editor/src/lib/exports/html-to-image/README.md
```diff
@@ -0,0 +1,16 @@
+# html-to-image
+
+The code in this folder has been very roughly adapted from html-to-image:
```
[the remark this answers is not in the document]
SomeHats: so since you reviewed i ran into some further issues with how we were doing things before - reading & writing in the same pass was causing the browser to constantly recalculate layout etc. 

I've since reworked the code here futher, and at this point there's pretty much zero left from html-to-image, so I've removed this entirely. I can bring it back if you think it's worth it though

packages/editor/src/lib/exports/html-to-image/README.md
```diff
@@ -0,0 +1,16 @@
...
+resources embedded. It then takes that clone, and embeds it (in with `<foreignObject>`) and the
+external resources in an SVG.
+
+We start with our existing SVG export generation, which includes `<foreignObject>`s for all shapes
+without a `toSvg` method. Then, in a single pass, we embed any relevant CSS and external resources.
+We also skip certain things that html-to-image does that aren't as relevant to us, like supporting
+same-origin iframes.
+
+Whilst the code here takes inspiration from html-to-image and in some cases is pretty closely
+adapted (as in `clonePseudos.ts`), most is written from scratch.
```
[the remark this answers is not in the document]
SomeHats: (resolving in favour of above discussion)

packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx
```diff
@@ -151,20 +153,19 @@ export class VideoShapeUtil extends BaseBoxShapeUtil<TLVideoShape> {
 		return <rect width={toDomPrecision(shape.props.w)} height={toDomPrecision(shape.props.h)} />
 	}
 
-	override toSvg(shape: TLVideoShape) {
-		return <image href={serializeVideo(shape.id)} width={shape.props.w} height={shape.props.h} />
+	override async toSvg(shape: TLVideoShape) {
+		const image = await serializeVideo(this.editor, shape)
```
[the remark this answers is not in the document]
SomeHats: we no longer update `props.time` as the video plays. We should really remove the prop entirely. I don't want to pass it in as there's no longer any way to control it

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -12,12 +12,13 @@ export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null, width: n
...
 	const [url, setUrl] = useState<string | null>(null)
 	const [isPlaceholder, setIsPlaceholder] = useState(false)
+	const isExport = !!useSvgExportContext()
 	const asset = assetId ? editor.getAsset(assetId) : null
 	const culledShapes = editor.getCulledShapes()
 	const isCulled = culledShapes.has(shapeId)
 	const didAlreadyResolve = useRef(false)
 
-	useEffect(() => {
+	useLayoutEffect(() => {
```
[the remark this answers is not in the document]
SomeHats: this was originally to get this resolving _before_ we did an SVG export, but it predates the `useDelaySvgExport` stuff. I've reverted and updated this to use that instead. 

fwiw, i think this would only have caused a double resolve in react strict mode in development

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -60,13 +62,19 @@ export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null, width: n
 		}
 
 		// If we already resolved the URL, debounce fetching potentially multiple image variations.
-		const timer = editor.timers.setTimeout(resolve, didAlreadyResolve.current ? 500 : 0)
-
-		return () => {
-			clearTimeout(timer)
-			isCancelled = true
+		if (didAlreadyResolve.current) {
+			const timer = editor.timers.setTimeout(resolve, 500)
```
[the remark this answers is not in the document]
SomeHats: this was originally to get this resolving _before_ we did an SVG export, but it predates the `useDelaySvgExport` stuff. I've reverted and updated this to use that instead. 

fwiw, i think this would only have caused a double resolve in react strict mode in development

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -127,6 +129,25 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
 
 		const pageRotation = this.editor.getShapePageTransform(shape)!.rotation()
 
+		if (svgExport) {
+			// for SVG exports, we show a blank embed
+			return (
+				<HTMLContainer className="tl-embed-container" id={shape.id}>
```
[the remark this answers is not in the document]
SomeHats: Maybe? I think it's tricky because we do sort of allocate space for it in the document. Right now we just render a blank rectangle in exports. Mind if i leave this as-is now as this is the same as the existing behaviour? We can follow up if we decide we want to do something like this

packages/editor/src/lib/exports/html-to-image/fetchCache.ts
```diff
@@ -0,0 +1,27 @@
+import { FileHelpers, assert, fetch } from '@tldraw/utils'
+
+export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
+	const cache = new Map<string, Promise<T | null>>()
```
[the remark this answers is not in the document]
SomeHats: y e s... i'm not sure what makes the most sense here.

On the one hand, it does make things _much_ faster. There are certain things I think we should sort of just always cache, like fonts. We could maybe do some sort of LRU cache for other things tho... I'll have a bit of a think about what we can do here

packages/editor/src/lib/exports/html-to-image/embedCss.ts
```diff
@@ -0,0 +1,185 @@
+import { bind, compact, safeParseUrl } from '@tldraw/utils'
+import { fetchCache, resourceToDataUrl } from './fetchCache'
+
+interface FontFace {
+	fontFace: string
+	urls: { original: string; resolved: string | null; embedded?: Promise<string | null> }[]
+	fontFamilies: Set<string>
+}
+
+const importsRegex =
```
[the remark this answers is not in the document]
SomeHats: I've linked to the grammars these are based on on MDN. These functions are tested in `parseCss.test.ts` - i think that should be sufficient for examples?

packages/editor/src/lib/exports/html-to-image/embedCss.ts
```diff
@@ -0,0 +1,185 @@
...
+	/@import\s+(?:"([^"]+)"|'([^']+)'|url\s*\(\s*(?:"([^"]+)"|'([^']+)'|([^'")]+))\s*\))([^;]+);/gi
+const fontFaceRegex = /@font-face\s*{([^}]+)}/gi
+const urlsRegex = /url\s*\(\s*(?:"([^"]+)"|'([^']+)'|([^'")]+))\s*\)/gi
+const fontFamilyRegex =
+	/(?:^|;)\s*font-family:\s*(?:([^'"][^;\n]+)|"([^"]+)"|'([^']+)')\s*(?:;|$)/gi
+
+function parseCssImports(css: string) {
+	return Array.from(css.matchAll(importsRegex), (m) => ({
+		url: m[1] || m[2] || m[3] || m[4] || m[5],
```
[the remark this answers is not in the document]
SomeHats: Named capturing groups don't have good browser support when using the same name across disjunctions at the moment

packages/editor/src/lib/exports/html-to-image/embedCss.ts
```diff
@@ -0,0 +1,185 @@
...
+		}
+
+		valueRegex.lastIndex = separatorRegex.lastIndex
+	}
+
+	return fontFamilies
+}
+
+export function shouldIncludeCssProperty(property: string) {
+	if (property.startsWith('-')) return false
```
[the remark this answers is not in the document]
SomeHats: We only apply values that are effective on an element. So if @keyframes are applied, the current value from them will be used. Same with @media. The actual keyframe/media blocks won't be included in the resulting SVG though

packages/editor/src/lib/exports/html-to-image/decorateAndEmbed.ts
```diff
@@ -0,0 +1,137 @@
...
+
+async function getVideoReplacement(video: HTMLVideoElement) {
+	try {
+		const dataUrl = await MediaHelpers.getVideoFrameAsDataUrl(video)
+		return createImage(dataUrl, video)
+	} catch (err) {
+		console.error('Could not get video frame', err)
+	}
+
+	if (video.poster) {
```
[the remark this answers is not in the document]
SomeHats: we don't, but this sort of thing is what this attribute is intended for

packages/editor/src/lib/exports/html-to-image/decorateAndEmbed.ts
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if (node.style.fontFamily) {
+		const parsed = parseCssFontFamilyValue(node.style.fontFamily)
+		for (const font of parsed) {
+			opts.onFoundUsedFont(font)
+		}
+	}
+
+	if (node.style.fontKerning === 'auto') {
```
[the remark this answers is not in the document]
SomeHats: added a comment here - essentially x-browser nonsense

packages/editor/src/lib/exports/html-to-image/decorateAndEmbed.ts
```diff
@@ -0,0 +1,137 @@
...
+		}
+		return node
+	}
+
+	// no support for iframes
+	return node
+}
+
+function applyInputValue(node: HTMLElement) {
+	if (node instanceof HTMLInputElement) {
```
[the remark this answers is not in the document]
SomeHats: added a comment

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: this was just for debugging - removed!

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: commented this whole file

apps/examples/src/misc/end-to-end.tsx
SomeHats: added a comment for this - it's so test cases can write CSS targeting an individual shape

packages/editor/src/lib/exports/getSvgJsx.tsx
SomeHats: no clue - i did a bit of digging, and this comes from the very first implementation of SVG export back in 2022: https://github.com/tldraw/brivate/commit/4a4b16a2b6c5a3c819b91719baa8a64d36f73b09

packages/editor/src/lib/exports/getSvgJsx.tsx
SomeHats: not sure - this was already in this file

packages/utils/src/lib/media/media.ts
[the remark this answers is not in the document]
mimecuvalo: just for my understanding: i guess maybe my question is - why not just wait til the `loadeddata` event? can `loadedmetadata` event happen after `loadeddata`? also, does doing `video.currentTime = time` cause a `seeked` event? `loadedData` afaik is when the desired frame is available which is what you want.

packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx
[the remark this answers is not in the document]
mimecuvalo: it's hard to say but it may be that the `props.time` is temporarily gone? even if we don't bring back sync, we might want to be able to save the time at a certain starting position.
but i don't feel too strongly about this.

packages/tldraw/src/lib/shapes/shared/useAsset.ts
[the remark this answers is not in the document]
mimecuvalo: sadly it wasn't the double resolve in React that was the issue. the problem was that upon page load the Camera started at an initial zoom which propagated to all the shapes. and _then_ once the editor finally loaded it would set the actual initial zoom. this caused two image requests to be happily sent off :-/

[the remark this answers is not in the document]
mimecuvalo: (unresolving this briefly just so you see it) i'll also add a comment about this in the code

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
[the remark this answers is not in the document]
mimecuvalo: cool and yes! it's totally fine for now, something to think about for future PRs

packages/editor/src/lib/exports/html-to-image/fetchCache.ts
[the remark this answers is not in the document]
mimecuvalo: it's funny because i can hear you actually saying 'y e s` in my head :P

no worries for this PR, I think for all intents and purposes it's probably fine. but then again, for bigger canvases, it might add up.

packages/editor/src/lib/exports/parseCss.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export interface ParsedFontFace {
+	fontFace: string
+	urls: { original: string; resolved: string | null; embedded?: Promise<string | null> }[]
+	fontFamilies: Set<string>
+}
+
+// parsing CSS with regular expressions doesn't really work. There are almost certainly edge cases
+// in the regular expressions we're using here. They're based formal grammars of CSS, but aren't
+// perfect. These were written with https://regexper.com/ - give it a go if you're editing these.
```
mimecuvalo, line 11: this is resource is very handy! nice find 🤘

packages/editor/src/lib/exports/html-to-image/embedCss.ts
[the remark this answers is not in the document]
mimecuvalo: that regexper resource is super cool. thank you for adding all those links! that'll be very handy for maintainability.

long-term for our codebase i'd love to have a pattern of making the regex's more split-up and readable, something along these lines or so (but don't worry about it for this PR): https://stackoverflow.com/questions/12317049/how-to-split-a-long-regular-expression-into-multiple-lines-in-javascript

packages/editor/src/lib/exports/html-to-image/embedCss.ts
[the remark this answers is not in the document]
mimecuvalo: ah interesting, good to know!

packages/editor/src/lib/exports/getSvgJsx.tsx
mimecuvalo: hah ok 🙃

packages/utils/src/lib/media/media.ts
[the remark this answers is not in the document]
SomeHats: couple things:
1. we can't set `currentTime` until after `loadedmetadata` or the video will error
2. i don't think we reliably get `loadeddata` called - i think if it's already loaded that part of the video when you seek, you don't always get that event. you do always get `seeked` though? but sometimes later, it seems

we can probably do some experimentation here and pare this down a bit? but what i have now is already the result of a good bit of trial and error

packages/tldraw/src/lib/shapes/shared/useAsset.ts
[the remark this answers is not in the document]
SomeHats: i think david's deep link work resolved that - we now initialize the camera at the correct place until waiting for the document to render before moving the camera

apps/docs/content/docs/editor.mdx
```diff
@@ -520,10 +520,32 @@ define a number of different callbacks describing how the binding should behave.
 [BindingUtil](?) reference or a [complete example of a
 binding](/examples/shapes/tools/sticker-bindings) for details.
 
-If the strucutre of your binding needs to change over time, you can provide
+If the structure of your binding needs to change over time, you can provide
 [migrations](/docs/persistence#Shape-props-migrations) describing how old stored bindings can be
 brought up to date.
 
+## Image exports
```
mimecuvalo, line 527: side rant: we should split up the Editor page - it's sooo loooong.
but this is neither here nor there for this PR :P

apps/examples/package.json
```diff
@@ -39,6 +39,8 @@
 		"@tldraw/dotcom-shared": "workspace:*",
 		"@tldraw/sync": "workspace:*",
 		"@vercel/analytics": "^1.1.1",
+		"ag-grid-community": "^32.1.0",
```
mimecuvalo, line 42: this is fine for this PR but i'm wondering in the future, for AG grid and pdf libs, that we delay load them/put them in a separate JS module. might start getting out of hand that the examples app has to get a bunch of code for only one of the examples.

packages/tldraw/src/lib/utils/export/copyAs.ts
```diff
@@ -29,6 +29,10 @@ export function copyAs(
 	if (window.navigator.clipboard.write) {
 		const { blobPromise, mimeType } = exportToBlobPromise(editor, ids, format, opts)
 
+		blobPromise.catch((err) => {
+			console.error(err)
```
mimecuvalo, line 33: should this do something more than just write to console? no toast or anything? does it basically silently fail?

packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -51,22 +61,23 @@ export function useAsset(shapeId: TLShapeId, assetId: TLAssetId | null, width: n
...
 				setUrl(resolvedUrl)
 				setIsPlaceholder(false)
+				isReady()
 			}
 		}
 
 		// If we already resolved the URL, debounce fetching potentially multiple image variations.
-		const timer = editor.timers.setTimeout(resolve, didAlreadyResolve.current ? 500 : 0)
-
+		const timer = editor.timers.setTimeout(resolve, didAlreadyResolve ? 500 : 0)
```
mimecuvalo, line 75: wait, why is the `.current` gone? typo?

packages/editor/src/lib/exports/fetchCache.ts
```diff
@@ -0,0 +1,27 @@
+import { FileHelpers, assert, fetch } from '@tldraw/utils'
+
+export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
+	const cache = new Map<string, Promise<T | null>>()
```
mimecuvalo, line 4: let's add a TODO here that we should consider making sure this doesn't grow unbounded in the future

packages/editor/src/lib/components/default-components/DefaultCanvas.tsx
```diff
@@ -497,6 +497,25 @@ function DebugSvgCopy({ id }: { id: TLShapeId }) {
 
 	if (!isInRoot || !image) return null
 
+	if (mode === 'iframe') {
```
mimecuvalo, line 500: when/why is this iframe used?

packages/editor/src/lib/exports/ExportDelay.tsx
```diff
@@ -0,0 +1,47 @@
+import { bind } from '@tldraw/utils'
+
+export class ExportDelay {
+	private isResolved = false
+	private readonly promisesToWaitFor: Promise<void>[] = []
+
+	constructor(private readonly maxDelayTimeMs: number) {}
+
+	@bind waitUntil(promise: Promise<void>): void {
+		// console.log('waitUntil', new Error().stack)
```
mimecuvalo, line 10: old comment

packages/editor/src/lib/exports/ExportDelay.tsx
```diff
@@ -0,0 +1,47 @@
...
+		if (this.isResolved) {
+			throw new Error(
+				'Cannot `waitUntil` - the export has already been resolved. Make sure to call `waitUntil` as soon as possible during an export - ie within the first react effect after rendering.'
```
mimecuvalo, line 13: this comment makes me think an example would be good. i would have expected the AG grid example to have a waitUntil but it looks like it's not in there? i thought you had mentioned the AG grid required it but maybe I misheard

packages/editor/src/lib/exports/ExportDelay.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+	private async resolvePromises() {
+		let lastLength = null
+		while (this.promisesToWaitFor.length !== lastLength) {
+			lastLength = this.promisesToWaitFor.length
+			await Promise.allSettled(this.promisesToWaitFor)
+
+			// wait for a cycle of the event loop to allow any of those promises to add more if needed.
+			// eslint-disable-next-line no-restricted-globals
+			await new Promise((r) => setTimeout(r, 0))
```
mimecuvalo, line 29: these setTimeouts are prbly fine. might be good in the future to have a dispose on this class to get rid of any pending timeouts (maybe a TODO)

packages/editor/src/lib/exports/ExportDelay.tsx
```diff
@@ -0,0 +1,47 @@
...
+			setTimeout(() => r('timeout'), this.maxDelayTimeMs)
+		)
+		const resolvePromise = this.resolvePromises().then(() => 'resolved' as const)
+
+		const result = await Promise.race([timeoutPromise, resolvePromise])
+		if (result === 'timeout') {
+			console.warn('[tldraw] Export delay timed out after ${this.maxDelayTimeMs}ms')
+		}
+
+		this.isResolved = true
```
mimecuvalo, line 45: this file in general is fancy! would be good to have an overview at the top explaining what it's doing. it can be a little hard to understand at first.

packages/editor/src/lib/exports/FontEmbedder.ts
```diff
@@ -0,0 +1,101 @@
...
+
+		for (const fontFace of this.fontFacesToEmbed) {
+			let fontFaceString = `@font-face {${fontFace.fontFace}}`
+
+			for (const url of fontFace.urls) {
+				if (!url.embedded) continue
+				const dataUrl = await url.embedded
+				if (!dataUrl) continue
+
+				fontFaceString = fontFaceString.replace(url.original, dataUrl)
```
mimecuvalo, line 56: worth mentioning with a comment here what this accomplishing, or perhaps better just an overview comment at the top of the file that you're doing some embedding here

packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -0,0 +1,310 @@
...
+	unwrapCustomElements() {
+		const visited = new Set<Node>()
+
+		const visit = (element: Element, clonedParent: Element | null) => {
+			if (visited.has(element)) return
+			visited.add(element)
+
+			const shadowRoot = element.shadowRoot
+
+			// if (shadowRoot || clonedParent) debugger
```
mimecuvalo, line 108: nix

packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -0,0 +1,310 @@
...
+				}
+
+				const clonedEl = element.cloneNode(false) as Element
+				this.styles.set(clonedEl, this.styles.get(element)!)
+
+				clonedParent.appendChild(clonedEl)
+
+				for (const child of getRenderedChildNodes(element)) {
+					if (child instanceof Element) {
+						visit(child, clonedEl)
```
mimecuvalo, line 139: this might be worth having a max-depth check here somewhere. i'm worried about an infinite loop potentially...

packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -0,0 +1,310 @@
...
+// when we're figuring out the default values for a tag, we need read them from a separate document
+// so they're not affected by the current document's styles
+let defaultStyleFrame:
+	| { iframe: HTMLIFrameElement; foreignObject: SVGForeignObjectElement; document: Document }
+	| undefined
+const defaultStylesByTagName: Record<string, ReadonlyStyles> = {}
+function getDefaultStyleFrame() {
+	if (!defaultStyleFrame) {
+		const frame = document.createElement('iframe')
+		frame.inert = true
```
mimecuvalo, line 276: worth having a comment here what `inert` is preventing

packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -0,0 +1,310 @@
...
+		frame.style.display = 'none'
+		document.body.appendChild(frame)
```
mimecuvalo, line 278: out of curiosity, is it strictly necessary to add to the document?
can something like `DocumentFragment` be used? https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment

packages/editor/src/lib/exports/domUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+	if (isShadowSlotElement(node)) {
+		const assignedNodes = node.assignedNodes()
+		if (assignedNodes?.length) {
+			return assignedNodes
+		}
+	}
+	return node.childNodes
+}
+
+export function* getRenderedChildren(node: Element) {
```
mimecuvalo, line 14: generator, fancy 😎

packages/editor/src/lib/exports/domUtils.ts
```diff
@@ -0,0 +1,46 @@
+export function getRenderedChildNodes(node: Element): Iterable<Node> {
+	if (node.shadowRoot) {
+		return node.shadowRoot.childNodes
+	}
+	if (isShadowSlotElement(node)) {
+		const assignedNodes = node.assignedNodes()
```
mimecuvalo, line 6: a comment here would be good, i don't think many of use are very familiar with shadow dom and what this is doing 🙃

packages/editor/src/lib/exports/embedMedia.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export async function embedMedia(node: HTMLElement) {
+	if (node instanceof HTMLCanvasElement) {
+		return replace(node, await getCanvasReplacement(node))
+	} else if (node instanceof HTMLVideoElement) {
+		return replace(node, await getVideoReplacement(node))
+	} else if (node instanceof HTMLImageElement) {
+		const src = node.currentSrc || node.src
+		const dataUrl = await resourceToDataUrl(src)
```
mimecuvalo, line 68: worth using the `createImage` fn?

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,128 @@
...
+	const result = getSvgJsx(editor, shapeIds, opts)
+	if (!result) return undefined
+
+	// we need to render that SVG into a real DOM element that's actually laid out in the document.
+	// without this CSS and layout aren't computed correctly, which we need to make sure any
+	// <foreignObject> elements have their styles and content inlined correctly.
+	const container = editor.getContainer()
+	const renderTarget = document.createElement('div')
+	renderTarget.className = 'tldraw-svg-export'
+	renderTarget.inert = true
```
mimecuvalo, line 28: this is slightly different than the above - you have `inert` but also a display: none but not the opacity: 0. and also this one has the tabIndex but the other one doesn't.
any reason they should be different? should they be consistent?

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+	// create a react root...
+	const root = createRoot(renderTarget)
+	try {
+		// ...and render the SVG into it.
+		flushSync(() => {
+			root.render(result.jsx)
+		})
+
+		// Some operation take a while - for example, waiting for an asset to load in. We give shape
```
mimecuvalo, line 49: sp: operations

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,128 @@
...
+	if (!result) return undefined
+
+	// we need to render that SVG into a real DOM element that's actually laid out in the document.
+	// without this CSS and layout aren't computed correctly, which we need to make sure any
+	// <foreignObject> elements have their styles and content inlined correctly.
+	const container = editor.getContainer()
+	const renderTarget = document.createElement('div')
+	renderTarget.className = 'tldraw-svg-export'
+	renderTarget.inert = true
+	renderTarget.tabIndex = -1
```
mimecuvalo, line 29: same question also with `DocumentFragment`

packages/editor/src/lib/exports/exportToSvg.tsx
```diff
@@ -0,0 +1,128 @@
...
+		// the document, these elements work exactly as we'd expect from other dom elements - they
+		// can load external resources, and any stylesheets in the document apply to them as we
+		// would expect them to. But when we pull the SVG into its own file or draw it to a canvas
+		// though, it has to be completely self-contained. We embed any external resources, and
+		// apply any styles directly to the elements themselves.
+		await applyChangesToForeignObjects(svg)
+
+		return { svg, width: result.width, height: result.height }
+	} finally {
+		// eslint-disable-next-line no-restricted-globals
```
mimecuvalo, line 67: i'm gonna be a stickler here and say you should use editor.timers :)

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -0,0 +1,383 @@
...
+		if (didRenderRef.current) {
+			throw new Error('SvgExport should only render once - do not use with react strict mode')
+		}
+		didRenderRef.current = true
+		;(async () => {
+			const shapeDefs: Record<string, ReactElement> = {}
+
+			const unorderedShapeElementPromises = renderingShapes.map(
+				async ({ id, opacity, index, backgroundIndex }) => {
+					// Don't render the frame if we're only exporting a single frame
```
mimecuvalo, line 176: typo: "render the shape" i think

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -0,0 +1,383 @@
...
+import { Box } from '../primitives/Box'
+import { Mat } from '../primitives/Mat'
+import { ExportDelay } from './ExportDelay'
+
+export function getSvgJsx(editor: Editor, ids: TLShapeId[], opts: TLImageExportOptions = {}) {
+	if (!window.document) throw Error('No document')
+
+	const {
+		scale = 1,
+		background = false,
```
mimecuvalo, line 41: i know this code precedes you but a comment explaining what `background` does would be helpful.

mimecuvalo (commented):
ok! phew some more comments - lookin' good!

packages/tldraw/src/lib/shapes/shared/useAsset.ts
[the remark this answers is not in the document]
mimecuvalo: ah neat. lemme take a look separately from this PR then

packages/utils/src/lib/media/media.ts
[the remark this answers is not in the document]
mimecuvalo: gotcha, makes sense. i'll be a broken record about it then: worth adding a comment about it :)
a year from now we'll be wondering about this and it'll look scarier to refactor if we don't know why they were all needed :)

packages/tldraw/src/lib/utils/export/copyAs.ts
SomeHats: there's already a toast saying that the copy failed, but without this we get a generic message in the console about it. This shows us the actual cause of the failure in the console

packages/tldraw/src/lib/shapes/shared/useAsset.ts
SomeHats: yes! good catch, ty

apps/docs/content/docs/editor.mdx
SomeHats: desperate to split this up 😭

apps/examples/package.json
SomeHats: we actually already do this automatically! each example is set up to be lazy-loaded when you navigate to it, so AG grid doesn't get loaded until you open the example that uses it

packages/editor/src/lib/exports/fetchCache.ts
SomeHats: good call

packages/editor/src/lib/components/default-components/DefaultCanvas.tsx
SomeHats: this is a debug utility. `img` give a more accurate/lighter-weight view of how things will be rendered, but switching to `iframe` is useful to inspect the generated CSS in the browser dev tools

packages/editor/src/lib/exports/ExportDelay.tsx
SomeHats: `waitUntil` is a bit low-level. The docs for it refer you to `useDelaySvgExport`, which is built on top of it and is what AG grid uses. I could mark `waitUntil` as internal for now?

packages/editor/src/lib/exports/ExportDelay.tsx
SomeHats: added!

packages/editor/src/lib/exports/FontEmbedder.ts
SomeHats: added a comment at the top giving an overview of what this class is for and how it works

packages/editor/src/lib/exports/StyleEmbedder.ts
SomeHats: Is there a way we could have a loop that wouldn't be prevented by the `visited` set?

packages/editor/src/lib/exports/StyleEmbedder.ts
SomeHats: yes - `getComputedStyle`/`computedStyleMap` needs the element to be in the DOM to have the right values & layout applied

packages/editor/src/lib/exports/domUtils.ts
SomeHats: 😎

packages/editor/src/lib/exports/embedMedia.ts
SomeHats: I wanted to keep the original image in place so we don't need to copy attributes etc.

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: sorry, the above where?

SomeHats: oh i see!

yea so for this one, we need full layout etc. which is why this goes in the document and is rendered at 0 opacity.

for the other one, we need `contentDocument`, which is only populated once the iframe is added to the document. we don't need full layout tho, which is why we can use `display: none`. we don't need `inert` there either because of that, so i've removed it

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: explained above

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: if i use editor.timers and the editor disposes, then these will never unmount

apps/examples/package.json
mimecuvalo: ah killer! 🙌

packages/editor/src/lib/exports/ExportDelay.tsx
mimecuvalo: ah no i see it now, somehow i had missed it - nevermind, it makes sense as is

packages/editor/src/lib/exports/ExportDelay.tsx
mimecuvalo: sounds good and makes sense!

packages/editor/src/lib/exports/StyleEmbedder.ts
mimecuvalo: i think the `visited` set suffices, it's just a paranoia thing. i guess seeing the `cloneNode`+recursion makes me think it'll be generating accidentally more nodes and then continue consuming them ad infinitum. 

the browser will throw anyway if we hit recursion so it's prbly fine

packages/editor/src/lib/exports/StyleEmbedder.ts
mimecuvalo: with all you shadow dom work, it makes me think we should be putting this in a shadow node heh. 

it's fine though but part of me is worrying about pollution of the rest of the dom potentially (if something isn't buttoned up either via this PR or future tweaks).

(actually, side-note, i wonder if we should have a "sandbox" node to add temporary nodes like this. since tldraw is meant to be integrated into a larger app, it feels like we should maybe stay within a generally `inert` div vs. adding it to the top-level of the `<body>` element)

packages/editor/src/lib/exports/exportToSvg.tsx
mimecuvalo: gotcha gotcha - ok, makes sense!

packages/editor/src/lib/exports/exportToSvg.tsx
mimecuvalo: ahh good point. then, maybe worth changing to be `container?.` since container might have been disposed potentially

mimecuvalo (approved):
LGTM! i can't believe you wrote an html-to-image generator. damn, alex, dust that dirt off your shoulders!

this may be minor but for a followup PR it's worth thinking about maybe appending these nodes to something else besides `document.body`. i think it'd be good to keep it contained, so-to-speak, to some special tldraw `<div>` (shadow?) of the DOM, sort of a defense-in-depth thing in my head, esp. as it gets integrated alongside other things.

packages/editor/src/lib/exports/exportToSvg.tsx
SomeHats: container wouldn't be set to `null` or anything as a result of that, right? we'd still have a reference to the node, the node just wouldn't be in the DOM any more. I don't think the `?` would do anything?

mimecuvalo: ah yes! nvm

PR #4404, opened 2024-08-21
fix the box shape tool not exiting the resizing state

SomeHats:
no release notes, this bug was introduced within this release

### Change type

- [x] `bugfix`


MitjaBezensek:
Nice catch 👏

PR #4410, opened 2024-08-22
inline nanoid

SomeHats:
We have a bunch of code working around the fact that nanoId is only distributed as an ES module, but we run both as es and commonjs modules. luckily, nanoid is nano! and even more so if you strip out the bits we don't use. This replaces the nanoid library with a vendored version of just the part we use.

### Change type

- [x] `improvement`


packages/utils/src/lib/id.ts
```diff
@@ -0,0 +1,100 @@
+/**
```
mimecuvalo, line 1: nit: for licenses do `/*!` to preserve the license comment from munging (we use this in other places in the codebase too)

packages/utils/src/lib/id.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * `nanoid` is currently only distributed as an ES module. some tools (jest, playwright) don't
+ * properly support ESM-only code yet, and tldraw itself is distributed as both an ES module and a
+ * CommonJS module. By including nanoid here, we can make sure it works well in every environment
+ * where tldraw is used. We can also remove some unused features like custom alphabets.
+ *
+ * @license
+ * The MIT License (MIT)
+ *
+ * Copyright 2017 Andrey Sitnik <andrey@sitnik.ru>
```
mimecuvalo, line 12: it's sufficient to just link to the license file, you don't have to re-include the MIT license every time. so,

```
/*!
 * MIT License
 * Modified code originally from <https://github.com/ai/nanoid>
 * Copyright 2017 Andrey Sitnik <andrey@sitnik.ru>
 */
 ```

PR #4412, opened 2024-08-22
support tc39 decorators

SomeHats:
Some of our tools (e.g. playwright) support tc39 decorators, but not the legacy typescript decorator syntax we currently use. This hasn't caused us issues because we've avoided using our decorators code in those tools, and because we haven't promoted those decorators to library consumers. 

The main motivation for this change is that I want to use some related code in playwright, but in general the ecosystem is moving more towards the TC39 decorator syntax. This diff updates our decorators to support both (and our transpilation/typechecking to use the TC39 version).

### Change type

- [x] `api`


ds300 (approved):
glad to have this taken care of! Thanks! Just one tiny thing I noticed to fix.

PR #4413, opened 2024-08-23
feat(video): integrate new audio controls with video shapes

mimecuvalo:
This PR is branched off of https://github.com/tldraw/tldraw/pull/4290
and integrates the new controls there with videos, in case we want to have them match.

Screenshot 2024-08-23 at 10 56 51

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Insert a video shape and verify the new media controls appear and function correctly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Updated video shapes to use the new unified media controls.


steveruizok:
Works great! Should get a design pass for consistency (font weights?) and I think the controls should only be displayed in a certain state (hovered after a pause?).

mimecuvalo:
> Should get a design pass for consistency (font weights?)

sounds good 👍 

> and I think the controls should only be displayed in a certain state (hovered after a pause?).

yeah, that's probably right. this diff (just to make it clear) is on top of the Audio diff. would Audio shapes also have the controls only hovered after a pause? (if they have album cover art)

packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx
```diff
@@ -89,6 +89,9 @@ export class VideoShapeUtil extends BaseBoxShapeUtil<TLVideoShape> {
 			}
 		}, [rVideo, prefersReducedMotion])
 
+		const zoom = editor.getZoomLevel()
+		const widthScaled = (shape.props.w * zoom) / 260
```
MitjaBezensek, line 93: Maybe pull 260 out in a const?

MitjaBezensek (approved):
Love that you can play / pause without having to go into edit mode.

I did notice that the icons are quite pixelated on a non-retina screen, but I guess we might update those assets down the line.

CleanShot 2024-08-29 at 09 44 57
CleanShot 2024-08-29 at 09 44 31
CleanShot 2024-08-29 at 09 44 27

mimecuvalo:
> I did notice that the icons are quite pixelated on a non-retina screen, but I guess we might update those assets down the line.

yeah, it's a bit of a shame, sigh. these icons are the radix icons but we do have this problem in general, tracked in this issue https://github.com/tldraw/tldraw/issues/2638

ds300:
Closing due to inactivity

PR #4414, opened 2024-08-23
Simplify getting of the text

MitjaBezensek:
We have a method now for getting the text. This should also make it work for any custom shapes with text.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Make copying for text also work for custom shapes that have text (they need to override the `getText` method in the shape util).


mimecuvalo (approved):
nice 👍

PR #4419, opened 2024-08-23
[experiment]: add captions to media

mimecuvalo:
This adds captions for images and videos. For videos you can double-click but for images you'll have to hit Enter for now.
If we wanted this for real we'd have to figure out the tension between cropping and adding a caption. Also, we'd have to improve the hit targets to be able to re-edit a text label.

Fun though!

https://github.com/user-attachments/assets/4301ea6f-a4af-42fe-bb45-67f499720263

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Test adding/editing/deleting captions.

### Release notes

- Feature: add captions to videos and images.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -1110,6 +1110,12 @@ input,
 	width: 100%;
 }
 
+.tl-shape[data-shape-type='image'] .tl-text-label,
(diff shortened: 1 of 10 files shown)


ds300:
Closing due to inactivity

PR #4424, opened 2024-08-27
Fix exports for dark mode frames and flipped images

SomeHats:
These weren't getting handled correctly before.

### Change type

- [x] `bugfix`

### Release notes

- Flipped images are now respected in exports
- Dark mode frames are exported with the correct label color


apps/examples/e2e/tests/export-snapshots.spec.tsx-snapshots/Export-snapshots-Images-light-1-chromium-linux.png
mimecuvalo, line 1: so good

mimecuvalo (approved):
woo!

Issue #4420, opened 2024-08-25
[Feature]: Add Page System with Notebook-like Backgrounds (Lines, Cubes, Graph)


alpha-alexxx:
### What's the feature?

**Description**:

I would like to request a feature enhancement to tldraw that introduces a "page system" along with customizable backgrounds similar to those found in notebooks. This feature would allow users to create multiple pages within a single document and apply different types of backgrounds to each page.

**Proposed Features:**

1. _Page System_:
Ability to create, navigate, and delete multiple pages within a drawing session.
Each page should be independent but remain part of the same document.
A sidebar or tab system for easy navigation between pages.
Background Options:

2. _Lines_: A background option featuring horizontal lines like those in a ruled notebook.
Cubes: A grid of cubes or squares for drawing 3D shapes or organizing content.
Graph Paper: A grid of smaller squares commonly used for mathematical plotting.
Saving and Loading:

Ensure that the page system and selected background are saved within the tldraw file format.
Users should be able to reopen their work without losing page layouts or background settings.

 **Rationale**:
This feature would significantly enhance the usability of tldraw for users who need structured layouts or multiple pages in their projects, such as educators, students, designers, or engineers. It would add versatility to the tool, making it more suitable for a broader range of tasks, from note-taking to detailed diagramming.

**Additional Context:**
I believe this feature would align well with tldraw's goal of providing a flexible, collaborative drawing environment. It could also attract users who need a lightweight alternative to more complex design tools but still require some structured drawing capabilities.

Thank you for considering this request!

### Where would you like this feature?

The tldraw.com website

### Contact Details

ankithome8@gmail.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ds300:
Hi! :wave: 

Thanks for the suggestion. We do have a page system so the first part of your proposal is taken care of already 👍🏼. We don't have a 'sidebar' or 'tabs' style UI built in, only a dropdown menu, but it should be easy to add any kind of custom page switcher you need.

As for the background types, we don't have that as a first class feature but again it would be easy to add a custom background to implement those things by overriding the `Grid` component, and you can store which 'type' of background each page should have on the page object `meta` property.

I think it's unlikely that we'd add the lines or cubes options to tldraw any time soon, but I can see us maybe adding a 'graph'-style grid rendering option in addition to our 'dots'-style default grid. At the very least I'd be happy to add an example for how to do this to our docs site.

ds300:
[...]

PR #4425, opened 2024-08-27
Add custom grid example

ds300:
#4420 highlighted for me that we don't have an example to demonstrate how to implement a custom grid, and this is definitely a common enough need to warrant an example IMO.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/examples/src/examples/custom-grid/CustomGridExample.tsx
+++ b/apps/examples/src/examples/custom-grid/CustomGridExample.tsx
@@ -1,0 +1,108 @@
+import { useLayoutEffect, useRef } from 'react'
+import { TLComponents, Tldraw, approximately, useEditor, useIsDarkMode, useValue } from 'tldraw'
+import 'tldraw/tldraw.css'
+
+/**
+ * There's a guide at the bottom of this file!
+ */
+
+const components: TLComponents = {
+	// [1]
+	Grid: ({ size, ...camera }) => {
+		const editor = useEditor()
+
+		// [2]
+		const screenBounds = useValue('screenBounds', () => editor.getViewportScreenBounds(), [])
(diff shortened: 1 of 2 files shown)


apps/examples/src/examples/custom-grid/CustomGridExample.tsx
```diff
@@ -0,0 +1,108 @@
...
+		const editor = useEditor()
+
+		// [2]
+		const screenBounds = useValue('screenBounds', () => editor.getViewportScreenBounds(), [])
+		const devicePixelRatio = useValue('dpr', () => editor.getInstanceState().devicePixelRatio, [])
+		const isDarkMode = useIsDarkMode()
+
+		const canvas = useRef<HTMLCanvasElement>(null)
+
+		useLayoutEffect(() => {
```
mimecuvalo, line 21: is `useLayoutEffect` necessary here? it seems like `useEffect` would suffice but lemme know if i'm missing something

mimecuvalo (approved):
LG! pre-approving modulo one comment about `useLayoutEffect` before landing though

apps/examples/src/examples/custom-grid/CustomGridExample.tsx
ds300: i think it is yeah. With `useEffect` the grid would lag behind the other canvas elements by one frame?

mimecuvalo: I think that might be ok btw to be one frame behind. Esp. for slower machines, it makes me wonder if we should be encouraging this pattern of strong consistency vs. something that might make the rendering slower potentially.

ds300: nah i think it's fine. drawing a few thousand lines on a canvas is so fast it's negligible.

PR #4428, opened 2024-08-28
[wip] Command bar

MitjaBezensek:
Adds a commandbar when pressing `ctlr/cmd+k`.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/packages/tldraw/src/lib/ui/components/ActionsMenu/DefaultActionsMenuContent.tsx
+++ b/packages/tldraw/src/lib/ui/components/ActionsMenu/DefaultActionsMenuContent.tsx
@@ -1,14 +1,6 @@
-import { useEditor, useValue } from '@tldraw/editor'
 import { PORTRAIT_BREAKPOINT } from '../../constants'
 import { useBreakpoint } from '../../context/breakpoints'
-import {
-	useAllowGroup,
-	useAllowUngroup,
-	useHasLinkShapeSelected,
-	useIsInSelectState,
-	useThreeStackableItems,
-	useUnlockedSelectedShapesCount,
-} from '../../hooks/menu-hooks'
+import { useAllowGroup, useAllowUngroup } from '../../hooks/menu-hooks'
 import { TldrawUiMenuActionItem } from '../primitives/menus/TldrawUiMenuActionItem'
 
 /** @public @react */
@@ -29,64 +21,48 @@ export function DefaultActionsMenuContent() {
 
 /** @public @react */
 export function AlignMenuItems() {
-	const twoSelected = useUnlockedSelectedShapesCount(2)
-	const isInSelectState = useIsInSelectState()
-	const enabled = twoSelected && isInSelectState
-
 	return (
 		<>
-			<TldrawUiMenuActionItem actionId="align-left" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-center-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-right" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="stretch-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-top" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-center-vertical" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-bottom" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="stretch-vertical" disabled={!enabled} />
+			<TldrawUiMenuActionItem actionId="align-left" />
+			<TldrawUiMenuActionItem actionId="align-center-horizontal" />
+			<TldrawUiMenuActionItem actionId="align-right" />
+			<TldrawUiMenuActionItem actionId="stretch-horizontal" />
+			<TldrawUiMenuActionItem actionId="align-top" />
+			<TldrawUiMenuActionItem actionId="align-center-vertical" />
+			<TldrawUiMenuActionItem actionId="align-bottom" />
+			<TldrawUiMenuActionItem actionId="stretch-vertical" />
 		</>
 	)
 }
 
 /** @public @react */
 export function DistributeMenuItems() {
-	const threeSelected = useUnlockedSelectedShapesCount(3)
-	const isInSelectState = useIsInSelectState()
-	const enabled = threeSelected && isInSelectState
-
 	return (
 		<>
-			<TldrawUiMenuActionItem actionId="distribute-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="distribute-vertical" disabled={!enabled} />
+			<TldrawUiMenuActionItem actionId="distribute-horizontal" />
+			<TldrawUiMenuActionItem actionId="distribute-vertical" />
 		</>
 	)
 }
 
 /** @public @react */
 export function StackMenuItems() {
-	const threeStackableItems = useThreeStackableItems()
-	const isInSelectState = useIsInSelectState()
-	const enabled = threeStackableItems && isInSelectState
-
 	return (
 		<>
-			<TldrawUiMenuActionItem actionId="stack-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="stack-vertical" disabled={!enabled} />
+			<TldrawUiMenuActionItem actionId="stack-horizontal" />
+			<TldrawUiMenuActionItem actionId="stack-vertical" />
 		</>
 	)
 }
 
 /** @public @react */
 export function ReorderMenuItems() {
-	const oneSelected = useUnlockedSelectedShapesCount(1)
-	const isInSelectState = useIsInSelectState()
-	const enabled = oneSelected && isInSelectState
-
 	return (
 		<>
-			<TldrawUiMenuActionItem actionId="send-to-back" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="send-backward" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="bring-forward" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="bring-to-front" disabled={!enabled} />
+			<TldrawUiMenuActionItem actionId="send-to-back" />
+			<TldrawUiMenuActionItem actionId="send-backward" />
+			<TldrawUiMenuActionItem actionId="bring-forward" />
+			<TldrawUiMenuActionItem actionId="bring-to-front" />
 		</>
 	)
 }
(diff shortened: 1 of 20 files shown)


mimecuvalo:
🤩

packages/tldraw/src/lib/ui/components/primitives/menus/TldrawUiMenuActionItem.tsx
```diff
@@ -11,5 +11,5 @@ export function TldrawUiMenuActionItem({ actionId = '', ...rest }: TLUiMenuActio
 	const actions = useActions()
 	const action = actions[actionId]
 	if (!action) return null
-	return <TldrawUiMenuItem {...action} {...rest} />
+	return <TldrawUiMenuItem {...action} disabled={!action.enabled?.()} {...rest} />
```
mimecuvalo, line 14: i think this logic should be:
```
const enabled = !('enabled' in action) || action.enabled()
```

feels like if `enabled()` is missing, we shouldn't be punishing previous versions of the library, to make this more backwards-compatible.

(right now, as it stands, if the function is missing, it'll disable everything)

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useTranslation } from './useTranslation/useTranslation'
+
+/** @public */
+export function useCommandBarActions(search: string) {
+	const actions = useActions()
+	const msg = useTranslation()
+	return Object.values(actions)
+		.filter((action) => {
+			// Don't show actions until we start searching
+			if (search === '') return false
```
mimecuvalo, line 12: i think part of the benefit of a cmd-k palette is to show popular actions at the top to spur ideas. it's not meant to be really like a cmd-space (spotlight on mac) type of feeling. i think typically cmd+k should show some default actions but happy to debate this!

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
```diff
@@ -0,0 +1,32 @@
...
+			// If we didn't want to show the disabled actions
+			// if (!action.enabled?.()) return false
+			const unwrapped = unwrapLabel(action.label, 'command-bar')
+			const value = msg(unwrapped as TLUiTranslationKey)
+			if (!value) return false
+			// Alternative to the search check above we can show all actions if we haven't started typing yet
+			// if (search === '') return true
+			return value.toLowerCase().includes(search.toLowerCase())
```
mimecuvalo, line 20: should we do a fuzzy search or this substring search?

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
```diff
@@ -0,0 +1,32 @@
...
+		})
+		.sort((a, b) => {
+			// Enabled actions before disabled ones
+			if (!a.enabled?.()) {
```
mimecuvalo, line 24: same comment here about the `enabled` function being present or not

packages/tldraw/src/lib/ui/hooks/menu-hooks.ts
```diff
@@ -217,3 +223,137 @@ export function useCanUndo() {
...
+				return false
+			}
+			return true
+		},
+		[editor]
+	)
+}
+
+/** @public */
+export function useCanRemoveFrame() {
```
mimecuvalo, line 286: some of these feels maybe a little too specific for a particular action to be in this general file...but i don't feel too strongly about it

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1114,6 +1338,9 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					editor.user.updateUserPreferences({ isSnapMode: !editor.user.getIsSnapMode() })
 				},
 				checkbox: true,
+				enabled() {
```
mimecuvalo, line 1341: seeing all the `return true` (and related to my comments about the `enabled` function implying true if missing), makes me think that maybe the function should be optional? on the other hand, it is good to force an author to think about the right value of `enable` should be when writing an action. probably it should be required, but just thinking out loud here.

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -218,13 +289,17 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					trackEvent('export-as', { format: 'json', source })
 					exportAs(ids, 'json', getExportName(editor, defaultDocumentName))
 				},
+				enabled() {
+					return pageHasShapes
```
mimecuvalo, line 293: so, i would have expected maybe in the `onSelect` function that we could get rid of the extra logic. i think with these changes that `onSelect` should correctly assume that it won't be called _unless_ `enabled()` returns true.

otherwise, it feels like we're duplicating a lot of logic can't really take advantage of this new paradigm. wdyt?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -238,13 +313,17 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 						getExportName(editor, defaultDocumentName)
 					)
 				},
+				enabled() {
+					return true
```
mimecuvalo, line 317: doing a random scan here, looking at the `onSelect` fn for this action, shouldn't this be checking `pageHasShapes`?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -253,13 +332,17 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					trackEvent('export-all-as', { format: 'png', source })
 					exportAs(ids, 'png', getExportName(editor, defaultDocumentName))
 				},
+				enabled() {
+					return true
```
mimecuvalo, line 336: actually....same here? 🤔

packages/tldraw/src/lib/ui/components/CommandBar/DefaultCommandBar.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { useCommandBarActions } from '../../hooks/useCommandBarActions'
+import { useMenuIsOpen } from '../../hooks/useMenuIsOpen'
+import { TLUiTranslationKey } from '../../hooks/useTranslation/TLUiTranslationKey'
+import { useTranslation } from '../../hooks/useTranslation/useTranslation'
+import { TldrawUiButton } from '../primitives/Button/TldrawUiButton'
+import { TldrawUiButtonLabel } from '../primitives/Button/TldrawUiButtonLabel'
+import { TldrawUiKbd } from '../primitives/TldrawUiKbd'
+
+export const COMMAND_BAR_ID = 'command bar'
+const MAX_ITEMS = 6
```
mimecuvalo, line 14: can the command bar just be scrollable? feels maybe limiting to just have it be the top 6 items

packages/tldraw/src/lib/ui/components/CommandBar/DefaultCommandBar.tsx
```diff
@@ -0,0 +1,223 @@
...
+		editor.deleteOpenMenu(COMMAND_BAR_ID)
+	}, [editor])
+
+	const shouldShowPreviousActions =
+		!showPreviousActions && search === '' && previousActionIds.length > 0
+
+	const onSelect = useCallback(
+		(index: number) => {
+			const action = actions[index]
+			if (!action || !action.enabled?.()) return
```
mimecuvalo, line 55: same comment here about the `enabled()` fn being present or not

mimecuvalo (commented):
neat stuff! put my initial comments on it, mainly around how `enabled()` works.

to answer your questions in the PR's description:
> We now also show disabled actions. Feels better to me since it might expose some actions that users are not aware of. It might also be confusing if you can't find an action you know exists, but it is currently disabled.

i think this is ok, but they should be sorted to the bottom of the list if disabled (which i think you're doing)

> I added an option to add an optional description for when an action is disabled. Try using align left when you have no shapes selected to see it. The hint is only displayed once you select the action though. It feels better than always showing it. It could add quite a few new translation strings though (one for each action that can be disabled).

that's pretty helpful actually. yeah, it'll generate more translations but that makes things a lot more clearer.

 > Do we show any example actions when you open the command bar? I've seen some tools do that so that people can discover new actions. Leaning towards no though.

i'm in the camp of showing defaults! added a comment about this below.

> I also added history of actions. You can press up or down arrow when you don't have any search term and it will show the previously used actions, ordered by the time of last use. Feels nice 🤷

neat, might be good to keep this list limited to maybe 3 max? this would also help the argument of showing something by default. it's a little hidden at the moment.

> There's a slight issue with the enabled function. 

i think i understand, maybe show me this offilne?

packages/tldraw/src/lib/ui/components/ActionsMenu/DefaultActionsMenuContent.tsx
```diff
@@ -29,64 +21,48 @@ export function DefaultActionsMenuContent() {
...
 		<>
-			<TldrawUiMenuActionItem actionId="align-left" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-center-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-right" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="stretch-horizontal" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-top" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-center-vertical" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="align-bottom" disabled={!enabled} />
-			<TldrawUiMenuActionItem actionId="stretch-vertical" disabled={!enabled} />
+			<TldrawUiMenuActionItem actionId="align-left" />
```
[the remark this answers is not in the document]
MitjaBezensek: Yeah, I like this part a lot as well, makes it so much cleaner.

packages/tldraw/src/lib/ui/components/primitives/menus/TldrawUiMenuActionItem.tsx
MitjaBezensek: Not sure about this one tbh. We then might show some actions as enabled, but when you select them nothing happens. Might prefer we go the opt in route 🤷

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
MitjaBezensek: Yeah. One idea I had was that we could show actions that users never used before. That said we'd need to store which ones they did use to implement that. Which is kinda problematic since we don't have auth 😄 

Another option would be to show the ones that the users are most likely to use. If you have a group selected there's a high chance you might want to ungroup it,...

Or just a predetermined set of actions we want to expose. Could be a good way of making people aware of new features if we always added new actions to the list.

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
MitjaBezensek: Yeah, I explored this a bit. Most fuzzy search libraries are quite big, but I guess we could go with something really simple.

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: Was wondering the same thing. That said the end users can use the actions without checking if it is enabled or not, so I guess it's better if we are a bit more defensive about applying these actions. Maybe `onSelect` could somehow call `enabled` though.

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: I converted everything 1:1 for now. So if an action had some restrictions before then I kept them, but if I didn't then I didn't add new ones. In this particular case we just don't do anything if we don't have any shapes.

Seems quite related to the discussion of whether we can move some of the restrictions that we currently have in `onSelect` to `enabled`.

packages/tldraw/src/lib/ui/components/CommandBar/DefaultCommandBar.tsx
MitjaBezensek: I guess so, but didn't really feel the need for it. Once you enter a few chars you quite quickly narrow the actions down to less than 6. I guess it might be more useful if we added fuzzy search which would match more actions.

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
mimecuvalo: we can make it smarter over time but having just a predetermined set to start with might be good. even just using our vercel analytics on highest used actions would be fine place to start.

packages/tldraw/src/lib/ui/hooks/useCommandBarActions.ts
mimecuvalo: yeah that sounds good. something for a future iteration sounds like a good plan 👍

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: i hesitate to say this...but should these be converted into classes (from being an object) so that they can call `enabled()`? 😬 

at the very least, even if we're not calling `enabled()` directly, can `onSelect()` just at least do:
`if (!pageHasShapes) return`

ds300:
Very sadly closing this due to inactivity, hope it comes back soon

https://github.com/tldraw/tldraw/pull/4428

mimecuvalo:
:sad-trombone: 
i feel like this might need to come back for a11y work btw at some point...

PR #4435, opened 2024-08-29
Allow publishing PR branches to npm for testing

SomeHats:
If you add the `publish-packages` label to a PR, we'll publish it to npm under the `internal` release tag.

### Change type

- [x] `other`

### DIFF
--- a/.github/workflows/publish-branch.yml
+++ b/.github/workflows/publish-branch.yml
@@ -1,0 +1,54 @@
+name: Publish branch packages
+
+on:
+  pull_request:
+    types: [labeled]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  deploy:
+    name: 'Publish branch packages'
+    environment: npm deploy
+    timeout-minutes: 60
+    runs-on: ubuntu-latest-16-cores-open
+    if: github.event.label.name == 'publish-packages'
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Remove the update-snapshots label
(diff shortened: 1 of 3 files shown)


scripts/publish-prerelease.ts
```diff
@@ -22,7 +23,15 @@ async function main() {
...
 
-main()
+const args = process.argv.slice(2)
+if (args.length === 0) {
+	throw new Error('Usage: publish-prerelease.ts <releaseTag>')
+}
+const releaseTag = args[0]
+if (releaseTag !== 'canary' && releaseTag !== 'latest') {
+	throw new Error('releaseTag must be "canary" or "latest"')
+}
```
ds300, line 36: won't this fail when you invoke it with `internal`?

.github/workflows/publish-branch.yml
```diff
@@ -0,0 +1,54 @@
...
+    environment: npm deploy
+    timeout-minutes: 60
+    runs-on: ubuntu-latest-16-cores-open
+    if: github.event.label.name == 'publish-packages'
+
+    permissions:
+      pull-requests: write
+
+    steps:
+      - name: Remove the update-snapshots label
```
ds300, line 23:
```suggestion
      - name: Remove the publish-packages label
```

ds300 (approved):
nice!

scripts/publish-prerelease.ts
SomeHats: yes, oops! i blame the robbot. thanks for pointing this out!

PR #4442, opened 2024-09-02
Rename TLSvgOptions

### LINKED ISSUE #4439: [Feature]: opt arg in exportToBlob() may be ambiguous.
### What's the feature?

I implemented a feature that allows users to export draws as blob & PNG format in absolute position & bound sizes.
Unfortunately ,I felt [opt in exportToBlob()](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/utils/export/export.ts#L140)'s type is ambiguous.
Because its type is  [TLSvgOptions](https://tldraw.dev/reference/editor/TLSvgOptions), I understood this to be simply about export as SVG format.
Fortunately, [exportAs()](https://tldraw.dev/reference/tldraw/exportAs) has description about opt.
However, I wish there was some mention that this isn't just about SVG.
Maybe it would be a good idea to change the type name. What do you think about this?
I think it would be good to add a related example.
``` typescript 
// ...
  const bounds = new Box(x, y, w, h); // Developers can set the values
  return await exportToBlob({
    editor,
    ids,
    format: "png",
    opts: {
      padding: 0,
      bounds: bounds,
    },
  });
};
```

The developer package

### Contact Details

nayounsang722@gmail.com

### Code of Conduct

ds300:
closes #4439

### Change type

- [x] `api`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Rename `TLSvgOptions` to `TLImageExportOptions`

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -1058,13 +1058,13 @@ export class Editor extends EventEmitter<TLEventMap> {
     getStateDescendant<T extends StateNode>(path: string): T | undefined;
     getStyleForNextShape<T>(style: StyleProp<T>): T;
     // @deprecated (undocumented)
-    getSvg(shapes: TLShape[] | TLShapeId[], opts?: TLSvgOptions): Promise<SVGSVGElement | undefined>;
-    getSvgElement(shapes: TLShape[] | TLShapeId[], opts?: TLSvgOptions): Promise<{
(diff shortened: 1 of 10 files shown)


mimecuvalo (approved):
👍

PR #4446, opened 2024-09-02
[feature] isShapeHidden option

ds300:
This PR adds an option to the Editor that allows people to control the visibility of shapes. This has been requested a couple of times for different use-cases:

- A layer panel with a visibility toggle per shape
- A kind-of 'private' drawing mode in a multiplayer app.

So to test this feature out I've implemented both of those in minimal ways as examples.

### Change type

- [x] `feature`

### Test plan

- [x] Unit tests

### Release notes

- Adds an `isShapeHidden` option, which allows you to provide custom logic to decide whether or not a shape should be shown on the canvas.


apps/examples/src/examples/layer-panel/ShapeList.tsx
```diff
@@ -0,0 +1,149 @@
...
+				>
+					{isEditingName ? (
+						<input
+							autoFocus
+							className="shape-name-input"
+							defaultValue={shapeName}
+							onBlur={() => setIsEditingName(false)}
+							onChange={(ev) => {
+								if (shape.type === 'frame') {
+									editor.updateShape({ ...shape, props: { ...shape.props, name: ev.target.value } })
```
fredrivett, line 75: do we need to spread `shape.props` here? thought we could just pass in a partial of what we want and it would only update that, so to avoid race conditions spreading should be avoided, but you folks would know far better on that!

apps/examples/src/examples/layer-panel/ShapeList.tsx
```diff
@@ -0,0 +1,149 @@
...
+								} else {
+									editor.updateShape({ ...shape, meta: { ...shape.meta, name: ev.target.value } })
```
fredrivett, line 77: same q here

apps/examples/src/examples/layer-panel/ShapeList.tsx
```diff
@@ -0,0 +1,149 @@
...
+								}
+							}}
+						/>
+					) : (
+						<div className="shape-name">{shapeName}</div>
+					)}
+					<button
+						className="shape-visibility-toggle"
+						onClick={(ev) => {
+							editor.updateShape({ ...shape, meta: { ...shape.meta, hidden: !shape.meta.hidden } })
```
fredrivett, line 93: same q here

apps/examples/src/examples/sync-private-content/SyncPrivateContentExample.tsx
```diff
@@ -0,0 +1,146 @@
...
+			() =>
+				editor
+					.getSelectedShapes()
+					.filter((shape) => !!shape.meta.private && shape.meta.ownerId === userId),
+			[editor, userId]
+		)
+
+		// [3]
+		const isPrivateDrawingMode$ = React.useContext(PrivateModeContext)!
+		const isPrivateDrawingMode = useValue(isPrivateDrawingMode$)
```
fredrivett, line 41: realise the meta is named just `private` which is great, but I'd be tempted to rename this to something more generic, like `PrivateEditingMode` or simple `isPrivateMode` as users aren't always drawing when wanting to work in private mode (can see examples of users editing docs privately before sharing with this sort of functionality at lex).

apps/examples/src/examples/sync-private-content/SyncPrivateContentExample.tsx
```diff
@@ -0,0 +1,146 @@
...
+					return !!shape.meta.private && shape.meta.ownerId !== userId
+				}}
+				onMount={(editor) => {
+					// [5]
+					editor.store.sideEffects.registerBeforeCreateHandler('shape', (shape) => {
+						if ('private' in shape.meta) return shape
+						return {
+							...shape,
+							meta: {
+								...shape.meta,
```
fredrivett, line 103: same q re: spreading `shape.meta` here

apps/docs/content/docs/editor.mdx
```diff
@@ -694,6 +694,36 @@ editor.run(
...
+- They will not be shown on the canvas
+- They will not be included in hit test results via `editor.getShapeAtPoint` and `editor.getShapesAtPoint`
+
+Otherwise, they will behave as normal. If you need to restrict certain interactions with hidden shapes, e.g. preventing the user from having a hidden shape selected, you need to handle that in your own code.
+
+e.g.
+
+```tsx
+<Tldraw
+	isShapeHidden={(shape) => shape.meta.hidden}
```
fredrivett, line 712: is the idea here that users may want to configure how the shape is hidden differently, so could pass `shape.meta.myHiddenProperty` and then tldraw would use that to indicate if the shape is hidden?

if so that's a nice addition, I wonder if it should default to this value so it doesn't need passing, though can see how that might conflict with existing users use of a `meta.hidden` value so this might be needed to be set explicitly.

apps/docs/content/docs/editor.mdx
```diff
@@ -694,6 +694,36 @@ editor.run(
...
+	onMount={(editor) => {
+		return react('clean up selection', () => {
+			const selectedShapeIds = editor.getSelectedShapeIds()
+			const filteredSelectedShapeIds = selectedShapeIds.filter((id) => !editor.isShapeHidden(id))
+			if (selectedShapeIds.length !== filteredSelectedShapeIds.length) {
+				editor.setSelectedShapes(filteredSelectedShapeIds)
+			}
+		})
```
fredrivett, line 720: intrigued on this part too, to me this feels more like the inner workings of tldraw rather than something the user should be controlling, is this something that could be included as part of mounting by default without needing the user to specify? I'd imagine so, as `isShapeHidden` should always return false if the user hasn't set up the `isShapeHidden` function on `Tldraw`.

if the user needs to specify it it might be worth adding a comment to clarify exactly what's going on.

apps/docs/content/docs/editor.mdx
```diff
@@ -694,6 +694,36 @@ editor.run(
...
+### Hiding shapes
+
+You can pass a `isShapeHidden` predicate when configuring tldraw. This predicate will be called for each shape to determine if it should be hidden.
+
+Hidden shapes will still be present in the store, but
+
+- They will not be shown on the canvas
+- They will not be included in hit test results via `editor.getShapeAtPoint` and `editor.getShapesAtPoint`
+
+Otherwise, they will behave as normal. If you need to restrict certain interactions with hidden shapes, e.g. preventing the user from having a hidden shape selected, you need to handle that in your own code.
```
fredrivett, line 706: this is interesting to think through, I would assume `getShapeAtPoint` and `getShapesAtPoint` would stop the shape from being selectable, but I would assume that if you selected a parent that is visible it would move the inner shape despite it being hidden (this seems to be the case in the example so that's good!).

also would assume bindings would move with them, but i think that comes with the meal as bindings inherently work that way, but thought i'd just double check.

other than that, is this note basically stating that another user might *already* have the shape selected so it could feasibly still be moved etc on its own, and if we want to disable that we should handle it? or is this saying that by default shapes will be selectable despite being hidden (if so, via which method would that be so?).

fredrivett (commented):
only quickly glanced over this before shooting out and some of my comments may show my lack of tldraw expertise, but thought best to look silly and double check stuff than attempt to avoid so and miss flagging something useful.

happy to re-review if there's any changes or when i've got more time later on.

apps/examples/src/examples/layer-panel/ShapeList.tsx
ds300: yes you are correct! Supplying a partial is definitely preferred. I'll fix.

apps/examples/src/examples/layer-panel/ShapeList.tsx
ds300: yep meta also supports partials, but only for bindings and shapes it seems.

apps/examples/src/examples/sync-private-content/SyncPrivateContentExample.tsx
ds300: here is different. we don't do any post-processing on the return values of store side effects. it's a much lower-level API. so you need to do a spread here.

apps/docs/content/docs/editor.mdx
ds300: The thinking is that the actual logic that defines when a shape should be hidden might be more complex than just a single boolean toggle. e.g. with the 'private mode' example it depends on the owner ID as well. So in the same way that React lets you define your UI as a function of your state, we let you define the 'hiddenness' of shapes as a function of your state, to make it easier to calculate correctly.

I don't think it makes sense for us to have a default implementation of this predicate since we make no prescriptions about what people put in meta properties, and it's very simple to add in userland.

apps/docs/content/docs/editor.mdx
ds300: We can't add this as builtin behaviour because it's not always what you'd want. E.g. with a layer panel it's entirely possible to select hidden shapes in the panel, and to delete them and so forth. We can't anticipate all the ways that the 'hiddenness' of a shape will inform how different interactions work.

Good call on the comment I'll add one.

apps/docs/content/docs/editor.mdx
ds300: > is this note basically stating that another user might already have the shape selected so it could feasibly still be moved etc on its own, and if we want to disable that we should handle it?

Yes.

> or is this saying that by default shapes will be selectable despite being hidden

Also yes. I'll reword to make this clearer.

> (if so, via which method would that be so?)

It mainly depends on the app and how 'hiddenness' is used to implement features. One that applies to all apps would be the `selectAll` feature (cmd+a). This selects all the unlocked shapes (including the hidden ones) because that might be what you want if you're building an app with a layers panel. But you wouldn't want cmd+a to select hidden shapes in the context of a multiplayer app with 'private' shapes.

apps/examples/src/examples/sync-private-content/SyncPrivateContentExample.tsx
fredrivett: ah good to know!

apps/docs/content/docs/editor.mdx
fredrivett: yeah that's fair, makes sense with more context!

apps/docs/content/docs/editor.mdx
fredrivett: ah great, that makes sense! I assume exposing a `hiddenShapesSelectable` or `preventSelectingHiddenShapes` that enables/disables this wouldn't work as it's more complex than just that in some instances?

just making sure the surface level API can't be simplified, all good if there's good reasons.

apps/docs/content/docs/editor.mdx
fredrivett: sweet, I think I'll have a better idea on any more thoughts when implementing on our end.

generally I think most folks will expect a hidden shape to work how it does in figma, so the more that can be baked in the better, but trust y'all to think through the bigger picture as realise you've got more use cases to bear in mind.

jshoung (approved):
Thanks a lot for doing this! I played around with the examples and they work well, the layer panel example is pretty cool too.

apps/docs/content/docs/editor.mdx
```diff
@@ -694,6 +694,42 @@ editor.run(
...
+
+You can prevent that by cleaning up the selection in a signal reaction that is set up in the `onMount` callback:
+
+```tsx
+<Tldraw
+	isShapeHidden={(shape) => shape.meta.hidden}
+	onMount={(editor) => {
+		// We don't prevent hidden shapes from being selected out of the box, because there are some situations where it's desirable.
+		// If you want to prevent hidden shapes from being selected, you can do so like this:
+		return react('clean up selection', () => {
```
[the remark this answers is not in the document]
ds300: It would be simple to add that prop, but to me it feels like a little too much to add it to the top-level tldraw props.

Plus I suspect that there will be unexpected ways in which hidden shapes conflict with other interactions, and this section of the docs pushes people towards figuring those things out on their own.

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -181,6 +181,15 @@ export interface TldrawEditorBaseProps {
...
 	deepLinks?: true | TLDeepLinkOptions
+
+	/**
+	 * Predicate for whether or not a shape should be hidden.
+	 *
+	 * Hidden shapes will not render in the editor, and they will not be eligible for hit test via
+	 * {@link Editor#getShapeAtPoint} and {@link Editor#getShapesAtPoint}. But otherwise they will
+	 * remain in the store and participate in all other operations.
+	 */
+	isShapeHidden?: TLEditorOptions['isShapeHidden']
```
SomeHats, line 192: we'll get better docs here if we inline the method type instead of referencing it like this

SomeHats (approved):
love it ty

PR #4448, opened 2024-09-03
Rename manually triggered actions

ds300:
Renames some github actions to make more sense

### Change type

- [x] `other`


mimecuvalo:
thank you!

Issue #4358, opened 2024-08-06
Double clicking a gif pauses AND enters cropping


TodePond:
1. Add a gif to the board.
2. Double click it.

Expected: It pauses it.
Actual: It pauses AND enters cropping state.

ds300:
what's the action to take here? disable cropping for gifs?

TodePond:
I think so!

PR #4451, opened 2024-09-03
images: dont stop playing a gif on double click

mimecuvalo:
Double-clicking currently pauses gif's which is odd. This changes is so that you can unpause them (if they were stopped for reduced motion reasons). But they can't be paused.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Images: dbl-clicking doesn't stop gifs


ds300:
related: #4358

steveruizok (approved):

steveruizok:
Works for me. We could probably cut deeper and remove the rest of the double click functionality? Or are you planning to pull that in a different PR?

mimecuvalo:
> We could probably cut deeper and remove the rest of the double click functionality? Or are you planning to pull that in a different PR?

I think we still need a way to play the gif though if it's initially paused (thru the reduced-motion setting).
Or do you mean removing the cropping behavior for gifs? I know in https://github.com/tldraw/tldraw/issues/4358 it was maybe suggesting that we drop cropping for gifs but I'm not entirely convinced.

steveruizok:
Easier done than said, so I've pushed a change that removes the "double click to toggle isPlaying".

PR #4454, opened 2024-09-04
Add sleep fn

SomeHats:
There are a bunch of places we do something like `await new Promise(resolve => setTimeout(resolve, 1000))`. Most of the time, these places have to come with an eslint disable. We have a rule enforcing that timers come from `editor.timers`, but in those cases this won't do what you would expect: when the editor is disposed, those timers will never resolve, which isn't really what you want if you're using the timers for control-flow around more expensive resources that need to be disposed of.

For example:
```tsx
const thing = createAnExpensiveResourceThatNeedsToBeExplicitlyDestroyed()
try {
    await doSomethingWithIt(thing)
    await new Promise(r => editor.timers.setTimeout(r, 100))
} finally {
    thing.dispose()
}
```

In this example, if the editor is disposed, then the `setTimeout` callback is never called, and the promise never resolves. The code in the finally block never runs, and the memory for `thing` leaks. 

### Change type

- [x] `other`

### Release notes

(internal-only change)

### DIFF
--- a/apps/docs/components/docs/docs-feedback-widget.tsx
+++ b/apps/docs/components/docs/docs-feedback-widget.tsx
@@ -5,7 +5,7 @@ import { useLocalStorageState } from '@/utils/storage'
 import { Field, Label, Textarea } from '@headlessui/react'
 import { ArrowLongRightIcon, CheckCircleIcon, HandThumbDownIcon } from '@heroicons/react/20/solid'
 import { HandThumbUpIcon } from '@heroicons/react/24/solid'
-import { track } from '@vercel/analytics'
+import { sleep } from '@tldraw/utils'
 import { usePathname } from 'next/navigation'
 import { FormEventHandler, useCallback, useState } from 'react'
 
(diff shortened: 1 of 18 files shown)


apps/docs/utils/ContentVectorDatabase.ts
```diff
@@ -139,7 +140,7 @@ export class ContentVectorDatabase {
 		}
 
 		// Sleep for 50ms or so to avoid rate limiting
-		await new Promise((r) => setTimeout(r, 35))
+		await sleep(35)
```
mimecuvalo, line 143: can you update the comment here? doesn't match

apps/docs/components/docs/docs-feedback-widget.tsx
```diff
@@ -17,9 +17,8 @@ async function submitFeedback(
 	_feedback: 1 | -1 | 0,
 	_note: string
 ) {
-	const feedback: { value: number; message?: string } = { value: _feedback }
-	if (_note !== '') feedback.message = _note
-	track('Feedback', feedback)
+	// todo
+	await sleep(2000)
```
mimecuvalo, line 21: ?

mimecuvalo (approved):
nice one! pre-approving modulo the couple comments (the feedback-widget in particular)

apps/docs/components/docs/docs-feedback-widget.tsx
SomeHats: oh good catch, this was a bad merge, sorry!

PR #4455, opened 2024-09-04
Serve optimised content from CDN

SomeHats:
- Updates our CDN to upload the optimized versions of assets
- Sets an immutable cache-control header on assets as they include the version in the URL and so will never change over time.
- Icons use the merged SVG instead of downloading each one individually

### Change type
- [x] `other`


mimecuvalo:
adding @MitjaBezensek as a reviewer

MitjaBezensek (approved):
Nice! Thought that we had some caching set up on the R2 bucket, but apparently not.

scripts/lib/upload-static-assets.ts
```diff
@@ -48,14 +53,22 @@ async function uploadDirectory(prefix: string, directoryPath: string) {
 
 export async function uploadStaticAssets(version: string) {
 	try {
+		await exec('yarn', ['refresh-assets'], { env: { ALLOW_REFRESH_ASSETS_CHANGES: '1' } })
```
mimecuvalo, line 56: `setAllVersions` _also_ does:
`await exec('yarn', ['refresh-assets', '--force'], { env: { ALLOW_REFRESH_ASSETS_CHANGES: '1' } })`

so it looks like you'll be doing this twice each time

SomeHats: this is mostly here because it does need to have run before we run any of the other code here. it's cached by `lazy` though, so won't run when we call it from a context where it's already done

mimecuvalo: mm, but setAllVersions uses `--force` to skip `lazy`, right?

MitjaBezensek: `setAllVersions` runs first, so it's fine if it's `--force`. This one in `uploadStaticAssets` runs after that and isn't `--force` so it should use the cache.

MitjaBezensek: PS: this all takes 2s to run, so not really an issue even if it did run twice in some cases
CleanShot 2024-09-06 at 10 11 27

PR #4460, opened 2024-09-05
Store bookmark images on our own asset server

SomeHats:
Sometimes, if you open an old room, you'll find that the bookmark image isn't working. This is often because the website the bookmark is pointing to has changed, and the image asset that we fetched when you first loaded the image isn't there any more. 

For our new exports with bookmarks, there's also an issue: we can only embed images that have CORS headers set. Many websites do set these headers on their open graph etc. images, but not all do. For example, the new dynamicland website looks like this when bookmarked: 
Screenshot 2024-09-05 at 16 51 49

This diff has our bookmark unfurler also save the OG images to our own servers. This is similar to how social media sites etc. store these images for display.

### Change type

- [x] `other`


apps/dotcom-worker/src/routes/extractBookmarkMetadata.tsx
```diff
@@ -0,0 +1,30 @@
+import { assert } from '@tldraw/utils'
+import { handleExtractBookmarkMetadataRequest } from '@tldraw/worker-shared'
+import { IRequest } from 'itty-router'
+import { Environment } from '../types'
+
+export async function extractBookmarkMetadata(request: IRequest, env: Environment) {
+	if (request.method === 'GET') {
+		// legacy route: extract metadata without saving image
+		return handleExtractBookmarkMetadataRequest({ request })
+	}
```
ds300, line 10: why not just keep only the GET route and do the uploading there?

ds300 (approved):

SomeHats:
> * should this bucket be purged of items after X days? the bucket probably won't get that big but is it reasonable expectation to have these hosted forever?

We currently store assets forever. I wouldn't want to add any purging for to these images unless we also had purging for rooms and their associated assets too. It would be weird for your bookmark images to suddenly disappear after a while.

> * _can_ we copy these over? are there copyright issues to hosting them ourselves? (instead of just hotlinking to them)

There's not much published on this, but as far as I can tell, there shouldn't be any issues here - or at least no more issue than we already have from allowing use-uploaded content. This was informative: https://webmasters.stackexchange.com/questions/125768/opengraph-image-and-license 

Essentially, there's a chance that with these images (or any other images/videos that people might upload to tldraw.com) that we may receive a DMCR notice to take down content. If/when that happens, we take down the content.

> * should it really be creating new images (instead of overriding) for the same URL? there'd prbly be a security issue if we overrode (i could imagine) so the answer is probably no, but just bringing up the discussion.

I would like to create new images because these resources may change over time, but in general i would like to avoid the contents of the board changing visually over time.

> * can we potentially create on-demand only for export cases?

We could do this, but I think it would be more complicated than what we have here

apps/dotcom-worker/src/routes/extractBookmarkMetadata.tsx
SomeHats: discussed offline

mimecuvalo:
> We currently store assets forever. I wouldn't want to add any purging for to these images unless we also had purging for rooms and their associated assets too. It would be weird for your bookmark images to suddenly disappear after a while.

To be clear, I wouldn't want the images to disappear either, per se. If we didn't have an expectation of image persistence (see discussion below), I think if an image is missing from our server, it would go refetch from the original url and repopulate.

> I would like to create new images because these resources may change over time, but in general i would like to avoid the contents of the board changing visually over time.

I could see that — I don't necessarily disagree. It'd be interesting to see how Facebook/Twitter treat this link as maybe a baseline of expectations:

   https://en.wikipedia.org

Today's og:image is a picture of "KNXV-TV studios in Phoenix" because the featured image changes on wikipedia from day-to-day. I wonder what the big websites do when the picture changes the next day? Might be worth comparing.

I don't think my comments block this PR — if there's followups here (vis-a-vis the image persistence), we can followup later.

PR #4461, opened 2024-09-06
csp: turn on; also, add data: for connect-src

mimecuvalo:
Turn on CSP for dotcom. Also, I found another case where `connect-src` needs to have `data:` added so that we can paste those types of images from other sites.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


SomeHats (approved):
it's great we're finally getting this enabled! amazing work :D

steveruizok:
Looks good, let's land this and get it out along with the watermark.

PR #4463, opened 2024-09-06
assets: no need for the setTimeout 0 anymore after deep links

mimecuvalo:
After David's Deep links PR https://github.com/tldraw/tldraw/pull/4333 the camera doesn't get set twice on page load (first at 1, and then at the actual zoom specified by the url). Because of this, we can drop this setTimeout 0 which was addressing that issue.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


SomeHats (approved):
nice!

PR #4468, opened 2024-09-09
consistent function style

SomeHats:
`const doSomething = () => {...}` won't produce as nice documentation as `function doSomething() {}`. This diff enforces function declaration style when function are written in export position.

### Change type

- [x] `other`


mimecuvalo:
✨

PR #4469, opened 2024-09-09
File save warning

steveruizok:
This PR adds a one-time warning about files in tldraw. It's only shown when the user loads a project with >500 records in the store. Hopefully this cuts down on the number of people who accidentally delete their local project. It is only shown once and will be dismissed once closed.

image

We could also pair this with a "local" chip in the top zone that pops up a screen like this.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Adds save warning.


mimecuvalo:
This is a good idea! Is 500 the right threshold? Why not 100?

steveruizok:
Just picked a number, something that seemed like the project would hurt to lose. I'll go with 128, another kind of random number, but early enough.

PR #4471, opened 2024-09-09
seo: add canonical url for rooms

mimecuvalo:
I need to test this with the Google Search tool but I believe this should consolidate URLs for googlebot. It happens after page load, not on server side, but I think it should still work. 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


ds300 (approved):
sorry, i missed this PR somehow! Great stuff

mimecuvalo:
bah, silly of me. after testing this further, i remembered that we disable indexing of `/r` rooms in `robots.txt` so this is moot.

PR #4475, opened 2024-09-10
Bisect issues using the PR preview links

MitjaBezensek:
Mimics git bisect, but uses the PR preview links so that you don't have to update deps, deal with migrations, and rebuilding when bisecting.

### How to use

1. Run `yarn dev-tools`. This will get the last 100 PR numbers, which you can then bisect over. If you need more PRs that that you can change it in `scripts/get-pr-numbers.ts` script.
2. Navigate to http://localhost:5420
3. Click `Start bisect`
4. Mark one of the PRs as good and one as bad. Use the preview links to open the preview deployment of the app at that specific PR.
5. Once you have both a good and a bad PR selected you will see the current PR marked in blue. Test it out and mark it accordingly. Continue doing so until the bisect finishes.

In some cases the preview builds are not available (PRs from the community might not have permissions to build them). In that case you can use the `Show all buttons` toggle and use a preview build close to that PR.

You can also use the `Automatically open preview links` to immediately open the next preview link after you mark the current PR. Helps with speeding things along.

### Limitations
Multiplayer functionality of preview links is removed after two days, so this is mostly useful for core editor functionality.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Adds biscecting on top of preview links to help us find PRs that broke some functionality.


apps/examples/src/misc/Bisect/pr-numbers.ts
```diff
@@ -0,0 +1,5 @@
+// This file is automatically generated by scripts/get-pr-numbers.ts.
+// Do not edit manually. Or do, I'm a comment, not a cop.
+
+/** @public */
+export const prNumbers = []
```
MitjaBezensek, line 5: Committing this with an empty array as it will be generated while running bisect. We probably need to have this one committed in, otherwise you might get compile errors until you run bisect for the first time.

mimecuvalo (commented):
i think this is fucking great. such a time saver and it makes me wish i had thought of it :)

my high-level feedback is: we should create a new app, otherwise this feels shoe-horned into the Examples app. we can create a simple `apps/devtools` Vite app, doesn't require much to stand up.

creating a Devtools app can be useful because we can add even more dev tools (maybe some SVG tools? license tools? .tldraw file debug tools?) or other things that can help those of us that work on tldraw do our job better.

i think if we keep it in Examples, and since that's deployed to examples.tldraw.com, it also feels pretty dirty. it's a particularity of our organization that we're open source - if we weren't open source, adding a devtool into something that runs in production would be a no-no.

so, let's create a new app! i think it could be a great foundation for a lot more innovative tools that folks can add over time!

apps/dev-tools/tsconfig.json
MitjaBezensek, line 1: Should we pull over the config from one of the other projects?

apps/dev-tools/src/App.tsx
MitjaBezensek, line 1: No routing for now, can always later when we'll have other tools.

apps/dev-tools/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
mimecuvalo, line 1: i don't feel strongly on this, but can the root `.gitignore` just apply the ignore rule to subdirectories?

mimecuvalo: (not blocking)

apps/dev-tools/package.json
```diff
@@ -0,0 +1,28 @@
...
+	"description": "tldraw dev tools",
+	"version": "2.0.0-beta.2",
+	"type": "module",
+	"scripts": {
+		"dev": "vite",
+		"build": "tsc -b && vite build",
+		"lint": "yarn run -T tsx ../../scripts/lint.ts",
+		"preview": "vite preview"
+	},
+	"dependencies": {
```
mimecuvalo, line 13: our other deps have:
```
"peerDependencies": {
		"react": "^18",
		"react-dom": "^18"
	},
```

let's make them match - i don't want to maybe have different versions of react

apps/dev-tools/src/App.tsx
```diff
@@ -0,0 +1,7 @@
+import Bisect from './Bisect/Bisect'
+
+function App() {
```
mimecuvalo, line 3: 🙌

apps/dev-tools/vite.config.ts
```diff
@@ -0,0 +1,12 @@
+import react from '@vitejs/plugin-react'
+import path from 'path'
+import { defineConfig } from 'vite'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+	plugins: [react()],
+	root: path.join(__dirname, 'src'),
+	server: {
+		port: 5420,
```
mimecuvalo, line 10: let's put this on a different port. 5420 is the examples app port. we can pick something else!

mimecuvalo (approved):
love it. pre-approving modulo some comments. let's gooooo 🚀

apps/dev-tools/.gitignore
MitjaBezensek: It already probably ignores a bunch of this generic stuff, let me see if I can clean it up (it's the base `.gitignore` from creating a new vite app).

apps/dev-tools/package.json
MitjaBezensek: Using just 18 gives me some yarn constraints issues
CleanShot 2024-09-12 at 12 16 26

We also seem to be using both 18.2 and just 18 all over the place. Will keep it as is for now.
image

apps/dev-tools/.gitignore
MitjaBezensek: Could completely remove it, nice catch 🙌

apps/dev-tools/vite.config.ts
MitjaBezensek: 4000 now.

PR #4480, opened 2024-09-10
[docs] Remove breadcrumbs (for now)

ds300:
These breacrumbs links either didn't have actual content, or the content was at a different route and it was already visible in the site nav. So the breadcrumbs weren't providing much value and it's easier right now to just remove them

- [x] `other`

before: 
image

after:
image


ds300:
breadcrumbs are a bit more valuable on mobile, where the main site nav is less visible, but not enough that it warrants fixing all the issues to make them work properly right now.

MitjaBezensek:
Fixes INT-273.

PR #4483, opened 2024-09-10
[docs] improve code block typography

ds300:
Bumped up the weight, switched to Noto Sans mono, turned off ligatures

Before
image

After
image

Before

image

After

image

just noticing now how incredibly tiny Geist Mono's descenders are. I think that's a big contributor to the illegibility.

### Change type

- [x] `other`


mimecuvalo:
oh i didn't see that you had created this PR before i created this another one https://github.com/tldraw/tldraw/pull/4487
i would agree about the local font!

ds300:
I ditched noto sans mono in favor of system default (specifically, using github's font-family string). Menlo is underrated.

gonna merge this. we can tweak the font family in #4487 if you like @mimecuvalo

mimecuvalo:
> gonna merge this. we can tweak the font family in https://github.com/tldraw/tldraw/pull/4487 if you like @mimecuvalo

sounds good - i'd imagine @steveruizok would want to tweak it so he can take it from here

PR #4487, opened 2024-09-10
docs: revert to system monospace font for code

mimecuvalo:
revert back to what we had before, i would agree that the original font is slightly more readable than the Geist Mono.
@steveruizok you probably have an opinion here :)

Before
Screenshot 2024-09-10 at 18 08 45

After
Screenshot 2024-09-10 at 18 09 21

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
@steveruizok i'm gonna close this as it was tweaked in https://github.com/tldraw/tldraw/pull/4483

PR #4491, opened 2024-09-11
[docs] better icon for 'Reference' nav item

ds300:
This looked a bit too much like play/pause in a media player.

Before
image

After
image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
way better - nice one!

PR #4493, opened 2024-09-11
Make docs site like 69x faster

SomeHats:
Right now, we render the entire docs site dynamically for each request. It's really slow! This is what the timing waterfall looks like for just fetching the initial HTML of the editor reference page:
Screenshot 2024-09-11 at 12 03 12

We don't need to do this, as the content is completely static and never changes between deploys. This diff uses `generateStaticParams` to make sure next pre-renders all of our content. Now, the waterfall graph looks like this:
Screenshot 2024-09-11 at 12 03 21

8.37s -> 121ms is a speed up of about 69x. Nice!

### Change type

- [x] `other`

### DIFF
--- a/apps/docs/app/(docs)/[...slug]/page.tsx
+++ b/apps/docs/app/(docs)/[...slug]/page.tsx
@@ -30,6 +30,14 @@ export async function generateMetadata({
 	return metadata
 }
 
+const nonDocsPaths = ['/blog/', '/legal/']
+export async function generateStaticParams() {
+	const paths = await db.getAllPaths()
+	return paths
+		.filter((path) => !nonDocsPaths.some((nonDocsPath) => path.startsWith(nonDocsPath)))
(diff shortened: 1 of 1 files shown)


apps/docs/app/(docs)/[...slug]/page.tsx
```diff
@@ -23,6 +23,13 @@ export async function generateMetadata({
 	return metadata
 }
 
+export async function generateStaticParams() {
+	const paths = await db.getAllPaths()
+	return paths
+		.filter((path) => !path.startsWith('/blog') && !path.startsWith('/legal/'))
```
Comment (mimecuvalo) [inline apps/docs/app/(docs)/[...slug]/page.tsx:29]: i'm worried we'll forget about this in the future if there's another top-level folder. possible to add something that checks against an array of paths it should ignore?

mimecuvalo (approved):
zomg.

approved modulo one comment

PR #4501, opened 2024-09-12
Use shiki for syntax highlighting in docs

ds300:
highlight.js was failing on some JSX-heavy code blocks. e.g. 
image

I first tried to replace it with PrismJS and ran into weird issues I didn't understand (in hindsight I might know how to troubleshoot them better now)

Then I tried Shiki and also ran into weird issues I didn't understand. Turns out we had this postprocessing step in `code.tsx` to do link injection and line lowlighting. It seemed to be kinda hard-coded to work with the output of highlight.js and was totally ruining the output of Shiki.

So I rewrote it to take advantage of Shiki's simpler output.

This results in things looking the same or even better on full code blocks and valid JS expressions

Before
image

After
image

However shiki is less gung-ho on context-free code snippets that wouldn't be valid in the top level of a file, like class members

Before

image

After

image

but it's still _fine_, I guess? We could maybe add custom styling for links (bold?) to make them pop a bit more in these cases.

EDIT:

Bold looked dreadful, but underline works nicely and also lets us omit the link icon, which was kinda messy in lots of places

Before

image

After

image

### Change type

- [x] `other`


SomeHats:
when i added these links back into the docs site we tried underlines and a couple of different icon options before settling on this one. It's not ideal, and probably less so now that we've lost the highlighting for some of these keywords, but i do find it more legible than underlines which sort of cause my eyes to slide of the text a bit? 

one thing we could try is icon, plus picking a default non-white color for when links aren't explicitly highlighted to help them stand out a bit.

ds300:
@SomeHats yeah agreed, that underline does make the text quite illegible. i tried applying a default color, but it's hard. How about a less intense underline? To my eye this looks pretty good

image
image

apps/docs/components/content/pre.tsx
```diff
@@ -23,7 +23,11 @@ export function Pre(props: DetailedHTMLProps<HTMLAttributes<HTMLPreElement>, HTM
 				// ref={container}
 				className={cn(
 					'bg-zinc-900 text-sm text-white shadow md:rounded-xl overflow-x-auto px-5 md:px-4 py-4 font-medium',
-					'[td_&]:bg-zinc-200 dark:[td_&]:bg-zinc-800 [td_&]:shadow-none [td_&]:rounded-lg [td_&]:p-0 [td_&]:px-1.5'
+					'[td_&]:bg-zinc-200 dark:[td_&]:bg-zinc-800 [td_&]:shadow-none [td_&]:rounded-lg [td_&]:p-0 [td_&]:px-1.5',
+					// this rule enables inverted dark/light mode inside of table cells
+					'[&_*]:text-[--shiki-dark]',
```
mimecuvalo, line 28: i don't love all these `*` rules - if there's a tighter alternative that would be great, but i also won't lose sleep over it :P

apps/docs/components/content/code.tsx
```diff
@@ -31,209 +30,148 @@ export function FocusLines({ children, lines }: { children: ReactNode; lines: nu
...
-		// I don't know why this happens, it feels like a bug. We handle this by keeping track of &
-		// merging consecutive tokens if they're identifiers with no non-identifier tokens in
-		// between.
-
-		// does this token look like a JS identifier?
-		function isIdentifier(token: string): boolean {
-			return /^[_$a-zA-Z\xA0-\uFFFF][_$a-zA-Z0-9\xA0-\uFFFF]*$/g.test(token)
+// split the code into an array of identifiers, and the bits in between them
+function tokenize(code: string): string[] {
+	const identifierRegex = /[_$a-zA-Z\xA0-\uFFFF][_$a-zA-Z0-9\xA0-\uFFFF]*/g
```
mimecuvalo, line 53: i see this predates this PR but can you have the `isIdentifier` fn above use this same regex? so it's not duplicated

mimecuvalo (approved):
neato. couple minor comments but good to go. nice discovery of shiki! hadn't heard of that library before.

apps/docs/components/content/code.tsx
ds300: they're not the same though, `isIdentifier` includes the `^` and `$` sentinels to match a whole string, that wouldn't work here.

apps/docs/components/content/pre.tsx
ds300: it could be `span` instead of `*` i guess

apps/docs/components/content/code.tsx
mimecuvalo: ah! i didn't notice that, my bad. well, they could be composed:

`new RegExp('^' + identifierRegexWithoutSentinels.source + '$')`

PR #4502, opened 2024-09-12
Fix parsing of older tldraw v2 documents

MitjaBezensek:
Parsing of some old files failed due to some record types that are no longer present in our store. 

See PRs that removed them: https://github.com/tldraw/tldraw/pull/1493/ ([user_document](https://github.com/tldraw/tldraw/pull/1493/files#diff-4a4bcc8b6ed0eb5c124a153c3d6e4de7a467ebc0a4d65ff4b18c74dd54fb3dd8R55)) and https://github.com/tldraw/tldraw/pull/1435/ ([user and user_presence](https://github.com/tldraw/tldraw/pull/1435/files#diff-901095d4d90bdb975afedd77a3b456547bb365634d99460d5a1b74d20955ee14R42)).

We could also migrate the records (just like we do a bit lower), but I guess we can't be certain if we even have a v2 document at this point 🤷‍♂️ 

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix a but with opening older v2 tldraw files.

### DIFF
--- a/packages/tldraw/src/lib/utils/tldr/file.ts
+++ b/packages/tldraw/src/lib/utils/tldr/file.ts
@@ -93,6 +93,20 @@ export type TldrawFileParseError =
 	| { type: 'migrationFailed'; reason: MigrationFailureReason }
 	| { type: 'invalidRecords'; cause: unknown }
 
+function parseAndRemoveOldRecordTypes(json: string) {
+	const document = JSON.parse(json)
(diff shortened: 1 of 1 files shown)


ds300 (approved):
yeah this should work. I think this has been broken since my big migrations PR early this year. I'm surprised it's taken so long to hear about it!

PR #4504, opened 2024-09-12
docs: try icons instead of categories for search

mimecuvalo:
screenshots:
Screenshot 2024-09-12 at 15 35 35
Screenshot 2024-09-12 at 15 35 22

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/docs/components/search/SearchAutocomplete.tsx
+++ b/apps/docs/components/search/SearchAutocomplete.tsx
@@ -1,6 +1,8 @@
+import { TldrawIcon } from '@/components/common/icon/tldraw'
 import { SearchEntry } from '@/utils/algolia'
 import { Combobox, ComboboxItem, ComboboxProvider, VisuallyHidden } from '@ariakit/react'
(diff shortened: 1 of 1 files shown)


SomeHats:
Small icons:
Screenshot 2024-09-12 at 16 44 41
Screenshot 2024-09-12 at 16 44 46

Text labels:
Screenshot 2024-09-12 at 16 55 51
Screenshot 2024-09-12 at 16 55 57

To be honest, I don't really like any of these. I think they're quite noisy and I don't think people would understand what the icons mean :/

SomeHats:
different approach here: https://github.com/tldraw/tldraw/pull/4507

mimecuvalo:
> To be honest, I don't really like any of these. I think they're quite noisy and I don't think people would understand what the icons mean :/

same! yeah just to be clear, this was just a test. i wasn't rooting for this solution either.

PR #4506, opened 2024-09-12
Fix a sentry issue with getting arrow bindings

MitjaBezensek:
[Issue](https://tldraw.sentry.io/issues/5396016166/?project=4504203639193600&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&stream_index=1).

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix a sentry issue when getting arrow bindings.

### DIFF
--- a/packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
+++ b/packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
@@ -43,12 +43,16 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
 
 	// when the binding itself changes
 	override onAfterCreate({ binding }: BindingOnCreateOptions<TLArrowBinding>): void {
-		arrowDidUpdate(this.editor, this.editor.getShape(binding.fromId) as TLArrowShape)
+		const arrow = this.editor.getShape(binding.fromId) as TLArrowShape | undefined
+		if (!arrow) return
+		arrowDidUpdate(this.editor, arrow)
 	}
 
 	// when the binding itself changes
 	override onAfterChange({ bindingAfter }: BindingOnChangeOptions<TLArrowBinding>): void {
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
```diff
@@ -43,12 +43,16 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
...
+		const arrow = this.editor.getShape(binding.fromId) as TLArrowShape | undefined
+		if (!arrow) return
+		arrowDidUpdate(this.editor, arrow)
 	}
 
 	// when the binding itself changes
 	override onAfterChange({ bindingAfter }: BindingOnChangeOptions<TLArrowBinding>): void {
-		arrowDidUpdate(this.editor, this.editor.getShape(bindingAfter.fromId) as TLArrowShape)
+		const arrow = this.editor.getShape(bindingAfter.fromId) as TLArrowShape | undefined
+		if (!arrow) return
```
MitjaBezensek, line 54: Not sure if we need to do some cleanup here.

SomeHats: i don't really understand how we end up with a change event for a binding that i think should be deleted :/ 

we can maybe land this, but i wonder if theres a fix further up the chain? i don't think these bindings should exists

SomeHats (approved):

packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
MitjaBezensek: Yeah, I also wasn't sure.

PR #4507, opened 2024-09-12
dynamically re-order search results for nicer-looking groups

SomeHats:
image

This runs through the search results, and tries to merge close-by small groups together. it works ok, and does a much better job at keeping more relevant results near the top of the list than absolute sorting. still some alternating though...

### Change type

- [x] `other`


mimecuvalo:
@SomeHats I figured out how to do this automatically via Algolia: https://github.com/tldraw/tldraw/pull/4516

closing this one up!

PR #4510, opened 2024-09-13
Improve loading of files

MitjaBezensek:
This is a followup to #4502 that makes sure we correctly migrate the data.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improve loading of files.

### DIFF
--- a/packages/tldraw/src/lib/utils/tldr/file.ts
+++ b/packages/tldraw/src/lib/utils/tldr/file.ts
@@ -297,7 +297,6 @@ export async function parseAndLoadDocument(
 	// this file before they'll get their camera etc.
 	// restored. we could change this in the future.
 	transact(() => {
-		console.log('result', parseFileResult.value.getStoreSnapshot())
 		editor.loadSnapshot(parseFileResult.value.getStoreSnapshot())
 		editor.clearHistory()
 
--- a/packages/tldraw/src/lib/utils/tldr/file.ts
+++ b/packages/tldraw/src/lib/utils/tldr/file.ts
@@ -14,6 +14,7 @@ import {
 	TLRecord,
 	TLSchema,
 	TLStore,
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/utils/tldr/file.ts
```diff
@@ -148,9 +133,10 @@ export function parseTldrawJsonFile({
 	// records. lets create a store with the records and migrate it to the
 	// latest version
 	let migrationResult: MigrationResult<SerializedStore<TLRecord>>
+	let storeSnapshot: SerializedStore<TLRecord>
 	try {
 		const records = pruneUnusedAssets(data.records as TLRecord[])
-		const storeSnapshot = Object.fromEntries(records.map((r) => [r.id, r]))
+		storeSnapshot = Object.fromEntries(records.map((r) => [r.id, r]))
 		migrationResult = schema.migrateStoreSnapshot({ store: storeSnapshot, schema: data.schema })
```
ds300, line 140: i think you can omit this line since it will be done again during createTLStore

ds300:
reading this PR made me realize the actual reason why the old v2 files were broken.

The problem is the `upgradeSchema` helper, which didn't create a version entry for the store schema migrations, which had the logic to delete the old record types.

This is the real fix 

image

My bad, I should have figured this out earlier.

ds300 (approved):
if we also apply the changes in my other comment this will be good to go

packages/tldraw/src/lib/utils/tldr/file.ts
MitjaBezensek: I does allow us to have a separate error message though. Not sure if we need it though 🤷‍♂️

PR #4530, opened 2024-09-16
.dev: use Git LFS for our blog images

mimecuvalo:
Convert the items in `/blog` to use Git LFS to avoid making our repo too heavy.

I turned on the setting in Vercel to enable this as well.

Screenshot 2024-09-16 at 11 27 26

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek (approved):
Looks good. We should update our `New machine setup` page in notion to include `git-lfs` installation.

ds300 (approved):
Never used this before but seems like a good idea and happy to give it a go.

PR #4534, opened 2024-09-17
[Example] Setting opts in export-canvas-as-image 

nayounsang:
- I thought it would be nice to have an example that would allow users to experiment with the results depending on the option when exporting.
- Rather than creating a new example, I added a feature to an existing example because it was appropriate.
### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan
- [ ] Unit tests
- [ ] End to end tests

### Release notes
- I add new feature at [export-canvas-as-image example](https://examples.tldraw.com/export-canvas-as-image).
image
image
- It is possible to open and close. It was placed so as not to affect the existing layout and drawing as much as possible.
- It can set opts: background, darkmode, padding, scale and Box's arg(x,y,w,h).
  - Background and darkmode are checkbox.
  - Padding, scale, box's arg are number input.
- I accidentally manipulated the script in root's package.json, but it was restored to its original state. Because I didn't know there was a script to run the example.
- `Control` component is implemented simply for DX.
- ⚠️ I haven't implemented preserveAspectRatio yet. I will implement it if necessary.
- ⚠️ If x, y, w, and h are only partially undefined in the box, no processing was performed. I thought it would be a good idea to check the results.


steveruizok:
Pushing some changes, then let's land this!

PR #4537, opened 2024-09-17
Add id for text shape

jshoung:
Resolves https://github.com/tldraw/tldraw/issues/4538

This PR adds an `id` for `TextShapeUtil`.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a text shape
2. Inspect the element with class `tl-text-wrapper`
3. It should have it's `shapeId` as it's `id` attribute

1. Create a geo shape
2. Inspect the element with class `tl-text-wrapper`
3. It should not have any `id` attribute — otherwise there would be duplicate elements with the same `id`

### Release notes

- Text shapes will now have an `id` of `shapeId`, similar to other shape types

### Evidence

|  GeoShapeUtil (no id on text wrapper) |  TextShapeUtil (has id on text wrapper) |
|---|---|
|  CleanShot 2024-09-17 at 17 52 49@2x | CleanShot 2024-09-17 at 17 52 31@2x |

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -617,7 +617,8 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 				</SVGContainer>
 				{showArrowLabel && (
 					<TextLabel
-						id={shape.id}
+						id={shape.id + '-label'}
+						shapeId={shape.id}
 						classNamePrefix="tl-arrow"
 						type="arrow"
 						font={shape.props.font}
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
@@ -465,7 +465,8 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 						}}
 					>
 						<TextLabel
-							id={id}
+							id={id + '-label'}
+							shapeId={id}
 							type={type}
 							font={font}
 							fontSize={LABEL_FONT_SIZES[size] * shape.props.scale}
--- a/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
@@ -214,7 +214,8 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
 					}}
 				>
 					<TextLabel
-						id={id}
+						id={id + '-label'}
+						shapeId={id}
 						type={type}
 						font={font}
 						fontSize={(fontSizeAdjustment || LABEL_FONT_SIZES[size]) * scale}
--- a/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
+++ b/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
@@ -193,7 +193,8 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 					/>
 				</svg>
 				<TextLabel
-					id={id}
+					id={id + '-label'}
+					shapeId={id}
 					type={type}
 					font={font}
 					textWidth={shape.props.w}
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -1769,8 +1769,6 @@ export const TextLabel: React_3.NamedExoticComponent<TextLabelProps>;
 // @public (undocumented)
 export interface TextLabelProps {
     // (undocumented)
-    addId?: boolean;
-    // (undocumented)
     align: TLDefaultHorizontalAlignStyle;
     // (undocumented)
     bounds?: Box;
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -1783,7 +1783,7 @@ export interface TextLabelProps {
     // (undocumented)
     fontSize: number;
     // (undocumented)
-    id: TLShapeId;
+    id: string;
     // (undocumented)
     isNote?: boolean;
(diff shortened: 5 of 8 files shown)


jshoung:

ds300:
I wonder whether a more general solution for this would be to add an id to the Shape's outer div wrapper here https://github.com/tldraw/tldraw/blob/4ea78e9140604019cb9d4a6fdd1ff898e7f5e38b/packages/editor/src/lib/components/Shape.tsx#L155

then you could find the text label with something like 
```ts
document.qureySelector(`#${shapeId} .tl-text-wrapper`)
```

Would that still solve whatever problem this PR aims to fix?

and @steveruizok what do you think about adding the shape id to the DOM? this feels like maybe it's so obvious that the fact we haven't done it yet might be deliberate.

jshoung:
> Would that still solve whatever problem this PR aims to fix?

Hmm that would cause duplicate `id`s because every other ShapeUtil already includes the `shapeId` in the returned component. `TextShapeUtil` is the only one missing it, because it only renders `TextLabel` and doesn't have any sibling or wrapper components (this is where the `id` is located for the other ShapeUtils`)

I could alternatively wrap `TextLabel` inside `TextShapeUtil` and add the `id` to that, but I was concerned there would be some unintended side effects from modifying the DOM structure.

ds300:
> every other ShapeUtil already includes the shapeId in the returned component

ah ok I didn't realise this. in that case this PR makes sense to be consistent with the other shapes 👍🏼

steveruizok:
Hey @jshoung, I see what you're after here! I'm pushing up some changes that differentiate between the `shapeId` of the TextLabel and the `id` that gets assigned to the element. I believe this would be a sneaky breaking change for folks who are using the TextLabel component (they'd need to add the `shapeId` prop and update the `id` prop).

Does this make sense for you?

steveruizok:
> I wonder whether a more general solution for this would be to add an id to the Shape's outer div wrapper here
> 
> https://github.com/tldraw/tldraw/blob/4ea78e9140604019cb9d4a6fdd1ff898e7f5e38b/packages/editor/src/lib/components/Shape.tsx#L155
> 
> then you could find the text label with something like
> 
> ```ts
> document.qureySelector(`#${shapeId} .tl-text-wrapper`)
> ```
> 
> Would that still solve whatever problem this PR aims to fix?
> 
> and @steveruizok what do you think about adding the shape id to the DOM? this feels like maybe it's so obvious that the fact we haven't done it yet might be deliberate.

Erp, missed this while working on the branch. We could?

ds300:
> We could?

no we couldn't alas. it'd be a breaking change because, as @jshoung mentioned above, the individual shapes already add the id into the DOM in their `component` methods.

jshoung:
> Hey @jshoung, I see what you're after here! I'm pushing up some changes that differentiate between the `shapeId` of the TextLabel and the `id` that gets assigned to the element. I believe this would be a sneaky breaking change for folks who are using the TextLabel component (they'd need to add the `shapeId` prop and update the `id` prop).
> 
> Does this make sense for you?

Yes this change makes sense! Our use case requires selecting shapes by their `id`, and `TextShapeUtil` was the only one that didn't support this. I imagine there'll also be some benefit to allowing the text labels within other shapes to be targeted as well.

I pushed an update to remove the unused `addId` prop and updated the tests to check for the new `-label` id as well.

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -617,7 +617,8 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 				</SVGContainer>
 				{showArrowLabel && (
 					<TextLabel
-						id={shape.id}
+						id={shape.id + '-label'}
+						shapeId={shape.id}
```
MitjaBezensek, line 621: We probably don't want to add this one as the `SVGContainer` sibling already sets the id.

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -465,7 +465,8 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 						}}
 					>
 						<TextLabel
-							id={id}
+							id={id + '-label'}
+							shapeId={id}
```
MitjaBezensek, line 469: We probably don't want to add this one as the `SVGContainer` sibling already sets the id.

packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -214,7 +214,8 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
 					}}
 				>
 					<TextLabel
-						id={id}
+						id={id + '-label'}
+						shapeId={id}
```
MitjaBezensek, line 218: Same here, parent `div` already sets the id.

jshoung:
@MitjaBezensek My understanding for how it works on the `*ShapeUtil` components is the `id` being set it different from the one on the sibling/parent (i.e. `shape.id`), since it is appended with `-label`, so it wouldn't exactly be a duplicate `id`.

jshoung:
@steveruizok @ds300 Anything else I can do for this PR to help get it merged? Would look to have this included in a future release!

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
steveruizok: > We probably don't want to add this one as the `SVGContainer` sibling already sets the id.

In this case, do we even need to do this PR? If a user can target this with JS / CSS already?

steveruizok: Not sure I follow, it will be <div class="tl-shape"><svg id={id}/><div id={id + '-label'}/></div>

jshoung:
Looks like this PR is no longer relevant because of https://github.com/tldraw/tldraw/pull/4694

PR #4539, opened 2024-09-17
Add some additional debugging tools.

MitjaBezensek:
Adds two context menu items to improve the developer experience.

The first one just logs the currently selected shapes. 

The second one allows you to select some shapes you'd want to track. It then prints them out when they change.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Adds some additional debugging tools to the develop experience.


apps/examples/src/misc/develop.tsx
```diff
@@ -1,20 +1,74 @@
...
 			<Tldraw
 				licenseKey={getLicenseKey()}
-				overrides={[performanceOverrides]}
+				overrides={[performanceOverrides, debuggingOverrides]}
 				persistenceKey="example"
 				onMount={(editor) => {
 					;(window as any).app = editor
 					;(window as any).editor = editor
+					const dispose = editor.store.sideEffects.registerAfterChangeHandler(
+						'shape',
```
mimecuvalo, line 58: nit: let's move this function to not be inline anymore (just for readability)

apps/examples/src/misc/diff.ts
```diff
@@ -0,0 +1,23 @@
...
+	const allKeys = new Set([...Object.keys(prev), ...Object.keys(next)])
+	for (const key of allKeys) {
+		const prevValue = prev[key]
+		const nextValue = next[key]
+		const keyPath = format(prefix, key)
+		if (prevValue !== nextValue) {
+			if (typeof prevValue === 'object' && typeof nextValue === 'object') {
+				getDiff(prevValue, nextValue, keyPath, diff)
+			} else {
+				diff[keyPath] = { prev: prev[key], next: next[key] }
```
mimecuvalo, line 18: i think a little tweak here would be to add the shape id to the table (otherwise if there's multiple things being tracked it's hard to tell what's what.

this isn't perfect, but you could do:
```
diff[keyPath] = { [`prev:${prev.id}`]: prev[key], [`next:${next.id}`]: next[key] }
```

mimecuvalo (approved):
🤩 i'm love dev tools. pre-approving modulo couple comments

Issue #4540, opened 2024-09-17
How do I create default shapes in external UI?


jdgalani:
### What's the example?

I have a custom toolbar UI placed on right side of the canvas.

I am replicating same shapes there, but I dont know what method to call for the select particular shape.

Example say for ellipse,
I have tried,

` editor.setCurrentTool("geo.ellipse");
`   
` editor.setCurrentTool("ellipse");
`

I have tried using internal MenuComponents it works out of the box, but it does not go well with the UI I am planning to build.
I am replacing the icons also.
So my best bet is to have a method.

I have done hand, select, note those things are working great.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MitjaBezensek:
Hey!

If your UI is inside the `ToolsContext` then you can do something like this:
```ts
const tools = useTools()
const oval = tools['oval']
oval?.onSelect('menu')
```

And if you are using it outside of the context then you can do it directly, just like we do it in [useTools hook](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/ui/hooks/useTools.tsx#L104-L108)):
```ts
editor.setStyleForNextShapes(GeoShapeGeoStyle, 'oval')
editor.setCurrentTool('geo')
```

You can see all the `geo` tool ids [here](https://github.com/tldraw/tldraw/blob/main/packages/tlschema/src/shapes/TLGeoShape.ts#L28-L47).

jdgalani:
Thanks, for quick response. First one works like charm.

PR #4545, opened 2024-09-17
Add an example of switching to the geo tool.

MitjaBezensek:
Setting the tool to the `geo` tool is a bit special, so it warrants being added to this example.

See #4540 

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Add an example of how you can set the current tool to a `geo` tool.


Taha-Hassan-Git (approved):
Looks good, is it worth adding to the other external ui example as well?

https://github.com/tldraw/tldraw/tree/main/apps/examples/src/examples/external-ui-context

PR #4548, opened 2024-09-17
Clean up `apps` directory

SomeHats:
Post 3.0 spring cleaning?

There a new `internal` folder with things that people who don't work at tldraw should never need to look at. The apps folder contains just our actual apps, with the various dotcom services under `apps/dotcom`.

vercel deploy will fail on this until it's ready to land, at which point i'll update the vercel config to point at the new script locations

### Change type

- [x] `other`


mimecuvalo:
this all looks good btw - thanks for landing quickly to avoid too many conflicts!

PR #4551, opened 2024-09-18
Fix cloudflare sync template deploy

phgn0:
Add a cloudflare worker name to `templates/sync-cloudflare/wrangler.toml`. This fixes the following deploy error:

```
Running yarn wrangler deploy produces the following error on main:

✘ [ERROR] You need to provide a name when publishing a worker. Either pass it as a cli arg with `--name <name>` or in your config file as `name = "<name>"`
To fix this, add a worker name to wrangler.toml.
```

Also add the `@cloudflare/workers-types` dev package referenced in `templates/sync-cloudflare/tsconfig.worker.json` to fix the following TypeScript warning:

```
Cannot find type definition file for '@cloudflare/workers-types'.
  The file is in the program because:
    Entry point of type library '@cloudflare/workers-types' specified in compilerOptionsts
```

### Change type

- [x] `bugfix`


phgn0:

phgn0:
Related, it would be helpful to document that https://github.com/tldraw/tldraw-sync-cloudflare is a mirror of `templates/sync-cloudflare`. To avoid pointless PRs to https://github.com/tldraw/tldraw-sync-cloudflare/pulls?q=is%3Apr+is%3Aclosed.

PR #4559, opened 2024-09-19
[sync] Allow doing CRUD directly on the server

ds300:
We've had a couple of requests for the ability to update the room data directly on the server, which totally make sense when implementing any kind of workflow that involves talking to an external service.

This PR adds a simple helper that gives you an async transactional context in which to perform updates that will be committed at the end, and which are isolated from any other concurrent updates happening in the server due to client activity or even other transactions. At the end of a transaction, the generated diff is calculated as if it were on a client and 'sent' to the room via a push request.

### Change type

- [x] `feature`
- [x] `api`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Adds the `updateStore` method to the `TLSocketRoom` class, to allow updating room data directly on the server.

### DIFF
--- a/packages/sync-core/api-report.md
+++ b/packages/sync-core/api-report.md
@@ -201,6 +201,18 @@ export interface RoomSnapshot {
     tombstones?: Record<string, number>;
 }
 
+// @public (undocumented)
+export interface RoomStoreMethods {
+    // (undocumented)
+    delete(recordOrId: string | UnknownRecord): void;
+    // (undocumented)
+    get(id: string): UnknownRecord;
+    // (undocumented)
+    getAll(): UnknownRecord[];
+    // (undocumented)
+    put(record: UnknownRecord): void;
+}
+
 // @internal (undocumented)
 export type SubscribingFn<T> = (cb: (val: T) => void) => () => void;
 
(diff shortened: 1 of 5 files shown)


strawhatduckk (approved):

packages/sync-core/src/lib/TLSocketRoom.ts
```diff
@@ -287,6 +279,32 @@ export class TLSocketRoom<R extends UnknownRecord = UnknownRecord, SessionMeta =
 		oldRoom.close()
 	}
 
+	/**
+	 * Allow applying changes to the store inside of a transaction.
+	 *
+	 * You can get values from the store by id with `store.get(id)`.
+	 * These values are safe to mutate, but to commit the changes you must call `store.set(...)` with the updated value.
```
ds300, line 286:
```suggestion
	 * These values are safe to mutate, but to commit the changes you must call `store.put(...)` with the updated value.
```

packages/sync-core/api-report.md
```diff
@@ -201,6 +201,18 @@ export interface RoomSnapshot {
     tombstones?: Record<string, number>;
 }
 
+// @public (undocumented)
+export interface RoomStoreMethods {
```
SomeHats, line 205: Could we make this `RoomStoreMethods<R>` instead of always returning/accepting `UnknownRecord`?

SomeHats (approved):
this is a really great change. thanks david!

PR #4565, opened 2024-09-20
Unify links for vs code.

MitjaBezensek:
Unify links in vs code with the links on dotcom.

Also make the watermark link work. You can't use anchors in the vs code extension - we actually have a system for handling this with the `runtime` and vs code defines it's own runtime.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open vs code extension.
2. The links in the main menu should be the same as on dotcom. 
3. Clicking on the watermark should open tldraw.dev.

### Release notes

- Unify vs code extension links. Make the watermark link work in the vs code extension.

### DIFF
--- a/apps/vscode/editor/src/Links.tsx
+++ b/apps/vscode/editor/src/Links.tsx
@@ -3,43 +3,42 @@ import { openUrl } from './utils/url'
 
 export function Links() {
 	return (
-		<TldrawUiMenuGroup id="links">
-			<TldrawUiMenuItem
-				id="github"
-				label="help-menu.github"
-				readonlyOk
-				icon="github"
-				onSelect={() => {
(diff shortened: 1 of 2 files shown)


apps/vscode/editor/src/Links.tsx
```diff
@@ -3,43 +3,42 @@ import { openUrl } from './utils/url'
 
 export function Links() {
 	return (
-		<TldrawUiMenuGroup id="links">
```
MitjaBezensek, line 6: Should probably pull this out somewhere, but we don't have a really good spot for it 🤷‍♂️

mimecuvalo (approved):
👍

PR #4567, opened 2024-09-20
hand tool: remove lockable attr

mimecuvalo:
noticed this in the How To meeting. i think this is an oversight, locking/unlocking the hand tool doesn't do anything.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix Hand tool being lockable when it already is.


MitjaBezensek (approved):
Nice catch 🙌

PR #4569, opened 2024-09-20
text: be able to keep tool locked

mimecuvalo:
Noticed during the How-To sesh that you can't lock text shapes 😢  This remedies that.

https://github.com/user-attachments/assets/b627b2b1-728d-4531-a604-30c41528cc27

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Make text shape be lockable


SomeHats (approved):

steveruizok:
anyone who uses this feature is nuts imo but looks good!

steveruizok:
I'd like to take a look at the logic in the state chart, there might be a way to do that without adding to the transitions

mimecuvalo:
> I'd like to take a look at the logic in the state chart, there might be a way to do that without adding to the transitions

Sure, lemme know 👍

steveruizok:
Ok, so I think you were fighting here with a really hidden part of the app: the side effect that transitions to editing automatically when we change the `editingShapeId`. The logic is pretty gnarly, but if we move it all there then we can get rid of some of the other distributed gnarliness.

packages/tldraw/src/lib/ui/hooks/useTools.tsx
```diff
@@ -57,7 +57,7 @@ export function ToolsProvider({ overrides, children }: TLUiToolsProviderProps) {
 				kbd: 'v',
 				readonlyOk: true,
 				onSelect(source) {
-					editor.setCurrentTool('select')
+					editor.setCurrentTool('select.idle')
```
steveruizok, line 60: be explicit about going to the select state rather than the editing state, because editing is a child of select

steveruizok (approved):

mimecuvalo:
Thanks for assisting but maybe this PR was closed a little too quickly. It's hard in the latest changes now to exit the mode — I'll create a followup PR to fix.

PR #4570, opened 2024-09-20
selection: allow cmd/ctrl to add to selection

mimecuvalo:
In the How-To sesh, I noticed that using Shift of course lets you add to a selection of shapes, but Cmd/Ctrl does not.
Typically, cmd/ctrl lets you do this in other contexts so some of that muscle memory doesn't get allowed in tldraw currently.
This enables cmd/ctrl to have the same behavior as shift.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Selection: allow cmd/ctrl to add multiple shapes to the selection.


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9571,6 +9571,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 				if (info.key === 'ShiftRight') info.key = 'ShiftLeft'
 				if (info.key === 'AltRight') info.key = 'AltLeft'
 				if (info.code === 'ControlRight') info.code = 'ControlLeft'
+				if (info.code === 'MetaRight') info.code = 'MetaLeft'
```
steveruizok, line 9574: Are we distinguishing between control and meta? We mix the two into the `info.ctrlKrey`. Is it possible to do this type of selection on Windows then?

mimecuvalo: - meta/ctrl key is currently combined in our logic but i think it's...fine, generally speaking. so on windows, this technically might not be correct if you're holding down the Win key (meta) but it will work as expected for the Ctrl key on Windows
- so yeah, you can do this on Windows with Ctrl key.
- and double yes, Miro lets Cmd/Ctrl be additive

steveruizok: ok, I'd like to try on a PC before landing just to be sure, but I think this will be good to land once we have tests up. Selection is fuuucking complicated, appreciate the patience while I go through the cases

steveruizok:
There was a little more to this than I thought (there always is with selection!)

The behavior is a _little_ different for shift and meta/ctrl. When you shift click a shape, we add it to your selected shapes on pointer down. However, because ctrl has this secondary function of brush selection (which also combines with shift to allow you to brush new shapes into your selection), we can't select shapes on pointer down. So a ctrl click will add / remove a shape from your selection on pointer up, but a click-and-drag will start brushing without selecting anything new.

Hideous graph:
image

steveruizok:
Good to go from my side! Would you test this on PC in the office before landing?

The main PC/Mac concerns are:
- on Mac, meta clicking a shape should toggle its selection
- on PC, control clicking a shape should toggle its selection
- on Mac, control clicking should select a shape (ie do a normal select on pointer down) and then open the context menu

steveruizok (approved):

mimecuvalo:
Damn, thanks for jumping in! I somehow successfully nerd snipe you every time hah.

> The behavior is a little different for shift and meta/ctrl. When you shift click a shape, we add it to your selected shapes on pointer down. However, because ctrl has this secondary function of brush selection (which also combines with shift to allow you to brush new shapes into your selection), we can't select shapes on pointer down. So a ctrl click will add / remove a shape from your selection on pointer up, but a click-and-drag will start brushing without selecting anything new.
> Also notice that for shift key: we add unselected shapes to the selection on shift + pointer down but we don't remove selected shapes until the shift + pointer up, because the shift + drag does have a secondary function when pointing selected shapes (drag along axis)

Ok, yeah, this all makes sense — thanks for letting me know about these subtleties!

> And for Macs, we do need to distinguish between ctrl and meta. I've added that to this PR (I'd avoiding distinguishing until now). The important thing is that we want to select shapes when we right click them, or control click them on macs.

👍 

> Good to go from my side! Would you test this on PC in the office before landing?

👍

mimecuvalo:
double-checked on PC - looks good. i added another commit to consolidate some of the ctrl/meta logic. in the Mozilla world we used to call this the "Accel" key, so i'm tying together some of that logic.

PR #4571, opened 2024-09-21
Fix label wrapping

steveruizok:
This PR fixes a bug that was causing broken wrapping on arrow labels.

We were calculating the bounds of the arrow label with an included padding when measuring the arrow label, but not removing that padding from the arrow label text.

Before:
image

After:
image

### Change type

- [x] `bugfix`

### Test plan

1. Use the text arrow labels
2. Check for weird line breaks

### Release notes

- Fixed a bug with arrow label text measurements.


mimecuvalo (approved):
LG - thanks!

PR #4566, opened 2024-09-20
Update installation.mdx

mbj36:
Update Licensing docs

PR #4575, opened 2024-09-23
Steve docs improvements

steveruizok:
pick from fdc254388ff24b3803f6de84d5fbda01a76ed11e


apps/docs/next.config.js
```diff
@@ -85,6 +85,22 @@ const nextConfig = {
 				destination: '/reference/:slug*',
 				permanent: true,
 			},
+			// redirects at the top level
+			{
+				source: '/examples',
+				destination: '/examples/basic/basic',
```
mimecuvalo, line 91: this should redirect to `/examples/` not basic/basic - since it's a permanent redirect we should redirect to something that'll be more evergreen

apps/docs/next.config.js
```diff
@@ -85,6 +85,22 @@ const nextConfig = {
...
+				permanent: true,
+			},
+			{
+				source: '/reference',
```
mimecuvalo, line 95: hm i see for `/reference/` and `/docs/` (they 404) we don't have the same redirect like `/examples/` - i'd have to dig into why they don't have the same behavior

mimecuvalo (commented):
might be worth cherry-picking this too https://github.com/tldraw/tldraw/pull/4566/files

PR #4577, opened 2024-09-23
Don't upload a snapshot from the snapshots page

MitjaBezensek:
This prevents uploading a snapshot when we are on a snapshot page. We can just use the current link instead of generating a new one.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Share a snapshot.
2. Open the snapshot.
3. Copy the share link. This should not upload the current snapshot to the server and should result copy the current url to the clipboard.

### Release notes

- Copying a snapshot link from a snapshot page now just returns the current url.

### DIFF
--- a/apps/dotcom/client/src/utils/clipboard.ts
+++ b/apps/dotcom/client/src/utils/clipboard.ts
@@ -1,0 +1,14 @@
+/** @internal */
+export async function writeToClipboard(result: Promise<Blob | ''>) {
+	if (navigator?.clipboard?.write) {
+		await navigator.clipboard.write([
+			new ClipboardItem({
+				'text/plain': result,
+			}),
+		])
+	} else if (navigator?.clipboard?.writeText) {
+		const link = await result
+		if (link === '') return
+		navigator.clipboard.writeText(await link.text())
+	}
+}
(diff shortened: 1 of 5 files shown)


apps/dotcom/client/src/utils/clipboard.ts
```diff
@@ -0,0 +1,14 @@
+/** @internal */
+export async function writeToClipboard(result: Promise<Blob | ''>) {
```
mimecuvalo, line 2: wait, i know this isn't your code and you're refactoring but why does one path `await result` and the other doesn't? shouldn't they both need to `await`?

mimecuvalo (approved):
LG - pre-approving one comment about that clipboard code

apps/dotcom/client/src/utils/clipboard.ts
MitjaBezensek: I think that in async contexts you need to pass a promise to `clipboard.write` to make Safari happy.
https://wolfgangrittner.dev/how-to-use-clipboard-api-in-firefox/

mimecuvalo: gotcha, ok makes sense. if you get a chance - add your comment in the code!

PR #4580, opened 2024-09-23
Fix collaboration shape indicator showing a line through the arrow's label

### LINKED ISSUE #4562: Multiplayer arrow indicators are drawn over labels

MitjaBezensek:
Masks for the collaborator indicators where somehow not getting correctly applied. First I played around with adding a random string to the end of the mask id which fixed it. But then I also found out this works.

I think there's some issue with ids for inline defined masks 🤷‍♂️ Was able to find a few similar issues online (but not exactly the same though).

Fixes #4562 

### Before

https://github.com/user-attachments/assets/6a6982d6-7984-4b4e-beb2-48c3094dce9d

### After

https://github.com/user-attachments/assets/0340fd96-c55e-4d61-a42d-8f1207b3c15c

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create an arrow with a label in a multiplayer project.
2. Select the arrow.
3. Open the same project in another browser / incognito.
4. The collaborator indicator should not be shown on top of the arrow's label.

### Release notes

- Fix an issue with arrow collaborator indicator showing on top of the arrow's label.


SomeHats (approved):

steveruizok:
Thanks @MitjaBezensek !

PR #4601, opened 2024-09-25
[botcom] purge prototype leftovers

ds300:
my god deleting code is fun and satisfying 💆‍♂️

### Change type

- [x] `other`


steveruizok:
Agree with the changes but I'll have to bookmark this commit, as there's some UI here that is still useful as we design those other views.

ds300:
Yeah that's why we deleted unused stuff _after_ merging your PR, so we could refer back here and pick stuff out as and when we need it.

PR #4604, opened 2024-09-25
[botcom] Share menu

steveruizok:
This PR adds UI for the share menu.

Fun fact: while writing this I found a much better way for us to do our QR codes with SVGs, if we want it!

localhost_3000_q_f_0 (1)

localhost_3000_q_f_0 (2)

localhost_3000_q_f_0

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
Some feature differences here:

- projects are either not shared, or shared as readonly or shared as edit. There aren't separate readonly links.
- the "copy link" button also turns on sharing (with an animation that should be noticeable). Beats disabled buttons.

steveruizok:
I'm feeling more convinced that we need separate UI for exporting images. It's too confusing, especially around selection, formats, and options like transparent backgrounds.

apps/dotcom/client/src/tla/utils/qrcode.ts
```diff
@@ -0,0 +1,12 @@
+export async function createQRCodeImageDataString(url: string) {
+	const QRCode = await import('qrcode')
+	return await QRCode.toString(url, {
```
mimecuvalo, line 3: neat!

apps/dotcom/client/src/tla/utils/copy.ts
```diff
@@ -0,0 +1,11 @@
+export async function copyTextToClipboard(text: string) {
```
mimecuvalo, line 1: feels like this would be just useful to expose (either by putting it in packages/utils or from tldraw, either way)

mimecuvalo: (to avoid duplication)

mimecuvalo:
lemme know if you wanna review this more / or just land and iterate. i think at our current stage of prototyping it's ok to just land since i see the designs are changing pretty fast. up to you!

apps/dotcom/client/src/tla/utils/copy.ts
steveruizok: It kind of is a duplication, though the one in the tldraw library takes a blob rather than a string.

steveruizok:
Ok, I think this is ready for a deeper review now.

- added tabs
- added qr code
- added export tab
- added export preview
- added toggle for inline help

ds300 (approved):
Love this. So clean and clear.

- the css modules are an improvement
- (nonblocking) the help icon acting as a toggle is weird. It causes some nasty jarring repaints, it doesn't look like a toggle and you can't tell which state it's in without clicking it, when you're on the export tab it doesn't do anything, in general it is not particularly well colocated with the bits of UI it's trying to provide clarity for. We should really adopt tooltips for this kind of thing, it's so much easier to integrate into designs, it's much better UX than this, and it's just a well established pattern that people use every day.

apps/dotcom/client/src/tla/components/TlaSwitch/switch.module.css
```diff
@@ -0,0 +1,48 @@
...
+	--switch-w: calc(var(--switch-h) * 1.618);
+	height: var(--switch-h);
+	width: var(--switch-w);
+	position: relative;
+	border-radius: 99px;
+	background-color: var(--tla-color-text-3);
+	transition: all 0.12s;
+}
+
+.switch:after {
```
mimecuvalo, line 27: nit: `::` here and below

apps/dotcom/client/src/tla/components/TlaFileShareMenu/file-share-menu.module.css
```diff
@@ -0,0 +1,151 @@
...
+	padding: var(--share-menu-padding);
+	flex-direction: column;
+	gap: var(--share-menu-padding);
+}
+
+.section:nth-of-type(n + 2) {
+	border-top: 1px solid var(--tla-color-border);
+}
+
+.controlGroup {
```
mimecuvalo, line 47: can i just say, that one of the things i like about css modules is getting away from the BEM stuff?

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenuSharePage.tsx
```diff
@@ -0,0 +1,192 @@
...
+			Copy snapshot link
+		</TlaShareMenuCopyButton>
+	)
+}
+
+function QrCode({ fileId }: { fileId: TldrawAppFileId }) {
+	const ref = useRef<HTMLImageElement>(null)
+
+	// Save the QR codes in local storage
+	const [qrCode, setQrCode] = useLocalStorageState<string | null>(fileId + 'qr-code-11', null)
```
mimecuvalo, line 164: what is the `code-11` for?

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenuExportPage.tsx
```diff
@@ -0,0 +1,333 @@
...
+	const handleSelectChange = useCallback(
+		(value: TldrawAppUser['exportTheme']) => {
+			app.setUserExportTheme(userId, value)
+		},
+		[app, userId]
+	)
+
+	return (
+		<TlaShareMenuControl>
+			<TlaShareMenuControlLabel>Theme...</TlaShareMenuControlLabel>
```
mimecuvalo, line 172: little nit: `…` instead of `...`

mimecuvalo (approved):
👍 

my only high-level feedback is the `Anyone with the link can...`:
i know we're trying to see if we can make it work so that there's only one link and it's either Edit or View. but seeing it in action on this PR, it _does_ feel like we're losing something, not being able to send an Edit link to some people and a View link to others. feels like we should still consider having two separate links, even if it as clean as single primary CTA.

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenuSharePage.tsx
steveruizok: just lazy bumps of the local storage key

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenuExportPage.tsx
```diff
@@ -0,0 +1,333 @@
...
+	const handleSelectChange = useCallback(
+		(value: TldrawAppUser['exportFormat']) => {
+			app.setUserExportFormat(userId, value)
+		},
+		[app, userId]
+	)
+
+	return (
+		<TlaShareMenuControl>
+			<TlaShareMenuControlLabel>Export as...</TlaShareMenuControlLabel>
```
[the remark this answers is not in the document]
steveruizok: Good catch. I'll drop, made more sense in a previous iteration

apps/dotcom/client/src/tla/components/TlaFileShareMenu/file-share-menu.module.css
steveruizok: Yeah I like it too!

steveruizok:
> my only high-level feedback is the `Anyone with the link can...`: i know we're trying to see if we can make it work so that there's only one link and it's either Edit or View. but seeing it in action on this PR, it _does_ feel like we're losing something, not being able to send an Edit link to some people and a View link to others. feels like we should still consider having two separate links, even if it as clean as single primary CTA.

Yeah, it does feel like one use case is gone. But I think it's a net improvement on the sort of mental model of how and whether a project is shared. I'd like to revisit this sort of UI again once we have groups and named users, because that could be the place for shared editors + readonly external viewers, or dynamic role switching.

steveruizok:
(This access stuff is such a strong conversion moment that I don't think it's been explored much outside of that.)

PR #4612, opened 2024-09-26
Support auto switching wheel behavior

### LINKED ISSUE #4579: [Feature]: Make scroll zoom with mouse, but pan with trackpad
### What's the feature?

I (and some others I've talked to) feel that it's most intuitive if scrolling with a mouse zooms, while moving two fingers on a touchpad (which sends scroll events) pans. I know about the `wheelBehavior` option, but that only allows the same behavior (either zoom or pan) for scroll on both mouse and trackpad.

Afaik this would require some heuristic to determine if a mouse or touchpad is used, as they send the same events.

For reference this is how Miro and Microsoft Whiteboard works.

The developer package

### Contact Details

_No response_

### Code of Conduct

trygve-aaberge-adsk:
This implements support for setting wheelBehavior to auto. What this means is that it will try to detect if you're using a mouse or a touchpad, and use wheelBehavior zoom when a mouse is used and pan when a touchpad is used. This is more intuitive behavior than always having it set to wheel or pan in my opinion.

As far as I know, it's unfortunately not possible to know if wheel events are generated from a mouse or a touchpad. The heuristic implemented is that if you scroll horizontally it assumes you use a touchpad and switch to pan, and if you pan with the middle mouse button it assumes you're using a mouse and switches to zoom. This seems to be similar to what other similar software does.

Fixes #4579

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Use a touchpad to pan the board by scrolling
2. Switch to a mouse and pan the board by dragging with the middle mouse button
3. Scroll and observe that it now zooms on scroll
4. Switch back to the touchpad, scroll horizontally and observe that it pans again

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Support automatically switching wheelBehavior when switching between touchpad and mouse (as determined by a heuristic). Use `pan` when using a touchpad and `zoom` when using a mouse.


packages/editor/src/lib/constants.ts
```diff
@@ -4,7 +4,7 @@ import { EASINGS } from './primitives/easings'
 /** @internal */
 export const DEFAULT_CAMERA_OPTIONS: TLCameraOptions = {
 	isLocked: false,
-	wheelBehavior: 'pan',
+	wheelBehavior: 'auto',
```
trygve-aaberge-adsk, line 7: I changed the default to `auto` because I think that's the most intuitive behavior, but that's of course up to you if you want to do.

trygve-aaberge-adsk:
No worries. And sorry for not signing the CLA yet, I had to talk to legal in my company (nothing particular about your CLA, just general policy for CLAs), but hopefully we'll get it sorted soon.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9165,6 +9165,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	/** @internal */
 	private _restoreToolId = 'select'
 
+	/** @internal */
+	private _autoWheelBehavior: 'pan' | 'zoom' = 'pan'
```
trygve-aaberge-adsk, line 9169: Whether the auto behavior starts off as `pan` or `zoom` should perhaps also be configurable.

It would also be nice to remember the last state the user was in in localstorage, so the same one can be used initially when loading the page again. Not sure if this is something tldraw should do automatically, or if it should emit an event on changes and allow this state to be set, so it can be implemented outside of tldraw?

packages/editor/src/lib/editor/types/misc-types.ts
```diff
@@ -52,11 +52,14 @@ export interface TLCameraOptions {
...
 	 *
+	 * - `auto`: Switches dynamically between `zoom` and `pan` based on the users interactions.
 	 * - `zoom`: The wheel will zoom in and out.
 	 * - `pan`: The wheel will pan the camera.
 	 * - `none`: The wheel will do nothing.
 	 */
-	wheelBehavior: 'zoom' | 'pan' | 'none'
+	wheelBehavior: 'auto' | 'zoom' | 'pan' | 'none'
+	/** The behavior used for the wheel when `wheelBehavior` is set to auto. */
+	autoWheelBehavior?: 'zoom' | 'pan'
```
trygve-aaberge-adsk, line 62: Do you think it makes sense to have this in camera options?

MitjaBezensek:
Sorry for some spam here, testing out the logic to skip some of our checks for external contributors.

SomeHats (changes requested):
We have a sort of a design principle for tldraw that whilst we do our best to extract intention from user actions, in places where that's not possible we aim for predictability instead. I'm feeling cautious about introducing this because i think that changing the behaviour of the mouse wheel during use violates that principle.

This works well on a mac, but testing this on windows/chromebook devices shows some pretty weird behaviour where after space-panning, scrolling the trackpad zooms initially then suddenly transforms into a pan when the x-axis passes a certain threshold, for example.

At the very least, I wouldn't like us to change the default behaviour. I'm hesitant to include this in the SDK at all though because I don't think we can guarantee a consistently good experience with it.

SomeHats:
I think after discussing internally a bit we don't think we can get this to a good enough quality across browsers/devices right now that we'd be keen to include it in the SDK. What are you missing to be able to implement this outside of the SDK?

trygve-aaberge-adsk:
> This works well on a mac, but testing this on windows/chromebook devices shows some pretty weird behaviour where after space-panning, scrolling the trackpad zooms initially then suddenly transforms into a pan when the x-axis passes a certain threshold, for example.

Yeah, the idea is that when you're using a trackpad there's not really any reason to use space-panning. But I agree that it in some cases behaves a bit weirdly since we have to guess if a user is using a mouse or a trackpad. For us the trade-off of having a better mouse and touchpad experience in most cases (in our opinion) makes it worth it, but I can understand you have stricter requirements in your SDK.

> I think after discussing internally a bit we don't think we can get this to a good enough quality across browsers/devices right now that we'd be keen to include it in the SDK. What are you missing to be able to implement this outside of the SDK?

Alright, implementing it on our side works for us. When I opened this PR I thought it wasn't possible to do that, and it still looks like it's not possible in a good/nice way, but looking at it again now I realize that it is possible if we override `editor.dispatch` something like this:

```
  const onMount = useCallback((editor: Editor) => {
    if (!editor._originalDispatch) {
      editor._originalDispatch = editor.dispatch
    }
    editor.dispatch = (info: TLEventInfo) => {
      const cameraOptions = editor.getCameraOptions()

      if (
        info.name === "wheel" &&
        cameraOptions.wheelBehavior !== "pan" &&
        !editor.inputs.isPanning &&
        (info.delta.x !== 0 || (info.delta.y !== 0 && editor.inputs.ctrlKey))
      ) {
        editor.setCameraOptions({ wheelBehavior: "pan" })
      }

      if (
        (info.name === "pointer_down" || info.name === "pointer_move") &&
        editor.inputs.isPanning &&
        cameraOptions.wheelBehavior !== "zoom"
      ) {
        editor.setCameraOptions({ wheelBehavior: "zoom" })
      }

      editor._originalDispatch(info)
      return editor
    }
  }, [])

  return <TldrawEditor onMount={onMount}>...</TldrawEditor>
```

It would be nice to be able to do this in a proper way though, without having to override `editor.dispatch`. In other words, what we need is a way to receive the events processed in `_flushEventForTick` (to determine which wheelBehavior should be used) and then be able to update the camera options before it runs, or in some other way choose which wheelBehavior that function uses.

SomeHats:
we have our "event" event but that runs after this stuff. Maybe we should introduce a "before-event" event instead?

trygve-aaberge-adsk:
> That might make sense. Though events are async, while we would preferably need to have updated camera options before _flushEventForTick runs.

Actually I see that eventemitter3 is calling the events synchronously, so this should be fine. I created PR #5319 which seem to be working fine for my use case and should be enough. Would love it if you could take a look at that.

PR #4615, opened 2024-09-27
Add eslint rule to check that tsdoc params match with function params

MitjaBezensek:
After this [conversation](https://discord.com/channels/859816885297741824/859816885801713728/1289133853822423050) I noticed that our tsdocs `@params` and our function / method definitions don't always match. We often miss tsdoc params or have outdated names for them.

This PR:
- Adds an eslint rule to check whether the tsdoc params and the function params / method arguments match. It allows for using `_` in function params (you can specifiy `@param shapes` in the tsdocs but use `_shapes: SomeType` in function. This helps with some functions that get overriden, like [this one](https://github.com/tldraw/tldraw/blob/a19c8f73898d59500accd87d416b4ae9a7a82ece/packages/editor/src/lib/editor/shapes/ShapeUtil.ts#L299-L308).
- It fixes the issues that the new rule found.
- Also adds the `select` argument back to the `createShape` and `createShapes` methods. We could also go the other way and remove the docs instead of adding the argument.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Add lint rules to check for discrepancies between tsdoc params and function params and fix all the discovered issues.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -7175,7 +7175,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 			this.store.put(shapeRecordsToCreate)
 
-			if (select) {
+			if (select && shapeRecordsToCreate.length) {
 				this.setSelectedShapes(shapeRecordsToCreate.map((s) => s.id))
 			}
 		})
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -3849,7 +3849,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * const idsOnPage2 = editor.getPageShapeIds(myPage2)
 	 * ```
 	 *
-	 * @param page - The page (or the page id) to get.
+	 * @param page - The page (or the page id) to get the shape ids for.
 	 *
 	 * @public
 	 **/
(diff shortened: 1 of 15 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -6988,7 +6992,10 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	createShapes<T extends TLUnknownShape>(shapes: OptionalKeys<TLShapePartial<T>, 'id'>[]): this {
+	createShapes<T extends TLUnknownShape>(
+		shapes: OptionalKeys<TLShapePartial<T>, 'id'>[],
+		select = false
```
MitjaBezensek, line 6997: Docs above mention this param, but the function didn't actually define it.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -2451,13 +2450,13 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * editor.setCamera(editor.getCamera())
 	 * ```
 	 *
-	 * @param options - The camera options to set.
+	 * @param opts - The camera options to set.
```
MitjaBezensek, line 2453: Renamed options to opts, so that we now consistently use opts (at least in the Editor, we might have other places to update still).

packages/editor/src/lib/editor/Editor.ts
SomeHats: i think we removed this a while ago because it complicates the API and it's pretty straightforward to do create then select. mind if we leave it out still?

SomeHats (approved):
love this! personally would drop that select param though

packages/editor/src/lib/editor/Editor.ts
MitjaBezensek: Happy to remove 👍

PR #4616, opened 2024-09-27
[botcom] clerk scaffolding

mimecuvalo:
this sets up the basic scaffolding for using Clerk in our app. as discussed, we'll go with Clerk with just keeping an eye for future self-hosting as an option.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/components/MultiplayerEditor.tsx
```diff
@@ -121,11 +122,16 @@ export function MultiplayerEditor({
...
 
 	const handleUiEvent = useHandleUiEvents()
 
 	const storeWithStatus = useSync({
-		uri: `${MULTIPLAYER_SERVER}/${RoomOpenModeToPath[roomOpenMode]}/${roomSlug}`,
+		uri: useCallback(
+			async () =>
+				`${MULTIPLAYER_SERVER}/${RoomOpenModeToPath[roomOpenMode]}/${roomSlug}?clerk_token=${await user.getToken()}`,
+			[roomOpenMode, roomSlug, user]
+		),
```
ds300, line 134: i think we should undo the change in here, `MultiplayerEditor` is used on current dotcom and we should leave it alone and do everything with the `TlaEditor` component.

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
```diff
@@ -1,11 +1,24 @@
...
+
 export function Component() {
+	if (!PUBLISHABLE_KEY) {
+		throw new Error('Missing Publishable Key')
+	}
+
 	return (
 		<AppStateProvider>
-			<Outlet />
+			<ClerkProvider publishableKey={PUBLISHABLE_KEY} afterSignOutUrl="/">
```
ds300, line 17: when merging with my stuff I think we'll need to hoist this so it can be available on the logged out page and if the user is logged in, redirect them.
```suggestion
			<ClerkProvider publishableKey={PUBLISHABLE_KEY} afterSignOutUrl="/q">
```

ds300 (approved):

PR #4622, opened 2024-09-28
Fix watermark link opening twice

vladh:
In 37e8df7 on 20 Sep, the watermark link was changed to make it work in VS Code. However, that change introduced a regression which is currently causing the tldraw website to open _twice_ when the watermark link is clicked on web.

This commit changes the `<a>` to not actually be a link, but rather to only trigger using the `onClick` event.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open web build.
2. Click on the watermark.
3. tldraw.dev should open once, not twice.

### Release notes

- Fix watermark link opening twice


vladh:

MitjaBezensek:
Hey! Thanks for letting us know about this issue.

I guess the issue occurs in Firefox? Looks like there are some browser inconsistencies here:
- Current solution opens the link twice in FF and once in Safari / Chrome.
- The proposed change opens the link once in FF, but doesn't work in Safari / Chrome.

Made a new PR which seems to work more consistently across browsers. Maybe you could give it a try?
https://github.com/tldraw/tldraw/pull/4627

vladh:
@MitjaBezensek I've updated my solution to give VS Code its own HTML. I think this is way better, because it allows us to serve normal browsers a normal `<a>` tag, which avoids all kinds of issues.

To summarise:

* Firefox: works normally
* Chrome: works — it required `stopEventPropagation()`
* VS Code: works — it required its own `runtime.openWindow()` code
* Mobile: works — but see below

I've uncovered what I think seems to be a different issue on mobile. The `<a>` works fine on mobile, but the (1) the clickable area is extremely small, I think 1px or 2px wide, much smaller than it appears, and (2) a double click is required for the link to trigger. I've only managed to get the link to trigger using a mobile simulator on Firefox — I can't get the link to trigger on my phone in either the development or production environment. I'm not sure why this is, but it seems to be a separate problem from this issue.

MitjaBezensek:
> @MitjaBezensek I've updated my solution to give VS Code its own HTML. I think this is way better, because it allows us to serve normal browsers a normal `<a>` tag, which avoids all kinds of issues.
> 
> To summarise:
> 
> * Firefox: works normally
> * Chrome: works — it required `stopEventPropagation()`
> * VS Code: works — it required its own `runtime.openWindow()` code
> * Mobile: works — but see below
> 
> I've uncovered what I think seems to be a different issue on mobile. The `<a>` works fine on mobile, but the (1) the clickable area is extremely small, I think 1px or 2px wide, much smaller than it appears, and (2) a double click is required for the link to trigger. I've only managed to get the link to trigger using a mobile simulator on Firefox — I can't get the link to trigger on my phone in either the development or production environment. I'm not sure why this is, but it seems to be a separate problem from this issue.
> 
> PS. During development, I came across an issue which I fixed in #4630.

I guess it makes sense to add another user agent since other developers might also use tldraw to develop vs code extensions 👍 

We could also do the switch based on `isMobile` check, since it doesn't make much sense for VS Code 🤷‍♂️ 

I also noticed that it's really hard to open the link on mobile. I guess the element is too small and too close to the edge to be a good touch target. @steveruizok should we maybe make the touch target a bit bigger here?

vladh:
> I also noticed that it's really hard to open the link on mobile. I guess the element is too small and too close to the edge to be a good touch target. @steveruizok should we maybe make the touch target a bit bigger here?

In my testing, the actual touch target is significantly smaller than the `<a>`'s bounding box. I don't quite understand why.

Also, it seems CI failures are unrelated. 🤔

steveruizok:
@MitjaBezensek could you review this one? Iirc the link is not intended to be clickable on touch devices, but should be on narrow pointer devices.

MitjaBezensek:
> @MitjaBezensek could you review this one? Iirc the link is not intended to be clickable on touch devices, but should be on narrow pointer devices.

Thanks for clarifying, I think it's then good to go 👍  

@vladh thanks for bringing this up and looking into it 🙌

MitjaBezensek:
@vladh to fix the tests you can add the `environment` prop to the editor fake [here](https://github.com/tldraw/tldraw/blob/751931e46970339c5c4a042600d63a726e734128/packages/editor/src/lib/license/Watermark.test.tsx#L7-L10).

vladh:
> @vladh to fix the tests you can add the `environment` prop to the editor fake [here](https://github.com/tldraw/tldraw/blob/751931e46970339c5c4a042600d63a726e734128/packages/editor/src/lib/license/Watermark.test.tsx#L7-L10).

Oh, neat, thank you! Fixed.

PR #4624, opened 2024-09-29
[dotcom] Add dialogs

steveruizok:
This PR adds some quick app-level dialogs. I'm not wed to the design here, and we could bring in the dialog system from the editor as we've discussed.

Screenshot 2024-09-29 at 00 20 18

For comparison, here's an example of the editor dialog:
Screenshot 2024-09-29 at 00 19 44

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/components/dialogs/TlaRenameFileDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/TlaRenameFileDialog.tsx
@@ -1,0 +1,58 @@
+import * as Dialog from '@radix-ui/react-dialog'
+import { useCallback, useRef } from 'react'
+import { useValue } from 'tldraw'
+import { useApp } from '../../hooks/useAppState'
+import { TldrawAppFileId } from '../../utils/schema/TldrawAppFile'
+import { TlaButton } from '../TlaButton/TlaButton'
+import { TlaFormGroup, TlaFormInput, TlaFormItem, TlaFormLabel } from '../tla-form/tla-form'
+import styles from './dialogs.module.css'
+
+export function TlaRenameFileDialog({ fileId }: { fileId: TldrawAppFileId }) {
+	const app = useApp()
+	const ref = useRef<HTMLInputElement>(null)
+
+	const fileName = useValue(
+		'file name',
+		() => {
+			const file = app.store.get(fileId)
+			if (!file) throw Error('expected a file')
+			return file.name
+		},
+		[app]
+	)
+
+	const handleClose = useCallback(() => {
+		// rename the file
+		const file = app.store.get(fileId)
+		if (!file) return
+		const elm = ref.current
+		if (!elm) return
+		const name = elm.value.trim() || fileName
+		app.store.put([{ ...file, name }])
+	}, [app, fileId, fileName])
+
+	return (
+		<>
+			<TlaFormGroup>
+				<TlaFormItem>
+					<Dialog.DialogTitle asChild>
+						<TlaFormLabel>Rename file</TlaFormLabel>
+					</Dialog.DialogTitle>
+					<TlaFormInput ref={ref} defaultValue={fileName} />
+				</TlaFormItem>
+			</TlaFormGroup>
+			<TlaFormGroup className={styles.actionGroup}>
+				<Dialog.Close asChild>
+					<TlaButton className={styles.actionButton} variant="secondary">
+						Cancel
+					</TlaButton>
+				</Dialog.Close>
+				<Dialog.Close asChild>
+					<TlaButton className={styles.actionButton} onClick={handleClose}>
+						Save
+					</TlaButton>
+				</Dialog.Close>
+			</TlaFormGroup>
+		</>
+	)
+}
--- a/apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
+++ b/apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
@@ -1,11 +1,28 @@
 import { Outlet } from 'react-router-dom'
-import { AppStateProvider } from '../hooks/useAppState'
+import { useValue } from 'tldraw'
+import { AppStateProvider, useApp } from '../hooks/useAppState'
 import '../styles/tla.css'
+import { TlaDialogsProvider } from './TlaDialogsProvider'
 
 export function Component() {
 	return (
 		<AppStateProvider>
-			<Outlet />
+			<Inner />
 		</AppStateProvider>
+	)
+}
+
+function Inner() {
+	const app = useApp()
+	const theme = useValue('theme', () => app.getSessionState().theme, [app])
+
+	return (
+		<div
+			className={`tla tl-container ${theme === 'light' ? 'tla-theme__light tl-theme__light' : 'tla-theme__dark tl-theme__dark'}`}
+		>
+			<TlaDialogsProvider>
+				<Outlet />
+			</TlaDialogsProvider>
+		</div>
 	)
 }
--- a/apps/dotcom/client/package.json
+++ b/apps/dotcom/client/package.json
@@ -19,6 +19,7 @@
 		"test": "yarn run -T jest"
 	},
 	"dependencies": {
+		"@radix-ui/react-alert-dialog": "^1.1.1",
 		"@radix-ui/react-dialog": "^1.0.5",
 		"@radix-ui/react-dropdown-menu": "^2.0.6",
 		"@radix-ui/react-popover": "^1.0.7",
--- a/apps/dotcom/client/package.json
+++ b/apps/dotcom/client/package.json
@@ -19,6 +19,7 @@
 		"test": "yarn run -T jest"
 	},
 	"dependencies": {
+		"@radix-ui/react-dialog": "^1.0.5",
 		"@radix-ui/react-dropdown-menu": "^2.0.6",
 		"@radix-ui/react-popover": "^1.0.7",
 		"@sentry/integrations": "^7.34.0",
--- a/apps/dotcom/client/src/tla/components/TlaDialog.tsx/TlaDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaDialog.tsx/TlaDialog.tsx
@@ -1,0 +1,32 @@
+import * as Dialogs from '@radix-ui/react-dialog'
+import { useCallback } from 'react'
+import { dialogsAtom, TldrawAppDialog } from '../../providers/TlaDialogsProvider'
+import styles from './dialog.module.css'
+
+export function TlaDialog({ dialog }: { dialog: TldrawAppDialog }) {
+	const handleOpenChange = useCallback(
+		(open: boolean) => {
+			if (!open) {
+				dialogsAtom.update((dialogs) => dialogs.filter((d) => d.id !== dialog.id))
+				dialog.onClose?.()
+			}
+		},
+		[dialog]
+	)
+
+	return (
+		<Dialogs.Root onOpenChange={handleOpenChange} modal open>
+			<Dialogs.Overlay className={styles.overlay} />
+			<Dialogs.Content
+				className={styles.container}
+				onPointerDownOutside={() => {
+					handleOpenChange(false)
+				}}
+			>
+				<div className={styles.containerInner}>
+					<dialog.Component />
+				</div>
+			</Dialogs.Content>
+		</Dialogs.Root>
+	)
+}
--- a/apps/dotcom/client/src/tla/components/TlaDialog.tsx/dialog.module.css
+++ b/apps/dotcom/client/src/tla/components/TlaDialog.tsx/dialog.module.css
@@ -1,0 +1,33 @@
+.overlay {
+	position: absolute;
+	inset: 0px;
+	background-color: var(--tla-color-overlay);
+	z-index: 2000;
+}
+
+.container {
+	outline: none;
+	display: flex;
+	align-self: center;
+	justify-self: center;
+	align-items: center;
+	justify-content: center;
+	pointer-events: none;
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%, -50%);
+	z-index: 2200;
+}
+
+.containerInner {
+	position: relative;
+	width: 100%;
+	max-width: 360px;
+	height: fit-content;
+	padding: 12px;
+	z-index: 2200;
+	color: var(--tla-color-text-1);
+	background-color: var(--tla-color-panel);
+	pointer-events: all;
+}
--- a/apps/dotcom/client/src/tla/components/TlaDialog.tsx/dialog.module.css
+++ b/apps/dotcom/client/src/tla/components/TlaDialog.tsx/dialog.module.css
@@ -3,6 +3,16 @@
 	inset: 0px;
 	background-color: var(--tla-color-overlay);
 	z-index: 2000;
+	animation: fade-in forwards 0.12s;
+}
+
+@keyframes fade-in {
+	0% {
+		opacity: 0;
+	}
+	100% {
+		opacity: 1;
+	}
 }
 
 .container {
@@ -16,18 +26,21 @@
 	position: absolute;
 	top: 50%;
 	left: 50%;
+	width: 100%;
+	max-width: 360px;
 	transform: translate(-50%, -50%);
 	z-index: 2200;
 }
 
 .containerInner {
 	position: relative;
-	width: 100%;
-	max-width: 360px;
 	height: fit-content;
+	width: 100%;
 	padding: 12px;
 	z-index: 2200;
 	color: var(--tla-color-text-1);
 	background-color: var(--tla-color-panel);
 	pointer-events: all;
+	border-radius: 9px;
+	box-shadow: var(--tla-shadow-3);
 }
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/TlaSidebar.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/TlaSidebar.tsx
@@ -13,6 +13,7 @@ import { getFileUrl, getShareableFileUrl } from '../../utils/urls'
 import { TlaAvatar } from '../TlaAvatar/TlaAvatar'
 import { TlaIcon, TlaIconWrapper } from '../TlaIcon/TlaIcon'
 import { TlaSpacer } from '../TlaSpacer/TlaSpacer'
+import { TlaRenameFileDialog } from '../dialogs/TlaRenameFileDialog'
 import styles from './sidebar.module.css'
 
 export function TlaSidebar() {
@@ -234,12 +235,10 @@ function TlaSidebarFileLinkMenu({ fileId }: { fileId: TldrawAppFile['id'] }) {
 			...d,
 			{
 				id: 'rename',
-				Component: () => {
-					return <div>hello!!!!!!</div>
-				},
+				Component: () => <TlaRenameFileDialog fileId={fileId} />,
 			},
 		])
-	}, [])
+	}, [fileId])
 
 	const handleDuplicateLinkClick = useCallback(() => {
 		// duplicate file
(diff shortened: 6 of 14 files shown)


packages/tldraw/src/lib/ui/hooks/useMenuIsOpen.ts
steveruizok, line 1: The `useMenuIsOpen` in tldraw now uses the `useGlobalMenuIsOpen`, but handles some callbacks with ui events.

packages/tldraw/api-report.md
```diff
@@ -2149,9 +2148,6 @@ export function TldrawUiMenuCheckboxItem<TranslationKey extends string = string,
 // @public (undocumented)
 export function TldrawUiMenuContextProvider({ type, sourceId, children, }: TLUiMenuContextProviderProps): JSX_2.Element;
 
-// @public (undocumented)
-export type TldrawUiMenuContextType = 'context-menu' | 'helper-buttons' | 'icons' | 'keyboard-shortcuts' | 'menu' | 'panel' | 'small-icons' | 'toolbar-overflow' | 'toolbar';
```
steveruizok, line 2153: There were a few types here that should have been `TLUi` instead of `TldrawUi`.

packages/tldraw/src/lib/ui/components/Toasts.tsx
steveruizok, line 1: For toasts and dialogs, I've modified a little so that the returned value from the hook is stable. The array of dialogs / toasts is now an atom. There's only one place where we actually use either, so it's an improvement over a footgun where you'd do something like this:

```tsx
const dialogs = useDialogs()

const handleClick = useCallback(() => {
  dialogs.addDialog({ ... })
}, [dialogs])
```

Previously, modifying the dialogs would cause the context to update. Now it does not, so the `dialogs` in the dependency array is okay.

packages/tldraw/src/lib/ui/components/Toasts.tsx
```diff
@@ -80,36 +82,16 @@ function Toast({ toast }: { toast: TLUiToast }) {
...
-				setHasToasts(false)
-			}, 1000)
-		}
-
-		return () => {
-			clearTimeout(timeoutId)
-		}
-	}, [toasts.length, setHasToasts, editor])
-
-	if (!hasToasts) return null
```
steveruizok, line 112: this is very much legacy from a time when I was trying to animate toasts out of the viewport

packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
```diff
@@ -23,6 +22,7 @@ export interface TLUiButtonPickerProps<T extends string> {
 	items: StyleValuesForUi<T>
 	theme: TLDefaultColorTheme
 	onValueChange(style: StyleProp<T>, value: T): void
+	onHistoryMark?(id: string): void
```
steveruizok, line 25: These components no longer depend on the editor, so I've extracted the history-effecting things and made them optional

packages/tldraw/src/lib/ui/context/TldrawUiContextProvider.tsx
```diff
@@ -60,23 +60,27 @@ export function TldrawUiContextProvider({
...
 									</TldrawUiComponentsProvider>
 								</BreakPointProvider>
-							</DialogsProvider>
-						</ToastsProvider>
-					</UiEventsProvider>
-				</TranslationProvider>
+							</TldrawUiDialogsProvider>
+						</TldrawUiToastsProvider>
+					</TldrawUiEventsProvider>
+				</TldrawUiTranslationProvider>
```
steveruizok, line 83: Renamed now that they're public

packages/tldraw/src/lib/ui/context/events.tsx
```diff
@@ -140,7 +140,7 @@ export interface EventsProviderProps {
 }
 
 /** @public @react */
-export function UiEventsProvider({ onEvent, children }: EventsProviderProps) {
+export function TldrawUiEventsProvider({ onEvent, children }: EventsProviderProps) {
```
steveruizok, line 143: Fixed the name here

packages/tldraw/src/lib/utils/tldr/__snapshots__/buildFromV1Document.test.ts.snap
steveruizok, line 1: The ids here are counted up from `uniqueId` calls. There's a new one in the editor (`contextId`) so that's why I've had to update the snapshot.

packages/utils/src/lib/timers.ts
steveruizok, line 1: Now that timers needs to be a global, rather than every editor having its own timers instance, I'd had to add a kind of `contextId` mapping. There might be a nicer way to do this (i.e. have a global map of Timers instead of a global timer with maps)

apps/dotcom/client/src/tla/components/TlaFileContent/TlaFileContent.tsx
steveruizok, line 1: Add file menu, in-place file renaming

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenu.tsx
steveruizok, line 1: use tldrawui stuff for the menu

apps/dotcom/client/src/tla/components/dialogs/TlaRenameFileDialog.tsx
steveruizok, line 1: add a dialog for file renames from menus

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
steveruizok, line 1: Lots more context providers here

packages/utils/src/lib/timers.ts
```diff
@@ -2,39 +2,67 @@
...
+		this.timeouts.set(contextId, [...current, id])
 		return id
 	}
 
 	/** @public */
-	setInterval(handler: TimerHandler, timeout?: number, ...args: any[]): number {
+	setInterval(contextId: string, handler: TimerHandler, timeout?: number, ...args: any[]): number {
 		const id = window.setInterval(handler, timeout, args)
-		this.intervals.push(id)
+		const current = this.timeouts.get(contextId) ?? []
```
MitjaBezensek, line 27:
```suggestion
		const current = this.intervals.get(contextId) ?? []
```

packages/utils/src/lib/timers.ts
```diff
@@ -2,39 +2,67 @@
...
+		this.intervals.set(contextId, [...current, id])
 		return id
 	}
 
 	/** @public */
-	requestAnimationFrame(callback: FrameRequestCallback): number {
+	requestAnimationFrame(contextId: string, callback: FrameRequestCallback): number {
 		const id = window.requestAnimationFrame(callback)
-		this.rafs.push(id)
+		const current = this.timeouts.get(contextId) ?? []
```
MitjaBezensek, line 35:
```suggestion
		const current = this.rafs.get(contextId) ?? []
```

packages/tldraw/src/lib/ui/overrides.ts
```diff
@@ -107,6 +96,7 @@ function useShallowArrayEquality<T extends unknown[]>(array: T): T {
 	return useMemo(() => array, array)
 }
 
+/**	@public */
```
MitjaBezensek, line 99:
```suggestion
/** public */
```

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1472,89 +1475,7 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-	 * certain interactions will behave differently; for example, when a
-	 * draw tool is selected and a menu is open, a pointer-down will not
-	 * create a dot (because the user is probably trying to close the menu)
-	 * however a pointer-down event followed by a drag will begin drawing
-	 * a line (because the user is BOTH trying to close the menu AND start
-	 * drawing a line).
-	 *
-	 * @public
-	 */
-	@computed getOpenMenus(): string[] {
```
MitjaBezensek, line 1486: We could make this a non breaking change by callin into `this.menus.getOpenMenus` from here? Similar with other similar changes.

packages/editor/src/lib/globals/menus.ts
```diff
@@ -0,0 +1,119 @@
...
+	 * certain interactions will behave differently; for example, when a
+	 * draw tool is selected and a menu is open, a pointer-down will not
+	 * create a dot (because the user is probably trying to close the menu)
+	 * however a pointer-down event followed by a drag will begin drawing
+	 * a line (because the user is BOTH trying to close the menu AND start
+	 * drawing a line).
+	 *
+	 * @public
+	 */
+	menus: atom<string[]>('open menus', []),
```
MitjaBezensek, line 16: Why not a map here, similar to timers?

packages/editor/src/lib/globals/menus.ts
```diff
@@ -0,0 +1,119 @@
...
+	 * @example
+	 * ```ts
+	 * getIsMenuOpen()
+	 * ```
+	 *
+	 * @param contextId - An optional context to check menus for.
+	 *
+	 * @public
+	 */
+	getIsMenuOpen(contextId?: string): boolean {
```
MitjaBezensek, line 105: Not really related to this pr, but `getIsAnyMenuOpen` would be a better name here. With the current name I'd expect to pass in a menu `id` and get the status just for that menu.

packages/editor/src/lib/hooks/useGlobalMenuIsOpen.ts
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		return () => {
+			if (rIsOpen.current) {
+				// Close menu on unmount
+				tlmenus.deleteOpenMenu(id)
+
+				// Close menu and all submenus when the parent is closed
+				tlmenus.getOpenMenus().forEach((menuId) => {
+					if (menuId.startsWith(id)) {
```
MitjaBezensek, line 51: Feels like we should be more exact here? Otherwise we might close some poorly named menus (like `context-menu` and `context-menu-2`)

packages/tldraw/src/lib/ui/components/Toasts.tsx
```diff
@@ -80,36 +82,16 @@ function Toast({ toast }: { toast: TLUiToast }) {
...
-
-	return toasts.map((toast) => <Toast key={toast.id} toast={toast} />)
-}
-
-export const Toasts = React.memo(_Toasts)
-
-export function ToastViewport() {
-	const editor = useEditor()
+/** @public @react */
+export const TldrawUiToasts = memo(function TldrawUiToasts() {
```
MitjaBezensek, line 86: Should unify the naming here. We use `TldrawUiToasts`, but then we use just `Toast` above. Might also make sense to update the filename to reflect the change?

packages/tldraw/src/lib/ui/hooks/useReadonly.ts
```diff
@@ -1,7 +1,7 @@
-import { useEditor, useValue } from '@tldraw/editor'
+import { useMaybeEditor, useValue } from '@tldraw/editor'
 
 /** @public */
 export function useReadonly() {
-	const editor = useEditor()
-	return useValue('isReadonlyMode', () => editor.getInstanceState().isReadonly, [editor])
+	const editor = useMaybeEditor()
+	return useValue('isReadonlyMode', () => editor?.getInstanceState().isReadonly, [editor])
```
MitjaBezensek, line 6: I'd prefer this to still return bool, now it's `boolean | undefined`

packages/tldraw/src/lib/ui/components/Dialogs.tsx
MitjaBezensek, line 1: Same file name change consideration here: `TldrawUIDialogs`?

packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
MitjaBezensek: They feel quite specific to tldraw so it feels like we might keep the editor. Or do we have some other uses where editor is not available?

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
```diff
@@ -74,14 +74,22 @@ export const TldrawUiInput = React.forwardRef<HTMLInputElement, TLUiInputProps>(
...
-						elm.select()
-					}
-				})
+				if (editor) {
+					editor.timers.requestAnimationFrame(() => {
+						if (autoSelect) {
+							elm.select()
+						}
+					})
+				} else {
```
MitjaBezensek, line 83: Not sure I like the `useMaybeEditor` pattern. Think I'd prefer passing in a prop called `raf` here. The only problem is how to communicate that to the end users.

packages/tldraw/src/lib/ui/components/primitives/TldrawUiSlider.tsx
```diff
@@ -11,13 +10,20 @@ export interface TLUiSliderProps {
 	label: string
 	title: string
 	onValueChange(value: number): void
+	onHistoryMark(id: string): void
```
MitjaBezensek, line 13: Same question here. `onHistoryMark` is so editor specific, do we see any uses here outside of editor. If nothing else then I'd at least rename it.

packages/tldraw/src/lib/ui/components/primitives/menus/TldrawUiMenuContext.tsx
```diff
@@ -29,7 +29,7 @@ export function useTldrawUiMenuContext() {
 
 /** @public */
 export interface TLUiMenuContextProviderProps {
-	type: TldrawUiMenuContextType
+	type: TLUiMenuContextType
```
MitjaBezensek, line 32: I guess it would make sense to unify the naming. `TLUiMenuContextProviderProps` for props and `TldrawUiMenuContextProvider` for the component. Feels weird.

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
```diff
@@ -1,11 +1,68 @@
...
+
+function InnerInner() {
+	const handleAppLevelUiEvent = useCallback<TLUiEventHandler>(() => {
+		// todo, implement handling ui events at the application layer
+	}, [])
+
+	return (
+		<AssetUrlsProvider assetUrls={assetUrls}>
+			<TldrawUiEventsProvider onEvent={handleAppLevelUiEvent}>
+				<TldrawUiTranslationProvider locale="en">
```
MitjaBezensek, line 56: Do we need this here? We'll probably have a different setup for app translations. We can then also keep the editor inside the provider. We also wouldn't needs `AssetUrlsProvider` then.  Unless we are reusing some of the translations?

packages/editor/src/lib/globals/menus.ts
```diff
@@ -0,0 +1,119 @@
...
+	 * ```ts
+	 * clearOpenMenus()
+	 * clearOpenMenus(myEditorId)
+	 * ```
+	 *
+	 * @param contextId - An optional context to clear menus for.
+	 *
+	 * @public
+	 */
+	clearOpenMenus(contextId?: string) {
```
MitjaBezensek, line 89: Same as above.

packages/editor/src/lib/globals/menus.ts
```diff
@@ -0,0 +1,119 @@
...
+		this.menus.set(contextId ? this.menus.get().filter((m) => !m.endsWith('-' + contextId)) : [])
```
MitjaBezensek, line 90: So calling `clearOpenMenus` without the `contextId` should clear all the the menus?

packages/utils/src/lib/timers.ts
steveruizok: this was an oversight but ironically iirc all the cleanTimeout / cancelIntervals are interchangeable

packages/editor/src/lib/editor/Editor.ts
steveruizok: That's true, and deprecate?

packages/editor/src/lib/globals/menus.ts
steveruizok: could do that

packages/editor/src/lib/hooks/useGlobalMenuIsOpen.ts
steveruizok: the map would solve this, let's do the map

packages/tldraw/src/lib/ui/components/Toasts.tsx
steveruizok: We could, the naming only really matters for exported things (we don't export Toast) but as a convention we might as well

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
steveruizok: I had a raf prop there before but it was tedious. I'd rather the ugly code be in the component I guess

packages/tldraw/src/lib/ui/components/primitives/TldrawUiSlider.tsx
steveruizok: maybe? If there's an app level undo/redo system? we could do the useMaybeEditor instead though

packages/tldraw/src/lib/ui/components/primitives/menus/TldrawUiMenuContext.tsx
steveruizok: The naming is intentionally separated between just types (TLUi) and components (TldrawUi). There are just a ton of exports from tldraw, so this was my way of wrangling the proliferation / difference

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
steveruizok: That's true. I added translations here from a "might as well" philosophy. The assets are important however for icons that are used in the UI components.

MitjaBezensek: I don't think we need assets. Tldraw will add the provider anyway, so we are kinda nesting providers.

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
MitjaBezensek, line 1: This is using `useTranslation`. Probably fine for now (if we keep the double translations provider as mentioned in the other comment), but we might want to pass in the label as a string instead and do the translations on the caller side.

MitjaBezensek: Noticed the same thing in `TldrawUIMenuGroup` and `TldrawUIMenuItem`. If we get rid of those three we can remove the assets provider, translation provider in `TlaAppProvider` and revert the locale change in the translations provider.

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
steveruizok: Without assets, the TldrawUiIcon component fails. Its ok to nest react contexts (we'll already be doing it for toasts, dialogs, etc, too)

packages/editor/src/lib/hooks/useGlobalMenuIsOpen.ts
steveruizok: My mistake, this is more to do with nested menus from the UI, ie when we close child menus when the parent menu closes. We should probably track nested menus in a better way, but this is how we've done it before.

packages/editor/src/lib/globals/menus.ts
steveruizok: Hmm, seems easier to keep this flat. Let's come back to this later and pick a nicer structure.

steveruizok: (the "front door" shouldn't change)

apps/dotcom/client/src/tla/providers/TlaAppProvider.tsx
MitjaBezensek: `TldrawUiIcon` is not used from tla at the moment. And for the usages in the tldraw package we already use the assets provider there.

PR #4627, opened 2024-09-30
Fix watermark link opening twice when clicking on the watermark.

MitjaBezensek:
Fix the link opening twice in FF.

Related #4622

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Start the app.
2. Click on the watermark. It should only open one new tab.
3. Try different browsers.

### Release notes

- Fix a bug with watermark link opening twice in FF.

### DIFF
--- a/packages/editor/src/lib/license/Watermark.tsx
+++ b/packages/editor/src/lib/license/Watermark.tsx
@@ -2,7 +2,7 @@ import { useQuickReactor, useValue } from '@tldraw/state-react'
 import { memo, useState } from 'react'
 import { useCanvasEvents } from '../hooks/useCanvasEvents'
 import { useEditor } from '../hooks/useEditor'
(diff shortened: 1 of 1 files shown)


vladh:
Hey @MitjaBezensek, sorry for not doing more testing on my previous PR.

In retrospect, I think a different strategy is needed here. While the solution in your PR works on Firefox for me, it does not work on mobile (where clicking does nothing), and it does not support keyboard navigation which is an accessibility issue.

I'll think about this more and get back to you. I wonder if there's a way to only apply this hack to VS Code users, and leave everyone else with a perfectly normal `<a>` tag?

vladh:
@MitjaBezensek There's a better solution in #4622 now!

MitjaBezensek:
Closing in favor of the mentioned pr.

PR #4632, opened 2024-10-01
text: followup fix to locking text shapes

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/4569
Once the text shape is locked you can't click the Select tool to switch out. This addresses that (because the tool is masked at the time). If there's a better fix for this, I'm all ears, I wasn't sure how else to resolve this (I see you pulled it out in the other PR @steveruizok )

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix bug with text shape locking.


steveruizok:
Oh I was sure that the deep transition approach had worked.

mimecuvalo:
Yeah, maybe there's a deeper bug to address here?

steveruizok:
Ok, looks like the correct behavior. Pushing a small modification that only does the exit/enter here when in the select state.

PR #4636, opened 2024-10-01
Improve perf for safari

MitjaBezensek:
Looks like safari has some issue with using masks. This PR switches from using masks to clipPaths. 
### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create some arrows.
2. Add labels to some of them.
3. Zoom in signficantly (especially in Safari) and pan around. You should not see a drop in perf. You can turn on perf measurement in the debug panel to see the fps for panning.

### Release notes

- Fix a performance issue with panning when zoomed in on arrows in Safari.

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -28,7 +28,6 @@ import {
 	mapObjectMapValues,
 	sanitizeId,
 	structuredClone,
-	tlenv,
 	toDomPrecision,
 	track,
 	useEditor,
@@ -63,8 +62,6 @@ import {
 	getArrowTerminalsInArrowSpace,
 	removeArrowBinding,
 } from './shared'
-
-let globalRenderIndex = 0
 
 enum ARROW_HANDLES {
 	START = 'start',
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -935,40 +927,26 @@ const ArrowSvg = track(function ArrowSvg({
 	const maskStartArrowhead = !(info.start.arrowhead === 'none' || info.start.arrowhead === 'arrow')
 	const maskEndArrowhead = !(info.end.arrowhead === 'none' || info.end.arrowhead === 'arrow')
 
-	// NOTE: I know right setting `changeIndex` hacky-as right! But we need this because otherwise safari loses
-	// the mask, see <https://linear.app/tldraw/issue/TLD-1500/changing-arrow-color-makes-line-pass-through-text>
-	const maskId = sanitizeId(shape.id + '_clip' + (tlenv.isSafari ? `_${changeIndex}` : ''))
```
MitjaBezensek, line 940: Looks like this does not apply to `clipPaths`, which is a nice win on top of fixing perf.

SomeHats (approved):
this is a great fix, ty!

PR #4639, opened 2024-10-01
[botcom] Use auth on backend

ds300:
This PR

- Renames the socket endpoint `/app/:userId` to just `/app`, and uses the access token to get the user id from clerk
- Uses the access token to gate 'owned' files.

TO DO in follow ups
- Allow shared files (will require asking the owner's DO for permission to establish a socket connection, and figuring out a way to revoke an already-established connection.
- Add web hook to update a user's info if they change it on google/whatever.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`


apps/dotcom/sync-worker/package.json
```diff
@@ -15,7 +15,7 @@
 		"test-ci": "lazy inherit",
 		"test": "yarn run -T jest",
 		"test-coverage": "lazy inherit",
-		"check-bundle-size": "yarn run -T tsx ../../../internal/scripts/check-worker-bundle.ts --entry src/worker.ts --size-limit-bytes 350000",
+		"check-bundle-size": "yarn run -T tsx ../../../internal/scripts/check-worker-bundle.ts --entry src/worker.ts --size-limit-bytes 450000",
```
ds300, line 18: this was due to importing clerk on the room Durable Object.

apps/dotcom/client/src/components/StoreErrorScreen.tsx
```diff
@@ -47,6 +47,11 @@ export function StoreErrorScreen({ error }: { error: Error }) {
 				message = 'The room you are trying to connect to does not exist.'
 				break
 			}
+			case TLIncompatibilityReason.NotAuthorized: {
+				header = 'Not authorized'
+				message = 'You are not authorized to view this room.'
```
mimecuvalo, line 52: Nitpick on language: "You need to be authorized to view this room." (missing authentication in this case, not authentication that's forbidden)

apps/dotcom/client/src/components/StoreErrorScreen.tsx
```diff
@@ -47,6 +47,11 @@ export function StoreErrorScreen({ error }: { error: Error }) {
 				message = 'The room you are trying to connect to does not exist.'
 				break
 			}
+			case TLIncompatibilityReason.NotAuthorized: {
```
mimecuvalo, line 50: We should prbly also have a Forbidden (403), might as well think ahead to that case if you like.

apps/dotcom/client/src/routes.tsx
```diff
@@ -5,32 +5,37 @@ import {
...
+			if (error instanceof TLRemoteSyncError) {
+				switch (error.reason) {
+					case TLIncompatibilityReason.RoomNotFound: {
+						header = 'Not found'
+						para1 = 'The file you are looking for does not exist.'
+						break
+					}
+					case TLIncompatibilityReason.NotAuthorized: {
+						header = 'Not authorized'
+						para1 = 'You are not authorized to view this file.'
```
mimecuvalo, line 33: ditto

apps/dotcom/client/src/components/StoreErrorScreen.tsx
mimecuvalo: actually, also it should be 'Unauthorized' instead of 'Not authorized' for the same reason.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -207,8 +210,19 @@ export function TlaEditor({
...
+	const user = useTldrawUser()
+
 	const store = useSync({
-		uri: `${MULTIPLAYER_SERVER}/app/file/${fileSlug}`,
+		uri: useCallback(async () => {
+			const url = new URL(`${MULTIPLAYER_SERVER}/app/file/${fileSlug}`)
+			if (user) {
+				url.searchParams.set('accessToken', await user.getToken())
+			}
+			if (temporary) {
```
mimecuvalo, line 221: I feel like when I see it say 'temporary' it makes me think the room will disappear after a certain time period. Would 'sandbox' / 'playground' be better options in general?

packages/sync/src/useSync.ts
```diff
@@ -136,6 +136,11 @@ export function useSync(opts: UseSyncOptions & TLStoreSchemaOptions): RemoteTLSt
 				client.close()
 				socket.close()
 				return
+			} else if (val === 'error' && closeCode === TLCloseEventCode.NOT_AUTHORIZED) {
```
mimecuvalo, line 139: same comment here on adding logic here in the future to handle asking the user to login, instead of just closing the connection

apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -42,15 +43,18 @@ const router = createRouter<Environment>()
...
+		if (!auth?.userId) return notFound()
 
 		// This needs to be a websocket request!
 		if (req.headers.get('upgrade')?.toLowerCase() === 'websocket') {
 			// Set up the durable object for this room
-			const id = env.TLAPP_DO.idFromName(userName)
-			return env.TLAPP_DO.get(id).fetch(req)
+			const id = env.TLAPP_DO.idFromName(auth.userId)
+			const url = new URL(req.url)
+			url.pathname = `/app/${auth.userId}`
```
mimecuvalo, line 55: it'd be great (in a followup is fine) to have a `urlFactory` place where the `/app` and related paths are in one place, we'll start to lose track.

`dotcom-shared/src/routes.ts` would make sense for this.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -298,6 +325,29 @@ export class TLDrawDurableObject {
 		const { 0: clientWebSocket, 1: serverWebSocket } = new WebSocketPair()
 		serverWebSocket.accept()
 
+		const closeSocket = (code: number, message: string) => {
+			serverWebSocket.close(code, message)
+			return new Response(null, { status: 101, webSocket: clientWebSocket })
```
mimecuvalo, line 330: i see this pattern is in other code files but it's surprising to me that after the `close` we send a 101, is that the correct status code for this situation?

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -298,6 +325,29 @@ export class TLDrawDurableObject {
...
+		}
+
+		if (this.isApp) {
+			const ownerId = await this.getOwnerId()
+
+			if (ownerId) {
+				const auth = await requireAuth(req, this.env)
+				if (ownerId !== auth.userId) {
```
mimecuvalo, line 338: ach, i was afraid i'd get to this point in the code and see this heh :)

i'm gonna be pendantic here and say that the wording here shouldn't be `NOT_AUTHORIZED` but should be `FORBIDDEN`.

it's the 401 vs. 403 distinction. it would be good to keep the lingo on this consistent with the HTTP wording because otherwise people will trip up on this distinction.

mimecuvalo (commented):
nice one! main comment is on the 401/403 type wording of things (unauthorized vs. forbidden)

apps/dotcom/client/src/components/StoreErrorScreen.tsx
ds300: good suggestions

apps/dotcom/client/src/components/StoreErrorScreen.tsx
ds300: this is actually for what '403 forbidden' means in HTTP (i.e. not authorized). it's confusing because http calls 401 'unauthorized' but really it means 'not authenticated'. 

still yeah, we might need to distinguish at some point. let's take care of it then.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
ds300: you can change the name later if you like. make sure you bump the documentInfo version number in the durable object. i'm fine with any of these.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: it might still work with other status codes? you can try it. but http status codes aren't exposed on the client for websocket requests, for some security reason. that's why we pass a code to the close method instead. so it's not helpful beyond making this code look more like normal http code

apps/dotcom/client/src/components/StoreErrorScreen.tsx
ds300: forget this, i read your comments out of order

PR #4642, opened 2024-10-02
menus: rework the open menu logic to be in one consistent place

mimecuvalo:
We have a lot of logic scattered everywhere to prevent certain logic when menus are open. It's a very manual process, easy to forget about when adding new shapes/tools/logic. This flips the logic a bit to be handled in one place vs. various places trying to account for this.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Rework open menu logic to be centralized.


packages/editor/src/lib/components/default-components/DefaultInFrontOfTheCanvas.tsx
```diff
@@ -0,0 +1,40 @@
+import { useValue } from '@tldraw/state-react'
+import { useEffect } from 'react'
+import { useEditor } from '../../hooks/useEditor'
+
+export const DefaultInFrontOfTheCanvas = () => {
```
SomeHats, line 5: (just passing by here, but given this is very often overridden by devs, is there a way we can either make sure they know to still include it or preferably include it in a different place so it won't get overridden by devs trying to make on-canvas UI? Maybe we put it in `DefaultCanvas` instead?)

mimecuvalo: yeah, fair enough!

i put it in `DefaultCanvas` — if you think it's something that should be overridable i think that could be alright too but i didn't that to start.

mimecuvalo:
@SomeHats or @steveruizok this ok to land?

steveruizok:
Let me review again, if you've handled that weird case with drawing then yes, I think so.

steveruizok:
That issue is still there, let's fix it. I think I know how. I would rather not give that up!

steveruizok:
Will target this with a PR

steveruizok:
Hm, not sure why this isn't picking up the mention, but here's that PR. https://github.com/tldraw/tldraw/pull/4731

Short story, we use the menu blocker as a temporary canvas if the user starts clicking and dragging while it's open.

PR #4644, opened 2024-10-02
text: followup to the followup of fix to locking text shapes

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/4632 (which is a followup to https://github.com/tldraw/tldraw/pull/4569)

@steveruizok i saw you had removed this line but it's necessary because otherwise the Select tool doesn't work when switching to it... (clicking on shapes doesn't do anything)

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix bug with text shape locking.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useTools.tsx
+++ b/packages/tldraw/src/lib/ui/hooks/useTools.tsx
@@ -63,6 +63,7 @@ export function ToolsProvider({ overrides, children }: TLUiToolsProviderProps) {
 						// text tool.
 						const currentNode = editor.root.getCurrent()!
 						currentNode.exit({}, currentNode.id)
+						currentNode.enter({}, currentNode.id)
 					}
 					editor.setCurrentTool('select.idle')
 					trackEvent('select-tool', { source, id: 'select' })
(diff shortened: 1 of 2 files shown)


steveruizok:
im losing my mind, this does not make sense to me

steveruizok:
Sorry, we really shouldn't have to be driving the state chart by hand here.

I think the clearest expression of what we want to happen is:

```ts
{
	id: 'select',
	label: 'tool.select',
	icon: 'tool-pointer',
	kbd: 'v',
	readonlyOk: true,
	onSelect(source) {
		editor.setEditingShape(null)
		editor.setCroppingShape(null)
		editor.setCurrentTool('select')
		trackEvent('select-tool', { source, id: 'select' })
	},
}
```

steveruizok:
Spending more time with this... I think this is a fundamental issue with the state chart. I'm going to add tests and let's land this PR, then I'm going to create a follow up PR with the change to `StateNode.transition` that should solve it more generally.

mimecuvalo:
cc @ds300 for your thoughts on this as well

steve had said this offline:
> We can solve this slightly more clearly by clearing the editing / cropping shapes, but then it would still happen for other substates later on
> the enter / exit trick seems to work
> I think the state chart should support that kind of "reset" or "parent transition"

PR #4646, opened 2024-10-02
Arrowhead clipping fix

MitjaBezensek:
Fixes issues with arrowheads not correctly clipping the arrow's body. The main issue was when we used the fill none option. There was also a difference in how it worked if arrows had labels vs if they didn't.

Also transitions the arrow indicator form using masks to using clip paths so that we also improve the performance of panning when we have arrows selected.

[Related discussion](https://discord.com/channels/859816885297741824/1290288009492824065/1290906937797640273).

### Before

image

### After

image

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix an issue introduced with #4636. The arrowheads did not correctly clip the arrow body.


packages/tldraw/src/lib/shapes/arrow/arrowheads.ts
```diff
@@ -55,7 +55,7 @@ function getTriangleHead({ point, int }: TLArrowPointsInfo) {
 	const PL = Vec.RotWith(int, point, PI / 6)
 	const PR = Vec.RotWith(int, point, -PI / 6)
 
-	return `M ${PL.x} ${PL.y} L ${point.x} ${point.y} L ${PR.x} ${PR.y} Z`
+	return `M ${PL.x} ${PL.y} L ${PR.x} ${PR.y} L ${point.x} ${point.y} Z`
```
MitjaBezensek, line 58: Had to reorder these to change the direction.

packages/tldraw/src/lib/shapes/arrow/arrowheads.ts
```diff
@@ -83,7 +83,7 @@ function getDiamondHead({ point, int }: TLArrowPointsInfo) {
 	const PQ = Vec.Lrp(PL, PR, 0.5)
 	PQ.add(Vec.Sub(PQ, point))
 
-	return `M ${PQ.x} ${PQ.y} L ${PL.x} ${PL.y} ${point.x} ${point.y} L ${PR.x} ${PR.y} Z`
+	return `M ${PQ.x} ${PQ.y} L ${PR.x} ${PR.y} ${point.x} ${point.y} L ${PL.x} ${PL.y} Z`
```
MitjaBezensek, line 86: Same here.

steveruizok:
Good fix! This must have been a recent regression, these were working correctly not too long ago.

steveruizok:
Also, do we have these in our e2e tests? I would have expected the snapshots to catch the differences.

MitjaBezensek:
> Good fix! This must have been a recent regression, these were working correctly not too long ago.

I actually caused this in a previous PR #4636 😓

MitjaBezensek:
> Also, do we have these in our e2e tests? I would have expected the snapshots to catch the differences.

Will take a look.

apps/examples/e2e/tests/export-snapshots.spec.tsx
```diff
@@ -487,6 +487,50 @@ const snapshots: Snapshots = {
 			),
 		},
 	},
+	Arrows: {
```
MitjaBezensek, line 490: Hm, now I'm not sure sure if this will catch the issue. We already have a snapshot test for Fills, which does include arrows. I ran it locally and the snapshot definitely changed, but the tests didn't detect it. I guess the change was not significant enough.

MitjaBezensek: Made the arrows `xl` now which should detect the change. Will keep this test instead of updating the fills one though.

PR #4647, opened 2024-10-02
Prevent some draw shape rerenders when changing the zoom

MitjaBezensek:
Any zoom change caused a rerender of draw shapes. Now we will only rerender when the conditions for the change in `forceSolid` or `dotAdjustment` are met.

[Related conversation](https://discord.com/channels/859816885297741824/1290589596417003571/1290963510963994626).

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a draw shape.
2. Change the zoom.
3. We shouldn't see renrenders for every change, only at certain zoom level changes (like going for 10 -> 20%)

### Release notes

- Reduce the number of times the draw shape has to rerender due to the zoom changes.


ds300 (approved):
nice! should be super noticeable on mobile/low-end machines.

PR #4649, opened 2024-10-02
Fix parent transitions

steveruizok:
While working on #4644, we ran into this curious problem where transitioning from a child state (like `select.editing_shape`) to a parent state (like `select`) did not cause the `select.editing_shape` to exit. It turns out there were a few issues with the state chart:

  1. Parent transitions had no effect on active children, were basically noops
  2. Transitioning to a partial branch did not exit the lower branch. For example, `a.b.c` to `a.x` did not exit `c`.

This PR aims to solve all that.

Everything seems to work alright, however there are a few things discovered along the way:
- creating transitions in a node's `onExit` adds a lot of complexity
- the side effects we've created around `editingShape` and `croppingShape` add a lot of complexity

In this PR, transitions to parents (e.g. from `a.b.c` to `a`) will exit the whole tree and re-enter it.

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed several rare bugs with state transitions.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -720,11 +720,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 			throw Error(`No state found for initialState "${initialState}".`)
 		}
 
-		this.root.enter(undefined, 'initial')
-
 		this.edgeScrollManager = new EdgeScrollManager(this)
 		this.focusManager = new FocusManager(this, autoFocus)
 		this.disposables.add(this.focusManager.dispose.bind(this.focusManager))
+
+		this.root.start()
 
 		if (this.getInstanceState().followingUserId) {
 			this.stopFollowingUser()
@@ -1380,16 +1380,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	getStateDescendant<T extends StateNode>(path: string): T | undefined {
-		const ids = path.split('.').reverse()
-		let state = this.root as StateNode
-		while (ids.length > 0) {
-			const id = ids.pop()
-			if (!id) return state as T
-			const childState = state.children?.[id]
-			if (!childState) return undefined
-			state = childState
-		}
-		return state as T
+		return this.root.getDescendant(path)
 	}
 
 	/* ---------------- Document Settings --------------- */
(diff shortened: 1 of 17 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -720,12 +720,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 			throw Error(`No state found for initialState "${initialState}".`)
 		}
 
-		this.root.enter(undefined, 'initial')
-
 		this.edgeScrollManager = new EdgeScrollManager(this)
 		this.focusManager = new FocusManager(this, autoFocus)
 		this.disposables.add(this.focusManager.dispose.bind(this.focusManager))
 
+		this.root.start()
```
steveruizok, line 727: Previously this was an `enter()`, but this does enough different that it should be its own method.

packages/editor/src/lib/editor/tools/StateNode.ts
```diff
@@ -129,39 +130,138 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
...
+			const childState = state.children?.[id]
+			if (!childState) return undefined
+			state = childState
+		}
+		return state as T
+	}
+
+	private exitBranch(id: string, info: any = {}) {
+		let nextExitingParent = this as StateNode | undefined
+		while (nextExitingParent) {
```
mimecuvalo, line 148: a comment would be good for this bit of logic, hurts my brain a bit

mimecuvalo: i think it's because i usually think of exiting the branch upwards,  it exits the parent, but then the logic references the `nextExitingChild` so it makes it look like it's going downwards 🙃

mimecuvalo (commented):
This looks good to me in general. I think @ds300 should give it a read too - I haven't delved into this bit as much yet.

packages/editor/src/lib/editor/tools/StateNode.ts
steveruizok: I believe that's right, we're disabling the branch here downwards (parent > child) when we exit an active branch. I don't know whether there's a strong convention here, but we could try either way and see if any tests fail?

ds300:
Closing due to inactivity

PR #4650, opened 2024-10-03
Fix an issue with nearest point and lines that start and end at the same point

MitjaBezensek:
This fixes an issue with getting the nearest point to a line for lines that start and end at the same point. 

Not sure how those lines got created though. Programatically we seem to allow that, so I also added a `onBeforeHandler` to line shape util that will nudge the end point just slightly if we try to create a line like that. We could also avoid creating it completely?

[Example sentry report](https://tldraw.sentry.io/issues/5936469482/?alert_rule_id=12855294&alert_timestamp=1727804797951&alert_type=email&environment=production&notification_uuid=3482b5d8-ad95-48ca-be09-40c77af5fcf2&project=4504203639193600&referrer=alert_email) for this issue.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a line shape programatically and make the start and end point be the same. Think it has to start and end at (0,0) but not sure.
2. Brush select.
3. This used to throw an error since we could not get the distance to line, but that should no longer be happening.

### Release notes

- Fix a bug with nearest points for lines that start and end at the same point.

### DIFF
--- a/packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
@@ -117,6 +117,33 @@ export class LineShapeUtil extends ShapeUtil<TLLineShape> {
 		}
 	}
 
+	override onBeforeCreate(next: TLLineShape): void | TLLineShape {
+		const {
+			props: { points },
+		} = next
+		const pointKeys = Object.keys(points)
+
+		if (pointKeys.length < 2) {
+			return
+		}
+
+		const firstPoint = points[pointKeys[0]]
+		const allSame = pointKeys.every((key) => {
+			const point = points[key]
+			return point.x === firstPoint.x && point.y === firstPoint.y
+		})
+		if (allSame) {
+			const lastKey = pointKeys[pointKeys.length - 1]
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx
```diff
@@ -117,6 +117,33 @@ export class LineShapeUtil extends ShapeUtil<TLLineShape> {
...
+		}
+
+		const firstPoint = points[pointKeys[0]]
+		const allSame = pointKeys.every((key) => {
+			const point = points[key]
+			return point.x === firstPoint.x && point.y === firstPoint.y
+		})
+		if (allSame) {
+			const lastKey = pointKeys[pointKeys.length - 1]
+			points[lastKey] = {
```
MitjaBezensek, line 137: Think we can mutate here.

steveruizok: Yep, you can mutate in beforeCreate

steveruizok (approved):

PR #4651, opened 2024-10-03
prevent accidental image drops

steveruizok:
This PR handles a case where users drop images or videos on top of our user interface (on tldraw.com).

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed a bug where dropping images or other things on user interface elements would navigate away from the canvas

### DIFF
--- a/apps/dotcom/client/src/components/SneakyOnDropOverride.ts
+++ b/apps/dotcom/client/src/components/SneakyOnDropOverride.ts
@@ -1,10 +1,36 @@
 import { useEffect } from 'react'
-import { parseAndLoadDocument, useDefaultHelpers, useEditor } from 'tldraw'
+import {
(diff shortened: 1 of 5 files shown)


mimecuvalo (changes requested):
bah! i was about to approve but then i checked Safari and of course this doesn't work on safari (makes an infinite loop) 🙃 arrrrgh

let's look again tomorrow

steveruizok:
> bah! i was about to approve but then i checked Safari and of course this doesn't work on safari (makes an infinite loop) 🙃 arrrrgh
> 
> let's look again tomorrow

I think I got it now

mimecuvalo (approved):
🙌

PR #4665, opened 2024-10-05
Add zoom buttons to collapsed navigation panel on desktop.

steveruizok:
This PR adds + and - buttons to the navigation panel when collapsed on the largest breakpoint (desktop). I'm pretty sure these would be useful for most people, but then again they're present on the default state of the navigation panel.

### Change type

- [x] `improvement`

### Test plan

1. On desktop, check the collapsed navigation panel.

### Release notes

- Adds zoom in and zoom out buttons on desktop.

### DIFF
--- a/packages/tldraw/src/lib/ui.css
+++ b/packages/tldraw/src/lib/ui.css
@@ -320,9 +320,9 @@
 
 /* Zoom button */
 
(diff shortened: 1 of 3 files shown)


steveruizok:
Closing this one. The default state (with mini-map open) is a good enough introduction.

PR #4669, opened 2024-10-06
Improve watermark tests

steveruizok:
This PR improves the tests for our watermark.

The `Watermark.test.tsx` file includes three unit tests that check whether the watermark is displayed or not displayed depending on whether the editor is licensed, unlicensed, or licensed with watermark. Previously, we were testing just the `Watermark` component on its own, which necessitated mocking the Editor and any other hooks used in the component.

In this PR, we instead test with the full Editor. I've created a custom hook, `useLicenseManagerState`, that returns the license manager state. In our tests, we now mock _only_ this hook, returning the result we wish to test and then checking whether the watermark component is in the document.

## Pros

No more mocking random parts of the Editor.

## Cons

An extra hook.

## One more thing

It's also possible to test this with the actual license keys, however that added some extra complexity in exchange for duplicating tests already covered by `LicenseManager.test.ts`.

### Change type

- [x] `other`

### DIFF
--- a/packages/editor/src/lib/license/LicenseProvider.tsx
+++ b/packages/editor/src/lib/license/LicenseProvider.tsx
@@ -1,3 +1,4 @@
+import { useValue } from '@tldraw/state-react'
 import { createContext, ReactNode, useContext, useState } from 'react'
 import { LicenseManager } from './LicenseManager'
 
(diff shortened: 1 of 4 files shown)


steveruizok:
Erp, I extracted this from #4662 but forgot to remove the changes from that PR, so they landed before this one.

PR #4682, opened 2024-10-07
Lod Perf (effect scheduler)

steveruizok:
This PR aims to improve on the LOD performance of image and video elements. Its main goal is to reduce the number of renders caused by the hooks in `useAsset`. It does this by moving the majority of the logic into an effect scheduler (via our `react`), keeping the result (url and asset) in React state, and only updating the result it has actually changed. We also debounce subsequent renders

### Change type

- [x] `improvement`

### Release notes

- The `useAssets` hook now is now specifically for `TLImageAsset`- or `TLVideoAsset`-type assets.

### DIFF
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -3682,9 +3682,9 @@ export function useActions(): TLUiActionsContextType;
 export function useAsset(options: {
     assetId: null | TLAssetId;
     shapeId: TLShapeId;
-    width: number;
 }): {
     asset: TLAsset | null | undefined;
+    isCulled: boolean;
     isPlaceholder: boolean;
     url: null | string;
 };
--- a/packages/tldraw/api-report.md
+++ b/packages/tldraw/api-report.md
@@ -3679,13 +3679,11 @@ export function unwrapLabel(label?: TLUiActionItem['label'], menuType?: string):
 export function useActions(): TLUiActionsContextType;
 
 // @public
-export function useAsset(options: {
+export function useAsset({ shapeId, assetId }: {
     assetId: null | TLAssetId;
     shapeId: TLShapeId;
 }): {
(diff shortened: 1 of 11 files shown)


packages/tldraw/src/lib/shapes/shared/useAsset.ts
```diff
@@ -19,81 +25,107 @@ import { useEffect, useRef, useState } from 'react'
...
+		let cancelDebounceFn: (() => void) | undefined
 
-			if (preview) {
-				setUrl(preview)
-				setIsPlaceholder(true)
-				isReady()
-				return
-			}
-		}
+		const cleanupEffectScheduler = react('update state', () => {
```
mimecuvalo, line 54: your previous work on this, in https://github.com/tldraw/tldraw/pull/4045, you had it using `useReactor` - do you think that's a better tact or is this take more what you were looking for? wdyt?

steveruizok: I think we might be able to use that. Originally there was some extra logic before the `react` that made me use the effect instead, but now it's pretty cleaned up

packages/tldraw/api-report.md
```diff
@@ -3679,13 +3679,11 @@ export function unwrapLabel(label?: TLUiActionItem['label'], menuType?: string):
...
+export function useAsset({ shapeId, assetId }: {
     assetId: null | TLAssetId;
     shapeId: TLShapeId;
-    width: number;
 }): {
-    asset: TLAsset | null | undefined;
-    isPlaceholder: boolean;
+    asset: (TLImageAsset | TLVideoAsset) | null;
     url: null | string;
 };
```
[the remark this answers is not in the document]
steveruizok: If we're considering this a breaking change, should we also rename them?

PR #4683, opened 2024-10-07
botcom: account menu [bk]

mimecuvalo:
[bk=burger king, as Alex says] kinda funky b/c i'm doing these MaybeProviders in `TlaRootProviders.tsx`. but it does work. lemme know what you think — we can rework from here.

Screenshot 2024-10-07 at 22 12 33

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/tla/hooks/useAppState.tsx
```diff
@@ -49,21 +54,33 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
 			return
 		}
 		let _app: TldrawApp
+		let _userId: RecordId<TldrawAppUser>
+
+		const userPreferencesRaw = getUserPreferences()
+		const userPreferences = atom<TLUserPreferences>('userPreferences', userPreferencesRaw)
```
ds300, line 60: we shouldn't be using this built-in user preferences or creating an atom here, we should be creating something which stores prefs on the `TldrawAppUser` record in the store. This could be an ordinary property on the TldrawApp class.

so we'd end up doing something like

```ts
export classs TldrawApp {
  ...
  tlUser = createTLUser({
    userPreferences: computed('user prefs', () => {
       const user = this.getUser(this.getCurrentUserId())
       return {
         ... extract prefs from user record
       }
    }),
    setUserPreferences: (prefs) => {
       this.store.put([{
          ... update the current user
       }])
    }
  })
}
```

looks like you added a property for this on TldrawAppUser but forgot to actually use it?

apps/dotcom/client/src/tla/layouts/TlaSidebarLayout/TlaSidebarLayout.tsx
```diff
@@ -13,14 +14,16 @@ export function TlaSidebarLayout({ children }: { children: ReactNode; collapsabl
...
 	usePreventAccidentalDrops()
+
 	return (
 		<div
 			className={styles.layout}
 			data-sidebar={isSidebarOpen}
 			data-sidebarmobile={isSidebarOpenMobile}
 		>
-			<TlaSidebar />
+			{currentEditor && <TlaSidebar />}
```
ds300, line 26: when would we want to hide the sidebar if there's not an editor running?

packages/dotcom-shared/src/tla-schema/TldrawAppUser.ts
```diff
@@ -6,6 +6,7 @@ import {
 	createRecordType,
 } from '@tldraw/store'
 import { T } from '@tldraw/validate'
+import { TLUserPreferences, getFreshUserPreferences, userTypeValidator } from 'tldraw'
```
mimecuvalo, line 9: @ds300 this is actually causing problems with the bundle-size.
i tried changing it to import from `@tldraw/editor`, that helps but it still increases the bundle size significantly. maybe we move `TLUserPreferences` to schema?

mimecuvalo: nvm, just seeing your other comment which invalidates this q

apps/dotcom/client/src/tla/layouts/TlaSidebarLayout/TlaSidebarLayout.tsx
mimecuvalo: yeah, right now the sidebar is really entangled with the Sidebar which isn't great:
- `TlaSidebar` → `useTldrFileDrop` → `useToasts` → `TldrawUiDialogsProvider` → `editor`
- `TldrawUiContextProvider` → `useEditor`
- `TlaFileMenu` → `useDialogs` → `TldrawUiDialogsProvider` → `editor`

several things like that :-/

steveruizok:
Second thought, this is going to have some rough conflicts to resolve.

packages/editor/src/lib/config/TLUserPreferences.ts
```diff
@@ -13,15 +13,20 @@ const USER_DATA_KEY = 'TLDRAW_USER_DATA_v3'
...
 	edgeScrollSpeed?: number | null
 	colorScheme?: 'light' | 'dark' | 'system'
 	isSnapMode?: boolean | null
 	isWrapMode?: boolean | null
 	isDynamicSizeMode?: boolean | null
 	isPasteAtCursorMode?: boolean | null
+	isToolLocked?: boolean | null
+	isGridMode?: boolean | null
+	isFocusMode?: boolean | null
+	isDebugMode?: boolean | null
```
steveruizok, line 29: These aren't supposed to be synced between projects / tabs, so we shouldn't add them to user preferences.

packages/tldraw/src/lib/ui/context/TldrawUiContextProvider.tsx
```diff
@@ -52,7 +52,7 @@ export interface TLUiContextProviderProps {
 }
 
 /** @public @react */
-export function TldrawUiContextProvider({
+export const TldrawUiContextProvider = track(function TldrawUiContextProvider({
```
steveruizok, line 55: What are we capturing here?

steveruizok:
Hm, let's think more about where / how we're splitting up user preferences.

I don't want to push instance-based preferences (e.g. tool lock) to the user preferences record, though we've combined them previously in the preferences panel. Language, dark mode / light mode will effect both the application and the editor, so it makes sense for those to be accessible from outside of the editor; the rest we can leave in a menu within the editor itself?

packages/editor/src/lib/config/TLUserPreferences.ts
mimecuvalo: You don't think so? I think it's worth a discussion — or at least, maybe an education for me on what I might be missing expectation-wise for users.

In my mind, these feel like App-level type prefs (say, especially focus or debug mode) but let's talk about it more in person.

packages/tldraw/src/lib/ui/context/TldrawUiContextProvider.tsx
mimecuvalo: Similar to the language menu hotfix that went out, it's so the lang menu is reactive: https://github.com/tldraw/tldraw/pull/4685

mimecuvalo:
(fixed up the merge btw, wasn't too hairy)

ds300:
to my mind these make sense to set globally, since they are legit preferences

- tool lock
- always snap
- select on wrap
- edge scrolling
- reduce motion
- paste at cursor


these only make sense per-document to me, because they are not long-lived preferences, they're temporary modal changes

- focus mode
- grid mode
- debug mode

Not sure about 'dynamic size', i can see arguments for both, but I'd lean towards per-document too.

think about it, if you have multiple documents open in different tabs and you change to 'focus mode' in one tab, do you expect the other tab to change to focus mode too? I wouldn't. Same for grid mode and debug mode and dynamic size.

mimecuvalo:
> think about it, if you have multiple documents open in different tabs and you change to 'focus mode' in one tab, do you expect the other tab to change to focus mode too? I wouldn't. Same for grid mode and debug mode and dynamic size.

this is a good litmus test — thanks David!

hah, however, whilst i agree about focus mode, i would disagree about debug mode, and maybe dynamic size.

we should have a lively debate about this in person :)

mimecuvalo:
i removed the controversial bits - gonna land the basics

PR #4686, opened 2024-10-08
Restore placeholder on .dev

steveruizok:
This PR restores the placeholder interaction on tldraw.dev.

### Change type

- [x] `other`


mimecuvalo:
@steveruizok one thing that's lost in this change is the Loading... state which was intentionally done this way so that slower networks could have a more responsive "Try it" button:

Screenshot 2024-10-08 at 14 12 13

this is why it's `className={cn('z-10 h-full', hidden ? 'hidden' : '')}` instead of just returning null

PR #4688, opened 2024-10-08
[botcom] Publishing

MitjaBezensek:
Adds publishing to botcom. 

This allows the users to publish the existing document. This is a point in time snapshot, which they can then update or delete at a later time. Only the owner of the file has the permission to do that, while everyone with a link can view the published document.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Add publishing to botcom.


apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
```diff
@@ -38,13 +40,16 @@ export const TlaButton = forwardRef<
 				className
 			)}
 		>
-			{icon && <TlaIcon icon={icon} />}
-			{children && <span>{children}</span>}
-			{iconRight && <TlaIcon icon={iconRight} className={styles.iconRight} />}
-			{isLoading && (
+			{isLoading ? (
```
MitjaBezensek, line 43: Looks like this logic was never used and was bulled. The spinner was show on top of other content.

packages/dotcom-shared/src/tla-schema/TldrawAppUser.ts
```diff
@@ -52,18 +53,33 @@ export const tldrawAppUserValidator: T.Validator<TldrawAppUser> = T.model(
...
+} as const)
+
 /** @public */
 export const tldrawAppUserMigrations = createRecordMigrationSequence({
 	sequenceId: 'com.tldraw-app.user',
 	recordType: 'user',
-	sequence: [],
+	sequence: [
+		{
+			id: Versions.AddPublishFlag,
```
MitjaBezensek, line 74: Feels a bit strange what we had to add a migration to introduce a feature flag. Maybe our flags could just be a map for string to bool?

mimecuvalo:
my intuition is that maybe we do away with snapshots, and yeah, go for a publish mode instead.
some reasons in no particular order, stream of consciousness:
- managing the snapshots seems annoying - a lot of UI just to keep track of them all
- higher-level product decision: it _feels_ like snapshots aren't as useful as published pages. in my head, they get into the territory of 'snapshot_final_final_v2_no_really_this_version' that you have to continually keep track of which is the version you want to be sending out to someone.
- because of share permissions, if someone had the intent to send two different versions to different people, it would start getting confusing (are they readonly? shouldn't i just share it with them so they can comment on them? oh, snapshots aren't something you can comment on?)
- especially because of the commenting consideration, then you, as the author, have to track down the particular snapshot that someone has potentially commented on in the future (if we were to do that)
- basically, if i wanted to send different snapshots to folks, it feels easier to do just Select All→Create a new board and share that separately vs. trying to do it in a sub-system called snapshots.
- related to that, because we're treating these as files, it then feels like snapshots are subfiles within files which feels confusing, data management-wise.

all the above is to say: feels like publishing is more in-line with what a collaborative product offering would provide.

#strong-opinion-weakly-held that is definitely worth more discussion!

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -218,18 +225,22 @@ export class TldrawApp {
...
 	toggleFileShared(fileId: TldrawAppFileId) {
-		const userId = this.getCurrentUserId()
-
 		const file = this.get(fileId) as TldrawAppFile
 		if (!file) throw Error(`No file with that id`)
 
-		if (userId !== file.ownerId) {
-			throw Error('user cannot edit that file')
-		}
+		if (!this.isFileOwner(fileId)) throw Error('user cannot edit that file')
```
MitjaBezensek, line 231: Flyby change since I added `isFileOwner` helper method.

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -271,6 +282,12 @@ export class TldrawApp {
 		return { newFile, editorStoreSnapshot }
 	}
 
+	isFileOwner(fileId: TldrawAppFileId) {
```
mimecuvalo, line 285: i assume eventually, non file owners should be able to publish, eh? but for now this is fine

apps/dotcom/client/src/tla/components/TlaEditor/TlaSnapshotsEditor.tsx
```diff
@@ -0,0 +1,83 @@
+import { getLicenseKey } from '@tldraw/dotcom-shared'
```
mimecuvalo, line 1: nit: can we rename this file to `TlaPublishedEditor`? throwing me off, heh (especially because it's explicitly read-only, it can kiiinda sound like it's editable but it's not)

apps/dotcom/client/src/tla/components/TlaEditor/TlaSnapshotsEditor.tsx
```diff
@@ -0,0 +1,83 @@
...
+			<Tldraw
+				licenseKey={getLicenseKey()}
+				assetUrls={assetUrls}
+				snapshot={snapshot}
+				overrides={[sharingUiOverrides, fileSystemUiOverrides]}
+				onUiEvent={handleUiEvent}
+				onMount={(editor) => {
+					;(window as any).app = editor
+					;(window as any).editor = editor
+					editor.updateInstanceState({ isReadonly: true })
```
mimecuvalo, line 71: this makes me think we should have a readonly property at the top-level - feels somewhat dirty in general that we have to wait until `onMount` to mark it as readonly... 🤔

mimecuvalo (commented):
lookin' good - mostly a bunch of nits, structurally i think everything is fine

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaFileShareMenuSharePage.tsx
```diff
@@ -1,14 +1,11 @@
 import { TldrawAppFile, TldrawAppFileId } from '@tldraw/dotcom-shared'
-import { useCallback, useEffect, useLayoutEffect, useRef } from 'react'
-import { FileHelpers, useLocalStorageState, useToasts, useValue } from 'tldraw'
+import { useCallback } from 'react'
```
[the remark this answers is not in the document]
MitjaBezensek: Yeah. Maybe we should make a directory for the tabs and put them there. Would avoid the `FileShareMenu` prefix for all of these.

apps/dotcom/client/src/tla/components/TlaEditor/TlaSnapshotsEditor.tsx
MitjaBezensek: Yeah, maybe. Think we did have it a while back, but unsure why it was removed.

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaPublishPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+		const result = await fetch(`${PUBLISH_ENDPOINT}/${publishedSlug}`, {
+			method: 'DELETE',
+			headers: {
+				'Content-Type': 'application/json',
+				Authorization: `Bearer ${token}`,
+			},
+		})
+		if (result.ok) {
+			app.toggleFilePublished(file.id)
```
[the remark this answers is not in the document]
MitjaBezensek: `publish` is used for both updates and first uploads and should only change the `published` status if it's a new upload. At least for now only the owner of a file can publish/delete things, so race conditions are not really a problem. We will have to think about them in the long term though.

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaPublishPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {
+	TlaMenuControl,
+	TlaMenuControlGroup,
+	TlaMenuControlLabel,
+	TlaMenuSection,
+} from '../tla-menu/tla-menu'
+import { QrCode } from './QrCode'
+import { TlaShareMenuCopyButton } from './file-share-menu-primitives'
+
+export function TlaPublishPage({ file }: { file: TldrawAppFile }) {
```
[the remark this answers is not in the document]
MitjaBezensek: We use the same name for other tabs, so I kept it like this.

ds300:
Some UX feedback:

- toggling the published state feels sluggish. it should toggle optimistically and do the backend stuff in the background.
- the publishing workflow is not obvious, this could use some tweaking. i'd suggest something along these lines 
   
   image

   This makes it more clear that new changes won't be automatically published and you have to come back to do that.

apps/dotcom/sync-worker/src/routes/deletePublishedRoom.ts
```diff
@@ -0,0 +1,20 @@
...
+	const parentSlug = await env.SNAPSHOT_SLUG_TO_PARENT_SLUG.get(roomId)
+	if (!parentSlug) return new Response('Parent slug is required', { status: 400 })
+
+	if (!(await isFileOwner(request, env, parentSlug))) {
+		return new Response('Unauthorized', { status: 401 })
+	}
+	await env.ROOM_SNAPSHOTS.delete(
+		getR2KeyForSnapshot({ parentSlug, snapshotSlug: roomId, isApp: true })
+	)
+	return new Response('ok', { status: 200 })
```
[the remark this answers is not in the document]
MitjaBezensek: We mostly just used the main status from each class (200, 400, 500) 😂 

That said we don't do much rest and it's been fine till now 🤷‍♂️

apps/dotcom/sync-worker/src/routes/getPublishedRoom.ts
```diff
@@ -0,0 +1,33 @@
+import { RoomSnapshot } from '@tldraw/sync-core'
+import { notFound } from '@tldraw/worker-shared'
+import { IRequest } from 'itty-router'
+import { getR2KeyForSnapshot } from '../r2'
+import { Environment } from '../types'
+import { R2Snapshot } from './createPublishedRoom'
+
+export async function getPublishedRoom(request: IRequest, env: Environment): Promise<Response> {
+	const roomId = request.params.roomId
+	if (!roomId) return notFound()
```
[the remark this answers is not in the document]
MitjaBezensek: Fixed this here https://github.com/tldraw/tldraw/pull/4688/commits/47aa07087efdcd1a684fedc72e1504a4bcc4dd63. Should give us a general way of handling 404s.

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaPublishPage.tsx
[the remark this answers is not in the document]
mimecuvalo: huh ok weird ¯\\_(ツ)_/¯

apps/dotcom/client/src/tla/components/TlaFileShareMenu/TlaPublishPage.tsx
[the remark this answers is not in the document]
mimecuvalo: sounds good, i think the same way the publish has a check `if (!published)` before toggling, we should just be defensive here.

apps/dotcom/client/src/tla/app/TldrawApp.ts
mimecuvalo: yes, definitely still useful — sorry, this code is good. i meant this as a comment more like we'll need an additional permissions check later for publishing by multiple editors later.

mimecuvalo (approved):
👍

apps/dotcom/client/src/tla/app/TldrawApp.ts
MitjaBezensek: Yeah. 

With @steveruizok we discussed different permissions levels. Publishing / deleting a published document is meant for `Full access` level. Right now that's only the owner, but when we will have groups then the whole group will have `Full access`.

MitjaBezensek:
> Some UX feedback:
> 
> * toggling the published state feels sluggish. it should toggle optimistically and do the backend stuff in the background.

Added optimistic updates 👍 

> * the publishing workflow is not obvious, this could use some tweaking. i'd suggest something along these lines

Will go ahead and merge this, just to avoid further conflicts, but lets discuss this tomorrow during the standup.

mimecuvalo:
> Will go ahead and merge this, just to avoid further conflicts, but lets discuss this tomorrow during the standup.

Sounds good to merge. I would agree with David's UX suggestions though — he's made a good callout about both "Last published" and "You have unpublished changes". Those definitely would help convey with more clarity to the user what the state of things is.

PR #4694, opened 2024-10-09
make sure DOM IDs are globally unique

SomeHats:
There are a lot of places where we currently derive a DOM ID from a shape ID. This works fine (ish) on tldraw.com, but doesn't work for a lot of developer use-cases: if there are multiple tldraw instances or exports happening, for example. This is because the DOM expects IDs to be globally unique. If there are multiple elements with the same ID in the dom, only the first is ever used. This can cause issues if e.g. 

1. i have a shape with a clip-path determined by the shape ID
2. i export that shape and add the resulting SVG to the dom. now, there are two clip paths with the same ID, but they're the same
3. I change the shape - and now, the ID is referring to the export, so i get weird rendering issues.

This diff attempts to resolve this issue and prevent it from happening again by introducing a new `SafeId` type, and helpers for generating and working with `SafeId`s. in tldraw, jsx using the `id` attribute will now result in a type error if the value isn't a safe ID. This doesn't affect library consumers writing JSX. 

As part of this, I've removed the ID that were added to certain shapes. Instead, all shapes now have a `data-shape-id` attribute on their wrapper.

### Change type

- [x] `bugfix`

### Release notes

- Exports and other tldraw instances no longer can affect how each other are rendered
- **BREAKING:** the `id` attribute that was present on some shapes in the dom has been removed. there's now a data-shape-id attribute on every shape wrapper instead though.


ds300 (approved):
looks great, a worthy breaking change!

PR #4698, opened 2024-10-09
[botcom] use single DurableObject for whole app

ds300:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


steveruizok (commented):
Looks good! I was getting a bug where the user wouldn't be found once the store synced, which was very confusing. The problem was solved by deleting my account on clerk and creating a new one in the UI.

Side note: There are a bunch of superficial changes that remove userId as a parameter to TldrawApp methods. (All good, I added some too that had come in last week)

steveruizok (approved):

PR #4699, opened 2024-10-09
botcom: scrap quick actions and scatter the ashes

mimecuvalo:
take the quick actions, compost it, and make something new. the Embed icon makes the overflow menu lopsided but i think we can rebalance it - nix the checkbox shape, for example?

- edit is covered by context menu / very common known kbd shortcuts
- view is covered by zoom corner menu already
- export moves to sidebar
- embed moves to overflow item
- media already is a toolbar item

before:
Screenshot 2024-10-09 at 15 32 11

after:
Screenshot 2024-10-09 at 15 31 46
Screenshot 2024-10-09 at 15 31 54

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
Hm, we shouldn't have the export options in the sidebar file menu like that. If they were in the sidebar, then because the sidebar also shows files that aren't open, we would have to spin up an editor in order to generate the image. We could do that, but it makes sense to me to have those options arranged for the current.

Perhaps we should separate the FileMenu into 1) the menu next to the title of the editing file and 2) the FileMenu that we see in the sidebar, and include those options in only the editing file's file menu.

packages/tldraw/src/lib/ui/components/Toolbar/DefaultToolbarContent.tsx
```diff
@@ -210,6 +211,11 @@ export function FrameToolbarItem() {
 	return <ToolbarItem tool="frame" />
 }
 
+/** @public @react */
+export function EmbedToolbarItem() {
+	return <ToolbarItem tool="embed" />
+}
```
steveruizok, line 217: If we were going to add a menu item like this, I would suggest we make it an "insert" type tool that combined images, embeds, and other content. I'd skip it for now and leave it as a bit of an easter egg for users who know to paste links. (The menu item is only used by < 1% of users)

steveruizok:
Title is misleading btw, this just scrabs the menu from the toolbar's quick actions panel; the rest of those little quick actions are still there. Actually we could make the embed / other menu items part of the actions menu? (Who named this stuff??)

mimecuvalo:
> Perhaps we should separate the FileMenu into 1) the menu next to the title of the editing file and 2) the FileMenu that we see in the sidebar, and include those options in only the editing file's file menu.

I think that makes sense, yeah. Let me play around with that.

> Title is misleading btw, this just scrabs the menu from the toolbar's quick actions panel; the rest of those little quick actions are still there. Actually we could make the embed / other menu items part of the actions menu? (Who named this stuff??)

Fair!

mimecuvalo:
superseded by https://github.com/tldraw/tldraw/pull/4764

PR #4707, opened 2024-10-11
ui: suppress aria warning about desc in dialogs

mimecuvalo:
small little tweak - noisy in dev logs. if in the future if we add descriptions for dialogs we can make this an optional parameter

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek (commented):
💅

PR #4710, opened 2024-10-11
ui: dont highlight menu triggers that dont have their submenus open

mimecuvalo:
the menu backgrounds were showing up even if the submenus weren't open - they should only be open when data-highlighted is active. also, fix up border-radiuses to match hover/open sub-menu state

the issue (you'll note File + Edit are both highlighted):
Screenshot 2024-10-11 at 09 58 10

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix submenu hover active state.


steveruizok:
I think actually the radii should be 0px here rather than more, so that we dont have this weird pinched contact between the radius and the submenu

image

There's still this super annoying bug with radix dropdown menu too

Kapture 2024-10-11 at 21 26 27

mimecuvalo:
> I think actually the radii should be 0px here rather than more, so that we dont have this weird pinched contact between the radius and the submenu

Hm, I think making it 0 makes it look worse unfortunately :-/ (if you zoom in, you can see the awkward hard edge next to the menu)
Screenshot 2024-10-17 at 09 29 54

 
> There's still this super annoying bug with radix dropdown menu too

Yeah, I had tried fixing that, but I'd have to go into Radix internals and it's just not worth at this time.

steveruizok:
Good shout, that does looks wrong to me too for the first / last items, though it looks better for the following items...

steveruizok:
Ok, I had to go back into my designs to figure out why the radii are the way they are.

image

The menu containers have a radius of 9px. While the container has no padding, the hover pseudo-elements have 4px of inset. This would normally suggest 5px of border radius on the hover elements. However, the containers have 1px "internal border" from their shadow. this is much more noticeable on dark mode than light mode, so light mode should probably have an actual border radius of 5px to look perfect. In other words, 4px is visually right half the time, 6px is visually wrong in both cases.

### Why was the highlight indicator radius only 2px?

Submenus are highlighted when open, using a gradient that points in the direction of the submenu content. The menus overlap at 6px. That leads to 2px overlap of the 4px radius, which gives it a pretty sharp pinch between the edge of the submenu and the highlight indicator.

image

Since the gradient hides the un-overlapped side of the indicator, I could reduce the radius in order to reduce that pinch. A perfect solution would be to only round the indicators if the menu item is the very first or the very last item in a menu, and thus effected by the sort of inset of the menu itself, but that's not possible to select for with CSS (given the different depths of the menu items).

There might be a way! But it's complicated. I've build the menus around "full size" touch targets on the menu items (ie with no dead space at the left or right) so using a regular border on the menu would be a problem. I did a quick test using a pseudo-element on the menu content itself that provides this inset border, and then remove the but that would effect group dividers, and mean those would need to be pseudo-elements as well, positioned in front of the menu content's inset border element. :psyduck:

Anyway, if we wanted to fix the pinch completely, we could either:

1. remove the indicator 
2. remove the overlap (and bump the radius back to --radius-2)
3. increase the overlap to hide the pinch

image

steveruizok:
My favorite design visually is:
- zero the radii
- increase the overlap

image

steveruizok:
Let's remove the radius change from this PR and I'll open a new PR with just that change. Everything else looks good! Would love to improve the appearance of the indicator in the menus as a followup, that's such an easy-to-produce bug and makes me cringe every time.

mimecuvalo:
> Ok, I had to go back into my designs to figure out why the radii are the way they are.
> 

omg steve, you got nerdsniped hard here 😮

PR #4711, opened 2024-10-11
[botcom] local session state, logged out view of files

steveruizok:
This PR implements a few changes inspired by the instantdb branch.

## Providers

I've flattened the component tree for react context providers in `TlaProviders`, which were getting complex. This should make it easier in the future to do clean diffs as we adjust providers. I started it, but we should avoid having too many components that return other wrappers and providers in this file.

## Local session state

The `TldrawAppSessionsState` is ignored. I have not removed it from the schema, as I don't want to handle that right now, but we no longer read it or update it. Instead, we keep a local atom that tracks data that was previously in the session record: ie sidebars, theme, auth user id, etc.

## Removing dependencies on `app`

The change to `localSessionStorage` removes several places where we depended on the `app`, as well as all of the places where we updated it.

## Offline preferences for export settings

Export settings are currently sitting on the `TldrawAppUser` record. In this PR, they also are at `localSessionStorage.exportSetttings` as a backup for signed-out users.

## Offline editor changes

When visiting a file while logged out, the editor now displays the `AnonLayout`, i.e. a framing around the editor.

image

### People menu

When visiting another user's file while logged out, I've included the `PeopleMenu` in a location next to the Style Panel.

image

### Export menu

For logged out users, I've (perhaps temporarily) replaced the toggle sidebar button with a button that opens an export menu.

image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/tla/providers/TlaProvider.tsx
```diff
@@ -18,43 +30,69 @@ if (!PUBLISHABLE_KEY) {
...
+	const handleAppLevelUiEvent = useCallback<TLUiEventHandler>(() => {
+		// todo, implement handling ui events at the application layer
+	}, [])
+
 	return (
 		<ClerkProvider publishableKey={PUBLISHABLE_KEY} afterSignOutUrl="/q">
-			<TlaRootProviders>
-				<SignedInProvider />
-			</TlaRootProviders>
+			<AppStateOrNotProvider>
```
mimecuvalo, line 39: fwiw, this is going to be improved in https://github.com/tldraw/tldraw/pull/4683 where I'm just using `TldrawUiContextProvider` instead of trying to recreate it manually (which will be error-prone)

steveruizok:
Any other reviews?

PR #4712, opened 2024-10-11
tla: rename icons without equal sign

mimecuvalo:
these equal signs are throwing me off - we shouldn't do this b/c typically in URLs equals are paired with parameters.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek (approved):

steveruizok:
For context, these are the file-names as they come out of Figma when the icons are exported. I'll set up a plugin that will follow this pattern when I do exports in the future.

PR #4719, opened 2024-10-16
botcom: scaffolding for i18n

mimecuvalo:
This starts rethinking our i18n story, at least at the botcom level.
- introduces (brings back) `react-intl` which lets start doing more complex translations (being able to include HTML in sentences), do pluralization, time, etc.
- adds a i18n middle layer to auto-generate ID's based on their hash, otherwise annoying to do this.
- you can see the different uses, there is the `<F>` tag for translated React items
- in other cases, you use `defineMessages`/`formatMessage` for things like alt/aria-labels/tooltips.
- adds some hidden languages like "accented" to show which where things are untranslated.
- similarly, adds a debug mode to add an annoying red text shadow to highlight things that are untranslated.
- also adds a looooooong language for languages like German where things start breaking the UI (basically tests UI to where we should have some ellipsis or other considerations)

- i18n cheat sheet: https://www.notion.so/tldraw/i18n-cheat-sheet-1223e4c324c080cf8e64c2b67f47378f

This has a new App-level set of debug items btw:
Screenshot 2024-10-25 at 23 02 44

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/tla/providers/TlaRootProviders.tsx
```diff
@@ -75,10 +99,26 @@ function InsideOfContainerContext({ children }: { children: ReactNode }) {
 
 function SignedInProvider({
```
MitjaBezensek, line 100: Won't we allow locale changes for signeout users?

MitjaBezensek (commented):
It's a bit hard to test the debug options since you have to login to turn them on/off, then log out to see the login screen where the debug options come into play.

We incorrectly tag editor strings:
image

It also seems that the current dotcom (not the tla routes) debugging does not react to the flags set there:
image

Feels like we should completely isolate these two: remove flags from the editor package and move it to notcom.

mimecuvalo:
> We incorrectly tag editor strings:
Yeah, sorry, I should have noted that I haven't fully fleshed this out for the editor strings, definitely those would be excluded before going forward. I should note that before landing that this PR is an RFC really. It's not fully-baked — I've marked the title of this PR to reflect that.

Sorry for the confusion!

apps/dotcom/client/src/tla/providers/TlaRootProviders.tsx
```diff
@@ -28,24 +30,46 @@ if (!PUBLISHABLE_KEY) {
 
 export function Component() {
 	const [container, setContainer] = useState<HTMLElement | null>(null)
+	const [locale, setLocale] = useState<string>('en')
+	const [messages, setMessages] = useState({})
 	const [theme, setTheme] = useState<'light' | 'dark' | 'system'>('light')
 	const handleThemeChange = (theme: 'light' | 'dark' | 'system') => setTheme(theme)
+	const handleLocaleChange = async (locale: string) => {
```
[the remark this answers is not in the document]
mimecuvalo: Good idea, I'll rework if we decide to move forward. 👍

apps/dotcom/client/src/tla/providers/TlaRootProviders.tsx
mimecuvalo: Hah, it's funny b/c right now there might not be a menu for logged out users, especially if we do this: https://github.com/tldraw/tldraw/pull/4699
I think we need to think of where a general App menu would go for logged out users...

mimecuvalo:
@MitjaBezensek this is ready for re-review. There's still more work to do, this just lays the groundwork.
We'll need to pull the app `raw()` strings to use this. I did a demo in Lokalise and it works fine as expected. So the next step is setting up an automatic pipeline to Lokalise as well.

MitjaBezensek (approved):
Looking good. Thanks for adding the docs, so helpful 🙌

PR #4723, opened 2024-10-17
Don't index multiplayer rooms, snapshots, history.

MitjaBezensek:
Prevent indexing of multiplayer rooms, snapshots, new room route (it just redirects), history, etc

This only touches current dotcom. Logged in experience will be hidden behind auth so should not be discoverable by search bots.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Prevent indexing of multiplayer rooms, snapshots, new room route (it just redirects), history, etc

### DIFF
--- a/apps/dotcom/client/src/pages/noindex.tsx
+++ b/apps/dotcom/client/src/pages/noindex.tsx
@@ -5,7 +5,7 @@ export function Component() {
 	return (
 		<>
 			<Helmet>
-				<meta name="robots" content="noindex" />
+				<meta name="robots" content="noindex, noimageindex, nofollow" />
 			</Helmet>
 			<Outlet />
 		</>
--- a/apps/dotcom/client/src/pages/noindex.tsx
+++ b/apps/dotcom/client/src/pages/noindex.tsx
@@ -1,0 +1,13 @@
+import { Helmet } from 'react-helmet-async'
+import { Outlet } from 'react-router-dom'
+
+export function Component() {
+	return (
+		<>
+			<Helmet>
+				<meta name="robots" content="noindex" />
+			</Helmet>
+			<Outlet />
+		</>
+	)
+}
--- a/apps/dotcom/client/public/robots.txt
+++ b/apps/dotcom/client/public/robots.txt
@@ -1,11 +1,2 @@
 User-agent: *
-Disallow: /r
-
-User-agent: *
-Disallow: /v
-
-User-agent: *
-Disallow: /s
-
-User-agent: *
 Allow: /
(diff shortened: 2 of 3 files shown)


mimecuvalo (commented):
aren't we already blocking these routes?
https://github.com/tldraw/tldraw/blob/main/apps/dotcom/client/public/robots.txt

MitjaBezensek:
> aren't we already blocking these routes? https://github.com/tldraw/tldraw/blob/main/apps/dotcom/client/public/robots.txt

Don't think that's enough. According to [google](https://developers.google.com/search/docs/crawling-indexing/robots/intro):

> A robots.txt file tells search engine crawlers which URLs the crawler can access on your site. This is used mainly to avoid overloading your site with requests; **it is not a mechanism for keeping a web page out of Google. To keep a web page out of Google, block indexing with noindex or password-protect the page**.

MitjaBezensek:
That said I guess we should then remove these routes from robots.txt
https://developers.google.com/search/docs/crawling-indexing/block-indexing

> Important: For the noindex rule to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex rule, and the page can still appear in search results, for example if other pages link to it.

mimecuvalo:
> Important: For the noindex rule to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex rule, and the page can still appear in search results, for example if other pages link to it.


this is bonkers

apps/dotcom/client/src/pages/noindex.tsx
```diff
@@ -0,0 +1,13 @@
+import { Helmet } from 'react-helmet-async'
+import { Outlet } from 'react-router-dom'
+
+export function Component() {
+	return (
+		<>
+			<Helmet>
+				<meta name="robots" content="noindex" />
```
mimecuvalo, line 8: looking into this a bit, i think we can do to make this stronger:
`<meta name="robots" content="NOINDEX, NOIMAGEINDEX, NOFOLLOW" />`

mimecuvalo (commented):
i'm wondering if we should do this on the Cloudflare level, where we would use:
`X-Robots-Tag: noindex, noimageindex, nofollow`

wdyt? it make it less complex for our router

MitjaBezensek:
> i'm wondering if we should do this on the Cloudflare level, where we would use: `X-Robots-Tag: noindex, noimageindex, nofollow`
> 
> wdyt? it might make it less complex for our router

Feels like doing it on code makes it more transparent when the routes won't be indexed. Adding another route is much easier to get right. Otherwise you need to be aware that you might also need to add another rule to cloudflare 🤷‍♂️

mimecuvalo:
> Otherwise you need to be aware that you might also need to add another rule to cloudflare 🤷‍♂️

yes! good point.

PR #4725, opened 2024-10-17
botcom: redirect to intended room when signing in

mimecuvalo:
also, fixes up the not authenticated/forbidden error msg (was always sending down forbidden accidentally)

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek (commented):
Should we do the same thing for profile (and debug) page?

mimecuvalo:
> Should we do the same thing for profile (and debug) page?

I would but in my other PRs I'm removing those anyway ¯\\_(ツ)_/¯
If we bring something back for those pages, then yeah, we can add a redirect.

PR #4731, opened 2024-10-18
Menu event juggling

steveruizok:
This PR restores the "prevent clicks but allow drags" logic when menus are open in https://github.com/tldraw/tldraw/pull/4731

### Change type

- [x] `improvement`


mimecuvalo (approved):
lookin good 😎

PR #4732, opened 2024-10-19
refactor: specify type at bbox

nayounsang:
I found that bbox's type is any.
Bbox is let variable, so I thought it would be good to improve.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. yarn run test
- Just I change the type, so logic is not changed. It just checks to see if there are any type errors.

- [ ] Unit tests
- [ ] End to end tests

### Release notes
- When I see the code in `packages/editor/src/lib/exports/getSvgJsx.tsx`, Improvements were found.
```
// L57
let bbox: = null // any type
```
- This is declared as `let`, but it is `any` type.
- I felt this was a risk for future maintenance.
- So I specify the type of `bbox`.
```
let bbox: null | Box = null
```


steveruizok (approved):

steveruizok:
Thanks @nayounsang !

PR #4733, opened 2024-10-19
[Refactor]: Get svg as image unused editor

nayounsang:
I found the unused arg `editor:Editor` in `getSvgAsImage`.
So I change it. To improve DX and context of func. 
### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. yarn run test

- [ ] Unit tests
- [ ] End to end tests

#### [5813541](https://github.com/tldraw/tldraw/pull/4733/commits/5813541915665195c4e45a260494784b635b891c)
- Change the arg name: `editor:Edtior` -> `_editor:Editor|null`
- Underbar means unused arg.
- it is nullable. Therefore, when using the func, developers can conveniently use it by specifying null regardless of the declaration of the editor class. 
- This can be used without error regardless of existing code. This is because it is simply a commit that changed the name and type.
#### [6f63f2c](https://github.com/tldraw/tldraw/pull/4733/commits/6f63f2c081c7df275501fbb030cc8784e1b87fed)
- The meaning has been clarified by deleting unused `editor`.
- I changed the code using `getSvgAsImage` inside the tldraw repo.
- But, there is concern about errors as it is not compatible with existing code.
- I think this commit should be applied when the major or minor version is released. 
- Therefore, I revert it and I have not yet tested this commit.
- If maintainers think this is correct, I will test it.

### DIFF
--- a/apps/docs/content/getting-started/releases-versioning.mdx
+++ b/apps/docs/content/getting-started/releases-versioning.mdx
@@ -15,19 +15,89 @@ Unlike many JavaScript packages distributed on [NPM](https://www.npmjs.com/), th
 
 {/* START AUTO-GENERATED CHANGELOG */}
 
-## Current release: [v3.2.0](/releases/v3.2.0)
+## Current release: [v3.3.0](/releases/v3.3.0)
 
-Welcome to the `v3.2.0` release of tldraw!
+Welcome to the 3.3.0 release of the tldraw sdk. This release includes a number of new features and bug fixes.
 
-### Oops
+Special thanks to new contributors [@nayounsang](https://github.com/nayounsang) and [@vladh](https://github.com/vladh)!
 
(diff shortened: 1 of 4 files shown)


steveruizok:
Probably not the best professional practice, but IIRC we kept this in there to avoid a breaking change; and given how common the editor is as a parameter, it might be better to keep it in so that we don't have to break things again if we start depending on it. wdyt @mimecuvalo ?

nayounsang:
> Probably not the best professional practice, but IIRC we kept this in there to avoid a breaking change; and given how common the editor is as a parameter, it might be better to keep it in so that we don't have to break things again if we start depending on it. wdyt @mimecuvalo ?

Oh, this makes sense if requirements and specifications of `getSvgAsImage` is changed.

PR #4752, opened 2024-10-22
[Fix] Colors from Excalidraw when pasting

steveruizok:
Excalidraw changed their colors, so this fixes the translation between excalidraw's colors and tldraw's colors when pasting excalidraw content into tldraw.

### Change type

- [x] `bugfix`

### Test plan

1. Paste some excalidraw shapes into tldraw

### Release notes

- Fixed a bug that prevented pasted Excalidraw content from keeping the right color.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/clipboard/pasteExcalidrawContent.ts
+++ b/packages/tldraw/src/lib/ui/hooks/clipboard/pasteExcalidrawContent.ts
@@ -437,7 +437,7 @@ function mapExcalidrawColorToTldrawColors(
 }
 
 const colorsToColors: Record<string, TLDefaultColorStyle> = {
-	...mapExcalidrawColorToTldrawColors('gray', 'grey', 'grey'),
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui/hooks/clipboard/pasteExcalidrawContent.ts
```diff
@@ -410,40 +411,44 @@ const fontFamilyToFontType: Record<number, TLDefaultFontStyle> = {
...
+	dark: TLDefaultColorStyle
+) {
+	const colors = [0, 1, 2, 3, 4].map((index) => oc[excalidrawColor][index])
+	return Object.fromEntries(colors.map((c, i) => [c, i < 3 ? light : dark]))
+}
+
 const colorsToColors: Record<string, TLDefaultColorStyle> = {
-	'#ffffff': 'grey',
-	// Strokes
+	...mapExcalidrawColorToTldrawColors('gray', 'grey', 'grey'),
```
mimecuvalo, line 440: gray vs grey? one of them seems to be a typo

PR #4757, opened 2024-10-22
chore: upgrade eslint v8 to v9

mimecuvalo:
As I worked on the i18n PR (https://github.com/tldraw/tldraw/pull/4719) I noticed that `react-intl` required a new version of `eslint`. That led me down a bit of a rabbit hole of upgrading v8 → v9. There were a couple things to upgrade to make this work.

- ran `npx @eslint/migrate-config .eslintrc.js` to upgrade to the new `eslint.config.mjs`
- `.eslintignore` is now deprecated and part of `eslint.config.mjs`
- some packages are no longer relevant, of note: `eslint-plugin-local` and `eslint-plugin-deprecation`
- the upgrade caught a couple bugs/dead code

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Upgrade eslint v8 → v9


apps/examples/e2e/tests/fixtures/fixtures.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
MitjaBezensek, line 1: Out of curiosity, why do we need this? Is it just the `use` name that's causing it?

packages/tldraw/src/lib/ui/components/StylePanel/DoubleDropdownPicker.tsx
```diff
@@ -41,14 +41,17 @@ function _DoubleDropdownPicker<T extends string>({
 	valueB,
 	onValueChange,
 }: DoubleDropdownPickerProps<T>) {
+	// eslint-disable-next-line react-hooks/rules-of-hooks
```
MitjaBezensek, line 44: Surprised about this one as well.

apps/docs/scripts/lib/generateApiContent.ts
```diff
@@ -7,7 +7,6 @@ const { log: nicelog } = console
 export async function generateApiContent(): Promise<GeneratedContent> {
 	const articles: Articles = {}
 	const CONTENT_DIRECTORY = path.join(process.cwd(), 'content')
-	// eslint-disable-next-line @typescript-eslint/no-var-requires
```
MitjaBezensek, line 10: I saw that some of these disables are dropped, but still feels like they could  be useful in other places?

apps/examples/e2e/tests/fixtures/fixtures.ts
mimecuvalo: oh interesting, you might be right actually - lemme take a look at renaming the `use`

packages/tldraw/src/lib/ui/components/StylePanel/DoubleDropdownPicker.tsx
mimecuvalo: ah looks like it's because it was `_DoubleDropdownPickerInner` vs. `DoubleDropdownPickerInner` - i'll rename

apps/examples/e2e/tests/fixtures/fixtures.ts
mimecuvalo: ah yeah, i'll rename here, good callout

SomeHats:
(i pushed a commit to this branch going from `} catch (_e) {` to `} catch {`)

SomeHats (approved):
nice! i've been putting this off 😅

PR #4760, opened 2024-10-23
e2e scaffolding

MitjaBezensek:
Adds e2e scaffolding (playwright):
- Tests depend on `auth-setup` so the tests run as a signed in user by default. There's an example spec of showing how to test not logged in functionality.
- These were a bit tricky to setup up, especially the sidebar toggle. Let's see if they'll be flaky 🤞 
- I created a separate `e2e-dotcom` environment, so that we can use development version of clerk api keys (other existing environments use the live ones). Clerk has some special dev mode features which allow for easier signing in during testing.
- Also had some issues with `examples` `e2e` tests failing (bookmark exports on dark were failing). Seems like it was caused if I used playwright version 1.48, so I went back to 1.46. We might face this issue in the future.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- e2e scaffolding for botcom.


.github/workflows/playwright-dotcom.yml
MitjaBezensek, line 1: Felt better to create a separate workflow than to cram this one in the examples playwright tests.

apps/dotcom/client/package.json
```diff
@@ -72,6 +76,9 @@
 		},
 		"setupFiles": [
 			"./setupTests.js"
+		],
+		"testPathIgnorePatterns": [
+			"/e2e/"
```
MitjaBezensek, line 81: Jest wanted to run these tests.

.github/workflows/playwright-dotcom.yml
```diff
@@ -0,0 +1,86 @@
...
+            ~/.cache/ms-playwright
+          key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
+
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps chromium chrome
+        if: steps.playwright-cache.outputs.cache-hit != 'true'
+
+      - name: Run Playwright tests
+        run: |
+          echo "CLERK_PUBLISHABLE_KEY=${{ secrets.VITE_CLERK_PUBLISHABLE_KEY }}" > apps/dotcom/sync-worker/.dev.vars
```
MitjaBezensek, line 53: Worker didn't pick up things from `env`, so using this instead.

apps/dotcom/client/playwright.config.ts
```diff
@@ -0,0 +1,89 @@
...
+		// {
+		//   name: 'Google Chrome',
+		//   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+		// },
+	],
+
+	/* Run your local dev server before starting the tests */
+	webServer: {
+		command: 'yarn dev-app',
+		url: 'http://localhost:3000',
```
mimecuvalo, line 84: should this go to /q to start with?

apps/dotcom/client/e2e/tests/homepage.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+test('can toggle sidebar', async ({ editor, sidebar }) => {
+	await editor.ensureSidebarClosed()
+	await expect(sidebar.sidebarLogo).not.toBeVisible()
+	await editor.toggleSidebar()
+	await expect(sidebar.sidebarLogo).toBeVisible()
+})
+
+test('can create new document', async ({ page, editor, sidebar }) => {
+	await editor.ensureSidebarOpen()
```
mimecuvalo, line 16: i think my only main feedback is it feels like we're going to need open up the sidebar for most, or at least a lot of tests. feels kind of annoying to have to do that every time.

i guess it's fine for now but i do wonder if we should default it to being open on e2e tests, not sure.

mimecuvalo (approved):
LG! sry for the slight delay

apps/dotcom/client/playwright.config.ts
MitjaBezensek: Playwright will do a simple check if the server at this url is available. If it is then it will start the tests, otherwise it will run the command above to start the it.

apps/dotcom/client/e2e/tests/homepage.spec.ts
MitjaBezensek: Yeah. We can also add a helper to the `HomePage` fixture: `gotoAndOpenSiderbar`. You need to go to some url to start the test anyway so combining it with opening the sidebar might be good enough.

apps/dotcom/client/playwright.config.ts
mimecuvalo: Sure sure, I understand, but I'm wondering if the initial `goto`'s might be just the tiniest bit faster if they go from the / to /q. Eh, it doesn't really matter I suppose b/c eventually it should be root / anyway when we officially launch

MitjaBezensek: Playwright won't load any website, until you tell it to. So the `goto` will be the first navigation that is performed. I think this just does a simple ping at the start, then forgets about it.

mimecuvalo: Oh ok, thanks for clearing that up then - I didn't realize that!

PR #4761, opened 2024-10-23
[botcom] Shared file fixes

ds300:
making it so that

- you can't rename other people's files
- you can't delete other people's files
- shared files appear immediately in your nav menu
- you can't change the share settings of other people's files

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -290,8 +290,19 @@ export class TldrawApp {
 		return file.ownerId === this.getCurrentUserId()
 	}
 
-	async deleteFile(_fileId: TldrawAppFileId) {
-		this.store.remove([_fileId])
+	async deleteOrForgetFile(fileId: TldrawAppFileId) {
+		if (this.isFileOwner(fileId)) {
+			this.store.remove([fileId])
+		} else {
+			const myId = this.getCurrentUserId()
+			const fileVisits = this.getAll('file-visit')
+				.filter((r) => r.fileId === fileId && r.ownerId === myId)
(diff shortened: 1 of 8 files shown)


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -271,8 +277,19 @@ export class TldrawApp {
...
 	}
 
-	async deleteFile(_fileId: TldrawAppFileId) {
-		this.store.remove([_fileId])
+	async deleteOrForgetFile(fileId: TldrawAppFileId) {
+		if (this.isOwner(fileId)) {
+			this.store.remove([fileId])
+		} else {
+			const myId = this.getCurrentUserId()
+			const fileVisits = this.getAll('file-visit')
```
mimecuvalo, line 285: i'm just realizing that this will grow prohibitively expensive as this list starts to include every visit x every user on the site, right?
any mitigations here to prevent this? (i see we do this in `getUserSharedFiles` as well)

mimecuvalo (commented):
lookin' good - just one concern with `getAll`

mimecuvalo:
looks like you and @MitjaBezensek might collide a bit with the publishing PR, just a small heads up

apps/dotcom/client/src/tla/app/TldrawApp.ts
ds300: scaling for this will be solved when we eventually switch to a real db. and anyway I'm about to remove this record type.

mimecuvalo: i can't tell what's fake and what's real anymore :insert-some-matrix-gif:

PR #4762, opened 2024-10-23
make options object stable

SomeHats:
If you write code like `<Tldraw options={{maxPages: 1}} />` that also e.g. sets state on mount or otherwise re-renders, getting a fresh options object will cause the editor to reload even though the options are the same. You have to memoize options, or specify them outside of your component.

This diff uses `useShallowObjectIdentity` to stop identical sets of options from causing the editor to get recreated.

### Change type

- [x] `api`

### Release notes

- Writing `options` inline in the Tldraw component will no longer cause re-render loops


mimecuvalo (commented):
neat! nice catch

PR #4764, opened 2024-10-23
botcom: alternative to multi-menu items

mimecuvalo:
We talked today about an alternative to the multi-menu options, an alternative to this PR https://github.com/tldraw/tldraw/pull/4699 . Instead of the hamburger menu being a QuickActions area, that we could combine it with the vertical dot menu at the top, like so:

Screenshot 2024-10-23 at 17 40 10

This might be a good compromise of the two worlds.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


ds300:
big improvement 🙌🏼

PR #4769, opened 2024-10-24
[botcom] fix inline renaming

ds300:
Clicking the name to rename was broken

### Change type

- [x] `other`


mimecuvalo:
doh, thank you!

PR #4771, opened 2024-10-24
make sure copy-as-png comes in at natural size

SomeHats:
Browsers sanitize image formats to prevent security issues when pasting between applications. For
paste within an application though, some browsers (only chromium-based browsers as of Nov 2024)
support custom clipboard formats starting with "web " which are unsanitized. Our PNGs include a
special chunk which indicates they're at 2x resolution, but that normally gets stripped - so if
you copy as png from tldraw, then paste back in, the resulting image will be 2x the expected
size. 

To work around this, this diff writes 2 version of the image to the clipboard - the normal png, and
the same blob with a custom mime type. When pasting, we check first for the custom mime type, and
if it's there, use that instead of the normal png.

### Change type

- [x] `bugfix`

### Release notes

-  Shapes copied as PNG will have the same size when pasted back into tldraw.


steveruizok:
You might have a different angle of attack here, but this bug is caused by the clipboard API stripping out certain metadata chunks in PNGs. It can be reproduced in Figma, etc. By comparison, export as PNG doesn't have the issue.

SomeHats:
> You might have a different angle of attack here, but this bug is caused by the clipboard API stripping out certain metadata chunks in PNGs. It can be reproduced in Figma, etc. By comparison, export as PNG doesn't have the issue.

This uses the new custom formats in the clipboard API (https://developer.chrome.com/blog/web-custom-formats-for-the-async-clipboard-api) to work around this. We write both a normal `image/png` (for pasting into other apps), and `web image/vnd.tldraw+png` (for pasting back into tldraw). The image/png gets metadata stripped, but the custom one doesn't

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -669,16 +675,27 @@ export function useNativeClipboardEvents() {
...
-			// First try to use the clipboard data on the event
-			if (e.clipboardData && !editor.inputs.shiftKey) {
-				handlePasteFromEventClipboardData(editor, e.clipboardData, point)
-			} else {
-				// Or else use the clipboard API
-				navigator.clipboard.read().then((clipboardItems) => {
-					if (Array.isArray(clipboardItems) && clipboardItems[0] instanceof ClipboardItem) {
-						handlePasteFromClipboardApi(editor, clipboardItems, point)
+			const pasteFromEvent = () => {
+				if (e.clipboardData) {
```
mimecuvalo, line 679: this isn't checking `shiftKey` anymore — shiftKey should do a "normal" / non-rich paste. is that being captured somewhere?

packages/tldraw/src/lib/utils/export/copyAs.ts
```diff
@@ -26,23 +37,18 @@ export function copyAs(
 	// https://bugs.webkit.org/show_bug.cgi?id=222262
 
 	if (!window.navigator.clipboard) return Promise.reject(new Error('Copy not supported'))
-	if (window.navigator.clipboard.write) {
+	if (window.navigator.clipboard.write as any) {
```
mimecuvalo, line 40: nit: `window.navigator` is unnecessary. you can just do `navigator`

packages/tldraw/src/lib/utils/clipboard.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// Browsers sanitize image formats to prevent security issues when pasting between applications. For
+// paste within an application though, some browsers (only chromium-based browsers as of Nov 2024)
+// support custom clipboard formats starting with "web " which are unsanitized. Our PNGs include a
+// special chunk which indicates they're at 2x resolution, but that normally gets stripped - so if
+// you copy as png from tldraw, then paste back in, the resulting image will be 2x the expected
+// size. To work around this, we write 2 version of the image to the clipboard - the normal png, and
+// the same blob with a custom mime type. When pasting, we check first for the custom mime type, and
+// if it's there, use that instead of the normal png.
+export const TLDRAW_CUSTOM_PNG_MIME_TYPE = 'web image/vnd.tldraw+png' as const
```
mimecuvalo, line 11: as mentioned above - we could allow the jpeg version too maybe

packages/tldraw/src/lib/utils/clipboard.ts
```diff
@@ -0,0 +1,61 @@
...
+	return (
+		typeof ClipboardItem !== 'undefined' &&
+		'supports' in ClipboardItem &&
+		(ClipboardItem.supports as (type: string) => boolean)(mimeType)
+	)
+}
+
+export function clipboardWrite(types: Record<string, Promise<Blob>>): Promise<void> {
+	// Note:  it's important that this function itself isn't async and doesn't really use promises -
+	// we need to create the relevant `ClipboardItem`s and call window.navigator.clipboard.write
```
mimecuvalo, line 37: nit: `navigator` instead of `window.navigator`

packages/tldraw/src/lib/utils/clipboard.ts
```diff
@@ -0,0 +1,61 @@
...
+	// synchronously to make sure safari knows that the user _wants_ to copy See
+	// https://bugs.webkit.org/show_bug.cgi?id=222262
+
+	const entries = Object.entries(types)
+
+	// clipboard.write will swallow errors if any of the promises reject. we log them here so we can
+	// understand what might have gone wrong.
+	for (const [_, promise] of entries) promise.catch((err) => console.error(err))
+
+	return window.navigator.clipboard.write([new ClipboardItem(types)]).catch((err) => {
```
mimecuvalo, line 47: ditto & below

mimecuvalo (commented):
lookin good 🙌 couple questions before landing

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -13,12 +14,24 @@ import {
...
 import { pasteTldrawContent } from './clipboard/pasteTldrawContent'
 import { pasteUrl } from './clipboard/pasteUrl'
 
+// Expected paste mime types. The earlier in this array they appear, the higher preference we give
+// them. For example, we prefer the `web image/png+tldraw` type to plain `image/png` as it does not
+// strip some of the extra metadata we write into it.
+const expectedPasteFileMimeTypes = [
+	TLDRAW_CUSTOM_PNG_MIME_TYPE,
+	'image/png',
+	'image/jpeg',
```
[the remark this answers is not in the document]
SomeHats: this is a hard one because most browser's just don't support writing jpeg to the clipboard in an interoperable way. We should remove that option I think.

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -669,16 +675,27 @@ export function useNativeClipboardEvents() {
...
-					if (Array.isArray(clipboardItems) && clipboardItems[0] instanceof ClipboardItem) {
-						handlePasteFromClipboardApi(editor, clipboardItems, point)
+			const pasteFromEvent = () => {
+				if (e.clipboardData) {
+					handlePasteFromEventClipboardData(editor, e.clipboardData, point)
+				}
+			}
+
+			// First try to use the clipboard API:
+			if (navigator.clipboard && navigator.clipboard.read) {
```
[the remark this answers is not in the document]
SomeHats: I couldn't find one as I was digging, and this works as you would expect in the testing I did 🤷

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
SomeHats: This shift key toggles "paste at cursor" mode, but as far as I could tell was otherwise not actually doing anything here. I don't know why it was included in that check to be honest.

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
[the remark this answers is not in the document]
mimecuvalo: ok, cool - i think removing the option is totally valid

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
mimecuvalo: ok, gotcha

mimecuvalo (approved):
👍

PR #4775, opened 2024-10-24
[botcom] fix copy for forbidden state

ds300:
Just reverting this copy change. 'Forbidden' is too archaic and too strong a word to be user-facing here. It's like an old bible word. cc @mimecuvalo 

### Change type

- [x] `other`


mimecuvalo:
Hah, ok - well, I think we obviously disagree about this (let's leave the PR open a bit longer to discuss next time).

I'll open a PR to discuss other options because not authorized isn't the right thing to convey here either. (sounds too police-like). Something like "You don't have access to this document" or something along those lines. It should convey that the document does exist, they're in the right place, they're logged in (we know who they are), but they just don't have the right perms.

PR #4778, opened 2024-10-24
[botcom] add `yarn reset-db` command

ds300:
oversight, should have added this when I added the script

### Change type

- [x] `other`


mimecuvalo:
ty ty ty ty

PR #4781, opened 2024-10-24
[infra] limit skew protection to one month

ds300:
vercel has a 15,000 file limit that we were exceeding by 300 or so

maybe it's time to ship assets to s3

### Change type

- [x] `other`

### DIFF
--- a/internal/scripts/deploy-dotcom.ts
+++ b/internal/scripts/deploy-dotcom.ts
@@ -387,9 +387,9 @@ async function coalesceWithPreviousAssets(assetsDir: string) {
 
 	// Always include the assets from the directly previous build, but also if there
 	// have been more deploys in the last six months, include those too.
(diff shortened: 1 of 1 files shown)


internal/scripts/deploy-dotcom.ts
```diff
@@ -387,9 +387,9 @@ async function coalesceWithPreviousAssets(assetsDir: string) {
 
 	// Always include the assets from the directly previous build, but also if there
 	// have been more deploys in the last six months, include those too.
-	const sixMonths = 1000 * 60 * 60 * 24 * 30 * 6
+	const oneMonth = 1000 * 60 * 60 * 24 * 30
```
mimecuvalo, line 390: can't we filter on count _and_ time? if the 15,000 limit is one of the factors to consider

PR #4786, opened 2024-10-25
Organize paths.

MitjaBezensek:
Organize file paths. Should help us stay a bit more consistent when defining and using paths of our app.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Help with tla route / paths organization.


apps/dotcom/client/src/tla/utils/urls.ts
```diff
@@ -1,11 +1,21 @@
-export function getFilePath(fileId: string): string {
-	return `/q/f/${fileId.split(':').pop()}`
+import { PREFIX } from '../../routes'
+
+export const PATH = {
```
mimecuvalo, line 3: i'd be reticent to have this one large object (eventually with a big app we could have a lot of endpoints).

feels like, for tree-shaking purposes especially, we can just have individual fn exports. i see the appeal for having it one `PATH` object but eventually things will get more complex.

two things to consider:
- constructing URL's in a relative vs. absolute manner down the line (useful, say, if and when this logic is shared on the backend and it needs to construct these URLs, very likely)
- parsing URLs so that we can detect where the user is at

here's an example of pattern i've used on before to consider, rough sketch:
```
export function profileUrl(username: string): string {
  return buildUrl({ pathname: `/${username}` });
}

export function buildUrl({
  pathname,
  searchParams,
  isAbsolute,
}: {
  pathname: string;
  searchParams?: { [key: string]: string };
  isAbsolute?: boolean;
}): string {
  let url = '';

  if (isAbsolute) {
    url += window.location.origin
  }
  url += `/${PREFIX.tla}`
  url += pathname

  if (searchParams) {
    url += '?' + new URLSearchParams(searchParams).toString();
  }

  return url;
}
```

so i tend to think of this as a URL factory type deal. it helps to avoid having to recreate the same pattern over-and-over (the `/${PREFIX.tla}` for example) and also helps to limit duplication of logic we'll need if there's search parameters to add later.

mimecuvalo (commented):
thx for tackling this! i love it when an app has this all consolidated.

let's discuss further the structure though, see my comment below.

apps/dotcom/client/src/tla/utils/urls.ts
MitjaBezensek: I think a single object works for now. It will take a while for us to have an app with so many routes that this becomes a problem (if ever). On top of that I really like the experience of one object: you only have to remember the top level name of the object then autocomplete helps with discovering the existing routes instead of having to jump to the file to see which ones are available and how they are called.

mimecuvalo (approved):
ty 🙏

PR #4793, opened 2024-10-26
Permit drag out of toolbar

steveruizok:
This PR adds the ability to drag tools out of the toolbar and onto the canvas.

Kapture 2024-10-26 at 11 11 44

## Interaction notes

A user should be able to interact with the toolbar items in a few ways:
1. Click on the item to select the tool / activate its `onSelect` (on touch start or pointer up).
2. Click and drag to create the shape, begin translating, and display the tool as active; then drop to create the shape and in the case of text and notes begin editing it
3. Click and drag and press Escape to cancel

## Implementation

This is very rough and involves a lot of duplication. It's not entirely clear to me what the right API here is, so **I'd love to get some input** on what we could do to make things easier.

Some open questions:

1. Should we abstract the "create on drag start" behavior, or provide better information (ie the current page position) as the parameters to that callback? What about behaviors like "select and start editing after drag end"?
2. Is it the right idea to move the pointer move event to the container, so that it occurs in front of the UI?

## Testing

Since this technically occurs at the UI layer, we'd have to write e2e tests to verify that things are working the way we expect them to.

### Change type

- [x] `improvement`

### Test plan

1. Drag tools from the toolbar onto the canvas.
1. Drag tools from the toolbar overflow menu onto the canvas.

### Release notes

- Adds the ability to drag shapes from the toolbar onto the canvas.

### API changes

- Adds `onDragFromToolbarToCreateShape`, `OnDragFromToolbarToCreateShapesOpts`, and `onDragStart` on `TldrawUiMenuItem` to facilitate dragging shapes out of a toolbar


mimecuvalo:
from a UX perspective we should absolutely do this!

> This is very rough and involves a lot of duplication. It's not entirely clear to me what the right API here is, so I'd love to get some input on what we could do to make things easier.

the first thing that comes to mind is maybe having a new class of tool state, so you could do something like: `editor.setCurrentTool('arrow.drag_create')` and having it delegate to the tool what should happen next.

> Should we abstract the "create on drag start" behavior, or provide better information (ie the current page position) as the parameters to that callback? What about behaviors like "select and start editing after drag end"?

i think yes (basically what i was saying above). i think it feels like passing control to the tool state would feel right.

steveruizok:
> from a UX perspective we should absolutely do this!
> 
> > This is very rough and involves a lot of duplication. It's not entirely clear to me what the right API here is, so I'd love to get some input on what we could do to make things easier.
> 
> the first thing that comes to mind is maybe having a new class of tool state, so you could do something like: `editor.setCurrentTool('arrow.drag_create')` and having it delegate to the tool what should happen next.
> 
> > Should we abstract the "create on drag start" behavior, or provide better information (ie the current page position) as the parameters to that callback? What about behaviors like "select and start editing after drag end"?
> 
> i think yes (basically what i was saying above). i think it feels like passing control to the tool state would feel right.

We'd still have

SomeHats:
@steveruizok I added a few changes here but mostly cleanup. Otherwise, I think this is good to go!

In response to questions:
> 1. Should we abstract the "create on drag start" behavior, or provide better information (ie the current page position) as the parameters to that callback? What about behaviors like "select and start editing after drag end"?

I went with a little helper method here that contains the basic information. These seems like about the right level of abstraction: it's there to use if you want and captures the most common case, but you can still do your own thing if you have a more advanced use-case. Take a look and let me know what you think!

> 2. Is it the right idea to move the pointer move event to the container, so that it occurs in front of the UI?
I actually took this further and moved it to the whole body. We should maybe discuss this offline, but I definitely think that right now, it can be a bit confusing that once your cursor leaves the canvas and is on the UI (or outside of it), tldraw just has no idea where it is. What are the potential down-sides to this approach? We should probably flag it the changelog if we go ahead with this

steveruizok:
Amazing! I'll take a look this afternoon.

mimecuvalo:
One little polish touch that is nice in Figma is making it _feel_ like you're dragging something out of the toolbar:

https://github.com/user-attachments/assets/bbb992a4-c0bf-407a-9618-dc8e812d0ef6

ds300:
Closing due to inactivity

mimecuvalo:
Resurrecting! We could use this for the workflow builder @steveruizok

mimecuvalo:
one issue that needs resolving before landing:

https://github.com/user-attachments/assets/d18a85b7-2b9c-4c67-b82f-87b85004e7be

PR #4795, opened 2024-10-26
Better support scale / quality in export utilities

steveruizok:
This PR allows for better control over scale when using the image export utilities in tldraw.

- `scale` is used for just the output scale of an image, relative to the size in tldraw
- `pixelRatio` is used for the pixelRatio scale, default is 2

A shape that is 100x100 in tldraw:
- with { scale: 1 } should export as a 100x100 svg
- with { scale: 2 } should export as a 200x200 svg
- with { scale: 1, pixelRatio: 1 } should export as a 100x100 png at 72dpi
- with { scale: 1, pixelRatio: 2 } should export as a 200x200 png at 144dpi
- with { scale: 2, pixelRatio: 2 } should export as a 400x400 png at 144dpi

Note, I tried deprecating scale instead of renaming to avoid the breaking change, but I don't think we can get away with that since it gets spread in from other places (where scale is used for actual 1x 2x scale).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Export an image at scale 1.
2. Export an image at scale 2.
2. Export an image at scale 3.

### Release notes

- Improved treatment of `scale` in image copy / export utilities.

### DIFF
--- a/packages/tldraw/src/lib/utils/export/export.ts
+++ b/packages/tldraw/src/lib/utils/export/export.ts
@@ -28,8 +28,8 @@ export async function getSvgAsImage(
 	const { type, quality, scale, width, height } = options
 
 	let [clampedWidth, clampedHeight] = await clampToBrowserMaxCanvasSize(
-		width * scale,
-		height * scale
+		width * scale * 2,
+		height * scale * 2
 	)
 	clampedWidth = Math.floor(clampedWidth)
 	clampedHeight = Math.floor(clampedHeight)
@@ -104,7 +104,7 @@ export async function getSvgAsImage(
 
 async function getSvgString(editor: Editor, ids: TLShapeId[], opts: TLImageExportOptions) {
 	const svg = await editor.getSvgString(ids?.length ? ids : [...editor.getCurrentPageShapeIds()], {
-		scale: 1,
+		scale: opts.scale ?? 1,
 		background: editor.getInstanceState().exportBackground,
 		...opts,
 	})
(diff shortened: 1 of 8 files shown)


packages/tldraw/src/lib/utils/export/export.ts
```diff
@@ -19,17 +19,17 @@ export async function getSvgAsImage(
...
 		type: 'png' | 'jpeg' | 'webp'
-		quality: number
-		scale: number
 		width: number
 		height: number
+		quality?: number
+		pixelRatio?: number
 	}
 ) {
-	const { type, quality, scale, width, height } = options
```
steveruizok, line 28: the use of scale in these options was confusing, as it's different from the scale that's in `TLImageExportOptions`

packages/tldraw/src/lib/utils/export/export.ts
```diff
@@ -19,17 +19,17 @@ export async function getSvgAsImage(
...
+	const { type, width, height, quality = 1, pixelRatio = 2 } = options
```
mimecuvalo, line 28: should this default to `window.devicePixelRatio` or `editor.getInstanceState().devicePixelRatio`?

steveruizok: no, even for 1x projects we want to create it at 2x (144ppi, rather than 72ppi).

PR #4796, opened 2024-10-26
Make default color theme light.

steveruizok:
This PR:
- makes the default user preference light, rather than system
- updates the theme when the media theme changes only if the user has selected system

## Rationale

Our light theme is better and presents a better initial experience of the app.
Our logged out pages only support light mode and we don't want to show a dark editor on a light page.
Our docs landing page has light mode as its default and we don't want to show a dark editor on a light page.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Sets the default color theme to light.

### DIFF
--- a/packages/editor/src/lib/editor/managers/UserPreferencesManager.ts
+++ b/packages/editor/src/lib/editor/managers/UserPreferencesManager.ts
@@ -12,7 +12,7 @@ export class UserPreferencesManager {
 		if (typeof window === 'undefined' || !('matchMedia' in window)) return
 
 		const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)')
-		if (this.getUserPreferences().colorScheme === 'system' && darkModeMediaQuery?.matches) {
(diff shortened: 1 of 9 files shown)


packages/editor/src/lib/editor/managers/UserPreferencesManager.ts
```diff
@@ -12,14 +12,16 @@ export class UserPreferencesManager {
...
-			} else {
-				this.systemColorScheme.set('light')
+			if (this.getUserPreferences().colorScheme === 'system') {
+				if (e.matches) {
+					this.systemColorScheme.set('dark')
+				} else {
+					this.systemColorScheme.set('light')
+				}
 			}
 		})
```
ds300, line 26: i'm gonna revert these changes, this introduces bugs around switching theme. i.e. if your system is set up to prefer dark and you select 'system' then it won't switch.

I think you already achieved the goal of this PR by just updating the default value to be 'light' instead of 'system'

PR #4798, opened 2024-10-26
[botcom] Duplicate / Publish / Create / Delete files on the server 

steveruizok:
This PR improves and creates functionality on the backend.

For the most part, simple changes such as creating a file or changing its name rely on optimistic updates. However, some actions rely on either sending information to the server directly. While we can make optimistic updates in these situations, we also need to revert those changes if the request fails.

## Duplicating files

This PR adds duplication on the server. Previously, the duplicate feature required that the duplicating file be open, so duplicating other files (from the sidebar) would duplicate the current open file instead. Now the server will pull the latest file from the tldraw worker and duplicate that instead.

If the user is viewing the duplicated page, we redirect them to the new page.

## Deleting / forgetting files

This PR adds deletion on the server. If the user owns a file, they will optimistically delete the file and all states associated with the file; if the user does not, then they'll only "forget" the file by optimistically deleting all of their own states associated with the file. Meanwhile, the server will actually delete the file and the file's published snapshot. If the delete fails, the user's optimistic updates will be reversed.

If the user is viewing the deleted / forgotten file, we redirect them to the root.

## Publishing / un-publishing files

This PR adds publishing on the server. Previously, the user would upload the current editor state to the server. Now, the server pulls the latest file from the tldraw worker and creates a published snapshot based on that instead.

I'd still like to build the "publish history" of versions that a user has published for a given file, rather than overwriting the single published snapshot file.

## Creating files

This PR adds support for .tldr file drops. Files are created on the server and then added to the user's own files.

## Cleanup

The sync-worker was getting pretty messy so I did some cleanup. Mostly, I've moved the endpoints that are related to the app to a `tla` folder, as I've done elsewhere. It wasn't clear to me which endpoints were "purely" working for the app layer and which weren't.

### Change type

- [x] `other`


steveruizok:
As with other PRs up right now, the end to end tests here will never pass because they've been replaced by the other e2e tests. Once approved, we'll need to bypass branch protections to land it.

apps/dotcom/sync-worker/src/routes/tla/duplicateFile.ts
```diff
@@ -0,0 +1,69 @@
...
+		// A user can duplicate other users' files only if they are shared
+		if (file.ownerId !== userId) {
+			if (!file.shared) {
+				throw Error('forbidden')
+			}
+		}
+
+		// Get the current serialized snapshot of the room (by waking up the worker,
+		// if we have to). Why not grab from the database directly? Because the worker
+		// only persists every ~30s while users are actively editing the room. If we
```
ds300, line 39:
```suggestion
		// only persists every ~10s while users are actively editing the room. If we
```

apps/dotcom/sync-worker/src/utils/tla/getCurrentSerializedRoomSnapshot.ts
```diff
@@ -0,0 +1,15 @@
+import { ROOM_PREFIX } from '@tldraw/dotcom-shared'
+import { Environment } from '../../types'
+
+export async function getCurrentSerializedRoomSnapshot(
+	roomId: string,
+	env: Environment
+): Promise<string> {
+	const id = env.TLDR_DOC.idFromName(`/${ROOM_PREFIX}/${roomId}`)
+	const worker = env.TLDR_DOC.get(id)
+	const snapshot = worker.getCurrentSerializedSnapshot()
```
ds300, line 10:
```suggestion
	const snapshot = await worker.getCurrentSerializedSnapshot()
```

apps/dotcom/sync-worker/src/utils/tla/getTldrawAppFileRecord.ts
```diff
@@ -0,0 +1,14 @@
...
+export async function getTldrawAppFileRecord(
+	slug: string,
+	env: Environment
+): Promise<TldrawAppFile | null> {
+	const d1Result = await env.DB.prepare('select * from records where topicId = ?1 and id = ?2')
+		.bind(APP_ID, `file:${slug}`)
+		.first()
+	if (!d1Result) return null
+	return JSON.parse((d1Result as any).record)
+}
```
ds300, line 14: this should probably talk to the durable object instead of the DB, to avoid potential race conditions

packages/sync-core/src/lib/TLSocketRoom.ts
```diff
@@ -268,6 +268,14 @@ export class TLSocketRoom<R extends UnknownRecord = UnknownRecord, SessionMeta =
 		return this.room.getSnapshot()
 	}
 
+	/**
+	 * Return a serialized snapshot of the document state, including clock-related bookkeeping.
+	 * @returns The serialized snapshot
+	 */
+	getCurrentSerializedSnapshot() {
+		return JSON.stringify(this.room.getSnapshot())
+	}
```
ds300, line 277: let's mark this as internal for now. don't want to pollute the public api with trivial helper stuff

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -219,123 +223,246 @@ export class TldrawApp {
...
+		if (this.isFileOwner(fileId)) {
+			// Optimistic update, remove file and file states
+			const userId = this.getCurrentUserId()
+			fileStates = this.getAll('file-state').filter(
+				(r) => r.fileId === fileId && r.ownerId === userId
+			)
+			this.store.remove([fileId, ...fileStates.map((s) => s.id)])
+		} else {
+			// If not the owner, just remove the file state
+			fileStates = this.getAll('file-state').filter((r) => r.fileId === fileId)
```
ds300, line 414:
```suggestion
			fileStates = this.getAll('file-state').filter(
				(r) => r.fileId === fileId
			)
			this.store.remove([fileId, ...fileStates.map((s) => s.id)])
		} else {
			// If not the owner, just remove the file state
			fileStates = this.getAll('file-state').filter((r) => r.fileId === fileId && r.ownerId === userId)
```

you mistakenly switched these predicates around

ds300 (changes requested):
good improvements here, love that you've started on error handling. noticed a couple of bugs but mostly looks great

steveruizok:
All good! Thanks for the feedback

ds300 (approved):

PR #4806, opened 2024-10-28
[botcom] add max file limit

steveruizok:
This PR adds a maximum limit to the number of files a user can create.

It adds a (provisional?) config property to the app for these constants.

### Change type

- [x] `other`


mimecuvalo (commented):
is the concern here abuse? i worry that we're gonna forget about this and then user's are going to be wondering what's going on. if it's provisional i might go for something higher like 500 perhaps.

PR #4811, opened 2024-10-28
update vite

ds300:
Update vite. Maybe this fixes that aborted loading thing

### Change type

- [x] `other`


mimecuvalo:
worth a shot ¯\\_(ツ)_/¯

Issue #4804, opened 2024-10-28
Line breaking bug


steveruizok:
### To Reproduce

1. Visit [https://www.tldraw.com/s/v2_c_88xK-DMuxdSzoZpMbPKlY?d=v0.0.1920.1113.page](https://www.tldraw.com/s/v2_c_88xK-DMuxdSzoZpMbPKlY?d=v0.0.1920.1113.page) and fork the project
2. Edit the label, deleting the space after "research" and creating the space again

### Expected

The measured size of the text box should exactly reflect the measured size of the text input.

### Observed

The two measurements differ.

mimecuvalo:
Ugh, annoying — I'll take a look.

steveruizok:
This too https://www.tldraw.com/r/Py5PfKCr5RCsBzr87SpVR?d=v443.162.1920.1113.page

GeorgVolkov:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

mimecuvalo:
I looked into this more and the root of this problem is actually more tied to the Draw font. The problem seems to be the slightly random variation in the font in some renders.

This is an issue we've seen before with this font, where it doesn't always consistently produce the same output.

steveruizok:
Right, that makes sense. @ds300 you did some tweaks with the font a while back, think there's a way to reduce the variation like this? Otherwise I can also reach out to the author @arrowtype

arrowtype:
The font is deliberately made to be a bit randomized. 

If you turn off the `calt` font feature, I believe it will prevent the randomization, and that may solve this. (It's possible it's using a different feature, and I'm forgetting which one. You can test it on a website like wakamai fondue.)

mimecuvalo:
Thanks for writing in @arrowtype. It's appreciated - also the wakamai fondue site is new to me, very useful.
It looks like the `calt` feature doesn't disable randomization. For example `iii` still randomizes:
<img width="263" alt="Screenshot 2024-11-04 at 09 39 51" src="https://github.com/user-attachments/assets/5f359b08-6b01-4ea7-b402-f22c860e451d">

I recognize that this is feature of the font to be randomized (and maybe not something to turn off) so we're trying some different tactics to try to measure the text in a more consistent way on our end. 

If you have any other ideas though, lemme know!

arrowtype:
Sorry, kept meaning to check on this while at a computer.

[...]

PR #4815, opened 2024-10-28
text: fix a text measurement edge case in rendering in Chrome.

mimecuvalo:
See the attached issue https://github.com/tldraw/tldraw/issues/4804

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix a text measurement edge case in rendering in Chrome.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -847,6 +847,8 @@ input,
 .tl-text-content {
 	position: absolute;
 	inset: 0px;
+	height: 100%;
+	width: 100%;
 	min-width: 1px;
 	min-height: 1px;
 	overflow: visible;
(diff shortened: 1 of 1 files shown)


mimecuvalo:
@steveruizok pushed up an alternate fix which isn't pretty either but lemme know what you think

mimecuvalo:
I looked into this more and the root of this problem is actually more tied to the Draw font. The problem seems to be the slightly random variation in the font in some renders.

This is an issue we've seen before with this font, where it doesn't always consistently produce the same output.

steveruizok:
> I looked into this more and the root of this problem is actually more tied to the Draw font. The problem seems to be the slightly random variation in the font in some renders.
> 
> This is an issue we've seen before with this font, where it doesn't always consistently produce the same output.

I think I replied on a different thread, but let's see if we can either turn off the randomization or create a fork of the font that is deterministic. Still worried that this will be moving the problem around rather...

mimecuvalo:
for record keeping here: @ds300 is working on recompiling the font in general. (which will also remove the randomization here)

PR #4817, opened 2024-10-29
Isolate tests by resetting the db between each test.

MitjaBezensek:
Resets the db between tests.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Isolates tests by resetting the db.

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/Database.ts
+++ b/apps/dotcom/client/e2e/fixtures/Database.ts
@@ -1,10 +1,14 @@
+import { Page } from '@playwright/test'
 import { execSync } from 'child_process'
 import path from 'path'
 
 export class Database {
-	constructor() {}
+	constructor(private readonly page: Page) {}
 
 	async reset() {
+		await this.page.evaluate(() => {
+			;(window as any).tla_dev_clear()
+		})
 		const scriptPath = path.resolve(__dirname, '../../../sync-worker/')
 		execSync(`${scriptPath}/reset-db.sh > /dev/null 2>&1`, { cwd: scriptPath })
 	}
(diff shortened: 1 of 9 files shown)


apps/dotcom/client/e2e/fixtures/Database.ts
```diff
@@ -0,0 +1,11 @@
+import { execSync } from 'child_process'
+import path from 'path'
+
+export class Database {
+	constructor() {}
+
+	async reset() {
+		const scriptPath = path.resolve(__dirname, '../../../sync-worker/')
+		execSync(`${scriptPath}/reset-db.sh > /dev/null 2>&1`, { cwd: scriptPath })
```
ds300, line 9: this doesn't work on it's own, you'd need to restart wrangler too otherwise the durable object will still have all the data in memory.

MitjaBezensek: Yeah, good catch. Added a test making sure we really clear it.

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -45,6 +46,9 @@ export class TldrawApp {
 
 	private constructor(store: Store<TldrawAppRecord>) {
 		this.store = store
+		if (isDevelopmentEnv) {
+			;(window as any).tla_dev_clear = () => this.store.clear()
```
MitjaBezensek, line 50: We expose this for e2e tests. Just resting the db was not enough since things still stayed in memory.

Think this should do for now, but I did notice some issues with running the e2e with the UI mode. In some cases the worker got stuck. If we find it problematic I can also look for an alternative solution.

PR #4782, opened 2024-10-24
chore(dotcom): refine forbidden vs not authorized error handling

mimecuvalo:
followup to the convo here: https://github.com/tldraw/tldraw/pull/4775
i'll leave this open so we can discuss our opinions!

### Change type

- [x] `other` 

### Test plan

1. Navigate to a forbidden or unauthorized route in the dotcom app and verify the error screen behavior.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Refined error handling for forbidden and unauthorized states on the website.

PR #4825, opened 2024-10-29
error page: wordsmith copy again

mimecuvalo:
continuing conversation from: https://github.com/tldraw/tldraw/pull/4790#discussion_r1819122178
and i guess rehashing out what i thought we had settled in https://github.com/tldraw/tldraw/pull/4782 :)

i'm gonna pull in @steveruizok here to mediate because i think we've changed this back-and-forth enough times! let's talk about it on this PR and bring forth our points.

- i understand that currently either things are fully private or fully public but soon it will encompass invited users. _and_ we do have `NOT_AUTHENTICATED` and `FORBIDDEN` distinctions for good reason: they help guide the user to know whether the error is temporary or permanent (i'm not logged in vs. i really don't have access even though i'm logged in)
- furthermore, something labeled as a "private file", when somebody wants to ask for access, is off-putting for a human to see. it feels intrusive to ask for access to something that is explicitly called private. 
- finally, i do want to callout that we should avoid any paradigm that publicly distinguishes to a non-authenticated/non-invited user that a file is private vs. invite-only. this can be a slight leak of security as it tells a snooping person what type of file is behind the access controls. we should keep it intentionally a little obscure to avoid that.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/tla/components/TlaFileError/TlaFileError.tsx
```diff
@@ -35,15 +35,15 @@ export function TlaFileError({ error }: { error: unknown }) {
 		case TLSyncErrorCloseEventReason.NOT_AUTHENTICATED: {
 			return (
 				<TlaFileErrorContent
-					header="Private file"
-					para1="Contact the owner to request access."
+					header="Sign in"
```
ds300, line 38: We're using `Log in` and `Sign in` inconsistently, would be good to pick one and stick with it.

apps/dotcom/client/src/tla/components/TlaFileError/TlaFileError.tsx
```diff
@@ -35,15 +35,15 @@ export function TlaFileError({ error }: { error: unknown }) {
...
+					para1="You are currently logged out and need sign in to view this file."
```
ds300, line 39: this statement will be false most of the time, and it will hurt UX a little. I don't think there's a security risk in keeping it how it was. What information does it give a 'snooper' that they don't already have? 'this URL points to a tldraw file' will already be known to them.

ds300: also 'you are currently logged out' is redundant. Just say 'you need to sign in'

apps/dotcom/client/src/tla/components/TlaFileError/TlaFileError.tsx
```diff
@@ -35,15 +35,15 @@ export function TlaFileError({ error }: { error: unknown }) {
...
 					cta={<TlaSignInButton />}
 				/>
 			)
 		}
 		case TLSyncErrorCloseEventReason.FORBIDDEN: {
 			return (
-				<TlaFileErrorContent header="Private file" para1={`Contact the owner to request access.`} />
+				<TlaFileErrorContent header="Invite only" para1={`Contact the owner to request access.`} />
```
ds300, line 46: fine with this given that we will be adding more fine-grained access controls at some point

ds300 (approved):

apps/dotcom/client/src/tla/components/TlaFileError/TlaFileError.tsx
mimecuvalo: > this statement will be false most of the time

Why would the statement be false most of the time? The majority of these cases will be:
- the owner is logged out, they have the link because they've bookmarked it, they need to sign in
The other case where a file was shared but then made private is honestly more rare. We can't confidently say you need to contact the owner if a person is logged out, we just don't have enough information. But all that is moot in my mind because we will add per-email invites sooner than later.

The current copy is confusing, esp. if you're the owner: "Contact the owner" when the person reading it is the owner, and then having a CTA to sign in are sending mixed messages. There's really no way to contact the owner at the moment (although we should add that, to 'ring the doorbell').

> What information does it give a 'snooper' that they don't already have? 'this URL points to a tldraw file' will already be known to them.

Again, it's slight but it's there. If someone comes across this link (say, it was posted publicly inadvertently, in a future where we have per-email invites), the person coming across this link could discover that this is either a truly private file they've come across or a file that is shared across multiple people. They could then infer from other signals online, say if it's shared, who possibly the other invitees are and can try to phish for access from one of those invitees.

ds300: > Again, it's slight but it's there. If someone comes across this link (say, it was posted publicly inadvertently, in a future where we have per-email invites), the person coming across this link could discover that this is either a truly private file they've come across or a file that is shared across multiple people. They could then infer from other signals online, say if it's shared, who possibly the other invitees are and can try to phish for access from one of those invitees.

how could they make that distinction though? i.e. 'truly private' vs 'selectively shared' - that's not exposed. All they know is that it exists and they don't have access.

I'm gonna relent here anyway. I think your suggestions create slightly worse UX for the (imo more common) use case where it's a guest trying to visit the file, but it's not a huge difference and I don't think it's worth spending all this time fussing over.

steveruizok (approved):

PR #4826, opened 2024-10-29
dev: restore the loading state for the demo placeholder

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/4684 and https://github.com/tldraw/tldraw/pull/4686
was on my todo list to restore some of the responsiveness, esp. on slower networks

https://github.com/user-attachments/assets/6a52039d-40b0-4d4b-98f4-fde08665dfa2

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
This starts loading the iframe when the button is clicked, right? I'd prefer the current experience, which is to load the editor immediately so that there isn't a delay after that interaction. Is there a reason why we wouldn't want to start loading the editor immediately?

If the user clicks the button before the editor is fully loaded, then sure, let's show a spinner while the editor loads in. But I don't want us to _start_ loading only when the user clicks the button.

steveruizok:
Pushed a shuffle / commentup while reviewing. Looks good to me!

steveruizok (approved):

PR #4827, opened 2024-10-29
[botcom] Sidebar file operations tests

Taha-Hassan-Git:
This PR adds E2E tests for sidebar file operations.

There's another optimistic update related bug caused by switching to a file which doesn't yet exist on the server, which results in the document not loading. Works fine locally but not in the CI for some reason. Skipping this test for now.

Also tests seem flakey, there's an issue I was running into locally with the websocket server occasionally returning a 500 and causing the entire app not to load and all the tests to fail. Re-running the tests usually results in them all passing. But if this is merged, potentially having to do that on every PR won't be very fun. Not sure what the way forward is with this.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [x] End to end tests


apps/dotcom/client/e2e/tests/homepage.spec.ts
```diff
@@ -1,21 +1,66 @@
...
+		const newName = await newFileLink?.innerText()
+		expect(newName).toBe(randomString)
+	})
+
+	await test.step('duplicate the document', async () => {
+		const fileName = await sidebar.getFileLink()
+		await sidebar.openFileMenu(fileName)
+		await page.getByRole('menuitem', { name: 'Duplicate' }).click()
+		await sleep(1000)
+		const copiedFileLink = page.getByRole('link').getByText(`${fileName} Copy`)
```
Taha-Hassan-Git, line 47: I'm expecting to get the link here, but the text is in nested inside a sibling div so this isn't returning anything. I think the text should be moved inside the link so I've started a minor refactor of the TlaSidebarFileLink component.

PR #4829, opened 2024-10-30
Store all the published snapshots instead of only the last one

MitjaBezensek:
We store all the published snapshots so that we could expose published file history in the future. The way it works is:
* `/parentSlug/publishedSlug` aways contains the latest published snapshot
* we now also store the snapshot under `/parentSlug/publishedSlug-timeOfPublish` to keep track of all the published changes.

image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/apps/dotcom/sync-worker/src/routes/tla/publishFile.ts
+++ b/apps/dotcom/sync-worker/src/routes/tla/publishFile.ts
@@ -42,6 +42,11 @@ export async function publishFile(request: IRequest, env: Environment): Promise<
 			getR2KeyForRoom({ slug: `${roomId}/${file.publishedSlug}`, isApp: true }),
 			serializedSnapshot
 		)
+		const currentTime = new Date().toISOString()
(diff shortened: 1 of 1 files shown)


mimecuvalo:
i think, ever so slightly annoying, is that `publishedSlug` can have `-` as one of the characters. maybe there should be a different delimiter between `publishedSlug-timeOfPublish` so that it's easier to segment it out later if needed. 

otherwise 👍 lgtm

MitjaBezensek:
> i think, ever so slightly annoying, is that `publishedSlug` can have `-` as one of the characters. maybe there should be a different delimiter between `publishedSlug-timeOfPublish` so that it's easier to segment it out later if needed.
> 
> otherwise 👍 lgtm

Good call, went with `|`.

PR #4835, opened 2024-11-01
i18n: pipeline

mimecuvalo:
wires up botcom i18n to be an automated process:
- uploads strings automatically when pushed to `main`
- downloads translated strings automatically on a daily basis and commits to `main`. 
- also, builds/compiles strings on `pre-commit`

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek (commented):
Looks good. The only thing I'm not so certain about is committing straight to main. Should we maybe open a PR instead?

mimecuvalo:
> The only thing I'm not so certain about is committing straight to main. Should we maybe open a PR instead?

Yeah, it's worth discussing. I think the issue will be that there will potentially be an annoying amount of PRs opened up, whenever one string gets added/edited. That kind of noise might be prohibitive for development.

The main concern probably comes from something committing straight to `main` which maybe is a security risk potentially.
I think what I'll do actually is limit the `git add -A` command so that it only adds `*.json` files to be more selective in what it can commit to `main`. If we had a `develop` branch I think this might be less controversial but we don't have that type of system.

In the end though, the goal of this automation is to get less human-involvement in the process - it's not a pipeline that needs  monitoring since the Lokalise cohort of translators is vetted/monitored anyway. We want to make this part of our system less distracting for us to deal with.

PR #4842, opened 2024-11-01
Try to make the site show up on Google

steveruizok:
This PR includes a handful of changes designed to improve the site's performance in search. This includes:

- tightening up meta tags in head
- adding some visually hidden HTML content
- adding a minimal sitemap
- moving core styles into a style tag so that things don't load wrong prematurely

If you have any other thoughts, let's try them!

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/styles/core.css
```diff
@@ -1,4 +1,10 @@
-@import url('https://fonts.googleapis.com/css2?family=Inter:wght@500;700&display=swap');
+/* NB: If we did position: fixed, Safari on iOS 18 and MacOS 15 would
```
mimecuvalo, line 1: not the biggest deal but removing the font import here affects the BoardHistorySnapshot (which doesn't pull in globals.css)

apps/dotcom/client/index.html
```diff
@@ -48,12 +92,18 @@
...
+
 		<title>tldraw</title>
 	</head>
 
 	<body>
-		<div id="root" class="site-wrapper"></div>
-		<script type="module" src="./src/main.tsx"></script>
 		<noscript>You need to enable JavaScript to run tldraw. ✌️</noscript>
+		<header>
+			<h1 class="visually-hidden">tldraw</h1>
```
mimecuvalo, line 103: crawlers are smart about hidden stuff - they don't index things like this, afaik. is this trying to avoid the 'soft 404'? i think Lu and i had tried this fwiw...

apps/dotcom/client/index.html
```diff
@@ -48,12 +92,18 @@
...
+			<p class="visually-hidden">A free and instant collaborative whiteboarding tool.</p>
+		</header>
+		<div id="root"></div>
+		<script type="module" src="./src/main.tsx"></script>
```
mimecuvalo, line 107: i would put `<noscript>` last btw

apps/dotcom/client/index.html
```diff
@@ -48,12 +92,18 @@
 		<meta property="og:site_name" content="tldraw" />
 		<meta property="og:url" content="https://www.tldraw.com/" />
 		<meta property="og:image" content="https://www.tldraw.com/social-og.png" />
+		<meta property="og:image:alt" content="A collaborative whiteboarding tool interface" />
```
mimecuvalo, line 95: interface? maybe a touch too geeky

apps/dotcom/client/index.html
```diff
@@ -1,34 +1,77 @@
...
-		<link rel="apple-touch-icon" href="/apple-touch-icon.png" />
-		<link rel="apple-touch-icon" sizes="152x152" href="/apple-touch-icon-152x152.svg" />
-		<link rel="apple-touch-icon" sizes="167x167" href="/apple-touch-icon-167x167.svg" />
-		<link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon-180x180.svg" />
-
 		<!-- $PRELOADED_FONTS -->
-		<link rel="preconnect" href="https://fonts.googleapis.com" />
-		<link rel="preconnect" href="https://fonts.gstatic.com" />
+		<link rel="preconnect" href="https://fonts.googleapis.com" crossorigin="anonymous" />
+		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous" />
```
mimecuvalo, line 74: looks like, judging from their docs it should be like this:
https://fonts.google.com/knowledge/using_type/using_web_fonts_from_a_font_delivery_service
```
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
```

apps/dotcom/client/index.html
```diff
@@ -1,34 +1,77 @@
...
+			#root {
+				position: relative;
+				inset: 0;
+				width: 100%;
+				height: 100%;
+				overflow: hidden;
+			}
 
+			.visually-hidden {
+				position: absolute !important;
```
mimecuvalo, line 37: why !important?

apps/dotcom/client/index.html
```diff
@@ -1,34 +1,77 @@
...
+			content="whiteboard, canvas, collaborative whiteboard, online drawing, team collaboration tool, free whiteboard app"
+		/>
+		<script type="application/ld+json">
+			{
+				"@context": "https://schema.org",
+				"@type": "SoftwareApplication",
+				"name": "tldraw",
+				"url": "https://www.tldraw.com/",
+				"applicationCategory": "BusinessApplication",
+				"operatingSystem": "WEB",
```
mimecuvalo, line 19: heh nice

apps/dotcom/client/index.html
```diff
@@ -1,34 +1,77 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" dir="ltr">
 	<head>
 		<meta charset="UTF-8" />
 
 		<link rel="manifest" href="/manifest.webmanifest" />
+		<link rel="canonical" href="https://www.tldraw.com/" />
```
mimecuvalo, line 7: can't use canonical here like this unfortunately :-/

the problem is that this `index.html` file is used on all paths for the site at the moment. so, effectively, on any room we'll be marking the canonical URL as the homepage. we'd have to add this in dynamically.

i tried doing this in https://github.com/tldraw/tldraw/pull/4471 but then realized it wasn't worth doing.

mimecuvalo (commented):
lookin' good - i hope it helps resolve some of the soft 404 stuff.

only blocker is the `canonical` bit which we should remove

apps/dotcom/client/index.html
steveruizok: not sure, but I have a feeling `visually-hidden` is a common class name... browser extensions maybe? I'm running on guides and references here, so cargo cult

apps/dotcom/client/index.html
steveruizok: I read some suggestions that anonymous improves things. Possible cargo culting

apps/dotcom/client/styles/core.css
```diff
@@ -1,4 +1,10 @@
...
+ * have several rendering issues. This must stay position: absolute. 
+ */
+.tldraw__editor {
+	height: 100%;
+	position: relative;
```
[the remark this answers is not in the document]
steveruizok: Removing, it's a bug with fixed but not with absolute.

apps/dotcom/client/index.html
steveruizok: Right on, though I'll keep it in for accessibility.

steveruizok: *hand waving*

PR #4846, opened 2024-11-03
Fix text layout bugs

steveruizok:
This PR fixes text layout bugs with labels. https://github.com/tldraw/tldraw/issues/4804

The issue was that text inputs and text elements sometimes have very small differences in how they measure text. Previously, we were using a `div` to set the width and height of the label. This size was inherited by the parent container. The `div` had a sibling, the `textarea`, that would take up its container width and height. However, if the textarea overflowed this size, we would have a problem where some text was not displayed. This disagreement _mostly_ happened when the computed size was a floating point number.

This PR removes the complex layout juggle and uses our text measurement as the authoritative size.

### Change type

- [x] `bugfix`

### Test plan

1. Create lots of text with various content, line breaks, widths, and heights.

### Release notes

- Fixed a bug with line breaks inside of text labels.

### DIFF
--- a/packages/editor/src/lib/editor/managers/TextManager.ts
+++ b/packages/editor/src/lib/editor/managers/TextManager.ts
@@ -39,6 +39,8 @@ const spaceCharacterRegex = /\s/
 /** @public */
 export class TextManager {
 	baseElm: HTMLDivElement
+	prevElm?: HTMLDivElement
+	debug = true
 
 	constructor(public editor: Editor) {
 		const container = this.editor.getContainer()
@@ -46,6 +48,7 @@ export class TextManager {
 		const elm = document.createElement('div')
 		elm.classList.add('tl-text')
 		elm.classList.add('tl-text-measure')
+
 		elm.tabIndex = -1
 		container.appendChild(elm)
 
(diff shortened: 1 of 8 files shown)


packages/editor/src/lib/editor/managers/TextManager.ts
```diff
@@ -40,12 +40,16 @@ const spaceCharacterRegex = /\s/
 export class TextManager {
 	baseElm: HTMLDivElement
 
+	prevElm?: HTMLDivElement
+	debug = false
```
steveruizok, line 44: got tired of writing and removing these lines every time I have to debug text measurement, it's blessed now

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -848,40 +853,52 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
...
+		if (!scaledLabelSizes[font]) {
+			scaledLabelSizes[font] = {}
+		}
+		minSize = editor.textMeasure.measureText('W', {
+			...TEXT_PROPS,
+			fontFamily: FONT_FAMILIES[font],
+			fontSize: LABEL_FONT_SIZES[size],
+			maxWidth: 100, // ?
+		})
+		scaledLabelSizes[font][size] = minSize
```
steveruizok, line 885: We were measuring the W character every time. Measurement is one of the slower things, so let's just cache a little bit instead.

steveruizok:
I'm seeing some bugs when a shape's scale is different, possibly due to how the DOM rounds decimals

Kapture 2024-11-03 at 15 01 52

PR #4851, opened 2024-11-04
Smart bringForward/sendBackward

ds300:
This PR makes it so that only overlapping shapes are considered when moving forward/backward incrementally. It feels great.
Before
Kapture 2024-11-04 at 12 36 22

After
Kapture 2024-11-04 at 12 36 22

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improved the 'bring forward' and 'send backward' actions by making them only consider nearby overlapping shapes when deciding the next ordering.

### DIFF
--- a/packages/editor/src/lib/utils/reorderShapes.ts
+++ b/packages/editor/src/lib/utils/reorderShapes.ts
@@ -178,6 +178,7 @@ function filterOverlappingShapes(editor: Editor, moving: Set<TLShape>, children:
 /**
  * Reorders the moving shapes forward in the parent's children.
  *
+ * @param editor The editor
  * @param moving The set of shapes that are moving
  * @param children The parent's children
  * @param changes The changes array to push changes to
@@ -228,6 +229,7 @@ function reorderForward(
 /**
  * Reorders the moving shapes backward in the parent's children.
  *
+ * @param editor The editor
  * @param moving The set of shapes that are moving
  * @param children The parent's children
  * @param changes The changes array to push changes to
--- a/packages/editor/src/lib/utils/reorderShapes.ts
+++ b/packages/editor/src/lib/utils/reorderShapes.ts
@@ -1,6 +1,8 @@
 import { TLParentId, TLShape, TLShapeId, TLShapePartial } from '@tldraw/tlschema'
 import { IndexKey, compact, getIndicesBetween, sortByIndex } from '@tldraw/utils'
 import { Editor } from '../editor/Editor'
+import { Vec } from '../primitives/Vec'
+import { polygonsIntersect } from '../primitives/intersect'
 
 export function getReorderingShapesChanges(
 	editor: Editor,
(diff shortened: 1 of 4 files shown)


packages/editor/src/lib/utils/reorderShapes.ts
```diff
@@ -147,14 +149,48 @@ function reorderToFront(moving: Set<TLShape>, children: TLShape[], changes: TLSh
...
  * @param changes The changes array to push changes to
  */
-function reorderForward(moving: Set<TLShape>, children: TLShape[], changes: TLShapePartial[]) {
+function reorderForward(
+	editor: Editor,
+	moving: Set<TLShape>,
+	children: TLShape[],
+	changes: TLShapePartial[]
+) {
+	children = filterOverlappingShapes(editor, moving, children)
```
SomeHats, line 192: Possibly handled by the jitter that we add to fractional indexes but I don't know if I like relying on that. Here's a scenario:
1. There are 3 shapes, A, B, C, & D. They have indexes 1, 2, 3, & 4 respectively. A+B+D overlap, but C does not.
2. The user tries to bring A forward. The children array is `[A, B, D]`.
3. We find B as the first selecting shape, then D as its next item in the array
4. `getIndexBetween(2, 4) => 3`
5. A now has the same index as C

Instead of filtering out all the children at the start, should we apply the filtering in the skipping/selecting loop? That way, when we look up the next child, we'll automatically get the correct one and avoid any chance of repeated IDs.

ds300: good suggestion I'll give that a go

packages/editor/src/lib/utils/reorderShapes.ts
```diff
@@ -147,14 +149,49 @@ function reorderToFront(moving: Set<TLShape>, children: TLShape[], changes: TLSh
...
+ * @param editor The editor
  * @param moving The set of shapes that are moving
  * @param children The parent's children
  * @param changes The changes array to push changes to
  */
-function reorderForward(moving: Set<TLShape>, children: TLShape[], changes: TLShapePartial[]) {
+function reorderForward(
+	editor: Editor,
+	moving: Set<TLShape>,
+	children: TLShape[],
```
steveruizok, line 190: I might move the filtering out of the function here and into whatever passes in the `children` (and maybe rename that argument to reflect this). There are times when reordering forward in a dumb way would be good (ie from a sidebar) so if we can keep this dumb then that would be best.

ds300: check alex's comment above, it's not totally safe to do that because then we'd be relying on the jitter algorithm to avoid clashing indexes. But we could definitely make smartness opt-in.

SomeHats (approved):
thanks david!

ds300:
ok I made it opt-out, merge o clock

PR #4852, opened 2024-11-04
botcom e2e tests

MitjaBezensek:
* Makes e2e test more stable.
* Brings the e2e (dotcom) workflow.
* Allows parallelization of tests by using multiple users and isolating the tests in the db. It seems to work fine till around 4 workers, then you might run into some postgres deadlocks.
* Improves [clerk usage in tests](https://clerk.com/docs/testing/playwright/overview).
* Adds missing tests.
* Pulls in @Taha-Hassan-Git tests from https://github.com/tldraw/tldraw/pull/4827
* Adds `@step` helper decorator that can be use in fixtures to make the test output a bit less noisy.
* Adds `repeatTest` helper which allows you to run a single test multiple times. It's a 1:1 replacement of `test` (so you can also use it with `.only`) to make it easier to switch between the two.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Bring back dotcom e2e tests, make them more stable, and add a few more of them.


apps/dotcom/client/e2e/tests/not-logged-in.spec.ts
```diff
@@ -10,30 +10,10 @@ test('can login', async ({ homePage, editor }) => {
 	await expect(editor.sidebarToggle).toBeVisible()
 })
 
-test.fixme('can visit a shared file', async () => {
```
MitjaBezensek, line 13: Added these in the sharing spec.

.github/workflows/playwright-dotcom.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          echo "CLERK_PUBLISHABLE_KEY=${{ secrets.VITE_CLERK_PUBLISHABLE_KEY }}" > apps/dotcom/sync-worker/.dev.vars
+          echo "CLERK_SECRET_KEY=${{ secrets.CLERK_SECRET_KEY }}" >> apps/dotcom/sync-worker/.dev.vars
+          yarn e2e-dotcom
+        env:
+          VITE_CLERK_PUBLISHABLE_KEY: ${{ secrets.VITE_CLERK_PUBLISHABLE_KEY }}
+          CLERK_SECRET_KEY: ${{ secrets.CLERK_SECRET_KEY }}
+        if: ${{!contains(github.event.pull_request.labels.*.name, 'e2e-x10')}}
+
+      - name: Run Playwright tests (x10)
```
MitjaBezensek, line 61: You can use `e2e-x10` label to trigger running tests 10 times. This should allow us to thoroughly test potentially flaky tests.

apps/dotcom/client/e2e/fixtures/tla-test.ts
```diff
@@ -1,34 +1,139 @@
...
+
+export function step(name: string) {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+	return function (target: Function, context: any) {
+		if (context.kind === 'method') {
+			return async function (...args: any[]) {
+				return await test.step(name, async () => {
+					// @ts-expect-error Parameter 'this' implicitly has an 'any' type.ts(7006)
+					return target.apply(this, args)
+				})
```
ds300, line 113: Love this decorator. I think you can infer the step names and just put `@step` without arguments everywhere.
```suggestion
				return await test.step(`${this.constructor.name}.${context.name}`, async () => {
					// @ts-expect-error Parameter 'this' implicitly has an 'any' type.ts(7006)
					return target.apply(this, args)
				})
```

ds300 (approved):
awesome, can't wait to try this with the refactored backend

.github/workflows/playwright-dotcom.yml
```diff
@@ -0,0 +1,97 @@
...
+        name: Upload S3
+        id: s3
+        with:
+          aws_key_id: ${{ secrets.AWS_S3_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.AWS_S3_SECRET_KEY}}
+          aws_bucket: playwright-reports.tldraw.xyz
+          source_dir: apps/dotcom/client/playwright-report
+
+      - name: Log report to summary
+        if: always()
```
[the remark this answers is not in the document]
MitjaBezensek: Think this makes sure that this step is always executed, even if any of the previous ones fails.

Issue #4860, opened 2024-11-05
Shapes get stuck in erasing mode


TodePond:
1. Make some shapes.
2. Pointer down on canvas.
3. Brush over the shapes to make them go translucent.
4. Press 'D' to change tool.
5. Pointer up.

Expected: The shapes should go opaque when you press 'D'.
Actual: The shapes get stuck as translucent.

PR #4861, opened 2024-11-05
Fix shapes getting stuck in erasing state

TodePond:
This PR fixes shapes getting stuck in the 'erasing' state.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

See: https://github.com/tldraw/tldraw/issues/4860

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with shapes getting stuck in the translucent erasing state.


TodePond:
@SomeHats is this really the thing to do here? we don't trigger `onCancel` when changing tools?

SomeHats (approved):
❤️ 
yeah i don't think we have a better way of doing this right now tbh

PR #4863, opened 2024-11-05
only cache assets from the current deploy

SomeHats:
We keep getting these weird 'Failed to fetch dynamically imported module' errors after deploys. From some debugging this time around, it looks like this is maybe related to cloudflare caching 404 responses, then continuing to server 404s long after they're no longer valid.

This was happening because we were setting an immutable cache-control header on all requests to /assets/ regardless of whether or not they were successful. This diff removes that blanket header, and instead only applies it to assets that we know definitely exist.

### Change type

- [x] `other`


ds300 (approved):
This seems like a sensible change. But I don't understand how vercel was serving up 404s for those files in the first place 🤔 happy to merge tho

apps/dotcom/client/package.json
```diff
@@ -67,6 +68,7 @@
 		"identity-obj-proxy": "^3.0.0",
 		"json5": "^2.2.3",
 		"lazyrepo": "0.0.0-alpha.27",
+		"regexgen": "^1.3.0",
```
mimecuvalo, line 71: this library seems unsupported (last update 7 yrs ago).
have you used it before / any alternatives?

apps/dotcom/client/scripts/build.ts
```diff
@@ -95,14 +98,21 @@ async function build() {
 						dest: `${multiplayerServerUrl}$1`,
 						check: true,
 					},
-					// cache static assets immutably
 					{
 						src: '^/assets/(.*)$',
+						continue: true,
```
mimecuvalo, line 103: it'd be good to note here what the `continue: true` here does.
we've scratched our heads on this config.json a couple times before so the more we can light the path here the better :)

apps/dotcom/client/package.json
SomeHats: no alternatives, but I have used it before - never really had issues with it, but i probably wouldn't put it somewhere it was accepting untrusted strings i guess?

PR #4864, opened 2024-11-06
Fix zoom to label

Taha-Hassan-Git:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/packages/tldraw/src/lib/shapes/frame/components/FrameHeading.tsx
+++ b/packages/tldraw/src/lib/shapes/frame/components/FrameHeading.tsx
@@ -67,13 +67,7 @@ export const FrameHeading = function FrameHeading({
 		}
 	}, [rInput, isEditing])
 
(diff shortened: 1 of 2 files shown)


Taha-Hassan-Git:
Should I keep going with this @steveruizok ?

steveruizok:
So my thinking here is that we should turn off this feature (zooming to an editing shape if the shape is off-screen).

If we do want to keep it, then we would really need a way to get the screenrect of the actual input and bring that onto the screen.

IMO this is currently a bug more than a feature, so let's just remove it now and then try to get it perfect someday.

Taha-Hassan-Git:
Sure, will do.

Issue #4845, opened 2024-11-02
[Bug]: Crash when rotating image while holding control


jamdotjar:
### What happened?

I tried to rotate an image while pressing control, and tldraw crashes with the error `TypeError: t is undefined`. When attempting to rotate other objects while holding control, tldraw doesn't allow it.
It also only crashes when control is held before rotation, if control starts to be pressed during a rotation nothing happens.
This behavior seems to have been introduced in 3.3.0.

### How can we reproduce the bug?

- Open a tldraw board on a site using SDK version 3.3.0 or later.
- Add an image
- Hold control
- Rotate the image while holding control.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Where did this happen?

The developer package

### Contact Details

darkestknight3d@outlook.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #4866, opened 2024-11-06
crop: fix issue with handle confusion when going from rotation->cropping

mimecuvalo:
fixes https://github.com/tldraw/tldraw/issues/4845
we shouldn't keep rotate handles around as it's confusing UI - it's not really what's going to happen when the ctrlKey is being pressed.

https://github.com/user-attachments/assets/74f6063d-9ef0-499b-8222-856b51d86f7a

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Update the handles when we're going to enter into a cropping mode.


packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
```diff
@@ -35,6 +38,22 @@ export const TldrawSelectionForeground = track(function TldrawSelectionForegroun
...
+			}
+		}
+		editor.addListener('event', handleEvent)
+		return () => {
+			editor.removeListener('event', handleEvent)
+		}
+	}, [editor, isCtrlKeyPressed])
+
+	const canEnterCropMode = isCtrlKeyPressed && getShouldEnterCropMode(editor)
+
```
steveruizok, line 56: We should really make editor.inputs into an atom, huh

steveruizok: Just took a shot at it, that would be a big old change, let's land this and then think about that.

steveruizok (dismissed):

mimecuvalo:
> Nevermind, I don't like this. The crop UI shows up too often, e.g. when pressing copy and paste. We show the handles when control+clicking the edges and after double clicking the shape, I think that's enough.

We still need to address the bug that this is causing though: https://github.com/tldraw/tldraw/issues/4845
(that was the original impetus for this PR)

steveruizok:
Is that fix in this PR? or should we treat it as a fresh problem?

ps for vibe / attitude etc, this PR works and is a good fix. When user intents overlap like this (here, is the cmd key down because the user is pressing a modifier for a shortcut, or are they trying to crop an image?) I try not to change anything in the UI as it'll be "noticeably wrong" in cases where the user has intent A instead of intent B, even if it hurts the discoverability of intent B.

mimecuvalo:
> Is that fix in this PR? or should we treat it as a fresh problem?

Yeah, the fix is in this PR — it's that you don't go accidentally into the rotate mode which causes a crash when holding `Ctrl`

> ps for vibe / attitude etc, this PR works and is a good fix. I have some poorly formed design principles around "collisions" baked into tldraw. When user intents overlap like this (here, is the cmd key down because the user is pressing a modifier for a shortcut, or are they trying to crop an image?) I try not to change anything in the UI as it'll be "noticeably wrong" in cases where the user has intent A instead of intent B, even if it hurts the discoverability of intent B.

Cool, no worries, I hear that. I think my take on it is similar, say, to when you're dragging an arrow and toggling `Cmd` (toggling binding) - the UI does change from underneath you to show intent. 

Let's discuss more offline and come up with a solution!

steveruizok:
PR here to fix https://github.com/tldraw/tldraw/pull/5087

PR #4874, opened 2024-11-07
dotcom: use Inter font via a <link> for better Cloudflare optimization

mimecuvalo:
talked with @SomeHats about this, following this PR: https://github.com/tldraw/tldraw/pull/4870
we're gonna turn on the Cloudflare optimization after this: https://developers.cloudflare.com/speed/optimization/content/fonts/
(Cloudflare wasn't doing the optimization correctly because we were using `@import` and it was stripping out the `preconnect` from the HTML on-demand)

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/dotcom/client/index.html
+++ b/apps/dotcom/client/index.html
@@ -73,8 +73,9 @@
 		<meta name="msapplication-tap-highlight" content="no" />
 
 		<!-- $PRELOADED_FONTS -->
-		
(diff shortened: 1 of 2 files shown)


SomeHats (approved):
looking good!

PR #4875, opened 2024-11-07
Snap to grid when creating shapes

Taha-Hassan-Git:
TLD-2817

TLD-2816

This PR makes sure that shapes snap to the grid when created. It adds a ```maybeSnapToGrid``` function, which can be used to push a shape onto the grid if grid mode is enabled, both when click-creating and when drag-creating.

1. Any shapes using the basebox shape tool (i.e frames)
2. Geo shapes
3. Both arrow handles
4. Line shapes, including shift-clicking
5. Note shapes (when translating, note shapes prefer adjacent note positions over grid)
6. Text shapes
7. Aligns uploaded assets using the top left of the selection bounds.
8. Does not snap to the grid when snap indicators are being shown

It also adds tests for this behaviour

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Enable grid
9. Click-create a note shape off the grid
10. It should snap to the grid
11. Add an asset, it should align with the grid

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Shapes snap to grid on creation, or when adding points.

### DIFF
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -143,6 +143,6 @@ export class Pointing extends StateNode {
 export function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
 	const isGridMode = editor.getInstanceState().isGridMode
 	const gridSize = editor.getDocumentSettings().gridSize
-	if (isGridMode) return point.snapToGrid(gridSize)
-	return point
+	if (isGridMode) return point.clone().snapToGrid(gridSize)
+	return point.clone()
 }
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -1,6 +1,7 @@
 import { TLShape, createShapeId } from '@tldraw/tlschema'
 import { structuredClone } from '@tldraw/utils'
 import { Vec } from '../../../../primitives/Vec'
+import { Editor } from '../../../Editor'
 import { TLBaseBoxShape } from '../../../shapes/BaseBoxShapeUtil'
 import { TLPointerEventInfo } from '../../../types/event-types'
 import { StateNode } from '../../StateNode'
@@ -108,8 +109,9 @@ export class Pointing extends StateNode {
 		}
 
 		const next = structuredClone(shape)
-		next.x = shape.x - delta.x
-		next.y = shape.y - delta.y
+		const newPoint = maybeSnapToGrid(new Vec(shape.x - delta.x, shape.y - delta.y), this.editor)
+		next.x = newPoint.x
+		next.y = newPoint.y
 		next.props.w = w
 		next.props.h = h
 
@@ -131,4 +133,11 @@ export class Pointing extends StateNode {
 	cancel() {
 		this.parent.transition('idle')
 	}
+}
+
+function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
+	const isGridMode = editor.getInstanceState().isGridMode
+	const gridSize = editor.getDocumentSettings().gridSize
+	if (isGridMode) return point.snapToGrid(gridSize)
+	return point
 }
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -135,7 +135,12 @@ export class Pointing extends StateNode {
 	}
 }
 
-function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
+/**
+ * Checks if grid mode is enabled and snaps a point to the grid if so
+ *
+ * @public
+ */
+export function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
 	const isGridMode = editor.getInstanceState().isGridMode
 	const gridSize = editor.getDocumentSettings().gridSize
 	if (isGridMode) return point.snapToGrid(gridSize)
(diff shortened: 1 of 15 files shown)


packages/tldraw/src/lib/utils/shapes/shapes.ts
```diff
@@ -18,3 +18,10 @@ export function getTextLabels(geometry: Geometry2d) {
 
 	return []
 }
+
+export function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
```
mimecuvalo, line 22: very handy, i'd say this should be a `@public` fn

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -168,12 +168,22 @@ export class Pointing extends StateNode {
...
-
-		this.editor.updateShape({
-			...shape,
-			x: shape.x + delta.x,
-			y: shape.y + delta.y,
-		})
+		const isGridMode = this.editor.getInstanceState().isGridMode
+		if (isGridMode) {
+			const gridSize = this.editor.getDocumentSettings().gridSize
+			const newPoint = new Vec(shape.x, shape.y).snapToGrid(gridSize)
```
mimecuvalo, line 174: this can use your `maybeSnapToGrid` fn, right?

packages/tldraw/src/lib/shapes/line/toolStates/Pointing.ts
```diff
@@ -45,8 +46,8 @@ export class Pointing extends StateNode {
 				this.editor.getShapeParentTransform(this.shape)!,
 				new Vec(this.shape.x, this.shape.y)
 			)
-
-			const nextPoint = Vec.Sub(currentPagePoint, shapePagePoint).addXY(0.1, 0.1)
+			const nudgedPoint = Vec.Sub(currentPagePoint, shapePagePoint).addXY(0.1, 0.1)
```
mimecuvalo, line 49: i know this isn't your code but a comment here about why it's being nudged would be helpful — do you know why?

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
```diff
@@ -168,12 +168,22 @@ export class Pointing extends StateNode {
 			const transform = this.editor.getShapeParentTransform(shape)
 			delta.rot(-transform.rotation())
 		}
-
-		this.editor.updateShape({
-			...shape,
-			x: shape.x + delta.x,
-			y: shape.y + delta.y,
-		})
+		const isGridMode = this.editor.getInstanceState().isGridMode
```
mimecuvalo, line 171: btw you could consider moving all this logic above line 144 - because most of this logic is about calculating `delta`

packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
```diff
@@ -132,3 +134,10 @@ export class Pointing extends StateNode {
 		this.parent.transition('idle')
 	}
 }
+
+function maybeSnapToGrid(point: Vec, editor: Editor): Vec {
```
mimecuvalo, line 138: duplicated from `shapes.ts`? let's have this function here be the canonical function, then you can export it as `@public` and use it elsewhere in the library

packages/tldraw/src/lib/utils/shapes/shapes.ts
mimecuvalo: ah nvm, i just saw this is also in `packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts` - let's just have one version of this

mimecuvalo (commented):
nice one 👏 , couple comments before landing

packages/tldraw/src/lib/utils/shapes/shapes.ts
Taha-Hassan-Git: Thanks, couldn't figure out a way to use it across packages!

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
Taha-Hassan-Git: I didn't think so, because it would strip the delta. But now I reckon I can actually.

packages/tldraw/src/lib/shapes/line/toolStates/Pointing.ts
Taha-Hassan-Git: It's to avoid zero-length lines, I can add a comment for sure.

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
Taha-Hassan-Git: Will do!

packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
Taha-Hassan-Git: Yes! Thanks!

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
Taha-Hassan-Git: I could either:

1. Check if we're in grid mode, then call maybeSnapToGrid, which would check gridMode again.
2. Change maybeSnapToGrid so that it handles text in a special way, or returns a bool to indicate whether it snapped or not
3. Leave as is, and duplicate the logic

Personally I prefer keeping the function simple, since it works fine everywhere else, and simply duplicating the logic, rather than adding the extra complexity. What do you think?

packages/tldraw/src/lib/shapes/text/toolStates/Pointing.ts
Taha-Hassan-Git: I actually want to calculate the delta as normal and then shift the top left of the shape to the nearest grid point, this keeps the text tool working in a similar way in grid mode and outside of it.

jpike88 (approved):

jpike88:
I'm desperate for a release on this one! Also I tested it locally it works fine

PR #4885, opened 2024-11-11
FE Fixes

MitjaBezensek:
* Fix an issue with not correctly creating a file when the user signs up. Fixes INT-462
* Fix an issue when signing out. Fixes INT-463
* Fix some asset loading issues. Fixes INT-460
* Add batching of messages. Fixes INT-448

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Creating new files for new users

1. Delete your user.
2. Sign in.
3. The app should create a file for you and you should be able to see it.

### Errors when signing out

1. Sign in.
3. Sign out.
4. You should not see any errors. Also you should not see errors in the console.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix an issue with not correctly creating a file when the user signs up.
- Fix an issue when signing out.
- Fix some asset loading issues. 
- Adds batching of messages. Now sending them every 50ms.

### DIFF
--- a/apps/dotcom/client/src/tla/app/zero-polyfill.ts
+++ b/apps/dotcom/client/src/tla/app/zero-polyfill.ts
@@ -15,6 +15,9 @@ import { Signal, computed, react, transact, uniqueId } from 'tldraw'
 export class Zero {
 	socket: ClientWebSocketAdapter
 	store = new OptimisticAppStore()
+	pendingUpdates: ZRowUpdate[] = []
+	timeout: NodeJS.Timeout | null = null
+	currentMutationId = uniqueId()
 
 	constructor(
 		private opts: { getUri(): Promise<string>; onMutationRejected(errorCode: ZErrorCode): void }
@@ -47,6 +50,12 @@ export class Zero {
 	}
 
 	dispose() {
+		if (this.pendingUpdates.length) {
+			this.sendPendingUpdates()
+		}
+		if (this.timeout) {
+			clearTimeout(this.timeout)
+		}
 		this.socket.close()
 	}
 
@@ -198,17 +207,29 @@ export class Zero {
 		})
 	}, this.____mutators)
 
+	private sendPendingUpdates() {
+		this.socket.sendMessage({
+			type: 'mutate',
+			mutationId: this.currentMutationId,
+			updates: this.pendingUpdates,
+		} satisfies ZClientSentMessage as any)
+
+		this.pendingUpdates = []
+		this.currentMutationId = uniqueId()
+		this.timeout = null
+	}
+
 	makeOptimistic(updates: ZRowUpdate[]) {
 		const mutationId = uniqueId()
-		console.log('apply local optimistic updates', mutationId, updates)
 		this.store.updateOptimisticData(updates, mutationId)
 
 		if (this.socket.isDisposed) return
 
-		this.socket.sendMessage({
-			type: 'mutate',
-			mutationId,
-			updates,
-		} satisfies ZClientSentMessage as any)
+		this.pendingUpdates.push(...updates)
+		if (!this.timeout) {
+			this.timeout = setTimeout(() => {
+				this.sendPendingUpdates()
+			}, 50)
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


apps/dotcom/client/src/tla/pages/local.tsx
```diff
@@ -11,19 +11,22 @@ import { getFilePath } from '../utils/urls'
...
 	const navigate = useNavigate()
+	const creatingFile = useRef(false)
 
 	useEffect(() => {
 		if (!app) return
 		if (app.getUserRecentFiles().length === 0) {
 			app.createFile().then((res) => {
 				if (res.ok) {
+					creatingFile.current = true
 					navigate(getFilePath(res.value.file.id), { state: { mode: 'create' } })
```
MitjaBezensek, line 22: This navigate does not kick in immediately, it ran the logic for redirecting to the recent file, which then removed the `creating` mode from the navigation.

apps/dotcom/client/src/tla/app/zero-polyfill.ts
```diff
@@ -203,6 +203,8 @@ export class Zero {
 		console.log('apply local optimistic updates', mutationId, updates)
 		this.store.updateOptimisticData(updates, mutationId)
 
+		if (this.socket.isDisposed) return
```
MitjaBezensek, line 206: Looks like the error is because:
- We first [dispose](https://github.com/tldraw/tldraw/blob/f716f149ff5c82ac9a901a46ce43ede2d9071d9b/apps/dotcom/client/src/tla/hooks/useAppState.tsx#L55) the app.
- Only then we call [onFileExit](https://github.com/tldraw/tldraw/blob/f716f149ff5c82ac9a901a46ce43ede2d9071d9b/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx#L200) from the `TlaEditor`

apps/dotcom/client/src/tla/components/TlaIcon/TlaIcon.tsx
```diff
@@ -31,7 +35,7 @@ export function TlaIcon({
 			ref={ref}
 			className={classNames(styles.icon, className)}
 			style={{
-				mask: `url(/icon-${icon}.svg) center 100% / 100% no-repeat`,
+				mask: getMaskStyle(icon),
```
MitjaBezensek, line 38: We weren't using the `/tla` prefix here.

apps/dotcom/client/src/tla/app/zero-polyfill.ts
MitjaBezensek, line 1: Batch the messages.

apps/dotcom/client/src/tla/app/zero-polyfill.ts
```diff
@@ -198,15 +207,29 @@ export class Zero {
...
-			mutationId,
-			updates,
-		} satisfies ZClientSentMessage as any)
+		if (this.socket.isDisposed) return
+
+		this.pendingUpdates.push(...updates)
+		if (!this.timeout) {
+			this.timeout = setTimeout(() => {
+				this.sendPendingUpdates()
+			}, 50)
```
MitjaBezensek, line 232: Not sure what we want to user here, feels like this should be ok for now 🤷‍♂️

apps/dotcom/client/src/tla/pages/local.tsx
```diff
@@ -11,19 +11,22 @@ import { getFilePath } from '../utils/urls'
...
 	const app = useMaybeApp()
 	const navigate = useNavigate()
+	const creatingFile = useRef(false)
 
 	useEffect(() => {
 		if (!app) return
 		if (app.getUserRecentFiles().length === 0) {
 			app.createFile().then((res) => {
 				if (res.ok) {
+					creatingFile.current = true
```
ds300, line 21: let's move this out to just before we call createFile()?

ds300 (approved):
good stuff! ❤️

PR #4888, opened 2024-11-11
Robustify replicator bootup

ds300:
- read the initial data in a transaction, at the same time write a unique id for the boot sequence to the db
- wait for the boot id update to come through before handling messages on the subscription queue.
- stream the initial data into sqlite, rather than loading it all into memory (which caps out at 130mb)
- use the zero schema format to define data types, and to provide runtime info to generate the initial load sql query.
- added a postgres config that you can opt in to, to log out the queries that are being executed.

TODO: 

- [ ] load test the db up to 1gb of data, figure out how to measure the data size, get a sense of how many users we can theoretically support with this setup.

### Change type

- [x] `other`


apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -61,36 +65,83 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
-								users.forEach((user) => this.insertRow(user, 'user'))
-								files.forEach((file) => this.insertRow(file, 'file'))
-								fileStates.forEach((fileState) => this.insertRow(fileState, 'file_state'))
+								// ok
+								await this.db.begin(async (db) => {
+									return db
+										.unsafe(sql, [], { prepare: false })
+										.simple()
+										.forEach((row: any) => {
+											this.insertRow(
```
MitjaBezensek, line 122: Might make sense to bulk insert here. We can do some tests when we'll test perf of other parts.

MitjaBezensek (approved):
👏

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
ds300: yeah let's wait, sqlite writes are supposed to be basically instant on DOs

ds300:
merging this now, gonna follow up with more robustness stuff before scalability testing

PR #4894, opened 2024-11-12
Only send updates to active users

MitjaBezensek:
This adds user durable object registration so that we can track which users are active and only send updates to them.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Only send updates to active users.


ds300 (approved):
this looks good! heads up that it overlaps with the work I'm doing on handling replica reboots https://linear.app/tldraw/issue/INT-472/make-sure-user-dos-can-handle-replica-reboots so it might get altered quite a bit in my next PR

PR #4898, opened 2024-11-12
Add rate limiting.

MitjaBezensek:
Should we rate limit something else? Used session id to rate limit no auth users, I guess that should be ok.

Resolves INT-458

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -342,6 +343,12 @@ export class TLDrawDurableObject extends DurableObject {
 				if (!auth && !file.shared) {
 					return closeSocket(TLSyncErrorCloseEventReason.NOT_AUTHENTICATED)
 				}
+				if (auth?.userId) {
+					const rateLimited = await isRateLimited(this.env, auth?.userId)
+					if (rateLimited) {
+						return closeSocket(TLSyncErrorCloseEventReason.FORBIDDEN)
```
MitjaBezensek, line 349: Not sure if we should introduce a new event reason here?

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -342,6 +343,17 @@ export class TLDrawDurableObject extends DurableObject {
...
+					}
+				} else {
+					const rateLimited = await isRateLimited(this.env, sessionId)
```
MitjaBezensek, line 352: When would the sessionId default to instanceId. Could that cause some issues?

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: yeah we should. `TLSyncErrorCloseEventReason.COOL_YOUR_JETS_HOMIE`

ds300: (or just `TLSyncErrorCloseEventReason.RATE_LIMITED`)

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: don't think so

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -117,10 +127,15 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 	}
 
 	override async webSocketMessage(_ws: WebSocket, message: string) {
+		const isRateLimited = await this.isRateLimited()
 		const msg = JSON.parse(message) as any as ZClientSentMessage
 		switch (msg.type) {
 			case 'mutate':
-				await this.handleMutate(msg)
+				if (isRateLimited) {
+					this.rejectMutation(msg.mutationId, ZErrorCode.rate_limit_exceeded)
```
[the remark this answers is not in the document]
ds300: yeah i think it's fine to just reject the mutation. the limit should be set so that it only affects bad actors anyway.

apps/dotcom/sync-worker/wrangler.toml
```diff
@@ -273,3 +273,21 @@ binding = "DB"
...
+type = "ratelimit"
+namespace_id = 1001
+simple = { limit = 300, period = 60 }
+
+[[env.staging.unsafe.bindings]]
+name = "RATE_LIMITER"
+type = "ratelimit"
+namespace_id = 1001
+simple = { limit = 300, period = 60 }
+
```
ds300, line 288: i think we'll need one for the preview env too

ds300 (approved):

PR #4912, opened 2024-11-13
handle postgres failures in replicator

ds300:
- each time the replicator reboots it gets a new random `bootId`
- each time the replicator sends a row update message to a user DO it includes the boot id.
  - if the boot id doesn't match the previous one, the user DO resets its state and makes the sockets reconnect to ensure consistent data.
- each time the replicator's postgres subscription fails it tries to reboot iself
- active user DOs ping the replicator every 10 seconds to make sure it wakes up if it was put to sleep for whatever reason (I think this is unlikely to happen, but cloudflare is technically allowed to stop DOs for whatever reason and they won't start up again unless we perform a fetch or RPC on them. Better safe than sorry)
- no longer using blockConcurrencyWhile in replicator init. Instead we have a mechanism for the replicator RPCs (fetchDataForUser et al) to wait for a 'connected' state to be achieved.
  - gonna add a ticket to add timeouts for these so that we can handle situations of low/zero availability more gracefully for end users.

also

- don't throw away the active_user table on sqlite init. This could lead to zombie user DOs that are active but don't receive updates.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -101,6 +101,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 	private queue = new ExecutionQueue()
 
 	private async reboot(delay = true) {
+		// TODO: set up analytics and alerts for this
 		this.debug('rebooting')
 		await this.queue.push(async () => {
 			if (delay) {
@@ -251,7 +252,6 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 					// on error
 					() => {
 						this.debug('on error')
-						// TODO: ping team if this fails repeatedly
 						const e = new Error('replication start failed')
 						this.captureException(e)
 						if (this.state.type === 'connecting') {
@@ -270,7 +270,6 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 				})
 				.catch((err) => {
 					this.debug('caught error')
-					// TODO: ping team if this fails
 					this.captureException(new Error('replication start failed (catch)'))
 					this.captureException(err)
 					resolve('error')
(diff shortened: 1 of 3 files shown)


apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -50,15 +50,21 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
-				this.replicator.registerUser(this.userId!)
-			})
+			if (this.lastReplicatorBootId === null) {
+				await this.ctx.blockConcurrencyWhile(async () => {
+					const { data, bootId } = await this.replicator.fetchDataForUser(this.userId!)
+					this.lastReplicatorBootId = bootId
+					this.store.initialize(data)
+					this.replicator.registerUser(this.userId!)
+				})
+			}
```
ds300, line 67: now that we have the bootId thing, we can avoid fetching the initial data on every connect since we are guaranteed it will be consistent

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -38,193 +37,319 @@ CREATE TABLE file_state (
...
+						}
+						// ignore other events until we get the boot id
+					},
+					// on connect
+					async () => {
+						this.debug('on connect')
+						try {
+							assert(this.state.type === 'connecting', 'state should be connecting')
+							const sql = getFetchEverythingSql(myId, bootId)
+							this.ctx.storage.sql.exec(seed)
```
MitjaBezensek, line 222: We could use the store instance variable here `this.sql`

MitjaBezensek (approved):
Looks good. It's getting a bit hairy though, a bit hard to track the state changes 😄 But hopefully we can remove all of this code later on.

PR #4914, opened 2024-11-13
Use location hints to hopefully spawn replicator closer to the db

MitjaBezensek:
Apparently the [first call ](https://developers.cloudflare.com/durable-objects/reference/data-location/#provide-a-location-hint) to `get` provides a location hint to where the DO should be located. Pulled out a helper function for getting the replicator and made all current uses go through that.

Looks like we created supabase databases in eu central, so west europe for the DO seemed best. We could [use a jurisdiction](https://developers.cloudflare.com/durable-objects/reference/data-location/#restrict-durable-objects-to-a-jurisdiction) and make sure it's located in EU? (location hints are not a guarantee of the location).

We have many users in US, it might be better to move the supabase instances? OTOH having them in EU brings us closer to GDPR.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Pin 🙏  the postgres replicator close to the DB.

### DIFF
--- a/apps/dotcom/sync-worker/src/utils/tla/permissions.ts
+++ b/apps/dotcom/sync-worker/src/utils/tla/permissions.ts
@@ -2,19 +2,6 @@ import { IRequest } from 'itty-router'
 import { Environment } from '../../types'
 import { getAuth } from './getAuth'
 
-export type FileOwnerStatusError = 'unauthorized' | 'not-found' | 'forbidden'
-
-export function fileOwnerStatusErrorResponse(error: FileOwnerStatusError) {
-	switch (error) {
-		case 'forbidden':
-			return new Response(JSON.stringify({ error: true, message: 'Forbidden' }), { status: 403 })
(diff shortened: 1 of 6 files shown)


apps/dotcom/sync-worker/src/utils/tla/permissions.ts
```diff
@@ -2,19 +2,6 @@ import { IRequest } from 'itty-router'
 import { Environment } from '../../types'
 import { getAuth } from './getAuth'
 
-export type FileOwnerStatusError = 'unauthorized' | 'not-found' | 'forbidden'
```
MitjaBezensek, line 5: Flyby delete.

MitjaBezensek:
Closing this for now, will create a new PR once https://github.com/tldraw/tldraw/pull/4926 merges.

PR #4918, opened 2024-11-14
Fix file name when exporting a single unnamed frame

trygve-aaberge-adsk:
If you have a frame without a name, select it and export it, the name before the extension would end up as an empty string. So the final name would just be a period and the extension (e.g. `.svg`). This would cause the SVG file to be downloaded as `svg.txt` with Chrome and as just `svg` with Firefox (and equivalent for JSON files, but PNG files would be named `png.png`).

With this change they will be named `frame.<extension>` instead.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a frame, don't name it.
2. Select the frame.
3. Right click and select export as SVG or JSON.
4. Observe that the downloaded file doesn't have the correct extension before this fix.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix file name when exporting a single unnamed frame

### DIFF
--- a/packages/tldraw/src/lib/utils/export/exportAs.ts
+++ b/packages/tldraw/src/lib/utils/export/exportAs.ts
@@ -28,7 +28,7 @@ export async function exportAs(
 		if (ids.length === 1) {
 			const first = editor.getShape(ids[0])!
 			if (editor.isShapeOfType<TLFrameShape>(first, 'frame')) {
-				name = first.props.name ?? 'frame'
+				name = first.props.name || 'frame'
 			} else {
 				name = `${sanitizeId(first.id)} at ${getTimestamp()}`
 			}


trygve-aaberge-adsk:
> Hey, thanks for your pull request! Before we can merge your PR, you will need to sign our [Contributor License Agreement](https://tldraw.notion.site/Contributor-License-Agreement-4d529dd5e4b3438b90cdf2a2f9d7e7e6?pvs=4) by posting a comment that reads:
> 
> > I have read and agree to the Contributor License Agreement.

You have received a corporate CLA from my employer (Autodesk), which should cover this instead of your normal CLA. I guess posting that I've agreed to your normal CLA wouldn't be completely correct. How do you want to handle this?

steveruizok:
> > Hey, thanks for your pull request! Before we can merge your PR, you will need to sign our [Contributor License Agreement](https://tldraw.notion.site/Contributor-License-Agreement-4d529dd5e4b3438b90cdf2a2f9d7e7e6?pvs=4) by posting a comment that reads:
> > > I have read and agree to the Contributor License Agreement.
> 
> You have received a corporate CLA from my employer (Autodesk), which should cover this instead of your normal CLA. I guess posting that I've agreed to your normal CLA wouldn't be completely correct. How do you want to handle this?

I think we'll need a whitelist for users who have signed alternative CLAs. @somehats can you help?

SomeHats:
I've manually added you. Should be good to go now @trygve-aaberge-adsk

trygve-aaberge-adsk:
Anything holding up this PR?

PR #4920, opened 2024-11-14
botcom: add react-select for better select menus

mimecuvalo:
I noticed while looking into https://linear.app/tldraw/issue/INT-468/issues-with-react-intl-wrapping-translations-into-spans-when-using that we should just be using a richer Select menu option that allows for content other than just strings.

This brings the radix-select library into use for our selects instead of just the raw/native browser one.

Screenshot 2024-11-14 at 12 24 00

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
This is great, IIRC they hadn't launched this when we did the AI originally. How does it work on mobile? That's often one of the places where the native select menu makes the most difference.

MitjaBezensek:
Looks like it doesn't close when clicking in the share menu:

https://github.com/user-attachments/assets/8bf0f769-0ccc-4b6f-aa20-0828dfbb0f96

Few other suggestions:
- Show a different cursor so that it's easier to see that you can interact with it,
- Maybe disable it / change it's appearance if the file isn't shared. Otherwise you click on it expecting something will happen.

mimecuvalo:
>  How does it work on mobile? That's often one of the places where the native select menu makes the most difference.

Mobile seems fine 👍 

> Looks like it doesn't close when clicking in the share menu:

Yeah, I had noticed this - ugh, it seems to be some kind of radix annoyance. It's because a popover is already open and this is yet another one? Lemme see if I can wrangle it into working.

> Show a different cursor so that it's easier to see that you can interact with it.

Mm, I would disagree on this - that's not a standard select dropdown behavior.

> Maybe disable it / change it's appearance if the file isn't shared. Otherwise you click on it expecting something will happen. Or only show it when the file is shared.

Ah yeah, it seems to have lost that styling, thx for catching.

mimecuvalo:
Ok, fixed up the issues 👍

PR #4926, opened 2024-11-15
Lazy replicator

ds300:
- the pg replicator only stores data for the users who are currently active
- even then it only stores the users' ids, along with any guest file ids they have access to.
- each user DO fetches their replica directly from postgres, and syncs via messages from the pg replicator
- user DOs keep their replica in memory rather than sqlite, since they are so small.

still need to debug and stress test this, but it should be a lot more scalable than the current thing.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -6,6 +6,7 @@ import postgres from 'postgres'
 import type { EventHint } from 'toucan-js/node_modules/@sentry/types'
 import type { TLDrawDurableObject } from './TLDrawDurableObject'
 import type { TLUserDurableObject } from './TLUserDurableObject'
+import { getPostgres } from './getPostgres'
 import { Environment } from './types'
 
 const seed = `
@@ -105,16 +106,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 			// preserve the promise so any awaiters do eventually get resolved
 			// TODO: set a timeout on the promise?
 			promise,
-			db: postgres(this.env.BOTCOM_POSTGRES_CONNECTION_STRING, {
-				types: {
-					bigint: {
-						from: [20], // PostgreSQL OID for BIGINT
-						parse: (value: string) => Number(value), // Convert string to number
-						to: 20,
-						serialize: (value: number) => String(value), // Convert number to string
-					},
-				},
-			}),
+			db: getPostgres(this.env),
 			sequenceId: uniqueId(),
 		}
 		const subscription = await this.state.db.subscribe(
@@ -199,7 +191,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 						sequenceId: this.state.sequenceId,
 						userId: row.userId,
 					})
-				} else {
+				} else if (!file) {
 					this.captureException(new Error(`File not found: ${row.fileId}`))
 				}
 			} else if (
@@ -328,7 +320,9 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 	}
 
 	async registerUser(userId: string) {
+		this.debug('registerUser', userId)
 		await this.waitUntilConnected()
+		this.debug('registerUser', userId, 'connected')
 		assert(this.state.type === 'connected', 'state should be connected in registerUser')
 		this.sql.exec(`INSERT INTO active_user (id) VALUES (?) ON CONFLICT (id) DO NOTHING`, userId)
 		const guestFiles = await this.state
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -60,7 +60,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 		// eslint-disable-next-line @typescript-eslint/no-deprecated
 		this.sentry?.captureException(exception, eventHint) as any
 		if (!this.sentry) {
-			console.error(exception)
+			console.error(`[TLPostgresReplicator]: `, exception)
 		}
 	}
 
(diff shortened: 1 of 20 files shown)


apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -0,0 +1,303 @@
...
+				this.reboot()
+			}
+		})
+	}
+
+	private updateStateAfterBootStep() {
+		assert(this.state.type === 'connecting', 'state should be connecting')
+		if (this.state.didGetBootId && this.state.data) {
+			// we got everything, so we can set the state to connected and apply any buffered events
+			assert(this.state.type === 'connecting', 'state should be connecting')
```
MitjaBezensek, line 149: Similar assertion in line 146, can possibly drop this.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -0,0 +1,303 @@
...
+			const promise = this.state.promise
+			const bufferedEvents = this.state.bufferedEvents
+			const data = this.state.data
+			this.state = {
+				type: 'connected',
+				bootId: this.state.bootId,
+				sequenceId: this.state.sequenceId,
+			}
+			this.store.initialize(data)
```
MitjaBezensek, line 158: Just noticed that `store.initialize` does not clear optimistic part of the store, but it probably should so that those updates don't stay in the store between reboots?

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -46,12 +46,11 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 			this.env.TL_PG_REPLICATOR.idFromName('0')
```
MitjaBezensek, line 46: Let's also use the helper here. I have a PR that will help with pinning things closer to db, so it will also help with that.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -62,17 +61,25 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
 
+	private async init() {
+		assert(this.userId, 'User ID not set')
+		this.debug('init')
+		this.cache = new UserDataSyncer(this.ctx, this.env, this.userId, (message) =>
+			this.broadcast(message)
+		)
+		this.debug('cache', !!this.cache)
+		await this.cache.waitUntilConnected()
+		this.replicator.registerUser(this.userId!)
```
MitjaBezensek, line 80: Looks like both the `TLUserDurableObject` and the `UserDataSyncer` register the user. Not sure if that's intended?

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -190,11 +237,12 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
 					)
 				}
-				if (nextFile.ownerId === this.userId) return
+				if (prevFile.ownerId === this.userId) return
 				if (prevFile.shared && prevFile.sharedLinkType === 'edit') return
 				throw new ZMutationError(
 					ZErrorCode.forbidden,
-					'Cannot update file that is not our own and not shared in edit mode'
+					'Cannot update file that is not our own and not shared in edit mode' +
+						` user id ${this.userId} ownerId ${nextFile.ownerId}`
```
MitjaBezensek, line 245: Should we print the `prevFile.ownerId here`?

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -120,190 +116,70 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
+			this.handleEvent.bind(this),
+			() => {
+				// re-register all active users to get their latest guest info
+				const users = this.sql.exec('SELECT id FROM active_user').toArray()
+				for (const user of users) {
+					const sequenceId = this.state.sequenceId
+					this.queue.push(async () => {
+						if (this.state.sequenceId !== sequenceId) return
+						this.messageUser(user.id as string, { type: 'force_reboot', sequenceId })
+						await sleep(10)
```
MitjaBezensek, line 139: Why do we need this wait?

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -314,130 +190,97 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
-						this.state.bootId,
-						row,
-						event.relation.table as 'user' | 'file' | 'file_state',
-						event.command,
-						userId
-					)
+					this.messageUser(userId, {
+						type: 'row_update',
+						row: row as any,
+						table: event.relation.table as 'user' | 'file' | 'file_state',
```
MitjaBezensek, line 270: We could use `as ZTable` here.

MitjaBezensek (commented):
Looking good. Left a few comments and questions.

MitjaBezensek:
Maybe we can add the durable object name to the debug messages. Right now we just print `booting`, so it's hard to tell which object printed that out if you turn on debugging for all of the DOs.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -0,0 +1,306 @@
...
+		private env: Environment,
+		private userId: string,
+		private broadcast: (message: ZServerSentMessage) => void
+	) {
+		this.sentry = createSentry(ctx, env)
+		this.replicator = env.TL_PG_REPLICATOR.get(
+			env.TL_PG_REPLICATOR.idFromName('0')
+		) as any as TLPostgresReplicator
+		this.db = getPostgres(env)
+		this.reboot(false)
```
MitjaBezensek, line 113: Not 100% on the naming here. `reboot` feels like something that happens after the boot already happened?

MitjaBezensek (approved):

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
ds300: to ease the pressure on the db during reboots. We might have a thousand or more actively connected users, and when the replicator reboots each of those users will reboot, and each user reboot triggers two sql queries (one in user DO to fetch their initial data, and one in replicator DO to fetch guest file subscriptions). 

I'd be worried about failures if kicking off a couple of thousand PG queries at exactly the same time. Staggering them individually by 10ms is definitely overkill, but I'll err one the side of caution for now and stress test this with batches of varying size in staging.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
ds300: i think what should happen is it should fetch the mutation number during reboots and clear any optimistic updates that happened up to and including that mutation number. I'll add that in this PR.

PR #4935, opened 2024-11-18
Improve panning performance when we have many not shapes and when we are zoomed out

### LINKED ISSUE #4923: Is redrawing while panning necessary?
I would have sworn I saw a tldraw demo at one point showing lightning fast panning despite a huge number of objects. Using it today, I pan a large number of objects (1024 built-in note shapes) the frame rate is quite low. This is when zoomed out so they're all visible. It's super fast when I'm zoomed in and only a few are visible.

Did you run into a reason why something other than transforming the canvas position is necessary during panning?

screenshot 2024-11-14 at 19 12 44@2x

MitjaBezensek:
When zoomed out we switched from using a shadow to using a border with opacity. This caused significant perf drop of rendering the note shapes.

Fixes #4923

###  Before
image

### After 
CleanShot 2024-11-18 at 10 50 46

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create many note shapes
2. Zoom out past 35%
3. Pan around.
4. It should be much smoother than on main.

### Release notes

- Improve performance of rendering note shapes when zoomed out past 35%.


packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -209,6 +209,8 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
 			this.editor,
 		])
 
+		const darkMode = useValue('dark mode', () => this.editor.user.getIsDarkMode(), [this.editor])
```
MitjaBezensek, line 212: I don't think it makes sense to show bottom border when using dark mode (we even have a question in this file if using the drop shadow makes sense for dark mode).

steveruizok:
Looks good! I updated the shadow color for dark mode / light mode.

mimecuvalo (approved):
surprising! nice discovery. pre-approving but i assume @steveruizok might wanna weigh in on colors

PR #4956, opened 2024-11-20
[botcom] fix sharing defaults

ds300:
The 'shared by default' and 'edit mode by default' settings got clobbered during the backend refactor.

### Change type

- [x] `other`


ds300:
these might mess with some e2e tests? need to check

ds300:
e2e tests didn't rely on the values.

MitjaBezensek (approved):

PR #4962, opened 2024-11-20
Improve the names for files that have no name set

MitjaBezensek:
Show different time formats for files from different periods.

CleanShot 2024-11-20 at 17 49 26

Feels like we have too many sections in the sidebar, should we pair them down?

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed a bug with…


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -104,6 +106,11 @@ export class TldrawApp {
 		)
 	}
 
+	private getIntl() {
+		const locale = this.user$.get()?.locale
+		return createIntl({ locale: locale ?? 'en' })
```
MitjaBezensek, line 111: @mimecuvalo looks like this is the way to use react intl outside of react?

apps/dotcom/client/src/tla/utils/dates.ts
```diff
@@ -0,0 +1,47 @@
...
+	const now = new Date()
+
+	return {
+		today: new Date(now.getFullYear(), now.getMonth(), now.getDate()).getTime(),
+		yesterday: new Date(now.getFullYear(), now.getMonth(), now.getDate() - 1).getTime(),
+		thisWeek: new Date(now.getFullYear(), now.getMonth(), now.getDate() - now.getDay()).getTime(),
+		thisMonth: new Date(now.getFullYear(), now.getMonth(), 1).getTime(),
+		thisYear: new Date(now.getFullYear(), 0, 1).getTime(),
+	}
+}
```
ds300, line 13: nice improvement here

apps/dotcom/client/src/tla/app/TldrawApp.ts
mimecuvalo: yes, correct. however, slightly annoying thing is you should add a bit of caching by calling `setupCreateIntl` (see in `i18n.tsx`). and you'll need to update that cache when the locale changes.

as long as you're looking at this can you update the `setupCreateIntl` comment to the correct link? i see i put an old one in there, should be https://formatjs.github.io/docs/react-intl/api/#createintl

ds300 (approved):
Nice. This is way better. I think we might still want to tweak the format settings but this is a good foundation, thanks!

mimecuvalo (commented):
> Feels like we have too many sections in the sidebar, should we pair them down?

I agree with you, mostly. I think this sidebar, honestly, will change to not just be by date anyways in the nearish future - we'll probably have a A-Z, by date, shared/private, that kind of thing — and I think then that will force us to pair down anyway.

mimecuvalo (approved):
thx!

PR #4965, opened 2024-11-21
[botcom] simplify replicator dispatch logic

ds300:
This stuff was a little bit of a mess. I made it synchronous again and hopefully caught an extra edge case or two.

- added a `isFileOwner` field to file_state, populated by triggers on both the file table and the file_state table.
- used `isFileOwner` to avoid checking ownership while dispatching events in the replicator
- simplified the event dispatching logic by moving it out into separate methods for each table, so it's easier to follow what's happening.
- fixed the 'file deleting isn't working in prod' bug, that was to do with r2 complaining about things it doesn't complain about in dev mode.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
@@ -163,7 +163,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 		if (this.sentry) {
 			// eslint-disable-next-line @typescript-eslint/no-deprecated
 			this.sentry.addBreadcrumb({
-				message: `[TLUserDurableObject]: ${args.join(' ')}`,
+				message: `[TLUserDurableObject]: ${args.map((a) => (typeof a === 'object' ? JSON.stringify(a) : a)).join(' ')}`,
 			})
 		}
 	}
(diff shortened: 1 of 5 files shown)


apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -163,7 +163,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 		if (this.sentry) {
 			// eslint-disable-next-line @typescript-eslint/no-deprecated
 			this.sentry.addBreadcrumb({
-				message: `[TLUserDurableObject]: ${args.join(' ')}`,
+				message: `[TLUserDurableObject]: ${args.map((a) => (typeof a === 'object' ? JSON.stringify(a) : a)).join(' ')}`,
```
ds300, line 166: objects were being sent to sentry as `[object Object]`

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -389,11 +389,15 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
+		if (publishedHistory.length > 0) {
+			await this.env.ROOM_SNAPSHOTS.delete(publishedHistory)
+		}
 		// remove edit history
 		const r2Key = getR2KeyForRoom({ slug: id, isApp: true })
 		const editHistory = await listAllObjectKeys(this.env.ROOMS_HISTORY_EPHEMERAL, r2Key)
-		await this.env.ROOMS_HISTORY_EPHEMERAL.delete(editHistory)
+		if (editHistory.length > 0) {
+			await this.env.ROOMS_HISTORY_EPHEMERAL.delete(editHistory)
+		}
```
ds300, line 400: this fixes the 'file not deleting' bug. in remote envs R2 complains if you call .delete with an empty array. i saw the errors popping up in sentry.

what's odd is that the mutations didn't seem to be rejected. I think that may not be working remotely too. i.e. if an error is thrown during handleMutation, inside the transaction maybe, it seems to crap out.

ds300: added ticket here https://linear.app/tldraw/issue/INT-520/investigate-mutation-rejection

MitjaBezensek (approved):
Makes things a lot more readable 👏 

The logic is getting pretty hairy though. Like, I wouldn't want to add support for groups on top of this. Makes me think we should maybe push more stuff to PG. Not sure if that would work for zero though 🤔

PR #4971, opened 2024-11-22
[WIP] Remove `TLAppDurableObject` now that prod has been deployed.

MitjaBezensek:
We removed the bindings a while back, but could not delete the durable object itself. We have since deployed production, which removed those bindings, so we should now be free to also remove the DO (see the linked issue for more info).
### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Remove the `TLAppDurableObject` since it is no longer used.


MitjaBezensek:
Moving back to WIP, looks like it still won't let us delete it.
>  Cannot apply delete-class migration to class 'TLAppDurableObject' which was not exported in the previous version of the script [code: 10074]

PR #4974, opened 2024-11-22
Scroll the sidebar to the top when creating or deleting files

MitjaBezensek:
If we create a new file it shows up at the top. When we delete a file we also select the top most one.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Make sure the active file is visible in the sidebar after creating a file or deleting it.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/TlaSidebar.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/TlaSidebar.tsx
@@ -77,7 +77,9 @@ export const TlaSidebar = memo(function TlaSidebar() {
 					<TlaSidebarWorkspaceLink />
 					<TlaSidebarCreateFileButton />
 				</div>
-				<TlaSidebarRecentFiles />
+				<div className={styles.content}>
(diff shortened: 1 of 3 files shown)


apps/dotcom/client/src/tla/components/TlaSidebar/TlaSidebar.tsx
```diff
@@ -171,6 +169,11 @@ function TlaSidebarRecentFiles() {
 		},
 		[app]
 	)
+	const sidebarContentRef = useRef<HTMLDivElement>(null)
+
+	useEffect(() => {
+		sidebarContentRef.current?.scrollTo({ top: 0 })
```
mimecuvalo, line 175: worth considering adding `behavior: smooth` if you like

MitjaBezensek: Added that (had to rework the logic a bit so that remote changes would not trigger the scroll).

PR #4977, opened 2024-11-22
Wait for the guest file to be loaded before showing an entry in the sidebar

MitjaBezensek:
### Before

https://github.com/user-attachments/assets/24e5bdaf-483f-4f39-a73f-fbd709293ef0

### After

https://github.com/user-attachments/assets/46aa6e15-06c6-4f83-8c92-200af64da5c7

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- We only show the guest file in the sidebar when the file record is loaded. Otherwise we get this file name flickering since we don't have the file name when we just visit the shared link.

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -184,6 +184,7 @@ export class TldrawApp {
 				continue
 			}
 			const file = myFiles[fileId]
+			if (!file) continue
 			const state = myStates[fileId]
 
 			nextRecentFileOrdering.push({


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -178,17 +184,26 @@ export class TldrawApp {
...
+					isLoaded: true,
+				})
+				continue
+			}
 			if (existing) {
 				nextRecentFileOrdering.push(existing)
 				continue
 			}
-			const file = myFiles[fileId]
-			const state = myStates[fileId]
```
ds300, line 202: I think the behavior of this PR can be achieved with a one-liner

```suggestion
if (!file) continue
```

ds300 (approved):
This works but I think it can be done with a much smaller diff (see other comment)

apps/dotcom/client/src/tla/app/TldrawApp.ts
MitjaBezensek: D'oh! 🤦

PR #4979, opened 2024-11-22
Click / right click on frame headings

steveruizok:
This PR improves right clicks on frames.

- adds geometry for frame labels, allowing them to be clicked and dragged using the normal hit detection system
- a few cleanups around the frame headings generally
- restores right clicking on frames

Kapture 2024-11-22 at 15 04 52

### Change type

- [x] `improvement`

### Test plan

1. Create a frame shape
2. Right click on its heading to select it and open the context menu
3. Click on its heading to select the shape
4. Click on its heading and drag to move it
5. Double click to begin editing the heading
6. Right click on frame body to select it (previously it would hit the canvas)

- [x] Unit tests

### Release notes

- Improved clicks for frame headings


trygve-aaberge-adsk:
@steveruizok: This made performance significantly worse on a canvas we have with a lot of frames. After this change it has very low fps when zooming in or out. This happens because it calls `measureTextSpans` in `getFrameHeadingInfo` which is expensive.

steveruizok:
> @steveruizok: This made performance significantly worse on a canvas we have with a lot of frames. After this change it has very low fps when zooming in or out. This happens because it calls `measureTextSpans` in `getFrameHeadingInfo` which is expensive.

Working on this!

PR #4983, opened 2024-11-24
Add eslint rule for react-intl.

steveruizok:
We want all of our files to use our wrapped version of react-intl. This _should_ fix auto imports resolving to react-int by mistake.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
Annoyingly I can't seem to get typescript to force auto-import to use the wrapped version instead.

mimecuvalo (commented):
nice one

PR #4985, opened 2024-11-24
Revert "Add eslint rule for react-intl."

steveruizok:
Reverts tldraw/tldraw#4983


mimecuvalo:
oh, what went wrong here?

PR #4986, opened 2024-11-25
Make sure our tests can test if data was propagated to the server

MitjaBezensek:
Adds a helper function `expectBeforeAndAfterReload` which allows you to easily test whether some assertion holds currently and also after a page reload.

Since playwright passes tests immediately after some condition becomes true (a new file is visible in the sidebar after creating it) it might be too fast to pass a test that might actually fail due to our server rejecting a change.

The new helper runs the assertion then reloads the page then tests the assertion again. This means that it also tests whether the mutation succeeded and the data got stored to the db / user durable object.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improve our testing story by also adding a helper for testing whether certain expectation is met before and after a page reload.


apps/dotcom/client/e2e/fixtures/helpers.ts
```diff
@@ -61,3 +57,22 @@ export function areUrlsEqual(
...
+export function createFixtures(page: Page) {
+	const newSidebar = new Sidebar(page)
+	const newEditor = new Editor(page, newSidebar)
+	const newHomePage = new HomePage(page, newEditor)
+	const newShareMenu = new ShareMenu(page)
+	const errorPage = new ErrorPage(page)
+	return { newSidebar, newEditor, newHomePage, newShareMenu, errorPage }
+}
+
+export function getRandomName() {
```
MitjaBezensek, line 70: Tried to use `uniqueId` but got some issue with `crypto` not being available.

ds300 (approved):
excellent idea!

PR #4989, opened 2024-11-25
[botcom] scroll whenever active file changes

ds300:
Slight tweak to #4974, to enable the scrolling on any route change, not just when creating/deleting/duplicating. The main use case here would be navigating back/forward through browser history.

### Change type

- [x] `other`


MitjaBezensek (approved):
👏

PR #4992, opened 2024-11-25
Soft deleting of files

MitjaBezensek:
Instead of deleting the files we now soft delete them. I kept the whole delete logic the same as before, which should allow us to delete the file from postgres which will trigger all other necessary file deletes. 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Add soft deleting of files.


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -454,7 +455,7 @@ export class TldrawApp {
 	async getOrCreateFileState(fileId: string) {
 		let fileState = this.getFileState(fileId)
 		if (!fileState) {
-			await this.z.mutate.file_state.create({
```
MitjaBezensek, line 457: Flyby change.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -611,6 +614,8 @@ export class TLDrawDurableObject extends DurableObject {
 		const roomIsReadOnlyForGuests = file.shared && file.sharedLinkType === 'view'
 
 		for (const session of room.getSessions()) {
+			if (file.isDeleted)
+				room.closeSession(session.sessionId, TLSyncErrorCloseEventReason.NOT_FOUND)
```
ds300, line 618:
```suggestion
			if (file.isDeleted) {
        room.closeSession(session.sessionId, TLSyncErrorCloseEventReason.NOT_FOUND)
        continue
      }
```

I think we need to do this to prevent the logic below from running

ds300 (approved):
Looks good code-wise. I'll do some manual testing tomorrow.

PR #4993, opened 2024-11-25
Use partials when mutating

MitjaBezensek:
This moves away for sending the whole entity when doing updates and instead just sends the changes. Should prevent:
- overwriting data (which has already hit PG, but has not reached the current user)
- rejection of valid mutations (user might have some invalid optimistic updates and if we then do any other changes to those same entities then these later, possibly valid, mutations would get rejected due to that first change).
- there's still [one case where we update the whole entity](https://github.com/tldraw/tldraw/blob/519f424459a980e423d6aad8a9f1a1184e99d4d1/apps/dotcom/client/src/tla/app/TldrawApp.ts#L148-L151), but it looks like that should maybe stay as is?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Only send the changed columns when doing update mutations.


ds300 (approved):
Nice, thank you! I think the 'setUserPreferences' impl could be partial too. Not sure why you suggested it would need to remain as-is?

MitjaBezensek:
> Nice, thank you! I think the 'setUserPreferences' impl could be partial too. Not sure why you suggested it would need to remain as-is?

Ah, you are right, updated that as well.

PR #4994, opened 2024-11-25
[botcom] owner info in side bar

ds300:
Appearance
image
Tooltip
Screenshot 2024-11-26 at 11 04 22
No avatar
image

No name
Screenshot 2024-11-26 at 11 09 00

I tried using the existing `TlaCollaborator` component in the hopes that we could consolidate with the people menu at some point, but it ended up looking like shit with the border on the avatar when the file is selected/hovered. Also not sure the 'first initial' style of avatar that we use in the people menu is appropriate here. It was feeling rabbit holey and I think it would take significant work to get that stuff all working nicely and this PR is not the place to do it. So I just removed the unused `TlaCollaborator` component for now.

### Change type

- [x] `other`


apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
```diff
@@ -88,18 +102,52 @@ export function TlaSidebarFileLinkInner({
...
+						e.preventDefault()
+					}}
+					className={styles.guestBadgeTrigger}
+				>
+					{avatar ? <img src={avatar} style={{ objectFit: 'cover' }} /> : <TlaIcon icon="group" />}
+				</TlaTooltipTrigger>
+				<TlaTooltipPortal container={container}>
+					<TlaTooltipContent
+						style={{ zIndex: 200 }}
+						// this is also needed to prevent the tooltip from closing when clicking the badge
```
MitjaBezensek, line 137: Don't think this works.

MitjaBezensek (approved):
Nice! This is really starting to come together 👏

apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
ds300: hm works for me 🤷🏼

PR #4996, opened 2024-11-26
Fix an issue when navigating back to a forgoten file did not restore it.

MitjaBezensek:
The issue was that recent files was still returning the forgotten file, which meant that we did not navigate to a new file. So when you then went back in the browser you landed on the same route which didn't rerun the logic to enter a file.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Open a shared file.
2. Forget the file.
3. Navigate back in the browser.
4. The forgotten file should reappear.

### Release notes

- Fixes an issue when navigating back to a forgotten file did not restore it.


ds300 (approved):
ah, much simpler fix than I expected!

PR #4998, opened 2024-11-26
[botcom] Wait 1000s before marking a file as re-visited 

steveruizok:
This PR fixes a minor problem where we would rapidly mark a file as entered and exited when it loads. This is mostly a development problem, but it's good to address it anyway.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek:
Think this doesn't work as intended, I see that some file visits don't get their `firstVisitAt` updated.

image

This could be since we send the whole file state to the server and it might get overwritten with stale data. Let's wait for #4993  to go in then I'll retest.

MitjaBezensek (approved):
This looks good. The mentioned issue with `firstVisitAt` is a bug which I address here https://github.com/tldraw/tldraw/pull/5002

PR #4999, opened 2024-11-26
Improve mutation rejection toasts

MitjaBezensek:
Throttles them (3s). Uses different messages for different error codes.

You can test this by setting `rateLimited` to `true` and sending different `ZErrorCodes` [here](https://github.com/tldraw/tldraw/blob/eb0cbe7e18aaee9f017de2e6c2ce360130cd2fe9/apps/dotcom/sync-worker/src/TLUserDurableObject.ts#L171-L188).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -123,6 +121,31 @@ export class TldrawApp {
 		await this.z.query.file.where('ownerId', this.userId).preload().complete
 	}
 
+	messages = defineMessages({
+		publish_failed: { defaultMessage: 'Unable to publish the file' },
```
MitjaBezensek, line 125: Happy to accept any copy changes.

PR #5002, opened 2024-11-26
Fix an issue with `firstVisitAt` not getting set

MitjaBezensek:
I noticed an issue with `firstVisitAt` not getting getting set when I was reviewing #4998. The problem was that we didn't set the `firstVisitAt` when we created a new file. We then didn't allow updating the `firstVisitAt` field, which meant that it never got populated.

Since we don't allow changing `firstVisitAt` I could also simplify some other parts of mutating file state.

File state gets created in two cases:
- when we create a file we immediately also create the file state (this one now correctly sets `firstVisitAt`).
- for guest files we call `onFileEnter` which in turn calls `getOrCreateFileState` which already correctly set the `firstVisitAt` in the insert mutation.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -484,39 +484,33 @@ export class TldrawApp {
 		return this.getUserFileStates().find((f) => f.fileId === fileId)
 	}
 
-	updateFileState(fileId: string, cb: (fileState: TlaFileState) => Partial<TlaFileState>) {
+	updateFileState(fileId: string, partial: Partial<TlaFileState>) {
 		const fileState = this.getFileState(fileId)
 		if (!fileState) return
 		// remove relationship because zero complains
-		const { file: _, ...rest } = fileState
-		this.z.mutate.file_state.update({ ...cb(rest), fileId, userId: fileState.userId })
+		this.z.mutate.file_state.update({ ...partial, fileId, userId: fileState.userId })
 	}
 
 	async onFileEnter(fileId: string) {
 		await this.getOrCreateFileState(fileId)
-		this.updateFileState(fileId, (fileState) => ({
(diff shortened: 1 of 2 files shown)


apps/dotcom/client/src/tla/app/zero-polyfill.ts
```diff
@@ -153,7 +153,7 @@ export class Zero {
 					{
 						table: 'file_state',
 						event: 'insert',
-						row: { fileId: data.id, userId: store.user.id } as any,
+						row: { fileId: data.id, userId: store.user.id, firstVisitAt: Date.now() } as any,
```
MitjaBezensek, line 156: We didn't set the `firstVisitAt` when we created the file, so it never got populated (`firstVisitAt` is marked as a immutable column).

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -484,39 +484,33 @@ export class TldrawApp {
...
 		// remove relationship because zero complains
-		const { file: _, ...rest } = fileState
-		this.z.mutate.file_state.update({ ...cb(rest), fileId, userId: fileState.userId })
+		this.z.mutate.file_state.update({ ...partial, fileId, userId: fileState.userId })
 	}
 
 	async onFileEnter(fileId: string) {
 		await this.getOrCreateFileState(fileId)
-		this.updateFileState(fileId, (fileState) => ({
-			firstVisitAt: fileState.firstVisitAt ?? Date.now(),
```
MitjaBezensek, line 498: This was the only reason why `updateFileState` had to receive a callback. Since we can't mutate the `firstVisitAt` anyway I could remove it.

ds300 (approved):

Anilu522:
Hola

Anilu522:
Hola

PR #5008, opened 2024-11-27
Use lower end machines.

MitjaBezensek:
Don't think these workflows require the higher end machines.

Kept the following workflows on higher machines:
- Checks. Almost always the slowest workflow. Anything we can do to speed this one up is worth it.
- Deploy-dotcom. We want fast deploys.
- Playwright tests. We parallelize the tests. I lowered the cores for dotcom since it only uses 2 workers. That said still I had to use 8 cores (even with 4 postgres didn't boot up in time). Examples seems to use 8 workers when running on 16 core machine and is quite slow, so I kept it at that.
- All the publish workflows (publish-patch, publish-branch, publish-canary,...). Unsure if we can lower these ones. We often build all projects in them so they might benefit from more cores and more ram?

Had to create the 8 core runner in our orgranization -> actions -> runners.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek:
Trying one last time with `ubuntu-latest-4-cores-open`, will leave it open till tomorrow morning 😂 🤞 
https://github.com/tldraw/tldraw/actions/runs/12053978230/job/33610928986?pr=5008

MitjaBezensek:
Had to define the new 4 and 8 core runners in our organization's action runner config.

PR #5010, opened 2024-11-27
Make sure notes snap to grid after position is updated

trygve-aaberge-adsk:
The notes are updated to be centered around the creation point. However, the possible snapping to the grid was done before they were moved. This meant that if the note width or height was not a multiple of the grid size, they would not be snapped to the grid. This could occur e.g. if dynamic size was on.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Enable "Show grid" and "Dynamic size"
2. Zoom all the way in
3. Create a note
4. See that it's snapped to the grid (and that it wasn't snapped to the grid before this change)

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with notes not snapping to the grid unless the size was a multiple of the grid size.


SomeHats:
amazing, ty!

PR #5011, opened 2024-11-27
Create a utility type for making undefined properties optional

trygve-aaberge-adsk:
This extracts a type used in the object validation for making all properties of an object that accepts undefined optional. This is useful when you have an object of validators (e.g. props for a shape) and want to create the typescript type for an object that will be accepted by it. With this, any of the properties that has a validator with `.optional()` will be optional in that object.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Expose a utility type for making undefined properties optional

### DIFF
--- a/packages/validate/api-report.md
+++ b/packages/validate/api-report.md
@@ -80,6 +80,13 @@ function literal<T extends boolean | number | string>(expectedValue: T): Validat
 // @public (undocumented)
 function literalEnum<const Values extends readonly unknown[]>(...values: Values): Validator<Values[number]>;
 
+// @public (undocumented)
+export type MakeUndefinedOptional<T extends object> = Expand<{
+    [P in ExtractRequiredKeys<T>]: T[P];
+} & {
+    [P in ExtractOptionalKeys<T>]?: T[P];
+}>;
+
 // @public
 function model<T extends {
     readonly id: string;
@@ -103,11 +110,7 @@ function numberUnion<Key extends string, Config extends UnionValidatorConfig<Key
 // @public
 function object<Shape extends object>(config: {
     readonly [K in keyof Shape]: Validatable<Shape[K]>;
-}): ObjectValidator<Expand<{
-    [P in ExtractRequiredKeys<Shape>]: Shape[P];
-} & {
-    [P in ExtractOptionalKeys<Shape>]?: Shape[P];
-}>>;
+}): ObjectValidator<MakeUndefinedOptional<Shape>>;
 
 // @public (undocumented)
 export class ObjectValidator<Shape extends object> extends Validator<Shape> {
@@ -187,6 +190,7 @@ declare namespace T {
         unknownObject,
         ExtractRequiredKeys,
         ExtractOptionalKeys,
+        MakeUndefinedOptional,
         jsonValue,
         linkUrl,
         srcUrl,
--- a/packages/validate/src/index.ts
+++ b/packages/validate/src/index.ts
@@ -9,6 +9,7 @@ export {
 	Validator,
 	type ExtractOptionalKeys,
 	type ExtractRequiredKeys,
+	type MakeUndefinedOptional,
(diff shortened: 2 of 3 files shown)


SomeHats (changes requested):
in general (and i know there are _a lot_ of counter examples in the SDK right now) we try to avoid making things that aren't related to the SDK and whiteboarding part of our public API. 

Instead of exposing this as a generic utility, what do you think about updating [`RecordPropsType`](https://github.com/tldraw/tldraw/blob/8329c84b1049d98fd94d1ff46e29901205394fea/packages/tlschema/src/recordsWithProps.ts#L20) to include this, and keeping `MakeUndefinedOptional` tagged as `@internal`?

`RecordPropsType` is the intended way of going from a props object to the typescript type for that object: https://github.com/tldraw/tldraw/blob/8329c84b1049d98fd94d1ff46e29901205394fea/apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx#L54

trygve-aaberge-adsk:
> in general (and i know there are _a lot_ of counter examples in the SDK right now) we try to avoid making things that aren't related to the SDK and whiteboarding part of our public API.

Ah, that probably makes sense.

> Instead of exposing this as a generic utility, what do you think about updating [`RecordPropsType`](https://github.com/tldraw/tldraw/blob/8329c84b1049d98fd94d1ff46e29901205394fea/packages/tlschema/src/recordsWithProps.ts#L20) to include this, and keeping `MakeUndefinedOptional` tagged as `@internal`?
> 
> `RecordPropsType` is the intended way of going from a props object to the typescript type for that object:

That would help! The only thing we're currently using `MakeUndefinedOptional` for is to wrap `RecordPropsType` like this:

```
export type InferProps<S extends RecordProps<any>> = MakeUndefinedOptional<RecordPropsType<S>>
```

SomeHats:
oh yup sorry my mistake - we can't actually export public types that reference internal ones :(

I've pushed some tweaks to this here and will land that version: https://github.com/tldraw/tldraw/pull/5055

trygve-aaberge-adsk:
Alright, thanks!

Btw, if you prefer, feel free to push to my branch next time so you don't have to open a new PR.

trygve-aaberge-adsk:
> Instead of exposing this as a generic utility, what do you think about updating [RecordPropsType](https://github.com/tldraw/tldraw/blob/8329c84b1049d98fd94d1ff46e29901205394fea/packages/tlschema/src/recordsWithProps.ts#L20) to include this, and keeping MakeUndefinedOptional tagged as @internal?

@SomeHats: So I ran into a problem with this when we have a union of props. The old RecordPropsType could be used to transform this union, but with the new one I only get an empty object.

Example:
```
// RecordPropsType before this PR
type OldRecordPropsType<Config extends Record<string, T.Validatable<any>>> = Expand<{
  [K in keyof Config]: T.TypeOf<Config[K]>
}>

// Same as RecordPropsType currently defined in tldraw
type NewRecordPropsType<Config extends Record<string, T.Validatable<any>>> = MakeUndefinedOptional<{
  [K in keyof Config]: T.TypeOf<Config[K]>
}>

const Props1 = {
  a: T.number,
}

const Props2 = {
  b: T.string,
}

type Props1Type = OldRecordPropsType<typeof Props1 | typeof Props2>
// Resolves to:
//     type Props1Type = {
//         a: number;
//     } | {
//         b: string;
//     }

type Props2Type = NewRecordPropsType<typeof Props1 | typeof Props2>
// Resolves to:
//    type Props2Type = {}
```

SomeHats:
What's your use-case for a union of props?

For now, I'd suggest just including the relevant helper in your own code

PR #5013, opened 2024-11-28
Fix an issue with drag and dropping the files

MitjaBezensek:
We got an error saying there was a bad request when drag and dropping files. This is because we wanted to create a file state as the file owner, but the file did not exist yet (which is not allowed, we only allow that for guest files). So we now also create a file together with the file state.

image

I also noticed that the `createFile` function does not need to be async and also extracted document names from the files if we have them.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Drag some files to the sidebar.
2. They should correctly appear (with the correct names if they have time) and you should not see any rejected mutation toasts (like the one above).

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -12,6 +12,7 @@ import { Result, assert, fetch, structuredClone, throttle, uniqueId } from '@tld
 import pick from 'lodash.pick'
 import {
 	Signal,
+	TLDocument,
 	TLSessionStateSnapshot,
 	TLStoreSnapshot,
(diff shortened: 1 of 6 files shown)


MitjaBezensek:
Btw @mimecuvalo is this the intended copy? Or should it be `Added 3 .tldr files`?
CleanShot 2024-11-28 at 09 35 50@2x

ds300 (approved):

mimecuvalo:
> Btw @mimecuvalo is this the intended copy? Or should it be Added 3 .tldr files?

doh! no it is not the intended copy :P i'll fix up

PR #5016, opened 2024-11-28
[botcom] add animations for file list insertions/removals

ds300:
This is especially important for deletions and duplications. it can be really hard to follow what's going on in the file list without it. you know what they say, 'a lil bit of juice makes the mouth happy'

Using https://auto-animate.formkit.com/ which is small and respects `prefers-reduced-motion`.

https://github.com/user-attachments/assets/10a65601-118c-4d8a-9cfb-6536cc798f56

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/package.json
+++ b/apps/dotcom/client/package.json
@@ -27,6 +27,7 @@
 	},
 	"dependencies": {
 		"@clerk/clerk-react": "^5.15.5",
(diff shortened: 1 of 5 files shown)


ds300:
eh, not into this today, plus it messes with e2e tests. deletions and duplications will hopefully be rare enough that this shouldn't matter too much.

mimecuvalo:
this looked nice fwiw!

PR #5018, opened 2024-11-28
Don't trigger navigation if we are already on the correct file page

MitjaBezensek:
Prevents an issue with editor remounting since we navigate.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create new file.
2. Click on it in the sidebar
3. tldraw editor should not remount.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
@@ -2,7 +2,7 @@ import { TlaFile } from '@tldraw/dotcom-shared'
 import classNames from 'classnames'
 import { useEffect, useState } from 'react'
 import { Link, useParams } from 'react-router-dom'
-import { useContainer, useValue } from 'tldraw'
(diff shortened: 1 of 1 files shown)


ds300 (approved):
yeah this seems to fix the repro case, feel free to merge, but I wonder whether the deeper issue was some memoization problem causing a component remount even though the pathname didn't change.

MitjaBezensek:
> yeah this seems to fix the repro case, feel free to merge, but I wonder whether the deeper issue was some memoization problem causing a component remount even though the pathname didn't change.

If I'm not mistaken the issue seems to be the `mode` parameter that we added for creating and duplicating rooms. When you first click the create button we add the `mode=creating` param, then when you just click on the link that params is lost.
https://github.com/tldraw/tldraw/blob/4df35e04adfbf14cdeb7956d8cd34149b7a71848/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx#L159-L176

ds300:
ah makes sense. we can merge this then. thanks for looking into it.

PR #5019, opened 2024-11-28
Allow renaming files when pressing enter

MitjaBezensek:
This allows renaming active files when the sidebar has focus and when you press enter. When the file link becomes active it focuses itself and when focused it enters rename mode if we press enter.

This also prevents creating a new file if we first click the create file button and then press enter immediately. The focus changes to the new file and we will start renaming it.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Click create file button.
2. Press enter. You should not see another file getting created. Instead, you should be in the rename mode for the newly added file.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarCreateFileButton.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarCreateFileButton.tsx
@@ -19,7 +19,7 @@ export function TlaSidebarCreateFileButton() {
 
 	const handleSidebarCreate = useCallback(async () => {
 		if (!rCanCreate.current) return
-		const res = await app.createFile()
+		const res = app.createFile()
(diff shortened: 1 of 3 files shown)


apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarCreateFileButton.tsx
```diff
@@ -19,7 +19,7 @@ export function TlaSidebarCreateFileButton() {
 
 	const handleSidebarCreate = useCallback(async () => {
 		if (!rCanCreate.current) return
-		const res = await app.createFile()
+		const res = app.createFile()
```
MitjaBezensek, line 22: Flyby change.

ds300 (approved):
nice this seems to work very well 👍🏼

PR #5021, opened 2024-11-28
Disable creating new multiplayer rooms and snapshots (the old logic)

MitjaBezensek:
We don't have a feature flag system that would cover both FE and BE so at least for now I added a simple flag to `dotcom-shared`.

When we decide to go live we can turn `OLD_FILE_CREATION_DISABLED` to `true`. Then after a while we can remove all the logic that is gated behind it.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Set `OLD_FILE_CREATION_DISABLED` to `true`
2. You should not see any UI that lets you create a new shared room or snapshot.

### DIFF
--- a/apps/dotcom/client/src/components/FileMenu.tsx
+++ b/apps/dotcom/client/src/components/FileMenu.tsx
@@ -1,3 +1,4 @@
+import { OLD_FILE_CREATION_DISABLED } from '@tldraw/dotcom-shared'
 import { TldrawUiMenuActionItem, TldrawUiMenuGroup, TldrawUiMenuSubmenu } from 'tldraw'
 import {
 	FORK_PROJECT_ACTION,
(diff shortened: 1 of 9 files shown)


ds300:
This looks good but I don't know if we need a feature flag for this. Is there any benefit to merging this and testing it out early? It's a small diff so we can just yolo it when we're ready to disable the old stuff i think.

MitjaBezensek:
> This looks good but I don't know if we need a feature flag for this. Is there any benefit to merging this and testing it out early? It's a small diff so we can just yolo it when we're ready to disable the old stuff i think.

Yeah, had the same thought. Let's keep the PR open and we can decide on what to do when we go live.

MitjaBezensek:
Closing in favor of https://github.com/tldraw/tldraw/pull/5385

PR #5028, opened 2024-11-29
Add gecko analytics

MitjaBezensek:
Adds basic analytics for our newly introduced durable objects.

I also created a few more visualizations in our on our Events grafana dashboard. You can now also choose which environment you want to look at using the filter on the top left of the dashboard.

Link to [check the events](https://tldraw.grafana.net/d/adgumkhou3chsd/events?orgId=1&from=now-2d&to=now&timezone=browser&var-environment=pr-5028) for this PR. Data is quite sparse now.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -25,6 +26,8 @@ CREATE TABLE user_file_subscriptions (
 );
 `
 
+const ONE_MINUTE = 60 * 1000
```
MitjaBezensek, line 29: We could also push an event for each postgres update, but this feels sufficient for now.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -126,15 +127,16 @@ export class UserDataSyncer {
 	private queue = new ExecutionQueue()
 
 	async reboot(delay = true) {
-		// TODO: set up analytics and alerts for this
 		this.debug('rebooting')
+		this.logEvent({ type: 'reboot', id: this.userId })
```
MitjaBezensek, line 131: We combine user durable object and user data syncer events under `user_durable_object` name. We could also separate them, but this felt good enough for now since these two act in tandem.

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -92,27 +99,44 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
+		const now = Date.now()
+		if (this.postgresUpdates > 0 && now - this.lastRpmLogTime > ONE_MINUTE) {
+			this.logEvent({
+				type: 'rpm',
+				rpm: this.postgresUpdates,
+			})
+			this.postgresUpdates = 0
+			this.lastRpmLogTime = now
+		}
 	}
```
ds300, line 115: one of the beautiful things about the worker analytics engine is the sampling. This means we don't need to worry about how many events we write, querying should remain fast. The only concern should be pricing, but it's $0.25 per million events so 🤷🏼 if we wanted we could definitely log an event for every postgres update.

I'm happy with this implementation though, just noting that we don't need to be aggressively optimizing these.

ds300 (approved):
this is excellent, love the new dashboard stuff.

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
MitjaBezensek: Sound good!

Issue #4953, opened 2024-11-20
Updating shape props to undefined is ignored


kazu-2020:
### What happened?

When attempting to update an arbitrary value (in the example below, props.text) within the props of a shape to undefined using editor.updateShape, the update to undefined is ignored.

```ts
const shape = editor.getShape(createShapeId('sample'))
if (!shape) return

const newShape = {
  ...shape,
  props: {
    ...shape.props,
    text: undefined,
  },
}
editor.updateShape(newShape)
```

> [!NOTE]
>  if `editor.store.put` is used instead, the update to undefined works as expected 🙆🏻‍♂️  

### How can we reproduce the bug?

[CodeSandbox](https://codesandbox.io/p/sandbox/tldraw-example-forked-ykz2v2)

[Prerequisites]
The text displayed on the shape references the value of `props.text`, which is 「I'm a shape!」.

1. Press the "reset" button
    - When the button is pressed, the `props.text` of the displayed shape is updated to `undefined`.
2. The update is not reflected, and 「I'm a shape!」 is still displayed 😱

![CleanShot 2024-11-20 at 12 15 35](https://github.com/user-attachments/assets/eaf38a43-c6cd-49f7-900b-15bd7aea39f0)

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The developer package

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kazu-2020:
This issue is caused by the following lines in the `applyPartialToRecordWithProp` function defined in Editor.ts:

https://github.com/tldraw/tldraw/blob/eb7ed748a6770b642119472f927c8ac636941cdb/packages/editor/src/lib/editor/Editor.ts#L9804-L9806

If there are no concerns, I’d like to propose removing the condition `nextValue !== undefined`.
I applied this change in my clone of the repository and ran the tests, all of which passed successfully! 👍🏻

kazu-2020:
> This issue is caused by the following lines in the `applyPartialToRecordWithProp` function defined in Editor.ts:
> 
> https://github.com/tldraw/tldraw/blob/eb7ed748a6770b642119472f927c8ac636941cdb/packages/editor/src/lib/editor/Editor.ts#L9804-L9806
> 
> If there are no concerns, I’d like to propose removing the condition `nextValue !== undefined`. I applied this change in my clone of the repository and ran the tests, all of which passed successfully! 👍🏻

@steveruizok 
Can I create a PR? 🤔

[...]

PR #5029, opened 2024-11-29
fix: Updating shape props to undefined when using editor.updateShape

kazu-2020:
When using `editor.store.put`, any property within the props of a shape can be updated to undefined. However, when using `editor.updateShape`, updates to undefined are ignored.

To address this, I have modified `editor.updateShape` to allow updating any property within props to undefined as well.

related: #4953
### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Updating shape props to undefined  when using editor.updateShape

### DIFF
--- a/packages/editor/src/lib/editor/Editor.test.ts
+++ b/packages/editor/src/lib/editor/Editor.test.ts
@@ -1,14 +1,58 @@
-import { Box, createShapeId, createTLStore } from '../..'
+import {
+	Box,
+	Geometry2d,
+	RecordProps,
+	Rectangle2d,
+	ShapeUtil,
+	T,
+	TLBaseShape,
+	createShapeId,
+	createTLStore,
+} from '../..'
 import { Editor } from './Editor'
+
+type ICustomShape = TLBaseShape<
+	'my-custom-shape',
+	{
+		w: number
+		h: number
+		text: string | undefined
+	}
+>
+
+class CustomShape extends ShapeUtil<ICustomShape> {
+	static override type = 'my-custom-shape' as const
+	static override props: RecordProps<ICustomShape> = {
+		w: T.number,
+		h: T.number,
+		text: T.string.optional(),
+	}
+	getDefaultProps(): ICustomShape['props'] {
+		return {
+			w: 200,
+			h: 200,
+			text: '',
+		}
+	}
+	getGeometry(shape: ICustomShape): Geometry2d {
+		return new Rectangle2d({
+			width: shape.props.w,
+			height: shape.props.h,
+			isFilled: true,
+		})
+	}
+	indicator() {}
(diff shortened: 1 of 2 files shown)


kazu-2020:

steveruizok:
I think we could land this! I'll wait for another approval, just in case there is something I'm missing. (FWIW I'd recommend using `null` rather than `undefined` for "optional" properties).

kazu-2020:
> I'd recommend using `null` rather than `undefined` for "optional" properties

Yes, that is what I thought when I read the code...

My question is, if I replace the props of the persisted data in this case from undefined to null, is using [tldraw's migration](https://tldraw.dev/docs/persistence#Shape-props-migrations) function a better approach? 🤔

SomeHats (approved):
lgtm!

steveruizok:
Pushing an empty commit to tell the robots it's cool

trygve-aaberge-adsk:
Hm, this is a pretty hidden breaking change. If someone has a conditional property they send to `updateShapes` which they set to a value if they want the prop updated and to undefined if they don't want it updated, it will now be set to undefined instead. This set will not be caught by type checking unless `exactOptionalPropertyTypes` is enabled (which is default off) and it will bypass the prop validation so you may end up with a property set to undefined which does not allow the value to be undefined. This in turn may make the app crash when trying to read it.

I don't think this has affected us, but having this released in a patch version (3.6.1) without any notes about it when it has the potential to destroy the data in shapes is pretty bad imo.

I can see that it's useful to be able to set a property to undefined, but having different behavior when you specify a property as undefined and when you not specify it seems a bit surprising to me.

mimecuvalo:
You make a great point @trygve-aaberge-adsk . I think it was a slip-up that that change ended up in 3.6.1. It looks like it was meant for the 3.7.0 release (and was noted in those docs). 

I think we'll have to discuss this more internally, you do raise some good points on the DX on this. Seems like a footgun for folks...

trygve-aaberge-adsk:
> You make a great point @trygve-aaberge-adsk . I think it was a slip-up that that change ended up in 3.6.1. It looks like it was meant for the 3.7.0 release (and was noted in those docs).

Yeah, good that the change was mentioned in 3.7.0, however I think it would be better if it was also mentioned that it was a breaking change that could have this impact.

> I think we'll have to discuss this more internally, you do raise some good points on the DX on this. Seems like a footgun for folks...

Thanks, great if you can discuss if there's a better way to look out for and handle changes like this. I feel like breaking changes, especially ones like this that can easily be missed and have a bad impact, are important to look out for and communicate clearly.

PR #5041, opened 2024-12-02
Expand helpers available in actions / toasts overrides.

steveruizok:
Our actions and tools overrides have access to certain UI helpers, such as copy, print, or toasts. This PR adds additional items to those helpers which were previously only available internally.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Makes new helper available via actions and tools overrides.

### DIFF
--- a/apps/examples/src/examples/action-overrides/ActionOverridesExample.tsx
+++ b/apps/examples/src/examples/action-overrides/ActionOverridesExample.tsx
@@ -1,4 +1,4 @@
-import { Tldraw } from 'tldraw'
+import { Tldraw, TLUiActionItem, TLUiActionsContextType } from 'tldraw'
 import 'tldraw/tldraw.css'
 
 export default function BasicExample() {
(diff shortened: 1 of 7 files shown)


mimecuvalo (commented):
i love that we're exposing these better.  i think there's something bothering me about it all being in a generic grabbag/messy kitchen drawer called "helpers". is there a better way to name this? should they even be grouped in the first place?

steveruizok:
Nothing really jumps out as to organization, so I think a garbage grab bag is probably best for now. It would be better to use React itself as an interface here, because all of these are basically hooks that we're passing into the overrides manually, but components don't feel like the right interface either. 🤷

steveruizok:
This is another one that could be solved by allowing the editor to be declared outside of react, or at least in a hook, so that it could be configured before being passed into a <Tldraw editor={editor}/>. But that's not a great interface either... this is a tough one!

PR #5044, opened 2024-12-02
[dotcom] Handle max connections properly

ds300:
This PR adds a first class failure mode for limiting the number of users in a room. Before this PR the websocket connection would just hang and the user would see a spinner forever. Now the socket is closed and users see this 

image

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


SomeHats (approved):
nice, ty!

PR #5057, opened 2024-12-03
Improve rerenedring of the page menu and quick actions

MitjaBezensek:
Page menu used to rerender when we created a new shape. Undo / redo buttons also often rerendered even though they enabled / disabled status did not change. This improves the number of times these items rerender.

### Before  

https://github.com/user-attachments/assets/0192ba66-4e87-4f7f-8686-8173d2da6241

### After

https://github.com/user-attachments/assets/11b5c697-bc54-4781-861e-b5abc1c95324

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improves rendering of the pages menu and quick actions.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -3750,7 +3750,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	@computed getPages(): TLPage[] {
-		return this._getAllPagesQuery().get().sort(sortByIndex)
+		return Array.from(this._getAllPagesQuery().get()).sort(sortByIndex)
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


steveruizok:
Looks good! I don't think the performance overhead for the shallow compare is anything to be concerned about, I'll wait on @ds300's nod

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3750,7 +3750,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	@computed getPages(): TLPage[] {
-		return this._getAllPagesQuery().get().sort(sortByIndex)
```
steveruizok, line 3753: I think the `records` method creates new arrays each time, so this might not be needed

steveruizok (approved):

packages/editor/src/lib/editor/Editor.ts
ds300: this is required because the records method returns a computed value and it's not safe to mutate it with the .sort method becuase then the new shallow equality checks fail.

ds300 (approved):
👍🏼

PR #5059, opened 2024-12-03
[botcom] slurp local files on sign in

ds300:
This PR replaces the temporary multiplayer room from the logged out root page with the previous local editor, and 'slurps' the data into a new room during the sign in flow.

If something goes wrong the user sees this:
image

follow up work:

- [ ] e2e tests
- [ ] add Terms and conditions checkbox to sign in

I'll create tickets for these upon merging.

### Change type

- [x] `other`


apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -163,7 +156,35 @@ function TlaEditorInner({
...
+							setIsReady()
+						}
+					})
+					.catch((_e) => {
+						// show migration error
+						// give people link to /local and tell them to
+						// create a tldr file and drag it in
+					})
+			} else {
+				uploadLegacyFiles(editor, abortController.signal)
```
MitjaBezensek, line 182: I guess this is retry logic in case loading of some assets in the `if` block failed? We run this for all the files though, even when they have nothing to do with legacy files.

apps/dotcom/client/src/utils/scratch-persistence-key.ts
```diff
@@ -9,11 +9,19 @@
...
-	getFromLocalStorage(defaultDocumentKey) ?? 'tldraw_document_v3'
-setInLocalStorage(defaultDocumentKey, SCRATCH_PERSISTENCE_KEY)
+export function getScratchPersistenceKey() {
+	return getFromLocalStorage(defaultDocumentKey) ?? 'tldraw_document_v3'
+}
+
+setInLocalStorage(defaultDocumentKey, getScratchPersistenceKey())
+
+export function resetScratchPersistenceKey() {
+	const id = 'tldraw_document_v3_' + uniqueId()
```
MitjaBezensek, line 24: Once consequence of this is that the user might delete local storage (but not indexed db) and loose access to their document. Before we would always default back to the same key, so that could not happen.

ds300: alas you can't rename an indexeddb database so in order to avoid this we'd have to copy the contents to another db. Not impossible but meh if someone is messing around with their localstorage and they incur data loss it's their own fault IMO.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
ds300: yep it will stop immediately if there's no slurpPersistenceKey though

apps/dotcom/client/src/utils/scratch-persistence-key.ts
MitjaBezensek: Yeah, fair.

apps/dotcom/client/src/tla/components/TlaIcon/TlaIcon.tsx
```diff
@@ -26,14 +28,20 @@ export function TlaIcon({
...
 
+	const _className = classNames({
+		[styles.icon]: true,
+		[styles.inline]: inline,
+		[className]: true,
+	})
+
 	if (icon === 'none') {
-		return <div className={classNames(styles.icon, className)} />
+		return <span className={_className} />
```
ds300, line 38: had to change the icons to spans because otherwise they can't be used as children of paragraph elements.

apps/dotcom/client/src/tla/hooks/useAppState.tsx
```diff
@@ -47,11 +46,6 @@ export function AppStateProvider({ children }: { children: ReactNode }) {
 					app.dispose()
 					return
 				}
-				const claimTemporaryFileId = getFromLocalStorage(TEMPORARY_FILE_KEY)
-				if (claimTemporaryFileId) {
-					deleteFromLocalStorage(TEMPORARY_FILE_KEY)
-					app.claimTemporaryFile(claimTemporaryFileId)
-				}
```
ds300, line 54: this moved to the useEffect in the root page (currently called local.tsx)

MitjaBezensek (approved):
This got way more difficult that I imagined. Feels good though! 

Love that we keep the unsuccessfully imported files around and that we give the user a manual way out.

Issue #5052, opened 2024-12-03
Pasting images between tldraw instances is flaky


ds300:
### What happened?

Steps to reproduce:

1. add a handful of images to a local tldraw document (e.g. go to https://tldraw.com)
2. select and copy the images
3. create a new tldraw document (either local or remote) in an incognito window
4. paste the images

### Where did this happen?

The tldraw.com website

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5061, opened 2024-12-03
assets: fix up resolving when copy/pasting multiple items; also, videos

mimecuvalo:
Fixes issue https://github.com/tldraw/tldraw/issues/5052

This is a regression from https://github.com/tldraw/tldraw/pull/4682 (which was rolled up into https://github.com/tldraw/tldraw/pull/4659 )

Couple things were problematic here:
- main issue was that `debounce`'d urls all rolled up into one single debounce call that was _shared_ for all assets. this needed to be split to be per asset, which is now the case. (see `getResolveAssetUrlDebounced`)
- As long as I was looking at this it turns out videos weren't working at all. We weren't accounting for `data:video` as well as `data:image` - that's now been corrected. Not sure how long that's been broken for...
- Noticed also that `isReady()` was not called appropriately in the non-preview case - that's now fixed as well.
- finally, made `VideoShapeUtil`'s calling of `useImageOrVideoAsset` consistent with `ImageShapeUtil`

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed bugs with copy/pasting multilple assets from one board to another.
- Fixed bug with copy/pasting videos from one board to another.

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
+++ b/packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
@@ -136,14 +136,14 @@ function resolveAssetUrl(
 		})
 }
 
-const debouncedFunctions = new Map<TLAssetId, typeof resolveAssetUrl>()
+const debouncedFunctions = new Map<TLAssetId, typeof resolveAssetUrl & { cancel(): void }>()
 
 function getResolveAssetUrlDebounced(assetId: TLAssetId) {
 	if (!debouncedFunctions.has(assetId)) {
 		const debouncedFn = debounce(resolveAssetUrl, 500)
 		debouncedFunctions.set(assetId, debouncedFn)
 	}
-	return debouncedFunctions.get(assetId) as typeof resolveAssetUrl & { cancel(): void }
+	return debouncedFunctions.get(assetId)!
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -134,7 +136,15 @@ function resolveAssetUrl(
 		})
 }
 
-const resolveAssetUrlDebounced = debounce(resolveAssetUrl, 500)
+const debouncedFunctions = new Map<TLAssetId, typeof resolveAssetUrl & { cancel(): void }>()
+
+function getResolveAssetUrlDebounced(assetId: TLAssetId) {
```
SomeHats, line 141: if we have two images pointing to the same asset (quite common - duplicating the image will do this, it was the original motivation for the asset resolution system) this will still cause issues as only one of the hooks will actually update. 

Instead of declaring this outside of the component, what if we added it in the component before the main `useLayoutEffect`:

```tsx
const resolveAssetUrlDebounced = useMemo(
    () => debounce(resolveAssetUrl, 500),
    [],
)
```

mimecuvalo: much simpler! love it

PR #5071, opened 2024-12-05
custom sync presence

SomeHats:
Allow supplying a custom presence derivation to `useSync` and `useSyncDemo`.

### Change type
- [x] `api`

### Release notes

- It's now possible to customise what presence data is synced between clients, or disable presence syncing entirely.


packages/sync/src/useSync.ts
```diff
@@ -83,6 +87,7 @@ export function useSync(opts: UseSyncOptions & TLStoreSchemaOptions): RemoteTLSt
 	const schema = useTLSchemaFromUtils(schemaOpts)
 
 	const prefs = useShallowObjectIdentity(userInfo)
+	const getUserPresence = useEvent(_getUserPresence ?? getDefaultUserPresence)
```
ds300, line 90: not sure it's appropriate to use `useEvent` here because if the fn impl changes it won't cause the presence state signal to update reactively.

should probably do something like

```ts
const gup = _getUserPresence ?? getDefaultUserPresence
const getUserPresence$ = atom('getUserPresence', gup)
useLayoutEffect(() => getUserPresence$.set(gup), [gup])
```

packages/sync/src/useSync.ts
```diff
@@ -275,4 +290,12 @@ export interface UseSyncOptions {
...
 	/** @internal */
 	trackAnalyticsEvent?(name: string, data: { [key: string]: any }): void
+
+	/**
+	 * A reactive function that returns a {@link @tldraw/tlschema#TLInstancePresence} object. The
+	 * result of this function will be synchronized across all clients to display presence
+	 * indicators such as cursors. See {@link @tldraw/tlschema#getDefaultUserPresence} for
+	 * the default implementation of this function.
+	 */
+	getUserPresence?(store: TLStore, user: TLSyncUserInfo): TLPresenceStateInfo | null
```
ds300, line 300: I think at the same time as doing this we should really make the cursor and camera properties of `TLInstancePresence` nullable, and to have default values of `null`. Otherwise there's no way to e.g. actually omit the user's cursor or camera positions and for us to handle that while rendering presence components. It just gets set to a meaningless default (0,0). 

Maybe the same for `currentPageId` and `screenBounds`?

Also if they forget to add a `lastActivityTimestamp` it will always be 0 and people's cursors will always be hidden by our UI at least. Maybe we should switch the default for that to `+Infinity` to make the cursors never disappear unless they give a `lastActivityTimestamp`?

I think the rest of the properties should be fine since their default states are also empty states.

packages/sync/src/useSync.ts
SomeHats: good shout!

packages/sync/src/useSync.ts
SomeHats: This makes sense. On `lastActivityTimestamp`, I think we might have issues as `Infinity` can't be serialized as JSON - it comes through as `null` instead. Maybe we treat `null` as `Infinity`?

ds300: sounds good

ds300 (approved):
good stuff

packages/sync/api-report.md
```diff
@@ -41,6 +45,7 @@ export interface UseSyncDemoOptions {
 // @public
 export interface UseSyncOptions {
     assets: TLAssetStore;
+    getUserPresence?(store: TLStore, user: TLSyncUserInfo): null | TLPresenceStateInfo;
```
ncnlinh, line 48:
```suggestion
    getUserPresence?(store: TLStore, user: TLPresenceUserInfo): null | TLPresenceStateInfo;
```
@SomeHats I got type error when using the options:

### Sample code

```ts
  const store = useSync({
    uri: `${HOST}/connect/${roomId}`,
    getUserPresence(store, user) {
      const defaults = getDefaultUserPresence(store, user)
      return defaults
    },
  })
```

### error:
```vscode
Argument of type 'TLSyncUserInfo' is not assignable to parameter of type 'TLPresenceUserInfo'.
  Types of property 'color' are incompatible.
    Type 'string | null | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2345)
```

ncnlinh (approved):

mimecuvalo:
I seem to be running into an issue @SomeHats 
Screenshot 2024-12-10 at 11 53 15

PR #5073, opened 2024-12-06
[botcom] retry user requests on connection failure

ds300:
Added a retry helper to utils, and customized it for retrying postgres stuff if there's a connection issue.

### Change type

- [x] `other`


apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -283,80 +284,86 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
+				this.cache.store.updateOptimisticData([update], msg.mutationId)
+			}
+		})
+	}
 
+	private async handleMutate(msg: ZClientSentMessage) {
+		this.assertCache()
+		try {
+			// we connect to pg via a pooler, so in the case that the pool is exhausted
+			// we need to retry the connection. (also in the case that a neon branch is asleep apparently?)
```
MitjaBezensek, line 365: I reached out and got a fast first reply from them asking for more detailed information about when the issues occurred. I've send them the exact times of all the errors I found in sentry 🤞

PR #5074, opened 2024-12-06
[botcom] resizable sidebar

ds300:
Kapture 2024-12-06 at 12 00 34

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
Great work David. I pushed a few minor changes:
- use a little state machine so that it's clearer what's happening
- capture the pointer while resizing
- ..but make the close action happen on pointer up only if we didn't resize
- position the hover area more to the left of the indicator rather than directly centered on it (convention)

PR #5076, opened 2024-12-06
Improve dragging files to the sidebar ux

MitjaBezensek:
Switch to the dragged in files instead of staying on the current file.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Drag some files to the sidebar.
2. It should switch to one of those dragged files.


ds300 (approved):
good one

PR #5078, opened 2024-12-06
[botcom] fancypants routes

ds300:
Some self-indulgent typescript shenanigans, as an end-of-week palette cleanser 💆‍♂️

I found the route definitions kinda hard to read and add to, with all the prefixes string interpolation and helper functions stuff. I wanted to make it obvious how to add a new route and whether/how to add a helper function for creating paths or urls for that route.

I remembered gary bernhardt came up with some clever typesafe router thingy a few years ago after TS added template string types, and figured I could do something similar that lets us specify the routes as normal (uninterpolated, easy to scan) strings and then extract the param types, and also to compile helper functions automatically.

That's what this PR does. it adds a routeDefs.ts file that lets you quickly scan the list of routes on the site, and if you want to add a new route you put it there and it automatically compiles a helper fn. Then in routes.tsx we reference the route paths when constructing the route component hierarchy.

I don't feel super strongly about whether or not to merge this, tbh it was fine how it was 🤷🏼 

### Change type

- [x] `other`


MitjaBezensek (approved):
Nice! 👏 
I guess this is the approach you mentioned a while ago.

I like that a single param case is simplified, but don't like the generic `param` name. It's easy to navigate to the route and check what we need to pass in though (and it's also somewhat obvious from the route name).

steveruizok:
image

steveruizok (approved):

PR #5097, opened 2024-12-09
Copywriting consolidate on file

MitjaBezensek:
This also changes it from the `tldraw` package. It's definitely what we'd want for dotcom, but unsure if we want to make this change for other users of the package? I also didn't rename the keys for the `tldraw` package so that we don't break existing translations.

Created a separate commit for the two changes.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


MitjaBezensek:
> This ticket was only meant to be scoped to botcom.

Makes sense, updated the PR.

PR #5099, opened 2024-12-09
[botcom] move sidebar toggles in to sidebar layout so they're always visible

ds300:
Previously the sidebar toggle button would disappear when you see an error page because it was only rendered as a child of the editor's MenuPanel.

This was mitigated on desktop because we forced the sidebar open on error pages, but that doesn't work on mobile because the sidebar is a modal drawer.

So here I've just moved the toggle button to be rendered by the sidebar layout component, and added a spacer div where it would be in the MenuPanel.

this should be a pixel perfect match for what was there before, but it obviously makes the menu panel code a tiny bit more fragile. but since we see it all the time it's unlikely we'll introduce style regressions by accident.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarToggleMobile.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarToggleMobile.tsx
@@ -23,7 +23,7 @@ export function TlaSidebarToggleMobile() {
 				})
 			}}
 		>
-			<TlaIcon icon="sidebar" />
+			<TlaIcon icon="sidebar-strong" />
 		</button>
 	)
 }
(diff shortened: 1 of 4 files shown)


MitjaBezensek:
Looks like it is still shown behind the logo when the sidebar is open?

CleanShot 2024-12-09 at 18 46 03

steveruizok:
Updated the index (and the icon, surprisingly), looks good!

apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarToggleMobile.tsx
```diff
@@ -23,7 +23,7 @@ export function TlaSidebarToggleMobile() {
 				})
 			}}
 		>
-			<TlaIcon icon="sidebar" />
+			<TlaIcon icon="sidebar-strong" />
```
steveruizok, line 26: must have missed this when I added the strong icon

PR #5100, opened 2024-12-09
Fix sync worker source maps.

MitjaBezensek:
Sync worker source maps are not working for the `production` and `staging` environments.

The issue was that our sync worker naming does not follow the `${env.TLDRAW_ENV}-tldraw-multiplayer` naming. Instead, our `staging` worker is `main-tldraw.multiplayer` and our production on is just `tldraw-multiplayer`. We then used these names to set up sentry releases. 

But when we are sending sentry issue reports we do use this naming since `WORKER_NAME` uses the mentioned naming:
https://tldraw.sentry.io/settings/projects/tldraw-sync/environments/
image

This means that we were linking sentry source maps to incorrect environments.

CleanShot 2024-12-09 at 15 43 41

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/internal/scripts/lib/deploy.ts
+++ b/internal/scripts/lib/deploy.ts
@@ -85,7 +85,7 @@ export async function wranglerDeploy({
 	sentry?: {
 		authToken: string
 		project: string
-		release?: string
+		enviroment?: string
 	}
 }) {
 	const varsArray = []
@@ -129,7 +129,7 @@ export async function wranglerDeploy({
 	}
 
 	if (sentry) {
-		const release = sentry.release ?? `${workerNameMatch[1]}.${versionMatch[1]}`
+		const release = `${sentry.enviroment ?? workerNameMatch[1]}.${versionMatch[1]}`
 
 		const sentryEnv = {
 			SENTRY_AUTH_TOKEN: sentry.authToken,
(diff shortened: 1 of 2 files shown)


internal/scripts/lib/deploy.ts
```diff
@@ -85,7 +85,7 @@ export async function wranglerDeploy({
 	sentry?: {
 		authToken: string
 		project: string
-		release?: string
+		environment?: string
```
ds300, line 88: shouldn't this still be called `release` ? environments are a different thing in sentry

ds300 (approved):

internal/scripts/lib/deploy.ts
MitjaBezensek: I guess the release is the combination of an environment and a random string to specify which release exactly it is. And that what was the `release` prop used to hold, both of these values. But now we just send the first part of a relase.

And it seems like we do then [forward this value](https://github.com/tldraw/tldraw/blob/2f6593e71b5f5a4ed1d7db92cd0bf6394db02cfa/packages/worker-shared/src/sentry.ts#L32) as `environment` when we report issues.

Issue #5091, opened 2024-12-09
Flip behavior is incorrect


dodo-Riley:
### What happened?

Flipping a non-square shape changes its position,  Flip should not change the position of a shape.
[Peek 2024-12-09 17-04.webm](https://github.com/user-attachments/assets/0668beb4-a6ab-420c-8883-600b6723f455)

### How can we reproduce the bug?

Add a triangle, rotate it a bit, and press shift+h to flip it horizontally, which will move it up a bit. Flip it vertically to move it down a little bit.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The tldraw.com website

### Contact Details

sjh7397@gmail.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SomeHats:
good bug!!!

MitjaBezensek:
Related https://github.com/tldraw/tldraw/issues/4110

PR #5104, opened 2024-12-09
[fix] Flipped shape walking

steveruizok:
This PR attempts to improve (but does not entirely fix) a bug where rotated shapes were not being offset correctly. Generally: we were using page transform centers rather than rotated page transform centers.

Kapture 2024-12-09 at 16 34 06

This part of our codebase is really rough to work with because it relies on making updates and then rmeasuring the changes. Some of the information will be stale until these first changes are applied. IIRC I simplified this once in a branch that was never merged, I'll see if I can dig that up.

Fixes issue https://github.com/tldraw/tldraw/issues/5091

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Rotate a triangle shape.
2. Use the flip horizontal / flip vertical shortcuts.
3. Select multiple shapes that include a rotated shape
4. Flip h / v

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with rotating individual shapes.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -6696,7 +6696,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		const initialShape = opts.initialShape ?? this.getShape(id)
 		if (!initialShape) return this
 
-		const scaleOrigin = opts.scaleOrigin ?? this.getShapePageBounds(id)?.center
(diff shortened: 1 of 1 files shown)


steveruizok:
Tagging @ds300 in case you have any ideas here!

Issue #4873, opened 2024-11-07
Improve TLCopyType jpeg support


mimecuvalo:
### What happened?

From discord:
> The copyAs function's format parameter has the type TLCopyType that includes 'jpeg'. However, when you try to enter 'jpeg' format the following error occurs: copyAs.ts:41 NotAllowedError: Failed to execute 'write' on 'Clipboard': Type image/jpeg not supported on write.

> I think that the cause is that the browser clipboard can't handle jpeg images, so there is now way around it. Shouldn't the TLCopyType type remove the 'jpeg' option?

We could either try to add the support for `web image/jpeg` (where supported: https://developer.chrome.com/blog/web-custom-formats-for-the-async-clipboard-api ) or we remove the `jpeg` option.

### Where did this happen?

The developer package

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5114, opened 2024-12-12
Exports DX pass

SomeHats:
Over the last few weeks we've had a lot of requests on discord around asset resolution, exports, and the two together. Some of the APIs here have evolved and change independently of each other over time, so I wanted to take a pass at making them make sense with each other a bit more. There are a few things going on in this diff:

1. **BREAKING** The export/copy-as JSON option has been removed. I think this was only ever there as a debug helper, and it's impossible to actually make use of the JSON once copied (it's not the same as .tldr json which confuses people). 
2. `exportToBlob` is deprecated in favour of a new `Editor.toImage` method. `exportToBlob` has been the canonical 'turn the canvas into an image' helper for a while, but it has quite a weird looking signature and isn't very discoverable.
3. the `copyAs` and `exportAs` helpers have had a couple of args merged into the options bag for consistency.
4. The `jpeg` format has been removed from `copyAs`. This is technically a breaking API change, but since it never actually worked anyway due to browser limitations i think its fine.
5. SVG exports now resolve assets according to how they'll be used:
    - if it's for an SVG, we still use the existing `shouldResolveToOriginal` behaviour
    - if it's for a bitmap export, we request an image downscaled according to the size it will appear in that resulting bitmap.
6. Better reference docs for several APIs around this stuff.
7. **BREAKING** the `useImageOrVideoAsset` hook now requires passing in `width`, instead of reading `shape.props.w`. This is so it can be used with shapes other than our own. Whilst this is technically a breaking change, this limitation means its unlikely that it was used with many custom shapes in practice.
8. The clamping that we used to apply in to `steppedScreenScale` in `resolveAssetUrl` has been moved to our own implementations of `TLAssetStore`. This is so implementors can make their own decisions about the range of scalings they might want to use.
9. The `steppedScreenScale` limit changed from 1/8 to 1/32. This is because when testing with full res photos from a modern smartphone, we were still downloading a 1000+px image in order to render it at a few hundred px across (and also we don't pay anything for these right now). Happy to change now/in the future if this doesn't seem right though.

### Change type

- [x] `api`

#### Breaking changes / user facing changes
- The copy/export as JSON option has been removed. Data copied/exported from here could not be used anyway. If you need this in your app, look into `Editor.getContentFromCurrentPage`.
- `useImageOrVideoAssetUrl` now expects a `width` parameter representing the rendered width of the asset.
- `Editor.getSvgElement` and `Editor.getSvgString` will now export all shapes on the current page instead of returning undefined when passed an empty array of shape ids.

#### Product improvement
- When exporting to an image, image assets are now downloaded at a resolution appropriate for how they will appear in the export.

#### API changes
- There's a new `Editor.toImage` method that makes creating an image from your canvas easier. (`exportToBlob` is deprecated in favour of it)
- `SvgExportContext` now exposes the `scale` and `pixelRatio` options of the current export
- `SvgExportContext` now has a `resolveAssetUrl` method to resolve an asset at a resolution appropriate for the export.
- `copyAs(editor, ids, format, opts)` has been deprecated in favour of `copyAs(editor, ids, opts)`.
- `exportAs(editor, ids, format, name, opts)` has been deprecated in favour of `exportAs(editor, ids, opts)`


apps/examples/src/examples/basic/BasicExample.tsx
```diff
@@ -1,10 +1,22 @@
-import { Tldraw } from 'tldraw'
+import { useState } from 'react'
+import { Tldraw, TldrawOptions } from 'tldraw'
 import 'tldraw/tldraw.css'
 
 export default function BasicExample() {
```
steveruizok, line 5: Is this meant to stay on BasicExample? Seems like the changes should be a different example.

steveruizok:
Good pickup btw Alex! This part of the SDK definitely needed attention.

packages/editor/src/lib/editor/managers/TextManager.ts
```diff
@@ -146,6 +147,7 @@ export class TextManager {
 				// the character itself. we're only interested in the character.
 				const rects = range.getClientRects()
 				const rect = rects[rects.length - 1]!
+				if (!rect) whyAmIRunning()
```
steveruizok, line 150: Should this throw an error?

.yarn/patches/canvas-size-npm-2.0.0-822398a242.patch
```diff
@@ -0,0 +1,21 @@
...
++++ b/package.json
+@@ -33,10 +33,15 @@
+   "files": [
+     "dist"
+   ],
+-  "type": "module",
+   "main": "dist/canvas-size.js",
+   "module": "dist/canvas-size.esm.js",
+   "unpkg": "dist/canvas-size.min.js",
++  "exports": {
```
mimecuvalo, line 13: ugh, fun

apps/examples/vite.config.ts
```diff
@@ -77,7 +77,7 @@ export default defineConfig(({ mode }) => ({
 		'process.env.TLDRAW_IMAGE_URL': urlOrLocalFallback(
 			mode,
 			env === 'development' ? undefined : 'https://images.tldraw.xyz',
-			8989
+			8786
```
mimecuvalo, line 80: why the change here?

packages/tlschema/src/TLStore.ts
```diff
@@ -48,10 +48,30 @@ export type TLStoreSnapshot = StoreSnapshot<TLRecord>
 
 /** @public */
 export interface TLAssetContext {
+	/**
+	 * The scale at which the asset is being rendered on-screen relative to its native dimensions.
+	 * If the asset is 1000px wide, but it's been resized/zoom so it takes 500px on-screen, this
+	 * will be 0.5.
+	 *
+	 * The scale measured CSS pixels, not device pixels.
```
mimecuvalo, line 56: nit: measures

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -184,22 +184,6 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					helpers.exportAs(ids, 'png', getExportName(editor, defaultDocumentName))
 				},
 			},
-			{
```
mimecuvalo, line 187: so, i agree with these being only useful for debugging and not practically useful. (and the confusion with the .tldr json format)

howeeeever...before you remove them, would it make sense to have the snapshot json format be available? it feels like vendor-lock-in to not have a proper non-lossy export option, ya know?

packages/tldraw/src/lib/utils/export/copyAs.ts
```diff
@@ -10,10 +10,10 @@ import {
 	doesClipboardSupportType,
 	getAdditionalClipboardWriteType,
 } from '../clipboard'
-import { exportToBlobPromise, exportToString } from './export'
+import { exportToImagePromise } from './export'
 
 /** @public */
-export type TLCopyType = 'svg' | 'png' | 'jpeg' | 'json'
+export type TLCopyType = 'svg' | 'png'
```
mimecuvalo, line 16: so for `jpeg`, have you considered using `web image/jpeg`? related issue: https://github.com/tldraw/tldraw/issues/4873
to be clear, i haven't messed around with that new API to see if it works or not.

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -177,8 +180,20 @@ function SvgExport({
 			isDarkMode,
 			waitUntil,
 			addExportDef,
+			scale,
+			pixelRatio,
+			async resolveAssetUrl(assetId, width) {
+				const asset = editor.getAsset(assetId)
+				if (!asset || asset.type === 'bookmark') return null
```
mimecuvalo, line 187: this asset.type bookmark check seems brittle. one day when this TLAudioAsset PR finally lands 🙄 https://github.com/tldraw/tldraw/pull/4290 it'll also break this check (i think). maybe we should have some propery that denotes whether it is exportable like this.

packages/tldraw/src/lib/utils/export/export.ts
```diff
@@ -1,113 +1,11 @@
...
-}
-
 async function getSvgString(editor: Editor, ids: TLShapeId[], opts: TLImageExportOptions) {
-	const svg = await editor.getSvgString(ids?.length ? ids : [...editor.getCurrentPageShapeIds()], {
-		scale: opts.scale ?? 1,
-		background: editor.getInstanceState().exportBackground,
-		...opts,
-	})
+	const svg = await editor.getSvgString(
+		ids?.length ? ids : [...editor.getCurrentPageShapeIds()],
```
mimecuvalo, line 6: maybe `editor.getSvgString()` can helpfully do this bit for you if ids isn't passed in? ¯\\_(ツ)_/¯ 
i see you basically have to do this logic several times. (e.g. copyAs and somewhere else i think)

packages/tldraw/src/lib/utils/export/exportAs.ts
```diff
@@ -1,28 +1,62 @@
...
  *
  * @public
  */
 export async function exportAs(
 	editor: Editor,
 	ids: TLShapeId[],
-	format: TLExportType = 'png',
-	name: string | undefined,
-	opts: TLImageExportOptions = {}
+	opts: TLImageExportOptions & { format: TLExportType; name?: string }
```
mimecuvalo, line 25: wait, why does this have `& { format: TLExportType }` that's already part of TLImageExportOptions

packages/tldraw/src/lib/TldrawImage.tsx
```diff
@@ -184,6 +167,12 @@ export const TldrawImage = memo(function TldrawImage(props: TldrawImageProps) {
 		pixelRatio,
 	])
 
+	useLayoutEffect(() => {
```
mimecuvalo, line 170: i could be wrong but i don't think this needs to be `useLayoutEffect` (i.e. synchronous)

i think in general, for the places we use `useLayoutEffect` we should have a justification comment for it: "We're using layoutEffect here b/c x,y,z..."

apps/examples/src/examples/screenshot-tool/ScreenshotTool/childStates/Dragging.ts
```diff
@@ -76,17 +76,19 @@ export class ScreenshotDragging extends StateNode {
 				copyAs(
 					editor,
 					shapes.map((s) => s.id),
-					'png',
-					{ bounds: box, background: editor.getInstanceState().exportBackground }
+					{ format: 'png', bounds: box, background: editor.getInstanceState().exportBackground }
```
mimecuvalo, line 79: can `editor.getInstanceState().exportBackground` just be the default if not specified?

apps/examples/src/examples/export-canvas-as-image/ExportCanvasImageExample.tsx
```diff
@@ -9,17 +9,13 @@ function ExportCanvasButton() {
...
-					opts: { background: false },
-				})
+				const { blob } = await editor.toImage([...shapeIds], { format: 'png', background: false })
 
 				const link = document.createElement('a')
-				link.href = window.URL.createObjectURL(blob)
+				link.href = URL.createObjectURL(blob)
 				link.download = 'every-shape-on-the-canvas.jpg'
 				link.click()
+				URL.revokeObjectURL(link.href)
```
mimecuvalo, line 18: for some reason this makes me worried: is this consistent across browsers in working? i.e. will the link work before this is revoked? i'm worried there's some timing issue potential here.
maybe try in a couple browsers with a large 10+MB image to be sure.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8591,12 +8597,65 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * @param opts - Options for the export.
+	 *
+	 * @returns A blob of the image.
+	 * @public
+	 */
+	async toImage(shapes: TLShapeId[] | TLShape[], opts: TLImageExportOptions = {}) {
+		const withDefaults: TLImageExportOptions = {
+			format: 'png',
+			scale: 1,
+			pixelRatio: opts.format === 'svg' ? undefined : 2,
```
mimecuvalo, line 8619: feels like this `opts.format === 'svg' ? undefined : 2,` bit of logic should be in deeper in `getSvgString`/`getSvgAsImage` to avoid footguns/forgetfulness

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -8591,12 +8597,65 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+			...opts,
+		}
+		const result = await this.getSvgString(shapes, withDefaults)
+		if (!result) throw new Error('Could not create SVG')
+
+		const format = opts.format ?? 'png'
```
mimecuvalo, line 8625: maybe just use `withDefaults.format` so that it's consistent (in case you change the above's default)

packages/editor/src/lib/editor/managers/TextManager.ts
mimecuvalo: this looks like maybe leftover from testing, but could be wrong

packages/editor/src/lib/editor/types/SvgExportContext.tsx
```diff
@@ -29,10 +30,29 @@ export interface SvgExportContext {
...
 	readonly isDarkMode: boolean
+
+	/**
+	 * The scale of the export - how much CSS pixels will be scaled up/down by.
+	 */
+	readonly scale: number
+
+	/**
+	 * If the ultimate target for this export is a raster image, this is the number of raster pixels
+	 * in the resulting bitmap per CSS pixel in the resulting SVG. If this value is null, the export
```
mimecuvalo, line 52: nit, seems hard to read: "If this value is null or the export is to SVG, then the asset won't be downscaled."

packages/editor/src/lib/editor/types/misc-types.ts
```diff
@@ -8,17 +8,67 @@ export type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T,
 export type OptionalKeys<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
 
 /** @public */
-export interface TLImageExportOptions {
+export interface TLSvgExportOptions {
+	/**
+	 * The bounding box, in page co-ordinates, of the area being exported.
```
mimecuvalo, line 13: i'll allow this co-ordinates british-ism this time hah ;)

packages/editor/src/lib/editor/types/misc-types.ts
```diff
@@ -8,17 +8,67 @@ export type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T,
...
+	/**
+	 * If the export is being converted to a lossy bitmap format (e.g. jpeg), this is the quality of
+	 * the export. This is a number between 0 and 1.
+	 */
+	quality?: number
+
+	/**
+	 * The format to export as. Defaults to 'png'.
+	 */
+	format?: 'svg' | 'png' | 'jpeg' | 'webp'
```
mimecuvalo, line 69: can this just be `TLExportType`?

packages/tldraw/src/lib/utils/export/exportAs.ts
mimecuvalo, line 25: this is a little buried/hard-to-tell that down below `png` is set as the default  - the options for the export on this function might mentioned some of the defaults (or to refer to TLImageExportOptions)

mimecuvalo (commented):
Nice one @SomeHats. My comments are mostly a lot of little things.

The main thing I think maybe to discuss is on the export JSON removal — it feels like we're losing the only non-lossy export option that's out-of-the-box. We should consider/discuss having the snapshot export or something like it exposed in lieu of the current JSON export, otherwise it feels like there's a hole in our API.

apps/examples/src/examples/basic/BasicExample.tsx
SomeHats: oh yeah sorry this wants cleaning up!

apps/examples/vite.config.ts
SomeHats: it was wrong, the images weren't working in the examples app

packages/tldraw/src/lib/ui/context/actions.tsx
SomeHats: the snapshot json format is available through `editor.getSnapshot` / `editor.loadSnapshot` or the `snapshot` prop on the `<Tldraw />` component

SomeHats: there's also the getcontent/putcontent APIs which are used by the clipboard

packages/tldraw/src/lib/utils/export/copyAs.ts
SomeHats: the issue with `web image/jpeg` is that whilst it works for copying a jpeg back into tldraw, it won't interop with most other software unfortunately - unless they've explicitly added support for these custom formats. I think to most users in most use-cases, this would make the feature appear broken :(

packages/editor/src/lib/exports/getSvgJsx.tsx
SomeHats: yeah... this is a bit of an issue with our exports as a whole at the moment. although the audio asset at least has a `src` property? But yeah i wonder if assets are even quite the right abstraction here :/

I'll invert this to explicitly check for image/video

packages/tldraw/src/lib/utils/export/export.ts
SomeHats: yeah good call! i wasn't sure as this would be a breaking change but i'll flag it

packages/tldraw/src/lib/TldrawImage.tsx
SomeHats: ah yeah - i wasn't sure if a standard effect would always run if a layout effect caused several syncronous renders, but looks like it does!

apps/examples/src/examples/screenshot-tool/ScreenshotTool/childStates/Dragging.ts
SomeHats: oh yeah this is actually the case now anyway - i can remove this line

apps/examples/src/examples/export-canvas-as-image/ExportCanvasImageExample.tsx
SomeHats: under the hood, object URLs like this work sort of like refcounted references to the underlying blob. ie as long as a URL is created but not revoked, the blob can't be garbage collected. Once the download is started, the download stream from the blob is the new reference stopping it from getting garbage collected. Revoking the object URL won't abort the stream (there are a few examples from around the web of people testing this out)

packages/editor/src/lib/editor/Editor.ts
SomeHats: I can't push this down further because those methods don't know what format they're exporting to - they're just generating an svg, so they generate SVG style. I guess i could push these down further, but `getSvg({format: 'png'})` feels a bit weird/confusing to me

packages/editor/src/lib/editor/types/SvgExportContext.tsx
SomeHats: This is a bit tricky to write for as in the context where this is used, it's something we give the user rather than the user giving to us. So they should be the ones applying the downscaling, if they're writing a custom shape where that's relevant. I've updated this to try and make it make more sense though.

packages/tldraw/src/lib/utils/export/exportAs.ts
SomeHats: updated the docs!

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: I know that, programmatically, those are available on the APIs. The problem is now that they're hidden from UI by default, which effectively neuters their efficacy. 
To be clear, I'm saying we need to keep some UI around this out-of-the-box.

packages/tldraw/src/lib/utils/export/copyAs.ts
mimecuvalo: fair enough 👍 thought i'd check

packages/tldraw/src/lib/utils/export/exportAs.ts
mimecuvalo: oh i see - gotcha

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: oh, i see - i keep forgetting that `format` is optional.

packages/tldraw/src/lib/ui/context/actions.tsx
burdiyan: OK. I discovered this thread after a frustration. I wanted to export my board into JSON to adjust some text in a whole bunch of places. There's no "Find and Replace" functionality in TLDraw either, so I figured using JSON to export, edit it, and then import back would be my option. Until I realized the option is not there anymore :(

PR #5119, opened 2024-12-14
[botcom] Add pinned files

steveruizok:
This PR adds a pinned files feature.

Screenshot 2024-12-14 at 14 52 32

- A file_state tracks whether a user has pinned a file
- A user may pin or unpin a file from the file menu
- A user's pinned files are displayed in a section at the top of their recent files
- The order of pinned files is determined by most recently-edited

### Change type

- [x] `other`

### Test plan

- [x] End to end tests


steveruizok:
This turned out to be not too much of a change. I added some unrelated comments to our migrations as a drive-by.

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -210,26 +214,42 @@ export class TldrawApp {
...
+			date: number
+		}[] = []
 
 		for (const fileId of myFileIds) {
 			const file = myFiles[fileId]
 			const state = myStates[fileId]
 			if (!file || !state) continue
 			const existing = this.lastRecentFileOrdering?.find((f) => f.fileId === fileId)
-			if (existing) {
+			if (existing && existing.isPinned === state.isPinned) {
```
steveruizok, line 228: Important: while we want the list to be stable with regard to recent edits (ie we don't want editing a file to change its order in the sidebar until the user reloads), we _do_ want the pinned state to change the appearance of the sidebar immediately.

steveruizok:
Drive-bv fix to offline indicator

ds300 (approved):
Looks great. Did you try adding a pin icon next to the 'Pinned' section heading in the sidebar? That'd be the only thing I'd try to improve this. Otherwise people don't get very good feedback when clicking the pinned button. It just jumps to the top of the list and it's not immediately obvious what the difference is, you have to read the (extremely-deemphasized) 'Pinned' title.

steveruizok:
I'll try that, looks a little off as it breaks the grid of titles...

image

But alternatives don't look especially better:

Screenshot 2024-12-16 at 09 46 47

Screenshot 2024-12-16 at 09 46 59

ds300:
image
this one is fine IMO, although i might try bumping the icon up a tiny bit so the righthand ridge is on the text baseline. and maybe to the right a tiny bit too.

steveruizok:
I'm going to go pin on left, right looks too strange to me.

steveruizok:
Screenshot 2024-12-16 at 10 03 47

I think no icon also looks good (original gut call) but let's land this and see

steveruizok:
Could also add a recent icon to the first recent date group

Screenshot 2024-12-16 at 10 23 58

PR #5121, opened 2024-12-14
docs: consistent CardShape type naming

c-ehrlich:
### Change type

Docs

### Description

This type is named `CardShape` everywhere else in this code block. Copy/pasting the current version of the code block into an editor would result in a type error.

Worth noting that the ["Custom Shape and Tool" example](https://tldraw.dev/examples/shapes/tools/custom-config) uses `ICardShape`.


c-ehrlich:

steveruizok (approved):

steveruizok:
Thanks @c-ehrlich !

PR #5123, opened 2024-12-15
[botcom] Support legacy routes

steveruizok:
This PR adds (stubs) support for legacy routes in botcom. Those are:
- shared room
- shared room (readonly)
- shared room (readonly, old url)
- snapshot
- history
- history snapshot
- touchscreen sidebar (maybe)

I think it's better for us to treat these as "new routes" within botcom so that we can develop out the ways that these routes should be handled. Many routes need new UI and interactions, while the more rare internal routes can be mostly left as they are.

We don't want to break shared rooms etc for people, however we do want to have the prompts to sign in, etc, present on these rooms, similar to what we display to signed out users for files. And for signed in users, we should have a way to "claim" or "slurp" the file into your account's files.

Some notes:

- titles appear in title bar for legacy routes
- legacy routes support copy tab in share menu

### Change type

- [x] `other`


apps/dotcom/client/src/components/LocalEditor.tsx
```diff
@@ -47,11 +47,13 @@ const components: TLComponents = {
 	},
 	MainMenu: () => (
 		<DefaultMainMenu>
-			<LocalFileMenu />
-			<EditSubmenu />
-			<ViewSubmenu />
-			<ExportFileContentSubMenu />
-			<ExtrasGroup />
```
steveruizok, line 54: I removed the group from the extras group and added a "basic group".

apps/dotcom/client/src/tla/components/TlaAppMenuGroup/TlaAppMenuGroup.tsx
steveruizok, line 1: Split the component overrides for the editor into their own files here

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -1,98 +1,51 @@
...
-				</div>
-			)
-		}
-		return null
-	},
+	ErrorFallback: TlaEditorErrorFallback,
+	KeyboardShortcutsDialog: TlaEditorKeyboardShortcutsDialog,
+	MenuPanel: TlaEditorMenuPanel,
+	TopPanel: TlaEditorTopPanel,
+	SharePanel: TlaEditorSharePanel,
```
steveruizok, line 48: tidier

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -254,57 +207,10 @@ function TlaEditorInner({ fileSlug, mode, deepLinks, duplicateId }: TlaEditorPro
...
-							saveFileNames.set(editor.store, handle.name)
-						}
-					},
-				}
-
-				return actions
-			},
-		}
-	}, [app, fileSlug, handleUiEvent, untitledProject])
+	const overrides = useFileEditorOverrides({ fileSlug })
```
steveruizok, line 210: Moved this into its own hook

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -254,57 +207,10 @@ function TlaEditorInner({ fileSlug, mode, deepLinks, duplicateId }: TlaEditorPro
...
 
 	return (
-		<div className={styles.editor} data-testid="tla-editor">
+		<TlaEditorWrapper>
```
steveruizok, line 213: We're re-using this now in the other editors, so pulled into own component

assets/translations/main.json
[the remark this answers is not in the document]
ds300: i reckon we should revert these copy changes. the legacy stuff already has a different notion of what a file is and it creates confusion. e.g. "Opening a file from a shared file is not supported."

Also people are barely gonna be seeing this stuff after we deploy new dotcom, it's not worth the trouble.

apps/dotcom/client/src/tla/pages/legacy-history-snapshot.tsx
```diff
@@ -0,0 +1,112 @@
...
+	})
+	if (!result.ok) return null
+	const data = (await result.json()) as RoomSnapshot
+
+	return { data, roomId, timestamp }
+})
+
+export { loader }
+
+export function Component({ error: _error }: { error?: unknown }) {
```
ds300, line 56: this loses the ability to restore to a previous snapshot. which is guess is fine after we make old stuff readonly, but no point removing it for now.

I think we should actually just leave the history pages entirely alone. it's a hidden and badly implemented feature, it doesn't need app UI chrome or slurp affordances.

ds300 (approved):
this looks good aside from a couple of things mentioned elsewhere:

- probably best not to update the old UI copy from 'project' to 'file' since 'file' is already used to refer to .tldr stuff, and the copy changes will go out with the next dotcom release even if botcom is not ready by then
- i don't think we need to bother with doing the history pages here

assets/translations/main.json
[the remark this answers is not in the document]
steveruizok: Makes sense, will roll them back

apps/dotcom/client/src/tla/pages/legacy-history-snapshot.tsx
steveruizok: sounds good, I'll take a look. They're pretty useful for error recovery so I would like to have them, even if we don't slurp on them

steveruizok:
Threaded in the restore functionality, not my finest work but again these routes are semi-disposable.

Issue #5047, opened 2024-12-02
[Bug]: Frame rendering crashes after Editor is recreated


trygve-aaberge-adsk:
### What happened?

If you change any of the props passed to `Tldraw`/`TldrawEditor` so `new Editor` is called again (so any dependencies of the `useLayoutEffect` that is calling `new Editor`), and then e.g. zoom or create a new shape, all frames will render as "Error" and this exception is thrown:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'top')
    at TextManager.measureElementTextNodeSpans (TextManager.ts:151:22)
    at TextManager.measureTextSpans (TextManager.ts:256:39)
    at getFrameHeadingInfo (frameHelpers.ts:41:35)
    at FrameShapeUtil.getGeometry (FrameShapeUtil.tsx:61:23)
    at Editor.ts:4178:40
    at __UNSAFE__Computed.derive (Store.ts:787:15)
    at __UNSAFE__Computed.__unsafe__getWithoutCapture (Computed.ts:223:24)
    at haveParentsChanged (helpers.ts:22:20)
    at __EffectScheduler__.maybeScheduleEffect (EffectScheduler.ts:91:31)
    at flushChanges (transactions.ts:128:6)
```

It seems like the old `Editor` instance is still doing some stuff, which leads to `FrameShapeUtil.getGeometry` being called which does some text measurement. However, the element the text measurement is using to measure has been removed from the document when the new `Editor` instance was created which makes the measurement fail.

This started happening in v3.5.0 when `FrameShapeUtil.getGeometry` started using `TextManager.measureTextSpans`.

### How can we reproduce the bug?

Run tldraw like this:

```
import { useState } from 'react'
import { Tldraw, TldrawOptions } from 'tldraw'
import 'tldraw/tldraw.css'

export default function CustomOptionsExample() {
	const [options, setOptions] = useState<Partial<TldrawOptions>>({
		maxPages: 3,
	})

	return (
		<div>
			<button
				style={{ position: 'fixed', zIndex: 1000 }}
				onClick={() => void setOptions({ maxPages: (options.maxPages ?? 0) + 1 })}
			>
				Change options
			</button>
			<div className="tldraw__editor">
				<Tldraw persistenceKey="example" options={options} />
			</div>
		</div>
	)
}
```

After page load click the "Change options" button and then zoom the canvas.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The developer package

### Code of Conduct

[...]

PR #5125, opened 2024-12-16
fix hot reload text measurement bug

ds300:
fixes https://github.com/tldraw/tldraw/issues/5047#issuecomment-2544070258

the problem was that during hot reload react was rerendering the container elem but the editor wasn't recreated at the same, so the sneaky div that the text measurement manager adds was being disconnected from the DOM.

### Change type

- [x] `bugfix`

### DIFF
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -2528,8 +2528,6 @@ export type TestEnvironment = 'development' | 'production';
 export class TextManager {
     constructor(editor: Editor);
     // (undocumented)
-    _baseElm: HTMLDivElement | null;
-    // (undocumented)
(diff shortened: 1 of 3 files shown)


packages/editor/api-report.md
```diff
@@ -2528,9 +2528,11 @@ export type TestEnvironment = 'development' | 'production';
 export class TextManager {
     constructor(editor: Editor);
     // (undocumented)
-    baseElm: HTMLDivElement;
+    _baseElm: HTMLDivElement | null;
```
SomeHats, line 2531: `private`?

SomeHats (approved):
lgtm! thanks david

PR #5129, opened 2024-12-16
Add name and timeout params.

MitjaBezensek:
Improve `getPostgres` helper.:
- We can now pass in a name, which allows us to see the who is holding an open connection in `pg_stat_activity`. See `application_name` column below.
CleanShot 2024-12-16 at 15 47 58@2x
- `idleTimeout` option with a default of 30.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


apps/dotcom/client/public/tla/locales-compiled/en.json
```diff
@@ -56,6 +56,7 @@
   "8ff3d0a705": "Rate limited",
   "902b0d55fd": "Error",
   "904a830405": "Rename",
+  "91e49bb0eb": "Copy to my files",
```
MitjaBezensek, line 59: Looks like this wasn't correctly added in one of the previous prs.

PR #5131, opened 2024-12-16
rich text: all in, nix the `text` property

mimecuvalo:
This nixes the `text` property in lieu of strictly the `richText` property. That's a lot of places to update.
This is a PR on top of the rich text PR so that we review this bit separately.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


packages/tlschema/src/shapes/TLGeoShape.ts
```diff
@@ -197,7 +196,10 @@ export const geoShapeMigrations = createShapePropsMigrationSequence({
 		},
 		{
 			id: geoShapeVersions.AddRichText,
-			up: () => {},
+			up: (props) => {
```
ds300, line 199: you can't modify migrations like this, they won't run for recent rooms, you should add a new migration

PR #5133, opened 2024-12-17
Execute reactor immediately on listen

### LINKED ISSUE #5130: [Bug]: history is sometimes not flushed on changes
### What happened?

I'm experiencing an issue where the `onHistory` callback to `.listen` on the store is sometimes not called on changes. After some debugging I've found that this seems to be caused by the store being disposed before the historyReactor function has run for the first time. This means that it will never register the signal parents and is therefore not triggered again when signals change.

The Store is disposed when the Editor is recreated, which seems to happen because we're using react strict mode. The problem is that the historyReactor is only set up when the store is created, and cancelled when it's disposed. But when a new Editor is created, the historyReactor is not set up again to ensure it has at least run once to register the signal parents.

The reason that the Editor and Store can be disposed before the historyReactor has run is that it uses throttleToNextFrame, and that uses requestAnimationFrame to schedule the run. So the Editor and Store is disposed after requestAnimationFrame is called, but before the callback to requestAnimationFrame is called.

Since this depends on the timing of requestAnimationFrame it's hard to give an example that reliably reproduces this. However, you can likely cause it to happen by using react strict mode and adding some delay to the requestAnimationFrame callback in throttle.ts.

Chrome

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

ds300:
Fixes #5130 

### Change type

- [x] `bugfix`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug during development with React Strict Mode enabled where store.listen might end up not calling the listener.

### DIFF
--- a/packages/store/src/lib/Store.ts
+++ b/packages/store/src/lib/Store.ts
@@ -660,11 +660,12 @@ export class Store<R extends UnknownRecord = UnknownRecord, Props = unknown> {
 			},
 		}
 
-		this.listeners.add(listener)
-
 		if (!this.historyReactor.scheduler.isActivelyListening) {
 			this.historyReactor.start()
+			this.historyReactor.scheduler.execute()
 		}
+
+		this.listeners.add(listener)
 
 		return () => {
 			this.listeners.delete(listener)
@@ -763,7 +764,7 @@ export class Store<R extends UnknownRecord = UnknownRecord, Props = unknown> {
 	 *
 	 * @param name - The name of the derivation cache.
 	 * @param derive - A function used to derive the value of the cache.
-	 * @param isEqual - A function that determins equality between two records.
+	 * @param isEqual - A function that determines equality between two records.
 	 * @public
 	 */
 	createComputedCache<Result, Record extends R = R>(


SomeHats:
this looks good, but i wonder - should we be calling `store.dispose()` on `editor.dispose()`? the store can in theory outlive the editor, or have the editor instance it's attached to change (e.g. because of props changing).

Should the store-creating-things (`useSync`, `useLocalSyncStore`, etc) manage calling `store.dispose()`

trygve-aaberge-adsk:
Yeah, I also think it would make sense to not call store.dispose in editor.dispose. We now end up with a store that's always disposed when running with react strict mode, which doesn't really make sense.

Since the only thing store dispose is currently doing is cancel the next historyReactor effect, it doesn't seem to matter much (after this change). But if dispose is changed to do more in the future, that could introduce problems. I would think since that is the only thing it's doing now, it would mean means it isn't that big of a change to change it?

Another thing, I wondered what the reason was for executing the historyReactor specifically in listen, instead of doing it in the start function of the reactor instead of scheduling it there? Is this something that's very specific to the store and history reactor, and not something you want to do for other reactors?

PR #5137, opened 2024-12-19
Add an onCrop handler to ShapeUtil

trygve-aaberge-adsk:
This is called when cropping and allows you to override the crop calculations. This is useful for instance to set a min width and/or height for the cropping.

It can also be used to provide a completely different cropping mechanism, e.g. to keep the size of the shape fixed and instead resize the image when cropping. This would require overriding the SelectionForeground component, but it's better to only have to override that than the whole selection tool. And maybe that component can be made more dynamic in the future.

This also defines specific crop interfaces instead of using the ones from the image shape and uses the correct util instead of the image util when cropping. Any shape can define that it supports cropping, so I don't think it should be tied to the image shape.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Add an onCrop handler to a util for a shape that can be cropped.
1. Create a shape with a crop prop.
2. Try to crop the shape and see that the behavior in onCrop is applied.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add support for an onCrop handler on shape utils that allows you to prevent or modify the crop.
- The `TLImageShapeCrop` type has been replaced by `TLShapeCrop`.


trygve-aaberge-adsk (commented):
I have implemented support for cropping in a custom shape, and I'm currently having to do this to prevent the shape from being cropped too small:

```
override onBeforeUpdate(prev: MyShape, next: MyShape) {
  if (prev.props.crop == null || next.props.crop == null) {
    return next
  }

  if (next.props.w < MIN_WIDTH && next.props.h < MIN_HEIGHT) {
    return prev
  } else if (next.props.w < MIN_WIDTH) {
    const crop: TLImageShapeCrop = {
      topLeft: { x: prev.props.crop.topLeft.x, y: next.props.crop.topLeft.y },
      bottomRight: { x: prev.props.crop.bottomRight.x, y: next.props.crop.bottomRight.y },
    }
    return { ...next, props: { ...next.props, w: prev.props.w, crop } }
  } else if (next.props.h < MIN_HEIGHT) {
    const crop: TLImageShapeCrop = {
      topLeft: { x: next.props.crop.topLeft.x, y: prev.props.crop.topLeft.y },
      bottomRight: { x: next.props.crop.bottomRight.x, y: prev.props.crop.bottomRight.y },
    }
    return { ...next, props: { ...next.props, h: prev.props.h, crop } }
  }
}
```

With the change in this PR, I can replace all of that with just:

```
override onCrop(shape: MyShape, info: TLCropInfo<MyShape>) {
  return cropBox(shape, info, { minWidth: MIN_WIDTH, minHeight: MIN_HEIGHT })
}
```

steveruizok:
Thanks for the PR Trygve. I don't expect this to get any reviews before the 6th of Jan but will check it. (Glad you were able to use `onBeforeUpdate` to solve the issue so far though!)

packages/tldraw/src/lib/shapes/shared/crop.ts
```diff
@@ -0,0 +1,130 @@
...
+/** @public */
+export function getDefaultCrop() {
+	return {
+		topLeft: { x: 0, y: 0 },
+		bottomRight: { x: 1, y: 1 },
+	}
+}
+
+/** @public */
+export function cropBox<T extends ShapeWithCrop>(
```
mimecuvalo, line 21: nit: `cropBox` sounds like an object instead of an action, let's name this to `getCropBox`

mimecuvalo: I see that you're trying to match `resizeBox` (which makes sense) but I think that `resizeBox` might be misnnamed as well since it's not _actually_ doing the cropping/resizing, it's just getting the dimensions to apply.

packages/tldraw/src/lib/shapes/shared/crop.ts
```diff
@@ -0,0 +1,136 @@
...
+	shape: T,
+	info: TLCropInfo<T>,
+	opts = {} as CropBoxOptions
+): TLShapePartial<ShapeWithCrop> | undefined {
```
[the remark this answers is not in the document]
mimecuvalo: This is a great catch. We should still send back a partial.

Maybe the types should be updated to be:
`T & {
    x: number;
    y: number;
    props: {
        w: number;
        h: number;
    };
}`

mimecuvalo (commented):
Sorry for the delay in reviewing this! I'd like @SomeHats a look before approving but I think this is looking great @trygve-aaberge-adsk . Thanks so much for the contribution!

packages/tldraw/src/lib/shapes/shared/crop.ts
trygve-aaberge-adsk: Makes sense, I renamed it now.

packages/tldraw/src/lib/shapes/shared/crop.ts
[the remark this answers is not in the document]
trygve-aaberge-adsk: > Maybe the types should be updated to be: `T & { x: number; y: number; props: { w: number; h: number; }; }`

That would still require it to include all the properties of `T`, which it does not.  I think the correct type for the current implementation of `resizeBox` would be:

```
TLBaseShape<T['type'], { w: number; h: number }>
```

If it was changed to return a partial shape, the type could be:

```
TLShapePartial<TLBaseShape<T['type'], { w: number; h: number }>>
```

However, this would make `x`, `y` and `props` optional even though they're always set, and it would include several optional properties that are never set. So it might be better to type it explicitly:

```
{
	id: TLShapeId
	type: T['type']
	x: number
	y: number
	props: { w: number; h: number }
}
```

I changed the return type of `getCropBox` to be similar to this now. Let me know what you think.

Issue #2900, opened 2024-02-21
Selection foreground hides for what feels like a long time after changing styles


TodePond:
When you change the styles of selected shapes, we hide the selection foreground for a while. But it feels like too long to me. Especially noticeable on touch screens.

![Image](https://uploads.linear.app/8ba8cf67-bb46-4faa-8a97-5eb61e0814c1/da746361-f3e6-4451-86a9-b40224e2360a/233eb0fa-3dea-42ea-b536-144ceb97abcc?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzhiYThjZjY3LWJiNDYtNGZhYS04YTk3LTVlYjYxZTA4MTRjMS9kYTc0NjM2MS1mM2U2LTQ0NTEtODZhOS1iNDAyMjRlMjM2MGEvMjMzZWIwZmEtM2RlYS00MmVhLWI1MzYtMTQ0Y2ViOTdhYmNjIiwiaWF0IjoxNzA5NDA4NTI4LCJleHAiOjMzMjc5OTY4NTI4fQ.BhvIKUrGuFPe3\_-6wMtEBcfnpjM9GRV32lCRf6sKg24)

Taha-Hassan-Git:
This still happens!

melnikkk:
Hi) 
Is this issue still actual and waiting for a fix? 
If yes, I have a small fix and am ready to open the PR)

kostyafarber:
Should this be closed? PR seems to have fixed the issue.

PR #5158, opened 2024-12-29
Style changing duration reduced from 2 to 1 second

melnikkk:
This PR fixes the issue #2900. The style changing delay reduced from 2 to 1 second. I tried different values of the duration, such as 800, 900, and 1000. 1000 looks good - not so fast and not so slow.

https://github.com/user-attachments/assets/e327be0f-0491-4e68-af0a-5e5dd0dc8b00

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Style changing duration reduced from 2 to 1 second


melnikkk:

mimecuvalo:
@steveruizok sound good to you? not sure if this had some interesting reason for being 2 sec

steveruizok:
I'm ok with this, though it only solves part of the problem.

On mobile, we should cancel the "hide the UI while changing styles" when the style menu closes.

steveruizok:
ps thanks @melnikkk !

PR #5161, opened 2024-12-29
Provided an ability to select multiple arrow labels using ctrl/cmd key

melnikkk:
This PR solves the issue #5065. It makes it possible to select multiple arrow descriptions using the cmd key, which was impossible before.

https://github.com/user-attachments/assets/193ee887-1011-481d-b2d3-16dfa753c0fc

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Provided an ability to select multiple arrow labels using ctrl/cmd key

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingArrowLabel.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingArrowLabel.ts
@@ -57,6 +57,15 @@ export class PointingArrowLabel extends StateNode {
 		this._labelDragOffset = Vec.Sub(labelGeometry.center, pointInShapeSpace)
 
 		this.markId = this.editor.markHistoryStoppingPoint('label-drag start')
+
+    const additiveSelectionKey = info.shiftKey || info.accelKey
+		if (additiveSelectionKey) {
+			const selectedShapeIds = this.editor.getSelectedShapeIds()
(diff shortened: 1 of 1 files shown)


melnikkk:

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingArrowLabel.ts
```diff
@@ -57,6 +57,14 @@ export class PointingArrowLabel extends StateNode {
 		this._labelDragOffset = Vec.Sub(labelGeometry.center, pointInShapeSpace)
 
 		this.markId = this.editor.markHistoryStoppingPoint('label-drag start')
+
+		if (this.info.ctrlKey) {
```
mimecuvalo, line 61:
```suggestion
    const additiveSelectionKey = info.shiftKey || info.accelKey
		if (additiveSelectionKey) {
```

mimecuvalo (commented):
thanks for the PR! mostly looks good — I've added a slight tweak to make it consistent with the other key modifiers.

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingArrowLabel.ts
melnikkk: I see. Submitted and squashed. Thanks for review)

PR #5172, opened 2025-01-06
Fix relative CSS import rules failing to be fetched

### LINKED ISSUE #4939: Failure to parse relative CSS rules
### What happened?

In FontEmbedder.ts of tldraw editor, `getCurrentDocumentFontFaces` will fail to fetch the href of a CSS rule if it is relative: https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/exports/FontEmbedder.ts#L97

I noticed this while using the export to PNG/SVG functionality. When doing an export, the FontEmbedder crawls through document stylesheets, following any import rules. It does this by fetching the rule's href. Let's say once of these rules are relative, for instance `@import "../style.css"`. If you try to fetch that directly, you would be effectively doing `GET domain.com/style.css`, resulting in 404. Let's say the stylesheet containing the rule lives at `domain.com/a/b/parent_style.css`, then you'd want the rule's href to resolve to `domain.com/a/style.css`.

A workaround could be to make the href absolute first, so instead of: `fontFaces.push(fetchCssFontFaces(rule.href))`, you could do something along the lines of:

```
const absoluteHref = new URL(rule.href, rule.parentStyleSheet?.href ?? undefined).href
fontFaces.push(fetchCssFontFaces(absoluteHref))
```

Make an SVG export while using a stylesheet with a relative import rule.

_No response_

### Where did this happen?

The developer package

### Code of Conduct

trygve-aaberge-adsk:
When getting fonts for an export, all stylesheets on the page are traversed. If any CSS import rules are found, they are downloaded. However this just directly used the href of the import rule, which makes it relative to the document. Import rule URLs should however be relative to the style sheet they are included in. We therefore have to construct an absolute URL by using the style sheet as the base. If the import rule doesn't have a parent style sheet it means it's an inline style sheet and we should fall back to using the baseURI as the base.

Fixes #4939

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Include a relative `@import` rule on your page.
2. Trigger an export or print.
3. Observe that the rule is resolved correctly after this change.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix relative CSS import rules failing to be fetched when exporting or printing.

### DIFF
--- a/packages/editor/src/lib/exports/FontEmbedder.ts
+++ b/packages/editor/src/lib/exports/FontEmbedder.ts
@@ -94,7 +94,8 @@ async function getCurrentDocumentFontFaces() {
 				if (rule instanceof CSSFontFaceRule) {
 					fontFaces.push(parseCssFontFaces(rule.cssText, styleSheet.href ?? document.baseURI))
 				} else if (rule instanceof CSSImportRule) {
-					fontFaces.push(fetchCssFontFaces(rule.href))
+					const absoluteUrl = new URL(rule.href, rule.parentStyleSheet?.href ?? document.baseURI)
+					fontFaces.push(fetchCssFontFaces(absoluteUrl.href))
 				}
 			}
 		} else if (styleSheet.href) {


SomeHats (approved):
amazing, thank you!

Issue #5173, opened 2025-01-07
[Bug]: maxImageDimension does not seem to work as expected in registerDefaultExternalContentHandlers


tofu2323:
### What happened?

I tried setting maxImageDimension to a smaller value (e.g., 300) to resize uploaded images, but the images appear at their original dimensions.

Looking at defaultExternalContentHandlers.ts, I noticed that in the following section:
```
editor.registerExternalContentHandler('files', async ({ point, files }) => {
	// ...
})
```

It seems the resizing logic using maxImageDimension is not included. In contrast, the logic is applied here:
```
editor.registerExternalAssetHandler('file', async ({ file, assetId }) => {
	// ...
	if (isFinite(maxImageDimension)) {
		const size = { w: assetInfo.props.w, h: assetInfo.props.h }
		const resizedSize = containBoxSize(size, { w: maxImageDimension, h: maxImageDimension })
		if (size !== resizedSize && MediaHelpers.isStaticImageType(file.type)) {
			assetInfo.props.w = resizedSize.w
			assetInfo.props.h = resizedSize.h
		}
	}
	// ...
})
```

**My Question:**
Should the maxImageDimension resizing logic also be applied in registerExternalContentHandler to ensure consistent resizing of uploaded images? Or is this the intended behavior?

### How can we reproduce the bug?

1. Set maxImageDimension to a smaller value, such as 300.
2. Upload an image using tldraw.
3. Observe that the image appears at its original dimensions instead of being resized to fit within the maxImageDimension limit.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The developer package

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5176, opened 2025-01-07
Fix max image dimension prop not getting applied.

MitjaBezensek:
The problem was twofold:
- we weren't awaiting the updating of the assets
- we used old asset values when creating the shapes.

fixes https://github.com/tldraw/tldraw/issues/5173 / TLD-2910

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. In `develop.tsx` set `maxImageDimension` to some value.
2. Paste an image.
3. The image should be within the max dimension.

### Release notes

- Fix a bug with `maxImageDimension` not getting applied to pasted images.


mimecuvalo (approved):
LG - thank you!

PR #5181, opened 2025-01-08
Allow slurping of legacy multiplayer routes.

MitjaBezensek:
When a logged in user visits a legacy multiplayer room (`/q/r/roomId`) they can now click the `Copy to my files` button to slurp the multiplayer room to a file that gets created.

This currently reuses all the existing assets - the files don't get re-uploaded. We will need to re-upload them so that we will be able to associate them with the user. Decided to skip this step for now and come back to it once we have the assets/user/file associations ready.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a legacy multiplayer room.
2. Login to the app and use the multiplayer room link, but prefix it with `/q` (so something like `/q/r/roomId`
3. Click the `Copy to my files` button.
4. You should now see the same contents in the newly created file.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopRightPanel.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopRightPanel.tsx
@@ -2,7 +2,7 @@ import { SignInButton } from '@clerk/clerk-react'
 import { TlaFileOpenState } from '@tldraw/dotcom-shared'
 import classNames from 'classnames'
 import { useCallback, useRef } from 'react'
-import { useNavigate, useParams } from 'react-router-dom'
+import { useNavigate } from 'react-router-dom'
 import { PeopleMenu, useEditor, usePassThroughWheelEvents, useTranslation } from 'tldraw'
 import { routes } from '../../../routeDefs'
 import { useMaybeApp } from '../../hooks/useAppState'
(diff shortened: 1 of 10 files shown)


ds300 (commented):
this is looking good. I was imagining it being done server-side, like how room duplication works, but this seems fine too.

ds300:
one thing i noticed that's missing is using the document title for the file name

MitjaBezensek:
Added default file name handling and moved it to server side 👍

packages/dotcom-shared/src/types.ts
```diff
@@ -139,7 +139,17 @@ export interface ZClientSentMessage {
 	updates: ZRowUpdate[]
 }
 
-export type TlaFileOpenMode = 'create' | 'duplicate' | null | undefined
+export type TlaFileOpenState =
+	| { mode: 'create' }
+	| { mode: 'duplicate'; duplicateId: string }
+	| {
+			mode: 'slurp-legacy-file'
+			duplicateId: string
```
MitjaBezensek, line 147: Used the same name as it does feel like a similar action.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopRightPanel.tsx
```diff
@@ -60,12 +63,36 @@ export function TlaEditorTopRightPanel({
 	)
 }
 
+function getFileName(editor: Editor) {
+	const documentName = editor.getDocumentSettings().name
+	if (documentName?.length > 0) return documentName
+	const firstPageName = editor.getPages()[0].name
+	if (firstPageName.length > 0 && !firstPageName.startsWith('Page 1')) return firstPageName
```
MitjaBezensek, line 70: Thought of using the localized copy here, but I think we have a long standing bug where we always use the 'Page 1' as the default name.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopRightPanel.tsx
```diff
@@ -60,12 +63,44 @@ export function TlaEditorTopRightPanel({
...
+	const msg = useTranslation()
+	const defaultPageName = msg('page-menu.new-page-initial-name')
+
+	const documentName = editor.getDocumentSettings().name
+	if (documentName?.length > 0) return documentName
+
+	const firstPageName = editor.getPages()[0].name
+	if (
+		firstPageName.length > 0 &&
+		!firstPageName.startsWith('Page 1') &&
```
MitjaBezensek, line 77: We still default to `Page 1` in some cases (think during ensure store usable, possibly other places).

ds300 (approved):
looking good!

PR #5183, opened 2025-01-08
Add API changes check

MitjaBezensek:
Motivation for this comes from us [failing to include](https://discord.com/channels/859816885297741824/859816885801713728/1316827904348520528) some of the breaking changes in our release notes.

This adds a check if a PR has any `api-report.md` file changes. If it does it then makes sure that the PR description contains `### API changes` section in the description of the PR. I also updated the PR template to include that section.

Example of a [failure](https://github.com/tldraw/tldraw/actions/runs/12674521637/job/35323392961#step:4:23) and [correctly finding the API changes section](https://github.com/tldraw/tldraw/actions/runs/12674521637/job/35323947163#step:4:23).

One drawback is that updating the description won't re-trigger the check, so you need to manually do it (or push an empty commit).

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -23,8 +23,28 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - uses: ./.github/actions/setup
+
+      - name: Check API changes in PR description
+        run: |
+          if git diff --name-only origin/${{ github.base_ref }} HEAD | grep -q "api-report.md$"; then
+            echo "Found changes to api-report.md file"
+            pr_body=$(gh pr view ${{ github.event.pull_request.number }} --json body -q '.body')
+            echo "Checking for '### API changes' in PR body..."
+            if [[ $pr_body != *"### API changes"* ]]; then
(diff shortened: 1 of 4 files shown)


.github/workflows/checks.yml
```diff
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
MitjaBezensek, line 27: This adds about 18s to the run, not that happy about that. Looks like it's needed to do a diff against base though?

SomeHats (approved):
I'm happy for us to land this, but it may be frustrating that it doesn't re-run when the diff description changes. 

I'm working on https://github.com/tldraw/huppy-bot, which is a trimmed down version of the old huppy-bot that used to power all the bublic stuff. It was hosted on fly because it needed git FS access which was a whole mess, but this new one is going to be a normal next app on vercel. I hope to deploy this and switch huppy-bot over to it today. If you wanted you could add this check there instead and have it respond to description changes? But it would mean checking for the api-report.md files by API instead which might be annoying.

I guess alternatively this could be its own task with a different set of triggers?

Issue #5188, opened 2025-01-09
Show toast instead of throwing Error for maxFilesAtOnce


engin-can:
### What happened?

For some odd reason, decision was made to trigger `Error` instead of showing toast for `maxFilesAtOnce`. Can we turn this into toast so users can see it on UI, since all other checks do display toast...

```
editor.registerExternalContentHandler('files', async ({ point, files }) => {
    if (files.length > editor.options.maxFilesAtOnce) {
	    throw Error('Too many files')
    }
```
    

### Where did this happen?

The developer package

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #5201, opened 2025-01-12
Added toast instead of throwing an error for the case when the amount…

melnikkk:
This PR solves the issue #5188. Throwing an error for the case when a user is trying to upload more files than maxFilesAtOnce is replaced with toast.

_(for testing purposes and screen recording, I reduced the max amount of files)_
https://github.com/user-attachments/assets/b7468fd0-f14f-41c0-96d9-01f7326a0ecd

P.S.: I'm not sure about the correction of the text, so feel free to notify me, and I'll change it :)

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- added toast instead of throwing an error for the case when the amount of files is bigger than `maxFilesAtOnce`


melnikkk:

melnikkk:
I am also not sure that I work correctly with the strings that need to be translated. What is the dev flow for such strings?

If this moment was described in docs, pls just throw it to me, I didn't see it)

packages/tldraw/src/lib/ui/hooks/useTranslation/defaultTranslation.ts
```diff
@@ -125,6 +125,7 @@ export const DEFAULT_TRANSLATION = {
 	'assets.files.size-too-big': 'File size is too big',
 	'assets.files.type-not-allowed': 'File type is not allowed',
 	'assets.files.upload-failed': 'Upload failed',
+	'assets.files.amount-too-big': 'Too many files',
```
mimecuvalo, line 128: nit: `amount-too-many`

mimecuvalo (commented):
thanks for taking this on! i agree, this should be a toast.

the translation system at the moment is a bit confusing. 🙃 
- don't edit the `defaultTranslation.ts` file
- instead, edit `main.json`
- then, run `yarn refresh-assets`

SomeHats (approved):
looks good! thank you :)

melnikkk:
@SomeHats done, conflicts resolved :)

PR #5207, opened 2025-01-13
[botcom] Use clerk cookie to control routing

ds300:
This PR remove the `q` prefix for tla routes, but if you go to `/q` or `/beta` you will be met with a very basic sign-in page. when you sign in, you get the tla routes instead of the legacy routes. We can share this link on twitter or whatever when we want to let people in. (need to update the clerk settings to allow non-tldraw.com emails first)

### Change type

- [x] `other`


MitjaBezensek (approved):
👍

PR #5211, opened 2025-01-14
Fix for resizing snapshot bug

steveruizok:
Fixes issue described here: https://tldraw.sentry.io/issues/5687288900/?project=4504203639193600

Interesting that the event just before the bug was a resize on mobile followed by clicking the delete button. I wonder if there are cases where the resizing state doesn't correctly enter or exit?

### Change type

- [x] `bugfix`

### Release notes

- Fixed a bug that could occur when resizing on mobile

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts
@@ -56,7 +56,14 @@ export class Resizing extends StateNode {
 		this.parent.setCurrentToolIdMask(info.onInteractionEnd)
 		this.creationCursorOffset = creationCursorOffset
 
-		this.snapshot = this._createSnapshot()
+		try {
+			// On rare and mysterious occasions, the user can enter the resizing state with no shapes selected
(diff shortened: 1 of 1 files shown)


mimecuvalo:
I think I'd like to dig in a little bit to see why it started happening end of November. Seems like the was some PR around that time that maybe caused this weird side effect that I'd like to understand.

mimecuvalo:
> I think I'd like to dig in a little bit to see why it started happening end of November. Seems like the was some PR around that time that maybe caused this weird side effect that I'd like to understand.

ach, my bad - i was referencing the wrong issue when i said this. nevermind!

mimecuvalo:
I was gonna approve but actually, before you land, I'd like to leave this PR open so that I can better test this Sentry improvements: https://github.com/tldraw/tldraw/pull/5221
I was looking at the events, and I think it's when people are "fighting" over a shape but it's a guess that I'd like to confirm with more context.

As an aside, I'm guessing for this particular bug, I think it might be when two people are collaborating, one person is resizing and then the shape is erased underneath them, but it's a guess

PR #5214, opened 2025-01-14
close sockets on 'close' event

ds300:
apparently it is necessary to call .close on the server socket to make the tcp connection close cleanly, even after the client has called .close() ?

before

Kapture 2025-01-14 at 11 27 52

after

Kapture 2025-01-14 at 11 26 22

### Change type

- [x] `improvement`

### Release notes

- sync: close the server sockets on disconnect to proactively clean up socket connections.


SomeHats (approved):
sure 😅

PR #4758, opened 2024-10-23
botcom: inline rename

mimecuvalo:
Renaming is currently clunky with the dialog. This gets rid of the dialog and makes renaming inline.

https://github.com/user-attachments/assets/54dec16c-b0f8-467c-8547-b9d7392d1dcd

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

PR #5217, opened 2025-01-14
[botcom] save document name on blur in side bar

ds300:
before

Kapture 2025-01-14 at 15 46 34

after
Kapture 2025-01-14 at 15 45 15

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarRenameInline.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarRenameInline.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames'
-import { useCallback, useEffect, useRef } from 'react'
+import { useCallback, useRef } from 'react'
 import { TldrawUiInput } from 'tldraw'
 import { useApp } from '../../../hooks/useAppState'
 import { TLAppUiEventSource, useTldrawAppUiEvents } from '../../../utils/app-ui-events'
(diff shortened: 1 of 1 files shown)


mimecuvalo (changes requested):
i think this is a regression in https://github.com/tldraw/tldraw/pull/5019
i did add the blurring behavior in https://github.com/tldraw/tldraw/pull/4758

it's in the `function handleClick(e: MouseEvent) {` code above, if you take a look!
i forget why i avoided `onBlur` though... 🤔

ds300:
> i think this is a regression in https://github.com/tldraw/tldraw/pull/5019

looks like all the behavior mentioned in that PR still works, what do you mean exactly?

> it's in the function handleClick(e: MouseEvent) { code above, if you take a look!

oh yeah, didn't notice that, seems like it wasn't working now though, will investigate

ds300:
So it looks like the 'click' event wasn't being fired, not sure why. 'mousedown' works but I can't see why onBlur is a bad idea? seems to work as expected. I'll go with this for now since it's simpler and if you remember why the document event was useful we can revert.

mimecuvalo:
> So it looks like the 'click' event wasn't being fired, not sure why. 'mousedown' works but I can't see why onBlur is a bad idea? seems to work as expected. I'll go with this for now since it's simpler and if you remember why the document event was useful we can revert.

sounds good to me! yeah, it's lost to me why i didn't use `onBlur` and not sure how this all bitrotted.

mimecuvalo (approved):
lg - i think i'm only hesitant b/c there seemed to be some kind of race condition before:
`We wait a tick because we don't want to immediately close the input.` - so as long as it seems like it's saving consistently.

MitjaBezensek (approved):

PR #5219, opened 2025-01-15
worker debug logging

ds300:
This PR begins to address some pains we've felt with regard to logging on cloudflare workers:

1. a console.log invocation is only flushed to the console (or cf dashboard) once a request completes. so if a request hangs for some reason you never see the log output.
2. logs are very eagerly sampled, which makes it hard to rely on them for debugging because you never know whether a log statement wasn't reached or whether it was just dropped.

so this PR

- adds a Logger durable object that you can connect to via a websocket to get an instant stream of every log statement, no waiting for requests to finish.
- wraps the Logger durable object with a Logger class to consolidate code.

The logger durable object is on in dev and preview envs, but is not available in staging or production yet because that would require auth and filtering user DO events by user.

You can try it on this PR by going to https://pr-5219-preview-deploy.tldraw.com/__debug-tail and then opening the app in another tab

also tackles https://linear.app/tldraw/issue/INT-648/investigate-flaky-preview-deploys somewhat

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/routes.tsx
+++ b/apps/dotcom/client/src/routes.tsx
@@ -155,6 +155,7 @@ export const router = createRoutesFromElements(
 		}}
 	>
 		{isClerkCookieSet || isOverrideFlagSet ? tlaRoutes : legacyRoutes}
+		<Route path="/__debug-tail" lazy={() => import('./tla/pages/worker-debug-tail')} />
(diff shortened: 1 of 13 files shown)


MitjaBezensek (approved):
Nice!

Had to test this manually, looks like the worker is stuck here as well. Did it work before and it only gets stuck after a while?

PR #5221, opened 2025-01-15
error logging: add more context for errors

mimecuvalo:
After starting to review https://github.com/tldraw/tldraw/pull/5211 it struck me that I didn't have information when I was trying to look at Sentry to see what conditions caused the bug.
This adds more context for Sentry when debugging things. (I'm guessing for this particular bug, I _think_ it might be when two people are collaborating, one person is resizing and then the shape is erased underneath them, but it's a guess).

Edit: I didn't see we had `createErrorAnnotations` - I'm adding more to that function now.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


SomeHats (approved):
makes sense!

PR #5230, opened 2025-01-17
[botcom] don't broadcast mutation rejections

ds300:
Bad mutations that happen on one tab should not incur toasts on all tabs.

### Change type

- [x] `other`


MitjaBezensek (approved):
Nice 💅

PR #5263, opened 2025-01-22
Welcome dialog for preview users

ds300:
- for new users we show a welcome dialog that says
  - this is a preview
  - log out to get back to the old experience
  - (if they have content in indexeddb) would you like us to slurp your local data

- add an option for dialogs to prevent clicking the background to dismiss
- refactor the dialog/toasts system to not duplicate UI/context, allowing people to implement their own dialog/toast ui wherever they want. I think this is what steve originally wanted to do back when tla first landed. This was necessary to get a full-screen dialog with the useDialogs() hook and it just makes a lot more sense in general.

### Change type

- [x] `api`

### Release notes

- Breaking SDK Changes
  - TldrawUiToasts renamed to DefaultToasts
  - TldrawUiDialogs renamed to DefaultDialogs
- New SDK stuff
  - Toasts overridable component added
  - Dialogs overridable component added


MitjaBezensek (approved):
A radix dailog warning is shown in the console when we show the preview dialog.
CleanShot 2025-01-23 at 08 39 55@2x

PR #5273, opened 2025-01-23
Add editor option to allow sticky note resizing by scale

steveruizok:
This PR adds an option to resize sticky notes. There are a few ways that sticky notes _could_ resize however these are all a bit more complicated and would take time to unpick This PR only adds "scale" resizing while leaving some doors open for other options in the future. 

This PR also:
- fixes a bug with `getNoteForPit`
- changes the zIndex of handles to be in front of selection foreground

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Visit the "resize note shape" example
2. Resize a note shape

- [x] Unit tests

### Release notes

- Added `options.noteShapeResizeMode` editor option to control how note shapes resize.


packages/editor/editor.css
```diff
@@ -39,9 +39,9 @@
 	--layer-overlays-user-brush: 50;
 	--layer-overlays-user-indicator-selected: 60;
 	--layer-overlays-user-indicator-hovered: 70;
-	--layer-overlays-user-handles: 80;
 	--layer-overlays-user-snapline: 90;
 	--layer-overlays-selection-fg: 100;
+	--layer-overlays-user-handles: 105;
```
mimecuvalo, line 44: this is probably worth a brief comment: we're gonna forget later why this is a little more subtle

packages/editor/src/lib/options.ts
```diff
@@ -111,4 +116,5 @@ export const defaultTldrawOptions = {
 	actionShortcutsLocation: 'swap',
 	createTextOnCanvasDoubleClick: true,
 	exportProvider: Fragment,
+	noteShapeResizeMode: 'none',
```
mimecuvalo, line 119: wdyt about the default being `scale`? seems like it might be a nice thing to enable by default

mimecuvalo: if not in the SDK, i think this would be interesting to turn on for dotcom

packages/editor/src/lib/options.ts
```diff
@@ -66,6 +66,11 @@ export interface TldrawOptions {
 	 * external context providers. By default, this is `React.Fragment`.
 	 */
 	readonly exportProvider: ComponentType<{ children: React.ReactNode }>
+	/**
+	 * How should the note shape resize? By default it does not resize (except automatically based on its text content),
+	 * but you can set it to be user-resizable using scale.
+	 */
+	readonly noteShapeResizeMode: 'none' | 'scale'
```
mimecuvalo, line 73: actually, why have this be a set of options vs. a boolean? i _guess_ we might have other options like scale-x or scale-y but ¯\\_(ツ)_/¯ 

i suppose, maybe eventually there's a `wideboi` resize? 🤔

mimecuvalo (approved):
nice one, this is gonna be great. couple comments maybe to discuss (about what the right default should be) but would be happy to land!

packages/editor/src/lib/options.ts
steveruizok: I'd like to try free resize! Our sticky positioning logic really complicates things so holding off for now.

packages/editor/src/lib/options.ts
steveruizok: We could, yeah. I'm not opposed to it, other than that it complicates the UI. I might try doing a single resize corner or something in a follower.

PR #5275, opened 2025-01-23
[botcom] handle existing google users going through the sign-up flow

ds300:
So it turns out that sign-up with google would fail if the user was already signed up. they would be redirected to e.g. https://staging.tldraw.com/#/sso-callback?sign_up_force_redirect_url=https%3A%2F%2Fstaging.tldraw.com%2Fpreview&sign_in_force_redirect_url=https%3A%2F%2Fstaging.tldraw.com%2Fpreview and it wouldn't complete the session initialisation. 

I'm not sure why this happens but I asked clerk about it. In the meantime I've added some edge case logic for this situation and now users should be signed in as expected.

### Change type

- [x] `other`


ds300:
I switched the configuration back to how it was when we were doing QA earlier and it's working fine on my phone now. No idea what was going on. utterly baffled.

PR #5279, opened 2025-01-24
hard code context id for dialogs

ds300:
Quick fix for the pointer capture issue. I can't see the downside to hard-coding the context id here.

Ultimately I feel this contextId stuff should be refactored away, it has been a repeated source of subtle bugs. 

global state im comin for u 🗡️ 

### Change type

- [x] `other`


MitjaBezensek:
💣 😂

PR #5281, opened 2025-01-24
Fix line wobble issue

jamesbvaughan:
This fixes an issue where line shapes would wobble if you dragged points in a way that changes the shape's overall size after moving the first point away from the shape's origin point.

This re-introduces part of this fix from #1915

### DIFF
--- a/packages/editor/src/lib/components/Shape.tsx
+++ b/packages/editor/src/lib/components/Shape.tsx
@@ -80,8 +80,8 @@ export const Shape = memo(function Shape({
 			}
 
 			// Width / Height
-			const width = Math.max(Math.ceil(bounds.width), 1)
-			const height = Math.max(Math.ceil(bounds.height), 1)
+			const width = Math.max(bounds.width, 1)
+			const height = Math.max(bounds.height, 1)
 
 			if (width !== prev.width || height !== prev.height) {
 				setStyleProperty(containerRef.current, 'width', width + 'px')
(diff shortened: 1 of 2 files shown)


jamesbvaughan:

ds300:
Thanks for this! I can't seem to reproduce this bug in any major browsers. Are you able to share what os/browser you're using and what zoom level the browser was at? (knowing window.devicePixelRatio would be helpful too). cheers!

ds300:
thanks, i was able to reproduce with Arc.

I was hesitant to merge this as-is because sdk users tend to expect that the shape container div will be exactly the same size as their custom shape's geometry, and even small differences of < 1px can create noticeable problems. 

But I noticed that the Arrow shape doesn't have the same issue in Arc, and when investigating why I discovered that it adds a minWidth and minHeight setting on its svg container, which prevents these issues from manifesting. I think it would be best to do the same thing here for the Line shape.

https://github.com/tldraw/tldraw/blob/b00da028e6073a245d41246f31605f1ffab974e7/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx#L613

https://github.com/tldraw/tldraw/blob/b00da028e6073a245d41246f31605f1ffab974e7/packages/tldraw/src/lib/shapes/line/LineShapeUtil.tsx#L166

jamesbvaughan:
Thanks @ds300, I've tried that out here and it does work! I've updated the PR with that fix.

ds300 (approved):
Thanks!

PR #5283, opened 2025-01-25
Add Frame colors

steveruizok:
This PR adds an editor option to enable frame colors.

When enabled:

Screenshot 2025-01-25 at 11 35 20
image

This PR also:

- fixes a bug with the minimum size of frame headings
- tightens up the size of frame headings

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Visit the Frame colors example.
2. Create frames

### Release notes

- Added `editor.options.showFrameColors` option to display colors for frames.

### DIFF
--- a/apps/examples/src/examples/frame-colors/FrameColorsExample.tsx
+++ b/apps/examples/src/examples/frame-colors/FrameColorsExample.tsx
@@ -1,0 +1,12 @@
+import { Tldraw } from 'tldraw'
+import 'tldraw/tldraw.css'
+
+export default function FrameColorsExample() {
+	return (
+		<>
+			<div className="tldraw__editor">
+				<Tldraw persistenceKey="example" options={{ showFrameColors: true }}></Tldraw>
+			</div>
+		</>
+	)
+}
--- a/apps/examples/src/examples/frame-colors/README.md
+++ b/apps/examples/src/examples/frame-colors/README.md
@@ -1,0 +1,14 @@
+---
+title: Frame colors
+component: ./FrameColorsExample.tsx
+category: editor-api
+keywords:
+  - frame
+  - color
+---
+
+Create a frame (keyboard shortcut **F**) and see how it looks with the `showFrameColors` option enabled.
+
+---
+
+Use the `editor.showFrameColors" editor option to display colored fills and headings on frame shapes.
--- a/apps/examples/src/examples/ui-events/README.md
+++ b/apps/examples/src/examples/ui-events/README.md
@@ -2,7 +2,6 @@
 title: UI events
 component: ./UiEventsExample.tsx
 category: editor-api
-priority: 0
 keywords: [ui, events, api, x-ray]
 ---
 
--- a/packages/editor/api-report.md
+++ b/packages/editor/api-report.md
@@ -746,6 +746,7 @@ export const defaultTldrawOptions: {
     readonly maxPointsPerDrawShape: 500;
     readonly maxShapesPerPage: 4000;
     readonly multiClickDurationMs: 200;
+    readonly showFrameColors: false;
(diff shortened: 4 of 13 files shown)


steveruizok:
Some design questions:

I'm currently using `theme[color].fill` for the heading background rather than duplicating it as `theme[color].frame.fill`. The color theme can be modified from user-land. If we duplicated it, then users could change the heading fill without changing the color.

The offset position feels wrong; it seems like the heading body should be left aligned with the shape rather than the text content of the heading, which feels correct when the heading background is transparent (or the same color as the background).
 
Screenshot 2025-01-25 at 11 42 15

Screenshot 2025-01-25 at 11 42 04

steveruizok:
@trygve-aaberge-adsk had you asked for this one? If so please would love your thoughts!

steveruizok:
I've had to add a hack that hides the colors when the frame is selected if the option is not enabled. I wonder if other styles need this sort of "please don't display me" option. Feels sloppy but I'm happy to sneak it in until a second case comes up.

trygve-aaberge-adsk:
> @trygve-aaberge-adsk had you asked for this one? If so please would love your thoughts!

Hm, I don't think I have. I asked @jorgen-braseth-adsk and he said we (as in our team, might be before my time in it) have talked about it before and might have mentioned it to you at some point though.

It seems a bit strange to me to migrate the existing frames to black and have new frames have a default color of black 
instead of keeping them white as they are today. I think having the existing frames change color would be a problem for our users (if we enabled the option).

I also feel it's weird that they share the last selected color with other shapes. E.g. if you draw a red line, the next frame you create will be red, but the colors of your lines and your frames are not really related in my mind. Though, this is a general thing, not really specific to frames, as it's the same with notes. In our app we have created our own note shape though, which among other things only uses the previous color you used for other notes, regardless of the colors used for other shapes.

steveruizok:
> > @trygve-aaberge-adsk had you asked for this one? If so please would love your thoughts!
> 
> Hm, I don't think I have. I asked @jorgen-braseth-adsk and he said we (as in our team, might be before my time in it) have talked about it before and might have mentioned it to you at some point though.
> 
> It seems a bit strange to me to migrate the existing frames to black and have new frames have a default color of black instead of keeping them white as they are today. I think having the existing frames change color would be a problem for our users (if we enabled the option).

@mimecuvalo had raised the same point. We could create a `null` value here to use the "uncolored" style. 

> I also feel it's weird that they share the last selected color with other shapes. E.g. if you draw a red line, the next frame you create will be red, but the colors of your lines and your frames are not really related in my mind. Though, this is a general thing, not really specific to frames, as it's the same with notes. In our app we have created our own note shape though, which among other things only uses the previous color you used for other notes, regardless of the colors used for other shapes.

Perhaps it should be a different property than color? We do this with `labelColor`, though there's no UI for it.

trygve-aaberge-adsk:
> @mimecuvalo had raised the same point. We could create a `null` value here to use the "uncolored" style.

Yeah, I think that makes sense.

> Perhaps it should be a different property than color? We do this with `labelColor`, though there's no UI for it.

Yeah, a separate style prop is what we did for our custom note shape. For the UI we sort of merge the two style props so there's still only one color option to select. The way we do that is:

When a tool is active: We check if the tool has one of the two style props, and use the last value from that style prop.

When multiple shapes are selected: We check if all the shapes has one of the style props or if some have one and some have the other. If all have one style prop, we just use that. If both are used, we check if the values are the same and have that value selected in the UI if they are, otherwise no value is selected. When setting a color we update the style props that are used by the shapes.

This is sort of a workaround to have the behavior we wanted (separate last color for each shape type, but a common option to set the color) with the way style props work today though. If you're considering this in tldraw, maybe you can find a better way.

PR #5293, opened 2025-01-27
support react 19

SomeHats:
This diff adds support for react 19. React 19 is backwards compatible with react 18, excluding the removal of some deprecated APIs. React 18.3 (which we already use) warns when these APIs are used.

"adding support" here involves two things:
1. upgrading our depenencies to ones that specify react 19 in their peer deps array. I just upgraded everything that had a non-breaking upgrade to go to.
2. adding react 19 to our own list of peer dependencies

We want to make sure that we can still be used with react 18, so we're not upgrading to react 19 internally because we don't want to make use of any of the new APIs that aren't 18-compatible.

I've made the decision here not to do anything like running tests against both react versions. I think that this would be more effort to set up (and have more on-going cost on CI times etc) than it would be worth, and I trust react's guarantees about react 19's backwards compatibility and the warnings in 18.3.

### Change type

- [x] `improvement`

### Release notes

- tldraw now supports react 19


apps/examples/e2e/tests/export-snapshots.spec.tsx-snapshots/Export-snapshots-Fonts-dark-1-Mobile-Chrome-linux.png
mimecuvalo, line 1: i guess kinda similar to the rich text stuff, this is changing the position of the labels in some cases, expected?

internal/config/api-extractor.json
```diff
@@ -136,7 +136,7 @@
 		 * SUPPORTED TOKENS: <packageName>, <unscopedPackageName>
 		 * DEFAULT VALUE: "<unscopedPackageName>.api.md"
 		 */
-		"reportFileName": "api-report.md",
+		"reportFileName": "api-report",
```
mimecuvalo, line 139: should this be api-report.api.md? or is that auto-appended?

internal/scripts/build-api.ts
```diff
@@ -48,8 +48,8 @@ async function buildApi(sourcePackageDir: string) {
 			'--no-color',
 			'--no-ext-diff',
 			'--no-index',
-			'api-report.md',
-			'api/temp/api-report.md',
+			'api-report.api.md',
```
mimecuvalo, line 51: why was this necessary?

internal/scripts/lint.ts
```diff
@@ -9,7 +9,7 @@ async function main() {
 	try {
 		await exec(
 			'yarn',
-			['prettier', shouldFix ? '--write' : '--check', '--log-level=error', '--cache', relativeCwd],
+			['prettier', shouldFix ? '--write' : '--check', '--log-level=warn', '--cache', relativeCwd],
```
mimecuvalo, line 12: we should leave this as an error methinks

package.json
```diff
@@ -123,15 +123,17 @@
 		"canvas": "npm:empty-npm-package@1.0.0",
 		"@microsoft/tsdoc@npm:0.14.2": "patch:@microsoft/tsdoc@npm%3A0.14.2#~/.yarn/patches/@microsoft-tsdoc-npm-0.14.2-9988282153.patch",
```
mimecuvalo, line 124: we can rm this old one yeah?

package.json
```diff
@@ -123,15 +123,17 @@
...
 		"postgres@npm:^3.4.4": "patch:postgres@npm%3A3.4.5#~/.yarn/patches/postgres-npm-3.4.5-8a680ccbcd.patch",
-		"canvas-size": "patch:canvas-size@npm%3A2.0.0#~/.yarn/patches/canvas-size-npm-2.0.0-822398a242.patch"
+		"canvas-size": "patch:canvas-size@npm%3A2.0.0#~/.yarn/patches/canvas-size-npm-2.0.0-822398a242.patch",
+		"@microsoft/tsdoc@npm:~0.15.1": "patch:@microsoft/tsdoc@npm%3A0.15.1#~/.yarn/patches/@microsoft-tsdoc-npm-0.15.1-e24295d9bd.patch",
+		"@microsoft/tsdoc@npm:0.15.1": "patch:@microsoft/tsdoc@npm%3A0.15.1#~/.yarn/patches/@microsoft-tsdoc-npm-0.15.1-e24295d9bd.patch"
```
mimecuvalo, line 128: why are there 2 here?

mimecuvalo (approved):
LG ⚛️

internal/config/api-extractor.json
SomeHats: auto-appended

internal/scripts/lint.ts
SomeHats: When prettier encounters a file it can't format, it won't output any information about that when the log level is error. Those are counted as warnings. It means that it fails silently without any feedback.

apps/examples/e2e/tests/export-snapshots.spec.tsx-snapshots/Export-snapshots-Fonts-dark-1-Mobile-Chrome-linux.png
SomeHats: I've been trying to chase this down but I think this is like... browser-level sub-pixel/rounding issues that have come from upgrading playwright which means upgrading the version of chrome we test against

PR #5298, opened 2025-01-28
separately export default external content/asset handlers

SomeHats:
Currently, if you want to e.g. augment but not override one of our default content handlers (e.g. to add audio file support to pasting), you have to copy-paste the entirety of that handler into your app. This kind of sucks. Now, you can call our default handlers yourself, so you can augment how they work without having to completely re-implement them.

For reviewers: it's easiest to read this diff with [whitespace changed hidden](https://github.com/tldraw/tldraw/pull/5298/files?diff=split&w=1#diff-51f3e244dba5247299a2153f96348efdca84e0e8fb8fe27cff4443dd9d8d4161)

### Change type

- [x] `api`

### Release notes

- You can now import each of our external asset/content handlers, so you can augment them without having to copy-paste them into your app

#### BREAKING
- `TLExternalAssetContent` has been renamed to `TLExternalAsset`


packages/editor/src/lib/editor/types/external-content.ts
```diff
@@ -2,57 +2,98 @@ import { TLAssetId } from '@tldraw/tlschema'
...
-	  }
-
-/** @public */
-export type TLExternalContent<EmbedDefinition> = {
+	| TLTldrawExternalContentSource
+	| TLExcalidrawExternalContentSource
+	| TLTextExternalContentSource
+	| TLErrorExternalContentSource
+
+/** @public */
```
mimecuvalo, line 38: dbl public

package.json
```diff
@@ -71,7 +71,7 @@
 	"packageManager": "yarn@4.0.2",
 	"lint-staged": {
 		"*.{js,jsx,ts,tsx,json}": [
-			"prettier --write --cache --log-level=error"
+			"prettier --write --cache --log-level=warn"
```
mimecuvalo, line 74: unrelated change methinks

mimecuvalo (commented):
before you land, there's maybe 2 spots (maybe others?) that could benefit from this new pattern:
- SneakyOnDropOverride
- SneakyFileDropHandler

wdyt? should they generically reference the registered handler or should they explicitly call out the new defaults?

package.json
SomeHats: unrelated but needed (prettier calls file errors warnings) and i cba to split it out - mind if it keep it in?

mimecuvalo (approved):

PR #5318, opened 2025-01-29
Clear the dev db if we are missing any applied migrations

MitjaBezensek:
This fixes an issue if you:
- were on a branch that has a new migration
- you then switch to a branch without that migration or a branch that has a different migration

Till now we would just error out in this case. After this change we will clear the db, then apply all the migrations from start.

### Change type

- [x] `improvement`

### Release notes

- Fix an issue with running the zero cache project when the db contains some migrations that the current branch doesn't know about.

### DIFF
--- a/apps/dotcom/zero-cache/check-migrations.ts
+++ b/apps/dotcom/zero-cache/check-migrations.ts
@@ -1,4 +1,4 @@
-import { exec } from '../../../internal/scripts/lib/exec'
+import { execSync } from 'child_process'
 import { getPostgresAndMigrations, waitForPostgres } from './postgres'
 
 async function checkMigrations() {
@@ -8,7 +8,7 @@ async function checkMigrations() {
 	for (const appliedMigration of appliedMigrations) {
 		if (!migrations.includes(appliedMigration.filename)) {
 			console.error(`Migration ${appliedMigration.filename} is missing. Will clean the db.`)
-			await exec('yarn', ['clean'])
+			execSync('yarn clean')
 			break
 		}
 	}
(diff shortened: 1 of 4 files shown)


apps/dotcom/zero-cache/check-migrations.ts
MitjaBezensek, line 1: Pulled this into a separate file and yarn script since we also run migrate to run our production migrations.

apps/dotcom/zero-cache/package.json
```diff
@@ -11,9 +11,11 @@
 		"defaults"
 	],
 	"scripts": {
-		"dev": "concurrently 'yarn docker-up' 'yarn migrate --signal-success'",
+		"dev": "yarn check-migrations && concurrently 'yarn docker-up' 'yarn migrate --signal-success'",
 		"docker-up": "docker compose --env-file .env -f ./docker/docker-compose.yml up",
+		"docker-up-detached": "docker compose --env-file .env -f ./docker/docker-compose.yml up --detach",
 		"docker-down": "docker compose --env-file .env -f ./docker/docker-compose.yml down",
+		"check-migrations": "yarn docker-up-detached && yarn tsx ./check-migrations.ts && yarn docker-down",
```
MitjaBezensek, line 18: We run docker in a detached state, that's why we use `docker-down` in the end to make sure we clear it up.

ds300:
Trying this out, I'm seeing this error

image


Maybe a node issue, what version of node are you on? v22.11.0 here

MitjaBezensek:
Was on v20. Tried it with 22 and also seeing the issue, nice catch. Will take a look.

ds300:
this adds quite a lot of time to the dev-app startup, is there any way to reset the db while it's running so we don't need all these extra steps? like maybe dropping and re-creating the `public` schema?

Or maybe there's some way to write it so that it only incurs the extra docker up/down cycles if there's actually a problem?

MitjaBezensek:
Will drop this for now and maybe pick it up at a later time (if we'll still need it).

PR #5319, opened 2025-01-29
Emit a before-event from Editor

trygve-aaberge-adsk:
This will emit an event from Editor when dispatch is called that can be used to handle events before they are handled by tldraw.

The specific thing we need this for is to update camera options on wheel and pointer events before tldraw handles them in order to have a different wheelBehavior based on if you are using a mouse or a touchpad (which is determined from a heuristic based on the wheel and pointer events).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Emit a `before-event` from Editor for events before they are handled by tldraw.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -9333,6 +9333,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	dispatch(info: TLEventInfo) {
+		this.emit('before-event', info)
 		this._pendingEventsForNextTick.push(info)
 		if (
 			!(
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9333,6 +9333,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	dispatch(info: TLEventInfo) {
+		this.emit('before-event', info)
```
SomeHats, line 9336: i think instead of dispatch we should put this at the top of `_flushEventForTick`? dispatch doesn't actually do any work, and some events circumvent it and call flush directly

SomeHats (commented):
Thanks for this @trygve-aaberge-adsk!

packages/editor/src/lib/editor/Editor.ts
trygve-aaberge-adsk: I moved it to after the crash check. Since the event isn't processed by tldraw in that case, I guess it makes sense to not emit the `before-event` event either.

SomeHats (approved):
amazing thank you!

PR #5324, opened 2025-01-30
Public preview opt-in

ds300:
Instead of a full-blown banner, I just added an item in the menu, and a little dot to entice people to click. this is a common pattern when apps add new stuff and I think it works pretty well here, especially since the goal is just to make sure the backend can handle the scale before we go full switchover.

Kapture 2025-01-30 at 11 23 56

### Change type

- [x] `other`


ds300:
think I might put this behind a posthog feature flag in case it doesn't work out and we need to revert quickly

MitjaBezensek:
The whole experience feels a bit overwhelming, so many things to click:
- Click try new tldraw
- Opens a modal where I have to click sign up
- A bunch of clicks to create the account
- Immediately shown the preview dialog which I have to dismiss
- Cookies agreement

ds300:
well customizing the copy of the sign-up modal would be a minor pain so after removing the preview modal i just left the sign up modal as-is. Not the best UX but they still get the info  in the welcome modal after signing up, and maximising conversion is not the goal for this weekend anyway.

MitjaBezensek (approved):
👍

PR #5325, opened 2025-01-30
Combined sign in/up for dotcom

SomeHats:
Clerk's API for this is that combined sign in/up mode is enabled when a `signInUrl` is provided but a `signUpUrl` isn't. They sort of seem to have an assumption that there's a page somewhere that does sign in, which we sort of don't have (its not in the app because we use modals) but sort of do (clerk creates one automatically at https://accounts.tldraw.com/sign-in).

This diff specifies the `signInUrl` (enabling combined sign in/up mode) and creates a placeholder sign in page at that path (just in case someone somehow gets sent to it, although I'm not sure how/why this would happen).

Clerk likes react-router to be searching for a wildcard under `/sign-in`. We don't have wildcard support in our routing stuff at the moment, but I don't think we'd need to construct those URLs ourself anyway, so I've just dynamically configured the route for now.

### Change type
- [x] `other`


ds300:
this is getting stuck during sign-up for me?

1. delete your account in clerk staging
2. go to /preview and sign up
3. stuck on https://accounts.staging.tldraw.com/sign-up with a blank screen

Also the modal doesn't look any different, what's supposed to be different here?

ds300:
hmmm now if I sign up with sso i land here https://accounts.staging.tldraw.com/user

apps/dotcom/client/src/main.tsx
```diff
@@ -19,7 +19,11 @@ if (!PUBLISHABLE_KEY) {
 }
 
 createRoot(document.getElementById('root')!).render(
-	<ClerkProvider publishableKey={PUBLISHABLE_KEY} afterSignOutUrl={routes.tlaRoot()}>
+	<ClerkProvider
+		publishableKey={PUBLISHABLE_KEY}
+		afterSignOutUrl={routes.tlaRoot()}
+		signInUrl={routes.signIn()}
```
ds300, line 25: can we just make `/` the sign-in url and then not have the extra route?

ds300 (approved):

MitjaBezensek (approved):
So much cleaner 👏

ds300 (approved):

PR #5326, opened 2025-01-30
Allow embedding other multiplayer routes and also tldraw app routes

MitjaBezensek:
Looks like we only allowed embeding of `/r` routes. This adds support for other multiplayer routes (like the readonly routes, snapshots routes) as well as the new tldraw app routes (`/f`, `/p`).

We might also want to hide the sidebar for the tldraw app routes? 

### Change type

- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Fixed a bug with…


packages/tldraw/src/lib/defaultEmbedDefinitions.ts
```diff
@@ -1,7 +1,7 @@
 import { safeParseUrl } from '@tldraw/editor'
 
 // Only allow multiplayer embeds. If we add additional routes later for example '/help' this won't match
-const TLDRAW_APP_RE = /(^\/r\/[^/]+\/?$)/
+const TLDRAW_APP_RE = /(^\/[f|p|r|ro|s|v]\/[^/]+\/?$)/
```
mimecuvalo, line 4: feels like this might bitrot over time.
is it possible to include some kind of mechanism such that this is linked to https://github.com/tldraw/tldraw/blob/main/packages/dotcom-shared/src/routes.ts ?

MitjaBezensek: Yeah, it was broken for quite a while, we only supported `/r`, but we had readonly rooms and snapshots for quite a while now.

I guess we could add a property to route definitions saying if the route should be embeddable, then parse them in the build step to generate this regex.

There's also a question of how do we show the app file embeds. We probably don't want to show the toolbar there, so we might need to rethink how embeds work in this case.

steveruizok:
huppy check

PR #5332, opened 2025-01-30
Alex/custom layouts

SomeHats:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


packages/tldraw/src/lib/ui/components/TldrawUiLayout/TldrawUiLayout.tsx
```diff
@@ -0,0 +1,228 @@
...
+			bottomRight,
+		]
+		const rects = elements.map((element) => element.getBoundingClientRect())
+		const containerRect = container.getBoundingClientRect()
+
+		// we use this weird set up with indexes so that we can do all the measurements in one go,
+		// then apply all the changes in one go. This prevents layout thrashing.
+		const info = { spacingPx, containerRect, elements, rects }
+
+		layout('x', [0, 1, 2], squishIndexes[squishTop], info)
```
mimecuvalo, line 88: i might be annoying and say the readability of this is a little tough :) especially since lots of things get passed into the `info` grabbag.

while more verbose, we might consider something like:
```
const elements = {
  [LAYOUT.TOP_LEFT]: topLeft,
  [LAYOUT.TOP_CENTER]: topCenter,
  ...
}
...
const topLayoutItem = [
  LAYOUT.TOP_LEFT,
  LAYOUT.TOP_CENTER,
  LAYOUT.TOP_RIGHT
][squishIndexes[squishTop]]
layout('x', elements[topLayoutItem], info)
```
besides, `squishIndex` only gets used in the `layout` function to index it things so you might as well just index it here. feels like `layout` shouldn't know about this bit.

packages/tldraw/src/lib/ui/components/TldrawUiLayout/TldrawUiLayout.tsx
```diff
@@ -0,0 +1,228 @@
...
+	topLeft?: React.ReactNode
+	topCenter?: React.ReactNode
+	topRight?: React.ReactNode
+	centerLeft?: React.ReactNode
+	centerRight?: React.ReactNode
+	bottomLeft?: React.ReactNode
+	bottomCenter?: React.ReactNode
+	bottomRight?: React.ReactNode
+	spacingPx?: number
+	squishTop?: TldrawUiLayoutHorizontalPosition
```
mimecuvalo, line 24: we should come up with a better term than squish :P
as i play around with it, it's non-intuitive as to how what will happen heh.
is there some terminology we can borrow from the `flex` or `grid` layout? or maybe something like `prefer<X>`?

mimecuvalo (commented):
neat!

some high-level issues/discussion points:
- debug mode seems busted?
- watermark gets hidden too easily (may be a pre-existing thing? i'm not sure)
- speaking of debug mode, it'd be interesting to talk about a general app bar (whether on the bottom or top) and how this fits into this scheme. e.g. a Maintenance or Announcement toolbar that goes at the top of the layout (a somewhat common ui paradigm). or, this becomes more important if tldraw is, say, used as PDF viewer and there's more scrolling involved and you have an app bar that is above everything.
- related also to the previous point: on mobile, there maybe is a desire for something like an affordance for a bottom tray/navigation area? 
- and to really pile on, that all makes me wonder further about drawers/sidebars 😆

SomeHats:
missing lock on vertical toolbar

packages/editor/src/lib/hooks/usePassThroughWheelEvents.ts
```diff
@@ -1,16 +1,23 @@
 import { RefObject, useEffect } from 'react'
 import { preventDefault } from '../utils/dom'
 import { useContainer } from './useContainer'
+import { useEvent } from './useEvent'
 
 /** @public */
-export function usePassThroughWheelEvents(ref: RefObject<HTMLElement>) {
+export function usePassThroughWheelEvents(
+	ref: RefObject<HTMLElement>,
+	condition?: (e: WheelEvent) => boolean
```
steveruizok, line 9: Not sure if condition is necessary here, we could always do this check. I can't think of anywhere scrollable that I'd want to pass through wheel events. (Could also be extracted out into a separate easy-to-land PR)

steveruizok:
Just reading through this now.

I did have an initial thought about our UI for the past few months, which is that every time I've wanted to change something in user-land, I ended up abusing a handful of layout components (like top left) and ignoring the rest. I wonder if the ideal would be less overall layout rather than more, with each component knowing more about its individual positioning.

There are a few places I can think of where this would fall apart are few (i.e. squishable scrolling containers with dynamic height based on some other UI element).

steveruizok:
Generally like this!

I actually don't see why `DefaultLayout` and `TldrawUiLayout` are separate components, they appear to be the same thing and I can't think of any place where someone would use `TldrawUiLayout` in a new component except to pass it different props... but those are determined by the component overrides anyway?

I think a vertical toolbar makes sense as a feature (even a built-in feature) but that it will need more work than is shown here. The vertical toolbar example has a few problems, such as spacing, vertical overlap, and mobile experience. For example, the number of tools shown is based on the horizontal break point, which doesn't make sense for a vertical sidebar. 

Screenshot 2025-02-05 at 12 22 01

steveruizok:
Agreed with the OP that the layout we have here doesn't actually fit into either CSS grid or flexbox. We have a kind of round-robin layout where certain elements need to flex based on their neighbors without having effecting their column / row buddy dimensions.

image

For folks who have made significant changes to our UI, I'm almost certain that they've just rebuilt it. We can make that easier, either with an entirely different <Layout> component as done here, or with positional <TopLeftArea> <TopRightArea> components. IIRC we did discuss this in when doing the override components for those but felt that, due to the specific layout constraints of our layout elements, it was more accurate to describe them for what we were using them as, rather than generic positional components (which they were not).

mimecuvalo:
> Was there a developer issue / community ask related to layout?

Lingoda were asking me and Phil for this.

> For example: no support for bottom bar / debug panel, mobile layout being broken by the size of the watermark

Alex and I discussed this at length (along with a top App Bar/Notification bar). Let's discuss more offline why this is probably the right choice.

steveruizok:
For what it's worth, folks can bail out of the layout and use our components wherever they want. For Lingoda / left-aligned bottom toolbar, I might have suggested something like this (with a few more detail to handle things like focus mode). 

```tsx
function CustomToolbar() {
	return (
		<div
			style={{
				position: 'absolute',
				left: 10,
				bottom: 10,
			}}
		>
			<DefaultToolbar />
		</div>
	)
}

export default function BasicExample() {
	return (
		<>
				<Tldraw
					persistenceKey="example"
					components={{
						NavigationPanel: null,
						Toolbar: null,
						InFrontOfTheCanvas: CustomToolbar,
					}}
				></Tldraw>
			</div>
		</>
	)
}
```

Remember that the tldraw UI layout is just our default layout. Many of the customers who did more ambitious things with our layout have thrown it out and built their own. That's not to say we can't improve the ways that it's used or customized but the "fully customizable layout" for tldraw just React / CSS.

mimecuvalo:
>  I might have suggested something like this (with a few more detail to handle things like focus mode).

Yes, I told them Lingoda solution over email as well.
However, the problem is that you still run into colliding UI "islands", which is what Alex is tackling here.

steveruizok:
I think the answer is to make our UI more composable (or find the places where it's not composable enough already), rather than bake in a more structured layout. I just don't see this as an actual solution to the problem since every app for whom this would be useful would have their own requirements. I think its a problem definition problem: the best people to write a layout for tldraw's UI are the developers building the app around tldraw.

PR #5333, opened 2025-01-31
fix duplicating + publishing rooms with big snapshots

ds300:
We had a bug where if the room snapshots were bigger than 1mb, duplicating and publishing would fail.

### Change type

- [x] `bugfix`


apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
```diff
@@ -117,6 +117,12 @@ function FileItems({
 		const file = app.getFile(fileId)
 		if (!file) return
 		const res = app.createFile({ id: newFileId, name: getDuplicateName(file, app) })
+		// copy the state too
+		const prevState = app.getFileState(fileId)
+		app.getOrCreateFileState(newFileId)
+		app.updateFileState(newFileId, {
+			lastSessionState: prevState?.lastSessionState,
+		})
```
ds300, line 125: drive-by ux fix, to preserve page and viewport locations

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -518,17 +519,10 @@ export class TLDrawDurableObject extends DurableObject {
...
-					const otherRoom = this.env.TLDR_DOC.get(
-						this.env.TLDR_DOC.idFromName(`/${ROOM_PREFIX}/${this.documentInfo.duplicateId}`)
-					) as any as TLDrawDurableObject
-					data = await otherRoom.getCurrentSerializedSnapshot()
-				} catch (_e) {
-					data = await this.r2.rooms
-						.get(getR2KeyForRoom({ slug: this.documentInfo.duplicateId, isApp: true }))
-						.then((r) => r?.text())
-				}
+				await getRoomDurableObject(this.env, this.documentInfo.duplicateId).awaitPersist()
```
MitjaBezensek, line 522: I guess this still does not guarantee the data to be up to date since we are using an `ExecutionQueue`? Push to it will happen, but the task might not run in time?

ds300: it waits for the enqueued task to complete

MitjaBezensek (approved):

PR #5340, opened 2025-02-01
Add option to disable numbered shortcuts on the toolbar

steveruizok:
This PR:

- adds an editor option to disable the numbered keyboard shortcuts for the default toolbar.
- improves the UX of keyboard shortcuts for the toolbar (allowing simultaneous keys, for fast changing)

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Use the keyboard shortcuts
2. When the option is off, keyboard shortcuts should be ignored

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- SDK: Added editor option to disable 0-9 keyboard shortcuts for the toolbar
- Improved keyboard shortcuts for the toolbar

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
@@ -1,6 +1,5 @@
 import { preventDefault, useEditor, useEvent, useUniqueSafeId } from '@tldraw/editor'
 import classNames from 'classnames'
-import hotkeys from 'hotkeys-js'
 import { createContext, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { PORTRAIT_BREAKPOINT } from '../../constants'
 import { useBreakpoint } from '../../context/breakpoints'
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
```diff
@@ -95,44 +122,24 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
...
-				)
-
-				const el = relevantEls[index]
-				if (el) el.click()
-			})
+		function handleKeyDown(event: KeyboardEvent) {
+			if (areShortcutsDisabled(editor)) return
+			// no accelerator keys
+			if (event.ctrlKey || event.metaKey || event.altKey || event.shiftKey) return
+			for (const [key, index] of KEYS) {
```
SomeHats, line 131: if we make KEYS an object instead of an array we could skip this iteration on every keypress and instead just look up the relevant index

SomeHats (approved):
👍

PR #5348, opened 2025-02-04
Move to server side slurping.

MitjaBezensek:
This moves duplication and legacy room slurping to the backend:
* It removes the the file open mode logic (no more setting it when navigating, no more sending search params to the backend).
* We now insert a file into the db with an additional column called `createSource` which is in the form of `f/kZTr6Zdx6TClE6I4qExV4`. In this case this means duplication of an existing app file. For legacy files we'd similarly use `/r|ro|v|s/kZTr6Zdx6TClE6I4qExV4` to use that existing legacy file as a starting point.
* Room DO then sees this column and reads the data from the appropriate source and initializes the room.
* We now retry getting the file record from the db, because otherwise we could still end up with a "Not found" issue since the navigation might happen before the replicator tells the room DO about the file creation.
* Assets will get re-uploaded with our existing `maybeAssociateFileAssets` logic, that makes sure all the assets are associated with the file.

We had to do this because in some cases the file got created in the db, then an file update was dispatched to the replicator (due to the trigger we have on the file table), which incorrectly the set room DO (its document info was set to a non app file without the creation params). If this happened before the user navigation hit the worker it would mean that duplication as well as slurping legacy files end up in a "Not found" error, even though the file was correctly inserted into the db.

### Change type
- [x] `improvement`

### Release notes

- Move duplicating and legacy file slurping to the server.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
@@ -1,7 +1,7 @@
-import { TlaFile, TlaFileOpenState } from '@tldraw/dotcom-shared'
+import { TlaFile } from '@tldraw/dotcom-shared'
 import classNames from 'classnames'
 import { KeyboardEvent, MouseEvent, useCallback, useEffect, useRef, useState } from 'react'
-import { Link, useLocation, useNavigate, useParams } from 'react-router-dom'
+import { Link, useNavigate, useParams } from 'react-router-dom'
 import { preventDefault, useContainer, useValue } from 'tldraw'
 import { routes } from '../../../../routeDefs'
 import { useApp } from '../../../hooks/useAppState'
@@ -81,7 +81,6 @@ export function TlaSidebarFileLinkInner({
 	const linkRef = useRef<HTMLAnchorElement | null>(null)
 	const app = useApp()
 	const intl = useIntl()
-	const state = useLocation().state as TlaFileOpenState | undefined
 	const focusCtx = useFileSidebarFocusContext()
 
 	const [isRenaming, setIsRenaming] = useState(debugIsRenaming)
@@ -98,11 +97,7 @@ export function TlaSidebarFileLinkInner({
 
 	useEffect(() => {
 		// on mount, trigger rename action if this is a new file.
-		if (
-			(state?.mode === 'create' || state?.mode === 'duplicate') &&
-			isActive &&
-			focusCtx.shouldRenameNextNewFile
-		) {
+		if (isActive && focusCtx.shouldRenameNextNewFile) {
 			focusCtx.shouldRenameNextNewFile = false
 			handleRenameAction()
 		}
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -245,8 +245,7 @@ export class TLDrawDurableObject extends DurableObject {
 			await getSlug(this.env, req.params.roomId, roomOpenMode),
 			'roomId must be present'
 		)
-		const url = new URL(req.url)
-		const isApp = url.pathname.startsWith('/app/')
+		const isApp = new URL(req.url).pathname.startsWith('/app/')
 
 		if (this._documentInfo) {
 			assert(this._documentInfo.slug === slug, 'slug must match')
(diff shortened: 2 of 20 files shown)


apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarFileLink.tsx
```diff
@@ -98,11 +97,7 @@ export function TlaSidebarFileLinkInner({
 
 	useEffect(() => {
 		// on mount, trigger rename action if this is a new file.
-		if (
-			(state?.mode === 'create' || state?.mode === 'duplicate') &&
-			isActive &&
-			focusCtx.shouldRenameNextNewFile
-		) {
+		if (isActive && focusCtx.shouldRenameNextNewFile) {
```
MitjaBezensek, line 100: Couldn't find any issues with omitting the mode in this check, but maybe you can double check?

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -398,14 +401,7 @@ export class TLDrawDurableObject extends DurableObject {
 						openMode = ROOM_OPEN_MODE.READ_ONLY
 					}
 				}
-			} else if (!this.documentInfo.appMode) {
```
MitjaBezensek, line 401: Feels like we don't need this any longer? If we didn't go through the proper file creation flow then the file won't exists in the db and then loading the file record will fail. Tried it by logging out then going to `/f/random-id` and I was correctly shown the `Not found` page.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -687,6 +705,18 @@ export class TLDrawDurableObject extends DurableObject {
...
+	async appFileRecordCreated(file: TlaFile) {
+		this._fileRecordCache = file
+
+		this.setDocumentInfo({
+			version: CURRENT_DOCUMENT_INFO_VERSION,
+			slug: file.id,
+			isApp: true,
+			deleted: false,
+		})
+		await this.getRoom()
```
MitjaBezensek, line 717: This bootstraps the room creation process.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -316,16 +307,28 @@ export class TLDrawDurableObject extends DurableObject {
...
-			return this._fileRecordCache
-		}
 		try {
-			this._fileRecordCache = await this.db
-				.selectFrom('file')
-				.where('id', '=', this.documentInfo.slug)
-				.selectAll()
-				.executeTakeFirstOrThrow()
-			return this._fileRecordCache
+			return await retry(
```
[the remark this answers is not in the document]
MitjaBezensek: Yeah, feels too slow for me too. Will bring back the create mode and see if there's something we can do for duplicating. That one also feels it should be pretty much immediate. 

I'm ok with slurping legacy stuff taking a slightly longer. File not found also feels quite slow (since it has to wait for all the attempts to fail). That's a less often scenario so I'm probably fine with that.

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -441,16 +440,17 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 			.map((x) => x.userId as string)
 		// if the file state was deleted before the file, we might not have any impacted users
 		if (event.command === 'delete') {
-			this.getStubForFile(row.id).appFileRecordDidDelete()
+			getRoomDurableObject(this.env, row.id).appFileRecordDidDelete()
```
MitjaBezensek, line 443: Use a general helper for this.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -319,12 +319,16 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
-							await tx
+							const { id, ...rest } = update.row as any
+							const result = await tx
 								.insertInto(update.table)
 								.values(update.row as any)
 								.onConflict((oc) => oc.column('id').doUpdateSet(rest))
+								.returningAll()
 								.execute()
+							if (update.table === 'file' && result.length > 0) {
+								getRoomDurableObject(this.env, id).appFileRecordCreated(result[0] as any as TlaFile)
```
MitjaBezensek, line 330: When a file gets created we now immediately call into the room DO to get it ready as soon as possible.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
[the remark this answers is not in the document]
MitjaBezensek: How does it feel now? I added some logic to the user durable object: when file insertion to the database succeeds we immediately call into the room DO to get it ready faster. We don't need to wait for the replicator to get notified, so it does seem to be faster.

ds300 (approved):
nice, much faster now 🙏🏼

PR #5349, opened 2025-02-04
Shape options

steveruizok:
This PR moves shape-specific options into shape utils as static objects.

I'd initially added shape options to the `options` prop, however like many other options, they aren't really options to the editor.

By comparison, adding static options to shape utils is less discoverable but easier to scale and document.

The only negative that I can think of, apart from discoverability, is that static changes like this effect all editors. This probably doesn't matter: I can't imagine a single app with two editors, one of which allows notes to resize and the other which does not.

```tsx
import { NoteShapeUtil, Tldraw } from 'tldraw'
import 'tldraw/tldraw.css'

NoteShapeUtil.options.resizeMode = 'scale'

export default function ResizeNoteExample() {
	return (
		<>
				<Tldraw persistenceKey="resize-note"></Tldraw>
			</div>
		</>
	)
}
```

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release Notes
- introduces shape options
- moves (unreleased) `noteShapeResizeMode` to `NoteShapeOptions.resizeMode`
- moves `maxDrawShapePoints` to `DrawShapeOptions.maxPoints`
- adds `maxPoints` to `HighlightShapeOptions.maxPoints`
- 💥 breaking change if someone was using `options.maxDrawShapePoints`.


packages/editor/src/lib/options.ts
```diff
@@ -67,10 +66,20 @@ export interface TldrawOptions {
...
+			maxPoints: number
+		}
+		note: {
+			/**
+			 * How should the note shape resize? By default it does not resize (except automatically based on its text content),
+			 * but you can set it to be user-resizable using scale.
+			 */
+			resizeMode: 'none' | 'scale'
+		}
+	}
```
SomeHats, line 82: the docs for this won't show up properly. For  the doc comments to show, you need to use a separate interface for each of these objects

SomeHats (commented):
personally not into this one - i think it just moves things around, but not in a way that really adds any clarity or helps separate things out at all? it also i think makes the generated docs for options a fair bit worse. I'm all for a common prefix within the options object, but unless we were doing like a more generalized version of "shape options" i don't think it's worth these being their own objects

steveruizok:
> personally not into this one - i think it just moves things around, but not in a way that really adds any clarity or helps separate things out at all? it also i think makes the generated docs for options a fair bit worse. I'm all for a common prefix within the options object, but unless we were doing like a more generalized version of "shape options" i don't think it's worth these being their own objects

We could do a separate `shapeOptions` prop, with the idea that these will be delivered to the specific shapes, though I don't think we could generate types for these automatically. I don't think that having a bunch of `noteShapeResizeMode`, `noteShapeDanceHeight`, `noteShapeHaircut` options is a good way to do it though.

mimecuvalo:
I think it might be more cognitively easier to have it under just `options` as this PR proposes. 
We already have:
- `options`
- `textOptions`
- `cameraOptions`
- we briefly had for a while `assetOptions`

Feels messy as far as DX is concerned and not great for discoverability to add yet another, i.e. `shapeOptions`
(I have to know as a developer that things aren't prefixed with `options` and that some aspects of tldraw have options and some things don't).

For consistency, @SomeHats isn't wrong, that probably `shapeOptions` fits with our existing pattern. But ugh, it feels like it's getting out of hand.

steveruizok:
That's a good point, we do have a lot of named options props.

steveruizok:
Wait what are text options?

steveruizok:
New approach here that I'm generally happier with. It does break the "lets put all the options in one spot" DX, though it improves the "configuration is close to the thing that is being configured" story.

packages/tldraw/src/test/resizing.test.ts
```diff
@@ -3563,6 +3563,9 @@ describe('editor.resizeNoteShape', () => {
 		expect(editor.getShape(noteBId)).toMatchObject({ x: 0, y: 0, props: { scale: 2.1 } }) // but scaled!
 
 		expect(editor.getShapePageBounds(noteBId)).toMatchObject({ x: 0, y: 0, w: 420, h: 420 })
+
+		// for the sake of future tests, set it back to normal
+		NoteShapeUtil.options.resizeMode = 'none'
```
SomeHats, line 3568: this might cause errors in other tests if this test fails. can we use a `berfore`/`after` instead?

SomeHats (commented):
the leaking between instances issue is a bit of a shame, but i do prefer this overall i think

steveruizok:
Ok, landing this one.

It will be a breaking change if someone was using `options.maxDrawShapePoints`.

PR #5350, opened 2025-02-04
Fix tldr file upload

ds300:
This PR uploads the assets separately from the room data json blobs, to avoid OOM exceptions on the server. I also consolidated the upload lifecycle management.

Unfortunately, the way it's currently set up, the images get duplicated on our backend because there's no way to assign the file ownership in postgres without the file row already existing, but making that happen with good UX is tricky. I looked into resolving it a number of ways but it was getting a bit yak shavey so I decided to ignore the duplication issue for now.

I think the ideal UX would be:

1. as soon as you drop the .tldr file onto the canvas we extract all the assets into indexeddb, upload the records to R2, and create a file row.
3. we gradually upload the assets in the background as soon as we know the file rows are in the db.
4. the files appear in the sidebar immediately, with loading spinners to indicate that they are still uploading. You can open them up and make edits and stuff while the uploading happens in the background

which is what I tried to implement to begin with, but it was getting too heavy and I want to move on so I fell back to this.

### Change type

- [x] `bugfix`


apps/dotcom/client/public/tla/locales-compiled/en.json
```diff
@@ -14,7 +14,6 @@
   "1da2e8106b": "An unexpected error occurred.",
   "1dd1c5fb7f": "Today",
   "1f9696d5f2": "Export the content as a .tldr file: Go to 'File' > 'Save a copy'",
-  "20530afec1": "{count, plural, one {Uploading {count} .tldr file…} other {Uploading {count} .tldr files…}}",
```
MitjaBezensek, line 17: Looks like this does not get removed from all other localizations. I forgot if we ever clean not used localizations from other languages? @mimecuvalo

apps/dotcom/client/src/tla/hooks/useTldrFileDrop.ts
```diff
@@ -45,39 +29,11 @@ export function useTldrFileDrop() {
...
-					keepOpen: true,
-				})
-				if (results.slugs.length > 0) {
-					navigate(routes.tlaFile(results.slugs[0]), {
-						state: { mode: 'create' } as TlaFileOpenState,
-					})
-				}
-			}
+			app.uploadTldrFiles(tldrawFiles, (file) => {
+				navigate(routes.tlaFile(file.id), { state: { mode: 'create' } satisfies TlaFileOpenState })
```
MitjaBezensek, line 33: My PR removes the `TlaFileOpenState` completely. Think just removing the state should work here, but let me know.

MitjaBezensek (approved):
Nice, love the progress toast, we should do something like that for our other long lasting actions.

When I drag a file to the sidebar we show the toast immediately and then it is immediately dismissed when we navigate to the file. This happened with a single small file.

https://github.com/user-attachments/assets/9786b2a8-7088-406c-a911-382962245964

apps/dotcom/client/public/tla/locales-compiled/en.json
mimecuvalo: Hey, yeah, they won't get proactively removed from other languages until the translations update in general (which is a weekly process). I could maybe do this in advance but it becomes tricky and not super worth it.

PR #5351, opened 2025-02-05
make websocket bootstrap time much faster

ds300:
This PR makes it return the user data early

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/UserDataSyncer.ts
+++ b/apps/dotcom/sync-worker/src/UserDataSyncer.ts
@@ -378,6 +378,24 @@ export class UserDataSyncer {
 		// ignore other events until we get the boot id
 	}
 
+	async waitUntilInitialData() {
+		while (
+			this.state.type !== 'connected' &&
+			!(this.state.type === 'connecting' && this.state.data)
+		) {
+			await sleep(50)
+		}
+	}
+
+	getInitialData() {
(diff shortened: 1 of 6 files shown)


apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -186,6 +186,12 @@ export class UserDataSyncer {
 
 	private updateStateAfterBootStep() {
 		assert(this.state.type === 'connecting', 'state should be connecting')
+		if (this.state.data) {
```
MitjaBezensek, line 189: We should remove the broadcast inside the if statement below so that we don't send twice.

MitjaBezensek (approved):
Looks good, left a comment about removing a broadcast that is no longer needed.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
ds300: if there are buffered events it will need to broadcast again. we could check whether or not it's necessary though

MitjaBezensek: We still send the state data there and it doesn't look like those buffered events ever make it there?

ds300: ah another good catch, fixed!

PR #5357, opened 2025-02-05
Improve frame heading perf

steveruizok:
This PR prevents rapid re-measurement of frame headings while zooming.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a lot of frame shapes
2. Zoom in and out

### Release notes

- Fix bug effecting performance when many frames are on the screen.

### DIFF
--- a/packages/tldraw/src/lib/shapes/frame/frameHelpers.ts
+++ b/packages/tldraw/src/lib/shapes/frame/frameHelpers.ts
@@ -18,7 +18,9 @@ export function getFrameHeadingSide(editor: Editor, shape: TLFrameShape): 0 | 1 
 
 /**
  * We use a weak map here to prevent re-measuring the text width of frames that haven't changed their names.
- * It's only really important for performance reasons while zooming in and out.
+ * It's only really important for performance reasons while zooming in and out. The measured text size is
+ * independent of the zoom level, so we can cache the expensive part (measurement) and apply those changes
+ * using the zoom level.
  */
 const measurementWeakmap = new WeakMap()
 
--- a/packages/tldraw/src/lib/shapes/frame/frameHelpers.ts
+++ b/packages/tldraw/src/lib/shapes/frame/frameHelpers.ts
@@ -17,6 +17,12 @@ export function getFrameHeadingSide(editor: Editor, shape: TLFrameShape): 0 | 1 
 }
 
 /**
+ * We use a weak map here to prevent re-measuring the text width of frames that haven't changed their names.
+ * It's only really important for performance reasons while zooming in and out.
+ */
+const measurementWeakmap = new WeakMap()
+
+/**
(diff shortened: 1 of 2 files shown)


MitjaBezensek (approved):
👍

packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -108,10 +105,19 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
...
 			if (didAlreadyResolve.current) {
-				resolveAssetUrlDebounced(editor, assetId, screenScale, exportInfo, (url) =>
-					resolve(asset, url)
-				)
-				cancelDebounceFn = resolveAssetUrlDebounced.cancel // cancel the debounce when the hook unmounts
+				let tick = 0
+
+				const resolveAssetAfterAWhile = () => {
+					tick++
+					if (tick > 32) {
```
mimecuvalo, line 112: why this change along with this PR? this makes the loading of the resized image asset delayed by about ~500ms when zooming is finished. is that the intent?

steveruizok: On the torture test that Mitja sent to me, the debounce was actually a performance hit (setting and removing the global timers). We could pull that from this PR and look at it separately, but my thinking here is that we have a tick going anyway, so we could use that.

mimecuvalo: (my bad, btw, the original one also does 500ms :doh:)

packages/tldraw/src/lib/shapes/frame/frameHelpers.ts
```diff
@@ -25,32 +33,28 @@ export function getFrameHeadingSide(editor: Editor, shape: TLFrameShape): 0 | 1
...
-			spans: [],
-		}
+		return new Box(0, -opts.height, shape.props.w, opts.height)
 	}
 
-	const spans = editor.textMeasure.measureTextSpans(
-		defaultEmptyAs(shape.props.name, 'Frame') + String.fromCharCode(8203),
-		opts
-	)
+	const frameTitle = defaultEmptyAs(shape.props.name, 'Frame') + String.fromCharCode(8203)
```
trygve-aaberge-adsk, line 46: Any reason this isn't in the if below? I guess running a regex every time this is called isn't super expensive, but seems like there's no reason to run it when the measurement is already cached.

MitjaBezensek: Nice catch, created a PR [here](https://github.com/tldraw/tldraw/pull/5413).

PR #5363, opened 2025-02-06
[proof-of-concept] contextual toolbar: mobile variant

mimecuvalo:
variant where the toolbar goes across the virtual keyboard to make it feel more integrated with the keyboard (h/t to Alex for the suggestion)

Screenshot_20250206-085426

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -2051,6 +2051,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	setEditingShape(shape: TLShapeId | TLShape | null): this {
 		const id = typeof shape === 'string' ? shape : (shape?.id ?? null)
+		this.setEditingShapeTipTapTextEditor(null)
 		if (id !== this.getEditingShapeId()) {
 			if (id) {
 				const shape = this.getShape(id)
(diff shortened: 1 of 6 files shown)


steveruizok:
Try testing this on iOS mobile. This one has the same issues that we talked about with the previous implementation: the toolbar stays active after the keyboard closes; the toolbar's position can change when moving between editing shapes; the toolbar can cover up text. If we can make this work then great, otherwise we need to decide between disabling the toolbar on the platforms where we can't guarantee it, or else skip the feature entirely on mobile.

https://github.com/user-attachments/assets/2c3e23e7-fe2e-4043-93f7-cdce17674250

mimecuvalo:
fixed the issues mentioned! should have said this variant wasn't fully baked, these issues were fixable.

this was a proof-of-concept to show during our Writer's Room to look at (hence no reviewers were assigned - i generally assign reviewers once it's ready to look at - but i'll add a [proof-of-concept/experiment] prefix next time to the title)

mimecuvalo:
i reverted the "disable precise editing on mobile" piece because it causes issue with newly created items: you can't pre-select an item in the toolbar and have that style be applied. would have to look for a workaround for that.

PR #5365, opened 2025-02-06
Commit the copy changes.

MitjaBezensek:
Looks like the changes from https://github.com/tldraw/tldraw/pull/5364 didn't make it into the localization files.

### Change type

- [x] `bugfix`


ds300 (approved):
nice good catch, added a linear ticket to prevent this https://linear.app/tldraw/issue/INT-723/make-sure-enjson-is-up-to-date-on-ci

PR #5369, opened 2025-02-06
Fix dancing icons.

MitjaBezensek:
Looks like this happened only on non-retina screens and was caused by subpixel rendering.

Resolves INT-720

### Before

https://github.com/user-attachments/assets/1018c07d-4f6e-4af1-ab2c-2b7ddd156e51

### After

https://github.com/user-attachments/assets/332316ec-b0b9-4b5f-afab-67658e2deb23

### Change type

- [x] `improvement`

### Release notes

- Fix the dancing icons when the sidebar width change.


steveruizok:
feature not bug

steveruizok (approved):

MitjaBezensek:
giphy

PR #5371, opened 2025-02-06
Improve published file experience

ds300:
This is by no means a perfect solution, but it's a lot better than the current situation for logged in users, which is bad mainly because it shows the sidebar. Showing the sidebar for published files:

- fails to give the file owner a sense of how other people will see the published version
- prevents the viewer from really _feeling_ how published files are different
- puts the sidebar in a confusing navigational state where nothing is selected

With this PR

1. Logged in users see the 'anonymous' view (no sidebar) but with a share button instead of the sign in button.
2. Logged in users also get a 'copy to my files' affordance in the three dots menu.
3. We replace the 'Save a copy' menu item (which opens the file picker) with a 'Download file' menu item (which downloads straight to the downloads folder).
4. Clicking the tldraw logo takes you back to the main app view

I think there's still a lot of room for improvement here, but this feels like a quick win to land ahead of launch

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


MitjaBezensek:
Should we do something similar for old legacy files? Feels like the use case is kinda the same there. Might make it a bit easier for users to differentiate between:
1. `My files`: These are the files that I can always get back to via the sidebar.
2. `Not my files`: These are files that I can only access via a link, but I can pull them into `My files`.

It would also allow us to unify the `Copy to my files` experience.

ds300:
yeah i agree, i'll add that here and we can see how it feels

ds300:
Didn't like where this was going with the legacy routes so decided to bail out. We can come back to that afterwards.

Gonna merge this as-is unless there's any objections @steveruizok ?

PR #5378, opened 2025-02-07
Numeric shortcuts were still getting triggered when used inside some inputs (like the file rename input)

MitjaBezensek:
We used `hotkeys` library [in the past t](https://github.com/tldraw/tldraw/pull/5340)o add the shortcuts for numeric tool shortcut keys. Looks like it handled the editable element filtering (not triggering keyboard shortcuts when the source was an editable element). Since we no longer use that we should prevent the shortcuts manually.

### Change type

- [x] `bugfix`

### Test plan

1. Edit the file name in the sidebar.
2. Pressing 1, 2, 3 should no longer change the tool and should add the pressed key to the name input.

### Release notes

- Fix an issue with numeric shortcuts working inside of editable elements.

### DIFF
--- a/packages/editor/src/lib/hooks/useDocumentEvents.ts
+++ b/packages/editor/src/lib/hooks/useDocumentEvents.ts
@@ -2,7 +2,7 @@ import { useValue } from '@tldraw/state-react'
 import { useEffect } from 'react'
 import { Editor } from '../editor/Editor'
 import { TLKeyboardEventInfo } from '../editor/types/event-types'
-import { preventDefault, stopEventPropagation } from '../utils/dom'
+import { activeElementShouldCaptureKeys, preventDefault, stopEventPropagation } from '../utils/dom'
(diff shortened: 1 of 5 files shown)


packages/editor/src/lib/hooks/useDocumentEvents.ts
```diff
@@ -274,15 +274,6 @@ export function useDocumentEvents() {
 	}, [editor, container, isAppFocused])
 }
 
-const INPUTS = ['input', 'select', 'button', 'textarea']
-
 function areShortcutsDisabled(editor: Editor) {
```
MitjaBezensek, line 277: We also seem to have `areShortcutsDisabled` function duplicated a couple of times, could maybe be unified.

ds300 (approved):
nice. not sure if we need to make `activeElementShouldCaptureKeys` public? but it's probably fine.

PR #5383, opened 2025-02-07
Make sure compiled localizations are up to date.

MitjaBezensek:
Make sure our compiled localizations are up to date.

Pre-commit [should update the them](https://github.com/tldraw/tldraw/blob/be65d368fd8b3aaa31b3d6a41b3f563f3312b379/.husky/pre-commit#L7), but just to be safe we'll double check in CI.

Example of a failed run [here](https://github.com/tldraw/tldraw/actions/runs/13200772367/job/36852008480?pr=5383#step:13:23) (and the [commit](https://github.com/tldraw/tldraw/pull/5383/commits/2608484e9489b7cc4d98162f0effb8a7fbd0c185) that caused it).

### Change type

- [x] `improvement`

### Release notes

- Make sure our compiled localizations are up to date.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -58,6 +58,11 @@ jobs:
       - name: Check API declarations and docs work as intended
         run: yarn api-check
 
+      - name: Check dotcom localizations
+        run: |
+          yarn build-i18n 
+          git diff --exit-code --quiet || (echo "Please run yarn build-i18n and commit the changes" && exit 1)
+
       - name: Test
         run: yarn test-ci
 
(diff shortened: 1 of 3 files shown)


.github/workflows/checks.yml
```diff
@@ -58,6 +58,11 @@ jobs:
       - name: Check API declarations and docs work as intended
         run: yarn api-check
 
+      - name: Check dotcom localizations
+        run: |
+          yarn build-i18n 
+          git diff --exit-code --quiet || (echo "Please run yarn build-i18n and commit the changes" && exit 1)
```
ds300, line 64: can we scope the diff to the locales-compiled folder? just to avoid potential false positives if we introduce any tooling that modifies the working tree later

ds300 (approved):

mimecuvalo:
Ah, nice - thx for doing this! Yeah, maybe there was some skipped pre-commit hooks or something.

PR #5384, opened 2025-02-07
add missing socket upgrade check

ds300:
- [x] `other`


MitjaBezensek:
Was just about to do the same 😂 👍

PR #5385, opened 2025-02-07
yeet legacy pages into sun

ds300:
# 🔪 

This PR gets rid of all the old versions of the legacy routes like /r/whatever and such.
The new app-enhanced versions still exist obvs.

### Change type

- [x] `other`


steveruizok:
rip

MitjaBezensek (commented):
Nice 👍 

Readonly rooms are missing qr code
CleanShot 2025-02-10 at 15 32 43@2x

PR #5400, opened 2025-02-10
Fix file permissions

MitjaBezensek:
Fixes:
* File permissions for non owners.
* File state permissions.
* An issue with the published file logic.

### Change type

- [x] `bugfix`

### Test plan

1. I tested this via `app.updateFile` since we don't expose UI elements for this.

### Release notes

- Fix a bug with changing files.

### DIFF
--- a/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
@@ -5,6 +5,7 @@ import {
 	TlaFile,
 	TlaFilePartial,
 	TlaFileState,
+	TlaFileStatePartial,
 	TlaUser,
 	Z_PROTOCOL_VERSION,
 	ZClientSentMessage,
@@ -292,7 +293,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 				)
 			}
 			case 'file_state': {
-				const nextFileState = update.row as TlaFileState
+				const nextFileState = update.row as TlaFileStatePartial
 				let file = s.files.find((f) => f.id === nextFileState.fileId)
 				if (!file) {
 					// The user might not have access to this file yet, because they just followed a link
@@ -303,11 +304,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 					throw new ZMutationError(ZErrorCode.bad_request, `File not found ${nextFileState.fileId}`)
 				}
 				if (file.ownerId === this.userId) return
-				if (
-					file.shared &&
-					(nextFileState.userId === this.userId || nextFileState.userId === undefined)
-				)
-					return
+				if (file.shared && nextFileState.userId === this.userId) return
 
 				throw new ZMutationError(
 					ZErrorCode.forbidden,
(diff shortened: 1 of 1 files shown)


apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -272,8 +273,18 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 						`Cannot create a file for another user ${nextFile.id}`
 					)
 				}
+				// Owners are allowed to make changes
 				if (prevFile.ownerId === this.userId) return
-				if (prevFile.shared && prevFile.sharedLinkType === 'edit') return
+
+				// We can make changes to updatedAt field in a shared, editable file
+				if (prevFile.shared && prevFile.sharedLinkType === 'edit') {
```
MitjaBezensek, line 280: We used to just allow everything if file was shared in edit mode (even for non owners). We know only allow changing `updateAt` field.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -360,7 +371,11 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 							await tx.updateTable(update.table).set(updates).where('id', '=', id).execute()
 							if (update.table === 'file') {
 								const currentFile = this.cache.store.getFullData()?.files.find((f) => f.id === id)
-								if (currentFile && currentFile.published !== rest.published) {
```
MitjaBezensek, line 363: This is another subtle error. We only relied on the published field to be different. But `rest` is a partial, so it might not even have the `published` field set. So we also need to check that we did receive some value here. Similar for the `lastPublished`, otherwise we'd compare a number to undefined.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -292,7 +303,11 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 					throw new ZMutationError(ZErrorCode.bad_request, `File not found ${nextFileState.fileId}`)
 				}
 				if (file.ownerId === this.userId) return
-				if (file.shared) return
+				if (
+					file.shared &&
+					(nextFileState.userId === this.userId || nextFileState.userId === undefined)
```
MitjaBezensek, line 308: This also feels like a bug. We should not allow users to change file state for other users.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -292,7 +304,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 					throw new ZMutationError(ZErrorCode.bad_request, `File not found ${nextFileState.fileId}`)
 				}
 				if (file.ownerId === this.userId) return
-				if (file.shared) return
+				if (file.shared && nextFileState.userId === this.userId) return
```
MitjaBezensek, line 307: We should not allow users to set file state for other users.

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -291,6 +303,12 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 				if (!file) {
 					throw new ZMutationError(ZErrorCode.bad_request, `File not found ${nextFileState.fileId}`)
 				}
+				if (nextFileState.userId !== this.userId) {
```
MitjaBezensek, line 306: Users should only be allowed to set their own file state right?

ds300 (approved):
great stuff, thanks

PR #5401, opened 2025-02-10
a11y: focus ring

mimecuvalo:
mini-side quest: bring a focus ring, when not mousing around, only when using the Tab key

https://github.com/user-attachments/assets/5280f40b-8460-476e-b2bc-7080809b31aa

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- a11y: enable focus ring.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
@@ -65,6 +65,9 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
 			#${id}_more > *:nth-child(-n + ${overflowIndex}) {
 				display: none;
 			}
+			#${id}_more > *:nth-child(-n + ${overflowIndex + 4}) {
+				margin-top: 0;
+			}
         `
 	}, [lastActiveOverflowItem, id, overflowIndex])
 
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -172,6 +172,7 @@
 		inset 0px 0px 0px 1px var(--color-panel-contrast);
 	--shadow-4: 0px 0px 3px hsl(0, 0%, 0%, 19%), 0px 5px 4px hsl(0, 0%, 0%, 16%),
 		0px 2px 16px hsl(0, 0%, 0%, 6%), inset 0px 0px 0px 1px var(--color-panel-contrast);
+	--shadow-focus-ring: inset 0 0 0 3px hsl(213, 100%, 68%);
 }
 
 .tl-theme__dark {
--- a/packages/editor/src/lib/editor/managers/FocusManager.ts
+++ b/packages/editor/src/lib/editor/managers/FocusManager.ts
@@ -30,6 +30,10 @@ export class FocusManager {
 			editor.updateInstanceState({ isFocused: !!autoFocus })
 		}
 		this.updateContainerClass()
+
+		const container = this.editor.getContainer()
+		container.addEventListener('keydown', this.handleKeyDown.bind(this))
+		container.addEventListener('mouseover', this.handleMouseOver.bind(this))
 	}
 
 	/**
@@ -52,6 +56,18 @@ export class FocusManager {
 		}
 	}
 
+	private handleKeyDown(keyEvent: KeyboardEvent) {
+		const container = this.editor.getContainer()
+		if (keyEvent.key === 'Tab') {
+			container.classList.remove('tl-container__no-focus-ring')
+		}
+	}
+
+	private handleMouseOver() {
+		const container = this.editor.getContainer()
+		container.classList.add('tl-container__no-focus-ring')
+	}
+
 	focus() {
 		this.editor.getContainer().focus()
 	}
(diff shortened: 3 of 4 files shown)


SomeHats (commented):
love this! but can you discuss (maybe a comment) why we need a special conditional classname to handle this? This seems like the exact use-case for `:focus-visible`

mimecuvalo:
> why we need a special conditional classname to handle this? This seems like the exact use-case for :focus-visible

ah yes, good call. in general what it's trying to achieve is only showing this added aide when using the keyboard, but not via the mouse.

packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
```diff
@@ -65,6 +65,9 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
 			#${id}_more > *:nth-child(-n + ${overflowIndex}) {
 				display: none;
 			}
+			#${id}_more > *:nth-child(-n + ${overflowIndex + 4}) {
```
mimecuvalo, line 68: this was currently broken / just more noticeable with the focus ring

steveruizok:
Quick QA here:

- on the examples site, the rings are shown by default prior to any keyboard interaction. This doesn't appear when full screened
- not shown when keyboard navigating with arrow keys (unless already visible)
- the toolbar overflow menu is a "trap", no way to tab out of it or within it
- toolbar menu is not accessible (it's not in the DOM until hovered, stretch goal since I want us to improve this anyway)

I think this is a good move overall. I might visually change the outline style before we ship it.

Some more context related to dropdown menus: Radix UI (which we use for dropdown menus) has built-in focus states. We were using these before, however the system has a bug when hovering over disabled items. This may or may not have been fixed since I disabled them. The bug would cause focus rings to appear in the parent menu when hovering the disabled items in the child menu.

steveruizok:
> it's a menu so you have to hit escape. (this is true of other menus too with radix) if we wanted to change this behavior we should do it outside of this PR.

Ah ok, all good then.
 
> maybe i'm not understanding this one, which menu did you mean here?

Here's a better description:
- In the "overflowing tools" submenu, the tab key has no action. You can navigate with the arrow keys though.
- In the "quick actions menu" the opposite is true: you can navigate within the menu using tab but cannot use the arrow keys. :psyduck:

mimecuvalo:
Gotcha, makes sense. Ok, this might be expanding the scope of this PR but I converted those places to be popovers now instead of menus!

steveruizok:
> Gotcha, makes sense. Ok, this might be expanding the scope of this PR but I converted those places to be popovers now instead of menus!

I think that's the right move. Good fix!

steveruizok:
Pushed up a few small improvements to the menus so that they display the hover state on highlight rather than hover. This makes them display correctly when using the keyboard to navigate.

steveruizok:
Did a variant that uses the ::after instead of the focus ring. I tend to like this better because it doesn't push against the styling so much, and we already use the after to indicate hovers / activity states.

Screenshot 2025-03-01 at 6 35 01 PM

image

image

steveruizok:
Let's see if we can solve these. The only one there that really bugs me is the expando-arrow contrast, though that can be fixed with CSS by making the focus color text/opacity 1. The blue-on-blue is annoying but is similarly okay-ish given that it's not a permanent state (ie you'd be tabbing to it, so you'd have experienced it without the background first). Adding a panel-colored outline would work too, though it might take some hackiness given that the icons are part of our SVG sheet.

We could try a black bg instead for the focus indicator, which I think would be better contrast to the toolbar's selection state. It would have the same issue as the blue-on-blue but for black.

mimecuvalo:
Do you mean black background when an item is selected, but blue otherwise? Or just black in general? Wasn't sure what you meant there.

I agree on the color palette being less of a big deal. I think the bigger deals are:
- not being able to see selected state
- expando-arrow

I mean, I dunno, feels like this might be bucking the trend here for a11y conventions to not do the ring.

One method to move forward is to ship with the ring, and if you wanna iterate/explore on a bg version on a separate PR that might work?

steveruizok:
Black background is not used anywhere else in the app, so it's unique and provides strong contrast (even with the selected tool). Works clearly on light and dark mode.

image

The focus hack: we do an invert on the focus for the black / white colors.
 
image

Sets the color to full opacity now when focused:

image

steveruizok:
Disappointing that radix doesn't support tab in dropdown menus, though the arrows still work as navigation.

steveruizok:
fixed this guy

Kapture 2025-03-07 at 16 44 20

steveruizok:
> I don't wanna be a party pooper about the bg exploration, but I think it's pushing against a11y conventions in a non-intuitive way. Especially since this is used within other applications downstream that would use more standard conventions. Furthermore, it's painting us into a corner for future UI that won't be able to use that same UI treatment.

I'm convinced! Let's stick with the same dimensions as the hover shape though: it doesn't get cut off by the edges of scrollable containers and generally fits better with the designs. Pushing that up now.

steveruizok:
That looks like this btw

image

Good call regarding downstream users. Even our own apps (dotcom) will need to centralize on focus styles.

steveruizok:
Also including a fix for this:

image

mimecuvalo:
Let's goooooo 🚀

PR #5402, opened 2025-02-10
Add tldraw and excalidraw to external content types

SomeHats:
Almost all paste operations currently turn into calls to `putExternalContent`, which means they can be customized using `registerExternalContentHandler`. tldraw and excalidraw content are different: they're handled directly in the clipboard code, and so can't be customized. 

This diff changes that: these types now have external content handlers, and default implementations for how they work. We also add an example demonstrating how you can use this API to override pasting of e.g. single frame shapes to match how figma handles these (finding a space for them instead of pasting them in place).

### Change type

- [x] `api`

### Release notes

- You can now customize how pasted tldraw and excalidraw content is handled with `registerExternalContentHandler`.


apps/examples/src/index.tsx
```diff
@@ -77,7 +77,7 @@ const router = createBrowserRouter([
 					return {
 						element: (
 							<NoIndex>
-								<ExampleWrapper example={example} component={Component} />,
```
mimecuvalo, line 80: lol

apps/examples/src/examples/custom-paste/CustomPasteExample.tsx
```diff
@@ -0,0 +1,87 @@
...
+const SPACING_BETWEEN_FRAMES = 50
+
+function handleCustomTldrawPaste(editor: Editor, { content, point }: TLTldrawExternalContent) {
+	// find the only shape in the pasted content
+	const onlyCopiedShape =
+		content.rootShapeIds.length === 1
+			? content.shapes.find((shape) => shape.id === content.rootShapeIds[0])
+			: null
+
+	// make sure that the shame is a frame. if it is, retrieve the current version of that frame
```
mimecuvalo, line 34: SHAME :P

mimecuvalo (approved):
👍 approved

something to consider for future architecture here:
I think about Miro, FigJam, OCIF (open canvas), and any other sources that might come to our canvas. Questions that arise in my head:
- do we end up having just a long list of external content handlers long-term?
- if I understand the waterfall correctly now, things fallback to the `text` if it's a format we don't recognize (e.g. `ExternalContentSourcesExample.tsx`) but that doesn't feel _great_. Feels like we might consider a more general, natural pasting hook into tldraw.

PR #5407, opened 2025-02-11
Add is_signed_in property to posthog events

ds300:
Adds an is_signed_in property to posthog events that we can use to differentiate pageviews.

### Change type

- [x] `other`


SomeHats (commented):
nice!

PR #5410, opened 2025-02-11
hide watermark in focus mode

ds300:
felt off for everything to disappear except for the watermark.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/styles/tla.css
+++ b/apps/dotcom/client/src/tla/styles/tla.css
@@ -98,6 +98,10 @@
 		inset 0px 0px 0px 1px var(--tla-color-panel);
 }
 
+.tla-focus-mode .tl-watermark_SEE-LICENSE {
+	visibility: hidden;
+}
+
 .tla .tl-container {
 	font-family: var(--tla-font-ui);
 	font-size: 12px;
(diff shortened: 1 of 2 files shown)


apps/dotcom/client/src/tla/styles/tla.css
```diff
@@ -98,6 +98,10 @@
 		inset 0px 0px 0px 1px var(--tla-color-panel);
 }
 
+.tla-focus-mode .tl-watermark_SEE-LICENSE {
+	visibility: hidden;
```
mimecuvalo, line 102: i'll let Steve weigh in on whether we should do this or not, but little nit here would be that `display: none` would feel slightly more correct.

steveruizok:
Sadly the demands of capital make this impossible.

We make a separate path for our special "licensed but still show the watermark" if we wanted this behavior on dotcom, however we'd have to keep the current behavior in the SDK (or further complicate our license language).

mimecuvalo:
> We make a separate path for our special "licensed but still show the watermark" if we wanted this behavior on dotcom,

just to make sure you saw @steveruizok - this is a dotcom-only change

steveruizok:
Ah in that case let's land it!

Issue #5232, opened 2025-01-20
When i try to drop Network Image from <Image> tag or <img> tag to tldraw canvas on MacOS it doesnt get dropped, but works fine on windows


mtalhastar:
### What happened?

Use Case:

When i try to drop Network Image from <Image> tag or <img> tag to tldraw canvas on **MacOS** it doesn't get dropped but works fine on windows.  but not on MacOS.

![Image](https://github.com/user-attachments/assets/6e269493-5a00-4b64-9c8c-06182b7804ec)

### How can we reproduce the bug?

In order to reproduce this issue open tldraw.com and open a simple localhost application side by side. In localhost application have <img tag and in src put a network image link> and try drop that on tldraw canvas. just like the example I mentioned

https://github.com/user-attachments/assets/47940de3-500b-4343-a480-3e8688ff04c9

### What browsers are you seeing the problem on?

Safari, Chrome

### Where did this happen?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

steveruizok:
Hey, thanks for the issue. I’ll check up on this, though I bet you would be able to debug it yourself by cloning and running the repo locally. The logic for file drop handling is in the “defaultExternalFileHandlers” function, which you can search for. We look at the data on the drop event.

mtalhastar:
Hi @steveruizok , I have tried but still no luck for MacOS, would appreciate support on this from your team

mtalhastar:
@steveruizok i also found that if you open google and search for google images and try to drop them into tldraw.com it never gets dropped in macOS but on windows it works perfectly

MitjaBezensek:
This looks like a bug.

It looks like the issue is that on Macs the dropped files don't get added to the `event.dataTransfer.files` array. So we don't handle them at all because of [this line](https://github.com/tldraw/tldraw/blob/7d33670703347049eb73100fb1d347f2a6da8e30/packages/editor/src/lib/hooks/useCanvasEvents.ts#L122).

We should probably also check `e.dataTransfer.getData('URL')` as it does seem that the url of the image is added there.

PR #5411, opened 2025-02-11
support dragging urls onto the canvas

SomeHats:
Right now, if you drag and drop something from another tab onto tldraw on mac nothing happens, and we don't offer any ways to customise that either. 

This diff makes our onDrop handler accept URLs then pass them onto the URL external asset handler. We could extend this to support e.g. dropping images from other tabs onto the canvas (they come through as URLs), but I've left that for now because it's pretty involved.

### Change type

- [x] `improvement`

### Release notes

- Dragging and dropping URLs onto the canvas is now supported (and creates a bookmark by default)

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -117,16 +117,28 @@ export function useCanvasEvents() {
 			async function onDrop(e: React.DragEvent<Element>) {
 				preventDefault(e)
 				stopEventPropagation(e)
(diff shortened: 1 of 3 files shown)


mimecuvalo:
this is the relevant issue, right? https://github.com/tldraw/tldraw/issues/5232

SomeHats:
yep! although only a sort of half-solution

mimecuvalo (approved):
👍 lgtm

zuoanCo:
How can I turn off this setting?

PR #5414, opened 2025-02-12
use correct domain for app socket connection

ds300:
this was using the raw process.env.MULTIPLAYER_SERVER when it should have been using the MULTIPLAYER_SERVER const defined in our config file, which uses the local origin in staging and prod.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -32,6 +32,7 @@ import {
 	parseTldrawJsonFile,
 	react,
 } from 'tldraw'
+import { MULTIPLAYER_SERVER } from '../../utils/config'
 import { multiplayerAssetStore } from '../../utils/multiplayerAssetStore'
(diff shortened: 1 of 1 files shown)


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -95,7 +96,7 @@ export class TldrawApp {
 				})
 				const token = await getToken()
 				params.set('accessToken', token || 'no-token-found')
-				return `${process.env.MULTIPLAYER_SERVER}/api/app/${userId}/connect?${params}`
+				return `${MULTIPLAYER_SERVER}/api/app/${userId}/connect?${params}`
```
MitjaBezensek, line 99:
```suggestion
				return `${MULTIPLAYER_SERVER}/app/${userId}/connect?${params}`
```

Looks like `MULTIPLAYER_SERVER` already adds the `/api`

MitjaBezensek (approved):

PR #5425, opened 2025-02-13
Actively prune users.

MitjaBezensek:
Track the time that users received their last updates then periodically prune them.

Resolves INT-748

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -135,6 +135,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 
 		const activeUsers = this.sql.exec('SELECT id FROM active_user').toArray() as { id: string }[]
 		const now = Date.now()
+		this.lastUserPruneTime = now
 		for (const user of activeUsers) {
 			this.usersWithUpdates.set(user.id, now)
 		}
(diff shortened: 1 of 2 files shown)


apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -187,6 +190,27 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
+		const activeUserIds = this.ctx.storage.sql
+			.exec('SELECT id FROM active_user')
+			.toArray()
+			.map((u) => u.id) as string[]
+		for (const id of activeUserIds) {
+			const lastUserUpdate = this.usersWithUpdates.get(id)
+			// User has had updates in the last prune period, so skip them
+			if (lastUserUpdate && now - lastUserUpdate < PRUNE_TIME) continue
+			// Their time is up
+			this.usersWithUpdates.delete(id)
```
ds300, line 209: should this line be moved into `unregisterUser`?

ds300 (approved):

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
MitjaBezensek: Their time is up in any case, so I think we should just evict them here. The only thing that will change is that it will short circuit the above if condition.

ds300: but if they're still active but we don't send them another update, they'll always be short-circuited right?

PR #5430, opened 2025-02-13
Replace postgres.js pt 1

ds300:
This replaces all the usages of postgres.js for normal (non-subscription) queries with Kysely. 

The next PR will be replacing the postgres.js subscription with [pg-logcial-replication](https://github.com/kibae/pg-logical-replication) and using a persistent replication slot to avoid the need to reboot user DOs whenever there is connection flake.

Then after that we can look at

- keeping a buffer of previous replication events so if a user DO misses a message or two it can request a catch-up rather than having to reboot.
- look at doing persistence in the user DOs and requesting catchups when starting up so during deploys we can avoid triggering a reboot on every active user DO.

And after that, new socket connections should be able to serve relatively up-to-date data immediately and we should have our immediate perf problems solved for a while.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/postgres.ts
+++ b/apps/dotcom/sync-worker/src/postgres.ts
@@ -1,32 +1,7 @@
 import { DB } from '@tldraw/dotcom-shared'
 import { Kysely, PostgresDialect } from 'kysely'
 import * as pg from 'pg'
-import postgres from 'postgres'
 import { Environment } from './types'
-
-/**
- * In most cases you want to use a pooled connection, which is what `createPostgresConnectionPool` does.
- * Use this one only for cases like subscriptions where you need to listen for notifications / WAL.
- */
-export function createPostgresConnection(
-	env: Environment,
-	{ name, idleTimeout = 30 }: { name: string; idleTimeout?: number }
-) {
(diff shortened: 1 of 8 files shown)


apps/dotcom/sync-worker/src/postgres.ts
```diff
@@ -39,7 +14,7 @@ export function createPostgresConnectionPool(env: Environment, name: string) {
 			connectionString: env.BOTCOM_POSTGRES_POOLED_CONNECTION_STRING,
 			application_name: name,
 			idleTimeoutMillis: 10_000,
-			max: 450,
```
ds300, line 42: i think this should be 1? since it's not a shared connection pool, it's just one pool per durable object, and really we should only need one connection per durable object?

MitjaBezensek (approved):
👍

apps/dotcom/sync-worker/src/postgres.ts
ds300: looks like kysely's postgres dialect straight up requires a connection pool, no option for a single client. 

The only potential downside might be accidentally blocking some operations? e.g. the replicator fetches guest file subscriptions, wouldn't want those to block each other. I'll bump this up to 100 for the replicator, and keep it at 1 for other things that don't serve multiple clients.

PR #5431, opened 2025-02-13
rich text: fonts

mimecuvalo:
Fonty font fonts:
Screenshot 2025-02-13 at 15 18 28

We had a long discussion about how to split up the fonts because of the increased page weight. There's a lot of interesting proposals to explore here, but just trying it out, it's not _too_ bad. We should avoid added complexity for loading at this point I think and we can get fancier later after rich text lands.
Screenshot 2025-02-13 at 15 18 40

Tldraw-ish font settings, I compiled (with David's help) a smaller variable font ~550kb vs ~750kb which excludes "bounciness" and non-informal variants:
Screenshot 2025-02-13 at 15 28 31

for my future reference:
- https://github.com/tldraw/shantell-sans
- https://shantellsans.com/
- https://github.com/IBM/plex/tree/%40ibm/plex-sans%401.1.0/packages
- https://wakamaifondue.com/
- https://fontdrop.info/


mimecuvalo:
moving focus to this PR: https://github.com/tldraw/tldraw/pull/5454

PR #5433, opened 2025-02-14
switch to pg-logical-replication

ds300:
follow up to #5430 

This PR swaps out our postgres.js subscription with pg-logical-replication, which allows us to use a persistent replication slot. This in turn means that when the postgres connection breaks we can reliably restart the replication slot from where we picked off with guaranteed consistency, and therefore we won't need to force our user durable objects to reboot.

In the next PR I will make it so that the replicator keeps a log of events so user DOs can resume via catch-up instead of doing a full reboot.

### Change type

- [x] `other`


MitjaBezensek:
Getting the following error when running locally:
`[TLPostgresReplicator]:  Error: no such column: sequenceId at offset 7: SQLITE_ERROR
 at TLPostgresReplicator.getCurrentSequenceId (file:///Users/mitja/Code/tldraw/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts:402:16)
`

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -120,29 +140,75 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
+			.blockConcurrencyWhile(async () => {
+				await this._migrate().catch((e) => {
+					this.captureException(e)
+					throw e
+				})
+				// if the slot name changed, we set the lsn to null, which will trigger a mass user DO reboot
+				if (this.sqlite.exec('select slotName from meta').one().slotName !== this.slotName) {
+					this.sqlite.exec('UPDATE meta SET slotName = ?, lsn = null', this.slotName)
+				}
+				await sql`SELECT pg_create_logical_replication_slot(${this.slotName}, 'wal2json') WHERE NOT EXISTS (SELECT 1 FROM pg_replication_slots WHERE slot_name = ${this.slotName})`.execute(
```
MitjaBezensek, line 205: Is `this.slotName` guaranteed to be always the same (not sure if DO ids are guaranteed to be the same across deploys, reboots,...)?

ds300:
try pulling! i removed the sequenceId stuff yesterday

MitjaBezensek:
> try pulling! i removed the sequenceId stuff yesterday

Oh sorry, I had the branch from before and didn't check for changes 😄

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -344,10 +344,7 @@ export class UserDataSyncer {
 	}
 
 	handleReplicationEvent(event: ZReplicationEvent) {
-		if (
-			event.type === 'force_reboot' ||
-			('sequenceId' in this.state && this.state.sequenceId !== event.sequenceId)
-		) {
+		if ('sequenceId' in this.state && this.state.sequenceId !== event.sequenceId) {
```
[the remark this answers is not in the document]
MitjaBezensek: Managed to get this error immediately after starting the server. I guess we should an early return for `maybe_force_reboots` if the above condition is not met?
```ts
dev::apps/dotcom/zero-cache [0] zstart_postgres-1  | 2025-02-18 09:25:00.858 UTC [137] STATEMENT:  START_REPLICATION SLOT "tlpr_91905267667cd883ee49290cd466714f429d489e97c46cfca3fa0944f8" LOGICAL 0/00000000
dev::apps/dotcom/sync-worker ✘ [ERROR] Error in messageUser Error: event type should be row_update got maybe_force_reboot
dev::apps/dotcom/sync-worker
dev::apps/dotcom/sync-worker       at null.<anonymous> (async file:///Users/mitja/Code/tldraw/apps/dotcom/sync-worker/.wrangler/tmp/dev-dBNJbc/worker.js:60596:21)
dev::apps/dotcom/sync-worker       at async TLPostgresReplicator.messageUser (file:///Users/mitja/Code/tldraw/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts:684:4) {
dev::apps/dotcom/sync-worker     remote: true
dev::apps/dotcom/sync-worker   }
```

MitjaBezensek (approved):

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
ds300: yeah the DO ID is guaranteed to be stable. 

this.slotName would only change if we manually alter the prefix and redeploy, and we'd only need to do that if something goes horribly wrong and we need to force all the active user DOs to reboot from scratch. I don't think we'll ever need to do that.

PR #5435, opened 2025-02-14
[botcom] Fix slow export menu in big files

steveruizok:
This PR fixes an issue where large files or files with GIFs / large images could freeze when opening the export menu.

https://github.com/user-attachments/assets/af0b21c2-d50b-4106-bef6-7738868f863e

## GIFs

Previously, we were inlining the entire GIF when creating an SVG. We now only inline the first frame. This was a crashing bug previously.

## Images / asset caching

Previously, we were loading a URL for each image or video shape. We now cache the result of the previous generation, so that we only load one image per asset. This is especially important for GIFs because the work to generate the export image src associated with the asset was expensive.

## File size

Previously, the image shown in the export panel could be very very large depending on the size of the canvas.

**Problem**

When we export images, we do so at 100% resolution. For large canvases, that can result in images which are very large (e.g. 5000x5000px or larger). Creating and displaying such a large image can be extremely hard on resources.

**Solution**

In this PR, we find a **scale** for the image based on the common bounds of the selected shapes, then apply that scale to keep the image small (under 500x500 pixels). We also use png instead of svg as the output format, which for small but dense images should work better. Together, these lead to preview images for even very large files (4000 draw shapes) being relatively quick. The actual exports are unaffected and slower.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Visit a very large file.
2. Open the export menu.

### Release notes

- Fixed a bug with export menu performance.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaFileShareMenu/Tabs/TlaExportTab.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaFileShareMenu/Tabs/TlaExportTab.tsx
@@ -1,6 +1,7 @@
 import classNames from 'classnames'
 import { useCallback, useRef, useState } from 'react'
 import {
+	Box,
 	Editor,
 	FileHelpers,
 	TLExportType,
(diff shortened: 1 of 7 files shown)


mimecuvalo:
Since you're looking into this, I wanted to make sure you saw this related issue @steveruizok  in case it wasn't on your radar:
https://linear.app/tldraw/issue/INT-701/export-just-previewing-a-heavy-room-in-export-causes-the-page-to-crash

steveruizok:
> Since you're looking into this, I wanted to make sure you saw this related issue @steveruizok in case it wasn't on your radar: https://linear.app/tldraw/issue/INT-701/export-just-previewing-a-heavy-room-in-export-causes-the-page-to-crash

This may be two different problems, possibly related to the GIFs in that file. Even with a handful of GIFs the image generation takes a very long time. Any ideas @mimecuvalo @MitjaBezensek?

steveruizok:
Are we trying to put the whole GIF into the SVG?

steveruizok:
lol yeah we were, ok this time we're only going to put in the first frame

apps/dotcom/client/src/tla/components/TlaFileShareMenu/Tabs/TlaExportTab.tsx
```diff
@@ -334,21 +335,25 @@ async function getEditorImage(
...
+	// image max is 216x216, so let's say 500 to be nice and safe
+	const scale = Math.min(500 / commonBounds.width, 500 / commonBounds.height)
+
+	const result = await editor.toImage(shapes, {
+		scale,
+		format: 'png',
+		padding: exportPadding ? editor.options.defaultSvgPadding : 0,
+		background: exportBackground,
+		darkMode: exportTheme === 'auto' ? undefined : exportTheme === 'dark',
+	})
```
steveruizok, line 350: Scale down the preview image on dotcom in fit within reasonable UI bounds.

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -121,17 +126,29 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
...
+				src = (await getDataURIFromURL(src)) || ''
+			}
+
+			// If it's animated then we need to get the first frame
+			if (getIsAnimated(this.editor, asset.id)) {
+				const { promise } = getFirstFrameOfAnimatedImage(src)
+				src = await promise
+			}
+			return src
+		})
```
steveruizok, line 149: For images (and later videos) we cache a promise that will resolve the src for an image.

MitjaBezensek:
Nice 👍

PR #5440, opened 2025-02-17
Add dotcom checks that will serve for alerting

MitjaBezensek:
Collect some basic stats which can be used for alerting purposes.

I investigated using checkly for some browser tests but decided against it. The main is is that it's not trivial to login from test automations like playwright. Even if you set the username / password for Oauth, you might get hit by captcha / automation detection. Two ways to get around that would be:
* Allowing sign in via username + password combo.
* Using some sort of email throwaway / burner service, which would allow api access to get the sign in codes.

This will use our existing updown.io health checks, but will extend them with some basic app specific health checks. One downside of this is that we use CF for collecting these checks, so we might report DB / Clerk being down, when it's in fact CF that is down. That said I don't think that's a big problem since if CF is down our app is down for sure.

I have added checks to updown.io, will enable them once this lands in prod.

### Change type

- [x] `improvement`


apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -116,6 +109,16 @@ const router = createRouter<Environment>()
 	})
 	.post('/app/submit-feedback', submitFeedback)
 	// end app
+	.all('/ph/*', (req) => {
```
MitjaBezensek, line 112: Move this one outside of `app` routes.

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -335,6 +340,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 
 	private handleEvent(row: postgres.Row | null, event: postgres.ReplicationEvent) {
 		this.lastPostgresMessageTime = Date.now()
+		getStatsDurableObjct(this.env).recordReplicatorPostgresUpdate()
```
ds300, line 343: could we throttle this? just worried about adding O(n) extra IO operations to the replicator, might make deploys even slower until we fix the reboot issue.

apps/dotcom/sync-worker/src/healthCheckRoutes.ts
```diff
@@ -0,0 +1,67 @@
...
+				return new Response('Could not reach clerk', { status: 500 })
+			}
+			return new Response('ok', { status: 200 })
+		} catch (_e) {
+			return new Response('Could not reach clerk', { status: 500 })
+		}
+	})
+	.get('/health-check/db', async (_, env) => {
+		try {
+			await createPostgresConnectionPool(env, 'TLStatsDurableObject')
```
ds300, line 56:
```suggestion
			await createPostgresConnectionPool(env, '/health-check/db')
```

apps/dotcom/sync-worker/src/TLStatsDurableObject.ts
```diff
@@ -0,0 +1,110 @@
...
+	async unusualNumberOfUserDOAborts() {
+		this.pruneUserDoAborts()
+		return this.userDoAborts.length > USER_DO_ABORT_THRESHOLD
+	}
+
+	async areUsersRecevingReplicationEvents() {
+		if (this.bootingUp()) return true
+
+		return this.lastReplicationEvent > this.getCutoffTime()
+	}
```
ds300, line 98: this feels prone to false positives to me. Maybe we're big enough that it'll never happen but I wouldn't put money on that.

ds300 (approved):
Happy to give this a go. Shame that having playwright-based tests didn't work out.

apps/dotcom/sync-worker/src/TLStatsDurableObject.ts
MitjaBezensek: Yeah, was also on the fence about it. We do seem to get the updates at all times, but there might be some down day (like a holiday) which trigger this. Removed it for now. 

Also feels like the users do aborts should cover some of this. The replicator getting stuck that we saw a while back would be triggered by the high amount of aborts.

PR #5445, opened 2025-02-17
Remove file creation route.

MitjaBezensek:
Remove the legacy new room creation route.

### Change type

- [x] `improvement`


apps/dotcom/client/src/pages/public-touchscreen-side-panel.tsx
MitjaBezensek, line 1: Looks like this was also creating new rooms. Asked @mimecuvalo about it and it seems it's safe to delete.

ds300 (approved):

PR #5454, opened 2025-02-18
wip: dynamically load fonts based on text styling

SomeHats:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


packages/editor/src/lib/editor/managers/FontManager.ts
```diff
@@ -0,0 +1,201 @@
...
+		const url = this.assetUrls?.[font.src.url] ?? font.src.url
+		const instance = new FontFace(font.fontFamily, `url(${JSON.stringify(url)})`, {
+			ascentOverride: font.ascentOverride,
+			descentOverride: font.descentOverride,
+			stretch: font.stretch,
+			style: font.style,
+			weight: font.weight,
+			featureSettings: font.featureSettings,
+			lineGapOverride: font.lineGapOverride,
+			display: 'swap',
```
mimecuvalo, line 160: let's consider `block` here instead of `swap`

packages/editor/src/lib/editor/managers/FontManager.ts
```diff
@@ -0,0 +1,201 @@
...
+	tech?: string
+}
+
+/**
+ * A font face that can be used in the editor. The properties of this are largely the same as the
+ * ones in the
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face | css `@font-face` rule}.
+ */
+export interface TLFontFace {
+	/**
```
mimecuvalo, line 27: there's some experimental ones here `variant` and `variationSettings` — do we care?

packages/editor/src/lib/editor/managers/FontManager.ts
```diff
@@ -0,0 +1,201 @@
...
+ * See {@link https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src | `src`} for details of the properties here.
+ */
+export interface TLFontFaceSource {
+	/**
+	 * A URL from which to load the font. If the value here is a key in
+	 * {@link TLEditorAssetUrls.fonts}, the value from there will be used instead.
+	 */
+	url: string
+	format?: string
+	tech?: string
```
mimecuvalo, line 18: tech?

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -65,6 +79,21 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
 		})
 	}
 
+	override getFontFaces(shape: TLTextShape): TLFontFace[] {
+		const list = getFontsFromRichText(
+			shape.props.richText,
+			tipTapDefaultExtensions,
+			{
+				family: `tldraw_${shape.props.font}`,
+				weight: '500',
```
mimecuvalo, line 88: i see 500 in lots of places as the default. our current base font weight is actually 400. wdyt?

mimecuvalo (commented):
we discussed offline several things - these are just some initial side-comments before i forget about them

packages/editor/src/lib/editor/managers/FontManager.ts
SomeHats: See https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src

packages/editor/src/lib/editor/managers/FontManager.ts
SomeHats: my sense is that right now there's not enough browser support that I feel confident including them - they largely won't work right now

packages/editor/src/lib/editor/managers/FontManager.ts
SomeHats: I was unable to see a noticeable difference for these. I think they're maybe intended for a different use-case than ours

packages/editor/src/lib/editor/managers/FontManager.ts
mimecuvalo: huh TIL

packages/editor/editor.css
```diff
@@ -1384,6 +1379,9 @@ input,
 	opacity: 0;
 	animation: fade-in 0.2s ease-in-out forwards;
 	animation-delay: 0.2s;
+	position: absolute;
```
mimecuvalo, line 1382: this is new? what is it fixing?

SomeHats: new preload screen for fonts

mimecuvalo (commented):
this is lookin solid 🙌 
i'm gonna cleanup any tests and then merge, we can continue any followup stuff (like the docs stuff you had in mind) on the main PR

PR #5460, opened 2025-02-20
retry slurping document records if need be

ds300:
There was a hole in the local file slurping logic where it would retry slurping assets but not the document records themselves. So if the user switched away from the slurped file before tlsync had a chance to finsih doing its thing, the data would seem to disappear.

This patches that hole by using the `createSource` column to signify that an app file should be initialized by slurping from a local file, rather than an ephemeral property on the TldrawApp instance. That way when we open such a file we can check whether the document.meta.slurpPersistenceKey has been added, and if so it means the records slurp happened successfully.

### Change type

- [x] `bugfix`

### DIFF
--- a/apps/dotcom/client/src/tla/utils/slurping.tsx
+++ b/apps/dotcom/client/src/tla/utils/slurping.tsx
@@ -73,8 +73,9 @@ export async function maybeSlurp(opts: SlurperOpts) {
 	if (opts.abortSignal.aborted) return
 	if (!opts.app.isFileOwner(opts.fileId)) return
 	const file = opts.app.getFile(opts.fileId)
-	if (!file?.createSource?.startsWith('lf/')) return
-	const persistenceKey = file.createSource.slice('lf/'.length)
+	const persistenceKey =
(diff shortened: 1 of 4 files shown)


apps/dotcom/client/src/tla/utils/slurping.tsx
```diff
@@ -75,13 +72,9 @@ interface SlurperOpts {
...
 	if (!opts.app.isFileOwner(opts.fileId)) return
-	let persistenceKey = null as string | null
-	if (opts.app._slurpFileId === opts.fileId) {
-		// we just landed on this file after signing in on the root page
-		persistenceKey = getScratchPersistenceKey()
-	} else {
-		persistenceKey = (opts.editor.getDocumentSettings().meta.slurpPersistenceKey as string) ?? null
-	}
+	const file = opts.app.getFile(opts.fileId)
+	if (!file?.createSource?.startsWith('lf/')) return
```
MitjaBezensek, line 76: Should we use `LOCAL_FILE_PREFIX` for all of these `lf` instances?

MitjaBezensek (approved):

PR #5461, opened 2025-02-20
Add sidebar context menu

MitjaBezensek:
Open a context menu when right clicking the file link. It shows the same items as the dropdown that is invoked by the three dots menu.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/client/package.json
+++ b/apps/dotcom/client/package.json
@@ -27,7 +27,7 @@
 	},
 	"dependencies": {
 		"@clerk/clerk-react": "^5.22.8",
-		"@radix-ui/react-context-menu": "^2.2.6",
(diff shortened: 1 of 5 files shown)


ds300 (approved):
This works well on desktop. Long press on iOS is kinda messed up because it opens both our menu and the native iOS link context menu. In safari it even looks bad because some of our menu gets cut off. We should only do one or the other. My vote would be to disable our context menu on mobile, since the three dots are always visible anyway.

MitjaBezensek:
> This works well on desktop. Long press on iOS is kinda messed up because it opens both our menu and the native iOS link context menu. In safari it even looks bad because some of our menu gets cut off. We should only do one or the other. My vote would be to disable our context menu on mobile, since the three dots are always visible anyway.

Makes sense yeah 👍

PR #5471, opened 2025-02-22
Consider `https://localhost` to be development

jamesbvaughan:


jamesbvaughan:
My current workaround for this is to add a domain in my `/etc/hosts` that's part of our license key and use that domain for local development rather than `localhost`.

With that workaround, this change no longer feels as important for me. I'll leave this open in case y'all think it would be useful to merge for others, but feel free to close it otherwise.

steveruizok:
Looks good to me!

mimecuvalo (commented):
I think your argument makes sense — thanks for the suggestion!
Before we land, do you want to modify the test in `LicenseManager.test.ts` → "Signals that it is development mode when appropriate" to reflect this change?

jamesbvaughan:
> Before we land, do you want to modify the test in `LicenseManager.test.ts` → "Signals that it is development mode when appropriate" to reflect this change?

Done!

PR #5478, opened 2025-02-23
[hotfixme] Fix copy as SVG

steveruizok:
This PR fixes a bug where Copy as SVG would throw an error.

Previously, copying as an SVG would copy the image as plain text. In #5114 we started writing the image blob to the clipboard for SVGs, though the mimeType we were using was still plain text. This would cause an error: `NotAllowedError: Failed to execute 'write' on 'Clipboard': Type text/plain does not match the blob's type image/svg+xml`.

In this PR, we now correctly write the image to the clipboard. We _also_ write the SVG string as plain text.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Select some shapes
2. Copy as SVG
3. Paste into a text editor (should paste the string)
4. Paste into a raster graphics program (should paste the SVG)
5. Paste into Figma (should paste as individual shapes)

### Release notes

- Fixed a bug with Copy as SVG

### DIFF
--- a/packages/tldraw/src/lib/utils/export/copyAs.ts
+++ b/packages/tldraw/src/lib/utils/export/copyAs.ts
@@ -73,10 +73,9 @@ export function copyAs(
 		}
 
 		if (opts?.format === 'svg') {
-			const blobPromise = editor
+			types['text/plain'] = editor
 				.getSvgString(ids, opts)
 				.then((result) => new Blob([result?.svg ?? ''], { type: 'text/plain' }))
-			types['text/plain'] = blobPromise
 		}
 
 		return clipboardWrite(types)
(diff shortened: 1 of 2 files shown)


SomeHats:
ooh i clocked this on my fonts branch but didn't realise it was on `main` too. Let's go with this one instead: https://github.com/tldraw/tldraw/pull/5482

Reasoning:
- Most browsers are really strict about what you can write to the clipboard. `image/svg+xml` (even `image/jpeg`) isn't allowed. that's why we have this text-writing stuff in the first place
- I don't want us to do a double-export, as they're pretty expensive. If we do want to write the SVG as two types, I'd rather pull the data from a single blob

PR #5479, opened 2025-02-23
Fix alignment, distribute

steveruizok:
This PR fixes alignment and distribution. It:

## Alignment

Fixes a bug where arrows were being taken into account when aligning shapes.

Kapture 2025-02-23 at 09 41 23

## Distribution

Adjusts the logic for distribution so that the result will space shapes out evenly between the first and last shapes, rather than distributing by centers.

Fixes a bug when distributing shapes if the same shape constituted the first and last shape.

Kapture 2025-02-23 at 09 44 11

### Change type

- [x] `bugfix`

### Test plan

1. Align a shape to an edge where that edge in the shape bounds is defined by an arrow
2. Distribute overlapping or irregular shapes

- [ ] Unit tests: NEEDS DOING

### Release notes

- Fixed a bug when aligning shapes with arrows
- Fixed a bug with distribution with overlapping shapes

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -2506,7 +2506,7 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
     constructor(editor: Editor);
     // @internal
     backgroundComponent?(shape: Shape): any;
-    canBeLaidOut(_shape: Shape): boolean;
+    canBeLaidOut(_shape: Shape, _type?: 'align' | 'distribute' | 'flip' | 'pack' | 'stack'): boolean;
     canBind(_opts: TLShapeUtilCanBindOpts<Shape>): boolean;
     canCrop(_shape: Shape): boolean;
     canDropShapes(_shape: Shape, _shapes: TLShape[]): boolean;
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -261,29 +261,31 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 
 					trackEvent('toggle-auto-size', { source })
 					editor.markHistoryStoppingPoint('toggling auto size')
-					const shapes = editor
-						.getSelectedShapes()
-						.filter(
-							(shape): shape is TLTextShape =>
-								editor.isShapeOfType<TLTextShape>(shape, 'text') && shape.props.autoSize === false
+					editor.run(() => {
```
mimecuvalo, line 264: not for this PR, but i think it would be good for DX to have `editor.run` take a history label (in this case, "toggling auto size") as part of the options. `editor.run` is basically a wrapper around history, so it would make sense to have it do `markHistoryStoppingPoint` under the hood. otherwise, people have to remember to do two separate things to make this work correctly.

steveruizok: I'm almost certain we had this at one point, where editor.mark and ye olde editor.batch were the same thing. There were so many cases of needing to have multiple frames-worth of stuff on a single undo/redo entry that the option felt like a trap so we separated it out when we refactored history.

PR #5482, opened 2025-02-24
fix copy as svg mime type

SomeHats:
The type here doesn't match the mimetype. They're different because browsers don't support writing `image/svg+xml` to the clipboard, so we write `text/plain` instead. But the blob was coming through with the `text/plain` format.

### Change type

- [x] `bugfix`

### Release notes

- Fix copy as svg

### DIFF
--- a/packages/tldraw/src/lib/utils/clipboard.ts
+++ b/packages/tldraw/src/lib/utils/clipboard.ts
@@ -1,4 +1,5 @@
 import { getOwnProperty, objectMapFromEntries } from '@tldraw/editor'
+import { TLCopyType } from './export/copyAs'
 
 // Browsers sanitize image formats to prevent security issues when pasting between applications. For
 // paste within an application though, some browsers (only chromium-based browsers as of Nov 2024)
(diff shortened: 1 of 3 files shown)


steveruizok:
Did https://github.com/tldraw/tldraw/pull/5478 introduce a bug? I wasn't getting errors on that fix

steveruizok:
Ah I see, I thought that had landed. All good

SomeHats:
> did #5478 introduce a bug? I wasn't getting errors on that fix

Yup - if you try it in something like firefox, you'll get an error and nothing will be copied

> If we are able to write the image to the clipboard (in certain browsers) that would be useful enough to try. But that's still a general limitation across browsers?

We have some stuff for writing to the clipboard with multiple mime types that we already use - i've adapted it to work for SVGs too. We should now be writing both if we detect that the browser supports writing svgs to the clipboard

steveruizok (approved):

steveruizok:
Great, let's go!

Issue #1084, opened 2022-12-22
[bug] The beta version does not show the pasted images, only their box


navakelvin:
Hello, the beta version does not show the images, only its box. To reproduce, simply copy and paste an image into a project in beta.

![image](https://user-images.githubusercontent.com/77164639/209035087-811923e3-57b0-4044-96b9-5d677b140220.png)

However, when exporting in the exported image if it is displayed.

PR #5487, opened 2025-02-24
Fix text padding, add context to shape geometry

steveruizok:
Long have we suffered from this:

image

Text has no horizontal padding. We've tried adding horizontal padding before (see #1084, #2855) however we always ran into a circular problem.

- Adding horizontal text padding makes arrows look better
- ...but then it runs alignment / snapping / selection
- ...and it makes text move around

image

What do we really want? **Contextual geometry**.

Kapture 2025-02-24 at 15 42 22

If I'm creating arrow info, I might want a different geometry than if I'm snapping or aligning. 

**Pros:**

 - Additive API, no other changes
 - No migrations needed
 - Opens the door for other contextual geometries
 - Might be useful also for lines, highlighters, draw shapes, and other shapes
 
**Cons**

- More geometries cached (ie any shape that I've ever pointed an arrow to)

### Change type

- [x] `improvement`
- [x] `api`

### Test plan

1. Create a text shape
2. Make a horizontal arrow pointing from the text shape to something else

### Release notes

- Improved horizontal padding for arrows bound to text shapes

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
@@ -51,11 +51,12 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
 		return sizeCache.get(shape.props, (props) => getTextSize(this.editor, props))
 	}
 
-	getGeometry(shape: TLTextShape) {
+	getGeometry(shape: TLTextShape, context: string) {
 		const { scale } = shape.props
 		const { width, height } = this.getMinDimensions(shape)!
 		return new Rectangle2d({
-			width: width * scale,
+			x: context === 'arrow' ? -10 : 0,
+			width: width * scale + (context === 'arrow' ? 20 : 0),
 			height: height * scale,
 			isFilled: true,
 			isLabel: true,
--- a/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
@@ -5,6 +5,7 @@ import {
 	Rectangle2d,
 	ShapeUtil,
 	SvgExportContext,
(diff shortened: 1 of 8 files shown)


ds300 (approved):
this is definitely better than what came before, but now I'd argue the padding for incoming arrows is a bit too much.

image

I wonder if it would be possible to break the 'arrow' context into 'arrow-point' and 'arrow-end' or something like that then keep the original geometry for the pointy bits.

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -51,11 +51,12 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
 		return sizeCache.get(shape.props, (props) => getTextSize(this.editor, props))
 	}
 
-	getGeometry(shape: TLTextShape) {
+	getGeometry(shape: TLTextShape, context: string) {
```
mimecuvalo, line 54: i think, looking forward, `context` might possibly end up being different, multiple kinds of things to help paint a picture. wdyt about it being an object? (with an initial property called `{originShape : 'arrow'}`)

steveruizok:
> this is definitely better than what came before, but now I'd argue the padding for incoming arrows is a bit too much.

Ah good call, let me take a look at that.

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
steveruizok: Makes sense. My trick for the different computed caches needs a string for the key but this could take an object.

steveruizok:
Ok, a few changes:

- Uses an options object for context
- Uses a more specific context key (`@tldraw/arrow-start`)

steveruizok:
(Great first use of the new shape options too)

PR #5488, opened 2025-02-25
Remove canvas size dependency

steveruizok:
This PR removes the canvas-size dependency and incorporates just the parts that we were using (I tipped the author $100). This was originally motivated by some silly mjs export stuff but the library also had a lot of additional support for web workers etc that we weren't using.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Try to create very large images; ie paste in a huge png.

### Release notes

- API: removes canvas-size dependency.


packages/editor/src/lib/utils/browserCanvasMaxSize.ts
```diff
@@ -1,39 +1,142 @@
...
+			}
+		}
+	}
+
+	// release memory
+	cropCvs.width = 0
+	cropCvs.height = 0
+
+	throw Error('Failed to determine maximum canvas dimension')
+}
```
steveruizok, line 136: No longer a promise as the underlying work is synchronous anyway. Was there a different reason for needing a promise before?

PR #5489, opened 2025-02-25
docs: Clarity on examples path

andrewginns:
Adds clarity that to run the local demo example app you need to be in the `apps/examples` folder.

Am unsure if this was just unclear to me due to unfamiliarity with Typescript project structures.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Adds example folder path to local demo instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,11 @@ Learn more at [tldraw.dev](https://tldraw.dev).
 
 The local development server will run our examples app. The basic example will show any changes you've made to the codebase.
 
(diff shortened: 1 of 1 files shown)


andrewginns:

SomeHats:
Thanks for this! This shouldn't actually be the case though - running `yarn dev` from the top should start the examples app (and a couple other services) just fine. Did you run into issues when doing that?

PR #5490, opened 2025-02-25
Use vite to load the assets

MitjaBezensek:
Moves icon assets outside of the public folder. Uses vite to load the merged sprite. Adds sprite preloading.

For now I didn't merge the two sprite files, didn't really feel like it's necessary for now.

### Change type

- [x] `improvement`


internal/scripts/refresh-assets.ts
```diff
@@ -152,7 +152,7 @@ async function copyIcons() {
 }
 
 async function optimizeAndMergeDotcomIcons() {
-	const sourceFolderPath = join(DOTCOM_FOLDER_PATH, 'client', 'public', 'tla')
+	const sourceFolderPath = join(DOTCOM_FOLDER_PATH, 'client', 'assets', 'icons', 'icon')
```
MitjaBezensek, line 155: Used the same folder structure as we use for our assets package. Might be overkill 🤷‍♂️

apps/dotcom/client/scripts/build.ts
```diff
@@ -63,24 +63,36 @@ async function build() {
 		'IBMPlexSans-Medium',
 		'IBMPlexMono-Medium',
 	]
-	const indexHtml = await readFileSync('.vercel/output/static/index.html', 'utf8')
-	await writeFileSync(
-		'.vercel/output/static/index.html',
-		indexHtml.replace(
-			'<!-- $PRELOADED_FONTS -->',
```
SomeHats, line 70: woah i had no idea we did this... I _think_ vite supports this out of the box and we don't need to be doing all this replacement stuff? If you put a URL starting with `./` in the href that goes from the filesystem location of index.html to the file system location of the asset, it _should_ pick it up automatically?

mimecuvalo: Yeah, when I implemented this I couldn't get it working with the Vite mechanism. But yeah, maybe worth trying the refresh-assets as you mentioned Mitja, not sure.

mimecuvalo (approved):
LGTM - i think if we wanna rethink the replacement stuff that can happen in a followup PR, i don't feel strongly

apps/dotcom/client/scripts/build.ts
SomeHats: I couldn't get `href="@tldraw/assets/icons/icon/0_merged.svg"` to work, but I could get `href="../../../packages/assets/icons/icon/0_merged.svg"` to work 🤷

So the main thing I'm worried about here is cases where assets share a prefix. For example, we have `icon-sidebar`, and `icon-sidebar-strong`. With the current `startsWith` approach, `icon-sidebar` might match `icon-sidebar-strong` and we'd end up with the wrong asset in use. 

I think it might be fine (maybe sorting them helps here) but an alternative to this or the embedded approach would be to do this as a vite plugin instead of in build.ts - that way we could leverage vite's resolution system and make sure the assets are mapped to the correct URLs.

That might be nicer than having loads of those preload links listed out? not sure

SomeHats: Anyway I don't want to blow this up into a huge amount of work - starting with just updating that one TLA one might be fine for now?

MitjaBezensek: Not sure I understand the `startsWith` comment: we are only loading `0_merged_tla-` and `0_merged-` asset files, not the individual icons themselves, so those should not match?

In any case I moved the tla sprite icon preloading to a direct link in `index.html` and it seems to be working 👍

PR #5495, opened 2025-02-25
Fix issue with duplicating bound arrows.

MitjaBezensek:
[Reversing the diff](https://github.com/tldraw/tldraw/blob/d799df28e99e17bb71d1ab198f7c910fc3987a88/packages/editor/src/lib/editor/Editor.ts#L9981) caused some Arrow side effects to trigger, which updated the indexes of the arrows. This then caused the [siblings array](https://github.com/tldraw/tldraw/blob/d799df28e99e17bb71d1ab198f7c910fc3987a88/packages/editor/src/lib/editor/Editor.ts#L5845) in `duplicateShapes` to get out of order which resulted in  errors like`Error: aU4P6 >= aTnD8` since we passed incorrect lower and upper values to `getIndexBetween`.

### Change type

- [x] `bugfix`

### Release notes

- Fix a bug with duplicating bound arrows.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -9978,7 +9978,7 @@ function withIsolatedShapes<T>(
 				}
 			})
 
-			editor.store.applyDiff(reverseRecordsDiff(changes))
+			editor.store.applyDiff(reverseRecordsDiff(changes), { runCallbacks: false })
 		},
 		{ history: 'ignore' }
 	)
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5788,7 +5788,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 			const { shapesToCreateWithOriginals, bindingsToCreate } = withIsolatedShapes(
 				this,
 				shapeIdSet,
-				false,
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/test/bindings.test.tsx
```diff
@@ -246,9 +246,6 @@ test('copying the from shape on its own does trigger isolation operations', () =
 		  "onBeforeDelete",
 		  "onAfterDelete",
 		  "onOperationComplete",
-		  "onBeforeCreate",
```
MitjaBezensek, line 249: @ds300 Saw that you wrote these tests. Are there any observable issues if these don't get triggered?

MitjaBezensek: Reworked this.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9942,6 +9943,7 @@ function pushShapeWithDescendants(editor: Editor, id: TLShapeId, result: TLShape
 function withIsolatedShapes<T>(
 	editor: Editor,
 	shapeIds: Set<TLShapeId>,
+	runCallbacks: boolean,
```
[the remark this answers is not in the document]
MitjaBezensek: Updated this: removed the option and fixed the tests.

ds300 (approved):
good fix!

PR #5496, opened 2025-02-25
Add `AtomMap` & refactor store

SomeHats:
This diff adds `AtomMap`, a drop-in replacement for `Map` that stores values in reactive atoms. `AtomMap` replaces the old storage mechanism for the `Store` - an `Atom<ReadonlyMap<Id, Atom<Record>>>`.

The main benefit (and motivation behind this change) is that adding/removing a record from an `AtomMap` no longer causes everything accessing a record from the store to recompute. This cuts down on the number cache invalidations on unrelated changes pretty dramatically. The fact that `AtomMap` is a motivation too - it does't have a second use in this diff, but in #4895, this replaces another adhoc atom-map like structure in the font manager.

There are a couple of additional changes in this diff that I pulled out of the rich text/fonts PR:
- `createSelectedComputedCache`, which had no uses throughout the codebase, has been deleted. There's a new `createCache` method that accepts a function returning a `computed` which can be used to create selector caches, or anything more complex you might need.
- `createComputedCache` now accepts an options object instead of a plain `isEqual` callback so we can have callbacks for record-equality and result-equality.

### Change type

- [x] `api`

### Release notes

- **BREAKING**. `store.createSelectedComputedCache` has been removed. Use `store.createCache` and create your own selector `computed` instead.
- **BREAKING**. `createComputerCache` no longer accepts a single `isEqual` fn as its 3rd argument. Instead, pass in an options object, with the `isEqual` fn named `areRecordsEqual`. You can now pass `areResultsEqual`, too.


steveruizok:
Just to double double check, this fixes the revalidation issue we reviewed with #4895?

packages/store/src/lib/AtomMap.ts
```diff
@@ -0,0 +1,201 @@
...
+
+	*values(): Generator<V, undefined, unknown> {
+		for (const key of this.presentKeysAtom.get()) {
+			const value = this.getAtom(key)!.get()
+			if (value === UNINITIALIZED) {
+				continue
+			}
+			yield value
+		}
+	}
```
ds300, line 183: unlike normal js maps these iterators won't be tolerant of updates (particularly insertions) because they iterate over the presentKeysAtom rather than the underlying JS Map.

Maybe they should deference the presentKeysAtom to get reactivity, and then iterate over this.valueMap.values() etc to make sure the iterator gets values that were inserted during iteration like a normal js map?

packages/store/src/lib/test/AtomMap.test.ts
```diff
@@ -0,0 +1,563 @@
...
+			expect(reactor).toHaveLastReturnedWith(true)
+
+			map.set('a', 2)
+			expect(reactor).toHaveBeenCalledTimes(3) // setting a value triggers a reaction
+			expect(reactor).toHaveLastReturnedWith(true)
+
+			map.delete('a')
+			expect(reactor).toHaveBeenCalledTimes(4)
+			expect(reactor).toHaveLastReturnedWith(false)
+		})
```
ds300, line 106: i'd be interested in checking that the reactivity remains after deletion to make sure reinsertions work
```suggestion

			map.set('a', 2)
			expect(reactor).toHaveBeenCalledTimes(5) // reinstating a previously deleted atom works
			expect(reactor).toHaveLastReturnedWith(true)
		})
```

ds300 (approved):
love this so much! big perf win.

packages/store/src/lib/AtomMap.ts
```diff
@@ -0,0 +1,201 @@
...
+			return undefined
+		}
+		return valueAtom
+	}
+
+	get(key: K): V | undefined {
+		const value = this.getAtom(key)?.get()
+		if (value === UNINITIALIZED) {
+			return undefined
+		}
```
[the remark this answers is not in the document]
SomeHats: good shout!

PR #5497, opened 2025-02-25
no reboot user data on deploy

ds300:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/tla/app/zero-polyfill.ts
+++ b/apps/dotcom/client/src/tla/app/zero-polyfill.ts
@@ -179,14 +179,7 @@ export class Zero {
 				if (store?.files.find((f) => f.id === data.id)) {
 					throw new Error('file already exists')
 				}
-				this.makeOptimistic([
-					{ table: 'file', event: 'insert', row: data },
-					{
-						table: 'file_state',
-						event: 'insert',
-						row: { fileId: data.id, userId: store.user.id, firstVisitAt: Date.now() } as any,
-					},
-				])
+				this.makeOptimistic([{ table: 'file', event: 'insert', row: data }])
 			},
 			update: (data: TlaFilePartial) => {
 				const existing = this.store.getFullData()?.files.find((f) => f.id === data.id)
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -1,4 +1,4 @@
-import { DB, TlaFile, TlaRow, TlaUser, ZTable } from '@tldraw/dotcom-shared'
+import { DB, TlaFile, TlaFileState, TlaRow, TlaUser, ZTable } from '@tldraw/dotcom-shared'
 import {
 	ExecutionQueue,
 	assert,
@@ -15,6 +15,7 @@ import { Kysely, sql } from 'kysely'
 
 import { LogicalReplicationService, Wal2Json, Wal2JsonPlugin } from 'pg-logical-replication'
 import { Logger } from './Logger'
+import { UserChangeCollator } from './UserChangeCollator'
 import { ZReplicationEventWithoutSequenceInfo } from './UserDataSyncer'
 import { createPostgresConnectionPool } from './postgres'
 import {
@@ -45,7 +46,8 @@ interface ReplicationEvent {
 
 interface Change {
 	event: ReplicationEvent
-	topicId: string
+	userId: string
+	fileId: string | null
 	row: TlaRow
 }
 
@@ -125,11 +127,12 @@ const migrations: Migration[] = [
         -- so we can do range queries on it
         walSegment INTEGER NOT NULL,
         walSegmentOffset INTEGER NOT NULL,
-				-- topicId is either a user id or a file id
-				topicId TEXT NOT NULL,
+				userId TEXT NOT NULL,
+				fileId TEXT,
         json TEXT NOT NULL
       );
-			CREATE INDEX history_walSegment_walSegmentOffset_topicId ON history (walSegment, walSegmentOffset, topicId);
+			CREATE INDEX history_walSegment_walSegmentOffset_userId ON history (walSegment, walSegmentOffset, userId);
+			CREATE INDEX history_walSegment_walSegmentOffset_fileId ON history (walSegment, walSegmentOffset, fileId);
 		`,
 	},
 ]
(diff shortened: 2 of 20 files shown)


apps/dotcom/client/src/tla/app/zero-polyfill.ts
```diff
@@ -179,14 +179,7 @@ export class Zero {
...
 					throw new Error('file already exists')
 				}
-				this.makeOptimistic([
-					{ table: 'file', event: 'insert', row: data },
-					{
-						table: 'file_state',
-						event: 'insert',
-						row: { fileId: data.id, userId: store.user.id, firstVisitAt: Date.now() } as any,
-					},
-				])
```
ds300, line 189: remember we had that "cannot update file state for another user: undefined" bug in sentry? this is why. I hoisted this out into the TldrawApp.createFile method and made sure it supplies all the right properties

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -434,166 +570,78 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
-
-					// alright we're good to go
-					this.messageUser(row.userId, {
-						type: 'row_update',
-						row: file as any,
-						table: 'file',
-						event: 'insert',
-						userId: row.userId,
-					})
-				})
```
ds300, line 562: it is no longer practical to send random updates to the user like this, so I made it the responsibility of the UserDataSyncer to fetch and delete the file records for guest files on an ad-hoc basis.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -196,84 +196,35 @@ export class UserDataSyncer {
...
-				this.commitMutations(mutationNumber)
-			} else {
-				this.broadcast({
-					type: 'initial_data',
-					initialData: this.state.data,
-				})
-			}
+	private async loadInitialDataFromR2() {
+		this.log.debug('loading snapshot from R2')
+		const res = await this.env.UPLOADS.get(this.getSnapshotKey())
```
MitjaBezensek, line 201: I'd feel more comfortable if we didn't reuse the uploads bucket. We have some api routes that allow fetching stuff from the bucket, so I'm afraid we might expose these snapshots down the line.

(for now it seems fine, using an url like this doesn't seem to work, but only because our router does not match the path https://pr-5497-preview-deploy.tldraw.com/api/app/uploads/user_do_snapshots/user_id).

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -370,47 +399,99 @@ export class UserDataSyncer {
...
+				}
+
+				assert(ev.type === 'row_update', `event type should be row_update got ${event.type}`)
+				this.handleRowUpdateEvent(ev)
+			}
+
+			this.store.commitLsn(event.lsn)
+		})
+
+		// make sure we have all the files we need
```
MitjaBezensek, line 430: When would these two cases happen?

Also, do we want to return in the if? feels like we'd want to add all of the missing files?

ds300: 1. for adding a missing guest file record, this would happen if the user adds a file_state but something went wrong when adding the guest file record after the mutation was applied. unlikely but better safe than sorry I reckon.

2. for deleting a guest file record that is no longer needed, this currently happens mainly if a shared file is set to private (we delete the file_state for guests in that case)

and no that return in the if shouldn't be there!

MitjaBezensek: Where do we delete the `file_state`, must have missed that?

ds300: https://github.com/tldraw/tldraw/blob/facfa5d1689b7722035e645627b63b5bcd78d011/apps/dotcom/zero-cache/migrations/000_seed.sql#L59-L66

MitjaBezensek: Thanks, completely forgot about this 😄

apps/dotcom/sync-worker/src/getFetchEverythingSql.ts
```diff
@@ -45,7 +45,7 @@ export function getFetchUserDataSql(userId: string) {
 	return sql`
 SELECT 'user' AS "table", null::text as "lsn", ${sql.raw(userColumns + ',' + fileNulls + ',' + fileStateNulls)} FROM public.user WHERE "id" = '${sql.raw(userId)}'
 UNION
-SELECT 'file' AS "table", null::text as "lsn", ${sql.raw(userNulls + ',' + fileColumns + ',' + fileStateNulls)} FROM public.file WHERE "isDeleted" = false AND ("ownerId" = '${sql.raw(userId)}' OR "shared" = true AND EXISTS(SELECT 1 FROM public.file_state WHERE "userId" = '${sql.raw(userId)}' AND public.file_state."fileId" = public.file.id))
+SELECT 'file' AS "table", null::text as "lsn", ${sql.raw(userNulls + ',' + fileColumns + ',' + fileStateNulls)} FROM public.file WHERE ("ownerId" = '${sql.raw(userId)}' OR "shared" = true AND EXISTS(SELECT 1 FROM public.file_state WHERE "userId" = '${sql.raw(userId)}' AND public.file_state."fileId" = public.file.id))
```
ds300, line 48: @MitjaBezensek I reverted this change from the other day because i needed to undelete a file for someone recently and it was a pain to figure out how to do it. After this change and the bits above in TldrawApp.ts we will be able to undelete files for people by just toggling 'isDeleted' in the backend. Yes it will mean sending people useless old deleted file records but that's a pretty low overhead for the convenience I think. With zero we should be able to solve this cleanly.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
ds300: yeah I meant to make a new bucket before submitting for review but it slipped my mind. will do that now.

apps/dotcom/sync-worker/src/getFetchEverythingSql.ts
MitjaBezensek: Sounds good. We do filter them out in the zero polyfill so they do not affect the max files calculation, so all good there.

MitjaBezensek (approved):

PR #5523, opened 2025-02-28
fix file extension for downloads

ds300:
before this, the 'Download file' option would return with a .json extension.

### Change type

- [x] `other`


MitjaBezensek (approved):
Nice catch.

PR #5531, opened 2025-03-03
Add a tests for guest files

MitjaBezensek:
Make sure they are visible in the sidebar, that we show a guest badge, and that the correct share menu tabs are visible.

### Change type

- [x] `improvement`


apps/dotcom/client/e2e/tests/sharing.spec.ts
```diff
@@ -23,11 +23,6 @@ test.describe('signed in user on own file', () => {
 		expect(await shareMenu.exportTabButton.isVisible()).toBe(true)
 		expect(await shareMenu.anonShareTabButton.isVisible()).toBe(false)
 
-		expect(await shareMenu.publishTabPage.isVisible()).toBe(false)
```
MitjaBezensek, line 26: These are the same as the ones just below.

PR #5532, opened 2025-03-03
Make better use of replicator history

ds300:
After making a couple of deploys with the new backend this morning we noticed that a large portion of user DOs still ended up doing a full db reset and this seems to be for two reasons:

1. we only stored about 10 minutes worth of history
2. a lot of users in our active_users table are idle at any given time (maybe like 70-80% of them) so a lot of them will not have received updates for more than 10 minutes.

This PR makes it so that if a user hasn't done anything for 5 minutes but their socket is still open, we execute a noop mutation number bump so that they receive a new `lsn`.

I also cleaned up the mutation handling logic so that it no longer does two transactions per mutation. This was only required before because postgres.js would not dispatch events grouped by transaction, and the ordering was not maintained. wal2json guarantees changes to be grouped by transaction with formatVersion: 1 (the default that we use) so as long as we commit the mutations after handling any state updates it should avoid data races.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -115,7 +115,7 @@ const migrations: Migration[] = [
 
 const ONE_MINUTE = 60 * 1000
 const PRUNE_INTERVAL = 10 * ONE_MINUTE
-const MAX_HISTORY_ROWS = 10_000
+const MAX_HISTORY_ROWS = 50_000
 
 type PromiseWithResolve = ReturnType<typeof promiseWithResolve>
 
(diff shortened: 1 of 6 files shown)


apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -863,7 +863,9 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 			this.reportActiveUsers()
 			this.log.debug('inserted active user')
 
+			const resumeStart = Date.now()
 			const resume = this.getResumeType(lsn, userId, guestFileIds)
+			this.logEvent({ type: 'resume_duration', duration: Date.now() - resumeStart })
```
ds300, line 868: ope, just remembered that this won't work when deployed remotely due to the security thing cloudflare do. dang.

MitjaBezensek (approved):
Nice! Really interested to see the effect of this one on redeploys.

Issue #5539, opened 2025-03-03
[Bug]: Chrome + IME: Enter key causes text duplication in TldrawUiInput component


banqinghe:
### What happened?

The issue I encountered is that in Chrome, using the Chinese input method, pressing the Enter key while inputting in the TldrawUiInput component causes text to be duplicated.

As shown in the gif below, my key sequence is `a` → `Enter`. The expected behavior should be that the input method exits and inputs the letter `a`, but the actual effect is that two `a` are entered.

![Image](https://github.com/user-attachments/assets/a8a03652-18b8-47b9-9fcd-f97f671bd582)

I have three different browsers on my computer: Chrome, Safari, and Zen. After testing, I found that only Chrome has this issue, which may be related to its internal event logic handling.

I found that the reason for this issue is that in Chrome, if `e.currentTarget.blur()` is executed in the `<input>`'s `onKeyDownCapture`, it will trigger an incorrect `onChange` event with duplicate text data when the input method exits.

By listening to the [composition event](https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent), it can be determined whether the current `Enter` was pressed during IME execution, thus avoiding this issue.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The tldraw.com website

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #5540, opened 2025-03-03
fix: prevent text duplication when using IME with Enter key in Chrome

banqinghe:
This PR prevents text duplication when using Input Method Editors (IME) with the Enter key in Chrome by handling the composition state correctly in the TldrawUiInput component.

Fixes https://github.com/tldraw/tldraw/issues/5539

before:

tldraw_issue

after

tldraw_pr

### Change type

- [x] `bugfix`

### Test plan

1. Open tldraw in Chrome.
2. Use an IME (e.g., Japanese or Chinese) to type text into an input.
3. Press Enter to confirm the composition.
4. Verify that the text is not duplicated.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where using an IME with the Enter key in Chrome could cause text duplication in UI inputs.

### DIFF
--- a/packages/editor/src/lib/utils/browserCanvasMaxSize.ts
+++ b/packages/editor/src/lib/utils/browserCanvasMaxSize.ts
@@ -19,9 +19,11 @@ function getBrowserCanvasMaxSize(): CanvasMaxSize {
 	return maxCanvasSizes
 }
 
-// Extracted from https://github.com/jhildenbiddle/canvas-size
-// MIT License: https://github.com/jhildenbiddle/canvas-size/blob/master/LICENSE
-// Copyright (c) John Hildenbiddle
+/*!
+ * Extracted from https://github.com/jhildenbiddle/canvas-size
+ * MIT License: https://github.com/jhildenbiddle/canvas-size/blob/master/LICENSE
+ * Copyright (c) John Hildenbiddle
+ */
 
 const MAX_SAFE_CANVAS_DIMENSION = 8192
 const MAX_SAFE_CANVAS_AREA = 4096 * 4096
--- a/packages/tldraw/src/lib/utils/tldr/file.ts
+++ b/packages/tldraw/src/lib/utils/tldr/file.ts
@@ -9,11 +9,14 @@ import {
 	SerializedSchemaV1,
 	SerializedSchemaV2,
 	SerializedStore,
+	StoreSnapshot,
 	T,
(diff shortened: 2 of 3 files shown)


banqinghe:

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
```diff
@@ -70,6 +70,8 @@ export const TldrawUiInput = React.forwardRef<HTMLInputElement, TLUiInputProps>(
 		const rInitialValue = React.useRef<string>(defaultValue ?? '')
 		const rCurrentValue = React.useRef<string>(defaultValue ?? '')
 
+		const isCompositioning = React.useRef(false)
```
mimecuvalo, line 73: nit: `isComposing`

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
```diff
@@ -176,6 +179,8 @@ export const TldrawUiInput = React.forwardRef<HTMLInputElement, TLUiInputProps>(
 					onChange={handleChange}
 					onFocus={handleFocus}
 					onBlur={handleBlur}
+					onCompositionStart={() => (isCompositioning.current = true)}
```
mimecuvalo, line 182: let's create some handlers above using `useCallback` to avoid any possible re-renders

packages/tldraw/src/lib/ui/components/primitives/TldrawUiInput.tsx
```diff
@@ -108,6 +110,7 @@ export const TldrawUiInput = React.forwardRef<HTMLInputElement, TLUiInputProps>(
 			(e: React.KeyboardEvent<HTMLInputElement>) => {
 				switch (e.key) {
 					case 'Enter': {
+						if (isCompositioning.current) return
```
mimecuvalo, line 113: Since this is a specific fix for Chrome, let's add a comment here specifying why this is necessary. It'd be great to mention that it's a Chrome issue, specifically tied with the `e.currentTarget.blur()` below that causes the double-text.

mimecuvalo (commented):
thank you for this PR and bringing the issue to our attention! couple comments before we land this.

mimecuvalo (approved):
ty!

PR #5541, opened 2025-03-03
Make collaboration hooks public

MathieuLoutre:
Makes `usePeerIds` and `usePresence` public so they can be used to customise collaboration components more easily.

I left the comment that maybe these two hooks should be computed properties, but if they're made public that may not be needed anymore. Unless it's better to make those computed properties right now and skip the public hooks?

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

Check that `usePeerIds` and `usePresence` can be accessed. Maybe an example?

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Makes `usePeerIds` and `usePresence` public

### DIFF
--- a/apps/docs/content/getting-started/releases-versioning.mdx
+++ b/apps/docs/content/getting-started/releases-versioning.mdx
@@ -15,72 +15,47 @@ Unlike many JavaScript packages distributed on [NPM](https://www.npmjs.com/), th
 
 {/* START AUTO-GENERATED CHANGELOG */}
 
-## Current release: [v3.8.0](/releases/v3.8.0)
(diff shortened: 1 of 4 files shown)


MitjaBezensek:
Looking good, thanks 👏 

The one thing we'd probably want to change is to revert the `releases-versioning.mdx` change.

MathieuLoutre:
Reverted that change – didn't see the file was auto-generated sorry!

MitjaBezensek:
> Reverted that change – didn't see the file was auto-generated sorry!

Thanks!

PR #5542, opened 2025-03-03
CTA analytics 

TodePond:
- Updated all dot dev links to use UTMs
  - In READMEs
  - On tldraw.com
  - In blog posts, eg: tldraw sync announcement
- Improved consistency of discord UTMs
- Removed unused GitHub discussions links

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,8 +1,8 @@
 # Frequently Asked Questions
 
-This document is meant to be provide answers to frequently asked questions about the [tldraw developer SDK](https://tldraw.dev). For questions related to **[tldraw.com](https://tldraw.com)**, please see our [tldraw.com guides](https://tldraw.notion.site/tldraw-com-1283e4c324c08086b585d438f088580f).
+This document is meant to be provide answers to frequently asked questions about the [tldraw developer SDK](https://tldraw.dev/?utm_source=github&utm_medium=repo&utm_campaign=faq). For questions related to **[tldraw.com](https://tldraw.com)**, please see our [tldraw.com guides](https://tldraw.notion.site/tldraw-com-1283e4c324c08086b585d438f088580f).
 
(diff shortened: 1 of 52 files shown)


steveruizok:
I don't think I understand the function of our UTMs here.

My understanding of UTMs is that they are a way to capture the success of a specific marketing initiative that may be spread over multiple mediums (e.g. blog posts, twitter, emails) in cases where referrer-based tracking wouldn't be enough. So that we can say, hey, our launch of the new tldraw washing machine had a ton of visitors and a lot of it came from our twitter videos.

The changes here seem like adding "UTMs for UTM's sake". Is this providing useful data for us? For example, is the `utm_source=github&utm_medium=readme&utm_campaign=learn-more` campaign useful if it's being used across all of our READMEs?

image

What I mean is: let's use UTMs more intentionally. They should be used in our marketing materials where we know which links we're posting. I don't think we need every link to tldraw.dev to have a UTM parameter, especially if the referrer is enough to tell us what we need to know. If we over-commit to this type of thing in places where it doesn't matter, then we'll inevitably not follow through with the links and so they'll be partial / useless anyway.

TodePond:
That's a fair concern.

In cases like GitHub readmes, I think we can do without the UTMs, as we do seem to get good referrer info there anyway, and you're right: We're likely to drop the ball on that.

I'd still like to keep the UTMs in some other places where it can help us to differentiate between sources that we don't get from referrer info. For example, we can't currently tell if someone came from clicking the "About" button versus the watermark on tldraw.com currently. I'd like to use UTMs for cases like that to help us learn how people make their way to dot dev. 

I'll go ahead and remove the UTMs from GitHub readmes.

TodePond:
Thanks @steveruizok!  See my [above comment](https://github.com/tldraw/tldraw/pull/5542#issuecomment-2697696859) for my thoughts.

- I've now removed UTMs from all github readmes and pages.
- I've removed UTMs from generic uses, like release notes.
- I've kept UTMs for specific campaigns.
- I've kept UTMs where we struggle to determine the source (eg: "About" button versus watermark on dotcom)
- I've kept UTMs for discord.tldraw.com links. We do get referrer info, but it's more limited than what we get for dotdev links, so I'd to keep using UTMs where we can for that.

Let me know what you think!

PR #5545, opened 2025-03-04
fix pasting files in safari

bluedot74:
This PR fixes a small bug with pasting files from your computer into Safari.
It only addresses pasting with Cmd+V, I don't know if there is any hope for the Menu paste in Safari.

It uses the existing "fallbackFiles" which appear to be designed to handle this but were only handling text files?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. On `main` copy a supported file from your local computer, such as a PNG.
2. In the editor, in Safari, press Cmd/Ctrl + V.
3. Note that nothing happens.
4. Repeat the process on this branch and the file[s] should be loaded into the editor as expected.

### Release notes

- Fixed a bug with pasting files from your computer in Safari


SomeHats:
looks good! thank you :)

PR #5548, opened 2025-03-04
setup pgbouncer for local dev (fix flaky e2e tests?)

ds300:
1. Running tests in parallel seems to be prone to slowness due to limits on db connections. For me adding pgbouncer improves things quite a bit.
1. If the page reloads before a mutation has been sent over the socket, the 'before and after reload' assertion helper we have can fail. So I've tentatively added a thing for the Sidebar operations where after making a mutation it waits for the optimistic updates log in the client to clear before continuing.
2. cleaned up some defunct infra code

### Change type

- [x] `bugfix`


MitjaBezensek (approved):
Kinda hate that we have to do stuff like this. Hopefully we can get rid of it soon 😄

PR #5554, opened 2025-03-04
events: explore using getCoalescedEvents

mimecuvalo:
Wanted to open a discussion on `getCoalescedEvents` and see if it's worth integrating into the canvas. The videos below show trying to draw with a CPU 20x slowdown.

(I know we've talked about `getPredictedEvents` but that's a separate topic)
Testbed: https://patrickhlauke.github.io/touch/coalesced-predicted-events/

Before:

https://github.com/user-attachments/assets/f4b0648e-22e5-43a3-874a-959d6011ada7

After:

https://github.com/user-attachments/assets/6a99db52-98e4-4a5c-aaab-9ccc1b970e9c

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improve draw fluidity on slower CPUs by using getCoalescedEvents.

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -12,7 +12,7 @@ import { useEditor } from './useEditor'
 
 export function useCanvasEvents() {
 	const editor = useEditor()
-	const currentToolId = useValue('current tool id', () => editor?.getCurrentToolId(), [editor])
+	const currentTool = useValue('current tool', () => editor.getCurrentTool(), [editor])
 
 	const events = useMemo(
 		function canvasEvents() {
@@ -53,9 +53,7 @@ export function useCanvasEvents() {
 
 				// For tools that benefit from a higher fidelity of events,
 				// we dispatch the coalesced events.
-				const events = ['draw', 'highlight'].includes(currentToolId)
-					? e.nativeEvent.getCoalescedEvents()
-					: [e]
+				const events = currentTool.useCoalescedEvents ? e.nativeEvent.getCoalescedEvents() : [e]
 				for (const singleEvent of events) {
 					editor.dispatch({
 						type: 'pointer',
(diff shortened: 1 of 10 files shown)


SomeHats:
this is so fucking cool??????

SomeHats:
so, thinking about this a bit more...  do we know of a reason we shouldn't do this?

mimecuvalo:
> so, thinking about this a bit more... do we know of a reason we shouldn't do this?

There could be a hesititation around when there's a lot of shapes being manipulated, say, selecting a lot of shapes on the canvas and moving them around all at once, maybe that affects perf? @MitjaBezensek wdyt?

Or, well, say using the Hand tool and doing the same, moving the canvas around.

MitjaBezensek:
I think this definitely makes sense for events where high precision brings some benefits - like while drawing since it allows us to make more drawings that more closely follow the user's input. For other cases it might actually be worse. The fact that browser coalesces events means that it determined the the device is not capable enough to process all of them, so it starts coalescing them.

We did run into some problems when we tried something like this before, see this PR, especially the coalescing part:
https://github.com/tldraw/tldraw/pull/3223

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -49,8 +51,12 @@ export function useCanvasEvents() {
 				lastX = e.clientX
 				lastY = e.clientY
 
-				const coalescedEvents = e.nativeEvent.getCoalescedEvents()
-				for (const singleEvent of coalescedEvents) {
+				// For tools that benefit from a higher fidelity of events,
+				// we dispatch the coalesced events.
+				const events = ['draw', 'highlight'].includes(currentToolId)
+					? e.nativeEvent.getCoalescedEvents()
+					: [e]
```
SomeHats, line 58: could we do this in a more extensible way rather than hard coding it? ie a property on the tool's state node?

mimecuvalo: yes, definitely — lemme cook that up 👍

mimecuvalo:
I spoke with @MitjaBezensek about this offline about his previous concerns and we should be good here.

Landing!

PR #5557, opened 2025-03-05
Migrate newsletter form to hubspot

MitjaBezensek:
Submit newsletter subscribers to hubspot instead of sendgrid.

The reason why we don't use a straight api call to add a contact is that you [can't set the contact to be a marketing contact](https://community.hubspot.com/t5/APIs-Integrations/How-do-I-set-a-contact-as-a-marketing-contact-using-Hubspot-s/m-p/895454) via an API. So I created a form in hubspot and we then [submit to that form](https://community.hubspot.com/t5/APIs-Integrations/Cannot-set-a-contact-s-marketing-status-via-API/m-p/800944/highlight/true#M64478). 

An alternative approach would be to use hubspot's form directly, but since this form is used in a couple of different places with different styling options it would make it a bit annoying to correctly style.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/docs/components/common/newsletter-signup.tsx
+++ b/apps/docs/components/common/newsletter-signup.tsx
@@ -33,12 +33,12 @@ export function NewsletterSignup({
 			setFormState('loading')
 			try {
 				const _email = new FormData(e.currentTarget)?.get('email') as string
-				const cookie = document.cookie
+				const hubspotCookie = document.cookie
 					.split('; ')
 					.find((row) => row.startsWith('hubspotutk='))
 					?.split('=')[1]
 				const url = window.location.href
-				const { error } = await submitNewsletterSignup(_email, cookie, url)
+				const { error } = await submitNewsletterSignup(_email, hubspotCookie, url)
 				if (error) throw error
 				setFormState('success')
 				// After a pause, we locally save that the user has submitted the form
--- a/apps/docs/components/common/newsletter-signup.tsx
+++ b/apps/docs/components/common/newsletter-signup.tsx
@@ -33,7 +33,11 @@ export function NewsletterSignup({
 			setFormState('loading')
 			try {
 				const _email = new FormData(e.currentTarget)?.get('email') as string
-				const { error } = await submitNewsletterSignup(_email)
+				const cookie = document.cookie
+					.split('; ')
+					.find((row) => row.startsWith('hubspotutk='))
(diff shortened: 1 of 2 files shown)


apps/docs/components/common/newsletter-signup.tsx
```diff
@@ -33,7 +33,12 @@ export function NewsletterSignup({
 			setFormState('loading')
 			try {
 				const _email = new FormData(e.currentTarget)?.get('email') as string
-				const { error } = await submitNewsletterSignup(_email)
+				const cookie = document.cookie
```
MitjaBezensek, line 36: I guess if the users allowed cookies well get it otherwise it should not be there.

mimecuvalo: just a nit here: we should call the variable hubspotCookie to. make it clear this isn't the whole cookie.

apps/docs/scripts/submit-newsletter-signup.ts
```diff
@@ -1,21 +1,36 @@
...
+			body: JSON.stringify({
+				submittedAt: Date.now(),
+				fields: [
+					{
+						name: 'email',
+						value: email,
+					},
+				],
+				context: {
+					hutk: cookie,
```
mimecuvalo, line 25: up above it's `hubspotutk` - does the context matter here that it's the same name or is `hutk` just fine?

mimecuvalo (approved):
approved, couple nits maybe

apps/docs/components/common/newsletter-signup.tsx
MitjaBezensek: Done 👍

apps/docs/scripts/submit-newsletter-signup.ts
MitjaBezensek: Feels like there are some inconsistencies on hubspots side:
> **hutk**
Body param
Used in context. The tracking cookie token value used for HubSpot lead activity tracking. You can retrieve this value from the "**hubspotutk**" cookie placed in the user's browser by the HubSpot JavaScript Tracking Code; the tracking code must be installed on the page that contains the form.
Please Note: While the **hutk** value is not required for the submission to be accepted, this token is used to associate analytics data with a contact record, so without this you will not see any page views or analytics source data for the contact record.

PR #5560, opened 2025-03-05
fix(ui): use useMaybeEditor in ui context to prevent remounts

ds300:
So we were facing a weird problem in dotcom where the editor would remount immediately after mounting for the first time, causing the room data to be fetched twice (once before initial editor mount, and once again on remount).

We tracked it down to this [UI context provider](https://github.com/tldraw/tldraw/blob/c661dfedf191a5d9b9fd168a881a99ab7e719f6b/packages/tldraw/src/lib/ui/context/TldrawUiContextProvider.tsx#L79-L86), and specifically the fact that reparenting this react-subtree inside a new context provider after the editor becomes available causes the whole thing to be remounted and the editor reinstantiates.

A quick and dirty way to fix this is to make the provider tree static and make sure that all components within it use `useMaybeEditor` so the app doesn't crash if there isn't an editor available.

The slow and clean way around this is to refactor to not have this UIContextProvider be so coupled to the Editor, but this is a bad perf issue that needs fixing asap, it's very noticeable for larger files.

### Change type

- [x] `bugfix` 

### Test plan

1. Open a large file in dotcom
2. Verify that the editor does not remount and room data is only fetched once

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a performance issue where the editor would remount on initial load.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLegacytSetDocumentTitle.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLegacytSetDocumentTitle.tsx
@@ -1,11 +1,11 @@
 import { Helmet } from 'react-helmet-async'
 import { useValue } from 'tldraw'
-import { globalEditor } from '../../../../utils/globalEditor'
+import { useGlobalEditor } from '../../../../utils/globalEditor'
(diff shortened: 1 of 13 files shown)


SomeHats (approved):
makes sense!

PR #5562, opened 2025-03-06
[dotcom] fix data leaking between preview branches

ds300:
Our user DO snapshots were leaking between branches 😬 

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/types.ts
+++ b/apps/dotcom/sync-worker/src/types.ts
@@ -69,6 +69,11 @@ export function isDebugLogging(env: Environment) {
 	return env.TLDRAW_ENV === 'development' || env.TLDRAW_ENV === 'preview'
 }
 
+export function getUserDoSnapshotKey(env: Environment, userId: string) {
+	const snapshotPrefix = env.TLDRAW_ENV === 'preview' ? env.WORKER_NAME + '/' : ''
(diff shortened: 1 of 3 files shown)


apps/dotcom/sync-worker/src/types.ts
```diff
@@ -69,6 +69,11 @@ export function isDebugLogging(env: Environment) {
 	return env.TLDRAW_ENV === 'development' || env.TLDRAW_ENV === 'preview'
 }
 
+export function getUserDoSnapshotKey(env: Environment, userId: string) {
+	const snapshotPrefix = env.TLDRAW_ENV === 'preview' ? env.WORKER_NAME + '/' : ''
```
MitjaBezensek, line 73: Should we also add some prefix for staging and production? Collision chance is very low, but might help if we ever want to clean just one of those but not the other.

MitjaBezensek (approved):

apps/dotcom/sync-worker/src/types.ts
ds300: staging and production have their own individual R2 buckets so there's no chance of leaking between them

MitjaBezensek: Ah right.

PR #5563, opened 2025-03-06
Add some randomness to prevent this happening in sync for online users.

MitjaBezensek:
Should help with having spiky traffic like the one during the last deploy:

image

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
@@ -29,6 +29,7 @@ import { getRoomDurableObject } from './utils/durableObjects'
 import { isRateLimited } from './utils/rateLimit'
 import { retryOnConnectionFailure } from './utils/retryOnConnectionFailure'
 
+const ONE_MINUTE = 60 * 1000
 export class TLUserDurableObject extends DurableObject<Environment> {
 	private readonly db: Kysely<DB>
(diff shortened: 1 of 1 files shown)


apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -125,8 +125,12 @@ export class TLUserDurableObject extends DurableObject<Environment> {
...
 				// do cache persist + cleanup
 				this.cache?.onInterval()
-				// do a noop mutation every 5 minutes
-				if (Date.now() - this.lastMutationTimestamp > 5 * 60 * 1000) {
+				// do a noop mutation every 5-6 minutes
+				// randomness prevents this happening in sync for all users that are online during reboot
+				const oneMinute = 60 * 1000
+				const offset = Math.random() * oneMinute
+				const interval = 5 * oneMinute + offset
+				if (Date.now() - this.lastMutationTimestamp > interval) {
```
ds300, line 133: not sure how effective this will be because this is evaluated once per second i expect we'll still get spikes after deploys but instead of being clustered around 1-2 seconds they'll be clustered around 4-6 seconds.

I was thinking of something more like setting the initial `lastMutationTimestamp` to be Date.now() + random number of milleseconds in the range of 2-7 minutes.

ds300 (approved):

PR #5565, opened 2025-03-06
fix reparentshapes preserve order

dodo-Riley:
When multiple shapes are selected and dragged over or out of a frame shape, they get reparented. However, during this process, the order of the selected shapes changes.

The issue arises because, inside reparentShapes(), the indices from indices are simply assigned to shapesToReparent in a loop. To preserve the original order, shapesToReparent should be sorted by their index before the loop runs. This ensures that new indices are assigned in the correct order.

I am attaching a video demonstrating the issue.

[without sorting]
https://github.com/user-attachments/assets/81130860-1578-476c-ba69-fe078ad44406

[with sorting]
https://github.com/user-attachments/assets/abb5f6a2-ddbb-4d80-9efd-2f416a1d6cb2

To fix the issue, modifying the reparentShapes() function as follows should work:

Change this line:
const shapesToReparent = compact(ids.map((id) => this.getShape(id)));
To:
const shapesToReparent = compact(ids.map((id) => this.getShape(id))).sort(sortByIndex);

This ensures that the shapes are sorted by their index before being processed, preserving their original selection order during reparenting.

[build result]

[test result]

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Add a frame shape along with multiple other shapes.
2. Select multiple shapes one by one while holding the Shift key to enter multi-select mode.
3. Drag the multi-selected shapes over the frame shape.

- [v] Unit tests
- [v] End to end tests

### Release notes

- Modify the reparentShapes() function to ensure that the original order of the shapes is preserved when reparenting.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5342,7 +5342,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		const invertedParentTransform = parentTransform.clone().invert()
 
-		const shapesToReparent = compact(ids.map((id) => this.getShape(id)))
+		const shapesToReparent = compact(ids.map((id) => this.getShape(id))).sort(sortByIndex)
 
 		// Ignore locked shapes so that we can reparent locked shapes, for example
 		// when a locked shape's parent is deleted.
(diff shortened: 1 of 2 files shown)


dodo-Riley:

MitjaBezensek:
Thanks! 👏

PR #5567, opened 2025-03-06
Use ImmutableMap inside AtomMap

ds300:
This PR fixes load perf for large files by updating `AtomMap` to use an immutable map (taken from Immutable.js).

On a large room like our old retro board, populating the `AtomMap` would take up more than a second in local dev on my machine, this brings it down to < 15ms. The cause for the slowness was updating the presentKeys `Set` via copy-on-write.

This also fixes another issue with AtomMap in that it didn't respect signia transaction rollbacks because it used a mutable Map to store the atoms.

### Change type

- [x] `bugfix`

### Release notes

- Fixed a perf regression that caused slowness mainly when loading documents.

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -41,7 +41,6 @@
 		"lint": "yarn run -T tsx ../../internal/scripts/lint.ts"
 	},
 	"dependencies": {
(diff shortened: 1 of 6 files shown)


SomeHats (approved):
this is amazing, ty!

PR #5586, opened 2025-03-08
Add page navigation kbds

steveruizok:
This PR adds shortcuts for moving between pages. Option + ArrowLeft will go to the prev page (eg from Page 2 to Page 1), Option + ArrowRight will go to the next page (from Page 1 to Page 2).

If the current page is NOT blank, then going to the next page from the last page will create a new page.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create some pages
2. Use the keyboard to navigate between pages
3. On a last page with shapes, create a new page by going "up" from the last page
3. On a last page without shapes, fail to create a new page by going "up" from the last page

### Release notes

- Added keyboard shortcuts (option + arrows) for navigating between pages.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
@@ -140,38 +140,56 @@ export function useKeyboardShortcuts() {
 	}, [actions, tools, isReadonlyMode, editor, isFocused])
 }
 
-// Shift is !
-// Alt is ?
-// Cmd / control is $
-// so cmd+shift+u would be $!u
+// convenience mapping since all real keys need to be single characters, sorry
+const MAPPED_KEYS: Record<string, string> = {
+	'→': 'right',
+	'←': 'left',
+}
+function getMappedKey(kbd: string) {
+	if (MAPPED_KEYS[kbd]) return MAPPED_KEYS[kbd]
+	return kbd
+}
 
 function getHotkeysStringFromKbd(kbd: string) {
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
```diff
@@ -140,39 +140,39 @@ export function useKeyboardShortcuts() {
...
+// up by comma, then parse each key to get the actual key and modifiers.
 function getHotkeysStringFromKbd(kbd: string) {
 	return getKeys(kbd)
 		.map((kbd) => {
 			let str = ''
-			const chars = kbd.split('')
-			if (chars.length === 1) {
-				str = chars[0]
+
+			const shift = kbd.includes('!')
```
mimecuvalo, line 151: i somehow didn't realize that the `!?$` was our own homebrew choices for these action characters. can we change these as long as your updating this code? it's super arcane/difficult to read, especially when trying to update the `actions.tsx` file.

if we can use `⇧ ⌥ ⌘` Unicode symbols that _massively_ help DX here and maintainability

steveruizok: IIRC it was due to the weirdness around option / alt and cmd / ctrl. If we can, mind if we do it as a separate PR?

mimecuvalo: yes, please!

PR #5587, opened 2025-03-09
Add eslint rule for "whilst"

steveruizok:
Every now and then the word "whilst" will appear in our codebase. This PR enforces an eslint rule that prevents it from appearing and suggests "while" as a fix.

We use American english and spelling in our codebase and communications. While whilst is common here in the UK (where our development team is) in the US the word sounds intentionally affected, like thou or thee or m'lady. It is absolutely critical that we avoid any perception of whimsy.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Use 'whilst' in a comment
2. Linting should fail

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -148,6 +148,7 @@ export default [
 			'local/tagged-components': 'error',
 			'local/prefer-class-methods': 'error',
 			'local/tsdoc-param-matching': 'error',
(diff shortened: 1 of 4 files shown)


mimecuvalo:
dear lordeth

PR #5603, opened 2025-03-11
perf: improve being able to pan around in larger rooms

mimecuvalo:
When I was looking at coalesced events, that had me slowdown my CPU more than usual. It made me notice that panning when there was a lot of shapes was chunky. Digging into the flamegraph, I noted that `updateHoveredShapeId` was a pretty big culprit. This splits off that function into a throttled or debounced version depending on context of what's happening on the canvas. 

Before (note FPS in corner):

https://github.com/user-attachments/assets/6f07c761-c16c-47e3-8dff-2ba1016203e0

After (note FPS in corner):

https://github.com/user-attachments/assets/8a1c3731-a327-40a3-8954-7b4916a64956

(Mitja, I tagged you since you've done perf stuff a lot but don't review while you're off!)

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve performance when panning in a large room with a lot of shapes.

### DIFF
--- a/packages/tldraw/src/lib/tools/selection-logic/updateHoveredShapeId.ts
+++ b/packages/tldraw/src/lib/tools/selection-logic/updateHoveredShapeId.ts
@@ -1,4 +1,4 @@
-import { Editor, TLShape, throttle } from '@tldraw/editor'
+import { Editor, TLShape, debounce, throttle } from '@tldraw/editor'
 
 function _updateHoveredShapeId(editor: Editor) {
 	// todo: consider replacing `get hoveredShapeId` with this; it would mean keeping hoveredShapeId in memory rather than in the store and possibly re-computing it more often than necessary
@@ -32,7 +32,13 @@ function _updateHoveredShapeId(editor: Editor) {
 }
 
 /** @internal */
-export const updateHoveredShapeId = throttle(
+export const updateHoveredShapeIdThrottled = throttle(
+	_updateHoveredShapeId,
+	process.env.NODE_ENV === 'test' ? 0 : 32
+)
+
+/** @internal */
+export const updateHoveredShapeIdDebounced = debounce(
 	_updateHoveredShapeId,
 	process.env.NODE_ENV === 'test' ? 0 : 32
 )
--- a/packages/tldraw/src/lib/shapes/text/toolStates/Idle.ts
+++ b/packages/tldraw/src/lib/shapes/text/toolStates/Idle.ts
@@ -1,5 +1,5 @@
 import { StateNode, TLKeyboardEventInfo, TLPointerEventInfo } from '@tldraw/editor'
-import { updateHoveredShapeId } from '../../../tools/selection-logic/updateHoveredShapeId'
+import { updateHoveredShapeIdThrottled } from '../../../tools/selection-logic/updateHoveredShapeId'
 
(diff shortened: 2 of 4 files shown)


steveruizok:
Hm, it should be really fast to get a hovered shape, <1 ms. I'd prefer to make that function faster rather than showing the wrong hovered shape during scroll or pan due to a debounce.

On my machine, the function is slow because we're rendering text as part of finding the hovered shape.

image

steveruizok:
Text perf fixed here: https://github.com/tldraw/tldraw/pull/5658

mimecuvalo:
> Hm, it should be really fast to get a hovered shape, <1 ms. I'd prefer to make that function faster rather than showing the wrong hovered shape during scroll or pan due to a debounce.

It's not the _getting_ of the hovered shape, it's the _setting_ of the hovered shape that's the issue when panning.

> On my machine, the function is slow because we're rendering text as part of finding the hovered shape.

For sure, this was an issue as well, but `updateHoveredShapeId` is still a problem here.

Combining testing with the text perf changes in the other PR, this PR still doubles the perf when panning around. So, we should still address the issue raised here too. Happy to show you offline - on my good machine I see an FPS increase from ~25fps to ~55fps

steveruizok:
Coming to the same conclusions here: updating the hovered shape should be a lower throttle while moving the camera (or based on pointer velocity).

We've improved a bunch of places that were making geometry etc slow... but I still feel like getting hovered shape is too slow.

mimecuvalo:
cool cool. i rebased this PR so that's up-to-date again. we can iterate on it or take a different tact, just lemme know.

packages/tldraw/src/lib/tools/selection-logic/updateHoveredShapeId.ts
```diff
@@ -32,7 +32,16 @@ function _updateHoveredShapeId(editor: Editor) {
 }
 
 /** @internal */
-export const updateHoveredShapeId = throttle(
+export const updateHoveredShapeIdThrottled = throttle(
```
SomeHats: wow i didn't realise we did this. this will not play nice with multiple editors on the page... Can we scope the throttling / cancelling to a single editor with a `WeakCache`? We can keep these as the exported functions, but they should look up editor-scoped throttle/debounce versions instead of global ones

packages/tldraw/src/lib/tools/selection-logic/updateHoveredShapeId.ts
```diff
@@ -32,7 +32,16 @@ function _updateHoveredShapeId(editor: Editor) {
...
 	_updateHoveredShapeId,
 	process.env.NODE_ENV === 'test' ? 0 : 32
 )
+
+/** @internal */
+export const updateHoveredShapeIdDebounced = debounce(
```
SomeHats: throttle vs debounce is one of those things that can take a bit of thinking to really understand the difference between. Are there names we can choose here that make it clearer which option should be used in which scenario? If not, could we add doc comments explaining which should be used in which scenario and why?

PR #5605, opened 2025-03-11
kbd: rework keyboard shortcut mapping symbols to be more intuitive

mimecuvalo:
followup to PR #5586 to update the symbols for kbd shortcuts

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

Breaking change: our keyboard shortcuts no longer use `$`/`!`/`?` but instead use `accel`/`shift`/`alt`, respectively.
- Rework keyboard shortcut mapping symbols to be more intuitive


apps/examples/src/examples/action-overrides/ActionOverridesExample.tsx
```diff
@@ -44,7 +44,7 @@ component's `overrides` prop. To learn more about using this actions via a custo
 custom actions menu example.
 
 [2]
-For kbds, Shift = !, Alt = ?, Cmd = $, i.e. Cmd+Shift+U is $!u. For more information on keyboard shortcuts see the
+For kbds, Shift = ⇧, Alt = ⌥, Cmd = ⌘, i.e. Cmd+Shift+U is ⇧⌘u. For more information on keyboard shortcuts see the
```
SomeHats, line 47: can we clarify that cmd is remapped to ctrl on windows here?

steveruizok:
I'm not against this PR but I think we should table it, as any decision here is going to be subjective. There's a pull toward keeping things as they are, so if it's _just_ competing vibes then I think we should keep it the way it is; if there's a strong feeling toward a change then let's do it.

For context, the reason I ended up on these symbols is due to cross-platform confusion. 

The **$** symbol in this context means **⌘ Command or ^ Control on Mac—or Control on PC**. The symbol itself is arbitrary. I picked it because there's no overlap with its usual meaning and its role here: no one would expect that to Undo, a user should press dollar sign and u.

The risk in using **⌘** as a signifier for **⌘ Command or ^ Control on Mac—or Control on PC**  is that its "default" reference (⌘ Command key on Mac) overlaps with _some of_ the reference here in this context, which could be misleading and IMO this is more confusing than using an arbitrary sign.

I can imagine a developer asking "⌘u is fine for Mac users, but what about on PC?" and having to respond "Oooh yeah well actually in this context ⌘ means ⌘ Command or ^ Control on Mac—or Control on PC"; where as with $ (or any other unrelated symbol) a developer would look to context for clues about what the symbol means. It's an easy problem to solve if you know that you have to solve it; it's a hard problem to solve if you think you've solved it already.

mimecuvalo:
> I'm not against this PR but I think we should table it, as any decision here is going to be subjective. There's a pull toward keeping things as they are, so if it's _just_ competing vibes then I think we should keep it the way it is; if there's a strong feeling toward a change then let's do it.

These are good points you raise, I see what you mean!

In that case, honestly, I would suggest that we use the keywords `Meta`/`Alt`/`Shift` literally to help this workflow.


So, for example, `kbd: '?!V',` → `kbd: alt+shift+V`. The shorthand doesn't get us anything, and just creates confusion.

Basically, we're doing all this work in `getHotkeysStringFromKbd` to construct this anyway but this shorthand we've constructed is basically our version of [brainfuck](https://en.wikipedia.org/wiki/Brainfuck) haha :P
I would stress this is less vibes and just being able to grok the code easily, so I'd feel strongly about this.

mimecuvalo:
ok this is reworked to be even more dead-simple. PTAL @steveruizok

mimecuvalo:
"meta" isn't a MacOS thing - this is a cross-OS wording convention: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey

> Meta is specifically on Mac so it has the same issue as Command symbol. If we want to go descriptive, CommandOrControl could work. FTR I can never "remember" the $?! meanings but again that might be a good thing (to know that I don't know).

The previous code was stating "or modifier keys (using ! for shift, $ for cmd, and ? for alt)" so it feels like we're being nitpicky here for language that we were already using that was MacOS-specific, yeah?

> I propose we host a small lunchtime debate session. 60s statements, one round of 30s responses?

👍

mimecuvalo:
er, sorry I didn't mean `meta` - I meant `accel`.
We have this as a util helper: https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/utils/keyboard.ts

mimecuvalo:
@steveruizok I think you're probably a little busy for the headspace on this one. To avoid nerd-sniping you further, @MitjaBezensek and @SomeHats wanna discuss here? I think the rest of the concerns have been allayed but lemme know if I'm missing something.

SomeHats:
I think switch what's currently `cmd-x` to `accel-x` resolves things? i don't feel strongly about any of this tho.

SomeHats:
maybe controversial but just throwing it out there: we could remove the 'special thing that means cmd and sometimes ctrl' key and ask people to write `cmd+d,ctrl+d` if that's what they mean 🤷

mimecuvalo:
yeah, to be fair, i think `cmd` is fine, i'm just tryin to make concessions here and feels like this PR is in bikeshedding mode :P
- yes, `alt` could have a similar problem but it doesn't map to possibly two different keys so maybe that's why it's ok.
- yeah `accel` terminology is a bit old school.
- it's true we're doing some magic to split out the cmd/ctrl key from one shorthand symbol/word. @SomeHats you might have a point of just being verbose around it. this does constitute a breaking change, so maybe we just stop being magical about this.

in the end, i do want us to something better than the confusing we currently have though!

mimecuvalo:
I just wanna reiterate here that `accelKey` is something that's already in our codebase in lots of places:
- https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/editor/types/event-types.ts#L52
but grep around and you'll see more.

mimecuvalo:
talked offline, we're going to do two things:
- make this verbose (split out cmd/ctrl)
- make it backwards compatibile

mimecuvalo:
ok updated. PTAL @SomeHats when you get the chance.

SomeHats (approved):
lgtm! can we document here? https://tldraw.dev/reference/tldraw/TLUiToolItem#kbd

Issue #5153, opened 2024-12-24
Make tldraw CSP compliant


luca-workera:
### What's the feature?

Hi,
thank you for the great work.

I like Tldraw and want to adopt it in the web app I am working on. The only issue is that it is not CSP-compliant.
For those who are not familiar with CSP, it is a feature that helps to prevent or minimize the risk of certain types of security threats, more here: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

CSP is very popular, I have seen it adopted in every company I have worked with; clients ask for it explicitly, so I think you should adjust to it, as a win-win (for Tldraw and who adopts it).

I browsed the code and it is very close to CSP-compliant; the only issue I can see is that a `<style>` tag is created to handle dynamic style. That happens only three times in the whole code, here:
- [Watermark component](https://github.com/tldraw/tldraw/blob/9e5acfe11f3dbb93e8a40d43d5dda936e34a62e2/packages/editor/src/lib/license/Watermark.tsx#L174)
- [defaultStyleDefs](https://github.com/tldraw/tldraw/blob/9e5acfe11f3dbb93e8a40d43d5dda936e34a62e2/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx#L38)
- [OverflowingToolbar component](https://github.com/tldraw/tldraw/blob/9e5acfe11f3dbb93e8a40d43d5dda936e34a62e2/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx#L141)

The strict CSP configuration blocks the inline style set by the `<style>` tag, so the browser won't parse it on an app on which CSP is applied, resulting in a Tldraw rendered with a broken style.

I would be happy to make a PR myself to fix the cases above, the code looks clear and I think I know what to do to fix the above cases, but first I would like to understand if you are interested in:
- have Tldraw CSP compliant
- keep it CSP compliant in the future as well

**This could be both a feature request or a bug report, happy to move it if you think it is the case.**

For me would be very appreciated to hear from you, I am happy to add more info if needed.
Thank you,
Luca

I am attaching a screenshot of the current issues with CSP and the broken style as the outcome:
_Tldraw with broken style due to CSP_
<img width="1500" alt="Screenshot 2024-12-24 at 11 38 46" src="https://github.com/user-attachments/assets/a6d51707-f5e4-4922-a19e-72329eaafafc" />

_CSP errors logged on the browser console_
[...]

PR #5607, opened 2025-03-11
security: provide a way to pass through `nonce` to the editor

mimecuvalo:
In our editor there are several places that use `<style>` tags. This can be in conflict with downstream consumers of our SDK (especially enterprise) that use CSP for their products.

After trying to see if we could do away with the `<style>` tags, we're just gonna provide a way to do `nonce`. While not ideal, it'll at least provide a way for our users to be compliant while (whilst? :P) still being able to use the tldraw library.

@ds300 I removed the `unsafe-eval` from the dotcom CSP, it was added in https://github.com/tldraw/tldraw/pull/4884 but I think it was added by mistake. Correct me if I'm wrong about that.

fixes https://github.com/tldraw/tldraw/issues/5153

drive-by cleanup of license cc @ds300 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Provide support to pass through `nonce` to the Editor.


SomeHats (approved):
api side looks good! idk about the csp but seems probably fine???

mimecuvalo:
yeah, i need @ds300 to weigh in about the "wasm for zero" csp change, seems unused

mimecuvalo:
@ds300 weighed in offline, safe to remove for now. we'll re-eval later.

PR #5614, opened 2025-03-12
'New user' -> 'Guest user' (dotcom only)

ds300:
1. This PR makes the previously-hard-coded 'New user' string subject to translations under the key `people-menu.anonymous-user`.
2. For dotcom this PR overrides the `'New user'` string with `'Guest user'`

### Change type

- [x] `improvement`

### Release notes

- Add a translation key 'people-menu.anonymous-user' with the default string 'New User'
- BREAKING CHANGE: `editor.user.getName()` no longer returns `'New user'` if the user has no name set. Instead it returns the empty string `''`.
- BREAKING CHANGE: `defaultUserPreferences.name` is no longer the string `'New user'`, it is now the empty string `''`


apps/dotcom/client/src/tla/components/TlaEditor/editor-messages.ts
```diff
@@ -5,4 +5,5 @@ export const editorMessages = defineMessages({
 	untitledProject: { defaultMessage: 'Untitled file' },
 	downloadFile: { defaultMessage: 'Download file' },
 	copyToMyfiles: { defaultMessage: 'Copy to my files' },
+	anonymousUser: { defaultMessage: 'Guest User' },
```
mimecuvalo, line 8: If I can help wordsmith here a bit, I think Guest User would be better serviced by just simply saying "Guest".

mimecuvalo: qttymueesbw71

ds300: tried this, it doesn't work quite as well. when you add 'user' in there it feels a tad crufty yes but in this situation that actually helps to convey that it's a default value provided by the app rather than a name chosen by the user, so it parses easier.

PR #5616, opened 2025-03-12
debug: move hard reset option to avoid fat finger

mimecuvalo:
little nit that i would run into occassionally, it's so close to the debug flags, and has no `confirm`.
i could have added a dialog to confirm but meh ¯\\_(ツ)_/¯ 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


SomeHats (approved):
makes sense

Issue #5622, opened 2025-03-13
Assets not being removed from indexedDB when calling deleteAssets


mimecuvalo:
### What happened?

From Discord:
> I understand that when a shape is deleted there is no asset cleanup. I've been using editor.deleteAssets to cleanup myself, and when you execute editor.getAssets() you see that they are removed. However, if you inspect the assets section in indexedDB, the assets are still there. Is this the intended behavior? I wasn't aware of that and ended with A LOT of unused assets in indexedDB

to investigate

### Where did this happen?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #5628, opened 2025-03-13
assets: be able to remove assets that are unused

mimecuvalo:
Be able to cleanup the assets when the assets are actually deleted. 
This came up in Discord here: https://discord.com/channels/859816885297741824/1349476132365537320/1349476132365537320

The IndexedDB can start getting crufty with old assets that are unused. This allows folks that proactively delete assets (not just the shapes) to be able to keep the local db clean.

fixes https://github.com/tldraw/tldraw/issues/5622

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Cleanup assets from the local indexedDB that are proactively deleted.

### DIFF
--- a/packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
+++ b/packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
@@ -303,6 +303,13 @@ export class LocalIndexedDb {
 			await assetsStore.put(blob, assetId)
 		})
 	}
+
+	async removeAsset(assetId: string) {
+		await this.tx('readwrite', [Table.Assets], async (tx) => {
+			const assetsStore = tx.objectStore(Table.Assets)
+			await assetsStore.delete(assetId)
+		})
+	}
 }
 
 /** @internal */
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -2105,6 +2105,8 @@ export class LocalIndexedDb {
     // (undocumented)
     pruneSessions(): Promise<void>;
     // (undocumented)
+    removeAsset(assetId: string): Promise<void>;
+    // (undocumented)
     storeAsset(assetId: string, blob: File): Promise<void>;
     // (undocumented)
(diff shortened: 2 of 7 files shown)


steveruizok:
IIRC we only keep things around due to undo/redo. For example, if I have lots of images on the page, I delete them all, and then undo the delete. I'm _pretty sure_ we could clean up unused assets whenever the client loads.

mimecuvalo:
@steveruizok this is only for the case when somebody _explicitly_ calls `deleteAssets` to cleanup manually.

packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
```diff
@@ -303,6 +303,13 @@ export class LocalIndexedDb {
 			await assetsStore.put(blob, assetId)
 		})
 	}
+
+	async removeAsset(assetId: string) {
+		await this.tx('readwrite', [Table.Assets], async (tx) => {
+			const assetsStore = tx.objectStore(Table.Assets)
+			await assetsStore.delete(assetId)
+		})
+	}
```
SomeHats, line 312: is it worth having this take the assetids array and do things in a single tx?

SomeHats (commented):
makes sense

packages/editor/src/lib/utils/sync/LocalIndexedDb.ts
mimecuvalo: maybe? I think i was worried that, if it's a lot of megabytes to delete it might start taking too long? I guess it's the same difference but maybe it could help the user mitigate freezing the ui potentially?

SomeHats: I'm pretty skeptical that this would impact that without some sort of explicit collaborative scheduling but maybe?

mimecuvalo: ok, fair enough, i'm sorta ambivalent on it. 👍

Issue #5215, opened 2025-01-14
WCAG compliance / accessibility megathread


steveruizok:
This issue is intended to capture all discussion and answer any questions around accessibility in tldraw, especially as it relates to [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) compliance. We've had several questions in the community around accessibility—this is the place to discuss!

## What does accessibility mean for tldraw?

People with disabilities should be able to use tldraw, just like they should be able to use the rest of the web.

The tldraw canvas is an unusual piece of media on the web. All of its unique characteristics present challenges for accessibility. Even a question like "what is the document" (e.g. to be read by a screen reader) is complicated by the fact that the document is multi-dimensional and often changing rapidly. There is no text cursor or insertion point. Many interactions involve visual relationships. Some standard interactions, such as text-based search, are harder to achieve on the canvas.

In the past, we've relied on conventions to solve many of tldraw's most difficult design problems. Discoverability is especially difficult, so we rely on conventions to avoid overloading the user interface. For example, the "rotate zone" at the outside corners of shapes is very difficult to discover if you've never used a canvas application before; but if you have, then it's obvious because all of the other apps do it the same way.

There are unfortunately few accessibility conventions to follow on the canvas, as the existing suite of canvas apps have not had accessibility features long enough for strong conventions to emerge. However, as an SDK, tldraw is in a unique position to "get accessibility right" in a way that helps _create_ those conventions, or at least to set a high standard across multiple products. Personally, I love the idea that many products (even competing products) would each inherit good accessibility patterns from tldraw. It's certainly an attractive value proposition to include these features with tldraw's SDK for customers.

### Gather information

Before we start shipping anything, there's research work to do to better understand the problem and its opportunities. 

- Inventory of accessibility features in tldraw. For the "normal stuff" like menu navigation, text, buttons, etc., what level of compliance are we reaching? Where can we improve on the areas that _don't_ require any creative solutions?
- Exploration of accessibility solutions in other apps. How do apps like Figma, Miro, Canva, Whimsical, Excalidraw, etc offer for accessibility in their apps? How well do these features work? Who worked on them? Are there patterns or conventions that these apps use that we can follow or adhere to? Innovation here is an anti-pattern, so if we can find something that works well and that we can replicate, it would be better for users.
- What are the "open questions" in accessibility for canvas applications? Which problems have not yet had solutions?
[...]

PR #5634, opened 2025-03-13
a11y: add a live region to announce selected tools

mimecuvalo:
Starts improving our voiceover, starting with the toolbar which was crap.
This follows the toasts pattern we have, since this is, in a way, a notification of what's happening on the screen, especially when selecting something using a hotkey instead of using the mouse pointer.

https://github.com/user-attachments/assets/95f7551a-e891-4c5a-bf93-784b86fe618f

related issue https://github.com/tldraw/tldraw/issues/5215

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Adds better voiceover support when selecting an action (a11y)


packages/tldraw/src/lib/ui/components/A11y.tsx
```diff
@@ -0,0 +1,26 @@
+import { useValue } from '@tldraw/editor'
+import { memo } from 'react'
+import { useA11y } from '../context/a11y'
+
+/** @public @react */
+export const DefaultA11y = memo(function TldrawUiA11y() {
```
SomeHats, line 6: We may add other a11y related components in the future, so I'd like to give this a slightly more specific name - `A11yAnnouncer`, perhaps?

packages/tldraw/src/lib/ui/context/a11y.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+/** @public @react */
+export function TldrawUiA11yProvider({ children }: A11yProviderProps) {
+	const currentMsg = useAtom<TLUiA11y>('a11y', { msg: '', priority: 'polite' })
+	const ctx = useContext(A11yContext)
+
+	const current = useMemo(
+		() => ({
+			currentMsg,
+			setMessage(msg: TLUiA11y) {
```
SomeHats, line 35: maybe `announce` instead of `setMessage`? To me `setMessage` doesn't imply that then might be immediately reflected to the user

packages/tldraw/src/lib/ui/hooks/useTools.tsx
```diff
@@ -44,8 +46,22 @@ export function ToolsProvider({ overrides, children }: TLUiToolsProviderProps) {
 	const editor = useMaybeEditor()
 	const trackEvent = useUiEvents()
 
+	const a11y = useA11y()
+	const msg = useTranslation()
 	const helpers = useDefaultHelpers()
 
+	const postSelectAction = React.useCallback(
```
SomeHats, line 53: nit: post makes me think of HTTP requests and i got confused for a sec

SomeHats (approved):
nice! some relatively minor thoughts on API and naming but otherwise this is looking great :)

packages/tldraw/src/lib/ui/context/a11y.tsx
mimecuvalo: good call 👍

PR #5639, opened 2025-03-13
error logging for i18n-upload-strings

ds300:
this seems to have been failing sometimes but it doesn't really say why because the underlying error seems to be eaten up by the 'unhandled promise rejection' reporter.

### Change type

- [x] `other`


mimecuvalo (approved):
cool thx!

PR #5643, opened 2025-03-13
Track embedded examples on docs site

TodePond:
This PR adds a URL query so that we can track which views on the examples site come from being embedded within our docs

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo (approved):
nice one 🤘

Issue #5645, opened 2025-03-14
Black highlight appears yellow


Ea-st-ring:
### What happened?

Please check the video below. When I use highlight with black color, it applied with yellow.
I made [PR](https://github.com/tldraw/tldraw/pull/5646) for this issue, Thanks.

https://github.com/user-attachments/assets/6969ec7c-130f-4dd1-90d5-08bc2d9484e1

### How can we reproduce the bug?

Just Do it in playground

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The developer package

### Contact Details

eastring@elicer.com

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
Ah, this one of those things that is actually unusual but expected in our app. The problem is that a "black" highlight isn't really a thing, and since our default color is black we want the highlighter to be the correct expected color.

Ea-st-ring:
Actually I do not fully understand your comment ㅜ.ㅜ It seemed like an issue but actually it was intended, right?
Then do I have to override css for highlight default color to black?
Our service client mentioned it as an issue

mimecuvalo:
Yeah, you would have to override the color with your own if you like! Take a look at this example to see how:
- https://examples.tldraw.com/changing-default-colors
- https://github.com/tldraw/tldraw/tree/main/apps/examples/src/examples/changing-default-colors

PR #5646, opened 2025-03-14
fix: update highlight color in DefaultColorThemePalette from yellow t…

Ea-st-ring:
…o dark gray

Please See [this Issue](https://github.com/tldraw/tldraw/issues/5645)

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [x] Unit tests
- [x] End to end tests

### Release notes

- Fixed a bug with…


mimecuvalo:
Ah, this one of those things that is actually unusual but expected in our app. The problem is that a "black" highlight isn't really a thing, and since our default color is black we want the highlighter to be the correct expected color.

PR #5658, opened 2025-03-15
[Fix] Rich text perf issue

steveruizok:
This PR improves the performance of the app with regard to text.

https://github.com/user-attachments/assets/3c468587-0cc1-4e72-8979-61fe14f6a0dc

The `renderPlaintextFromRichText` method is slow. It depends on TipTap's `getText` method. This parses the rich text and flattens it to plain text. As usual, we can improve performance by: 

1. making `renderPlaintextFromRichText` faster
2. not calling `renderPlaintextFromRichText`

## Making it faster 

We can make it faster by using a weak cache so that we don't need to re-flatten text if we've flattened it before. We can also make it faster by skipping the flatten if the text content is empty. This can be done very quickly without using `getText`.

## Avoiding it entirely

It turns out that most of the places where we call `renderPlaintextFromRichText` are checks against empty text, meaning we don't need to call `renderPlaintextFromRichText` at all. In these cases we _just_ check whether the text is empty and continue based on that result. 

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improved performance related to rich text.

### DIFF
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
@@ -20,10 +20,8 @@ import {
 	TLGeoShape,
 	TLGeoShapeProps,
 	TLResizeInfo,
-	TLRichText,
 	TLShapeUtilCanvasSvgDef,
 	Vec,
-	WeakCache,
 	exhaustiveSwitchError,
 	geoShapeMigrations,
 	geoShapeProps,
@@ -37,7 +35,6 @@ import {
 
 import isEqual from 'lodash.isequal'
 import {
-	isEmptyRichText,
 	renderHtmlFromRichTextForMeasurement,
 	renderPlaintextFromRichText,
 } from '../../utils/text/richText'
@@ -65,8 +62,6 @@ import {
 import { getLines } from './getLines'
 
 const MIN_SIZE_WITH_LABEL = 17 * 3
-
-const textCache = new WeakCache<TLRichText, string>()
 
 /** @public */
 export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
@@ -416,11 +411,7 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 	}
 
 	override getText(shape: TLGeoShape) {
-		const { richText } = shape.props
-		return textCache.get(richText, () => {
-			if (isEmptyRichText(richText)) return ''
-			return renderPlaintextFromRichText(this.editor, richText)
-		})
+		return renderPlaintextFromRichText(this.editor, shape.props.richText)
 	}
 
 	override getFontFaces(shape: TLGeoShape): TLFontFace[] {
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
@@ -20,8 +20,10 @@ import {
 	TLGeoShape,
 	TLGeoShapeProps,
 	TLResizeInfo,
+	TLRichText,
 	TLShapeUtilCanvasSvgDef,
 	Vec,
+	WeakCache,
 	exhaustiveSwitchError,
 	geoShapeMigrations,
 	geoShapeProps,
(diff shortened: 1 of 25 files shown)


packages/tldraw/src/lib/utils/text/richText.ts
```diff
@@ -89,6 +90,23 @@ export function renderHtmlFromRichTextForMeasurement(editor: Editor, richText: T
...
+	}
+	return false
+}
+
+export function isSingleCharacterRichText(richText: TLRichText) {
+	if (richText.content.length === 1) {
+		if ((richText.content[0] as any)?.content?.length === 1) return true
+	}
+	return false
+}
```
SomeHats, line 108: these are fine for us, but we should double-check they work with non-texty extensions people might add like images, mentions, or horizontal rules

packages/tldraw/src/lib/shapes/shared/useEditableRichText.ts
```diff
@@ -9,7 +9,7 @@ export function useEditableRichText(shapeId: TLShapeId, type: string, richText?:
 	const isEditing = commonUseEditableTextHandlers.isEditing
 	const editor = useEditor()
 	const rInput = useRef<HTMLDivElement>(null)
-	const isEmpty = richText ? renderPlaintextFromRichText(editor, richText).length === 0 : true
+	const isEmpty = richText && !isEmptyRichText(richText)
```
mimecuvalo: patching this up, this was inverted

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -706,9 +708,9 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 		}
 
 		// If we got rid of the text, cancel out any growY from the prev text
-		const prevPlaintext = renderPlaintextFromRichText(this.editor, prev.props.richText)
-		const nextPlaintext = renderPlaintextFromRichText(this.editor, next.props.richText)
-		if (prevPlaintext && !nextPlaintext) {
+		const wasEmpty = isEmptyRichText(prev.props.richText)
+		const isEmpty = isEmptyRichText(next.props.richText)
+		if (wasEmpty && !isEmpty) {
```
mimecuvalo, line 713: going to fix this inversion too :)

packages/tldraw/src/lib/utils/text/richText.ts
```diff
@@ -89,6 +90,23 @@ export function renderHtmlFromRichTextForMeasurement(editor: Editor, richText: T
...
+export function isEmptyRichText(richText: TLRichText) {
+	if (richText.content.length === 1) {
+		if (!(richText.content[0] as any).content) return true
+	}
+	return false
+}
+
+export function isSingleCharacterRichText(richText: TLRichText) {
+	if (richText.content.length === 1) {
+		if ((richText.content[0] as any)?.content?.length === 1) return true
```
mimecuvalo, line 105: this should be checking the `text`

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -116,7 +116,7 @@ export class EditingShape extends StateNode {
 				// N.B. One nuance here is that we want empty text fields to be removed from the canvas when the user clicks away from them.
 				const isEmptyTextShape =
 					this.editor.isShapeOfType<TLTextShape>(editingShape, 'text') &&
-					renderPlaintextFromRichText(this.editor, editingShape.props.richText).trim() === ''
+					isEmptyRichText(editingShape.props.richText)
```
mimecuvalo, line 119: this isn't quite the same - we want the shapes to be removed if they're just whitespace, going to revert this bit

steveruizok: Good catch, I'd left the one in onShapeUpdate or EditEnd or whatever that deletes the shape, though with a note that we should see if we can do a version of tiptap/pm's `getText` that bails as soon as it finds any text, so that we don't block after editing large pieces of text.

steveruizok:
Looks good, landing! Thanks for the help @mime

PR #5625, opened 2025-03-13
Track Enter key to accept search query

TodePond:
This PR makes us also track enter key presses in search queries too. This will help us evaluate how helpful our search is to developers.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

PR #5666, opened 2025-03-17
Hotfix: Track Enter key to accept search query (#5625)

TodePond:
This is a manual hotfix as it cherrypicks a PR that landed *before* the new hotfix infra arrived.

---

This PR makes us also track enter key presses in search queries too. This will help us evaluate how helpful our search is to developers.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
@TodePond you can use the `docs-hotfix-please` label now fwiw :D

TodePond:
@mimecuvalo yes I know - this one snuck in before that, so it has to be manual

PR #5667, opened 2025-03-17
Lu/better example focus

TodePond:
This PR fixes and improves the focus behavior of examples on tldraw.dev

- The current focus behavior was broken, as the written `preserveFocus` URL query appears to be a typo. The actual query we listen for is `tldraw_preserve_focus`, so our query was doing nothing.
- I think a better behavior would be to make it more explicit as to when the user clicks onto the example. Currently it's quite annoying that it hijacks your scroll when you scroll down to see the example code. From our interviews, we know that many developers use these examples pages as references for code, often going back to the same example again and again.
- Currently, it isn't clear when you can or cannot use keybinds. After opening an example, you can immediately pan around the canvas, but you can't use keybinds. This PR makes it a more explicit choice to start interacting with the example, as it prompts you to click on the example itself, and there's a blue focus indicator around the example whenever you can use keybinds.

TODO:
- [ ] push focus indicator
- [ ] gif/screenshot comparison

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Try out the examples on the docs site.
2. Click in and out of the examples, use keybinds, check if all is working as intended.

### Release notes

- Docs: Improved UX of examples on the tldraw.dev


steveruizok:
Great improvements! Did you try showing the **Try example** UI again when the user clicks out of the example?

TodePond:
> Did you try showing the **Try example** UI again when the user clicks out of the example?

I was originally worried it would be too annoying, but I've tried it now and it's not too bad. It now appears and disappears as you click on and off it.

2025-03-18 at 10 35 16 - Olive Chimpanzee

PR #5670, opened 2025-03-17
Make `type` optional for updateShapes

ds300:
- Adds a new type `TLShapeUpdatePartial` which is just a copy of `TLShapePartial` `type` is optional.
- Updates all code that previously used `TLShapePartial` in updateShape codepaths to use `TLShapeUpdatePartial`.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Make the `type` property optional when calling `updateShape`, `updateShapes`, `animateShape`, and `animateShapes`.

### DIFF
--- a/packages/tlschema/api-report.api.md
+++ b/packages/tlschema/api-report.api.md
@@ -1450,6 +1450,13 @@ export type TLShapePartial<T extends TLShape = TLShape> = T extends T ? {
 } & Partial<Omit<T, 'id' | 'meta' | 'props' | 'type'>> : never;
 
 // @public (undocumented)
+export type TLShapeUpdatePartial<T extends TLShape = TLShape> = T extends T ? {
+    id: TLShapeId;
+    meta?: Partial<T['meta']>;
+    props?: Partial<T['props']>;
+} & Partial<Omit<T, 'id' | 'meta' | 'props'>> : never;
+
+// @public (undocumented)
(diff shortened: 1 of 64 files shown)


packages/tlschema/api-report.api.md
```diff
@@ -1449,6 +1449,13 @@ export type TLShapePartial<T extends TLShape = TLShape> = T extends T ? {
     type: T['type'];
 } & Partial<Omit<T, 'id' | 'meta' | 'props' | 'type'>> : never;
 
+// @public (undocumented)
+export type TLShapeUpdatePartial<T extends TLShape = TLShape> = T extends T ? {
```
SomeHats, line 1453: For bindings, we use `TLBindingCreate` and `TLBindingUpdate` as the types for the create/update partials. Could this either match those, or could we update the binding ones to reflect this naming scheme?

steveruizok: I've stumbled over those before! No preference on which we go with but I'd like us to have the same solution for both.

steveruizok:
> Historically, the reason we included `type` was to allow typescript to correctly narrow the type. However, that broke when we added `TLUnknownShape`, and we use generics instead.

Actually now that we have better types, having "type" has a purpose again!

PR #5673, opened 2025-03-17
Reactive inputs

SomeHats:
Currently, `editor.inputs` stores a bunch of state in a big mutable object. This works pretty well, because that state is _mostly_ only accessed in event handlers. 

There are a few places in the UI where we access it, where we've previously relied on a `useTick` hook to re-render the component on every frame to get around our inability to subscribe to the input values. That's been fine though, because the only bits of UI that have used it have been in debug panels, so it hasn't really mattered.

With elbow arrows, I'm expecting that to change - we'll be using the current page position in the UI, and wanting it to update reactivly. This diff replaces the existing `editor.inputs` object with a new `InputManager` that has the exact same API, but is backed by atoms instead of plain mutable objects.

I don't expect that to be controversial, but the one thing this does add that might be is `@atom accessor foo = 'bar'` syntax, for creating a field on a class called foo, which is backed by an atom, and is initialised with `'bar'`. We've supported the new standards-based decorator syntax (as well as the legacy ts decorators syntax) for a little while, but to enable this syntax I've switched our transpilation tooling over to standards-based decorators too. There are a few places where we _could_ switch to this syntax, but I've only done so for the inputs where we needed it, and one other proof-of-concept in the license manager.

Typescript, esbuild, eslint, and prettier support this syntax out of the box. I've had to make some tweaks to enable it for recast, vite, jest, and next.

### Change type

- [x] `api`

### Release notes

The values in `editor.inputs` are now reactive.

### DIFF
--- a/packages/editor/src/lib/primitives/Vec.ts
+++ b/packages/editor/src/lib/primitives/Vec.ts
@@ -1,3 +1,4 @@
+import { atom, Atom } from '@tldraw/state'
 import { VecModel } from '@tldraw/tlschema'
 import { EASINGS } from './easings'
 import { toFixed } from './utils'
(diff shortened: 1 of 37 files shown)


packages/editor/src/lib/primitives/Vec.ts
```diff
@@ -594,4 +595,36 @@ export class Vec {
...
+				get: () => this._z.get(),
+				set: (value: number) => this._z.set(value),
+			},
+		})
+	}
+
+	private _x: Atom<number>
+	private _y: Atom<number>
+	private _z: Atom<number>
+}
```
ds300, line 628: one of our big customers uses terser in their build pipeline with the `pure_getters` setting enabled, which prevents signals from being tracked in lots of cases. Which is why we originally removed all signal-capturing getters from our codebase a long while back.

I'm not sure if defining the setters like this would cause the same problem, but we definitely need to check and address it if so.

ds300 (commented):
I'm lukewarm on the `@atom` decorator. It's concise but it also kinda hides side effects in a way I'm not crazy about. Happy to give it a go though.

packages/editor/src/lib/primitives/Vec.ts
ds300: > Does that match your understanding? 

I vaguely remember it being more complicated than that, like our customer had a combination of terser settings (or maybe terser + some other compiler?) that meant it would unsafely memoize the getter value sometimes. I could be wrong, it was a long time ago.

> I think the alternative is making each of these into a function which is obviously a much bigger change from an api point of view, but maybe that's the way to go here?

Yeah maybe. The alternative is to try it like this and cross fingers that our customer doesn't notice any bugs, but that doesn't feel like a winning strategy to me 😵‍💫

SomeHats:
closing in favour of #5721

Issue #5661, opened 2025-03-16
[Bug]: SVG exports unexpectedly large on tldraw.com


zetter:
### What happened?

When exporting simple diagrams containing as SVG, the exported files are large (over 1 MB).

This happens on https://www.tldraw.com. I was unable to reproduce it in https://tldraw.dev or the sandbox so I'm unsure if they are running different versions of tldraw or there is custom code on tldraw.com that is causing the issue. 

It looks like the difference comes from 28 embeds of the Inter font at various weights and unicode ranges even though the font isn't used in the SVG.

[Example SVG from the sandbox (213 KB)](https://github.com/user-attachments/assets/4d430985-5dd9-4725-b82e-20e7882393d5)

[Example SVG from tldraw.com (1.3 MB)](https://github.com/user-attachments/assets/5093b84a-3a42-420d-ac5a-018cccac0fe7)

### How can we reproduce the bug?

1. Visit https://www.tldraw.com
2. Create a text node with some text in it
3. Go to Export -> SVG

The file size will be approximately 1.3mb

### What browsers are you seeing the problem on?

Latest Chrome and Safari

### Where did this happen?

The tldraw.com website

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
Thanks for reporting the issue — we'll look into it. cc @SomeHats

zetter:
@mimecuvalo Thanks for taking a look. I'm still able to reproduce the issue on tldraw.com after cleaning my cache and using a private browsing window - would tldraw.com have been deployed after this was merged?

mimecuvalo:
> [@mimecuvalo](https://github.com/mimecuvalo) Thanks for taking a look. I'm still able to reproduce the issue on tldraw.com after cleaning my cache and using a private browsing window - would tldraw.com have been deployed after this was merged?

tldraw.com doesn't have the latest fixes yet — the PR has landend but tldraw.com won't be updated until maybe Friday or Monday.

PR #5676, opened 2025-03-17
exports: fix Inter being embedded; reduce excessive styling

mimecuvalo:
- fix Inter being embedded in our exports, fixes https://github.com/tldraw/tldraw/issues/5661
- recursively look at inherited parent styles
- add more inheritable properties
- cross-reference [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) related properties so that the properties don't explode into all of the `-color` properties
- maybe a controversial change @SomeHats but I changed `shouldRespectDefaults: true,` — probably it's necessary but I want to understand why it is so, can we get away with not having it?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix issue with exports embedding Inter and having excessive styling.


packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -27,7 +32,7 @@ export class StyleEmbedder {
 	readRootElementStyles(rootElement: Element) {
 		// when reading a root, we always apply _all_ the styles, even if they match the defaults
 		this.readElementStyles(rootElement, {
-			shouldRespectDefaults: false,
+			shouldRespectDefaults: true,
```
mimecuvalo, line 35: this is the controversial line i'd like to understand/discuss

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -229,6 +229,7 @@ export function RichTextSVG({
 		verticalAlign === 'middle' ? 'center' : verticalAlign === 'start' ? 'flex-start' : 'flex-end'
 	const wrapperStyle = {
 		display: 'flex',
+		fontFamily: DefaultFontFamilies[font],
```
mimecuvalo, line 232: this fixes the Inter problem

packages/editor/src/lib/exports/StyleEmbedder.ts
```diff
@@ -53,9 +58,14 @@ export class StyleEmbedder {
...
-			? (this.styles.get(element.parentElement as Element)?.self ?? NO_STYLES)
-			: NO_STYLES
+		let parentStyles = NO_STYLES
+		if (shouldSkipInheritedParentStyles) {
+			let el = element.parentElement
+			while (el) {
+				parentStyles = Object.assign({}, this.styles.get(el)?.self, parentStyles)
+				el = el.parentElement
+			}
+		}
```
SomeHats, line 68: maybe minor, maybe not, but as these are called for every element in the dom, for large trees i could see this (allocating a new object and iterating over everything that came before to assign to it) getting pretty expensive for large/deep trees. The ideal thing here I think would be to build up the parent styles object as we traverse down the tree embedding styles so we don't have to constantly traverse up again. That's obvs a big change though, so maybe for now we could replace the `Object.assign` with a for/in and normal assignment so we don't end up with so many intermediate copies of the style map?

packages/editor/src/lib/exports/StyleEmbedder.ts
SomeHats: From the comment above this, I _think_ it's because this is a root element? Which means that in order for the inheritance stuff to be set up we need to pro-actively apply styles to it?

I think I would test this a bit with global/inherited styles. I believe there were some quirks I found with this when you're also using tailwind?

SomeHats: could also be worth tweaking this so that it only applies defaults that may later be inherited from?

packages/editor/src/lib/exports/parseCss.ts
```diff
@@ -111,11 +111,43 @@ export function parseCssValueUrls(value: string) {
...
+	'border-top-color',
+	'caret-color',
+	'column-rule-color',
+	'outline-color',
+	'text-decoration',
+	'text-decoration-color',
+	'text-emphasis-color',
+])
+
+export function isPropertyCoveredByCurrentColor(
```
SomeHats, line 131: this is brilliant ty

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
SomeHats: good find

SomeHats (commented):
love this!

PR #5677, opened 2025-03-18
examples: make nav not do full page reload

mimecuvalo:
this was driving me nuts, especially when filtering

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/examples/src/ExamplePage.tsx
+++ b/apps/examples/src/ExamplePage.tsx
@@ -153,7 +153,7 @@ function ExampleSidebarListItem({
 
 	return (
 		<li className="examples__sidebar__item" data-active={isActive}>
-			<Link to={example.path} className="examples__sidebar__item__link hoverable" />
+			<Link replace to={example.path} className="examples__sidebar__item__link hoverable" />
 			<div className="examples__sidebar__item__title">
 				<span>{example.title}</span>
 			</div>


apps/examples/src/ExamplePage.tsx
```diff
@@ -153,7 +153,7 @@ function ExampleSidebarListItem({
 
 	return (
 		<li className="examples__sidebar__item" data-active={isActive}>
-			<Link to={example.path} className="examples__sidebar__item__link hoverable" />
+			<Link replace to={example.path} className="examples__sidebar__item__link hoverable" />
```
SomeHats, line 156: wait why does this prevent a full page reload? i would have thought it would just disabled the back button?

mimecuvalo: these are ajax links/dynamic links, the current ones cause a full navigation, this just uses `history.replace`

mimecuvalo: actually, now that i write that out-loud, i'm hearing my mistake, you're right, i have no idea why this is working - this does break Back behavior.
but without `replace` it doesn't do a dynamic load... weird

mimecuvalo: reverted https://github.com/tldraw/tldraw/pull/5724

PR #5680, opened 2025-03-18
dotcom/sdk sync-up: port back some of the new UI treatment to the SDK/examples

mimecuvalo:
With the new dotcom release, our SDK styling has gotten a little crusty. Living in the examples/development mode, I'm stuck in the old world 😭 

- unifies the styling in the top-left area with the new dotcom look 💅 
- updates the `actionShortcutsLocation` to the new default of `toolbar` - let me know if this is controversial though...
- adds a regular sidebar toggle to our examples app instead of the "full" button which was hard to discover

Screenshot 2025-03-18 at 14 03 38

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Update styling for our top-left menu area

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
@@ -214,7 +214,6 @@ function TlaEditorInner({ fileSlug, deepLinks }: TlaEditorProps) {
 				onMount={handleMount}
 				onUiEvent={handleUiEvent}
 				components={components}
-				options={{ actionShortcutsLocation: 'toolbar' }}
 				deepLinks={deepLinks || undefined}
 				overrides={overrides}
 				isShapeHidden={isShapeHidden}
--- a/apps/dotcom/client/src/tla/components/TlaEditor/TlaLegacyFileEditor.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/TlaLegacyFileEditor.tsx
@@ -110,7 +110,6 @@ function TlaEditorInner({
 				onUiEvent={handleUiEvent}
 				components={components}
 				deepLinks
-				options={{ actionShortcutsLocation: 'toolbar' }}
 			>
 				<ThemeUpdater />
 				<SneakyDarkModeSync />
--- a/apps/dotcom/client/src/tla/components/TlaEditor/TlaLegacySnapshotEditor.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/TlaLegacySnapshotEditor.tsx
@@ -129,7 +129,6 @@ function TlaEditorInner({
 				onUiEvent={handleUiEvent}
 				components={components}
 				deepLinks
-				options={{ actionShortcutsLocation: 'toolbar' }}
 			>
 				<ThemeUpdater />
 				<SneakyDarkModeSync />
--- a/apps/dotcom/client/src/tla/pages/local.tsx
+++ b/apps/dotcom/client/src/tla/pages/local.tsx
@@ -73,7 +73,6 @@ function LocalTldraw() {
 						}
 					})
 				}}
-				options={{ actionShortcutsLocation: 'toolbar' }}
 			>
 				<SneakyDarkModeSync />
 			</LocalEditor>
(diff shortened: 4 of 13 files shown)


SomeHats:
one principle that we've tried to stick to in the examples app is a sort of "what you see is what you get" - that is, anything within the area of the example can be replicated ~perfectly by duplicating the code of the example. We've been pretty careful to avoid chrome from the outside examples app bleeding into the example area itself to preserve this.

For example, an earlier version of the syncing examples used a floating controls in the top-center of the screen, but made it look too much like a part of the example, so now we have the current design where the sync controls are clearly outside of the example itself.

I think that we should maintain this principle, and not add the in-example sidebar button. 

What's the motivation behind making the full mode more discoverable? I see it as mostly being useful for people developing tldraw. For actual users, we should be driving them away from the examples app towards tldraw.dev/examples instead.

As for the default `actionShortcutsLocation`... i'm uncertain, but generally i think skeptical. This is a pretty big change to be sending down the wire to our end users. I guess it's easy enough to revert though? Not sure.

Finally... if we do this, we should put the ... after the page name, as it is on tldraw.com. the spacing on it looks wierd otherwise.

mimecuvalo:
> one principle that we've tried to stick to in the examples app is a sort of "what you see is what you get" - that is, anything within the area of the example can be replicated ~perfectly by duplicating the code of the example. We've been pretty careful to avoid chrome from the outside examples app bleeding into the example area itself to preserve this.

yeah i hear this. i think i'm not too worried about this being confusing? it's not in any individual example code so i think it might be a bit overblown that people will get confused. so i'm just speaking to your point about duplicating the code of the example — i think this might be coddling the devs a bit heh :P

> For example, an earlier version of the syncing examples used a floating controls in the top-center of the screen, but made it look too much like a part of the example, so now we have the current design where the sync controls are clearly outside of the example itself.

i think it's fine personally ¯\\_(ツ)_/¯  but we should debate it more sounds like!

> What's the motivation behind making the full mode more discoverable? I see it as mostly being useful for people developing tldraw. For actual users, we should be driving them away from the examples app towards tldraw.dev/examples instead.

it's just pretty undiscoverable, and i want to make it feel more like a sidebar would. no normal sidebar has a fullscreen button embedded in it, that breaks expectations, a fullscreen button is usually in the content / embed itself. and the sidebar can be toggled globally, not on an individual item level.

> As for the default actionShortcutsLocation... i'm uncertain, but generally i think skeptical. This is a pretty big change to be sending down the wire to our end users. I guess it's easy enough to revert though? Not sure.

yeah, i don't feel strongly on this point, but wanted to push on it to see if it's worth budging.

> Finally... if we do this, we should put the ... after the page name, as it is on tldraw.com. the spacing on it looks wierd otherwise.

it felt a bit weird but i don't disagree, i'm happy to do that as well!

feels like @steveruizok would wanna weigh in, oh and possibly @ds300

mimecuvalo:
cc @TodePond as well!

TodePond:
> For actual users, we should be driving them away from the examples app towards tldraw.dev/examples instead.

Yes this is correct. Though some of our developers do use the examples app locally to try out things.

I'm also conscious that the `/full` routes of these examples becomes embedded on `tldraw.dev`, and that's where most of our developers _first_ explore these.

I agree with Alex here: Whenever you copy-paste from an example... it should create exactly the same thing as you should see in the example. If we want to refresh the look of the examples, we should refresh the out-of-the-box look of the SDK itself.

steveruizok:
This is a bigger project. If we can make small changes then let's do that, otherwise let's use our designers for a proper redesign.

Some places where it makes sense to improve:
- When opening an example, scroll to the example in the sidebar
- Slow fast refresh during development, could it be improved?

Some constraints:
- we want to show the SDK as vanilla as possible
- we want to make / keep our examples FAST for development and hacking

mimecuvalo:
I think I muddied the waters here by including/discussing any changes to the Examples stuff.

I backed that piece of it out to make the discussion clearer. The PR now just _only_ changes the SDK to make the top-left corner styling up-to-date.

> we want to show the SDK as vanilla as possible

totally agreed, that's why I think we should make it look like our vanilla tldraw.com setup.

Several times in the comments above folks have stated that the SDK should look like what we have out-of-the-box. The problem I think is that tldraw.com has now skewed from what is out-of-the-box because of the different visual treatment.

So let's just focus on these targeted CSS changes for the top-left.

steveruizok:
Closing this one; the examples app "needs" a re-design, as do our starter kits. I have some ideas about a unified place for developers (a kind of platform.tldraw.com) but would prefer small fixes/changes to examples site for now.

PR #5685, opened 2025-03-18
fix sentry csp

ds300:
seems like sentry is using this other domain, at least on our preview envs

image

### Change type

- [x] `other`


ds300:
this affects staging only, and it is required because i updated the DSN for the staging env kinda by accident. You can see here https://tldraw.sentry.io/settings/projects/dotcom-staging/keys/

Issue #5611, opened 2025-03-11
Rich text disappears with React StrictMode


bluedot74:
### What happened?

If React StrictMode is enabled and you're in a development environment, editing Rich Text causes the text to disappear.

### How can we reproduce the bug?

1. In a React development environment enable StrictMode.
2. Add text to the canvas
3. Click twice on the text to enter Edit mode
4. Note the text disappears and additional keystrokes don't do anything

### Where did this happen?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bluedot74:
https://github.com/user-attachments/assets/ec0d492f-a13f-4f26-9576-425dadd8c596

mimecuvalo:
There is a known issue in development with this combination: rich text + strict mode + iOS

Is this happening on iOS for you or some other environment?

bluedot74:
> There is a known issue in development with this combination: rich text + strict mode + iOS
> 
> Is this happening on iOS for you or some other environment?

It's happening on Mac in Chrome and Safari, I haven't tried on any mobile devices.

orwashams:
Same problem here, im using nextjs 15.2.1, tldraw 3.10.2

- used Brave browser with shield off
- Tried incognito mode it didn't work
- Tried Firefox and still the same issue
- There are no extensions on my browser

An important note: If you select the text tool then immediately change the color the cursor appears and you can type text.

arybitskiy:
Same here: react-router 7 + react 19

mimecuvalo:
I'm still keeping an eye on this. I've tried on both Next.js and Vite and various configs and still can't repro 😞

orwashams:
React Strict Mode causes the component to mount twice in the dev environment, so there’s definitely some code that can’t handle that.

Perhaps some cleanup or leftover state is making it non-idempotent, hence the issue only occurs for some users and not others—it could be on tldraw's side or Tiptap's. 

I’ve tried importing the Tldraw component dynamically and using different package managers (pnpm, bun, npm) but neither solved the problem. I’m using Next.js 15.2.1 and tldraw 3.10.2. 

Thankfully, the issue disappears in production.

SomeHats:
@orwashams would you be able to share a repo or stackblitz etc. that demonstrates the issue?

orwashams:
https://github.com/orwashams/testing-tldraw

A fresh NextJS app I just made, with default options upon creating the project, hope it helps.

[...]

PR #5689, opened 2025-03-18
text: fix editing for React 19/StrictMode

mimecuvalo:
There were several layers of things that made this tricky to debug on our end:
- our nextjs-template repo (https://github.com/tldraw/nextjs-template) has an old yarn.lock that has accidentally been committed
- upgrading React to 19 was prevented by our Yarn version: https://github.com/tldraw/tldraw/pull/5687
- We've been using React 18.3 for a while which should be functionally (mostly?) the same as React 19 except warning about future issues. However our monorepo wasn't exhibiting the issue. It seems that Next.js 14.2 also exhibits the problem even if it's just React 18.3, because maybe they're doing some monkeypatching or something to make it more like React 19, not sure.

All that being said, to fix the text issue, we need to make sure we don't call `setRichTextEditor` in succession because in turn calls `editor.destroy()` which makes a mess.

fixes https://github.com/tldraw/tldraw/issues/5611#issuecomment-2726908498

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

We do want to have some better testing of React 18 vs React 19 but difficult to setup at the moment.

### Release notes

- Fix developing with StrictMode + React 19 when editing text.


an5rag (approved):

SomeHats (approved):
nice one! thanks mime :)

mimecuvalo:
It's actually not working at the moment, I'm discovering 😞 I've been looking at alternates all morning ugggh

PR #5690, opened 2025-03-18
templates: rm yarn.lock files

mimecuvalo:
these have been pinned to hella old versions 😬 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


SomeHats (approved):
ty! Could we also add these to `.gitignore`? Not in the template folders themselves as that'll get copied across, but maybe in `templates/.gitignore`?

PR #5693, opened 2025-03-19
Draw shape: Guard against zero pressure pointer events

ds300:
We had a user of a Xiaomi tablet report that they couldn't draw anything, and after [much debugging](https://discord.com/channels/859816885297741824/859816885801713728/1350835984736059455) we discovered that their tablet was setting the pen pressure to 0 for every point. Interesting choice.

This PR attempts to account for that in a couple of ways

1. if the first point on a line has a pressure of 0, assume the device is broken and use pressure simulation.
2. if most of the points on a line get stripped out when filtering for low pressure at the start and end of the line, use the original points as they were.

### Change type

- [x] `other`

### Release notes

- Tweaked the draw shape svg rendering to be more forgiving for tablets which report low pen pressures.
- Fix an edge case where some tablets report 0 pressure for all pointer move events.

### DIFF
--- a/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
+++ b/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
@@ -168,7 +168,9 @@ export class Drawing extends StateNode {
 		const { z = 0.5 } = this.info.point
 
 		this.isPen = isPen
-		this.isPenOrStylus = isPen || (z > 0 && z < 0.5) || (z > 0.5 && z < 1)
+		// if z === 0 on the initial point, treat this pen as a mouse because it's likely a broken pen
+		// or a broken OS.
+		this.isPenOrStylus = (isPen && z !== 0) || (z > 0 && z < 0.5) || (z > 0.5 && z < 1)
 
 		const pressure = this.isPenOrStylus ? z * 1.25 : 0.5
 
--- a/packages/tldraw/src/lib/shapes/shared/freehand/getStrokePoints.ts
+++ b/packages/tldraw/src/lib/shapes/shared/freehand/getStrokePoints.ts
@@ -52,6 +52,11 @@ export function getStrokePoints(
 		}
 	}
 
+	// if we accidentally stripped out most of the line, revert to the original points
+	if (!simulatePressure && pts.length < rawInputPoints.length * 0.5) {
+		pts = rawInputPoints.map(Vec.From)
+	}
+
 	if (pts.length === 0)
 		return [
 			{


ds300:
hmm testing with light strokes on a samsung tablet it can sometimes end a line with a 0 pressure point, which suggests that it might also sometimes start a line with a 0 pressure point. or if not samsung then maybe other tablets.

ds300:
also looking at the lines the affected user drew with this pr's preview deploy, all the points still have `z: 0` and it's not doing pressure simulation for some reason (should be 0.5 and pressure simulation) so i must not have been faking their devices behavior correctly while testing.

packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
```diff
@@ -168,7 +168,9 @@ export class Drawing extends StateNode {
 		const { z = 0.5 } = this.info.point
 
 		this.isPen = isPen
-		this.isPenOrStylus = isPen || (z > 0 && z < 0.5) || (z > 0.5 && z < 1)
+		// if z === 0 on the initial point, treat this pen as a mouse because it's likely a broken pen
+		// or a broken OS.
+		this.isPenOrStylus = (isPen && z !== 0) || (z > 0 && z < 0.5) || (z > 0.5 && z < 1)
```
steveruizok, line 173: We'd need to test this across browsers. IIRC Safari (?) reports zero pressure in unusual scenarios, such as at the start of lines when using the Apple Pencil. I vaguely remember removing a zero check.

I actually wouldn't mind a "Input device: auto / mouse / pen" setting, similar to the "mousewheel behavior". Better input controls have been on my list forever.

steveruizok:
Do we still want to work on this?

steveruizok:
I replaced the strip-and-revert approach in `getStrokePoints` with a simpler pressure clamp.

**Why:** The original approach (strip low-pressure points from start/end, then revert if >50% were stripped) had two issues:
- The 50% threshold is arbitrary and fragile
- Reverting puts back points with `z=0`, which still produce near-zero-width stroke sections downstream in `setStrokePointRadii` — so the stroke doesn't vanish entirely, but it looks broken

**What changed:** Instead of stripping points and potentially reverting, we now clamp any point with `z < 0.025` up to `0.025`. This means:
- No points are ever removed, so there's no risk of stripping most of the stroke
- Zero-pressure points render as very thin (minimum pressure) rather than invisible
- Existing saved documents with `z=0` data get fixed at render time
- Legitimate light-pressure strokes still work naturally — they just can't go below 2.5%

The `isPenOrStylus` guard on the first point (Change 1) is kept as-is — if the pen reports zero pressure from the start, falling back to simulated pressure mode is the right call.

PR #5696, opened 2025-03-19
Add & fix missing state docs

TodePond:
This PR adds and fixes some missing state docs on our API reference. It also adds a new command `yarn refresh-api`, which you can use within the `apps/docs` folder to quickly refresh local API pages (after running something like `yarn build-api` at root).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Docs: Added more docs to the `@tldraw/state` and `@tldraw/state-react` API reference,

### DIFF
--- a/packages/state-react/src/lib/useAtom.ts
+++ b/packages/state-react/src/lib/useAtom.ts
@@ -4,7 +4,7 @@ import { useState } from 'react'
 /**
  * Creates a new atom and returns it. The atom will be created only once.
  *
- * See {@link state#atom}
+ * See `atom`.
  *
  * @example
  * ```ts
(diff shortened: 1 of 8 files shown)


packages/state-react/src/lib/useAtom.ts
```diff
@@ -4,7 +4,7 @@ import { useState } from 'react'
 /**
  * Creates a new atom and returns it. The atom will be created only once.
  *
- * See {@link state#atom}
```
SomeHats, line 7: You want `{@link @tldraw/state#atom}` I believe - probably better for this to link?

packages/state-react/src/lib/useComputed.ts
```diff
@@ -2,11 +2,12 @@
 import { Computed, ComputedOptions, computed } from '@tldraw/state'
 import { useMemo } from 'react'
 
+/** @public */
+export function useComputed<Value>(name: string, compute: () => Value, deps: any[]): Computed<Value>
+
 /**
  * Creates a new computed signal and returns it. The computed signal will be created only once.
```
SomeHats, line 9:
```suggestion
 * Creates a new {@link @tldraw/state#computed} signal and returns it. The computed signal will be created only once.
```

SomeHats (approved):
love it! i'm assuming the link removal was because the existing syntax wasn't working correctly, but if it was deliberate please ignore my suggestions

TodePond:
Thanks, yeah those type of links are completely broken even when we prefix it with the package
eg: https://tldraw.dev/reference/editor/TldrawEditorWithStoreProps
image

See ticket: https://linear.app/tldraw/issue/INT-1015/fix-broken-links-in-api-reference

PR #5714, opened 2025-03-20
Improve search config

TodePond:
This PR changes our algolia config to fix some common failure cases and improve experience. It also adds some keywords to pages.

- Section name is now considered in search ranking. This means that searching for "Examples" will actually come up with examples first.
- We now group sections of results into threes instead of fours. This means you don't have to scroll so far down to find what you need (which is usually an example).
- We now support advanced search syntax, which means you can use quotes to search for an exact string. (You can also use a minus to specify something to exclude from the results).

image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/docs/scripts/update-algolia-index.ts
+++ b/apps/docs/scripts/update-algolia-index.ts
@@ -258,8 +258,22 @@ async function updateAlgoliaIndex() {
 				{ autoGenerateObjectIDIfNotExist: true }
 			)
 			await index.setSettings({
-				searchableAttributes: ['titleHeadingFirst', 'title', 'description', 'keywords', 'content'],
-				camelCaseAttributes: ['titleHeadingFirst', 'title', 'description', 'keywords', 'content'],
+				searchableAttributes: [
+					'section',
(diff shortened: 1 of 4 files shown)


apps/docs/scripts/update-algolia-index.ts
```diff
@@ -273,8 +287,9 @@ async function updateAlgoliaIndex() {
 					'asc(articleIndex)',
 					'asc(headingIndex)',
 				],
-				distinct: 4,
+				distinct: 3,
 				attributeForDistinct: 'section',
```
SomeHats, line 291: idk if you've played with removing these two entirely but it could be worth a look (although a bit weird having results from different sections interleaved)

SomeHats (approved):

apps/docs/scripts/update-algolia-index.ts
TodePond: Maybe! Will probably revisit, but this feels like a small drive-by improvement for now

PR #5715, opened 2025-03-20
Add example for mass rich text updates

ds300:
This PR adds a naive example based on no research about how to toggle bold/italic/highlight for a whole text shape at once, and how to integrate that logic into a simple context toolbar that can operate on multiple shapes at once.

### Change type

- [x] `other`


SomeHats:
some of your utilities could probably be replaced if you did:
```ts
import {Node} from '@tiptap/pm/model'

const node = Node.fromJson(richText)

...
```

There are a bunch of utilities for checking/traversing nodes, and e.g. applying marks to them: https://prosemirror.net/docs/ref/#model

PR #5719, opened 2025-03-20
Add dedicated search page

TodePond:
Pressing the enter key when using the search bar takes you through to a dedicated search page... `/search?query=[query]`.

We have a large API surface! Often, our small search results window doesn't have enough room to display all the results we need. It's hard to navigate and cuts off too many things. A dedicated search page makes this easier. We can fit more on the screen, and there's room for us to show more results. We can also share search queries with people.

This PR also includes some other small changes:
- Fixed mobile menu bar sliding off the screen when you scroll.
- Fixed page title on some docs pages.
- Removed some unused helpers.
- Renamed some search components for clarity.
- Added a tiny hover effect to search results (affects both the full page and the dialog).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. On the docs site...
2. Open search by pressing Cmd+K, or by clicking on the search bar on the docs page...
3. Type in a query, eg: "test"
4. Press enter.
5. It should take you to a dedicated search page.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Docs: Added full page search.


mimecuvalo (approved):
👍 pre-approving.
my design feedback would be that it's hard to tell that i'm on a search page. it's missing a header or something that says "Results". esp. if i hit Enter accidentally without having selected an item, arriving to the page might be a little surprising/lack context.

PR #5731, opened 2025-03-22
less eager hard reboot

ds300:
Looking at what was going on during the deploy yesterday, I noticed that there were thousands of hard reboots while there were only a handful of times where the replicator specifically requested a hard reboot. That meant we were being too eager to do a full hard reboot when encountering errors during deploys. We didn't see any errors on sentry during the last deploy so it seems like literally all of the failures we interpreted as 'errors' during deploy were due to timeouts, i.e. the replicator taking more than 5 seconds to service user registration requests. This makes sense. if we have 1500 users connected during a deploy and let's be very optimistic and say it takes 50ms to service a registration request, that's already 75 seconds if done serially, and parallelism isn't going to cut down enough to make it < 5 seconds guaranteed if all the user DOs are restarting over the course of a short span of time (10 seconds to a minute ish?)

So yes, we do need to go a bit easier on the replicator while there's only one of them, and we shouldn't immediately do a hard reboot if there was a timeout while trying to connect, that will only exacerbate the problem.

This pr

- bumps the boot timeout from 5s to 10s
- makes is so that booting needs to fail 4 times to trigger a hard reboot, instead of just 1 time.
- adds some more nuanced tracking for the reboot event

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -76,6 +76,8 @@ export class TLDrawDurableObject extends DurableObject {
 
 	_room: Promise<TLSocketRoom<TLRecord, SessionMeta>> | null = null
 
+	sentry: ReturnType<typeof createSentry> | null = null
+
 	getRoom() {
 		if (!this._documentInfo) {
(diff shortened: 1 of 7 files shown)


ds300:
closing in favor of #5745

PR #5734, opened 2025-03-23
When editing a text shape, don't mount the text editor for non-editing empty shapes unless they're hovered

steveruizok:
This PR improves text editing.

## The bug

Previously, we would mount all of the text editors for all of the text-editable shapes when the user began editing any text-editable shape. For large projects this would block the main thread as all of those text editors were mounted. Even on my Very Fast Laptop this would introduce ~200-500ms of lag between when I press "Enter" to begin editing a shape and when the app would begin responding to my key strokes. This often dropped the first letter of whatever I was typing.

## The fix

We now mount the text editor for a shape if it's:
- editing; or
- has text; or
- is hovered

If a shape is neither of those, then we don't mount the text editor until the user hovers the shape. From a user perspective, the interactions are identical to what they were before.

## Proof

Before: (starting to edit a shape with ~1000 empty shapes on the page)

image

After:

There's nothing to see, we don't drop any frames.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create many empty shapes
2. Begin editing any shape.

### Release notes

- Fixed a bug causing a performance delay when editing text.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -897,14 +897,15 @@ input,
 	cursor: var(--tl-cursor-default);
 }
 
+.tl-rich-text[data-is-ready-for-editing='true'],
+.tl-text-wrapper[data-is-ready-for-editing='true'] .tl-text-input {
+	cursor: var(--tl-cursor-text);
+}
+
 .tl-text-input::selection {
 	background: var(--color-selected);
 	color: var(--color-selected-contrast);
 	text-shadow: none;
-}
-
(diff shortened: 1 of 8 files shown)


mimecuvalo (commented):
thanks for the fix! this is lookin' good - i'm gonna rename some of it because it's gonna start getting confusing with the _other_ `iseditinganything` found in `DefaultCanvas` and we don't want to start getting mixed up about that.

PR #5735, opened 2025-03-23
Cache the fonts extracted from rich text.

steveruizok:
This PR improves the performance of shapes with rich text content.

## The bug

Due to reasons I cannot understand, we extract the fonts from rich text on every frame while translating shapes. Maybe we shouldn't, I don't know. Anyway, this is expensive when many shapes are moving around. Actually, not even that many. Try it yourself!

## The fix

The fix here is the same as #5658, which is to skip the work if the text is empt; or use a weak cache to re-use the result from last time.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create lots of geo shapes.
2. Drag all the geo shapes around.

### Release notes

- Improved performance while editing many geo shapes or text shapes.

### DIFF
--- a/packages/editor/src/lib/utils/richText.ts
+++ b/packages/editor/src/lib/utils/richText.ts
@@ -1,6 +1,7 @@
 import { getSchema, JSONContent, Editor as TTEditor } from '@tiptap/core'
 import { Node } from '@tiptap/pm/model'
 import { EditorProviderProps } from '@tiptap/react'
+import { EMPTY_ARRAY } from '@tldraw/state'
 import { TLRichText } from '@tldraw/tlschema'
 import { assert, WeakCache } from '@tldraw/utils'
 import { Editor } from '../editor/Editor'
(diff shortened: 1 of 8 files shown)


packages/editor/src/lib/utils/richText.ts
```diff
@@ -39,34 +40,53 @@ export type RichTextFontVisitor = (
...
 export function getFontsFromRichText(
 	editor: Editor,
 	richText: TLRichText,
 	initialState: RichTextFontVisitorState
 ) {
-	const { tipTapConfig, addFontsFromNode } = editor.getTextOptions()
-	assert(tipTapConfig, 'textOptions.tipTapConfig must be set to use rich text')
-	assert(addFontsFromNode, 'textOptions.addFontsFromNode must be set to use rich text')
+	if (isEmptyRichText(richText)) return EMPTY_ARRAY
+	return fontsFromTextCache.get(richText, () => {
```
SomeHats, line 66: I don't think this is quite right - the fonts found will depend largely on `initialState`, which encodes stuff like the initial font (from the styles panel). Right now, i think if you change the font, you won't invalidate the cache.

SomeHats (changes requested):

packages/editor/src/lib/utils/richText.ts
SomeHats: You can probably see this by removing the font preloading that we do in `onMount` in `Tldraw`

mimecuvalo:
pushed up the fix @SomeHats

SomeHats (approved):
nice! ty

PR #5743, opened 2025-03-24
fix translation perf regression from rich text font detection

SomeHats:
We were re-computed this way too eagerly. We only need to check when the props change.

### Change type

- [x] `bugfix`

### Release notes

- Fix a performance regression when dragging many shapes at the same time.

### DIFF
--- a/packages/editor/src/lib/editor/managers/FontManager.ts
+++ b/packages/editor/src/lib/editor/managers/FontManager.ts
@@ -94,7 +94,7 @@ export class FontManager {
 				const shapeUtil = this.editor.getShapeUtil(shape)
 				return shapeUtil.getFontFaces(shape)
 			},
-			{ areResultsEqual: areArraysShallowEqual }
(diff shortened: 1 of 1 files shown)


mimecuvalo (approved):
👍 but this does make me wonder what the right defaults should be for `createComputedCache`. it feels footgun-y for us, let alone others, and it makes me wonder if it shouldn't force areResultsEqual/areRecordsEqual to be considered when calling the function (if nothing to just specify explicitly undefined or noop).

like, for example, `editor.store.createComputedCache('handle snap geometry'` in HandleSnaps. should this also consider areRecordsEqual? it's unclear to me if it should be added there or whether it's only on a case-by-case basis.

SomeHats:
Yeah, it's a bit complicated and nuanced. I would love if we had an easier way for us to track specifically which bits of the shape a method uses. for handle snaps, it's not really an issue because we don't need it on every frame i guess. not sure what the correct thing is tbh

PR #5746, opened 2025-03-25
Better whyAmIRunning

ds300:
@SomeHats I think this should fix the flakiness you were seeing. can you confirm?

Leaving as draft because I still need to check the perf and probably add tests now that it touches normal code paths.

### Change type

- [x] `other`

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -6,6 +6,7 @@ import {
 	react,
 	transact,
 	unsafe__withoutCapture,
+	whyAmIRunning,
 } from '@tldraw/state'
 import {
 	ComputedCache,
@@ -4424,6 +4425,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	/** @internal */
 	@computed private _getShapePageBoundsCache(): ComputedCache<Box, TLShape> {
 		return this.store.createComputedCache<Box, TLShape>('pageBoundsCache', (shape) => {
+			whyAmIRunning()
 			const pageTransform = this._getShapePageTransformCache().get(shape.id)
 
 			if (!pageTransform) return new Box()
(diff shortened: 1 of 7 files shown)


SomeHats:
this works really well! thanks david :)

SomeHats:
For perf concerns I wonder if it would be worth using like a NODE_ENV check to strip the debug code entirely in production? 

We could probably inline that for now, but we could look at production/development versions of the library in the future. it'd be a good way to strip a bunch of unused dev tools from tldraws bundle size too

ds300:
i did some benchmarking of before and after, there's no real difference so will just add tests and call it a day

packages/editor/src/lib/components/Shape.tsx
```diff
@@ -104,22 +104,18 @@ export const Shape = memo(function Shape({
...
+		const bgContainer = bgContainerRef.current
+
+		// Opacity
+		setStyleProperty(container, 'opacity', opacity)
+		setStyleProperty(bgContainer, 'opacity', opacity)
+
+		// Z-Index
+		setStyleProperty(container, 'z-index', index)
+		setStyleProperty(bgContainer, 'z-index', backgroundIndex)
+	}, [opacity, index, backgroundIndex])
```
SomeHats, line 118: any reason not to have these set in react at this point?

ds300: might be better for perf this way? i'll test and if it's the same we can just use regular react

ds300: surprisingly there's a significant difference. Changing the opacity of 600 shapes took ~85ms on average with useLayoutEffect and ~95ms on average with the style prop

ds300:
Alright I made this much better because now it

- lists the entire path of changes, up to root atoms. before it only listed the immediate parent changes, so it took longer to debug memoization issues.
- doesn't pollute any existing code besides adding two null checks, one in startCapturingParents and one in stopCapturingParents.

Issue #4945, opened 2024-11-18
Camera zoom speed inverted on Safari trackpad


Karavil:
### What happened?

Safari (and possibly other non-Chromium browsers) exhibit incorrect zoom speed behavior with trackpad pinch-to-zoom when adjusting camera options. When setting lower zoom speeds, the camera paradoxically zooms **faster** and becomes unstable.

## Reproduction
1. Open Safari
2. Visit [Camera Options Demo](https://examples.tldraw.com/camera-options/full?preserveFocus=true)
3. Find the "Zoom Speed" control in the demo interface
4. Set value to `0.5` 
5. Try zooming using trackpad pinch gesture
  *(Note: May also affect scroll wheel zoom, but untested)*

https://github.com/user-attachments/assets/4ee128ff-50bb-441f-a50c-663e70f7dd79

### Expected
- Zoom should be 2x *slower* than default (matching Chrome)
- Motion should remain smooth throughout zoom

#### Safari
- ~2x **faster** zoom than default
- Choppy/unstable motion
- Zoom sometimes resets to start position

*Note*: After noticing the inverse relationship, I tried setting `Zoom Speed` to `2` to attempt slowing it down by 50%. While this partially works (zoom is slower), after each zoom gesture completes, the view snaps back to the original zoom level (1).

#### Chrome
- Correctly shows 2x slower zoom
- Motion remains smooth
- Zoom level persists after gesture

## Environment
- macOS Sequoia 15.1 (24B83), reproduced on Sonoma
- Safari latest
- May affect other non-Chromium browsers

Really interested in understanding TLDraw's camera system better and potentially contributing a fix. Happy to pair on this or take pointers toward relevant codebase sections for investigation :)

### Where did this happen?

The developer package

### Contact Details

alp@hey.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

steveruizok:
Oh good catch! Definitely not working correctly.

SomeHats:
Hey folks - the fix for this is now out in 3.11.1

PR #5749, opened 2025-03-25
Fix zoom speed on pinch events

arpit-goblins:

### Change type

- [ y] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan
(I have done these steps, but you're more than welcome to recreate them)

1. Open TlDraw on either safari or on a touchscreen device
2. Navigate to /camera-options
3. Check that zooming & panning work as intended when using safari trackpad or pinching on touchscreen
4. Change around zoomSpeed & retest.
5. Consider also testing existing zooming & panning work when not using gestures

- [N/A] Unit tests
- [N/A] End to end tests

### Release notes

- Fixed a bug where gesture pinch events (e.g. safari trackpad, touchscreen) would not interact correctly with zoomSpeed, leading to jerky and unreliable motion.
[Issue](https://github.com/tldraw/tldraw/issues/4945)

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -9929,6 +9929,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 							const { x, y } = this.inputs.currentScreenPoint
 							let delta = dz
 
+							// If we're forcing zoom, then we need to do the wheel normalization math here
 							if (wheelBehavior === 'zoom') {
 								if (Math.abs(dy) > 10) {
 									delta = (10 * Math.sign(dy)) / 100
@@ -9938,14 +9939,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 							}
 
 							const zoom = cz + (delta ?? 0) * zoomSpeed * cz
-							this._setCamera(
-								new Vec(
-									cx + (x / zoom - x) - (x / cz - x),
-									cy + (y / zoom - y) - (y / cz - y),
-									zoom
-								),
-								{ immediate: true }
-							)
+							this._setCamera(new Vec(cx + x / zoom - x / cz, cy + y / zoom - y / cz, zoom), {
+								immediate: true,
+							})
 							this.maybeTrackPerformance('Zooming')
 							return
 						}
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -9676,7 +9676,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	private _restoreToolId = 'select'
 
 	/** @internal */
-	private _pinchStart = 1
+	private _pinchZ = 1
 
 	/** @internal */
 	private _didPinch = false
@@ -9819,7 +9819,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 						if (inputs.isPinching) return
 
 						if (!inputs.isEditing) {
-							this._pinchStart = this.getCamera().z
+							this._pinchZ = this.getCamera().z
 							if (!this._selectedShapeIdsAtPointerDown.length) {
 								this._selectedShapeIdsAtPointerDown = [...pageState.selectedShapeIds]
 							}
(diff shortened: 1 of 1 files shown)


arpit-goblins:

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9939,14 +9943,9 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 							const zoom = cz + (delta ?? 0) * zoomSpeed * cz
-							this._setCamera(
-								new Vec(
-									cx + (x / zoom - x) - (x / cz - x),
-									cy + (y / zoom - y) - (y / cz - y),
-									zoom
-								),
-								{ immediate: true }
-							)
+							this._setCamera(new Vec(cx + x / zoom - x / cz, cy + y / zoom - y / cz, zoom), {
```
SomeHats, line 9946: took me a sec to figure out what was going on here but good catch on simplifying this!

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9859,12 +9859,17 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 						const { x: cx, y: cy, z: cz } = unsafe__withoutCapture(() => this.getCamera())
 
+						let delta = z - this._pinchZ
+						this._pinchZ = z
+
 						const { panSpeed, zoomSpeed } = cameraOptions
+						const zoom = cz + (delta ?? 0) * zoomSpeed * cz
```
SomeHats, line 9866: this isn't quite right at the moment unfortunately. The perceived zoom speed increases / decreases as I zoom in & out with this. It's particularly noticeable when zooming out:


https://github.com/user-attachments/assets/e387b696-4cb1-4490-a9fd-9fd46e649ea7

The perceived zoom speed should remain the same regardless of zoom level.

SomeHats (changes requested):
Thanks for this!

packages/editor/src/lib/editor/Editor.ts
arpit-goblins: That makes sense!

I've fixed this in some sense. However, I'm wondering if I should override the way Gesture events are handled to be purely linear, rather than a scale.

What i mean by this is - suppose you start your fingers 1 inch apart. If you move them 0.1 inches apart, this will register as a 'z value' of 1.1, e.g. zooming you out 10% (scaled by zoom speed).
However, if you start your fingers 2 inches apart, and then move your fingers 0.1 inches apart, this will register as a 'z value' of 1.05, zooming you out 5%.

I could scale linearly so that it solely zooms you based off the delta of how much your fingers move, irrespective of how close/far your fingers start.

Pros:

zoomSpeed is consistent on gesture pinch events, regardless of how close or far you start your fingers.
Without this, the user can almost always zoom in/out faster than they do the other (depending on whether i calculate delta as z - this._pinchZ or z/this._pinchZ - 1. With the former, zooming in is super easy (start with fingers really close together, and you can scale z really quickly by pinching your fingers really far apart.
Emulates non-Safari trackpad zooming
Cons:

Might be intuitively different than expected behavior of zooming on touchscreens/safari trackpad
I would have to update TLPinchEventInfo to give details about the exact positions of the initial pointers, rather than just the centerpoint of them.

SomeHats: one thing i'm noticing is that these changes have made zooming feel messed up on mobile, we might have to dig a bit deeper into the gesture detection stuff we're using

arpit-goblins: > one thing i'm noticing is that these changes have made zooming feel messed up on mobile, we might have to dig a bit deeper into the gesture detection stuff we're using

I dag deeper.

steveruizok:
Sorry @arpit-goblins, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #5750, opened 2025-03-25
Update validation.ts to allow blob: as src for Image

Pierre-Mike:
This allows local browser images to be used as a Source. 

  
Screenshot 2025-03-25 at 22 34 22

sorry for the unit test. I haven't made this repo work locally yet. 

### Change type

- [ x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. load image from a Blob 
2.

- [ ] Unit tests 
- [ ] End to end tests

### Release notes

- Fixed a bug to allow Blob to be use as Image/asset source

### DIFF
--- a/packages/validate/src/lib/validation.ts
+++ b/packages/validate/src/lib/validation.ts
@@ -974,7 +974,7 @@ export const linkUrl = string.check((value) => {
 })
 
 // N.B. asset: is a reference to the local indexedDB object store.
-const validSrcProtocols = new Set(['http:', 'https:', 'data:', 'asset:'])
+const validSrcProtocols = new Set(['http:', 'https:', 'data:', 'asset:', 'blob:'])
 
 /**
  * Validates that a valid is a url safe to load as an asset.


Pierre-Mike:

SomeHats:
Hey @Pierre-Mike! Thank you for creating a PR for this.

This is actually working as intended though - it's a deliberate decision on our part not to allow `blob:` URLs in the store. This is because in the vast majority of use-cases, the data in the store becomes what gets stored on our users servers to then restore in a later session, or sync to other users. Because blob URLs only last for a single page-load in a single browser tab, they should never end up in the store for these use-cases.

There is a workaround though! We allow a special custom `asset:` protocol for assets. This is used for example in our `persistenceKey` function that stores the document in `IndexedDb`: when you drop in an image, we store it in indexed DB against it's ID. Then, when the image gets rendered, it's read from indexeddb, a blob URL is created, and that's what's returned from the resolver. 

The trick here is to prefix your blob URLs with `asset:`, then use a custom (`assetStore`)[https://tldraw.dev/reference/tlschema/TLAssetStore] with a `resolve` method that'll strip the `asset` prefix.

Out of curiosity, what's your use-case for storing blob URLs in the store?

Pierre-Mike:
@SomeHats, 

Thank you for your response and the workaround. I may not be explaining my issue clearly, so let me clarify my use case:

- I have a list of images.
- I need to load each image and lock it in place. However, to load an asset, I need the dimensions (width and height), which I cannot determine until I download the image.

To address this, I used React Query to load the image myself, returning a Blob that I can reuse on the canvas. This approach prevents me from loading the image twice—once to get the dimensions and once to display it on the canvas. 

In the long term, I’d like to load all images in the background, prioritizing the image currently on the canvas.

Do you have any recommendations? While the documentation is helpful, it contains a lot of information, and I sometimes find it overwhelming. Thank you!

SomeHats:
I responded with some notes on that other issue, but I'd also say to check whether loading the images twice actually causes them to be loaded twice or if there's some caching etc. going on - browsers are often pretty smart about this sort of stuff.

If you don't need to manipulate your images at all, you could try not making shapes for them at all, and instead rendering them on the canvas directly. Take a look at the 'things on the canvas' example for that

PR #5754, opened 2025-03-26
Geometry2d Improvements

SomeHats:
This diff adds a number of `Geometry2d` improvements back-ported from my work on elbow arrows.

1. Intersection helpers. We have hit tests, distance to, nearest point etc. on geometry, but no intersections. This diff adds support for `Geometry2d.intersectLineSegment` and `Geometry2d.intersectCircle`. We previously were using these downstream in kind of a hack way, but having them directly on the geometry and able to play nicer with groups etc. is very helpful.
2. Transformation. `Geometry2d.transform` allows you to efficiently transform a geometry by some matrix. Where possible, we avoid transforming every single point and instead forward the methods to the original untransformed geometry. This also allows for some efficiency gains by e.g. caching geometries in page space. For now, I've only ported some really obvious / simple use-cases over to using this, but there are many parts of the code that could now be simplified by switching to transformed geometries.
3. Filters. Almost all geometry methods do some sort of filtering - e.g. to ignore labels, etc. Sometimes this was hard-coded, sometimes method accepted an optional boolean `includeFilters` parameter. Now, all methods than can have filters applied accept a filters argument which determines which parts do/don't get included in the shape. New here (& motivating this change) is a new way of designating parts of a group as "internal" geometry - e.g. the lines within a geo shape that we might not want to partake in arrow binding. 

### Change type

- [x] `api`

### Release notes

- It's now easier to work with `Geometry2d` objects, with methods for intersections, transforming geometries, and filtering.

### DIFF
--- a/packages/editor/src/lib/primitives/geometry/Geometry2d.ts
+++ b/packages/editor/src/lib/primitives/geometry/Geometry2d.ts
@@ -1,8 +1,14 @@
+import { assert } from '@tldraw/utils'
 import { Box } from '../Box'
 import { Mat, MatModel } from '../Mat'
 import { Vec, VecLike } from '../Vec'
-import { intersectLineSegmentPolygon, intersectLineSegmentPolyline } from '../intersect'
-import { pointInPolygon } from '../utils'
+import {
+	intersectCirclePolygon,
+	intersectCirclePolyline,
+	intersectLineSegmentPolygon,
+	intersectLineSegmentPolyline,
+} from '../intersect'
+import { approximately, pointInPolygon } from '../utils'
 
 /** @public */
 export interface Geometry2dFilters {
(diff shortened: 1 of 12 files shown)


packages/editor/src/lib/primitives/geometry/Geometry2d.ts
```diff
@@ -73,10 +109,30 @@ export abstract class Geometry2d {
...
+		if (intersections) yield* intersections
+	}
+
+	*intersectCircle(center: VecLike, radius: number, filters?: Geometry2dFilters) {
+		if (this.isExcludedByFilter(filters)) return
+		const intersections = this.isClosed
+			? intersectCirclePolygon(center, radius, this.vertices)
+			: intersectCirclePolyline(center, radius, this.vertices)
+
+		if (intersections) yield* intersections
```
ds300, line 132: why use a generator here? I'm a bit worried about perf vs just returning the actual array. Geometry operations are liable to be called frequently in hot loops and on code paths that block UI updates, so we should probably avoid this extra allocation and indirection if it's not really needed

ds300 (approved):

packages/editor/src/lib/primitives/geometry/Geometry2d.ts
SomeHats: My main thinking here was groups, where you might have intersections from several pieces of geometry that you'd otherwise need to merge, but i guess that's maybe less allocations overall than using a generator (?)

PR #5761, opened 2025-03-27
a11y: navigable shapes

mimecuvalo:
As part of a [larger push](https://github.com/tldraw/tldraw/issues/5215) to add accessibility to our SDK, a big piece of that work is being able to navigate through our shapes in some kind of predictable fashion. This builds upon @Taha-Hassan-Git 's great work and knowledge in this area, thanks man. :tip-o-the-hat:

Things that were tackled in this PR:
- navigating shapes using the Tab key, when in the Select tool.
- navigating shapes using Cmd/Ctrl+Arrow keys, when in the Select tool.
- only allowing certain shapes to be navigated to. We ignore draw/highlighter/arrow/group/line. Groups need exploration and will be tackled later.
- panning the camera to the selected shape, but avoiding doing so in a jarring way. We don't center the shape to avoid too much whiplashy-ness.

An initial foray into this was relaying purely on DOM but it had a bunch of browser quirks which forced making this purely a programmatic control on our end. Things like ensuring culled shapes are still accessible even though they're not rendered was one of the issues but also tab order became unpredictable at times which steered me away from that direction.

We coud have considered using something like rbush for some spatial indexing of the shapes. For the intents and purposes of this PR, it seemed like overkill at the moment. But we might cross that bridge down the line, we'll see.

The reading-direction heuristics are a combination of dividing the pages into rows and then looking at distance and angles to see what is the spatially "next" shape to be read. It takes _all_ of the shapes and sorts them into a logical order so that nothing is missed/skipped when tabbing around.
The directional-arrow heuristics don't divide things into rows and don't create a sorted set of shapes. Instead, they decide based on the current shape and direction which is the next spatially to go to, depending on distance+angle.

There's a decent amount of nuance in this kind of navigation but it's not all covered in this PR, for separate PRs, we'll look at:
- [x] adding a "skipping to content" button
- [ ] question whether maybe directional navigation visits ‘canTabTo’ shapes, maybe yes?
- [ ] tackling what Enter/Escape should do when on the canvas shapes
- [ ] how to deal with hierarchy / parent-child / frame / group shapes
- [ ] and more

https://github.com/user-attachments/assets/49b6b34e-2553-4047-846f-5d3383e1e3c6

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [x] End to end tests

### Release notes

- a11y: navigable shapes using Tab and Cmd/Ctrl+Arrow


Taha-Hassan-Git:
Omg!!

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3019,6 +3019,41 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * Zoom the camera to the current selection if offscreen.
+	 *
+	 * @example
+	 * ```ts
+	 * editor.zoomToShapeIfOffscreen()
+	 * ```
+	 *
+	 * @public
+	 */
+	zoomToShapeIfOffscreen(padding = 16) {
```
SomeHats, line 3032: should this be called `zoomToSelectionIfOffscreen`?

packages/editor/src/lib/utils/shapeNavigation.ts
```diff
@@ -0,0 +1,199 @@
...
+import { Vec } from '../primitives/Vec'
+
+const SHALLOW_ANGLE = 20
+const ROW_THRESHOLD = 100
+
+/** @public */
+export type AdjacentDirection = 'next' | 'prev' | 'left' | 'right' | 'up' | 'down'
+
+/** @public */
+export function selectAdjacentShape(editor: Editor, direction: AdjacentDirection) {
```
SomeHats, line 12: Instead of having these as free functions, could we either make these methods on `Editor` or put them in a `manager` like `editor.snaps`? So you'd have either `editor.selectAdjacentShape(...)` or `editor.navigation.selectAdjacentShape()`? I think free functions are probably the least discoverable of the possible API options here, they're no good for autocomplete-driven-development

packages/editor/src/lib/utils/shapeNavigation.ts
```diff
@@ -0,0 +1,199 @@
...
+	editor.zoomToShapeIfOffscreen(256)
+}
+
+/**
+ * Generates a reading order for shapes based on rows grouping.
+ * Tries to keep a natural reading order (left-to-right, top-to-bottom).
+ *
+ * @public
+ */
+export function getShapesInReadingOrder(editor: Editor): TLShape[] {
```
SomeHats, line 44: Can we rename this to `getCurrentPageShapesInReadingOrder`?

Also, is it worth sticking this in a `computed` so it's cached between invocations? I can imagine this getting expensive on big documents. 

If you keep this as a free function, you can do this with:
```
const shapesInReadingOrderStore = new WeakCache<Editor, Computed<TLShape[]>>()
export function getShapesInReadingOrder(editor: Editor) {
  return shapesInReadingOrderStore.get(editor, () => computed('shapes in reading order', () => {
    // existing impl goes here
  })).get()
}
```

or if you make it a method u can just stick `@computed` on it

packages/editor/src/lib/utils/shapeNavigation.ts
```diff
@@ -0,0 +1,199 @@
+import { TLShape, TLShapeId } from '@tldraw/tlschema'
+import { Editor } from '../editor/Editor'
+import { Vec } from '../primitives/Vec'
+
+const SHALLOW_ANGLE = 20
+const ROW_THRESHOLD = 100
```
SomeHats, line 6: should these be options??? idk

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1478,6 +1481,66 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					trackEvent('change-page', { source, direction: 'next' })
 				},
 			},
+			{
```
SomeHats, line 1484: not 100%, but i feel like these should be part of the select tool rather than part of the UI. I don't think i would expect disabling the UI to also disable keyboard navigation. wdyt?

SomeHats (commented):
this is really great work, thanks @mimecuvalo! i'll dive in deeper soon but i've left some initial thoughts :)

packages/editor/src/lib/utils/shapeNavigation.ts
mimecuvalo: yeah, I don't disagree! i'd prefer it honestly it on the editor as well. (even better for `editor.navigation.selectAdjacentShape()` but it feels like we always do a flat structure, with maybe camera as the exception?)

sometimes i can't tell if we just don't like putting too many things on the editor

packages/editor/src/lib/utils/shapeNavigation.ts
mimecuvalo: maybe? i think esp. if this is going in the Select tool that maybe it isn't, at least in this first iteration.

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: ok, yeah i think this is the better call 👍

packages/editor/src/lib/utils/shapeNavigation.ts
mimecuvalo: moved the logic into the Select Tool in the end

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1778,6 +1778,192 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		for (const row of rows) {
+			if (row.length <= 2) continue
+
+			for (let i = 0; i < row.length - 2; i++) {
+				const currentShape = row[i]
+				const nextShape = row[i + 1]
+				const nextNextShape = row[i + 2]
+
+				// Only consider adjustment if the next two shapes are relatively close to each other.
+				const dist1 = Vec.Dist(currentShape.center, nextShape.center)
```
mimecuvalo, line 1869: I'm debating in my head whether we need precision here and maybe `Dist2` would be fine, your thoughts @SomeHats ?

SomeHats: You can use dist2 here. A < B * 0.9 and A² < B² * 0.9² are equivalent (i think), so there's no loss of precision.

packages/tldraw/src/lib/ui/components/A11y.tsx
```diff
@@ -1,6 +1,38 @@
...
+
+	return (
+		<TldrawUiButton
+			ref={button}
+			type="normal"
+			tabIndex={1}
+			className="tl-skip-to-main-content"
+			onClick={handleNavigateToFirstShape}
+		>
+			{msg('a11y.skip-to-main-content')}
```
SomeHats, line 32: as discussed offline maybe tweak the language here?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1778,6 +1778,192 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	selectAdjacentShape(direction: TLAdjacentDirection) {
+		const currentShapeId = this.getSelectedShapeIds()[0]
+
+		let shapes: TLShape[] = []
+		let adjacentShapeId: TLShapeId
+		if (direction === 'next' || direction === 'prev') {
+			shapes = this.getCurrentPageShapesInReadingOrder()
+			const shapeIds = shapes.map((shape) => shape.id)
+
+			const currentIndex = shapeIds.indexOf(currentShapeId)
```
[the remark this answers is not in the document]
mimecuvalo: ehh, kind of - it's written this way more for maintainability and to highlight that it wraps around. so i'd have to write, like, a lookahead? seems gnarlier to parse for folks.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1778,6 +1778,192 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+				distance * 1.0 + // Base distance
+				offAxisDeviation * 2.0 + // Heavy penalty for off-axis deviation
+				(distance - directionalDistance) * 1.5 + // Penalty for diagonal distance
+				angleDeviation * 0.5 // Slight penalty for angular deviation
+
+			return { shape, score }
+		})
+
+		scoredShapes.sort((a, b) => a.score - b.score)
+		return scoredShapes[0].shape.id
```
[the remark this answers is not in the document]
mimecuvalo: more than a nit! great idea

packages/editor/src/lib/hooks/useDocumentEvents.ts
```diff
@@ -10,6 +10,11 @@ import { useEditor } from './useEditor'
 export function useDocumentEvents() {
 	const editor = useEditor()
 	const container = useContainer()
+	const hasSelectedShapes = useValue(
+		'hasSelectedShapes',
+		() => !!editor.getSelectedShapeIds().length,
+		[editor]
+	)
```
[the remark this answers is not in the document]
mimecuvalo: oh yeah, i have no idea why i did it this way, oops

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1778,6 +1778,192 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this
 	}
 
+	selectAdjacentShape(direction: TLAdjacentDirection) {
+		const currentShapeId = this.getSelectedShapeIds()[0]
```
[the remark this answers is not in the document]
mimecuvalo: you're right, it's sort of non-deterministic, lemme tweak this 👍

packages/editor/src/lib/editor/Editor.ts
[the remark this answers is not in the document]
SomeHats: ok! i don't really understand this, sorry - I thought `shapes.map(shape => shape.id).indexOf(someId)` and `shapes.findIndex(shape => shape.id === someId)` would always return the same value?

obvs not a big deal tho!

mimecuvalo:
talked with Alex offline and this is good to go 👍

PR #5762, opened 2025-03-27
isShapeHidden => getShapeVisibility, to allow children of hidden shapes to be visible

ds300:
This PR was motivated by a very reasonable [discord request
](https://discord.com/channels/859816885297741824/1353628236487327857/1353628236487327857)

> We are working on a feature where we need to hide all of the shapes except for the "focused shape" (we are using isShapeHidden prop) - this is largely working as expected, except for one challenge - when the "focused shape" is child of a frame/ group shape, then hiding the parent shape also hides the "focused shape", which makes sense in general context, but we need the ability to hide the parent shape without hiding the "focused shape".

Ended up being a fairly small diff.

Kapture 2025-03-27 at 12 00 09

I'm not crazy about this 'force_show' API and would appreciate alternative suggestions if you have any.

### Change type

- [x] `other`

### Release notes

- Allow the children of a hidden shape to show themselves by returning a 'force_show' override from the `isShapeHidden` predicate.

### DIFF
--- a/packages/editor/src/lib/TldrawEditor.tsx
+++ b/packages/editor/src/lib/TldrawEditor.tsx
@@ -207,6 +207,10 @@ export interface TldrawEditorBaseProps {
 	 * getShapeVisibility={(shape, editor) => shape.meta.hidden ? 'hidden' : 'inherit'}
 	 * ```
 	 *
+	 * - `'inherit' | undefined` - (default) The shape will be visible unless its parent is hidden.
+	 * - `'hidden'` - The shape will be hidden.
+	 * - `'visible'` - The shape will be visible.
+	 *
 	 * @param shape - The shape to check.
 	 * @param editor - The editor instance.
 	 */
--- a/packages/editor/src/lib/TldrawEditor.tsx
+++ b/packages/editor/src/lib/TldrawEditor.tsx
@@ -213,7 +213,7 @@ export interface TldrawEditorBaseProps {
 	getShapeVisibility?(
 		shape: TLShape,
 		editor: Editor
-	): 'visible' | 'hidden' | 'inherit' | null | undefined | void
+	): 'visible' | 'hidden' | 'inherit' | null | undefined
 
 	/**
 	 * The URLs for the fonts to use in the editor.
--- a/packages/editor/src/lib/TldrawEditor.tsx
+++ b/packages/editor/src/lib/TldrawEditor.tsx
@@ -191,13 +191,29 @@ export interface TldrawEditorBaseProps {
 	/**
 	 * Predicate for whether or not a shape should be hidden.
 	 *
+	 * @deprecated Use {@link TldrawEditorBaseProps#getShapeVisibility} instead.
+	 */
+	isShapeHidden?(shape: TLShape, editor: Editor): boolean
+
+	/**
+	 * Provides a way to hide shapes.
+	 *
 	 * Hidden shapes will not render in the editor, and they will not be eligible for hit test via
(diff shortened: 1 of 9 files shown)


steveruizok:
This seems really hacky, I think our default position with regard to these kind of "half APIs" should be to say no. My suggestion would be for the user here to patch the library following your suggestions here because I don't think an API that ties into meta would be the right solution. If we did have an API, I dunno, something like editor.getForcedVisibleShapeIds would be most similar to our other APIs?

As a potential alternative, @mimecuvalo was working on a component override for Shape that might actually make sense her (ie that would let some external logic override the default container around shape, and which here could override the logic for visibility).

ds300:
>  I don't think an API that ties into meta would be the right solution

that's just in the layer panel example. the SDK api does not involve meta, it works by augmenting the return value of the `isShapeHidden` prop to be either `true`, `false`, or the string `'force_show'` which is like `false !important`

ds300:
> I think my suggestion here would be to keep the vague structure of this, but to deprecate isShapeHidden and replace it with an API that fits with better with the capabilities of the new one. I think my suggestion would be isShapeVisible(shape: TLShape, parentVisibility: boolean): boolean. The default implementation would be return parentVisibility. parentVisibility would also be true for shapes belonging to the current page.

> We could also do like getShapeVisibility and have it return 'visible' (for force_show), 'hidden' (for 'true') or undefined | null for inherit from parent

i like both of these a lot more, will think it over and try one of them out

steveruizok:
> > I don't think an API that ties into meta would be the right solution
> 
> that's just in the layer panel example. the SDK api does not involve meta, it works by augmenting the return value of the `isShapeHidden` prop to be either `true`, `false`, or the string `'force_show'` which is like `false !important`

My mistake! Makes sense

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -191,11 +191,29 @@ export interface TldrawEditorBaseProps {
 	/**
 	 * Predicate for whether or not a shape should be hidden.
 	 *
+	 * @deprecated Use {@link TldrawEditorBaseProps#getShapeVisibility} instead.
+	 */
+	isShapeHidden?(shape: TLShape, editor: Editor): boolean
+
+	/**
+	 * Provides a way to hide shapes.
```
SomeHats, line 199: nit - fairly self explanatory, but it could be worth listing and explaining each possible return value here

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -191,11 +191,29 @@ export interface TldrawEditorBaseProps {
...
+	 * ```
+	 *
+	 * @param shape - The shape to check.
+	 * @param editor - The editor instance.
 	 */
-	isShapeHidden?(shape: TLShape, editor: Editor): boolean
+	getShapeVisibility?(
+		shape: TLShape,
+		editor: Editor
+	): 'visible' | 'hidden' | 'inherit' | null | undefined | void
```
SomeHats, line 216: Do we want `void` here? My understanding of it is that its meaning is closer to "we dont care about the return value of this function" than "nothing": https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgOrDACwGoHtgAmyA3gFACQCARgBQCUAXMgG74GkC+ppoksiKAHK48hEhWr0mAVxAEIMUBHZdSMWQjDBcIZJLrjyUCGGlRdAcjjIAzmCigA5hc7cEOu8gDuGHGyboWKJEALwkelTIXO4gniAi-sjCwchhxBFRQA

SomeHats (approved):
nice!

PR #5773, opened 2025-03-28
a11y: announce shapes as they're visited

mimecuvalo:
Building off of https://github.com/tldraw/tldraw/pull/5634 and https://github.com/tldraw/tldraw/pull/5761 this is adding a11y live text to be read aloud when visiting a shape.

We add an overridable method for shapes to customize this called `getAriaLiveText`.
Furthermore, we lay the groundwork here to start letting media shapes have `altText`.
Drive-by fix of `heart` being missing in `geo-styles` list.
Also, drive-by fix of us calling our Image button "Asset" (what are we selling financial instruments here? :P) "Media" is a better word for this button, more human.

Some of the i18n translation is funky. It's a shortcoming of our current system that we don't support interpolation :-/
It sucks, and we'll revisit in the future.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: announce shapes as they're visited


packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -448,6 +448,14 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 		return undefined
 	}
 
+	getAriaLiveText(_shape: Shape): string | undefined {
+		return undefined
+	}
+
+	getAriaAltText(_shape: Shape): string | undefined {
+		return undefined
+	}
```
SomeHats, line 457: could we add doc comments explaining what the difference between these two is?

SomeHats (commented):
So my main reservation here is which parts of the system are responsible for what.

I feel like a11y should be a core part of the canvas. The way this is right now, if you're using tldraw without our UI, some parts of the system work as expected (you can navigate with keys) but others don't (your navigations won't be announced).

I think in an ideal world, we'd move the announcer into `@tldraw/editor` - and probably sink most of `generateShapeAnnouncementMessage` into shape utils too.

The big challenge there is that the raw editor doesn't have any notion of a UI string for localization. I have a few ideas about how we could add that, but obviously that's a pretty significant change. Maybe one to discuss offline. What do you think?

packages/tldraw/src/lib/ui/components/A11y.tsx
```diff
@@ -41,8 +52,123 @@ export function SkipToMainContent() {
...
+		const shapeIndex = msg('a11y.shape-index')
+			.replace('{num}', currentShapeIndex)
+			.replace('{total}', totalShapes)
+
+		// Get describing text (alt text or shape text)
+		const describingText =
+			shapeUtil.getAriaDescriptor(shape) || shapeUtil.getText(shape) || msg('a11y.shape-empty')
+
+		// Build the full announcement
+		a11yLive = `${describingText}, ${shapeType}. ${shapeIndex}`
```
SomeHats, line 125: instead of "Empty, star. 2 of 4" can/should we just do "Star, 2 of 4"? I think we should put the most useful/actionable info first

packages/tldraw/src/lib/ui/components/A11y.tsx
```diff
@@ -41,8 +52,123 @@ export function SkipToMainContent() {
...
+
+		scheduler.attach()
+		scheduler.execute()
+
+		announceSelectedShapes(editor.getSelectedShapeIds())
+		return () => {
+			scheduler.detach()
+			announceSelectedShapesDebounced.cancel()
+		}
+	}, [editor, a11y, msg])
```
SomeHats, line 167:
```suggestion
		const stopListening = react('useSelectedShapesAnnouncer', () => {
			const selectedShapes = editor.getSelectedShapeIds()
			announceSelectedShapesDebounced(selectedShapes)
		})

		return () => {
			stopListening()
			announceSelectedShapesDebounced.cancel()
		}
	}, [editor, a11y, msg])
```

SomeHats (commented):
looking good! main question mark for me is the thottling - that one feels like a miss to me. otherwise this lgtm!

packages/tldraw/src/lib/ui/components/A11y.tsx
```diff
@@ -41,8 +52,123 @@ export function SkipToMainContent() {
...
+				selectedShapeIds,
+				msg,
+			})
+
+			if (a11yLive) {
+				a11y.announce({ msg: a11yLive })
+			}
+		}
+
+		const announceSelectedShapesDebounced = throttle(announceSelectedShapes, 1000)
```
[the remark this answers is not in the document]
mimecuvalo: Oh oops, this is leftover from when it was "polite" instead of "assertive" - removed!

PR #5774, opened 2025-03-29
Fix typo in Local Images example

mootari:
Fixes a typo in the footer text of the Local Images example.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

n/a

### Release notes

n/a

### DIFF
--- a/apps/examples/src/examples/local-images/LocalImagesExample.tsx
+++ b/apps/examples/src/examples/local-images/LocalImagesExample.tsx
@@ -73,7 +73,7 @@ Each image has an associated asset record, so we'll create that first. We need a
 `assetId` so that we can later associate it with the image.
 
 [3]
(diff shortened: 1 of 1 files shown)


mootari:
It seems that the run-on sentence immediately after could also use some light restructuring: https://github.com/tldraw/tldraw/blob/5fcdc0f29d81f74d55b7e5d08d4bdb6cc968c58d/apps/examples/src/examples/local-images/LocalImagesExample.tsx#L77-L80

mimecuvalo (approved):
👍 ty!
Agreed about the run-on sentence too if you wanna take a stab at that.

steveruizok:
Good catch!

PR #5783, opened 2025-04-01
a11y: more fine-grained panning when in Hand tool

mimecuvalo:
We have spacebar panning embedded in the Editor logic at the moment which moves the camera by viewport w/h. 
This is in addition to that logic to give more granular control to moving around.
I'm not _entirely_ convinced it's necessary but it does feel like some more fine-grained tooling would be needed here.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- a11y: provide more fine-grained camera controls via keyboard while in the Hand tool


mimecuvalo:
all good thoughts, thanks @SomeHats .

I'm actually gonna put this PR into draft mode, I'm not feeling super strongly/confident about it.

Issue #5527, opened 2025-03-02
can't load an svg -- At shape(type = image).props.w: Expected a non-zero positive number, got 0


haydenflinner:
Hey, I ran into an error while using tldraw:

```js
ValidationError@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:10315:5
prefixError@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:10326:13
model/<@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:10851:14
validate@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:10361:28
validate@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:8260:27
validateRecord@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:10068:25
put/<@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:9338:32
atomic/<@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:9779:24
transact@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:7046:12
atomic@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:9769:12
put@https://vitejsviteahoswhus-fcdx--5173--31ca1d38.local-corp.webcontainer.io/node_modules/.vite/deps/tldraw.js?v=53385dff:9309:10
[...]

PR #5789, opened 2025-04-01
images: fix not being able to insert SVGs into Firefox

mimecuvalo:
fixes https://github.com/tldraw/tldraw/issues/5527

Not a pretty fix, sigh. @SomeHats lemme know if you have thoughts.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fix not being able to insert SVGs into Firefox

### DIFF
--- a/packages/utils/src/lib/media/media.ts
+++ b/packages/utils/src/lib/media/media.ts
@@ -151,11 +151,11 @@ export class MediaHelpers {
 			}
 			img.crossOrigin = 'anonymous'
 			img.referrerPolicy = 'strict-origin-when-cross-origin'
-			img.src = src
 			img.style.visibility = 'hidden'
 			img.style.position = 'absolute'
 			img.style.opacity = '0'
 			img.style.zIndex = '-9999'
+			img.src = src
 		})
 	}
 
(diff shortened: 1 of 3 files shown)


packages/utils/src/lib/media/media.ts
```diff
@@ -119,17 +119,33 @@ export class MediaHelpers {
...
+			img.onload = () => {
+				document.body.appendChild(img)
+				// Sigh, Firefox doesn't have naturalWidth or naturalHeight for SVGs. :-/
+				// We have to attach to dom and use clientWidth/clientHeight.
+				const dimensions = {
+					w: img.naturalWidth || img.clientWidth,
+					h: img.naturalHeight || img.clientHeight,
+				}
+				document.body.removeChild(img)
+				resolve({ ...dimensions, image: img })
```
SomeHats, line 136: is it worth checking for `naturalWidth` before adding to the dom?

SomeHats (approved):
fun 🙃

packages/utils/src/lib/media/media.ts
mimecuvalo: good call 👍

PR #5791, opened 2025-04-01
fix github workflow sometimes using sh instead of bash

mimecuvalo:
I realized, after digging deeper, that the issues with this github action was because it was `sh` instead of `bash`.
This PR makes sure we're always using `bash`.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -4,6 +4,10 @@ on:
   schedule:
     - cron: '0 0 1 * *'
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   close-issues:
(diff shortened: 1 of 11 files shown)


MitjaBezensek:
Looks like e2e have a problem with this change?

What were the issues you were seeing?

mimecuvalo:
@MitjaBezensek looks like the e2e failure was some temporary flake thing.

sorry, i forgot to paste a link of what i was talking about, i was talking about the failures for this action: https://github.com/tldraw/tldraw/actions/workflows/i18n-download-strings.yml

MitjaBezensek (approved):
💅

PR #5805, opened 2025-04-02
Prune sst stages based on the active amazon clusters.

MitjaBezensek:
Gets active clusters from aws (sst does not have a list command), then we prune sst stages for prs that have been closed for a while.

Ran this locally and got the arns and then the stage names correctly. Let's see if the auth is set up correctly when this next runs.

```
💡[424]: test.ts:10: arn= arn:aws:ecs:eu-north-1:571248596162:cluster/tldraw-pr-5551-clusterCluster-trnfsoeh
💡[424]: test.ts:10: arn= arn:aws:ecs:eu-north-1:571248596162:cluster/tldraw-davidsheldrick-clusterCluster-xtawvuow
💡[424]: test.ts:10: arn= arn:aws:ecs:eu-north-1:571248596162:cluster/tldraw-pr-5804-clusterCluster-fhmtdafa
💡[423]: test.ts:16: names= [ 'pr-5551', 'pr-5804' ]
```

Resolves INT-1245
(together with changes in #5795)

### Change type

- [x] `improvement`

### DIFF
--- a/.github/workflows/prune-preview-deploys.yml
+++ b/.github/workflows/prune-preview-deploys.yml
@@ -1,7 +1,6 @@
 name: Prune preview deploys
 
 on:
-  pull_request:
   schedule:
     # run once per day at midnight or whatever
     - cron: '0 0 * * *'
--- a/.github/workflows/prune-preview-deploys.yml
+++ b/.github/workflows/prune-preview-deploys.yml
@@ -31,6 +31,12 @@ jobs:
       - name: Setup Fly IO
         uses: superfly/flyctl-actions/setup-flyctl@master
 
+      - name: Configure AWS credentials using OIDC
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::571248596162:role/huppy-bot
+          aws-region: 'eu-north-1'
+
       - name: Prune preview deploys
         run: yarn tsx internal/scripts/prune-preview-deploys.ts
         env:
--- a/.github/workflows/prune-preview-deploys.yml
+++ b/.github/workflows/prune-preview-deploys.yml
@@ -6,6 +6,9 @@ on:
     # run once per day at midnight or whatever
     - cron: '0 0 * * *'
 
+permissions:
+  id-token: 'write'
+
 env:
   CI: 1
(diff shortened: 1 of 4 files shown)


.github/workflows/prune-preview-deploys.yml
```diff
@@ -31,6 +31,12 @@ jobs:
       - name: Setup Fly IO
         uses: superfly/flyctl-actions/setup-flyctl@master
 
+      - name: Configure AWS credentials using OIDC
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::571248596162:role/huppy-bot
```
MitjaBezensek, line 37: We only prune preview, so this is hard coded (for deploys we have to switch between different profiles).

.github/workflows/prune-preview-deploys.yml
```diff
@@ -31,6 +31,12 @@ jobs:
       - name: Setup Fly IO
         uses: superfly/flyctl-actions/setup-flyctl@master
 
+      - name: Configure AWS credentials using OIDC
+        uses: aws-actions/configure-aws-credentials@v4
```
ds300, line 35: you could test the permissions for this by briefly adding `  pull_request:` to the `on:` section at the top of this file

internal/scripts/prune-preview-deploys.ts
```diff
@@ -163,13 +164,35 @@ async function listFlyioPreviewApps() {
 	return appNames.filter((name) => ZERO_CACHE_APP_REGEX.test(name))
 }
 
+async function listAmazonClusters() {
+	const client = new ECSClient({ region: 'eu-north-1', profile: 'preview' })
```
ds300, line 168: i wonder if we might get into trouble with `profile: 'preview'` when running on CI like we did with the sst config, where we had to use `process.env.CI` to make it undefined to prevent aws complaining about not being able to find the config for the profile?

ds300 (approved):

internal/scripts/prune-preview-deploys.ts
MitjaBezensek: Yeah, you were right. Remove the profile now.

PR #5812, opened 2025-04-03
Include assetUrls in ExplodedExample

trygve-aaberge-adsk:
Not including assetUrls meant that the exploded example didn't load fonts correctly. It would just try to load the fonts e.g. from /tldraw_draw.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape with text in the exploded example
2. Reload the page to make sure the fonts aren't loaded from previous requests
3. See that the font isn't loaded and the fallback font is used

- [ ] Unit tests
- [ ] End to end tests

### DIFF
--- a/apps/examples/src/examples/exploded/ExplodedExample.tsx
+++ b/apps/examples/src/examples/exploded/ExplodedExample.tsx
@@ -13,6 +13,7 @@ import {
 	TldrawUi,
 	defaultAddFontsFromNode,
 	defaultBindingUtils,
+	defaultEditorAssetUrls,
 	defaultShapeTools,
 	defaultShapeUtils,
 	defaultTools,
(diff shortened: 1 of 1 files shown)


trygve-aaberge-adsk:
It wasn't immediately clear that this is what I had to do to load fonts correctly after `usePreloadAssets` was removed, since it wasn't mentioned in the release notes.

mimecuvalo:
thanks for the PR!

> It wasn't immediately clear that this is what I had to do to load fonts correctly after usePreloadAssets was removed, since it wasn't mentioned in the release notes.

sorry about that, you are right, we should have mentioned this.

mimecuvalo:
ah, looks like it's failing b/c of the imports, mind fixing that up? and then it should land.

trygve-aaberge-adsk:
Ah, I didn't notice that `useDefaultEditorAssetsWithOverrides` was internal (and lint didn't run/show in my editor for some reason, and I didn't notice the build error here because there's always some failures because of permissions).

I changed it to use `defaultEditorAssetUrls` directly now.

PR #5821, opened 2025-04-04
Indicators perf

steveruizok:
This PR improves the rendering for indicators. Now that we're leaving them in the DOM (and hiding them) to avoid expensive mounting / unmounting, I've seen the rendering cost of indicators effecting performance. This really only happens when dragging large numbers of shapes around, though on slower computers it might show up with smaller numbers of shapes.

In this PR, we do two things:
- we now don't update the position of indicators if they're hidden
- we only render the indicator if its associated shape's props have changed

And a third, secret thing:

- for geo shapes, we render the "solid" style indicator when zoomed out

### Change type

- [x] `improvement`

### Test plan

1. Make many shapes
2. Drag them about the canvas

### Release notes

- Improved performance when dragging many shapes at once


packages/editor/src/lib/components/default-components/DefaultShapeIndicator.tsx
```diff
@@ -9,13 +9,18 @@ import { useEditorComponents } from '../../hooks/useEditorComponents'
...
+const EvenInnererIndicator = memo(
+	({ shape, util }: { shape: TLShape; util: ShapeUtil<any> }) => {
+		return useStateTracking('Indicator: ' + shape.type, () =>
+			// always fetch the latest shape from the store even if the props/meta have not changed, to avoid
+			// calling the render method with stale data.
+			util.indicator(util.editor.store.unsafeGetWithoutCapture(shape.id) as TLShape)
+		)
+	},
+	(prevProps, nextProps) => {
+		return prevProps.shape.props === nextProps.shape.props
```
ds300, line 21: we should also use meta to bust the cache, we had problems with that on the Shape component
```suggestion
		return prevProps.shape.props === nextProps.shape.props && prevProps.shape.meta === nextProps.shape.meta
```

ds300 (approved):

packages/editor/src/lib/components/default-components/DefaultShapeIndicator.tsx
steveruizok: Good catch, thanks

PR #5826, opened 2025-04-04
a11y: add resize kbd shortcut

mimecuvalo:
Cmd+opt+shift +/- to resize a shape.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [ ] Unit tests
- [x] End to end tests

### Release notes

- a11y: add resize kbd shortcut


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -102,6 +102,49 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 			return editor.isIn('select') && editor.getSelectedShapeIds().length > 0
 		}
 
+		function scaleShapes(scaleFactor: number) {
```
SomeHats, line 105: Instead of trying to implement resizing again here and missing out all our shape utils lifecycle callbacks etc, lets have this use `resizeShape` under the hood.

It should also scale shapes relative to the center of the current selection, rather from an arbitrary point on each shape.

SomeHats (changes requested):
what's the motivation here?

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: hah, sometimes I don't even know our own API 🙃 i looked for resize and couldn't find it for some reason.

i don't know about using the center point actually. but let's discuss more offline along w the motivation.

mimecuvalo: updated center point - thx @SomeHats

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -107,6 +108,35 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				const shapes = selectedShapeIds
+					.map((id) => editor.getShape(id))
+					.filter(Boolean) as TLShape[]
+
+				// Update each shape
+				shapes.forEach((shape) => {
+					if (!('w' in shape.props) || !('h' in shape.props)) return
+
+					// Don't make it too small.
+					if (scaleFactor < 1.0 && (shape.props.w < 32 || shape.props.h < 32)) return
```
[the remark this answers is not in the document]
mimecuvalo: Ah, this is a good point with the multiple selection, ok nixing this (and yeah we can look into adding it into resizeShape some other time)

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1483,6 +1513,22 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				onSelect: async (source) => {
+					scaleShapes(1.1)
+					trackEvent('embiggen-shapes', { source })
+				},
+			},
+			{
+				id: 'emsmallen-shapes',
+				kbd: '$!?-',
+				onSelect: async (source) => {
+					scaleShapes(0.9)
```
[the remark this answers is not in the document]
mimecuvalo: very good point! nice one

SomeHats (approved):
amazing! thanks @mimecuvalo

PR #5827, opened 2025-04-04
a11y: add way to focus styling menu via kbd

mimecuvalo:
Cmd+Enter to focus the styling menu. This was originally going to open up the context menu but that seemed secondary for now to the styling menu.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: add way to focus the styling menu via kbd


SomeHats (changes requested):
When I was testing this, I noticed a couple of weird behaviours:
1. Sometimes, when opening the context menu, the selected shape would go back to the first in the document. I couldn't repro this on video though, not sure what it was about
2. The mobile rotate handle shows up when I do this. I can interact with it once, but then it goes away. Spooky.
3. There's a weird delay between hitting the shortcut and the menu opening. The mobile handle makes this extra noticeable because it appears instantly
4. I can't tab through the context menu. Maybe that's not the convention here, but it's the first thing I tried
5. When it opens, should the first item be focused with the ring around it? 

Kapture 2025-04-07 at 13 55 47

Any way we can resolve these?

mimecuvalo:
> Sometimes, when opening the context menu, the selected shape would go back to the first in the document. I couldn't repro this on video though, not sure what it was about

This might have been resolved in https://github.com/tldraw/tldraw/pull/5837 which had issues with tab capture sometimes. But hard to say.

mimecuvalo:
> I can't tab through the context menu. Maybe that's not the convention here, but it's the first thing I tried

For menus, you have to use up/down fwiw.

packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx
```diff
@@ -20,6 +20,24 @@ export const DefaultContextMenu = memo(function DefaultContextMenu({
...
+	}, [])
+
+	useEffect(() => {
+		// Make sure to remove the event listener when the component unmounts.
+		return () => {
+			document.body.removeEventListener('keydown', preventEscapeFromLosingShapeFocus, {
+				capture: true,
+			})
+		}
+	}, [preventEscapeFromLosingShapeFocus])
```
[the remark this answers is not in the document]
mimecuvalo: good call 👍

mimecuvalo:
Ok, fixed up the mobile rotate handle, silly issue there.

I tried resolving the delay of opening but couldn't figure that out, it's a head scratcher. As far as I can tell, the event is dispatched and then useCanvasEvent dispatches it internally. It's hard to say why it takes an extra tick.
The other topics we discussed offline, thanks for the review!

SomeHats:
OK I managed to capture the selection moving issue: 

https://github.com/user-attachments/assets/11bf0764-36b3-4a1b-b0fd-3cb88691f7af

I also noticed when this was happening that tab was moving me around the examples list, but cmd-enter was still opening the context menu of the current selected shape

mimecuvalo:
> OK I managed to capture the selection moving issue:
> 
>  Kapture.2025-04-10.at.11.53.07.mp4 
> I also noticed when this was happening that tab was moving me around the examples list, but cmd-enter was still opening the context menu of the current selected shape

Huh, I think I see, I might have to see this in person. (it's harder to tell without the keyboard plugin in Kap to show when keys are being pressed to follow along, heh)

mimecuvalo:
Ok, this has been reworked. PTAL @SomeHats 
I think one missing thing, for a separate PR, is having Escape return to the previously focused area, instead of losing context.

steveruizok:
Some notes:

High level:

I thought we were focusing for now on reading / navigating the canvas before exploring accessibility-relate features for editing / modifying canvas content, which seems to be a separate and much larger question. I still think this is a good idea. 

If we want to roll forward with these sorts of PRs, I'd flag them all as possibly-changing if we come up with other solutions. For example, we may do kbd sequences for different panels, i.e. p-c-1 for style-panel, colors, and black; and free uo Cmd+Enter for some other interaction.

Low level:

- When I press Cmd+Enter, the style panel does not display a focus indicator until I use the Tab or arrow keys to move between items in the menu
- Almost seems like there should be a display change in the style panel itself, i.e. some focus ring or other indication that focus has moved to that panel. Same really for the toolbar and other places where we're tabbing from panel to panel.

mimecuvalo:
> I thought we were focusing for now on reading / navigating the canvas before exploring accessibility-relate features for editing / modifying canvas content, which seems to be a separate and much larger question. I still think this is a good idea.

The core issue with navigating is that you tabbing away between the canvas and the surrounding chrome means that the shapes aren't selected. So, while you _can_ tab to the style panel, it's useless unless there's a shape selected. This provides  the crucial tie between the UI and canvas.

> If we want to roll forward with these sorts of PRs, I'd flag them all as possibly-changing if we come up with other solutions. 

I think we discussed this offline yesterday but I can't tell if your comment was after/before that conversation. Miro, for example has the same paradigm for Cmd+Enter of focusing the contextual toolbar (in our case the Style Panel).

> For example, we may do kbd sequences for different panels, i.e. p-c-1 for style-panel, colors, and black; and free uo Cmd+Enter for some other interaction.

I would strongly, strongly argue that this too many shortcuts for a person to remember. It's not about having Vim/Emacs-like mappings to everything (that's hard even for coders sometimes! let alone regular folk). I don't think we should be protective about Cmd+Enter being useful for some other interaction.

> When I press Cmd+Enter, the style panel does not display a focus indicator until I use the Tab or arrow keys to move between items in the menu

I mentioned this on another issue: what you're probably running into is you haven't crossed the browser heuristic for being in "tabbing mode" for the focus-visible to start showing up. You need to not click a shape and then hit Cmd+Enter, that won't work initially (w/o doing tab/arrow as you say). You'll need to be tabbing around shapes, then hit Cmd+Enter, as a person using accessibility features in the real world would. If we felt strongly that we would want this to not be just for a11y users, we would use `:focus` instead of `:focus-visible` - we can clarify this with the consultant, but I would say we should still stick with `:focus-visible` imo.

MitjaBezensek:
Not sure if it's connected to this PR, but it looks like tabbing to the size part of the style panel does not land on the correct element (see how I land on L, where the selected option is M).

https://github.com/user-attachments/assets/4937afaf-3807-412d-b5b6-862990048324

Similarly here with a few other properties:

https://github.com/user-attachments/assets/20d7afa8-84dc-4fae-b81d-c2b8e0e8bbe9

We'd probably want to show the focus ring when we open the shape menu?

https://github.com/user-attachments/assets/c0030efe-a2ca-426d-b34e-e55af363d347

Another thing that might not be connected is that I'd want the up and down arrows to also work while moving between elements in the style panel (like moving between colors since we have multiple rows).

> The core issue with navigating is that you tabbing away between the canvas and the surrounding chrome means that the shapes aren't selected. So, while you can tab to the style panel, it's useless unless there's a shape selected. This provides the crucial tie between the UI and canvas.

It does look like you can still tab to the style panel with no shapes selected? Also, I guess we do want to allow that, so that you can change the styles for next shape?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1510,6 +1510,21 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					trackEvent('change-page', { source, direction: 'next' })
 				},
 			},
+			{
+				id: 'adjust-shape-styles',
+				label: 'a11y.adjust-shape-styles',
+				kbd: 'cmd+Enter,ctrl+Enter',
+				readonlyOk: true,
```
MitjaBezensek, line 1517: Don't think we even show the style menu in readonly?

mimecuvalo:
> Not really connected to this PR, but it looks like tabbing to the size part of the style panel does not land on the correct element (see how I land on L, where the selected option is M).

yeah, this is ok. the browser / radix "remembers" where you last had focus.

mimecuvalo:
> Another thing that might not be connected is that I'd want the up and down arrows to also work while moving between elements in the style panel (like moving between colors since we have multiple rows).

maaaybe, i think for that, i want to talk w/ our consultant on what best practices is for that. it's unclear to me (and Alex)

> It does look like you can still tab to the style panel with no shapes selected? Also, I guess we do want to allow that, so that you can change the styles for next shape?

yeah, we still want to allow this case 👍

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: good point! fixed

MitjaBezensek (approved):

PR #5829, opened 2025-04-06
Update quick-start.mdx

tulavamsidheeraj:
The props.text is rising an error of unexpected property. This change will fix it.


tulavamsidheeraj:

SomeHats:
Thanks @tulavamsidheeraj! I wanted to make sure we got the `import` correct too so I moved your change over to #5836. Thank you for flagging this!

PR #5834, opened 2025-04-07
Fix flyio app prunning and deployment.

MitjaBezensek:
Looks like we need to pass the `-y` flag to [accept the confirmation](https://fly.io/docs/flyctl/apps-destroy/#options) when pruning the apps. [Failure](https://github.com/tldraw/tldraw/actions/runs/14297908654/job/40067462974#step:6:346) of the last run.

Preview deploys were also failing to start correctly because of permissions deployment issue. We tried to deploy permissions before deploying zero, which failed since the zero hasn't had the chance to create the tables yet.

### Change type

- [x] `bugfix`

### DIFF
--- a/internal/scripts/deploy-dotcom.ts
+++ b/internal/scripts/deploy-dotcom.ts
@@ -253,22 +253,6 @@ async function deployAssetUploadWorker({ dryRun }: { dryRun: boolean }) {
 	})
 }
 
-async function deployPermissionsToFlyIo() {
-	const schemaPath = path.join(REPO_ROOT, 'packages', 'dotcom-shared', 'src', 'tlaSchema.ts')
-	const permissionsFile = 'permissions.sql'
-	await exec('npx', [
-		'zero-deploy-permissions',
-		'--schema-path',
-		schemaPath,
-		'--output-file',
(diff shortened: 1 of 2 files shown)


internal/scripts/deploy-dotcom.ts
```diff
@@ -415,6 +399,22 @@ function updateFlyioToml(appName: string): void {
...
+	const schemaPath = path.join(REPO_ROOT, 'packages', 'dotcom-shared', 'src', 'tlaSchema.ts')
+	const permissionsFile = 'permissions.sql'
+	await exec('npx', [
+		'zero-deploy-permissions',
+		'--schema-path',
+		schemaPath,
+		'--output-file',
+		permissionsFile,
+	])
+	const result = await exec('psql', [env.BOTCOM_POSTGRES_CONNECTION_STRING, '-f', permissionsFile])
```
MitjaBezensek, line 412: The command failed, but it didn't stop the deployment. Added a check for it so that it now fails: https://github.com/tldraw/tldraw/actions/runs/14306723803/job/40092470607?pr=5834#step:12:1017

PR #5837, opened 2025-04-07
a11y: fix focus ring on share menu/embed dialog; also slider/watermark

mimecuvalo:
- fix dotcom share menu (we were using dropdown, and we should have used popover
- fix embed menu on dotcom
- fix slider and watermark capturing tab focus when shapes were selected

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
going to land this now and @MitjaBezensek's gonna do an async review

PR #5844, opened 2025-04-08
Fix e2e tests.

MitjaBezensek:
This implements `run` api that zero has added.

Also had to make some of our apis async. Example issue that this fixes is duplicating files. There we first called `createFile` and then navigated to it. But since `insertWithFileState` is `async` we got the file state update from the navigation before the file was actually created, which caused some issues.

### Change type

- [x] `bugfix`


apps/dotcom/client/src/tla/app/zero-polyfill.ts
```diff
@@ -182,9 +180,21 @@ export class Zero {
 		}
 		return {
 			...stuff,
-			where: (column: string, _ownerId: string) => {
+			where: (column: string, op: string, value: any) => {
+				assert(op === '=', 'Only = operator is supported')
```
MitjaBezensek, line 184: Made this explicit, just so we don't start using some operators we currently don't support.

PR #5872, opened 2025-04-10
a11y: make toolbars more compliant with keyboard navigation

mimecuvalo:
The toolbars as they stand right now aren't set up right yet. To be compliant, they need to support left/right and setting tabIndex on items as you go.
This requires quite a rework of the toolbars and their buttons.

- [x] add aria-labels to toolbars
- [x] zoom menu - dropdown
- [x] zoom menu - keep focus when toggling
- [x] main toolbar
- [x] main overflow toolbar
- [x] top-left toolbar
- [x] style panel stuff
- [x] toolbar extra menu
- [x] loading screen
- [x] contextual toolbar
- [x] drive-by fix number shortcuts not working when toolbar selected
- [x] going to need to override branding instead of tldraw
- [x] add more aria-labels to a11y tree
- [x] padding at top of style panel wtf
- [x] slider left/right/top/bottom
- [x] sub-menus
- [x] check dotcom
- [ ] escape should return back to the original spot

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: make toolbars more compliant with keyboard navigation


packages/editor/src/lib/components/default-components/DefaultSpinner.tsx
```diff
@@ -1,7 +1,7 @@
 /** @public @react */
 export function DefaultSpinner() {
 	return (
-		<svg width={16} height={16} viewBox="0 0 16 16">
+		<svg width={16} height={16} viewBox="0 0 16 16" role="alert" aria-busy="true">
```
SomeHats, line 4: is `role="alert"` appropriate for a loading spinner? from mdn:
> Warning: Because of its intrusive nature, the alert role must be used sparingly and only in situations where the user's immediate attention is required.

packages/tldraw/src/lib/ui/components/ActionsMenu/DefaultActionsMenu.tsx
```diff
@@ -47,27 +48,30 @@ export const DefaultActionsMenu = memo(function DefaultActionsMenu({
...
 		<TldrawUiPopover id="actions-menu">
 			<TldrawUiPopoverTrigger>
-				<TldrawUiButton
-					type="icon"
-					data-testid="actions-menu.button"
-					title={msg('actions-menu.title')}
-				>
-					<TldrawUiButtonIcon icon="dots-vertical" small />
-				</TldrawUiButton>
+				<PopoverPrimitive.Anchor>
```
SomeHats, line 51: I thought `Anchor` was only needed when the anchor was something different to the trigger? Does `TldrawUiPopoverTrigger` not imply this?

packages/tldraw/src/lib/ui/components/Spinner.tsx
```diff
@@ -1,9 +1,20 @@
...
+
 	return (
-		<svg width={16} height={16} viewBox="0 0 16 16" {...props}>
+		<svg
+			width={16}
+			height={16}
+			viewBox="0 0 16 16"
+			{...props}
+			aria-label={msg('app.loading')}
+			role="alert"
```
SomeHats, line 15: ditto re: `alert` for this

packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
```diff
@@ -123,64 +123,72 @@ export function CommonStylePickerSet({ styles, theme }: ThemeStylePickerSetProps
...
-						title={msg('style-panel.color')}
-						uiType="color"
-						style={DefaultColorStyle}
-						items={STYLES.color}
-						value={color}
-						onValueChange={handleValueChange}
-						theme={theme}
-						onHistoryMark={onHistoryMark}
-					/>
+					<TldrawUiToolbar label={msg('style-panel.color')}>
```
SomeHats, line 128: i thought we had discussed making these radio groups instead of toolbars so it's easier to change the values within them and navigate between the different controls?

packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
```diff
@@ -172,19 +174,22 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
...
 							<TldrawUiPopoverTrigger>
-								<TldrawUiButton
-									title={msg('tool-panel.more')}
-									type="tool"
-									className="tlui-toolbar__overflow"
-									data-testid="tools.more-button"
-								>
-									<TldrawUiButtonIcon icon="chevron-up" />
-								</TldrawUiButton>
+								<PopoverPrimitive.Anchor>
```
SomeHats, line 177: ditto - why is this needed?

SomeHats: ohhhhh is it because the new toolbar components don't take a `ref`?

packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
```diff
@@ -172,19 +174,22 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
...
+										className="tlui-toolbar__overflow"
+										data-testid="tools.more-button"
+									>
+										<TldrawUiButtonIcon icon="chevron-up" />
+									</TldrawUiToolbarButton>
+								</PopoverPrimitive.Anchor>
 							</TldrawUiPopoverTrigger>
 							<TldrawUiPopoverContent side="top" align="center">
-								<div
+								<TldrawUiToolbar
```
SomeHats, line 189: i'm really uncertain about this overflow grid being a horizontal toolbar. it just sort of feels like my up/down arrow keys should work here? wdyt?

assets/translations/main.json
```diff
@@ -364,7 +366,9 @@
 	"style-panel.opacity": "Opacity",
 	"style-panel.size": "Size",
 	"style-panel.spline": "Spline",
+	"tool-panel.title": "Tools",
 	"tool-panel.more": "More",
+	"navigation-zone.title": "Navigation toolbar",
```
SomeHats, line 371: Screenshot 2025-04-14 at 11 14 57

I don't think we need to include "toolbar" in these

packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
SomeHats: It's also a bit odd to me that these are now all announced as toolbars.

For colors, it's surprising that the up/down arrow keys move my selection right/left

assets/translations/main.json
```diff
@@ -244,7 +244,9 @@
 	"tool.rich-text-link-remove": "Remove link",
 	"tool.rich-text-header": "Header",
 	"tool.rich-text-bulletList": "Bulleted list",
+	"tool.rich-text-toolbar-title": "Rich text toolbar",
```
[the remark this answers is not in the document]
mimecuvalo: you're right — good call!

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -285,6 +285,7 @@ export const TldrawEditor = memo(function TldrawEditor({
 			onPointerDown={stopEventPropagation}
 			tabIndex={-1}
 			role="application"
+			aria-label={_options?.branding ?? 'tldraw'}
```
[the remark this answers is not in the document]
mimecuvalo: yeah good question, it's to help with the accessibility tree and voiceover. also, esp. since tldraw tends to be a sub-frame of a larger application, it'll help delineate to the user that you're in the "padlet whiteboard" section or "clickup board" section, that kind of thing.

Screenshot 2025-04-14 at 12 37 00
Screenshot 2025-04-14 at 12 36 41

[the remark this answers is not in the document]
SomeHats: does 'move focus to canvas' not already imply this?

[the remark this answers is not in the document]
SomeHats: I would be tempted to call this "board" or something, end-users of SDK customers are unlikely to know what tldraw is

packages/editor/src/lib/components/default-components/DefaultSpinner.tsx
mimecuvalo: good push, i re-read what best practices are on loading/spinner and saw that i was reading some outdated/refuted practices. i updated it to more accepted practices.

packages/tldraw/src/lib/ui/components/ActionsMenu/DefaultActionsMenu.tsx
mimecuvalo: ah sorry, this was leftover from some testing i was doing, removing!

assets/translations/main.json
mimecuvalo: nice catch 👍

packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
mimecuvalo: yeah, couple things here that made me rethink the radio vs. toolbar thing:
- the main reason was that the Radix UI component for radio/radiogroups didn't fit right: https://www.radix-ui.com/primitives/docs/components/radio-group they have items/indicators and it felt awkward with what we were using inputs/indicators/labels (well, we'd leave out the labels). yeah i'm sure we could restyle it all to look like buttons but it just felt awkward.
- furthermore, looking a bit at Miro made me think it was more natural to use buttons than to use the radio/input paradigm.

but let's talk further if you think we should think about this some more.

packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
mimecuvalo: - i hear you on the up/down working. maybe? it makes it a bit complex since we have to account for people changing the column count and widows/orphans in rows. i think that's a nice-to-have instead of a must-have and we could do make this fancier in the future regarding the up/down stuff.

> Also, the way this works with keyboard nav & voiceover feels a bit weird. I use arrow keys to move to the "more" button, then enter to select it. Then, my arrow keys stop working. I have to press tab once to move focus to the first item in the popup. Then tab no longer moves between those items, i'm back to using my arrow keys.

very much agreed! i did try for hours to get this to work but couldn't for the life of me :-/ i tried focusing the first element in various ways but it seemed like i couldn't change `:focus-visible` for some reason! i'm trying more today to make that work b/c it's so annoying. related to that i wanted hitting Escape to go back to the trigger button but that was also proving to be a pain-in-the-ass 😭 

> what if we treat the whole overflowing toolbar as one thing. So when you go right from the last item on the main toolbar, we open the overflow and move your focus to the next element in there and vice-versa?

that's not a half-bad idea. i think we'd have to go custom to make that work though because you'd be contending with the radix context states that it uses internally :-/ i think we could make a more custom pass in the future without radix but ejecting now out of radix for this feels like not the right move at the moment.

packages/editor/src/lib/TldrawEditor.tsx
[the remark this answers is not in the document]
mimecuvalo: sorry, "move focus to canvas" was the accidental screen capture there (that's the "skip to content" button), it would be, just:

Screenshot 2025-04-14 at 13 19 16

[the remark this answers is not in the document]
mimecuvalo: board isn't a bad idea though, i think that might make sense
- board
- or, canvas

the eternal question :P

packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
mimecuvalo: > Also, the way this works with keyboard nav & voiceover feels a bit weird. I use arrow keys to move to the "more" button, then enter to select it. Then, my arrow keys stop working. I have to press tab once to move focus to the first item in the popup. Then tab no longer moves between those items, i'm back to using my arrow keys.

I should also regarding this, that this is how I see it working in other apps too with popovers. I think only Radix menus auto-focus the first item?

packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
mimecuvalo: I messed around with this some more today to see how it felt like, and ach, maybe?
To be honest, it's very easy to swap out for the RadioGroup if we want. It feels like it might be one of those things that our a11y auditor/consultant will say: we should use RadioGroup vs. Toolbar. I would trust their opinion on which way to go more than ours maybe :)

SomeHats: I'd love to have a play with the radio button version if u have a link

packages/editor/src/lib/TldrawEditor.tsx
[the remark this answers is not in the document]
mimecuvalo: it's a moot point here at the moment actually - i don't have translation at this level currently :-/
we need to wait for i18n work to happen

PR #5876, opened 2025-04-11
Fix custom mutators deploy

ds300:
We forgot to add the ZERO_PUSH_URL to our deploy pipeline.

### Change type

- [x] `other`


MitjaBezensek:
Unsure why release label check is failing. Some issue with huppy?

PR #5897, opened 2025-04-15
Reapply coalesced events and fix the local dev issue

MitjaBezensek:
Reapplies the reverted change in https://github.com/tldraw/tldraw/pull/5895 and fixes the local dev issue.

Looks like accessing local dev server via an ip address (instead of localhost) makes the browser not allow access to this function - it doesn't exists at all.

image

### Change type

- [x] `improvement`

### Release notes

- Improve draw fluidity on slower CPUs by using getCoalescedEvents.

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -1,3 +1,4 @@
+import { useValue } from '@tldraw/state-react'
 import React, { useMemo } from 'react'
 import { RIGHT_MOUSE_BUTTON } from '../constants'
 import {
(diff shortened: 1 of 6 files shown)


packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -49,12 +51,20 @@ export function useCanvasEvents() {
...
-				editor.dispatch({
-					type: 'pointer',
-					target: 'canvas',
-					name: 'pointer_move',
-					...getPointerInfo(e),
-				})
+				// For tools that benefit from a higher fidelity of events,
+				// we dispatch the coalesced events.
+				let events: PointerEvent[] | React.PointerEvent<Element>[] = [e]
+				if (currentTool.useCoalescedEvents && e.nativeEvent.getCoalescedEvents) {
```
MitjaBezensek, line 57: Additional check to make sure the browser supports this api - it doesn't seem to support it when accessing a local server via an ip address (possibly in other cases as well).

MitjaBezensek:
Closing in favor of https://github.com/tldraw/tldraw/pull/5898

PR #5902, opened 2025-04-15
Fix bad arrow bindings to text shapes

SomeHats:
This was caused by filtering out labels too aggressively.

### Change type

- [x] `bugfix`

### DIFF
--- a/packages/editor/src/lib/primitives/geometry/Geometry2d.ts
+++ b/packages/editor/src/lib/primitives/geometry/Geometry2d.ts
@@ -11,7 +11,13 @@ import {
 } from '../intersect'
 import { approximately, pointInPolygon } from '../utils'
 
-/** @public */
+/**
+ * Filter geometry within a group.
+ *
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/primitives/geometry/Geometry2d.ts
```diff
@@ -73,7 +80,6 @@ export abstract class Geometry2d {
 	abstract nearestPoint(point: Vec, filters?: Geometry2dFilters): Vec
```
mimecuvalo, line 80: do we need the `isExcludedByFilter` function at all anymore? nix?

mimecuvalo: i see it's used in `Group2d` - should it be specific to `Group2d` / check that it's only used in that context or something?

SomeHats: still part of the Geometry2d contract

PR #5908, opened 2025-04-15
stickies: make sure our custom tab behavior takes over the rich text behavior

mimecuvalo:
our stickies custom tab/shift+tab behavior is getting mixed up with indentation behavior in rich text, oops.

https://github.com/user-attachments/assets/753cc4f3-eaa1-4262-a47d-b2d8c0e1d058

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Added `hasCustomTabBehavior` prop to `RichTextLabel` so developers can also opt out of this behavior where necessary

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -2040,6 +2040,8 @@ export interface RichTextLabelProps {
     // (undocumented)
     fontSize: number;
     // (undocumented)
+    hasCustomTabBehavior?: boolean;
+    // (undocumented)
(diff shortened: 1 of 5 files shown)


MitjaBezensek (approved):
Nice catch. Maybe we can add a note to the release notes that we have added `hasCustomTabBehavior` prop to `RichTextLabel` so developers can also opt out of this behaviour.

PR #5923, opened 2025-04-17
Optimize query.

MitjaBezensek:
Our file permission logic caused zero to use some unoptimized queries that could take quite a while to load. This adds the suggested improvements from https://github.com/grgbkr/tldraw/pull/1

### Change type

- [x] `improvement`


MitjaBezensek:
> nice. i was surprised this worked with the polyfill without any changes!

It's a very good polyfill 😂

PR #5928, opened 2025-04-17
Fix getSelectedShapes caching

ds300:
### Change type

- [x] `bugfix`

### Release notes

- Small perf fix to make sure `editor.getSelectedShapes` doesn't cause signals to fire too often.


mimecuvalo:
sweet! thx!

PR #5938, opened 2025-04-21
fix: focus editor after shortcut selection

nihaocami:
When you select a shortcut like `r` for rectangle, the behaviour is that it basically click the rectangle shape tool/component. That takes focus away from the editor. Therefore I can't immediately draw unless I refocus by hand. Here I am just automatically refocusing the editor for a better UX.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Press a shortcut
2. You should be able to immediately draw the shape without weird double clicking to refocus editor.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve shortcut UX by refocusing the editor


nihaocami:

mimecuvalo:
(sorry for the delay in looking into this)
Hm, our current behavior is as such:
- if your focus is on the canvas, and then you hit `r`, the focus remains on the canvas.
- if your focus is on the toolbar, you hit `r`, the focus remains on the toolbar.

I saw your change is specific to the 1-9 numeric shortcuts, was that intentional? 
Is there a particular browser you're seeing the issue on?

steveruizok:
This is strange, the editor should "start drawing" on pointer down even when it's not focused.

nihaocami:
Hi all, just confirmed this seems to be my touchpad. Specifically I am on windows surface book 4. Touch screen behaviour is normal but the touch pad requires a mouse event before it can start painting. So I guess this may not be a universal issue after all. We can close it if you feel it's a non issue. Thx again!

steveruizok:
> Hi all, just confirmed this seems to be my touchpad. Specifically I am on windows surface book 4. Touch screen behaviour is normal but the touch pad requires a mouse event before it can start painting. So I guess this may not be a universal issue after all. We can close it if you feel it's a non issue. Thx again!

Annoying! If you can figure out why, I'd be happy to address it. Actually it's a good question of where the focus goes when a panel closes—it might go to the body but it should go back to the editor.

nihaocami:
Awesome. OK let me close this for now until I come up with an actual solution. Thx all!

PR #5942, opened 2025-04-22
Fix Safari issue.

MitjaBezensek:
Fixes an issue with artefacts in Safari.

### Before

https://github.com/user-attachments/assets/fae215ba-7116-43c8-8d5b-fc51df446eca

### After

https://github.com/user-attachments/assets/af3c38a1-fc9d-490e-8c72-19ea2b01f694

### Change type

- [x] `bugfix`

### Release notes

- Fix and issue with Safari artefacts when switching between multiple "snapshots" / cards.

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -1,5 +1,11 @@
 /* @tldraw/editor */
 
+.tl-safari-wrapper {
+	transform: translateZ(0);
+	height: 100%;
+	width: 100%;
+}
+
 .tl-container {
 	width: 100%;
 	height: 100%;
@@ -290,7 +296,7 @@ input,
 .tl-shapes {
 	position: relative;
 	z-index: var(--layer-canvas-shapes);
-	will-change: transform;
+	/* will-change: transform; */
 }
 
 .tl-overlays {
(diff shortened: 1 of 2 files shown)


SomeHats:
So we've avoided this historically for a couple reasons. 

1. it creates some rendering nasties at high zoom levels on chrome: 
Screenshot 2025-04-22 at 17 21 44
2. it forces every shape onto its own GPU compositing layer. can impact perf, but the main issue is actually with memory. As each shape ends up as an uncompressed texture on the GPU, you blow through your graphics memory really quickly which will crash devices like iphones :(

MitjaBezensek:
A little update on this:
- `will-change: transform` in this case didn't cause every shape to use it's own layer. It only added one layer for the `tl-shapes` div - the parent that holds the shapes and where I applied this css rule. It's still not a fix since it does cause rendering issues in Safari (and possibly other browser) at low zoom levels.

https://github.com/user-attachments/assets/1c36a717-af5e-40d7-949d-eebb565925ba

- I pushed another change that adds a wrapper that uses a known Safari fix for many of these rendering issues `-webkit-transform: translateZ(0);`.

For the time being we did not decide to include this fix in the library, but we can do this later on if needed. We decided not to go forward with this fix because:
* It changes the DOM structure and might cause some issues with existing users that rely on it.
* We only heard about this problem from one of the users so it might just a really strange combination of css rules and Safari that causes this. Since adding the mentioned fix might have other consequences (possible performance implications) we decided to suggest the mentioned fix to the user. We might revisit this if we get reports from other users.

PR #5949, opened 2025-04-24
templates: Fix simple server node example dropping messages on initial connection

### LINKED ISSUE #5946: Sync custom server example fails when makeOrLoadRoom is slow
### What happened?

Hello,

I've been experiencing an issue with @tldraw/sync and was finally able to reproduce it, even on the official example.

Symptoms:
Initial load of the room never succeeds, room stuck in the loading state, socket connection is never established despite the code reaching `room.handleSocketConnect` call
A refresh fixes it because the room is already made and loaded from cache

After a looot of debugging, I've pinpointed the issue, and it can be reproduced by modifying the line 36. here: https://github.com/tldraw/tldraw/blob/main/templates/simple-server-example/src/server/rooms.ts#L36
and adding the following to the beginning of the mutex promise resolve:
```
mutex = mutex
  then(async () => {
    // It seems like just waiting for a ~100ms is enough to cause the client never to connect to a room
    await new Promise((resolve) => setTimeout(resolve, 100))

    if (rooms.has(roomId)) {
      // ...
```

It seems that a lengthy enough work in this async method which I would expect to be resilient to longer loads is enough to cause the client never to connect, despite sending repeated connect attempts once it stays in the connecting state long enough.

Confirmed reproduced by @MitjaBezensek  

Chrome

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

OriginalEXE:

Fixes #5946 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

No tests

### Release notes

- Fix node server failing to connect on first try

### DIFF
--- a/templates/simple-server-example/package.json
+++ b/templates/simple-server-example/package.json
@@ -23,6 +23,7 @@
 	"devDependencies": {
 		"@types/bun": "^1.1.6",
 		"@types/express": "^4.17.21",
+		"@types/ws": "^8.18.1",
 		"concurrently": "^9.0.1",
 		"lazyrepo": "0.0.0-alpha.27",
 		"tsx": "^4.19.2",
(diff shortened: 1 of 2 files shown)


SomeHats (approved):
Thanks for this!

PR #5952, opened 2025-04-24
feat(editor): add Overlays component for camera-following overlays

SomeHats:
Tldraw has two main places to add stuff to the canvas: the `OnTheCanvas` and the `InFrontOfTheCanvas` components. 

`OnTheCanvas` follows the camera, and allows you to place content within shapes and below overlays. `InFrontOfTheCanvas` does not follow the camera, and allows you to place content above overlays.

This diff introduces a 3rd slot, `Overlays`, which exists between the two. It follows the camera, is above the shapes (and overlays, although that's configurable) but is below the `InFrontOfTheCanvas` layer. 

I've felt the need for something like this a few times, but the main motivation is from the hints we show with elbow arrows. For these, the existing things aren't really appropriate. Selection foreground (for example) can't be used because it's those overlays show when the selection is hidden, and are on a different shape to the one selected (the arrow's target).  

### Change type

- [x] `api`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- A new `Overlays` component lets you add overlays on top of the canvas that still follow the camera.


packages/editor/editor.css
```diff
@@ -472,6 +473,10 @@ input,
 	stroke-width: calc(2.5px * var(--tl-scale));
 }
 
+.tl-custom-overlays {
```
mimecuvalo, line 476: "custom" here sorta implies that it's not being used by tldraw and you can use it as an escape hatch for SDK users (but it will be used by elbow arrows). similarly, it implies that the other layers aren't customizable heh.

i'd just nix the "custom" part of this.

mimecuvalo (approved):
makes sense, pre-approving but i'd rethink the "custom" terminology.

packages/editor/editor.css
SomeHats: `tl-overlays` already exists! it contains a bunch of pre-defined things like selection fg/bg, handles, indicators, etc. This is like a way of getting something extra into the overlays? I couldn't think of a better name but open to other suggestions!

PR #5956, opened 2025-04-25
[VSCode] Allow users to drag files from the sidebar.

MitjaBezensek:
This allows users to drag image files from the sidebar into the editor. In this case we don't get the files (so the files content handler does not get called). Instead we get an url prefixed by `file://`. We have to send the url to the extension, since the webview runs in a sandbox and does not have file system access. The extension gets the file and sends it back.

Resolves ENG-3111

### Change type

- [x] `bugfix`

### DIFF
--- a/apps/vscode/editor/src/utils/externalUrlContentHandler.ts
+++ b/apps/vscode/editor/src/utils/externalUrlContentHandler.ts
@@ -9,7 +9,6 @@ export function registerExternalUrlContentHandler(editor: Editor) {
 		if (info.url.startsWith('file://')) {
 			// We cannot fetch the file directly since we are sandboxed, we have to ask the extension manager to get it for us
 			const data = await rpc('vscode:get-file', { url: info.url })
-
 			const uint8Array = new Uint8Array(data.file)
 			const blob = new Blob([uint8Array], { type: data.mimeType })
 			const file = new File([blob], data.fileName, { type: data.mimeType })
--- a/apps/vscode/editor/src/utils/externalUrlContentHandler.ts
+++ b/apps/vscode/editor/src/utils/externalUrlContentHandler.ts
@@ -1,0 +1,26 @@
+import { Editor } from 'tldraw'
+import { rpc } from './rpc'
+
+export function registerExternalUrlContentHandler(editor: Editor) {
+	const currentFilesContentHandler = editor.externalContentHandlers['files']
+	const currentUrlContentHandler = editor.externalContentHandlers['url']
+	editor.registerExternalContentHandler('url', async (info) => {
+		// This happens when you shift drag a file from the file explorer to the editor
+		if (info.url.startsWith('file://')) {
+			// We cannot fetch the file directly since we are sandboxed, we have to ask the extension manager to get it for us
+			const data = await rpc('vscode:get-file', { url: info.url })
+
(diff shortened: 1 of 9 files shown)


apps/vscode/extension/src/WebViewMessageHandler.ts
```diff
@@ -194,3 +219,22 @@ export class WebViewMessageHandler {
 		}
 	}
 }
+
+function getMimeTypeFromPath(filePath: string): string {
+	const extension = filePath.split('.').pop()?.toLowerCase()
+	if (!extension) throw new Error('No extension found in file path')
+
+	const mimeTypes: Record<string, string> = {
```
mimecuvalo, line 227: should we support videos too?

apps/vscode/extension/src/WebViewMessageHandler.ts
```diff
@@ -194,3 +219,22 @@ export class WebViewMessageHandler {
...
+		jpg: 'image/jpeg',
+		jpeg: 'image/jpeg',
+		png: 'image/png',
+		webp: 'image/webp',
+		gif: 'image/gif',
+		apng: 'image/apng',
+		avif: 'image/avif',
+		svg: 'image/svg+xml',
+	}
+	const mimeType = mimeTypes[extension]
```
mimecuvalo, line 237: might be also worth checking `DEFAULT_SUPPORTED_IMAGE_TYPES` (or `DEFAULT_SUPPORTED_MEDIA_TYPE_LIST` if we support videos too, and maybe support audio one day)

apps/vscode/editor/src/utils/externalUrlContentHandler.ts
```diff
@@ -0,0 +1,25 @@
+import { Editor } from 'tldraw'
+import { rpc } from './rpc'
+
+export function registerExternalUrlContentHandler(editor: Editor) {
+	const currentFilesContentHandler = editor.externalContentHandlers['files']
+	const currentUrlContentHandler = editor.externalContentHandlers['url']
+	editor.registerExternalContentHandler('url', async (info) => {
+		// This happens when you shift drag a file from the file explorer to the editor
+		if (info.url.startsWith('file://')) {
```
mimecuvalo, line 9: can it be multiple files?

mimecuvalo (commented):
couple questions, just new territory for me

apps/vscode/editor/src/utils/externalUrlContentHandler.ts
MitjaBezensek: I couldn't drag multiple files in my testing. We also don't support `urls` external handlers, only a singular `url`.

apps/vscode/extension/src/WebViewMessageHandler.ts
MitjaBezensek: Added some logic for this now 👍

mimecuvalo (approved):
nice!

apps/vscode/editor/src/utils/externalUrlContentHandler.ts
mimecuvalo: i wasn't sure if you could press `shift` and select multiple files / drag them that way

PR #5973, opened 2025-04-29
a11y: be able to navigate in/out of container shapes

mimecuvalo:
Adds Cmd+Shift+Up/Down for container navigation.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- a11y: Adds Cmd+Shift+Up/Down for container navigation.


mimecuvalo:
lemme develop this some more

assets/translations/main.json
```diff
@@ -255,6 +255,7 @@
 	"a11y.multiple-shapes": "{num} shapes selected",
 	"a11y.select-shape": "Select next shape",
 	"a11y.select-shape-direction": "Select shape in direction",
+	"a11y.enter-leave-container": "Enter/leave container",
```
MitjaBezensek, line 258: It might not be obvious to users what container means?

MitjaBezensek (approved):

assets/translations/main.json
mimecuvalo: Yeah, it's tricky - we need a generic enough word to capture both the Frame and Group shapes.
Container is what's used in Miro so ¯\\_(ツ)_/¯

PR #5975, opened 2025-04-29
Restore DPR constrained width/height on shapes.

SomeHats:
This reverts commit efa1a4b4b6fd0e266d06be80b1b86913143562f1. Without this DPR-based rounding logic, we see jitter on some shapes like this:

Kapture 2025-04-29 at 14 35 35

Fixes ENG-3193

### Change type

- [x] `bugfix`

### Release notes

- Prevent arrows and lines from jittering in some browsers as their width/height changes

### DIFF
--- a/packages/editor/src/lib/components/Shape.tsx
+++ b/packages/editor/src/lib/components/Shape.tsx
@@ -27,6 +27,7 @@ export const Shape = memo(function Shape({
 	index,
 	backgroundIndex,
 	opacity,
+	dprMultiple,
 }: {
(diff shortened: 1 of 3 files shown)


mimecuvalo:
👍 honestly i thought we had already landed this 😬

mimecuvalo (approved):

SomeHats:
We may be able to remove this fix again as chrome fixed this upstream: https://chromium-review.googlesource.com/c/chromium/src/+/6498206

PR #5977, opened 2025-04-29
Examples cleanup

TodePond:
The purpose of this PR is to do a big clean up of ALL of our examples to get things ready for a redesign / rehaul.

These changes are based on the surveys and interviews we carried out with developers, in an attempt to resolve some of the common hurdles we've seen. For example, developers have sometimes failed to see what the purpose of a specific example is, or they've stumbled across out-of-date information via our examples, or they've failed to find the example they need because it isn't labelled/positioned correctly.

From looking through each example one-by-one, I can see why many of these hurdles were happening. This PR attempts to resolve a lot of them.

# Changes

- Copy pass on titles, subtitles and descriptions.
  - All examples now have subtitles and descriptions. In practice, this means that every single example now explains itself when viewed on tldraw.dev. Nothing is there with no explanation.
  - All titles changed to sentence case.
  - Consistent punctuation and grammar.
  - Many examples have been renamed to better reflect what they are showing. Their names may have made sense at some point in the past, but now that our SDK has changed+grown, some of them benefit from a new name. eg: `Custom UI` -> `Replace the entire UI`.
  - Some examples have been renamed to use consistent vocab, eg: `Inline` -> `Inset`.

- URLs of examples shortened.
  - eg: `tldraw.dev/examples/shapes/tools/shape-with-geometry` -> `tldraw.dev/examples/shape-with-geometry`
  - The old URLs still work (they redirect to the new ones).
  - The purpose of this change is that we can reorganize examples and their categories without breaking old links and URLs (eg: within discord messages).

- Some specific examples have had their URL changed.
  - Typo fix: `tldraw.dev/examples/peristence-key` -> `tldraw.dev/examples/persistence-key`, and a redirect setup for the old link, which is in the wild in various places.
  - Renamed an example that we refer people to within our code in various places: 'State and store' -> 'Signals'. `tldraw.dev/examples/state-store` -> `tldraw.dev/examples/signals`. "State and store" is too vague and people have been struggling to find info on how to use our signals. This should help that. A redirect is setup for the old link, which is in the wild in various places.

- Out-of-date examples updated.
  - Many out-of-date APIs have been replaced with more up-to-date ones.
  - Some broken examples have been fixed.

- Some examples have been edited to be more suitable for the tldraw.dev page.
  - eg: The inset editor example now fits within the embedded example.

- **Categories refreshed.**
  - More categories have been added to split up some overblown categories. The "getting started" category was the most overblown, and included some very non-basic things.
    - Old: Getting started, UI & theming, Shapes & tools, Data & assets, Editor API, Collaboration, Use Cases
    - New: Getting started, Configuration, Editor API, UI & theming, Page layout, Events & effects, Shapes & tools, Collaboration, Data & assets, Use cases
  - Many examples have changed categories to make our categorization more consistent. eg: 'Note resizing' and 'Frame colors' are now in the same category, as they both deal with configuring shape options.
  - Priority levels added on every example, surfacing more basic and important examples nearer the top of the list.
  - The new 'getting started' category is now very small (two examples), with the sync demo example being hardcoded to appear there as well as the collaboration category. Previously we saw a huge number of people looking at the custom renderer example, just because it's near the top.

- Some bug fixes.
  - eg: Stopped inconsistent word wrapping in the sidebar depending on whether something's selected or not.
  - eg: Large images like PDFs and images getting included in the code panels, often crashing the page.

- Other changes
  - I removed the persistence key from the basic example. I know this is annoying for local development, but it didn't match what's listed in the example description - and it's the first thing people see! I'm open to changing this if it's considered too annoying.

# Next steps

There is plenty more that I want to do.
- Present our examples in a way that's easier to navigate around them.
  - eg: By making 'related examples' a first-class thing. Each example should link to other related ones, and potentially relevant docs, starters, references, etc...
  - eg: By designing an overview page, with search, filtering, etc...
- Better support our "bottom-of-the-file guides". We use a consistent format, and we could make these more visible within the examples pages themselves, while keeping them good for local development.
- There are plenty more examples that we should add, and plenty that we should improve and/or simplify.
- Make our examples even more copy-paste-able. eg: Make sure they're all truly standalone. eg: Add buttons to copy, or open in stackblitz, etc...
- All of this and more, without giving the wrong impression that these are marketing examples. We need to be clear that these are API examples to be used as a reference.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- [Docs]: Restructured example categories.
- [Docs]: Updated all examples to contain up-to-date info.
- [Docs]: Updated copy for various examples for clarity.


steveruizok:
I'm still hesitant to redesign this page in a way that foregrounds our "API reference examples" in the same way that most companies foreground more designed starter kits or marketing examples.

Our examples are designed as developer reference, they're not designed to be browsed or to give an idea of what can be made with the tldraw SDK. Many of our examples are visually identical to each-other. They intentionally use the least amount of styling code as possible. Their code is meant to be read.

This is a very different user story than "let's see what this tldraw SDK can be used for", which would require a very different kind of example. We're totally capable of _making_ that example but it would be embarrassing to use our develop examples for that purpose. In a way we're already doing this on tldraw.dev, which I'm barely comfortable with, because it's made us feel like the examples problem is solved when it's very not.

If this content is meant for developer users, what is the optimal presentation of that content for that user?

If there are other users who we want to address, who are they? What would be their optimal content? What would be the optimal presentation of that content for that user?

TodePond:
I believe this is ready, but still to come: Tomorrow morning I'll add a description with all the changes and thinking behind them.

cc @SomeHats would be keen on getting your take on these when I've done that.

TodePond:
Hey @SomeHats I've added a description outlining the changes. Would love to get your input and to get this merged before our release this week.

apps/docs/scripts/lib/generateSection.ts
```diff
@@ -190,13 +190,18 @@ function getArticleData({
 
 	const { content } = parsed
 
+	// if (sectionId === 'examples') {
+	// 	console.log('order', order, 'priority', priority)
+	// }
```
SomeHats, line 195: nit: remove

apps/examples/src/examples/custom-renderer/README.md
```diff
@@ -1,8 +1,9 @@
 ---
 title: Custom renderer
 component: ./CustomRendererExample.tsx
-category: basic
+category: ui
```
SomeHats, line 4: not sure if i agree with "ui" categorization for this one but i can't think of a better one.

I kinda hate this example though and would be down to set `hide: true` on this one

apps/examples/src/examples/hosted-images/README.md
```diff
@@ -6,3 +6,10 @@ priority: 2
 hide: true
```
SomeHats, line 6: not sure why this one is hidden but maybe remove this?

apps/examples/src/examples/inset-canvas/inset-canvas.css
```diff
@@ -1,4 +1,7 @@
 .tldraw__editor-with-inset-canvas .tl-canvas {
 	position: absolute;
-	inset: 100px 200px 100px 100px;
+	/* inset: 100px; */
```
SomeHats, line 3: remove

apps/examples/src/examples/toSvg-method-example/CustomShapeToSvgExample.tsx
```diff
@@ -67,15 +65,15 @@ export class MyShapeUtil extends ShapeUtil<ICustomShape> {
...
-	}
+	// override toSvg(
+	// 	shape: ICustomShape,
+	// 	ctx: SvgExportContext
+	// ): ReactElement | null | Promise<ReactElement | null> {
+	// 	// ctx.addExportDef(getFontDef(shape))
+	// 	const isDarkmode = ctx.isDarkMode
+	// 	const fill = isDarkmode ? DARK_FILL : LIGHT_FILL
+	// 	return this.getSvgRect(shape, { fill })
+	// }
```
SomeHats, line 76: should this be commented?

SomeHats (approved):

SomeHats:
this looks great! big improvement

apps/examples/src/examples/custom-renderer/README.md
TodePond: There are definitely some people I speak to who want that sort of thing. The category ID is "ui" but the title of that category is "UI & theming" so I think it works okay here. I also considered having this one and some others in an "Advanced" category - all themed around using tldraw as an engine. But that felt too vague

TodePond:
Thanks @SomeHats will do a quick PR resolving some of these, thanks for catching

apps/examples/src/examples/hosted-images/README.md
TodePond: Interestingly, we don't hide ANY examples within tldraw.dev, and that's what I was looking at improving. But either way, yes, will remove this

apps/examples/src/examples/toSvg-method-example/CustomShapeToSvgExample.tsx
TodePond: no good spot

PR #5981, opened 2025-04-30
Fix active states for buttons in toolbars

steveruizok:
This PR fixes a bug where active items were not shown in style panel toolbars.

Before:

image

After:

image

### Change type

- [x] `bugfix`

### Test plan

1. Open some menus

### Release notes

- Fixed a bug in the UI where active items in dropdowns weren't displayed as active

### DIFF
--- a/packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
+++ b/packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
@@ -132,9 +132,7 @@ export const TldrawUiButtonPicker = memo(function TldrawUiButtonPicker<T extends
 						data-testid={`style.${uiType}.${item.value}`}
 						aria-label={label}
 						value={item.value}
-						data-state={
-							value.type === 'shared' && value.value === item.value ? 'hinted' : undefined
-						}
+						data-isactive={value.type === 'shared' && value.value === item.value}
 						title={label}
 						className={classNames('tlui-button-grid__button')}
 						style={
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
```diff
@@ -132,9 +132,7 @@ export const TldrawUiButtonPicker = memo(function TldrawUiButtonPicker<T extends
 						data-testid={`style.${uiType}.${item.value}`}
 						aria-label={label}
 						value={item.value}
-						data-state={
```
mimecuvalo, line 135: actually, now that this is a ToggleItem, we might have to, optionally also do `data-state=on|off`: https://www.radix-ui.com/primitives/docs/components/toolbar#toggleitem
i think it works without it though

mimecuvalo (approved):
huh, looks like this has broken for a long while, i can't tell when it regressed after looking back a bit

packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
steveruizok: Looks like that system is still being worked on (i.e. a value for the toggle group that would be picked up by the toggle item), but if it ends up being the same logic as isactive then that would be better.

mimecuvalo: yeah i wouldn't worry about it, this is good to land

PR #5988, opened 2025-05-01
Better computed cache memoization

ds300:
This fixes one bug and makes two minor perf adjustments.

1. [Bug] The geometry cache now uses shape.meta as part of the cache key, while previously it only used shape.props. This meant that SDK users couldn't do conditional geometry based on meta properties on their custom shapes. I guess we expected custom shapes not to really use meta, but it happened!
2. [Perf] Narrowed the cache key for two of our other caches to avoid wasted recomputations.

### Change type

- [x] `bugfix`

### Release notes

- Allow changes to shape.meta to bust the shape geometry cache.


steveruizok:
That's great. Any tests or behaviors where this would make a visible change? I remember looking at this code when working on #5821 but don't remember exactly what led me there.

ds300:
>  Any tests or behaviors where this would make a visible change

for the perf improvements, not really. Maybe if you had a shape with loads of handles (like a really long line shape with hundreds of segments) and then someone else dragged it around while you had it selected?

For the bugfix, it's only noticeable if a SDK user has a custom shape type that uses shape.meta for properties that could affect the geometry of the shape. I'll add a regression test for this.

PR #5994, opened 2025-05-01
a11y: fix popover focus

mimecuvalo:
(this is on top of another PR, the Cmd+Enter one so it's more easily visible how the change affects tabbing through shapes)

This selects the first, visible, non-disabled button in a popover when opened. I made the default true but I'm not opposed and could be convinced that the default should be false.

- Drive-by fix here to add a placeholder to the Edit link dialog.
- also, a fix for the Slider, in MobileStylePanel stealing focus when being opened.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: focus first item in popover when opened


mimecuvalo:
Yeah, the heuristic here for the browser is that you have to be tabbing around a bit for focus-visible to turn on.
So, you can't just click on the popover button and expect the focus ring to show up immediately.
You need to tab to the toolbar in question, hit Enter on the popover to open it up and then you'll see the intended result of this PR.

steveruizok:
Sorry, what I meant is:

- Select a shape
- Press Cmd+Enter to focus the style panel
- Expect to see focus ring

But understand now that this is addressing something else!

mimecuvalo:
> Sorry, what I meant is:
> Select a shape/Press Cmd+Enter to focus the style panel/Expect to see focus ring
> But understand now that this is addressing something else!

Right, I think I understood that in the first go around — what I was explaining was though is that `:focus-visible` won't be set by the browser though if the "Select a shape" step came from a mouse click though, you'll need to do a keyboard navigation to get there first.

MitjaBezensek (approved):

PR #6000, opened 2025-05-02
Export menu improvements

ds300:
1. Use a checkerboard background to indicate transparency
2. Move the export button above the preview so we can allow the preview to be dynamically sized, which in turn fixes a lot of UX weirdness that was going on there.

@mimecuvalo code is not final, just looking for UX feedback right now, I know you had some opinions about this component :)

### Change type

- [x] `other`


mimecuvalo:
lookin' great! :chefs-kiss:

ds300:
@steveruizok thanks for the feedback. I felt the same about the narrow images issue. I fixed that by tweaking the borders and adding some padding

image

> they force us to put the button above the preview in order to maintain location consistency, which is inconsistent vs other views

Which other views? The QR code in the sharing tab is below the copy link button, so for this menu it actually is more consistent to keep the preview below the export button?

> seems unrelated to this bug

At first I did fix just the image size bug, but it still looked bad due to the existing UX issues that we also have tickets for, so I thought it might be worth tackling everything at once.

Pretty happy with it now, gonna move ahead, let me know if you have any more feedback or feel strongly enough to veto.

mimecuvalo (approved):
:chefs-kiss: nice one!

steveruizok:
going to start blocking PRs that introduce UI changes, cmon

PR #6001, opened 2025-05-02
a11y: address icons and handles

mimecuvalo:
We have some English hard coded in places and also some missing strings on icons.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: add labels to icons and handles


assets/translations/main.json
```diff
@@ -408,5 +415,29 @@
...
+	"handle.rotate.bottom_right_rotate": "Rotate bottom right",
+	"handle.rotate.mobile_rotate": "Rotate",
+	"handle.crop.top": "Crop top",
+	"handle.crop.bottom": "Crop bottom",
+	"handle.crop.left": "Crop left",
+	"handle.crop.right": "Crop right",
+	"handle.crop.top-left": "Crop top left",
+	"handle.crop.top-right": "Crop top right",
+	"handle.crop.bottom-left": "Crop bottom left",
+	"handle.crop.bottom-right": "Crop bottom right",
```
steveruizok, line 439: truly hope that screen readers aren't reading the overlay UI

assets/translations/main.json
```diff
@@ -310,6 +310,7 @@
 	"status.offline": "Offline",
 	"people-menu.title": "People",
 	"people-menu.change-name": "Change name",
+	"people-menu.avatar-color": "Avatar color",
```
steveruizok, line 313: This be "Color", I think, given that the change message is "change-color". We don't use the word "Avatar" anywhere else.

assets/translations/main.json
mimecuvalo: The thing is that it's part of the a11y tree by definition (even if tabbing over things skips over things at the moment):

Screenshot 2025-05-06 at 15 04 10

mimecuvalo: example here is that the "Clone note" buttons are for the nibs

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -109,6 +109,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 			{
 				id: 'tail',
 				type: 'vertex',
+				label: 'Move tail',
```
MitjaBezensek, line 112: Shouldn't we be localizing things? Or do we only use english for a11y?

packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
```diff
@@ -477,12 +467,15 @@ export const RotateCornerHandle = function RotateCornerHandle({
 	'data-testid'?: string
 }) {
 	const events = useSelectionEvents(corner)
+	const msg = useTranslation()
+	const label = msg(`handle.rotate.${corner}`)
```
MitjaBezensek, line 471: I guess this is not typechecked properly?

MitjaBezensek (approved):

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
mimecuvalo: this is apps/examples so it doesn't matter :P

packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
mimecuvalo: 😬 yeah it's not really typechecked b/c of the signature: `msg(id?: Exclude<string, TLUiTranslationKey> | string)`
i think we should move away from this system in general in the coming year.

apps/examples/src/examples/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
MitjaBezensek: Ah, didn't see 😅

PR #6009, opened 2025-05-06
[RFC] i18n: wire through to the editor package

mimecuvalo:
This starts allowing i18n to be accessed at the editor level. We have some places (notably in shapes) where there is a necessity for text to be translated but no way to do so in a general way.

This came more to light via the a11y work and forcing this conversation when Alex and I were discussing.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- i18n: wire through to the editor package


steveruizok:
Making an issue for general i18n modernization.

PR #6012, opened 2025-05-06
clipboard: fix pasting shape data accidentally into a text area

### LINKED ISSUE #5871: Pasting shape in rich text field causes issues
### What happened?

If you copy a shape from tldraw and paste it in a rich text field it will paste the base64 encoded data into the text field.

I'm not sure what the preferred behavior is here. Perhaps just doing nothing? But at least not pasting all the base64 encoded data.

1. Copy a shape
2. Edit e.g. a note
3. Paste

*No response*

### Where did this happen?

The [tldraw.com](http://tldraw.com) website

### **Possible Expected Behaviour**

* If it's a text shape then we paste the text content ? 
* If its text paste it inside otherwise paste the shape ?

### Contact Details

*No response*

### Code of Conduct

mimecuvalo:
fixes #5871 so that we don't accidentally get base64 data into a shape

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- clipboard: fix pasting shape data accidentally into a text area


trygve-aaberge-adsk:
Maybe not a big deal, but I see that with this it pastes nothing into a rich text shape, while if you try to paste it into the text field of an arrow, it will paste the text (which was also the behavior for all shapes before rich text).

Another thing, if you try to paste into an application that can handle HTML, you will still just get the base64 text (this isn't a regression btw, it was like that before rich text too). Could this be improved somehow? For instance, instead of the pasted content being a div with the base64 text inside it, could it be a HTML version of the rich text, with the base64 data for tldraw in a data attribute instead?

steveruizok:
Ah wait I think there's a better way of doing this! Selecting text and pasting tldraw shapes clears the text. We should really prevent it by a check for the tldraw data onPaste, rather than transforming the content. (This will work for plain text too)

steveruizok:
Pushing up a commit that gets this done, @mimecuvalo could you review (feel free to revert if you catch issues)

Issue #5881, opened 2025-04-11
[Bug]: `list-style: none` in `.tl-rich-text` at tldraw.dev


prichey:
### What happened?

I noticed that if I create an `ol` or `ul` in a text element in the demo at tldraw.dev, the disc / number prefix on the `li` is not visible. I believe this is due to the [Tailwind reset](https://tailwindcss.com/docs/preflight#lists-are-unstyled):

```css
ol, ul, menu {
  list-style: none;
}
```

This is not an issue at tldraw.com, where Tailwind doesn't appear to be used.

tldraw.dev|tldraw.com
-|-
<img width="152" alt="Image" src="https://github.com/user-attachments/assets/5cad3f5b-951b-4bd2-b664-44c8f18ce6ce" />|<img width="157" alt="Image" src="https://github.com/user-attachments/assets/5dbad5c8-b229-479c-aeba-fcff12cdcde5" />

### How can we reproduce the bug?

In the demo at tldraw.dev, add a text element with the following content `* hello`. The disc on the `li` isn't visible.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The tldraw.dev website

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nayounsang:
Can I try this issue?

mimecuvalo:
Ach, good catch.
@nayounsang sure! We might tackle it ourselves today though.

SomeHats:
we should stop relying on browser css for rich text styling and explicitly write out all the styles that we're assuming will be present in the SDK css. otherwise people will be running into this sort of thing with their own CSS resets

Issue #5870, opened 2025-04-10
Blank lines render differently in edit mode


trygve-aaberge-adsk:
### What happened?

For shapes using rich text, unless the `TextDirection` extension is loaded, blank paragraphs will only take up space when editing, not when the shape is not in editing state. So when editing I have a blank line between two paragraphs, and when exiting editing the blank line disappears.

The `TextDirection` extension isn't exported, so that means I have to use `tipTapDefaultExtensions` in order to avoid this behavior.

What happens is that when not editing it's using `renderHtmlFromRichText` and this inserts a `<br />` into `<p dir="auto"></p>`, but when `TextDirection` isn't loaded the paragraphs don't have a `dir` attribute.

### How can we reproduce the bug?

1. Open tldraw with a config that doesn't use `tipTapDefaultExtensions`. E.g. the `rich-text-custom-extension` example.
2. Create a text shape.
3. Write something, then press enter twice and write something more.
4. Observe that there's a blank line between the two texts.
5. Deselect the shape.
6. Observe that the blank line disappears, and the texts are right beneath each other.

### Where did this happen?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

trygve-aaberge-adsk:
It's also an issue if you have paragraphs with other styling, even if `TextDirection` is included. E.g. when using `@tiptap/extension-text-align` and setting an alignment on the paragraph. Then it won't be matched by the replacement because it has a style attribute.

trygve-aaberge-adsk:
Maybe instead of replacing empty paragraphs, we could set something like `min-height: ${lineHeight}em` on `p` elements?

trygve-aaberge-adsk:
It also happens with `br` elements at the end of text. I.e. if you don't have the `KeyboardShiftEnterTweakExtension` and press shift+enter at the end of editing a note, you'll see the text move when deselecting the note.

mimecuvalo:
Great report @trygve-aaberge-adsk. Let me start with the easy stuff  by exporting some extensions and then I'll take a look at addressing some of the other issues too in the coming weeks.

PR #6013, opened 2025-05-06
rich text: fix up styling of empty lines and lists

mimecuvalo:
fix https://github.com/tldraw/tldraw/issues/5881 and https://github.com/tldraw/tldraw/issues/5870

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- rich text: fix styling of empty lines in some cases and lists in some cases

### DIFF
--- a/packages/editor/editor.css
+++ b/packages/editor/editor.css
@@ -969,6 +969,8 @@ input,
 
 .tl-rich-text p {
 	margin: 0;
+	/* Depending on the extensions, <p> tags can be empty, without a <br />. */
+	min-height: 1lh;
 }
 
 .tl-rich-text ul,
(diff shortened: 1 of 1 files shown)


packages/editor/editor.css
```diff
@@ -969,13 +969,17 @@ input,
...
+	min-height: 1lh;
 }
 
 .tl-rich-text ul,
 .tl-rich-text ol {
 	text-align: left;
 	margin: 0;
 	padding-left: 3.25ch;
+	/* Some resets, like Tailwind, nix the list styling. */
+	list-style: revert;
```
MitjaBezensek, line 982: Are there any other styles that we rely on, but might be reset?

mimecuvalo: as far as i know, i _think_ this will be sufficient. 🤞

Issue #5966, opened 2025-04-28
[Feature]: Support letterSpacing for PlainTextLabel and textMeasure


kwonuj:
### What's the feature?

I'm having trouble finding the exact value when I want to measure the value with letterSpacing applied using the textMeasure.measureText function.

### Where would you like this feature?

The developer package

### Contact Details

kwonuj96@gmail.com

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
Thanks for the bug report. Just to make sure I'm understanding your usecase right, is the measurement of text producing incorrect results because letter-spacing rules in your CSS? Is it causing inconsistencies in rendering?

I'm also wondering, as an aside to myself, if it's not just letter-spacing but other CSS rules we should be capturing more holistically in `measureText`

kwonuj:
I’m trying to get the exact width of text with various CSS properties applied, including letterSpacing. However, I can’t pass letterSpacing as a parameter to the measurement function provided by `TextManager` class.

For now, I’m fine with just adding support for the letterSpacing option, but I think I’ll need to consider whether there are other properties to handle as well.

PR #6014, opened 2025-05-06
text: be able to add other styling to measure by if applicable

mimecuvalo:
fixes https://github.com/tldraw/tldraw/issues/5966
We want to be able to specify things like `letter-spacing` in text measurement. However, at the same time, we don't need to enumerate the rest of the ~20 CSS text properties explicitly. (e.g. `font-optical-sizing`, `font-stretch`, etc.). So, this let's people specify any other styles that are relevant for measurement.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- text: be able to add other styling to measure by if applicable

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -2934,6 +2934,7 @@ export class TextManager {
         fontWeight: string;
         lineHeight: number;
         minWidth?: null | number;
+        otherStyles?: Record<string, string>;
         padding: string;
     }): BoxModel & {
         scrollWidth: number;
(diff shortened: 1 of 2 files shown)


MitjaBezensek (approved):

steveruizok:
This is going to cause problems if we re-use the same element each time, which is an optimization I'm pretty keen to try, because these properties won't be turned off and will be re-used. We may need to allow users to specify default properties, too.

PR #6016, opened 2025-05-07
Allow setting of autoplay video attribute via TldrawOptions

elstob:
This pull request is a low impact attempt to resolve https://github.com/tldraw/tldraw/issues/5716.

Automatically playing videos can be a significant performance and user experience issue on canvases where multiple videos are included. The original issue proposes adding an autoplay prop to video shapes as well as a default option for initial video creation. After experiencing the issue for our use case it feels that the second point of being able to globally control autoplay behaviour is the lowest effort way to solve the main pain points.

Here `autoplayVideos` is introduced as a new `TldrawOption` and it is used solely in the `VideoShape` to set the video element attribute.

Although it would be amazing from a personal perspective if this could be added as is I'm more than happy to discuss further improvements to the PR. I feel possibilities may include:

-  Add autoplay prop to video shapes as per original issue (and maybe context menu option for videos to toggle)
- Change `autoplayVideos` boolean option to a `videoAttributes` subset object of configurable video element props. (e.g muted, loop etc) so as to be able to define other currently hardcoded properties via tldraw options.

Look forward to hearing your thoughts!

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Add a video to a canvas with persistence.
2. Reload canvas to note video autoplays.
3. Update canvas instance with option `autoplayVideos` set to `false` e.g. `<Tldraw persistenceKey="example" options={{ autoplayVideos: false }} />`.
4. Reload canvas to note video no longer autoplays.

### Release notes

- Allow video autoplay behaviour to be controlled by a new `autoplayVideos` option.


elstob:

mimecuvalo (commented):
Hey there! 

In general, the way we would add this is by letting people using `configure`/`options` on a particular shape, see this example: https://github.com/tldraw/tldraw/blob/main/apps/examples/src/examples/configure-shape-util/ConfigureShapeUtilExample.tsx
For example, you could change ArrowShapeUtil's options for various things in this way.

For this particular idea, I'm not sure we'd add it, we'd have to think about it. The issue is mainly when there are multiple videos on the canvas, you don't want 50 videos playing all at once :P I could see something more like wanting to have something autoplay on a per instance, not _all_ of a particular shape. And even then, when having autoplay on a canvas, it's worth considering that can be potentially jarring for users not expecting that behavior.

elstob:
@mimecuvalo - thanks for the reply, I could definitely change the PR so that the VideoShapeUtil is configurable with an autoplay parameter if that is preferable.

What you said about having 50 videos autoplaying on the canvas is definitely true - in fact that is the current behaviour within tldraw as the autoplay value is hardcoded on all videos at the moment! (nightmare).

If this isn't intended then I would just as happily remove [this line](https://github.com/tldraw/tldraw/blob/4389e597566020cb54a60eceee877c14ba8c737d/packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx#L169) from `VideoShapeUtil.tsx` in this PR as that would solve our issue with videos autoplaying in the bluntest way possible, but this would also be a behavioural change that people might currently be depending on so I guess not acceptable.

The PR as is does at least give the option to toggle this which is why I hoped it might be sufficient but please let me know if it is possible to proceed in some what as right now having multiple videos autoplay is not great for our UX. Thanks again!

elstob:
@mimecuvalo - I'm going to close this PR and instead open another one for discussion that uses the shape `configure` approach you suggested. Hope that's acceptable!

mimecuvalo:
> @mimecuvalo - I'm going to close this PR and instead open another one for discussion that uses the shape `configure` approach you suggested. Hope that's acceptable!

sounds good! lemme check internally whether we would like autoplay to be default `true` - that's surprising to me

PR #6021, opened 2025-05-07
Docs: Remove section on rich text exports

TodePond:
This PR removes a small section about rich text exports that doesn't help much. Noticed it while looking through our docs and examples - pulled out this small change into its own PR.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/docs/content/docs/shapes.mdx
+++ b/apps/docs/content/docs/shapes.mdx
@@ -282,11 +282,6 @@ export default function RichTextCustomExtensionExample() {
 
 On our [examples site](https://examples.tldraw.com/) please filter by "rich text" and you will find examples on how to add custom extensions or UI to tailor the rich text.
 
-#### Exports for rich text
-
-One of the other challenges right now with rich text is the exports. We use the SVG `foreignObject` to insert rich text into the shapes.
-This isn't ideal and we're looking in the future to help improve this.
-
 ### Migrations
 
 You can add migrations for your shape props by adding a `migrations` property to your shape's util class. See [the persistence docs](/docs/persistence#Shape-props-migrations) for more information.


apps/docs/content/docs/shapes.mdx
```diff
@@ -282,11 +282,6 @@ export default function RichTextCustomExtensionExample() {
 
 On our [examples site](https://examples.tldraw.com/) please filter by "rich text" and you will find examples on how to add custom extensions or UI to tailor the rich text.
 
-#### Exports for rich text
-
-One of the other challenges right now with rich text is the exports. We use the SVG `foreignObject` to insert rich text into the shapes.
-This isn't ideal and we're looking in the future to help improve this.
```
trygve-aaberge-adsk, line 288: Isn't it useful to know that rich text shapes are using `foreignObject` when exporting to SVG? I found that useful to know at least.

TodePond: Hey, maybe! In what way did this help you?

trygve-aaberge-adsk: Typical SVG editors can't edit SVGs with foreignObject, so in general it's good to be aware what shapes use that (e.g. if our users ask us about editing exported SVGs).

We have some custom shapes which don't implement `toSvg`, so they use the fallback which uses foreignObject when exporting to SVG. Knowing that rich text uses foreignObject means that there's not really any point of implementing `toSvg` until rich text generates SVG without foreignObject (though, this wasn't on our priority list anyways).

Also, it's nice to know that you're looking to improve this in the future. But if you're not planning this, that part should of course be removed.

TodePond: Okay, thank you for that feedback. That's helpful!
I think we should keep that information somewhere - though it should perhaps be somewhere more discoverable. I'll have a look.

steveruizok: My issue here is with the language, which isn't helpful. A callout that text is exported as foreign objects is plenty, no need to apologize or make it sound like this is still being worked on.

PR #6023, opened 2025-05-07
Improve performance of a programatically rotating shape when it is selected

MitjaBezensek:
Looks like the main slowdown is downloading the updated cursors over and over again and then rendering them. Check how many requests the browser makes in the before case.

### Before

https://github.com/user-attachments/assets/34a6f317-e605-4367-b8b7-f6e26cef07e6

### After

https://github.com/user-attachments/assets/2ab46925-3f66-446a-b4d2-c4715b3d2fe6

Hovering still results in slowdown, but that's just the the result of updating the cursor that many times.

https://github.com/user-attachments/assets/5ea221e5-3bbc-4da0-9105-c5507af77ee4

### Change type

- [x] `improvement`

### Release notes

- Improve the performance of shape rotation. This slowdown could be seen when the shape was rotated programmatically and you selected it. Or in a multiplayer scenario if you selected a shape that somebody else was rotating.

### DIFF
--- a/packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
+++ b/packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
@@ -14,7 +14,7 @@ import {
 	useValue,
 } from '@tldraw/editor'
 import classNames from 'classnames'
-import { useRef } from 'react'
+import { PointerEventHandler, useRef, useState } from 'react'
 import { useReadonly } from '../ui/hooks/useReadonly'
 import { TldrawCropHandles } from './TldrawCropHandles'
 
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
```diff
@@ -459,6 +443,51 @@ export const TldrawSelectionForeground = track(function TldrawSelectionForegroun
...
+			className={className}
+			data-testid={dataTestId}
+			role="button"
+			aria-label={`${position} target`}
+			pointerEvents="all"
+			x={x}
+			y={y}
+			width={width}
+			height={height}
+			onMouseEnter={() => setMouseOver(true)}
```
mimecuvalo, line 483: very nice, so simple!

mimecuvalo (approved):
🎉

Issue #5387, opened 2025-02-08
[Bug]: Inconsistent pan behavior


jamesbvaughan:
### What happened?

When `panSpeed` is set to something other than `1`, the resulting pan behavior is inconsistent between hand tool panning and middle-click-drag panning.

See the following recordings, both with `panSpeed: 2`:

Panning via middle-click-drag:

![Image](https://github.com/user-attachments/assets/830e4b82-bce6-42ea-96a1-21882718b7fd)

Panning via the hand tool:

![Image](https://github.com/user-attachments/assets/935e7799-dd5d-40ce-b8d3-a703584d32a8)

Personally, I like how the hand tool behaves, as if `panSpeed` is always `1` and I'd want middle-click-drag panning to behave similarly. That way, the `panSpeed` setting can be used to control how panning via trackpad swipes feels, while more direct pans (both hand tool panning and middle-click-drag panning) continue to feel natural because the page "sticks" to the cursor.

I might take a crack at this myself, but it's not clear to me whether my preference is actually aligned with others'.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Where did this happen?

The developer package

### Contact Details

james@arcol.io

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #6024, opened 2025-05-07
Fix inconsistent panning.

MitjaBezensek:
Makes the pan speed consistent between the hand tool and space / middle mouse button panning.

After talking to Steve we decided to go with the [suggestion](https://github.com/tldraw/tldraw/issues/5387) of keeping the panning at speed 1 for those two actions and only take panning speed into account for mouse wheel / trackpad panning. Otherwise the contents of the canvas drift underneath your cursor, which feels wrong.

### Change type

- [x] `bugfix`

### Test plan

1. Use the camera options example.
2. Increase / decrease the pan speed.
3. Panning should be consistent between the hand tool and space bar / middle mouse button panning. It should use pan speed of 1 (shapes should not drift away from the hand cursor when doing it).
4. Panning speed should still be taken into account for wheel panning and trackpad panning.

### Release notes

- Don't take pan speed into account for hand tool panning so that it's consistent with middle mouse button / spacebar panning.

### DIFF
--- a/packages/tldraw/src/test/commands/setCamera.test.ts
+++ b/packages/tldraw/src/test/commands/setCamera.test.ts
@@ -206,7 +206,7 @@ describe('CameraOptions.wheelBehavior', () => {
 })
 
 describe('CameraOptions.panSpeed', () => {
-	it('Effects wheel panning (2x)', () => {
+	it('Affects wheel panning (2x)', () => {
 		editor
 			.setCameraOptions({ ...DEFAULT_CAMERA_OPTIONS, panSpeed: 2, wheelBehavior: 'pan' })
 			.dispatch({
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/test/commands/setCamera.test.ts
```diff
@@ -206,7 +206,7 @@ describe('CameraOptions.wheelBehavior', () => {
 })
 
 describe('CameraOptions.panSpeed', () => {
-	it('Effects wheel panning (2x)', () => {
+	it('Affects wheel panning (2x)', () => {
```
mimecuvalo, line 209: your place in heaven is secure :P

mimecuvalo (approved):
👍

PR #6027, opened 2025-05-08
Make sure the file exists before uploading it to the bucket.

MitjaBezensek:
This makes sure that the file exists before we upload an asset to the R2 bucket. Otherwise you could upload assets and not have them be associated with a file.

This also uses a queue for inserting the asset <> file association. Otherwise the insert might fail and we might still not correctly associate assets to files. Using a queue allows us to retry the inserts with an exponential backoff.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/sync-worker/src/types.ts
+++ b/apps/dotcom/sync-worker/src/types.ts
@@ -1,5 +1,6 @@
 // https://developers.cloudflare.com/analytics/analytics-engine/
 
+import { Queue } from '@cloudflare/workers-types'
 import type { RoomSnapshot } from '@tldraw/sync-core'
 import type { TLDrawDurableObject } from './TLDrawDurableObject'
 import type { TLLoggerDurableObject } from './TLLoggerDurableObject'
@@ -63,6 +64,8 @@ export interface Environment {
 	HEALTH_CHECK_BEARER_TOKEN: string | undefined
 
 	RATE_LIMITER: RateLimit
+
+	QUEUE: Queue<QueueMessage>
 }
 
 export function isDebugLogging(env: Environment) {
@@ -153,3 +156,9 @@ export type TLUserDurableObjectEvent =
 	  }
 	| { type: 'reboot_duration'; id: string; duration: number }
 	| { type: 'cold_start_time'; id: string; duration: number }
+
+export interface QueueMessage {
+	type: 'asset-upload'
+	objectName: string
+	fileId: string
+}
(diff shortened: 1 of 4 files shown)


apps/dotcom/sync-worker/src/types.ts
```diff
@@ -153,3 +156,9 @@ export type TLUserDurableObjectEvent =
 	  }
 	| { type: 'reboot_duration'; id: string; duration: number }
 	| { type: 'cold_start_time'; id: string; duration: number }
+
+export interface QueueMessage {
+	type: 'asset-upload'
```
MitjaBezensek, line 161: Went with a more generic approach, which should allow us to process different kinds of messages here.

Alternately we could create a separate queue for each use case.

apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -195,6 +197,21 @@ export default class Worker extends WorkerEntrypoint<Environment> {
...
+	override async queue(batch: MessageBatch<QueueMessage>): Promise<void> {
+		const db = createPostgresConnectionPool(this.env, 'sync-worker-queue')
+		for (const message of batch.messages) {
+			const { objectName, fileId } = message.body
+			try {
+				await db.insertInto('asset').values({ objectName, fileId }).executeTakeFirstOrThrow()
+				message.ack()
+			} catch (_e) {
+				message.retry({
+					delaySeconds: QUEUE_BASE_DELAY ** message.attempts,
```
ds300, line 210: should we put some max limit on this? like after 10 attempts maybe we should stop trying altogether, because the file id is most likely wrong? It's unlikely that the db would be down for 2^10 seconds (~20 minutes) at least.

ds300 (approved):
looks good, just one question about handling failures

apps/dotcom/sync-worker/src/worker.ts
ds300: i guess we already validated the file id by this point 🤔 

still technically possible we could hard delete a file and cause this to get in an infinite failure loop.

maybe stop after 12 attempts or something?

MitjaBezensek: Yeah, no pointing in retying into perpetuity 😁

Added a `max_retries` option and a dead letter queue for production:
https://github.com/tldraw/tldraw/pull/6027/commits/11d613e4e33a0be074de3851ba53eb0a3fc4311e

PR #6029, opened 2025-05-08
arrows: create min-width for label indicator

mimecuvalo:
With elbow arrows, we have a lot more vertical segments. When starting editing on the label, you can't see the cursor blinking b/c it overlaps with the shape. This makes the rectangle indicator show up when editing, even if we don't have text in the label yet.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- arrows: create min-width for label indicator


SomeHats:
Discussed offline, 3 issues:
1. Geometry is always present and clickable, it should only be included when this is the editing shape
2. There is a background on text when editing that shouldn't be there (pre-existing)
3. The occluded/masked part of the arrow doesn't match up with the textbox outline

(didn't mention this, but I would also probably make the min width maybe ~1/2 the size that it is now - a lot of characters cause it to shrink rather than grow at the moment.

PR #6041, opened 2025-05-10
Prevent announcements during interactions

steveruizok:
This PR fixes a bug causing accessibility announcements to occur too often. The code as written was firing during interactions (translating a shape, etc) and on hover/unhovers, due to the announcement function being captured. This narrows it to only when appropriate.

This announcement isn't free so if there's a way to know whether the user is consuming this announcements, we can turn them off until that's true.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Turn on screenreader
2. Drag a shape around
3. Select it
4. Deselect it

### Release notes

- Fixed a bug causing too many screenreader announcements

### DIFF
--- a/packages/tldraw/src/lib/ui/components/A11y.tsx
+++ b/packages/tldraw/src/lib/ui/components/A11y.tsx
@@ -142,6 +142,8 @@ export const useSelectedShapesAnnouncer = () => {
 	useEffect(() => {
 		if (!editor) return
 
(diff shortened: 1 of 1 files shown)


steveruizok:
Note that you need to have at least one non-selected shape in order for the "announce on every frame" bug to occur on main.

PR #6045, opened 2025-05-12
Fix border radius on focused buttons

steveruizok:
The focus styles are setting border radius on elements. Our buttons etc should determine their own radius.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed a bug with focused buttons in the application UI receiving a border radius


steveruizok:
@mimecuvalo what's a better way to handle the different focus:visible states for these items? Should I run through every button style and add them there? I have a feeling that the app-wide solution in tla.css is going to be trouble.

mimecuvalo:
> @mimecuvalo what's a better way to handle the different focus:visible states for these items? Should I run through every button style and add them there? I have a feeling that the app-wide solution in tla.css is going to be trouble.

Discussed offline, would be nice to have some kind of mixin for this sort of thing but we don't have a compiled css system that does that sort of thing.

steveruizok:
did things the hard way

PR #6053, opened 2025-05-12
Simplify the cursor logic

MitjaBezensek:
Simplifies the logic from https://github.com/tldraw/tldraw/pull/6023 and also makes it more performant.

### Change type

- [x] `improvement`


mimecuvalo (approved):
👍 thx!

PR #6056, opened 2025-05-12
Fix queue previews

MitjaBezensek:
We need to have a separate queue for each preview deploy.

### Change type

- [x] `bugfix`

### DIFF
--- a/internal/scripts/deploy-dotcom.ts
+++ b/internal/scripts/deploy-dotcom.ts
@@ -255,7 +255,9 @@ async function deployTlsyncWorker({ dryRun }: { dryRun: boolean }) {
 	if (previewId && !didUpdateTlsyncWorker) {
 		await setWranglerPreviewConfig(worker, { name: workerId })
 		didUpdateTlsyncWorker = true
-		await exec('yarn', ['wrangler', 'queues', 'create', `tldraw-multiplayer-queue-${previewId}`])
+		await exec('yarn', ['wrangler', 'queues', 'create', `tldraw-multiplayer-queue-${previewId}`], {
+			pwd: worker,
+		})
 	}
 	await exec('yarn', ['workspace', '@tldraw/zero-cache', 'migrate', dryRun ? '--dry-run' : null], {
 		env: {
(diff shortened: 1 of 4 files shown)


internal/scripts/prune-preview-deploys.ts
```diff
@@ -83,6 +83,18 @@ async function deletePreviewWorkerDeployment(id: string) {
...
+
+	const queueUrl = `https://api.cloudflare.com/client/v4/accounts/${env.CLOUDFLARE_ACCOUNT_ID}/queues/tldraw-multiplayer-queue-${id}`
+	const queueRes = await fetch(queueUrl, {
+		method: 'DELETE',
+		headers: {
+			Authorization: `Bearer ${env.CLOUDFLARE_API_TOKEN}`,
+			'Content-Type': 'application/json',
+		},
+	})
+	if (!res.ok) {
```
MitjaBezensek, line 95: We don't want to error out here as some old PRs don't have queues associated with them.

internal/scripts/deploy-dotcom.ts
```diff
@@ -253,8 +253,14 @@ let didUpdateTlsyncWorker = false
...
 	const workerId = `${previewId ?? env.TLDRAW_ENV}-tldraw-multiplayer`
 	if (previewId && !didUpdateTlsyncWorker) {
-		await setWranglerPreviewConfig(worker, { name: workerId })
+		const queueName = `tldraw-multiplayer-queue-${previewId}`
+		await setWranglerPreviewConfig(worker, { name: workerId }, queueName)
 		didUpdateTlsyncWorker = true
+		try {
+			await exec('yarn', ['wrangler', 'queues', 'info', queueName], { pwd: worker })
+		} catch (_e) {
+			await exec('yarn', ['wrangler', 'queues', 'create', queueName], { pwd: worker })
```
MitjaBezensek, line 262: This is not idempotent, so we first try to list it, then we create it if that fails. Also considered deleting then creating again, but cloudflare doesn't allow it since the queue might have active consumers.

PR #6064, opened 2025-05-13
Only create a queue if not a dry run.

MitjaBezensek:
Looks like the queue creation does not support a dry run flag, so let's just omit creating it for a dry run.

### Change type

- [x] `improvement`


ds300 (approved):
good catch!

PR #6066, opened 2025-05-13
Associate users with file uploads.

MitjaBezensek:
If a logged in user uploads a file we also add their user id to the assets table.

### Change type

- [x] `improvement`


apps/dotcom/sync-worker/src/routes/tla/uploads.ts
```diff
@@ -2,9 +2,12 @@ import { handleUserAssetUpload } from '@tldraw/worker-shared'
 import { IRequest } from 'itty-router'
 import { createPostgresConnectionPool } from '../../postgres'
 import { Environment } from '../../types'
+import { getAuth } from '../../utils/tla/getAuth'
 
 export async function upload(request: IRequest, env: Environment): Promise<Response> {
 	const { body, url, headers } = request
+	const auth = await getAuth(request, env)
```
MitjaBezensek, line 9: We get the auth from cookies - didn't need to change anything on the frontend as these get automatically passed along.

ds300 (approved):

PR #6070, opened 2025-05-13
Configurable VideoShapeUtil autoplay

elstob:
This pull request is a second attempt to resolve https://github.com/tldraw/tldraw/issues/5716.

It was mentioned during a previous attempt (https://github.com/tldraw/tldraw/pull/6016) by @mimecuvalo that the preferred approach to defining behaviour like this was to allow for configuration of the shape util in question rather than a global setting. This PR allows for `VideoShapeUtil` to be configured using the `defaultAutoplay` boolean property.

Some things to note:

- This parameter sets the `playing` prop based on the value of `defaultAutoplay` and then uses that prop in the rendering of the shape. This choice was made because a) that prop is currently just hard coded to `true` and not actually used anywhere and b) Adding an `autoPlay` prop would introduce additional complexity in additional code requirements such as migrations.
- I hope this is sufficient to allow this behaviour to be introduced because as noted in the closed PR - the current situation of multiple videos all autoplaying at once is pretty egregious, so please use this as a starting point for further improvements that may allow this PR to go in.

Thanks in advance.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Note all videos currently added to a canvas autoplay
2. Configure your canvas to use a configured `VideoShapeUtil` e.g. `VideoShapeUtil.configure({ defaultAutoplay: true }`
3. Not that videos now do not autoplay.

### Release notes

- Allows for `VideoShapeUtil` autoplay behaviour to be configured using the `defaultAutoplay` boolean property.


steveruizok:
Good PR @elstob! This makes sense to me, let's go.

steveruizok:
Wow, I never noticed how half-implemented the playing property is. At one time we synced the playing state across users but that was broken; we removed that behavior but didn't clean up the properties.

Anyway, I'm going to tweak the name to just autoplay, since it's still not configurable.

steveruizok:
It would be good to follow this up with a PR that allows the video to be controlled by its `playing` and `time` prop, even if we're not updating them when the video plays or pauses, so that a developer could control the start time of a video or start it playing programatically.

PR #6074, opened 2025-05-14
DRAFT Buda/eng 3384 dont open links when the shift key is pressed

budatl:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


packages/tldraw/src/lib/shapes/shared/HyperlinkButton.tsx
```diff
@@ -19,6 +26,12 @@ export function HyperlinkButton({ url }: { url: string }) {
 			onPointerUp={stopEventPropagation}
 			title={url}
 			draggable={false}
+			onClick={(e) => {
+				if (selectToolActive && editor.inputs.shiftKey) {
+					e.preventDefault()
+					e.stopPropagation()
```
steveruizok, line 32: We have our own stopEventPropagation and preventDefault helpers that let us track these event kills. (They're really hard to debug otherwise).

packages/editor/src/lib/hooks/useReactiveKeyPressed.ts
```diff
@@ -0,0 +1,57 @@
...
+ *       Visit
+ *     </a>
+ *   )
+ * }
+ * ```
+ *
+ * @param keys - Array of key names to track (e.g. `['Shift', 'Meta', 'Escape', 'a']`).
+ * @returns A boolean indicating whether any of the specified keys are currently pressed.
+ *
+ * @public
```
ds300, line 34:
```suggestion
 * @internal
```
@SomeHats had a PR a couple of months ago to make Editor.inputs reactive, and I think I'd much rather see us resurrect that idea  than expose a new hook specifically for keys. I think the PR was more or less done ? https://github.com/tldraw/tldraw/pull/5721

no need to bring that back for this PR, but maybe we keep the hook as internal until we have reactive inputs?

packages/editor/src/lib/hooks/useReactiveKeyPressed.ts
```diff
@@ -0,0 +1,57 @@
...
+ */
+export const pressedKeysSignal = atom('pressedKeys', {} as { [key: string]: boolean })
+
+export function useReactiveKeyPressed(keys: string[]) {
```
ds300, line 38: you could make this strongly typed as well

```ts
export function useIsKeyPressed<K extends string>(keys: K[]): {[k in K]: boolean } {
```

SomeHats:
I'm pretty reluctant to add another input-management helper like this just for this one-off use case. I'd rather we either:
1. Remove `useIsKeyPressed` and just write a more standard react `useEffect` + `useState` just for this one component
2. Revive #5721

The reason we closed out #5721 was mostly because i changed my approach for something in elbow arrows, so we no longer needed it. Otherwise though, apart from writing release notes, it's pretty much ready to go: it introduces the new `getX()` helpers, and keeps the old properties around as deprecated getters.

I don't wanna blow up the scope of this change, but @budatl how do you feel about maybe reviving the reacting inputs PR instead? Otherwise, I'd rather we stick to a spot-change rather than introducing a new utility that does the same thing as the existing `editor.inputs` API.

steveruizok:
We're overthinking this, it should be an easy fix https://github.com/tldraw/tldraw/pull/6136

Reactive inputs is a good separate idea, we should do that anyway if the UI or components are depending on inputs.

PR #6075, opened 2025-05-15
Consolidate mutation logic

ds300:
# Consolidate Mutation Logic

This PR makes it so that we don't have separate mutation and authorization logic for zero. Making this possible involved cleaning up and simplifying a lot of other parts of the codebase. I'm gonna walk though the changes here:

1. **Simplifying the ZStoreData type**

   https://github.com/tldraw/tldraw/blob/c0d0f8a755fdff0e9f2f33617dc45109d9e8f400/packages/dotcom-shared/src/types.ts#L78-L90
  
   Previously this type used camelCase versions of our table names, and also for the 'user' table it used a single nullable row rather than an array of rows. This required a lot of special casing in the code. I changed this to use snake_case for the table names, and made the user table an array of rows. This means that we can now use the same logic for all tables, and we don't need as much glue code.

   Making this change naively would break older clients, so I added a down-migration step when sending the 'initial_data' message.

   https://github.com/tldraw/tldraw/blob/c0d0f8a755fdff0e9f2f33617dc45109d9e8f400/apps/dotcom/sync-worker/src/TLUserDurableObject.ts#L191-L201

2. **Simplifying the client CRUD/Query logic, making them disposable**

   As mentioned above, having a simpler ZStoreData type means that we can use the same logic for all tables so I deleted a lot of the boilerplate code while moving the Client CRUD/Query logic into their own classes [ClientCRUD](https://github.com/tldraw/tldraw/blob/c0d0f8a755fdff0e9f2f33617dc45109d9e8f400/apps/dotcom/client/src/tla/app/ClientCRUD.ts#L5) and [ClientQuery](https://github.com/tldraw/tldraw/blob/c0d0f8a755fdff0e9f2f33617dc45109d9e8f400/apps/dotcom/client/src/tla/app/ClientQuery.ts#L4).

   These classes get instantiated per-mutation, and receive an abort signal so we can throw errors if they are used after the mutation has 'finished' and the crud tracking has stopped.

3. **Replacing _doMutate ad-hoc CRUD logic with a ZQL-style CRUD class backed by postgres**

   To eventually use zero-style mutators in our custom backend we need to implement a CRUD class that points to postgres. 
   
   https://github.com/tldraw/tldraw/blob/a1ea48518d89627f522b06a31d6f8d5cc0986ccc/apps/dotcom/sync-worker/src/zero/ServerCrud.ts#L9

   We can also use this to handle legacy 'update'-style mutations from older clients during a transition period. 

   https://github.com/tldraw/tldraw/blob/d043ca6e8bd2353f59ea4dac1980bebd42f66807/apps/dotcom/sync-worker/src/TLUserDurableObject.ts#L362-L366

   This meant moving one bit of ad-hoc validation logic that was in _doMutate into the `assertValidMutation` method. Namely, checking that MAX_FILES hasn't been exceeded. And `assertValidMutation` itself was moved to a separate file and marked as legacy because it's only needed for the legacy update-style mutations.

4. **Sending 'mutator'-style mutations to the server**

   In order to let us eventually consolidate our authorization logic, we need to be sending 'mutator'-style mutations to the server instead of 'update'-style mutations. Then our custom mutators will be entirely responsible for auth checks.

   https://github.com/tldraw/tldraw/blob/d043ca6e8bd2353f59ea4dac1980bebd42f66807/packages/dotcom-shared/src/types.ts#L157-L168

   In order to handle the mutators on the server, we need to implement an polyfill of the zql query interface on the server backed by postgres.

   https://github.com/tldraw/tldraw/blob/d043ca6e8bd2353f59ea4dac1980bebd42f66807/apps/dotcom/sync-worker/src/zero/ServerQuery.ts#L5

## Testing

I've manually tested the downgrading by running an older version of the client against this version of the backend.

e2e tests are all passing.

### Change type

- [x] `other`

### DIFF
--- a/packages/dotcom-shared/src/OptimisticAppStore.ts
+++ b/packages/dotcom-shared/src/OptimisticAppStore.ts
@@ -5,9 +5,7 @@ import { schema } from './tlaSchema'
 import { ZRowUpdate, ZStoreData } from './types'
 
 export class OptimisticAppStore {
-	private _gold_store = atom('zero store', null as null | ZStoreData, {
-		isEqual: isEqual,
-	})
+	private _gold_store = atom('zero store', null as null | ZStoreData, { isEqual })
 
 	private _optimisticStore = atom<
 		Array<{
@@ -37,18 +35,22 @@ export class OptimisticAppStore {
 		})
 	}
 
-	private store = computed('store', () => {
-		const gold = this._gold_store.get()
-		if (!gold) return null
-		let data = gold
-		const optimistic = this._optimisticStore.get()
-		for (const changes of optimistic) {
-			for (const update of changes.updates) {
-				data = this.applyUpdate(data, update)
+	private store = computed(
+		'store',
+		() => {
+			const gold = this._gold_store.get()
+			if (!gold) return null
+			let data = gold
+			const optimistic = this._optimisticStore.get()
+			for (const changes of optimistic) {
+				for (const update of changes.updates) {
+					data = this.applyUpdate(data, update)
+				}
 			}
-		}
-		return data
-	})
+			return data
+		},
+		{ isEqual }
+	)
 
 	getCommittedData() {
 		return this._gold_store.get()
--- a/packages/dotcom-shared/src/OptimisticAppStore.ts
+++ b/packages/dotcom-shared/src/OptimisticAppStore.ts
@@ -104,6 +104,7 @@ export class OptimisticAppStore {
 		const { row, table, event } = update
 		const tableSchema = assertExists(schema.tables[table], 'table not found')
 		const rows = prev[table]
+		assert(rows, 'table not found in store ' + table)
 		const matchExisting = (existing: any) =>
 			tableSchema.primaryKey.every((key) => existing[key] === (row as any)[key])
 
--- a/packages/dotcom-shared/src/OptimisticAppStore.ts
+++ b/packages/dotcom-shared/src/OptimisticAppStore.ts
@@ -1,5 +1,5 @@
 import { atom, computed, react, transact } from '@tldraw/state'
-import { assert, assertExists } from '@tldraw/utils'
+import { assert, assertExists, exhaustiveSwitchError } from '@tldraw/utils'
 import isEqual from 'lodash.isequal'
 import { schema } from './tlaSchema'
 import { ZRowUpdate, ZStoreData } from './types'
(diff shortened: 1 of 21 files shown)


packages/state/src/lib/transactions.ts
```diff
@@ -300,3 +301,17 @@ export function transact<T>(fn: () => T): T {
...
+export function pauseReactions() {
+	assert(!inst.currentTransaction, 'Cannot pause reactions inside a transaction')
+	const txn = (inst.currentTransaction = new Transaction(null))
+
+	return () => {
+		assert(txn === inst.currentTransaction, 'Cannot resume reactions while in a transaction')
+		inst.currentTransaction = null
+		txn.commit()
+	}
+}
```
ds300, line 317: This was allows us to prevent rerendering the app in the middle of a mutator, which might lead to inconsistent/buggy states.

MitjaBezensek (commented):
Some feedback below.

packages/dotcom-shared/src/OptimisticAppStore.ts
```diff
@@ -107,59 +102,34 @@ export class OptimisticAppStore {
...
+					[table]: rows.map((existing) => {
+						if (matchExisting(existing)) {
+							return { ...existing, ...row }
+						}
+						return existing
+					}),
 				}
-			} else {
-				assert(event === 'insert', 'invalid event')
+			case 'delete':
```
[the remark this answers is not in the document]
ds300: yes however the place to prevent user row deletions is really in the mutation/mutator layer. Investigating just now I noticed we weren't preventing user row deletions for legacy update-style mutations (😰) so have added a commit to this PR for that. Also prevented users from hard deleting files through the legacy mutations.

packages/state/src/lib/transactions.ts
ds300: good catch! i'm gonna remove this for now and add a follow up ticket to do it properly later. The place where I was seeing issues that required this (adding a new file not triggering the input to be focused) was really caused by something else (bad immutable data handling causing too many rerenders and then us relying on that happening) which I solved separately. So it should be fine to add this later.

ds300:
@MitjaBezensek thanks for your excellent feedback! I've incorporated it and this is ready for review again.

MitjaBezensek (approved):
Nice 👏

PR #6086, opened 2025-05-18
Minor perf tweak to notVisibleShapes

steveruizok:
This PR drops the check for masked page bounds when analyzing notVisibleShapes. This isn't really needed and is one more computed for the system to check on a relatively hot path (e.g. getting the hovered shape on a crowded canvas).

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Slight performance improvement of complex boards.

### DIFF
--- a/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
+++ b/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
@@ -1,49 +1,48 @@
 import { computed, isUninitialized } from '@tldraw/state'
 import { TLShapeId } from '@tldraw/tlschema'
-import { Box } from '../../primitives/Box'
 import { Editor } from '../Editor'
 
-function isShapeNotVisible(editor: Editor, id: TLShapeId, viewportPageBounds: Box): boolean {
-	const maskedPageBounds = editor.getShapeMaskedPageBounds(id)
-	// if the shape is fully outside of its parent's clipping bounds...
-	if (maskedPageBounds === undefined) return true
-
-	// if the shape is fully outside of the viewport page bounds...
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/primitives/Box.ts
```diff
@@ -418,7 +418,7 @@ export class Box {
 	}
 
 	static Includes(A: Box, B: Box) {
-		return Box.Collides(A, B) || Box.Contains(A, B)
+		return Box.Collides(A, B)
```
steveruizok, line 421: Logically equal

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -1,49 +1,48 @@
...
-		shapes.forEach((id) => {
-			if (isShapeNotVisible(editor, id, viewportPageBounds)) {
-				notVisibleShapes.add(id)
-			}
-		})
-		return notVisibleShapes
-	}
-	return computed<Set<TLShapeId>>('notVisibleShapes', (prevValue) => {
+export function notVisibleShapes(editor: Editor) {
+	return computed<Set<TLShapeId>>('notVisibleShapes', function updateNotVisibleShapes(prevValue) {
```
steveruizok, line 28: fwiw this implementation seems like it was designed for something else. This is also nested in a different computed in the Editor

MitjaBezensek: I think this might be some leftovers from when we used rbush.

MitjaBezensek (approved):

PR #6098, opened 2025-05-20
analytics: tweak vercel.json routing, take 2

mimecuvalo:
bah bah

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/analytics/vercel.json
```diff
@@ -16,6 +12,10 @@
 				{ "key": "Access-Control-Allow-Methods", "value": "GET" },
 				{ "key": "Cache-Control", "value": "public, max-age=31536000, immutable" }
```
SomeHats, line 13: this cache-control header seems really aggressive for an ingestion endpoint... is this definitely what we want?

PR #6100, opened 2025-05-20
Compile fetchEverythingSql ahead of time

ds300:
This PR adds ahead-of-time compilation for our big SQL query that the user data syncer uses to boot.

The aim is to make it easier and less error prone to add new tables, and to minimize the number of nulls in the result set (which result from UNION-ing many tables together).

This is achieved by:

2. Compiling the sql query from a minimal description of the tables/queries to fetch.
3. Reusing columns in the result set, so they can contain data from multiple tables.

### Change type

- [x] `other`


MitjaBezensek (approved):
👍

PR #6108, opened 2025-05-21
Revert "scale scrubbing"

SomeHats:
Reverts tldraw/tldraw#6057


TodePond:
why 🥲

SomeHats:
https://github.com/user-attachments/assets/e5f68680-7863-48e6-b2ec-4e11d17c4484

it just feels so random & underbaked for dotcom and we're about to do a release. 

* Makes it really easy to get broken-looking documents (see above)
* As an interaction, it feels a long way from tldraw-quality to me - different shapes respond to it in different ways which is surprising, the way it just goes forever is very weird.
* Super low-priority feature that afaik has never been asked for

PR #6112, opened 2025-05-21
Use PathBuilder for geo & line shapes

SomeHats:
Before #5572, we had two geometric style shapes that required inky rendering: lines and geo shapes. Both of these shapes implemented their own inky algorithm, plus approaches for turning into dashed and dotted lines too. For geo shapes, this was particularly bad - many geo shapes needed their own completely custom implementations for inky & dashed lines, as well as for geometry.

In #5572, we introduced a third: elbow arrows (straight and arc arrows don't get draw-style rendering currently). To facilitate this, that PR introduced the foundations of a new abstraction for working with paths in tldraw: `PathBuilder`. `PathBuilder` has an API similar to that of the webs [`Path2D`](https://developer.mozilla.org/en-US/docs/Web/API/Path2D). It allowed the construction of a path using familiar, canvas-like commands, that could then be converted into `Geometry2d` or SVG - including in inky and dashed styles. The plan when I introduced that API was always to use it for line and geo shapes too, but that was out of scope of the initial elbow arrows PR.

This PR revamps `PathBuilder` to suit the needs of lines and geo shapes, and replaces all the many custom implementations of draw and dash style rendering with the single approach defined in `PathBuilder`. Among other things, this means it's now much easier to add new geo shapes (just add the definition in `getGeoShapePath` and the geometry / rendering will be handled automatically) & much easier for developers to create shapes that use tldraws draw/dash styles (define the path using `PathBuilder`, then call `toSvg` with the appropriate methods).

There are some extremely subtle differences in rendering as a result of this PR as there were subtle differences between many of the previous algorithms.

### Change type

- [x] `api`

### Release notes

- Introduce `PathBuilder`, a new API to make constructing tldraw-style inky and dashed SVGs much easier.


ds300:
There's a fairly jarring downgrade for cloud shapes with dashed lines 
image

Dashes seem fine for the other shapes.

packages/tldraw/src/lib/shapes/shared/PathBuilder.tsx
```diff
@@ -39,143 +42,399 @@ export interface DashedPathBuilderOpts extends BasePathBuilderOpts {
...
+		largeArcFlag: boolean,
+		sweepFlag: boolean,
+		xAxisRotationRadians: number,
+		x2: number,
+		y2: number,
+		opts?: PathBuilderCommandOpts
+	) {
+		// As arc flags make them very sensitive to offsets when we render them in draw mode, we
+		// approximate arcs by converting them to up to 4 (1 per 90° segment) cubic bezier curves.
+		// This is alogirthm is a Claude special:
```
mimecuvalo, line 225: nit: just a typo

mimecuvalo (commented):
I think in general it's looking good! 
- any performance concerns?
- when you say "some extremely subtle differences in rendering as a result of this PR as there were subtle differences between many of the previous algorithms." - can you elaborate on that? that would help me understand what I should review more rigorously
- David has a point on the dashed-Cloud shape - why did that not affect e2e screenshot tests? (the geo shapes one)

SomeHats:
> * any performance concerns?

The main one is that using beziers to approximate arcs results in a few more vertices being generated for each arc that we need to check against, but i'm not worried about it. it's all in JS land, and generally JS is fast and it's the dom/rendering we need to worry about

> * when you say "some extremely subtle differences in rendering as a result of this PR as there were subtle differences between many of the previous algorithms." - can you elaborate on that? that would help me understand what I should review more rigorously

Honestly i don't think you need to focus anywhere. I think that the clouds example was one of the possible changes, but i hadn't noticed it as it doesn't always happen. I've fixed it now. Others are things like using beziers to approximate arcs aren't 100% right, but the margin for error is low enough to be inperceptible

> * David has a point on the dashed-Cloud shape - why did that not affect e2e screenshot tests? (the geo shapes one)

As far as I can tell, it's because the dash issue didn't show on all cloud shapes, and otherwise the rendering differences are within the margin of error

SomeHats:
Looking a bit more closely, I think our margin of error for snapshots was a bit high - i've lowered it

steveruizok:
This seems like an sure improvement, though at the same time I remember the geometry shapes requiring a lot of special casing in order to make them look right, in particular with dashes.

Before:
image

After:
image

SomeHats:
I've ported across most of the special casing, but from your before/after i've noticed a little bug which i'm going to fix. thanks!

PR #6116, opened 2025-05-22
[signia] add async effect deferral

ds300:
Basic support for async transactions in signia. This is somewhat limited in that these aren't transactions because state can leak between them. Really it's just a way to defer effects until the end of some async logic.

If a number of calls to `deferAsyncEffects` overlap in time:
- effects will not run until the last one completes
- if any of them throws, the whole lot of them will be reverted. This is unfortunate, but it's better than leaving things in a broken state i think.

Let's keep this internal for now, not 100% sure whether there is need in the community.

### Change type

- [x] `other`


SomeHats:
so this suggestion is only half thought out, but i'm worried that the nesting constraint currently is a bit foot-gunny. Imagine I have a UI, with two buttons. Both buttons trigger an async transaction. If I press the first button then the second, these transactions aren't really "nested" - they're distinct operations. Right now, assuming these both take the same time, I'd get an error because the transactions would be interpreted as being nested, but the second would outlive the first.

It might be that this is fine and we can accept this, but I wonder if it might be better to:
1. Not allow async transactions at all by queuing each call to `asyncTransaction`
2. Delaying the end of of the parent transaction until the child transaction is complete

Appreciate that none of these are good options and for now maybe just sticking with the footgun-y one is the way forward. It's a shame the [async context](https://github.com/tc39/proposal-async-context) api isn't around yet i feel like we could build something really good on that

ds300:
> queuing each call to asyncTransaction

hm nested transactions would result in deadlocks in this case, unless there's one queue per call site or something like that.

> Delaying the end of of the parent transaction until the child transaction is complete

yeah this is the approach I'd want to take, and I gave it a little go but it got hairy and I wanted to ship this sooner rather than later. 

But you've convinced me the potential for danger here is greater than I imagined so I'll give it another go and see if I can figure something out.

ds300:
@SomeHats I figured something better out.

Calling this 'asyncTransaction' was misleading because state changes weren't isolated so I changed the name to `deferAsyncEffects` and made it work more like `transact` in that it doesn't nest per se, rather it just coalesces changes, while still allowing to roll back if something throws at any point.

packages/state/src/lib/transactions.ts
```diff
@@ -300,3 +310,61 @@ export function transact<T>(fn: () => T): T {
...
+	// is done.
+	if (inst.currentTransaction?.isSync) {
+		throw new Error('deferAsyncEffects cannot be called during a sync transaction')
+	}
+
+	// Can't kick off async transactions during a reaction phase at the moment,
+	// because the transaction stack is cleared after the reaction phase.
+	// So wait until the path ahead is clear
+	while (inst.globalIsReacting) {
+		await sleep(0)
```
SomeHats, line 329: tiniest of nits but maybe a microtask would be more appropriate than a timeout here? i could also see this just throwing and forcing the callee to handle this, the initial call of the callback being async in some conditions could be surprising

SomeHats (approved):

PR #6131, opened 2025-05-26
Use hand tool as default on mobile

MitjaBezensek:
After landing on a document on mobile it's really easy to move things when you want to move around the document. Using hand as the default on mobile might be better.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
@@ -26,6 +26,7 @@ import { multiplayerAssetStore } from '../../../utils/multiplayerAssetStore'
 import { useMaybeApp } from '../../hooks/useAppState'
 import { ReadyWrapper, useSetIsReady } from '../../hooks/useIsReady'
 import { useTldrawUser } from '../../hooks/useUser'
+import { getIsCoarsePointer } from '../../utils/getIsCoarsePointer'
 import { maybeSlurp } from '../../utils/slurping'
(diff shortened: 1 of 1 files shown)


MitjaBezensek:
> +1 to this, since mobile is more about consumption than creation, but i'd be curious if there were counter-arguments to this!

Yeah, we saw the problem here at local first conf. There was a document that everyone used to collaborate on the topics to discuss. Quite a few people mentioned that they accidentally moving things when they tried to move around the document.

steveruizok:
> +1 to this, since mobile is more about consumption than creation, but i'd be curious if there were counter-arguments to this!

I'd say it's less about consumption vs. creation and more about the risk of moving things accidentally while trying to first orient yourself on a new project, as Mitja said. First experience with the app shouldn't be "oops I moved the thing they were showing me".

There may be a related change where a user lands on an empty part of a project when they get a mobile link. We could try to zoom to fit if it's 1) the first time the user has visited a page, and 2) there is content on the page. On large projects this might also be a problem, perhaps a "navigate to content" would be better.

Related: a lot of apps use a different interaction system completely on mobile that combines selection and navigation. I've always wanted to try implementing that but I also think it overcompensates in favor of for first time users (making "actual use" impossibly slow) and might not be that good for first time users either.

steveruizok:
I'm going to slide in here though because I think the logic is more complex than coarse pointer = hand. I'd like the hand to be the default if they're on a page that has content but keep select as the default tool if they're on an empty page.

steveruizok (approved):

PR #6136, opened 2025-05-27
Fix shift key opening link

steveruizok:
This PR fixes a bug where holding the shift key while clicking a link would open the link. (This is often accidental while adding a shape to the user's selection).

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a bookmark shape
2. Hold shift and click the hyperlink icon

### Release notes

- Fixed a bug with opening bookmark hyperlinks accidentally while holding shift

### DIFF
--- a/packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
@@ -16,10 +16,11 @@ import {
 	stopEventPropagation,
 	tlenv,
 	toDomPrecision,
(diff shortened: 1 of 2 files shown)


mimecuvalo:
I feel like I'm so confused on how this works for some reason:
`if (!editor.inputs.shiftKey) stopEventPropagation(e)`
so, if the shift key isn't pressed we stop propagation? shouldn't it be the opposite? on staging it seems to work, maybe it's late in the day and I'm not seeing it, but how does this work??

Issue #5930, opened 2025-04-18
[Bug]: Zoom Behavior Is Off


seflless:
### What happened?

I'm not sure what it's internally called. But when you press the **Z** key, Tldraw goes into a mode where you can click drag a region and it will zoom to that. Let's called "region-zoom" for now. 

A few things that are off:
1. You can press z, but it's not included in the keyboards dialog
2. When you first use it, it stays in a mode where click dragging zooms in
3. Pressing escape doesn't return you to select mode (like other tools do)
4. Once you've used it once, manually switched back to select mode, it will now require z being held when dragging to zoom to selection. It should either be a mode or a modifier key based approach, not both.

### How can we reproduce the bug?

Here are the repro steps for each sub issue. Note: I'd guess it's not suppose to be a mode, but a modifier key based approach.

**The keyboard shortcut isn't documented**
1. Open the keyboard shortcut menu
2. Note how there is no mention of pressing/holding z to zoom to region

<img width="70%" src="https://github.com/user-attachments/assets/4db47c1c-e88f-408d-909f-877d92bb93c2"/>

**You can't use escape key, only manually leave region-zoom mode**
1. Start from a freshly loaded tldraw.com page with some shapes on it
2. Tap z to go to region-zoom tool/mode
3. Press escape
4. You'll not go back to default/select mode/tool

**It acts as a mode AND as a modifier key**
1. Start from a freshly loaded tldraw.com page with some shapes on it
2. Tap z. You can now repeatedly click and drag to zoom in over and over
3. Click the select tool button in the toolbar at bottom
4. Tap z and you'll see it doesn't permanently go into a mode where you can repeatedly zoom on click dragging
5. From this point on you must hold z and click drag to zoom to region.

Hope that makes sense. Have to rush to work at the moment and my video recording wasn't picking up shortcut keys to demonstrate this.

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The tldraw.com website

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

SomeHats:
Thanks for filing this @seflless! We'll investigate.

mimecuvalo:
So, this is an interesting one and it's uncovered something deeper at play here.
This is a regression since this PR: https://github.com/tldraw/tldraw/pull/4481/files#r1752235094
Specifically, in that comment that's referenced in the link.

The issue is as such:
- we don't proactively `focus()` the container anymore since that PR.
[...]

PR #6140, opened 2025-05-28
[BUG] Complete region zoom when editor loses focus

budatl:
This PR fixes inconsistencies with the region-zoom tool (Z key), which was behaving as both a mode and a modifier. It simplifies the interaction and aligns it with other tools.

**Changes**
- Added Z shortcut to the keyboard help dialog
- When editor loses focus while Z is still pressed, zoom completes and reverts back to selector mode.

Notes
- Resolves unexpected behavior when switching in and out of region-zoom
- Improves UX consistency across tools

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- Go to editor
- Press Z to enter zoom mode
- Zoom repeatedly by dragging
- Manually click away from editor to lose focus (for example the "Page 1" dropdown)
- Editor should reset to Selector mode 

Open the keyboard shortcut dialog; Z is missing
- [ ] Unit tests
- [ ] End to end tests

### Release notes
Complete region zoom when editor loses focus.


SomeHats (changes requested):
Couple issues here:

1. I just noticed that pressing z tldraw.com actually does exhibit the behaviour described if it's the first thing you press after loading the page. This is still the case in the dotcom preview on this branch. I think this is maybe a wider issue relating to focus management and @mimecuvalo's a11y work. For example, if you interact with the sidebar, then try to spacebar-drag the editor, instead the sidebar will scroll and you'll draw a normal selection instead of panning around.
2. The way you've added an action here means that now pressing z zooms in. I don't think this is the intended behaviour?

Sorry about this - should have thought about / noticed this when we were pairing yesterday.

mimecuvalo:
> Couple issues here:
> 
> 1. I just noticed that pressing z tldraw.com actually does exhibit the behaviour described if it's the first thing you press after loading the page. This is still the case in the dotcom preview on this branch. I think this is maybe a wider issue relating to focus management and @mimecuvalo's a11y work. For example, if you interact with the sidebar, then try to spacebar-drag the editor, instead the sidebar will scroll and you'll draw a normal selection instead of panning around.
> 2. The way you've added an action here means that now pressing z zooms in. I don't think this is the intended behaviour?
> 
> Sorry about this - should have thought about / noticed this when we were pairing yesterday.

As discussed in the related bug: https://github.com/tldraw/tldraw/issues/5930#issuecomment-2919728084
This is an issue due to focus dating back from September. We'll close this PR and look at this more deeply.

PR #6162, opened 2025-05-31
Geometry bug fixes

steveruizok:
This PR fixes a few minor bug fixes with geometry.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed bug where `Vec.uni()` could return a null value on zero length vectors
- Fixed a bug with SVG output of negative-width Rectangles

### DIFF
--- a/apps/examples/src/examples/before-create-update-shape/BeforeCreateUpdateShapeExample.tsx
+++ b/apps/examples/src/examples/before-create-update-shape/BeforeCreateUpdateShapeExample.tsx
@@ -14,7 +14,7 @@ function constrainShapeToRadius(editor: Editor, shape: TLShape, radius: number) 
 
 	// if the shape is outside the radius, move it to the edge of the radius:
 	if (distanceFromCenter > radius) {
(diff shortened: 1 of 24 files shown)


mimecuvalo:
> Fixed bug where Vec.uni() could return a null value on zero length vectors

fwiw, we had this discussion @steveruizok in https://github.com/tldraw/tldraw/pull/2742#discussion_r1479873709
At the time, we decided that the caller should handle the divide by zero issues, it's been a year since that PR so I can't quite recall the nuance but I think the argument was that changing the behavior might change some downstream usage that relied on the way it was at the moment.

steveruizok:
I'm going to give this one a tentative "probably fine". We don't have any code (I don't think) that relies on catching an error with `uni`, though we do have protection against zero length geometries (ie arcs with start and end handles at the same position) since those cause other problems anyway. I agree that it's a risk but I think we'll be okay here.

PR #6165, opened 2025-06-01
Only re-evaluate fonts when props.richText changes

steveruizok:
As feared, the font recognition and caching code that we wrote for rich text has some ugly performance effects, especially when resizing shapes.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a few hundred shapes
2. Resize them

### Release notes

- Improved performance when resizing or editing many shapes.

### DIFF
--- a/packages/editor/src/lib/editor/managers/FontManager.ts
+++ b/packages/editor/src/lib/editor/managers/FontManager.ts
@@ -94,7 +94,11 @@ export class FontManager {
 				const shapeUtil = this.editor.getShapeUtil(shape)
 				return shapeUtil.getFontFaces(shape)
 			},
-			{ areResultsEqual: areArraysShallowEqual, areRecordsEqual: (a, b) => a.props === b.props }
+			{
+				areResultsEqual: areArraysShallowEqual,
+				// @ts-expect-error
+				areRecordsEqual: (a, b) => a.props.richText === b.props.richText,
+			}
 		)
 
 		this.shapeFontLoadStateCache = editor.store.createCache<(FontState | null)[], TLShape>(


packages/editor/src/lib/editor/managers/FontManager.ts
```diff
@@ -94,7 +94,11 @@ export class FontManager {
 				const shapeUtil = this.editor.getShapeUtil(shape)
 				return shapeUtil.getFontFaces(shape)
 			},
-			{ areResultsEqual: areArraysShallowEqual, areRecordsEqual: (a, b) => a.props === b.props }
+			{
+				areResultsEqual: areArraysShallowEqual,
+				// @ts-expect-error
+				areRecordsEqual: (a, b) => a.props.richText === b.props.richText,
```
steveruizok, line 100: Not sure how best to handle this because the editor doesn't know which properties a shape depends on for its `getFontFaces`. I think it's best if we paint over this for now. At worst, this feature won't work unless the shape has a property named rich text.

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -167,6 +168,9 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 	}
 
 	override getFontFaces(shape: TLGeoShape): TLFontFace[] {
+		if (isEmptyRichText(shape.props.richText)) {
+			return EMPTY_ARRAY
+		}
```
steveruizok, line 173: I added guards like this around text measurement (most geo shapes don't have text) but here they're only a partial solution. Parsing even "empty" text is slow because `getFontsFromRichText` is slow because it does a lot of work.

mimecuvalo:
Hm, I'm worried about landing this without Alex around because there some kind of nuance here that is being lost.
The previous PR that touched this: https://github.com/tldraw/tldraw/pull/5743
added this line: `areRecordsEqual: (a, b) => a.props === b.props`

And that makes me worried that @SomeHats didn't make it `a.props.richText` on purpose.

Furthermore, my comment in that other PR becomes even more relevant because we're redoing it in this PR:
>  this does make me wonder what the right defaults should be for createComputedCache. it feels footgun-y for us, let alone others, and it makes me wonder if it shouldn't force areResultsEqual/areRecordsEqual to be considered when calling the function (if nothing to just specify explicitly undefined or noop).
> 
> like, for example, editor.store.createComputedCache('handle snap geometry' in HandleSnaps. should this also consider areRecordsEqual? it's unclear to me if it should be added there or whether it's only on a case-by-case basis.

and Alex's response was:
> Yeah, it's a bit complicated and nuanced. I would love if we had an easier way for us to track specifically which bits of the shape a method uses. for handle snaps, it's not really an issue because we don't need it on every frame i guess. not sure what the correct thing is tbh

So, I'm worried that none of us quite _really_ know what's going on here? We should discuss more I think about this system together.

steveruizok:
I dunno, I think we know what's going on. If we don't have that check, then the fonts get re-evaluated too often, but if we add anything more specific that `props` (ie `props.richText`) then there's a type error because `props.richText` isn't a default shape prop and so it doesn't exist at the editor layer.

If for example I had a special shape that used `props.text` for rich text, then the fonts would never be re-evaluated because props.richText would be undefined and undefined would always equal undefined. However, rich text is a very new API and I doubt anyone is using it; we could do further "just for tldraw" guards (i.e. if shape.type === 'geo'...) to protect against this.

This is causing a pretty bad performance hit (even on my Very Fast Laptop), so I'm willing to take a jump here.

steveruizok:
Put another way: if tldraw were not an SDK then this would be an easy decision, we'd want that code to run only when props.richText changed, because that's the only change that would require us to re-evaluate which fonts are used.

I'm also starting to think that this is the Wrong Way to re-evaluate fonts and that an imperative API + side effects would be better. For example, a 'registerFont' method that adds new fonts to an atom, and that I could call from an `shape.onChange` side effect that narrowed the hell out of the circumstances that would require us to do the expensive scan for new fonts. That would give more control to developers to control how / when new fonts are registered, as well as causing new fonts to load for reasons other than shape changes.

steveruizok:
I'm going to land this one, resizing lots of shapes is just too embarrassing on perf to leave in.

PR #6173, opened 2025-06-03
improvement(dotcom): bump max files and add signout route

steveruizok:
This PR bumps the max files from 100 to 200.

This PR also adds a secret /signout route, which I keep reaching for in development (and sometimes have to reset application data instead).

### Change type

- [x] `improvement`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Increased the maximum files for users on tldraw.com

### DIFF
--- a/apps/dotcom/client/src/pages/signout.tsx
+++ b/apps/dotcom/client/src/pages/signout.tsx
@@ -1,9 +1,0 @@
-import { SignOutButton } from '@clerk/clerk-react'
-
-export function Component() {
-	return (
-		<div>
-			<SignOutButton />
-		</div>
-	)
-}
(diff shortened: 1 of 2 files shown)


ds300:
did you hit the limit?

apps/dotcom/client/src/pages/signout.tsx
```diff
@@ -0,0 +1,9 @@
+import { SignOutButton } from '@clerk/clerk-react'
+
+export function Component() {
+	return (
+		<div>
+			<SignOutButton />
+		</div>
+	)
+}
```
ds300: to avoid having to click the button you could also use useClerk and just call `clek.signOut()` immediately then render a redirect to '/'

steveruizok:
> did you hit the limit?

Yeah I did. I actually had started working on a feature to delete all empty projects without names on load, though I see that the `isEmpty` flag never gets set to `false`, so we'd have to actually pull data in order to find out what is actually empty. That might be a more convenient way to keep peoples file count / list tidy.

PR #6184, opened 2025-06-04
Replicator topics graph

ds300:
Sorry for the big PR!

- Replaced the `user_file_subscription` table with a new graph-based topic subscription model. This is much more flexible and allows us to have transitive dependencies like user:steve -> group:tldraw -> file:retro rather than having to worry about managing separate subscriptions for each group file and each user in a group.
- Isolated, refactored, and unit-tested a bunch of the replicator core code, moving much of it into a new `replicator` folder.
- Created a new `ChangeCollator` abstraction for converting postgres changes into effects (sending messages to users, publishing/unpublishing, and notifying file DOs of updates). There are two types of collators: `LiveChangeCollator` and `CatchUpChangeCollator`. The `LiveChangeCollator` is used for live replication, and the `CatchUpChangeCollator` is used while resuming user DO connections.
- Fixed a few bugs.

### Change type

- [x] `other`

### Test plan

- [x] **Create new files** - ensure proper subscription setup for new files
- [x] **Delete files you own** - verify all collaborators are properly unsubscribed
- [x] **Share a file with another user** - verify they receive real-time updates when you edit
- [x] **Join a shared file** - confirm you see live changes from the file owner
- [x] **Leave a shared file** - ensure you stop receiving updates after leaving
- [x] **File permissions changes** - verify subscription updates when sharing settings change
- [x] **File ID validation** - try creating files with invalid characters (commas, special chars)

### **User Session Management**
- [x] **User connects/disconnects** - verify proper subscription registration/cleanup
- [x] **Multiple browser tabs** - test sync across different sessions for same user
- [x] **Network interruption recovery** - disconnect/reconnect and verify catch-up works
- [x] **Cross-device sync** - test same user on different devices

### **Database Replication & Sync**
- [x] **Large file operations** - test with files that have many collaborators
- [x] **Subscription graph traversal** - verify complex sharing hierarchies work correctly
- [x] **Rapid subscription changes** - quickly add/remove file access multiple times
- [x] **Memory usage under load** - monitor subscription graph size with many users
- [x] **Check subscription query performance** - especially the recursive CTE queries


apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -205,6 +205,12 @@ function TlaEditorInner({ fileSlug, deepLinks }: TlaEditorProps) {
 		userInfo: app?.tlUser.userPreferences,
 	})
 
+	// we need to prevent calling onFileExit if the store is in an error state
+	const storeError = useRef(false)
+	if (store.status === 'error') {
+		storeError.current = true
+	}
```
ds300, line 212: this fixes that 'row not found' issue you brought up earlier

.github/actions/setup/action.yml
```diff
@@ -13,7 +13,7 @@ runs:
     - name: Setup Node.js Environment
       uses: actions/setup-node@v3
       with:
-        node-version: 20.11.0
+        node-version: 22.16.0
```
ds300, line 16: had to update node to be able to use the builtin `node:sqlite` api for the replicator unit tests.

apps/docs/scripts/lib/generateSection.ts
```diff
@@ -305,7 +305,9 @@ function getComponentCodeFiles({
 			)
 			// For each of these component files, read the file and add it to the componentCodeFiles object
 			.forEach((file) => {
-				componentCodeFiles[file.name] = fs.readFileSync(path.join(file.path, file.name)).toString()
+				componentCodeFiles[file.name] = fs
+					.readFileSync(path.join(file.parentPath, file.name))
+					.toString()
```
ds300, line 310: this was due to the node upgrade, we were using a deprecated `dirent.path` property, it was renamed to `dirent.parentPath` a while ago.

apps/dotcom/asset-upload-worker/package.json
```diff
@@ -9,7 +9,7 @@
 	},
 	"main": "src/index.ts",
 	"scripts": {
-		"dev": "yarn run -T tsx ../../../internal/scripts/workers/dev.ts --persist-to tmp-assets",
+		"dev": "yarn run -T tsx ../../../internal/scripts/workers/dev.ts --persist-to tmp-assets --inspector-port 9449",
```
ds300, line 12: this change helps us use debuggers with wrangler.

The problem is that we have multiple wrangler dev scripts running, and they each try to use port 9229 by default. So when you attach to the wrangler process, you get whichever one grabbed the port first, which is only rarely the sync-worker. 

So I've assigned the less important workers other ports and allowed sync-worker to use 9229 by default so that the standard wrangler debugging profile Just Works.

apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
```diff
@@ -363,16 +247,22 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
...
+		const numTopicSubscriptions = this.sqlite.exec('select count(*) from topic_subscription').one()
+			.count as number
 		const activeUsers = this.sqlite.exec('select count(*) from active_user').one().count as number
 		const meta = this.sqlite.exec('select * from meta').one()
 		return {
 			earliestHistoryRow,
 			latestHistoryRow,
+			historyDurationMinutes: (latestHistoryRow.timestamp - earliestHistoryRow.timestamp) / 60_000,
+			numHistoryRows,
+			numTopicSubscriptions,
```
ds300, line 260: this is just some extra diagnostics that will be helpful to have on the secret admin dashboard.

apps/dotcom/sync-worker/src/UserDataSyncer.ts
```diff
@@ -464,6 +464,18 @@ export class UserDataSyncer {
...
+		const topicUpdates = getSubscriptionChanges(
+			event.changes
+				.filter((c): c is ZRowUpdateEvent => c.type === 'row_update')
+				.map((ev) => ({ row: ev.row, event: { command: ev.event, table: ev.table } }))
+		)
+
+		if (topicUpdates.newSubscriptions?.length || topicUpdates.removedSubscriptions?.length) {
+			this.reboot({ hard: true, delay: false, source: 'handleReplicationEvent(hard reboot)' })
+			return
+		}
```
ds300, line 477: I deleted some 'fixup' code below that was attempting to detect invalid states due to subscription changes so it could add and remove file rows. but subscriptions change rarely happen (even with upcoming groups) so I figure it's fine to just do a hard reboot in those situations. We can optimise this later by handing the subscription events individually if need be, but I don't expect we'll need to.

packages/dotcom-shared/src/mutators.ts
```diff
@@ -56,25 +56,30 @@ export function createMutators(userId: string) {
 			},
 		},
 		file: {
-			insert: async (tx, file: TlaFile) => {
-				assert(file.ownerId === userId, ZErrorCode.forbidden)
-				await assertNotMaxFiles(tx, userId)
-
-				await tx.mutate.file.insert(file)
-			},
```
ds300, line 64: we weren't using this, so I removed it for now

packages/dotcom-shared/src/mutators.ts
```diff
@@ -56,25 +56,30 @@ export function createMutators(userId: string) {
...
 				tx,
 				{ file, fileState }: { file: TlaFile; fileState: TlaFileState }
 			) => {
 				assert(file.ownerId === userId, ZErrorCode.forbidden)
 				await assertNotMaxFiles(tx, userId)
+				assert(file.id.match(/^[a-zA-Z0-9_-]+$/), ZErrorCode.bad_request)
+				assert(file.id.length <= 32, ZErrorCode.bad_request)
+				assert(file.id.length >= 16, ZErrorCode.bad_request)
+				assert(file.id === fileState.fileId, ZErrorCode.bad_request)
+				assert(fileState.userId === userId, ZErrorCode.forbidden)
```
ds300, line 69: since this pr relies on file ids not having commas, I added these extra checks to ensure it.

packages/dotcom-shared/src/mutators.ts
```diff
@@ -56,25 +56,30 @@ export function createMutators(userId: string) {
...
 			},
 			deleteOrForget: async (tx, file: TlaFile) => {
 				await tx.mutate.file_state.delete({ fileId: file.id, userId })
 				if (file?.ownerId === userId) {
+					if (tx.location === 'server') {
+						// todo: use a sql trigger for this like we do for setting shared to false
+						await tx.dbTransaction.query(`delete from public.file_state where "fileId" = $1`, [
+							file.id,
+						])
+					}
```
ds300, line 82: drive-by fix to clean up all files states after soft deleting a file. we definitely should be using a trigger for this but it can wait for a follow up.

apps/dotcom/sync-worker/src/replicator/ChangeCollator.ts
```diff
@@ -0,0 +1,396 @@
...
+				FROM topic_subscription ts
+				JOIN subscribers s ON ts.toTopic = s.fromTopic
+			)
+			SELECT DISTINCT fromTopic FROM subscribers WHERE fromTopic LIKE 'user:%'
+		`,
+				...nonUserTopics
+			)
+			.toArray()
+
+		const recursiveUsers = result.map((row) => row.fromTopic.replace('user:', ''))
```
MitjaBezensek, line 159: Could pull out this replacing into a function as we do use it twice.

apps/dotcom/sync-worker/src/replicator/ChangeCollator.ts
```diff
@@ -0,0 +1,396 @@
...
+			changes.push(change)
+		}
+	}
+
+	handleEvent(change: ChangeV2) {
+		// we update subscriptions during both normal operation and catch up
+		const { command, table } = change.event
+
+		if (!this.isCatchUp) {
+			const effects = getEffects(change)
```
MitjaBezensek, line 190: Could return an empty array instead of null if we have no effects, then we wouldn't need the check below.

apps/dotcom/sync-worker/src/replicator/ChangeCollator.ts
```diff
@@ -0,0 +1,396 @@
...
+			if (effects) {
+				this.effects.push(...effects)
+			}
+		}
+
+		// Check if any of the topics have listeners
+		const hasAnyListener = this.hasListenerForTopics(change.topics)
```
MitjaBezensek, line 197: Got thrown off a bit by this as I thought this is always `true` since it lives inside `LiveChangeCollator`. I guess I'd probably pull this method up to an abstract class `ChangeCollator`, then have two proper subclasses of that. Also fine if you keep it like it is though.

MitjaBezensek (approved):
Nice, this is looking really good 👍 Should allow for easy additions of new entities to our system if / when the need arises.

PR #6187, opened 2025-06-05
Use socket hibernation for user durable object

ds300:
This switches to using the socket hibernation api for the user durable to reduce the number of cloudflare CPU minutes used by idling browser tabs.

### Change type

- [x] `other`


steveruizok:
Still seems like a good idea, right?

steveruizok:
Closing this one, we can pick it up again or take some of the other PRs built on top

PR #6193, opened 2025-06-06
Add history page behind debug menu for tla files

ds300:
This PR adds the history page routes for new files. Only the owner of the file can access them or revert changes.

I've snuck a UI affordance in via a file menu option that is only visible if you have debug mode switched on.

### Change type

- [x] `other`


apps/dotcom/client/src/tla/components/TlaEditor/TlaLegacySnapshotEditor.tsx
```diff
@@ -28,47 +27,23 @@ export const components: TLComponents = {
 }
 
 export function TlaLegacySnapshotEditor({
-	fileSlug,
 	snapshot,
-	timeStamp,
-	context,
-	token,
+	fileSlug,
 }: {
```
ds300, line 32: refactored this to only work on the old snapshot routes `/s/:slug` and made a new editor component specifically for the history snapshot.

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopLeftPanel.tsx
```diff
@@ -248,6 +261,20 @@ export function TlaEditorTopLeftPanelSignedIn() {
 				<TldrawUiMenuGroup id="preferences">
 					<PreferencesGroup />
 				</TldrawUiMenuGroup>
+				{isDebugMode && (
```
MitjaBezensek: We could also check if we are the owner here and hide the menu item for non-owners?

PR #6195, opened 2025-06-06
[Do not merge] Prevent editing of legacy multiplayer rooms

MitjaBezensek:
Changes to old multiplayer rooms:
* make them read-only
* make it easier to copy them (made sure to only show the modal once for the case where you not not logged in, then you log in - we only show the dialog when not logged in)
* disable sharing 

### Change type

- [x] `improvement`


apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLegacyModal.tsx
```diff
@@ -39,15 +76,36 @@ function LegacyChangesModal({ onClose }: { onClose(): void }) {
...
+			return
+		}
+
+		// We might have shown the modal before. If that's the case, we don't show it again.
+		// In case the user has no files, we show the modal anyway, so they can copy the room to their files.
+		const dontShowModal =
+			searchParams.get('shownLegacyModal') === 'true' && app?.getUserOwnFiles().length !== 0
+		if (dontShowModal) {
+			return
+		}
```
ds300: playing around with this I kinda think it would be fine to always show the modal. wdyt?

ds300 (approved):
seems to work well!

apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLegacyModal.tsx
MitjaBezensek: Sure, let's go with always showing it. Was a bit undecided myself 👍

PR #6202, opened 2025-06-06
Remove `getShapePageGeometry`

steveruizok:
This PR removes `getShapePageGeometry`. 

### Background

Geometries are complex and so implement lots of internal caching to avoid recomputing vertices or bounds. These properties are often referenced many times during interactions like snapping, selecting, or translating.

Geometry transforms were added in https://github.com/tldraw/tldraw/pull/5754 while working on elbow arrows. When a geometry is transformed, its vertices etc and other things are no longer cached and they're recomputed instead. This is fine if the transform is part of the creation of the shape (as in the PathBuilder2d class) but not if we're creating new geometries as we access them, which is done by `getShapePageGeometry`.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/packages/editor/src/lib/utils/reorderShapes.ts
+++ b/packages/editor/src/lib/utils/reorderShapes.ts
@@ -1,8 +1,6 @@
 import { TLParentId, TLShape, TLShapeId, TLShapePartial } from '@tldraw/tlschema'
 import { IndexKey, compact, getIndicesBetween, sortByIndex } from '@tldraw/utils'
 import { Editor } from '../editor/Editor'
-import { Vec } from '../primitives/Vec'
-import { polygonsIntersect } from '../primitives/intersect'
 
 export function getReorderingShapesChanges(
 	editor: Editor,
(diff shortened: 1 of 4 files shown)


packages/tldraw/src/lib/shapes/arrow/elbow/elbowArrowSnapLines.tsx
```diff
@@ -71,13 +71,16 @@ export function getElbowArrowSnapLines(editor: Editor) {
 					if (!shapeBounds || !viewportBounds.includes(shapeBounds)) continue
 
 					const bindings = getArrowBindings(editor, shape)
+					const pageTransform = editor.getShapePageTransform(id)
+					if (!pageTransform) continue
 
-					const geometry = editor.getShapePageGeometry(id)
-					const vertices = geometry.getVertices({ includeInternal: false, includeLabels: false })
```
steveruizok, line 76: This causes the vertices to be recomputed

packages/editor/src/lib/utils/reorderShapes.ts
```diff
@@ -155,19 +153,18 @@ function reorderToFront(moving: Set<TLShape>, children: TLShape[], changes: TLSh
 }
 
 function getOverlapChecker(editor: Editor, moving: Set<TLShape>) {
```
steveruizok, line 155: We don't really need to check all of the vertices, since this is a minor optimization anyway. Bounds should be fine.

PR #6217, opened 2025-06-09
Fix renaming from the main menu.

MitjaBezensek:
We had to call `onEnd` so that we would stop renaming in this case. Otherwise [this](https://github.com/tldraw/tldraw/blob/0e269ff1b1e8852afc97bcab6773fa70e05622b7/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopLeftPanel.tsx#L301-L306) `useEffect` would immediately put us back into renaming.

See the linked issue for bug description.

Fixes INT-1064
 
### Change type

- [x] `bugfix`


steveruizok:
I think I broke that last week, let me check why

steveruizok:
Nope, wasn't me. This looks good!

MitjaBezensek:
> Nope, wasn't me. This looks good!

Yeah, just found this reported 3 months ago as well, so looks like it's been there for a while 😅
https://linear.app/tldraw/issue/INT-1064/renaming-file-glitch

PR #6218, opened 2025-06-09
Default to draw tool if the page is empty

MitjaBezensek:
* Move all the sneaky components into the existing sneaky folder.
* Fix the comment in the `SneakyHandToolEmptyPage`
* Default to draw tool if we don't have any shapes on the current page.

Even if we don't go with the last option we should at least merge the first two changes (which are in separate PRs).

### Change type

- [x] `improvement`


steveruizok:
Oh I've thought about this too. This is just on tldraw.com, right? I'm good to try it there

MitjaBezensek:
> Oh I've thought about this too. This is just on tldraw.com, right? I'm good to try it there

Yes, just tldraw.com.

apps/dotcom/client/e2e/fixtures/Editor.ts
```diff
@@ -81,6 +81,7 @@ export class Editor {
 
 	@step
 	async createTextShape(text: string) {
+		await this.page.getByTestId('tools.select').click()
```
MitjaBezensek, line 84: Since we now default to draw we have to switch to select first, otherwise double click won't create a shape.

apps/dotcom/client/e2e/fixtures/HomePage.ts
```diff
@@ -54,6 +56,7 @@ export class HomePage {
 	async isLoaded() {
 		await expect(async () => {
 			await expect(this.tldrawEditor).toBeVisible({ timeout: 10000 })
+			await expect(this.tldrawCanvas).toBeVisible({ timeout: 10000 })
```
MitjaBezensek, line 59: Some file menu actions failed due to the defaulting to the draw tool. Think the issue was:
1. We opened the file menu in the sidebar (sidebar loaded before the editor).
2. Editor loaded and the tool switcher switched to the draw tool, which closed the menu.
3. We then tried to rename the file, but the menu was closed.

Adding this check makes sure our editor is properly loaded first, only then we start applying sidebar actions.

PR #6221, opened 2025-06-09
Allow downloading of the original image / video

MitjaBezensek:
Allow users to download original assets. Didn't make this generic (like adding `canDownloadOriginal` to shape util), but if the users would want it we could support it down the line. 

### Change type

- [x] `feature`

### Release notes

Allow users to download the original asset via a built in action. Supports `TLImageShape` and `TLVideoShape`. Users can access this from the main menu as well as the context menu when only shapes that support downloading are selected.


mimecuvalo:
this is brilliant

ds300:
this is good we should do a merge

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1565,6 +1578,47 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				onSelect: async (source) => {
+					const selectedShapes = editor.getSelectedShapes()
+					if (selectedShapes.length === 0) return
+
+					const mediaShapes = selectedShapes.filter((s): s is TLImageShape | TLVideoShape =>
+						supportsDownloadingOriginal(s, editor)
+					)
+
+					// We only want to download originals if we only have media shapes selected
+					if (mediaShapes.length !== selectedShapes.length) return
```
mimecuvalo: i don't know about this bit. i worry that, say, there's some images and they're connected by some arrows. the download will fail b/c the connector shapes are stopping the action. wdyt?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1565,6 +1578,47 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+						if (!url) return
+
+						const link = document.createElement('a')
+						link.href = url
+
+						if (
+							(asset.type === 'video' || asset.type === 'image') &&
+							!asset.props.src.startsWith('asset:')
+						) {
+							link.download = asset.props.name
```
mimecuvalo, line 1609: hm, i wonder in the case of multiple downloads if the browser throws a "looks like the website is trying to do many downloads at once". 

what do you think about using `archiver` to bundle the files together?
https://www.npmjs.com/package/archiver
my understanding of `archiver` is it can output a zip file without compression applied (we don't need compression, just a bundler type file format). plus it seems that npm package is pretty lightweight.

mimecuvalo (commented):
sorry for the delayed review! some thoughts added

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: It will only fail if you select both images and arrows yeah. Not sure, but to me it feels worse to have a bunch of stuff selected and then only some of it ends up getting downloaded, no? With having a strict rule to only show the `Download original` button when you have media shapes selected we are more close to setting the right expectation.

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: Yeah, you do get the prompt to allow downloading of multiple files.

I'm fine with going either way. I kinda prefer that I don't have to unpack things and the delete the archive, but I can also see the benefit of not seeing the prompt and only having one file in the end. 

I guess the archive would work well for the use case of making a copy of the assets for backup purposes, while the multiple files are more useful if you want to immediately do something with them 🤷‍♂️

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: In my mind, I think it should be less programmatically strict and more on the human-level of thinking.
So the way I would frame it is a graceful degradation of actions around the content that you're interacting with.

For example, on a website, if I copy-paste a page on Wikipedia that includes both text, images, tables. I would expect copy-paste to still work and paste into VSCode, even though I accidentally selected some images which won't be copied into my text editor.
So, I would expect if I selected some shapes which were images that in the context menu the 'Download original' option would be available to click on b/c it's not enforcing a strictness of selection.

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: Yeah my worry is that folks will miss the prompt and then wonder what's going on.
When you mention the backup use-case - that does sound compelling actually!

steveruizok: I think we can skip the question for now and just download multiple files

mimecuvalo: To be clear on my comment: I think it shouldn't archive if it's just one single image. But if it's multiple, an archive would be killer. Maybe needs a discussion with the team if we wanna discuss how this fiddly bit of the feature would work.

packages/tldraw/src/lib/ui/context/actions.tsx
steveruizok: I think Mime's right, I'd expect it to download everything that can be downloaded. I'd look to other apps here for guidance, it must be a common pattern.

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -65,6 +67,17 @@ export interface ActionsProviderProps {
...
 }
 
+/** @public */
+export function supportsDownloadingOriginal(
+	shape: TLShape,
+	editor: Editor
+): shape is TLImageShape | TLVideoShape {
+	return (
+		(editor.isShapeOfType(shape, 'image') || editor.isShapeOfType(shape, 'video')) &&
+		!!(shape as any).props.assetId
```
steveruizok, line 77: my API head goes to `ShapeUtil.getDownload` or `getDownloadUrl` or `getOriginal` or something, where the downloadable asset is generated. This would be really useful for example in tldraw computer for images / audio / videos, which we've had to implement separately.

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: Looks like most apps don't even have this feature. They use copy to clipboard or selection export (Figma, FigJam, Whimsical, Excalidraw)

Miro has a download, but only for a single image. Once you select multiple images you don't get that option. Same for Notion.
CleanShot 2025-06-13 at 12 13 22
CleanShot 2025-06-13 at 12 16 55

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: > When you mention the backup use-case - that does sound compelling actually!

Yeah, it's compelling but we also need to take the likelihood of that use case into consideration. I'd think it's less likely than just downloading some images for further use. For full backup you'd probably rather use download the whole file 🤷‍♂️

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: I'll loosen the requirement so that this works if at least one downloadable shape is selected 👍

packages/tldraw/src/lib/ui/context/actions.tsx
MitjaBezensek: Yeah, I considered that, but was thinking more towards:
* `ShapeUtil` would only have a method like `canDownload`.
* We'd continue relying on `editor.resolveAssetUrl` to get the actual asset.

It could also work the other way around, but then we'd do something like:
```ts
function getDownloadUrl(){
    return this.editor.resolveAssetUrl(...)
}
```

Which feels a bit convoluted? Unless we had shapes that allowed downloading without having a backing asset.

packages/tldraw/src/lib/ui/context/actions.tsx
steveruizok: Let's talk archiving as a follow up

packages/tldraw/src/lib/ui/context/actions.tsx
steveruizok: I wouldn't rely too much on the assets, that seems like an implementation detail. Our asset system is really only needed if you're trying to minimize duplication (such as with base64'd image sources) or if you expect the asset's source to change and need to react to that change. For example in computer the shapes just have the url of the source image / video. There could be lots of third party shapes where the asset isn't stored in our asset system.

steveruizok: I don't think this is a blocker, I'd like to have this action in there (I separately implemented it in my branch of mime's image toolbar PR

PR #6238, opened 2025-06-11
Improve text measurement perf

steveruizok:
This PR improves the performance of text measurement.

- reuses the same element for measurement rather than creating / mounting new elements for each measurement
- adds `measureScrollWidth` as an option to measureText because this is only used in note shapes.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape with a label
2. Resize it

### Release notes

- Improved rich text performance


steveruizok:
Broke this one down for parts, closing

PR #6243, opened 2025-06-12
Stroke and font size customization example.

MitjaBezensek:
Show how to customize stroke and font sizes.

### Change type

- [x] `improvement`


apps/examples/src/examples/custom-stroke-and-font-sizes/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Custom stroke and font sizes
+component: ./CustomStrokeAndFontSizes.tsx
+category: configuration
+priority: 2
+keywords: [stroke, font, size, style]
+---
+
+Customize the available stroke and font sizes in tldraw.
```
steveruizok, line 9: These kinds of examples are tricky because their changes will also apply to all of the other examples (similar to the "customize color" example).

steveruizok: Not saying we shouldn't do it, but I'd added a warning / note to the custom color example about it.

MitjaBezensek: Ah, right. Added a note for now. We could also have some cleanup logic, but it would complicate the example and possibly make it confusing.

steveruizok (approved):

PR #6244, opened 2025-06-12
Add focus mode example.

MitjaBezensek:
### Change type

- [x] `improvement`

### DIFF
--- a/apps/examples/src/examples/focus-mode/README.md
+++ b/apps/examples/src/examples/focus-mode/README.md
@@ -1,5 +1,5 @@
 ---
-title: Focus mode
+title: Toggle focus mode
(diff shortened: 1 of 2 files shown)


apps/examples/src/examples/focus-mode/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Focus mode
```
TodePond:
```suggestion
title: Toggle focus mode
```
Looking at the list, it could confuse with some of the 'focus' examples. Let's rename it to this maybe?

MitjaBezensek: Sounds good 👍

PR #6248, opened 2025-06-13
Redact the url.

MitjaBezensek:
This allows you to still probably get the file (with a combination of the user's email if they decided to share it + the part of the slug), but prevents directly opening user's files.

### Change type

- [x] `improvement`


steveruizok:
As mentioned in DMs, since this is to protect links posted to our internal feedback system, I think it would be better to add something to the front of the URL, like `pleasebemindful`, so that the url can't be opened accidentally but could be more easily opened if we need it, i.e. `pleasebemindful-https://tldraw.com/f/etc`.

MitjaBezensek:
> As mentioned in DMs, since this is to protect links posted to our internal feedback system, I think it would be better to add something to the front of the URL, like `pleasebemindful`, so that the url can't be opened accidentally but could be more easily opened if we need it, i.e. `pleasebemindful-https://tldraw.com/f/etc`.

Sounds good, reworked it to add a prefix.

mimecuvalo:
thank you both!

PR #6251, opened 2025-06-13
Massively speed up 'fetch everything' query

ds300:
This uses joins (fast) instead of `where exists` (slow) to get a big old speed boost on the main 'fetch everything' query. Also simplified the query generation quite a bit.

This takes queries from the order of ~100ms down to the order of ~1ms.

Also switched from `UNION` (which dedupes rows) to `UNION ALL` (which doesn't). We don't need to make postgres check the rows for deduping because our queries don't introduce duplicates. This seems to save only a small amount (fraction of a ms) but I reckon it could be a bigger difference later on with groups when people's result set sizes will grow significantly.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/fetchEverythingSql.snap.ts
+++ b/apps/dotcom/sync-worker/src/fetchEverythingSql.snap.ts
@@ -3,7 +3,10 @@
 
 export const fetchEverythingSql = `
 WITH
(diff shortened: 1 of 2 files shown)


steveruizok:
hell yeah

MitjaBezensek (approved):
👍

PR #6254, opened 2025-06-15
perf(geo): skip measurement for minimum width

steveruizok:
This PR improves geo shape resizing performance by skipping the measurement we were doing for the shape's minimum width. This should not produce any changes in existing projects but it will allow geo shapes to be made more narrow if their text contents are single characters.

image

Someday I'd like to turn off these minimums completely and allow for a "tinyX / tinyY" flags to be set that allow users to create geo shapes that ignore fitting their labels.

### Change type

- [x] `improvement`

### Test plan

1. Create lots of geo shapes
2. Resize them, it should be faster
3. Mush shapes down

- [x] Unit tests (if present)

### Release notes

- Improved performance of resizing geo shapes.

### DIFF
--- a/packages/tldraw/src/test/commands/__snapshots__/getSvgString.test.ts.snap
+++ b/packages/tldraw/src/test/commands/__snapshots__/getSvgString.test.ts.snap
@@ -70,14 +70,14 @@ exports[`Matches a snapshot: Basic SVG 1`] = `
       transform="matrix(1, 0, 0, 1, 0, 0)"
     >
       <path
-        d="M 0 0 L 100 0 L 100 428 L 0 428 Z"
+        d="M 0 0 L 100 0 L 100 100 L 0 100 Z"
         fill="none"
         stroke="#1d1d1d"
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/test/commands/__snapshots__/getSvgString.test.ts.snap
```diff
@@ -202,14 +202,14 @@ exports[`Returns all shapes when no ids are provided: All shapes 1`] = `
...
-        d="M 0 0 L 100 0 L 100 428 L 0 428 Z"
+        d="M 0 0 L 100 0 L 100 100 L 0 100 Z"
         fill="none"
         stroke="#1d1d1d"
         stroke-width="3.5"
       />
       <foreignobject
         class="tl-export-embed-styles tl-rich-text tl-rich-text-svg"
-        height="428"
+        height="100"
```
steveruizok, line 212: not sure why this was being set incorrectly before

PR #6275, opened 2025-06-18
Prevent duplication of locked shapes, and unlock pasted locked shapes

TodePond:
This PR makes it so that...
- You can't duplicate locked shapes anymore. We were already disabling the button. Now the keyboard shortcut's behavior reflects that too.
- When you paste a locked shape, it now unlocks the pasted shape. Note: Only root level shapes get unlocked. If you paste a  frame that... say... has a locked child inside it, the pasted locked child will stay locked.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Prevented duplication of locked shapes.
- Pasting a locked shape onto the canvas now unlocks it.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -6211,11 +6211,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	duplicateShapes(shapes: TLShapeId[] | TLShape[], offset?: VecLike): this {
 		this.run(() => {
-			const ids =
+			const _ids =
 				typeof shapes[0] === 'string'
 					? (shapes as TLShapeId[])
 					: (shapes as TLShape[]).map((s) => s.id)
 
+			const ids = this._shouldIgnoreShapeLock ? _ids : this._getUnlockedShapeIds(_ids)
 			if (ids.length <= 0) return this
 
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -6211,11 +6211,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	duplicateShapes(shapes: TLShapeId[] | TLShape[], offset?: VecLike): this {
 		this.run(() => {
-			const ids =
+			const _ids =
 				typeof shapes[0] === 'string'
 					? (shapes as TLShapeId[])
 					: (shapes as TLShape[]).map((s) => s.id)
 
+			const ids = this._shouldIgnoreShapeLock ? _ids : this._getUnlockedShapeIds(_ids)
```
mimecuvalo, line 6219: mind adding a test to `duplicate.test.ts`? 🙏

mimecuvalo (approved):
pre-approving modulo adding a quick test pls!

packages/editor/src/lib/editor/Editor.ts
TodePond: will do!

PR #6278, opened 2025-06-18
image cropping: make sure we setCroppingShape to null when exiting the state machine

mimecuvalo:
we never noticed that finishing a crop sometimes leaves the cropping state in a half-state.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Make sure we setCroppingShape to null when exiting the state machine.

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Crop/children/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Crop/children/Idle.ts
@@ -24,6 +24,7 @@ export class Idle extends StateNode {
 	}
 
 	override onExit() {
+		this.editor.setCroppingShape(null)
 		this.editor.setCursor({ type: 'default', rotation: 0 })
 	}
 


mimecuvalo:
oh wait doh lemme rework

mimecuvalo:
hm, there doesn't seem to be a good way to fix this at the base level because we don't know the next state really. we'll revisit more fully in the future

packages/editor/api-report.api.md
```diff
@@ -2700,7 +2700,7 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
     // (undocumented)
     enter(info: any, from: string): void;
     // (undocumented)
-    exit(info: any, from: string): void;
+    exit(info: any, to: string): void;
```
mimecuvalo, line 2703: drive-by typo fix

packages/tldraw/src/lib/ui/components/Toolbar/DefaultImageToolbar.tsx
```diff
@@ -46,7 +46,10 @@ function ContextualToolbarInner({
 
 	const isInCropTool = useValue('editor path', () => editor.isIn('select.crop.'), [editor])
 	const isCropping = useValue('editor path', () => editor.isIn('select.crop.cropping'), [editor])
-	const handleManipulatingEnd = useCallback(() => editor.setCurrentTool('select.idle'), [editor])
+	const handleManipulatingEnd = useCallback(() => {
+		editor.setCroppingShape(null)
```
mimecuvalo, line 50: we can't do this `onExit` in the crop.idle tool - we'll have to rethink that another time

mimecuvalo:
gonna land in interests of deploy

PR #6282, opened 2025-06-19
Require modern browsers

steveruizok:
image

We use Array.findLastIndex which was shipped across browsers in 2023. We keep seeing this error from users with old browsers. At the moment the app just hangs indefinitely. You can reproduce it like this:

```ts
commitMutations(mutationIds: string[]) {
	this._optimisticStore.update((prev) => {
		if (!prev) return prev
		delete prev.prototype.findLastIndex
		const highestIndex = prev.findLastIndex((p) => mutationIds.includes(p.mutationId))
		return prev.slice(highestIndex + 1)
	})
}
```

Which will result in an infinite spinner as the error occurs behind the scenes. This PR at least tells the user to update. We could improve this experience though it's a very rare one.

Closes https://github.com/tldraw/tldraw/issues/6281

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Fixed a silent bug effecting out of date browsers. Update your browser!

### DIFF
--- a/packages/dotcom-shared/src/OptimisticAppStore.ts
+++ b/packages/dotcom-shared/src/OptimisticAppStore.ts
@@ -17,7 +17,7 @@ export class OptimisticAppStore {
 	constructor() {
 		// this is for the one guy in morocco who is still on firefox 102
 		if (!('findLastIndex' in Array.prototype)) {
-			window.alert('come on, update your browser')
+			window.alert('hey update your browser')
 			throw Error('browser too old')
 		}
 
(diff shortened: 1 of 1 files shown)


mimecuvalo (changes requested):
i mean...lemme rework this - we can make this better :)

mimecuvalo:
PR here https://github.com/tldraw/tldraw/pull/6302

ds300:
closing in favor of #6302

PR #6285, opened 2025-06-19
test(dotcom): add e2e tests for existing document compatibility

MitjaBezensek:
Run e2e tests on existing document to make sure we didn't break any of them.

### Change type

- [x] `other`

### Test plan

1. Run the new e2e tests in the dotcom package.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Added e2e tests to verify compatibility with existing documents.

### DIFF
--- a/.github/workflows/staging-e2e.yml
+++ b/.github/workflows/staging-e2e.yml
@@ -8,7 +8,6 @@ on:
         required: false
         default: '10'
         type: number
-  pull_request:
 
 permissions:
   contents: read


.github/workflows/staging-cleanup.yml
```diff
@@ -0,0 +1,37 @@
...
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Run our setup
+        uses: ./.github/actions/setup
+
+      - name: Cleanup test files
+        run: yarn tsx internal/scripts/dotcom/cleanup-test-files.ts
```
MitjaBezensek: We could also call cleanup after we run the test.

apps/dotcom/client/e2e/fixtures/HomePage.ts
```diff
@@ -37,6 +37,22 @@ export class HomePage {
 		}).toPass()
 	}
 
+	async loginWithEmailAndPassword(email: string, password: string) {
```
MitjaBezensek, line 40: Enabled logging in with a password for staging.

apps/dotcom/client/e2e/fixtures/helpers.ts
```diff
@@ -76,3 +77,25 @@ export function getRandomName() {
 	}
 	return result
 }
+
```
MitjaBezensek, line 80: Pulled this part out as it's used in multiple places now.

apps/dotcom/client/e2e/fixtures/staging-test.ts
MitjaBezensek, line 1: Created a separate setup for the staging tests. Using a different login method, so it made it easier to do it like this.

apps/dotcom/client/package.json
```diff
@@ -20,10 +20,11 @@
 		"test": "yarn run -T jest",
 		"i18n:extract": "formatjs extract 'src/{tla,pages,components}/**/*.ts*' --out-file public/tla/locales/en.json --additional-component-names F --id-interpolation-pattern '[md5:contenthash:hex:10]' --format lokalise",
 		"i18n:compile": "formatjs compile-folder public/tla/locales public/tla/locales-compiled --format lokalise",
-		"e2e": "rm -rf 'e2e/.auth' && yarn playwright test",
-		"e2e-x10": "yarn playwright test --repeat-each=10",
-		"e2e-debug": "yarn playwright test --debug",
-		"e2e-ui": "yarn playwright test --ui"
+		"e2e": "rm -rf 'e2e/.auth' && yarn playwright test --project=chromium",
```
MitjaBezensek, line 23: Had to scope these to a specific project, so that we don't run the staging tests as well.

apps/dotcom/sync-worker/migrations/001_create_test_file_table.sql
MitjaBezensek, line 1: Added storing of the test ids in d1 so that we can delete them later.

apps/dotcom/sync-worker/src/utils/tla/getAuth.ts
MitjaBezensek, line 1: Pulled this out.

internal/scripts/dotcom/copy-random-files-to-staging.ts
```diff
@@ -0,0 +1,193 @@
...
+			__dirname,
+			'../../../apps/dotcom/client/e2e/fixtures/test-files.json'
+		)
+		writeFileSync(testFilesPath, JSON.stringify(copiedIds, null, 2))
+		console.log('Updated test files')
+	}
+}
+
+async function main() {
+	const n = 10
```
MitjaBezensek: 10 while we test this, can decide on exact number later.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -382,6 +395,12 @@ export class TLDrawDurableObject extends DurableObject {
 				if (file.isDeleted) {
 					return closeSocket(TLSyncErrorCloseEventReason.NOT_FOUND)
 				}
+
+				const testFile = await isTestFile(this.env, file.id)
+				if (testFile && !(await canAccessTestProductionFile(this.env, auth))) {
```
MitjaBezensek: Only limited this to `@tldraw.com` email, but we could pass in the `STAGING_OWNER_ID` and only allow access to that user 🤷‍♂️

apps/dotcom/sync-worker/migrations/001_create_test_file_table.sql
ds300: can we just prefix them with `test_` to avoid the need for a separate database?

internal/scripts/dotcom/copy-random-files-to-staging.ts
```diff
@@ -0,0 +1,193 @@
...
+	const query = `
+		SELECT * FROM file
+		TABLESAMPLE BERNOULLI(30)
+		WHERE "shared" = true
+		  AND ("updatedAt" - "createdAt") >= $1
+		LIMIT $2;
+	`
+
+	const { rows } = await client.query(query, [threeDays, n])
+	await client.end()
```
ds300, line 48: Just in case the query errors, so the script still exits cleanly.
```suggestion
	const { rows } = await client.query(query, [threeDays, n]).finally(() => client.end())
```

internal/scripts/dotcom/copy-random-files-to-staging.ts
```diff
@@ -0,0 +1,193 @@
...
+				testFile.sharedLinkType,
+				testFile.published,
+				testFile.lastPublished,
+				testFile.publishedSlug,
+				testFile.createdAt,
+				testFile.updatedAt,
+				testFile.isEmpty,
+				testFile.isDeleted,
+			])
+			await stagingClient.end()
```
ds300: same here
```suggestion
			]).finally(() => stagingClient.end())
```

apps/dotcom/sync-worker/migrations/001_create_test_file_table.sql
MitjaBezensek: Sure. I guess staging won't have that many files so listing the elements and filtering should be fine. Or even better, we can query supabase and get all the test user's files and then clean up based on that.

apps/dotcom/sync-worker/src/utils/tla/getAuth.ts
```diff
@@ -105,3 +105,17 @@ export async function requireWriteAccessToFile(
 		await db.destroy()
 	}
 }
+
+export async function requireAdminAccess(env: Environment, auth: { userId: string } | null) {
```
ds300, line 109: nice

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: admin emails is fine i think

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -382,6 +395,12 @@ export class TLDrawDurableObject extends DurableObject {
...
+					return closeSocket(TLSyncErrorCloseEventReason.NOT_FOUND)
+				}
```
ds300, line 401: should there be an `else if (testFile) { also fail }` clause here? I guess if it has `shared: false` it's no biggie

apps/dotcom/client/e2e/tests/staging.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+	if (existsSync(testFilesPath)) {
+		try {
+			testFiles = JSON.parse(readFileSync(testFilesPath, 'utf-8')) as string[]
+		} catch (error) {
+			console.warn('Failed to read test-files.json:', error)
+			testFiles = []
+		}
+	}
+
+	test.skip(testFiles.length === 0, 'No test files available')
```
ds300: what's this for?

apps/dotcom/client/e2e/tests/staging.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { existsSync, readFileSync } from 'fs'
+import { join } from 'path'
+import { expectBeforeAndAfterReload } from '../fixtures/helpers'
+import { expect, test } from '../fixtures/staging-test'
+
+test.describe.configure({ mode: 'serial' })
```
ds300: is 'serial' required for some reason?

ds300 (commented):
looking really good

.github/workflows/staging-cleanup.yml
ds300: What about running it at the _start_ of the next test run? That way we keep the test files around in case there's a failure but also we don't need to worry about manually cleaning them up.

internal/scripts/dotcom/copy-random-files-to-staging.ts
MitjaBezensek: Reworked this part to only create the client once and then end it in a finally 👍

apps/dotcom/client/e2e/tests/staging.spec.ts
MitjaBezensek: We will use a single user, so this should mean a bit better test isolation and might prevent some strange behaviours if we run tests in parallel. Not sure if a real issue though 🤷‍♂️

apps/dotcom/client/e2e/tests/staging.spec.ts
MitjaBezensek: Some leftover, will remove 👍

.github/workflows/staging-cleanup.yml
MitjaBezensek: Yeah, probably best if we go with that.

I'd say let's keep it like it is for now, just so I can run these two workflows separately, then once they are fully tested I'll move it to run at the start of the next test run.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
MitjaBezensek: Think our regular check a bit further down should cover that case?

.github/workflows/staging-cleanup-daily.yml
```diff
@@ -0,0 +1,38 @@
+name: Staging E2E Daily Cleanup
+
+on:
+  schedule:
+    - cron: '45 1 * * *'
```
MitjaBezensek, line 5: Let's do it during the night and not at an exact hour.

PR #6286, opened 2025-06-19
improvement(a11y): allow switching back-and-forth from the style panel

mimecuvalo:
When using the Cmd+Enter shortcut, you can then adjust styles in the style panel.
However, the issue is that it's very tricky to get back to the shape you were manipulating. Hitting Escape takes you back to an unfocused canvas at the moment.
This change makes it so that you can hit Escape and focus the container again properly.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Select a shape and press Cmd+Enter to focus the style panel.
2. Press Escape and verify focus returns to the shape container.

- [ ] Unit tests (if present)

### Release notes

- a11y: allow switching back to the shape when doing adjusting styles.

### DIFF
--- a/apps/examples/e2e/tests/test-a11y-style-panel-navigation.spec.ts
+++ b/apps/examples/e2e/tests/test-a11y-style-panel-navigation.spec.ts
@@ -1,0 +1,259 @@
+import { Page, expect } from '@playwright/test'
+import { Editor } from 'tldraw'
+import { setupPage } from '../shared-e2e'
+import test from './fixtures/fixtures'
+
+declare const editor: Editor
+
+test.describe('A11y Style Panel Navigation', () => {
+	test.beforeEach(async ({ page }) => {
+		await setupPage(page)
+	})
+
+	test('can navigate between geo shape label editing and style panel with keyboard', async ({
+		page,
+		isMobile,
+	}) => {
+		if (isMobile) {
+			// Skip on mobile as the keyboard navigation behavior is different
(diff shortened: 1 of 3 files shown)


steveruizok:
Adding end to end tests! I noticed / fixed one issue which is that pressing Cmd+Enter to complete editing rich text (in geo shapes) would not re-focus the canvas.

mimecuvalo:
> It's like that in production as well, so not related to this PR, but just wanted to share it.

@MitjaBezensek (didn't get a chance to answer your question before this merged). this is expected, the browser has its own internal heuristics that determine when the user has gone into an "accessibility mode", basically when `:focus-visible` becomes enabled. so you have to tab around a bit first and then hit cmd+enter and it will always show up. if you just click on a shape and hit cmd+enter the heuristic won't trigger.

MitjaBezensek:
> > It's like that in production as well, so not related to this PR, but just wanted to share it.
> 
> @MitjaBezensek (didn't get a chance to answer your question before this merged). this is expected, the browser has its own internal heuristics that determine when the user has gone into an "accessibility mode", basically when `:focus-visible` becomes enabled. so you have to tab around a bit first and then hit cmd+enter and it will always show up. if you just click on a shape and hit cmd+enter the heuristic won't trigger.

I guess the confusing part for me was that it works like I'd expect the second time 🤷‍♂️

PR #6288, opened 2025-06-19
feat(a11y): add keyboard shortcut to open context menu

mimecuvalo:
Builds off of our Cmd+Enter shortcut. Allows access to our context menu via kbd shortcut.
This was originally explored in https://github.com/tldraw/tldraw/pull/5827 but it was using the wrong dispatch event (`pointerdown` instead of `contextmenu`) which now solves the issues that we had initially seen there.

https://github.com/user-attachments/assets/bdec0a1c-0c87-4360-b54a-1ed94e6e08e4

### Change type

- [x] `improvement`

### Test plan

1. Select an object on the canvas.
2. Press Cmd+Enter (or the designated shortcut).
3. Verify the context menu opens at the selection.

- [ ] Unit tests (if present)

### Release notes

- Added a keyboard shortcut to open the context menu for improved accessibility.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx
+++ b/packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx
@@ -1,6 +1,6 @@
 import { preventDefault, useContainer, useEditor, useEditorComponents } from '@tldraw/editor'
 import { ContextMenu as _ContextMenu } from 'radix-ui'
-import { ReactNode, memo, useCallback } from 'react'
+import { ReactNode, memo, useCallback, useEffect } from 'react'
 import { useMenuIsOpen } from '../../hooks/useMenuIsOpen'
 import { useTranslation } from '../../hooks/useTranslation/useTranslation'
 import { TldrawUiMenuContextProvider } from '../primitives/menus/TldrawUiMenuContext'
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1537,6 +1537,39 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+					const centerY = selectionBounds.y + selectionBounds.height / 2
+
+					// Convert page coordinates to screen coordinates
+					const screenPoint = editor.pageToScreen(new Vec(centerX, centerY))
+
+					// Dispatch a contextmenu event directly at the center of the selection
+					editor
+						.getContainer()
+						.querySelector('.tl-canvas')
+						?.dispatchEvent(
```
MitjaBezensek, line 1562: Would maybe dispatching via the editor be better? Or does it end up there anyway?

MitjaBezensek (approved):

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: I couldn't get that to work — it's because the `DefaultContextMenu` uses `_ContextMenu.Trigger` which is expecting a native pointer click event.

PR #6290, opened 2025-06-19
Support configuring when arrows should be exact and/or bind

trygve-aaberge-adsk:
By default arrows stop at the edge of the target shape and bind to it. To stop the arrow at the pointer location you have to hold down alt and to avoid binding you have to hold down ctrl. These bindings were hard coded.

In our application we'd like the default to be that it stops at the pointer, and to stop at the edge of the shape you would hold down alt. To support this I added two new options to the arrow util. To make it more generic, I made these options be functions that let's you control these behaviors as you like.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Configure the arrow util, e.g.
```
ArrowShapeUtil.configure({
	shouldBeExact: (editor) => !editor.inputs.altKey,
})
```
2. Create any shape
3. Create an arrow from the shape
4. Observe that it starts/stops at the pointer location by default, and at the edge of the shape when you hold down alt.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Support configuring when arrows should be exact and/or bind


packages/tldraw/src/lib/shapes/arrow/arrow-types.ts
```diff
@@ -85,6 +85,18 @@ export interface ArrowShapeOptions {
 	 * should we wait before we assume the user is targeting precisely instead of imprecisely.
 	 */
 	readonly pointingPreciseTimeout: number
+
+	/**
+	 * When creating an arrow, should it stop exactly at the pointer, or should
+	 * it stop at the edge of the target shape.
+	 */
+	// eslint-disable-next-line @typescript-eslint/method-signature-style
```
trygve-aaberge-adsk, line 93: I disabled this rule because TypeScript doesn't support using readonly on method shorthands (as eslint suggests), and the rule is disabled similarly in `packages/tldraw/src/lib/defaultEmbedDefinitions.ts`.

steveruizok: Looks good

steveruizok:
Looks good! I've made some minor changes but happy to have this one land. Thanks for the PR @trygve-aaberge-adsk !

PR #6296, opened 2025-06-22
fix(text): align line height measurement with rendering

steveruizok:
This PR fixes a small discrepancy between the line height when measuring text vs. rendering. When rendering text, we calculated line height at `floor(lineHeight * fontSize)` while we only the un-floored `lineHeight * fontSize` when measuring. This would cause the height of a note shape, for example, to become incorrectly calculated when the text grew very long.

This PR unifies that system by using the provided `lineHeight` value, e.g. `1.35`, which seems less likely to produce confusing results.

Kapture 2025-06-22 at 09 12 00

### Alternatives

Alternatively, we could floor the `lineHeight * fontSize` when measuring text, though IMO this is less predictable. @mimecuvalo may have guidance on why line-height is done this way when working on rich text. If it was passing forward an earlier decision, I don't remember why we might have chose to do things using a calculated pixel value (consistency across browsers? SVG exports before foreign object?).

### Change type

- [x] `bugfix` 

### Test plan

1. Create a note shape with a lot of text
2. As the text gets bigger, the note shape's top and bottom gaps should remain correct

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fixed a bug with text measurement


steveruizok:
Tagging a few people here for review in case anyone remembers why we did line heights this way.

steveruizok:
> ### Bug: Line Height vs Min Height Mismatch

stale

mimecuvalo:
> Alternatively, we could floor the lineHeight * fontSize when measuring text, though IMO this is less predictable. @mimecuvalo may have guidance on why line-height is done this way when working on rich text. If it was passing forward an earlier decision, I don't remember why we might have chose to do things using a calculated pixel value (consistency across browsers? SVG exports before foreign object?).

yeah, looks like this is very old code, predating any rich text stuff (from the "transfer-out: transfer out" commit): https://github.com/tldraw/tldraw/blame/398bd352ae0601b543f53172e2a655d590f444c3/packages/tldraw/src/lib/shapes/shared/TextLabel.tsx#L78-L82

LGTM

steveruizok:
Ok, updating our snapshots and will land this one

PR #6300, opened 2025-06-23
Fix for now.

MitjaBezensek:
Prevent the commit from triggering another publish.

### Change type

- [x] `bugfix`


internal/scripts/publish-vscode-extension.ts
```diff
@@ -80,7 +80,6 @@ async function packageAndPublish(version: string) {
 	switch (env.TLDRAW_ENV) {
 		case 'production':
 			await exec('yarn', ['package'], { pwd: EXTENSION_DIR })
-			await copyExtensionToReleaseFolder(version)
```
MitjaBezensek, line 83: Don't do it for production as it would retrigger a deploy. Figure out how to do it correctly.

PR #6302, opened 2025-06-23
app: old browsers error page

mimecuvalo:
alt take on https://github.com/tldraw/tldraw/pull/6282
- this gets us a translated string
- we shouldn't do the check in dotcom-shared, but at the app level, esp. since the checks will grow eventually to have other things, not just `findLastIndex`

Screenshot 2025-06-23 at 16 00 12

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
good feedback! updated to be simpler.

ds300:
Can you also hide the 'back to tldraw' link for this error screen?

mimecuvalo:
> Can you also hide the 'back to tldraw' link for this error screen?

sure thing, done.

ds300 (approved):

PR #6303, opened 2025-06-23
Skip ci actions after we commit the extension

MitjaBezensek:
This should not trigger an additional push event for the branch and should prevent cycles / running actions twice.

### Change type

- [x] `bugfix`

### DIFF
--- a/.github/workflows/publish-vscode-extension.yml
+++ b/.github/workflows/publish-vscode-extension.yml
@@ -21,10 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-    # Prevent multiple runs from happening at the same time
     concurrency: vscode-extension-publish
-    # Skip if the commit was made by the bot to avoid infinite loops
-    if: github.event.head_commit.author.name != 'huppy-bot[bot]'
 
     steps:
       - name: Generate GH token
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-vscode-extension.yml
```diff
@@ -21,10 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
-    # Prevent multiple runs from happening at the same time
     concurrency: vscode-extension-publish
-    # Skip if the commit was made by the bot to avoid infinite loops
-    if: github.event.head_commit.author.name != 'huppy-bot[bot]'
```
MitjaBezensek, line 27: Used this till now to avoid cycles.

internal/scripts/publish-vscode-extension.ts
```diff
@@ -80,6 +80,7 @@ async function packageAndPublish(version: string) {
 	switch (env.TLDRAW_ENV) {
 		case 'production':
 			await exec('yarn', ['package'], { pwd: EXTENSION_DIR })
+			await copyExtensionToReleaseFolder(version)
```
MitjaBezensek, line 83: I guess it should be now safe to also commit to the production branch?

ds300: i guess!

PR #6308, opened 2025-06-24
Use wrangler for deletions.

MitjaBezensek:
Use wrangler instead of the api as it does seem to delete things more consistently.

### Change type

- [x] `improvement`

### DIFF
--- a/internal/scripts/prune-preview-deploys.ts
+++ b/internal/scripts/prune-preview-deploys.ts
@@ -112,14 +112,15 @@ async function deletePreviewWorkerDeployment(id: string) {
 		} catch (err) {
 			nicelog(`Failed to delete consumer ${scriptName}: ${err}`)
 		}
-
+		await deletePreviewWorker(id)
 		try {
 			await deleteQueue(queueName)
 		} catch (err) {
 			nicelog(`Failed to delete queue ${queueName}: ${err}`)
 		}
+	} else {
+		await deletePreviewWorker(id)
 	}
-	await deletePreviewWorker(id)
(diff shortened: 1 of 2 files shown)


internal/scripts/prune-preview-deploys.ts
```diff
@@ -85,45 +85,50 @@ async function listPreviewWorkerDeployments() {
...
-				await deleteQueueConsumer({ id, consumerId })
-			}
-			await deleteQueue(id)
+
+		try {
+			await deleteQueueConsumer(queueName, id)
+		} catch (err) {
+			nicelog(`Failed to delete consumer ${id}: ${err}`)
+		}
+		await deletePreviewWorker(id)
```
MitjaBezensek, line 115: Looks like deleting the worker here removes the binding. If we wanted to delete the queue first we'd get an error that a binding is still active 🤷‍♂️

internal/scripts/prune-preview-deploys.ts
```diff
@@ -85,56 +80,47 @@ async function listPreviewWorkerDeployments() {
...
-			}
-			await deleteQueue(id)
-		}
-	}
-	await deletePreviewWorker(id)
-}
-
-const queuesMap = new Map<string, { id: string; consumerId: string | undefined }>()
 
-async function getQueues() {
```
MitjaBezensek, line 128: Deleting via the api required consumer id, where as wrangler allows using the worker id, so this is no longer needed.

ds300 (approved):

PR #6315, opened 2025-06-24
Prevent crash when creating shapes when max shapes is reached

trygve-aaberge-adsk:
The non-null assertion after creating the note shape caused it to crash when max shapes is reached, as the shape would not exist after createShape in that case.

I looked at how the text shape handles this, and handled it similarly.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Set `maxShapesPerPage: 0` in `CustomOptionsExample.tsx`
2. Open the example and try to create a note.

Before this PR it would crash, now it shows the max shapes reach toast.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Prevent crash when creating notes when max shapes is reached

### DIFF
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -35,6 +35,12 @@ export class Pointing extends StateNode {
 				])
 				.select(id)
 
+			const shape = this.editor.getShape(id)
+			if (!shape) {
+				this.cancel()
+				return
+			}
+
 			const parent = this.parent as BaseBoxShapeTool
 			this.editor.setCurrentTool(
 				'select.resizing',
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -11,35 +11,33 @@ export class Pointing extends StateNode {
 	static override id = 'pointing'
 
 	override onPointerMove(info: TLPointerEventInfo) {
-		if (this.editor.inputs.isDragging) {
-			const { originPagePoint } = this.editor.inputs
+		const { editor } = this
+		if (editor.inputs.isDragging) {
+			const { originPagePoint } = editor.inputs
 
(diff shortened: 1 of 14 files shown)


trygve-aaberge-adsk:
After creating this I noticed that arrow, geo and frame shapes had a similar problem, so I added some commits to fix those as well.

trygve-aaberge-adsk:
I found another related bug where it undoes your latest change when the tool cancels. I pushed [a commit](https://github.com/tldraw/tldraw/pull/6315/commits/0a7eea2e408102257a63015a338b8e541e65634a) to fix that too. The bug is described in the commit message.

steveruizok:
Pushing some helpers here, and buttoning up a few other places where we'd run into bugs on this.

steveruizok:
Remind me next time I create an API not to chaining APIs (e.g. "every method returns this") so that we can actually throw errors

packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
```diff
@@ -11,35 +11,33 @@ export class Pointing extends StateNode {
...
-					},
-				])
-				.select(id)
+			const creatingMarkId = editor.markHistoryStoppingPoint(`creating_box:${id}`)
+			const newPoint = maybeSnapToGrid(originPagePoint, editor)
 
-			const shape = this.editor.getShape(id)
-			if (!shape) {
-				this.cancel()
-				return
```
trygve-aaberge-adsk: @steveruizok: Did you mean to remove this `return`? That causes a bug when using this tool. If you have reached max shapes and e.g. select the frame tool, and then hold down the mouse button and drag, it will continue here and try to set the tool to resizing, and you will get `Error: Resizing but nothing is selected` in the console.

steveruizok: !

steveruizok: Fixed here https://github.com/tldraw/tldraw/pull/6359

trygve-aaberge-adsk: Thanks!

PR #6327, opened 2025-06-26
minimap: omit arrows/lines

mimecuvalo:
opening up a mini-PR for a discussion about this, came up in our retro a couple weeks ago.

before:

Screenshot 2025-06-26 at 12 36 20

after:

Screenshot 2025-06-26 at 12 36 08

our analytics pipeline board's minimap (before+after):

Screenshot 2025-06-26 at 12 36 35
Screenshot 2025-06-26 at 12 36 41

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Minimap: omit arrow/lines from rendering for clearer navigation

### API changes

- add `hideInMinimap` to `ShapeUtil` to optionally hide shapes from the minimap.


mimecuvalo:
adding @steveruizok and @MitjaBezensek

MitjaBezensek:
I like it!

Few things to think about:
* Should we maybe expose this in shape util, which would allow users to also opt in into this behaviour for their custom shapes?
* It feels like it would be quite strange if I create a single line / arrow and I don't see it in the minimap. Should we maybe only start doing this after there are a few other shapes present? Or maybe the arrows need to be bound to omit them?

mimecuvalo:
> I like it!
> 
> Few things to think about:
> 
> * Should we maybe expose this in shape util, which would allow users to also opt in into this behaviour for their custom shapes?
> * It feels like it would be quite strange if I create a single line / arrow and I don't see it in the minimap. Should we maybe only start doing this after there are a few other shapes present? Or maybe the arrows need to be bound to omit them?

- good call, made it a method on `ShapeUtil`
- for the 2nd point, eh, i don't know, i think it would be stranger still if something is there but then disappears after 2 or 3 shapes are on the canvas. better to be consistent in that way then to somewhat conditionally render.

MitjaBezensek:
> * for the 2nd point, eh, i don't know, i think it would be stranger still if something is there but then disappears after 2 or 3 shapes are on the canvas. better to be consistent in that way then to somewhat conditionally render.

Yeah, I agree 👍

PR #6331, opened 2025-06-26
feat(editor): add auto-layout feature to remove shape overlaps

covolute-release:
This PR introduces a new method for automatically arranging shapes and a new web worker architecture for improved performance.

### Change type

- [x] `feature`

### Test plan

1. Navigate to example "Layout Shapes (Remove overlaps between shapes)"
2. Create many shapes which overlap
3. Press buttons "?" below
4. Read instructions

1. Setup empty Next.js/Vite-based projects
2. Install tldraw
3. Check that editor.layoutShapes() function is working

- [ ] Unit tests
- [ ] End to end tests

### Release notes

This update introduces two significant enhancements to tldraw: a new method for automatically arranging shapes and a new web worker architecture for improved performance.

### New Feature: Shape Layout Method

A new `layoutShapes` method has been added to the editor to solve the problem of overlapping shapes. This method accepts an array of shapes and automatically repositions them to ensure they are all visible and do not obscure one another.

The method includes several optional parameters for customization:
*   **Animation:** Standard tldraw animation config.
*   **Weighting:** A system to assign weights to shapes, allowing "heavier" shapes to remain fixed while "lighter" shapes are repositioned.
*   **Padding:** A configurable spacing value, which can be a single number or a `[top, right, bottom, left]` array.

The layout calculations are performed by the **WebCola** library.

### New Paradigm: Web Worker Integration for Performance

Previously, all computations were executed on the main UI thread, which could lead to performance degradation during intensive tasks. This release introduces a web worker architecture to offload these tasks.

**Implementation Details:**
*   **Delayed Loading:** Auto-layout adds +35kb to package, but since it is imported via await import('...'), it is loaded only when computation is needed.
*   **Off-Thread Calculation:** The computationally expensive `layoutShapes` logic is now executed in a web worker to avoid blocking the UI.
*   **Zero-Configuration Setup:** The web worker is embedded within the JavaScript output and instantiated via a Blob URL, eliminating any need for special configuration by consuming applications.
*   **Asynchronous Loading:** The worker's code is placed in a separate chunk that is loaded asynchronously only upon the first call to `layoutShapes()`, preserving a small initial bundle size. This has been verified in Next.JS and Vite-based applications.
*   **Build Process:** This integration is managed by `@ast-grep/napi`, a high-performance, native tree-sitter-based parser, during the build step.

Additionally, a usage example and complete method annotations have been included.

Jun-26-2025 2-48-07 PM

### DIFF
--- a/apps/examples/src/examples/layout-shapes/LayoutShapesExample.tsx
+++ b/apps/examples/src/examples/layout-shapes/LayoutShapesExample.tsx
@@ -1,0 +1,112 @@
+import {
+	DefaultToolbar,
+	DefaultToolbarContent,
+	TLComponents,
+	Tldraw,
+	TldrawUiMenuItem,
+	TLUiOverrides,
+	useIsToolSelected,
+	useTools,
+} from 'tldraw'
+import 'tldraw/tldraw.css'
+
+const uiOverrides: TLUiOverrides = {
+	tools(editor, tools) {
+		tools['simple'] = {
+			id: 'simple',
+			icon: 'simple',
+			label: 'Simple',
+			onSelect: () => {
+				editor.layoutShapes(editor.getCurrentPageShapes())
+			},
+		}
+		tools['animation'] = {
+			id: 'animation',
+			icon: 'animation',
+			label: 'Animation',
+			onSelect: () => {
+				editor.layoutShapes(editor.getCurrentPageShapes(), { animation: { duration: 1000 } })
+			},
+		}
+		tools['padding'] = {
+			id: 'padding',
+			icon: 'padding',
+			label: 'Padding',
+			onSelect: () => {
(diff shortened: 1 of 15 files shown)


steveruizok:
cooool

steveruizok:
I don't think we can land this on main... but I think we _can_ add it as an example.

skyyash:
is it supposed to work on clicking `?` buttons? not working for me.

steveruizok:
Sorry @covolute-release, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #6333, opened 2025-06-26
Delete page shapes before deleting the page itself

ds300:
I can't quite believe that we never caught this before now 😳 

When you delete a page the shapes on that page don't get deleted and just hang around invisibly but taking up space forever.

### Change type

- [x] `bugfix`

### Release notes

- Fixed a bug when deleting pages that prevented the shapes on that page from being garbage collected.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -666,6 +666,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 							])
 						}
 					},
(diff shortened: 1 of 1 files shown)


steveruizok:
I almost remember that we did this intentionally so that undos weren't huge... but, like assets, we were supposed to follow up by removing the unused shapes / assets the next time that the document loaded.

mimecuvalo (approved):
👍 worth doing a test @ds300 ?

MitjaBezensek (approved):

Issue #5728, opened 2025-03-21
Some functionality does not work cross-realm


derekcicerone:
### What happened?

We are using tldraw to support drawing annotations over screen shares in the Roam app (https://ro.am). Our app is based on Electron and opens multiple windows so it uses tldraw in cross-realm (aka multi-realm) scenarios. This basically means that we have one window (the main one) which runs all the JavaScript code but its rendering to multiple windows which each have their own DOM. Somewhat confusingly each DOM has its own JavaScript objects for nodes so `instanceof` checks do not work. In our testing we found that only two small changes were needed to make tldraw work in this environment:

```diff
diff --git a/node_modules/@tldraw/editor/dist-esm/lib/editor/Editor.mjs b/node_modules/@tldraw/editor/dist-esm/lib/editor/Editor.mjs
index f0a5444..12d1c11 100644
--- a/node_modules/@tldraw/editor/dist-esm/lib/editor/Editor.mjs
+++ b/node_modules/@tldraw/editor/dist-esm/lib/editor/Editor.mjs
@@ -2589,7 +2589,7 @@ class Editor extends (_a = EventEmitter, _getIsShapeHiddenCache_dec = [computed]
    * @public
    */
   updateViewportScreenBounds(screenBounds, center = false) {
-    if (screenBounds instanceof HTMLElement) {
+    if (screenBounds.nodeType === Node.ELEMENT_NODE) {
       const rect = screenBounds.getBoundingClientRect();
       screenBounds = new Box(
         rect.left || rect.x,
diff --git a/node_modules/@tldraw/editor/dist-esm/lib/utils/dom.mjs b/node_modules/@tldraw/editor/dist-esm/lib/utils/dom.mjs
index 1578001..bc614a5 100644
--- a/node_modules/@tldraw/editor/dist-esm/lib/utils/dom.mjs
+++ b/node_modules/@tldraw/editor/dist-esm/lib/utils/dom.mjs
@@ -1,6 +1,6 @@
 import { debugFlags, pointerCaptureTrackingObject } from "./debug-flags.mjs";
 function loopToHtmlElement(elm) {
-  if (elm instanceof HTMLElement) return elm;
+  if (elm.nodeType === Node.ELEMENT_NODE) return elm;
   if (elm.parentElement) return loopToHtmlElement(elm.parentElement);
   else throw Error("Could not find a parent element of an HTML type!");
 }
```

[...]

PR #6338, opened 2025-06-27
fix: resolve cross-realm compatibility issues

### LINKED ISSUE #6337: [Bug]: Keyboard shortcuts do not work cross-realm
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `tldraw@3.13.3` for the project I'm working on.

This is another cross-realm issue similar to the ones reported in https://github.com/tldraw/tldraw/issues/5728.

The most recent tldraw upgrade broke our app because we were using built-in shapes with the `TldrawEditor` component. After upgrading we got errors that translations were missing because the translation context wasn't in our React component tree. We didn't find a good way to make it work via the public APIs so we switched to using the `Tldraw` component instead. That shifted our approach from adding things we want on top of `TldrawEditor` to removing things we don't want from `Tldraw`. We previously used our own undo/redo handlers but now we use the built-in ones. The built-in ones didn't work cross-realm but with the change below they now work.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/tldraw/dist-esm/lib/ui/hooks/useKeyboardShortcuts.mjs b/node_modules/tldraw/dist-esm/lib/ui/hooks/useKeyboardShortcuts.mjs
index 1bf365f..82439e9 100644
--- a/node_modules/tldraw/dist-esm/lib/ui/hooks/useKeyboardShortcuts.mjs
+++ b/node_modules/tldraw/dist-esm/lib/ui/hooks/useKeyboardShortcuts.mjs
@@ -1,6 +1,7 @@
 import {
   isAccelKey,
   preventDefault,
+  useContainer,
   useEditor,
   useValue
 } from "@tldraw/editor";
@@ -18,6 +19,7 @@ const SKIP_KBDS = [
   "asset"
 ];
 function useKeyboardShortcuts() {
+  const container = useContainer();
   const editor = useEditor();
   const isReadonlyMode = useReadonly();
   const actions = useActions();
@@ -27,13 +29,13 @@ function useKeyboardShortcuts() {
     if (!isFocused) return;
     const disposables = new Array();
     const hot = (keys, callback) => {
-      hotkeys(keys, { element: document.body }, callback);
+      hotkeys(keys, { element: container.ownerDocument.body }, callback);
       disposables.push(() => {
         hotkeys.unbind(keys, callback);
       });
     };
     const hotUp = (keys, callback) => {
-      hotkeys(keys, { element: document.body, keyup: true, keydown: false }, callback);
+      hotkeys(keys, { element: container.ownerDocument.body, keyup: true, keydown: false }, callback);
       disposables.push(() => {
         hotkeys.unbind(keys, callback);
       });
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

ds300:
A few small changes to make tldraw work in cross-realm based on suggestions by @derekcicerone

Fixes #6337 and partially #5728 

### Change type

- [x] `bugfix` 

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix a few code paths that didn't work for cross-realm situations.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -3662,7 +3662,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	updateViewportScreenBounds(screenBounds: Box | HTMLElement, center = false): this {
-		if (screenBounds instanceof HTMLElement) {
+		if (!(screenBounds instanceof Box)) {
 			const rect = screenBounds.getBoundingClientRect()
 			screenBounds = new Box(
 				rect.left || rect.x,
--- a/packages/editor/src/lib/utils/dom.ts
+++ b/packages/editor/src/lib/utils/dom.ts
@@ -18,7 +18,7 @@ import { debugFlags, pointerCaptureTrackingObject } from './debug-flags'
 
 /** @public */
 export function loopToHtmlElement(elm: Element): HTMLElement {
-	if (elm instanceof HTMLElement) return elm
+	if (elm.nodeType === Node.ELEMENT_NODE) return elm as HTMLElement
 	if (elm.parentElement) return loopToHtmlElement(elm.parentElement)
 	else throw Error('Could not find a parent element of an HTML type!')
 }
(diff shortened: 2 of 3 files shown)


mimecuvalo (approved):
👍 feels like we'll be playing whack-a-mole here without a linter rule (as mentioned in https://github.com/tldraw/tldraw/issues/5728 ) but it's all good.

derekcicerone:
Awesome, thanks! We have a linter rule for these that catches most but not all issues. For example we block all direct references to the global document, navigator, and window objects. We have a React context which components can use to get the correct versions of those objects. But we still find instanceof issues via manual testing. I suppose we could have a rule for those too but it would be a bit tricker because there’s no readily available list for all the things to block.

Issue #5095, opened 2024-12-09
[Bug]: Copying images from the board gets slower the more images you have


jorgen-braseth-adsk:
### What happened?

If you add a lot (100+) of images to a board, copying a single image FROM the board starts to slow down. Depending on the size of the images, it can take several seconds, or even completely crash the browser tab.

### How can we reproduce the bug?

1. Put a lot of images on a board - the larger the images, the more prominent the error.
2. Copy a single image
3. Observe that the tab freezes for a time while copying.

The issue happens both on our own solution built on the SDK and on TLDraw.com

This actual time consumed seems to be inside the `compressToBase64` function in the `ls-string` library, where we see the nested call to `compress` seemingly being called more times (once for each image on the board?) the more images you have on the board. The individual calls to compress aren't too bad, but the total adds up:

<img width="1355" alt="image" src="https://github.com/user-attachments/assets/aeb1bfd0-ad14-47d0-9eb0-d4d610d5ecd1">
<img width="730" alt="image" src="https://github.com/user-attachments/assets/3928201a-921b-406e-bebb-02257fbb9fa3">

### What browsers are you seeing the problem on?

Chrome

### Where did this happen?

The tldraw.com website

### Contact Details

jorgen.aars.braseth@autodesk.com

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mimecuvalo:
Thanks for the bug report @jorgen-braseth-adsk ! There's probably some optimization we could do here for sure.
I'm having trouble reproducing the specific point of "Copy a single image" though. I see the issue happening when trying to copy all of them - but I'm surprised about the callout of it happening when a single image is being copied. Is there maybe an example board you could share that exhibits the issue?

jorgen-braseth-adsk:
I can no longer reproduce this by copying a single issue. The issue persists, as you say, when copying multiple images.
This board is an example: https://www.tldraw.com/r/FhQTCgTCNzRu2F7c6XNlr?d=v-6467.-3785.9362.4398.page
Incidentally: creating a shared board with these images on it also took a very long time. (1min+)

jorgen-braseth-adsk:
We have observed that the board in the link above is exceptionally slow and copying all images freezes the tab for about 10 seconds on my machine (Chrome on Macbook M2), while on @trygve-aaberge-adsk 's machine (Chromium on Arch Linux, Thinkpad X1Carbon Gen9) it works fine.

mimecuvalo:
[...]

PR #6344, opened 2025-06-30
clipboard: only compress non-assets for greatly improved perf

mimecuvalo:
fixes https://github.com/tldraw/tldraw/issues/5095

I created a versioning scheme so that this doesn't break copy/paste from other tldraw boards. Adding the versioning will help any future changes, if any.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improve perf of copy+paste when copying lots of large images.

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -352,25 +352,57 @@ async function handleClipboardThings(editor: Editor, things: ClipboardThing[], p
 
 							if (tldrawHtmlComment) {
 								try {
-									// If we've found tldraw content in the html string, use that as JSON
-									const jsonComment = lz.decompressFromBase64(tldrawHtmlComment)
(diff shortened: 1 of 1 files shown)


steveruizok:
Do we really need to compress anything here?

mimecuvalo:
> Do we really need to compress anything here?

Yeah, you know it's maybe debatable.
I did check and, with say, 2000 shapes on the board copying the shapes make the clipboard go from 4MB → 1MB.
But then again ¯\\_(ツ)_/¯ we're just saving some MB on the clipboard which doesn't really matter. Like, one image on it's own is easily 4MB.

Compressing 2000 shapes (without assets) on my pretty nice machine takes ~500ms which I could see taking longer on a slower machine.
So, yeah I could see nixing compressing altogether.

steveruizok:
let's go!

steveruizok (approved):

PR #6347, opened 2025-06-30
Simplify UI of education example

TodePond:
This PR removes the pages menu and most tools from the education example to simplify it.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


steveruizok:
Better IMO to remove the actions using the overrides, so that the keyboard shortcuts don't work either!

TodePond:
good shout, cursor updated the PR for me

PR #6353, opened 2025-06-30
Fix geo shape checkmark selection bug

steveruizok:
Fixes a bug where the checkmark on a filled geo shape (`check-box`) was not selectable when positioned in front of another filled shape.

The issue stemmed from the hit testing logic in `Editor.ts` not correctly excluding internal, unfilled geometries (like the checkmark lines) when determining which shape was clicked. This PR modifies the hit testing to skip these internal geometries, ensuring the main shape is correctly selected.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1.  **Filled checkbox over filled shape:**
    *   Create a filled `check-box` geo shape.
    *   Place it in front of another filled shape.
    *   Clicking on the area where the checkmark would be should select the `check-box` geo shape, not the background shape.
2.  **Unfilled checkbox over filled shape:**
    *   Create an unfilled `check-box` geo shape.
    *   Place it in front of another filled shape.
    *   Clicking on the border of the `check-box` should select it.
    *   Clicking inside the empty center of the `check-box` (where the checkmark would be) should select the background shape.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where the checkmark on a filled geo shape (`check-box`) was not selectable when positioned in front of another filled shape.

### DIFF
--- a/checkbox-hit-testing-fix.md
+++ b/checkbox-hit-testing-fix.md
@@ -1,0 +1,32 @@
+# Checkbox Geo Shape Hit Testing Fix
+
+## Problem
+The checkmark on geo shapes with `geo: 'check-box'` was not selectable when the shape was filled and positioned in front of another filled shape. This was because the internal geometry (checkmark lines) was being included in hit testing even though it couldn't be clicked on.
+
+## Root Cause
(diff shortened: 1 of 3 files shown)


steveruizok:
bad bot

PR #6363, opened 2025-07-01
a11y: add keyboard shortcuts preference

mimecuvalo:
From the a11y audit:

> Currently, there are keyboard shortcuts to easily jump to certain functionality. For example, D activates Darwing, O activates an Oval shape, N for a new Note, cmd +, Option V aligns vertically, and so on.
> 
> This is useful and allows for quick keyboard navigation (good), but one of the following should be true for the shortcuts that consist of only alphanumeric characters (N, O, etc) to include one or more non-printable keys (like alt, ctrl, shift, etc):
> 
> It should be possible to turn off keyboard shortcuts
> 
> It should be possible to remap keyboard shortcuts that only consist of 
> 
> The reason for this is that those are easy to trigger, especially by voice controls, and might interfere with other assistive technologies that come with their own shortcuts. People who experience tremors or other impairments related to coordination might also accidentally press keys on their keyboard and accidentally trigger a keyboard shortcut by doing so.
> 
> ### Suggested fixes
> 
> Several options:
> 
> Option A: The quickest fix would be to provide a mechanism (that's accessible and easily discoverable) to toggle all the keyboard shortcuts off or on. That way, those who experience issues with them, can just turn them off. They won't get the benefit of the easy controls, but the controls won't get in the way of their work either.
> 
> Option B: If keyboard shortcuts are an important feature, then allowing them to be remapped to include non-printable characters as well is the more user-sensitive solution. Someone who for example accidentally keeps hitting the N key while using the trackpad (and by doing so creating a new note), will be able to still benefit from a shortcut (quickly creating a new note), but without the accidental triggering.
> 
> WCAG requires this just for the shortcuts consisting of only alphanumeric/printable characters. But from a UX point of view, it might make more sense to allow the remapping/turning off for all shortcuts, to keep it consistent.
> 
> Also remember: These settings should be remembered when revisiting the page. A user who needs them to be turned off or remapped, should not have to go through this flow each time they re-open the page.
> 
> ### Related WCAG
> 
> [SC 2.4.1 Character key shortcuts (A)](https://www.w3.org/TR/WCAG22/#character-key-shortcuts): If a keyboard shortcut is implemented in content using only letter, punctuation, number, or symbol characters, then a mechanism is available to turn the shortcut off, or a mechanism is available to remap the shortcut to include one or more non-printable keyboard keys.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


mimecuvalo:
also i'm gonna kvetch here: it's wild how many files it takes to wire up 1 preference 😆

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -48,6 +48,7 @@ export interface TLUiActionItem<
 	label?: TransationKey | { [key: string]: TransationKey }
 	readonlyOk?: boolean
 	checkbox?: boolean
+	isRequiredA11yAction?: boolean
```
MitjaBezensek, line 52: I really don't like optional keys since they make these features so hard to discover. If I'll add a new action there's almost no way I'd know I might need to add this as well. Prefer some duplication over convenience here.

apps/dotcom/client/e2e/fixtures/Database.ts
```diff
@@ -30,6 +30,7 @@ const defaultUser = {
 	flags: '',
 	locale: null,
 	animationSpeed: null,
+	keyboardShortcuts: true,
```
MitjaBezensek: The name makes me think we store actual keyboard shortcuts. Maybe `keyboardShortcutsEnabled` or something like that? In some cases we also use the `isSomething` naming (`isDarkMode`).

apps/dotcom/zero-cache/migrations/019_add_keyboard_shortcuts_pref.sql
```diff
@@ -0,0 +1,8 @@
+-- Purpose: Add an allowAnalyticsCookie column to the user table that indicates whether the user has
```
MitjaBezensek: Remove / update the comment?

packages/editor/src/lib/editor/managers/UserPreferencesManager/UserPreferencesManager.ts
```diff
@@ -75,6 +76,12 @@ export class UserPreferencesManager {
 		return this.user.userPreferences.get().animationSpeed ?? defaultUserPreferences.animationSpeed
 	}
 
+	@computed getKeyboardShortcuts() {
```
MitjaBezensek: Same thought on the naming. Do we return the shortcuts?

apps/dotcom/client/e2e/fixtures/Database.ts
mimecuvalo: Ah good call 👍

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: Yeah, it's a good point, however I'd say maybe not, given the context here. I mean, honestly, I have the same problem with `readonlyOk`, I feel like we've forgotten in the past to add it and only think of it later, seems like a footgun for `readonlyOk`. On the other hand, these a11y actions will be a _very_ limited set of things — I don't expect the set to change, it really is meant to only be the small subset of things just so it keeps working.

MitjaBezensek: Yeah, `readonlyOk` is really problematic. You could not log out from the dotcom in readonly rooms because of it [being optional ](https://github.com/tldraw/tldraw/pull/6199)across the codebase🤦‍♂️

I'm fine with it going either way, I just prefer explicitness. Like if the sdk users will add actions will they even know about this prop? I think not, even though they might add some actions the would require setting it to true 🤷‍♂️

MitjaBezensek (approved):

PR #6371, opened 2025-07-02
fix(svg): restore text shadows and fix outline widths

steveruizok:
This PR fixes the text shadows on SVG exports.

It:
- restores text shadows on rich text svgs
- fixes the width of outlines on plain text svgs

image

image

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Fixed a bug with text outlines not appearing for rich text in SVGs
- Fixed stroke width for plain text in SVGs


ds300 (dismissed):
this accidentally applied the border to highlighted text which is not supposed to have a border (it looks so bad lol)

PR #6377, opened 2025-07-02
Remove license key from dotcom

MitjaBezensek:
Think we had this so that we didn't show missing license outputs for development and dotcom. I think this no longer applies as we now allow using the package without a license while showing the watermark (which is what we do on dotcom anyway).

@mimecuvalo the only consideration is that we will hit the watermark track, but I guess we filter our own webpage?

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/client/src/components/LocalEditor.tsx
+++ b/apps/dotcom/client/src/components/LocalEditor.tsx
@@ -1,4 +1,3 @@
-import { getLicenseKey } from '@tldraw/dotcom-shared'
 import { ReactNode } from 'react'
 import { Editor, TLComponents, Tldraw, TldrawOptions, useEvent } from 'tldraw'
 import { SneakyToolSwitcher } from '../tla/components/TlaEditor/sneaky/SneakyToolSwitcher'
@@ -38,7 +37,6 @@ export function LocalEditor({
 	return (
 		<div className="tldraw__editor" data-testid={dataTestId}>
 			<Tldraw
-				licenseKey={getLicenseKey()}
 				assetUrls={assetUrls}
 				persistenceKey={persistenceKey ?? getScratchPersistenceKey()}
 				onMount={handleMount}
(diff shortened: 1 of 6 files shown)


mimecuvalo:
> @mimecuvalo the only consideration is that we will hit the watermark track, but I guess we filter our own webpage?

this is what currently happens though, so it keeps parity in that regard.
if anything, we want to track when logged out too (although might be too much traffic). actually now i'm still confused on why it's not sending a signal when logged out 🤔

MitjaBezensek:
> > @mimecuvalo the only consideration is that we will hit the watermark track, but I guess we filter our own webpage?
> 
> this is what currently happens though, so it keeps parity in that regard. if anything, we want to track when logged out too (although might be too much traffic). actually now i'm still confused on why it's not sending a signal when logged out 🤔

I think it's due to these lines: https://github.com/tldraw/tldraw/blob/4bdb4828ecf8536685b7fbf7d0fbcdab960c772c/packages/editor/src/lib/license/LicenseManager.ts#L95-L97

[LocalEditor](https://github.com/tldraw/tldraw/blob/4bdb4828ecf8536685b7fbf7d0fbcdab960c772c/apps/dotcom/client/src/components/LocalEditor.tsx#L41) (which is used when logged out) sets the license, so it is not unlicensed and it does not send the tracking. [TlaEditor](https://github.com/tldraw/tldraw/blob/4bdb4828ecf8536685b7fbf7d0fbcdab960c772c/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx#L253-L271) on the other hand does not set the license, so it is unlicensed and it does send it.

mimecuvalo (commented):
I think after sleeping on this, that maybe I'm changing my mind: we _shouldn't_ remove this.

I'm remembering the reason we added this to the dotcom in the first place: we need to keep dogfooding our license. We saw this in action yesterday when we noticed some of it was busted. We should keep dotcom as something that actively consumes the license so that we're aware of issues downstream before they reach customers.

MitjaBezensek:
> I think after sleeping on this, that maybe I'm changing my mind: we _shouldn't_ remove this.
> 
> I'm remembering the reason we added this to the dotcom in the first place: we need to keep dogfooding our license. We saw this in action yesterday when we noticed some of it was busted. We should keep dotcom as something that actively consumes the license so that we're aware of issues downstream before they reach customers.

Sounds good. I'll then go ahead and add the license to all the other editors. Will go ahead and close this as it's easier to start from scratch.

PR #6379, opened 2025-07-02
New release process (again)

ds300:
Follow up to #6365 which was reverted

This time we keep the package.json and version.ts version numbers up to date as before. I think the cleanest way to do this on the main branch is to create a new action that can be triggered after each `latest` release on npm.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


.github/workflows/bump-versions.yml
```diff
@@ -0,0 +1,53 @@
...
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  bump_versions:
+    name: 'Bump package versions to latest tldraw version'
+    environment: npm deploy
+    timeout-minutes: 30
+    runs-on: ubuntu-latest-16-cores-open
```
mimecuvalo, line 21: this doesn't need the beefy 16 cores, right?

PR #6051, opened 2025-05-12
tsc: typecheck our templates as well

mimecuvalo:
related to https://github.com/tldraw/tldraw/issues/5970#issuecomment-2867749835

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

PR #6325, opened 2025-06-26
arrows: add rich text

mimecuvalo:
We've been debating this internally whether to add rich text for arrow labels. After some soul searching, I personally think it would make sense to add it in. This PR enables it! @steveruizok lemme know your thoughts.

Screenshot 2025-06-26 at 12 09 03

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add rich text for arrow labels.

PR #6381, opened 2025-07-02
Make templates first-class members of our monorepo

SomeHats:
This diff aims to make working on our templates a first-class part of our monorepo in preparation for the starter kits work we've been talking about.

There are a few big differences here:
- Templates are now a part of the yarn workspace. Their dependencies are installed with the others
- They're typechecked alongside everything else in the repo
- They run against the in-repo versions of our packages instead of the release versions
- There's a new release script that hopefully will make adding new starter kits really easy
- They're linted by our monorepo lint tooling
- There's a new top-level `yarn dev-template <template name>` script to make it easy to start work on a template.

This has resulted in a few side effects for the examples themselves:
- There are some config options needed only when in the monorepo. These can be wrapped in special comments to exclude them from the published versions
- There's a special declaration in package.json that describes how they should be published, including removing certain things that we don't want published
- The published examples no longer include linting
- I simplified the tsconfig setups of several of them. They now all use a single tsconfig file, which fits better with our system. 
- Simplifying tsconfig prompted some other changes - i switched the cloudflare sync example over to use the vite cloudflare plugin instead of separate vite/wrangler builds.
- I fixed some mistakes, type errors, and linting issues that enabling these checks exposed

I also fixed up / deleted some unrelated cruft & mistakes that i think come from vibe-coding. Working on this caused a few dependencies to get updated when the lockfile was invalidated (prettier, typescript) so I've upgraded the package.json files across the board to reflect the newer versions.

There's some prior art in #6051 which aimed to integrate the templates with our type checking while keeping them outside the rest of the monorepo flow. This diff goes a fair bit further.

### Change type

- [x] `other`


apps/dotcom/client/package.json
```diff
@@ -67,18 +67,16 @@
...
-		"eslint-plugin-formatjs": "^5.2.13",
 		"fast-glob": "^3.3.3",
 		"identity-obj-proxy": "^3.0.0",
 		"json5": "^2.2.3",
 		"kysely": "^0.27.5",
 		"lazyrepo": "0.0.0-alpha.27",
 		"pg": "^8.13.1",
 		"regexgen": "^1.3.0",
-		"vite": "^5.4.14",
+		"vite": "^7.0.1",
```
mimecuvalo, line 79: decently big version jump. any upgrade concerns?

packages/editor/src/lib/license/LicenseManager.test.ts
```diff
@@ -417,7 +417,7 @@ function importPrivateKey(pemContents: string) {
 	// base64 decode the string to get the binary data
 	const binaryDerString = atob(pemContents)
 	// convert from a binary string to an ArrayBuffer
-	const binaryDer = str2ab(binaryDerString) as Uint8Array
```
mimecuvalo, line 420: how did you find this tweak? were you looking for extra casting?

packages/tldraw/src/test/Editor.test.tsx
```diff
@@ -899,3 +899,70 @@ describe('the geometry cache', () => {
 		expect(editor.getShapePageBounds(A)!.width).toBe(200)
 	})
 })
+describe('editor.getShapePageBounds', () => {
```
mimecuvalo, line 902: why new tests?

mimecuvalo: oic nvm

templates/ai/worker/do/openai/getTldrawAiChangesFromSimpleEvents.ts
```diff
@@ -82,7 +86,7 @@ function getTldrawAiChangesFromSimpleCreateOrUpdateEvent(
 					x: shape.x,
 					y: shape.y,
 					props: {
-						richText: toRichText(shape.text ?? ''),
+						richText: toRichTextIfNeeded(shape.text ?? ''),
```
mimecuvalo, line 89: shape.text is rich text sometimes? 🤔

templates/ai/worker/do/openai/schema.ts
```diff
@@ -119,7 +124,7 @@ const SimpleArrowShape = z.object({
 	x2: z.number(),
 	y2: z.number(),
 	color: SimpleColor.optional(),
-	text: SimpleLabel.optional(),
+	text: z.string().optional(),
```
mimecuvalo, line 127: i'm considering adding rich text to arrows btw: https://github.com/tldraw/tldraw/pull/6325
maybe this should be SimpleLabel just in case?

templates/sync-cloudflare/worker-configuration.d.ts
```diff
@@ -0,0 +1,8202 @@
+/* eslint-disable */
```
mimecuvalo, line 1: omg this was crashing github code review — thx @SomeHats :P

mimecuvalo (approved):
great stuff! just a couple minor comments.

my only request for future PRs, please let's do prettier changes on separate PRs :D
makes it hard to separate out the noise hahaha — thx!

apps/dotcom/client/package.json
SomeHats: nope - i tested both these out. 6.0 mostly bought a some new APIs around SSR that we don't use, and 7.0 mostly removed some deprecated stuff we don't use

packages/editor/src/lib/license/LicenseManager.test.ts
SomeHats: with typescript getting bumped this became an error

templates/ai/worker/do/openai/getTldrawAiChangesFromSimpleEvents.ts
SomeHats: yeah this stuff is a nightmare. these "simple" shape definitions sometimes use rich text (e.g. geo does, arrow doesnt), and we can't convert rich -> plain cus this runs on the server and we don't have access to our tiptap stuff there :|

templates/ai/worker/do/openai/schema.ts
SomeHats: that means we have to handle rich -> plain conversion on the server which we can't do right now

templates/sync-cloudflare/worker-configuration.d.ts
SomeHats: sorry this is what cloudflare has decided the new way of doing worker types is 😅

apps/dotcom/client/package.json
mimecuvalo: thanks! appreciate the due dilligence here 👍

PR #6382, opened 2025-07-02
analytics: bring back reo; add redaction to reo+posthog

mimecuvalo:
- this brings back Reo but within an iframe to only send the iframe url.
- also, redact the URL in posthog as well

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/utils/analytics.tsx
```diff
@@ -32,6 +32,20 @@ const shouldUsePosthog = POSTHOG_KEY !== undefined
 const GA4_MEASUREMENT_ID: string | undefined = import.meta.env.VITE_GA4_MEASUREMENT_ID
 const shouldUseGA4 = GA4_MEASUREMENT_ID !== undefined
 
+const PROPERTIES_TO_REDACT = ['url', 'href', 'pathname']
```
MitjaBezensek, line 35: Maybe use a set here? Would make the check below more succinct.

MitjaBezensek (approved):

apps/dotcom/client/src/utils/analytics.tsx
mimecuvalo: yeah the check below isn't so obvious, i'll add a comment. it's because posthog sends up events like this:
`$pathname` (posthog events) but also possibly something like `pathname` (custom events).

PR #6394, opened 2025-07-04
Reduce hooks in useValue

steveruizok:
This PR simplifies `useValue` (in the same spirit of `useReactor`). Previously, we were using two memos and a ref; now it uses one memo.

### Change type

- [x] `improvement`

### Release notes

- Small improvement to the performance on `useValue`.

### DIFF
--- a/packages/state-react/src/lib/useValue.ts
+++ b/packages/state-react/src/lib/useValue.ts
@@ -1,6 +1,6 @@
 /* eslint-disable prefer-rest-params */
 import { Signal, computed, react } from '@tldraw/state'
-import { useMemo, useRef, useSyncExternalStore } from 'react'
+import { useMemo, useSyncExternalStore } from 'react'
 
 /** @public */
 export function useValue<Value>(value: Signal<Value>): Value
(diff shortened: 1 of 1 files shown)


packages/state-react/src/lib/useValue.ts
```diff
@@ -49,38 +49,36 @@ export function useValue() {
 	const deps = args.length === 3 ? args[2] : [args[0]]
 	const name = args.length === 3 ? args[0] : `useValue(${args[0].name})`
 
-	const isInRender = useRef(true)
-	isInRender.current = true
+	let isInRender = true
```
ds300, line 52: this is not equivalent to using the useRef, and it won't work as intended because this let binding is scoped to each render individually while a useRef is scoped to the component's lifetime. setting isInRender to true will have no effect on a computed that was created during a previous render.

ds300 (changes requested):

packages/state-react/src/lib/useValue.ts
steveruizok: I might be getting tripped up on react lifecycle here. I can drop this back to use the ref.

You've got a unit test for this edge case though, right? It passes on this PR, which is weird if the ref is needed.

ds300: your change here actually affects untested behvaior. namely that the useValue should throw synchronously during render if the computed throws an error and the component is still alive.

And now that I try to test this I realise it's not working as expected. lemme see if i can come up with a fix and I'll make an alternative pr to this

steveruizok:
Much better in #6405 🙌

PR #6395, opened 2025-07-04
feat(api): improve indicators API by moving logic to state nodes

steveruizok:
This PR solves a long-standing todo, which replaces the hard-coded list of "show indicators" to state nodes instead. As a result, we can remove the `<TldrawShapeIndicators>` component that provided that logic.

### Change type

- [x] `api`

### Test plan

1. Verify that shape indicators still appear correctly in various tool states (Idle, Brushing, etc.)
2. Verify that indicators are hidden when expected based on the new `shouldShowIndicators()` logic.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- SDK: Added `shouldShowIndicators()` to `StateNode`, used to configure whether indicators should be shown when in the state

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -40,6 +40,10 @@ export class Idle extends StateNode {
 
 	selectedShapesOnKeyDown: TLShape[] = []
 
+	override getShouldShowIndicators() {
+		return true
+	}
+
 	override onEnter() {
 		this.parent.setCurrentToolIdMask(undefined)
 		updateHoveredShapeId(this.editor)


packages/editor/src/lib/components/default-components/DefaultShapeIndicators.tsx
```diff
@@ -28,27 +28,28 @@ export const DefaultShapeIndicators = memo(function DefaultShapeIndicators({
...
+
+			const path = editor.getPath()
+			const nodeIds = path.split('.')
+			for (let i = 0; i < nodeIds.length; i++) {
+				const subpath = nodeIds.slice(0, i).join('.')
+				const stateNode = editor.getStateDescendant(subpath)
+				if (!stateNode?.getShouldShowIndicators()) {
+					return prev
+				}
+			}
```
steveruizok: The main new logic

packages/editor/api-report.api.md
```diff
@@ -2735,6 +2735,7 @@ export abstract class StateNode implements Partial<TLEventHandlers> {
     getCurrentToolIdMask(): string | undefined;
     getIsActive(): boolean;
     getPath(): string;
+    getShouldShowIndicators(): boolean;
```
SomeHats, line 2738: For similar flags in e.g. shape utils we drop the `get` and just do `canDoX()` etc.

For things like this where the function name is unambiguously returning a boolean and not performing an action, I think we can drop the `get` and have a shorter, easier to read name

packages/editor/src/lib/components/default-components/DefaultShapeIndicators.tsx
```diff
@@ -28,27 +29,31 @@ export const DefaultShapeIndicators = memo(function DefaultShapeIndicators({
...
 
-			const instanceState = editor.getInstanceState()
+			let shouldShowIndicators = false
+			let state: StateNode | undefined = editor.getCurrentTool()
+			// allow deeper state nodes to override the should show indicators value of shallower nodes
+			while (state) {
+				shouldShowIndicators = !!state.getShouldShowIndicators()
+				state = state.getCurrent()
+				if (!state) break
+			}
```
SomeHats, line 43: i'd be tempted to make this a method on `editor` or the root state node for easier reuse. wdyt?

steveruizok:
Closing due to merge conflicts

PR #6396, opened 2025-07-04
clipboard: fix up html entities being escaped

### LINKED ISSUE #6428: Pasting text with angle brackets crashes canvas
### What happened?

When copy-pasting a textbox that contains text in angle brackets (e.g. `example 123 <bracketed> 456`), the content in the angle brackets is removed. If the text is entirely in angle brackets (e.g. `<example 123>`) and it is pasted, the canvas crashes.

the text object that is copied
Image

after pasting
Image

1. Make a text object.
2. Type in `<text>`.
3. Copy-paste the object (canvas crashes at this step).

Firefox

### Where did this happen?

The tldraw.com website

### Contact Details

_No response_

### Code of Conduct

mimecuvalo:
followup fix for https://github.com/tldraw/tldraw/pull/6344
because we don't base64 encode anymore, a knock-on effect is that any characters within this html blob gets treated as such. we need to unescape these upon paste.

hat-tip to Phil for the report.

Screenshot 2025-07-04 at 17 40 56

this also fixes #6428

Image

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- clipboard: fix up html entities being escaped

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -355,12 +355,7 @@ async function handleClipboardThings(editor: Editor, things: ClipboardThing[], p
 									// First try parsing as plain JSON (version 2 format)
 									let json
 									try {
-										json = JSON.parse(
-											tldrawHtmlComment
-												.replace(/&gt;/g, '>')
-												.replace(/&lt;/g, '<')
-												.replace(/&amp;/g, '&')
-										)
+										json = JSON.parse(tldrawHtmlComment)
 									} catch {
 										// Fall back to LZ decompression (legacy format)
 										const jsonComment = lz.decompressFromBase64(tldrawHtmlComment)
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -355,7 +355,12 @@ async function handleClipboardThings(editor: Editor, things: ClipboardThing[], p
 									// First try parsing as plain JSON (version 2 format)
 									let json
 									try {
-										json = JSON.parse(tldrawHtmlComment)
+										json = JSON.parse(
+											tldrawHtmlComment
+												.replace(/&gt;/g, '>')
+												.replace(/&lt;/g, '<')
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -355,7 +355,12 @@ async function handleClipboardThings(editor: Editor, things: ClipboardThing[], p
 									// First try parsing as plain JSON (version 2 format)
 									let json
 									try {
-										json = JSON.parse(tldrawHtmlComment)
+										json = JSON.parse(
+											tldrawHtmlComment
+												.replace(/&gt;/g, '>')
+												.replace(/&lt;/g, '<')
+												.replace(/&amp;/g, '&')
+										)
```
SomeHats: We have some issues here at the moment: 

Image

Looking at the code for actually copying this stuff, it seems like we're not actually explicitly html-encoding these things ourselves. Where does the encoding come from? If we're using a regex here instead of a more robust html escaping system, why were these three chosen? I would have expected to see the reverse of this encoding explicitly happening somewhere

SomeHats:
btw can we sdk + dotcom hotfix this when it's ready?

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
mimecuvalo: had a discussion with Alex offline. annoyingly we need to go back to the original revision of this:
https://github.com/tldraw/tldraw/pull/6344/commits 
we still need to base64 encode the regular non-assets json.

ugh, this means we need a version 3...

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -584,13 +597,21 @@ const handleNativeOrMenuCopy = async (editor: Editor) => {
 		return
 	}
 
-	// Version 2: Don't compress anything.
-	const stringifiedClipboard = JSON.stringify({
+	// Use versioned clipboard format for better compression
+	// Version 2: Don't compress assets, only compress other data
```
SomeHats: version 3?

PR #6404, opened 2025-07-07
Faster User DO Resume

ds300:
Our deploys have been overloading the replicator for a minute or two at busy times. I think the new history-log-based resume code (introduced in https://github.com/tldraw/tldraw/pull/6184) is significantly slower than what we had before.

This PR aims to reduce the amount of work the replicator needs to do during deploys by optimizing for the fairly common case that no relevant updates occurred during the brief window when active users' DOs were rebooting. If we store the sequence ID + number in the user DO snapshot, then we can check whether the user DO is up to date much much more quickly when registering.

This should do a lot to help ease the load on the replicator during a deploy 🤞🏼  

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -546,7 +546,7 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 		return { sequenceId: this.slotName }
 	}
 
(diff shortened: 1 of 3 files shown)


ds300:
if this works as well as I think it will we could probably actually completely remove the history log and the smartass fast-forwarding logic. especially now that our hard reboot 'fetch everything' query is a couple of orders of magnitude faster.

PR #6411, opened 2025-07-08
Show max shapes alert when duplicating shapes.

MitjaBezensek:
`editor.duplicateShapes` already does the check so we don't need to do an early return.

### Change type

- [x] `improvement`

### Release notes

- Alert that maximum shapes were reached when duplicating shapes. Until now we just silently rejected the duplication action.


mimecuvalo:
feels like we're playing whack-a-mole still with these - would love to solve this deeper down one day! 
but this is good to go

PR #6412, opened 2025-07-08
Workflow builder starter kit

SomeHats:
Adds a workflow builder template. See the README for more information.

### Change type

- [x] `other`

### API Changes

- `TldrawUiToolbar` now accepts an optional `orientation` prop, which can be either horizontal or vertical. It's horizontal by default.


SergioRiquelmeSanchez:

PR #6414, opened 2025-07-08
feat(examples): add math shape proof-of-concepts

TodePond:
This PR adds two math shape examples as proof-of-concept. This isn't intended to be merged. It's just so that I can share it with other developers who might be interested in building custom shapes like this.

- The `math-shape-mathlive` example uses the MathLive library. Try it [here](https://examples-git-lu-math-shape-tldraw.vercel.app/math-shape-mathlive/full).
- The `math-shape` example uses the MathQuill library. Try it [here](https://examples-git-lu-math-shape-tldraw.vercel.app/math-shape/full). It uses the `react-mathquill` package. Note: Out-of-the-box, MathQuill has some rendering issues with vertical positioning when not at 100% zoom. To fix this, I've disabled some of its dynamic vertical positioning, which means that nested indices don't grow as much as they should. However, I think this is preferred to the default, which involves things moving around the screen. If it was up to me, I would use the MathLive library instead, or if I wanted to stick with MathQuill, I would consider forking it and/or patching it. I think this is what the Desmos team does.
- For another example of math shapes in tldraw, check out community member Max Whittaker's example repo [here](https://github.com/jungian6/TLDraw_MathsDemo).

If I were to take these further, I'd polish up interactions and UX papercuts, and I'd hone in on what features to offer within menus. There's also some unexplored MathLive features that I disabled for the time being, eg: The virtual keyboard. 

### Change type

- [x] `other` 

### Test plan

1. Open the math-shape or math-shape-mathlive examples.
2. Interact with the math input fields.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added proof-of-concept math shape examples using MathLive and MathQuill.

### DIFF
--- a/apps/examples/src/examples/math-shape-mathlive/MathLiveExample.tsx
+++ b/apps/examples/src/examples/math-shape-mathlive/MathLiveExample.tsx
@@ -5,7 +5,6 @@ import { useCallback, useEffect, useRef } from 'react'
 import {
 	DefaultToolbar,
 	DefaultToolbarContent,
-	Editor,
 	Geometry2d,
 	HTMLContainer,
 	RecordProps,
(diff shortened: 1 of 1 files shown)


steveruizok:
Reference PR

PR #6417, opened 2025-07-08
Fix diagonal arrow issue

### LINKED ISSUE #6391: Arrow positioning bug at specific positions
### What happened?

Image

When:
- Adding a perfect circle taking up 1 unit of the grid
- Adding a similar circle in diagonal within 1 unit of distance (either up-right or down-right)
- Adding an arrow from center to center
- The arrow is clipped inside the origin circle

This doesn't happen when the target circle is on the left

_No response_

Firefox

### Where did this happen?

The tldraw.com website

### Contact Details

yleflour@pairprog.io

### Code of Conduct

SomeHats:
This fixes #6391. This turned out to be a very sneaky issue where sometimes, when doing intersections between line segments and polylines, if the line segment crossed the poly line exactly at one of it's vertices, floating point errors would cause the intersection not to be detected. 

The fix here is to use our approximate equality helpers to account for the possibility of floating point errors when working with these intersections.

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests
- [x] End to end tests

### Release notes

- Fix an issue that would cause arrows between circles to sometimes render incorrectly


steveruizok:
Nice catch!

PR #6418, opened 2025-07-09
refactor(dotcom): user DO registration and replicator history

ds300:
we can delete the replicator's history log entirely if we use this fast-resume thingy from #6404 along with conveying registration requests via the postgres wal to avoid race conditions for hard reboots. We also don't need to keep the lsn in the user data syncer any more, and there's no need to update it regularly.

### Change type

- [x] `other`

### Test plan

1. [Describe manual test steps]

- [x] Unit tests (if present)

### Release notes

- Refactor User DO registration to use postgres WAL and remove replicator history log.


ds300:
still a potential thing to do, but let's try to make zero happen first.

PR #6419, opened 2025-07-09
Fix text measurement style leakage

ds300:
A couple of weeks ago steve refactored the text manager to keep a single reusable div around for text measurement, presumably to reduce allocation/gc/dom-insertion/removal overhead? There was an attempt made to 'reset' the styles on the div for each measurement operation, but it wasn't effective.

This PR makes it so that every style that is applied during `measureText` is reset back to its original value afterwards. Resetting every single style might be slightly overkill because some styles are _always_ set. But it doesn't seem to affect perf so 🤷🏼 we can optimize later if need be.

### Change type

- [x] `other`

### Release notes

- Fixed a bug with text measurement where styles could leak from one measurement to the next.


SomeHats (approved):
thanks david!

PR #6423, opened 2025-07-09
Fix fixed-size text measurement

ds300:
Before

Kapture 2025-07-09 at 13 47 50

After

Kapture 2025-07-09 at 13 53 20

### Change type

- [x] `bugfix`

### Release notes

- Fixed a bug where fixed-size text shapes were rendered with the wrong height.

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
@@ -303,30 +303,26 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
 }
 
 function getTextSize(editor: Editor, props: TLTextShape['props']) {
-	const { font, richText, autoSize, size, w } = props
+	const { font, richText, size, w } = props
 
 	const minWidth = 16
 	const fontSize = FONT_SIZES[size]
 
-	const manualWidth = autoSize ? null : Math.max(minWidth, Math.floor(w))
+	const maybeFixedWidth = props.autoSize ? null : Math.max(minWidth, Math.floor(w))
 
 	const html = renderHtmlFromRichTextForMeasurement(editor, richText)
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -305,20 +305,17 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
...
 
-	const minWidth = autoSize ? 16 : Math.max(16, w)
+	const minWidth = 16
 	const fontSize = FONT_SIZES[size]
 
-	const cw = autoSize
-		? null
-		: // `measureText` floors the number so we need to do the same here to avoid issues.
-			Math.floor(Math.max(minWidth, w))
+	const manualWidth = autoSize ? null : Math.max(minWidth, Math.floor(w))
```
mimecuvalo: would be good to keep the comment for future reference

mimecuvalo: nit: `manualWidth` → `maybeManualWidth`

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -329,7 +326,7 @@ function getTextSize(editor: Editor, props: TLTextShape['props']) {
 	}
 
 	return {
-		width: Math.max(minWidth, result.w),
+		width: manualWidth ?? Math.max(minWidth, result.w),
```
mimecuvalo: nit: found it hard to follow the logic here safely (b/c autoSize has a different branch of logic)

suggestion: `autoSize ? Math.max(minWidth, result.w) : maybeManualWidth` to indicate intent better

mimecuvalo (approved):
pre-approving, just some nits

PR #6426, opened 2025-07-09
chore: improve changelog generation

MitjaBezensek:
Pull out api changes. Make sure we parse all the bullets and not just the first one.

### Change type

- [x] `other` 

### Test plan

1. Run the changelog generation script and verify it parses multiple bullets and API changes correctly.

- [ ] Unit tests (if present)

### Release notes

- Improved changelog generation to include API changes and multiple bullet points.

### DIFF
--- a/internal/scripts/lib/auto-plugin.js
+++ b/internal/scripts/lib/auto-plugin.js
@@ -18,9 +18,10 @@ module.exports = class AutoPlugin {
 
 				const parseSection = (sectionHeader) => {
 					// Match the section between the header and the next header or any empty line or end of text.
-					const match = new RegExp(`${sectionHeader}\\n\\n([\\s\\S]*?)(?=\\n### |\\n\\n|$)`).exec(
-						body
-					)
+					const match = new RegExp(
+						`${sectionHeader}\\n\\n([\\s\\S]*?)(?=\\n### |\\n\\n|$)`,
+						'i'
+					).exec(body)
 					if (!match) return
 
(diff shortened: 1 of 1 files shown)


internal/scripts/lib/auto-plugin.js
```diff
@@ -11,15 +11,46 @@ module.exports = class AutoPlugin {
...
+
+					const bulletPoints = match[1]
+						.split('\n')
+						.map((line) => line.trim())
+						.filter((line) => line.startsWith('-'))
+						.map((line) => line.substring(1).trim())
+
+					if (bulletPoints.length === 0) return
+
+					const prefix = sectionHeader === '### API Changes' ? '[API Change]: ' : ''
```
MitjaBezensek: Decided to add this prefix so that we can distinguish which lines are the api changes lines. Don't think we can rely on the `api` label, we don't seem to be using that one at all. Also, some bug fixes / other changes might also include api changes.

internal/scripts/lib/auto-plugin.js
```diff
@@ -11,15 +11,46 @@ module.exports = class AutoPlugin {
...
+					output.push(formattedPoints)
+				}
+
+				let output = []
+
+				parseSection('### Release Notes')
+				parseSection('### API Changes')
+
+				if (output.length > 0) {
+					return output.join('\n')
```
MitjaBezensek, line 50: I guess this should work even though we return a multiline string? Not that familiar with `yarn auto`.

PR #6432, opened 2025-07-10
chore: add dotcom hotfix label and trigger scripts

MitjaBezensek:
Makes a hotfix for PR's that have the `dotcom-hotfix-please` label.

### Change type

- [x] `other`

### Test plan

1. Verify the new label triggers the dotcom hotfix workflow.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added automation for triggering dotcom hotfixes via labels.

### DIFF
--- a/internal/scripts/trigger-dotcom-hotfix.ts
+++ b/internal/scripts/trigger-dotcom-hotfix.ts
@@ -36,7 +36,9 @@ async function main() {
 	await discord.step(`Creating hotfix branch and cherry-picking changes`, async () => {
 		const HEAD = (await exec('git', ['rev-parse', 'HEAD'])).trim()
 		await exec('git', ['fetch', 'origin', 'hotfixes'])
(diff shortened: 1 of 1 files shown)


internal/scripts/trigger-dotcom-hotfix.ts
```diff
@@ -0,0 +1,102 @@
...
+			head: hotfixBranchName,
+			base: 'hotfixes',
+		})
+
+		await octokit.rest.pulls.merge({
+			owner: 'tldraw',
+			repo: 'tldraw',
+			pull_number: createdPr.data.number,
+			merge_method: 'squash',
+			auto_merge: true,
```
MitjaBezensek, line 75: My understanding is that as soon as the checks pass this will auto merge the PR. We are not using a merge queue for hotfixes, but the concurrency option on the workflow should prevent multiple hotfixes to go in at the same time.

PR #6436, opened 2025-07-10
Streamed persist/load 

ds300:
We keep getting complaints of rooms with tons of drawing not being able to persist new data. This should help alleviate that in the near term, though really I think we need to start treating path data for draw shapes as binary assets somehow.

1. Use multipart uploads to stream json chunks to r2 (would be nice if you could just pass a ReadableStream to `bucket.put()` but you have to specify the file size ahead of time with that API for some reason.
2. Use a streaming json parser to load the snapshot. (would be nice if they did this under the hood for us when calling `res.json()` but it's currently literally implemented as `JSON.parse(await res.text())` https://github.com/cloudflare/workerd/blob/250d4b41ba11bcff1e8b1fad8c334654c183e9f4/src/workerd/api/r2-bucket.c%2B%2B#L1168

### Change type

- [x] `other`

### API changes
- (Only internal APIs were affected)
- Changed the sync protocol to allow passing the initial 'connect' msg diff as a gzipped base64 encoded string.


SomeHats (commented):
This is really cool. 

My one concern with this is switching from the super-mega-optimized V8 native JSON parsing & serialization methods to parsing (& some serializing) implemented in JS. It might be fine, but I'd be interested in maybe seeing some numbers to understand the impact of this, or a way we can try it out on some big rooms that are causing problems at the moment before rolling it out to everyone.

When we were talking about streaming parsing / serialization the other day, the thing I had in mind was [line-separated json](https://jsonlines.org/). This would mean a new format for serialized documents (I guess we'd detect if they start with `{"v": 2}\n` or something), but would also mean that we could get the benefits of streaming while sticking to native parsing/serialisation for the bulk of work. 

If switching away from native JSON methods isn't an issue then i think we should def go with this, but if it does have drawbacks I think it could be worth exploring something with JSON lines too.

MitjaBezensek:
Is the streaming parser more memory efficient than the regular one? If it is then that might allow us to increase our file size limit, but if it is not it might even make things worse.

Let's say we now allow users to upload bigger documents. Those documents end up in r2, but when the DO will try to load them into memory it will fail with the `Durable Object's isolate exceeded its memory limit and was reset.` error and you won't be able to load the room at all. 

I tried this PR on the big file mentioned in the `Out of memory issues in large rooms` project proposal and the memory issue still happened.

Maybe we can try to load a few of the large documents and make sure we don't start hitting the limit sooner with this change?

ds300:
> My one concern with this is switching from the super-mega-optimized V8 native JSON parsing & serialization methods to parsing (& some serializing) implemented in JS. It might be fine

it's fine. it's actually about 10% faster for large files and the difference is negligible for small files.

ds300:
> Is the streaming parser more memory efficient than the regular one

yeah, that's reason i'm adding it! the regular one loads the whole json blob into memory then converts it to a string and then calls native JSON.parse on it. This streaming one takes chunks (sized 4kb, on my machine at least) and tokenizes+parses them incrementally. so it never has a huge amount of data in memory beyond what accumulates as the final result.

Though I haven't done any heap profiling to confirm this.

> I tried this PR on the big file mentioned in the Out of memory issues in large rooms project proposal and the memory issue still happened.

I think it's most likely running out of memory when it's sending the initial data websocket connection. Will need to check this on monday though.

ds300:
yeah this stuff is not that useful if we still need to `JSON.stringify()` the whole dang document when sending it to newly-connected clients. And that's a big task to fix, so gonna put this on hold while we discuss the project proposal

ds300:
think we extracted everything useful from here

PR #6444, opened 2025-07-11
a11y: add preference to show labels/tooltips

mimecuvalo:
This is to address these issues:
- https://linear.app/tldraw/issue/ACC-72/add-labels-to-the-inputs
- https://linear.app/tldraw/issue/ACC-4/the-light-grey-states-in-the-panel-on-the-right-side-dont-have-enough

https://github.com/user-attachments/assets/3e4f68f0-1468-4e77-98d4-5a10c904ac4a

Screenshot 2025-07-11 at 12 43 39
Screenshot 2025-07-11 at 12 43 48
Screenshot 2025-07-11 at 13 35 45
Screenshot 2025-07-11 at 13 35 59

Focus ring for selected items:
Screenshot 2025-07-14 at 15 07 02

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: add preference to show labels/tooltips

### API changes

- add `showUiLabels` as a preference
- add `TldrawUiTooltip` as a new component


SomeHats:
might be worth sticking pointer-events: none on these, they make it hard to mouse over the thing below them in grids:
Kapture 2025-07-11 at 16 50 52

Also, is it possible to get rid of the slight flicker you get when switching from one thing to the next? when moving along e.g. the toolbar, the way it disappears for a frame or two and then reappears is quite distracting. I this it's because I see the tooltip as a single UI item that moves, rather than something appearing and disappearing. Example really leaning into this that isn't right for tldraw but makes the 'single ui item' thing clear: 
Kapture 2025-07-11 at 16 56 26

Also, if we can get this feeling really smooth and nice and maybe finesse the timings a bit, what do you think of having the tooltips enabled by default?

mimecuvalo:
> Also, if we can get this feeling really smooth and nice and maybe finesse the timings a bit, what do you think of having the tooltips enabled by default?

I refer you to "mime's musings" from the 1st week when I joined — I've wanted tooltips since day 1 😆

mimecuvalo:
ported over styles from the dotcom styling of tooltips.

> might be worth sticking pointer-events: none on these, they make it hard to mouse over the thing below them in grids: 

good call, adding `disableHoverableContent` fixed this.

> Also, is it possible to get rid of the slight flicker you get when switching from one thing to the next?

i'll work on that next! i'll make a singleton tooltip manager

mimecuvalo:
@SomeHats @steveruizok  PTAL! the tooltips have been reworked

SomeHats:
looking good! the animation should respect reduce motion, but personally i would remove it entirely - i don't think it's right for tldraw. sorry, i realise looking back at my comment where i showed an animated example that it looked like i was suggesting we add animation

mimecuvalo:
> looking good! the animation should respect reduce motion, but personally i would remove it entirely - i don't think it's right for tldraw. sorry, i realise looking back at my comment where i showed an animated example that it looked like i was suggesting we add animation

yeah this is a @steveruizok call for the animation - not sure how you felt it / assumed you'd tweak this anyway :D

mimecuvalo:
@SomeHats PTAL!

packages/tldraw/src/lib/shapes/shared/usePrefersReducedMotion.tsx
```diff
@@ -1,10 +1,20 @@
+import { useMaybeEditor, useValue } from '@tldraw/editor'
 import { useEffect, useState } from 'react'
 
 /** @public */
 export function usePrefersReducedMotion() {
+	const editor = useMaybeEditor()
+	const animationSpeed = useValue('animationSpeed', () => editor?.user.getAnimationSpeed(), [
```
mimecuvalo, line 7: @MitjaBezensek or @steveruizok I noticed that we don't actually use the user pref for this hook??
any historical reason for that or maybe just has been overlooked thus far?

PR #6446, opened 2025-07-11
a11y: fix up directional navigation, accounting for parentIds

mimecuvalo:
I noticed in testing that Tab does the right thing, skipping over child shapes (and staying within Frame shapes when tabbing around). But we didn't apply the same logic to directional navigation, oops.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- a11y: fix up directional navigation, accounting for parentIds

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1211,7 +1211,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     getIsReadonly(): boolean;
     // @internal
     getMarkIdMatching(idSubstring: string): null | string;
-    getNearestAdjacentShape(currentShapeId: TLShapeId, direction: 'down' | 'left' | 'right' | 'up'): TLShapeId;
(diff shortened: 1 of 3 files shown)


MitjaBezensek (approved):
👍

One thing I noticed (which is unrelated to this PR) is that pressing escape (when a shape within a group is selected) will then move the selection to the group. But that's not the case for frames. Not sure if that's intended or not.

mimecuvalo:
> One thing I noticed (which is unrelated to this PR) is that pressing escape (when a shape within a group is selected) will then move the selection to the group. But that's not the case for frames. Not sure if that's intended or not.

this is super interesting, i didn't know about this functionality / code. after digging around, i see it's from `this.editor.popFocusedGroupId()` call in `SelectTool/childStates/Idle.ts` so it's very specific to `group` shapes.

@SomeHats maybe has context on why this behavior wasn't mimicked in frames?

PR #6455, opened 2025-07-14
Use a different api to get the checks.

MitjaBezensek:
The current logic does not correctly check whether all the required statuses have passed. We were only checking the `mergeable` status on the PR, but it seems that only reports whether the PR does not contain merge conflicts.

With this update we use a different api for checking if the required checks have passed. This uses:
- branch api to get the `hotfixes` [branch protections](https://docs.github.com/en/rest/branches/branch-protection?apiVersion=2022-11-28#get-branch-protection)
- commit statuses via the [combined status for ref api](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#get-the-combined-status-for-a-specific-reference)
- then check the commit statuses against the required branch protections to make sure all the required checks passed

### Change type

- [x] `other`

### DIFF
--- a/.github/workflows/trigger-dotcom-hotfix.yml
+++ b/.github/workflows/trigger-dotcom-hotfix.yml
@@ -14,8 +14,6 @@ defaults:
 permissions:
   contents: write
   pull-requests: write
-  checks: read
-  statuses: read
 
 jobs:
   trigger:
--- a/.github/workflows/trigger-dotcom-hotfix.yml
+++ b/.github/workflows/trigger-dotcom-hotfix.yml
@@ -14,6 +14,8 @@ defaults:
 permissions:
   contents: write
   pull-requests: write
+  checks: read
+  statuses: read
 
 jobs:
   trigger:
--- a/internal/scripts/trigger-dotcom-hotfix.ts
+++ b/internal/scripts/trigger-dotcom-hotfix.ts
@@ -48,7 +48,7 @@ async function main() {
 		await exec('git', ['push', 'origin', hotfixBranchName])
 	})
 
-	await discord.step('Creating hotfix PR and adding to merge queue', async () => {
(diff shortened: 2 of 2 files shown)


internal/scripts/trigger-dotcom-hotfix.ts
```diff
@@ -72,20 +72,40 @@ This PR cherry-picks the changes from the original PR to the hotfixes branch for
...
-				nicelog(`PR #${createdPr.data.number} status still being calculated, waiting...`)
+				nicelog(`PR #${createdPr.data.number} merge status still being calculated, waiting...`)
 				await new Promise((resolve) => setTimeout(resolve, 15 * 1000))
 				continue
 			}
 
-			// GitHub has finished calculating, now check the result
-			if (prStatus.mergeable) {
+			if (!prStatus.mergeable) {
+				nicelog(`PR #${createdPr.data.number} has conflicts and cannot be merged`)
```
MitjaBezensek: Looks like mergeable only tells you whether you have merge conflicts.

.github/workflows/trigger-dotcom-hotfix.yml
```diff
@@ -14,6 +14,8 @@ defaults:
 permissions:
   contents: write
   pull-requests: write
+  checks: read
+  statuses: read
```
MitjaBezensek: Think these are the correct permissions, but not 100%.

MitjaBezensek:
> I think we can use `prStatus.mergeable_state === 'clean'` instead of doing all the branch protection checks ourselves.
> 
> The `mergeable_state` property is available in the graphql api and the rest api, but it's only documented in the graphql api (where it is called `PullRequest#mergeStateStatus`). It's been around for at least 6 years though so probably safe to use it at this point.

Sounds good, updated it with that. I guess if they ever do decide to remove it (unlikely) the run will fail and we can update the logic.

PR #6458, opened 2025-07-14
Feat/llm agent editor v1

Innovationlzc:


SomeHats:
Hi @Innovationlzc! Thanks for the PR! 

Could you talk through what you're trying to achieve with this and why you want to land it in the tldraw repo? As far as I can tell, the example you're adding doesn't actually use tldraw at all.

steveruizok:
This seems to be an agent's PR, sus account, closing.

PR #6462, opened 2025-07-15
feat(sdk): support custom colors and theme extension

steveruizok:
This PR is a rough update that supports custom colors.

Kapture 2025-07-15 at 22 20 38

This PR:
- Allows all places that accept colors from the theme object to accept a plain string instead
- Allows customization of the theme object

**Plain string colors**

When using a plain string as a color:

- Default colors still work the way they did before, however this adds minimal support for any other color, too.
- Complex color properties such as semi-fills, pattern colors, and light and dark mode colors, are not supported.

**Extending the default colors**

When extending the default colors object:

- Default colors still work the way they did before, but can be modified as they were before.
- A user can define new colors with complex color properties such as semi-fills, pattern colors, and light and dark mode colors.

### Change type

- [x] `feature`

### Test plan

1. See the custom colors example.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- SDK: Added support for arbitrary color values in the `color` prop of default shapes.
- SDK: Added support for extending default theme (adding new colors)

### API changes

- Updates `TLColorStyle` to support strings.
- Adds new `EnumStylePropOrString` prop type.
- Adds `StyleProp.defineEnumOrString`

### DIFF
--- a/apps/examples/src/styles.css
+++ b/apps/examples/src/styles.css
@@ -176,8 +176,6 @@ li.examples__sidebar__item {
 	right: 0px;
 }
 
-
-
 .examples__sidebar__item__title {
 	flex-grow: 2;
 	overflow: hidden;
@@ -230,7 +228,6 @@ li.examples__sidebar__item {
 	padding-right: 64px;
 }
 
-
 /* -------------------- List Item ------------------- */
 
 .example__sidebar__item__buttons {
@@ -267,8 +264,6 @@ li.examples__sidebar__item {
 .example__sidebar__item__button:nth-of-type(n + 1) {
 	margin-left: -4px;
 }
-
-
 
 .example {
 	position: absolute;


mimecuvalo:
waaaaat, omg

steveruizok:
Some things that I'd like to try:

- making the theme a class that can have colors accessed from it, i.e. `theme.getColor(myColor, "semi")`, to replace this PR's weird ternaries (though they're fine)
- allowing users to register additional color subtypes, like semi or solid etc, for custom colors

SomeHats:
Overall I'm pretty into this! It works well for colors, but I'm not sure how well we could extend this concept to other styles.

Problems I would like to solve:
1. Fill styles kinda break with this. Your theme class might help there though! It would be cool to do an automatic `semi` color or something
2. Some of our palette/theme colors overlap with CSS colors - ie "blue" is valid CSS, but also a palette color. It's very minor, but I wonder if it's worth making these values `tl-blue` or `palette-blue` or something
3. Another minor one, but for highlighter pen we use HDR colors, which I think need an sRGB fallback. Might be hard to do tho.

It might be interesting to split `color` into `fillColor` and `strokeColor` while we're working in this area.

I'm also pretty cautious about the APIs we might introduce with this, and the precedent it might set / we might need to be consistent with down the line. I think the ternaries are a bit awkward, so something like a theme class would help a lot, but then I'm not sure if that's a concept that would still make a load of sense for things like strokes, fill styles, fonts, etc. 

I'd still rather we take some time to properly explore some more general options here which just isn't something we've ever really done. But I wouldn't mind us pushing forward with this approach, so long as we're happy to potentially throw some of these APIs away down the line.

SomeHats:
I've been thinking a bit more about this and would really love if we could come up with a story for at least stroke sizes (comes up almost as often as colors) before moving forward with this approach. I'm worried that we'll back ourselves into a bit of a weird API design corner if we don't consider other style types before going ahead with this.

One idea might be to introduce something like a style util or something that can answer questions about the true value of a style.

I could imagine an API that looks like this:
```tsx
/********** in tlshema: **********/
// we define 'base styles' that sort of describe the interface for the styles used by our own shapes
abstract class BaseSizeStyle<T> extends StyleProp<T> {
    // they have IDs similar to shape utils and state nodes
    static id = 'tldraw:size'

    // this size style needs a helper method for turning whatever value it's mean to be into a px value
    abstract toPxSize(value: T, context: string): number
}

// then, we implement our styles by extending those base styles:
class DefaultSizeStyle extends BaseSizeStyle<'s' | 'm' | 'l' | 'xl'> {
    // out default looks up values in objects like we do currently
    toPxSize(value, context) {
        // handle different values in different contexts:
        if (context === 'stroke') return STROKE_SIZES[value]
        if (context === 'font') return FONT_SIZES[value]
        return DEFAULT_STROKE_SIZES[value]
    }
}

/********** in custom code: **********/
// a custom implementor can write this however they want
class NumberSizeStyle extends BaseSizeStyle<number> {
    toPxSize(value, context) {
        if (value === 'stroke') return value * 0.2 // or whatever
        return value
    }
}

class ComplexSizeStyle extends BaseSizeStyle<{size: number, font: number}> {
    toPxSize(value, context) {
        return value[context]
    }
}

// they can pass it in to tldraw. it gets injected into the schema and made available to look up on the editor
return <Tldraw styleProps={[NumberSizeStyle]} />

/********** in shapes: **********/
// we can look up the actual impls and use the values accordingly: 
const actualSize = editor.getStyleProp('tldraw:size').toPxSize(shape.props.size, 'stroke')
return <rect strokeWidth={actualSize} />
```

zb-builds-things:
Is this one back in the backlog? Very interested in it 🙏

steveruizok:
Closing due to merge conflicts

PR #6465, opened 2025-07-16
fix(TldrawImage): ensure fonts load and asset URLs are passed

MitjaBezensek:
We weren't passing the asset urls and we weren't waiting for the fonts to load.

### Before
image

### After
image

### Change type

- [x] `bugfix`

### Test plan

1. Open the image component example and verify fonts render correctly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where fonts would not load correctly in TldrawImage.


mimecuvalo (approved):
oofa, nice catch

PR #6467, opened 2025-07-16
fix(dotcom): resolve CSP error in development

MitjaBezensek:
Not sure why I only saw this error in dev, but I don't see it in prod.

I did see that we handle csp a bit differently in [head](https://github.com/tldraw/tldraw/blob/c5fa4794492228ee4dd7bdc149f7728438ace4ba/apps/dotcom/client/src/components/Head/Head.tsx#L27), so maybe that's it. @mimecuvalo maybe you have a better solution?

### Change type

- [x] `bugfix` 

### Test plan

- [ ] Unit tests (if present)

### Release notes

- Fix a CSP error occurring in development environments.

### DIFF
--- a/apps/dotcom/client/src/utils/csp.ts
+++ b/apps/dotcom/client/src/utils/csp.ts
@@ -26,6 +26,7 @@ export const cspDirectives: { [key: string]: string[] } = {
 		'https://stats.g.doubleclick.net',
 		'https://*.google-analytics.com',
 		'https://api.reo.dev',
+		'https://fonts.googleapis.com',
 	],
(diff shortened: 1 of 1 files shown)


mimecuvalo:
Hm, how do you repro this in dev?

The reason it doesn't happen in prod is because we don't include the Inter font on prod.

dev:
Screenshot 2025-07-17 at 11 51 50


prod (inserted at build time at the `PRELOADED_FONTS` placeholder):
Screenshot 2025-07-17 at 11 51 56

MitjaBezensek:
> Hm, how do you repro this in dev?

Here are the steps:
https://linear.app/tldraw/issue/ENG-3569/errors-with-loading-fonts-when-doing-svg-exports-in-developoment

PR #6470, opened 2025-07-17
perf: adaptive FPS

mimecuvalo:
The throttle PR (https://github.com/tldraw/tldraw/pull/6409) reminded me of a wee conversation we had when the performance improvements went out last year (https://github.com/tldraw/tldraw/pull/2977#discussion_r1505742741) about how setting the FPS to be a constant 60 sort of short-changed the faster machines that could achieve 120fps.

When it hits 120 fps on my machine it starts feeling butter smooooth.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- perf: adaptive FPS

### DIFF
--- a/packages/utils/src/lib/throttle.ts
+++ b/packages/utils/src/lib/throttle.ts
@@ -5,16 +5,85 @@ const isTest = () =>
 	!globalThis.__FORCE_RAF_IN_TESTS__
 
 const fpsQueue: Array<() => void> = []
-const targetFps = 60
-const targetTimePerFrame = Math.floor(1000 / targetFps) // 16ms
+
+// Adaptive FPS system
+const MIN_FPS = 30
+const MAX_FPS = 120
+const DEFAULT_FPS = 60
+const PERFORMANCE_SAMPLE_SIZE = 10
+const ADJUSTMENT_THRESHOLD = 0.8 // 80% of target time
+
+let targetFps = DEFAULT_FPS
+let targetTimePerFrame = Math.floor(1000 / targetFps) // 16ms
 let frameRaf: undefined | number
 let flushRaf: undefined | number
 let lastFlushTime = -targetTimePerFrame
 
+// Performance monitoring
+let performanceSamples: number[] = []
+let lastPerformanceCheck = 0
+let consecutiveGoodFrames = 0
+let consecutiveBadFrames = 0
+
+const updateTargetFps = () => {
+	if (performanceSamples.length < PERFORMANCE_SAMPLE_SIZE) return
+
+	const avgFrameTime = performanceSamples.reduce((a, b) => a + b, 0) / performanceSamples.length
+	const currentTargetTime = 1000 / targetFps
+
+	// If we're consistently performing well, try to increase FPS
+	if (avgFrameTime < currentTargetTime * ADJUSTMENT_THRESHOLD) {
+		consecutiveGoodFrames++
+		consecutiveBadFrames = 0
+
+		// After 3 consecutive good performance checks, try to increase FPS
+		if (consecutiveGoodFrames >= 3 && targetFps < MAX_FPS) {
+			const newFps = Math.min(MAX_FPS, targetFps + 10)
+			targetFps = newFps
+			targetTimePerFrame = Math.floor(1000 / targetFps)
+			consecutiveGoodFrames = 0
+		}
+	}
+	// If we're consistently performing poorly, decrease FPS
+	else if (avgFrameTime > currentTargetTime * 1.2) {
+		consecutiveBadFrames++
+		consecutiveGoodFrames = 0
+
+		// After 2 consecutive bad performance checks, decrease FPS
+		if (consecutiveBadFrames >= 2 && targetFps > MIN_FPS) {
+			const newFps = Math.max(MIN_FPS, targetFps - 10)
+			targetFps = newFps
+			targetTimePerFrame = Math.floor(1000 / targetFps)
+			consecutiveBadFrames = 0
+		}
+	}
+
+	// Reset samples for next measurement period
+	performanceSamples = []
+}
+
+const recordPerformance = (frameTime: number) => {
+	performanceSamples.push(frameTime)
+
+	// Check performance every second
+	const now = Date.now()
+	if (now - lastPerformanceCheck > 1000) {
+		updateTargetFps()
+		lastPerformanceCheck = now
+	}
+}
+
 const flush = () => {
+	const startTime = Date.now()
 	const queue = fpsQueue.splice(0, fpsQueue.length)
 	for (const fn of queue) {
 		fn()
+	}
+	const endTime = Date.now()
+
+	// Record performance if not in test mode
+	if (!isTest()) {
+		recordPerformance(endTime - startTime)
 	}
 }
 
@@ -116,4 +185,37 @@ export function throttleToNextFrame(fn: () => void): () => void {
 			fpsQueue.splice(index, 1)
 		}
 	}
+}
+
+/**
+ * Gets the current adaptive target FPS.
+ * @returns the current target FPS (between 30-120)
+ * @internal
+ */
+export function getCurrentFps(): number {
+	return targetFps
+}
+
+/**
+ * Resets the adaptive FPS system back to default settings.
+ * Useful for testing or manual performance adjustments.
+ * @internal
+ */
+export function resetAdaptiveFps(): void {
+	targetFps = DEFAULT_FPS
+	targetTimePerFrame = Math.floor(1000 / targetFps)
+	performanceSamples = []
+	lastPerformanceCheck = 0
+	consecutiveGoodFrames = 0
+	consecutiveBadFrames = 0
+}
+
+/**
+ * Manually sets the target FPS (will be overridden by adaptive system).
+ * @param fps - the target FPS to set (will be clamped between 30-120)
+ * @internal
+ */
+export function setTargetFps(fps: number): void {
+	targetFps = Math.max(MIN_FPS, Math.min(MAX_FPS, fps))
+	targetTimePerFrame = Math.floor(1000 / targetFps)
 }
--- a/packages/tldraw/src/lib/ui/components/DefaultDebugPanel.tsx
+++ b/packages/tldraw/src/lib/ui/components/DefaultDebugPanel.tsx
@@ -1,5 +1,6 @@
 import {
 	debugFlags,
+	getCurrentFps,
 	track,
 	useEditor,
 	usePassThroughWheelEvents,
@@ -110,7 +111,7 @@ function FPS() {
 					isSlow = !isSlow
 				}
 
-				fpsRef.current!.innerHTML = `FPS ${fps.toString()}`
+				fpsRef.current!.innerHTML = `FPS ${fps.toString()} (target: ${getCurrentFps()})`
 				fpsRef.current!.className =
 					`tlui-debug-panel__fps` + (isSlow ? ` tlui-debug-panel__fps__slow` : ``)
 
--- a/packages/utils/api-report.api.md
+++ b/packages/utils/api-report.api.md
@@ -142,6 +142,9 @@ export function fpsThrottle(fn: {
 // @internal (undocumented)
 export function getChangedKeys<T extends object>(obj1: T, obj2: T): (keyof T)[];
 
+// @internal
+export function getCurrentFps(): number;
+
 // @internal (undocumented)
 export function getErrorAnnotations(error: Error): ErrorAnnotations;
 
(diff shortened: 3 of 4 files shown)


ds300:
at the moment TLSyncClient uses this to throttle socket messages to 60fps and I don't think we'd want to be sending messages any faster than that. Maybe as part of this refactor you could allow overriding the upper fps bound for some throttled operations?

SomeHats:
@mimecuvalo have you had a change to see how this works across mobile devices? @MitjaBezensek can probably weigh in more but i'm guessing that's the main area where we want to avoid pushing the browser to is max so we don't get throttled

mimecuvalo:
> at the moment TLSyncClient uses this to throttle socket messages to 60fps and I don't think we'd want to be sending messages any faster than that. Maybe as part of this refactor you could allow overriding the upper fps bound for some throttled operations?

Good call out! Yeah, it doesn't maybe feel right that TLSyncClient should work at 60fps (at the moment). It does feel like either a) this uses a different, lodash-unmodified throttle function or b) we set an upper bound as you suggest — do you have a sense of a number that you'd suggest? Given network latency, I wouldn't expect this to be very high, I think 30fps is just fine for this - I know we use it for collaborative mouse cursors.

> @mimecuvalo have you had a change to see how this works across mobile devices? @MitjaBezensek can probably weigh in more but i'm guessing that's the main area where we want to avoid pushing the browser to is max so we don't get throttled

Good push. Had a chat with @MitjaBezensek about this offline, it's possible that a variant on this PR is to just suss out the max FPS and then let the browser go from there. So, we have a target FPS that we deduce, and then the browser can throttle the `requestAnimationFrame` as needed, which it does anyway — this instead of us trying to change the target on the fly.

MitjaBezensek:
I'm trying to understand the benefits here. So:
1. We will run at higher frame rates on devices that can handle it. That's definitely great.
2. What happens for the devices that can't handle it though? Is adding this logic better than just relying on the browser doing the throttling? Did you test it out?
I'd really want to make sure that the additional logic does not make things worse for lower end devices. I could definitely see that doubling the rate at which we do things cause even more pressure on these devices and cause them to throttle sooner. Will this logic kick in and make it better? I'd definitely like us to answer these questions before proceeding. Especially since it does introduce quite some complexity to already complex part of the codebase.

mimecuvalo:
> We will run at higher frame rates on devices that can handle it. That's definitely great.

yeah exactly.

> Will this logic kick in and make it better? I'd definitely like us to answer these questions before proceeding. Especially since it does introduce quite some complexity to already complex part of the codebase.

this logic is adaptive in both directions (as opposed to the other variant PR). if slower operations start happening, the target FPS is lowered to match what the device can handle.

> What happens for the devices that can't handle it though?

One thing worth mentioning here about this thought is that we currently have a fixed 60 FPS target — if this were true, it would also be affecting current devices that can't catch up. I think this is where the browser steps and calls the RAF at a lower FPS if the workload is high. It's a target, so the browser isn't necessarily forced to work faster just b/c we change the target.

> Did you test it out?

There's an easy way to test this, just set all the targets to 1000 🙃 And it's still works just fine.

Screenshot 2025-07-24 at 14 22 22

> Is adding this logic better than just relying on the browser doing the throttling? 

I think the good fix about Steve's PR (https://github.com/tldraw/tldraw/pull/6409) is that if the browser is having a hard time catching up, when it finally does, the flush will happen. If they keep piling up and getting delayed, the target FPS will be more wishful thinking and just basically happen when the RAF is finally called. The adaptive piece of this tries to more align with what the browser is actually able to handle at the moment, work-load wise. So I think for slower devices it just brings the targets more in line, shouldn't affect perf by "overclocking" or trying to make it run faster. But the benefit then is strictly for faster devices that are detected to be able to run faster.

MitjaBezensek:
> > Will this logic kick in and make it better? I'd definitely like us to answer these questions before proceeding. Especially since it does introduce quite some complexity to already complex part of the codebase.
> 
> this logic is adaptive in both directions (as opposed to the other variant PR). if slower operations start happening, the target FPS is lowered to match what the device can handle.

But browser already does that, what's the benefit of us doing it on top of that? We base this whole logic on the rate that browser calls us. So browser is already doing this work, no?

> > Did you test it out?
> 
> There's an easy way to test this, just set all the targets to 1000 🙃 And it's still works just fine.

I can also set the target to 1000 on main and it works. Where's the benefit of this diff (show me a use case where us adapting the frame rate works better than just us relying on browser doing that)?

MitjaBezensek:
>  But the benefit then is strictly for faster devices that are detected to be able to run faster.

Think we can achieve that just by increasing the target frame rate in a separate raf, which would be much simpler?

mimecuvalo:
> But browser already does that, what's the benefit of us doing it on top of that? We base this whole logic on the rate that browser calls us. So browser is already doing this work, no?

Yeah, again, I think it was a good push, which is why I created the variant PR based on your feedback: https://github.com/tldraw/tldraw/pull/6497

> I can also set the target to 1000 on main and it works. Where's the benefit of this diff (show me a use case where us adapting the frame rate works better than just us relying on browser doing that)?

We might be going in circles 🙃 The benefit is for the _faster_ FPS, right? Where 60fps is slower than what the machine can handle (like my M1 mac). But hey, we could also just say, let the browser handle it all maybe?

> Think we can achieve that just by increasing the target frame rate in a separate raf, which would be much simpler?

I think I know what you mean but I'm not sure? Do you wanna create a variant PR to show what you mean?

MitjaBezensek:
> I think I know what you mean but I'm not sure? Do you wanna create a variant PR to show what you mean?

Something like this https://github.com/tldraw/tldraw/pull/6500/files
We only do the fps check 60 times, but we could keep it running indefinitely, just like in this pr. Not sure if it's needed though 🤷‍♂️

mimecuvalo:
> > I think I know what you mean but I'm not sure? Do you wanna create a variant PR to show what you mean?
> 
> Something like this https://github.com/tldraw/tldraw/pull/6500/files We only do the fps check 60 times, but we could keep it running indefinitely, just like in this pr. Not sure if it's needed though 🤷‍♂️

btw, you said "Think we can achieve that just by increasing the target frame rate in a separate raf," but that's what this PR and the maximum variant both do, right? The measuring is happening in a separate RAF cycle.

your PR i think might cause a separate issue potentially:
- since it runs as a side effect on module load, and since tldraw might be part of a heavier page that's loading, a 500ms wait time to start measuring might measure a heavy initialization sequence and then we won't get the right measurement
- the max checks idea is interesting and could be applied to this existing PR, but, again i would worry about measuring at some wrong moment where we don't get the right reading.

MitjaBezensek:
> btw, you said "Think we can achieve that just by increasing the target frame rate in a separate raf," but that's what this PR and the maximum variant both do, right? The measuring is happening in a separate RAF cycle.

Your PR also down regulates the fps, which my PR does not do. And this is the part that I have been questioning all along.

> 
> your PR i think might cause a separate issue potentially:
> 
> * since it runs as a side effect on module load, and since tldraw might be part of a heavier page that's loading, a 500ms wait time to start measuring might measure a heavy initialization sequence and then we won't get the right measurement
> * the max checks idea is interesting and could be applied to this existing PR, but, again i would worry about measuring at some wrong moment where we don't get the right reading.

I only wanted to show you what I had mind, it's in no way production ready, just trying to showcase the idea of "what if we only increase the fps on devices that have support for it". We can achieve that in many ways and my PR is just a quick example of that.

mimecuvalo:
Had a chat offline with @MitjaBezensek to sort out what we meant. Merged the maximum variant on to this PR, so now this only ratchets up instead of down.

@ds300 we were curious if we could forgo having a special function/special logic for `TLSyncClient` and just use a regular throttle? Is there a reason this needed `fpsThrottle` vs. a regular throttle?

ds300:
> @ds300 we were curious if we could forgo having a special function/special logic for TLSyncClient and just use a regular throttle? Is there a reason this needed fpsThrottle vs. a regular throttle?

i think regular throttle would be fine, just need to make sure the fn gets called on the trailing edge

mimecuvalo:
@MitjaBezensek / @ds300 any remaining feedback?

packages/utils/src/lib/throttle.ts
```diff
@@ -117,3 +226,41 @@ export function throttleToNextFrame(fn: () => void): () => void {
...
+export function getCurrentFps(): number {
+	return targetFps
+}
+
+/**
+ * Resets the FPS detection system back to default settings.
+ * Useful for testing or when you want to re-detect the device's natural FPS.
+ * @internal
+ */
+export function resetAdaptiveFps(): void {
```
MitjaBezensek, line 315: Do we still need `resetAdaptiveFps` and `setTargetFps`, it looks like they are not being used? Maybe they are useful for testing?

packages/utils/src/lib/throttle.ts
```diff
@@ -5,17 +5,126 @@ const isTest = () =>
...
+
 const flush = () => {
 	const queue = fpsQueue.splice(0, fpsQueue.length)
 	for (const fn of queue) {
 		fn()
 	}
+
+	// Start measuring actual browser frame timing
+	if (!isTest()) {
+		startFrameTimingMeasurement()
```
MitjaBezensek, line 176: Should we maybe do this outside of flush? This is a really hot code path so might be worth it to avoid this call which early returns all the time?

packages/utils/src/lib/throttle.ts
```diff
@@ -5,17 +5,126 @@ const isTest = () =>
...
+
+	const avgFps = fpsCheckHistory.reduce((a, b) => a + b, 0) / fpsCheckHistory.length
+
+	// Track the maximum FPS we've detected (likely the device's natural refresh rate)
+	if (avgFps > maxDetectedFps) {
+		maxDetectedFps = Math.min(MAX_FPS, Math.round(avgFps))
+
+		// Only update target FPS if user hasn't manually set it
+		if (!isManuallySet) {
+			targetFps = maxDetectedFps
```
MitjaBezensek, line 51: Might be worth stopping frame measurements if we already hit 120 fps?

mimecuvalo:
> This looks like it lowers the performance somehow? Here's me resizing and translating 400 shapes.
> 
> Was testing with `yarn dev-app`. So might be due to the `TLSyncClient` changes?
> 

Huh, lemme take a look, it's weird b/c the target is still 60 in your videos, so it shouldn't be much of a change? Curious...

packages/utils/src/lib/throttle.ts
mimecuvalo: Actually, now I'm wondering if 120 should be the max or not 😆 
There _are_ higher high-end gaming monitors that have 240hz... maybe we should allow even higher??

packages/utils/src/lib/throttle.ts
mimecuvalo: yeah it's useful for testing but i'm happy to remove. either way

packages/utils/src/lib/throttle.ts
MitjaBezensek: Not sure we can reliably process things at that rate 😁

packages/utils/src/lib/throttle.ts
MitjaBezensek: Happy to keep it then👍

mimecuvalo:
> This looks like it lowers the performance somehow? Here's me resizing and translating 400 shapes.
> 
> Was testing with `yarn dev-app`. So might be due to the `TLSyncClient` changes?

I reverted the last changes to `TLSyncClient` (that removed fpsThrottle) - those last changes indeed did make it worse.
It basically ends up worse that there's a separate queue of operations happening vs. the TLSyncClient queue being tied in specifically to our main FPS queue — this was an interesting learning!

I've been exploring how I could modify TLSyncClient. One option is that when you're solo in a room that the updates don't have to happen quite as often as 60fps(!), another option is having some kind of backoff if we're starting to send/receive a lot of operations over the wire and our fpsQueue is starting to fall behind. Def. worth a discussion to think about this some more.

mimecuvalo:
Oh, the other thing is possibly being able to designate something on the `fpsQueue` as having a lower FPS, something worth considering as well.

Just spoke with @ds300 offline about the TLSyncClient and adding a "solo mode". Basically to:
- not send up mouse cursor and presence state to the server when in a room by yourself
- lower the FPS (even to once a second) when you're just in a room by yourself

mimecuvalo:
I marked this PR as `[wip]` and made some significant changes in advance of @ds300's work for a `isSolodMode`.

Basically now there are two queues, one that can have a slower FPS if needed (like the network syncing), and one that can have a the more performant FPS.

mimecuvalo:
closing in favor of https://github.com/tldraw/tldraw/pull/6868!

PR #6472, opened 2025-07-17
refactor(draw): optimize pressure property storage

steveruizok:
This PR refactors how draw shape point pressure is stored and handled, leading to storage optimization and improved clarity.

Previously, pressure was stored as a float `0-1` in the `z` property of a point, with `0` or `0.5` indicating no pressure. This update changes the storage to:
1.  **Omit the `z` property entirely** when no pressure information is available (e.g., mouse input).
2.  Store pressure as an **integer between `0` and `100`** when pressure is provided (e.g., pen/stylus input).

A new migration (`OptimizePressure: 3`) is included to safely convert existing draw shapes to the new format, ensuring backward and forward compatibility. This change reduces the data size of draw shapes and makes the pressure values more intuitive.

### Change type

- [x] `improvement`

### Test plan

1.  **Draw with mouse:**
    *   Select the draw tool.
    *   Draw a shape using a mouse.
    *   Verify the shape is drawn correctly. (Internally, points should *not* have a `z` property).
2.  **Draw with pen/stylus (if applicable):**
    *   Select the draw tool.
    *   Draw a shape using a pen or stylus with varying pressure.
    *   Verify the shape's stroke width changes with pressure. (Internally, points *should* have a `z` property with an integer value between 0-100).
3.  **Load old files:**
    *   Open a file created before this PR with draw shapes.
    *   Verify the draw shapes load and display correctly, and their pressure values are migrated.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Optimized draw shapes by storing pressure as an integer (0-100) only when provided, reducing file size and improving data clarity. Non-pressure inputs (e.g., mouse) no longer store a pressure value. Existing draw shapes will be automatically migrated.


ds300:
This won't make a significant difference to runtime memory usage, and only a marginal difference to serialization output size.

If the latter is the goal it would be a lot more impactful to encode points as base64 float16 arrays (see the david/base-64-draw-shape branch for a POC), but by far the best thing for serialization output size would be gzip.

For runtime performance the size of shape.props.segments is dwarfed by the size of the derived cached data we create based on it (vec arrays, svg path strings, Geometry2D, etc).

PR #6474, opened 2025-07-17
feat(create-tldraw): add npm create tldraw CLI tool

SomeHats:
A cli tool for quickly creating a tldraw app from one of our templates.

Kapture 2025-07-17 at 17 42 10

### Change type

- [x] `feature`

### Test plan

1. Run `npm create tldraw` and follow the prompts.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added `npm create tldraw`, a CLI tool for quickly creating a tldraw app from templates.

### DIFF
--- a/internal/scripts/export-template.ts
+++ b/internal/scripts/export-template.ts
@@ -89,7 +89,10 @@ async function main() {
 	console.log('Cleaning package.json...')
 	delete packageJson[EXPORT_CONFIG_KEY]
 
-	if (exportConfig?.scripts && packageJson.scripts) {
+	if (exportConfig?.scripts) {
+		if (!packageJson.scripts) {
+			packageJson.scripts = {}
+		}
 		for (const [name, script] of Object.entries(exportConfig.scripts)) {
 			if (script === null) {
 				delete packageJson.scripts[name]


apps/bemo-worker/package.json
```diff
@@ -34,7 +34,7 @@
 		"@cloudflare/workers-types": "^4.20250703.0",
 		"@types/react": "^18.3.18",
 		"@types/react-dom": "^18.3.5",
-		"esbuild": "^0.25.0",
+		"esbuild": "^0.25.6",
```
mimecuvalo, line 37: fyi dependabot updating this again here: https://github.com/tldraw/tldraw/pull/6482

packages/create-tldraw/src/wrap-ansi.ts
```diff
@@ -0,0 +1,249 @@
+/*!
```
mimecuvalo, line 1: instead of forking, this maybe feels like `yarn patch` territory?
or, might be worth making a PR upstream, i bet Sindre would be receptive

packages/create-tldraw/src/group-select.ts
```diff
@@ -0,0 +1,148 @@
+/*!
```
mimecuvalo, line 1: hm, checking here, are these not useful?
- https://github.com/bombshell-dev/clack/blob/1adb270af5509ead3d22c4fa602b465f39c7a8c2/packages/prompts/src/group.ts
- or, https://github.com/bombshell-dev/clack/blob/1adb270af5509ead3d22c4fa602b465f39c7a8c2/packages/prompts/src/group-multi-select.ts

packages/create-tldraw/src/main.ts
```diff
@@ -0,0 +1,232 @@
...
+	if (overwrite === 'yes') {
+		emptyDir(targetDir)
+	}
+}
+
+async function downloadTemplate(template: Template, targetDir: string) {
+	const s = spinner()
+	s.start(`Downloading github.com/${template.repo}...`)
+	try {
+		const url = `https://github.com/${template.repo}/archive/refs/heads/main.tar.gz`
```
mimecuvalo, line 194: oo neat

mimecuvalo: actually, checking here, if this is pulling from main, possible that there's a version skew / types might be off until the tldraw version is matched with the template?

actually, for my understanding, how does this handle the `tldraw` version as they're currently set to `workspace:*`?

packages/create-tldraw/src/main.ts
```diff
@@ -0,0 +1,232 @@
...
+const HELP = `\
+Usage: create-tldraw [OPTION]... [DIRECTORY]
+
+Create a new project using tldraw.
+With no arguments, start the CLI in interactive mode.
+
+Options:
+  -h, --help                 display this help message
+  -t, --template NAME        use a specific template
+  -o, --overwrite            overwrite the target directory if it exists
```
mimecuvalo, line 37: ehh, i don't expect `overwrite` to be an option for these sorts of things. people can make a new folder if they need to and remove the old one. i don't want to give people a loaded gun with this option.

mimecuvalo (commented):
excited for this!

packages/create-tldraw/src/group-select.ts
SomeHats: sadly not

SomeHats: group is a different thing, and group multi select doesn't work for single selection

packages/create-tldraw/src/wrap-ansi.ts
SomeHats: this was involved enough that i wanted to be able to test it and write types for it, which is why i didn't want to do it as a patch.

I will look at upstreaming when i have some free time

packages/create-tldraw/src/main.ts
SomeHats: i'm just matching `npm create vite` here. happy to remove, but imo this option is pretty unambiguous so i'm not sure it's that much of a footgun

packages/create-tldraw/src/main.ts
SomeHats: template publishing is tied to sdk releases. when we release a new SDK version (which includes this package), we also publish templates, with all `workspace:` dependencies set to the newly published SDK version. This pulls from those published repos, so should always have up to date dependencies.

packages/create-tldraw/src/group-select.ts
mimecuvalo: sigh kk

packages/create-tldraw/src/main.ts
mimecuvalo: oh ok, i guess i'm thrown off b/c this references `main.tar.gz` - shouldn't be `production`?

packages/create-tldraw/src/main.ts
mimecuvalo: i'm surprised Vite has this (i'm more used to create-react-app/create-next-app), footgun in the sense where if you use a command-line completion or history it can accidentally be reused.

it's fine though, you do have a confirmation `select` so it doesn't do without confirming which is fine.

packages/create-tldraw/src/main.ts
```diff
@@ -0,0 +1,232 @@
...
+	await ensureDirectoryEmpty(targetDir, args.overwrite)
+	await downloadTemplate(template, targetDir)
+	await renameTemplate(name, targetDir)
+
+	const manager = getPackageManager()
+	const doneMessage = ['Done! Now run:', '']
+	if (targetDir !== process.cwd()) {
+		doneMessage.push(`   cd ${relative(process.cwd(), targetDir)}`)
+	}
+	doneMessage.push(`   ${getInstallCommand(manager)}`)
```
mimecuvalo, line 76: i forget what Vite does but create-next-app auto-installs the deps which i think is handy - you already know what the command manager is anyway.

packages/create-tldraw/src/main.ts
SomeHats: nope - this pulls from the published template repositories, not our monorepo

packages/create-tldraw/src/main.ts
SomeHats: I did investigate this originally. There's enough cases where it's unreliable and can cause errors that I prefer the vite approach here of leaving it up to the developer. for example, if you're creating an app in an existing monorepo, it's easy for the install to fail if you don't already have workspaces correctly set up for your new app

packages/create-tldraw/src/main.ts
mimecuvalo: ah right right, i see it now

packages/create-tldraw/src/main.ts
mimecuvalo: ok, fair enough

steveruizok:
Screenshot 2025-07-28 at 9 45 36 AM

This didn't work for me

PR #6484, opened 2025-07-21
chore: add performance tracking tooling and e2e tests

mimecuvalo:
We need a way to see if our performance is stable over time. This starts laying a foundation of being able to track via e2e tests and Posthog our performance. @MitjaBezensek esp. since you have the most experience here I believe.

Screengrab of the tests:

https://github.com/user-attachments/assets/8bdb3cfc-f23f-4344-9eea-dc29c1b6449f

Goes up to Posthog:
<img width=

### DIFF
--- a/apps/examples/e2e/tests/baselines/fps-baselines.json
+++ b/apps/examples/e2e/tests/baselines/fps-baselines.json
@@ -23,6 +23,17 @@
 					"browser": "chromium"
 				}
 			},
+			"resize_shapes": {
+				"avgFps": 24,
+				"minFps": 18,
+				"maxFps": 67,
+				"timestamp": "2025-07-21T14:28:23.924Z",
+				"environment": {
+					"platform": "linux",
+					"viewport": "1280x720",
+					"browser": "chromium"
+				}
+			},
 			"canvas_panning": {
 				"avgFps": 54,
 				"minFps": 28,
@@ -37,17 +48,6 @@
 			"canvas_zooming": {
 				"avgFps": 51,
 				"minFps": 19,
-				"maxFps": 67,
-				"timestamp": "2025-07-21T14:28:23.924Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"resize_shapes": {
-				"avgFps": 41,
-				"minFps": 34,
 				"maxFps": 67,
 				"timestamp": "2025-07-21T14:28:23.924Z",
 				"environment": {
--- a/apps/examples/e2e/tests/fixtures/baseline-manager.ts
+++ b/apps/examples/e2e/tests/fixtures/baseline-manager.ts
@@ -190,7 +190,7 @@ export class BaselineManager {
 
 		const isRegression = avgFpsChange < -this.regressionThreshold
 		const isImprovement = avgFpsChange > this.warningThreshold
-		const isSignificantChange = Math.abs(avgFpsChange) > this.warningThreshold
+		const isSignificantChange = Math.abs(avgFpsChange) > this.regressionThreshold
 
 		let status: 'pass' | 'warning' | 'fail' = 'pass'
 		let message = `Performance stable. Average FPS: ${metrics.averageFps} (${avgFpsChange.toFixed(1)}% change)`


apps/examples/e2e/tests/README-PERFORMANCE.md
```diff
@@ -0,0 +1,262 @@
...
+yarn e2e --grep "Performance Tests"
+
+# Run specific performance test
+yarn e2e --grep "Shape Rotation Performance"
+
+# View current baselines
+node scripts/manage-perf-baselines.mjs view
+
+# Establish all baselines manually
+node scripts/manage-perf-baselines.mjs establish
```
ds300: this stuff seems to not apply?

mimecuvalo:
>  This might be a perfect excuse to start running e2e tests on the mac mini?

yeah that'd be killer! i'm tweaking the baseline snapshot now to account for the linux distro but a mac mini would be awesome.

apps/examples/e2e/tests/README-PERFORMANCE.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+## 📊 What It Tests
+
+The suite measures FPS performance for key tldraw interactions:
+
+- **Shape Rotation**: Multi-select rotation of 10+ shapes
+- **Shape Dragging**: Dragging 5+ shapes around the canvas
+- **Canvas Panning**: Space+drag panning across heavy boards
+- **Canvas Zooming**: Wheel zoom in/out cycles
```
MitjaBezensek: It might also be nice to test out switching between two different areas on the canvas to test out culling of shapes. I guess zooming can also do that if we zoom in / out of an area with many shapes.

apps/examples/e2e/tests/README-PERFORMANCE.md
```diff
@@ -0,0 +1,256 @@
...
+- Severity levels (high/medium)
+- Detailed regression information
+
+## 🎯 Usage Examples
+
+### Running Individual Tests
+
+```bash
+# Test specific interaction
+yarn e2e --grep "Shape Rotation Performance"
```
MitjaBezensek: This doesn't seem to work, I guess lazy is not forwarding the arguments:
```ts
lazyrepo 0.0.0-alpha.27
-----------------------
Unknown option `--grep`
```

apps/examples/e2e/tests/fixtures/heavy-board-generator.ts
```diff
@@ -0,0 +1,278 @@
...
+	async selectRandomShapes(count: number): Promise<void> {
+		await this.page.evaluate((count) => {
+			const editor = (window as any).editor
+			if (!editor) return
+
+			const allShapes = editor.getCurrentPageShapes()
+			const selectedCount = Math.min(count, allShapes.length)
+
+			// Select random shapes
+			const shuffled = allShapes.sort(() => 0.5 - Math.random())
```
MitjaBezensek: Should we use the seeded random here as well?

apps/examples/e2e/tests/README-PERFORMANCE.md
MitjaBezensek, line 1: This seems quite verbose, not sure anyone will fully read this file?

apps/examples/e2e/tests/README-PERFORMANCE.md
```diff
@@ -0,0 +1,256 @@
...
+				}
+			}
+		}
+	}
+}
+```
+
+### Regression Detection
+
+- **🔴 Fail**: >10% performance drop (configurable)
```
MitjaBezensek: Feels quite low, but we can always adjust down the line if we feel we need to.

apps/examples/e2e/tests/fixtures/perf-utils.ts
```diff
@@ -0,0 +1,478 @@
...
+				await page.mouse.up()
+			},
+			{ warmupMs, measureMs }
+		)
+
+		// Clear selection
+		const page = (this.fpsTracker as any).page as Page
+		await page.keyboard.press('Escape')
+
+		const comparison = this.baselineManager.compareWithBaseline(
```
MitjaBezensek: Feels like this last part of the test is repeated in all tests, let's pull it out?

MitjaBezensek (approved):
Looking good 👍

Few observations:
- Maybe we should make this into a separate workflow? It would make it easier to see that the performance workflow is failing. Also, we could make it a non required check while we make sure that it's not flaky.
- I wonder if testing via the editor apis vs browser interactions would make any difference? Testing as the user would use it would definitely be a more representative vs just using the apis. Like right now these tests would not catch if we added `sleep(1000)` to our event handlers. So we circumvent all event handler logic, even the whole state machine which also has quite complex logic in some cases.
- It would be really great if we could have something like this running in production. This is testing just a few use cases, while running it in production would run against various shape combinations, user devices, browsers,...

mimecuvalo:
> Maybe we should make this into a separate workflow? It would make it easier to see that the performance workflow is failing. Also, we could make it a non required check while we make sure that it's not flaky.

good call!

> I wonder if testing via the editor apis vs browser interactions would make any difference? Testing as the user would use it would definitely be a more representative vs just using the apis. Like right now these tests would not catch if we added sleep(1000) to our event handlers. So we circumvent all event handler logic, even the whole state machine which also has quite complex logic in some cases.

yeah, i tried doing just browser interactions but it's quite tricky to get dragging and mouse wheel events working properly :-/
i'm open to it for a future revision, or maybe if you wanna tinker around in a future rev to see if you get it working? i had a hard time trying to do so...

> It would be really great if we could have something like this running in production. This is testing just a few use cases, while running it in production would run against various shape combinations, user devices, browsers,...

agreed! i think the tradeoff though is that i would like to also work for PRs. so i think maybe you're suggesting a separate workflow (outside of PR checks) that just logs perf in prod?

apps/examples/e2e/tests/README-PERFORMANCE.md
mimecuvalo: yeah, the zooming is trying to do achieve this (and the panning).

apps/examples/e2e/tests/README-PERFORMANCE.md
mimecuvalo: yeah, i think we'll have to gauge this, not sure if 10 is too high or too low yet 🤔

MitjaBezensek:
> agreed! i think the tradeoff though is that i would like to also work for PRs. so i think maybe you're suggesting a separate workflow (outside of PR checks) that just logs perf in prod?

Yeah, something that runs in production on end user's machines. We'd probably need to add some hooks into our code, things like timing how long resizing takes (when entering resize state start tracking. when we exit the node we'd then stop the timing and somehow emit the duration. then the app could use that to send it to posthog).

.github/workflows/playwright-examples.yml
```diff
@@ -49,6 +49,10 @@ jobs:
       - name: Run Playwright tests
         run: 'yarn e2e'
         working-directory: apps/examples
+        env:
+          PERFORMANCE_ANALYTICS_ENABLED: true
```
MitjaBezensek, line 53: We probably want to remove these, otherwise we'll run perf tests with this workflow as well?

mimecuvalo: i'll fix this in another PR, nice catch - it doesn't matter at the moment because we exclude the perf tests via package.json but i'll clean it up

PR #6488, opened 2025-07-22
perf(sync): optimize TLSyncRoom and improve tombstone pruning

ds300:
We were allocating an awful lot of large collections in this class. This was a serious problem for many hot paths (updating records being the main one).

This PR fixes the biggest issues by

- Using AtomMap to store the main document state
- Using lazy iteration when doing many processes that previously used eagerly-evaluated collections.

While working on this I noticed and fixed an issue with how we handled tombstone pruning. Basically we weren't updating the `tombstoneHistoryStartsAtClock` field properly and it left a hole where deletes might have not been propagated to clients with flaky connections.

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Improve server-side performance when dealing with large rooms in tldraw sync

### API changes
- `TLSocketRoom#getRecord` return type now reflects that it returns `undefined` if the record was not found.
- `StoreSchema#migrateStoreSnapshot` has a new option `mutateInputStore: boolean` which is false by default. If left false, it will copy the input store with `structuredClone`. If set to true the mutators will be allowed to mutate the input store.
- (other internal changes)

### DIFF
--- a/packages/sync-core/src/test/pruneTombstones.test.ts
+++ b/packages/sync-core/src/test/pruneTombstones.test.ts
@@ -150,37 +150,6 @@ describe('TLSyncRoom pruneTombstones', () => {
 		expect(room.tombstones.size).toBe(initialSize)
 	})
 
-	it('should handle needsPrune flag correctly', () => {
-		// Set needsPrune to false
-		;(room as any).needsPrune = false
-
-		// Add tombstones above threshold
-		for (let i = 0; i < 3200; i++) {
-			room.tombstones.set(`doc${i}`, i + 1)
-		}
-
-		// Call pruneTombstones (should not prune since needsPrune is false)
-		;(room as any).pruneTombstones()
-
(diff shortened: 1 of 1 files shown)


packages/store/src/lib/StoreSchema.ts
```diff
@@ -286,13 +310,13 @@ export class StoreSchema<R extends UnknownRecord, P = unknown> {
...
 						if (result) {
-							store[id as keyof typeof store] = structuredClone(result) as any
+							store[id as keyof typeof store] = result as any
 						}
 					}
 				} else if (migration.scope === 'store') {
 					const result = migration.up!(store)
 					if (result) {
-						store = structuredClone(result) as any
+						store = result as any
```
ds300, line 319: these two clones were kinda over-the-top in terms of mutation protection I think. Can't imagine a situation where the data returned from one of these would be retained elsewhere.

PR #6492, opened 2025-07-23
feat(dotcom): notify users when rooms start to get big

MitjaBezensek:
We now update the document and add info about the room usage percentage to the document meta. We then use that in the front end to show file size dialogs to the users.

### Change type

- [x] `improvement`

### Test plan

1. Open a room and add a large number of shapes or assets.
2. Verify that the room usage percentage is updated in the document meta.
3. Verify that a file size dialog appears when the room exceeds certain thresholds.

- [ ] Unit tests (if present)

### Release notes

- Added notifications and dialogs to inform users when their room size is becoming large.

### DIFF
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -83,7 +83,7 @@ async function canAccessTestProductionFile(
 }
 
 const MB = 1024 * 1024
-const ROOM_SIZE_LIMIT_MB = 30
+const ROOM_SIZE_LIMIT_MB = 25
 
 export class TLDrawDurableObject extends DurableObject {
 	// A unique identifier for this instance of the Durable Object
--- a/apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLargeFileHandler.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/sneaky/SneakyLargeFileHandler.tsx
@@ -79,7 +79,7 @@ export function SneakyLargeFileHander() {
 				component: ({ onClose }: { onClose(): void }) => <RoomSizeLimitDialog onClose={onClose} />,
 			})
 			setSizeLimitShown(true)
-		} else if (!sizeLimitShown && !sizeWarningShown && storageUsedPercentage > 80) {
+		} else if (!sizeLimitShown && !sizeWarningShown && storageUsedPercentage > 75) {
 			trackEvent('room-size-warning-dialog-shown', { source: 'dialog' })
 			addDialog({
 				component: ({ onClose }: { onClose(): void }) => (


ds300:
I'm hesitant to go with this approach for the server-side changes. Mainly I don't want to publish these API and protocol changes because they seem very provisional and would be hard to walk back later.

As an alternative, do you think we could use the `document.meta` for this? We could add something like a `percentageOfStorageUsed` value, then after loading the room and after persisting the room we can use the `TLSocketRoom.updateStore` method to set it. Then we can leave it up to the client to decide when to show toasts etc.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -701,10 +719,13 @@ export class TLDrawDurableObject extends DurableObject {
 				this.maybeAssociateFileAssets()
 
 				const key = getR2KeyForRoom({ slug: slug, isApp: this.documentInfo.isApp })
-				await Promise.all([
+				const [roomObject] = await Promise.all([
 					this.r2.rooms.put(key, snapshot),
 					this.r2.versionCache.put(key + `/` + new Date().toISOString(), snapshot),
 				])
+				if (roomObject) {
+					await this.addRoomStorageUsedPercentage(room, roomObject.size / MB, true)
```
MitjaBezensek, line 728: This is not 100% correct. We'll persist the document with the old value and we'll then update the value for the users with the result of the persist. But I think that's probably fine.

MitjaBezensek:
> As an alternative, do you think we could use the `document.meta` for this? We could add something like a `percentageOfStorageUsed` value, then after loading the room and after persisting the room we can use the `TLSocketRoom.updateStore` method to set it. Then we can leave it up to the client to decide when to show toasts etc.

Sounds good, updated the PR with the suggested change.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: if the percent value changes it will persist again, but that's no biggie. it shouldn't happen too often as long as we round the percent value to an integer.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -685,6 +684,26 @@ export class TLDrawDurableObject extends DurableObject {
...
+			for (const record of records) {
+				if (record.typeName !== 'document') continue
+				const document = record as TLDocument
+				const meta = document.meta
+				// In some cases we don't want to update the document if it already has percentage set.
+				// Example for that is when we load the room. If it has a percentage set, we don't want to overwrite it.
+				if (!shouldUpdate && meta.storageUsedPercentage !== undefined) return
+				meta.storageUsedPercentage = Math.ceil((roomSizeMB / ROOM_SIZE_LIMIT_MB) * 100)
+				store.put(document)
+			}
```
ds300: the document id is always `'document:document'` and we have a const for it called `TLDOCUMENT_ID`, so there's no need to iterate over the collection of documents here to find it
```suggestion
		const document = store.get(TLDOCUMENT_ID) as TLDocument
		const meta = document.meta
		// In some cases we don't want to update the document if it already has percentage set.
		// Example for that is when we load the room. If it has a percentage set, we don't want to overwrite it.
		if (!shouldUpdate && meta.storageUsedPercentage !== undefined) return
		meta.storageUsedPercentage = Math.ceil((roomSizeMB / ROOM_SIZE_LIMIT_MB) * 100)
		store.put(document)
```

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -569,7 +563,7 @@ export class TLDrawDurableObject extends DurableObject {
 		const serialized = typeof data === 'string' ? data : JSON.stringify(data)
 		const snapshot = typeof data === 'string' ? JSON.parse(data) : data
 		await this.r2.rooms.put(this._fileRecordCache.id, serialized)
-		return { type: 'room_found' as const, snapshot }
+		return { type: 'room_found' as const, snapshot, roomSizeMB: 0 }
```
ds300: why return 0 here instead of getting the size from the .put result on the line above?

MitjaBezensek: Ah, nice catch.

MitjaBezensek:
> I think as a follow up we might want to do some kind of persistent banner for when you're in the danger zone, so you can delete stuff and get the feedback when you've left the danger zone. But this is good for now.

Yeah, I played with the banner a bit, but it didn't really feel that great. I guess we'd have to explore a few different ideas to come with something good. We could also use something like the offline indicator to show the status.

PR #6494, opened 2025-07-23
feat(editor): add updatePointer method

SomeHats:
Add `editor.updatePointer`, which dispatched a pointer move event in the same screen position as the last pointer event the editor received. This is useful when you want an interaction to update in response to an external change, such as moving the camera or creating a shape.

### Change type

- [x] `api`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add `editor.updatePointer()`, which is useful when you want an interaction to update in response to an external change, such as moving the camera or creating a shape.


TodePond (approved):
Nice!

Maybe beyond the scope of this PR, but this 'auto-pointer-move' still isn't happening when you're following someone.

steveruizok:
For future, this was added for the workflow app's sidebar item dragging interaction.

PR #6510, opened 2025-07-28
Chunked persist.

MitjaBezensek:
When I explored the room size limits I noticed that it's actually persisting that often pushes the durable object to run out of memory. This should definitely help with that.

### Change type

- [x] `improvement`

### DIFF
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -757,10 +757,20 @@ export class TLDrawDurableObject extends DurableObject {
 			// 5MB buffer
 			const fiveMB = 5 * MB
 			const buffer = new Uint8Array(fiveMB)
-			const parts1 = []
-			const parts2 = []
+			const parts1: R2UploadedPart[] = []
+			const parts2: R2UploadedPart[] = []
 			let partNumber = 1
 			let offset = 0
+
+			const uploadBuffer = async (data: Uint8Array) => {
+				const [p1, p2] = await Promise.all([
+					out1.uploadPart(partNumber, data),
+					out2.uploadPart(partNumber, data),
+				])
+				parts1.push(p1)
+				parts2.push(p2)
+				partNumber++
+			}
 
 			for (const chunk of generateSnapshotChunks(snapshot)) {
 				let remainingChunk = chunk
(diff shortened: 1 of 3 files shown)


apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -935,3 +997,23 @@ async function listAllObjectKeys(bucket: R2Bucket, prefix: string): Promise<stri
 
 	return keys
 }
+
+function* generateSnapshotChunks(snapshot: RoomSnapshot, clock: number): Generator<Uint8Array> {
+	const encoder = new TextEncoder()
+
+	yield encoder.encode(`{"clock":${clock},"tombstones":`)
```
MitjaBezensek: Copied this from https://github.com/tldraw/tldraw/pull/6436. Why can't we use the snapshot clock here? Looks like that's what we used before. @ds300

ds300: yeah should be fine to. We should also include the new property `tombstoneHistoryStartsAtClock` that was added in #6488

MitjaBezensek: Sounds good, just added that. 

Feels like this change might make it a bit prone to bugs if we do future updates to the structure of the snapshot. Maybe we should improve the chunk generation to be more generic and pull things out of the snapshot instead of hardcoding the values.

ds300 (approved):
looks good!

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: yeah sounds smart, go for it.

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -935,3 +996,38 @@ async function listAllObjectKeys(bucket: R2Bucket, prefix: string): Promise<stri
 
 	return keys
 }
+
+function* generateSnapshotChunks(snapshot: RoomSnapshot): Generator<Uint8Array> {
```
ds300: might be worth adding some tests for this? could check that serializing and then parsing again produces the same result as the input snapshot?

PR #6512, opened 2025-07-28
feat(sync-core): compress connect payload with gzip

ds300:
I added some inline base64 encoders/decoders for this because `btoa` and `atob` are `string -> string` and we needed `string -> Uint8Array` and `Uint8Array -> string` and i figured it would help us avoid extra allocation but not sure it's worth it now. Will do some perf testing tomorrow to check.

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Opts into gzip for initial data fetch in tldraw sync where available.

### Api changes

all internal


steveruizok:
Closing?

PR #6514, opened 2025-07-29
add custom shape wrapper component

SomeHats:
Sometimes, it's useful to customize how shapes are rendered without digging into the shape itself - for example, to apply custom css classnames to a subset of shapes.

This diff introduces a `ShapeWrapper` component (extracted from the `Shape` component) and makes it overridable in the same way other editor components are. It also adds an example showing how this new API can be used to style a specific shape.

### Change type

- [x] `api`
- [ ] `other`

### API Changes

- Add `ShapeWrapper` to the `components` prop to allow customising how every single shape is rendered to the DOM.

### DIFF
--- a/packages/editor/src/lib/components/default-components/DefaultShapeWrapper.tsx
+++ b/packages/editor/src/lib/components/default-components/DefaultShapeWrapper.tsx
@@ -1,0 +1,33 @@
+import { TLShape } from '@tldraw/tlschema'
+import { forwardRef, ReactNode } from 'react'
+
+/** @public */
+export interface TLShapeWrapperProps {
+	/** The shape being rendered. */
+	shape: TLShape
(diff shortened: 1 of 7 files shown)


packages/editor/src/lib/components/default-components/DefaultShapeWrapper.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { forwardRef, ReactNode } from 'react'
+
+/** @public */
+export interface TLShapeWrapperProps {
+	/** The shape being rendered. */
+	shape: TLShape
+	/** Whether this is the shapes regular, or background component. */
+	isBackground: boolean
+	/** The shape's rendered component. */
+	children: ReactNode
```
ds300, line 12: why not add div props here and spread them in?

PR #6515, opened 2025-07-30
Adds a `data-state` attribute with the current state-tree path to tl-container 

SomeHats:
This means you can write CSS targeting certain tool states. e.g.

```css
/* only interactive in select.idle: */
.tl-container[data-state="select.idle"] .myThing {
  pointer-events: all;
}

/* interactive in all select substates: */
.tl-container[date-state^="select."] .myThing {
  pointer-events: all;
}
```

### Change type
- [x] `api`

### Release notes

- There's now a `data-state` attribute on `.tl-container` which contains the current state tree path. This makes it easier to write styles based on the current interaction state.


mimecuvalo (approved):
🤩

PR #6517, opened 2025-07-30
Improve room history logic.

MitjaBezensek:
There was an issue with rooms that have a large history. We tried to fetch all the timestaps (could be thousands) which caused the worker to take longer than 120s to respond, which means it's execution got canceled due to worker limits.

We now:
* Fetch an initial batch of 1000 items from the room history. If we have fewer than 1000 items we just return them. This keeps the old logic intact for rooms that don't have a big history.
* For rooms with more than 1000 history entries we start from the current month and work our way back to find the first month that has some history entries. Once we find that month we start collecting history entries. Once we get to 1000 (which could span multiple months), we return those.
* We also check if we have any older entries so that users can request more.

### Change type

- [x] `improvement`


ds300 (approved):
nice this looks good.

PR #6524, opened 2025-07-31
Add solo sync mode

ds300:
This makes it so that presence updates only send if someone else is in the room.

### Change type

- [x] `improvement`

### Release notes

- `useSync` now only sends presence updates, like mouse and camera positions, when there are more than one unique user in a room.

### API changes

Internal only

### DIFF
--- a/packages/sync/src/useSync.ts
+++ b/packages/sync/src/useSync.ts
@@ -166,12 +166,12 @@ export function useSync(opts: UseSyncOptions & TLStoreSchemaOptions): RemoteTLSt
 			})
 		})
 
-		const otherUserPresences = store.query.records('instance_presence', () => ({
+		const otherUserPresences = store.query.ids('instance_presence', () => ({
 			userId: { neq: userPreferences.get().id },
 		}))
 
 		const presenceMode = computed<TLPresenceMode>('presenceMode', () => {
-			if (otherUserPresences.get().length === 0) return 'solo'
+			if (otherUserPresences.get().size === 0) return 'solo'
 			return 'full'
 		})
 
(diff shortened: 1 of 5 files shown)


packages/sync/src/useSync.ts
```diff
@@ -165,6 +166,15 @@ export function useSync(opts: UseSyncOptions & TLStoreSchemaOptions): RemoteTLSt
 			})
 		})
 
+		const otherUserPresences = store.query.ids('instance_presence', () => ({
+			userId: { neq: userPreferences.get().id },
+		}))
+
+		const presenceMode = computed<TLPresenceMode>('presenceMode', () => {
```
mimecuvalo, line 173: the main feedback is that we need this to update when someone joins the room. it should be able to:
- switch to `full` mode when someone joins
- make sure we proactively send an initial presence to the other users when they first join
- and of course switch back to solo if everyone leaves.

i see this is `computed` - is it supposed to be updating? i can't get it to switch modes when trying an incognito window and joining the same room 🤔

mimecuvalo (commented):
excited about this :D

main comment is on switching modes — i don't see the `computed` value updating?

packages/sync/src/useSync.ts
ds300: yeah it should update, indeed it's working for me both locally and in the PR preview. So you don't see any presence at all when opening the same room in two different browser contexts?

ds300: i wonder if the initial presence update is not going through on your machine for some reason. I could probably make that more robust by waiting for the presence derivation to become truthy before sending instead of just assuming it will be truthy after the initial diff is handled. I wonder if there's some race condition in the app bootstrap process that makes the user data not present on inital load for you.

Does it work if you open a shared room in two different logged out browser contexts?

mimecuvalo: sry! nevermind about my comment — i checked again and it's fine! i must not have tested i right when i was checking it out

PR #6525, opened 2025-07-31
Fix edge hitting on arrows

steveruizok:
This pull request adds the ability to separate inner and outer margins for shape detection. It fixes a bug where hollow shapes could not be bound-to by arrows when overlapping solid filled shapes.

https://github.com/user-attachments/assets/3866a869-90f3-4400-b20d-02774b71cab0

### Change type

- [x] `bugfix`

### Test plan

1. Create a solid shape
2. Create a hollow shape on top
3. Bind an arrow to the hollow shape

- [x] Unit tests

### Release notes

- Fixed a bug where hollow shapes could not be bound-to by arrows when overlapping a filled shape.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5158,7 +5158,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 		point: VecLike,
 		opts = {} as {
 			renderingOnly?: boolean
-			margin?: number
+			margin?: number | [number, number] // [inside, outside]
+			insideMargin?: number
 			hitInside?: boolean
 			hitLocked?: boolean
 			// TODO: we probably need to rename this, we don't quite _always_
@@ -5166,6 +5167,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 			hitLabels?: boolean
 			hitFrameInside?: boolean
 			filter?(shape: TLShape): boolean
+			debug?: boolean
 		}
 	): TLShape | undefined {
 		const zoomLevel = this.getZoomLevel()
@@ -5178,6 +5180,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 			hitInside = false,
 			hitFrameInside = false,
 		} = opts
+
+		const [innerMargin, outerMargin] = Array.isArray(margin) ? margin : [margin, margin]
 
 		let inHollowSmallestArea = Infinity
 		let inHollowSmallestAreaHit: TLShape | null = null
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5224,13 +5228,13 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 
-			if (this.isShapeOfType(shape, 'frame')) {
+			if (this.isShapeOfType<TLFrameShape>(shape, 'frame')) {
 				// On the rare case that we've hit a frame (not its label), test again hitInside to be forced true;
 				// this prevents clicks from passing through the body of a frame to shapes behind it.
 
 				// If the hit is within the frame's outer margin, then select the frame
 				const distance = geometry.distanceToPoint(pointInShapeSpace, hitInside)
-				if (Math.abs(distance) <= margin) {
+				if (distance < outerMargin || distance > -innerMargin) {
```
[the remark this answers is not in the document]
steveruizok: not a bug

steveruizok:
tests are correct, code is wrong

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5158,7 +5158,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		point: VecLike,
 		opts = {} as {
 			renderingOnly?: boolean
-			margin?: number
+			margin?: number | [number, number] // [inside, outside]
```
SomeHats: mind pulling this out as its own interface and doing this comment as a proper doc comment?

SomeHats: Or you can also write `[inside: number, outside: number]` which is a weird typescript syntax but useful documentation

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5224,13 +5226,24 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+					hitFrameInside
+						? // On hitInside, the distance will be negative for hits inside
+							// If the distance is positive, check against the outer margin
+							(distance > 0 && distance <= outerMargin) ||
+							// If the distance is negative, check against the inner margin
+							(distance <= 0 && distance > -innerMargin)
+						: // If hitInside is false, then sadly _we do not know_ whether the
+							// point is inside or outside of the shape, so we check against
+							// the max of the two margins
+							distance <= outerMargin
```
SomeHats: This comment doesn't match the logic here (no checking against the max of the two margins)

personally i would still pull this out to an `isDistanceWithinMargins(distance, innerMargin, outerMargin, hitInside)` helper given it's complicated enough and i think intended to be the same here and below

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5283,12 +5296,15 @@ export class Editor extends EventEmitter<TLEventMap> {
 				}
 			}
 
+			if (!distance) distance = 0
```
SomeHats: what's this about? as far as i can tell, every branch of the if thing above sets distance. even then, shouldn't this be `Infinity` or something to indicate a miss? At the very least this could use a comment

steveruizok: Thought there was an issue with -0 results, all good to remove.

packages/editor/src/lib/editor/Editor.ts
steveruizok: fixed comments (and a code issue). I'm not going to create a helper fn since I'm not sure it's exactly the same logic as elsewhere.

PR #6526, opened 2025-07-31
refactor(ui): rework toolbar classnames and layout components

SomeHats:
For workflow builder, we want a vertical toolbar. I'm backporting it to the SDK because it's a common enough request (#6532), but in the process i got super confused by two things:

* We have a `tlui-toolbar` classname and a `TldrawUiToolbar` component. These are completely unrelated.
* We have some `tlui-buttons` classnames that are _almost always_ used with `TldrawUiToolbar`, and sometimes contain buttons as the name would suggest.

So this diff tweaks those. `tlui-toolbar` (used for the main actual toolbar at the bottom of the screen) and co are now `tlui-main-toolbar`.

`tlui-buttons` has been replaced with some components that have the same effect: `TldrawUiRow` and `TldrawUiGrid`. These are now used by `TldrawUiToolbar` depending on the `orientation` prop it gets passed. I wanted to use components instead of classnames for these because if you're customizing our UI by copy-pasting chunks of it, random CSS classnames are even less well documented than components like these. I also wanted names that reflected the way these were used a bit better (ie often not with buttons).

### Change type

- [x] `api`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added `TldrawUiRow` and `TldrawUiGrid` layout classes. 
- The `tlui-buttons__horizontal` and `tlui-buttons__grid` class names have been replaced by `tlui-row` and `tlui-grid`
- The `tlui-toolbar` classnames have been renamed to `tlui-main-toolbar`

### DIFF
--- a/apps/docs/utils/TldrawApiModel.ts
+++ b/apps/docs/utils/TldrawApiModel.ts
@@ -131,6 +131,21 @@ export class TldrawApiModel extends ApiModel {
 				return this.resolveToken(component, tokens[2])
 			}
 
+			if (
+				tokens.length === 9 &&
+				tokens[0].text === 'import("react").' &&
+				tokens[1].text === 'ForwardRefExoticComponent' &&
+				tokens[2].text === '<' &&
+				tokens[3].kind === ExcerptTokenKind.Reference &&
+				tokens[4].text === ' & import("react").' &&
+				tokens[5].text === 'RefAttributes' &&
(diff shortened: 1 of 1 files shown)


MitjaBezensek (approved):
Nice 👍

Might be worth adding some of the css class name changes to api changes section? Some users might already be overriding our styles and this might break them.

PR #6527, opened 2025-07-31
SDK changes to the AI module

TodePond:
This PR contains all of the SDK changes to the AI module from the agent starter PR (#6496).

- The `applyChange` function is no longer hardcoded. You can now specify how AI changes are applied (or not) by passing an `apply` function in the same way that you pass your own `generate` and `stream` functions. In the agent template, we use this to extract the diffs from those applied changes and putting them into chat history. The old behavior is exported as `defaultApplyChange`, which is what the option defaults to.
- There's a new `TldrawAi` type. It's the object you get from calling `useTldrawAi`. Having this as a named type makes it easier to pass around the `ai` object into other functions and components.
- The `generate` and `getPrompt` functions now both accept `TldrawAiPromptOptions` as their argument, similar to the `prompt` function. This is partly for consistency and partly so we can pass through more properties, ie: `meta`.
- Speaking of which... the `meta` property is now passed all the way through. It doesn't get abandoned halfway through anymore.
- When using `ai.prompt`, the `message` property is now optional. For example, you might want to prompt with the canvas contents, but no messages.
- Fixed the returned `handleChanges` function so that it actually applies changes now.
- You can now specify the context bounds and prompt bounds of your request. Previously, the `ai.prompt` function didn't accept them as options, so you were effectively hardcoded to the viewport.
- Added some padding to the generated image so that it doesn't error out when you prompt with zero-height/width shapes. Note: We should *actually* be getting an image of the context bounds, including any empty space. That could be a future improvement.
- Errors that happen during generation now get passed back all the way up.
- Some JSDoc comments were moved so that they get picked up by API extractor.

There's also one function, `doesGeometryOverlapPolygon`, that is now exported from the editor, with JSDoc comments.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### DIFF
--- a/packages/ai/src/lib/TldrawAiModule.ts
+++ b/packages/ai/src/lib/TldrawAiModule.ts
@@ -1,19 +1,14 @@
-import {
-	Box,
-	Editor,
-	exhaustiveSwitchError,
-	FileHelpers,
-	structuredClone,
-	TLShapePartial,
-} from 'tldraw'
+import { Box, Editor, FileHelpers, structuredClone } from 'tldraw'
 import { TldrawAiTransformConstructor } from './TldrawAiTransform'
-import { TLAiChange, TLAiContent, TLAiMessages, TLAiPrompt } from './types'
+import { TLAiChange, TLAiContent, TLAiPrompt } from './types'
+import { TldrawAiApplyFn, TldrawAiPromptOptions } from './useTldrawAi'
 import { asMessage } from './utils'
 
 /** @public */
 export interface TldrawAiModuleOptions {
 	editor: Editor
 	transforms?: TldrawAiTransformConstructor[]
+	apply: TldrawAiApplyFn
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/ai/src/lib/TldrawAiModule.ts
```diff
@@ -1,19 +1,14 @@
...
-import { TLAiChange, TLAiContent, TLAiMessages, TLAiPrompt } from './types'
+import { TLAiChange, TLAiContent, TLAiPrompt } from './types'
+import { TldrawAiApplyFn, TldrawAiPromptOptions } from './useTldrawAi'
 import { asMessage } from './utils'
 
 /** @public */
 export interface TldrawAiModuleOptions {
 	editor: Editor
 	transforms?: TldrawAiTransformConstructor[]
+	apply: TldrawAiApplyFn
```
SomeHats: can this be optional?

packages/ai/src/lib/useTldrawAi.ts
```diff
@@ -23,19 +60,32 @@ export type TldrawAiStreamFn = (opts: {
...
 
+/**
+ * The function signature for applying a change to the editor.
+ * @public
+ */
+export type TldrawAiApplyFn = (opts: { change: TLAiChange; editor: Editor }) => void
+
 /** @public */
-export interface TldrawAiOptions extends Omit<TldrawAiModuleOptions, 'editor'> {
+export interface TldrawAiOptions extends Omit<TldrawAiModuleOptions, 'editor' | 'apply'> {
```
SomeHats: I think this is only pulling in transforms at this point. can we just inline it instead of using `omit`? the generated docs will be better

packages/ai/src/lib/TldrawAiModule.ts
TodePond: Done. If you don't set it, it defaults to the default apply change (ie: the old behavior)

packages/ai/src/lib/useTldrawAi.ts
TodePond: Actually, with the change to make `apply` optional within `TldrawAiModuleOptions`, it makes this type definition cleaner again. I've simplified it but kept it more similar to before. WDYT?

PR #6529, opened 2025-08-01
Remove unnecessary conditions

the-moisrex:
`rot(0)` will not change anything, and will cause no problem, the if statements might be there for performance reasons if rotation 0 is common, if that's the case, then we should keep the ifs.

I was just skimming through the project when I saw it, and thought why not do a PR about it. Thanks.

### DIFF
--- a/packages/editor/src/lib/primitives/Vec.ts
+++ b/packages/editor/src/lib/primitives/Vec.ts
@@ -33,7 +33,6 @@ export class Vec {
 	}
 
 	rot(r: number) {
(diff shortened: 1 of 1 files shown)


MitjaBezensek:
Hey!

Thanks for exploring our repo.

This is definitely there as a performance measure as we do call `rot` for many of our actions. One example being during resizing - if you resize a shape that hasn't been rotated we will call this with 0 a bunch of times.

So we should keep this optimization.

PR #6530, opened 2025-08-01
Track room load duration

MitjaBezensek:
Track room load duration and send some basic metrics. Also segment the metrics based on the room size.

[Posthog chart](https://eu.posthog.com/project/45921/insights/URk77Mgl?dashboard=127276&variables_override=%7B%7D) - tried it on preview link for this PR.

### Change type

- [x] `improvement`


ds300 (approved):
Nice. the bucketing is a good idea

PR #6531, opened 2025-08-01
add EditorAtom helper

SomeHats:
When I was working on the workflows starter, I found myself needing to store a lot of state in atoms. This includes:
* The current exectution
* Which ports on which shapes should be highlighted when dragging a connection
* Whether the on-canvas component picker was visible, and which shape it should be attached to

All of this state should really belong to a specific `Editor` instance. It's referencing data from that instance. If I was building a full app and navigated from one workflow to another, I would want all the state to be cleared due to the editor getting re-created. It's possible to get this sort of functionality by using a weak map from editors to atoms, but it takes a fair bit of management. 

This diff adds an `EditorAtom` helper which handles the weak-map juggling. The interface is similar to that of `Editor`, but each method takes an `editor` for scoping.

### Change type

- [x] `api`

### API Changes

- Added `EditorAtom`, a helper for storing editor UI state in an `Atom`

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1549,6 +1549,19 @@ export class Editor extends EventEmitter<TLEventMap> {
     zoomToUser(userId: string, opts?: TLCameraMoveOptions): this;
 }
 
+// @public
+export class EditorAtom<T> {
+    constructor(name: string, getInitialState: (editor: Editor) => T);
+    // (undocumented)
+    get(editor: Editor): T;
+    // (undocumented)
+    getAtom(editor: Editor): Atom<T>;
+    // (undocumented)
+    set(editor: Editor, state: T): T;
+    // (undocumented)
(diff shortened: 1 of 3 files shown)


ds300:
Can you give an example of how you'd use this? I don't really understand why leakage was happening in the workflow builder

SomeHats:
> Can you give an example of how you'd use this? I don't really understand why leakage was happening in the workflow builder

Sure! So here's an example where i use this: https://github.com/tldraw/tldraw/blob/alex/workflow-builder/templates/workflow-builder/src/ports/portState.ts

This `portState` is effectively editor instance state - `hintingPort` is conceptually very similar (and is used in the same way when creating bindings) as `hintingShapeIds` in the editor instance state.

In the workflow builder, I could make this a single global atom without any issue. If I were doing this on tldraw.com though, I wouldn't want to do that. If a user changed documents, then the state from an old document would be trying to get applied to the new document.

PR #6532, opened 2025-08-01
Vertical toolbar example

SomeHats:
Vertical toolbar! Use it like this: 

```tsx
<Tldraw components={{ Toolbar: () => <DefaultToolbar orientation="vertical" /> }} />
```

Right now, I've only supported a vertical toolbar on the left of the screen because to put it elsewhere we'd need a better way of moving other parts of the UI around. This also supports groups in the toolbar, and re-works the overflowing toolbar  component to work both horizontally and vertically.

### Change type

- [x] `api`

### API Changes

- You can now position the toolbar vertically down the right of the screen by setting it's orientation to `vertical`. You can also now use `<TldrawUiMenuGroup />` to group together toolbar items.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
@@ -41,10 +41,14 @@ const NUMBERED_SHORTCUT_KEYS: Record<string, number> = {
 /** @public */
 export interface OverflowingToolbarProps {
 	children: React.ReactNode
+	orientation?: 'horizontal' | 'vertical'
 }
 
 /** @public @react */
-export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
+export function OverflowingToolbar({
+	children,
+	orientation = 'horizontal',
+}: OverflowingToolbarProps) {
 	const editor = useEditor()
 	const id = useUniqueSafeId()
 	const breakpoint = useBreakpoint()
@@ -156,7 +160,7 @@ export function OverflowingToolbar({ children }: OverflowingToolbarProps) {
 		<>
 			<style nonce={editor.options.nonce}>{css}</style>
 			<TldrawUiToolbar
-				orientation="horizontal"
+				orientation={orientation}
 				className={classNames('tlui-main-toolbar__tools', {
 					'tlui-main-toolbar__tools__mobile': breakpoint < PORTRAIT_BREAKPOINT.TABLET_SM,
 				})}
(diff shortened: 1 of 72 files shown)


packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
```diff
@@ -214,3 +326,14 @@ export const isActiveTLUiToolItem = (
...
+function findParentWithClassName(startingElement: HTMLElement, className: string): HTMLElement {
+	let element: HTMLElement | null = startingElement
+	while (element) {
+		if (element.classList.contains(className)) {
+			return element
+		}
+		element = element.parentElement
+	}
+	throw new Error('Could not find parent with class name ' + className)
+}
```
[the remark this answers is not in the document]
SomeHats: intended behaviour

packages/tldraw/src/lib/ui/components/primitives/layout.tsx
```diff
@@ -1,10 +1,55 @@
...
+	tooltipSide?: 'top' | 'right' | 'bottom' | 'left'
+}
+/** @public @react */
+export function TldrawUiOrientationProvider({
+	children,
+	orientation,
+	tooltipSide,
+}: TldrawUiOrientationProviderProps) {
+	const prevContext = useTldrawUiOrientation()
+	const tooltipSideToUse =
```
MitjaBezensek, line 34: Not sure I understand why we need to do this? I guess you might pass in orientation but not the `tooltipSide` and then you want to inherit the parent's tooltip side? Feels like it might lead to some confusion and it might be better to make it required?

MitjaBezensek (commented):
Nice 👏 Quite a few people wanted to have this, you're gonna make them happy 😄

Noticed a few possible UI improvements.

* Looks like there's a bit of additional space on the left side of `More` button when using groups:
image

* We overlap with the minimap in some cases. Maybe we should also have some calculation to start hiding items sooner if this happens?
image

* Also noticed that the quick actions gets covered up by the style panel on mobile, but I think that's fine since you have to intentionally open the style panel and probably don't want to use the quick actions at the same time.
IMG_254E86C7E7A5-1

SomeHats:
> Looks like there's a bit of additional space on the left side of More button when using groups:

fixed! 

> We overlap with the minimap in some cases. Maybe we should also have some calculation to start hiding items sooner if this happens?

Also fixed :)

> Also noticed that the quick actions gets covered up by the style panel on mobile, but I think that's fine since you have to intentionally open the style panel and probably don't want to use the quick actions at the same time.

yup, im gonna leave as-is for now

packages/tldraw/src/lib/ui/components/primitives/layout.tsx
SomeHats: I've added a comment here trying to explain this.

Essentially, the default tooltip side is almost always fine - but when you do change it, you then want it to cascade down until you explicitly change it again. However, if you then switch orientation your tooltip side will be wrong - it'll be covering items in your layout. So in that case, we reset it to a default.

PR #6533, opened 2025-08-01
Unsaved changes examples

steveruizok:
This pull request introduces a new example for tracking and handling unsaved changes in `tldraw` documents. It keeps track of the set of diffs that have accumulated since the last save point and checks whether the diff is empty

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`

### Release notes

- Add an option to `squashRecordDiffs` to allow mutating the first diff instead of creating a new diff.

### API changes

- Add an option to `squashRecordDiffs` to allow mutating the first diff instead of creating a new diff.


ds300:
I think we should ditch the 'json compare' one. The diff squashing one needs a small amount of work (using the mutable squash operation and checking that the diff is empty before setting the state).

I'll fix this up and get it merged.

PR #6543, opened 2025-08-04
Fix dragging within multiple editors

TodePond:
This PR fixes dragging interactions when using multiple tldraw components on the same page. eg: [Inset editors with fixed sizes](https://examples-canary.tldraw.com/inline) and [Common practices with inset editors](https://examples-canary.tldraw.com/inline-behavior)

The bug happened because we're now handling pointer move events on a document basis, not an editor-by-editor basis. This meant that different tldraw editors were killing the pointer move event, preventing other editors from handling it.
This PR fixes the bug by *no longer killing the event*. If we do need to kill the event, perhaps we could kill the event on an editor-by-editor basis by tracking this as a map instead of a boolean or something.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Try the "Inset editor (fixed sizes)" and "Inset editor (common practices)" examples.
2. Try drawing in multiple editors. Make sure you can draw in more than just one of them.

- [ ] Unit tests
- [ ] End to end tests

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -151,9 +151,6 @@ export function useCanvasEvents() {
 		let lastX: number, lastY: number
 
 		function onPointerMove(e: PointerEvent) {
-			if ((e as any).isKilled) return
-			;(e as any).isKilled = true
-
 			if (e.clientX === lastX && e.clientY === lastY) return
 			lastX = e.clientX
 			lastY = e.clientY


packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -151,9 +151,6 @@ export function useCanvasEvents() {
 		let lastX: number, lastY: number
 
 		function onPointerMove(e: PointerEvent) {
-			if ((e as any).isKilled) return
-			;(e as any).isKilled = true
```
MitjaBezensek, line 155: Looks like this was added in the drag out of toolbar PR. @SomeHats did this prevent some issues with that feature? Can we safely remove it?

https://github.com/tldraw/tldraw/pull/6563/files#diff-980beb0aa0ee9aa6d1cd386cef3dc05a500c030638ffb58d45fd11b79126103fR155

SomeHats: i'm not sure - i think @steveruizok added this?

MitjaBezensek: To me it looks like the early return was there before, but the killing of the event was added in that PR https://github.com/tldraw/tldraw/pull/6563/files#diff-980beb0aa0ee9aa6d1cd386cef3dc05a500c030638ffb58d45fd11b79126103fL48

steveruizok: We do this to avoid loops when events are re-dispatched to the canvas, or when events would bubble through shapes. I don't think we rely on this as much since moving away from DOM based hit testing in 2023.

PR #6546, opened 2025-08-05
Exam marking example

max-drake:
This adds an example, built on the `pdf-editor` example, for marking exams. It adds a custom `exam-mark` shape with an input used for marking questions (and associated tool for placing that shape), as well as a widget that sums the total score of all of the the `exam-mark` shapes.

Most of the diff is from copy-pasting the components from the `pdf-editor` example, see
- `add-mark-util.tsx`,
- `add-mark-tool.tsx`,
- `ExamScoreLabel.tsx`, and
- `ui-overrides.tsx`

for the bulk of the actual changes.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`


apps/examples/src/examples/exam-marking/ExamScoreLabel.tsx
```diff
@@ -0,0 +1,62 @@
...
+			}, 0)
+			setScore(newScore)
+		}
+
+		updateScore()
+
+		const unlisten = editor.store.listen(updateScore, { scope: 'document' })
+
+		return unlisten
+	}, [editor])
```
SomeHats:
```suggestion
const score = useValue('score', () => {
    let score = 0;
    for (const shape of editor.getCurrentPageShapes()) {
        if (!editor.isShapeOfType<IExamMarkShape>(shape, 'exam-mark')) continue
        score += shape.props.score
    }
    return score
}, [editor])
```

couple tricks here:
1. isShapeOfType with the type passed in to keep typescript happy without the `as`
2. iterate over the shapes in a single pass instead of 3 (Array.from, filter, reduce)
3. use `useValue` to subscribe to tldraw data in a react context. tldraw is built on this really neat reactivity library that david built, so we don't usually need extra state + effects and whatever to do things like this. you can read a bit here: https://signia.tldraw.dev/docs/what-are-signals. the rough idea tho is that whenever any 'atom' of data that we read in the `useValue` callback changes, we'll re-evaluate that callback. If it produces a different value, then we'll re-render the component with the new value.

apps/examples/src/examples/exam-marking/add-mark-util.tsx
```diff
@@ -0,0 +1,144 @@
...
+		// [b]
+		/* eslint-disable react-hooks/rules-of-hooks */
+		const [score, setScore] = useState<number | string>(shape.props.score)
+
+		const inputRef = useRef<HTMLInputElement>(null)
+
+		// [c]
+		useEffect(() => {
+			if (isEditing && inputRef.current) {
+				inputRef.current.focus()
```
SomeHats, line 60: let's maybe throw an `inputRef.current.selectAll()` in here too

apps/examples/src/examples/exam-marking/add-mark-tool.tsx
```diff
@@ -0,0 +1,25 @@
...
+	override onPointerDown() {
+		const { w, h } = examMarkShapeDefaultProps
+
+		this.editor.createShape<IExamMarkShape>({
+			type: 'exam-mark',
+			x: this.editor.inputs.currentPagePoint.x - w / 2,
+			y: this.editor.inputs.currentPagePoint.y - h / 2,
+		})
+		if (!this.editor.getInstanceState().isToolLocked) {
+			this.editor.setCurrentTool('select')
```
SomeHats, line 21: do you think it's worth doing `this.editor.setEditingShape` in here to immediately start editing the shape? otherwise you have to click then double click to edit

apps/examples/src/examples/exam-marking/add-mark-util.tsx
```diff
@@ -0,0 +1,144 @@
...
+							this.editor.setEditingShape(null)
+						}}
+					/>
+				</div>
+			</HTMLContainer>
+		)
+	}
+
+	override indicator(shape: IExamMarkShape) {
+		return <rect width={shape.props.w} height={shape.props.h} />
```
SomeHats: can we add a radius here to match the one on the text box?

apps/examples/src/examples/exam-marking/add-mark-util.tsx
```diff
@@ -0,0 +1,144 @@
...
+	{
+		w: number
+		h: number
+		score: number
+	}
+>
+
+export const examMarkShapeDefaultProps: IExamMarkShape['props'] = {
+	w: 80,
+	h: 40,
```
SomeHats: if these never change, we can just use constants instead of storing them as props

apps/examples/src/examples/exam-marking/add-mark-tool.tsx
max-drake: ended up being slightly more complicated than this but working now!

SomeHats (approved):
nice!!

PR #6548, opened 2025-08-05
changes room name to avoid overcrowding

sylwiavargas:
This PR introduces a cosmetic change in Quick Start that can otherwise be a blocker for users.

### Problem description

If you just copy-paste the code from the existing Quick Start, you may hit an overused room:

https://github.com/user-attachments/assets/cabd59f6-18db-41ed-8ed0-ca5a9074ef0b

Users who are not familiar with multi-player can have a hard time figuring out the way out.

### Solution

The current change makes it more explicit that the room should have a unique ID by making the example more explicit.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [X] `other`

### DIFF
--- a/apps/docs/content/getting-started/quick-start.mdx
+++ b/apps/docs/content/getting-started/quick-start.mdx
@@ -82,7 +82,7 @@ import { useSyncDemo } from '@tldraw/sync'
 import 'tldraw/tldraw.css'
 
 export default function App() {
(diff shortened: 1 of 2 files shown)


sylwiavargas:

TodePond (approved):

PR #6555, opened 2025-08-06
Add `zoomToBounds` example

max-drake:
Adds an API example showing how to use the editor's `zoomToBounds` function.
re: [this linear issue](https://linear.app/tldraw/issue/ENG-3171/example-request-zooming-to-bounds)

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`


apps/examples/src/examples/zoom-to-bounds/ZoomToBoundsExample.tsx
```diff
@@ -0,0 +1,146 @@
...
+		</div>
+	)
+}
+
+export default function ZoomToBoundsExample() {
+	const position1 = new Vec(1000, 500)
+	const size1 = new Vec(500, 400)
+
+	const position2 = new Vec(50, 100)
+	const size2 = new Vec(900, 720)
```
TodePond: Because these positions are getting hard-coded here, if you move either of the boxes, it moves the camera to a blank space.

apps/examples/src/examples/zoom-to-bounds/README.md
```diff
@@ -0,0 +1,21 @@
...
+category: editor-api
+priority: 1
+keywords: [zoom, camera, bounds, lock, animation, force,]
+---
+
+Demonstrate zooming the camera to specific bounds using the editor's `zoomToBounds` function.
+
+---
+
+This example shows how to use the `zoomToBounds` method to programmatically zoom the camera to specific areas of the canvas. The example creates two boxes at different positions and provides buttons to zoom to each one. Note that the bounds the camera ultimately moves to may be different than the bounds you pass into the function. This is because the values of the bounds are massaged to maintain your viewport's aspect ratio. Also,if an inset is not specified, a default one will be set, adding inset padding to the bounds.
```
TodePond:
```suggestion
This example shows how to use the `zoomToBounds` method to programmatically zoom the camera to specific areas of the canvas. The example creates two boxes at different positions and provides buttons to zoom to each one. Note that the bounds the camera ultimately moves to may be different than the bounds you pass into the function. This is because the values of the bounds are massaged to maintain your viewport's aspect ratio. Also, if an inset is not specified, a default one will be set, adding inset padding to the bounds.
```

apps/examples/src/examples/zoom-to-bounds/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The example demonstrates:
+
+- **Zooming**: Programmatically zoom to different bounds. 
+- **Forced Zooming**: Use the `force` parameter to zoom even when the camera is locked.
+- **Animation Control**: If oyu like, you can animate the canera movement.
```
TodePond:
```suggestion
- **Animation Control**: If you like, you can animate the camera movement.
```

TodePond (commented):
I think it would be worth simplifying this example down to the core thing it's showing, ie: `zoomToBounds`. I think that introducing camera locking distracts from the main thing it's showing. The `force` prop would be good to explore in a separate example, as it's present on all camera-y editor methods, not just `zoomToBounds`.

By reducing the scope of this example, it could probably also lose the CSS file (instead only using a few inline styles if needed).

max-drake:
> I think it would be worth simplifying this example down to the core thing it's showing, ie: `zoomToBounds`. I think that introducing camera locking distracts from the main thing it's showing. The `force` prop would be good to explore in a separate example, as it's present on all camera-y editor methods, not just `zoomToBounds`.
> 
> By reducing the scope of this example, it could probably also lose the CSS file (instead only using a few inline styles if needed).

Thanks for the feedback! I stripped back most of the other stuff so now it's just 2 buttons, one to zoom to each box. 
I ended up not removing the `inset` prop because
a) it's specific to `zoomToBounds()` (not part of `TLCameraMoveOptions`), and 
b) as I mention in the readme, it could potentially be the cause of some behavior someone new to working with the camera in tldraw might not expect because it's set to a nonzero value by default,
so I wanted to call attention to its existence. But happy to delete if you think its better to strip the example back even more.

apps/examples/src/examples/zoom-to-bounds/README.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Zoom to bounds
+component: ./ZoomToBoundsExample.tsx
+category: editor-api
+priority: 1
+keywords: [zoom, camera, bounds, inset]
+---
+
+Demonstrate zooming the camera to specific bounds using the editor's `zoomToBounds` function.
```
TodePond:
```suggestion
Zoom the camera to specific bounds using the editor's `zoomToBounds` function.
```

PR #6558, opened 2025-08-08
Fix: Arrow labels can be truncated when exporting

### LINKED ISSUE #5224: Arrow labels truncated when exporting
### What happened?

When an arrow label overflows the bounding box created by other items, it ends up being truncated in exports.

Given the following starting point

CleanShot 2025-01-15 at 14 12 42

doing a drag-selection over the entire area results in

CleanShot 2025-01-15 at 14 13 01

which if you then export (via the top-level export menu or by right-clicking on the selection) you get

PNG | SVG
---- | ----
truncated | truncated

Firefox

### Where did this happen?

The tldraw.com website

### Contact Details

_No response_

### Code of Conduct

Slowhand0309:
This PR fixes an issue where shapes containing labeled arrows could be truncated when exported.
Fixes #5224.

A patch has been applied to the SVG export process to address this.
If this approach does not meet expectations, please feel free to close this PR.

|before|after|
|-----|-----|
| before | after |

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Exported shapes as PNG with arrow labels extending beyond their bounding box (as shown above) to confirm they were not cut off.
2. Also tested a case where the arrow label did not protrude.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where shapes containing labeled arrows could be truncated when exported.

### DIFF
--- a/packages/editor/src/lib/exports/getSvgJsx.tsx
+++ b/packages/editor/src/lib/exports/getSvgJsx.tsx
@@ -1,5 +1,6 @@
 import { useAtom, useValue } from '@tldraw/state-react'
 import {
+	TLArrowShape,
 	TLFrameShape,
 	TLGroupShape,
 	TLShape,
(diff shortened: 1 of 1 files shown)


mimecuvalo:
thx for the PR @Slowhand0309 !

@SomeHats you're the arrow expert, i think this looks good but could use your eye!

SomeHats:
Thanks for the PR @Slowhand0309! We generally try really hard not to add this sort of per-shape special-casing to the editor (there are a few places in the codebase where it happens, but we're trying to reduce them over time!)

I've put up a PR with an alternative approach [here](https://github.com/tldraw/tldraw/pull/6635) that achieves the same sort of thing without the special casing.

Slowhand0309:
@mimecuvalo @SomeHats 
Thank you for your comment. I understand the direction. The alternative PR is great!

PR #6559, opened 2025-08-08
Add consent form for logged out users.

MitjaBezensek:
Add cookie consent for logged out users.

1. We store this in local storage.
2. If the user then logs in we sync it with the user's signed in options, so that they have a consistent experience when logged in or logged out.
3. If a logged in user changes their preference we also update it in local storage so that it's the same if they log out.
4. For logged out users we also show an additional entry in the main menu, which allows them to change their cookie preferences.
5. Decided to go with a more generic storage of the cookie consent, which can allow us to add consent for cookies with other purposes (like marketing,...)

We'll adjust the appearance of the consent dialog in a follow up PR.

### Change type

- [x] `improvement`


apps/dotcom/client/src/tla/hooks/useAnalyticsConsent.tsx
```diff
@@ -0,0 +1,89 @@
...
+			return user.allowAnalyticsCookie
+		} else {
+			return getStoredAnalyticsConsent()
+		}
+	}, [isSignedIn, user, app])
+
+	const [consent, setConsent] = useState<boolean | null>(getCurrentConsent)
+
+	// Update consent when external state changes
+	useEffect(() => {
```
SomeHats: i think this is fine, but I'd be tempted to drop all the event stuff and instead use an atom in `analytics.tsx`? i think it would be simple. but also happy to go ahead as is!
like:
```tsx
// analytics.tsx
export const cookieConsent = atom('cookie consent', getFromLocalStorage)
react(() => {
    setInLocalStorage(cookieConsent.get())
})

// useAnalyticsConsent.tsx
import {cookieConsent} from './analytics.tsx'
export function useAnalyticsConsent() {
    const storedConsent = useValue(cookieConsent)
    const currentConsent = isSignedIn && user && app ? user.allowAnalyticsCookie : storedConsent

    const updateConsent = (newConsent) => {
        cookieConset.set(newConsent)
        // plus the other stuff
    }
}
```

MitjaBezensek: Sounds good, updated the code with an atom.

PR #6560, opened 2025-08-08
Add delete user button to admin dashboard

ds300:
1. Get cursor to spruce up the admin panel. Some of the styling was fucked.
2. Add a 'delete user' button to the dashboard. User deletions might take some time, especially if the user has a lot of files. So it reports progress via server-sent events.

### Change type

- [x] `other`

### DIFF
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -1,4 +1,4 @@
-import { ZStoreData } from '@tldraw/dotcom-shared'
+import { TlaFile, ZStoreData } from '@tldraw/dotcom-shared'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { Navigate } from 'react-router-dom'
 import { fetch } from 'tldraw'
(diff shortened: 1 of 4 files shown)


apps/dotcom/client/src/pages/admin.tsx
```diff
@@ -1,17 +1,96 @@
...
+	)
+}
+
+// Helper component for user data summary
+function UserDataSummary({ userData }: { userData: any }) {
+	const getUserInfo = () => {
+		const user = userData?.USER?.[0] || userData?.user || userData
+		const files = userData?.FILE || []
+		const deletedFiles = files.filter((f: any) => f.deletedAt)
+		const activeFiles = files.filter((f: any) => !f.deletedAt)
```
MitjaBezensek: This doesn't work, we probably need to look at `isDeleted` field? Also not sure if we are correctly getting the `files` or maybe the `user` in the first place.

MitjaBezensek (approved):
Nice! Found a small bug in display the user data.

apps/dotcom/client/src/pages/admin.tsx
ds300: yeah! i had fixed this. cursor must have sneakily unfixed it at some point.

PR #6564, opened 2025-08-08
feat(examples): add lasso select tool example and export freehand utils

max-drake:
Adds an example that shows how to build a lasso select tool. It also addresses a new api feature: `EditorAtom`s, as well as `Overlays`, (and the difference between Overlays and `InFrontOfTheCanvas`) which aren't addressed in any example yet.

lasso-select-final

### API Changes
I needed some freehand stuff which hadn't been exported from the `tldraw` package yet, @SomeHats seemed to think this was an oversight so I exported it and included that in this PR. If that's an issue then lmk and I will fix :)

### Change type

- [x] `feature`

### Test plan

1. Open the Lasso Select Tool example.
2. Use the tool to select shapes on the canvas.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added a new example for building a lasso select tool and exported freehand SVG utilities.


apps/examples/src/examples/lasso-select-tool/LassoSelectTool.ts
```diff
@@ -0,0 +1,154 @@
...
+//[1]
+export class LassoingState extends StateNode {
+	static override id = 'lassoing'
+
+	info = {} as TLPointerEventInfo
+
+	markId = null as null | string
+
+	//[a]
+	points = new EditorAtom<VecModel[]>('lasso points', () => [])
```
SomeHats: Because this belongs to a `StateNode`, it's already scoped to the editor, so we actually don't need `EditorAtom` here - we can use a normal `atom` instead.

The thing that `EditorAtom` lets you do is store this data in a module-scoped variable. That means that in `LassoSelectSvgComponent`, you don't need to do all the drilling into the statenode to get at this state. instead, you can import the `EditorAtom` and access it directly. 

I think either of these (regular `atom` + drilling or `EditorAtom` + import) are valid approaches, but i wouldn't do both together.

apps/examples/src/examples/lasso-select-tool/LassoSelectToolExample.tsx
```diff
@@ -0,0 +1,156 @@
...
+			const currentTool = editor.getCurrentTool()
+			if (currentTool.id === 'lasso-select') {
+				const lassoTool = currentTool as LassoSelectTool
+				const currentState = lassoTool.getCurrent()
+				if (currentState?.id === 'lassoing') {
+					const lassoingState = currentState as LassoingState
+					return lassoingState.points.get(editor)
+				}
+			}
+			return []
```
SomeHats: If we stick to drilling here, I think you could do something like this instead:
```tsx
if (!editor.isIn('lasso-select.lassoing')) return []
const lassoing = editor.getStateDescendant('lasso-select.lassoing') as LassoingState
return lassoing.points.get(editor)
```

max-drake: i went with the drilling approach! its much cleaner esp with this, ty

SomeHats (approved):

apps/examples/src/examples/lasso-select-tool/LassoSelectTool.ts
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	private doesLassoFullyContainShape(lassoPoints: VecModel[], shape: TLShape): boolean {
+		const { editor } = this
+
+		const geometry = editor.getShapeGeometry(shape)
+		const pageTransform = editor.getShapePageTransform(shape)
+		const shapeVertices = pageTransform.applyToPoints(geometry.vertices)
+
+		const allVerticesInside = shapeVertices.every((vertex) => {
```
steveruizok, line 89: This is the "fine pass", but we could do a "broad pass" using the bounding box.

PR #6568, opened 2025-08-11
namespace css variables

SomeHats:
css variables cascade down through the tree. we've had reports in the past of people having issues because our css variable names, which are fairly generic (`--space-3`, `--color-warning`) clash with those from customers design systems.

this diff takes advantage of the coming 4.0 to rename all our variables to namespace them with `--tl-`, like we do with our classnames.

### Change type

- [x] `api`

### Release notes

- **BREAKING:** all of tldraw's css variables now start with `--tl`. e.g. `--color-background` is now `--tl-color-background`, `--space-4` is now `--tl-space-4`, etc. Names which already started with `--tl-` remain the same.


MitjaBezensek (approved):
Nice!

PR #6569, opened 2025-08-11
fix resize/rotate cursors with tailwindcss

SomeHats:
tailwind's preflight includes [a rule](https://github.com/tailwindlabs/tailwindcss/blob/40f506f31765b004df6b3b6eb9e965f8cfc7cd2d/src/css/preflight.css#L339-L346) that sets `cursor: pointer` on anything with `role="button"`. Properties set with CSS always have higher precedence that ones set with attributes. So the `cursor` attributes we set get overridden by the default tailwind ones. 

This diff switches `cursor` attributes to use style attributes instead.

### Change type

- [x] `bugfix`

### Release notes

- Make sure the correct cursors show for resizing and rotating on pages that use tailwindcss.

### DIFF
--- a/packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
+++ b/packages/tldraw/src/lib/canvas/TldrawSelectionForeground.tsx
@@ -459,7 +459,7 @@ export const ResizeHandle = function ResizeHandle({
 			y={y}
 			width={width}
 			height={height}
-			cursor={cursor}
+			style={{ cursor }}
 			{...events}
 		/>
 	)
@@ -497,7 +497,7 @@ export const RotateCornerHandle = function RotateCornerHandle({
 			y={toDomPrecision(cy - targetSize * 3)}
 			width={toDomPrecision(Math.max(1, targetSize * 3))}
 			height={toDomPrecision(Math.max(1, targetSize * 3))}
-			cursor={cursor}
+			style={{ cursor }}
 			{...events}
 		/>
 	)


steveruizok:
@SomeHats Still want to land this?

SomeHats:
yeah but i couldn't get it past the perf test. @mimecuvalo any ideas?

mimecuvalo:
> yeah but i couldn't get it past the perf test. @mimecuvalo any ideas?

in general, i would take the perf test as advisory for now (it's fairly new still, a month old) and fwiw, it's not a required test to land.

however, that being said, possibly there is a regression here:

i re-ran the test and the rotation is failing, here's the report: https://playwright-reports.tldraw.xyz/VYMqUjbnM/index.html#?testId=066b8d0b8374e0b9b1da-45a432196045ca64b914
you can look at the video and Attachments to see the `stdout`:
```
🔄 Completed 2.1 rotations during measurement
rotate_shapes Performance: 16 FPS (min: 8)
```

other ones are more like:
```
🔄 Completed 4.8 rotations during measurement
rotate_shapes Performance: 35 FPS (min: 24)
```

you can examine the baselines here @SomeHats https://github.com/tldraw/tldraw/blob/main/apps/examples/e2e/tests/baselines/fps-baselines.json

mimecuvalo:
@SomeHats oh yeah this is def. a perf regression. if you look at the `git blame` here, the prev. code was designed _because_ there was a performance issue here that caused issues:
- https://github.com/tldraw/tldraw/pull/6053
- https://github.com/tldraw/tldraw/pull/6023

mimecuvalo:
cc @MitjaBezensek

MitjaBezensek:
Yeah, looks like this suffers from the mentioned issue of constantly downloading the cursor mentioned in #6023. Made a quick pr to help with testing here https://github.com/tldraw/tldraw/pull/6674

As soon as you select the shape you should see the network tab go crazy:

https://github.com/user-attachments/assets/a899cb30-5c0d-422b-a4e8-6f84844e15cb

MitjaBezensek:
Looks like adding `!important` here works as it overrides the style:
https://github.com/tldraw/tldraw/blob/01a46c1f0aea61411222c6921959fd8f4f7ce14c/packages/editor/editor.css#L1136

steveruizok:
Not solving this one

PR #6571, opened 2025-08-11
fix(ui): respect KeyboardShortcutsDialog override and fix TipTap shortcut

cubeghost:
One of my teammates noticed the `Text formatting > Header` shortcut wasn't working, and I found that [TipTap lists it as](https://tiptap.dev/docs/editor/core-concepts/keyboard-shortcuts#paragraph-formatting) `Cmd-Alt-1`, not `Cmd-Shift-1`. 

While testing a fix, I also noticed that using the shortcut for opening the keyboard shortcuts dialog would always bring up the default dialog, even when an override was present. I updated the action to default to the override component using `useTldrawUiComponents()`. I left the default dialog as a fallback thinking that it's potentially an accessibility issue to fully disable this dialog, but I'm not 100% sure.

I tried to write a unit test to cover this (based on [`ContextMenu.test.tsx`](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/test/ui/ContextMenu.test.tsx)) but haven't been able to get the dialog to open with a keyboard event 😩 

| Before | After |
|--------|--------|
| Kapture 2025-08-08 at 20 01 38 | Kapture 2025-08-08 at 20 00 00 |

### Change type

- [x] `bugfix` 

### Test plan

1. Go to the "Changing menus" example, where there is an override for `KeyboardShortcutsDialog` 
2. Open the dialog using the main menu, confirm the contents include "Like my posts"
3. Open the dialog using the keyboard shortcut (`cmd+alt+/`), confirm the contents include "Like my posts"

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- KeyboardShortcutsDialog override is now respected in more places
- Fix an incorrectly listed keyboard shortcut


cubeghost:

SomeHats:
This is great, thank you!

PR #6580, opened 2025-08-13
improvement(ui): remove delay from info tooltips

TodePond:
In a few places, we use tooltips as a kind of info popup. In these cases, we shouldn't have a delay for the tooltip: It feels sluggish / unresponsive, especially on touch devices.

To achieve this, I added a `delayDuration` option to tooltips to let you override your chosen default option.

https://github.com/user-attachments/assets/d45a6611-4c54-437e-a604-37813108447b

https://github.com/user-attachments/assets/67b14ab4-7db8-4b96-bbe2-eff079b5960e

If we decide against this change, an alternative approach could be to keep the delay but to immediately show the tooltip when you click.

### Change type

- [x] `improvement`

### Test plan

1. Open the manage cookies dialog in the dotcom app.
2. Hover the info buttons (or tap them on mobile)
3. Make sure the tooltips appear instantly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added a `delayDuration` option to tooltips to allow overriding the default delay.

### API changes

- Added the `delayDuration` option to tooltips to let you override the delay for a single tooltip.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/primitives/TldrawUiTooltip.tsx
+++ b/packages/tldraw/src/lib/ui/components/primitives/TldrawUiTooltip.tsx
@@ -279,7 +279,9 @@ export const TldrawUiTooltip = forwardRef<HTMLButtonElement, TldrawUiTooltipProp
 		if (!hasProvider) {
 			return (
 				<_Tooltip.Root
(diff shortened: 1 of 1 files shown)


TodePond:
Moved to here
https://github.com/tldraw/tldraw/pull/6591

PR #6582, opened 2025-08-13
Tooltip fixes

SomeHats:
- Remove the animation (fixes INT-2151)
- Remove the title attribute from dropdown menus (fixes INT-2149)
- Prevent tooltips showing on devices that don't support hover (fixes INT-2137)
- Make sure `pointer-events: none` is correctly set on tooltips (fixes INT-2134)

### API Changes

Removes "panel" from `TLUiMenuContextType`, which was unused.

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -4101,7 +4101,7 @@ export interface TLUiMenuContextProviderProps {
 }
 
 // @public (undocumented)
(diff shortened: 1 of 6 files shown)


TodePond:
As I feared, this is preventing info tooltips from appearing on mobile now. Might need to add a special case or prop for those cases.

image

steveruizok:
This should never have been using tooltips :(

PR #6593, opened 2025-08-14
chore: migrate from jest to vitest

SomeHats:
Migrate our test runner from jest to vitest. We've been speaking about this for a little while - we fight jest a fair bit, and vitest just seems to work better out of the box. Vitest has some really good monorepo features, so we switch to it as a task runner for tests instead of lazyrepo so we can get things like global test watch mode.

### Change type

- [x] `other` (or improvement, feature, api, other - pick ONE)

### Test plan

1. Run `yarn test` to ensure all tests pass under vitest

- [x] Unit tests (if present)

### Release notes

- Migrated test runner from Jest to Vitest

### DIFF
--- a/packages/editor/src/lib/editor/managers/SnapManager/SnapManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/SnapManager/SnapManager.test.ts
@@ -1,4 +1,11 @@
-import { TLFrameShape, TLGroupShape, TLPageId, TLShapeId, createShapeId } from '@tldraw/tlschema'
+import {
+	TLFrameShape,
+	TLGroupShape,
+	TLPageId,
+	TLShape,
+	TLShapeId,
+	createShapeId,
+} from '@tldraw/tlschema'
 import { Mocked, vi } from 'vitest'
 import { Box } from '../../../primitives/Box'
 import { Vec } from '../../../primitives/Vec'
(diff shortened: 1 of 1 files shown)


internal/config/eslint-preset-react.js
ds300, line 1: why did these two eslint config files get deleted?

SomeHats: i deleted them because they're unused

packages/editor/src/lib/editor/managers/SnapManager/SnapManager.test.ts
```diff
@@ -330,7 +331,7 @@ describe('SnapManager', () => {
 
 			editor.getSortedChildIdsForParent.mockReturnValue([frameId])
 			editor.getShape.mockReturnValue(frameShape as any)
-			editor.isShapeOfType.mockImplementation((_shape, type) => type === 'frame')
+			editor.isShapeOfType.mockImplementation((_shape: any, type: any) => type === 'frame')
```
ds300: There's more addition of `: any` than i'd like with the mocks. does vitest just have worse types or did we lose some specificity in the TestEditor?

ds300: ah I guess it's the vitest `Mocked` wrapper type being less good than the jest one?

SomeHats: looks like different handling of overloads, but some of these were unneeded - i've removed them

MitjaBezensek:
🎉

PR #6605, opened 2025-08-19
fix(ui): memoize asset urls merging

ds300:
A couple of our users have complained that custom `assetUrls` trigger remounts. This might be why?

### Change type

- [x] `bugfix` 

### Test plan

1. Provide custom assetUrls to the Tldraw component
2. Verify that the component does not remount unnecessarily

- [ ] Unit tests (if present)

### Release notes

- Fixed a bug where the editor would remount in a loop if given custom assetUrls.


SomeHats (approved):
oh yeah that would probably do it...

PR #6612, opened 2025-08-20
fix(ui): disable context menu on style panel buttons

TodePond:
This PR disables the context menu when you right click buttons in the style panel.

Previously, opening the context menu in this way would get your pointer stuck 'down'. And it felt broken because it showed the native context menu instead of our own. Let's disable it instead!

Any a11y considerations here, Mime? Would be good to get your thoughts on this when you're back. If this is the way to go, we could do it for other menus too (eg: toolbar).

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Right click on a button in the style panel
2. Make sure it doesn't show a context menu, but instead selects that style.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- We no longer show the native context menu when you right click a style in the style panel.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
+++ b/packages/tldraw/src/lib/ui/components/primitives/TldrawUiButtonPicker.tsx
@@ -148,6 +148,9 @@ export const TldrawUiButtonPicker = memo(function TldrawUiButtonPicker<T extends
 							onPointerDown={handleButtonPointerDown}
 							onPointerUp={handleButtonPointerUp}
 							onClick={handleButtonClick}
+							onContextMenu={(e) => {
+								e.preventDefault()
+							}}
 						>
 							<TldrawUiButtonIcon icon={item.icon} />
 						</TldrawUiToolbarToggleItem>


mimecuvalo (commented):
after some research, i _think_ this is fine, a11y-wise. i'll bring it up during the next a11y review with Sarah if we decide to go forward with this.

i'm also agreed that if you do it here, you might as well do it on all the buttons.

however, i do think since these are the primitives that it might prevent functionality perhaps from people using our UI components downstream. perhaps a stretch, but it could be people add a set of links in the ButtonPicker (wikipedia, reddit, chatgpt) and then they want to right click→open in a new tab? i dunno, that's a made-up example.

have you thought about just examining `e.button !== 1` in `handleButtonPointerDown`? that seems like it would fix your issue of the button getting stuck.

if we did decide that we want to block right-click's it perhaps needs to be an option that people can turn on/off.

steveruizok:
Not seeing the stuck cursor issue. I'm going to close this one. If we decide to prevent right clicking, which feels strange given how rare that is, then let's do everything.

PR #6613, opened 2025-08-20
fix extra dropdown slot and bad gradient in style panel

SomeHats:
Before: 
Screenshot 2025-08-20 at 10 43 14

After: 
Screenshot 2025-08-20 at 10 43 31

Fixes INT-2225

### Change type

- [x] `other`


MitjaBezensek:
Fixes INT-2225

PR #6619, opened 2025-08-20
Chat template starter kit

SomeHats:
A starter kit demonstrating how you can use tldraw to add visual context to a chat product.

### Change type

- [x] `other`


CLAUDE.md
```diff
@@ -16,31 +16,26 @@ This is the tldraw monorepo - an infinite canvas SDK for React applications. It'
 - `yarn dev-app` - Start tldraw.com client app development
 - `yarn dev-docs` - Start documentation site development
 - `yarn dev-vscode` - Start VSCode extension development
+- `yarn dev-template <template name>` - Runs a template
 - `yarn context` - Find and display nearest CONTEXT.md file (supports -v, -r, -u flags)
 - `yarn refresh-context` - Update CONTEXT.md files using Claude Code CLI
 
-### Building
-
-- `yarn build` - Build all packages using lazy (incremental build system)
```
mimecuvalo, line 24: why rm these?

templates/chat/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+	"compilerOptions": {
+		"target": "es5",
```
mimecuvalo, line 3: `tsconfig.base.json` does `esnext`.
feels like we should change this here and, as i grep around, the 3 other places that do es5

templates/chat/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+		"lib": ["dom", "dom.iterable", "esnext"],
```
mimecuvalo, line 4: actually since i'm nitpicking here anyway, all the templates have pretty different `tsconfig.json`
some have:
`"lib": ["DOM", "DOM.Iterable", "ES2022"],`
another has:
`"lib": ["ES2020", "DOM", "DOM.Iterable"],`
and chat here has yet another one.

can we standardize this somehow?

templates/chat/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+MIT License
+
+Copyright (c) 2024 tldraw Inc.
```
mimecuvalo, line 3: well, 2025 :)

templates/chat/README.md
```diff
@@ -0,0 +1,76 @@
...
+- **`src/components/MessageList.tsx`:** Scrollable message history with loading states
+- **`src/components/ChatMessage.tsx`:** Individual message display component
+- **`src/components/ChatInput.tsx`:** Input field with send functionality
+- **`src/components/WhiteboardModal.tsx`:** Modal component that integrates tldraw for drawing and sketching
+- **`src/app/api/chat/route.ts`:** Next.js API route using Vercel AI SDK for OpenAI integration
+- **`src/app/styles.css`:** CSS with responsive design for all components
+
+## License
+
+This project is provided under the MIT license found [here](https://github.com/tldraw/nextjs-template/blob/main/LICENSE.md). The tldraw SDK is provided under the [tldraw license](https://github.com/tldraw/tldraw/blob/main/LICENSE.md).
```
mimecuvalo: /template/chat

templates/chat/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Trademarks
+
+Copyright (c) 2024-present tldraw Inc. The tldraw name and logo are trademarks of tldraw. Please see our [trademark guidelines](https://github.com/tldraw/tldraw/blob/main/TRADEMARKS.md) for info on acceptable usage.
```
mimecuvalo, line 61: 2025

templates/chat/src/components/ChatInputImage.tsx
```diff
@@ -0,0 +1,23 @@
...
+	image: WhiteboardImage
+	onRemove: () => void
+	onEdit: () => void
+}
+
+export function ChatInputImage({ image, onRemove, onEdit }: ChatInputImageProps) {
+	return (
+		<div className="input-image">
+			{/* the whole image is a clickable area that opens the whiteboard modal. */}
+			<button type="button" className="input-image-edit" onClick={onEdit}>
```
mimecuvalo: a11y nits: some aria-labels on these buttons to indicate what they do

templates/chat/src/components/ChatMessage.tsx
```diff
@@ -0,0 +1,61 @@
...
+export function ChatMessage({ message, onImageClick }: ChatMessageProps) {
+	// Split the message parts into image and text ones
+	const textParts = message.parts.filter((part) => part.type === 'text' && part.text.trim())
+	const imageParts = message.parts.filter((part) => part.type === 'file')
+
+	// For AI messages with no content, show thinking state
+	if (message.role === 'assistant' && textParts.length === 0 && imageParts.length === 0) {
+		return (
+			<div className="message assistant-message thinking-message">
+				<div className="thinking-text">Thinking…</div>
```
mimecuvalo, line 22: totally unnecessary but if you wanted to do a nice animation with the ellipsis might be a nice "loading" touch

templates/chat/src/components/ChatMessage.tsx
```diff
@@ -0,0 +1,61 @@
...
+				const handleImageClick = () => {
+					// if we have a tldraw snapshot, we open the tldraw modal when it's clicked:
+					if (tldrawMetadata) {
+						onImageClick(tldrawMetadata)
+					}
+				}
+
+				return (
+					<button
+						key={index}
```
mimecuvalo: aria-label

templates/chat/src/components/WhiteboardModal.tsx
```diff
@@ -0,0 +1,186 @@
...
+	const handleOverlayClick = (e: React.MouseEvent) => {
+		if (e.target === e.currentTarget) {
+			onCancel()
+		}
+	}
+
+	return (
+		<div className="modal-overlay" onClick={handleOverlayClick}>
+			<Tldraw
+				components={components}
```
mimecuvalo, line 124: something odd, when the modal shows (for an existing whiteboard), it auto-focuses a random(?) shape. why is that? might not be related to your PR

templates/chat/README.md
```diff
@@ -0,0 +1,76 @@
...
+- Click the whiteboard button to open the tldraw canvas
+- Draw, sketch, and create diagrams to supplement conversations
+- Annotate images and visual content directly on the canvas
+
+## Environment Setup
+
+Create a `.env.local` file in the root directory and add your OpenAI API key:
+
+```
+OPENAI_API_KEY=your_openai_api_key_here
```
mimecuvalo, line 38: b/c i'm a lazy eng i loaded up the template without setting this (knowing that i prbly needed to).
i was surprised that there's no error/or toast in the UI to notify that the key is missing. prbly some feedback to the dev that the key should be set?

templates/chat/src/components/ChatInput.tsx
```diff
@@ -0,0 +1,230 @@
...
+		if (textareaRef.current) {
+			// Reset height to auto to get the correct scrollHeight
+			textareaRef.current.style.height = 'auto'
+			// Set height based on scrollHeight, with max height for ~5 lines
+			const maxHeight = 5 * 24 // Approximate line height * 5 lines
+			textareaRef.current.style.height = `${Math.min(textareaRef.current.scrollHeight, maxHeight)}px`
+
+			// Scroll to bottom when textarea grows
+			if (scrollToBottom) {
+				scrollToBottom('instant')
```
mimecuvalo: i don't think it should scroll while you're typing (compare to chatgpt).
sometimes you want to reference something above to write out a prompt, but then the chat is scrolling away while you type so it feels frustrating

templates/chat/src/components/ChatInput.tsx
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}, [disabled])
+
+	// Auto-resize textarea and scroll to bottom when content changes.
+	useLayoutEffect(() => {
+		if (textareaRef.current) {
+			// Reset height to auto to get the correct scrollHeight
+			textareaRef.current.style.height = 'auto'
+			// Set height based on scrollHeight, with max height for ~5 lines
+			const maxHeight = 5 * 24 // Approximate line height * 5 lines
```
mimecuvalo: to make this easier to reuse as a template, prbly you could use `5lh` if that's the intent:
https://developer.mozilla.org/en-US/docs/Web/CSS/length#lh

templates/chat/src/components/ChatInput.tsx
```diff
@@ -0,0 +1,230 @@
...
+			)}
+
+			{/* the main input is a text area. we resize it automatically to fit its content. */}
+			<div className="input-container">
+				<textarea
+					ref={textareaRef}
+					value={input}
+					onChange={(e) => dispatch({ type: 'setInput', input: e.target.value })}
+					onKeyDown={handleKeyDown}
+					placeholder={disabled ? '' : 'Type your message...'}
```
mimecuvalo: nit: ellipsis `…`

templates/chat/src/components/ChatInput.tsx
```diff
@@ -0,0 +1,230 @@
...
+					</div>
+				)}
+			</div>
+
+			{/* below the input we have several controls: */}
+			<div className="chat-input-bottom">
+				{/* a button to upload an image */}
+				<TldrawUiTooltip content="Upload an image">
+					<button
+						type="button"
```
mimecuvalo, line 184: aria-label in this file

templates/chat/src/app/layout.tsx
```diff
@@ -0,0 +1,14 @@
+import './styles.css'
+
+export const metadata = {
+	title: 'tldraw chat starter kit',
```
mimecuvalo, line 4: weird, i just noticed that `<title>` and this metadata is put at the bottom of the `body` tag, maybe only in dev? weird

templates/chat/src/app/page.tsx
```diff
@@ -0,0 +1,14 @@
+'use client'
+import { TldrawUiTooltipProvider } from 'tldraw'
+import { Chat } from '../components/Chat'
+
+export default function Home() {
+	return (
+		<main className="tl-container tl-theme__light">
```
mimecuvalo: took me a sec to figure this out, annoying thing about `tl-container` is that it sets `user-select: none` which prevents copy/pasting text from the chat. let's override that pls!

CLAUDE.md
SomeHats: i found claude only ever using this in times when they weren't helpful

templates/chat/src/components/WhiteboardModal.tsx
SomeHats: it will select what you had selected before! i may switch that off tho

templates/chat/LICENSE.md
SomeHats: this is everywhere i'll update it separately

templates/chat/README.md
SomeHats: will do for everything in a followup

templates/chat/src/components/ChatMessage.tsx
SomeHats: i'm gonna skip this for now

templates/chat/tsconfig.json
SomeHats: hard to standardize because these all have to have their own tsconfigs, so we can't use a single shared one. i'm gonna leave this as-is for now, it doesn't really make a difference.

mimecuvalo: no worries, i think i'll create a ticket somewhere to standardize this a bit though, in case it causes some inscrutable bug down the line

templates/chat/LICENSE.md
mimecuvalo: yeah it's everywhere, but for new stuff liike this we should have the newer date

PR #6626, opened 2025-08-20
Fix misaligned InFrontOfCanvas ui when canvas is inset

SomeHats:
Before: 
Screenshot 2025-08-20 at 17 14 45

After: Screenshot 2025-08-20 at 17 15 06

In order to make this work, we need to:
1. Make `InFrontOfTheCanvas` a child of `tl-canvas`
2. Put the on-canvas parts of our default UI in `InFrontOfTheCanvas`.

### Change type

- [x] `bugfix`

### Release notes

- On canvas UI renders in the correct place when the canvas is inset from the tldraw container.

### API Changes

- **BREAKING**: `InFrontOfTheCanvas` now renders as a child of `.tl-canvas`, in a `.tl-canvas__in-front` wrapper.

### DIFF
--- a/packages/tldraw/src/lib/Tldraw.tsx
+++ b/packages/tldraw/src/lib/Tldraw.tsx
@@ -33,7 +33,7 @@ import { registerDefaultSideEffects } from './defaultSideEffects'
 import { defaultTools } from './defaultTools'
 import { EmbedShapeUtil } from './shapes/embed/EmbedShapeUtil'
 import { allDefaultFontFaces } from './shapes/shared/defaultFonts'
-import { TldrawUi, TldrawUiProps } from './ui/TldrawUi'
+import { TldrawUi, TldrawUiInFrontOfTheCanvas, TldrawUiProps } from './ui/TldrawUi'
 import { TLUiAssetUrlOverrides, useDefaultUiAssetUrlsWithOverrides } from './ui/assetUrls'
 import { LoadingScreen } from './ui/components/LoadingScreen'
 import { Spinner } from './ui/components/Spinner'
@@ -118,6 +118,18 @@ export function Tldraw(props: TldrawProps) {
 
 	const _components = useShallowObjectIdentity(components)
 
+	const CustomInFrontOfTheCanvas = components?.InFrontOfTheCanvas
+	const InFrontOfTheCanvas = useMemo(() => {
+		if (rest.hideUi) return CustomInFrontOfTheCanvas ?? null
+		if (!CustomInFrontOfTheCanvas) return TldrawUiInFrontOfTheCanvas
+
+		return () => (
+			<>
+				<TldrawUiInFrontOfTheCanvas />
+				<CustomInFrontOfTheCanvas />
+			</>
+		)
+	}, [rest.hideUi, CustomInFrontOfTheCanvas])
 	const componentsWithDefault = useMemo(
 		() => ({
 			Scribble: TldrawScribble,
@@ -129,8 +141,9 @@ export function Tldraw(props: TldrawProps) {
 			Spinner,
 			LoadingScreen,
 			..._components,
+			InFrontOfTheCanvas,
 		}),
-		[_components]
+		[_components, InFrontOfTheCanvas]
 	)
 
 	const _shapeUtils = useShallowArrayIdentity(shapeUtils)
--- a/packages/tldraw/src/lib/ui/TldrawUi.tsx
+++ b/packages/tldraw/src/lib/ui/TldrawUi.tsx
@@ -223,13 +223,23 @@ const TldrawUiContent = React.memo(function TldrawUI() {
 					</div>
 				</>
 			)}
-			{RichTextToolbar && <RichTextToolbar />}
-			{ImageToolbar && <ImageToolbar />}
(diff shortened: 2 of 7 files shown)


packages/tldraw/src/lib/ui/TldrawUi.tsx
```diff
@@ -223,13 +219,23 @@ const TldrawUiContent = React.memo(function TldrawUI() {
 					</div>
 				</>
 			)}
+			{Toasts && <Toasts />}
```
mimecuvalo, line 221: why aren't these in front of the canvas as well?

packages/tldraw/src/lib/Tldraw.tsx
```diff
@@ -118,6 +118,18 @@ export function Tldraw(props: TldrawProps) {
...
+		if (rest.hideUi) return CustomInFrontOfTheCanvas ?? null
+		if (!CustomInFrontOfTheCanvas) return TldrawUiInFrontOfTheCanvas
+
+		return () => (
+			<>
+				<TldrawUiInFrontOfTheCanvas />
+				<CustomInFrontOfTheCanvas />
+			</>
+		)
+	}, [rest.hideUi, CustomInFrontOfTheCanvas])
```
mimecuvalo, line 132: hah, funny enough, this was the original plan to create TldrawUiInFrontOfTheCanvas for the ImageToolbar: https://github.com/tldraw/tldraw/pull/4443/commits/f66e026f9ff05e9cb496c80c79773070a52c8b51#diff-397ba121a4ac79fc30fd86ca1b8e900a1061c4947a192f6166a1a2958978acf5

but we decided against it because it would override people's use of InFrontOfTheCanvas. i see you've neatly added the ability for a Custom one to go on top of the `TldrawUiInFrontOfTheCanvas`.

however...i guess now people can't override that component specifically which doesn't seem great...
say, if they want to create a custom toolbar set, they won't be able to. say, if they want to only use the ImageToolbar:
```
CustomInFrontOfTheCanvas:
<>
			{ImageToolbar && <ImageToolbar />}
			<MyCustomToolbar />
</>
```
they won't be able to opt out of the RichTextToolbar, FollowingIndicator and whatever else we put in there (unless they go nuclear and use hideUi?)

basically, can we just keep the toolbars within `TldrawUiContent`?

mimecuvalo (commented):
btw, how did you recreate the issue you have on your "Before:" screenshot, was that specific to the Chat template? I'm looking at the https://examples.tldraw.com/inset but it dosen't have that issue.

SomeHats:
> btw, how did you recreate the issue you have on your "Before:" screenshot, was that specific to the Chat template? I'm looking at the https://examples.tldraw.com/inset but it dosen't have that issue.

https://examples.tldraw.com/inset-canvas

packages/tldraw/src/lib/ui/TldrawUi.tsx
SomeHats: The toasts and dialogs don't depend on the canvas position, they can be anywhere. they're much more a part of the wrapping UI than a part of the canvas in the way that the other ones are.

packages/tldraw/src/lib/Tldraw.tsx
SomeHats: You can disable or override each of these with `components`:
```tsx
<Tldraw components={{ ImageToolbar: null, RichTextToolbar: null, etc }} />
```

We were missing the ability to override `FollowingIndicator` in this way, so I've added that.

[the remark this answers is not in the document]
SomeHats: this is intentional

mimecuvalo: - FollowingIndicator was added in a separate PR? maybe it wasn't pushed up
- i think maybe my bigger concern is the composing here (this deviates from the consistency of the other editor components vs. other slots that simply override). it's less that you can add the `FollowingIndicator` overridability, more that it ends up being whack-a-mole when we inevitably end up adding more things in here, right? feels like that'll be surprising for people that expected to have more control over this slot.

i think maybe what i'm saying is, as long as this is a breaking change anyway, maybe you create a new slot altogether? `InFrontOfCanvasUi`? i dunno, some naming that isn't too confusing. wdyt?

SomeHats: oh yeah i forgot to push up following indicator sorry.

i don't think i buy the whack-a-mole thing here. We never needed to override just these few specific components as a group before, why would we need that now? My reluctance here is that the part of the API we'd need to add this to is `<TldrawEditor />` - not `<Tldraw />`. That already has `InFrontOfTheCanvas`, and the purpose of that has always been "here's a blank slot for you chuck your own things" - which hasn't changed here. Adding a second one that's like "here's another blank slot! ignore it or use it just to turn off these things" feels like it would simplify things for us, but complicate the API story? but maybe that's actually the right tradeoff here im not sure

mimecuvalo: I don't wanna be a blocker here btw, I disagree somewhat but commit :)
It's prbly fine :)

PR #6627, opened 2025-08-20
Allow custom per-shape padding in exports

SomeHats:
When working with images, it's often really annoying that tldraw adds padding when exporting back to an image - it's not needed if it'm using tldraw to make some annotations on top of an image, or composing some images together. This diff makes the export padding option accept a function as well as a number that can return a different amount of padding for different shapes. By default, we return 32 (the previous default), or 0 for image shapes.

Also taking the opportunity of 4.0 to rename this to something more consistent with other export related options.

Pulled out of #6619. Closes ENG-3651.

### Change type

- [x] `improvement`

### Release notes

- Image exports of images no longer add unneeded padding

### API Changes

- You can now pass a function to the export `padding` option to add different padding to different shapes in the export
- **BREAKING:** `TldrawOptions.defaultSvgPadding` has been renamed to `TldrawOptions.defaultExportPadding`

### DIFF
--- a/packages/editor/src/lib/exports/getSvgJsx.tsx
+++ b/packages/editor/src/lib/exports/getSvgJsx.tsx
@@ -44,7 +44,7 @@ export function getSvgJsx(editor: Editor, ids: TLShapeId[], opts: TLImageExportO
 		scale = 1,
 		// should we include the background in the export? or is it transparent?
 		background = editor.getInstanceState().exportBackground,
-		padding = editor.options.defaultSvgPadding,
+		padding = editor.options.defaultExportPadding,
 		preserveAspectRatio,
 	} = opts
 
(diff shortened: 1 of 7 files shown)


packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -61,13 +61,18 @@ export function getSvgJsx(editor: Editor, ids: TLShapeId[], opts: TLImageExportO
...
+			const expandedBounds =
+				paddingForShape === 0
+					? maskedPageBounds
+					: maskedPageBounds.clone().expandBy(paddingForShape)
 			if (bbox) {
-				bbox.union(maskedPageBounds)
+				bbox.union(expandedBounds)
 			} else {
-				bbox = maskedPageBounds.clone()
+				bbox = expandedBounds
```
ds300, line 75: isn't this gonna end up with weird stuff happening like if you draw something near the edge of the image it'll add padding to that edge?

image

SomeHats: hmmmmm yes i will rethink

SomeHats:
hmmmmm yes

PR #6628, opened 2025-08-21
Fix mobile touch handling inside editing shape

### LINKED ISSUE #6617: Cannot interact with editable custom shape example on touch screens
### What happened?

On this example https://examples.tldraw.com/editable-shape

You can double tap to get into the 'editable' state, but pressing the button has no effect on touch devices.

Reported on discord https://discord.com/channels/859816885297741824/1407034158080983141/1407034158080983141

_No response_

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

ds300:
Fixes #6617 

Touch events (specifically touchend), on anything other than anchors or text inputs, were being suppressed inside of editable shapes on mobile.

I think this fix preserves the previous callout for '.tl-text-content' by just allowing any pointer events on the children of an editable shape's wrapper div. @mimecuvalo you added this I think, is there a situation where we'd want to allow pointer events on a .tl-text-content but the shape itself isn't in the 'editing' state?

### Change type

- [x] `bugfix`

### Release notes

- Fixed a bug that prevented certain pointer events from firing in editable shapes on mobile.

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -79,15 +79,16 @@ export function useCanvasEvents() {
 				// check that e.target is an HTMLElement
 				if (!(e.target instanceof HTMLElement)) return
 
+				const editingShapeId = editor.getEditingShape()?.id
 				if (
+					// if the target is not inside the editing shape
+					!(editingShapeId && e.target.closest(`[data-shape-id="${editingShapeId}"]`)) &&
+					// and the target is not an clickable element
+					e.target.tagName !== 'BUTTON' &&
 					e.target.tagName !== 'A' &&
+					// or a text input ??
 					e.target.tagName !== 'TEXTAREA' &&
-					!e.target.isContentEditable &&
-					// When in EditingShape state, we are actually clicking on a 'DIV'
-					// not A/TEXTAREA/contenteditable element yet. So, to preserve cursor position
-					// for edit mode on mobile we need to not preventDefault.
-					// TODO: Find out if we still need this preventDefault in general though.
-					!(editor.getEditingShape() && e.target.className.includes('tl-text-content'))
+					!e.target.isContentEditable
 				) {
 					preventDefault(e)
 				}


packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -79,15 +79,16 @@ export function useCanvasEvents() {
 				// check that e.target is an HTMLElement
 				if (!(e.target instanceof HTMLElement)) return
 
+				const editingShapeId = editor.getEditingShape()?.id
 				if (
+					// if the target is not inside the editing shape
+					!(editingShapeId && e.target.closest(`[data-shape-id="${editingShapeId}"]`)) &&
+					// and the target is not an clickable element
+					e.target.tagName !== 'BUTTON' &&
```
mimecuvalo: the tricky thing is that you could have not just buttons, but a lot of interactive ui elements, input sliders, radio buttons, select menus, etc...this PR might be just fixing this narrow example we have, but really we want to improve interactivity with these React components in general

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -79,15 +79,16 @@ export function useCanvasEvents() {
...
 					e.target.tagName !== 'A' &&
+					// or a text input ??
 					e.target.tagName !== 'TEXTAREA' &&
-					!e.target.isContentEditable &&
-					// When in EditingShape state, we are actually clicking on a 'DIV'
```
mimecuvalo, line 86: would be worth preserving this comment

mimecuvalo (commented):
i...am not sure :) i think, first, this editable/focus/touch stuff is tricksy enough where i'd have to test with some physical devices when i'm back in town, specifically iOS is the naughty one usually. (testing double-tapping and preservation of editable state/focus, as per the deleted comment). so you should test that for sure with a physical device, esp that the virtual keyboard doesn't become wonky.

but also, i'm not sure that specifically whitelisting only buttons is only a patch fix for this example bug. we'd want to step back and think about interactivity within editable components a little more, see my related comment. (it's also a little unclear to me why the A/anchor tag is in here specifically... bookmarks?)

this is def. hitting on a general problem with rich React components within our shapes: we tell people to override event handling, do a lot of preventDefault/stopPropagation to make things work, but it's a bit nutty at the moment in that regard.

ds300:
I removed the button whitelist. Forget about that.

the actual change here is to make all touchend events work inside of editable shapes. That's why the comment was deleted, it no longer applied since it was talking about a carveout for only certain touchend events inside editing shapes. But I'm pretty sure we should allow all touchend events inside editing shapes?

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -79,15 +79,15 @@ export function useCanvasEvents() {
 				// check that e.target is an HTMLElement
 				if (!(e.target instanceof HTMLElement)) return
 
+				const editingShapeId = editor.getEditingShape()?.id
 				if (
+					// if the target is not inside the editing shape
+					!(editingShapeId && e.target.closest(`[data-shape-id="${editingShapeId}"]`)) &&
+					// and the target is not an clickable element
 					e.target.tagName !== 'A' &&
+					// or a text input ??
```
mimecuvalo: to help your `??` confusion → 
our `TextArea.tsx` uses `<textarea>`

mimecuvalo (approved):
pre-approving modulo:
- nit on the textarea comment, it's `TextArea.tsx`
- make sure to test on iOS!

PR #6629, opened 2025-08-21
feat(dotcom): track new file creation

MitjaBezensek:
We can use this as a baseline of how long it takes to load a room. We can break it down by creation source (duplication, slurping, regular blank file creation).

### Change type

- [x] `feature` 

### Test plan

1. Create a new file from various sources (blank, duplication, slurp)
2. Verify tracking events are fired correctly

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added tracking for new file creation sources to monitor room loading performance.


mimecuvalo (approved):
ty!

PR #6633, opened 2025-08-22
fix: always render hidden dialog title in dialog to satisfy radix and a11y

### LINKED ISSUE #6518: Dialog example throws error when showing simple dialog
### What happened?

If you open our [dialog example](https://tldraw.dev/examples/toasts-and-dialogs) and then click on the `Show simple dialog` you get this error in the console:

Image

The reasons seems to be the fact that we are using radix as our dialog wrapper. Radix then expect to have a radix title element as one of it's children (or at least a VisuallyHidden wrapped title if you don't want to show the title). 

This makes the example a bit confusing, as it is not clear how to really use a custom dialog

_No response_

_No response_

### Where did this happen?

The tldraw.dev website

### Contact Details

mitja.bezensek@gmail.com

### Code of Conduct

pokycookie:
Fixes #6518

### Problem
- In the simple dialog example on the docs site ([toasts-and-dialogs](https://tldraw.dev/examples/toasts-and-dialogs))
- Radix UI Dialog expects a `<Dialog.Title>` to exist in the subtree

image

### Changes
- `packages/tldraw/src/lib/ui/components/Dialogs.tsx`
- Add `aria-label` attribute to `<_Dialog.Content>`.
- Always render a hidden `<_Dialog.Title>` wrapped in `<VisuallyHidden.Root>` with `aria-hidden`, inside `<_Dialog.Content>`.

### Impact
- No API changes. No breaking changes.
- Removes Radix warnings in simple cases while preserving accessibility defaults.
- Improves DX by making dialogs safe-by-default even when users don’t render a visible title.

1) When users pass a component without a visible title (e.g., the “Simple dialog” example), Radix’s rule fails because there’s no `<Dialog.Title>` in the subtree. The hidden title ensures Radix’s requirement is satisfied; `aria-label` provides a fallback accessible name.
2) When users provide `<TldrawUiDialogTitle>`, Radix will use that as the accessible name. The hidden title does not affect rendering (`VisuallyHidden`) and does not interfere with a11y (`aria-hidden`).

### Change type

- [x] `bugfix`

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Dialogs.tsx
+++ b/packages/tldraw/src/lib/ui/components/Dialogs.tsx
@@ -1,5 +1,5 @@
 import { useContainer, useValue } from '@tldraw/editor'
-import { Dialog as _Dialog } from 'radix-ui'
+import { Dialog as _Dialog, VisuallyHidden } from 'radix-ui'
 import { memo, useCallback, useRef } from 'react'
 import { TLUiDialog, useDialogs } from '../context/dialogs'
 
@@ -49,6 +49,7 @@ const TldrawUiDialog = ({
 						dir="ltr"
 						className="tlui-dialog__content"
 						aria-describedby={undefined}
+						aria-label="Dialog"
 						onMouseDown={() => (mouseDownInsideContentRef.current = true)}
 						onMouseUp={() => (mouseDownInsideContentRef.current = false)}
 						onInteractOutside={(e) => {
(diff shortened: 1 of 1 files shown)


pokycookie:

mimecuvalo:
Ach, after reviewing this PR I don't think this will be a good direction forward.
I know I encouraged you to create this PR but I think we'll have to step back and think more holistically on our dialog API. I'm thinking we should probably add a required `title` to `TLUiDialog` (or something along those lines) b/c a title _should_ be required for a dialog to be created properly (even if hidden). But that requires a fairly large refactor which we would need to do internally.

pokycookie:
Apologies for the delayed reply.
Thanks for the review and guidance. I fully understand and respect the decision, and I agree it’s the right call.
I’ll close this PR. Thank you!

PR #6636, opened 2025-08-22
feat(editor): skip padding on image exports

SomeHats:
Take 2 of #6627. instead of custom per-shape padding, shapes can declare themselves as 'export bounds containers'. If all shapes in an export are within the bounds of one of these containers, the padding is skipped: 
image

There're still cases where you might suddenly get padding where you don't expect it (e.g. the bounds of a shape is outside of the container even if it's visually contained) but at least in that case you get even padding all the way aroung.

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests (if present)

### Release notes

- When exporting an annotated image, we no longer add padding unless the annotations stray over the bounds of the image.

### API changes

- Adds `ShapeUtil.isExportBoundsContainer`. If this is true, exports of your shape won't have padding if their bounds contain all other shapes in the export.

### DIFF
--- a/packages/editor/src/lib/exports/getSvgJsx.tsx
+++ b/packages/editor/src/lib/exports/getSvgJsx.tsx
@@ -147,7 +147,7 @@ export function getExportDefaultBounds(
 			// things like annotated images.
 			if (isContainer && Box.ContainsApproximately(maskedPageBounds, bbox)) {
 				isBoundedByContainer = true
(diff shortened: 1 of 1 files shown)


steveruizok:
Seems like a good idea, let's use it on the frame shape too!

ds300 (approved):
Good solution!

Issue #6141, opened 2025-05-28
Crash when duplicating objects with duplicate fractional index


rudi-c:
### What happened?

When two sibling shapes have the same fractional index, `duplicateShapes` crashes:

```
    at generateKeyBetween (index.js:401:11)
    at generateJitteredKeyBetween (index.js:416:15)
    at spreadGeneratorResults (index.js:438:13)
    at generateNJitteredKeysBetween (index.js:424:10)
    at getIndexBetween (reordering.ts:79:9)
    at Editor.ts:5936:19
    at Array.forEach (<anonymous>)
    at Editor.ts:5929:32
    at transact (transactions.ts:299:10)
    at HistoryManager.batch (HistoryManager.ts:95:5)
```

In our case, the issue with duplicate fractional indices probably comes from our own mistake creating a new shape by copy of another, so I wouldn't expect this to be a common issue. However, tldraw does try to support multiplayer and even with jittered fractional indices, it's possible to get duplicate indices.

This is another example of application code [should never throw an exception](https://github.com/tldraw/tldraw/issues/6076), especially in the context of using tldraw as a library where there's no way to wrap the code in a try-catch. But tricky in this case given the rarity of the crash & the fact that it's throw in a third-party library (fractional-indexing-jittered).

### Where did this happen?

The developer package

### Contact Details

rudi@arcol.io

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
Thanks for the feedback. Yeah, definitely not ideal at the moment, as you mention, it's in third-party code so it's a touch hard to mitigate. It is quite rare for it to happen but yes, you're right this is something we want to handle better in the future 👍

giorgiogross:
Also running into this; probably my own fault bc I've migrated my existing objects in my db to use all the same default key (tldraw seems to be ok with that) but then when copying one of them it always creates a dupe with the same fractional index, and throws this error.

steveruizok:
This issue was addressed in #6646, which upgraded the fractional indexing system to use a more robust implementation (jittered-fractional-indexing package). The new package better handles edge cases with duplicate fractional indices and prevents the crash described in this issue.

Issue #5864, opened 2025-04-09
Fractional indices can contain trailing zeroes


rudi-c:
### What happened?

An invariant of fractional indices is that they should not end with trailing zeroes.

The fractional library that tldraw uses doesn't seem to preserve this invariant. In fact, if you compare that library (TMeerhof/fractional-indexing-jittered) with the one it's originally based on (rocicorp/fractional-indexing)

https://github.com/TMeerhof/fractional-indexing-jittered/blob/725dfd034ab3efd5c4f4164d76bc0e0829aca299/src/integer.ts#L18
```
export function validateOrderKey(
  orderKey: string,
  charSet: IndexedCharacterSet
) {
  // TODO more checks
  getIntegerPart(orderKey, charSet);
}
```

https://github.com/rocicorp/fractional-indexing/blob/51050cdfeb8e30733b7351e2d3096995eac8b5c5/src/index.js#L115
```
function validateOrderKey(key, digits) {
  if (key === "A" + digits[0].repeat(26)) {
    throw new Error("invalid order key: " + key);
  }
  // getIntegerPart will throw if the first character is bad,
  // or the key is too short.  we'd call it to check these things
  // even if we didn't need the result
  const i = getIntegerPart(key);
  const f = key.slice(i.length);
  if (f.slice(-1) === digits[0]) {
    throw new Error("invalid order key: " + key);
  }
}
```

The former does not inspire confidence.

We ran into this because other parts of our app uses rocicorp/fractional-indexing because reasons. I can't imagine others would have the same problem, but there's a potential for other bugs whenever invariants are not correctly maintained, it's just hard to recreate the exact edge case.

### How can we reproduce the bug?

I got this to happen by creating a frame, a rectangle, and just dragging the rectangle in and out of the frame repeatedly.

### Where did this happen?

The developer package

### Contact Details

rudi@arcol.io

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

SomeHats:
This is a great issue, thanks rudi! 

I think the path forward for us here would be to:

1. Make sure this invariant is respected 
2. Add a migration to all records that use fraction indexes to strip trailing 0s that might already exist in stored data. Off the top of my head these records would be:
   1. Pages
   2. Shapes
   3. Lines (props.points)

mimecuvalo:
- If we want to respect the invariant, I assume we'd have to make an upstream patch yeah @SomeHats. Was that something you wanted to raise?
[...]

PR #6646, opened 2025-08-26
improvement(utils): use jittered-fractional-indexing package

mimecuvalo:
We added jittered indexes in this PR https://github.com/tldraw/tldraw/pull/4312 building off of strict types in this PR https://github.com/tldraw/tldraw/pull/2827

This definitely helped with collisions that we were having, esp. pages with a large # of shapes.

However, the package we were using (`[fractional-indexing-jittered](https://github.com/TMeerhof/fractional-indexing-jittered/tree/main)`) wasn't perfect. As described astutely in one ticket, "it doesn't inspire confidence".

@nathanhleung took it upon himself to create a new fork of the original package and created: https://github.com/nathanhleung/jittered-fractional-indexing
which is based off of
https://github.com/rocicorp/fractional-indexing

which after review does inspire more confidence.

Here are the related issues:
- https://github.com/tldraw/tldraw/issues/5864
- https://github.com/tldraw/tldraw/issues/6141

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improved shape ordering reliability by switching to the jittered-fractional-indexing package.

### DIFF
--- a/packages/tlschema/src/store-migrations.ts
+++ b/packages/tlschema/src/store-migrations.ts
@@ -89,6 +89,10 @@ export const storeMigrations = createMigrationSequence({
 					}
 				}
 			},
+			down: () => {
+				// noop
+				// Enables tlsync to support older clients so as to not force people to refresh immediately after deploying.
+			},
 		},
 	],
 })
--- a/packages/tlschema/src/migrations.test.ts
+++ b/packages/tlschema/src/migrations.test.ts
@@ -228,7 +228,7 @@ describe('Store removing Icon and Code shapes', () => {
 })
 
 describe('Fixing index keys', () => {
-	const { up } = getTestMigration(storeVersions.FixIndexKeys)
+	const { up, down } = getTestMigration(storeVersions.FixIndexKeys)
 	test('up works as expected', () => {
 		const snapshot = [
 			ShapeRecord.create({
@@ -255,6 +255,19 @@ describe('Fixing index keys', () => {
 		expect(fixed.find((s) => s.id === 'shape:2')?.index).not.toBe('a00')
 		expect(fixed.find((s) => s.id === 'shape:2')?.index).toMatch(/^a0[1-9A-Za-z]{3}$/)
 		expect(fixed.find((s) => s.id === 'shape:3')?.index).toBe('a111')
+	})
+
+	test('down works as expected', () => {
+		const snapshot = [
+			ShapeRecord.create({
+				type: 'shape',
+				id: 'shape:1',
+				parentId: 'page:any',
+				index: 'a00',
+			} as any),
+		]
+		const unchanged = snapshot.map((shape) => down(shape))
+		expect(unchanged.find((s) => s.id === 'shape:1')?.index).toBe('a00')
 	})
 })
 


packages/tlschema/src/store-migrations.ts
```diff
@@ -63,5 +67,42 @@ export const storeMigrations = createMigrationSequence({
 				}
 			},
 		},
+		{
+			id: Versions.FixIndexKeys,
+			scope: 'store',
+			up: (store) => {
```
SomeHats: miggghhht be worth doing this as record-specific migrations so we can write `down` migrations for them (no-op) - otherwise we're forcing everyone to reload their browser when this goes live

ds300: it can still be just one migration, only with a 'record' scope.

@mimecuvalo 'store' scope is only needed for when you need to add/remove records usually. or if there are data dependencies between different records that need to be evaluated during the migration.

nathanhleung:
> love this! personally i don't think we need to increase the random bits - the place where the jittering is useful is when two clients create a shape in the same index at the same time. it's already extremely rare

glad the new package could be helpful! and yes — to clarify, the [collision probability formula](https://github.com/nathanhleung/jittered-fractional-indexing/blob/main/README.md#generatekeybetween) i put in the [jittered-fractional-indexing readme](https://github.com/nathanhleung/jittered-fractional-indexing/blob/main/README.md) is specific to repeated `generateKeyBetween` calls with the **same** arguments `a` and `b`.

assuming fractional indices are unique, each adjacent pair of indices in your app splits the space into discrete chunks that can never collide across each other. concretely, the ~0.742% collision probability applies when someone creates two shapes and 3998 clients all try to insert a new shape between those two shapes at the same time

mimecuvalo:
> assuming fractional indices are unique, each adjacent pair of indices in your app splits the space into discrete chunks that can never collide across each other. concretely, the ~0.742% collision probability applies when someone creates two shapes and 3998 clients all try to insert a new shape between those two shapes at the same time

makes sense! i think the theoretical case i'm (prbly) over-compensating for here is bringing a ton of shapes forward/back at the same time. so if i select 2000 shapes, say that are on top of another 2000 shapes, and then i bring them all forward/back at once. see the example video below (correct me if i'm wrong/overthinking here @SomeHats or @ds300 )

https://github.com/user-attachments/assets/e9731281-52da-4cc8-b761-c1e8c8113479

SomeHats:
@mimecuvalo we allocate all of those shapes different indexes according to their initial position in that case. 

So even if you and I both selected 2000 shapes and both brought them to the front at the exact same time, you would still only have n=2 for the purposes of collisions, but 2000 times.

mimecuvalo:
> @mimecuvalo we allocate all of those shapes different indexes according to their initial position in that case.
> 
> So even if you and I both selected 2000 shapes and both brought them to the front at the exact same time, you would still only have n=2 for the purposes of collisions, but 2000 times.

this is why it came to mind, we've had exceptions of this nature, not too long ago:
Screenshot 2025-08-27 at 12 58 58

SomeHats:
this looks like invalid keys getting generated, not collisions?

packages/tlschema/src/store-migrations.ts
mimecuvalo: ah killer, thanks @ds300. pushed up the change, can you verify?

packages/tlschema/src/store-migrations.ts
```diff
@@ -63,5 +66,40 @@ export const storeMigrations = createMigrationSequence({
 				}
 			},
 		},
+		{
+			id: Versions.FixIndexKeys,
+			scope: 'record',
```
ds300, line 71: can you add a noop down migration please. This will enable tlsync to support older clients and it won't force people to refresh immediately after deploying.

ds300 (changes requested):

packages/tlschema/src/store-migrations.ts
mimecuvalo: cool, pushed up

PR #6657, opened 2025-08-27
revert: "a11y: add tooltip for opacity slider"

SomeHats:
Reverts tldraw/tldraw#6647

This causes the tooltip on the cropping indicator to appear when it shouldn't when entering cropping mode. Also, tooltips on on-canvas controls don't correctly follow their target's positions: 
Kapture 2025-08-27 at 14 19 56

### Change type

- [x] `bugfix` 

### Test plan

1. Enter cropping mode and verify the tooltip does not appear incorrectly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where tooltips would incorrectly appear when entering cropping mode.


mimecuvalo:
nice catch, thanks

PR #6663, opened 2025-08-27
a11y: add tooltip for opacity slider and image/text toolbars

mimecuvalo:
re-introduces https://github.com/tldraw/tldraw/pull/6647 which was reverted b/c there was issues with dbl-clicking on cropping (https://github.com/tldraw/tldraw/pull/6657)

it also fixes image toolbar missing tooltips, the toolbar examples, and the alt and link editors.

@SomeHats i didn't resolve the tooltip not following the camera but meh, i don't think shows up very much after this change.
i _could_ add something in `TooltipSingleton` to track the `editor.getCamera()` to hide the tooltip if the camera moves, maybe that's worth doing?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- a11y: add tooltip for opacity slider and image/text toolbars


packages/tldraw/src/lib/ui/components/primitives/TldrawUiSlider.tsx
```diff
@@ -49,9 +52,24 @@ export const TldrawUiSlider = React.forwardRef<HTMLDivElement, TLUiSliderProps>(
...
+		tooltipManager.hideAllTooltips()
 		onHistoryMark('click slider')
 	}, [onHistoryMark])
 
+	// N.B. This is a bit silly. The Radix slider auto-focuses which
+	// triggers TldrawUiTooltip handleFocus when we dbl-click to edit an image,
+	// which in turn makes the tooltip display prematurely.
+	// This makes it wait until we've focused to show the tooltip.
+	useEffect(() => {
+		tltime.setTimeout(
```
SomeHats: because this is in react we should use a proper teardown here instead of relying on our custom timer fallback teardown stuff

SomeHats (commented):
we're missing a translation string here:
Screenshot 2025-08-28 at 10 56 05

also, let's set `tooltipSide="top"` on the wrapping `TldrawUiToolbar` so the tooltip doesn't obscure the image.

otherwise lgtm!

PR #6665, opened 2025-08-28
Fix internal license logic.

MitjaBezensek:
Looks like we never properly errored out on a expired internal license. 

We should probably hotfix this so it goes out in the 3.x branch?

### Change type

- [x] `bugfix`

### Release notes

- Fix a bug with expired internal license check.

### API changes

* Add an `internal-expired` state to the license manager.

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.ts
+++ b/packages/editor/src/lib/license/LicenseManager.ts
@@ -73,10 +73,9 @@ export class LicenseManager {
 	public isDevelopment: boolean
 	public isTest: boolean
 	public isCryptoAvailable: boolean
-	state = atom<'pending' | 'licensed' | 'licensed-with-watermark' | 'unlicensed'>(
-		'license state',
-		'pending'
-	)
+	state = atom<
+		'pending' | 'licensed' | 'licensed-with-watermark' | 'unlicensed' | 'internal-expired'
+	>('license state', 'pending')
 	public verbose = true
 
 	constructor(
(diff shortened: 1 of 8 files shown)


packages/editor/src/lib/license/LicenseProvider.tsx
```diff
@@ -16,5 +17,12 @@ export function LicenseProvider({
 	children: ReactNode
 }) {
 	const [licenseManager] = useState(() => new LicenseManager(licenseKey))
+	const licenseState = useValue(licenseManager.state)
+
+	// If internal license has expired, don't render the editor at all
+	if (licenseState === 'internal-expired') {
+		return null
```
mimecuvalo: good call on not throwing anymore, that'll line up with our new idea to just have a blank screen.
there should be something in the `console` though to show the dev that something is wrong

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -367,15 +375,18 @@ export class LicenseManager {
 	static className = 'tl-watermark_SEE-LICENSE'
 }
 
-export function isEditorUnlicensed(result: LicenseFromKeyResult) {
-	if (!result.isLicenseParseable) return true
-	if (!result.isDomainValid && !result.isDevelopment) return true
+export function isEditorUnlicensed(result: LicenseFromKeyResult): {
```
mimecuvalo: i think this function is doing a bit too much now, helper-wise. esp. if we have, like, a `trial-version-expired` it'll get messy. let's keep this just to return a boolean and then we can handle other cases above

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -89,21 +88,30 @@ export class LicenseManager {
...
-		this.getLicenseFromKey(licenseKey).then((result) => {
-			const isUnlicensed = isEditorUnlicensed(result)
+		this.getLicenseFromKey(licenseKey)
+			.then((result) => {
+				const { unlicensed, internalExpired } = isEditorUnlicensed(result)
 
-			if (!this.isDevelopment && isUnlicensed) {
-				fetch(WATERMARK_TRACK_SRC)
-			}
+				if (!this.isDevelopment && unlicensed) {
```
mimecuvalo: nit: keep this as `isUnlicensed` — just makes it readable/denotes as a boolean quickly

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -89,21 +88,30 @@ export class LicenseManager {
...
+					fetch(WATERMARK_TRACK_SRC)
+				}
 
-			if (isUnlicensed) {
+				if (unlicensed) {
+					if (internalExpired) {
```
mimecuvalo: here we can just check `result.isInternalLicense` directly — i'd rather the helper function not try to do this for me

MitjaBezensek: What if it unlicensed since it's not parseable? We'd need to repeat all the checks that `isEditorUnlicensed` is doing.

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -89,21 +94,20 @@ export class LicenseManager {
...
-				fetch(WATERMARK_TRACK_SRC)
-			}
+				if (!this.isDevelopment && licenseState === 'unlicensed') {
+					fetch(WATERMARK_TRACK_SRC)
+				}
 
-			if (isUnlicensed) {
+				this.state.set(licenseState)
+			})
+			.catch((error) => {
```
mimecuvalo, line 107: good call

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -89,21 +94,20 @@ export class LicenseManager {
...
-		this.getLicenseFromKey(licenseKey).then((result) => {
-			const isUnlicensed = isEditorUnlicensed(result)
+		this.getLicenseFromKey(licenseKey)
+			.then((result) => {
+				const licenseState = getLicenseState(result)
 
-			if (!this.isDevelopment && isUnlicensed) {
-				fetch(WATERMARK_TRACK_SRC)
-			}
+				if (!this.isDevelopment && licenseState === 'unlicensed') {
```
mimecuvalo, line 101: hm, `licensed-with-watermark` should also track, yeah? i forget actually

MitjaBezensek: It didn't do so before, so unsure. Do we actually have any of these out there or is it just our own license?

PR #6671, opened 2025-08-28
Fix over-measuring arrow labels

steveruizok:
This PR fixes a performance issue on arrows with empty labels. Previously we would measure the size of the letter "i" every time an arrow without a label rendered. We would only ever have to do this when the shape is editing, so that the empty state still has the correct bounds... but we can also just stash those "i" sizes somewhere so that we don't have to go down the slow prosemirror flatten + dom measurement process.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Improved arrow performance

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowPath.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowPath.tsx
@@ -10,8 +10,8 @@ export function getArrowBodyPath(shape: TLArrowShape, info: TLArrowInfo, opts: P
 				.moveTo(info.start.point.x, info.start.point.y, { offset: 0, roundness: 0 })
 				.lineTo(info.end.point.x, info.end.point.y, { offset: 0, roundness: 0 })
 				.toSvg(opts)
-		case 'arc': {
-			const path = new PathBuilder()
+		case 'arc':
+			return new PathBuilder()
 				.moveTo(info.start.point.x, info.start.point.y, { offset: 0, roundness: 0 })
 				.circularArcTo(
 					info.bodyArc.radius,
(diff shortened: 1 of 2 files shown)


SomeHats:
did you notice this was causing an issue? i'm really trying to avoid these sorts of non-reactive caches because they've caused so many bugs for us in the past (we should account at least for the things bugbot mentioned).

given this is only used for the initial editing state and then is cached for that particular shape, i'm surprised this was causing problems. you can't modify the arrow while you're editing the label, so this wont happen every frame, and if it does it's no more expensive than it would be from there being a label anyway?

mimecuvalo:
talked offline a bit - we should solve this up higher in the call stack. we have some checks to not call `getArrowLabelPosition` when unnecessary (in `getGeometry`) but those checks are missing in `component`, for example. so either we avoid calling where necessary or `getArrowLabelPosition` should just have a short-circuit path for the empty case.

steveruizok:
> did you notice this was causing an issue? i'm really trying to avoid these sorts of non-reactive caches because they've caused so many bugs for us in the past (we should account at least for the things bugbot mentioned).
> 
> given this is only used for the initial editing state and then is cached for that particular shape, i'm surprised this was causing problems. you can't modify the arrow while you're editing the label, so this wont happen every frame, and if it does it's no more expensive than it would be from there being a label anyway?

That's not what's happening. For labels with text content, html is rendered and cached but is remeasured on every update (this is ok). For labels without text, html is rendered on every update and measured on every update. It does not need to be.

https://github.com/user-attachments/assets/dcd2ea37-a174-4a21-8c85-c57725a9edf4

mimecuvalo:
@steveruizok created an alt take here fwiw https://github.com/tldraw/tldraw/pull/6673

SomeHats:
let's use @mimecuvalo's version of this?

steveruizok:
Closed in favor of #6673, thanks @mimecuvalo

PR #6678, opened 2025-09-01
feat: add agent starter template

TodePond:
This PR adds a starter template for building an agent chat app with tldraw.

To try the template, run `yarn dev-template agent` from root.
You'll also need to place a `.dev.vars` file into the `templates/agent` folder. There, you'll want to set api keys for these values, or at the very least, at least Anthropic.

```
OPENAI_API_KEY=
ANTHROPIC_API_KEY=
GOOGLE_API_KEY=
```

### Change type

- [x] `feature`

### Test plan

1. Run `yarn dev-template agent` from the root.
2. Add a `.dev.vars` file with an `ANTHROPIC_API_KEY` to `templates/agent`.
3. Verify the agent chat app loads and functions.

- [ ] Unit tests (if present)

### Release notes

- Added agent starter template.

### DIFF
--- a/templates/agent/client/agent/TldrawAgent.ts
+++ b/templates/agent/client/agent/TldrawAgent.ts
@@ -741,7 +741,6 @@ function requestAgent({ agent, request }: { agent: TldrawAgent; request: AgentRe
 
 	const requestPromise = (async () => {
 		const prompt = await agent.preparePrompt(request, helpers)
-		console.log('PROMPT SENT', prompt)
 		let incompleteDiff: RecordsDiff<TLRecord> | null = null
 		const actionPromises: Promise<void>[] = []
 		try {


templates/agent/worker/models.ts
```diff
@@ -0,0 +1,75 @@
...
+		name: 'gemini-2.5-pro',
+		id: 'gemini-2.5-pro',
+		provider: 'google',
+		thinking: true,
+	},
+	'gpt-4.1': {
+		name: 'gpt-4.1',
+		id: 'gpt-4.1-2025-04-14',
+		provider: 'openai',
+	},
```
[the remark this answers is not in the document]
TodePond: Don't worry bug bot it's fine we're in the future

templates/agent/README.md
```diff
@@ -0,0 +1,578 @@
...
+		_type: z.literal('clear'),
+	})
+	// A title and description tell the model what the action does
+	.meta({
+		title: 'Clear',
+		description: 'The agent deletes all shapes on the canvas.',
+	})
+
+// Infer the action's type
+type IClearAction = z.infer<typeof ClearAction>
```
SomeHats: super nit but i find the `I` prefix so confusing, i can never remember when i need it and when i don't. Could this be `type ClearAction = z.infer<typeof ClearAction>` instead?

templates/agent/README.md
```diff
@@ -0,0 +1,578 @@
...
+	transform: AgentRequestTransform
+) {
+	if (!action.complete) return
+	const { agent } = transform
+
+	// Schedule a follow-up request so the agent can use the data
+	agent.schedule("Here's a random Wikipedia article.")
+
+	// Fetch from the external API
+	return await fetchRandomWikipediaArticle()
```
SomeHats: don't need to go into depth but it might be worth having this example show how you pass the article back into the agent - like:
```tsx
const article = fetchRandomWikipediaArticle()

return {
  // i don't know what an action looks like
  actionContents: article
}
```

also: is there ever a situation where you would want to return data _without_ scheduling a follow-up request?

templates/agent/README.md
```diff
@@ -0,0 +1,578 @@
...
+The agent can see and move, delete, and arrange any custom shapes out of the box. However, it will see the shape's `type` as `'unknown'`, with a `subType` field that will show the internal name of the shape's `type`. It will also only be able to see the shape's `shapeId` and `x` and `y` coordinates. If you want the agent to be able to see and edit other props, you can add your custom shape to the schema we use to help the model understand shapes.
+
+Let's add support for in-canvas embeds, and allow the model to read and update their urls if they like. TLdraw has a builtin `TLEmbedShape` already, so we'll be using that.
+
+1. Define the fields of your new shape in `shared/format/SimpleShape.ts` using a zod object. Every shape is required to have a `_type` and a `shapeId` field, and it's strongly recommended that you give it a `note` field as well, as that is where the model stores information about the shape's purposes when creating or updating it.
+   These fields are what the model will see, so it's worthwhile to give them descriptive names.
+
+```ts
+const SimpleEmbedShape = z.object({
+	_type: z.literal('bookmark'),
```
SomeHats:
```suggestion
	_type: z.literal('embed'),
```

??

templates/agent/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+## License
+
+This project is part of the tldraw SDK. It is provided under the [tldraw SDK license](https://github.com/tldraw/tldraw/blob/main/LICENSE.md).
+
+You can use the tldraw SDK in commercial or non-commercial projects so long as you preserve the "Made with tldraw" watermark on the canvas. To remove the watermark, you can purchase a [business license](https://tldraw.dev#pricing). Visit [tldraw.dev](https://tldraw.dev) to learn more.
+
+## Trademarks
+
+Copyright (c) 2024-present tldraw Inc. The tldraw name and logo are trademarks of tldraw. Please see our [trademark guidelines](https://github.com/tldraw/tldraw/blob/main/TRADEMARKS.md) for info on acceptable usage.
```
SomeHats: 2025

templates/agent/client/agent/TldrawAgent.ts
```diff
@@ -0,0 +1,647 @@
...
+		for (const util of Object.values(promptPartUtils)) {
+			const untransformedPart = await util.getPart(request, this)
+			const transformedPart = util.transformPart(untransformedPart, transform)
+			if (!transformedPart) continue
+			transformedParts.push(transformedPart)
+		}
+
+		const agentPrompt = Object.fromEntries(transformedParts.map((part) => [part.type, part]))
+
+		console.log('PROMPT', agentPrompt)
```
SomeHats: remove

templates/agent/client/agent/TldrawAgent.ts
```diff
@@ -0,0 +1,647 @@
...
+		transform = new AgentRequestTransform(this)
+	): AgentActionResult {
+		const { editor } = this
+		const util = this.getAgentActionUtil(action._type)
+		this.isActing = true
+		let value: AgentActionResult['value']
+		const diff = editor.store.extractingChanges(() => {
+			value = util.applyAction(structuredClone(action), transform) ?? undefined
+		})
+		this.isActing = false
```
SomeHats:
```suggestion
		this.isActing = true
		let value: AgentActionResult['value']
		try {
			const diff = editor.store.extractingChanges(() => {
				value = util.applyAction(structuredClone(action), transform) ?? undefined
			})
		} finally {
			this.isActing = false
		}
```

in case applying fails

templates/agent/client/agent/dedupeShapesContextItem.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Remove duplicate shapes from a shapes context item.
+ * If there's only one shape left, return it as a shape item instead.
+ *
+ * @param item - The shapes context item to strip duplicates from.
+ * @param existingItems - The existing context items to check against.
+ * @returns The new context items with duplicates removed.
+ */
+export function dedupeShapesContextItem(
```
SomeHats: i guess there's some fancy ai magic reason why we need a separate shape + shapes type right? we couldn't do just `shapes`?

templates/agent/client/agent/requestAgent.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Send a request to the agent and handle its response.
+ * This is a helper function that is used internally by the agent.
+ *
+ * @param agent - The agent to send the request to.
+ * @param request - The request to send to the agent.
+ * @param onError - A callback to call if an error occurs.
+ * @returns A promise for the request and a function to cancel it.
+ */
+export function requestAgent({
```
SomeHats: ok so there's a few things like this where by putting them in their own files, i think they start to look like either parts of the API or like... bigger concepts than they need to be. This fn for example is only used by `TldrawAgent`, and is effectively part of the implementation of `.request`. I might be wrong but I don't think you'd ever want to call this on its own, it should always be done via `.request`.

For this starter, I would make this a non-exported function defined in `TldrawAgent.ts`. I think having this as a separate file would be fine in a library where we can control what's made public or not, but for a starter I think we only really have files as a way of encapsulating parts of the implementation / API.

Obvs this is pretty bike-sheddy/nitpicky, but i'd be curious to get your take. not just this file either but generally how we structure this and other starters.

templates/agent/client/agent/requestAgent.ts
```diff
@@ -0,0 +1,114 @@
...
+		}
+		agent.$chatHistory.update((prev) => [...prev, promptHistoryItem])
+	}
+
+	let cancelled = false
+	const controller = new AbortController()
+	const signal = controller.signal
+	const transform = new AgentRequestTransform(agent)
+
+	const promise = new Promise<AgentActionResult[]>((resolve) => {
```
SomeHats: whenever i see raw promises mixed with async i get uncomfortable because it's really easy to mess up stuff like error handling and end up with promises that never resolve etc. can we do this like this instead?
```tsx
const promise = (async () => {
  const prompt = await agent.preparePrompt(...)
  // ...
})()

File: templates/agent/client/components/highlights/ContextHighlights.tsx [hunk 11]
```diff
@@ -0,0 +1,201 @@
...
+					pagePoint: item.point,
+					generating: true,
+					color: 'var(--tl-color-selected)',
+				}
+			})
+		},
+		[activeContextItems]
+	)
+
+	const allAreaHighlights = useValue(
```
SomeHats: i think this should be ` useMemo`?

templates/agent/client/components/highlights/ContextHighlights.tsx
```diff
@@ -0,0 +1,201 @@
...
+			selectedAreas,
+			selectedShapes,
+			selectedShapesAreas,
+			activeAreas,
+			activeShapes,
+			activeShapeAreas,
+		]
+	)
+
+	const allPointsHighlights = useValue(
```
SomeHats: ditto

templates/agent/shared/AgentRequestTransform.ts
```diff
@@ -0,0 +1,450 @@
...
+ * relative to the current chat origin. The `removeOffsetFromShape` method
+ * reverses it. This is helpful because it helps to keep numbers low, which is
+ * easier for the model to deal with.
+ *
+ * Many transformation methods save some state. For example, the
+ * `ensureShapeIdIsUnique` method changes a shape's ID if it's not unique, and
+ * it saves a record of this change so that further actions can continue to
+ * refer to the shape by its untransformed ID.
+ */
+export class AgentRequestTransform {
```
SomeHats: so i have some suggestions around this transforms api:

1. Update the `AgentActionUtil` base class so it has `editor` and `agent` properties on the instance. This means that most actions don't need their second argument at all, and i think makes more sense with how i understand the lifecycle of one of these utils.
2. Rename from `Transform` to `Helpers`, and de-emphasise it a little in the docs. I feel like right now transforms feel like too much of an important concept, and I think it's mostly a hold-over from the ai module days. at this point "a transform" isn't really a _thing_, we just have this bag of slightly generic helpers. It's still useful but i don't think warrants its current status

templates/agent/shared/actions/AddDetailActionUtil.ts
```diff
@@ -0,0 +1,41 @@
...
+			icon: 'pencil' as const,
+			description,
+		}
+	}
+
+	override applyAction(action: Streaming<IAddDetailAction>, transform: AgentRequestTransform) {
+		if (!action.complete) return
+		const { agent } = transform
+		agent.schedule('Add detail to the canvas.')
+	}
```
SomeHats: re: adding agent + editor to the util, i mean so this function would now look like this:
```tsx
	override applyAction(action: Streaming<IAddDetailAction>) {
		if (!action.complete) return
		this.agent.schedule('Add detail to the canvas.')
	}
```

templates/agent/shared/actions/AgentActionUtil.ts
```diff
@@ -0,0 +1,61 @@
...
+	getInfo(_action: Streaming<T>): Partial<ChatHistoryInfo> | null {
+		return {}
+	}
+
+	/**
+	 * Transform the action before saving it to chat history.
+	 * Useful for sanitizing or correcting actions.
+	 * @returns The transformed action, or null to reject the action
+	 */
+	transformAction(action: Streaming<T>, _transform: AgentRequestTransform): Streaming<T> | null {
```
SomeHats: again re: "transforms"... I would rename this to `sanitizeAction` or similar.

templates/agent/shared/actions/CountShapesActionUtil.ts
```diff
@@ -0,0 +1,47 @@
...
+		transform: AgentRequestTransform
+	) {
+		if (!action.complete) return
+		const { agent, editor } = transform
+
+		// Schedule a follow-up agent request
+		agent.schedule("Here's the number of shapes on the canvas.")
+
+		// Count the shapes
+		return editor.getCurrentPageShapes().length
```
SomeHats: I find this API of both returning and scheduling slightly confusing. I feel like either it should be that I call `.schedule` and have to pass that all the results, or returning results implies scheduling and i have to include my schedule message in the return value. it's one of these:

image

may just be a docs problem, but possibly some API tweaks would make combinations that don't make sense impossible

templates/agent/client/agent/applyAiChange.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Apply an AI change to the canvas.
+ *
+ * @param change - The change to apply
+ * @param editor - The editor to apply the change to
+ *
+ * @public
+ */
+export function applyAiChange({ change, editor }: { change: TLAiChange; editor: Editor }) {
```
SomeHats: this & the `TLAiChange` type are now as far as I can tell only used by the create and update actions. at this point can we just delete this entirely?

templates/agent/shared/format/SimpleFill.ts
```diff
@@ -0,0 +1,30 @@
...
+	tint: 'solid',
+	pattern: 'pattern',
+}
+
+const SHAPE_TO_SIMPLE_FILLS: Record<TLDefaultFillStyle, ISimpleFill> = {
+	none: 'none',
+	fill: 'solid',
+	semi: 'background',
+	solid: 'tint',
+	pattern: 'pattern',
```
SomeHats, line 21: god we should really just rename these for real lol

templates/agent/shared/parts/PeripheralShapesPartUtil.ts
```diff
@@ -0,0 +1,122 @@
...
+		return { ...part, clusters }
+	}
+
+	override buildContent({ clusters }: PeripheralShapesPart): string[] {
+		if (clusters.length === 0) {
+			return []
+		}
+
+		return [
+			"There are some groups of shapes in your peripheral vision, outside the your main view. You can't make out their details or content. If you want to see their content, you need to get closer. The groups are as follows",
```
SomeHats, line 64: this is cool

templates/agent/worker/do/AgentDurableObject.ts
```diff
@@ -0,0 +1,80 @@
+import { DurableObject } from 'cloudflare:workers'
+import { AutoRouter, error } from 'itty-router'
+import { AgentAction } from '../../shared/types/AgentAction'
+import { AgentPrompt } from '../../shared/types/AgentPrompt'
+import { Streaming } from '../../shared/types/Streaming'
+import { Environment } from '../environment'
+import { AgentService } from './AgentService'
+
+export class AgentDurableObject extends DurableObject<Environment> {
```
SomeHats, line 9: I don't think this needs to be a durable object. can we put this in the normal request handler? it'll make it easier for people to adapt to not-cloudflare.

templates/agent/worker/do/closeAndParseJson.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * JSON helper. Given a potentially incomplete JSON string, return the parsed object.
+ * The string might be missing closing braces, brackets, or other characters like quotation marks.
+ * @param string - The string to parse.
+ * @returns The parsed object.
+ */
+export function closeAndParseJson(string: string) {
```
SomeHats, line 7: fun one: give the agent this prompt: `Create a shape with the label 'hi " {'`

it should be fairly straightforward to have this handle escape characters - maybe just a claude job?

templates/agent/README.md
TodePond: > don't need to go into depth but it might be worth having this example show how you pass the article back into the agent

hmm you can return anything you want here, it doesn't need to be a specific type of object - am I understanding you right?

> also: is there ever a situation where you would want to return data without scheduling a follow-up request?

You might want an action to return data to be used in the next request, just in case there *is* a follow-up request.

templates/agent/README.md
TodePond: Let's get rid of this - we replaced its example in the readme

templates/agent/client/agent/dedupeShapesContextItem.ts
TodePond: It's more of a UX and capturing user intent thing. The user might add a bunch of shapes as a single context item, or they might add an individual shape. You might want to say to the model "Put this shape in the center of these shapes" or something.

templates/agent/client/agent/requestAgent.ts
TodePond: yeah I agree with this. I think it's something I would export as "internal" if this was in the SDK, but seeing as that's not available here, let's move it into TldrawAgent.ts

templates/agent/shared/AgentRequestTransform.ts
TodePond: > Rename from Transform to Helpers

Yeah let's do that.

> de-emphasise it a little in the docs

Done.

> Update the AgentActionUtil base class so it has editor and agent properties on the instance

We did that at a certain point, but we want the worker to be able to use the utils too without the agent or editor, so we pulled them into an argument, but agree that it would be better to be able to get em from the instance. Will take another look at what we can do here.

templates/agent/shared/actions/AgentActionUtil.ts
TodePond: Okay I read your mind and did this without reading this comment

templates/agent/shared/actions/CountShapesActionUtil.ts
TodePond: We think youre right - we're covering some more niche use cases with this current system that might be worth sacrificing for a clearer API.

templates/agent/client/agent/applyAiChange.ts
TodePond: I'll remove the dependency. Won't remove it from the SDK in this PR though.

templates/agent/shared/format/SimpleFill.ts
TodePond: 4.0 thing!?

templates/agent/shared/parts/PeripheralShapesPartUtil.ts
TodePond: Nice one @max-drake

templates/agent/worker/do/AgentDurableObject.ts
TodePond: I was mimicking what we have in the AI template - do you think we should remove that there too?

TodePond:
> overall this is looking great! my main thought is just that like... it's a lot???? and i wonder where it sits on the starter kit <> product demo <> module spectrum thingy.

Thanks for the review alex, we're going through your points and making changes now. This is something I've been wodnering too, and I think it does fit best as a starter kit still as I feel it's too much of a tech demo to be a product demo, and I wouldn't want to make this part of an SDK at this point. I think it's okay for the starter to be bigger than the others, but there are possibly things we could pull out if we wanted to reduce its size

templates/agent/worker/do/AgentDurableObject.ts
SomeHats: i don't think that needs to be a durable object either, but if you'd rather call that out of scope im fine with keeping it as-is for now

templates/agent/README.md
TodePond: done

templates/agent/README.md
TodePond: done

templates/agent/README.md
TodePond: done

templates/agent/client/agent/TldrawAgent.ts
TodePond: done

templates/agent/client/agent/TldrawAgent.ts
TodePond: done

templates/agent/client/agent/requestAgent.ts
TodePond: done

templates/agent/shared/AgentRequestTransform.ts
TodePond: We added them!

PR #6679, opened 2025-09-01
Export `doesGeometryOverlapPolygon`

TodePond:
This PR exports `doesGeometryOverlapPolygon` publicly. The agent starter makes use of it.

In the end, this is the only change to the SDK needed by the agent starter as we removed the dependency on the AI module. We'll still need to fix canvas pointer events when multiple editors are open on the same page, but that's an SDK fix, not a feature.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- Added the `overlapsPolygon` method to the `Geometry2D` class.

### API Changes

- Added the `overlapsPolygon` method to the `Geometry2D` class.

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1895,6 +1895,8 @@ export class Group2d extends Geometry2d {
     // (undocumented)
     nearestPoint(point: VecLike, filters?: Geometry2dFilters): Vec;
     // (undocumented)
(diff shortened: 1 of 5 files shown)


TodePond:
> happy for us to land as-is, but I would rather we do this as a method on the `Geometry2d` class instead of as a free method

lets do that

TodePond:
okay moved it to Geometry2D

PR #6681, opened 2025-09-01
Bump AI template's packages

TodePond:
This PR bumps zod and openai within the AI template so that it's on the same version as the agent starter. It also removes the unused openai package from the docs site.

It also fixes a bug that was preventing the AI template from working by adding a transform that turns rich text into strings.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


SomeHats (approved):
lgtm!

PR #6682, opened 2025-09-02
improvement(ui): improve RTL support for Arabic and other right-to-left languages

blurname:
I found that the tldraw SDK lacks the support of rtl, which is not convenient for people using Arabic. So I've done some attempts to improve it.

Mainly changed

1. UI primitives accept `dir`
2. ContextMenu positioning on cursor's left
3. place `TldrawUiTranslationProvider` above `TldrawUiTooltipProvider`

Now UI in Arabic

image
image
image
image

### Change type

- [x] `improvement`

### Test plan

1. Switch language to Arabic or another RTL language.
2. Verify UI primitives (menus, dialogs, tooltips) respect the direction.
3. Verify ContextMenu positioning.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improved RTL support for Arabic and other right-to-left languages.


blurname:

packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx
```diff
@@ -101,9 +102,11 @@ export const DefaultContextMenu = memo(function DefaultContextMenu({
 	// so skip rendering the menu.
 	const content = children ?? <DefaultContextMenuContent />
 
+	const dir = useDir()
+
 	return (
-		<_ContextMenu.Root dir="ltr" onOpenChange={handleOpenChange} modal={false}>
-			<_ContextMenu.Trigger onContextMenu={undefined} dir="ltr" disabled={disabled}>
+		<_ContextMenu.Root dir={dir} onOpenChange={handleOpenChange} modal={false}>
+			<_ContextMenu.Trigger onContextMenu={undefined} dir={'ltr'} disabled={disabled}>
```
mimecuvalo, line 109: this should be `{dir}`, right?

packages/tldraw/src/lib/ui/components/primitives/TldrawUiPopover.tsx
```diff
@@ -53,13 +55,15 @@ export interface TLUiPopoverContentProps {
 export function TldrawUiPopoverContent({
 	side,
 	children,
-	align = 'center',
+	align,
```
mimecuvalo, line 58: let's just keep this as is, as it's clearer: it'll be `center` as the default which is what you still have anyway

packages/tldraw/src/lib/ui/TldrawUi.tsx
```diff
@@ -170,6 +170,7 @@ const TldrawUiContent = React.memo(function TldrawUI() {
 
 	return (
 		<div
+			data-dir={useDir()}
```
mimecuvalo, line 173: let's not put a hook as prop like this, feels icky :P

packages/tldraw/src/lib/ui/context/TldrawUiContextProvider.tsx
```diff
@@ -73,12 +73,12 @@ export const TldrawUiContextProvider = track(function TldrawUiContextProvider({
 	const editor = useMaybeEditor()
 	return (
 		<MimeTypeContext.Provider value={mediaMimeTypes}>
-			<TldrawUiTooltipProvider>
-				<AssetUrlsProvider assetUrls={useDefaultUiAssetUrlsWithOverrides(assetUrls)}>
-					<TldrawUiTranslationProvider
-						overrides={useMergedTranslationOverrides(overrides)}
-						locale={editor?.user.getLocale() ?? defaultUserPreferences.locale}
-					>
+			<TldrawUiTranslationProvider
```
mimecuvalo, line 76: 👍 ty!

packages/tldraw/src/lib/ui.css
```diff
@@ -525,31 +525,31 @@
...
-	margin: 0px;
-	padding: 2px;
-	border-radius: 2px;
+	align-self: bottom;
+	min-width: 1em;
+	padding-left: var(--tl-space-2);
+}
+
+.tlui-kbd > span:nth-last-of-type(n + 2) {
+	width: 15px;
```
mimecuvalo, line 542: hm, why 15px? that seems like this would be brittle, depending on what's in the `<span>`

packages/tldraw/src/lib/ui/components/MainMenu/DefaultMainMenu.tsx
mimecuvalo, line 39: take a look at https://www.radix-ui.com/primitives/docs/components/dropdown-menu#content
instead of modifying `tlui-menu` down below with '!important' positioning rules, you probably should be changing these values where needed. unfortunately it's a bit scattered throughout the codebase!

packages/tldraw/src/lib/ui.css
```diff
@@ -2121,3 +2171,24 @@ tldraw? probably.
...
+[dir='rtl'] .tlui-page-menu__name {
+	text-align: right;
+}
+
+[dir='rtl'] .tlui-page-menu__item__button > span {
+	text-align: right;
+}
+
+/* RTL Context Menu Horizontal Positioning */
+.tlui-menu[data-dir='rtl'] {
```
mimecuvalo, line 2189: take a look at my comment in `DefaultMainMenu`, you'll probably want to change the Radix props, not override it like this

packages/tldraw/src/lib/ui/TldrawUi.tsx
mimecuvalo: also, is this used at all?

packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx
```diff
@@ -115,6 +118,7 @@ export const DefaultContextMenu = memo(function DefaultContextMenu({
 						alignOffset={-4}
 						collisionPadding={4}
 						onContextMenu={preventDefault}
+						data-dir={dir}
```
mimecuvalo, line 121: is this unused?

packages/tldraw/src/lib/ui.css
```diff
@@ -525,31 +525,31 @@
...
+	padding-left: var(--tl-space-2);
+}
+
+.tlui-kbd > span:nth-last-of-type(n + 2) {
+	width: 15px;
 }
 
-.tlui-kbd > span:last-child {
-	padding-right: 0;
+div[dir='rtl'] .tlui-kbd {
```
mimecuvalo, line 545: i'll nitpick on this one a bit. i think the general selector of `[dir='rtl']` might be a bit too generic.
instead, it would be interesting to put `dir="rtl"` on the `.tl-container` class (you'd have to pass it from `Tldraw.tsx` → `TldrawEditor.tsx` as a prop I believe).

in this way, you could have a rule like:
`.tl-container[dir='rtl'] .tlui-kbd`
which makes it clearer that the whole UI is in RTL mode.
and if we made that change, it makes me wonder whether we should possibly remove the `dir` attributes in all the various places potentially. it's unclear to me whether you would want to have an app that has RTL at the top-level but individual UI components can be potentially LTR? probably not 🤔 

i _would_ say to put it on the `.tlui-layout` div but we do unfortunately have UI that isn't completely in that div, e.g. the contextual toolbars.

mimecuvalo (commented):
sorry for the delay in reviewing! when we saw this PR come up, the team got excited to see it :D (we were just very busy with our 4.0 launch)

got a couple comments / questions before moving forward, but keen to get this in!

steveruizok:
Sorry @blurname, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #6683, opened 2025-09-02
Fix panning perf.

MitjaBezensek:
Every tooltip component tracked camera changes, causing hundreds of re-renders during panning. Moved camera tracking to `TooltipSingleton` using `getCameraState()` API.

This is what caused the perf tests to start failing. Probably worth hotfixing.
 
### Change type

- [x] `bugfix`


steveruizok:
Great fix!

PR #6699, opened 2025-09-05
Dynamically sized shape example

SomeHats:
An example of a shape that pulls its size from the DOM. This also adds a `canCull()` flag to shape utils, and these shapes need their dom layout at all times, not just when on screen.

### Change type

- [x] `api`

### API Changes

- Add `ShapeUtil.canCull`, which can be used to disable culling on certain shapes.

### DIFF
--- a/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
+++ b/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
@@ -284,6 +284,17 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -283,6 +283,17 @@ export abstract class ShapeUtil<Shape extends TLUnknownShape = TLUnknownShape> {
 		return true
 	}
 
+	/**
+	 * Whether this shape can be culled when rendering. By default, shapes are culled for
```
ds300: this condition is confusing I think?

```suggestion
	 * Whether this shape can be culled. By default, shapes are culled for
```

PR #6700, opened 2025-09-05
chore: update license year to 2025

SomeHats:
Update the license year from 2024 to 2025 across all licenses and READMEs.

### Change type

- [x] `other`

### Test plan

1. Verify license files show 2025.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Updated license year to 2025.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ You can use the tldraw SDK in commercial or non-commercial projects so long as y
 
 ## Trademarks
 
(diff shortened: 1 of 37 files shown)


steveruizok:
We would want to keep 2024 in 2024-present. We might actually want to keep 2022, which is the year that the company started.

mimecuvalo:
> We would want to keep 2024 in 2024-present. We might actually want to keep 2022, which is the year that the company started.

yeah Steve is right, this should be where it started, not the currenty year

PR #6702, opened 2025-09-05
prioritised topological sort for migrations

SomeHats:
Replace basic topological sort with priority-based Kahn's algorithm that minimizes the total distance between migrations and their explicit dependencies. The new algorithm respects both implicit sequence dependencies (`foo/1` → `foo/2`) and explicit dependsOn relationships while scheduling dependent migrations as close as possible to their dependencies.

This is to make `dependsOn` migrations more predictable. If `foo/1` and `foo/2` both make changes to `foo` records, and `store/1` depends on `foo/1`, it's better that `store/1` runs before `foo/2` in case `foo/2` makes changes that affect `store/1`.

Key changes:
- Use Kahn's algorithm with urgency scoring to prioritize migrations that unblock explicit dependencies
- Add comprehensive test coverage for distance minimization scenarios
- Maintain all existing behavior including cycle detection
- Add detailed documentation explaining the algorithm and its benefits

🤖 Generated with [Claude Code](https://claude.ai/code)

### Change type

- [x] `improvement`

### Release notes

- Migrations with explicit `dependsOn` constraints will now be scheduled as close as possible to the things they depend on

### DIFF
--- a/packages/store/src/lib/migrate.ts
+++ b/packages/store/src/lib/migrate.ts
@@ -160,38 +160,127 @@ export interface MigrationSequence {
 	sequence: Migration[]
 }
 
+/**
+ * Sorts migrations using a distance-minimizing topological sort.
+ *
+ * This function respects two types of dependencies:
+ * 1. Implicit sequence dependencies (foo/1 must come before foo/2)
+ * 2. Explicit dependencies via `dependsOn` property
+ *
+ * The algorithm minimizes the total distance between migrations and their explicit
+ * dependencies in the final ordering, while maintaining topological correctness.
+ * This means when migration A depends on migration B, A will be scheduled as close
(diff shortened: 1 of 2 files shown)


packages/store/src/lib/migrate.ts
```diff
@@ -160,38 +160,127 @@ export interface MigrationSequence {
...
+					urgencyScore += 100 // High priority for explicit dependencies
+				}
+			}
+
+			// Priority 2: Count all unprocessed dependents (to break ties)
+			for (const depId of dependents.get(m.id) || []) {
+				if (!processed.has(depId)) {
+					urgencyScore += 1
+				}
+			}
```
ds300: seems like these two loops can be trivially combined into one?

ds300 (approved):
thank you!!

PR #6718, opened 2025-09-10
Add timeline scrubber example

SomeHats:
It's basic but demonstrates the principle.

### Change type

- [x] `other`


mimecuvalo:
this is sick! i love it so much. i'd love to expose this somewhere (debug menu?) in general.

future wishlist would be adding more coarse "stops" along the way for bigger changes (creations/deletions).

ship ship ship!

PR #6722, opened 2025-09-11
docs: automate documentation and add jsdoc comments

steveruizok:
This PR adds Claude slash commands for DOCS.md and jsdoc comments for directories. The bigger idea here is a next-best-thing documentation system that we can update automatically and review.

### Change type

- [x] `improvement`

### Test plan

1. Verify documentation generation scripts

- [x] Unit tests (if present)

### Release notes

- Added jsdoc comments to all exported functions across the state, store, sync-core, sync, tlschema, utils, and validate packages


steveruizok:
@claude remove the jsdoc @description notation from all files modified in this branch. We don't use that identifier, we just write the description without it.

PR #6726, opened 2025-09-11
feat(license): add new NATIVE_LICENSE type

mimecuvalo:
We have a need for some customers that embed tldraw inside of native apps, whether it's Electron/Tauri/wkwebview. 
This opens up the ability to denote that kind of usage in our licensing scheme.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- license: add new NATIVE_LICENSE type

### API changes

- license: adds `isNativeLicense type internally

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.ts
+++ b/packages/editor/src/lib/license/LicenseManager.ts
@@ -6,11 +6,21 @@ import { importPublicKey, str2ab } from '../utils/licensing'
 const GRACE_PERIOD_DAYS = 30
 
 export const FLAGS = {
+	// -- MUTUALLY EXCLUSIVE FLAGS --
+	// Annual means the license expires after a time period, usually 1 year.
 	ANNUAL_LICENSE: 1,
+	// Perpetual means the license never expires up to the max supported version.
 	PERPETUAL_LICENSE: 1 << 1,
+
+	// -- ADDITIVE FLAGS --
+	// Internal means the license is for internal use only.
 	INTERNAL_LICENSE: 1 << 2,
+	// Watermark means the product is watermarked.
 	WITH_WATERMARK: 1 << 3,
+	// Evaluation means the license is for evaluation purposes only.
 	EVALUATION_LICENSE: 1 << 4,
+	// Native means the license is for native apps which switches
+	// on special-case logic.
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -11,6 +11,7 @@ export const FLAGS = {
 	INTERNAL_LICENSE: 1 << 2,
 	WITH_WATERMARK: 1 << 3,
 	EVALUATION_LICENSE: 1 << 4,
+	NATIVE_LICENSE: 1 << 5,
```
MitjaBezensek, line 24: We should maybe document all the different flags and what they do, which ones can be combined, and which ones are mutually exclusive.

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -323,10 +326,20 @@ export class LicenseManager {
 				}
 			}
 
+			// Native license support
+			// In this case, `normalizedHost` is actually a protocol, e.g. `app-bundle:`
+			if (this.isNativeLicense(licenseInfo) && window.location.protocol === normalizedHost) {
```
MitjaBezensek: Feels a bit strange to read this. Maybe `normalizedHost` name should be changed?

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -323,10 +337,20 @@ export class LicenseManager {
 				}
 			}
 
+			// Native license support
+			// In this case, `normalizedHost` is actually a protocol, e.g. `app-bundle:`
+			if (this.isNativeLicense(licenseInfo) && window.location.protocol === maybeProtocol) {
```
SomeHats, line 342: i don't think we should use a protocol here, but instead an overall prefix on the URL. I know we handle vscode specifically, but if we encountered another similar platform where the format is `some-protocol://some-unique-id/index.html`, then we wouldn't be able to use this solution

mimecuvalo: followup here https://github.com/tldraw/tldraw/pull/6732

PR #6731, opened 2025-09-12
fix(editor): ignore events originating within InFrontOfCanvas

SomeHats:
Fixes an issue where things in `InFrontOfTheCanvas` behave erratically because their events are also being handled by the canvas event handler. Previously, `InFrontOfTheCanvas` was outside of the `tl-canvas` wrapper, but this caused issues where its contents would become mis-aligned from the main canvas when that canvas didn't fully fill `tl-container`. We moved it inside in #6626, which introduced this issue.

There are a few approaches that we can take here:

**1. stopPropagation on all events from `InFrontOfTheCanvas`**

This would stop tl-canvas handling any other events, but that would cause problems for anyone who might have event handlers higher up the dom tree, for example on `body` providing `onClickOutside` type functionality.

**2. stopPropagation on specific events in the `InFrontOfTheCanvas` implementation**

This would be a breaking change, but it's 4.0 so that's OK. People would need to deal with similar quirks as they do with shapes, adding things like `onPointerDown={stopEventPropagation}` to specific elements. This has the same `stopPropagation` drawback that option 1 has, but we also encourage that pattern all over the place (maybe we shouldn't?). This also has some quirks where the events we have by default aren't applied super consistently - in the case of the image toolbar, it works fine on desktop, but not on mobile. It's tricky for consumers to know which events they need to `preventDefault` on and when.

**3. Ignore all events from `InFrontOfTheCanvas`**

This is the approach this diff currently takes. Instead of using `stopPropagation` at all, we specifically ignore problematic events that are already handled by `InFrontOfTheCanvas`. They still get propagated as normal otherwise. Implementation-wise this is a bit more complex, but I think it provides the best overall experience??

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where events originating in InFrontOfCanvas were incorrectly handled by the canvas.

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -157,10 +157,6 @@ export function useCanvasEvents() {
 				inFrontOfTheCanvas: {
 					onPointerDown: markEventAsHandledByInFrontOfTheCanvas,
 					onPointerUp: markEventAsHandledByInFrontOfTheCanvas,
-					onPointerEnter: markEventAsHandledByInFrontOfTheCanvas,
-					onPointerLeave: markEventAsHandledByInFrontOfTheCanvas,
-					onDragOver: markEventAsHandledByInFrontOfTheCanvas,
-					onDrop: markEventAsHandledByInFrontOfTheCanvas,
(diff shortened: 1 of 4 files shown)


SomeHats:
Closing in favor of #6733

PR #6732, opened 2025-09-12
license: have NATIVE_LICENSE use a regex

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/6726

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- license: have NATIVE_LICENSE use a regex

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.ts
+++ b/packages/editor/src/lib/license/LicenseManager.ts
@@ -323,11 +323,8 @@ export class LicenseManager {
 
 			// Native license support
 			// In this case, `normalizedHost` is actually a protocol, e.g. `app-bundle:`
-			if (
-				this.isNativeLicense(licenseInfo) &&
-				new RegExp(normalizedHostOrUrlRegex).test(window.location.href)
-			) {
-				return true
+			if (this.isNativeLicense(licenseInfo)) {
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -322,6 +321,15 @@ export class LicenseManager {
...
+			// In this case, `normalizedHost` is actually a protocol, e.g. `app-bundle:`
+			if (
+				this.isNativeLicense(licenseInfo) &&
+				new RegExp(normalizedHostOrUrlRegex).test(window.location.href)
+			) {
+				return true
+			}
+
 			// Glob testing, we only support '*.somedomain.com' right now.
 			if (host.includes('*')) {
```
SomeHats, line 331: we should maybe only do this if `!this.isNativeLicense`?

SomeHats: or make the above be something like:
```tsx
if (this.isNativeLicense(licenseInfo)) {
    return new RegExp(normalizedHostOrUrlRegex).test(window.location.href)
}
```

PR #6733, opened 2025-09-12
replace stopEventPropagation with markEventAsHandled and kill `e.isKilled`

SomeHats:
An alternative and expanded take on #6731. It's probably worth reading the description of that diff too.

The canvas element adds a bunch of event handlers for controlling input to the canvas as a whole. Then, we have things on that canvas that we want to handle differently. We currently have two major patterns for how we handle those things:
1. we write `;(e as any).isKilled = true`. This is mostly in SDK internals. By marking this special extra property on the event, we stop it from being picked up elsewhere when we want to use it.
2. we use `stopEventPropagation(e)`, which is the same as writing `e.stopPropagation()`. This is less "weird looking" so we often use it more, and it stops the event from bubbling up to the canvas, but also from bubbling up beyond the canvas as a whole.

In general, `stopPropagation` is sometimes seen as a bit of an anti-pattern because it's a pretty course hammer - once something has called `.stopPropagation()`, that event is _gone_. Other things above the tree - even things that aren't tldraw - can't see that event any more. In theory, this can cause issues. For example, an 'on click outside' type handler for a modal might not get picked up if the click gets swallowed by tldraw.

This diff replaces `e.isKilled` with two new utilities: `markEventAsHandled` and `wasEventAlreadyHandled`. These do essentially the same thing as `e.isKilled`, but use a `WeakSet` instead of a property for type safety and to avoid any potential collisions. It also replaces most of the existing `.stopPropagation()` use with this new utility, so events still bubble up above tldraw, while continuing to be ignored by tldraw itself.

I've gone through and tested all the places we were using `stopPropagation` that we're now using `wasEventAlreadyHandled`.

### Change type

- [x] `api`

### API Changes

* Deprecate `stopEventPropagation`. Most places that used it should be replaced with `markEventAsHandled` which causes tldraw to ignore an event, but allows it to propagate out above tldraw. If you definitely still want stop propagation semantics, you can call `event.stopPropagation()` instead.
* **MAYBE BREAKING**: tldraw no longer swallows most events that happen within it. pointer and click events happening within tldraw will now bubble up to any listeners you have above tldraw in the dom.

### DIFF
--- a/packages/editor/src/lib/TldrawEditor.tsx
+++ b/packages/editor/src/lib/TldrawEditor.tsx
@@ -44,7 +44,7 @@ import { LicenseProvider } from './license/LicenseProvider'
 import { Watermark } from './license/Watermark'
 import { TldrawOptions } from './options'
 import { TLDeepLinkOptions } from './utils/deepLinks'
-import { stopEventPropagation } from './utils/dom'
+import { markEventAsHandled } from './utils/dom'
 import { TLTextOptions } from './utils/richText'
 import { TLStoreWithStatus } from './utils/sync/StoreWithStatus'
 
@@ -275,7 +275,7 @@ export const TldrawEditor = memo(function TldrawEditor({
 			data-tldraw={version}
 			draggable={false}
 			className={classNames(`${TL_CONTAINER_CLASS} tl-theme__light`, className)}
-			onPointerDown={stopEventPropagation}
+			onPointerDown={markEventAsHandled}
 			tabIndex={-1}
 			role="application"
 			aria-label={_options?.branding ?? 'tldraw'}
--- a/packages/editor/src/lib/hooks/useSelectionEvents.ts
+++ b/packages/editor/src/lib/hooks/useSelectionEvents.ts
@@ -3,9 +3,10 @@ import { RIGHT_MOUSE_BUTTON } from '../constants'
 import { TLSelectionHandle } from '../editor/types/selection-types'
 import {
 	loopToHtmlElement,
+	markEventAsHandled,
 	releasePointerCapture,
 	setPointerCapture,
-	stopEventPropagation,
+	wasEventAlreadyHandled,
 } from '../utils/dom'
 import { getPointerInfo } from '../utils/getPointerInfo'
 import { useEditor } from './useEditor'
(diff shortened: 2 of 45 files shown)


packages/editor/src/lib/hooks/useSelectionEvents.ts
```diff
@@ -54,14 +55,14 @@ export function useSelectionEvents(handle: TLSelectionHandle) {
 					handle,
 					...getPointerInfo(e),
 				})
-				stopEventPropagation(e)
+				markEventAsHandled(e)
```
mimecuvalo, line 58: remind me here why this is setup like this:
- onPointerDown did stopPropagation
- onPointerMove/onPointerUp do not though

questions are:
1. do onPointerMove/onPointerUp need to also have `markEventAsHandled`?
2. does switching the explicit stopEventPropagation for `onPointerDown` have a knock-on effect?

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -275,7 +275,7 @@ export const TldrawEditor = memo(function TldrawEditor({
 			data-tldraw={version}
 			draggable={false}
 			className={classNames(`${TL_CONTAINER_CLASS} tl-theme__light`, className)}
-			onPointerDown={stopEventPropagation}
+			onPointerDown={markEventAsHandled}
```
mimecuvalo, line 278: this one makes me a little more worried than other ones, this won't have knock-on effects maybe?

packages/editor/src/lib/hooks/useSelectionEvents.ts
SomeHats: 1. Nope! `onPointerMove` we always want so we keep the pointer position up to date. `onPointerUp` will only have an effect during an editor interaction, and in that case we generally want it handled regardless of what you're pointer-up-ing over. Although generally those interactions will use pointer locking anyway so a non-issue.
2. nope - i tested all of these and couldn't find any knock-on effects

packages/editor/src/lib/TldrawEditor.tsx
SomeHats: not on tldraw. it may have knock-on effects on tldraw consumers as they'll now be able to see these events whereas before they were swallowed completely, but that's part of the reason for this diff i think?

packages/editor/src/lib/hooks/useSelectionEvents.ts
mimecuvalo: ok, i think it's maybe putting that somewhere as a comment, for clarity in this file. thx!

SomeHats:
@claude can you create a follow-up PR removing the deprecated `stopEventPropagation` API?

PR #6739, opened 2025-09-14
Augmentation-based shape types

Andarist:
### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

WIP

### Release notes

WIP

### DIFF
--- a/apps/examples/src/examples/after-create-update-shape/AfterCreateUpdateShapeExample.tsx
+++ b/apps/examples/src/examples/after-create-update-shape/AfterCreateUpdateShapeExample.tsx
@@ -14,6 +14,10 @@ function ensureOnlyOneRedShape(editor: Editor, shapeId: TLShapeId) {
 	const otherRedShapesOnPage = Array.from(editor.getPageShapeIds(pageId))
 		.map((id) => editor.getShape(id)!)
 		.filter((otherShape) => otherShape.id !== shape.id && isRedShape(otherShape))
+		.filter(
+			(shape): shape is Extract<TLShape, { props: { color: string } }> =>
+				'color' in shape.props && typeof shape.props.color === 'string'
+		)
 
 	// set the color of all those shapes to black:
 	editor.updateShapes(
--- a/apps/examples/src/examples/after-delete-shape/AfterDeleteShapeExample.tsx
+++ b/apps/examples/src/examples/after-delete-shape/AfterDeleteShapeExample.tsx
@@ -51,7 +51,7 @@ function createDemoShapes(editor: Editor) {
 		},
 		...[50, 180, 310].map((x) => ({
 			id: createShapeId(),
-			type: 'geo',
+			type: 'geo' as const,
 			parentId: frameId,
 			x,
 			y: 120,
--- a/apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
+++ b/apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
@@ -6,6 +6,12 @@ import 'ag-grid-community/styles/ag-grid.css'
 import 'ag-grid-community/styles/ag-theme-quartz.css'
 import 'tldraw/tldraw.css'
 
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		'ag-grid': AgGridShape
+	}
+}
+
 type AgGridShape = TLBaseShape<
 	'ag-grid',
 	{ w: number; h: number; rowData: any[]; columnDefs: any[] }
--- a/apps/examples/src/examples/drag-and-drop/DragAndDropExample.tsx
+++ b/apps/examples/src/examples/drag-and-drop/DragAndDropExample.tsx
@@ -11,6 +11,13 @@ import {
 } from 'tldraw'
 import 'tldraw/tldraw.css'
 
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		'my-grid-shape': MyGridShape
+		'my-counter-shape': MyCounterShape
+	}
+}
+
 // [1]
 type MyGridShape = TLBaseShape<'my-grid-shape', Record<string, never>>
 type MyCounterShape = TLBaseShape<'my-counter-shape', Record<string, never>>
@@ -139,21 +146,21 @@ export default function DragAndDropExample() {
 
 /*
 [1]
-Define custom shape types using TLBaseShape. Each shape type needs a unique identifier and can have custom 
-properties. Here we use Record<string, never> since our shapes don't need any custom properties. These are 
+Define custom shape types using TLBaseShape. Each shape type needs a unique identifier and can have custom
+properties. Here we use Record<string, never> since our shapes don't need any custom properties. These are
 very basic custom shapes: see the custom shape examples for more complex examples.
 
 [2]
-Create a ShapeUtil for the counter shape. This defines how the shape behaves and renders. We disable resizing 
+Create a ShapeUtil for the counter shape. This defines how the shape behaves and renders. We disable resizing
 and use Circle2d geometry for collision detection. The component renders as a red circle using HTMLContainer.
 
 [3]
-Create a ShapeUtil for the grid shape. This creates a rectangular grid that can accept dropped shapes. We use 
+Create a ShapeUtil for the grid shape. This creates a rectangular grid that can accept dropped shapes. We use
 Rectangle2d geometry and render it with CSS grid lines using background gradients.
 
 [5]
 Override onDragShapesIn to handle when shapes are dragged into the grid. We filter for counter shapes that
-aren't already children of this grid, then reparent them to become children. This makes them move with the grid. 
+aren't already children of this grid, then reparent them to become children. This makes them move with the grid.
 
 [6]
 Override onDragShapesOut to handle when shapes are dragged out of the grid. If they're not being dragged to
--- a/apps/examples/src/examples/editable-shape/EditableShapeUtil.tsx
+++ b/apps/examples/src/examples/editable-shape/EditableShapeUtil.tsx
@@ -11,7 +11,7 @@ import {
 
 const ANIMAL_EMOJIS = ['🐶', '🐱', '🐨', '🐮', '🐴']
 
-type IMyEditableShape = TLBaseShape<
+export type IMyEditableShape = TLBaseShape<
 	'my-editable-shape',
 	{
 		w: number
@@ -96,28 +96,28 @@ export class EditableShapeUtil extends BaseBoxShapeUtil<IMyEditableShape> {
 	}
 }
 
-/* 
+/*
 
-This is our shape util, which defines how our shape renders and behaves. For 
+This is our shape util, which defines how our shape renders and behaves. For
 more information on the shape util, check out the custom shape example.
 
 [1]
 We override the canEdit method to allow the shape to enter the editing state.
 
 [2]
-We want to conditionally render the component based on whether it is being 
+We want to conditionally render the component based on whether it is being
 edited or not.
 
 	[a] We can check whether our shape is being edited by comparing the
 		editing shape id to the shape's id.
-	
+
 	[b] We want to allow pointer events when the shape is being edited,
 		and stop event propagation on pointer down. Check out the interactive
 		shape example for more information on this.
 
 	[c] We render a button to change the animal emoji when the shape is being
-		edited. 
-		
+		edited.
+
 	[e]	We also render a message when the shape is not being edited.
 
 [3]
--- a/apps/examples/src/examples/external-content-sources/ExternalContentSourcesExample.tsx
+++ b/apps/examples/src/examples/external-content-sources/ExternalContentSourcesExample.tsx
@@ -9,11 +9,17 @@ import {
 } from 'tldraw'
 import 'tldraw/tldraw.css'
 
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		'dangerous-html': IDangerousHtmlShape
+	}
+}
+
 // There's a guide at the bottom of this page!
 
 // [1]
 export type IDangerousHtmlShape = TLBaseShape<
-	'html',
+	'dangerous-html',
 	{
 		w: number
 		h: number
@@ -23,11 +29,11 @@ export type IDangerousHtmlShape = TLBaseShape<
 
 // [2]
 class DangerousHtmlExample extends BaseBoxShapeUtil<IDangerousHtmlShape> {
-	static override type = 'html' as const
+	static override type = 'dangerous-html' as const
 
 	override getDefaultProps() {
 		return {
-			type: 'html',
+			type: 'dangerous-html',
 			w: 500,
 			h: 300,
 			html: '<div>hello</div>',
@@ -61,7 +67,7 @@ export default function ExternalContentSourcesExample() {
 				const center = content.point ?? editor.getViewportPageBounds().center
 
 				editor.createShape({
-					type: 'html',
+					type: 'dangerous-html',
 					x: center.x - 250,
 					y: center.y - 150,
 					props: {
@@ -82,7 +88,7 @@ export default function ExternalContentSourcesExample() {
 	)
 }
 
-/* 
+/*
 Introduction:
 This example shows how to handle content pasted from external sources, this could be
 embeds, files, svgs, text, images, or urls. In this case we will handle text/html content.
@@ -96,11 +102,11 @@ This is our shape util. It's a class that extends BaseBoxShapeUtil. For a more d
 example of how to create a custom shape, see the custom config example.
 
 [3]
-We use the onMount prop to get access to the editor instance via 
-the handleMount callback (check out the API example for a more detailed look at this). Then we 
-call the registerExternalContentHandler method, we could choose to handle embeds, files, svgs, 
+We use the onMount prop to get access to the editor instance via
+the handleMount callback (check out the API example for a more detailed look at this). Then we
+call the registerExternalContentHandler method, we could choose to handle embeds, files, svgs,
 text, images, or urls. For this example we will handle text/html content. The handler is called
-with the point where the user pasted the content and an array of sources. We will find and 
+with the point where the user pasted the content and an array of sources. We will find and
 return the html source, then create a new shape with that html content.
 
 */
--- a/apps/examples/src/examples/pin-bindings/PinExample.tsx
+++ b/apps/examples/src/examples/pin-bindings/PinExample.tsx
@@ -14,7 +14,9 @@ import {
 	TLBaseShape,
 	TLEditorComponents,
 	TLPointerEventInfo,
+	TLShape,
 	TLShapeId,
+	TLShapePartial,
 	TLShapeUtilCanBindOpts,
 	TLUiComponents,
 	TLUiOverrides,
@@ -29,6 +31,12 @@ import {
 	useTools,
 } from 'tldraw'
 import 'tldraw/tldraw.css'
+
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		pin: PinShape
+	}
+}
 
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
 type PinShape = TLBaseShape<'pin', {}>
@@ -240,15 +248,15 @@ class PinBindingUtil extends BindingUtil<PinBinding> {
 			}
 		}
 
-		const updates = []
+		const updates: TLShapePartial<TLShape>[] = []
 		for (const [shapeId, position] of currentPositions) {
 			const delta = Vec.Sub(position, initialPositions.get(shapeId)!)
 			if (delta.len2() <= 0.01) continue
 
 			const newPosition = this.editor.getPointInParentSpace(shapeId, position)
 			updates.push({
+				...this.editor.getShape(shapeId)!,
 				id: shapeId,
-				type: this.editor.getShape(shapeId)!.type,
 				x: newPosition.x,
 				y: newPosition.y,
 			})
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -55,7 +55,6 @@ import {
 	TLStore,
 	TLStoreSnapshot,
 	TLUnknownBinding,
-	TLUnknownShape,
 	TLVideoAsset,
 	createBindingId,
 	createShapeId,
@@ -974,7 +973,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	shapeUtils: { readonly [K in string]?: ShapeUtil<TLUnknownShape> }
+	shapeUtils: { readonly [K in string]?: ShapeUtil<TLShape> }
 
 	styleProps: { [key: string]: Map<StyleProp<any>, string> }
 
@@ -993,8 +992,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	getShapeUtil<S extends TLUnknownShape>(shape: S | TLShapePartial<S>): ShapeUtil<S>
-	getShapeUtil<S extends TLUnknownShape>(type: S['type']): ShapeUtil<S>
+	getShapeUtil<S extends TLShape>(shape: S | TLShapePartial<S>): ShapeUtil<S>
+	getShapeUtil<S extends TLShape>(type: S['type']): ShapeUtil<S>
 	getShapeUtil<T extends ShapeUtil>(type: T extends ShapeUtil<infer R> ? R['type'] : string): T
 	getShapeUtil(arg: string | { type: string }) {
 		const type = typeof arg === 'string' ? arg : arg.type
@@ -1008,8 +1007,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @param shape - A shape, shape partial, or shape type.
 	 */
-	hasShapeUtil<S extends TLUnknownShape>(shape: S | TLShapePartial<S>): boolean
-	hasShapeUtil<S extends TLUnknownShape>(type: S['type']): boolean
+	hasShapeUtil(shape: TLShape | TLShapePartial<TLShape>): boolean
+	hasShapeUtil(type: TLShape['type']): boolean
 	hasShapeUtil<T extends ShapeUtil>(
 		type: T extends ShapeUtil<infer R> ? R['type'] : string
 	): boolean
@@ -5501,15 +5500,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	isShapeOfType<T extends TLUnknownShape>(shape: TLUnknownShape, type: T['type']): shape is T
-	isShapeOfType<T extends TLUnknownShape>(
-		shapeId: TLUnknownShape['id'],
-		type: T['type']
-	): shapeId is T['id']
-	isShapeOfType<T extends TLUnknownShape>(
-		arg: TLUnknownShape | TLUnknownShape['id'],
-		type: T['type']
-	) {
+	isShapeOfType<T extends TLShape>(shape: TLShape, type: T['type']): shape is T
+	isShapeOfType<T extends TLShape = TLShape>(shapeId: TLShapeId, type: T['type']): boolean
+	isShapeOfType(arg: TLShape | TLShapeId, type: TLShape['type']) {
 		const shape = typeof arg === 'string' ? this.getShape(arg) : arg
 		if (!shape) return false
 		return shape.type === type
@@ -7769,9 +7762,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	canCreateShape<T extends TLUnknownShape>(
-		shape: OptionalKeys<TLShapePartial<T>, 'id'> | T['id']
-	): boolean {
+	canCreateShape(shape: OptionalKeys<TLShapePartial<TLShape>, 'id'> | TLShape['id']): boolean {
 		return this.canCreateShapes([shape])
 	}
 
@@ -7782,8 +7773,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	canCreateShapes<T extends TLUnknownShape>(
-		shapes: (T['id'] | OptionalKeys<TLShapePartial<T>, 'id'>)[]
+	canCreateShapes(
+		shapes: (TLShape['id'] | OptionalKeys<TLShapePartial<TLShape>, 'id'>)[]
 	): boolean {
 		return shapes.length + this.getCurrentPageShapeIds().size <= this.options.maxShapesPerPage
 	}
@@ -7801,7 +7792,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	createShape<T extends TLUnknownShape>(shape: OptionalKeys<TLShapePartial<T>, 'id'>): this {
+	createShape<TShape extends TLShape>(shape: OptionalKeys<TLShapePartial<TShape>, 'id'>): this {
 		this.createShapes([shape])
 		return this
 	}
@@ -7819,7 +7810,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	createShapes<T extends TLUnknownShape>(shapes: OptionalKeys<TLShapePartial<T>, 'id'>[]): this {
+	createShapes<TShape extends TLShape = TLShape>(
+		shapes: OptionalKeys<TLShapePartial<TShape>, 'id'>[]
+	): this {
 		if (!Array.isArray(shapes)) {
 			throw Error('Editor.createShapes: must provide an array of shapes or shape partials')
 		}
@@ -8316,7 +8309,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	updateShape<T extends TLUnknownShape>(partial: TLShapePartial<T> | null | undefined) {
+	updateShape<T extends TLShape = TLShape>(partial: TLShapePartial<T> | null | undefined) {
 		this.updateShapes([partial])
 		return this
 	}
@@ -8333,7 +8326,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	updateShapes<T extends TLUnknownShape>(partials: (TLShapePartial<T> | null | undefined)[]) {
+	updateShapes<T extends TLShape>(partials: (TLShapePartial<T> | null | undefined)[]) {
 		const compactedPartials: TLShapePartial<T>[] = Array(partials.length)
 
 		for (let i = 0, n = partials.length; i < n; i++) {
@@ -10766,7 +10759,10 @@ function alertMaxShapes(editor: Editor, pageId = editor.getCurrentPageId()) {
 
 function applyPartialToRecordWithProps<
 	T extends UnknownRecord & { type: string; props: object; meta: object },
->(prev: T, partial?: Partial<T> & { props?: Partial<T['props']> }): T {
+>(
+	prev: T,
+	partial?: T extends T ? Omit<Partial<T>, 'props'> & { props?: Partial<T['props']> } : never
+): T {
 	if (!partial) return prev
 	let next = null as null | T
 	const entries = Object.entries(partial)
--- a/packages/editor/src/lib/editor/derivations/parentsToChildren.ts
+++ b/packages/editor/src/lib/editor/derivations/parentsToChildren.ts
@@ -11,7 +11,7 @@ function fromScratch(
 ) {
 	const result: ParentShapeIdsToChildShapeIds = {}
 	const shapeIds = shapeIdsQuery.get()
-	const shapes = Array(shapeIds.size) as TLShape[]
+	const shapes = Array<TLShape>(shapeIds.size)
 	shapeIds.forEach((id) => shapes.push(store.get(id)!))
 
 	// Sort the shapes by index
--- a/packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
+++ b/packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
@@ -1,4 +1,4 @@
-import { TLBaseShape } from '@tldraw/tlschema'
+import { TLShape } from '@tldraw/tlschema'
 import { lerp } from '@tldraw/utils'
 import { Geometry2d } from '../../primitives/geometry/Geometry2d'
 import { Rectangle2d } from '../../primitives/geometry/Rectangle2d'
@@ -7,7 +7,7 @@ import { ShapeUtil, TLResizeInfo } from './ShapeUtil'
 import { resizeBox } from './shared/resizeBox'
 
 /** @public */
-export type TLBaseBoxShape = TLBaseShape<string, { w: number; h: number }>
+export type TLBaseBoxShape = Extract<TLShape, { props: { w: number; h: number } }>
 
 /** @public */
 export abstract class BaseBoxShapeUtil<Shape extends TLBaseBoxShape> extends ShapeUtil<Shape> {
--- a/packages/tldraw/src/lib/shapes/note/NoteShapeTool.test.ts
+++ b/packages/tldraw/src/lib/shapes/note/NoteShapeTool.test.ts
@@ -1,3 +1,4 @@
+import { TLNoteShape } from '@tldraw/tlschema'
 import { TestEditor } from '../../../test/TestEditor'
 import { NoteShapeTool } from './NoteShapeTool'
 
@@ -242,7 +243,7 @@ describe('Adjacent note position helpers (sticky pits)', () => {
 
 	it('Falls into correct pit below notes with growY', () => {
 		editor.createShape({ type: 'note', x: 0, y: 0 }).updateShape({
-			...editor.getLastCreatedShape(),
+			...editor.getLastCreatedShape<TLNoteShape>(),
 			props: { growY: 100 },
 		})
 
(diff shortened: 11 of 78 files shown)


Andarist:

SomeHats:
@claude merge main and resolve conflicts

apps/examples/src/examples/after-create-update-shape/AfterCreateUpdateShapeExample.tsx
```diff
@@ -14,6 +14,10 @@ function ensureOnlyOneRedShape(editor: Editor, shapeId: TLShapeId) {
 	const otherRedShapesOnPage = Array.from(editor.getPageShapeIds(pageId))
 		.map((id) => editor.getShape(id)!)
 		.filter((otherShape) => otherShape.id !== shape.id && isRedShape(otherShape))
```
SomeHats: nit: i'd maybe just move the `: shape is ...` onto this `.filter` call in this case rather than add another one

apps/examples/src/examples/after-delete-shape/AfterDeleteShapeExample.tsx
```diff
@@ -51,7 +51,7 @@ function createDemoShapes(editor: Editor) {
 		},
 		...[50, 180, 310].map((x) => ({
```
SomeHats, line 52: mayyybe return annotation here instead of the `as const`? so you get proper autocomplete on the rest

apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
```diff
@@ -6,6 +6,12 @@ import 'ag-grid-community/styles/ag-grid.css'
 import 'ag-grid-community/styles/ag-theme-quartz.css'
 import 'tldraw/tldraw.css'
 
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		'ag-grid': AgGridShape
```
SomeHats: So, some options for what the API here could be:

1.
```tsx
type AgGridShape = TLBaseShape<'ag-grid', {w: number, h: number}>

declare module '@tldraw/tlschema' {
    // Not `GlobalShapePropsMap`, because it contains shapes not props
    export interface GlobalShapeMap {
        // text label here is arbitrary
        'ag-grid': AgGridShape
    }
}
```

2.
```tsx
type AgGridShape = TLBaseShape<'ag-grid', {w: number, h: number}>

declare module '@tldraw/tlschema' {
    export interface GlobalShapePropsMap {
        // text label here is arbitrary, but we're just specifying the props
        'ag-grid': {w: number, h: number}
    }
}
```

3.
```tsx
declare module '@tldraw/tlschema' {
    export interface GlobalShapePropsMap {
        // text label here becomes part of the shape def
        'ag-grid': {w: number, h: number}
    }
}

// something like this:
type AgGridShape = TLShape<'ag-grid'>
```

I think 1. has the benefit of requiring the fewest changes from existing code, but i kind of prefer the API of 3 - feels like there are maybe less "loose ends"?

apps/examples/src/examples/bounds-snapping-shape/types.ts
```diff
@@ -0,0 +1,7 @@
+import { IPlayingCard } from './PlayingCardShape/playing-card-util'
+
+declare module '@tldraw/tlschema' {
```
SomeHats: nit: i maybe wouldn't put these in separate files, i think we're already more spread out than is helpful in these examples. I also don't want to suggest that devs have to / should use `.d.ts` files for these things.

apps/examples/src/examples/pin-bindings/PinExample.tsx
```diff
@@ -240,15 +248,15 @@ class PinBindingUtil extends BindingUtil<PinBinding> {
 			}
 		}
 
-		const updates = []
+		const updates: TLShapePartial<TLShape>[] = []
```
SomeHats: (i haven't read down to the newest def of `TLShapePartial` yet) could this be just `TLShapePartial` without the type param?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -965,8 +964,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	getShapeUtil<S extends TLUnknownShape>(shape: S | TLShapePartial<S>): ShapeUtil<S>
-	getShapeUtil<S extends TLUnknownShape>(type: S['type']): ShapeUtil<S>
+	getShapeUtil<S extends TLShape>(shape: S | TLShapePartial<S>): ShapeUtil<S>
+	getShapeUtil<S extends TLShape>(type: S['type']): ShapeUtil<S>
```
SomeHats: I think right now `editor.getShapeUtil('arrow')` will return `ShapeUtil<TLShape>`. Is there a way we can tweak this so it'll return `ShapeUtil<TLArrowShape>`? type-level tests for this would be awesome

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5391,15 +5390,9 @@ export class Editor extends EventEmitter<TLEventMap> {
...
-	isShapeOfType<T extends TLUnknownShape>(
-		shapeId: TLUnknownShape['id'],
-		type: T['type']
-	): shapeId is T['id']
-	isShapeOfType<T extends TLUnknownShape>(
-		arg: TLUnknownShape | TLUnknownShape['id'],
-		type: T['type']
-	) {
+	isShapeOfType<T extends TLShape>(shape: TLShape, type: T['type']): shape is T
+	isShapeOfType<T extends TLShape = TLShape>(shapeId: TLShapeId, type: T['type']): boolean
```
SomeHats: ditto here: would love to make sure that `editor.isShapeOfType(shape, 'arrow')` is refining the type of `shape` as expected

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -7686,7 +7677,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 *
 	 * @public
 	 */
-	createShape<T extends TLUnknownShape>(shape: OptionalKeys<TLShapePartial<T>, 'id'>): this {
+	createShape<TShape extends TLShape>(shape: OptionalKeys<TLShapePartial<TShape>, 'id'>): this {
```
SomeHats, line 7721: not urgent, but it's maybe worth spending some time looking at these type parameters and deciding whether we think they're still all that useful in this world

packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
```diff
@@ -7,7 +7,7 @@ import { ShapeUtil, TLResizeInfo } from './ShapeUtil'
 import { resizeBox } from './shared/resizeBox'
 
 /** @public */
-export type TLBaseBoxShape = TLBaseShape<string, { w: number; h: number }>
+export type TLBaseBoxShape = Extract<TLShape, { props: { w: number; h: number } }>
```
SomeHats: half-formed API thought: we use this pattern a lot. maybe a `ShapesMatching<{w: number, h: number }>` helper would be useful. but also maybe this isn't a pattern we want to encourage

packages/tldraw/src/lib/shapes/note/NoteShapeTool.test.ts
```diff
@@ -1,3 +1,4 @@
+import { TLNoteShape } from '../../..'
```
SomeHats: nit: import from the actual file

packages/tldraw/src/lib/utils/excalidraw/putExcalidrawContent.ts
```diff
@@ -143,7 +143,7 @@ export async function putExcalidrawContent(
 						align,
 						dash: getDash(element),
 						fill: getFill(element),
-					},
+					} as any,
```
SomeHats: possibly more trouble than it's worth but i would love to find a way not to use `as any` so much here, we've gotten tripped up by stuff getting missed in this excalidraw pasting stuff before

packages/tlschema/src/records/TLShape.ts
```diff
@@ -52,12 +52,16 @@ export type TLDefaultShape =
...
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface GlobalShapePropsMap {}
+
 /**
  * The set of all shapes that are available in the editor, including unknown shapes.
  *
  * @public
  */
-export type TLShape = TLDefaultShape | TLUnknownShape
+export type TLShape = TLDefaultShape | GlobalShapePropsMap[keyof GlobalShapePropsMap]
```
SomeHats: one thing we'll need is a way for people to override or exclude our default shapes like
```tsx
interface GlobalShapePropsMap {
    arrow: null
}
```

do u think we can maybe get creative here in a way that'll allow that?

Also optional, but I'm quite a fan of being able to do something like `TLShape<'arrow'>` to be able to access a shape of a specific type. wdyt?

Andarist:
1. The labels in the API don't really matter so they can be arbitrary:
```ts
declare module '@tldraw/tlschema' {
	export interface GlobalShapePropsMap {
		'my-grid-shape': MyGridShape
		'my-counter-shape': MyCounterShape
	}
}
```

There are other options on the table though so we should discuss further about them [here](https://github.com/tldraw/tldraw/pull/6739/files#r2388595654).


2. sometimes contextual return types don't work, especially with code like:
```ts
editor.createShapes([
    ...stuff.map(x => ({
        // no autocompletes here
        // and the below string literal widens to `string`
        type: 'geo',
    }))
])
```

It is a TS issue though: [TS playground](https://www.typescriptlang.org/play/?ts=5.9.3#code/C4TwDgpgBAysCuAzRUC8AoKUA+UDeUokAXFAEQCGZA3FAM4AWFAThACYAKzA9mKQHbwAtgCMIzKAF9MOfIXARSZETXpNWnHnyiDR4qdXTo2EAMYAbFtETx+p4AEtu-KKdYVgEOEkQAKOgjIpN7IANoAugCUpLYA1vzcAO78huhuEB5egX6hMgB0BaEAjAA0UABM4XlCFGC+vgD6kWgAfFC+eDJYRIrkVCVdalaavKSlMpKRkQNRhkA)

Workarounds:
- `as const`
- an explicit return type annotation

3. Similarly contextual typing doesn't apply to class fields, so `as const` is required in:
```ts
override shapeType = 'PlayingCard' as const
```

This is also a TS issue~ (this time more like a design decision for *reasons*).

4. Due to how TS might "flatten unions" it's necessary to use spread operator when your source data contains more than one type of a shape:
```diff
editor.updateShapes(
    shapes.map((shape, i) => ({
-        type: shape.type,
+        ...shape,
        id: shape.id,
        isLocked: shape.isLocked,
        index: indexes[i],
    }))
```

This allows TS to distribute `shapes` and output correct (and compatible!) union elements like `{ type: 'a', ...rest, ...common } | { type: 'b', ...rest, ...common }` instead of mixing up everything into a flat: `{ type: 'a' | 'b', ...allRests, ...common }`.

Changes in `templates/agent/client/components/chat-history/TldrawDiffViewer.tsx` were made for similar reasons.

5. This is kind of an internal annoyance. But given module augmentation is global~ per tsconfig.json location - you might have to use different string types in tests and examples so you don't accidentally use different types for different shape types. This leads to usage of `type: "test1"`, `type: "test2"` etc for one-off shapes in tests. Luckily [this is an error](https://www.typescriptlang.org/play/?ts=5.9.3#code/JYOwLgpgTgZghgYwgAgCoQM5mQbwL4BQBoksiK6WuByycAXNbbWAJ4AOEjARHNwNw1m7KAHt2jEAFcAtgCNog2nkGFi4aPCRpM2HEIZNmbTjz5LhYicixRQAcwsqChIA) so there are some builtin guardrails that will help you to avoid accidental mistakes.

6. In the past, a bunch of methods had generics on them. Technically, with this change - they wouldn't have to. Unions should be nicely discriminated based on the type and the generics are somewhat obsolete. However, they can still serve validation~ purpose. One can use generics to constraint what kind of arguments can actually be passed at some location:
```ts
editor.createShape<ArrowShape>(onlyArrowsAcceptedHere)
```

Without generics the above can only be achieved with (or with multiple statements etc):
```ts
editor.createShape(onlyArrowsAcceptedHere satisfies ArrowShape)
```

There are also still some APIs that are not type-safe out of the box so the generic acts as an assertion of sorts:
```ts
const note = editor.getLastCreatedShape<TLNoteShape>()
```

In reality, this is as good as an assertion:
```ts
const note = editor.getLastCreatedShape() as TLNoteShape
```

You could try to eliminate those APIs or at least require `type` to be passed in and perform a runtime-check based on that. Some APIs might already require `type` but IIRC some don't.

There are also some APIs that are just a little bit too complex today and fail to infer - even if theoretically they could. I could look into them to see if the situation can be improved in user-land but I'm not sure if it can.

So all in all, the generic type parameters should probably still be preserved on your signatures.

7. At times, I had to reach for things like this:
```ts
export type TLBaseBoxShape = Extract<TLShape, {
    props: {
        h: number;
        w: number;
    };
}>;
```

It's functional but perhaps not ideal when done in the user-space so aliases for things like this should be considered. I don't think this is a big issue though, users should rarely have to do things like this. 

It just means that previously you could declare a base box shape by typing out the required props and you wouldn't care about the `.type` at all. Now it has to be derived so your "generic" utilities functioning on a certain common props are just bound to that base union. That said, you likely want to phase out usage of such common props altogether and rely on standardized root properties for things like this and only hold shape-specific props in `props`. 

8. When writing migration code (in tests) variables like `oldArrow` had to be `as any`ed. This is pretty common problem and it's just up to the user to write a type-safe migration code if they don't want to risk anything. 

9. To make things more type-safe, I had to now introduce some local helpers like `isShapeWithUrl`/`assertShapeWithUrl`. This is pretty standard. I'm just pointing it out because similar checks might have to be written more often by users now if they don't want to use casting. The previous version of the touched code just wasn't type-safe enough.

10. No more filtering by intersection:
```ts
editor.getShape<TLShape & { props: { altText: string } }>(shapeId)
```

The code above should be written now like this:
```ts
const shape = editor.getShape<Extract<TLShape, { props: { altText: string } }>>(shapeId)
```

In those instances, the intersection is just not enough to filter out/reduce the `TLShape` union.

11. In `packages/tldraw/src/lib/utils/excalidraw/putExcalidrawContent.ts` I resorted to some temporary `as any`s. Similarly, there are temporary `as any`s in `createShapeRecordType` - that deals with your custom runtime validators. This stuff can surely be improved and I'll look into it.

apps/examples/src/examples/after-create-update-shape/AfterCreateUpdateShapeExample.tsx
Andarist: yeeah, this is definitely some oversight here - I probably have tried to leverage inferred type predicates here but, unfortunately, "deep properties" don't contribute to narrowing their parent object types 😢 [TS playground](https://www.typescriptlang.org/play/?ts=5.9.3#code/C4TwDgpgBAglC8UDeBYAUFKpIC4oCIBDfAbnUzACcB7MAZz1Q0ygGNqAbayvO4SgJYA7AOZlmAX3FT06bNABCCZOSzgIefACNSqqrQYrmmasAAWEHlCEBXALZbL4zDLSv0AEwisOhStHYhPigBYAg7PDgAHygFcXQBADMoAAp8di5KfBChELC7ADp9egBKI0xQ8OcoAHoaqAA9AH50CXQgA). It would be so sweet to get this https://github.com/microsoft/TypeScript/pull/38839

apps/examples/src/examples/after-delete-shape/AfterDeleteShapeExample.tsx
Andarist: This actually should be typed as `OptionalKeys<TLShapePartial<TLGeoShape>, 'id'>` and that is rather complex for the user to write. I think this is somewhat fine - but you are right this suffers from the lack of autocompletion and that's not really great. 

So I'd consider introducing a helper for this, maybe something like:
```ts
type PartialCreateShape<T extends TLShape> = OptionalKeys<TLShapePartial<T>, 'id'>
```

apps/examples/src/examples/bounds-snapping-shape/types.ts
Andarist: In principle, I like the idea of those examples being somewhat isolated - it aids copy-pasting. I don't mind introducing the requested change if requested though.

The `.d.ts` thing is a separate issue - and I agree, those should be regular `.ts` files and I'll rename those files.

apps/examples/src/examples/pin-bindings/PinExample.tsx
Andarist: ye, this works without the explicit type param given this type alias has the same default

packages/editor/src/lib/editor/Editor.ts
Andarist: I'll add this but tbf this overload will now be somewhat obsolete given this works:
```ts
if (shape.type === 'arrow') { shape /* TLArrowShape */ }
```

Discriminated unions ❤️

packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
Andarist: I'd probably wait with it to test out the waters how often the need for this actually comes up. It's up to you though.

packages/editor/src/lib/editor/Editor.ts
Andarist: Sure thing! I just added this locally and that helps me with fixing `as any`s in `packages/tldraw/src/lib/utils/excalidraw/putExcalidrawContent.ts` given I can now do this:
```ts
tldrawContent.shapes.push({
	...base,
	type: 'geo',
	props: {
		...editor.getShapeUtil('geo').getDefaultProps(),
		/* rest below */
	},
})
```

packages/tlschema/src/records/TLShape.ts
Andarist: > do u think we can maybe get creative here in a way that'll allow that?

yeah, it's definitely possible - I'll discuss those requirements with @ds300 

> Also optional, but I'm quite a fan of being able to do something like TLShape<'arrow'> to be able to access a shape of a specific type. wdyt?

Yeah, this would be neat :) and its totally possible with something like:
```ts
type AllTLShapes = TLDefaultShape | GlobalShapePropsMap[keyof GlobalShapePropsMap]
export type TLShape<K extends AllTLShapes['type'] = AllTLShapes['type']> = K extends K ? Extract<AllTLShapes, { type: K }> : never
```

A version I like better is something like this:
```ts
type IndexByProp<T extends Record<P, string>, P extends keyof T> = {
	[E in T as E[P]]: E
}

type AllTLShapes = TLDefaultShape | GlobalShapePropsMap[keyof GlobalShapePropsMap]

export type TLShape<K extends keyof IndexByProp<AllTLShapes, 'type'> = keyof IndexByProp<AllTLShapes, 'type'>> = IndexByProp<AllTLShapes, 'type'>[K]
```

But for some reason, I run into a spurious:
> Type alias 'TLShape' circularly references itself.ts(2456)

This technique should very much work though ([TS playground](https://www.typescriptlang.org/play/?ts=5.9.3#code/C4TwDgpgBAkgdgEwgDwEIgAoCcD2YA8AKlCsBIgM5QBKEAxjlgvhgDRQXBYCWcA5gD52GEsjKUoAawggcAMyiEBUALxQA3gFgAUAEgA2gFEovRVACGVQ-owBdWwC4ohnQF8dO0JEUAZAIJYuADuAMoAFubeaupQXhBOAOTmgThBCVDu2p7g0IQ+qDjI4ZHQ0bE5iQBGhemZ2d5+ADaNecWQVGp5AcFt0AA+vgVFEZAe2ihgjMDl3q0jEPgA0qLiCFTSsgrwSGiYuARNLT69FOwJcQnKahvysIgo6Nh4+IdzJadQ5zmXV3c7j-sXs03u0zhcBPpFrYxnEoIdVL5ejoAPTIqDogB6AH4YTkoAApACunG6qQRXRSoXmKLRmKxQA)) and the circularity error I get is surprising

packages/editor/src/lib/editor/Editor.ts
Andarist: I feel I posted my take on this [here](https://github.com/tldraw/tldraw/pull/6739#issuecomment-3397094418)

packages/editor/src/lib/editor/derivations/parentsToChildren.ts
```diff
@@ -11,7 +11,7 @@ function fromScratch(
 ) {
 	const result: ParentShapeIdsToChildShapeIds = {}
 	const shapeIds = shapeIdsQuery.get()
-	const shapes = Array(shapeIds.size) as TLShape[]
+	const shapes = Array<TLShape>(shapeIds.size)
```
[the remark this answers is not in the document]
Andarist: Nice catch, https://github.com/tldraw/tldraw/pull/6936

apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
Andarist: I like the 3rd option the most too. @ds300 what are your thought on this one?

ds300: yep 3rd is best IMO. Simplest and cleanest. And it has interesting implications for the `TLShape` type that I think would be good.

apps/examples/src/examples/drag-and-drop/DragAndDropExample.tsx
```diff
@@ -11,6 +11,13 @@ import {
 } from 'tldraw'
 import 'tldraw/tldraw.css'
 
+declare module '@tldraw/tlschema' {
+	export interface GlobalShapePropsMap {
+		'my-grid-shape': MyGridShape
```
ds300: if possible would be nice to have a pattern where the shape type name is stored in a const and can be a single source of truth
```ts
export const MY_GRID_SHAPE = 'my-grid-shape'

declare module '@tldraw/tlschema' {
	export interface GlobalShapePropsMap {
		[MY_GRID_SHAPE]: MyGridProps
	}
}

export type MyGridShape = TLShape<MY_GRID_SHAPE>

export class MyGridShapeUtil {
  override shapeType = MY_GRID_SHAPE
}
```

etc etc

apps/examples/src/examples/editable-shape/EditableShapeUtil.tsx
```diff
@@ -89,28 +89,28 @@ export class EditableShapeUtil extends BaseBoxShapeUtil<IMyEditableShape> {
...
-This is our shape util, which defines how our shape renders and behaves. For 
+This is our shape util, which defines how our shape renders and behaves. For
 more information on the shape util, check out the custom shape example.
 
 [1]
 We override the canEdit method to allow the shape to enter the editing state.
 
 [2]
-We want to conditionally render the component based on whether it is being 
+We want to conditionally render the component based on whether it is being
```
ds300, line 106: any idea what part of your dev environment is cleaning up whitespace in comments?

not complaining, i think it'd be good for us to have in our stack.

apps/examples/src/examples/external-content-sources/ExternalContentSourcesExample.tsx
```diff
@@ -61,7 +67,7 @@ export default function ExternalContentSourcesExample() {
 				const center = content.point ?? editor.getViewportPageBounds().center
 
 				editor.createShape({
-					type: 'html',
+					type: 'dangerous-html',
```
ds300, line 64: one thing that I used to see a lot with copilot (pre agents) was that it expected `editor.createShape` to take the shape type as a first parameter, e.g. it would try to do

```ts
editor.createShape('dangerous-html', {
  props: { ... }
})
```

and i kinda feel like it would make sense to add that as a signature overload as part of this change. although of course it could happen later if that would balloon the diff.

apps/examples/src/examples/pin-bindings/PinExample.tsx
```diff
@@ -240,15 +247,15 @@ class PinBindingUtil extends BindingUtil<PinBinding> {
...
 			if (delta.len2() <= 0.01) continue
 
 			const newPosition = this.editor.getPointInParentSpace(shapeId, position)
 			updates.push({
+				...this.editor.getShape(shapeId)!,
 				id: shapeId,
-				type: this.editor.getShape(shapeId)!.type,
 				x: newPosition.x,
 				y: newPosition.y,
 			})
```
ds300, line 263: the need for this change hints at another related issue that has cropped up before:

The `type` param isn't strictly necessary for update partials. It's mainly useful for type checking the props, or providing intellisense. I almost made it optional a few months ago. I think this project would be a good opportunity to try that again and see if it makes sense.

apps/examples/src/examples/bounds-snapping-shape/types.ts
ds300: i'm with alex on this one, generally I like to group things by behavior rather than by kind, and in this case I think it makes more sense to define the types alongside the code that uses them (i.e. the shape util, for the most part)

packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
ds300: I think we should definitely make an internal helper type for this, but yeah I don't think it should be a common pattern for sdk users so no need to make it public until we have evidence to the contrary.

packages/tldraw/src/test/Editor.test.tsx
```diff
@@ -866,9 +917,17 @@ describe('instance.isReadonly', () => {
...
+	export interface GlobalShapePropsMap {
+		'my-custom-shape': MyCustomShape
+	}
+}
+
 describe('the geometry cache', () => {
-	class CustomShapeUtil extends BaseBoxShapeUtil<any> {
-		static override type = 'custom'
+	class CustomShapeUtil extends BaseBoxShapeUtil<MyCustomShape> {
+		static override type = 'custom' as const
```
ds300: shouldn't this be changed to 'my-custom-shape' too? is this test passing? i'd be surprised if so.

packages/tldraw/src/test/groups.test.tsx
```diff
@@ -35,14 +35,16 @@ const ids = {
...
+
 const box = (
 	id: TLShapeId,
 	x: number,
 	y: number,
 	w = 10,
 	h = 10,
-	fill = 'solid'
-): TLShapePartial => ({
+	fill: BoxShape['props']['fill'] = 'solid'
```
ds300, line 47: nit: it already knows the type it's just failing to narrow the string so we can use 'as const'

```suggestion
	fill = 'solid' as const
```

packages/tlschema/src/records/TLShape.ts
ds300: > do u think we can maybe get creative here in a way that'll allow that?

> yeah, it's definitely possible - I'll discuss those requirements with @ds300

One easy way to do this would be to have a separate augmentable interface for specifying which builtin shapes to exclude

```ts
export interface GlobalExcludeBuiltinShapesMap { }

...
// MyGeoShape.tsx
declare module '@tldraw/tlschema' {
  export interface GlobalExcludeBuiltinShapesMap {
    geo: true
  }
	export interface GlobalShapePropsMap {
		geo: MyGeoShapeProps
	}
}
```

and then we just use the keys of that interface to `Exclude` from the `TLDefaultShape` type or whatever it ends up being.

ds300 (commented):
This is fantastic work, thanks @Andarist Let's catch up on discord to discuss next steps.

packages/tlschema/src/records/TLShape.ts
ds300: wait we wouldn't even need a separate interface if default shapes aren't declared in the `GlobalShapePropsMap` like the current state of this PR.

apps/examples/src/examples/editable-shape/EditableShapeUtil.tsx
Andarist: I'm not entirely sure but it's probably this in my VS Code:
```
"files.trimTrailingWhitespace": true
```

`.editorconfig` has similar settings and should be portable between IDEs:
```
trim_trailing_whitespace = true
```

apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
Andarist: I'll make the changes

apps/examples/src/examples/drag-and-drop/DragAndDropExample.tsx
Andarist: Sure, that makes sense. I'll make the changes

apps/examples/src/examples/external-content-sources/ExternalContentSourcesExample.tsx
Andarist: Yeah, I think if you are open to considering alternative APIs then we could discuss/experiment/implement that separately.

apps/examples/src/examples/pin-bindings/PinExample.tsx
Andarist: Hm, ye - I get why it's needed here. Similarly, changing `type: ` to a spread was done here solely to satisfy types. Given the target here (`updates`) allows different strictly-typed~ shapes, it needs that discriminator in some form. I'm not sure what we could really do about it

packages/tldraw/src/test/groups.test.tsx
Andarist: It would typecheck *here* but not at the call sites as this would produce this signature:
```ts
const box: (id: TLShapeId, x: number, y: number, w?: number, h?: number, fill?: "solid") => TLShapePartial<BoxShape>
```

So now 2 out of 3 calls here wouldn't succeed:
```ts
		editor.createShapes([
			box(ids.boxA, 0, 0, 10, 10, 'none'),
			box(ids.boxB, 20, 0, 10, 10, 'solid'),
			box(ids.boxC, 0, 50, 10, 10, 'none'),
		])
```

packages/tldraw/src/test/Editor.test.tsx
Andarist: Good catch. It should be `'my-custom-shape'`. I'll look into why this test is passing rn

Andarist: `applyPartialToRecordWithProps` ignores `type` updates (rightfully so). This actually gets created with `custom` type through:
```ts
const { A } = editor.createShapesFromJsx([<TL.custom ref="A" x={0} y={0} w={100} h={100} />])
```

It means that I've missed `TL.` magic~ and I have not updated its types. Gonna do that and fixup this place.

packages/tldraw/src/test/groups.test.tsx
ds300: ah sorry my bad. for some reason when i read this i thought it was in a destructuring expression.

apps/examples/src/examples/external-content-sources/ExternalContentSourcesExample.tsx
ds300: let's actually forget about this for now. I don't wanna distract from the main work here and I'm pretty certain it will be possible to add this kind of API later no matter what.

apps/examples/src/examples/pin-bindings/PinExample.tsx
ds300: > it needs that discriminator in some form

Only if you're updating the props (and you care about type safety). 

e.g. `updateShape({id, x, y})` only updates properties that are common to all shapes, so it's type safe without specifying the `type` property.

So i think we could make `type` optional for situations like this where you're not updating the props, or when you don't care about type safety.

But also this is kinda distracting from the main task here so, again, let's forget about it for now.

packages/tldraw/src/test/Editor.test.tsx
Andarist: Fixed.

packages/tlschema/src/records/TLShape.ts
Andarist: From the implementation's PoV a separate interface would be cleaner but I could totally squeeze this into a single interface and perhaps that would be cleaner from the usage's PoV.

What do we need to support? 
1. adding new shapes
2. excluding existing shapes
3. overriding existing shapes

Could you confirm that all 3 are required? Are there any other things I should think about in this context?

packages/editor/src/lib/editor/shapes/BaseBoxShapeUtil.tsx
Andarist: Added it as `ExtractShapeByProps`

Andarist: Actually, I had to make it a public util - internal ones are stripped from the public `.d.ts` output and that creates issues given the location of the util (`@tldraw/tlschema`) and its various consumers in different packages (editor, tldraw, public api part of the tlschema itself)

apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
Andarist: The requested `TLShape` runs into a TS issue: [TS playground](https://www.typescriptlang.org/play/?moduleResolution=100&ts=6.0.0-dev.20251021#code/C4TwDgpgBAShDGB7ATgEwJKoDwyhAHsBAHaoDOUAqsQNbGIDuxcSaAfFALxRnDICWxAOZQAZFADeUAPrTQkWQC5YUAL4BuAFCb50anUbMEKVFygAhAIZkILE1l4DhAGljG0mLPvpM77NlqagkTIAGaW8NBWNn7YmlBQACrgEABylgC20AREpBSOgkLO8VCYeIQk5G6sGNjehrFsxRwSJcgQlqiIxAA2IFD8qMqYWgntnd19ULrpWcrJkLMQWqraAPRr0wAW0ABG1tC6UDvtUPDjRBSWZ-zI8ACuPZYCoJobpQDkpmSW-cA7UGQiEQwDOlnuNgG+X4GX4T2Q00Qx2AwDAZEUGyE-H+912ADokBk1rDzogyIhQsA1gsIABlc78MBU-hkMj3CBkNYANgATABGAAcAAY3ptdvdQcAkbtoGzTtioAxEI9TFtLAA3Q7S6CCTW8fhCSxEUyhe7If4QZBBYghcKRJIAGWidLVkCwNPKuSqBRcUAACkC0Z7KhRELsAFYIYBsEo5EMWA6xLAAIjIrogydciQdtPTmBaJV08xSoyg+GUxHuGRlyFLIArVZrpbAgfR-tbK3WmwVSuQNAoDGxW22ewgPUYUE1yDI-G62yNA16gg5I6g+0hui7i9tEWg2educsbo9cbyPD4hVcAcQQdPVTDkfg0ckooSgI6XV6-UG8xzedQWjvGMH6TH8KRLMoqbpsmgGbAkRZJCWr4JOWUCVtWlqwW+9ZoY2mHIVALY3m215olhqw6CkjoAILIECDCHpAZj7gcjEQCmzz0Zmkg8PwABeEANhhCKqAE2hHNmAAiEDhI8wBscxDq0fRbGBEc1E9D0bEUNwUkyeCPTyemgTvP8LJQF0HLEB8oK9jQeBDpaiiiq+RzaeYIAetwrRAVAADaiSLlAGlaemZB+cmujJgAutFygAKKEMgETAFgIXaa4UgIYFomvho4lUe5nlUd5JR+fFQXpWFUDWFA8URVFsUJSsalUdmbFYAA0sGZ40BAIAUlARVeVAfUDaEQ1hR5NIcJwJTDSkfmddFrVMe1-5mLEnjrUeEBiUAA). 

While I have a workaround already (presented in the playground above), I'll play around with it more to see if I can come up with something better

packages/tlschema/src/records/TLShape.ts
ds300: yes exactly, all 3 of those are required.

ds300: i can't think of any other nuance here, it feels pretty straightforward to do.

We'd have one internal shape type

```ts
interface TLDefaultShapes {
  geo: ...
  note: ...
  ...
}
```

and then the userland ones

```ts
interface TLGlobalShapePropsMap {
  geo: null
  note: MyCustomNoteShape
  ...
}
```

And then when we create the final TLShape we simply `Omit<TLDefaultShapes, keyof TLGlobalShapePropsMap>` before merging them together?

apps/examples/src/examples/ag-grid-shape/DataGridExample.tsx
ds300: the workaround seems fine to me. Is there something about it you don't like?

but also wouldn't the correct id type for TLBaseShape be something like `RecordId<this>` instead of `TLShapeId` ? (it also seems to work ok from what I can tell)

Andarist: > the workaround seems fine to me. Is there something about it you don't like?

Just the duplication and lack of a direct relationship between this type and its "base type". 

> but also wouldn't the correct id type for TLBaseShape be something like RecordId<this> instead of TLShapeId ? (it also seems to work ok from what I can tell)

In the codebase, this introduces a lot of errors. I'd leave it aside for now but I could take a stab at fixing them later

packages/tlschema/src/records/TLShape.ts
Andarist: The use case of overriding core shapes with custom ones is a little bit tricky from the types PoV. It's not hard to implement it as proposed *but* we there is an important thing to be said about it. Your codebase gets compiled assuming the core shapes are **not** overriden. 

This could lead to some runtime issues if any codepaths would assume anything about those shapes which isn't unlikely. At the same time, this might not be a strong concern because - I assume - this is an advanced use case and the compatibility burden gets shifted onto the user overriding it.

What is safe is an *extension* so I spent some time playing around with ideas how one could implement that while forbidding invalid/incompatible extensions: [TS playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251027#code/CYUwxgNghgTiAEYD2A7AzgF3hiwZQHcAFGJANwEtRgAueAVxQoEd6E0BPAWwCMkIA3ACghGDgAcEAMSgRO8ALzwU9CBHgAfBilAAzCihDBN8XbLQhhoiQgDCSOAGUAFlEkAeEknFoAfIvgvHxMxSSRdbFx8YlJKahMZOQ4rAwwQGDMwBAAVABl7J1dJNHgAbyF4RFhaeAKQFzcQd1LEfgc6TBgDAHN4AF9fYUq+AA86OoaPFrHlel50gXgOOhV5mH7BivhYUgJxh3qippbgChgOjC6UXoGh+DiQJH3CxubsCi4QFbmeBfhxaAcHp0Pj8EBQFAbYR9ESpdKZBAAcQgSB4skmICCaAAsm54CARmkdCU8hMjiVypUwNVvmpFgB6en4kaQeigeAECgYZymcwcLZgeiYJBcOgtNEALx6UAuVxuDKZBKJxk53O2rS4AJAaQgHEQQowIvgaCOWweTzK70+3zWiwBUCB1xBSDBEMWsgw2SVsp6-QVzOV8AM6uQmqgGAoPAgCAIDqEMKEcIyUCy8GRqPRRyxuPEAZAxPgpIOGIpW0Z2xguzFHNcaTI6R91368HLBlD4nDkejefQFFQiDg4ZAJQh+MrDiDEQ4SHo2lDnxQWC58dhi-hKZyuQAQlALBj3NkbAAaQKkHz+SnYGx0Q+SO7iM9oOhY6EiUIIACCajyJYCl4A2gA0kGkIANYgNOETpmiEAYtmbgALo0GWTIUBEXLGhGajwK4JTcggSYoLIkR4IQXgPMYnC8Pw2DOPmQZYLhyhIFgvz0Ya9BgHRxg8Hq+HbPQ3QLhgnb9hCHDOEgBAofA0YYAA5CUABWBrwHAXDkNIfL3LIbAlAAFNy4bbGoUk9qcTaqjyiTyIazKsuy-Emo0ACU2w6LR7DcHw6jRlAlBNqguqef8j5bO+4QkdE5FUEY5klNBmaNPB4hAQhWyVAA-IW267ocrwZZURWAUehVFfAACiLIQGyTSJbBWaPjmaUnjZepaBFEQ4KRMTkLFwC+GV8CDeVrZdXR2zdHM+ZpMYZC6QRJRtZ5kIYUxbGQgSDlGCecBoPw9bYEg8AAAaGPWMAnUdxqgRQubLuVdD1XBTVuGl8XwG1hXZed6SFeWgrClwxrkmp4JcbRFAlDw+AoFxhU3rle5HO4JVpiiMEvd4OJvYBCGbH0-7gZB6MZg1yWveICFWOWe0HXFX4QD+5LIYyQjlu+8CM8zjQlEoeQ7sjrwAESA4aXDCyeLQAwaRrOZI7NMuVkrSo23TCOWAwmALeX7sL5qS5a5YAFRKvm1DG-csVIIr5VWl8sy2rb5X2o63TOq6KAa0rRUel6hJq97Gza0j+UeMLoyG9LTLIHA8CjM7RUzKsvwwEH5XLI7qdB1rWg60L4c7FJUctjHBwVrsieVKc5yYXKOeDeW5YACIAPIVY4ABy8nZDh-kIKb1W1cAlvUjAwBCEAA)

The core of that idea is that `TLGlobalShapePropsMap` itself could extend a registry of core shapes. This way, we could leverage TS protection against invalid extensions. 

But... as I mentioned, this very well might be considered YAGNI and redundant complexity. Let me know if you like this or if I should go with the simpler approach you have proposed above.

templates/agent/shared/actions/CreateActionUtil.ts
```diff
@@ -98,10 +98,9 @@ const SHARED_DEFAULTS = {
 	id: 'shape:shape' as TLShapeId,
 }
 
-const SHAPE_DEFAULTS: Record<TLShape['type'], TLShapePartial> = {
+const SHAPE_DEFAULTS = {
```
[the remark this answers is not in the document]
Andarist: This change unified the code with its counterpart/copy in `packages/fairy-shared`. The `type` is always set in the respective `convertX` functions (like `convertArrowShapeToTldrawShape`). I don't think those `type`s here were needed and removing them allowed types to pass.

That said, all of those `convertX` functions have asserts in them so clearly the types flowing here through those layers are not exactly airtight right now.

PR #6744, opened 2025-09-15
Update earthquake node label to 'Earthquake data'

SomeHats:
Updates the earthquake node tooltip label from "earthquake" to "earthquake data" as requested.

This changes the title property in the EarthquakeNodeType class, which is used for the tooltip text in the workflow toolbar.

🤖 Generated with [Claude Code](https://claude.ai/code)


SomeHats:
@claude also on the workflow builder, i notice that our node defs read like `class AddNodeType extends NodeDefinition`. Can you go through and rename all the node definitions to end with `Definition` instead of `Type` like this: `class AddNodeDefinition extends NodeDefinition`

PR #6745, opened 2025-09-15
feat(ui): add maximum file size to asset upload toast

kostyafarber:
When a user tries to upload a file that is larger than `maxAssetSize`, it would be useful to add a description of what that size is. This PR adds that description to the toast that is triggered by this behaviour.

## Before
image

## After
image

### Change type

- [x] `improvement`

### Test plan

1. Attempt to upload a file larger than the maximum asset size.
2. Verify the toast notification displays the maximum allowed size in the description.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added maximum file size to the description of the file too big toast, when uploading a file.


kostyafarber:

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -901,8 +901,22 @@ export function notifyIfFileNotAllowed(file: File, options: TLDefaultExternalCon
...
+
+			const value = bytes / Math.pow(base, unitIndex)
+			const formatted = value % 1 === 0 ? value.toString() : value.toFixed(2)
+
+			return `${formatted} ${units[unitIndex]}`
+		}
+
 		toasts.addToast({
 			title: msg('assets.files.size-too-big'),
+			description: `Maximum file size is ${formatBytes(maxAssetSize)}`,
```
ds300: we should use a localized string for this. Add it to main.json as an additional key (not replacing the existing one) and have it be parameterized by the size string.

e.g. `"assets.files.maximum-size": "Maximum file size is {size}"` and then in here do `msg('assets.files.maximum-size').replace('{size}', formatBytes(maxAssetSize))`

kostyafarber: Good shout. Done!

ds300 (approved):
Thanks! Good one!

PR #6752, opened 2025-09-16
feat(templates): make sync template URL copy-pasteable

ds300:
We already have localStorage user management stuff built in so this was all that's needed.

### Change type

- [x] `other`

### Test plan

1. Open the sync-cloudflare template
2. Copy the URL and paste it into a new tab
3. Verify the room state is preserved

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Sync template: make URL copy-pasteable


SomeHats:
@claude can you:

1. upgrade react-router to `^7.9.1` using the declarative mode, make sure `yarn constraints` passes, and run `yarn dedupe`
2. wrap any localstorage accesses in helpers that do a try/catch in case the browser prevents us from accessing it

thx

PR #6755, opened 2025-09-16
feat(editor): allow users to set trackpad/mouse preference for wheel behavior

kostyafarber:
This change introduces a new preference of either `Trackpad` or `Mouse` or no preference (in which case the `wheelBehavior` field of `TLCameraOptions` is used) to control wheel behaviour. 

For example, if selecting `Mouse` and using a mouse peripheral, scrolling with the wheel will zoom in towards the cursor and if holding command/control it pans the screen.

### New
image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added preferences to set wheel behaviour for pointer peripherals

### API changes

- Added a new translation key
- Added a new preference to allow users to control wheel behaviour `getInputMode`

### DIFF
--- a/packages/editor/src/lib/utils/reparenting.ts
+++ b/packages/editor/src/lib/utils/reparenting.ts
@@ -162,7 +162,13 @@ function getOverlappingShapes<T extends TLShape[] | TLShapeId[]>(
 	const parentPageTransform = editor.getShapePageTransform(shape)
 	const parentPageCorners = parentPageTransform.applyToPoints(parentGeometry.vertices)
 
-	const parentPageMaskVertices = editor.getShapeMask(shape)
+	const _shape = editor.getShape(shape)
+	if (!_shape) return EMPTY_ARRAY
+
+	const pageTransform = editor.getShapePageTransform(shape)
+	const clipPath = editor.getShapeUtil(_shape.type).getClipPath?.(_shape)
+
+	const parentPageMaskVertices = clipPath ? pageTransform.applyToPoints(clipPath) : undefined
 	const parentPagePolygon = parentPageMaskVertices
 		? intersectPolygonPolygon(parentPageMaskVertices, parentPageCorners)
 		: parentPageCorners
--- a/packages/tldraw/src/test/frames.test.ts
+++ b/packages/tldraw/src/test/frames.test.ts
@@ -1680,3 +1680,18 @@ it('drops into the top-most frame, if there is one', () => {
 
 	expect(editor.getShape(rect)?.parentId).toBe(editor.getCurrentPageId())
 })
+
+it('does not get drop children of nested frame if they are occluded from the outer frame', () => {
+	const frame1Id = dragCreateFrame({ down: [100, 100], move: [300, 300], up: [300, 300] })
+	const frame2Id = dragCreateFrame({ down: [150, 150], move: [290, 290], up: [290, 290] })
+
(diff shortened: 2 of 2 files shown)


ds300:
In terms of UX theres a couple of things I'd change here:

1. 'Mouse or trackpad' is a slightly odd name for this submenu. I'd change it to "Mouse mode".
2. It looks a bit odd when neither option is selected so we could maybe have a separate 'Auto' option for when the preference value is `null/undefined`?.

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1366,6 +1366,60 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					})
 				},
 			},
+			{
+				id: 'toggle-trackpad',
+				label: 'action.toggle-trackpad',
+				onSelect(source) {
```
ds300: I think we should nix (!) these actions and supply the `onSelect` prop to the `TldrawUiMenuActionCheckboxItem` components instead of the `actionId` prop. should be a lot simpler.

mimecuvalo:
> 1. 'Mouse or trackpad' is a slightly odd name for this submenu. I'd change it to "Mouse mode".

"Input method"?

I agree with David but Mouse mode also sounds a bit funny as well in my head. 

Also makes me think of Larry Tesler :P https://arc.net/l/quote/wxcwecki

kostyafarber:
> > 1. 'Mouse or trackpad' is a slightly odd name for this submenu. I'd change it to "Mouse mode".
> 
> "Input method"?
> 
> I agree with David but Mouse mode also sounds a bit funny as well in my head.
> 
> Also makes me think of Larry Tesler :P https://arc.net/l/quote/wxcwecki

Yeah it's a tough one to name! I got my inspiration from Miro for naming

image

Maybe "Pointer peripheral settings"?

mimecuvalo:
> Pointer peripheral settings

😆 hehe too nerdy

kostyafarber:
OK I've gone with "Input mode" and made some refactor changes per the comments :)

packages/tldraw/src/lib/ui/components/InputModeMenu.tsx
```diff
@@ -0,0 +1,65 @@
...
+	])
+
+	const isModeChecked = (mode: string) => {
+		if (mode === 'auto') {
+			return inputMode === null
+		}
+		return inputMode === mode
+	}
+
+	const getLabel = (mode: string, wheelBehavior: 'zoom' | 'pan' | 'none') => {
```
ds300, line 28: if `wheelBehavior` is set to `'none'` will we end up with a invalid label ID? if so we should add a special case in here

ds300 (approved):
Nice! Just one more thing to maybe clean up? Otherwise all good.

packages/tldraw/src/lib/ui/components/InputModeMenu.tsx
kostyafarber: yeah good point! I think the label without parenthesis works in this case? 

```ts
"action.toggle-auto-none": "Auto",
```

If we are happy with that, I'll add it and then merge :)

PR #6768, opened 2025-09-17
Docs: change license page language

sylwiavargas:
Needs Steve's approval.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/docs/content/community/license.mdx
+++ b/apps/docs/content/community/license.mdx
@@ -48,9 +48,9 @@ function App() {
 }
 ```
 
-License keys are validated **on the client**. You can use them offline.
+License keys are validated **on the client**. You can use them offline. They can be public.
 
-It safe for license keys to be public. Business and Hobby license keys will only work on pre-agreed domains. Trial license keys will work on any domain.
+Business and Hobby license keys will only work on pre-agreed domains. Trial license keys will work on any domain.
 
 The tldraw SDK will not work in production without a valid license key.
 
@@ -64,6 +64,6 @@ When using the tldraw SDK in production under a trial license, the tldraw SDK wi
 
 While the tldraw SDK is [source available](https://github.com/tldraw/tldraw), it is not permissively licensed. Many of our examples and demos are available under an MIT licenses, and you should feel free to use this code as if it were your own; however the any source code or packages licensed by the tldraw SDK would not be [Open Source](https://opensource.org/osd) by any definition.
 
-If you wish to include tldraw in an open source project, you may do so but the SDK itself must remain under its original license. This means that you and your down-stream users will require their own trial, commercial, or hobby license in order to use the SDK in production. In this case, please also see our [trademark guidelines](https://staging.tldraw.dev/legal/trademark-guidelines) that limit the use of the tldraw name and branding.
+If you wish to include tldraw in an open source project, you may do so but the SDK itself must remain under its original license. This means that you and your down-stream users will require their own trial, commercial, or hobby license in order to use the SDK in production. In this case, please also see our [trademark guidelines](https://tldraw.dev/legal/trademark-guidelines) that limit the use of the tldraw name and branding.
 
 The tldraw SDK is a product of a company. See [tldraw.dev](https://tldraw.dev/company) for more information about the company and the team.
--- a/apps/docs/content/community/license.mdx
+++ b/apps/docs/content/community/license.mdx
@@ -6,101 +6,148 @@ date: 9/13/2023
 order: 2
 ---
 
-tldraw's source code and distributed packages (the tldraw SDK) are provided under the [tldraw license](/legal/tldraw-license).
+tldraw's source code and distributed packages (the tldraw SDK) are provided under the [tldraw license](/legal/tldraw-license). Our licensing model supports different use cases, from local development to commercial applications and hobby projects.
 
-You can use the tldraw SDK in commercial or non-commercial projects so long as you preserve the "Made with tldraw" watermark. If you purchase a [business license](https://tldraw.dev#pricing), we will provide you with a license key that removes the watermark.
+## License types at a glance
 
-For more information, visit the [pricing page](https://tldraw.dev#pricing).
+- **Default license** - Free for local development with watermark.
+- **Trial license** - 100 days free for production environments.
+- **Commercial license** - Full commercial rights for production environments.
+- **Hobby license** - Free for non-commercial hosted projects with watermark (by application).
 
-<TwoImages
-	left={{
-		src: '/images/with-watermark.png',
-		title: 'With a valid license key, the watermark is removed.',
-		alt: 'tldraw with watermark',
-	}}
-	right={{
-		src: '/images/without-watermark.png',
-		title: 'When a valid license key is provided, the canvas does not display the watermark.',
-		alt: 'tldraw without watermark',
-	}}
-/>
+For more information, visit the [pricing page](/pricing).
+
+---
+
+## How to request your license key
+
+Requesting license key depends on the license type:
+
+- **Default license** - No license needed.
+- **Trial license** - Fill out a Trial License form(LINK) and we will email you a license key.
+- **Commercial license** - Speak with our sales team(LINK) to determine your product needs.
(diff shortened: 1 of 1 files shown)


apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
 ## How to use your license key
 
-To use your license key, provide it to the `Tldraw` component’s `licenseKey` prop.
+License keys control functionality and remove watermarks in production environments. Here's how to integrate them:
+
+### Basic integration
 
 ```tsx
 import { Tldraw } from 'tldraw'
+;<Tldraw
```
mimecuvalo: typo: semicolon in front

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
+	licenseKey={process.env.TLDRAW_LICENSE_KEY}
+	// ... other props
+/>
 ```
 
-<Callout>The [TldrawEditor](?) component also has a `licenseKey` prop.</Callout>
+### Key management best practices
 
-If the license key is valid and the license is active, then the canvas should no longer display the watermark. If you still see a watermark, check the JavaScript console for information about any problems and how to fix them.
+- Store keys as environment variables, not in source code.
```
mimecuvalo: we say the best practice here of storing as env variable, but the first example is storing as source code 🙃 
i think maybe we just leave out the source code example

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
 
-### Active licenses
+### What license keys control
 
-Your license key encodes the **type of license** that you have and the **expiration date** for that license.
+License keys encode several pieces of information:
 
-To be valid, a license key must be associated with an active license. An **annual** **license** is valid if the current date is earlier than the license’s expiration date.
+- **Validity period** - Expiration date for the license.
+- **Feature access** - Watermark display and functionality limits.
```
mimecuvalo: this isn't accurate for how we distribute licenses. it doesn't gate features at the moment (and the watermark display thing is something we only do for our dogfooded license)
we should explicitly say that it's Domain access

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
-### Domain verification
+### Technical enforcement
 
-Your license key also encodes a list of **approved domains** where the keys should be valid.
+The SDK includes built-in validation that:
 
-In production, a license key is only valid when used on an approved domain. If your key is used on an unapproved domain, then the client will consider the key invalid.
+- Verifies license key authenticity and expiration.
+- Detects whether you're running locally or in a production environment.
+- Based on license type, enforces watermark display.
```
mimecuvalo: we should mention about domains/hosts, that's the key thing it checks for besides expiry date

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
+---
+
+## Special conditions and requirements
+
+### Environment detection
+
+The SDK automatically detects your deployment environment:
+
+- **Development environment**: Single computer without network deployment.
+- **Production environment**: Servers, cloud platforms, web applications, or network-accessible deployments.
```
mimecuvalo: network-accessible isn't required. it also works in offline mode and it's worth mentioning that

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
 
-### What happens if my license key is missing / invalid?
+### Watermark requirements
 
-If you don’t provide a license key, or your license key is invalid, then the canvas will display the “Made with tldraw” watermark. If you provide an invalid license key in development, we will log notes about the problem in the JavaScript console.
+- Default and Hobby licenses must display the tldraw watermark.
```
mimecuvalo: "Default" is trial? a bit confusing, makes it sound like the startup "default" license has a watermark

apps/docs/content/community/license.mdx
mimecuvalo: it's mentioned below i see but this seesm to contradict it

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
+- You should not disable, hide, remove, or alter the watermark under Default license.
+- Trial and Commercial licenses remove the watermark requirement.
 
-### Does tldraw make any network calls to validate my license?
+### Domain and usage restrictions
 
-No. The license key check runs on the client and does not rely on any network calls. The watermark image is loaded from cdn.tldraw.com, however this is only loaded if a license key is missing or invalid.
+- License keys may include domain restrictions (check your specific license terms).
```
mimecuvalo: might as well say: all license keys include domain restrictions except for trial ones

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,148 @@ date: 9/13/2023
...
 
-### Can I still use tldraw while offline?
+### Offline usage
 
-Yes. The license key check runs on the client and does not rely on any network calls, so you can use the tldraw SDK for offline apps.
+License does not require network connectivity.
 
-### Can I expose my license key in production?
+- Initial validation happens during SDK initialization.
+- In case of the watermark, a single fetch is sent for the watermark svg but the license would still work if the fetch would fail.
```
mimecuvalo: nit: SVG

mimecuvalo (commented):
some feedback!

apps/docs/content/community/license.mdx
steveruizok:
```suggestion
- In case of the watermark, a single fetch is sent for the watermark SVG but the license would still work if the fetch would fail.
```

apps/docs/content/community/license.mdx
steveruizok: not accurate, trial licenses don't

apps/docs/content/community/license.mdx
steveruizok: default is not trial

apps/docs/content/community/license.mdx
steveruizok: Yeah, re-working this in the license itself

apps/docs/content/community/license.mdx
mimecuvalo: ^ that's what i said, yeah? "except for trial ones"

mimecuvalo:
Why do we call it "Default license"? That was confusing me — why not just "Development license"?
Or, I guess the cognitive dissonance in my head is that the default is sort of the code is unlicensed (there is no license key) — albeit there's a different terms of service/licensing in the sense of legal requirements.

One tweak, we could edit the section where it says "Default license - No license needed." → "Default license - No license key needed." There's the license that we hand out in the form of the key and then there's license as far as our governance and it's getting conflated there heh.

Lookin' good overall though!

sylwiavargas:
@mimecuvalo, thanks for re-read 💜 
Good catch about "No license key needed," corrected it! 

Regarding the question about Default License, initially the name was different but the current phrasing is a result of a discussion on Discord (I believe?). Should we bring your question to Discord to continue?

apps/docs/content/community/license.mdx
```diff
@@ -6,101 +6,64 @@ date: 9/13/2023
...
-
-Yes. The license key check runs on the client so keys must be exposed in your application client. In production, the [domain verification](#Domain-verification) phase of our license key check should ensure that your license key is valid only on your approved domains.
-
-### Should I use my license key in development and testing?
+```tsx
+const myLicenseKey = 'tldraw-***********************'
 
-Yes. We suggest using your license key in development and testing so that the experience matches your production environment. When creating your license key, be sure to provide us with all deployment and preview domains where the license key should be valid.
+function App() {
+	return <Tldraw licenseKey={yourLicenseKey} />
```
TodePond:
```suggestion
	return <Tldraw licenseKey={myLicenseKey} />
```

PR #6771, opened 2025-09-17
new extract draft changelog script

SomeHats:
Adds a new script for extracting draft changelogs. This is based entirely on my personal workflow for writing changelogs, which usually involves checking each PR anyway. So this script includes almost the entire commit message, just cleaning it up a little by removing sections and commits we don't need.

### Change type

- [x] `other`

### DIFF
--- a/internal/scripts/extract-draft-changelog.tsx
+++ b/internal/scripts/extract-draft-changelog.tsx
@@ -1,0 +1,176 @@
+import { join } from 'path'
+import { exec } from './lib/exec'
+import { REPO_ROOT, writeStringFile } from './lib/file'
+import { nicelog } from './lib/nicelog'
+
+async function main() {
+	const args = process.argv.slice(2)
+
+	if (args.length < 1 || args.length > 2) {
+		console.error('Usage: yarn extract-draft-changelog <start-ref> [end-ref]')
+		console.error('Example: yarn extract-draft-changelog v2.0.0 HEAD')
+		console.error('Example: yarn extract-draft-changelog v2.0.0  # uses current HEAD')
+		process.exit(1)
+	}
+
+	const startRef = args[0]
+	const endRef = args[1] || 'HEAD'
+
+	nicelog(`Extracting changelog from ${startRef} to ${endRef}`)
+
+	try {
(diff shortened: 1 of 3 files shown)


internal/scripts/extract-draft-changelog.tsx
```diff
@@ -0,0 +1,176 @@
...
+	const startRef = args[0]
+	const endRef = args[1] || 'HEAD'
+
+	nicelog(`Extracting changelog from ${startRef} to ${endRef}`)
+
+	try {
+		// Get list of commits between the two refs
+		const gitLogOutput = await exec(
+			'git',
+			['log', '--oneline', '--no-merges', `${startRef}..${endRef}`],
```
ds300: i think before doing this you'd need to make sure both refs are available locally. also does this work if there is no linear history between the refs? e.g. if we do a cherry pick or sneaky hotfix to production and then do `git log production..main` will it give a sensible result?

internal/scripts/extract-draft-changelog.tsx
```diff
@@ -0,0 +1,176 @@
...
+				// Clean up excessive newlines
+				.replace(/\n{3,}/g, '\n\n')
+				// remove trailing horizontal rule
+				.trim()
+				.replace(/-{3,}$/, '')
+				.trim()
+
+			// Only include entries that touch packages/ folder
+			const touchesPackages = Array.from(folders).some((folder) => folder.startsWith('packages/'))
+			if (touchesPackages) {
```
ds300: could hoist this check up. Also seems like the folders list is only used for this check, so I think we could do the whole thing simpler like

`if (await exec('git diff-tree {commit} -- packages'))` which should return the empty string if no files inside the packages dir were changed.

ds300:
managed to find a slightly better way to get the list of unapplied commits between two diffs `git cherry <fromRef> <toRef>` which allows us to filter out commits that were already cherry-picked, but thinking about this right now I realise we're probably not going to run into that issue with SDK releases. silly me.

anyway I also tweaked the output of this and handled some edge cases and set it up as a replacement for the auto changelog step in the patch release script. thanks ❤️

PR #6773, opened 2025-09-17
fix(examples): resolve PdfJS worker src resolution

ds300:
I think this is a more robust way to do it. Seemed like vite was struggling with the `new URL` situation

### Change type

- [x] `bugfix` 

### Test plan

1. Open the PDF editor example and verify PDFs load correctly.

- [ ] Unit tests (if present)

### Release notes

- Fixed an issue where PDF worker source resolution failed in some environments.


ds300:
@claude can you open another PR to do the same thing but for the exam marker example?

Issue #6764, opened 2025-09-17
Add option to configure ArrowShapeUtil preciseTimeout behavior


kristoffer-adam-tronerud-adsk:
### What's the feature?

When using the arrow tool, you can hover above the target for short while to trigger `isPrecise` behaviour. This get you into a state where you can precisely place your endpoints: 

https://github.com/user-attachments/assets/9ae9dec3-887c-40c8-8dab-2a11a55bf922

However, I'd like to make the timeout to trigger `isExact` behaviour where the endpoint is rendered at cursor on top of the target. I have managed to hackishly get this to work by configuring `shouldBeExact` to use exactly the same timeout as the preciseTimeout. But, this feels brittle as it relies on the timeout's to be the same as well as duplicating the behaviour. 

https://github.com/user-attachments/assets/14d8954a-9f2a-40ec-a429-510e172e98e0

It would be great to have an option to configure ArrowShapeUtil e.g where the timeout's already in place could make the arrow endpoints use `isExact` snapping instead of `isPrecise`. 

### Where would you like this feature?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

SomeHats:
Hey @kristoffer-adam-tronerud-adsk! This is a great suggestion - we'd happily accept a PR adding this if you feel comfortable making one.

kristoffer-adam-tronerud-adsk:
> Hey [@kristoffer-adam-tronerud-adsk](https://github.com/kristoffer-adam-tronerud-adsk)! This is a great suggestion - we'd happily accept a PR adding this if you feel comfortable making one.

Thanks, have made a suggestion here! https://github.com/tldraw/tldraw/pull/6775

SomeHats:
Fixed in #6781

PR #6775, opened 2025-09-17
feat: Add option 'preciseToExact' which overrides precise movements to exact

kristoffer-adam-tronerud-adsk:
- Suggestion implementation of https://github.com/tldraw/tldraw/issues/6764

> When using the arrow tool, you can hover above the target for short while to trigger isPrecise behaviour. This get you into a state where you can precisely place your endpoints...

> ...It would be great to have an option to configure ArrowShapeUtil e.g where the timeout's already in place could make the arrow endpoints use isExact snapping instead of isPrecise.

This PR introduces a boolean option `preciseToExact` that makes the arrows have exact behaviour instead of precise after the hover-timeout:

https://github.com/user-attachments/assets/c84a5eb8-0445-4ba9-96cc-84d23aaf84e5

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. Configure the arrow util: `ArrowShapeUtil.configure({ preciseToExact: true })`
2. Create any shape
3. Create an arrow targeting the shape, observe that the arrow uses exact instead of precise arrows on hover
4. Create an arrow originating from the shape, observe that when hovering, the origin uses exact arrows instead of precise arrows

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Support configuring if arrows should be exact on hover instead of precise

### API changes
- Added option `preciseToExact` in ArrowShapeUtil

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -229,6 +229,7 @@ export interface ArrowShapeOptions {
     readonly minElbowHandleDistance: number;
     readonly minElbowLegLength: Record<TLDefaultSizeStyle, number>;
     readonly pointingPreciseTimeout: number;
+    readonly preciseToExact: boolean;
     readonly shouldBeExact: (editor: Editor) => boolean;
     readonly shouldIgnoreTargets: (editor: Editor) => boolean;
 }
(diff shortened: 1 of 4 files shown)


SomeHats:
Hey @kristoffer-adam-tronerud-adsk! Thanks for making this. Seeing how this looked in practice got me thinking a bit and I think I'd rather we solve this in a slightly different way. So I've put up an alternative take here, which includes an example: https://github.com/tldraw/tldraw/pull/6781

kristoffer-adam-tronerud-adsk:
> Hey @kristoffer-adam-tronerud-adsk! Thanks for making this. Seeing how this looked in practice got me thinking a bit and I think I'd rather we solve this in a slightly different way. So I've put up an alternative take here, which includes an example: #6781

@SomeHats Great, thanks! We can close this one then. Will https://github.com/tldraw/tldraw/pull/6781 be available in the next release, when can we expect it to be available?

SomeHats:
yes, it'll be in the next release! which should be... some time today, I hope!

PR #6787, opened 2025-09-18
Use pointingPreciseTimeout in DraggingHandle

### LINKED ISSUE #6767: ArrowShapeUtil pointingPreciseTimeout is not applied
### What happened?

This is related to https://github.com/tldraw/tldraw/issues/6764 and https://github.com/tldraw/tldraw/issues/6765 where we'd like to change how the arrow snapping looks and feels. 
---

My understanding is that `hoverPreciseTimeout` and `pointingPreciseTimeout` should change the amount of time needed to hover the start and endpoints of the target shape before `isPrecise` is applied.   

However, I've tried to configure ArrowShapeUtil's `pointingPreciseTimeout` doesn't work

Here's an example of defaults being used which to my understanding is 
```		
hoverPreciseTimeout: 600,
pointingPreciseTimeout: 320,
```
https://github.com/user-attachments/assets/b7912eab-b750-41ab-99f7-e896826064a9

But, when applying 3 seconds timeout to both, the pointingPreciseTimeout doesn't seem to work:

https://github.com/user-attachments/assets/60a2a982-341b-4222-ab23-76780478cca5

Configure pointingPreciseTimeout to have no delay, or a long delay, without it seeming to have an effect. 
```
ArrowShapeUtil.configure({
    pointingPreciseTimeout: 0, // this doesn't seem to work
})

ArrowShapeUtil.configure({
    pointingPreciseTimeout: 3000, // this doesn't seem to work
})
```

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

ds300:
This fixes #6767 – we still have some hard-coded arrow logic in the select tool which was not using the arrow shape util's configuration values around precision interactions.

### Change type

- [x] `bugfix`

### Test plan

1. Create a shape...
2.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where the Arrow tool's `pointingPreciseTimeout` option was not being respected.

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
@@ -139,6 +139,9 @@ export class DraggingHandle extends StateNode {
 
 	// Only relevant to arrows
 	private resetExactTimeout() {
+		const arrowUtil = this.editor.getShapeUtil<any>('arrow')
+		const timeoutValue = (arrowUtil.options as any)?.pointingPreciseTimeout ?? 750
+
 		if (this.exactTimeout !== -1) {
 			this.clearExactTimeout()
 		}
@@ -150,7 +153,7 @@ export class DraggingHandle extends StateNode {
 				this.update()
 			}
 			this.exactTimeout = -1
-		}, 750)
+		}, timeoutValue)
 	}
 
 	// Only relevant to arrows
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
@@ -13,6 +13,7 @@ import {
 	sortByIndex,
 	structuredClone,
 } from '@tldraw/editor'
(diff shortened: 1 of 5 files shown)


packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
```diff
@@ -139,6 +139,9 @@ export class DraggingHandle extends StateNode {
 
 	// Only relevant to arrows
 	private resetExactTimeout() {
+		const arrowUtil = this.editor.getShapeUtil<any>('arrow')
+		const timeoutValue = (arrowUtil.options as any)?.pointingPreciseTimeout ?? 750
```
SomeHats:
```suggestion
		const arrowUtil = this.editor.getShapeUtil<ArrowShapeUtil>('arrow')
		const timeoutValue = arrowUtil.options.pointingPreciseTimeout
```

SomeHats (approved):

packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
ds300: ah right, for some reason I though this was in the editor package

PR #6790, opened 2025-09-18
fix(ui): wrap style panel labels for long translations

mimecuvalo:
followup to https://github.com/tldraw/tldraw/pull/6783
we just need to have this wrap in general, it barely works in English as it is, let alone other languages.

before:
Screenshot 2025-09-18 at 10 35 35

after:
Screenshot 2025-09-18 at 10 41 47

also, drive-by fix for buttons losing their margin when the tooltip is hovered
Screenshot 2025-09-18 at 10 41 52

### Change type

- [x] `bugfix`

### Test plan

1. Open the style panel.
2. Verify that labels wrap correctly (especially in non-English languages).
3. Verify that buttons maintain their margin when tooltips are hovered.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where style panel labels would not wrap for long translations.

### DIFF
--- a/packages/tldraw/src/lib/ui.css
+++ b/packages/tldraw/src/lib/ui.css
@@ -1039,6 +1039,11 @@ tldraw? probably.
 	display: none;
 }
 
+/*
+ * This is used in a couple places, like Align and Vertical Align.
+ * It's because we have a toolbar with a Toggle Group but then an adjacent button
+ * next to it that opens a popup.
+ */
 .tlui-style-panel__section .tlui-toolbar:has(.tlui-toolbar) {
 	flex-wrap: wrap;
 }


packages/tldraw/src/lib/ui.css
```diff
@@ -1039,10 +1039,14 @@ tldraw? probably.
 	display: none;
 }
 
-.tlui-style-panel[data-ismobile='true'] .tlui-style-panel__section .tlui-row.tlui-toolbar {
+.tlui-style-panel__section .tlui-toolbar:has(.tlui-toolbar) {
```
SomeHats, line 1047: this "thing contains itself" selector is a little confusing and could maybe use a comment

SomeHats (approved):

PR #6791, opened 2025-09-18
docs: automatic rewrites based on sitemap

MitjaBezensek:
This is is the first step in allowing easier adding of new pages via framer. We only need to allow a redeploy of the docs app via a button to press to make this fully work.

### Change type

- [x] `improvement` 

### Test plan

1. Run `yarn dev` in the docs app then go to one of the framer urls (like http://localhost:3001/careers).

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fetch sitemap from framer and generate rewrites for the sites that are found.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,7 +34,8 @@ This is the tldraw monorepo - an infinite canvas SDK for React applications. It'
 - `yarn format` - Format code with Prettier
 - `yarn api-check` - Validate public API consistency
 
-IMPORTANT: NEVER run bare `tsc` - always use `yarn typecheck`
+IMPORTANT: NEVER run bare `tsc` - always use `yarn typecheck`.
+If the `typecheck` command is not found, it's because you're not running it from the root of the repo.
 
 ## Architecture Overview
 
--- a/apps/bemo-worker/package.json
+++ b/apps/bemo-worker/package.json
@@ -31,12 +31,12 @@
 		"react-dom": "^18.3.1"
 	},
 	"devDependencies": {
-		"@cloudflare/workers-types": "^4.20250703.0",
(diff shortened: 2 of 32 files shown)


apps/docs/next.config.js
```diff
@@ -3,6 +3,40 @@
...
+// Function to fetch Framer sitemap and generate dynamic rewrites
+async function getFramerRewrites() {
+	try {
+		const response = await fetch(`https://${REWRITE_DOMAIN}/sitemap.xml`)
+		if (!response.ok) {
+			throw new Error(`Failed to fetch Framer sitemap for rewrites: ${response.statusText}`)
+		}
+
+		const xmlText = await response.text()
+		const urlRegex = /<url>\s*<loc>(.*?)<\/loc>/g
```
mimecuvalo, line 15: do we need the `<url>`?

mimecuvalo: i see in https://github.com/tldraw/tldraw/pull/6789/files that it has a different regex is why i ask, prbly good to keep them consistent

MitjaBezensek: Unified the regexes, I guess it's a bit safer with the url 🤷‍♂️

MitjaBezensek:
> so, if you add a new page in Framer and hit Publish, does a person have to then manually go deploy the docs site? or would it kick off a deploy automatically somehow, you think?

Yeah, they'd need to redeploy. I think re-running the action that last deployed it would work. Don't think framer has webhooks for actions like publishing the site.

MitjaBezensek:
I'm actually not 100% if we should proceed with this. If you choose a slug that already exists you can easily accidentally shadow an existing docs site, so some supervision might be safer?

PR #6801, opened 2025-09-18
Only deploy dotcom previews with a label?

ds300:
How do people feel about not having dotcom preview deploys unless you explicitly opt in? It feels like they're only needed on <5% of PRs right now. If this becomes annoying we can always revert, or maybe set up a thing where it checks whether files inside dotcom or dotcom-shared have been edited? 

This might lead to situations where PRs get merged that break dotcom deployment to staging, but we get discord messages about that and it's not too hard to debug. Also I can't remember the dotcom deployment ever breaking for reasons other than Vercel/cloudflare api flake.

### Change type

- [x] `other`

### DIFF
--- a/.github/workflows/deploy-dotcom.yml
+++ b/.github/workflows/deploy-dotcom.yml
@@ -2,6 +2,7 @@ name: Deploy .com
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


mimecuvalo:
🚢 🚢 🚢

PR #6819, opened 2025-09-19
improvement(tldraw): allow embeds to opt-out of paste-to-embed behavior

kostyafarber:
We should make the default for pasted tldraw links bookmarks instead of an embedded iframe. This change adds a parameter to `EmbedDefinition` called `skipOnPaste` in which we can control this behaviour. 

### Change type

- [x] `improvement`

### Test plan

1. Copy a tldraw URL.
2. Paste it into the editor.
3. Verify it creates a bookmark instead of an embed.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Tldraw links default to pasting as bookmarks rather than embeds

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -619,6 +619,7 @@ export function DebugFlags(): JSX_2.Element | null;
 // @public (undocumented)
 export const DEFAULT_EMBED_DEFINITIONS: readonly [{
     readonly doesResize: true;
+    readonly embedOnPaste: false;
     readonly fromEmbedUrl: (url: string) => string | undefined;
     readonly height: 500;
     readonly hostnames: readonly ["beta.tldraw.com", "tldraw.com", "localhost:3000"];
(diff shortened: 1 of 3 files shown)


mimecuvalo:
this might be a @steveruizok question. i think it's been a part of our cool demos to be able to have infinitely-nested embedded tldraw's and show how neat that is :P

what's the angle coming from that prompted the change? was there some community asks for it or did it feel like a bug maybe.

kostyafarber:
> this might be a @steveruizok question. i think it's been a part of our cool demos to be able to have infinitely-nested embedded tldraw's and show how neat that is :P
> 
> what's the angle coming from that prompted the change? was there some community asks for it or did it feel like a bug maybe.

@steveruizok raised this one, so I'll leave it to him to answer! I will add you can still do this if you use the "Insert embed" option in the preferences menu, this is only affects when you try paste from the clipboard.

steveruizok:
I think the "paste as bookmark" is a better default. We have the right click option to switch between bookmark and embeds.

kostyafarber:
> This seems good to me. I think I'd prefer to flip the logic and have it named `embedOnPaste` with a default value of `true`?
> 
> Or maybe we take steve's suggestion and have it be opt-in? (tbh I'm not sure about that since it's a backwards-incompatible change and as far as I'm aware nobody has asked for this feature before so probably people are more likely to expect the embeds to be pasteable by default)

Good shout! I'll flip it the boolean around and then put the PR in the queue 👍🏻

kostyafarber:
@claude review the PR again please

ds300:
@kostyafarber the way you did the switch to `embedOnPaste` made it a breaking change, i.e. now anyone who has custom embed definitions will have to add `embedOnPaste: true` if they want to preserve the existing behavior.

You should either add a clear callout to the PR description explaining the breaking change and how SDK users should address it, or you could make the change backwards compatible by making `embedOnPaste` true by default, which could be done by e.g.

```diff
- if (embedInfo && embedInfo.definition.embedOnPaste) {
+ if (embedInfo && embedInfo.definition.embedOnPaste !== false) {
```

I don't mind either way, both are fine.

kostyafarber:
Good shout, I'll make it true by default, I'll open another PR.

PR #6826, opened 2025-09-23
fix: use correct translation key for too many files

MitjaBezensek:
Looks like we were using the incorrect label. Not sure why our type system didn't catch this?

### Change type

- [x] `bugfix` 

### Test plan

1. Attempt to upload more than the allowed number of files.
2. Verify the error message displays correctly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a translation error when uploading too many files.

### DIFF
--- a/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
+++ b/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
@@ -382,7 +382,7 @@ export async function defaultHandleExternalFileContent(
 ) {
 	const { acceptedImageMimeTypes = DEFAULT_SUPPORTED_IMAGE_TYPES, toasts, msg } = options
 	if (files.length > editor.options.maxFilesAtOnce) {
-		toasts.addToast({ title: msg('assets.files.amount-too-big'), severity: 'error' })
+		toasts.addToast({ title: msg('assets.files.amount-too-many'), severity: 'error' })
 		return
 	}
 


packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -382,7 +382,7 @@ export async function defaultHandleExternalFileContent(
 ) {
 	const { acceptedImageMimeTypes = DEFAULT_SUPPORTED_IMAGE_TYPES, toasts, msg } = options
 	if (files.length > editor.options.maxFilesAtOnce) {
-		toasts.addToast({ title: msg('assets.files.amount-too-big'), severity: 'error' })
+		toasts.addToast({ title: msg('assets.files.amount-too-many'), severity: 'error' })
```
mimecuvalo, line 385: looks like our type on `msg` is `function msg(id?: Exclude<string, TLUiTranslationKey> | string) {`
which is prbly because people can bring different translation packs to tldraw.

but it does feel like this used to work 🤔 would be interesting to bisect back to a year ago and see if this would have thrown an error then

MitjaBezensek: I'll go ahead and merge this. We can take a look at this alongside our work on better types.

MitjaBezensek: Looks like it has been broken for a while:
https://github.com/tldraw/tldraw/pull/2198/files#diff-199b3949bfe48f2dd1e85914c12df72ea8b6d0eae36188b1a894acdbdfd58224R108

PR #6830, opened 2025-09-24
feat: add Numpad Enter support for text shape editing

### LINKED ISSUE #6669: Numpad Enter is ignored when editing text
### What happened?

i should be able to select a text object, hit the Enter key on my Numpad and have the same behavior as if I had hit the regular Enter key.

1. Create a text object
2. Enter some text
3. Click away to confirm the text entry
4. Click on the text object to select it
5. Hit the Enter key on the Numpad

Current behavior:
Nothing happens

Expected behavior:
Just like hitting the regular Enter key, the whole text should be selected and highlighted - ready to be replaced.

Using Comet on macOS 15.6. but happens on Chrome and Firefox too. Haven't tested on Windows.

Chrome

### Where did this happen?

The tldraw.com website

### Contact Details

tldraw@shinjipons.com

### Code of Conduct

terrydkim:
Fixes #6669 

### Problem
Numpad Enter is ignored when wanting to edit text.

### Change
- Add NumpadEnter to onKeyup function
- Write tests for Enter, NumpadEnter
### Impact
This PR adds Numpad Enter support for text editing in shapes. Users can now use the Numpad Enter key to edit text in shapes, arrows, and other elements, providing better accessibility and consistency with standard keyboard interactions.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a text shape and add content
2. Exit editing mode
3. Select the text shape
4. Switch to text tool
5. Test regular Enter key - should start editing
6. Test Numpad Enter key - should start editing
7. Verify both keys work consistently

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added Numpad Enter support for text editing in shapes and other elements. Users can now use Numpad Enter to enter text editing mode consistently across all shape types.

### After

https://github.com/user-attachments/assets/8eadb67b-a9bf-4f57-88bf-d6cb8df3f244

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -517,7 +517,8 @@ export class Idle extends StateNode {
 
 	override onKeyUp(info: TLKeyboardEventInfo) {
 		switch (info.code) {
-			case 'Enter': {
+			case 'Enter':
+			case 'NumpadEnter': {
 				// Because Enter onKeyDown can happen outside the canvas (but then focus the canvas potentially),
 				// we need to check if the canvas was initially selecting something before continuing.
 				if (!this.selectedShapesOnKeyDown.length) return
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -516,9 +516,8 @@ export class Idle extends StateNode {
 	}
 
 	override onKeyUp(info: TLKeyboardEventInfo) {
-		switch (info.code) {
-			case 'Enter':
-			case 'NumpadEnter': {
+		switch (info.key) {
+			case 'Enter': {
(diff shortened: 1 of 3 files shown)


terrydkim:

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -517,7 +517,8 @@ export class Idle extends StateNode {
 
 	override onKeyUp(info: TLKeyboardEventInfo) {
 		switch (info.code) {
```
mimecuvalo: the better fix here would be to change this to be `info.key`, please!

as long as you're making this fix, it would be absolutely grand to have the same fix applied in:
`SelectTool/childStates/Crop/children/Idle.ts` as well :)

mimecuvalo (commented):
thanks for tackling this issue! i have a tweak to the logic that will be more robust going forward.

terrydkim:
@mimecuvalo Thanks for the feedback! I've updated the code as you suggested. Please take a look when you can :>

PR #6835, opened 2025-09-26
Add ascii branding console message on dot-com

kostyafarber:
Add a branded ascii art console log for tldraw.com. We experimented with having "tldraw" after the logo but would require a pretty wide console to render it correctly.

image

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


apps/dotcom/client/src/utils/consoleBranding.ts
```diff
@@ -0,0 +1,21 @@
+const TLDRAW_BRANDING_MESSAGE = `
+.*@@@@@@@@@@@@@@@@@@@@@@@@@#-                                                                                           
+%@@@@@@@@@@@@@@@@@@@@@@@@@@@@-                                                                                          
+@@@@@@@@@@@@@%%%@@@@@@@@@@@@@=     
+@@@@@@@@@@@@+:..-%@@@@@@@@@@@=     
+@@@@@@@@@@@@-    #@@@@@@@@@@@=     
+@@@@@@@@@@@@%*+*%@@@@@@@@@@@@=     
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@=      Build with tldraw at tldraw.dev
+@@@@@@@@@@@@*-::+@@@@@@@@@@@@=                
+@@@@@@@@@@@@:    *@@@@@@@@@@@=      Work with tldraw at tldraw.dev/careers
```
MitjaBezensek: If we add the protocol prefix these two links can then be clicked, feels like it's a better experience.

It does make the required console width a bit wider though, so not sure if that's the reason why protocol is omitted?

kostyafarber: Agreed, yeah it would be a much better way to display the links. Not sure, I only removed the "tldraw" ascii art bit. I'll add full link :)

mimecuvalo:
i love this idea :D and it couldn't come from anybody but you @kostyafarber :D

might i make a suggestion? might be interesting to have the inverse of colors (via https://www.asciiart.eu/image-to-ascii ), especially maybe the Arrow one suits us hah:

Code page 437:
Screenshot 2025-09-26 at 14 19 45

arrow :)
Screenshot 2025-09-26 at 14 20 24

black & white:
Screenshot 2025-09-26 at 14 20 49

normal:
Screenshot 2025-09-26 at 14 10 41

kostyafarber:
This is what an inverted example with the arrows could look like @mimecuvalo (I thought it would make sense to move the CTAs in the gap between the characters)

image

Not bad, but I do think the inversion in the other direction looks better imo, but I'm easy either way :) 

However we did generate the one in the commit with AI tools, it seems like that website is much better at creating the logo. 

Let me create some examples with screenshots of how it looks in the console with this website, and we can choose one.

kostyafarber:
We think this looks pretty rad! 

image

PR #6864, opened 2025-10-01
fix e2e tests in newer node versions

SomeHats:
Node 22.18 (and 23+ i think) introduced a change to the default flags that broke our playwright setup. See:

* https://github.com/microsoft/playwright/issues/34868
* https://github.com/nodejs/node/issues/59364

Until this is settled or we come up with a better solution here, this uses `NODE_OPTIONS` to get back to the old behaviour for playwright.

### Change type

- [x] `other`


MitjaBezensek:
Thanks❤️

PR #6868, opened 2025-10-02
Bump up the target rate to 120fps.

MitjaBezensek:
Browser will throttle us on screens that can suport it (raf won't run more often than the device supports).

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…


packages/utils/src/lib/throttle.ts
```diff
@@ -76,7 +90,23 @@ export function fpsThrottle(fn: { (): void; cancel?(): void }): {
...
+	// Store custom FPS getter if provided
+	if (getTargetFps) {
+		customFpsGetters.set(fn, getTargetFps)
+	}
+
 	const throttledFn = () => {
+		// Custom FPS - check timing before queuing
+		if (getTargetFps) {
+			const lastRun = customFpsLastRunTime.get(fn) ?? -Infinity
+			const customTimePerFrame = Math.floor(1000 / getTargetFps()) * 0.9
```
mimecuvalo: nit: might as well make 0.9 a constant

packages/sync-core/src/lib/TLSyncClient.ts
```diff
@@ -561,6 +567,11 @@ export class TLSyncClient<R extends UnknownRecord, S extends Store<R> = Store<R>
 		this.flushPendingPushRequests()
 	}
 
+	/** Get the target FPS for network operations based on presence mode */
+	private getSyncFps(): number {
+		return this.presenceMode?.get() === 'solo' ? SOLO_MODE_FPS : COLLABORATIVE_MODE_FPS
```
mimecuvalo, line 811: nice, much clearer than my version!

packages/utils/src/lib/throttle.ts
```diff
@@ -5,15 +5,25 @@ const isTest = () =>
 	!globalThis.__FORCE_RAF_IN_TESTS__
 
 const fpsQueue: Array<() => void> = []
-const targetFps = 60
-const targetTimePerFrame = Math.floor(1000 / targetFps) * 0.9 // ~15ms - we allow for some variance as browsers aren't that precise.
+const targetFps = 120
+const targetTimePerFrame = Math.floor(1000 / targetFps) * 0.9 // ~7ms - we allow for some variance as browsers aren't that precise.
```
mimecuvalo: actually, is the 0.9 still necessary? i'm forgetting a bit on why this was needed... let's expand the comment if we still need this, we're gonna forget again in the future why we did this 🙃

mimecuvalo (commented):
nice! i'm down with this change.

- nit: can we also pull the change from https://github.com/tldraw/tldraw/pull/6470 that updates `DefaultDebugPanel.tsx‎`?
- does this work as expected on iOS?

packages/utils/src/lib/throttle.ts
MitjaBezensek: Added a comment here.

MitjaBezensek:
> nice! i'm down with this change.
> 
> * nit: can we also pull the change from #6470 that updates `DefaultDebugPanel.tsx‎`?
> * does this work as expected on iOS?

* Pulled the debug panel changes. It felt best to show the max instead of the target as the target is now always the same.
* It does seem to work, but you have to enable safari to use higher refresh rates:
https://www.reddit.com/r/iphone/comments/1gzltgr/guide_how_to_enable_120hz_in_safari/

> Go to settings Apps
Safari
Advanced
Feature flags
Turn off prefer page rendering near 60hz
This was introduced in iOS 18 but a lot of people do not know about it!

mimecuvalo:
> * does this work as expected on iOS?

awesome, thanks for the info. actually to be crisper with my question: does this cause any issues on iOS with the browser trying to go too fast than it can go? (less of a question of whether it can go to 120hz but is it "overclocking" and causing it to freak out? :P)

MitjaBezensek:
> > * does this work as expected on iOS?
> 
> awesome, thanks for the info. actually to be crisper with my question: does this cause any issues on iOS with the browser trying to go too fast than it can go? (less of a question of whether it can go to 120hz but is it "overclocking" and causing it to freak out? :P)

I didn't see any issues while testing 🤷‍♂️

I guess it would work pretty much the same as it did till now: we were setting the target to 60 but some actions were too slow so the fps dropped. Now we set a higher default so more actions will fall into that category. But from my tests the rendering stabilizes at the same fps as before. 

With the default being 60fps in Safari I'd expect many people on iOS won't even know the difference. Would be great if we could play with some older devices, though. Maybe some old androids with high refresh screens.

mimecuvalo:
woo woo! 🎉

PR #6876, opened 2025-10-03
localStorageAtom

steveruizok:
This PR adds a `localStorageAtom` to the state package.

See Jotai's `atomWithStorage`: https://jotai.org/docs/utilities/storage https://github.com/pmndrs/jotai/pull/394.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Adds `localStorageAtom` to @tldraw/state


ds300:
whoops looks like you left the shader template in this PR

ds300:
we should make this reactive across tabs with https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

steveruizok:
> we should make this reactive across tabs with https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

Oh nice!

Cleaned up the accidental stuff from the shader work. Do you think this is the right way to go about this, or should we extend the Atom to have storage in its class? I could imagine a StorageAtom with `persist()` and `restore()` methods, rather than doing everything in the wrapper like we're doing here.

PR #6880, opened 2025-10-05
fix: text paste positioning to prevent shapes from running off screen

naaa760:
### Description

-fixes text paste behavior where text shapes would run off the screen when pasting at cursor position or when pasting into auto-sizing text boxes with long content.

**Before:** Text could extend beyond viewport boundaries, making it partially or completely invisible.

**After:** Text shapes are automatically constrained to fit within screen boundaries with proper wrapping and positioning.

### Change type

- [x] `bugfix`


steveruizok:
Sorry @naaa760, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #6883, opened 2025-10-06
feat(editor): add independent x and y self snapping to shapes

kostyafarber:
This change introduces independent x and y line snapping for handles to points on the same shape. Snapping already exists for individual points but not for x and y lines. 

The motivation for this change came from work on a cubic bezier shape, where snapping control handles to start and end points on the x and y axes aids in creating smooth curve shapes, as seen below. 

2025-10-06 at 13 59 53 - Violet Ptarmigan

Since users may not always want handles to snap to x and y lines, I've introduced a `snapType` enum parameter to the `TLHandle` class that allows to specify snap types `point` current behaviour and `align` new behaviour. If it is not provided then we do not snap.

### Change type

- [x] `improvement`

### Test plan

1. Create a shape with handles (like a line or bezier).
2. Drag a handle and observe it snapping to the x or y axis of other points on the same shape.

- [x] Unit tests (if present)

### Release notes

- Added a `snapType` which allows handles to snap to other snap points on the x and y axis.

### DIFF
--- a/packages/editor/src/lib/editor/managers/SnapManager/HandleSnaps.ts
+++ b/packages/editor/src/lib/editor/managers/SnapManager/HandleSnaps.ts
@@ -246,6 +246,7 @@ export class HandleSnaps {
 	}): SnapData | null {
 		const currentShapeTransform = assertExists(this.editor.getShapePageTransform(currentShapeId))
 		const handleInPageSpace = currentShapeTransform.applyToPoint(handle)
+		// eslint-disable-next-line @typescript-eslint/no-deprecated
 		const snapType = handle.canSnap ? 'point' : handle.snapType
 
 		if (snapType === 'point') {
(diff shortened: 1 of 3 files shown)


packages/tlschema/src/misc/TLHandle.ts
```diff
@@ -25,6 +25,7 @@ export interface TLHandle {
 	label?: string
 	type: TLHandleType
 	canSnap?: boolean
+	canIndependentSnap?: boolean
```
ds300: I wonder whether it makes sense to replace these two booleans with one `snapType` enum to make the api a bit less messy.

i.e. we could deprecate `canSnap`, then add a `snapType?: 'point' | 'align' | 'none'`

? 

then we could log an error if they supply both `canSnap` and `snapType`.

kostyafarber: Yeah agreed it would make the API clearer. I've gone with `point | align` and made it optional which means `none`.

ds300 (approved):
Nice! this looks good to me. Claude is right about adding a unit test or two. 

Not sure if the handle snapping logic already has tests, if not would be good to cover that too.

kostyafarber:
> Nice! this looks good to me. Claude is right about adding a unit test or two.
> 
> Not sure if the handle snapping logic already has tests, if not would be good to cover that too.

Sure, there are no tests for handle snapping logic either, so I'll add that too

PR #6893, opened 2025-10-08
style panel: address zooming in/out wrapping unnecessarily sometimes

mimecuvalo:
open to other suggestions / hacks but this is as best as I can come up with for this predicament.

solves this issue when zoomed at 90% (browser zoom). this also happens on zoom in, btw.
Screenshot 2025-10-08 at 12 25 20

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- style panel: address zooming in/out wrapping unnecessarily sometimes

### DIFF
--- a/packages/tldraw/src/lib/ui.css
+++ b/packages/tldraw/src/lib/ui.css
@@ -993,8 +993,16 @@
 	position: relative;
 	z-index: var(--tl-layer-panels);
 	pointer-events: all;
-	width: 148px;
-	max-width: 148px;
+	/*
+	  This is a fractional width because when zooming in/out with the browser the
+	  width of this panel can become something like 147.995. The browser tries it's
+	  best but since the buttons within are exact size they start wrapping where
+		we don't want them to.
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui.css
```diff
@@ -993,8 +993,16 @@
...
+	/*
+	  This is a fractional width because when zooming in/out with the browser the
+	  width of this panel can become something like 147.995. The browser tries it's
+	  best but since the buttons within are exact size they start wrapping where
+		we don't want them to.
+		We need the flex-wrap on the items btw because if "Enhanced a11y mode" is on,
+	  the labels and buttons need to wrap appropriately.
+	*/
+	width: 148.05px;
+	max-width: 148.05px;
```
ds300, line 1005: a larger project, but maybe using css grid for these instead of flexbox would avoid the issue?

mimecuvalo:
closing in favor of https://github.com/tldraw/tldraw/pull/6920

PR #6894, opened 2025-10-08
fix(bookmark): consolidate logic and expose createBookmarkFromUrl

ds300:
Consolidated the logic we use for pasting and the conversion action, and exposed it as a helper function for our users.

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Added `createBookmarkFromUrl` helper function for creating bookmark shapes more easily.
- Fixed an issue where the convert-to-bookmark action used different logic than pasting.

### DIFF
--- a/packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
@@ -1,20 +1,12 @@
 import {
-	AssetRecordType,
 	BaseBoxShapeUtil,
-	Editor,
 	HTMLContainer,
-	Result,
 	T,
-	TLAssetId,
 	TLBookmarkAsset,
 	TLBookmarkShape,
 	TLBookmarkShapeProps,
-	TLShapePartial,
 	bookmarkShapeMigrations,
 	bookmarkShapeProps,
-	createShapeId,
-	debounce,
-	getHashForString,
 	lerp,
 	tlenv,
 	toDomPrecision,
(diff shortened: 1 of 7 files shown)


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -387,45 +390,41 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+					trackEvent('convert-to-bookmark', { source })
+					const shapes = editor.getSelectedShapes()
 
-						const createList: TLShapePartial[] = []
-						const deleteList: TLShapeId[] = []
+					const markId = editor.markHistoryStoppingPoint('convert shapes to bookmark')
+
+					const creationPromises: Promise<Result<any, any>>[] = []
+
+					await deferAsyncEffects(async () => {
```
mimecuvalo: i'm unfamiliar with `deferAsyncEffects` — we don't seem to use it anywhere? why was it added for this bit of logic?

packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
```diff
@@ -318,3 +318,58 @@ const createBookmarkAssetOnUrlChange = debounce(async (editor: Editor, shape: TL
...
 	})
 }, 500)
+
+/**
+ * Creates a bookmark shape from a URL with unfurled metadata.
+ *
+ * @returns A Result containing the created bookmark shape or an error
+ * @public
+ */
+export async function createBookmarkFromUrl(
```
mimecuvalo: part of me thinks: this makes sense to put this function in this file.
OTOH, it makes `BookmarkShapeUtil.tsx` a dep for `actions.tsx` and `defaultExternalContentHandlers.ts` and downstream users might not want BookmarkShapeUtil to be a dep/valid shape in their remix.

i think this logic goes into the `bookmark` directory but as a separate file.

PR #6895, opened 2025-10-08
Improve mobile breakpoints

TodePond:
This PR makes more things disappear from the top of the screen on dotcom on small mobile screens. This mostly affects logged out users. There are still more improvements we could make but this covers most cases.

Before

https://github.com/user-attachments/assets/0a8d73cc-22b6-4c3f-a2b2-d1c67e0dafe8

After

https://github.com/user-attachments/assets/569e3743-6131-43f9-bff3-81d249b8dbc1

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`


ds300 (approved):
nice

PR #6897, opened 2025-10-08
refactor(embed): remove excalidraw embed definition

kostyafarber:
Excalidraw stopped supporting collaboration inside iframes (pr for this is [here](https://github.com/excalidraw/excalidraw/pull/6646)) which means nothing syncs between the iframe inside the tldraw canvas and your excalidraw room. This removes the usefulness of this particular embed, so lets remove it.

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Removed the excalidraw embed

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
+      pull-requests: write
       issues: write
       id-token: write
 
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -563,6 +563,15 @@ export interface CopyAsOptions extends TLImageExportOptions {
 // @public (undocumented)
 export function CopyMenuItem(): JSX_2.Element;
 
+// @public
+export function createBookmarkFromUrl(editor: Editor, { url, center, }: {
+    center?: {
+        x: number;
+        y: number;
+    };
+    url: string;
+}): Promise<Result<TLBookmarkShape, string>>;
+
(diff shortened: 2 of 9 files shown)


ds300:
sorry this slipped off my radar

I tried this and the bookmark was being rendered at the embed size so it looked kinda big

(the one on the right is how it should look IMO) 

image

I'm pretty sure you could achieve this in a small diff, by overriding the `getGeometry` method for the embed shape util to detect when it can't render the embed, and then return a rect sized like the base bookmark instead, otherwise use the super.getGeometry method (from the BoxShapeUtil class) for normal valid embeds.

kostyafarber:
> sorry this slipped off my radar
> 
> I tried this and the bookmark was being rendered at the embed size so it looked kinda big
> 
> (the one on the right is how it should look IMO)
> 
> image
> I'm pretty sure you could achieve this in a small diff, by overriding the `getGeometry` method for the embed shape util to detect when it can't render the embed, and then return a rect sized like the base bookmark instead, otherwise use the super.getGeometry method (from the BoxShapeUtil class) for normal valid embeds.

No worries! Yeah OK I'll patch that up and send over, it does look a bit weird at that size.

kostyafarber:
@ds300 done

ds300:
@kostyafarber I'm seeing this gray bar up top, any idea where that's coming from?

image

kostyafarber:
> @kostyafarber I'm seeing this gray bar up top, any idea where that's coming from?
> 
> image

yeah the bookmark image placeholder element is still being rendered, I've added a boolean to the `BookmarkShapeComponent` that can toggle whether we want to display any image related content, which in this case is a no. This removes the little gray strip.

PR #6901, opened 2025-10-08
fix(agent): stop prompting when agent is stopped

PidgeyBE:
Fixed a bug in the AI Agent template where stopping the agent was not possible if there are open items on the todo list

### Change type

- [x] `bugfix`

### Test plan

1. Prompt the AI agent
2. make sure there are multiple items in the todo list
3. hit the `stop` button
4. -> agents doesnt stop without the fix in this PR

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug in the AI Agent template where stopping the agent was not possible if there are open items on the todo list


PidgeyBE:

PR #6908, opened 2025-10-10
feat(examples): add bezier curve shape example

kostyafarber:
Added an example that showcases how to create a bezier curve shape, useful for vector editing.

## Demo
2025-10-10 at 9 43 08 - Teal Rattlesnake

### Change type

- [x] `other`

### Test plan

1. Open the cubic-bezier-shape example.
2. Interact with the handles to modify the curve.
3. Verify undo/redo behavior while editing.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added a new bezier curve example

---

> [!NOTE]
> Adds a cubic bezier curve shape example with custom handles, snapping, editing/translation interactions, and undo/redo while editing.
> 
> - **Examples**:
>   - Add cubic bezier curve example in `apps/examples/src/examples/cubic-bezier-shape/` with entry `README.md` and `CubicBezierShapeExample.tsx`.
> - **Shape Util** (`CubicBezierShape.tsx`):
>   - Implement `bezier-curve` shape (`BezierCurveShapeUtil`) with `CubicBezier2d` geometry, SVG rendering, bounds/handle snapping, and resize behavior.
>   - Provide four handles (`start`, `end`, `cp1`, `cp2`), auto-collapse control points near endpoints, and meta-key gestures for handle drag behaviors and curve bending during translate.
>   - Show dashed control lines when selected during edit/drag/translate.
> - **Editor State Overrides** (`CubicBezierShapeExample.tsx`):
>   - Customize select tool states to: collapse cp1/cp2 on meta-click, remain in editing after handle drag, and allow translating while in editing.
>   - Register the custom shape and mount an initial curve at viewport center.
> - **UI/Handles** (`CustomHandles.tsx`):
>   - Custom `Handles` component to display handles for `bezier-curve` during edit/drag/translate and default behavior for other shapes.
> - **Undo/Redo** (`SneakyUndoRedoWhileEditing.tsx`):
>   - Add keybinding hook to perform undo/redo while preserving editing state.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16a4811aa1c3f45139f944a51d9a9eae66792dd4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mimecuvalo:
🤩

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function ControlHandles({ children }: TLHandlesProps) {
+	const editor = useEditor()
+
+	const shouldDisplayHandles = useValue(
+		'shouldDisplayHandles',
+		() => {
+			if (editor.isInAny('select.editing_shape', 'select.dragging_handle')) {
+				const onlySelectedShape = editor.getOnlySelectedShape()
+				return onlySelectedShape && editor.isShapeOfType(onlySelectedShape, 'bezier-curve')
```
ds300: you could maybe hoist this check to the start of the function? i think it would make the rest of this logic easier to follow

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
```diff
@@ -0,0 +1,38 @@
...
+			}
+
+			if (editor.isInAny('select.idle', 'select.pointing_shape')) {
+				const onlySelectedShape = editor.getOnlySelectedShape()
+				return !(onlySelectedShape && editor.isShapeOfType(onlySelectedShape, 'bezier-curve'))
+			}
+
+			if (editor.isInAny('select.idle', 'select.pointing_handle', 'select.pointing_shape')) {
+				return true
+			}
```
ds300: seems like this only covers the 'select.pointing_handle' state since the previous two clauses would catch 'select.idle' and 'select.pointing_shape', which also would make this a bug I think since it would show the handles for the bezier if you're pointing at the handle of a completely different shape like an arrow?

i think that's what's going on here

Kapture 2025-10-10 at 13 06 43

apps/examples/src/examples/cubic-bezier-shape/CubicBezierShapeExample.tsx
```diff
@@ -0,0 +1,421 @@
...
+			},
+		}
+	}
+
+	override getBoundsSnapGeometry(shape: IBezierCurveShape): BoundsSnapGeometry {
+		const snapPoints = [shape.props.start, shape.props.end]
+		return {
+			points: snapPoints,
+		}
+	}
```
ds300: nit
```suggestion
	override getBoundsSnapGeometry(shape: IBezierCurveShape): BoundsSnapGeometry {
		return {
			points: [shape.props.start, shape.props.end],
		}
	}
```

apps/examples/src/examples/cubic-bezier-shape/CubicBezierShapeExample.tsx
```diff
@@ -0,0 +1,421 @@
...
+			},
+		]
+
+		if (Vec.Dist(shape.props.cp1, shape.props.start) < threshold) {
+			handles = handles.filter((handle) => handle.id !== 'cp1')
+		}
+
+		if (Vec.Dist(shape.props.cp2, shape.props.end) < threshold) {
+			handles = handles.filter((handle) => handle.id !== 'cp2')
+		}
```
ds300: might make sense to do these distance calculations in screen space rather than shape space, so that the threshold is about visual distance rather than document distance, and therefore feels the same at any zoom level.

apps/examples/src/examples/cubic-bezier-shape/PenTool/Pen.ts
```diff
@@ -0,0 +1,10 @@
+import { StateNode, TLPointerEventInfo, TLShapeId } from 'tldraw'
+
+export class Pen extends StateNode {
+	static override id = 'pen'
+
+	override onEnter(info: TLPointerEventInfo & { shapeId: TLShapeId }) {
+		this.editor.updateInstanceState({ isToolLocked: false })
+		this.editor.setEditingShape(info.shapeId)
+	}
+}
```
ds300: might be a good opportunity to design and use custom icon(s) for this tool too?

kostyafarber:
@ds300 thanks for the comprehensive review :) I'll cover the points you made, and make progress on these in the meantime.

* Yeah this mirrors what figma does, if you snap a control point to its associated start or end point it "deletes" it so you end up in state in which you are hovering the start/end point
* This interaction could be better, I was emulating figmas approach in which holding command gets you in "Bend Mode". 
image

This has some interesting behaviour depending on the states of the curve:
1. If you hit the curve and drag then it changes the shape of the curve (I think by just moving the control points by some delta)
2. If you hit a start or end point, it's as if you've selected the control point and can start dragging it (useful if you've deleted a control point)

2025-10-10 at 13 54 09 - Turquoise Basilisk
* In terms of undo/redo I was looking into this, [shortcuts are disabled](https://github.com/tldraw/tldraw/blob/ae12c0a5a37beecc2ea4662a6dfea3ca7d590cea/packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts#L151) whilst in editing mode and thus undo/redo (via `cmd+z` anyway), not sure what the best way to get around this is, I could intercept the shortcut and implement some custom behaviour. 
* Hmm, good catch I'll look into the snapping logic and see whats up there.

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
kostyafarber: Not a problem, agreed!

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
kostyafarber: yep definitely a bug! good catch, I'll patch that

apps/examples/src/examples/cubic-bezier-shape/CubicBezierShapeExample.tsx
kostyafarber: agreed!

apps/examples/src/examples/cubic-bezier-shape/PenTool/Pen.ts
kostyafarber: yeah we can, this is a bit of a weird one. I really only needed this state as a transition so I could go back into editing mode after dragging a handle (for which I had to rebind some `PointingHandle` handlers to achieve). 

But if I improve the interactions for `cmd` based clicking ("Bend Mode") it'll probably involve adding another state, so I could create a new tool with some icons yeah if it'll be more feature rich

apps/examples/src/examples/cubic-bezier-shape/CubicBezierShapeExample.tsx
kostyafarber: yes agreed, I'll change that

steveruizok:
Pushing some structure changes here.

- Renaming the IBezierCurveShape to MyBezierCurveShape (we should standardize on this across examples)
- Moving the BezierCurveShapeUtil into its own file
- Removing the pen tool, we can solve this another way

The `onInteractionEnd` pattern is really annoying, one of the biggest "cracks" in the interaction system. I'm going to make it take an optional callback, rather than a string for a tool name, which will prevent the default action. We could deprecate the string in the future. I regret it!

packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
```diff
@@ -248,11 +257,11 @@ export class DraggingHandle extends StateNode {
...
 		const { onInteractionEnd } = this.info
-		if (onInteractionEnd) {
-			// Return to the tool that was active before this one,
-			// whether tool lock is turned on or not!
+		if (typeof onInteractionEnd === 'string') {
+			// Return to the tool that was active before this one, whether tool lock is turned on or not!
 			this.editor.setCurrentTool(onInteractionEnd, { shapeId: this.shapeId })
-			return
+		} else {
+			onInteractionEnd?.()
```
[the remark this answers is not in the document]
steveruizok: Not true, the return is just a little lower down

packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
```diff
@@ -220,10 +222,17 @@ export class DraggingHandle extends StateNode {
...
+			if (typeof onInteractionEnd === 'string') {
+				if (this.editor.getInstanceState().isToolLocked && onInteractionEnd) {
+					// Return to the tool that was active before this one,
+					// but only if tool lock is turned on!
+
+					this.editor.setCurrentTool(onInteractionEnd, { shapeId: this.shapeId })
+				}
+			} else {
+				onInteractionEnd?.()
+			}
```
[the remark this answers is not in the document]
steveruizok: Not true. If `onInteractionEnd` is a function, then we defer to the user's `onInteractionEnd` whether tool locked or not. That callback would need to incorporate tool lock.

apps/examples/src/examples/cubic-bezier-shape/PenTool/Pen.ts
kostyafarber: we removed the tools

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
kostyafarber: fixed

apps/examples/src/examples/cubic-bezier-shape/ControlHandles.tsx
kostyafarber: I've cleaned up this logic

kostyafarber:
@steveruizok @ds300 I've cleaned up the code and restructured a few things and added some functionality:
* fixed the bugs David found 
* you can now bend the curve when holding command and translating
* you can translate in edit mode
* command clicking and dragging from the start and end points work without having to be in pointing handle state

let me know if you find anything else! cheers

kostyafarber:
alright! good to review again, fixed the things from @steveruizok and myself's meeting from last week

cc @ds300 

- [x] undo/redo for handle operations
- [x]  Implement collapse distance in screen space, not page space (this relies on the snapping system, so I've set this example to set `alwaysSnap` to true)
- [x]  Add check to prevent handle collapse to start/end when user holds accelerator key
- [x]  Fix handling of handle visibility during translation (don't show control lines or handles during translate ops)
-  Improve command+curve dragging to match reference implementation **(agreed to do this in a follow up, will make for interesting work)** 
- [x]  Update handle indicator dashed lines to be in screen space (divide dash by zoom)

ds300 (approved):
Noticed a couple of issues with shift+drag on the handles:

1. The control point shift+drag angle-snapping behavior is only working correctly for one of the control points, for the other it seems broken?

https://github.com/user-attachments/assets/5d23f9e1-f704-462d-b026-62635cd3d1bb

2. Shift+drag on the start/end points seems to be broken in horizontal directions. (i think it should constrain the point to move only up/down/left/right or along 45 degree diagonals? like when translating shapes)

I'm not sure if these are fixable in 'user land' right now, I think it might be hard-coded behavior in the dragging_handle tool state? Do you mind looking into it and if it's a tiny fix just take care of it? Otherwise, if it's a tricky one or there are some api design decisions to make, create an issue for it?

> Yeah this mirrors what figma does, if you snap a control point to its associated start or end point it "deletes" it so you end up in state in which you are hovering the start/end point

right but i think it's a bug that when you drag your mouse away and the control point is 'created' again it shows the hover state _under the control point_ ?

 Kapture 2025-10-10 at 13 14 40

1. start dragging a control point
2. drag over it's originating start/end point
3. show the hovering state _for the start/end point_, fine
4. drag away again
6. there should be no hovering state on either the start/end point or the control point

no?

otherwise I think this is looking great. Nice job!

kostyafarber:
@ds300 thanks! 

1. Yeah the snapping behaviour will require an api change I think. Right now we have no control over what an "adjacent handle" is, which is the relative snap point to the dragged handle. See [here](https://github.com/tldraw/tldraw/blob/459f8092e51a14e3adaebace822f39c754ffff98/packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx#L84). The reason this works for one handle and not the other funnily enough is the way Bezier curve points are laid out: start,cp1,cp2,end. The adjacent handle for cp1 is cp2 and the adjacent handle for cp2 is end (which is correct).
2. imo this relates to the above, for start and end points I think its better to have the snap point be where your pointer was when you started dragging, or something similar. I know Glyphs (font editor) uses your previous start point (ghosted position) and figma uses your initial drag point.
2025-10-21 at 12 26 51 - Sapphire Sailfish
 
Id like some API where I could specify the snap point for this. I'll open an issue.

P.S I fixed the weird pointing handle hover state thing.

apps/examples/src/examples/cubic-bezier-shape/SneakyUndoRedoWhileEditing.tsx
```diff
@@ -0,0 +1,30 @@
...
+				const editingShape = editor.getEditingShape()
+				if (!editingShape) return
+
+				if (e.shiftKey) {
+					editor.redo()
+					editor.setEditingShape(editingShape)
+				} else {
+					editor.undo()
+					editor.setEditingShape(editingShape)
+				}
```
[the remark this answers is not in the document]
steveruizok: You could use the editing shape id here, but I think the setEditingShape has a guard too

PR #6917, opened 2025-10-13
fix(editor): disable coalesced events on iOS to fix broken ink

steveruizok:
This PR fixes ink/drawing issues on iOS by disabling coalesced pointer events on that platform. iOS sometimes doesn't have `getCoalescedEvents` available, causing broken drawing behavior. It adds a platform check to only use coalesced events when not on iOS.

### Change type

- [x] `bugfix` 

### Test plan

1. Test drawing/ink tool on iOS devices
2. Verify drawing works smoothly without breaks
3. Ensure drawing still works correctly on other platforms (desktop, Android)

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where drawing with the ink tool could be broken on iOS devices.

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -162,6 +162,9 @@ export function useCanvasEvents() {
 			// For tools that benefit from a higher fidelity of events,
 			// we dispatch the coalesced events.
 			// N.B. Sometimes getCoalescedEvents isn't present on iOS, ugh.
+			// Specifically, in local mode (non-https) mode, iOS does not `useCoalescedEvents`
+			// so it appears like the ink is working locally, when really it's just that `useCoalescedEvents`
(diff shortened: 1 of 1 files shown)


mimecuvalo (approved):
LGTM

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -162,7 +163,10 @@ export function useCanvasEvents() {
 			// we dispatch the coalesced events.
 			// N.B. Sometimes getCoalescedEvents isn't present on iOS, ugh.
```
mimecuvalo, line 164:
```suggestion
			// N.B. Sometimes getCoalescedEvents isn't present on iOS, ugh.
			// Specifically, in local mode (non-https) mode, iOS does not `useCoalescedEvents`
			// so it appears like the ink is working locally, when really it's just that `useCoalescedEvents`
			// is disabled. The intent here is to have `useCoalescedEvents` disabled for iOS.
```

PR #6919, opened 2025-10-13
feat(editor): add onInteractionEnd callback option

kostyafarber:
Simplifies and standardizes the logic for handling onInteractionEnd across SelectTool child states by using early returns and reducing nested conditionals. This improves code readability and ensures consistent behavior when transitioning between tool states (picked up from @steveruizok commit).

Lets ship this change in a separate PR from the https://github.com/tldraw/tldraw/pull/6908.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Add an option to add `onInteractionEnd` callback instead of passing a tool state string

### DIFF
--- a/apps/analytics/src/analytics.tsx
+++ b/apps/analytics/src/analytics.tsx
@@ -163,7 +163,9 @@ export default function Analytics() {
 					className="tl-analytics-button tl-analytics-button-primary"
 					onClick={() => onConsentChanged(true)}
 				>
-					Accept
+					<div className="tl-analytics-button-text-wrapper">
+						<div className="tl-analytics-button-text">Accept all</div>
+					</div>
 				</button>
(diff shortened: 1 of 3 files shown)


steveruizok:
Looks good! Can you add unit tests for this?

kostyafarber:
@steveruizok added tests!

kostyafarber:
Nice! Yeah I ran into this issue with the `onDragEnd` handler for instance, which I thought would be called at the very end of drag operations but there were some cleanup functions that undid some of the behaviour I wanted essentially.

Sure, we can write something up or scope out the design.

Lets ship it!

PR #6962, opened 2025-10-17
fix(analytics): identify users after account creation

steveruizok:
This PR makes two fixes to analytics:
- identifies users after they create an account
- spreads the properties object rather than creating a new property named `properties`

I've had a hell of a time trying to figure out why we couldn't get reliable data on user activity before account creation. It turned out we are not identifying users after the user creates their account, meaning the user's first session cannot be joined to the rest of our data about the identified user.

Separately, we also are not spreading the `properties` into the identification properties object, so instead we have a nice property named `properties` on our event, which is opaque to a lot of searches. Luckily this only seems to have mattered on tldraw.computer, where the analytics' identify method is called. 

### Change type

### Test plan

1. Create a new account on dotcom.
2. Verify in PostHog that the user is identified and properties are spread correctly.

- [ ] Unit tests (if present)

### Release notes

- Fixed an issue where users were not correctly identified in analytics after account creation.

---

> [!NOTE]
> Identify users in PostHog after account creation (respecting cookie consent) and spread `properties` in the identify payload.
> 
> - **Analytics**:
>   - Update `apps/analytics/src/analytics.tsx` to spread `...properties` in `posthog.identify` instead of nesting under `properties`.
> - **App (dotcom)**:
>   - In `apps/dotcom/client/src/tla/app/TldrawApp.ts`, on first user creation (`didCreate`), check consent via `getStoredCookieConsent()` and call `posthog.identify` with `email`, `name`, and `analytics_consent: true`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9897c07824ec236e5d0e162c6a89f902eb09c352. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/analytics/src/analytics.tsx
+++ b/apps/analytics/src/analytics.tsx
@@ -184,7 +184,7 @@ export function identify(userId: string, properties?: { [key: string]: any }) {
 	if (storedHasConsent !== 'opted-in') return
 
 	posthog.identify(userId, {
-		properties,
+		...properties,
 		analytics_consent: true,
 	})
 	ReactGA.set({ userId })
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -39,6 +39,7 @@ import {
 	uniqueId,
 } from '@tldraw/utils'
 import pick from 'lodash.pick'
+import posthog from 'posthog-js'
 import {
 	Atom,
 	Signal,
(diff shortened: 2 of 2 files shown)


steveruizok:
Seems to be more complexity here. I'd be sad if the problem weren't in our code.

steveruizok:
It seems that the in-memory user id is lost when the page reloads after Clerk completes authentication.

PR #7003, opened 2025-10-24
improvement: remove 'noreferrer' from external links

steveruizok:
Updated all external link components and related code to use rel="noopener" instead of rel="noopener noreferrer". Adjusted ESLint configuration to allow this change. This improves referrer tracking while maintaining security against window.opener vulnerabilities.

- Canvas items (e.g. Bookmark Shape, any items with links) still use `noreferrer`
- All other items (e.g. links to tldraw.dev, watermark, etc) use only `noopenner`

For reference, noreferrer = noopenner + strip referrer info:

> The noreferrer keyword for the [rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel) attribute ... instructs the browser, when navigating to the target resource, to omit the [Referer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Referer) header and otherwise leak no referrer information — and additionally to behave as if the [noopener](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener) keyword were also specified. [Source MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noreferrer)

### Change type

- [x] `improvement`

### Test plan

1. Open the application and click on various external links (e.g., documentation, social links, watermark).
2. Verify that the links open in a new tab.
3. Inspect the link elements to ensure `rel="noopener"` is present and `noreferrer` is absent where expected.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Updated external links to use `noopener` instead of `noreferrer` to improve analytics tracking while maintaining security.

---

> [!NOTE]
> Standardizes external links to use noopener, adds optional allowReferrer across link utilities and runtime, removes legacy openUrl helpers, and updates ESLint and API signatures.
> 
> - **Link handling**
>   - Replace `rel="noreferrer"` with `rel="noopener"` across UI components (`apps/analytics`, `apps/docs`, `apps/examples`, `apps/dotcom`, `internal/dev-tools`).
>   - Fix invalid attributes and URLs (e.g., `target`/`rel` combos; `twitter` → `x.com`).
> - **Runtime/API**
>   - Extend `runtime.openWindow(url, target, allowReferrer?)` and `openWindow(...)` wrapper to conditionally include `noreferrer`.
>   - Update usage sites to `runtime.openWindow`/`window.open` with appropriate features; remove `utils/url.openUrl`.
>   - Update API report to reflect new signatures.
> - **Dotcom client**
>   - `useOpenUrlAndTrack` now accepts `{ allowReferrer }` and uses `window.open(..., 'noopener[ noreferrer]')`.
>   - Propagate `allowReferrer` to callers (e.g., menu items, sidebar link); add `allowReferrer` prop to `ExternalLink`.
> - **VS Code**
>   - Switch to `runtime.openWindow` in menu links; adapt `setRuntimeOverrides(openWindow)` to new signature.
> - **Editor/UI**
>   - Use `runtime.openWindow` in watermark and label/link actions; open embed links without referrer by default.
> - **ESLint**
>   - Configure `react/jsx-no-target-blank` to `['error', { allowReferrer: true }]`; re-enable strict rule in `packages/*`.
> - **Tests/fixtures**
>   - Update rich text snapshot `rel` to drop `noreferrer` (keep `noopener nofollow`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71a5ec02bfa33a24692995eca855f73408728971. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/examples/e2e/tests/export-snapshots-data.tsx
```diff
@@ -74,7 +74,7 @@ export const richText = {
 											attrs: {
 												href: 'https://tldraw.dev',
 												target: '_blank',
-												rel: 'noopener noreferrer nofollow',
+												rel: 'noopener',
```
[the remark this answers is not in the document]
steveruizok: @claude fix, should restore nofollow

apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
```diff
@@ -59,7 +59,7 @@ export function TlaMenuControlInfoTooltip({
 					<a
 						onClick={onClick}
 						href={href}
-						target="_blank nofollow noreferrer"
+						target="_blank noopener"
```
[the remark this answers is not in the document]
steveruizok: @claude fix

steveruizok:
> we need to keep `noreferrer` across the SDK and keep the eslint rule, at least for the SDK, b/c you can leak the secret/unguessable page URL to 3rd-party domains.

Ah good shout for shared rooms, I forgot that the referrer would include the full (file) URL not just the host name.

This PR keeps noreferrer for UCG, like bookmark shapes on the canvas and shape links. So the only links where we'd be including the referrer would be for our own sites and GitHub and Notion, which we link to from the menus. I could restore noreferrer to those.

We could create an eslint rule scoped to the SDK packages?

mimecuvalo:
> This PR keeps noreferrer for UCG, like bookmark shapes on the canvas and shape links.

mm, it does? this was my main concern. example: in `BookmarkShapeUtil.tsx`, `noreferrer` was nixed. intentional?

> So the only links where we'd be including the referrer would be for our own sites and GitHub and Notion, which we link to from the menus. I could restore noreferrer to those.

yeah, wouldn't hurt for github/notion to not get referrers (although they're inherently more trustable destinations)

> We could create an eslint rule scoped to the SDK packages?

yes, please 🙏

packages/editor/src/lib/utils/runtime.ts
```diff
@@ -5,7 +5,7 @@ export const runtime: {
 	hardReset(): void
 } = {
 	openWindow(url, target) {
-		window.open(url, target, 'noopener noreferrer')
```
mimecuvalo, line 8: specifically this one

packages/tldraw/src/lib/shapes/bookmark/BookmarkShapeUtil.tsx
```diff
@@ -195,7 +195,7 @@ export function BookmarkShapeComponent({
 						className="tl-bookmark__link"
 						href={url || ''}
 						target="_blank"
-						rel="noopener noreferrer"
+						rel="noopener"
```
mimecuvalo: and this one

packages/tldraw/src/lib/ui/components/EmbedDialog.tsx
```diff
@@ -92,7 +92,7 @@ export const EmbedDialog = track(function EmbedDialog({ onClose }: TLUiDialogPro
 									<>
 										<a
 											target="_blank"
-											rel="noopener noreferrer"
+											rel="noopener"
```
mimecuvalo, line 95: this one, technically as well, although limited in scope

steveruizok:
Ok, second pass.

- keep `noopener noreferrer` as the default
- explicitly opt in to passing referrer with just `noopener`

I also stumbled on some interesting behavior with `runtime.openWindow`. It seems like the vscode extension handles new tab opens differently, and so we should prefer to use runtime.openWindow instead of window.open? But this is used very inconsistently.

I'm going to create a PR that just explicitly opts the watermark / tldraw.dev links in to referrer but it feels like there's a refactor to be had with how we're handling links in general.

PR #7007, opened 2025-10-24
feat(sync-core): add stream operation for string append optimization

quasor:
## Summary

Introduces a new `ValueOpType.Stream` operation that optimizes network diffs when strings are only appended to (`nextString.startsWith(prevString)`). This reduces bandwidth usage when streaming LLM generation into shapes by sending only the appended text instead of the full string value.

## Changes

- Added `ValueOpType.Stream` constant and `StreamOp` type
- Modified `diffObject()` and `diffValue()` to detect string streaming scenarios
- Added `Stream` case to `applyObjectDiff()` for applying stream operations
- Bumped `TLSYNC_PROTOCOL_VERSION` from 7 to 8 for backward compatibility
- Added 15 comprehensive tests for string streaming behavior

## Test plan

- [x] All 313 tests passing in sync-core package
- [x] String append detection working for top-level and nested properties
- [x] Stream operations correctly applied and validated (offset checks)
- [x] Protocol version bump ensures no silent failures with old clients

## Example use case

This optimization is particularly beneficial for AI-powered features where LLM responses are streamed token-by-token into text shapes. Instead of sending the entire text content on each update, only the newly generated tokens are transmitted.

### Change type

- [x] improvement

### Release notes

Add stream operation for string append optimization in sync-core.

### API changes
internal only

---

> [!NOTE]
> Adds string append support in diffs/patches with protocol v8 and session capability gating, plus RoomSession refactor and related API/test updates.
> 
> - **Diff/Apply mechanics**:
>   - Support string appends in `AppendOp` (`value: string | unknown[]`) and handle in `applyObjectDiff`.
>   - Extend `diffRecord`/internals to detect string appends; add `legacyAppendMode` to fall back to `put`.
> - **Protocol & compatibility**:
>   - Bump `TLSYNC_PROTOCOL_VERSION` to `8`.
>   - Handshake sets `supportsStringAppend` false for v7 clients; introduce `getCanEmitStringAppend()` to decide legacy mode.
> - **Room/session**:
>   - Refactor `RoomSession` into union over `RoomSessionBase`; add `supportsStringAppend` field.
>   - `TLSyncRoom` threads `legacyAppendMode` into `replaceState`/`mergeDiff`; stores and propagates `supportsStringAppend` across lifecycle.
> - **API surface**:
>   - Export `RoomSessionBase`; update types for `AppendOp`, `diffRecord`, `DocumentState.mergeDiff/replaceState`, and `RoomSession`.
> - **Tests**:
>   - Add/adjust tests for string append behavior, session lifecycle, and protocol v8; update expectations to use `append` for strings.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4b6d5143494fc3043a879cdce14bccccaa90691b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sync-core/api-report.api.md
+++ b/packages/sync-core/api-report.api.md
@@ -147,15 +147,15 @@ export type RoomSession<R extends UnknownRecord, Meta> = (RoomSessionBase<R, Met
 });
 
 // @internal
-export type RoomSessionBase<R extends UnknownRecord, Meta> = {
+export interface RoomSessionBase<R extends UnknownRecord, Meta> {
+    isReadonly: boolean;
     meta: Meta;
     presenceId: null | string;
+    requiresLegacyRejection: boolean;
     sessionId: string;
     socket: TLRoomSocket<R>;
-    isReadonly: boolean;
-    requiresLegacyRejection: boolean;
     supportsStringAppend: boolean;
-};
+}
 
 // @internal
 export const RoomSessionState: {
--- a/packages/sync-core/src/lib/RoomSession.ts
+++ b/packages/sync-core/src/lib/RoomSession.ts
@@ -69,7 +69,7 @@ export const SESSION_IDLE_TIMEOUT = 20000
  *
  * @internal
  */
-export type RoomSessionBase<R extends UnknownRecord, Meta> = {
+export interface RoomSessionBase<R extends UnknownRecord, Meta> {
 	/** Unique identifier for this session */
 	sessionId: string
 	/** Presence identifier for live cursor/selection tracking, if available */
--- a/packages/sync-core/src/lib/TLSyncRoom.ts
+++ b/packages/sync-core/src/lib/TLSyncRoom.ts
@@ -178,6 +178,7 @@ export class DocumentState<R extends UnknownRecord> {
 	 *
 	 * @param diff - The object diff to apply
 	 * @param clock - The new clock value
+	 * @param legacyAppendMode - If true, string append operations will be converted to Put operations
 	 * @returns Result containing the final diff and new DocumentState, or null if no changes, or validation error
 	 */
 	mergeDiff(


quasor:

quasor:
Thoughts on naming, and a new op vs extending Append to handle strings? Happy to adjust.

mimecuvalo:
cc @TodePond and @ds300 for your radar

ds300:
Thanks for getting this ball rolling! I have been wondering when someone would ask for this :)

> Thoughts on naming, and a new op vs extending Append to handle strings? Happy to adjust

Very keen to see what it would look like to extend the existing Append operation. Should be a significantly smaller diff?

Also I think we can and should add backwards compatibility for older clients here. I'm happy to do that before merging.

quasor:
@ds300 Combining with array append does tighten it up a bit. Lmk what you think, and if I can help move it forward anymore or if you'd like to take it from here?

packages/sync-core/src/lib/diff.ts
```diff
@@ -213,6 +221,15 @@ function diffObject(prev: object, next: object, nestedKeys?: Set<string>): Objec
...
+			} else if (typeof prevVal === 'string' && typeof nextVal === 'string') {
+				if (nextVal.startsWith(prevVal)) {
+					const appendedText = nextVal.slice(prevVal.length)
+					if (!result) result = {}
+					result[key] = [ValueOpType.Append, appendedText, prevVal.length]
+				} else {
+					if (!result) result = {}
+					result[key] = [ValueOpType.Put, nextVal]
+				}
 			} else {
```
ds300: i think we could simplify this a little (probably need to make this change locally and run prettier)
```suggestion
			} else if (typeof prevVal === 'string' && typeof nextVal === 'string' && nextVal.startsWith(prevVal)) {
				const appendedText = nextVal.slice(prevVal.length)
				if (!result) result = {}
				result[key] = [ValueOpType.Append, appendedText, prevVal.length]
			} else {
```

packages/sync-core/src/lib/diff.ts
```diff
@@ -201,7 +201,15 @@ function diffObject(prev: object, next: object, nestedKeys?: Set<string>): Objec
...
-			if (nestedKeys?.has(key) && prevVal && nextVal) {
+			if (
+				nestedKeys?.has(key) &&
+				prevVal &&
+				nextVal &&
+				typeof prevVal === 'object' &&
+				typeof nextVal === 'object' &&
+				!Array.isArray(prevVal) &&
+				!Array.isArray(nextVal)
+			) {
```
ds300: 👍🏼 nice to have these extra sanity checks, thanks

packages/sync-core/src/lib/diff.ts
```diff
@@ -182,7 +182,7 @@ export interface ObjectDiff {
  * @internal
  */
 export function diffRecord(prev: object, next: object): ObjectDiff | null {
-	return diffObject(prev, next, new Set(['props']))
+	return diffObject(prev, next, new Set(['props', 'meta']))
```
quasor: @ds300 fyi, would like to include this. Do not intend to sneak this in.

quasor:
@ds300 I _think_ I'm done. All yours.

packages/sync-core/src/lib/diff.ts
ds300: good one

mimecuvalo:
~@quasor when you get the chance, to pass the tests you'll need to run `yarn build-api` to generate the latest API docs and commit that output.~

sorry ignore this

PR #7008, opened 2025-10-25
feat(ui): add user avatar sync in share panel

Digital39999:
This pull request introduces user avatar support and synchronization in the share panel.
Users can now see live avatars of collaborators in real time when sharing or editing together.

### Change type

- [x] `feature` 
- [x] `api` 

### API changes
- Added `avatar` field to `TLUserPreferences`, `TLPresenceUserInfo`, and `TLInstancePresence`.
- Added `getAvatar()` method to `UserPreferencesManager`.
- Added `AvatarSize` type (`'sm' | 'md' | 'lg' | 'xl'`).
- Added `avatarSize` prop to `PeopleMenu` and `DefaultSharePanel`.
- Added `DefaultSharePanelProps` extending `PeopleMenuProps`.
- Modified `DefaultSharePanel` to accept props for avatar customization.
- Modified `PeopleMenu` and `UserPresenceEditor` with `allowChangeColor` and `allowChangeName` props.

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added user avatar support and synchronization in the share panel.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/SharePanel/PeopleMenu.tsx
+++ b/packages/tldraw/src/lib/ui/components/SharePanel/PeopleMenu.tsx
@@ -47,10 +47,6 @@ export function PeopleMenu({
 
 	if (!userIds.length) return null
 
-	console.log('Current user avatar:', userAvatar)
-	console.log('Current user name:', userName)
-	console.log('Current user color:', userColor)
-
 	return (
 		<_Popover.Root onOpenChange={onOpenChange} open={isOpen}>
 			<_Popover.Trigger dir="ltr" asChild>


packages/tldraw/src/lib/ui/components/SharePanel/UserPresenceEditor.tsx
```diff
@@ -44,8 +54,8 @@ export function UserPresenceEditor() {
 
 	return (
 		<div className="tlui-people-menu__user">
-			<UserPresenceColorPicker />
-			{isEditingName ? (
+			{allowChangeColor && <UserPresenceColorPicker />}
```
ds300: I think it makes sense to have this option but setting it to true results in this

Image

I'd suggest just making the button non-interactive rather than removing it altogether

ds300:
cross-posting feedback from [the discord thread](https://discord.com/channels/859816885297741824/1431282449211723876/1433124197764304917)


> really appreciate the work on this but there's a few minor technical issues here, and a few design issues here that I think we'd need to discuss internally.
> 
> design:
> the avatars look too small at the default size. and if you bump the size up at all the circle seems too big when there's no avatar. i dunno if we need to find a sweet spot where both states look minimally bad, or if we just optimize for whether the collaborators have avatars or not.
> the colors are not visible in the face pile, and the faces are not visible in the people menu dropdown. this is a UX problem because it makes it harder to keep track of who is who.
> 
> technical:
> the extra options (allowChangeColor, allowChangeName) aren't handled well, they mess up the layout of the menu
> the size prop isn't a pattern we usually follow for customization. we tend to push people towards using css overrides.
> 
> So yeah sorry this needs a bit more time in the oven, and i'd like to get design input from other tldraw product folks.
> 
> Probably an easier route to making this a reality on the design side wold be to make the people menu components overrideable like most of our other ui components. I'm wondering why that hasn't happened yet tbh.

steveruizok:
Sorry @Digital39999, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #7014, opened 2025-10-27
fix(dotcom): resolve file operation and sidebar UI issues

MitjaBezensek:
- Fixed file operations including pinning and dragging files between groups.
- Changed how we handle the home group. We used `my-files` as the group id, but we now use the real id.
- Fixed UI issues with the drag-and-drop indicator being clipped at the top of the sidebar.
- Cleaned up the context menu to prevent the home group from appearing twice (both as "My files" and the user's name).

### Change type

### Test plan

1. Drag files between groups in the sidebar and verify the indicator is not clipped.
2. Pin/unpin files and verify the home group doesn't appear twice in the context menu.

- [ ] Unit tests (if present)

### Release notes

- Fixed file operations and UI issues in the sidebar, including drag-and-drop improvements and group ID resolution.

---

> [!NOTE]
> Adds shared flag helpers and consistently resolves "my-files" to the user’s home group for pin/move/delete/drag operations, with minor admin, backend, and UI tweaks.
> 
> - **Core/shared**:
>   - Add `parseFlags` and `userHasFlag` in `packages/dotcom-shared/src/mutators.ts`; adopt across app.
> - **App logic**:
>   - In `TldrawApp`, use `parseFlags` for `getUserFlags` and add `resolveGroupId("my-files"|id)`.
> - **File actions & drag/drop**:
>   - Use `resolveGroupId` for pin/unpin, delete, and pinned checks in `TlaFileMenu`, `TlaSidebarFileLink`, `TlaSidebarRenameInline`.
>   - Translate group IDs (including targets) during drag operations in `useDragTracking` and send normalized IDs to `handleFileDragOperation`.
> - **UI/UX**:
>   - Exclude home group from "Move to" group list to avoid duplicate "My files" entry.
>   - Tweak sidebar content padding to prevent drag indicator clipping.
> - **Admin/E2E/Backend**:
>   - Use `userHasFlag` for groups migration checks in `admin.tsx`, e2e `Database.ts`, and worker `acceptInvite.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 80fe11d6c49d8fda94064a4db807dd889d20bcd5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -367,17 +367,6 @@ export class TldrawApp {
 		return this.userId
 	}
 
-	/**
-	 * Resolves a group ID for backend operations.
-	 * Translates "my-files" to the home group ID (user ID), passes through actual group IDs unchanged.
-	 *
-	 * @param groupId - Either "my-files" or an actual group ID
-	 * @returns The actual group ID to use in backend operations
-	 */
-	resolveGroupId(groupId: string): string {
-		return groupId === 'my-files' ? this.getHomeGroupId() : groupId
-	}
-
 	@computed({ isEqual })
 	getGroupMemberships() {
 		return this.groupMemberships$.get().slice(0).sort(sortByIndex)
--- a/apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
@@ -115,18 +115,17 @@ export function FileItems({
 	const trackEvent = useTldrawAppUiEvents()
 	const copiedMsg = useMsg(messages.copied)
 	const hasAdminRights = useHasFileAdminRights(fileId)
-	const isPinned = useIsFilePinned(fileId, groupId ? app.resolveGroupId(groupId) : '')
+	const isPinned = useIsFilePinned(fileId, groupId ?? '')
 	const isActive = useCurrentFileId() === fileId
 	const hasGroups = useHasFlag('groups_frontend')
 
 	const file = useValue('file', () => app.getFile(fileId), [app, fileId])
 
-	// Get groups data (excluding the home group, which is shown as "My files")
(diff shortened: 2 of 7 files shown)


apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css
```diff
@@ -83,7 +83,7 @@
 	flex-direction: column;
 	flex: 1 1 auto;
 	margin: 0px -12px;
-	padding: 0px 12px 40px 12px;
+	padding: 2px 12px 40px 12px;
```
MitjaBezensek, line 86: Drag indicator was getting cut off for my files.

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -366,6 +367,17 @@ export class TldrawApp {
...
+	 * Resolves a group ID for backend operations.
+	 * Translates "my-files" to the home group ID (user ID), passes through actual group IDs unchanged.
+	 *
+	 * @param groupId - Either "my-files" or an actual group ID
+	 * @returns The actual group ID to use in backend operations
+	 */
+	resolveGroupId(groupId: string): string {
+		return groupId === 'my-files' ? this.getHomeGroupId() : groupId
+	}
+
```
ds300: i think I'd rather just get rid of any instances of 'my-files'? this feels like a flaky pattern to have in the codebase to me, gotta remember to call it all the time

MitjaBezensek: Sure can take a look into that. 

Any special gotchas that you might remember? Just want to avoid hitting the same reason why it was done in this way in the first place.

ds300: no. i only used 'my-files' because the initial spike version of the data model tried to keep the existing data model for the 'home' group. part of the reason I redid the data model to use the groups schema for the 'home' group was to avoid having this extra 'my-files' conditional branching everywhere.

PR #7037, opened 2025-10-29
fix(ui): add file input to DOM for iOS Safari compatibility

kostyafarber:
Apparently, for iOS, an [undocumented](https://stackoverflow.com/questions/47664777/javascript-file-input-onchange-not-working-ios-safari-only) requirement is for `input` elements to be added to the DOM for `onchange` events to fire reliably. 

We don't do this for uploading files locally, and therefore image uploads fail silently and flakily for iOS users using Safari. Lets add the `input` to the DOM. I also made sure that we resolve the promise if we find no files, to avoid hanging promises.

### Change type

- [x] `bugfix` 

### Test plan

1. Open tldraw on iOS Safari
2. Attempt to upload an image
3. Verify the upload succeeds reliably

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with iOS image uploads on Safari

### DIFF
--- a/packages/tldraw/src/lib/ui/getLocalFiles.ts
+++ b/packages/tldraw/src/lib/ui/getLocalFiles.ts
@@ -21,6 +21,7 @@ export function getLocalFiles(options?: {
 			const fileList = (e.target as HTMLInputElement).files
 			if (!fileList || fileList.length === 0) {
 				resolve([])
+				dispose()
 				return
 			}
 			const files = Array.from(fileList)


mimecuvalo:
what a find :) nice sleuthing.

i'm curious: was this always broken you think or did it regress at some point?

kostyafarber:
you mean on our end or WebKit? I think I saw some WebKit bug reports saying they fixed something similar to this but they didn't.

packages/tldraw/src/lib/ui/getLocalFiles.ts
```diff
@@ -9,17 +9,22 @@ export function getLocalFiles(options?: {
...
 		input.multiple = allowMultiple
+		input.style.display = 'none'
 
 		async function onchange(e: Event) {
 			const fileList = (e.target as HTMLInputElement).files
-			if (!fileList || fileList.length === 0) return
+			if (!fileList || fileList.length === 0) {
+				resolve([])
+				return
+			}
```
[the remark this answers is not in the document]
mimecuvalo: yeah, this was gonna be my comment as well, we need to make sure we're doing this now esp. since we're adding it to the DOM.

packages/tldraw/src/lib/ui/getLocalFiles.ts
```diff
@@ -9,17 +9,22 @@ export function getLocalFiles(options?: {
...
 			const files = Array.from(fileList)
 			input.value = ''
 			resolve(files)
 			input.removeEventListener('change', onchange)
 			input.remove()
 		}
 
+		document.body.appendChild(input)
 		input.addEventListener('change', onchange)
```
mimecuvalo, line 40: you'll also need to listen to the `cancel` event so that we clean up things properly.
(prbly have a `dispose` function that we call in both cases to clean up stuff

kostyafarber:
> couple comments!

done! :)

ds300:
all good on my iPhone! nice one

packages/tldraw/src/lib/ui/getLocalFiles.ts
```diff
@@ -9,17 +9,33 @@ export function getLocalFiles(options?: {
...
+			input.removeEventListener('change', onchange)
+			input.removeEventListener('cancel', oncancel)
+			input.remove()
+		}
 
 		async function onchange(e: Event) {
 			const fileList = (e.target as HTMLInputElement).files
-			if (!fileList || fileList.length === 0) return
+			if (!fileList || fileList.length === 0) {
+				resolve([])
```
mimecuvalo, line 23: gotta dispose here too :)

kostyafarber: good catch :)

PR #7058, opened 2025-10-31
fix(dotcom): init user data server-side to prevent race condition

ds300:
Seems like when users signed up there was some kind of race condition where it would try to create a file and the userId foreign key check would fail because the user row hadn't been created yet.

This PR makes it create a user row before even accepting a websocket connection and should bonk the problem on the head.

### Change type

- [x] `bugfix` 

### Test plan

1. Sign up as a new user and ensure the user row is created server-side before the websocket connection is established.

- [ ] Unit tests (if present)

### Release notes

- Fixed a race condition where new user creation could fail during file creation.

---

> [!NOTE]
> Create user/group records on first server contact and remove client-side user creation, updating client bootstrap accordingly.
> 
> - **Backend (sync-worker)**:
>   - Initialize missing `user`, `group`, and `group_user` rows on first `/app/:userId/*` request by fetching profile from Clerk; clear `USER_DO_SNAPSHOTS` and seed defaults (e.g., `color: '___INIT___'`).
>   - Add `getClerkClient` usage; ensure rate limiting and cache init proceed after bootstrap.
> - **Client**:
>   - Remove client-driven user creation flow: `TldrawApp.preload()` no longer accepts `initialUserData`; `TldrawApp.create` no longer requires `fullName/email/avatar`.
>   - After preload, if user has `color === '___INIT___'`, set user preferences and emit `create-user` event; return `userId` from fetched user.
>   - Update hook `useAppState` to match new `TldrawApp.create` signature.
> - **Shared**:
>   - Mark `mutators.init` as deprecated.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 881ca5ed285093f7f3d920de5bbcdd240b612a7d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -66,7 +66,6 @@ import { getScratchPersistenceKey } from '../../utils/scratch-persistence-key'
 import { TLAppUiContextType } from '../utils/app-ui-events'
 import { getDateFormat } from '../utils/dates'
 import { createIntl, defineMessages, setupCreateIntl } from '../utils/i18n'
-import { updateLocalSessionState } from '../utils/local-session-state'
 import { Zero as ZeroPolyfill } from './zero-polyfill'
 
 export const TLDR_FILE_ENDPOINT = `/api/app/tldr`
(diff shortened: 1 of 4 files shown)


ds300:
@claude review this PR please, be super critical

PR #7091, opened 2025-11-09
feat(editor): augmentation-based shape and binding types

Andarist:
continues the work done in https://github.com/tldraw/tldraw/pull/6739

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Tested manually in https://github.com/Andarist/tldraw-augmentation-test 
2. The most complex `TLIndexedShapes` also tested manually in [TS playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251027#code/JYOwLgpgTgZghgYwgAgCoBkBCcDOEDKAFnAA4QA8qAnmQDTIAKUA9iTgHzIDeAUMsmBoQAXGiEBuPshIs2opqxySAvjx6CyadAEEoLAO5FSKALxbseI2XIAiOHub6b9LsgAmwKKJxgooAObIyuySGigYmMwAHlam5rgExNY2AEbRztzIUaIgAK4AtinQ4shUOQVFUEEh6kJaAML2brHIZhEJsbYITRmuCMwANsxeyD5+IIHBoXUYAOIsuSQtbVgdSRQ2-gskvcjAbt6+AdXTmhgAavsQzMvxluu2AG5XzLtgwPkiyHmFxdIDcCoAVEaUGEDgIBOajCWgAIhB4LkBmBllIAD5aXQGWLo8zRHH8DEYRpQZrrXFzbYE5BE9CXNzXWKSHgQKIkYZgPbgaDwJBaWZDFJwAaxBRsACypG4qhZbI5XMgsEQ4XQAuYQpF6zFOElJG4Um6pPKAwGJQA9GbkKyEANcgzkPpgGBCMh4AMcFQDbkfMx8qJXEKAF4BOCHcaTc2W1mQEBuB1Ol1wZD9fIkAYQSADKjJ71gX2jcn8Z4M5j+gQfL4-SolNOA4HIUHpiElYVgVDRsPHZSRq1RGNxhmI5EF4xSLbMRbG03IC17MAAchwyCT-SgKBw63oTuTEPnnKK7mAODgKXTbjNzEe0D8DJAPBlrPZUE5MIwAElY6yIGTjEuzLx+FnAB1FAnWgOBIGQS9oGXZBchAYBmEhZgYGXXJ-E+bk4wAawgKglwhAcETgJFOQ3YwBCEHAADopFnVBCCPZcTUcJdvQEZguRtO0UAACmGZMkI8d4kOFLNe24+0BOgvQrgASgEQgUEHEjh3Isgl1AKCQBQfxmFowCzToy0P0UlAAANcHM5MAW9FB9BQewUBgYBkWgY4J05Dxj1Pb8Rw0gzkAAbQAaS5KRCWQXCqBQ-lBWFUVZB1UgIppOFiNI2IgpsMIbAAXWXJcwujCBYyXDB4SHFF1my3K8tSgB+GdLVXdd1gItdmP0QECMeOBXJPdNl1jHcQBAZh9xQGSb1vUZOIc5AACtcy4217WdCB8lS-hir7Uq3CXTZthsbbkCakLTtEXb+yXaLYrmeLNWMbVdVOprZ2AVDtxKsrYLyE0jxdPrbRQDa4LwKpiCXbzBr850mPU0G6hwebnNcxU5yg3JOUdZ1vggK8oFO-gHvVBKtSS3VQoKn6Du+JEBjS+DB1Ab9ibO-HCfZq7LuQC7+H4K68qu3sbuQI6Jx2RrmsEzrEY6xyBm6-Dlz6gbfOGuNujGiaGymwmZtKuaHRQZafFWnizK2gWBYAUT7KBEDASh0Eq1TquMFxKLIEXglSkXafK12MuRLKcqEfLpY+r6FwI9wQ7I9ZvfslAdLhziECUhBsL2GPFz1o3pv2W9Tuu-bbrw+7VUexLFFem3+Hey1PsxqGC+Qg2rkhDbITssycx9fJ-OUhFQCdRC7wbgX2nuYxyBC+hSY1WuJVIan2F56PW9wb5daKQvO9mnvkDXPvWSPd4JmT5jgASJcYAEoowAxmFvNrT0p-4e3fCdl23cyzcmQwi+w3jbUQs4+5gwQLmfMiN44uQQiJSeDcZ6JDngvOKZMnpkBemvMKAAyKKldUJL3Js9SmpA8o1G7GoWca4UYDCvHGd8n4ojfliDgYQPALTcMtK+dAH4GRsJ-BpVo+pZw22gYPUQqDOg2CkXmfIuxZwKNgeSCRn8gwhk7BMSQGj+BTF4Q3Ys1wZGrFnskExrwvazgAFS02-LY5AVijGf3LJ8covwoB6MtG49+9ZGzghAD4txzE2wdlGEcXRriBaGP0cuBw+gzFYkcEyGWrUEkGBifwNI2Q8QxHWD2DJuSjGqFnLOWEAB5W2+AABy85UDIGIFeZA9ioiSUcTuUkPAgA) (previous iteration can be found in: [TS playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251027#code/JYOwLgpgTgZghgYwgAgCoBkBCcDOEDKAFnAA4QA8qAnmQDTIAKUA9iTgHzIDeAUMsmBoQAXGiEBuPshIs2opqxySAvjx6CyadAEEoLAO4K2yALzdkAE2BRROMFFABzZKo0oMug0VIozGbHjeZOQA5HB6zPoh9B4RhrIc6kJamMwAHkY4puZpoiAArgC2AEbQ4shUeUWlUC5Jmv7pQb4puATEwSHF6dEp6Zns9e7oAMLhFpnZXMgIzAA2zDbIdg4gzq7JGGNQFs3Z-m3NoQjjvVvjA0NaAGrAFhDMk2bTYMCFIsgFJWXSc3BUTlE3XmEDgIDqbhudweez8WEOHQoIQAbtDmGd0Ld7o8EoMrhgACIQeD5OZgWFaTyRPYAHz6aVpWm2u0RyDpGCxMMRanuCD+UBQsxAdgEcxGi3aPlE+RAwAAjvkUDgqCV5jwIGkSIswAJkuKBUdUJw-GzdWRmDBRfrJZoAPS2mYS5aIrInEAgZg60rIZjI6AOCz3EA8Xn8wXMYU6sBzIkksnNaWyhVKlXdObqzXas0oADyfqg+gcYDgxTmEFjcFJ5MRlGNaFNBTmc1NMvuMFAEAsprcFtFFarzTUoEgsEQw378cRmQAsqRuFJwgZRHnoIXgMXS+XiZXJz5KDo4gNJPwTjtl-m1xuyxPq3vzjsj1JurlkNbDVh+rjj8hUdjz6uixLa9twHGsOTRR9VAzLUoB1YdoHgJAtAAcQWYo4DmZoZznDVIBACwskJEDdzIbCSG4KCNRguDwAQscULQjCsISWdyN4E9xiqJtyntZANT5fJ7mQfR10IZB4DmZUpAQfI7GYQpRGmdCAC8nDgWx7CcFweIdXCIHw4TROQOBHUKEgy0gOYqBmWSwHk50fCkX8HkUgQ3g+L4anKcz-kBZBgTLMFygwsBUFwjTVnWHS+LSPCu1AYzTJIOBXk3YT-h4SjM1g5B4NHJCMFQ5h0MwqcWLned+F4xdIlc-RiEgfMIq05RkF40BZjMlLgDSvThWACMZgFFKICyME+IiWpgEtKhmHyZAZU695wFysBMrUKis0hDAAEl8I1TtmiyZ4eAASHa8EwEIFAAANcBumY-lklB9BesEo2YcTgDJaAfXyHUrBwIDOwcsgcDOgBtABpXLwQAawgWbLUKxjSp8MjjKyGG+oI5AEaRrQb2YxRWLO06AH42odWYBVB0bjNpjD9H+MbkTgb7geMgy3Q9L0UF9f07iDZZPte5AACtbNhgShKuiBCjJ06e0taM31ZHHCPQImypJ0hoYAXUVymocV0R2stdcYrisbPlJOZgBwMS2bmRUBGuha8FqYgskB4Guyuh26ezEXhJQdsfqmnU5p1ESrs+CB80V06UeKpidbYViDat-TccbZs6VbYkOwsJPKZABPoCT0QTdO07q-1s3qadR2fDGxm5mZqhWfZv40rBLsec9fz+fzAMgzO7HYpzrJ8d7aZoYAJlhvHEd7Ii41vUjyvI3Bs17VWJT2DXCeIzeIDIxf9eQY2l7yCuoAbk+N+JjO9ahher9UWvKYAUViqBEBgH3NrHw9AXhCGri4QYtdRBVQdNNN2KZVTNkDnYb6zY5YCnoJbd4YIshyw9r9b2nwh4CwLEWPCq1Fa8RujDAAZCvAm68dxnzIg9TSCA4a5SyM9Ls11aZ2T4sKfIAjCCB2KAAkACAxIRissZEg5lgD00EQQtsLCg5uBwLQahDpPT8JEngGyclCg+nLj7Iu5cuz2wRsgG6JBQCKWUA9fQc05hdmSjgfBYjXTXU4YrZWfZT5HynvhTWRUSovxwJnehjC15ayCenKJpBDa1wpvwKmHstI3QOIEGsUN6B-xlhQcJad0bbwNvQPOLZ9rtksd2IQ+8YwJJ8OwdgzjXFdhcVALh-CIBJ14sUf6AhPqFDgDYyEgMfJd2QIUJ0QoEAOwgPQMWz1EGOj0KNLU+EtJtmLsHDCwA2jg1SadP+9hAHAOaXQcwbhIHKGgTA9JvFakYSssslAw1pFrJksYumWRen0FUQ7P2fzQ7t2GhYay4cRwg2jv5ayBC7o4AenySseAk45JtBQfJyAAByEZcV22BvuEpaMt66xIAbVp60eC8QFDgeYfouy7X2mkQ6LphC0ttFy4OLL7hspZK3KYPL0lGLsgpVouS9wACIfniulWAjJcr7ItzICK0V6SVJqWamsSQvFRX3L1Q6UVzlmCiExUcaVpqFXmF4gAKhxp2O1P40Tqo1W5d4VRvhQCNe69JUy-IBVBCAX1frjJkjCrFHVjhQ3pMNW6hmS5KRxGaNFGmKAar6ATc+c1H4GSIjTU6Z8PLVC8V4gSHMP98C4pCKgZAxA-TIAdWkIpFhnWngsEAA))

### Description

The main goal of this PR is to allow users to provide users to strictly type all their custom shapes' props. 

Previously people would have to resort to using an unknown shape, casting, or "casting" through generics (where tldraw types would just accept their type arguments happily). With the new system in place users can augment a builtin tldraw interface like this:
```ts
const CIRCLE_CLIP_TYPE = 'circle-clip'

declare module 'tldraw' {
	export interface TLGlobalShapePropsMap {
		[CIRCLE_CLIP_TYPE]: {
			w: number
			h: number
		}
	}
}

export type CircleClipShape = TLShape<typeof CIRCLE_CLIP_TYPE>
```

Once this is done, users can now use the shape type like this:
```ts
editor.updateShape({
	id: shapeId,
	type: 'circle-clip',
	props: {
		w: 100,
		h: 100,
	},
})
```

Thanks to the power of discriminated unions, this will now error if the user tries to update the shape with the wrong props.

That said, for the time-being, I kept generic parameters in the editor methods to:
- maintain backwards compatibility
- keep this documented pattern working: https://tldraw.dev/docs/shapes#Meta-information

:warning: this allows users to bypass the created augmentation system and locally call methods with shapes having *more* properties than prescribed by the augmentation system. Given the TS structural type system... this is something that can always happen when arguments are not passed inline so it's largely fine as the users should start using the new system anyway. This should allow them to do it gradually though.

This PR consists of:
- the mentioned augmentation system. Related changes are mostly local to [`TLShape` file](https://github.com/tldraw/tldraw/pull/7091/files#diff-74a7534c2bdfff2f6ef9e068496c69dca2a7b09d05c1ff482485fd30480694d3)
- using the new system in the [examples](https://github.com/tldraw/tldraw/pull/7091/files#diff-5f23c6b955fcdcf8a402446d96178e12241ca0cb01af4e73535cc1ca261f81f0), [templates](https://github.com/tldraw/tldraw/pull/7091/files#diff-5845112b17d3eb903a41e90e834858e5b8ab71b6350348811d183137824e4f07) and tests
- updating the docs to reflect the new system [here](https://github.com/tldraw/tldraw/pull/7091/files#diff-1b0ad2aab7ced6fc26f40f9929d2d16269b8fa505bdf4b18b5b746facc5bcefc) and in the "guides" in the examples
- followup code changes required by the new system, they can be found all over the place.
- the same kind of changes related to the bindings (this PR includes the work done in https://github.com/tldraw/tldraw/pull/7133 )

### API changes

Numerous typings changes! make @ds300 compile a list

---

> [!NOTE]
> Introduces module-augmentation-driven typing for custom shapes and bindings, updates core editor APIs and utilities to support it, and migrates docs, examples, templates, and tests accordingly.
> 
> - **Types & Schema**:
>   - Add `TLGlobalShapePropsMap`/`TLGlobalBindingPropsMap`, `TLIndexedShapes`/`TLIndexedBindings`, `ExtractShapeByProps`, and `TLCreateShapePartial`.
>   - Refactor `TLBaseShape`/`TLBaseBinding` and shape/binding record validators to work with augmented types.
> - **Editor API**:
>   - Overload `getShapeUtil`, `isShapeOfType`, and `getBindings*` for typed shape/binding keys.
>   - Update `createShape(s)`/`updateShape(s)` generics and accept typed partials.
>   - Tighten types across selection, snapping, export, reparenting, etc.
> - **Bindings**:
>   - Generic `BindingUtil`/callbacks now use typed `TLBinding`; connection/arrow binding utilities updated.
> - **Docs & Guides**:
>   - Rewrite shape/binding guides to use module augmentation (`TLShape<typeof TYPE>`), update API docs.
> - **Examples & Templates**:
>   - Migrate all examples and app templates to declare module augmentations and use typed shapes; remove legacy generic casts.
> - **Tests**:
>   - Update tests to new typing model and API overloads; add tests for type narrowing and create/update generics.

---

- [x] Unit tests (if present)

### Release notes

- Introduces module-augmentation-driven typing for custom shapes and bindings.

### DIFF
--- a/apps/examples/src/examples/custom-relative-snapping/CustomRelativeSnappingExample.tsx
+++ b/apps/examples/src/examples/custom-relative-snapping/CustomRelativeSnappingExample.tsx
@@ -5,9 +5,9 @@ import {
 	HTMLContainer,
 	RecordProps,
 	ShapeUtil,
-	TLBaseShape,
 	TLHandle,
 	TLHandleDragInfo,
+	TLShape,
 	Tldraw,
 	Vec,
 	VecLike,
@@ -17,20 +17,26 @@ import {
 } from 'tldraw'
 import 'tldraw/tldraw.css'
 
+const Y_SHAPE_TYPE = 'y-shape'
+
 // [1]
-type YShape = TLBaseShape<
-	'y-shape',
-	{
-		center: VecLike
-		armTop: VecLike
-		armLeft: VecLike
-		armRight: VecLike
+declare module 'tldraw' {
+	export interface TLGlobalShapePropsMap {
+		[Y_SHAPE_TYPE]: {
+			center: VecLike
+			armTop: VecLike
+			armLeft: VecLike
+			armRight: VecLike
+		}
 	}
->
+}
 
 // [2]
+type YShape = TLShape<typeof Y_SHAPE_TYPE>
+
+// [3]
 class YShapeUtil extends ShapeUtil<YShape> {
-	static override type = 'y-shape' as const
+	static override type = Y_SHAPE_TYPE
 	static override props: RecordProps<YShape> = {
 		center: vecModelValidator,
 		armTop: vecModelValidator,
@@ -67,7 +73,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 		return true
 	}
 
-	// [3]
+	// [4]
 	getGeometry(shape: YShape): Geometry2d {
 		const { center, armTop, armLeft, armRight } = shape.props
 		const c = Vec.From(center)
@@ -84,7 +90,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 		})
 	}
 
-	// [4]
+	// [5]
 	override getHandles(shape: YShape): TLHandle[] {
 		const indices = [ZERO_INDEX_KEY, ...getIndicesAbove(ZERO_INDEX_KEY, 3)]
 
@@ -102,7 +108,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 				x: shape.props.armTop.x,
 				y: shape.props.armTop.y,
 				index: indices[1],
-				// [5]
+				// [6]
 				snapReferenceHandleId: 'center',
 			},
 			{
@@ -111,7 +117,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 				x: shape.props.armLeft.x,
 				y: shape.props.armLeft.y,
 				index: indices[2],
-				// [6]
+				// [7]
 				snapReferenceHandleId: 'center',
 			},
 			{
@@ -120,7 +126,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 				x: shape.props.armRight.x,
 				y: shape.props.armRight.y,
 				index: indices[3],
-				// [7]
+				// [8]
 				snapReferenceHandleId: 'center',
 			},
 		]
@@ -137,7 +143,7 @@ class YShapeUtil extends ShapeUtil<YShape> {
 		}
 	}
 
-	// [8]
+	// [9]
 	component(shape: YShape) {
 		const { center, armTop, armLeft, armRight } = shape.props
 
@@ -198,7 +204,7 @@ export default function CustomRelativeSnappingYShapeExample() {
 					const centerY = viewportPageBounds.center.y
 
 					editor.createShape({
-						type: 'y-shape',
+						type: Y_SHAPE_TYPE,
 						x: centerX - 100,
 						y: centerY - 100,
 					})
@@ -223,24 +229,31 @@ The shape has three arms radiating from a center junction point:
 - armRight (bottom-right arm endpoint)
 
 [1]
-Define the shape type with four points representing a Y-shaped connector.
+First, we need to extend TLGlobalShapePropsMap to add our shape's props to the global type system.
+This tells TypeScript about the shape type with four points representing a Y-shaped connector.
 
 [2]
-The shape util with validators for each point.
+Define the shape type using TLShape with the shape's type as a type argument.
 
 [3]
-Use Group2d geometry containing three line segments from center to each arm.
+The shape util with validators for each point.
 
 [4]
-Four handles in array order: [center, armTop, armLeft, armRight]
+Use Group2d geometry containing three line segments from center to each arm.
 
 [5]
-With `snapReferenceHandleId: 'center'`, when you shift+drag armTop, it will snap to the center point.
+Four handles in array order: [center, armTop, armLeft, armRight]
 
 [6]
-Similarly, armLeft would snap relative to the center point.
+With `snapReferenceHandleId: 'center'`, when you shift+drag armTop, it will snap to the center point.
 
 [7]
+Similarly, armLeft would snap relative to the center point.
+
+[8]
 And armRight would snap to the center point.
+
+[9]
+The component method defines how our shape renders.
 
 */
--- a/apps/examples/src/examples/easter-egg-styles/EasterEggStylesExample.tsx
+++ b/apps/examples/src/examples/easter-egg-styles/EasterEggStylesExample.tsx
@@ -1,4 +1,4 @@
-import { Editor, TLGeoShape, Tldraw, toRichText } from 'tldraw'
+import { Editor, Tldraw, toRichText } from 'tldraw'
 import 'tldraw/tldraw.css'
 
 export default function EasterEggStylesExample() {
@@ -6,7 +6,7 @@ export default function EasterEggStylesExample() {
 		<div className="tldraw__editor">
 			<Tldraw
 				onMount={(editor: Editor) => {
-					editor.createShapes<TLGeoShape>([
+					editor.createShapes([
 						{
 							type: 'geo',
 							x: 0,
--- a/apps/examples/src/examples/interaction-end-callback/InteractionEndExample.tsx
+++ b/apps/examples/src/examples/interaction-end-callback/InteractionEndExample.tsx
@@ -1,4 +1,4 @@
-import { StateNode, TLGeoShape, TLPointerEventInfo, Tldraw, createShapeId } from 'tldraw'
+import { StateNode, TLPointerEventInfo, Tldraw, createShapeId } from 'tldraw'
 import 'tldraw/tldraw.css'
 
 // [1]
@@ -13,7 +13,7 @@ class QuickShapeTool extends StateNode {
 		const { currentPagePoint } = this.editor.inputs
 		const shapeId = createShapeId()
 
-		this.editor.createShape<TLGeoShape>({
+		this.editor.createShape({
 			id: shapeId,
 			type: 'geo',
 			x: currentPagePoint.x - 50,
@@ -32,7 +32,7 @@ class QuickShapeTool extends StateNode {
 			// [3]
 			onInteractionEnd: () => {
 				// Change fill to semi-transparent after dragging
-				this.editor.updateShape<TLGeoShape>({
+				this.editor.updateShape({
 					id: shapeId,
 					type: 'geo',
 					props: { fill: 'pattern' },


packages/tlschema/src/records/TLShape.ts
```diff
@@ -79,12 +79,75 @@ export type TLDefaultShape =
...
+	video: OverwritableDefaultShape<TLVideoShapeProps>
+}
+
+/** @public */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface TLGlobalShapePropsMap extends TLDefaultShapePropsMap {}
+
+/** @public */
+// prettier-ignore
+export type TLIndexedShapes = {
```
Andarist, line 88: I'll admit this is a little bit funky. You can play with this "system" in this [TS playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251027#code/JYOwLgpgTgZghgYwgAgCoBkBCcDOEDKAFnAA4QA8qAnmQDTIAKUA9iTgHzIDeAUMsmBoQAXGiEBuPshIs2opqxySAvjx6CyadAEEoLAO4K2yALzdkAE2BRROMFFABzZKo0oMug0VIozGbHjeZOQA5HB6zPoh9B4RhrIc6kJamMwAHkY4puZpoiAArgC2AEbQ4shUeUWlUC5Jmv7pQb4puATEwSHF6dEp6Zns9e7oAMLhFpnZXMgIzAA2zDbIdg4gzq7JGGNQFs3Z-m3NoQjjvVvjA0NaAGrAFhDMk2bTYMCFIsgFJWXSc3BUTlE3XmEDgIDqbhudweez8WEOHQoIQAbtDmGd0Ld7o8EoMrhgACIQeD5OZgWFaTyRPYAHz6aVpWm2u0RyDpGCxMMRanuCD+UBQsxAdgEcxGi3aPlE+RAwAAjvkUDgqCV5jwIGkSIswAJkuKBUdUJw-GzdWRmDBRfrJZoAPS2mYS5aIrInEAgZg60rIZjI6AOCz3EA8Xn8wXMYU6sBzIkksnNaWyhVKlXdObqzXas0oADyfqg+gcYDgxTmEFjcFJ5MRlGNaFNBTmc1NMvuMFAEAsprcFtFFarzTUoEgsEQw378cRmQAsqRuFJwgZRHnoIXgMXS+XiZXJz5KDo4gNJPwTjtl-m1xuyxPq3vzjsj1JurlkNbDVh+rjj8hUdjz6uixLa9twHGsOTRR9VAzLUoB1YdoHgJAtAAcQWYo4DmZoZznDVIBACwskJEDdzIbCSG4KCNRguDwAQscULQjCsISWdyN4E9xiqJtyntZANT5fJ7mQfR10IZB4DmZUpAQfI7GYQpRGmdCAC8nDgWx7CcFweIdXCIHw4TROQOBHUKEgy0gOYqBmWSwHk50fCkX8HkUgQ3g+L4anKcz-kBZBgTLMFygwsBUFwjTVnWHS+LSPCu1AYzTJIOBXk3YT-h4SjM1g5B4NHJCMFQ5h0MwqcWLned+F4xdIlc-RiEgfMIq05RkF40BZjMlLgDSvThWACMZgFFKICyME+IiWpgEtKhmHyZAZU695wFysBMrUKis0hDAAEl8I1TtmiyZ4eAASHa8EwEIFAAANcBumY-lklB9BesEo2YcTgDJaAfXyHUrBwIDOwcsgcDOgBtABpXLwQAawgWbLUKxjSp8MjjKyGG+oI5AEaRrQb2YxRWLO06AH42odWYBVB0bjNpjD9H+MbkTgb7geMgy3Q9L0UF9f07iDZZPte5AACtbNhgShKuiBCjJ06e0taM31ZHHCPQImypJ0hoYAXUVymocV0R2stdcYrisbPlJOZgBwMS2bmRUBGuha8FqYgskB4Guyuh26ezEXhJQdsfqmnU5p1ESrs+CB80V06UeKpidbYViDat-TccbZs6VbYkOwsJPKZABPoCT0QTdO07q-1s3qadR2fDGxm5mZqhWfZv40rBLsec9fz+fzAMgzO7HYpzrJ8d7aZoYAJlhvHEd7Ii41vUjyvI3Bs17VWJT2DXCeIzeIDIxf9eQY2l7yCuoAbk+N+JjO9ahher9UWvKYAUViqBEBgH3NrHw9AXhCGri4QYtdRBVQdNNN2KZVTNkDnYb6zY5YCnoJbd4YIshyw9r9b2nwh4CwLEWPCq1Fa8RujDAAZCvAm68dxnzIg9TSCA4a5SyM9Ls11aZ2T4sKfIAjCCB2KAAkACAxIRissZEg5lgD00EQQtsLCg5uBwLQahDpPT8JEngGyclCg+nLj7Iu5cuz2wRsgG6JBQCKWUA9fQc05hdmSjgfBYjXTXU4YrZWfZT5HynvhTWRUSovxwJnehjC15ayCenKJpBDa1wpvwKmHstI3QOIEGsUN6B-xlhQcJad0bbwNvQPOLZ9rtksd2IQ+8YwJJ8OwdgzjXFdhcVALh-CIBJ14sUf6AhPqFDgDYyEgMfJd2QIUJ0QoEAOwgPQMWz1EGOj0KNLU+EtJtmLsHDCwA2jg1SadP+9hAHAOaXQcwbhIHKGgTA9JvFakYSssslAw1pFrJksYumWRen0FUQ7P2fzQ7t2GhYay4cRwg2jv5ayBC7o4AenySseAk45JtBQfJyAAByEZcV22BvuEpaMt66xIAbVp60eC8QFDgeYfouy7X2mkQ6LphC0ttFy4OLL7hspZK3KYPL0lGLsgpVouS9wACIfniulWAjJcr7ItzICK0V6SVJqWamsSQvFRX3L1Q6UVzlmCiExUcaVpqFXmF4gAKhxp2O1P40Tqo1W5d4VRvhQCNe69JUy-IBVBCAX1frjJkjCrFHVjhQ3pMNW6hmS5KRxGaNFGmKAar6ATc+c1H4GSIjTU6Z8PLVC8V4gSHMP98C4pCKgZAxA-TIAdWkIpFhnWngsEAA).

My main motivation behind this complexity was to disallow invalid extensions of default shapes (like arrows, video, etc). I think it's quite dangerous to allow users to skip defining some of the existing props. It poses a risk where of tldraw codebase thinking everything is fine (given it can't see the user's augmentation) and typechecking fine but then crashing accidentally at runtime when some props are not provided by the user's incompatible shape. 

Then `group` is a "core shape" that just can't be overwritten at all. This setup here ensures it's not possible to overwrite.

And the user might also get errors on those invalid augmentations. However, that requires `skipLibCheck: false` (the default) and a lot of people might have the opposite value configured. 

You might ask... why all of this is inlined and somewhat unreadable... I don't like it either but it's on purpose. Basically, any type that is used here has to be exported from this package too and I didn't feel like exporting a bunch of one-off type helpers that would clobber auto-imports, completions etc. That's enforced by `api-extractor`.

apps/docs/content/docs/shapes.mdx
```diff
@@ -95,15 +95,29 @@ You can create your own custom shapes. In the examples below, we will create a c
 
 In tldraw's data model, each shape is represented by a JSON object. Let's first create a type that describes what this object will look like.
 
+First, we need to declare a module augmentation to register our shape's props:
```
ds300:
```suggestion
First, we register the shape's props using TypeScript module augmentation:
```

apps/examples/src/examples/cubic-bezier-shape/CubicBezierShape.tsx
```diff
@@ -54,7 +55,7 @@ export class BezierCurveShapeUtil extends ShapeUtil<MyBezierCurveShape> {
 		return true
 	}
 
-	// [3]
+	// [4]
```
ds300, line 58: i think we'll need to renumber the ones below this too. sorry 😞

apps/examples/src/examples/pin-bindings/PinExample.tsx
```diff
@@ -30,13 +31,21 @@ import {
...
 import 'tldraw/tldraw.css'
 
-// eslint-disable-next-line @typescript-eslint/no-empty-object-type
-type PinShape = TLBaseShape<'pin', {}>
+const PIN_TYPE = 'pin'
+
+declare module 'tldraw' {
+	export interface TLGlobalShapePropsMap {
+		// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+		[PIN_TYPE]: {}
```
ds300: I know it was like this already, but elsewhere we've used `Record<string, never>` for this, which is much safer, and it'd be good to make that consistent here since we're refactoring.

packages/tlschema/src/records/TLShape.ts
ds300: > to disallow invalid extensions of default shapes (like arrows, video, etc)

i understand your thinking here, but actually it's totally fine for people to have completely incompatible types for shapes with the same name as our default shapes. They aren't always going to be 'augmenting' the built-in types but maybe replacing them entirely. 

We can and should avoid any danger here by making our own internal shape types like `TLArrowShape`, `TLGeoShape`, etc non-augmentable. i.e. not defined in terms of the global shape props map.

For the core shapes stuff here, I think we can safely hard-code that to just 'group' for now and avoid some of the helper types and mapped types here. It's exceedingly unlikely that we'll add more core shapes in the future, and the risk of allowing people to override them is low when we error at runtime, so it's not really worth paying the complexity cost here IMO.

ds300 (commented):
This is looking fantastic! 

I'm not quite finished reviewing, but it's the end of my day and I have a few small bits of actionable feedback so I figure I'll leave them here now and finish the review tomorrow.

apps/examples/src/examples/pin-bindings/PinExample.tsx
Andarist: I wouldn't necessarily call it safer ([TS playground](https://www.typescriptlang.org/play/?ts=6.0.0-dev.20251030#code/CYUwxgNghgTiAEYD2A7AzgF3gBxk7aAXPAErhIzAA8mMAligOYA08KIAbiDAHwDcAKAG58aAHQAzJEjEB3ABZQMnbn3gB6dfG54YrFHTAgBydFgDe8BUpUx4AX3gBeHHgKTpazWyTaYu-SQsAyN4PmwhU0x4WmJaBkZnK0VlLhgvLSgICCRZEGB4ACNwKABXNAQAA3Y0yvg6NHgoNDQ6RhQoQogEDF9bAE8MeQSBIA)) ;p
```ts
declare const props: Record<string, never>;

props.foo.whatever; // error, nice
const { whatever } = props.foo; // no error, not nice ;p

const str: string = whatever; // allowed because `never` is assignable to everything
```

I'll go ahead and just implement the requested change though

apps/dotcom/client/src/fairy/actions/LabelActionUtil.ts
```diff
@@ -17,8 +31,17 @@ export class LabelActionUtil extends AgentActionUtil<LabelAction> {
 		if (!action.complete) return action
 
 		const shapeId = helpers.ensureShapeIdExists(action.shapeId)
-		if (!shapeId) return null
-
+		if (!shapeId || !this.agent?.editor) {
+			return null
+		}
+		const shape = this.agent.editor.getShape(`shape:${shapeId}` as TLShapeId)
```
Andarist, line 37: @mimecuvalo could you take a look at the changes here? https://github.com/tldraw/tldraw/pull/7102/files#r2526219226

packages/tldraw/src/lib/tools/SelectTool/childStates/DraggingHandle.tsx
```diff
@@ -228,7 +228,7 @@ export class DraggingHandle extends StateNode {
 			}
 			const endChanges = util.onHandleDragEnd?.(shape, handleDragInfo)
 			if (endChanges) {
-				this.editor.updateShapes([{ ...endChanges, id: shape.id, type: shape.type }])
+				this.editor.updateShapes([{ ...endChanges, id: shape.id }])
```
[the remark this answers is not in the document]
Andarist: `endChanges` already has the `type` property, we don't need to repeat it here.

packages/tlschema/src/records/TLShape.ts
Andarist: My main concern was that the code could accidentally depend on the in-repo shape props. I have experimented with replacing props on `TLArrowShape` to see what error would pop up and there were some. I thiiiink all of them were isolated to tests and to the code called by the util itself and the associated binding. I'd assume one could use a custom arrow shape without using a custom arrow binding util and then things would throw at runtime if that custom arrow shape wouldn't conform to the default arrow props. 

I have removed default shapes now from the globals props maps, see: https://github.com/tldraw/tldraw/pull/7091/commits/2502efb74f9a35595dcb168e7f443e1987a5e0ce

ds300: love the improvement here, thank you!

apps/dotcom/client/src/fairy/actions/LabelActionUtil.ts
mimecuvalo: i dig it - thank you!

PR #7103, opened 2025-11-11
perf(draw): use delta-encoded segments for draw shapes

steveruizok:
This PR updates our draw shape to use deltas rather than absolute points. It changes the data structure for a segments points from an array of VecModels to an array of integers. The resulting shapes are roughly to 75-80% smaller.

For example:

image

Before: 8826 characters of JSON
After: 2212 characters of JSON
Size: 25% of original

image

Before: 19318 characters of JSON
After: 4458 characters of JSON
Size: 24% of original

image

Created at 800% zoom. 9953 -> 2216, 22% of original

image

Created at 10% zoom. 10030 -> 3047, 30% of original

There's still an optimization that I haven't tried here, which is to use the zoom level at which the user began the drawing as a means to further scale the input shapes. In other words, a very large shape created when the user is zoomed out will have bigger deltas between points as compared to when the user is very zoomed in, however in both cases this precision can be a factor of the zoom and so the deltas scaled by that. Smaller deltas that can be expanded later!

### Change type

- [x] `improvement`

### Test plan

1. Try drawing
2. Try highlighting
3. Zoom in and draw
4. Zoom out and draw
5. Turn on dynamic size and draw
6. Try drawing with a pen or stylus

- [x] Unit tests
- [x] End to end tests

### Release notes

- Improved storage size of draw shapes.

---

> [!NOTE]
> Switches draw/highlight shapes to firstPoint + integer-delta segments with a new zoom prop, updating schema/migrations, tooling, rendering, imports, and tests accordingly.
> 
> - **Schema & Migrations**
>   - Add `zoom` to `TLDrawShape`/`TLHighlightShape`; change `TLDrawShapeSegment` to `{ firstPoint, points:number[], type }`.
>   - New migrations (`AddEfficiency`) convert absolute points ⇄ delta-encoded segments; validators/tests updated.
> - **Core Rendering & Geometry**
>   - Update path reconstruction, resizing, interpolation, and stroke generation to read delta-encoded segments.
>   - Custom canvas renderer decodes deltas for straight/free segments.
> - **Tools & Creation**
>   - Drawing tool records segments as deltas; computes `isClosed` using reconstructed last point; carries `zoom`.
>   - Pen action and toolbar drag icons emit delta-encoded segments.
> - **Import/Export**
>   - Excalidraw import builds delta-encoded draw segments.
>   - v1 document import converts draw points to new segment format.
> - **Tests & Fixtures**
>   - Snapshots, unit/e2e fixtures updated to new segment shape; add FPS baselines and last-run log.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 763f1597ab0f015519f719f498b1310126481972. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/interpolate-props.ts
+++ b/packages/tldraw/src/lib/shapes/shared/interpolate-props.ts
@@ -23,15 +23,15 @@ export const interpolateSegments = (
 
 	// Interpolate points
 	const interpolatedPoints = pointsToUseStart.map((point, k) => {
-		let z = 0.5
-		if (pointsToUseEnd[k].z !== undefined && point.z !== undefined) {
-			z = lerp(point.z, pointsToUseEnd[k].z as number, progress)
-		}
-		return {
+		const result: VecModel = {
 			x: lerp(point.x, pointsToUseEnd[k].x, progress),
 			y: lerp(point.y, pointsToUseEnd[k].y, progress),
-			z,
 		}
+		// Only include z if both start and end points have z values (pen input)
+		if (pointsToUseEnd[k].z !== undefined && point.z !== undefined) {
+			result.z = lerp(point.z, pointsToUseEnd[k].z as number, progress)
+		}
+		return result
 	})
 	// Return all interpolated points in a single segment
 	// Convert points back to segment format (firstPoint + deltas)


ds300:
steve I appreciate the clever tricks to make json less verbose here, but I think we'd have a easier time now and going forward if we pick up the base64 branch. It's the same or better compression, it's a smaller lift, and now that our diffs support string append operations it shouldn't have any negative perf consequences. Plus UInt8Array got toBase64 and fromBase64 methods added to the spec so it should be lightning fast.

https://github.com/tldraw/tldraw/tree/david/base-64-draw-shape

steveruizok:
> steve I appreciate the clever tricks to make json less verbose here, but I think we'd have a easier time now and going forward if we pick up the base64 branch. It's the same or better compression, it's a smaller lift, and now that our diffs support string append operations it shouldn't have any negative perf consequences. Plus UInt8Array got toBase64 and fromBase64 methods added to the spec so it should be lightning fast.
> 
> https://github.com/tldraw/tldraw/tree/david/base-64-draw-shape

This one is new to me!

Looking more at your PR, I'd actually love to combine the approaches and see what the gains would be. But I agree, if we're going to use base64, then let's do that first.

PR #7133, opened 2025-11-17
improvement(editor): augmentation-based binding types

Andarist:
### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

It relies on the existing test suite and additional manual testing.

- [x] Unit tests (if present)

### Description

This continues the work made in https://github.com/tldraw/tldraw/pull/7091 . The set of changes is pretty much identical o the ones from that other PR but this time the augmentation system for *bindings* is being introduced.

With this PR one can define custom bindings as follows:
```ts
const PIN_TYPE = 'pin'

declare module 'tldraw' {
	export interface TLGlobalBindingPropsMap {
		[PIN_TYPE]: {
			anchor: VecModel;
		}
	}
}

export type PinBinding = TLBinding<typeof PIN_TYPE>
```

---

> [!NOTE]
> Migrates bindings to a module-augmentation TLBinding system with typed helpers and overloads, updating schema, editor APIs, docs, examples, and tests accordingly.
> 
> - **Core/Schema (tlschema)**
>   - Add augmentation entrypoint `TLGlobalBindingPropsMap`, `TLIndexedBindings`, and generic `TLBinding<K>`.
>   - Refactor `TLBaseBinding` structure; update `TLBindingId`, `TLBindingCreate/Update` to be distributive.
>   - Update `createBindingRecordType` and API report.
> - **Editor API**
>   - Make `BindingUtil` and binding callbacks generic on `TLBinding`.
>   - Add typed overloads for `editor.getBindingsFromShape/ToShape/InvolvingShape` and `getBindingUtil`.
>   - Replace usages of `TLUnknownBinding`; improve `isShapeOfType` usage.
> - **Docs & Examples**
>   - Document new custom binding pattern using `declare module 'tldraw'` and `TLBinding<typeof TYPE>`.
>   - Migrate examples (layout, pin, sticker, connection, arrows, etc.) to augmentation-based bindings.
> - **Refactors/Tests**
>   - Remove unnecessary generic type args on `createShape/updateShape(s)`; simplify `isShapeOfType` calls.
>   - Update tests and internal code to new binding APIs and typing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4cb328ac2b60cb76efbf230f0bc2984d938e989d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/test/test-jsx.tsx
+++ b/packages/tldraw/src/test/test-jsx.tsx
@@ -49,29 +49,30 @@ type PropsForAsset<Type extends string> = Type extends TLAsset['type']
 	? Partial<AssetByType<Type>['props']>
 	: Record<string, unknown>
 
-interface CommonBindingProps {
+interface BindingReactConnections {
 	from?: string | TLShapeId
 	to: string | TLShapeId
+}
+
+interface CommonBindingReactProps extends BindingReactConnections {
 	ref?: string
 	id?: TLBindingId
 }
 
-type PropsForBinding<Type extends TLBinding['type']> = CommonBindingProps &
+type ReactPropsForBinding<Type extends TLBinding['type']> = CommonBindingReactProps &
 	Partial<TLBinding<Type>['props']>
 
 type BindingToCreate = TLBinding extends infer E
 	? E extends TLBinding
 		? {
 				type: E['type']
-				props: PropsForBinding<E['type']>
+				props: Partial<TLBinding<E['type']>['props']>
+				id: TLBindingId | undefined
 				parentId: TLShapeId | undefined
 				ref: string | undefined
+				connections: BindingReactConnections
 			}
 		: never
-	: never
-
-type WithPartialProps<T extends { props: unknown }> = T extends T
-	? Omit<T, 'props'> & { props: Partial<T['props']> }
 	: never
 
 const createElement = (
@@ -104,7 +105,7 @@ const tlBinding = new Proxy(
 		},
 	}
 ) as {
-	[K in TLBinding['type']]: (props: PropsForBinding<K>) => null
+	[K in TLBinding['type']]: (props: ReactPropsForBinding<K>) => null
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5876,7 +5872,15 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		type: K
+	): Extract<TLBinding, { type: K }>[]
+	getBindingsFromShape<Binding extends TLBinding = TLBinding>(
+		shape: TLShape | TLShapeId,
+		type: Binding['type']
+	): Binding[]
+	getBindingsFromShape<Binding extends TLBinding = TLBinding>(
 		shape: TLShape | TLShapeId,
 		type: Binding['type']
 	): Binding[] {
```
ds300, line 5886: what's the purpose of the extra overload here? seems to work with only the top signature. I guess backwards compatibility?
```suggestion
	getBindingsFromShape<K extends TLBinding['type']>(
		shape: TLShape | TLShapeId,
		type: K
	): Extract<TLBinding, { type: K }>[] {
```

packages/tldraw/src/test/test-jsx.tsx
```diff
@@ -275,18 +283,19 @@ export function shapesFromJsx(shapes: React.JSX.Element | Array<React.JSX.Elemen
...
+		delete bindingCopy.props.from
+		delete bindingCopy.props.to
 
 		bindings.push({
 			id: bindingId,
-			type: binding.type,
 			fromId,
 			toId,
-			props,
+			...bindingCopy,
```
ds300: hmm this looks like it might overwrite the id(s?). can we keep it roughly how it was before? should we spread it at the top of the object literal?

packages/editor/src/lib/editor/Editor.ts
Andarist: >  I guess backwards compatibility?

This. The last one is an implementation overload so it isn't available publicly. So then when adding a new one I had to "duplicate" the existing one to make it public. But I've only done this to maintain backwards compatibility. If you don't need it then we could remove this (and similar extra overloads I added in other places). 

We could even go further and remove type parameters from a lot of those methods and rely more on discriminated unions alone, see this part of the description in [the shapes PR](https://github.com/tldraw/tldraw/pull/7091):

> That said, for the time-being, I kept generic parameters in the editor methods to:
> - maintain backwards compatibility
> - keep this documented pattern working: https://tldraw.dev/docs/shapes#Meta-information

> :warning: this allows users to bypass the created augmentation system and locally call methods with shapes having *more* properties than prescribed by the augmentation system. Given the TS structural type system... this is something that can always happen when arguments are not passed inline so it's largely fine as the users should start using the new system anyway. This should allow them to do it gradually though.

packages/tldraw/src/test/test-jsx.tsx
Andarist: It wouldn't override `id` as far as I can tell - the `bindingCopy` only had `type` and `props` properties. I only need to keep them together to preserve the output type after spreading to stay as a discriminated union. But ye, I'd consider spreading at the top to be better and derisking this 👍 

I took another pass on this and I think I came up with a cleaner solution that just converts the react props into an easier to work with `BindingToCreate` earlier: https://github.com/tldraw/tldraw/pull/7133/commits/4cb328ac2b60cb76efbf230f0bc2984d938e989d . This ensures that `bindingToCreate.props` are bindings props and not element props (as those have extra properties "mixed" into the binding props) 

Tbf, it doesn't require any casts or extra tricks mostly thanks to an element being typed as `ReactElement<any, any>` but it feels nice to remove the need for those extra `delete`s that were here originally and stuff.

packages/editor/src/lib/editor/Editor.ts
ds300: alright let's err on the side of backcompat. thanks

PR #7147, opened 2025-11-19
feat(editor): allow registering custom drop handler

vantezzen:
Currently, customizing tldraw's drop integrated handling is only possible within the "url" and "files" external content handlers.
This PR adds an additional "drop" handler that allows handling any other type of drop event. This allows registering arbitrary handlers, e.g. for on-page drop events of elements outside of the tldraw canvas:

Screen Recording 2025-11-19 at 11 03 00

### Change type

- [x] `feature`

### Test plan

1. Create a draggable element on the page
2. Register a "drop" handler

```
editor.registerExternalContentHandler("drop", (info) => {
          console.log("Handling external drop:", info);
});
```

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Allow registering custom drop handler

### API changes

- Add "drop" as a handler type to "registerExternalContentHandler"

---

> [!NOTE]
> Introduces `TLDropExternalContent` and routes canvas `onDrop` to a registered custom handler before default files/url handling.
> 
> - **API / Types**:
>   - Add `TLDropExternalContent` (`type: 'drop'`, carries React `DragEvent`) and include in `TLExternalContent` union; export from `src/index.ts`.
>   - API report updated (imports `DragEvent`, new interface/type entries).
> - **Editor**:
>   - Extend `externalContentHandlers` with `drop`; `registerExternalContentHandler` now supports `'drop'`.
> - **Canvas Events**:
>   - `onDrop` calls `editor.putExternalContent({ type: 'drop', event, point })` when a custom drop handler is registered, else falls back to files/url handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7c9f7156139f498d0fb01c693b923bd25a38443. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


vantezzen:

mimecuvalo:
I like this. Thanks for your PR! @ds300 any concerns?

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -122,6 +122,12 @@ export function useCanvasEvents() {
 					})
 					return
 				}
+
+				await editor.putExternalContent({
+					type: 'drop',
```
mimecuvalo: we should probably be more targeted here though.
this won't fire if it's a `url` or `file`. so we would want to call this something like `unspecified-drop`/`generic-drop`.

or, i could see this being something that handles just HTML content, like in your screencap.

or or, alternatively, this goes above as the first handler, and if nothing is specified it goes on to the files/url logic.

vantezzen: I like the last idea of being able to choose between handling any generic drop event yourself or opting in to tldraw preparing them into more specific formats. Although it would be nice to be able to signal back if the other handlers should still run, e.g if your top handler doesn't support that format.

But I'm open to change however you see fits best into the overall project.

ds300: chiming in here to say i agree it should go first if there's a 'drop' handler defined.

```ts
if (editor.externalContentHandlers.drop) {
   editor.putExternalContent({type: 'drop', ...})
   return
}
// ... otherwise handle files/urls specifically
```

then the drop handler can decide if it wants to recursively call putExternalContent with files or a url.

mimecuvalo: cool, we're in agreement that this gets reworked so that the option that the drop handler goes first. 👍
@vantezzen lemme know if you want to take a stab at it :)

vantezzen: @mimecuvalo Nice - just pushed David's suggested reordering

mimecuvalo: (sorry for the delay here, i've been sick)

thx for making the change! the placement is good. i think the last change is this still needs to be renamed. the rest of the items are nouns (text/files/url/etc) and this is sort of a verb; or, at least it's hard to tell that the other ones won't have effect when this is specified.

i would say this could be called TlCatchAllExternalContent/'catch-all' because otherwise it the mutually exclusivity of this handler will get lost on people.

steveruizok:
Looking at this again, it feels like a weird use of the external content handler. We just want a way to interrupt the external content handler when a drop occurs? I'd throw that in an editor option.

steveruizok:
See escape here: https://github.com/tldraw/tldraw/pull/7527

mimecuvalo:
made a comment on the other PR

steveruizok:
Sorry @vantezzen, we're not accepting pull requests from external contributors at this time. [See discussion here](https://github.com/tldraw/tldraw/issues/7695). If you'd like to report a bug or suggest a feature, please [open an issue](https://github.com/tldraw/tldraw/issues/new/choose) instead.

PR #7210, opened 2025-11-26
feat(dotcom): add usage based limits for fairies

MitjaBezensek:
Add usage based rate limiting.
- Store weekly usage in `user_fairies.weeklyUsage` as JSONB (`{"2025-W01": 12.34}`)
- Check rate limit before streaming actions (429 if exceeded)
- Record actual cost after each generation completes
- ISO week-based tracking
- Admins bypass limits - we still store their usage so that we can have a rough estimate of how much things cost
- We talk directly to User durable object that is defined in sync worker (using `script_name` to link these things together)
- Also fixes fairy auth.

### Change type

- [x] `other`

### Test plan

- [ ] Unit tests (if present)

### Release notes

- Add usage based limits for fairies

---

> [!NOTE]
> Implements weekly, cost-based fairy usage tracking and rate limiting across workers with DO-to-DO checks, updates client error messaging, and adds schema/migration support.
> 
> - **Backend (sync-worker)**:
>   - Add `user_fairies.weeklyUsage` JSONB with migration `026_fairy_usage_tracking.sql` and schema typing.
>   - TLUser DO: new endpoints `GET /app/:userId/fairy/has-access`, `POST /fairy/check-rate-limit`, `POST /fairy/record-usage` (ISO-week based; atomic JSONB updates).
>   - Admin: ensure `weeklyUsage: {}` on upsert.
> - **Fairy worker**:
>   - Introduce `INTERNAL_BASE_URL` and bind `TL_USER` DO; env/wrangler bindings updated (all envs + preview via deploy script).
>   - Before streaming: check rate limit via TLUser DO; on finish: record actual cost; propagate user headers (`X-User-Id`, `X-Is-Admin`).
>   - Add robust SSE error/abort handling and CORS headers mutability fix.
> - **Client/UI**:
>   - Show specific toasts for weekly fairy limit (`fairy_rate_limit_title`/`fairy_rate_limit_exceeded`); add i18n strings.
>   - Agent fetch: fail fast on non-OK responses with server error payload; suppress console noise for rate-limit errors.
> - **Auth**:
>   - `hasFairyAccess` now queries TLUser DO instead of hardcoded allowlist.
> - **Deploy/Config**:
>   - Preview deploy script injects `TL_USER` DO binding for fairy worker.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f201ed02dd473f810a33b252cedd4f1872b867b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


apps/dotcom/client/src/fairy/FairyApp.tsx
MitjaBezensek, line 1: We should decide how we message this, so I just added something for now.

apps/dotcom/fairy-worker/src/do/AgentDurableObject.ts
```diff
@@ -11,6 +12,47 @@ export class AgentDurableObject extends DurableObject<Environment> {
...
+		// Admins bypass rate limits
+		if (userIsAdmin) return null
+
+		const checkRes = await userStub.fetch(
+			`${INTERNAL_BASE_URL}/app/${userId}/fairy/check-rate-limit`,
+			{
+				method: 'POST',
+				headers: { 'Content-Type': 'application/json' },
+			}
+		)
```
ds300, line 34: interesting that you opted for `fetch` instead of using rpc methods. did RPC not work?

apps/dotcom/zero-cache/migrations/026_fairy_usage_tracking.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE user_fairies
+ADD COLUMN "weeklyUsage" JSONB DEFAULT '{}'::jsonb;
```
ds300, line 2: when you asked about how to do this i suggested using postgres because I assumed the data would need to be read by the client. It looks like that's _not_ happening (yet?). If it will never need to happen, your initial approach of using workers KV (or even durable object storage on the fairy agent DO) is gonna be a lot simpler to implement, with the downside that it will be a bit harder for us to access the usage data for analytics, but not impossible.

apps/dotcom/fairy-worker/src/do/AgentDurableObject.ts
MitjaBezensek: It doesn't seem to be supported in dev, unsure about prod.
> ✘ [ERROR] Error: Cannot access "TLUserDurableObject#testRpc" as Durable Object RPC is not yet supported between multiple dev sessions.

Fetch worked, so I went with that as it was easier to test.

Issue #7077, opened 2025-11-07
canEdit cannot be disabled for some shapes


vbrajon:
### What happened?

canEdit works on arrow but not on other shapes.

Try to create an rectangle and double click on it and it will crash.

This is true for tldraw@3, I can't verify on tldraw@4 as I don't have licence key.

SIDE NOTE: I suggest making a playground to submit reproduction easily, making it with v0 was simple enough but still not ideal in my opinion. I suggest also making an error message on TLDraw v4 when there is no licence, instead of removing the UI after few seconds. I suggest also making a free licence key, at least to display the app for 5min for reproduction purpose. (and ideally free for open-source projects)

Thanks for tldraw

### How can we reproduce the bug?

https://v0.app/chat/tldraw-canvas-options-bDZqPxIbuxY

```ts
"use client"

import { Tldraw, defaultShapeUtils } from "tldraw"
import "tldraw/tldraw.css"

export default function TldrawCanvas() {
  const shapeUtils = defaultShapeUtils.map((Util) => {
    // NOTE: no way to disable all shapes, it crashes on ovals/ellipse
    // if (Util.type !== "arrow") return Util
    return class extends Util {
      canEdit(shape: any) {
        return false
      }
    }
  })

  return (
    <div className="w-full h-full">
      <Tldraw shapeUtils={shapeUtils} />
    </div>
  )
}
```

### Where did this happen?

The developer package

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
Thanks for the bug report and suggestion. Hear you on the license requirement (in prod/v0) being annoying. It's on our radar to improve for these environments 👍

mimecuvalo:
fixed in https://github.com/tldraw/tldraw/pull/7361 and related PRs

PR #7276, opened 2025-12-02
feat(editor): add startEditingShape/stopEditingShape; deprecate setEditingShape

mimecuvalo:
this is based on top of https://github.com/tldraw/tldraw/pull/7298

Right now it's hard to start editing shape without doing several checks to make sure that you're safe to start editing.
It currently involves calling `setEditingShape`, transitioning or setting the current tool to `EditingShape` all while checking if we're in readonly mode and making sure the shape is editable (via the `canEdit` shape API check). 

Addresses https://github.com/tldraw/tldraw/issues/7077

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- editor: add startEditingShape; deprecate setEditingShape

### API changes

editor: add startEditingShape to better encapsulate proper ergonomics around doing an edit. deprecate setEditingShape

---

> [!NOTE]
> Adds Editor.canEditShape/canCropShape and a startEditingShape helper used across tools/selection; enables embeds to be editable while locked; relocates TLResizeShapeOptions; tweaks paste parent logic and event order.
> 
> - **Editor API**:
>   - Add `editor.canEditShape()` and `editor.canCropShape()` guards; introduce private `_setEditingShapeInternal`.
>   - Move `TLResizeShapeOptions` to `types/misc-types` and re-export from index.
>   - Change misc event flow to `emit('event')` before `root.handleEvent`.
>   - Remove parent-at-pointer override when pasting content at a point in `putContentOntoCurrentPage`.
> - **Selection/Tools**:
>   - Add `startEditingShape` util and update states (`arrow`, `geo`, `note`, `text`, `select`) to use `editor.canEditShape`/`editor.canCropShape`.
>   - Remove `getShouldEnterCropModeOnPointerDown`.
> - **Embeds**:
>   - Add `canEditWhileLocked` to `ShapeUtil` and `EmbedDefinition`; implement in `EmbedShapeUtil`.
>   - Default embed definitions set `canEditWhileLocked: true` for supported hosts.
> - **Tests/Examples**:
>   - Update tests and examples/comments to reflect new editing flow and API changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 74c181972070be71eb2856f8c1b7d3299e47ad11. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/text/toolStates/Idle.ts
+++ b/packages/tldraw/src/lib/shapes/text/toolStates/Idle.ts
@@ -33,16 +33,11 @@ export class Idle extends StateNode {
 			this.editor.setCurrentTool('select')
 			if (
 				!startEditingShape(this.editor, onlySelectedShape, {
-					info: { ...info, target: 'shape', shape: onlySelectedShape },
+					info,
 				})
 			) {
 				return
 			}
-			this.editor.root.getCurrent()?.transition('editing_shape', {
-				...info,
-				target: 'shape',
-				shape: onlySelectedShape,
-			})
 		}
 	}
 
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -18,7 +18,7 @@ import { isOverArrowLabel } from '../../../shapes/arrow/arrowLabel'
 import { getHitShapeOnCanvasPointerDown } from '../../selection-logic/getHitShapeOnCanvasPointerDown'
 import { selectOnCanvasPointerUp } from '../../selection-logic/selectOnCanvasPointerUp'
 import { updateHoveredShapeId } from '../../selection-logic/updateHoveredShapeId'
-import { startEditingShape as startEditingShapeHelper } from '../selectHelpers'
+import { startEditingShape } from '../selectHelpers'
 
 const SKIPPED_KEYS_FOR_AUTO_EDITING = [
 	'Delete',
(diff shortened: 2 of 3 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -2278,6 +2267,52 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+			...(options.info ?? {}),
+			target: 'shape',
+			shape,
+		})
+
+		if (options.selectAll) {
+			this.emit('select-all-text', { shapeId: shape.id })
+		}
+
+		return true
```
steveruizok: The pattern with Editor methods is to return the Editor class instance. I'd add a `canEdit` method here if the user needs to check, or else check after. Its annoying but that's the pattern, I'd rather not change it.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -2278,6 +2267,52 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+
+		this.select(shape.id)
+		this._setEditingShapeInternal(shape.id)
+		if (!this.getEditingShape()) return false
+
+		this.setCurrentTool('select.editing_shape', {
+			...(options.info ?? {}),
+			target: 'shape',
+			shape,
+		})
```
steveruizok: select is a state that only exists at the tldraw level

steveruizok:
Hey, let me put up a PR that shows a different way of getting to the same outcome.

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: hah! silly mistake

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: gotcha, we could still do a `return this`. i don't need to do the `return true` here

steveruizok:
Closing this one, I've moved lots of ideas following our conversation!

PR #7291, opened 2025-12-04
fix(editor): exit editing mode when dragging and blurring input

MitjaBezensek:
When in shape editing mode and we drag the mouse (which causes blurring text input) we now exit editing mode. 

### Before 
After dragging the mouse slightly we stay in edit mode, but the text input is no longer focused - see how we have no handles around the shape. Copying (the shape) at this point does not work, as we still have editing shape id and we have some logic to prevent copying while editing.

https://github.com/user-attachments/assets/e0a1aa0a-6e33-4b29-a335-683ddad2bc1e

### After
We now exit editing and start translating.

https://github.com/user-attachments/assets/13214068-1262-48dd-8cbf-85f5cd948b2c

### Change type

- [x] `bugfix`

### Test plan

1. Enter edit mode on a shape with text.
2. Drag the mouse away from the label so the input blurs.
3. Verify the editor exits editing mode and starts translating.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- When editing a shape and dragging a mouse causes the text input to blur we now exit editing.

---

> [!NOTE]
> Exits shape editing and begins translating when dragging from a label causes the text input to blur, with a new e2e test and related EditingShape state updates.
> 
> - **Editing behavior (SelectTool `EditingShape`)**:
>   - Detects input focus via new `isTextInputFocused()` and tracks presses on the currently edited shape (`didPointerDownOnEditingShape`).
>   - When dragging from the editing shape and the text input is blurred, exits edit mode and transitions to `translating`.
>   - Refactors hit tracking: rename `hitShapeForPointerUp` to `hitLabelOnShapeForPointerUp`; update `onPointerDown`, `onPointerMove`, and `onPointerUp` accordingly.
> - **E2E tests**:
>   - Adds test `exits edit mode when dragging from text label with blurred input` in `apps/examples/e2e/tests/test-focus.spec.ts` to validate the new behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b63101c5b278372f07847d98c1dbe9df36774bf5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
@@ -27,7 +27,8 @@ export class EditingShape extends StateNode {
 		return (
 			container.contains(document.activeElement) &&
 			(document.activeElement?.nodeName === 'INPUT' ||
-				document.activeElement?.nodeName === 'TEXTAREA')
+				document.activeElement?.nodeName === 'TEXTAREA' ||
+				(document.activeElement as HTMLElement)?.isContentEditable)
 		)
 	}
 


packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -21,6 +21,21 @@ export class EditingShape extends StateNode {
 	hitShapeForPointerUp: TLShape | null = null
 	private info = {} as EditingShapeInfo
 
+	private isTextInputFocused(): boolean {
```
MitjaBezensek, line 25: Do we have a better way to know this?

steveruizok:
Good catch!

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
steveruizok: We have something like this for keyboard shortcuts

https://github.com/tldraw/tldraw/blob/bea35c2d049db05c851e72bf83a60f4f68aae990/packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts#L148-L156

steveruizok: Though this is actually something we should fix (i.e. whether a shape should prevent keyboard shortcuts while editing). We used to have a check on whether the user was focused on text inputs or text areas...

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -21,6 +21,21 @@ export class EditingShape extends StateNode {
...
+		const container = this.editor.getContainer()
+		return (
+			container.contains(document.activeElement) &&
+			(document.activeElement?.nodeName === 'TEXTAREA' ||
+				(document.activeElement as HTMLElement)?.isContentEditable)
```
mimecuvalo, line 31: oh man, i went down a rabbit hole here, our code is a bit messy for this stuff. we have like 10 different checks for this around the codebase that all do this a differently 🤦 

for now, make sure you check nodeName is either `INPUT` or `TEXTAREA`. i'll make a followup PR to unify this better across the codebase

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -120,7 +149,8 @@ export class EditingShape extends StateNode {
 					) {
 						// it's a hit to the label!
 						if (selectingShape.id === editingShape.id) {
-							// If we clicked on the editing geo / arrow shape's label, do nothing
+							// Track click on editing shape's label for drag detection
+							this.hitShapeForPointerUp = selectingShape
```
mimecuvalo: ah hm, this is a poorly named variable by me. this should have been named `this.hitLabelOnShapeForPointerUp`. i think you were trying to add it b/c it looked missing but this was intentional.
as such, this is changing the expectation for `onPointerUp`/`onPointerMove` which has logic that is only supposed to execute if it's a label that is in question.

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
mimecuvalo: made comment below - didn't see this one, we have some other checks in the codebase that do this: `activeElementShouldCaptureKeys` and `areShortcutsDisabled` (two different `areShortcutsDisabled`!) but they're scattered and work differently. i'll make a followup to unify them

mimecuvalo (changes requested):
blocking b/c of the hitShape thing

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
MitjaBezensek: Rename the variable. Also introduced a new variable to handle my use case 👍

steveruizok:
Let's get this landed before this week's release!

PR #7297, opened 2025-12-05
feat(editor): add getCanEditShape and getCanCropShape methods

steveruizok:
Add a new public `Editor.getCanEditShape` method that checks whether a shape can be edited. This centralizes the logic for determining shape editability and makes it easier for consumers to query this state.

The method checks:
- Whether the shape exists
- Whether the shape is already being edited
- Whether the shape's util allows editing (`canEdit`)
- Whether the editor is in readonly mode (and whether the shape has an exception via `canEditInReadonly`)

Also refactors `setEditingShape` to use early returns for better clarity, separating the null case from the valid shape case.

### Change type

- [x] `api`

### Test plan

1. Call `editor.getCanEditShape(shapeId)` to check if a shape can be edited
2. The method should return `false` if:
   - The shape doesn't exist
   - The shape is already being edited
   - The shape's util returns `false` from `canEdit()`
   - The editor is readonly and the shape's util doesn't allow editing in readonly mode
3. The method should return `true` otherwise

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added new `Editor.getCanEditShape` method to check if a shape can be edited.

### API changes

- Added new `Editor.getCanEditShape(shape: TLShape | TLShapeId): boolean` method that returns whether a given shape can be edited.

---

> [!NOTE]
> Introduce `Editor.getCanEditShape` and `Editor.getCanCropShape`, refactor editing/cropping flows to use them, and add `canEditWhileLocked` (incl. embed support).
> 
> - **Editor/Core**:
>   - Add `Editor.getCanEditShape(shape)` and `Editor.getCanCropShape(shape)` with checks for existence, current editing state, util permissions, readonly, and locks (incl. ancestor locks via `canEditWhileLocked`).
>   - Refactor `setEditingShape` to early-return logic and to use `getCanEditShape`; improves cleanup of previous edit state and rich text editor.
>   - Extend `ShapeUtil` with `canEditWhileLocked` hook.
> - **Embeds**:
>   - Add `EmbedDefinition.canEditWhileLocked` and implement `EmbedShapeUtil.canEditWhileLocked` (reads definition); set `canEditWhileLocked: true` for default embed hosts.
> - **Tools/Interactions**:
>   - Replace ad‑hoc checks with `editor.getCanEditShape` / `editor.getCanCropShape` in select, text, geo, arrow, and arrow‑label states; enable Ctrl+resize-handle to enter crop when allowed.
> - **Cleanup**:
>   - Remove `getShouldEnterCropModeOnPointerDown` helper.
> - **API**:
>   - Update API reports to reflect new methods and flags.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dd7323e94c0167d30dc04ed6c8a34dfb3829ba94. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mimecuvalo:
yeah i was gonna originally go with this path but it feels arduous and it's way too easy to forget/footgun-y to do this check.
- on top of this, any prev. code in the wild is certainly not making the right checks (isReadOnly, canEdit)
- it's too easy of a footgun to forget this (our code has sooo many places where we forget to make this check) which is what steered me to doing a `startEditingShape`.

the issue is that you need to do alll this boilerplate to get it right:
```
  if (this.getIsReadonly() && !util.canEditInReadonly(shape)) return false
  if (!util.canEdit(shape)) return false
  this.setEditingShape(shape.id)
  if (!this.getEditingShape()) return false // we're not guarenteed that setEditingShape succeeded
  this.setCurrentTool('select.editing_shape', info) // to pass info
```
and sometimes, `this.emit('select-all-text', { shapeId: shape.id })` after all that.

even if we add `getCanEditShape` it only reduces the boilerplate to:
```
  if (this.editor.canEditShape()) return false
  this.setEditingShape(shape.id)
  if (!this.getEditingShape()) return false // we're not guarenteed that setEditingShape succeeded
  this.setCurrentTool('select.editing_shape') // to pass info
```

so now, you have to have guards/checks on _both_ sides of the setEditingShape, to see if you can edit _and_ to see if it succeeded (which is not guaranteed). in a different naming, instead of `startEditingShape` proposed in the other PR it could be named really something like `tryStartEditingShape` b/c it's kind of like a try/catch in a way.

steveruizok:
Closing in favor of #7298

PR #7298, opened 2025-12-05
feat(api): add canEditShape and canCropShape with edit context

steveruizok:
This PR adds centralized methods for checking shape editability and cropability, along with a new `TLEditStartInfo` type that provides context about how an edit action was triggered.

**New Editor methods:**
- `Editor.canEditShape(shape, info?)` - Type-safe check if a shape can be edited (considers `canEdit`, readonly mode, locked state, and the new `canEditWhileLocked` extension point)
- `Editor.canCropShape(shape)` - Type-safe check if a shape can be cropped (considers `canCrop` and locked state)

**New ShapeUtil extension points:**
- `ShapeUtil.canEditWhileLocked(shape)` - Override to allow editing when locked (e.g., embeds for playing videos)
- `ShapeUtil.canEdit(shape, info)` - Now receives `TLEditStartInfo` context

**TLEditStartInfo types:**
- `click` - Single click (e.g., on already selected shape)
- `click-header` - Click on frame header
- `double-click` - Double-click on shape
- `double-click-handle` - Double-click on shape handle
- `double-click-corner` - Double-click on corner handle
- `press_enter` - Press Enter key

### Change type

- [x] `api`

### Test plan

1. Create a shape (geo, arrow, text, embed)
2. Double-click to edit - should work
3. Lock the shape, double-click - should not edit (except embeds)
4. Add an image, double-click to crop - should work
5. Lock the image, double-click - should not crop
6. Create a frame, click on header to rename - should work
7. Create a frame, press Enter - should NOT edit (frames only edit via header click)

- [x] Unit tests
- [ ] End to end tests

### API changes

- Added `Editor.canEditShape<T>(shape: T | null, info?: TLEditStartInfo): shape is T` - type-safe check if a shape can be edited
- Added `Editor.canCropShape<T>(shape: T | null): shape is T` - type-safe check if a shape can be cropped
- Added `ShapeUtil.canEditWhileLocked(shape: Shape): boolean` - extension point for shapes that can be edited while locked
- Added `TLEditStartInfo` type to describe how shape editing was triggered
- Changed `ShapeUtil.canEdit(shape, info)` to accept optional `TLEditStartInfo` parameter
- Added `canEditWhileLocked?: boolean` property to `EmbedDefinition` interface
- Added `startEditingShapeWithRichText` to public exports
- Frames now only edit when clicking on header (not on Enter key press)

### Release notes

- Added `Editor.canEditShape()` and `Editor.canCropShape()` methods for centralized shape editability/cropability checks.
- Added `ShapeUtil.canEditWhileLocked()` extension point for shapes that can be edited when locked.
- Embed shapes can now be interacted with (e.g., play videos) even when locked.
- Added `TLEditStartInfo` type providing context about how edit actions are triggered.
- Exported `startEditingShapeWithRichText` helper function.

---

> [!NOTE]
> Introduce Editor.canEditShape/canCropShape with TLEditStartInfo and update ShapeUtil and UI/tools to use them; frames edit only via header and embeds can be edited while locked.
> 
> - **Editor API**:
>   - Add `Editor.canEditShape(shape, info?)` and `Editor.canCropShape(shape)` with type guards.
>   - Add `TLEditStartInfo` to describe edit triggers; expose `canUndo/canRedo` (and keep `getCanUndo/getCanRedo`).
>   - Refactor `setEditingShape`/`setCroppingShape` logic to use new guards; minor input manager cleanup.
> - **ShapeUtil API**:
>   - `canEdit(shape, info)`, new `canEditWhileLocked(shape)`; standardize method signatures (remove unused underscored params).
>   - Tighten frames: `FrameShapeUtil.canEdit(shape, info)` only on `click-header`.
> - **Embeds**:
>   - Add `canEditWhileLocked?: boolean` to `EmbedDefinition`; `EmbedShapeUtil.canEditWhileLocked` honors per-embed setting.
>   - Mark default/embed schema providers (tlschema) as editable while locked for supported hosts.
> - **tldraw UI/tools**:
>   - Export `startEditingShapeWithRichText`; replace ad-hoc checks with `editor.canEditShape` / `editor.canCropShape` across tools and states.
>   - Selection/arrow/note/geo/text interactions updated to use new APIs and respect readonly/locked.
> - **Tests**:
>   - Update frame tests: editing via header click only (Enter no longer edits).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 908028fda0f24224890998bd6538baf187958562. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -7767,8 +7767,14 @@ export class Editor extends EventEmitter<TLEventMap> {
 		// then if the shape is flipped in one axis only, we need to apply an extra rotation
 		// to make sure the shape is mirrored correctly
 		if (Math.sign(scale.x) * Math.sign(scale.y) < 0) {
-			let { rotation } = Mat.Decompose(options.initialPageTransform)
-			rotation -= 2 * rotation
+			// We need to compute the new local rotation that will result in the negated page rotation.
+			// For a shape with local rotation `localRot` and parent page rotation `parentRot`:
+			// - pageRot = parentRot + localRot
+			// - newPageRot = -pageRot (we want to negate the page rotation)
+			// - newPageRot = parentRot + newLocalRot (parent hasn't changed)
+			// - Therefore: newLocalRot = -pageRot - parentRot = -(parentRot + localRot) - parentRot = -localRot - 2*parentRot
+			const parentRotation = this.getShapeParentTransform(id).rotation()
+			const rotation = -options.initialShape.rotation - 2 * parentRotation
 			this.updateShapes([{ id, type, rotation }])
 		}
 
@@ -7788,9 +7794,13 @@ export class Editor extends EventEmitter<TLEventMap> {
 		)
 
 		// now calculate how far away the shape is from where it needs to be
-		const pageBounds = this.getShapePageBounds(id)!
 		const pageTransform = this.getShapePageTransform(id)!
-		const currentPageCenter = pageBounds.center
+		// We need to use the local bounds center transformed to page space, not the axis-aligned
+		// page bounds center. This is because the page bounds are axis-aligned and their center
+		// changes when the rotation changes, but we want to use the same reference point as
+		// preScaleShapePageCenter (which used initialBounds.center transformed by the page transform).
+		const currentLocalBounds = this.getShapeGeometry(id).bounds
+		const currentPageCenter = Mat.applyToPoint(pageTransform, currentLocalBounds.center)
 		const shapePageTransformOrigin = pageTransform.point()
 		if (!currentPageCenter || !shapePageTransformOrigin) return this
 		const pageDelta = Vec.Sub(postScaleShapePageCenter, currentPageCenter)
--- a/packages/tldraw/src/test/flipShapes.test.ts
+++ b/packages/tldraw/src/test/flipShapes.test.ts
@@ -637,3 +637,35 @@ it('Updates the image shape flip properties when flipped', () => {
 	editor.flipShapes(editor.getSelectedShapeIds(), 'vertical')
 	expect(editor.getLastCreatedShape<TLImageShape>().props.flipY).toBe(true)
 })
+
+it('Restores flipped shape positions when shape is rotated', () => {
+	editor.selectAll().rotateSelection(PI / 2.5)
+	const before = editor.getSelectedShapes()
+	editor.flipShapes(editor.getSelectedShapeIds(), 'horizontal')
+	editor.flipShapes(editor.getSelectedShapeIds(), 'horizontal')
+	const after = editor.getSelectedShapes()
+	expect(after.length).toBe(before.length)
+	for (let i = 0; i < before.length; i++) {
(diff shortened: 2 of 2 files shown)


mimecuvalo:
i'm not sure if you saw my comment on the other PR before closing, copying here:
yeah i was gonna originally go with this path but it feels arduous and it's way too easy to forget/footgun-y to do this check.

on top of this, any prev. code in the wild is certainly not making the right checks (isReadOnly, canEdit)
it's too easy of a footgun to forget this (our code has sooo many places where we forget to make this check) which is what steered me to doing a startEditingShape.
the issue is that you need to do alll this boilerplate to get it right:
```
  if (this.getIsReadonly() && !util.canEditInReadonly(shape)) return false
  if (!util.canEdit(shape)) return false
  this.setEditingShape(shape.id)
  if (!this.getEditingShape()) return false // we're not guarenteed that setEditingShape succeeded
  this.setCurrentTool('select.editing_shape', info) // to pass info
```
and sometimes, `this.emit('select-all-text', { shapeId: shape.id })` after all that.

even if we add getCanEditShape it only reduces the boilerplate to:
```
  if (this.editor.canEditShape()) return false
  this.setEditingShape(shape.id)
  if (!this.getEditingShape()) return false // we're not guarenteed that setEditingShape succeeded
  this.setCurrentTool('select.editing_shape') // to pass info
```
so now, you have to have guards/checks on both sides of the setEditingShape, to see if you can edit and to see if it succeeded (which is not guaranteed). in a different naming, instead of startEditingShape proposed in the other PR it could be named really something like tryStartEditingShape b/c it's kind of like a try/catch in a way.

steveruizok:
> the issue is that you need to do alll this boilerplate to get it right:

That's not accurate: if `getCanEditShape` passes then `setEditingShape` will succeed, too. The only way for it to fail is if something very sneaky happens like an `onEditEnd` of one shape deleting the newly editing shape, which would produce an error. In that case, all the involved code would be written by the user, though, so I consider that high intent precision foot-gunning or sabatoge.

The actual code would be:

>  if (this.canEditShape(shapeId) {
>   this.setEditingShape(shapeId)
> }

The calling site would still need to handle state transitions (ie moving to select.editing) because the select tool is in the tldraw layer. The select tool doesn't exist in the Editor, so we can't add that there. Same story for `select-all` or other shape-conditional behaviors because those involve tldraw's shapes and so need to happen outside of the editor. Annoying but we want to keep those splits.

Looking at the implementations, I think this is a pretty clean solution—and it works for cropping, too!

mimecuvalo:
discussed offline, we're gonna combine the PR ideas basically from https://github.com/tldraw/tldraw/pull/7276 and this. i'll rebase https://github.com/tldraw/tldraw/pull/7276 on top of this PR

steveruizok:
Following chat with @mimecuvalo, we'll be exporting the ~`startEditingShapeWithLabel`~ `startEditingShapeWithRichText` helper. This helper ensures that the shape has a rich text property and then does the things necessary to start editing that rich text.

steveruizok:
Closing in favor of the even-cleaner #7361

PR #7304, opened 2025-12-05
fix(text): resolve duplicate TextDirection extension and update core

mimecuvalo:
following up after this PR: https://github.com/tldraw/tldraw/pull/7175
one thing i didn't notice was that our custom TextDirection now (sorta) conflicted with the core extension, also named TextDirection, this spit out a warning, not an error. so, nothing broke but we need to clean it up.

this PR was more annoying than i expected to get right, took some iterations, oh well.

### Change type

### Test plan

- [x] Unit tests (if present)

### Release notes

- editing: fix duplicate TextDirection extension, use new core extension

### API changes

- [breaking] adds `attrs` to the rich text validator: properties using the `richTextValidator` now have an optional `attrs` property. A migration might be necessary if you have older clients/custom shapes that won't expect that property to be synced across.
- removes `TextDirection` from our exports, use TipTap's native one instead

---

> [!NOTE]
> Replaces custom TextDirection with TipTap’s core extension, adds `attrs` to rich text with migrations, removes the old export, and upgrades TipTap packages.
> 
> - **Text/Rich Text**:
>   - Replace custom `TextDirection` with TipTap core; remove `lib/utils/text/textDirection` and its public export; re-add via `extensions.TextDirection.configure({ direction: 'auto' })`.
>   - Update `RichTextArea` TipTap config to disable core textDirection in core list and set `textDirection: 'auto'`.
> - **Schema & Migrations**:
>   - Extend `richTextValidator` to allow optional `attrs`.
>   - Add `AddRichTextAttrs` migrations for `text`, `geo`, `note`, and `arrow` shapes (down removes `richText.attrs`); add tests.
> - **Dependencies**:
>   - Upgrade TipTap packages across repo to `^3.12.1` (lockfile updated).
> - **API**:
>   - Remove `TextDirection` from `tldraw` public API.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1ce022ab538e42711d8407a300642c3fd96c83f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9088,273 +9088,273 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@tiptap/core@npm:^3.11.0, @tiptap/core@npm:^3.6.2":
-  version: 3.11.0
-  resolution: "@tiptap/core@npm:3.11.0"
+"@tiptap/core@npm:^3.12.1, @tiptap/core@npm:^3.13.0":
+  version: 3.13.0
+  resolution: "@tiptap/core@npm:3.13.0"
   peerDependencies:
-    "@tiptap/pm": ^3.11.0
-  checksum: 10/e9f23828147dd5167d56a139b9902a2f75d0c0b62723fb19bde58826ca407c656892a3e252d77d4632d34522ffa3adfe2be8e5e77b080758fce4abd770ef0d2c
+    "@tiptap/pm": ^3.13.0
+  checksum: 10/f44f15f6b873eb8a91a38f2987a7a20d90f1746a83c9a5536417ec8b5c15dbd2dec9c7c9b5ab8aaef1ff474eb216517c57531f126ce91e08890f93de425df162
   languageName: node
   linkType: hard
 
-"@tiptap/extension-blockquote@npm:^3.11.0":
-  version: 3.11.0
-  resolution: "@tiptap/extension-blockquote@npm:3.11.0"
+"@tiptap/extension-blockquote@npm:^3.13.0":
+  version: 3.13.0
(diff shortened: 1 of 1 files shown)


mimecuvalo:
ugh, @ds300 can i get your eyes on this? actually i'm thinking i need to write a migration for each shape that uses `richText` b/c now we have a slightly different `richTextValidator`. any way to avoid this?

steveruizok:
> ugh, @ds300 can i get your eyes on this? actually i'm thinking i need to write a migration for each shape that uses `richText` b/c now we have a slightly different `richTextValidator`. any way to avoid this?

Since the property is optional, I believe you would only have to write the "down" migration that removes the attr property for earlier versions.

mimecuvalo:
ok, added the migrations - @ds300 / @MitjaBezensek PTAL

packages/tlschema/src/shapes/TLArrowShape.test.ts
```diff
@@ -492,6 +493,130 @@ describe('TLArrowShape', () => {
 		// so we don't test it
 	})
 
+	describe('arrowShapeMigrations - AddRichTextAttrs migration', () => {
```
steveruizok: We gotta stop the bots from adding these silly migration tests when they're also tested in migrations.test.ts. I'm yanking them all!

steveruizok: (I think they were introduced by me/max when we did the big clauddening)

PR #7361, opened 2025-12-12
feat(editor): add canEditShape and canCropShape methods

steveruizok:
This PR centralizes shape editing and cropping permission logic by introducing two new `Editor` methods that consolidate scattered permission checks into single, consistent APIs.

### Key changes

- **`Editor.canEditShape()`** - Checks if a shape can be edited, considering the shape's `canEdit()` method with `TLEditStartInfo` context, readonly mode, and lock state
- **`Editor.canCropShape()`** - Checks if a shape can be cropped, considering the shape's `canCrop()` method and lock state
- **New `TLEditStartInfo` type** - Describes how an edit was triggered (`press_enter`, `click`, `double-click`, `double-click-corner`, `click-header`, `unknown`)
- **New `ShapeUtil.canEditWhileLocked()`** - Allows shapes like embeds to remain interactive even when locked
- **`FrameShapeUtil.canEdit()` update** - Frames can only be edited via header click, not Enter key
- **Minor cleanup** - Refactored `canUndo`/`canRedo` to use computed pattern, renamed `startEditingShapeWithLabel` to `startEditingShapeWithRichText`

Original implementation: steveruizok/add-get-can-edit-shape-clean

### Change type

- [x] `api`

### Test plan

1. Select a frame shape and press Enter - it should NOT enter edit mode
2. Click on a frame's header - it should enter edit mode
3. Lock an embed shape - it should still be interactive

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added `Editor.canEditShape()` method to centralize shape editing permission checks
- Added `Editor.canCropShape()` method to centralize shape cropping permission checks
- Added `TLEditStartInfo` type to describe how shape editing was triggered
- Added `ShapeUtil.canEditWhileLocked()` method for shapes that remain editable when locked
- Frame shapes can now only be edited by clicking their header, not by pressing Enter

### API changes

- Added `Editor.canEditShape(shape, info?)` method that returns whether a shape can be edited
- Added `Editor.canCropShape(shape)` method that returns whether a shape can be cropped
- Added `TLEditStartInfo` type with `type` field (`'press_enter' | 'click' | 'double-click' | 'double-click-corner' | 'click-header' | 'unknown'`)
- Added `ShapeUtil.canEditWhileLocked(shape)` method (returns `false` by default)
- Updated `ShapeUtil.canEdit(shape, info)` signature to accept `TLEditStartInfo` parameter
- Renamed `startEditingShapeWithLabel` to `startEditingShapeWithRichText` and exported it
- Refactored `Editor.getCanUndo()`/`Editor.getCanRedo()` to `Editor.canUndo()`/`Editor.canRedo()`

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -841,6 +841,10 @@ export class Editor extends EventEmitter<TLEventMap> {
     cancelDoubleClick(): void;
     canCreateShape(shape: OptionalKeys<TLShapePartial<TLShape>, 'id'> | TLShape['id']): boolean;
     canCreateShapes(shapes: (OptionalKeys<TLShapePartial<TLShape>, 'id'> | TLShape['id'])[]): boolean;
+    canCropShape<T extends TLShape | TLShapeId>(shape: null | T): shape is T;
+    canEditShape<T extends TLShape | TLShapeId>(shape: null | T, info?: TLEditStartInfo): shape is T;
+    canRedo(): boolean;
+    canUndo(): boolean;
     // @internal (undocumented)
     capturedPointerId: null | number;
     centerOnPoint(point: VecLike, opts?: TLCameraMoveOptions): this;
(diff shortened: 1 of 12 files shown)


on5sl:
Hi! When would this be released? We are experiencing a bug related to canedit = false but the arrow label is still editable and gives an error. This would fix it we think.

Issue #7377, opened 2025-12-14
Optimize TextManager with batch measurements to reduce layout thrashing


steveruizok:
## Problem

The `TextManager` class in `@tldraw/editor` measures text by setting styles/content on a DOM element and calling `getBoundingClientRect()`. When measuring many shapes sequentially, each measurement triggers a forced layout reflow, causing layout thrashing.

This is one of the most expensive operations in tldraw, particularly noticeable when:
- Resizing many shapes with labels
- Initial render of documents with many text-containing shapes
- Any operation that invalidates text measurements across multiple shapes

## Current behavior

```typescript
// Each call triggers a layout
shapes.forEach(shape => textManager.measureText(shape.text, opts))
// Results in N layouts for N shapes
```

## Proposed solution: Batch measurements

Instead of measuring one-at-a-time, collect all pending measurements and process them together:

### Option A: Explicit batch API

```typescript
interface MeasureRequest {
  text: string
  opts: TLMeasureTextOpts
}

measureTextBatch(items: MeasureRequest[]): Array<BoxModel & { scrollWidth: number }>
```

Implementation approach:
1. Create N measurement divs (or clone from a template element)
2. Set all content/styles in one write pass
3. Append all to a container already in DOM
4. Read all measurements in one read pass
5. Clean up elements

This changes from O(N) layouts to O(1) layouts.

### Option B: Pooled elements with read/write separation

Maintain a pool of measurement elements. When batch measuring:
1. Acquire N elements from pool (create if needed)
2. Write phase: set all innerHTML/styles
3. Read phase: call getBoundingClientRect on all
4. Return elements to pool

### Option C: DocumentFragment approach

```typescript
measureTextBatch(items: MeasureRequest[]) {
  const fragment = document.createDocumentFragment()
  const elements: HTMLDivElement[] = []
  
  // Create and configure all elements
  for (const item of items) {
    const el = this.template.cloneNode(true) as HTMLDivElement
    this.applyStyles(el, item.opts)
    el.innerHTML = item.text
    elements.push(el)
    fragment.appendChild(el)
  }
  
  // Single DOM insertion
  this.container.appendChild(fragment)
  
  // Batch read all measurements
  const results = elements.map(el => {
    const rect = el.getBoundingClientRect()
    return { x: 0, y: 0, w: rect.width, h: rect.height, scrollWidth: el.scrollWidth }
  })
  
  // Cleanup
  elements.forEach(el => el.remove())
  
  return results
}
```

[...]

PR #7378, opened 2025-12-15
perf(geo): optimize label measurement and refactor sizing logic

steveruizok:
This PR improves performance of geo shape label handling by adding caching for expensive label size measurements and refactoring the sizing logic for better maintainability.

The main optimization is using WeakCache to cache label size measurements per shape instance. During resize operations, the code now uses the initialShape (which is stable) for label measurement instead of creating temporary shapes with new dimensions. This prevents expensive measureHtml calls from running on every resize frame.

Original branch: https://github.com/tldraw/tldraw/tree/geo-shape-labels-trace

### Change type

- [x] `improvement`

### Test plan

1. Create a geo shape (rectangle, ellipse, etc.)
2. Add a label to the shape
3. Resize the shape - should feel smooth
4. Try with multiple labeled geo shapes

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Removed `attrs` property from `richTextValidator`
- Removed `AddRichTextAttrs` from `arrowShapeVersions`

### Release notes

- Improved performance when resizing geo shapes with labels by caching label size measurements.
- Fixed redundant offsetWidth call in CenteredTopPanelContainer.

### DIFF
--- a/.claude/commands/analyze-trace.md
+++ b/.claude/commands/analyze-trace.md
@@ -1,0 +1,119 @@
+# Analyze Chrome DevTools trace
+
+Analyze the Chrome DevTools performance trace file: $ARGUMENTS
+
+## Instructions
+
+1. First, find and read the trace file. If no path is given, look for `.json` files in the repo root that look like traces (large files with "trace" in the name or Chrome's default naming pattern).
+
+2. The trace file is likely too large to read directly. Write a temporary Node.js script to analyze it:
+
+```javascript
+// analyze-trace-temp.mjs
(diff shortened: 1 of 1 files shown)


mimecuvalo:
Are we going to do this for each shape separately? I'm curious if we should address the text measurement lower down. I mentioned one strategy with OffscreenCanvas+workers in https://github.com/tldraw/tldraw/issues/7377 - might be worth exploring that in case that hasn't been done before @ tldraw, not sure.

PR #7413, opened 2025-12-17
docs: add comprehensive documentation system

steveruizok:
This PR is an experiment in generating docs with Opus 4.5 / SOTA models. Currently we have only moderate coverage within our official docs and these are quite old. The ideal outcome of course would be for one of us to rent a hotel room and a week's worth of Turkish coffee and write lots and lots of docs, but there are some next-best things.

**Risks**. I have some obvious concerns about publishing AI-generated documentation. It has the same risks as any documentation, such as out-dated advice, errors, or broken code snippets. It does not have the risk of being poorly written from a language and technical perspective, which is something that turns me off from a project's docs, and to erode trust simply by being written by AI.

An ideal outcome—and one being explored by products like Mintlify—would be to have documentation that "auto updates" with each PR, adding or modifying content and ensuring that any changes to the codebase are represented in the documentation. And for this documentation to be accurate, consistent and correct in tone, and genuinely helpful.

Having experimented with this to some extent on a few of our side projects, my impression is that O45 is _very_ good at writing documentation. Significantly better than previous models. It is also very good at updating documentation to match new or changed implementations.

So some questions that would need answering, and that hopefully we can explore in this PR:

- How complete are the current docs?
- How good is Opus 4.5 (or other SOTA models) at writing documentation?
- How good is Opus 4.5 at maintaining documentation?
- How useful are these docs to human readers?
- How useful are these docs to AI agents?
- How much review is needed?

---

This PR introduces a comprehensive documentation system for AI agents and contributors.

### Change type

- [x] `other`

### Test plan

Documentation changes are markdown files that don't require functional testing.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added comprehensive internal documentation system for AI agents and contributors


mimecuvalo:
rm'd this from the PR desc: 
```

### API changes

- **Breaking**: Removed `getStroke` function (use `getStrokePoints` instead)
- **Breaking**: Removed `getStrokeOutlinePoints` function
- **Breaking**: Removed `setStrokePointRadii` function
- **Breaking**: Changed `startEditingShapeWithRichText` signature - now takes `(editor, shape, selectAll?)` instead of `(editor, shapeOrId, options?)`
- Removed `TLEventInfo` re-export from tldraw package
- Added new `TextDirection` export
```

steveruizok:
Yeah, not sure where that came from, thanks.

PR #7418, opened 2025-12-17
perf: introduce FpsScheduler for independent UI and network queues

mimecuvalo:
once more with feeling. the previous attempts were trying to put schedule UI and network events on the same queue, which ... I don't know what we were thinking :P 
Anyway, this creates a proper FpsScheduler class that can take different target rates which solves the issues we were seeing. But also, it solves the fact that even without the 120fps change, we shouldn't be combining these queues.

I recommend reviewing this PR with "hide whitespace" on. you'll note that it had less changes than you would expect. it's really more about just creating a JS class to encapsulate the throttle queue.

this lays the groundwork for the child PR here that does the network bit of this: https://github.com/tldraw/tldraw/pull/7657

previous PRs: https://github.com/tldraw/tldraw/pull/6868 to https://github.com/tldraw/tldraw/pull/6470

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests (if present)

### Release notes

- Improved performance by separating UI and network scheduling queues.

### API changes

- adds `FpsScheduler` to be able to create a FPS-throttled queue of functions to execute

---

> [!NOTE]
> **Introduce per-instance FPS scheduling**
> 
> - Add `FpsScheduler` class in `lib/throttle` with its own queue/state and configurable target FPS; create a default 120fps instance and have `fpsThrottle`/`throttleToNextFrame` delegate to it
> - Export `FpsScheduler` from `utils` (`src/index.ts`); update API report accordingly
> - Add comprehensive unit tests (`lib/throttle.test.ts`) covering throttling, next-frame batching, cancelation, and real-world scenarios
> - UI tweak: `DefaultDebugPanel` FPS readout now includes `maxKnownFps` (`FPS ${fps} (max: ${maxKnownFps})`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 871ad9510e89bab7336a6c29d8cb90cfea633c37. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/utils/src/lib/throttle.test.ts
+++ b/packages/utils/src/lib/throttle.test.ts
@@ -45,54 +45,6 @@ describe('FpsScheduler class', () => {
 		}
 	}
 
-	describe('separate instances have separate queues', () => {
-		it('should maintain separate queues for different instances', () => {
-			const throttle1 = new FpsScheduler(60)
-			const throttle2 = new FpsScheduler(30)
-
-			const fn1 = vi.fn()
-			const fn2 = vi.fn()
-
-			// throttleToNextFrame directly queues the function and returns a cancel function
-			throttle1.throttleToNextFrame(fn1)
-			throttle2.throttleToNextFrame(fn2)
-
-			// Neither should be called yet
-			expect(fn1).not.toHaveBeenCalled()
-			expect(fn2).not.toHaveBeenCalled()
-
-			// Trigger animation frames
-			flushAnimationFrames()
-
-			// Both should be called
-			expect(fn1).toHaveBeenCalledTimes(1)
-			expect(fn2).toHaveBeenCalledTimes(1)
-		})
-
-		it('should not mix functions between different instances', () => {
-			const throttle1 = new FpsScheduler(120)
-			const throttle2 = new FpsScheduler(120)
-
-			const fn1 = vi.fn()
-			const fn2 = vi.fn()
-
-			const throttled1 = throttle1.fpsThrottle(fn1)
-			const throttled2 = throttle2.fpsThrottle(fn2)
-
-			// Call multiple times
-			throttled1()
-			throttled1()
-			throttled2()
-			throttled2()
-
-			flushAnimationFrames()
-
-			// Each function should only execute once (throttled)
-			expect(fn1).toHaveBeenCalledTimes(1)
-			expect(fn2).toHaveBeenCalledTimes(1)
-		})
-	})
-
 	describe('FPS throttling', () => {
 		it('should throttle to the target FPS', () => {
 			const throttle = new FpsScheduler(60) // ~16.67ms per frame, ~15ms with variance
--- a/packages/utils/src/lib/throttle.ts
+++ b/packages/utils/src/lib/throttle.ts
@@ -8,7 +8,7 @@ const isTest = () =>
 // by aiming slightly below the theoretical frame duration (e.g., ~7.5ms instead of 8.33ms for 120fps)
 const timingVarianceFactor = 0.9
 const getTargetTimePerFrame = (targetFps: number) =>
-	Math.floor(1000 / targetFps) * timingVarianceFactor // ~7ms
+	Math.floor(1000 / targetFps) * timingVarianceFactor
 
 /**
  * A scheduler class that manages a queue of functions to be executed at a target frame rate.
@@ -72,7 +72,7 @@ export class FpsScheduler {
 			// eslint-disable-next-line no-restricted-globals
 			this.flushRaf = requestAnimationFrame(() => {
 				this.flushRaf = undefined
-				this.lastFlushTime = now
+				this.lastFlushTime = Date.now()
 				this.flush()
 			})
 		}
@@ -181,11 +181,6 @@ const defaultScheduler = new FpsScheduler(120)
  *
  * // Cancel pending calls if needed
  * updateCanvas.cancel?.()
- *
- * // Custom FPS throttling for less critical updates
- * const slowUpdate = fpsThrottle(() => {
- *   heavyComputation()
- * }, () => 30) // Throttle to 30fps
  * ```
  *
  * @internal


steveruizok:
Could you separate the "solo mode" improvements from the "120fps tick" improvements? The 120fps change is going to produce some problems across different browsers and may require us to change the way that we handle draw shapes. (Same bugs we ran into last few times we tried this, or with coalescing). But I would _love_ to save the CPU cost of "syncing with myself" when working alone on tldraw.com.

mimecuvalo:
> Could you separate the "solo mode" improvements from the "120fps tick" improvements? The 120fps change is going to produce some problems across different browsers and may require us to change the way that we handle draw shapes. (Same bugs we ran into last few times we tried this, or with coalescing). But I would _love_ to save the CPU cost of "syncing with myself" when working alone on tldraw.com.

sounds good 👍 

@MitjaBezensek @steveruizok i moved the solo mode changes into a separate PR (child branch of this PR): https://github.com/tldraw/tldraw/pull/7657

packages/utils/src/lib/throttle.ts
```diff
@@ -4,73 +4,188 @@ const isTest = () =>
...
-
-const flush = () => {
-	const queue = fpsQueue.splice(0, fpsQueue.length)
-	for (const fn of queue) {
-		fn()
+// Browsers aren't precise with frame timing - this factor prevents skipping frames unnecessarily
+// by aiming slightly below the theoretical frame duration (e.g., ~7.5ms instead of 8.33ms for 120fps)
+const timingVarianceFactor = 0.9
+const getTargetTimePerFrame = (targetFps: number) =>
+	Math.floor(1000 / targetFps) * timingVarianceFactor // ~7ms
```
MitjaBezensek:
```suggestion
	Math.floor(1000 / targetFps) * timingVarianceFactor
```
I'd remove the comment as the exact number depends on the passed in param.

packages/utils/src/lib/throttle.test.ts
```diff
@@ -0,0 +1,491 @@
...
+		while (rafCallbacks.length > 0 && iterations < maxIterations) {
+			const callbacks = [...rafCallbacks]
+			rafCallbacks = []
+			callbacks.forEach((cb) => cb(performance.now()))
+			iterations++
+		}
+	}
+
+	describe('separate instances have separate queues', () => {
+		it('should maintain separate queues for different instances', () => {
```
MitjaBezensek: Not sure this tests what is says it tests? How do we know that we maintain separate queues for different instances? I guess we could add the same function to different instances and then expect it gets called multiple times.

packages/utils/src/lib/throttle.test.ts
```diff
@@ -0,0 +1,491 @@
...
+
+			// Trigger animation frames
+			flushAnimationFrames()
+
+			// Both should be called
+			expect(fn1).toHaveBeenCalledTimes(1)
+			expect(fn2).toHaveBeenCalledTimes(1)
+		})
+
+		it('should not mix functions between different instances', () => {
```
MitjaBezensek: Also not sure about this one? This looks like it tests that throttling works, unsure about the instances logic?

packages/utils/src/lib/throttle.ts
```diff
@@ -4,73 +4,188 @@ const isTest = () =>
...
+			if (this.flushRaf) return // ...though if there's a flush raf, that means we'll be flushing on this frame already, so we can do nothing here.
+			this.lastFlushTime = now
+			this.flush()
+		} else {
+			// If we haven't already waited for the next frame to run the tick, we need to wait until the next frame to flush.
+			if (this.flushRaf) return // ...though if there's a flush raf, that means we'll be flushing on the next frame already, so we can do nothing here.
+			// eslint-disable-next-line no-restricted-globals
+			this.flushRaf = requestAnimationFrame(() => {
+				this.flushRaf = undefined
+				this.lastFlushTime = now
```
MitjaBezensek: Think now here will be stale as it is inside the raf, we probably need to get a new value?

packages/utils/src/lib/throttle.ts
```diff
@@ -4,73 +4,188 @@ const isTest = () =>
...
+ * // Call as often as you want - automatically throttled to 120fps
  * document.addEventListener('mousemove', updateCanvas)
  *
  * // Cancel pending calls if needed
  * updateCanvas.cancel?.()
+ *
+ * // Custom FPS throttling for less critical updates
+ * const slowUpdate = fpsThrottle(() => {
+ *   heavyComputation()
+ * }, () => 30) // Throttle to 30fps
```
MitjaBezensek: Seems a stale comment, we only accept a single param?

packages/utils/src/lib/throttle.ts
mimecuvalo: oh man this is a good catch. this isn't new in this PR, this was an issue an the existing code! will fix up

packages/utils/src/lib/throttle.test.ts
mimecuvalo: not gonna lie, this test was vibed 🙃

mimecuvalo: nixed

packages/utils/src/lib/throttle.test.ts
mimecuvalo: gonna nix this one - it's pretty well covered by the one below: "should respect different FPS settings"

random-yang:
Does this mean I can unlock the 120 Hz TLdraw canvas experience?

mimecuvalo:
> Does this mean I can unlock the 120 Hz TLdraw canvas experience?

Yes, this will be available in the SDK in the next release.
This will be available on tldraw.com in a week or so.

PR #7441, opened 2025-12-24
feat(editor): add WebGL-based shape indicators for improved performance

steveruizok:
This PR introduces a WebGL-based indicator rendering system as an alternative to the current SVG-based approach. The new system renders selection, hover, and hint indicators using WebGL triangles for better performance with many selected shapes.

## Key changes

- **WebGLIndicatorManager**: Core class that manages the WebGL rendering lifecycle, geometry caching, and dirty-marking system
- **WebGLShapeIndicators**: React component that mounts the WebGL canvas
- **Geometry triangulation**: Converts shape outlines to triangle strips for GPU rendering
- **Shader-based stroke width**: Stroke width is applied in the vertex shader, allowing geometry to be cached independently of zoom level
- **Dirty-marking system**: Shapes are marked dirty when they change, and geometry is only recomputed when indicators are actually displayed

## Technical details

- Geometry format: Each vertex stores `(centerX, centerY, offsetX, offsetY)` where offset is the normalized perpendicular direction
- The shader computes: `screenPos = (pagePos + camera.xy) * camera.z` then applies stroke offset in screen space
- Geometry cache is zoom-independent since stroke width is a uniform, not baked into vertices

### Change type

- [x] `improvement`

### Test plan

1. Create multiple shapes on the canvas
2. Select shapes and verify indicators render correctly
3. Zoom in/out and pan - indicators should follow shapes correctly
4. Hover over unselected shapes - hover indicator should appear

- [ ] Unit tests (if present)

### Release notes

- Added experimental WebGL-based shape indicator rendering for improved performance

---

> [!NOTE]
> Replaces the SVG indicator path with a performant WebGL pipeline and adds a dedicated indicator-geometry API.
> 
> - Adds `WebGLIndicatorManager` and `WebGLShapeIndicators` to render selection/hover indicators via WebGL (triangulated outlines, shader-based stroke width, color updates)
> - Integrates WebGL indicators into default canvas/components (replaces `DefaultShapeIndicators`; updates overlay ordering) and exports in `index.ts`
> - Adds `Editor.getShapeIndicatorGeometry` and `ShapeUtil.getIndicatorGeometry` (default = `getGeometry`), enabling shape-specific indicator outlines
> - Implements `getIndicatorGeometry` overrides for arrow, draw, frame, geo, highlight, and note shapes; adds smoothed stroke sampling for freehand
> - Introduces WebGL geometry/GL setup helpers (`indicator-webgl-geometry`, `indicator-webgl-setup`); caches geometry with dirty-marking and viewport-reactive rendering
> - Updates API reports/types (`TLWebGLShapeIndicatorsProps`, `WebGLIndicatorManager`, `WebGLShapeIndicators`) and switches app usage (`TldrawShapeIndicators`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2145be1d6ab07143cdbead42b032f12209941ba5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -1912,6 +1912,8 @@ export class HighlightShapeUtil extends ShapeUtil<TLHighlightShape> {
     // (undocumented)
     getGeometry(shape: TLHighlightShape): Circle2d | Polygon2d;
     // (undocumented)
+    getIndicatorGeometry(shape: TLHighlightShape): Circle2d | Polyline2d;
+    // (undocumented)
     getInterpolatedProps(startShape: TLHighlightShape, endShape: TLHighlightShape, t: number): TLHighlightShapeProps;
     // (undocumented)
     hideResizeHandles(shape: TLHighlightShape): boolean;
--- a/packages/tldraw/src/lib/shapes/highlight/HighlightShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/highlight/HighlightShapeUtil.tsx
@@ -3,6 +3,7 @@ import {
 	Circle2d,
 	Editor,
 	Polygon2d,
+	Polyline2d,
 	SVGContainer,
 	ShapeUtil,
 	TLHighlightShape,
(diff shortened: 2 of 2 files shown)


steveruizok:
This is a spike PR but I think it has promise.

Our current indicators are pretty annoying: in order to prevent mount / unmount performance hits, we leave all indicators in the tree and hide or show them. However it still adds a lot of latency when duplicating shapes / deleting them, etc. And while we try not to over-render, even the cost of evaluating hooks on indicators gets slower on very large documents.

This strategy would reduce the cost to render to basically zero. We could cache indicator geometry in page space for all shapes and then transform the geometry while we render it. (The transforms themselves are very very fast, we likely wouldn't need to cache this).

steveruizok:
This PR uses the shape geometry to show indicators. We'd want to let shapes describe their own indicators. I don't think I've ever seen any shape that really uh innovates on indicators so hopefully the migration path would be easy.

steveruizok:
Going to try this again with canvas to benefit from regular line-width / Path2D caching.

PR #7506, opened 2025-12-29
fix(editor): handle fragment-only URLs in SVG export

PidgeyBE:
SVGs support some default markers e.g.
```
marker-end="url(#arrow)"
```
see https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/marker

This is bugged in TLDraw png/svg export though because all `url(...)`'s references are fetched and injected into svgs while exporting.
`#arrow` isn't a valid URL so the marker doesn't get exported correctly.

This PR fixes that.

### Change type

- [x] `bugfix` 

### Test plan

1. Put an SVG on tldraw canvas, using one of the "default supported shapes" in svg, like #arrow. E.g. Used in a mermaid sequencdiagram
2. Export tldraw canvas to svg/png. 
3. Bug: Arrows got lost

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed export of svg markers (https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/marker)

---

> [!NOTE]
> Ensures fragment-only CSS `url()` values aren’t treated as external resources during export.
> 
> - Update `parseCssValueUrls` to filter out URLs starting with `#` (e.g., `url(#arrowhead)`), preventing incorrect fetching/injection
> - Add unit test in `parseCss.test.ts` validating `url(#arrowhead)` returns no matches
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fe99aeeb31e90f7e63ec3d3a332a9857de7e7a62. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


steveruizok:
@claude run lint

steveruizok:
I can't replicate the original bug. Do you happen to have an SVG image that will reproduce this?

PidgeyBE:
> I can't replicate the original bug. Do you happen to have an SVG image that will reproduce this?

@steveruizok  sorry for the late reply.

I couldn't reproduce the issue using the `toSvg` method on ShapeUtils, but it is reproducible with the `component()` method.
E.g. If you make a shapeUtil with 
```
    override component(shape): React.ReactElement {
        return (
            <svg viewBox="0 0 300 100" xmlns="http://www.w3.org/2000/svg">
  <defs>
    <marker
      id="arrow"
      viewBox="0 0 10 10"
      refX="5"
      refY="5"
      markerWidth="6"
      markerHeight="6"
      orient="auto-start-reverse">
      <path d="M 0 0 L 10 5 L 0 10 z" />
    </marker>
  </defs>

  <line
    x1="10"
    y1="10"
    x2="90"
    y2="90"
    stroke="black"
    marker-end="url(#arrow)" />

  <path
    d="M 110 10
       C 120 20, 130 20, 140 10
       C 150 0, 160 0, 170 10
       C 180 20, 190 20, 200 10"
    stroke="black"
    fill="none"
    marker-start="url(#arrow)"
    marker-mid="url(#arrow)"
    marker-end="url(#arrow)" />
</svg>
        );
    }
```

The arrows don't get rendered correctly.
image


The reason we dont use useSvg directly is because our components are a combination of svg images and other html.

steveruizok:
Ok, great. I was trying with SVG images, though those would have been handled as .svg assets rather than inline SVGs.

steveruizok:
Landing. Thanks @PidgeyBE !

PR #7527, opened 2026-01-01
feat(editor): add onDropOnCanvas option for custom drop handling

steveruizok:
In order to give developers full control over drop event handling, this PR adds a new `onDropOnCanvas` option to the editor that intercepts all canvas drop events. This is an escape hatch feature.

## Design

The callback receives the page coordinates where content was dropped and the original drag event:

```typescript
onDropOnCanvas?(options: { point: VecLike; event: React.DragEvent<Element> }): boolean
```

Returning `true` prevents default file and handling (by the external content API), while returning `false` or `undefined` allows the built-in handlers to process the drop. 

## Implementation details

- Added `onDropOnCanvas` to `TLEditorOptions` interface
- Stored as private `_onDropOnCanvas` property on the Editor instance, following the same pattern as `getShapeVisibility`
- Integrated into the drop event flow in `useCanvasEvents` - called first, allows returning boolean to control whether default handlers run
- The callback has access to the raw drag event for maximum flexibility, while keeping the API clean with the page position already converted

## Motivation

This allows developers to:
- Intercept and handle drops of custom data types
- Prevent default handling when needed
- Access the full drag event details for complex scenarios

### Change type

- [x] `feature`

### API changes

- Added new `onDropOnCanvas` option to `TLEditorOptions` that allows developers to intercept canvas drop events and control whether default file/URL handling occurs

---

> [!NOTE]
> **Medium Risk**
> Changes the core canvas drop event path; incorrect callback usage or return values could unintentionally bypass default file/URL drop behavior.
> 
> **Overview**
> Adds a new `experimental__onDropOnCanvas` editor option that runs on canvas drop events with `{ point, event }` and can return `true` to short-circuit the built-in drop handlers.
> 
> Updates the canvas `onDrop` flow to compute the drop page point once, invoke the callback before default file/URL handling, and wires the option into `defaultTldrawOptions` and the generated API report.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dcc4c1237c7765afcd5a77beffcfef990270251a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/options.ts
+++ b/packages/editor/src/lib/options.ts
@@ -1,6 +1,7 @@
 import { ComponentType, Fragment } from 'react'
 import { DEFAULT_CAMERA_OPTIONS } from './constants'
 import { TLCameraOptions } from './editor/types/misc-types'
+import { VecLike } from './primitives/Vec'
 import { TLDeepLinkOptions } from './utils/deepLinks'
 import { TLTextOptions } from './utils/richText'
 
@@ -155,6 +156,15 @@ export interface TldrawOptions {
 	 * viewport's page dimensions regardless of overview zoom changes.
 	 */
 	readonly quickZoomPreservesScreenBounds: boolean
+	/**
+	 * Called when content is dropped on the canvas. Provides the page position
+	 * where the drop occurred and the underlying drag event object.
+	 * Return true to prevent default drop handling (files, URLs, etc.)
+	 */
+	experimental__onDropOnCanvas?(options: {
+		point: VecLike
+		event: React.DragEvent<Element>
+	}): boolean
 }
 
 /** @public */
@@ -217,4 +227,5 @@ export const defaultTldrawOptions = {
 	text: {},
 	deepLinks: undefined,
 	quickZoomPreservesScreenBounds: true,
+	experimental__onDropOnCanvas: undefined,
 } as const satisfies TldrawOptions


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -253,6 +254,12 @@ export interface TLEditorOptions {
 		shape: TLShape,
 		editor: Editor
 	): 'visible' | 'hidden' | 'inherit' | null | undefined
+	/**
+	 * Called when content is dropped on the canvas. Provides the page position
+	 * where the drop occurred and the underlying drag event object.
+	 * Return true to prevent default drop handling (files, URLs, etc.)
+	 */
+	onDropOnCanvas?(options: { point: VecLike; event: React.DragEvent<Element> }): boolean
```
mimecuvalo: mm i don't think we should make a one-off event handler on `TLEditorOptions` — this would make it too bifurcated for folks to know where event handlers should be wired into tldraw

steveruizok: IMO options is acceptable as a "least worst bad place" for things. I personally don't like options at all but we did make a design decision to centralize our "configuration debt" into editor.options. Lets try to come up with a better place using existing APIs, but if we can't then I'd rather options -> wait for obvious place or second related thing (e.g. oh we need an `onDropOnToolbar` too?) -> make that design.

mimecuvalo: heard. let's still have a discussion, it's worth having a couple minds in the room to design this in a way that will rhyme with other similar items.

steveruizok: moving it into options, calling experimental

steveruizok:
Clean copied this bad boy onto an option. I need to remember where we keep the skeletons in the SDK

Issue #6164, opened 2025-06-01
Add scroll wheel zoom direction preference


visnup:
### What's the feature?

For whatever reason, my muscle memory wants scrolling up on the scroll wheel (actually, Magic Mouse on a Mac) to zoom into the canvas (objects get larger). I know everyone has their own preference though, so I'd suggest just adding a user preference that is "Invert zoom direction".

I think I'm used to it from that pinching out causes my index finger to move up, which is the same finger that would move the "scroll wheel" on a magic mouse. For what it's worth, the Maps app on macOS agrees with me, though Photos does not.

### Where would you like this feature?

The developer package

### Contact Details

visnup@gmail.com

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

visnup:
Would still enjoy this small feature

steveruizok:
This issue has been reopened as it remains unaddressed.

## Summary

The feature requested in this issue has **not been implemented** despite comprehensive camera options being added in PR #3282.

## Analysis

The current `TLCameraOptions` interface includes controls for:
- `panSpeed`: Adjust pan sensitivity
- `zoomSpeed`: Adjust zoom sensitivity  
- `wheelBehavior`: Choose between zoom, pan, or none

However, **there is no `invertZoomDirection` or equivalent option** to reverse the scroll zoom direction.

## Current behavior
- Scrolling UP on scroll wheel: Zooms OUT
- Scrolling DOWN on scroll wheel: Zooms IN

## Requested behavior
- Option to INVERT this direction (like macOS Maps, Apple Photos)

## Why this matters
The current behavior doesn't match the muscle memory of users familiar with:
- macOS Maps and Photos apps
- Pinch gestures on touch devices (pinching out = index finger moving up)
- Many mapping applications

## Implementation Path
1. Add `invertZoomDirection?: boolean` to `TLCameraOptions`
2. Update zoom calculation logic in `Editor.ts` to respect this option
3. Add tests and documentation

This is a reasonable UX preference that aligns with established patterns in other applications.

PR #7552, opened 2026-01-02
feat(editor): add zoom direction preference

kostyafarber:
Some users (#6164) prefer inverted scroll wheel zoom direction - scrolling up to zoom out and down to zoom in (similar to macOS "natural" scrolling). This PR adds an `isZoomDirectionInverted` user preference that allows users to toggle this behavior.

### Change type

- [x] `feature`

### Test plan

1. Open the main menu and navigate to Preferences
2. Toggle "Invert zoom" on
3. Scroll on the canvas - zoom direction should now be inverted (scroll up to zoom out, scroll down to zoom in)
4. Toggle "Invert zoom" off
5. Verify zoom direction returns to normal (scroll up to zoom in, scroll down to zoom out)

- [x] Unit tests

### API changes

- Added `isZoomDirectionInverted` property to `TLUserPreferences`
- Added `getIsZoomDirectionInverted()` method to `UserPreferencesManager`
- Added `ToggleInvertZoomItem` UI component for use in menus

### Release notes

- Added a new "Invert zoom" preference that allows users to invert the scroll wheel zoom direction.

---

> [!NOTE]
> Adds a user preference to invert mouse wheel zoom and wires it through storage, editor behavior, UI, and i18n.
> 
> - **Editor behavior**: Applies inverted zoom in `Editor` wheel handling when `inputMode === 'mouse'` and `isZoomDirectionInverted` is true; exposes `getIsZoomDirectionInverted()` via `UserPreferencesManager`; updates defaults, schema validation, migrations, and versioning.
> - **Persistence & schema**: New `isZoomDirectionInverted` on `TLUserPreferences` and DB (`030_add_zoom_direction_preference.sql`); sync worker SQL and column maps updated; test helpers default updated.
> - **UI**: Adds `ToggleInvertZoomItem` under `Input device` submenu; disables toggle unless mouse is selected; styling updated to respect `[data-disabled]`.
> - **Translations**: New strings for invert zoom and rename menu label from `menu.input-mode` to `menu.input-device` across locales.
> - **Tests & API reports**: Unit tests for preferences manager and camera zoom behavior; API reports updated to include new keys and menu item.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1571510e8a7e250e9684e76a3381a5e43a5297c2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mimecuvalo:
it's crazy how many files it takes to add a pref 🙃

steveruizok:
Kind of not sure about this one, I think its significant that I haven't heard this request in the last five years. Not sure I'd add a preference for this if the convention is very strong.

steveruizok:
I'd be okay having this in the SDK as a developer option, just thinking about the preferences menu.

kostyafarber:
> Kind of not sure about this one, I think its significant that I haven't heard this request in the last five years. Not sure I'd add a preference for this if the convention is very strong.

yeah thats fair this did come off of a request here #6164. fwiw this is an option in Figma on Desktop. We could add it to the sdk but I don't feel crazy strongly about it either way.

steveruizok:
Aha, well that might be enough evidence for me.

steveruizok:
What do you think of this? Since we have an input dropdown here: 
image

steveruizok:
We might have more (e.g. disabling auto pen mode is another one I sometimes get asked about)

kostyafarber:
> We might have more (e.g. disabling auto pen mode is another one I sometimes get asked about)

ah so you think we move all these input-y ones if and when they come in to that input mode submenu. 

The pen mode one, in that submenu makes sense to me, but zoom direction I dunno. We do only zoom with input devices, but I would not naturally look for this option in an input devices menu. I'd probably heuristically think it would exist at the top level menu.

steveruizok:
It looks like Figma started in a direction of putting input stuff together and should have included zoom direction here, too.

image

steveruizok:
Hm, invert zoom also inverts the panning on a trackpad, which is wrong. This really is "invert wheel zoom direction" and only have an effect if mousewheel zoom is on. Which also makes me think it should be placed next to that control.

kostyafarber:
Ok fair, good catch, it should only affect zooming. I didn't appreciate that mouse wheel zoom option, makes sense to put it with input devices then! I'll take a look.

kostyafarber:
cc @steveruizok 
* we only invert zoom if user has specified mouse mode as their preference
* changed the menu parent name to `Input device`
* moved `Invert zoom` into that menu per your image
* added some tests

kostyafarber:
closed in favour of the clean pr

PR #7592, opened 2026-01-05
Reduce culling recalculation frequency with expanded bounds

### LINKED ISSUE #7436: Explore hysteresis strategy for shape culling
## Summary

Investigate a hysteresis-based optimization for shape culling to reduce computation frequency during panning and zooming operations.

## Current behavior

The `notVisibleShapes` computed value in `packages/editor/src/lib/editor/derivations/notVisibleShapes.ts` recomputes whenever the viewport bounds change. During pan/zoom operations, this means culling checks run on every frame where the camera moves, even for small movements.

The current algorithm:
1. Iterates through all shapes on the page
2. Checks each shape's bounds against viewport bounds (O(n) per viewport change)
3. Updates the culled set when shapes enter/exit the viewport

While recent optimizations (commit 2293c7f) improved per-shape check performance, the computation still runs frequently during continuous pan/zoom.

## Proposed optimization

Implement a **hysteresis strategy** that reduces culling recalculation frequency:

### Core concept

1. **Expanded bounds**: Instead of testing against exact viewport bounds, test against expanded "hysteresis bounds" that are larger than the viewport by some margin (e.g., 20% on each side)

2. **Deferred recalculation**: Only recalculate culling when the actual viewport exits the previously computed hysteresis bounds

3. **Reset on significant change**: When recalculation triggers, compute new expanded bounds around the current viewport

### Visual explanation

```
┌──────────────────────────────────┐
│     Hysteresis bounds            │
│   ┌────────────────────────┐     │
│   │   Actual viewport      │     │
│   │                        │     │
│   │                        │     │
│   └────────────────────────┘     │
│                                  │
└──────────────────────────────────┘
```

During pan/zoom:
- Shapes inside hysteresis bounds are rendered (slightly more than visible)
- Culling only recalculates when viewport edge touches hysteresis edge
- This reduces O(n) calculations from every-frame to periodic

### Trade-offs

| Benefit | Cost |
|---------|------|
| Fewer culling recalculations during pan/zoom | Slightly more shapes rendered at any time |
| Smoother performance during navigation | Additional memory for tracking hysteresis state |
| Reduced reactive cascade frequency | Complexity in managing the hysteresis bounds |

### Implementation considerations

1. **Margin calculation**: Could be fixed pixels, percentage of viewport, or dynamic based on zoom level

2. **State management**: Need to track:
   - Current hysteresis bounds
   - When to invalidate/recalculate

3. **Integration points**:
   - Modify `notVisibleShapes.ts` to use hysteresis bounds
   - May need new atoms/signals to track hysteresis state
   - Consider whether to apply to both pan and zoom, or just pan

4. **Edge cases**:
   - Initial load (no previous hysteresis bounds)
   - Page changes (reset hysteresis)
   - Window resize (recalculate hysteresis)
   - Shape additions/deletions (may need partial update)

## Questions to investigate

- What margin size provides the best balance between render overhead and recalculation frequency?
- Should the margin be zoom-dependent (larger margin at high zoom)?
- How does this interact with the existing incremental computation?
- Are there cases where we'd want to disable hysteresis (e.g., very large documents)?

## Related files

- `packages/editor/src/lib/editor/derivations/notVisibleShapes.ts` - Main culling logic
- `packages/editor/src/lib/editor/Editor.ts` - `getViewportPageBounds()` and related methods
- `packages/editor/src/lib/components/Shape.tsx` - DOM application of culling
- `packages/tldraw/src/test/getCulledShapes.test.tsx` - Test suite

## Possible benchmarks

1. Measure culling recalculation frequency during typical pan gestures
2. Compare frame times with/without hysteresis during continuous zoom
3. Test on documents with varying shape counts (100, 1000, 10000 shapes)

MitjaBezensek:
Reduces culling recalculation frequency during pan/zoom by using expanded culling bounds with 20% margin around the viewport. Shapes are checked against these expanded bounds instead of exact viewport bounds. Culling bounds only recalculate when:
  1. Viewport moves outside existing culling bounds
  2. Bounds become too large after zooming in (>2x area of the regular culling bounds)
  3. Page changes

  **Implementation:**
  - `_cullingBounds`: Computed (uses `prevValue`) to maintain expanded bounds
  - Returns cached bounds when viewport still inside, reducing recomputation
  - Automatically tracks page changes via reactive dependencies
  - Handles zoom-in by recalculating when new bounds would be <50% of current area

Fixes #7436 

### Change type

- [x] `improvement`

### Release notes

- Improve performance of zooming and panning by reducing the frequency of culled shape calculation.
- Introduce an Editor option called `cullingMargin` which controls how big of a margin around the viewport do we wish to use when culling. This improves performance of certain actions like panning and zooming.

### Api changes
* Introduce an internal computed (`_cullingBounds ` for tracking the culled bounds.
* Introduce an Editor option called `cullingMargin` which controls how big of a margin around the viewport do we wish to use when culling. This improves performance of certain actions like panning and zooming.

---

> [!NOTE]
> Improves pan/zoom performance by culling against expanded bounds instead of exact viewport.
> 
> - Introduces internal computed `Editor._cullingBounds` and `derive/calculateCullingBounds` to cache expanded viewport bounds
> - Adds `cullingMargin` option (default `0.2`) to `TldrawOptions` and `defaultTldrawOptions` (reflected in `api-report.api.md`)
> - Updates `notVisibleShapes` to use `_cullingBounds` for visibility checks
> - Adds/updates tests (`getCulledShapes.test.tsx`) to validate behavior, reduced recomputation, and custom `cullingMargin`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5155ce4adcfc7d8e6c99c6aaa262bb53cbc6b62. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -2740,6 +2742,29 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		this.getCurrentPageId()
+		const viewport = this.getViewportPageBounds()
+
+		// Recalculate if first run or viewport moved outside bounds
+		// (page changes trigger re-run via getCurrentPageId dependency)
+		if (isUninitialized(prev) || !prev || !prev.contains(viewport)) {
+			return calculateCullingBounds(viewport)
+		}
+
+		// Also recalculate if bounds are too large (e.g., after zooming in)
```
MitjaBezensek: If we didn't do this we would get stuck on overly large culled bounds after zooming out, thus not using the culling logic at all.

steveruizok:
swooping by to extract the constant into an editor option

steveruizok:
Can you explain more about what happens when a user zooms? Zooming out makes sense, because the zoom will trigger new bounds as the bounds no longer contain the frame. Zooming in I'm less sure of because the viewport will remain inside of the culled bounds. Maybe we should trigger a fresh culling bounds anyway when the camera mode switches back to idle (since that's often a good time to drop a frame if we have to).

edit is this what you mean by "If we didn't do this we would get stuck on overly large culled bounds after zooming out, thus not using the culling logic at all."?

MitjaBezensek:
### Zooming out
When you zoom out we will reach a point when the viewport is no longer inside the culling bounds, which will trigger an update to the culling bounds. This will then trigger a not visible shapes recalculation.

### Zooming in
The above logic does not work when zooming in as the viewport will remain in the culling bounds when zooming in. This is fine, but it can defeat the whole purpose of culling if we don't update the culling bounds at a certain point. Imagine this:

1. We zoom out as much as possible.
2. This would make the culling bounds really large.
3. When then zoom in for regular work at 100%.
4. The culling logic might not cull any shapes as the bounds are really large compared to the current viewport.

At a certain point (right now I set it to twice the regular culling margin) we want to update the culling bounds.

CleanShot 2026-01-06 at 09 57 16@2x

You can also think of this as having an outer cutoff and an inner cutoff:
CleanShot 2026-01-06 at 10 07 45@2x

So when our margins increase (which happens when zooming in) will we trigger an update to culling bounds. And this also allows some leeway (we don't update on every zoom in), similarly to how we only update when zooming out only when you reach the edges. Small zooms should not trigger it.

steveruizok:
> snip

Love this! Could go right into a twitter thread, too.

I think the .2 default might be too cautious, I might suggest doing 1x or even 2x, since the cost of rendering more things is likely lower than the cost to evaluate. I'll test a bit and see when / where I notice.

mimecuvalo:
Checking in on this one, any more feedback before landing?

MitjaBezensek:
> Checking in on this one, any more feedback before landing?

I'm not sure if we should even land this. The reason being is that this works in the following way:
1. While we are inside the expanded bounds the perf is good. We don't change the visibility of shapes, but it does mean that we show more shape than we would have to due to this expansion. This means that small changes in viewport don't update any shapes, but we do show more shapes (by show I mean more shapes that don't have `display: none` set).
2. As soon as we go outside it usually means a big change in dom (a bunch of shapes might go from being hidden to now shown and vice versa). This can actually cause a bit of a long frame when it happens. With our regular logic we do many, but smaller changes, but here it's fewer, but bigger changes.

So it's hard to say if this is an improvement. Depends on how you use the space. If you are staying in a limited area then it might be worth it (there's still a penalty of showing more shapes), but if you significantly pan around then I don't think it's worth it. Would be interesting to understand user behaviour to better understand whether this is worth it or not.

steveruizok:
Closing!

PR #7602, opened 2026-01-05
feat(editor): add zoomToFitPadding option to TldrawOptions

### LINKED ISSUE #7599: Add zoomToFitPadding option to TldrawOptions
## Problem

The `zoomToBounds` and `zoomToFit` methods use a hardcoded padding value (`ZOOM_TO_FIT_PADDING = 128`) when calculating the viewport inset. This padding determines how much space is left around content when zooming to fit.

Currently, the only way to customize this behavior is to pass `inset` on every individual call to `zoomToBounds`:

```ts
editor.zoomToBounds(bounds, { inset: 64 })
```

This is inconvenient when you want consistent padding across your entire application, especially since many zoom operations are triggered by UI actions (like "Zoom to fit" in the menu) that don't expose this option.

## Proposed solution

Add a `zoomToFitPadding` option to `TldrawOptions` that allows configuring the default padding used for zoom-to-fit operations:

```ts
const options: Partial<TldrawOptions> = {
  zoomToFitPadding: 64, // default is 128
}

function MyApp() {
  return <Tldraw options={options} />
}
```

This follows the existing pattern used for other configurable values like `defaultSvgPadding`, `hitTestMargin`, etc.

## Use cases

- **Embedded editors**: Apps embedding tldraw in smaller containers may want less padding
- **Presentation modes**: Full-screen presentations might want minimal padding
- **Dense UIs**: Apps with sidebars/panels may want to account for UI chrome
- **Consistency**: Set once and have all zoom-to-fit operations respect the same value

## Alternatives considered

1. **Per-call `inset` option**: Already exists but requires passing on every call and doesn't affect UI-triggered zooms
2. **CSS-based solution**: Wouldn't work as this is a camera calculation, not visual styling

steveruizok:
This PR adds a configurable `zoomToFitPadding` option to `TldrawOptions` that allows customizing the default padding used when zooming to fit content in the viewport.

Previously the padding was hardcoded to 128 pixels. Now users can configure it via the options prop:

```tsx
const options: Partial<TldrawOptions> = {
  zoomToFitPadding: 64, // default is 128
}

function MyApp() {
  return <Tldraw options={options} />
}
```

This follows the existing pattern used for other configurable values in `TldrawOptions`. The change affects methods like `zoomToFit()`, `zoomToSelection()`, and `zoomToBounds()`.

Closes #7599

### Change type

- [x] `feature`
- [x] `api`

### Test plan

1. Create a tldraw instance with default options
2. Add some shapes and use "Zoom to fit" from the menu
3. Verify the padding around content matches the default (128px or 28% of viewport width, whichever is smaller)
4. Create a tldraw instance with `options={{ zoomToFitPadding: 64 }}`
5. Verify the padding is now 64px (or 28% of viewport width if that's smaller)

- [x] Unit tests

### Release notes

- Added `zoomToFitPadding` option to `TldrawOptions` to allow customizing the default padding used by zoom-to-fit operations.

### API changes

- Added `zoomToFitPadding: number` to the `TldrawOptions` interface. Defaults to 128 pixels.

---

> [!NOTE]
> Introduces a configurable padding for zoom-to-fit operations and removes the hardcoded constant.
> 
> - Adds `zoomToFitPadding` to `TldrawOptions` (default `128`) and exposes it in `defaultTldrawOptions`
> - Updates `Editor.zoomToBounds()` to use `this.options.zoomToFitPadding` instead of the removed `ZOOM_TO_FIT_PADDING`
> - Removes exports of `ZOOM_TO_FIT_PADDING` and `DEFAULT_SNAP_THRESHOLD` from `constants`/`index`
> - Updates unit tests to reference the option and adds a test for custom padding
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 05c929634e8fc76c012f147a922c038dae7f099c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/options.ts
+++ b/packages/editor/src/lib/options.ts
@@ -1,5 +1,4 @@
 import { ComponentType, Fragment } from 'react'
-import { DEFAULT_SNAP_THRESHOLD, ZOOM_TO_FIT_PADDING } from './constants'
 
 /**
  * Options for configuring tldraw. For defaults, see {@link defaultTldrawOptions}.
@@ -170,6 +169,6 @@ export const defaultTldrawOptions = {
 	debouncedZoom: true,
 	debouncedZoomThreshold: 500,
 	spacebarPanning: true,
-	zoomToFitPadding: ZOOM_TO_FIT_PADDING,
-	snapThreshold: DEFAULT_SNAP_THRESHOLD,
+	zoomToFitPadding: 128,
+	snapThreshold: 8,
 } as const satisfies TldrawOptions
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -615,9 +615,6 @@ export const DEFAULT_ANIMATION_OPTIONS: {
 // @internal (undocumented)
 export const DEFAULT_CAMERA_OPTIONS: TLCameraOptions;
 
-// @internal (undocumented)
-export const DEFAULT_SNAP_THRESHOLD = 8;
-
 // @public (undocumented)
 export function DefaultBackground(): JSX.Element;
 
--- a/packages/editor/src/index.ts
+++ b/packages/editor/src/index.ts
@@ -120,12 +120,7 @@ export {
 	userTypeValidator,
 	type TLUserPreferences,
 } from './lib/config/TLUserPreferences'
-export {
-	DEFAULT_ANIMATION_OPTIONS,
-	DEFAULT_CAMERA_OPTIONS,
-	DEFAULT_SNAP_THRESHOLD,
-	SIDES,
-} from './lib/constants'
+export { DEFAULT_ANIMATION_OPTIONS, DEFAULT_CAMERA_OPTIONS, SIDES } from './lib/constants'
 export {
 	BindingUtil,
 	type BindingOnChangeOptions,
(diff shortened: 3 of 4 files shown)


packages/editor/src/lib/options.ts
```diff
@@ -104,9 +104,12 @@ export interface TldrawOptions {
...
 	readonly spacebarPanning: boolean
 	/**
-	 * The distance (in screen pixels) at which shapes snap to guides and other shapes.
+	 * The default padding (in pixels) used when zooming to fit content in the viewport.
+	 * This affects methods like `zoomToFit()`, `zoomToSelection()`, and `zoomToBounds()`.
+	 * The actual padding used is the minimum of this value and 28% of the viewport width.
+	 * Defaults to 128 pixels.
 	 */
-	readonly snapThreshold: number
+	readonly zoomToFitPadding: number
```
mimecuvalo, line 112: make this optional - to make it easier for folks to transition, API-wise

mimecuvalo (changes requested):
tiny nit before landing

packages/editor/src/lib/options.ts
steveruizok: we partial this so it is optional. confusing that some are explicitly marked as optional, I'll make an issue for that.

PR #7610, opened 2026-01-06
fix(editor): distanceToLineSegment now returns actual distance instead of squared

arpit-goblins:
This PR fixes a bug in `Geometry2d.distanceToLineSegment` where it was returning squared distance instead of actual distance.

The method uses `Vec.Dist2` (squared distance) internally for efficient comparison, but was returning the raw squared value. This caused `hitTestLineSegment` to compare squared distance against non-squared margin values (like `hitTestMargin / zoomLevel`), making hit detection far too strict.

For example, with a default `hitTestMargin` of 8 pixels:
- Before: shapes only hit if within ~2.8px (`√8`) of the eraser stroke
- After: shapes correctly hit if within 8px of the eraser stroke

### Change type

- [x] `bugfix`

### Test plan

1. Use the eraser tool to erase shapes
2. Notice that shapes are now erased when the stroke comes within the expected margin (~8 screen pixels)

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where `distanceToLineSegment` returned squared distance instead of actual distance, causing hit testing (eraser, scribble select) to be too strict.

---

> [!NOTE]
> Corrects geometry distance calculations used by hit testing.
> 
> - Fixes `Geometry2d.distanceToLineSegment` to return `Math.sqrt(dist)` (actual distance) rather than squared distance
> - Ensures `hitTestLineSegment` compares consistent units with margins, improving eraser detection
> - Updates `groups.test.tsx` erasing assertion to expect `getErasingShapeIds().length` of 3 after moving the eraser
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 393174d8cf63ccd6101f3446e1793528a213f807. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


arpit-goblins:

steveruizok:
We use Dist2 to avoid having to square distances when comparing them. The closest point according to Vec.Dist2 will be the same as the shortest Vec.Dist but will be faster. If it is important to have the distance in actual units then we should only be squaring the result.

steveruizok:
@arpit-goblins mind signing the CLA from your other account?

arpit-goblins:
> @arpit-goblins mind signing the CLA from your other account?

Yeah, sorry about that.

arpit-goblins:
> We use Dist2 to avoid having to square distances when comparing them. The closest point according to Vec.Dist2 will be the same as the shortest Vec.Dist but will be faster. If it is important to have the distance in actual units then we should only be squaring the result.

Sure - I see you modified it to square root it at the end.

tyranasaurus:

PR #7612, opened 2026-01-06
fix(tldraw): don't debounce asset resolution if asset changed

### LINKED ISSUE #7611: Bug: Updating the asset of an ImageShape should not be debounced
## Background

Asset debouncing was introduced in https://github.com/tldraw/tldraw/pull/5361 to improve performance during zooming. However, this change has caused a regression where updating the asset of an image shape is now noticeably slower.

## How to reproduce

You can reproduce this issue by updating a shape using the code below:

```ts
        editor.updateShape<TLImageShape>({
          id: imageShapeId,
          type: 'image',
          parentId: panelFrameId,
          x,
          y,
          isLocked: false,
          meta: {
            name: 'Image',
            panelIndex,
            mediaId,
            taskId,
            width,
            height,
          },
          props: {
            assetId: updatedAssetId, // <-- Update assetId
            w,
            h,
          },
        })
```

steveruizok:
Asset debouncing (PR #5361) introduced a regression where updating the assetId of an ImageShape is noticeably slow due to a 500ms debounce. This PR fixes the issue by only skipping the debounce when the asset id has changed.

Closes #7611

### Change type

- [x] `bugfix`

### Test plan

1. Add an image shape to the canvas
2. Update its `assetId` programmatically via `editor.updateShape()`
3. The new image should appear immediately without delay
4. Zoom in/out rapidly - debouncing should still apply during camera movement for performance

### Release notes

- Fixed a regression where updating an image shape's asset was delayed by 500ms due to incorrect debouncing.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to a React hook’s debouncing control flow; main risk is subtle timing/cleanup issues around tick listener cancellation.
> 
> **Overview**
> Fixes a regression in `useImageOrVideoAsset` where swapping an image/video shape’s `assetId` could be delayed by the existing 500ms debounce.
> 
> The hook now tracks the previous `assetId` and **bypasses debouncing on asset changes**, resolving the new URL immediately (and cancelling any pending debounced tick handler), while keeping the debounce behavior for repeated scale/zoom-driven URL variations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8715f2a40193ee225dcea0b88421ec4b337f7b59. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
+++ b/packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
@@ -62,8 +62,8 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 	// Track the previous assetId to detect when the asset itself changes
 	const previousAssetId = useRef<TLAssetId | null>(null)
 
-	// Track whether we should skip debouncing for the next immediate resolution
-	const shouldSkipDebounce = useRef(false)
+	// Track whether we should run immediately (skip debouncing) for the next resolution
+	const shouldRunImmediately = useRef(false)
 
 	// The last URL that we've seen for the shape
 	const previousUrl = useRef<string | null>(null)
@@ -73,9 +73,9 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 		const assetIdChanged = previousAssetId.current !== assetId
 		previousAssetId.current = assetId
 
-		// Set flag to skip debouncing for the next immediate resolution
+		// Set flag to run immediately (skip debouncing) for the next resolution
 		if (assetIdChanged) {
-			shouldSkipDebounce.current = true
+			shouldRunImmediately.current = true
 		}
 
 		if (!assetId) return
@@ -124,7 +124,7 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 
 			// Debounce fetching potentially multiple image variations (e.g. during zoom or resize).
 			// Don't debounce when the asset itself changes - resolve immediately.
-			if (didAlreadyResolve.current && !shouldSkipDebounce.current) {
+			if (didAlreadyResolve.current && !shouldRunImmediately.current) {
 				let tick = 0
 
 				const resolveAssetAfterAWhile = () => {
@@ -145,7 +145,7 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 				cancelDebounceFn?.()
 				resolveAssetUrl(editor, assetId, screenScale, exportInfo, (url) => resolve(asset, url))
 				// Reset the flag after immediate resolution so subsequent updates are debounced
-				shouldSkipDebounce.current = false
+				shouldRunImmediately.current = false
 			}
 		})
 


packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -107,8 +107,9 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 				exportIsReady() // let the SVG export know we're ready for export
 			}
 
-			// If we already resolved the URL, debounce fetching potentially multiple image variations.
-			if (didAlreadyResolve.current) {
+			// Debounce fetching potentially multiple image variations while the camera is moving (zooming).
+			// Don't debounce when the asset changes (e.g. assetId update) - resolve immediately.
+			if (didAlreadyResolve.current && editor.getCameraState() === 'moving') {
```
mimecuvalo: this isn't just above the camera moving - this is so you don't download many different _resolutions_ of images when zooming in/out

mimecuvalo: oh wait zoom is covered in 'moving' - nvm

mimecuvalo: mm, this is still tricky b/c it's not just camera state that affects whether resolution changes will happen. 
for example, this could also happen when resizing an image shape. but also, people might have other criteria for resolution changes (e.g. network changes). i dunno, i think this PR might require a different tact.

steveruizok: Ah right, so its almost like there's a queue of things that need to be resolved but that queue is blocked by the resize or by the camera moving or by a timeout?

steveruizok: This has to be like the third queue / debounced flush thing that we have in the app. Maybe we need a plumber (a flush manager).

packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -59,12 +59,19 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
...
+
 	// The last URL that we've seen for the shape
 	const previousUrl = useRef<string | null>(null)
 
 	useEffect(() => {
 		if (!assetId) return
 
+		// Check if the assetId changed (not just resolution/scale updates)
+		const assetIdChanged = previousAssetId.current !== assetId
+		previousAssetId.current = assetId
```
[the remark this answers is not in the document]
steveruizok: @claude fix

packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -107,8 +114,9 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 				exportIsReady() // let the SVG export know we're ready for export
 			}
 
-			// If we already resolved the URL, debounce fetching potentially multiple image variations.
-			if (didAlreadyResolve.current) {
+			// Debounce fetching potentially multiple image variations (e.g. during zoom or resize).
+			// Don't debounce when the asset itself changes - resolve immediately.
+			if (didAlreadyResolve.current && !assetIdChanged) {
```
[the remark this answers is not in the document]
steveruizok: @claude evaluate and fix

packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
```diff
@@ -59,10 +59,25 @@ export function useImageOrVideoAsset({ shapeId, assetId, width }: UseImageOrVide
 	// A flag for whether we've resolved the asset URL at least once, after which we can debounce
 	const didAlreadyResolve = useRef(false)
 
+	// Track the previous assetId to detect when the asset itself changes
+	const previousAssetId = useRef<TLAssetId | null>(null)
+
+	// Track whether we should skip debouncing for the next immediate resolution
+	const shouldSkipDebounce = useRef(false)
```
mimecuvalo: nit: i'd rename this to `shouldRunImmediately` b/c the double-negatives below throw me off a bit

mimecuvalo (approved):
pre-approved - thx!

packages/tldraw/src/lib/shapes/shared/useImageOrVideoAsset.ts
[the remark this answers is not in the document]
steveruizok: @claude please

PR #7638, opened 2026-01-06
docs(editor): clarify getClipPath should return inner clip boundary

### LINKED ISSUE #7627: Child shapes protrude beyond stroke boundary in custom clipping shapes
## Summary

When using a custom shape with clipping enabled via `getClipPath`, the stroke width of the shape is not accounted for when calculating the clip boundary. This causes child shapes to visually protrude beyond the inner edge of the stroke.

## Background

This issue wasn't noticeable when only frames supported clipping, because frames have very thin strokes (or effectively no visible stroke boundary to consider). Now that custom shapes can implement `getClipPath` for arbitrary clip boundaries, shapes with thicker strokes reveal this limitation.

## Reproduction

1. Create a custom clipping shape with a visible stroke (e.g., the `CircleClipShapeUtil` example with a thicker stroke)
2. Add child shapes that extend to the edge of the clipping boundary
3. Observe that children appear to protrude slightly beyond the inner stroke boundary

The issue is that `getClipPath` returns points that define the shape's geometry boundary, but when the shape is rendered with a stroke, that stroke is drawn centered on the boundary line (half inside, half outside). The clipping calculation doesn't account for this, so children clip to the geometric center of the stroke rather than its inner edge.

## Expected behavior

Child shapes should be clipped to the inner edge of the parent shape's stroke, not the geometric center line where the stroke is drawn.

## Relevant code

The clipping mask is calculated in `_getShapeMaskCache` (Editor.ts:4926-4951), which calls `util.getClipPath?.(ancestor)` to get the clip boundary points. There's currently no mechanism to adjust this boundary inward by half the stroke width.

## Potential solutions

1. Add an optional `getClipPathStrokeWidth` method to ShapeUtil that returns the stroke width to inset by
2. Modify `getClipPath` to accept an options parameter that includes stroke compensation
3. Have `getClipPath` implementations return an already-inset path (putting the burden on shape implementers)
4. Add a `getClipPathInset` method that returns the inset amount

## Related

- Custom clipping example: `apps/examples/src/examples/custom-clipping-shape/`
- `_getShapeMaskCache`: `packages/editor/src/lib/editor/Editor.ts:4926`

steveruizok:
Fixes the issue where child shapes protrude beyond the stroke boundary in custom clipping shapes. Rather than adding complex polygon offset utilities, this PR takes the simpler approach of clarifying that `getClipPath` should return the **inner** clip boundary (already accounting for stroke width).

Closes #7627

### Changes

1. **Updated `getClipPath` documentation** in `ShapeUtil.ts` to clarify that:
   - The returned points should define the inner clip boundary
   - If your shape has a stroke, you should inset the clip path by half the stroke width
   - Includes a code example showing the pattern

2. **Updated custom-clipping-shape example** to demonstrate the correct pattern:
   - Added a `STROKE_WIDTH` constant for consistency
   - Modified `getClipPath` to use `clipRadius = outerRadius - STROKE_WIDTH / 2`
   - Added comments explaining why the inset is needed

3. **Updated tests** to expect the correct clip radius (99 instead of 100 for a 200x200 shape with stroke width 2)

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn dev` and navigate to the custom clipping shape example
2. Create a circle clip shape and add child shapes inside it
3. Observe that children are clipped to the inner edge of the stroke, not the center line

- [x] Unit tests

### Release notes

- Fixed child shapes protruding beyond the stroke boundary in custom clipping shapes by clarifying that `getClipPath` should return the inner clip boundary (accounting for stroke width)

---

> [!NOTE]
> Clarifies the clipping contract and aligns examples/tests to clip to the inner stroke edge.
> 
> - **Docs**: Expand `ShapeUtil.getClipPath` JSDoc to specify returning the inner clip boundary and include an inset-by-half-stroke example.
> - **Example**: Update `CircleClipShapeUtil` to use `STROKE_WIDTH`, compute `clipRadius = outerRadius - STROKE_WIDTH / 2`, and derive polygon segments from circumference via `PI2`/`clamp`; circle render now uses the constant stroke width.
> - **Tests**: Update custom clipping tests to expect the inset radius (e.g., 99 for 200×200 with stroke 2) and mirror the example’s `clipRadius` logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 031ab3d9c99073de6f3811cf9cd33936dfc62fc5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
+++ b/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
@@ -324,7 +324,7 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
 	 * @example
 	 * ```ts
 	 * override getClipPath(shape: MyShape): Vec[] | undefined {
-	 *   const strokeWidth = 3
+	 *   const strokeWidth = 2
 	 *   const inset = strokeWidth / 2
 	 *   // Return points inset by half the stroke width
 	 *   return [


packages/tldraw/src/test/custom-clipping.test.ts
```diff
@@ -31,6 +31,9 @@ export type CircleClipShape = TLShape<typeof CIRCLE_CLIP_TYPE>
 
 export const isClippingEnabled$ = atom('isClippingEnabled', true)
 
+// The stroke width used when rendering the circle
+const STROKE_WIDTH = 2
```
kostyafarber, line 35: should this be 3? it's three in the examples if not then lgtm

Issue #7641, opened 2026-01-07
An extra line is inserted underneath bullet points, and cannot be removed


nxb9626:
# Summary

When a text box's last line is a bullet point a line is inserted below and there seems to be no way to remove it. Found on the tldraw.com site but recreated locally by following the sdk basic example setup so I'm fairly sure this is an sdk issue. Noticed while trying to align  the spacing appears by running the `Arrangement` > `Distrubute V` on a few boxes with bulleted lists that are close together,  since there is now extra empty space below each text box.

# Problem

This extra new line is inserted and cannot be removed. Currently clicking Backspace with the moves the cursor up but does not remove that new line. 

<img width="415" height="317" alt="Image" src="https://github.com/user-attachments/assets/5242e58c-3f8e-4e0a-befa-9ffbd220e486" />

# Example

Probably unnecessary but other issues had an example so I'm including an example :\) 

https://www.tldraw.com/f/1SLjDo9ZvSF3a4-qpjxwG?d=v-57.-317.2151.1380.page

Can provide edit access if that is helpful, but it seems to happen on every bullet text box whose last line is a bullet point.

# Expected Behavior

- It seems like it should not be auto inserted from the start
- Pressing Backspace (Delete on mac) on that bottom line should remove it
- Pressing the Delete(Windows/Linux) key on the line above (end of the bullet point should remove it

# Actual Behavior

- On inserting a bullet point a new line is also inserted
- Pressing Backspace (MacOS Delete) on that new line does not remove it
- Pressing Delete (Windows/Linux Delete) on the line above inserts a new bullet point, which is odd and may be related

# Technical Context

Have tested in Firefox and Chrome on Linux and MacOS, occurring in all places.

If I get the chance I'll dive in and get an idea of what is causing this code-wise.

mimecuvalo:
fixed in https://github.com/tldraw/tldraw/pull/7643

PR #7643, opened 2026-01-07
fix: prevent extra line after bullet/ordered lists (#7641)

sahiee-dev:
## Description

This PR fixes an issue where an extra, non-removable empty line was automatically inserted after bullet and ordered lists when they were the last line in a text shape.

### What was happening
- When a text shape ended with a bullet or numbered list, an empty paragraph was forcibly added below it
- Pressing **Backspace** on the empty line would move the cursor but not remove the line
- Pressing **Delete** at the end of a list item would incorrectly create a new empty list item
- This caused unexpected spacing, especially noticeable when using **Arrange → Distribute V** across multiple text shapes

### Root cause
`tldraw` uses TipTap’s `StarterKit`, which includes the `TrailingNode` extension.

By default, `TrailingNode` inserts a trailing paragraph after block nodes unless they are explicitly excluded via `notAfter`.  
List-related nodes (`bulletList`, `orderedList`, `listItem`) were not excluded, so a trailing paragraph was always inserted after lists.

### Fix
Configured `TrailingNode` to skip inserting trailing paragraphs after list-related nodes by extending the `notAfter` configuration.  
This prevents the extra empty line and restores correct cursor behavior at the end of lists.

### Visual comparison
Before and after images are included below to demonstrate:
- Removal of the phantom trailing line
- Correct cursor behavior
- Proper vertical spacing when distributing text shapes

Screenshot 2026-01-07 at 3 08 03 PM (BEFORE)

Screenshot 2026-01-07 at 3 07 35 PM (AFTER)

---

### Change type

- [x] `bugfix`

---

### Test plan

1. Create a text shape
2. Add a bullet or ordered list
3. Ensure the list is the final line in the text
4. Verify:
   - No extra empty line is inserted
   - Backspace correctly removes the last list item
   - Delete does not create a phantom list item
5. Repeat with multiple text shapes and use **Arrange → Distribute V** to confirm spacing is correct

- [ ] Unit tests  
- [ ] End to end tests

---

### Release notes

- Fixed an issue where bullet and ordered lists added an extra, non-removable empty line at the end of text shapes

---

> [!NOTE]
> Prevents phantom trailing lines after lists by adjusting the TipTap StarterKit configuration.
> 
> - Adds `trailingNode.notAfter` for `paragraph`, `bulletList`, `orderedList`, `listItem` in `richText.ts` to stop auto-inserted trailing paragraphs
> - Removes the extra empty line after lists and restores expected backspace/delete behavior at list ends
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87d316636dccf3bdf2c9e88194c362d876792b41. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


steveruizok:
Will close https://github.com/tldraw/tldraw/issues/7641

kostyafarber:
nice! I think you added something to the claude settings by accident

mimecuvalo:
Yeah alright, let's disable it for these. This was originally b/c it's an opinionated take on making the usability of navigating around editing text more intuitive. Other text editors also have had to make this tradeoff.

I think in previous versions of TipTap it was much more harder to exit "list-mode" and go back to editing regular unblocked text. But it feels better now, even when disabled, so let's do this 👍 

As Kostya mentions above, let's revert the unrelated change and we should be good to go. Thanks!

sahiee-dev:
Reverted the unrelated change to `.claude/settings.json` as per reviewer feedback. 

The PR now only includes the trailing-node fix for bullet and ordered lists (#7641).  

All other behavior remains unchanged.

steveruizok:
> Yeah alright, let's disable it for these. This was originally b/c it's an opinionated take on making the usability of navigating around editing text more intuitive. Other text editors also have had to make this tradeoff.
> 
> I think in previous versions of TipTap it was much more harder to exist "list-mode" and go back to editing regular unblocked text. But it feels better now, even when disabled, so let's do this 👍
> 
> As Kostya mentions above, let's revert the unrelated change and we should be good to go. Thanks!

Could we make a change to show the extra space while editing but not while not-editing?

sahiee-dev:
> > Yeah alright, let's disable it for these. This was originally b/c it's an opinionated take on making the usability of navigating around editing text more intuitive. Other text editors also have had to make this tradeoff.
> > I think in previous versions of TipTap it was much more harder to exist "list-mode" and go back to editing regular unblocked text. But it feels better now, even when disabled, so let's do this 👍
> > As Kostya mentions above, let's revert the unrelated change and we should be good to go. Thanks!
> 
> Could we make a change to show the extra space while editing but not while not-editing?

Good question, I agree that would be a nicer UX.

I think that would require distinguishing between “editing” vs “static” render states (e.g. via editor/view state or render-time spacing), which feels a bit broader than this bugfix and might need some design discussion.

For this PR, I’ve kept it scoped to restoring expected list behavior and removing the non-removable trailing paragraph.

Happy to explore the “show while editing only” approach in a follow-up PR if that sounds reasonable 👍

mimecuvalo:
> Could we make a change to show the extra space while editing but not while not-editing?

Eh, I think it makes it jarring. The shape size might change underneath you b/c it would be adding an extra line on enter/exit. 
Again, I think exiting list mode was more cumbersome in prev. versions of TipTap. I think it's easier to keyboard out in later versions.

Issue #7581, opened 2026-01-04
Clicking on selection outline should keep selection even when another shape is above


steveruizok:
## Description

When a shape is selected and another shape is positioned above it, clicking on the blue selection outline of the selected shape incorrectly selects the overlapping shape instead of keeping the current selection.

The selection outline/foreground should have hit testing priority over other shapes, so that users can interact with resize handles, rotation handles, and the selection bounds even when the selected shape is behind other shapes.

## Steps to reproduce

1. Create two overlapping shapes (e.g., a body shape and a tail shape like in the screenshot)
2. Select the shape that is behind the other (lower in z-order)
3. Click on the visible blue selection outline of the selected shape where it overlaps with the shape above
4. Observe: The overlapping shape gets selected instead

## Expected behavior

Clicking anywhere on the selection outline (the blue stroke around the selection bounds) should maintain the current selection and allow interaction with the selected shape, regardless of what other shapes might be above it.

## Actual behavior

The selection outline does not get hit testing priority, so clicking on it where another shape overlaps causes that shape to be selected instead.

## Visual reference

In the example below, clicking on the blue outline of the selected tail (behind the body) should not select the body shape:

- The tail is currently selected (shown with blue outline and handles)
- Clicking on the blue selection outline where it's visible behind the body incorrectly selects the body instead

steveruizok:
The research here was wrong—the fix has to do with the `Editor.getShapeAtPoint` rather than CSS.

sahiee-dev:
I implemented Solution #2 (prioritizing selection bounds before shape hit-testing). The fix works but changes interaction semantics, shapes inside selection bounds cannot be selected without first clearing selection. Please confirm this is the desired behavior.

max-drake:
closing bc can no longer repro:
https://github.com/user-attachments/assets/8ef4d680-d34f-4a69-82ea-f09ea7321691

PR #7654, opened 2026-01-08
fix: make selection outline intercept pointer events (tldraw #7581)

sahiee-dev:
## Descripton
When a shape is selected and another shape is positioned above it, clicking on the blue selection outline of the selected shape was incorrectly selecting the overlapping shape instead of keeping the current selection.

**Root cause:** The selection outline had `pointer-events: none` in CSS, so clicks passed through to the canvas, which then hit-tested shapes and found the overlapping one.

**Solution:** Make the selection outline itself interactive by:
1. Adding `'outline'` to the [TLSelectionHandle](cci:2://file:///Users/lulu/Desktop/tldraw/packages/editor/src/lib/editor/types/selection-types.ts:3:0-3:90) type
2. Changing `.tl-selection__fg__outline` from `pointer-events: none` to `pointer-events: stroke`
3. Adding an explicit `'outline'` case in the selection handler that transitions to `pointing_selection`
4. Attaching [useSelectionEvents('outline')](cci:1://file:///Users/lulu/Desktop/tldraw/packages/editor/src/lib/hooks/useSelectionEvents.ts:7:0-96:1) to the outline rect element

Using `pointer-events: stroke` (not `all`) ensures clicks pass through the interior of selection bounds, only capturing clicks on the visible stroke itself. This preserves the ability to click shapes inside the selection bounds.

### Before
Clicking on the selection outline where it overlaps another shape incorrectly selects the overlapping shape.

### After
Clicking on the selection outline maintains the current selection, allowing proper interaction with the selected shape.

### Change type

- [x] `bugfix`

### Test plan

1. Create two overlapping shapes (e.g., Rectangle A at 300,300 and Rectangle B at 400,400)
2. Select the shape that is behind (lower in z-order)
3. Click on the visible blue selection outline where it overlaps with the shape above
4. Verify the original shape stays selected
5. Verify clicking inside the selection bounds (not on the outline) still allows selecting shapes underneath
6. Verify dragging from the outline moves the selected shape correctly

- [x] Unit tests

### Release notes

- Fixed a bug where clicking on the selection outline of a selected shape would incorrectly select an overlapping shape instead of maintaining the current selection.

---

> [!NOTE]
> Ensures clicks on the visible selection stroke keep the current selection instead of passing through to overlapping shapes.
> 
> - Add `'outline'` to `TLSelectionHandle` and wire `useSelectionEvents('outline')` to the selection rect in `TldrawSelectionForeground`
> - Change `.tl-selection__fg__outline` CSS to `pointer-events: stroke` so only the stroke intercepts events (interior still passes through)
> - Update select tool states:
>   - In `select.idle`, clicking `handle: 'outline'` transitions to `pointing_selection`
>   - In `select.crop.idle`, clicking `outline` transitions to `select.crop.pointing_crop` (drag crop)
>   - In `PointingSelection`, skip hit-testing when `handle === 'outline'` on click/double-click to preserve selection
> - Add unit tests covering outline click behavior and state transitions
> - Minor CSS selector/whitespace tidy-ups
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 184034e8b318cf307bff121cbca47d0333030640. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sahiee-dev:
fixes #7581

steveruizok:
Sorry, this is the wrong solution to the problem. The research bot came up with a bad suggestion.

sahiee-dev:
Thanks for the clarification understood.
I see now that the outline should remain visual-only and that hit-testing priority needs to be resolved in the select tool logic rather than via pointer event interception.
I’ll take a look at reworking this with selection-bounds checks happening before shape hit-testing.

PR #7683, opened 2026-01-13
feat(sdk): built-in shape style overrides proposal

ds300:

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -839,24 +839,63 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		// Priority order (highest wins):
 		// 1. getShapeStyleOverrides callback
-		// 2. defaultStyles (from ShapeUtil)
+		// 2. StyleProp resolvers ($ props) sorted by priority
+		// 3. defaultStyles (from ShapeUtil)
 
-		const callbackOverrides = this._getShapeStyleOverrides?.(shape, this)
-
-		// Fast path: no overrides
-		if (!callbackOverrides || Object.keys(callbackOverrides).length === 0) {
-			return defaultStyles
-		}
-
-		// Merge overrides, resolving any themeable values
 		const merged = { ...defaultStyles }
 
-		for (const [key, value] of Object.entries(callbackOverrides)) {
-			if (typeof value === 'undefined') continue
-			if (isThemedValue(value)) {
-				;(merged as any)[key] = ctx.isDarkMode ? value.dark : value.light
-			} else {
-				;(merged as any)[key] = value
+		// Apply StyleProp resolvers ($ props) in priority order
+		const stylePropsForShape = this.styleProps[shape.type]
+		if (stylePropsForShape) {
+			// Collect style props with resolvers, sorted by priority
+			const propsWithResolvers: Array<{ styleProp: StyleProp<any>; propKey: string }> = []
+
+			for (const [styleProp, propKey] of stylePropsForShape) {
+				if (styleProp.hasResolver() && styleProp.appliesToShape(shape.type)) {
+					propsWithResolvers.push({ styleProp, propKey })
+				}
+			}
+
+			// Sort by priority (lower first), then by id for stable ordering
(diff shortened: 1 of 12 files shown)


packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -221,6 +234,8 @@ export function RichTextSVG({
 	labelColor,
 	padding,
 	showTextOutline = true,
+	fontFamily,
+	fontWeight,
```
mimecuvalo, line 238: something about this feels like a losing proposition, or like, a moving target.
there are at least these CSS font properties:
```
font-family
font-kerning
font-optical-sizing
font-palette
font-size
font-size-adjust
font-style
font-variant
font-weight
...
and so many more
```

 i guess eventually, outside of this PR, we might consider some general `cssProperties` property

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -252,6 +255,34 @@ export interface TLEditorOptions {
...
+	 *   }
+	 *   return undefined
+	 * }}
+	 * ```
+	 *
+	 * @param shape - The shape to compute overrides for.
+	 * @param editor - The editor instance.
+	 * @returns Style overrides to apply, or undefined/null for no overrides.
+	 */
+	getShapeStyleOverrides?(shape: TLShape, editor: Editor): TLShapeStyleOverrides | null | undefined
```
mimecuvalo, line 286: this is feeling good 👍

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -536,6 +549,39 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
...
+	 *     strokeWidth: STROKE_SIZES[shape.props.size] * shape.props.scale,
+	 *     strokeColor: getColorValue(theme, shape.props.color, 'solid'),
+	 *     // ... other styles - all fully resolved, no Themeable wrappers
+	 *   }
+	 * }
+	 * ```
+	 *
+	 * @public
+	 */
+	getDefaultStyles(_shape: Shape, _ctx: TLStyleContext): TLShapeResolvedStyles | undefined {
```
mimecuvalo: also feeling good about this 👍

packages/editor/src/lib/editor/Editor.ts
mimecuvalo: what i like especially here is that centering around this mechanism allows for folks to keep it semantic.
say, if they made a color set like: `success | error | warning | info` then those could be mapped later to whatever the latest colors are for those values. that, instead of having to hard code hex values that might quickly go out-of-date.

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -211,6 +212,27 @@ export interface TldrawEditorBaseProps {
...
+	/**
+	 * Provides a way to compute style overrides for shapes at runtime.
+	 *
+	 * This callback is called when computing resolved styles for a shape. The returned
+	 * overrides are merged on top of the shape's default styles.
+	 *
+	 * @example
+	 * ```ts
+	 * getShapeStyleOverrides={(shape, editor) => {
+	 *   if (shape.meta.highlighted) {
```
mimecuvalo, line 224: do we still think `meta` is where the overrides should live? yes, it's a flexible grab-bag. but, say, in a year where maybe we have extensions/plugins and different conflicting plugins both have conflicting names on `meta`. we should at least consider `styleOverrides` being the canonical/blessed place for these kinds of things to live on `meta`. but also, the `styleOverrides` being on the top-level is not necessarily a terrible idea, i see you removed that (which was a 3rd-level before) but maybe we still use `styleOverrides` on the top-level for these functional overrides?

PR #7811, opened 2026-02-03
fix(menu): add useCanApplySelectionAction hook to disable menu items when not in select tool

### LINKED ISSUE #7810: Zoom to selection button non-functional when not in select tool
## Description

When a user has shapes selected and then switches to a different tool (e.g., draw, geo, etc.), the "Zoom to selection" button in the zoom menu remains enabled, but clicking it does nothing.

## Steps to reproduce

1. Create some shapes on the canvas
2. Select one or more shapes with the select tool
3. Switch to a different tool (e.g., press 'D' for draw tool)
4. Open the zoom menu and observe that "Zoom to selection" is still enabled
5. Click "Zoom to selection"
6. **Nothing happens** - the camera doesn't zoom to the selection

## Expected behavior

Either:
- The button should be disabled when not in the select tool (consistent with action behavior), OR
- Clicking the button should work and zoom to the selection regardless of current tool

## Actual behavior

The button appears enabled but clicking it has no effect. This creates a confusing UX where the button seems broken.

## Root cause

There's a mismatch between the menu item's enabled state and the action's behavior:

- `ZoomToSelectionMenuItem` checks only `editor.getSelectedShapeIds().length > 0` to determine if enabled
- The action's `onSelect` handler checks `canApplySelectionAction()` which requires `editor.isIn('select')` AND has selected shapes
- When not in the select tool, the action silently returns early, doing nothing

## Environment

- tldraw version: main branch
- Browser: All
- OS: All

max-drake:
In order to properly disable menu items when selection-dependent actions cannot be applied, this PR adds a reusable `useCanApplySelectionAction` hook that checks whether the user is in the select tool AND has shapes selected. This corresponds to the `canApplySelectionAction()` check used by many actions in `actions.tsx`.

Closes #7810

https://github.com/user-attachments/assets/481bf941-4cd0-4882-9cf8-3415b0bd8012

### Changes

Updated menu items to use the new hook:
- `ZoomToSelectionMenuItem` - now uses `useCanApplySelectionAction`
- `CutMenuItem` - now checks `useCanApplySelectionAction` + unlocked shapes
- `CopyMenuItem` - now uses `useCanApplySelectionAction`
- `DeleteMenuItem` - now checks `useCanApplySelectionAction` + unlocked shapes

### Change type

- [x] `bugfix`

### Test plan

1. Create some shapes on the canvas
2. Select one or more shapes with the select tool
3. Switch to a different tool (e.g., press 'D' for draw tool)
4. Open menus and verify that selection-dependent items (zoom to selection, cut, copy, delete) are disabled
5. Switch back to select tool and verify items are enabled again

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Added `useCanApplySelectionAction()` hook that returns true when in select tool with shapes selected

### Release notes

- Add `useCanApplySelectionAction` hook for checking if selection actions should be enabled
- Fix menu items (zoom to selection, cut, copy, delete) being enabled when not in select tool

### DIFF
--- a/.claude/skills/write-pr/SKILL.md
+++ b/.claude/skills/write-pr/SKILL.md
@@ -105,3 +105,5 @@ Search for and link relevant issues that this PR addresses.
 
 - Never include "Generated with Claude Code" unless the PR directly relates to Claude Code
 - Never use title case for descriptions - use sentence case
+- Never put yourself as co-author of any commits
+- Always include an API changes section if the PR has changes to any api-report.md


steveruizok:
piggy backing on this to solve many other similar issues

PR #7821, opened 2026-02-04
feat(editor): pass shapes to ShapeUtil.canBind()

### LINKED ISSUE #7748: Allow ShapeUtil.canBind to receive full shape objects, not just types
I have a use case where it would be helpful to use shape props in my determination of whether shapes can bind together, rather than only shape types.

MitjaBezensek:
Closes #7748

`ShapeUtil.canBind()` previously only received shape type strings via `fromShapeType`/`toShapeType`, making it impossible to make binding decisions based on shape props. This PR replaces those with `fromShape`/`toShape` as `TLShape | { type }` unions — matching the pattern already used by `Editor.canBindShapes()`.

When the full shape exists, it's passed directly. When it doesn't (e.g. during arrow creation before the arrow is committed to the store), a `{ type }` stub is passed instead. Consumers can use `'id' in shape` to narrow to `TLShape` when they need prop access.

### Change type

- [x] `api`

### Test plan

1. Create arrows and verify they still bind to shapes correctly
2. Verify group shapes still reject bindings
3. Check pin-bindings, layout-bindings, and sticker-bindings examples still work

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Breaking! Removed `fromShapeType` and `toShapeType` from `TLShapeUtilCanBindOpts`
- Added `fromShape: TLShape | { type: TLShape['type'] }` to `TLShapeUtilCanBindOpts`
- Added `toShape: TLShape | { type: TLShape['type'] }` to `TLShapeUtilCanBindOpts`

### Release notes

- Pass full shapes (when possible) to `ShapeUtil.canBind()` so binding decisions can be based on shape props, not just types.

---

> [!NOTE]
> **Medium Risk**
> Touches a core editor API used by all binding checks and introduces a signature/typing change that may break downstream `canBind` implementations or edge cases where only type stubs are available.
> 
> **Overview**
> `ShapeUtil.canBind()` now receives `fromShape`/`toShape` (either full `TLShape` objects or `{ type }` stubs) via an expanded `TLShapeUtilCanBindOpts`, enabling binding decisions based on actual shape data when available.
> 
> `Editor.canBindShapes()` constructs and passes these new fields while keeping `fromShapeType`/`toShapeType` as *deprecated* compatibility fields, and example utils (layout bindings, pin bindings) plus `ArrowShapeUtil` are updated to use `fromShape.type`/`toShape.type`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b971312fde5739d575d4735119266d5513303351. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
+++ b/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
@@ -50,6 +50,16 @@ export interface TLShapeUtilCanBindOpts<Shape extends TLShape = TLShape> {
 	toShape: TLShape | { type: TLShape['type'] }
 	/** The type of binding. */
 	bindingType: string
+	/**
+	 * The type of shape referenced by the `fromId` of the binding.
+	 * @deprecated Use `fromShape.type` instead.
+	 */
+	fromShapeType: TLShape['type']
+	/**
+	 * The type of shape referenced by the `toId` of the binding.
+	 * @deprecated Use `toShape.type` instead.
+	 */
+	toShapeType: TLShape['type']
 }
 
 /**
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -4343,9 +4343,13 @@ export interface TLShapeUtilCanBindOpts<Shape extends TLShape = TLShape> {
     fromShape: {
         type: TLShape['type'];
     } | TLShape;
+    // @deprecated
+    fromShapeType: TLShape['type'];
     toShape: {
         type: TLShape['type'];
     } | TLShape;
+    // @deprecated
+    toShapeType: TLShape['type'];
 }
 
 // @public (undocumented)
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -6256,6 +6256,8 @@ export class Editor extends EventEmitter<TLEventMap> {
 			fromShape: typeof fromShape === 'string' ? { type: fromShape } : fromShape,
 			toShape: typeof toShape === 'string' ? { type: toShape } : toShape,
 			bindingType,
+			fromShapeType,
+			toShapeType,
 		}
 
 		if (fromShapeType === toShapeType) {


packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -35,15 +35,19 @@ export interface TLShapeUtilConstructor<T extends TLShape, U extends ShapeUtil<T
...
  *
  * @public
  */
 export interface TLShapeUtilCanBindOpts<Shape extends TLShape = TLShape> {
-	/** The type of shape referenced by the `fromId` of the binding. */
-	fromShapeType: TLShape['type']
-	/** The type of shape referenced by the `toId` of the binding. */
-	toShapeType: TLShape['type']
+	/** The shape referenced by the `fromId` of the binding, or a `{ type }` stub if unavailable. */
+	fromShape: TLShape | { type: TLShape['type'] }
```
mimecuvalo, line 48: can we avoid making this a breaking change? we're getting feedback that we're causing a lot of downstream thrash for customers.
possible to make the previous options deprecated, for now? 🙏

MitjaBezensek: Good shout! Done.

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -129,9 +129,9 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 	override canEdit() {
 		return true
 	}
-	override canBind({ toShapeType }: TLShapeUtilCanBindOpts<TLArrowShape>): boolean {
+	override canBind({ toShape }: TLShapeUtilCanBindOpts<TLArrowShape>): boolean {
 		// bindings can go from arrows to shapes, but not from shapes to arrows
-		return toShapeType !== 'arrow'
+		return toShape.type !== 'arrow'
```
mimecuvalo, line 134: should this be: `(toShape.type ?? toShapeType)`

MitjaBezensek: No, don't think we need that. We pass the same thing in both of them.

PR #7826, opened 2026-02-04
perf(editor): skip hover updates while panning

steveruizok:
This PR improves performance when panning in large documents by skipping expensive hover hit-testing while the camera is moving.

Riffing on https://github.com/tldraw/tldraw/pull/5603

https://github.com/user-attachments/assets/2e01018d-05b5-455c-b9a2-40832135b72d

### How it works

Hit-testing shapes is expensive in large documents. When panning, we don't need continuous hover updates—we just need to resume when the camera stops.

The logic:
1. Camera idle → update hover normally
2. Camera moving + locked → skip entirely (no hit-testing)
3. Camera moving + no current hover → lock immediately
4. Camera moving + same shape → keep current hover
5. Camera moving + different shape → clear hover and lock

This means: when you start panning over a shape, it stays hovered until your cursor moves off it, then hover clears and we stop hit-testing until the camera stops.

### Changes

- Added `cameraState: 'idle' | 'moving'` to `TLInstance` (promoted from private atom)
- Added side effect that triggers `updateHoveredShapeId` when camera becomes idle
- Added hover-locking logic in `updateHoveredShapeId.ts`

### Change type

- [x] `improvement`

### API changes

- Added `TLInstance.cameraState: 'idle' | 'moving'` - tracks whether the camera is currently moving or idle

### Test plan

1. Open a document with many shapes
2. Pan around the canvas
3. Verify hover states don't flicker or cause slowdown
4. Verify hover resumes correctly when panning stops

### Release notes

- Improved performance when panning in large documents by skipping hover updates during camera movement.

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how `cameraState` is stored/migrated on `TLInstance` and alters hover hit-testing behavior during camera movement, which could cause hover/selection regressions in edge cases.
> 
> **Overview**
> Improves panning performance by **skipping expensive hover hit-testing while the camera is moving** and only resuming hover updates once the camera returns to idle.
> 
> Promotes camera movement state into `TLInstance.cameraState` (with defaults + migration), updates `Editor.getCameraState()` to read/write via instance state, and adds a default side effect to run `updateHoveredShapeId` when `cameraState` transitions to `idle`.
> 
> Reworks `updateHoveredShapeId` to implement a per-editor “hover lock” during camera movement: keep the current hovered shape if unchanged, otherwise clear hover and stop hit-testing until the camera stops; adds tests covering these scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af79115add89f9648a46a66703035def02d57d19. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/test/commands/cameraState.test.ts
+++ b/packages/tldraw/src/test/commands/cameraState.test.ts
@@ -230,6 +230,115 @@ describe('getDebouncedZoomLevel', () => {
 	})
 })
 
+describe('hover updates during camera movement', () => {
+	// Note: These tests verify the hover locking optimization during camera movement.
+	// Important: pan({ x, y }) moves the camera, which changes how screen coordinates
+	// map to page coordinates. After pan(10, 10), screen point (150, 150) maps to
+	// page point (140, 140). Tests account for this by positioning pointer well
+	// inside shapes so small pans don't move the effective page point outside.
+
+	// Note: Default geo shapes have fill: 'none' (hollow), so hit testing only works on edges.
+	// For interior points to register hits, shapes must have fill: 'solid'.
+
+	it('keeps hover when camera starts moving and pointer is over same shape', () => {
+		// Create a solid shape so we can hit anywhere inside it
+		editor.createShapes([{ type: 'geo', x: 100, y: 100, props: { w: 200, h: 200, fill: 'solid' } }])
+		const shape = editor.getLastCreatedShape()
+
+		// Move pointer to center of shape (200, 200 is center of 100-300 range)
+		editor.pointerMove(200, 200)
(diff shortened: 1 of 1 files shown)


steveruizok:
Now that we've improved the indicators performance, this really doesn't seem to make much of a difference on large rooms. On my computer, the difference is .91ms to .25ms per frame. The UX is good though, I don't think anyone would notice. We're still spending a huge amount of time on culled shapes though.

mimecuvalo:
great change! yeah, i thought in my original PR that we had to still keep updating the hovered shaped. 
i would have agreed that i don't think we need to update it all during camera movement!

PR #7832, opened 2026-02-04
fix(sync): fix Durable Object SQLite migration in multiplayer template

### LINKED ISSUE #7804: Multiplayer template Durable Objects fail due to SQLite migration incompatibility
## Problem

The deployed `multiplayer-template` worker at `multiplayer.templates.tldraw.dev` is failing because the Durable Objects were originally created with `new_classes` but the current code expects SQLite-backed storage (`new_sqlite_classes`).

Cloudflare doesn't allow converting existing Durable Object classes to SQLite storage without deleting the existing data.

## Error

```
This Durable Object is not backed by SQLite storage, so the SQL API is not available. 
SQL can be enabled on a new Durable Object class by using the `new_sqlite_classes` 
instead of `new_classes` under `[[migrations]]` in your wrangler.toml, but an 
already-deployed class cannot be converted to SQLite (except by deleting the existing data).
```

Stack trace:
```
at DurableObjectSqliteSyncWrapper.exec (index.js:6859:22)
at migrateSqliteSyncStorage (index.js:11632:13)
at new SQLiteSyncStorage (index.js:11733:5)
```

## Current configuration

In `templates/sync-cloudflare/wrangler.toml`:

```toml
[[migrations]]
tag = "v1"
new_sqlite_classes = ["TldrawDurableObject"]
```

## Options

1. **Delete existing DOs** - Loses all room data but fixes the issue
2. **Rename the DO class** - Use a new class name (e.g., `TldrawDurableObjectV2`) so fresh SQLite-backed DOs are created. Old rooms become inaccessible.
3. **Add migration path** - Create a new class with SQLite, migrate data from old class, then deprecate old class

## Affected

- Production deployment at `multiplayer.templates.tldraw.dev`
- 108 errors logged in the last observation period

MitjaBezensek:
In order to fix the multiplayer template's broken Durable Object migrations (where `classes_with_sqlite` was used as an invalid Cloudflare migration directive), this PR replaces the old `TldrawDurableObject` with a new `TldrawDurableObjectSqlite` class using proper `new_sqlite_classes` migration.

Follows up on #7829 which attempted to fix this by collapsing migrations, but that approach doesn't work for existing deployments that already ran v1/v2. Instead, this adds a v3 migration that deletes the old class and creates a new SQLite-backed one.

Closes #7804

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn dev-template sync-cloudflare`
2. Verify the Durable Object deploys and rooms work correctly
3. Verify existing deployments can migrate from v1/v2 to v3

### Release notes

- Fix Durable Object SQLite migration in the multiplayer Cloudflare template

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes Cloudflare Durable Object class names and migration directives, including deleting the previous class and introducing a new SQLite-backed class, which can affect existing deployments during upgrade.
> 
> **Overview**
> Fixes the `sync-cloudflare` multiplayer template’s broken Durable Object SQLite rollout by switching the durable object binding/export from `TldrawDurableObject` to a new `TldrawDurableObjectSqlite` implementation.
> 
> Updates `wrangler.toml` migrations to add a `v3` step that **deletes** the old class and creates the replacement via Cloudflare’s supported `new_sqlite_classes` directive (and annotates `v2` as a no-op), and refreshes generated types in `worker-configuration.d.ts` to point at the new class.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1f73f0b8a74ff73bb3d69b02cffcd2a3f7ea218. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/templates/sync-cloudflare/worker-configuration.d.ts
+++ b/templates/sync-cloudflare/worker-configuration.d.ts
@@ -3,7 +3,9 @@
 // Runtime types generated with workerd@1.20250617.0 2025-05-08 nodejs_compat
 declare namespace Cloudflare {
 	interface Env {
-		TLDRAW_DURABLE_OBJECT: DurableObjectNamespace<import('./worker/worker').TldrawDurableObject>
+		TLDRAW_DURABLE_OBJECT: DurableObjectNamespace<
+			import('./worker/worker').TldrawDurableObjectSqlite
+		>
 		TLDRAW_BUCKET: R2Bucket
 	}
 }


templates/sync-cloudflare/wrangler.toml
```diff
@@ -20,18 +20,25 @@ directory = "./dist/client"
...
 [[migrations]]
 tag = "v1"
 new_classes = ["TldrawDurableObject"]
 
+# v2 was a no-op: classes_with_sqlite is not a valid Cloudflare migration directive
 [[migrations]]
 tag = "v2"
 classes_with_sqlite = ["TldrawDurableObject"]
 
+# New SQLite-backed class to replace KV-backed TldrawDurableObject
```
MitjaBezensek, line 36: Think we needed to create a new class:
> You cannot enable a SQLite storage backend on an existing, deployed Durable Object class, so setting new_sqlite_classes on later migrations will fail with an error. Automatic migration of deployed classes from their key-value storage backend to SQLite storage backend will be available in the future.

https://developers.cloudflare.com/durable-objects/reference/durable-objects-migrations/

PR #7836, opened 2026-02-04
eagle vision: tweaks to zoom/cursor/edge scrolling

mimecuvalo:
couple tweaks to eagle vision 🦅 
- zoom to fit instead of 5% - feels like 5% is too high of a limit for a lot of boards, and then you can't actually see what you're zooming into, reduces the usability of the feature
- using zoom to fit fixes the position being lost when going in/out of eagle vision
- allow "quick peek" - don't updateBrush until the user has the intention to actually go to a different part of the board
- brings back the logic to fix the cursor being (-) if alt key is being pressed, from original PR
- adds edge scrolling

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### API changes

Added `quickZoomPreservesScreenBounds` to `TldrawOptions` (default `true`). When true, the quick-zoom brush maintains constant screen-pixel size as the user zooms the overview — zooming in shrinks the target viewport, zooming out expands it. Set to `false` to keep fixed page dimensions (original behavior).

```tsx
<Tldraw options={{ quickZoomPreservesScreenBounds: false }} />
```

### Release notes

- eagle vision: tweaks to zoom/cursor/edge scrolling

---

> [!NOTE]
> **Medium Risk**
> Touches core viewport/camera behavior (quick-zoom camera math, brush sizing, and edge-scrolling conditions), so regressions could affect navigation/interaction feel across tools, though changes are well-covered by updated tests.
> 
> **Overview**
> Quick zoom (`zoom.zoom_quick`) is reworked to compute a *zoom-to-fit* overview (instead of a fixed 5% zoom), preserve the page point under the cursor when zooming out, and only “commit” to moving/teleporting after crossing a drag threshold (idle vs moving), returning to the original viewport if released while idle.
> 
> Adds `quickZoomPreservesScreenBounds` to `TldrawOptions` (default `true`) to control whether the quick-zoom brush keeps a constant screen-pixel size as the user zooms the overview, plus UI/translation plumbing to show “Quick zoom” (`shift+z`) in the keyboard shortcuts dialog.
> 
> Edge scrolling is tightened to respect camera lock in `EdgeScrollManager`, and select-tool tick handlers now call edge scrolling only while actively dragging and not panning; ZoomTool cursor logic also avoids showing zoom-out while in quick-zoom even if Alt is held. Tests are updated/expanded to cover the new quick-zoom state machine, cursor/exit semantics, and the new option behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 44f0267a62d062a39bc8ab5ae8a751c6cec37b9d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


steveruizok:
I still think the right zoom is minimum zoom (5%), preserving your cursor position, rather than zoom to fit. Its strange to zoom out and press this key and end up zooming back in, and on large rooms (where the whole canvas won't fit in 5% viewport) it feels strange to have the zoomed-out viewport centered on nothing. I would want to replicate the "zoomed out temporarily to minimum from my cursor location", though perhaps the right number isn't 5%?

steveruizok:
Ok, more work on this... I think we were conflating the zoom brush and the mode of the interaction. Splitting this up so that we can show the zoom brush immediately, while still preserving the "require move before ending at the new bounds". Also switching from pointer moves to pointer position offsets to account for wheeling.

packages/tldraw/src/lib/tools/ZoomTool/childStates/ZoomQuick.ts
```diff
@@ -1,34 +1,47 @@
...
+
+		editor.setCursor({ type: 'zoom-in', rotation: 0 })
+
+		// Zoom to the 2nd zoom step while preserving cursor position
+		// If already below that, zoom to the 1st zoom step instead
+		const baseZoom = editor.getBaseZoom()
+		const { zoomSteps } = editor.getCameraOptions()
+		const currentZoom = editor.getCamera().z
+		const targetZoom =
+			currentZoom < zoomSteps[1] * baseZoom ? zoomSteps[0] * baseZoom : zoomSteps[1] * baseZoom
```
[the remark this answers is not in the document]
steveruizok: @claude fix

mimecuvalo:
> I still think the right zoom is minimum zoom (5%), preserving your cursor position, rather than zoom to fit. Its strange to zoom out and press this key and end up zooming back in, and on large rooms (where the whole canvas won't fit in 5% viewport) it feels strange to have the zoomed-out viewport centered on nothing. I would want to replicate the "zoomed out temporarily to minimum from my cursor location", though perhaps the right number isn't 5%?

Sounds like we need to some design discussion/writer's room in person like I mentioned on Discord. I'm finding the 10%/5% choice unsatisfying in lots of cases, can show you when we meet.

Might be a HMR thing but I'm also finding that sometimes the tool gets stuck in Zoom mode — have you seen this?

packages/editor/src/lib/editor/managers/EdgeScrollManager/EdgeScrollManager.ts
```diff
@@ -108,8 +108,8 @@ export class EdgeScrollManager {
 	private moveCameraWhenCloseToEdge(proximityFactor: { x: number; y: number }) {
 		const { editor } = this
 		if (
-			!editor.inputs.getIsDragging() ||
-			editor.inputs.getIsPanning() ||
+			(!editor.isIn('zoom.zoom_quick') &&
```
mimecuvalo: i made this check the instanceState.zoomBrush b/c technically this is at the editor level and the editor doesn't know about the zoom tool (or at least it shouldn't), since it's at the tldraw package level

steveruizok: ah great catch, that's right. Do we really need this check? I thought it should be the caller that decides when to call this function (based on whatever local state).

mimecuvalo: edge scrolling wouldn't work without it - unless you want to a slightly bigger refactor to the EdgeScrollManager some more

steveruizok:
> Might be a HMR thing but I'm also finding that sometimes the tool gets stuck in Zoom mode — have you seen this?

I have, yeah.

packages/editor/src/lib/editor/managers/EdgeScrollManager/EdgeScrollManager.ts
steveruizok: I'll move the guards into the call sites so you can see what I mean.

steveruizok:
Removing edge scrolling from this feature, since we can't get the user's mouse position after it leaves the window (the user has not clicked so we can't capture the pointer). We could track the mouseleave and get an estimated exit position based on the last known position and the last known velocity but that feels like a separate system.

steveruizok:
> Sounds like we need to some design discussion/writer's room in person like I mentioned on Discord. I'm finding the 10%/5% choice unsatisfying in lots of cases, can show you when we meet.

My rationale here is:

- I want to reproduce the same effect as "zooming out from the current pointer position", so its important to preserve the page position of the pointer. This gives a lot of control to the user about where the camera will be.
- I want to allow the user to select new empty space

In practice, a user would probably zoom out to the level that lets them see everything on the screen and then stop there rather than going out to 5%. That would be a kind of expanded union of the common page shape bounds and the user's viewport, that would still preserve their pointer page position. :psyduck:

steveruizok:
Ok, some changes summarized. This should replicate the "zoom out from my mouse cursor until everything is on screen".

### More realistic zoom level

The old `ZoomQuick` zoomed to a fixed 5% and used a small brush (1/4 screen size). The new version:

- **Zoom-to-fit overview**: computes the minimum zoom needed to show the union of the current viewport and all page shapes, preserving the cursor's page position. Applies a 0.85x multiplier for breathing room, clamped to the minimum zoom step.
- **Brush matches original viewport size**: instead of a small fixed brush, the brush dimensions equal the initial viewport page bounds, so you're choosing *where* to look at the same zoom level.
- **Cursor-proportional positioning**: the brush offset within the viewport is normalized to the cursor's screen position, so the brush tracks naturally as you move.

### Viewport while zooming / `quickZoomPreservesScreenBounds` option

New `TldrawOptions` field (default `true`). A reactive effect watches `editor.getCamera().z` and rescales the brush page dimensions by `overviewZoom / currentZoom` whenever the overview zoom changes (wheel, pinch, shortcuts). This keeps the brush the same pixel size on screen — zooming in on the overview shrinks the target viewport (higher return zoom), zooming out expands it. Set to `false` to keep fixed page dimensions regardless of overview zoom changes.

### Edge scrolling cleanup

- `EdgeScrollManager.updateEdgeScrolling` now bails immediately when the camera is locked, rather than checking `isLocked` deeper in `moveCameraWhenCloseToEdge`.
- The `isDragging`/`isPanning` guard was moved out of `moveCameraWhenCloseToEdge` and into each caller (`Brushing`, `Resizing`, `Translating` `onTick` handlers), so edge scrolling only ticks while actually dragging — not while panning with spacebar.

### ZoomTool cursor fix

`ZoomTool.updateCursor` no longer shows the alt "zoom-out" cursor while in `zoom_quick` state — it always shows "zoom-in" during quick zoom regardless of alt key state.

steveruizok:
Landing this. The remaining issues are UX opinions but I'm reasonably happy with them, enough to learn more in the hand. My one thing is whether the zoom really should just be  10% / 5% in order to be more static and predictable, rather than dynamic based on where the cursor is on the screen.

mimecuvalo:
we're discussing at the moment!

PR #7840, opened 2026-02-04
perf(editor): reduce allocations and hash computation overhead

### LINKED ISSUE #7833: ImmutableMap hashes short string keys on every lookup
## Summary

Shape ID strings are hashed on every `ImmutableMap.get()` call because they're shorter than the 16-character cache threshold. This causes significant overhead in hot paths like visibility culling during camera panning.

## Problem

The `ImmutableMap` implementation (extracted from Immutable.js) uses a string hash cache, but only for strings >= 16 characters:

```typescript
const STRING_HASH_CACHE_MIN_STRLEN = 16

case 'string':
  return v.length > STRING_HASH_CACHE_MIN_STRLEN 
    ? cachedHashString(v)  // cached
    : hashString(v)        // NOT cached - called every time
```

Shape IDs like `shape:abc123` are typically ~14 characters, so they **always bypass the cache**. Every `store.get(shapeId)` re-computes the hash.

## Impact

In derived computations that iterate over shapes (e.g., `notVisibleShapes`), we call `editor.getShape(id)` for each shape on the page. With 100 shapes at 60fps, this results in hundreds of redundant hash computations per frame during camera panning.

Call chain:
```
editor.getShape(id)
  → store.get(id)
    → AtomMap.getAtom(key)
      → ImmutableMap.get(key)
        → BitmapIndexedNode.get(shift, keyHash, key)
          → if (keyHash === undefined) keyHash = hash(key)
            → hashString(key)  // every single time
```

## Potential solutions

1. **Lower `STRING_HASH_CACHE_MIN_STRLEN`** - Quick fix, trades memory for speed. The threshold exists because cache lookup has overhead, but for repeated lookups of the same keys it's a net win.

2. **Use branded ID objects with `hashCode()` method** - The hash function already fast-paths objects with a `hashCode()` method. Shape IDs could be objects that pre-compute their hash once.

3. **Use native `Map` for `AtomMap` internals** - Native Map uses internal object identity, no explicit hashing. Would need to maintain immutability semantics another way.

4. **Cache at the lookup layer** - Add memoization in `AtomMap.get()` or `Store.get()` for hot-path access patterns.

## Files involved

- `packages/store/src/lib/ImmutableMap.ts` - Hash function and cache threshold (lines 87-100, 149)
- `packages/store/src/lib/AtomMap.ts` - Uses ImmutableMap for key storage
- `packages/store/src/lib/Store.ts` - Uses AtomMap for record storage

steveruizok:
In order to improve rendering performance in large canvases and multiplayer rooms, this PR bundles three related performance optimizations that reduce memory allocations and computation overhead in hot paths.

Closes #7833

**Dragging a single shape on a canvas with ~2.5K shapes**:

Before: 
image

After:
image

### Changes (in order of gains)

**Replace lodash `isEqual` with optimized Set comparisons**

`lodash.isEqual` doesn't handle Sets efficiently - it likely converts them to arrays for deep comparison. With thousands of shapes in large multiplayer rooms, this caused significant overhead during the equality checks in computed signals. The new comparison functions iterate Sets directly with early-exit on size mismatch.

Files: `CanvasShapeIndicators.tsx`, `usePeerIds.ts`

**Avoid unnecessary Set allocations in `notVisibleShapes`**

The `notVisibleShapes` derivation was creating a new Set on every recompute even when the result hadn't changed. This PR adds:
- A fast path returning a cached empty set when all shapes are visible
- Change detection before allocating a new Set
- Return of the previous value when contents are unchanged

This reduces GC pressure during panning and zooming.

`notVisibleShapes.ts`

**Simplify string hash caching in ImmutableMap**

Shape IDs (~14 characters) were bypassing the string hash cache because of a 16-character minimum threshold. This caused redundant hash computations on every `store.get(shapeId)` call. The fix:
- Caches all strings regardless of length
- Removes cache eviction logic (strings are bounded by document size)

`ImmutableMap.ts`

### Change type

- [x] `improvement`

### Test plan

1. Open a canvas with many shapes (100+)
2. Pan and zoom around - should feel smoother
3. In a multiplayer room with multiple users, verify no regressions

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve performance in large canvases and multiplayer rooms by optimizing Set comparisons, reducing memory allocations, and caching string hashes

---

> [!NOTE]
> **Medium Risk**
> Touches core rendering/culling and reactive equality logic; while behavior should be equivalent, mistakes could cause shapes/indicators to not update or to flicker under certain viewport/collaboration states.
> 
> **Overview**
> Improves large-canvas rendering performance by **centralizing shape culling**: shapes now register their DOM containers into a new `ShapeCullingProvider`, and `DefaultCanvas` runs a single `CullingController` reactor to toggle `display` for all culled shapes instead of each `Shape` owning its own culling subscription.
> 
> Reduces reactive overhead in multiplayer/indicator code by replacing deep equality (`isEqual`) with purpose-built Set/array comparators for computed render data (`CanvasShapeIndicators`) and active peer IDs (`useActivePeerIds$`), and reduces allocations in `notVisibleShapes` via fast paths and returning the previous Set when unchanged.
> 
> Cuts store hashing work by changing `ImmutableMap` to **always** cache string hashes (with a larger bounded cache), avoiding repeated hash computations for short string keys (e.g. shape IDs).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f1ddcae0704dd583c128db9d4601d494c718865. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
+++ b/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
@@ -38,40 +38,27 @@ export function notVisibleShapes(editor: Editor) {
 			return nextValue
 		}
 
-		// Subsequent runs: check if result differs before creating new Set
-		let count = 0
-		let hasDiff = false
-
+		// Subsequent runs: single pass to collect IDs and detect changes
+		const notVisibleIds: TLShapeId[] = []
 		for (let i = 0; i < allShapes.length; i++) {
 			shape = allShapes[i]
 			if (visibleIds.has(shape.id)) continue
 			if (!editor.getShapeUtil(shape.type).canCull(shape)) continue
+			notVisibleIds.push(shape.id)
+		}
 
-			count++
-			if (!hasDiff && !prevValue.has(shape.id)) {
-				hasDiff = true
+		// Check if the result changed
+		if (notVisibleIds.length === prevValue.size) {
+			let same = true
+			for (let i = 0; i < notVisibleIds.length; i++) {
+				if (!prevValue.has(notVisibleIds[i])) {
+					same = false
+					break
+				}
 			}
+			if (same) return prevValue
 		}
 
-		// Check if size changed
-		if (!hasDiff && prevValue.size !== count) {
-			hasDiff = true
-		}
-
-		// If nothing changed, return the previous value
-		if (!hasDiff) {
-			return prevValue
-		}
-
-		// Build the new Set (only when needed)
-		const nextValue = new Set<TLShapeId>()
-		for (let i = 0; i < allShapes.length; i++) {
-			shape = allShapes[i]
-			if (visibleIds.has(shape.id)) continue
-			if (!editor.getShapeUtil(shape.type).canCull(shape)) continue
-			nextValue.add(shape.id)
-		}
-
-		return nextValue
+		return new Set(notVisibleIds)
 	})
 }


packages/store/src/lib/ImmutableMap.ts
```diff
@@ -73,11 +72,6 @@ function cachedHashString(string: string) {
 	let hashed = stringHashCache[string]
 	if (hashed === undefined) {
 		hashed = hashString(string)
-		if (STRING_HASH_CACHE_SIZE === STRING_HASH_CACHE_MAX_SIZE) {
```
MitjaBezensek, line 76: This could lead to memory issues. `stringHashCache` would never get cleaned up, would take in ids from shapes from different pages, even different documents. Let's at least set an upper bound, could even be some high number.

packages/editor/src/lib/components/default-components/CanvasShapeIndicators.tsx
```diff
@@ -14,6 +14,50 @@ interface CollaboratorIndicatorData {
...
 }
 
+interface RenderData {
+	idsToDisplay: Set<TLShapeId>
+	renderingShapeIds: Set<TLShapeId>
+	hintingShapeIds: TLShapeId[]
+	collaboratorIndicators: CollaboratorIndicatorData[]
+}
+
+function setsEqual<T>(a: Set<T>, b: Set<T>): boolean {
```
MitjaBezensek, line 24: Same function here, let's reuse it instead of defining it again:
https://github.com/tldraw/tldraw/blob/8edd4041a9869dc36df6c33990cf822ff3606bef/packages/editor/src/lib/hooks/usePeerIds.ts#L10-L16

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -9,34 +9,69 @@ import { Editor } from '../Editor'
...
+		if (!hasDiff) {
+			return prevValue
+		}
+
+		// Build the new Set (only when needed)
+		const nextValue = new Set<TLShapeId>()
+		for (let i = 0; i < allShapes.length; i++) {
+			shape = allShapes[i]
+			if (visibleIds.has(shape.id)) continue
+			if (!editor.getShapeUtil(shape.type).canCull(shape)) continue
```
mimecuvalo: i think doing this canCull check twice might be a bit much.
you could still collect the id's above in the first loop, but in an array instead?

packages/editor/src/lib/hooks/usePeerIds.ts
```diff
@@ -8,6 +7,14 @@ import {
 import { uniq } from '../utils/uniq'
 import { useEditor } from './useEditor'
 
+function setsEqual<T>(a: Set<T>, b: Set<T>): boolean {
+	if (a.size !== b.size) return false
+	for (const item of a) {
+		if (!b.has(item)) return false
+	}
+	return true
+}
```
mimecuvalo, line 16: agreed with Cursor, would be great to have a util file for this sort of thing

packages/store/src/lib/ImmutableMap.ts
```diff
@@ -145,10 +139,7 @@ const symbolMap = Object.create(null)
 
 let _objHashUID = 0
 
-const STRING_HASH_CACHE_MIN_STRLEN = 16
-const STRING_HASH_CACHE_MAX_SIZE = 255
-let STRING_HASH_CACHE_SIZE = 0
-let stringHashCache: Record<string, number> = {}
+const stringHashCache: Record<string, number> = {}
```
mimecuvalo: i don't think we should make this unbounded. either use an LRU or have a max that's closer to our max shape count

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -9,34 +9,69 @@ import { Editor } from '../Editor'
...
 
-		for (const prev of prevValue) {
-			if (!nextValue.has(prev)) return nextValue
+		// If nothing changed, return the previous value
+		if (!hasDiff) {
+			return prevValue
+		}
+
+		// Build the new Set (only when needed)
+		const nextValue = new Set<TLShapeId>()
```
MitjaBezensek: This looks to me like we are only optimizing for the case where we have no changes?
1. It speeds up the use case where not much work is needed (no shape visibility changes).
2. It makes it slower for cases when work is needed (visibility changes) as we now have to allocate and have a more complex loop over more shapes.

In simple pans (which show / hide additional shapes) we will always do an allocation as well as another loop after this change. So it feels like we are making things worse for the case where we already have to do more work.

packages/editor/src/lib/hooks/useShapeCulling.tsx
MitjaBezensek, line 1: Nice, I was also thinking about something like this 👍

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -9,34 +9,69 @@ import { Editor } from '../Editor'
...
+		let hasDiff = false
+
+		for (let i = 0; i < allShapes.length; i++) {
+			shape = allShapes[i]
+			if (visibleIds.has(shape.id)) continue
+			if (!editor.getShapeUtil(shape.type).canCull(shape)) continue
+
+			count++
+			if (!hasDiff && !prevValue.has(shape.id)) {
+				hasDiff = true
```
MitjaBezensek: We can break from the loop here. We detected a diff, we don't need to do the full count in this case as we already know we'll have to allocate.

packages/store/src/lib/ImmutableMap.ts
mimecuvalo: also, in a SPA, i'm pretty sure this would presist across pages/boards, no? 😬

MitjaBezensek: Yeah, it definitely would. It's a global, so only a refresh would reset it.

Issue #7755, opened 2026-01-28
TypeScript OOM when using TLGlobalShapePropsMap with complex discriminated union prop types


thesnups:
# TypeScript OOM when using TLGlobalShapePropsMap with complex discriminated union prop types

## Summary

After upgrading to tldraw 4.3.0 and using the new `TLGlobalShapePropsMap` module augmentation pattern, TypeScript runs out of memory when custom shape props contain deeply nested discriminated unions. We isolated the OOM to the `TLIndexedShapes` mapped type and found a low‑blast‑radius fix that preserves **literal shape typing** while avoiding expensive union expansion for the **unspecified** `TLShape` case. Because that fix makes `TLShape` cheap by default, we also propose a small Editor‑API typing tweak that restores automatic `createShape` / `updateShape` inference without re‑introducing the OOM.

## Root Cause

**Source file:** `packages/tlschema/src/records/TLShape.ts` ([link](https://github.com/tldraw/tldraw/blob/v4.3.0/packages/tlschema/src/records/TLShape.ts#L88-L115))

The issue is in `TLIndexedShapes`:

```typescript
export declare type TLIndexedShapes = {
    [K in keyof TLGlobalShapePropsMap | TLDefaultShape['type'] as ...]: 
      ... TLBaseShape<K, TLGlobalShapePropsMap[K]> ...
};
```

**Problem:** `TLIndexedShapes` is a **mapped type over a union of keys**. TypeScript eagerly evaluates all values in mapped types when any part of the type is accessed. When `TLGlobalShapePropsMap[K]` contains a complex discriminated union (e.g., `CustomShapeProps`), the mapped type forces evaluation of `TLBaseShape<..., CustomShapeProps>` and triggers exponential type expansion. This is amplified by:
- `TLShape` defaulting to `keyof TLIndexedShapes` (so the compiler eagerly expands the full union).
- Utility types like `ExtractShapeByProps` and unions referencing `TLShape` (they trigger distribution across the full union).

**Tested fixes that did NOT work:**
1. Extracting computation into a helper type `TLComputeShape<K>` - still OOM
2. Removing the default type parameter from `TLShape` - still OOM
3. Simplifying `TLAnyShape` / `TLRecord` / `ExtractShapeByProps` alone - still OOM
4. Rewriting `TLIndexedShapes` with “lazy” conditional types only - still OOM

This confirms the issue is fundamental to mapped type evaluation and the `TLShape` union default, not just how `TLShape` is used in isolation.

[...]

PR #7843, opened 2026-02-04
Use a different formulation of `TLShape` and `TLBinding`

Andarist:
### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

I do believe the provided repro case hits a TS issue (I'll be reporting it) but in the meanime we can use a different formulation of those types to fix the reported issue. 

### Test plan

- [x] manual tests in the provided repro

### Release notes

_placeholder_

fixes https://github.com/tldraw/tldraw/issues/7755

---

> [!NOTE]
> **Medium Risk**
> Type-level refactor in exported schema types (`TLShape`, `TLBinding`) and removal/rename of the underlying indexed-map aliases may break downstream TypeScript builds that relied on the old (undocumented) types, though there are no runtime behavior changes.
> 
> **Overview**
> Refactors the `@tldraw/tlschema` type model for `TLShape` and `TLBinding` to avoid a TypeScript inference issue by switching from indexed-map aliases (`TLIndexedShapes`/`TLIndexedBindings`) to value-union aliases (`TLUnionizedShapes`/`TLUnionizedBindings`) and defining `TLShape`/`TLBinding` via `Extract<...,{type: K}>`.
> 
> Adds a new shared utility type `Values<T> = T[keyof T]` (exported from `@tldraw/utils`) and updates the public exports/API reports accordingly; test JSX helpers update their binding creation conditional type to a distributive `_BindingToCreate` helper to match the new formulation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b577f0e38679377934df4717305124898ac6d6b5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tlschema/src/records/TLShape.ts
+++ b/packages/tlschema/src/records/TLShape.ts
@@ -5,7 +5,7 @@ import {
 	createRecordMigrationSequence,
 	createRecordType,
 } from '@tldraw/store'
-import { mapObjectMapValues, uniqueId } from '@tldraw/utils'
+import { Values, mapObjectMapValues, uniqueId } from '@tldraw/utils'
 import { T } from '@tldraw/validate'
 import { SchemaPropsInfo } from '../createTLSchema'
 import { TLPropsMigrations } from '../recordsWithProps'
@@ -85,7 +85,7 @@ export interface TLGlobalShapePropsMap {}
 
 /** @public */
 // prettier-ignore
-export type TLIndexedShapes = {
+export type TLUnionizedShapes = Values<{
 	// We iterate over a union of augmented keys and default shape types.
 	// This allows us to include (or conditionally exclude or override) the default shapes in one go.
 	//
@@ -112,7 +112,7 @@ export type TLIndexedShapes = {
 					Extract<TLDefaultShape, { type: K }>
 			: // use the custom shape definition
 				TLBaseShape<K, TLGlobalShapePropsMap[K & keyof TLGlobalShapePropsMap]>
-}
+}>
 
 /**
  * The set of all shapes that are available in the editor.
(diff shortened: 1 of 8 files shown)


packages/tlschema/src/records/TLShape.ts
```diff
@@ -143,7 +143,10 @@ export type TLIndexedShapes = {
  *
  * @public
  */
-export type TLShape<K extends keyof TLIndexedShapes = keyof TLIndexedShapes> = TLIndexedShapes[K]
+export type TLShape<K extends TLUnionizedShapes['type'] = TLUnionizedShapes['type']> = Extract<
```
Andarist, line 146: this is the core of the fix - I just changed this to use `Extract` from a union instead of looking up the result through an indexed access (they should really be interchangeable here)

thesnups:
Thank you so much for your help!

packages/tlschema/src/records/TLShape.ts
```diff
@@ -85,7 +85,7 @@ export interface TLGlobalShapePropsMap {}
 
 /** @public */
 // prettier-ignore
-export type TLIndexedShapes = {
+export type TLUnionizedShapes = Values<{
```
mimecuvalo, line 88: two options here:
- to avoid having a breaking change, we could make `type TLIndexedShapes = TLUnionizedShapes` although i guess they're not exactly the same.
- or, in the PR comment, let's call out that this is technically a breaking API change

packages/tlschema/src/records/TLShape.ts
```diff
@@ -85,7 +85,7 @@ export interface TLGlobalShapePropsMap {}
...
 	// We iterate over a union of augmented keys and default shape types.
```
mimecuvalo, line 89: related to the changes, can you please add `### API changes` to the PR description? this will make it pass the Huppy check.

packages/tlschema/src/records/TLShape.ts
mimecuvalo: elegant change!

Andarist:
I'm closing for the reasons mentioned in https://github.com/tldraw/tldraw/issues/7755#issuecomment-3869775806

PR #7849, opened 2026-02-05
Display an indicator to alert on local(anon) large file/session

kaneel:
Adds a local file handler that will check periodically wether a file is getting quite large or if session spent since mount exceed a certain time.

The motivation here is to nudge users to create an account or logging to slurp file before they reach a point where slurping may fail.

It will:

- Checks file size and time session on mount + periodically every `CHECK_INTERVAL_MS`
- Display a tiny face that express our concerns over file size or sesison time 
- On click on the faces, a dialog appears explaining a bit more (should we add a button for direct login?)
- On acknowledgement (`onClose`) of the modal we remember the time in localStorage and won't pester user again
- On every future interval, we carefully check if the user has already acknowledged one or the other dialogs

Right now this is still a draft, I need to check wether or not this is what we want, how we want it, and I need to make sure the constant numbers are well thought + actually get real icons for it (I am nearing the moment I actually am enjoying them too much :D)

BIG DOC!
image

OLD DOC!
Screenshot 2026-02-05 at 11 48 21

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
2.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

---

> [!NOTE]
> **Medium Risk**
> Adds new periodic checks and UI dialogs in the anonymous editor, which could introduce performance or UX regressions if thresholds/asset sizing are off, but it’s scoped to anon/local state only.
> 
> **Overview**
> Adds a new `LocalFileHandler` for anonymous/scratch-pad users that periodically checks local document size (shape count and asset bytes) and session duration, then shows a small indicator that opens a warning dialog prompting the user to sign in to save their work.
> 
> The handler persists acknowledgement timestamps in `localStorage` (with expiry) and prioritizes large-file warnings over long-session warnings. This is wired into the anon top-right editor panel, adds new i18n strings for the dialog copy, and includes Playwright E2E coverage for indicator visibility, priority, and acknowledgement/expiry behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4aa7e466184e6a783d9f5f299fab9e40f84ec4cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


kaneel:

apps/dotcom/client/src/components/local_file_handler.constants.ts
mimecuvalo, line 1: dashes instead of snake case. or i might just create a folder called `LocalFile` and put relevant local files in there:
- LocalFileHandler.tsx
- LocalEditor.tsx
- constants.ts

apps/dotcom/client/src/components/local_file_handler.constants.ts
```diff
@@ -0,0 +1,8 @@
+export const MAX_SHAPES = 4000
```
mimecuvalo: instead of creating another MAX_SHAPES - let's use what the editor option has for `maxShapesPerPage`

apps/dotcom/client/src/components/local_file_handler.constants.ts
```diff
@@ -0,0 +1,8 @@
...
+export const MAX_ASSETS_MB = 100
+export const MAX_SESSION_TIME_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const ACKNOWLEDGEMENT_EXPIRY_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const CHECK_INTERVAL_MS = 1000 * 60 * 10
```
mimecuvalo: nit: `// 10 minutes`

apps/dotcom/client/src/components/local_file_handler.constants.ts
```diff
@@ -0,0 +1,8 @@
+export const MAX_SHAPES = 4000
+export const MAX_ASSETS_MB = 100
```
mimecuvalo: let's use `DEFAULT_MAX_ASSET_SIZE` instead

apps/dotcom/client/src/components/local_file_handler.constants.ts
```diff
@@ -0,0 +1,8 @@
...
+export const MAX_SESSION_TIME_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const ACKNOWLEDGEMENT_EXPIRY_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const CHECK_INTERVAL_MS = 1000 * 60 * 10
+
+export const LOCAL_STORAGE_KEY_FILESIZE = 'tldraw_localFileLargeSizeAcknowledged'
+export const LOCAL_STORAGE_KEY_SESSION = 'tldraw_localFileSessionAcknowledged'
```
mimecuvalo: i think our storage keys use snake case actually

apps/dotcom/client/public/tla/locales/en.json
```diff
@@ -474,6 +480,9 @@
   "cea291e45e": {
     "translation": "Learn more about publishing."
   },
+  "d01bd00f78": {
+    "translation": "This file is getting quite large, maybe consider saving it by creating an account or logging in!"
```
mimecuvalo, line 484: nit: "This file is getting large. Consider saving it by creating an account or logging in."
@steveruizok might have some other wording on this

apps/dotcom/client/public/tla/locales/en.json
```diff
@@ -420,6 +423,9 @@
   "b758336699": {
     "translation": "Copy link"
   },
+  "b995b72679": {
+    "translation": "You've been working on this file for a while, maybe consider saving it by creating an account or logging in!"
```
mimecuvalo, line 427: nit: `’` for apostrophe - typography nerd here, hah.
"You've been working on this file for a while. Consider consider saving it by creating an account or logging in."

apps/dotcom/client/public/tla/locales/en.json
```diff
@@ -62,6 +62,9 @@
   "1dd1c5fb7f": {
     "translation": "Today"
   },
+  "2319670994": {
+    "translation": "Been a long time!"
```
mimecuvalo, line 66: this might be too vague :P
or, to be concrete, i don't think this will translate very well

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditorTopRightPanel.tsx
```diff
@@ -47,6 +48,7 @@ export function TlaEditorTopRightPanel({
 	if (isAnonUser) {
 		return (
 			<div ref={ref} className={styles.topRightPanel}>
+				<LocalFileHandler />
```
mimecuvalo, line 51: hm, since this is actually not being referenced by the LocalEditor, then I think we can put the LocalFileHandler in the TlaEditor folder instead

apps/dotcom/client/styles/globals.css
```diff
@@ -488,3 +488,13 @@ a {
 	/* https://stackoverflow.com/a/23735769/1810018 */
 	-webkit-transform: translateZ(0);
 }
+
+/* ------------------- File indicator ------------------- */
+
+.tlui-file-indicator {
```
mimecuvalo: use `TldrawUiButtonIcon` instead of restyling a new icon button

apps/dotcom/client/src/components/LocalFileHandler.tsx
```diff
@@ -0,0 +1,420 @@
...
+					<TldrawUiButtonLabel>
+						<F defaultMessage="Got it" />
+					</TldrawUiButtonLabel>
+				</TldrawUiButton>
+			</TldrawUiDialogFooter>
+		</>
+	)
+}
+
+function ChungusBoi() {
```
mimecuvalo: let's put the icons in a separate file - otherwise makes this file too big hah

apps/dotcom/client/src/components/LocalFileHandler.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+		const pages = editor.getPages()
+		const totalShapes = pages.reduce((sum, page) => sum + editor.getPageShapeIds(page).size, 0)
+		const lastFileSizeAcknowledged = Number(getFromLocalStorage(LOCAL_STORAGE_KEY_FILESIZE) ?? '0')
+		const lastTimeSessionAcknowledged = Number(
+			getFromLocalStorage(LOCAL_STORAGE_KEY_SESSION) ?? '0'
+		)
+
+		if (Date.now() - lastFileSizeAcknowledged > ACKNOWLEDGEMENT_EXPIRY_MS) {
+			const totalAssetsSizeInMB = editor.getAssets().reduce((acc, asset) => {
```
mimecuvalo: i think the assets size shouldn't matter at all for the slurping. was this something Steve was asking for?

apps/dotcom/client/src/components/LocalFileHandler.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+export function LocalFileHandler() {
+	const isMounted = useRef(false)
+	const intervalRef = useRef<number | undefined>(undefined)
+	const mountTime = useRef(0)
+	const [localFileState, setLocalFileState] = useState<LocalFileState>(LocalFileState.NONE)
+	const editor = useEditor()
+
+	const handleFileSizeAcknowledged = useCallback(() => {
+		setLocalFileState(LocalFileState.NONE)
```
mimecuvalo: does this just warn once and then never again? 🤔 
or does it ask again in 10minutes?

apps/dotcom/client/src/components/LocalFileHandler.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+		if (Date.now() - lastFileSizeAcknowledged > ACKNOWLEDGEMENT_EXPIRY_MS) {
+			const totalAssetsSizeInMB = editor.getAssets().reduce((acc, asset) => {
+				if (asset.type === 'image' || asset.type === 'video') {
+					return acc + (asset.props.fileSize ?? 0)
+				}
+				return acc
+			}, 0)
+
+			if (totalShapes > MAX_SHAPES || totalAssetsSizeInMB > MAX_ASSETS_MB) {
```
mimecuvalo: hm, totalShapes should never get over MAX_SHAPES, if we're handling it correctly in the editor.
this should be checking something like `totalShapes > 0.9 * MAX_SHAPES` or something like that

apps/dotcom/client/public/tla/locales/en.json
kaneel: agreed, very much done in a rush to validate the behaviour but it requires some proper wording

apps/dotcom/client/src/components/LocalFileHandler.tsx
kaneel: This was indeed discussed, and that's why I insisted on commiting this one as a draft as it really calls for discussion on what we consider important to communicate to users (as in: what becomes "incentive to create an account")

apps/dotcom/client/src/components/LocalFileHandler.tsx
kaneel: I've set a constant for this, acknowledgment expires every 24 hours for now, subject to change

apps/dotcom/client/src/components/LocalFileHandler/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const DEFAULT_MAX_ASSET_SIZE = 100
```
mimecuvalo, line 1: use the `DEFAULT_MAX_ASSET_SIZE` defined in the `tldraw` package - you can rm this one here

apps/dotcom/client/src/components/LocalFileHandler/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+export const MAX_SESSION_TIME_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const ACKNOWLEDGEMENT_EXPIRY_MS = 24 * 60 * 60 * 1000 // 24 hours
+export const CHECK_INTERVAL_MS = 1000 * 60 * 10 // 10 minutes
+
+export const LOCAL_STORAGE_KEY_FILESIZE = 'localFileLargeSizeAcknowledged'
```
mimecuvalo, line 6: snake-case i believe for these keys

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const handleClick = useCallback(() => {
+		addDialog({
+			id: 'session-time-warning',
+			onClose: onClose,
+			component: ({ onClose }) => <SessionTimeDialog onClose={onClose} />,
+		})
+	}, [onClose, addDialog])
+
+	return (
+		<button
```
mimecuvalo, line 142: use `<TldrawUiButton type="icon"` pls

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.module.css
```diff
@@ -0,0 +1,7 @@
+.indicatorButton {
```
mimecuvalo, line 1: we can nix this styling

apps/dotcom/client/src/components/LocalFileHandler/dialogs.tsx
```diff
@@ -0,0 +1,56 @@
...
+	return (
+		<>
+			<TldrawUiDialogHeader>
+				<TldrawUiDialogTitle>
+					<F defaultMessage="File is getting large" />
+				</TldrawUiDialogTitle>
+				<TldrawUiDialogCloseButton />
+			</TldrawUiDialogHeader>
+			<TldrawUiDialogBody style={{ maxWidth: 350 }}>
+				<F defaultMessage="This file is getting quite large, maybe consider saving it by creating an account or logging in!" />
```
mimecuvalo, line 22: i left some comments previously on improving the wording here - let's apply those please!

apps/dotcom/client/src/components/LocalFileHandler/dialogs.tsx
```diff
@@ -0,0 +1,56 @@
...
+	return (
+		<>
+			<TldrawUiDialogHeader>
+				<TldrawUiDialogTitle>
+					<F defaultMessage="Been a long time!" />
+				</TldrawUiDialogTitle>
+				<TldrawUiDialogCloseButton />
+			</TldrawUiDialogHeader>
+			<TldrawUiDialogBody style={{ maxWidth: 350 }}>
+				<F defaultMessage="You've been working on this file for a while, maybe consider saving it by creating an account or logging in!" />
```
mimecuvalo, line 45: ditto

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const handleTimeSessionAcknowledged = useCallback(() => {
+		setLocalFileState(LocalFileState.NONE)
+		setInLocalStorage(LOCAL_STORAGE_KEY_SESSION, Date.now().toString())
+	}, [])
+
+	const handleShapesChecked = useCallback(async () => {
+		if (!isMounted.current) return
+
+		const pages = editor.getPages()
+		const maxShapes = 0.9 * editor.options.maxShapesPerPage
```
mimecuvalo, line 49: brief comment here would be good to understand what the 0.9 is doing for future devs

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+		const totalShapes = pages.reduce((sum, page) => sum + editor.getPageShapeIds(page).size, 0)
+		const lastFileSizeAcknowledged = Number(getFromLocalStorage(LOCAL_STORAGE_KEY_FILESIZE) ?? '0')
```
mimecuvalo, line 51: nit here: use `parseInt` instead of the Number constructor.

some notes on this: https://journal.artful.dev/what-you-risk-when-using-number-to-parse-an-integer-from-a-string-in-typescript/

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+		const lastTimeSessionAcknowledged = Number(
+			getFromLocalStorage(LOCAL_STORAGE_KEY_SESSION) ?? '0'
+		)
+
+		if (Date.now() - lastFileSizeAcknowledged > ACKNOWLEDGEMENT_EXPIRY_MS) {
+			const totalAssetsSizeInMB = editor.getAssets().reduce((acc, asset) => {
+				if (asset.type === 'image' || asset.type === 'video') {
```
mimecuvalo, line 58: actually i don't think we need to filter on asset.type (in the future we might have `audio` for example).

let's see if we can check if `fileSize` is present on props instead (i think maybe this was added for type safety)

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
mimecuvalo, line 49: mm, and we should rename `maxShapes` because that'll be confusing below. this isn't maxShapes anymore — it's more like `alertThresholdForShapes`, meh that kinda sounds awkward. i'd just put the 0.9 below in the if statement where you actually do the check

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+		setLocalFileState(LocalFileState.NONE)
+		setInLocalStorage(LOCAL_STORAGE_KEY_SESSION, Date.now().toString())
+	}, [])
+
+	const handleShapesChecked = useCallback(async () => {
+		if (!isMounted.current) return
+
+		const pages = editor.getPages()
+		const maxShapes = 0.9 * editor.options.maxShapesPerPage
+		const totalShapes = pages.reduce((sum, page) => sum + editor.getPageShapeIds(page).size, 0)
```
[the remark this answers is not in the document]
mimecuvalo: this is a valid concern too - let's address!

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const handleClick = useCallback(() => {
+		addDialog({
+			id: 'large-file-warning',
+			onClose: onClose,
+			component: ({ onClose }) => <LargeFileDialog onClose={onClose} />,
+		})
+	}, [onClose, addDialog])
+
+	return (
+		<button
```
mimecuvalo, line 119: `<TldrawUiButton type="icon"`

mimecuvalo (commented):
some more comments!

apps/dotcom/client/src/components/LocalFileHandler/constants.ts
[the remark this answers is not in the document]
kostyafarber: good shout

apps/dotcom/client/src/components/LocalFileHandler/LocalFileHandler.tsx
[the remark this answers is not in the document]
kaneel: Right, so this is confusing, because of https://github.com/tldraw/tldraw/pull/7849#discussion_r2773136982 I used maxShapesPerPage but we're not really trying to assess wether or not the user is adding the max amount of shapes per pages, we're trying to define what's actually a good number for assessing if users has enough pages to warrant an alert for "big document".

kaneel:
Not sure where this is going. 

I don't think we're 100% sure what we want to do, this was started as some exploration but not necesarily meant – from my point of view – to be merged. If we have an idea of what represents "a large document" or "a long session", we should put that into words somewhere. Same for dialogs, I'm not sure we really decided what it is we want to communicate and even if we want to communicate it.

PR #7858, opened 2026-02-06
feat(hand): add one-finger zoom gesture for touch devices

### LINKED ISSUE #7677: Add double-tap-and-drag zoom gesture for mobile
## Problem statement

Currently, zooming on mobile requires a two-finger pinch gesture. While this works well, it requires both hands or awkward single-hand positioning. Many mobile apps (Google Maps, Apple Maps, etc.) support an alternative single-handed zoom gesture: double-tap and hold, then drag up/down to zoom in/out.

## Proposed solution

Add support for a "double-tap-and-drag" zoom gesture on touch devices:

1. User double-taps and **holds** on the second tap (finger stays down)
2. While holding, dragging **up** zooms in
3. While holding, dragging **down** zooms out
4. Releasing the finger ends the zoom gesture

The zoom should be centered on the initial tap point, similar to how pinch-to-zoom centers between the two fingers.

## Use cases

- **Single-handed phone use**: Users holding their phone in one hand can zoom without needing a second hand
- **Accessibility**: Easier for users who have difficulty with two-finger gestures
- **Power users**: Faster zooming for users familiar with this gesture from maps apps

## Alternatives considered

- **Volume buttons for zoom**: Less intuitive, not standard
- **Zoom slider UI**: Takes up screen space, less direct
- **Keep pinch-only**: Works but less accessible for single-handed use

## Implementation notes

The gesture detection would likely need to:
- Track double-tap timing (similar to existing click handling)
- Detect when the second tap is held vs released
- On hold + drag, transition to a "zooming" state
- Map vertical drag distance to zoom level changes

max-drake:
Closes #7677

In order to give touch users an accessible way to zoom without requiring a two-finger pinch gesture, this PR adds a double-tap-and-drag zoom interaction to the HandTool.

On coarse-pointer (touch) devices, double-tapping the canvas and dragging vertically zooms in/out — dragging down zooms in, dragging up zooms out, matching the Google Maps convention. The zoom anchors to the initial touch point so content stays stable under the finger. On pointer-up, any remaining velocity triggers a momentum-based zoom slide.

Original branch: `max/one-finger-zoom`

### Change type

- [x] `feature`

### Test plan

1. Open tldraw on a touch device (or enable touch simulation in devtools)
2. Double-tap the canvas and hold on the second tap
3. Drag down to zoom in, drag up to zoom out
4. Verify zoom anchors to the tap point
5. Release quickly and verify momentum zoom continues briefly

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add double-tap-and-drag zoom gesture for touch devices

---

> [!NOTE]
> **Medium Risk**
> Touches core camera animation math and hand-tool input handling, which can affect pan/zoom behavior across devices. Risk is moderate because changes are localized but interact with high-frequency tick/input code paths.
> 
> **Overview**
> Adds a **double-tap-and-drag one-finger zoom gesture** for coarse pointers: `PointingCanvas` transitions into a new `OneFingerZooming` state on double-click down, where vertical drag adjusts camera zoom while keeping the initial touch point anchored, then returns to idle.
> 
> Extends `editor.slideCamera` to support **z-direction zoom animation** (in addition to x/y panning), adjusting camera x/y to keep the viewport center stable during zoom; `OneFingerZooming` uses this to apply momentum-based zoom on pointer-up.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0d5c39b9f752904c4347de0d3df472e1c4226e92. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


max-drake:
converting to draft pending https://github.com/tldraw/tldraw/pull/8897

packages/tldraw/src/lib/tools/HandTool/childStates/OneFingerZooming.ts
```diff
@@ -0,0 +1,79 @@
...
+		this.editor.menus.clearOpenMenus()
+
+		const currentScreenY = this.editor.inputs.getCurrentScreenPoint().y
+
+		// Dragging up = zoom out, dragging down = zoom in (same as google maps default)
+		// we won't respect the user's zoom direction preference because it only applies
+		// to mouse input
+		const dy = (this.originScreenY - currentScreenY) * -1
+
+		// ~200px of drag ≈ 2x zoom change.
```
frolic, line 30: where'd this number come from? should this be % of viewport so we can factor in different device sizes and densities?

packages/tldraw/src/lib/tools/HandTool/HandTool.ts
```diff
@@ -9,15 +10,27 @@ export class HandTool extends StateNode {
...
-			const currentScreenPoint = this.editor.inputs.getCurrentScreenPoint()
-			this.editor.zoomIn(currentScreenPoint, {
-				animation: { duration: 220, easing: EASINGS.easeOutQuint },
-			})
+		switch (info.phase) {
+			case 'settle-down': {
+				// A double-tap whose second press is still held down: begin one-finger
+				// drag-to-zoom. This is a touch gesture, so only enter it on a coarse pointer.
+				if (this.editor.getInstanceState().isCoarsePointer) {
+					this.transition('one_finger_zooming', info)
```
frolic, line 22: does this need any sort of minimum movement threshold to activate so a subtle movement during double tap doesn't prevent the "double tap to zoom out" from working? or does the OS itself handle this already?

max-drake: there might be some os stuff working here as well for touch inputs, but we also handle this ourselves in `ClickManager.ts` (the `MAX_CLICK_DISTANCE` const), agnostic of if it's a coarse pointer, pen, or normal pointer

packages/tldraw/src/lib/tools/HandTool/childStates/OneFingerZooming.ts
max-drake: this was kind of just trial and error? I think I'd want to keep it tied to a fixed distance because if someone had a really big screen, they'd have to drag a massive distance to get a reasonable zoom, ditto with a super small screen. some quick searching tells me that the end result of like "pixels per inch of drag" doesn't vary a _massive_ amount between devices, and I don't think we have any way of accessing device pixel density save like, a lookup table, which is doable but imo not worth it.

mimecuvalo:
updated title + desc to say this is HandTool — got confused at first on why the Select tool wasn't doing what i thought it should be doing 🙃

packages/tldraw/src/lib/tools/HandTool/childStates/OneFingerZooming.ts
```diff
@@ -0,0 +1,79 @@
...
+
+	private complete() {
+		const pointerVelocity = this.editor.inputs.getPointerVelocity()
+		const velocityAtPointerUp = Math.min(pointerVelocity.len(), 2)
+
+		if (velocityAtPointerUp > 0.1) {
+			// Use velocity y-sign for momentum direction (positive y = moving down = zoom in)
+			this.editor.slideCamera({
+				speed: velocityAtPointerUp,
+				direction: { x: 0, y: 0, z: Math.sign(pointerVelocity.y) * 0.01 },
```
mimecuvalo, line 66: TIL about `Math.sign`! how useful

mimecuvalo (approved):
nice one 👍

packages/tldraw/src/lib/tools/HandTool/childStates/OneFingerZooming.ts
frolic: same! very cool

PR #7878, opened 2026-02-10
fix: geo shapes with text could not be resized to a smaller size

mimecuvalo:
this regressed in https://github.com/tldraw/tldraw/pull/7412 and we didn't notice til now, oops.

before:

https://github.com/user-attachments/assets/1d02cfdb-5f6b-4b1c-bdc6-f4c7122863ba

after:

https://github.com/user-attachments/assets/3c34687c-3ff9-4839-87c0-574b2873581a

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- fix: geo shapes with text could not be resized to a smaller size

---

> [!NOTE]
> **Medium Risk**
> Touches `GeoShapeUtil.onResize` sizing constraints and bypasses the label-size cache during drag-resize, which could affect resize behavior/perf for all labeled geo shapes; changes are covered by new unit tests.
> 
> **Overview**
> Fixes geo-shape resizing when a label is present by **re-measuring label size at the in-progress resize dimensions** (via `measureUnscaledLabelSize` on a temp shape) so constraints reflect text wrapping, rather than being locked to the initial cached measurement.
> 
> Adds unit tests covering resizing labeled geo shapes narrower than the initial label width, enforcing minimum width/height based on wrapped label size, and ensuring unlabeled shapes still resize freely; also disposes the `TestEditor` after each test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e326b65e4b31b4e706b749f608d79162ee77c6ab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.test.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.test.tsx
@@ -1,4 +1,4 @@
-import { Group2d, IndexKey, TLShapeId } from '@tldraw/editor'
+import { Group2d, IndexKey, TLGeoShape, TLShapeId, createShapeId, toRichText } from '@tldraw/editor'
 import { TestEditor } from '../../../test/TestEditor'
 import { TL } from '../../../test/test-jsx'
 
@@ -7,6 +7,10 @@ let ids: Record<string, TLShapeId>
 
 beforeEach(() => {
 	editor = new TestEditor()
+})
+
+afterEach(() => {
+	editor?.dispose()
 })
 
 describe('Handle snapping', () => {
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.test.tsx
```diff
@@ -9,6 +9,10 @@ beforeEach(() => {
 	editor = new TestEditor()
 })
 
+afterEach(() => {
+	editor?.dispose()
```
kaneel, line 13: since we recreate the global variable before every test, can you walk me through why this is needed? Do we have some values that are saved in some memory and restored by `new TestEditor`?

kaneel (commented):
I have some questions for my own personal knowledge. 

Also, I get that we cannot test directly that we get the value we we want because of the nature of the component and this is why you used `toBeLessThan` or `toBeGreaterThan`, I assume these tests won't pass at all with the previous code?

PR #7886, opened 2026-02-11
revert(tldraw): revert SVG sanitization

### LINKED ISSUE #7876: Sanitize SVG content in the SDK to prevent XSS
## Problem

SVG files are executable documents, not inert images. When users paste or import SVGs into tldraw, there's no sanitization step — meaning malicious SVGs with embedded scripts, cross-origin references, or phishing links can make it through to rendering.

### Attack vectors in the current codebase

1. **SVG text pasting** — User-pasted SVG text is parsed via `DOMParser` and stored as an asset with no sanitization (`defaultExternalContentHandlers.ts:314`)
2. **SVG file drops** — SVG files dropped onto the canvas are handled as image assets without content inspection
3. **Bookmark favicons** — Favicon URLs extracted from fetched pages could point to malicious SVGs (`BookmarkShapeUtil.tsx:216`)
4. **SVG export round-trips** — Exported SVGs re-imported could carry injected content

While `` tags don't execute scripts, SVGs opened as top-level documents or injected via other means can.

## Proposed solution

Add SVG sanitization as a defense-in-depth measure in the SDK. One strong candidate is [svg-hush](https://github.com/cloudflare/svg-hush) from Cloudflare:

- Allowlist-based approach — removes elements/attributes not explicitly permitted
- Strips scripting, cross-domain links, and cross-origin resource references
- Written in Rust (could compile to WASM for browser use)
- MIT licensed, actively maintained

### Where to sanitize

- In `defaultExternalContentHandlers.ts` before SVG text is stored as an asset
- In the asset creation pipeline for SVG files
- Optionally as a configurable `sanitizeSvg` hook so SDK users can customize

### Alternatives

- **DOMPurify** — widely used HTML/SVG sanitizer, pure JS, larger scope than needed
- **Manual strip** — remove `<script>`, event attributes, etc. manually (error-prone, incomplete)
- **CSP headers only** — helps but is a dotcom-level mitigation, not an SDK-level one

## Considerations

- Bundle size impact of WASM module vs. JS-only solution
- Whether sanitization should be opt-in or on-by-default
- Performance impact on large/complex SVGs
- Need to preserve valid SVG features (filters, gradients, masks) while stripping dangerous ones

MitjaBezensek:
Reverts #7880 (SVG sanitization). Our SVG export relies on `<foreignObject>` for text rendering (used by `RichTextSVG` in text, geo, note, arrow shapes + the fallback HTML renderer in `getSvgJsx`). The sanitizer correctly strips `foreignObject` (DOMPurify does the same — it's in their `svgDisallowed` list), but this breaks copy/paste of our own SVGs since all text content is lost.

Keeps the docs recommendation to host assets on a separate domain — that's the primary defense against same-origin SVG attacks.

Closes #7876

### Change type

- [x] `improvement`

### Test plan

1. Copy as SVG, paste into tldraw — text should render correctly with fonts
2. Drop SVG files onto canvas — should work as before

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Revert SVG sanitization that broke copy/paste of SVGs containing text

### DIFF
--- a/apps/docs/content/sdk-features/assets.mdx
+++ b/apps/docs/content/sdk-features/assets.mdx
@@ -298,6 +298,12 @@ Asset records use the migration system to evolve their schema. Each asset type h
 
 Validators ensure asset data matches the expected structure at runtime. The [assetValidator](?) uses a discriminated union on the `type` field. If you're creating custom asset types, create validators following the same pattern and add migration sequences to handle schema changes.
 
+## Security
+
+### Hosting assets on a separate domain
+
+We recommend serving user-uploaded assets from a completely separate domain (e.g. `example-assets.com` rather than a subdomain like `assets.example.com`). This provides an extra layer of protection: if a malicious file somehow bypasses sanitization, browser same-origin policies prevent it from accessing cookies, storage, or APIs on your main domain.
+
 ## Related examples
 
 - [Hosted images](https://github.com/tldraw/tldraw/tree/main/apps/examples/src/examples/hosted-images) - Implement a TLAssetStore that uploads images to a server


steveruizok:
Thanks for catching this before we landed it.

PR #7914, opened 2026-02-13
perf(ci): add shared TypeScript incremental build cache across CI workflows

kaneel:
It's just me assessing things using Claude, it's an experiment at iterating with Claude figuring what can be improved and moved around. All in all, there's a way to separate cache to make sure we don't reuse lazy cache but stil reuse TS cache in some places where it may matter.

## What's actually new vs what was there before

**Before this PR:** only checks.yml had any build caching (both .lazy and .tsbuildinfo, inline). Every other workflow built from scratch every time.
**After this PR:**
8 additional workflows get TypeScript incremental caching via the shared action. This is the main improvement.
checks.yml build job keeps its .lazy cache (now in a separate step with its own key namespace so it doesn't leak to other workflows).
actions/cache@v3 bumped to @v4 everywhere.

## What the PR actually does 

There are two independent caching mechanisms at play: 

### 1. TypeScript incremental build cache (.tsbuild/.tsbuildinfo) 

These files are TypeScript's own incremental compilation state. When tsc runs with --incremental, it reads .tsbuildinfo to figure out which files changed since last compilation, and only recompiles those. The .tsbuild directories contain the compiled .d.ts outputs. 

**Where this helps:** any workflow that runs yarn build-types (which calls tsc under the hood). That's checks.yml test job, all the deploy workflows, and publish-editor-extensions. The Playwright workflows also benefit because preview-app builds the app which includes type compilation. 

**How much time it saves:** on a typical PR that changes a few files, instead of compiling the entire monorepo from scratch (~60-90s), tsc only recompiles what changed (~10-20s). 

### 2. Lazyrepo manifest cache (.lazy) 

These files track which lazyrepo tasks ran successfully for which input hashes.
They let lazyrepo skip tasks entirely when inputs haven't changed. 

**Where this helps:** only the checks.yml build job, and only for tasks whose inputs didn't change between commits. For example, if a PR only touches packages/editor/, lazyrepo can skip refresh-assets, build-eslint-plugin, build-i18n, and building other unchanged packages. 

**How much time it saves:** those dependency tasks are fast (a few seconds each). Skipping unchanged package builds is more meaningful but depends on how many packages a PR touches. Ballpark: 5-30s in the best case. 

**Why it can't be shared:** the .lazy manifests say "this task completed" but the actual build outputs (dist/ dirs) aren't cached. If another workflow restores those manifests, lazyrepo skips the build and the outputs are missing. This is what broke Playwright. 

## Where each workflow stands

Workflow | What it builds | Benefits from .tsbuildinfo? | Benefits from .lazy?
-- | -- | -- | --
checks.yml test | build-types, tests | Yes | Has its own inline cache (unchanged)
checks.yml build | build-package (all packages) | Yes | Yes (separate lazyrepo-* key, safe)
deploy-dotcom | build-types + deploy script | Yes | No need
deploy-bemo | build-types + deploy script | Yes | No need
deploy-analytics | build-types + deploy script | Yes | No need
publish-editor-extensions | build-types + vsce publish | Yes | No need
publish-canary | Full publish (builds everything) | Yes | No (would poison)
playwright-dotcom | Full app build + e2e | Yes | No (would poison)
playwright-examples | Full app build + e2e | Yes | No (would poison)

---

> [!NOTE]
> **Low Risk**
> CI-only caching changes; main risk is cache key/path mistakes causing slower builds or stale/incorrect incremental artifacts, not production behavior changes.
> 
> **Overview**
> Introduces a reusable composite action `./.github/actions/restore-build-caches` that restores TypeScript incremental build artifacts (`**/.tsbuild`, `**/*.tsbuildinfo`) via `actions/cache@v5`.
> 
> Updates several workflows to use this shared cache restore step (deploy, publish, and Playwright), and restructures `checks.yml` so the build job restores TS caches via the shared action while restoring `.lazy` via a **separate** `lazyrepo-*` cache key; the test job’s cache step is renamed to “Restore test caches” but keeps its broader cache set.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df1f589ec377eb93ca0722ecbdb12545486c7c08. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Check circular dependencies
         run: yarn check-circular-deps
 
-      - name: Restore build tool caches
+      - name: Restore test caches
         uses: actions/cache@v5
         with:
           path: |
@@ -102,7 +102,7 @@ jobs:
       - uses: ./.github/actions/restore-build-caches
 
       - name: Restore lazyrepo cache
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             .lazy


mimecuvalo:
be sure to add the Release Label @kaneel ! that's this bit:
Screenshot 2026-02-13 at 13 04 30

kaneel:
I don't think it does much this one

PR #7916, opened 2026-02-13
[wip] Custom styles

MitjaBezensek:
> Updated API in #8053## Summary

This is not feature complete, more a PoC.

Introduces a layered style system for shapes. Instead of shapes hardcoding pixel values from style tokens, styles are now resolved through a pipeline: **global tokens → per-shape defaults → runtime overrides**.

All customization options are fully type-safe — new color/size/font values use module augmentation to extend the type system, and `useShapeStyles()` returns correctly typed objects via `TLShapeStylesMap`.

### Customization options

**1. Per-shape-type style configuration** (`ShapeUtil.configure({ styles })`)

Use when one shape type needs different values than others. E.g. geo shapes should have thicker strokes than arrows at the same "medium" size. Only affects that shape type, others keep global defaults. Supports colors, sizes, and fonts.

```tsx
const shapeUtils = [
  GeoShapeUtil.configure({
    styles: {
      sizes: {
        s: { stroke: 1, labelFont: 14 },
        m: { stroke: 3, labelFont: 18 },
        l: { stroke: 6, labelFont: 24 },
        xl: { stroke: 12, labelFont: 32 },
      },
    },
  }),
]

<Tldraw shapeUtils={shapeUtils} />
```

**2. Override existing styles** (`styles` prop on `<Tldraw>`)

Use when you want to rebrand the whole canvas. E.g. your app's "red" is `#dc0000` not tldraw's default, or "medium" should mean 5px stroke instead of 3.5px. Affects all shapes globally.

Possibly name this `shapeStyles` to go along with existing `shapeUtils` naming. Another option would be to use a similar config as above, just on the base `ShapeUtil` class, something like `ShapeUtil.configure....`.

```tsx
const styles: TLStylesConfig = {
  colors: {
    red: {
      light: { solid: '#e03131', semi: '#ffe3e3' },
      dark: { solid: '#ff6b6b', semi: '#4a2020' },
    },
  },
  sizes: {
    m: { stroke: 5, font: 30, labelFont: 28, arrowLabelFont: 24 },
  },
  fonts: {
    sans: "'Inter', sans-serif",
  },
}

<Tldraw styles={styles} />
```

**3. Add or remove style values** (`styles` prop + module augmentation)

Use when the built-in palette isn't enough — or has too much. New values defined in the `styles` prop are automatically added to the set of allowed values for that style prop (e.g. `DefaultColorStyle`), so the store accepts them. Requires `declare module` for type safety.

To **add** a new color like "coral":

```tsx
declare module '@tldraw/tlschema' {
  interface TLColorStyleExtensions {
    coral: true
  }
}

const styles: TLStylesConfig = {
  colors: {
    coral: {
      light: { solid: '#ff6b6b', fill: '#ff6b6b', semi: '#ffe0e0', pattern: '#ff8787' },
      dark: { solid: '#ff8787', fill: '#ff8787', semi: '#4a2020', pattern: '#ff6b6b' },
    },
  },
}

<Tldraw styles={styles} />
```

To **remove** an existing value, set it to `null`. This removes it from the allowed set:

```tsx
const styles: TLStylesConfig = {
  colors: {
    'light-violet': null, // remove light-violet from the color palette
    'light-red': null,
  },
  sizes: {
    xl: null, // remove XL from the size options
  },
}

<Tldraw styles={styles} />
```

**4. Runtime style overrides** (`getShapeStyleOverrides` prop on `<TldrawEditor>`)

Use when styles depend on dynamic state. E.g. grey out locked shapes, highlight selected shapes, dim shapes based on a filter. Runs per-shape at render time, supports `{ light, dark }` values for theme-aware overrides.

```tsx
const getShapeStyleOverrides: TLShapeStyleOverrides = (shape) => {
  if (shape.isLocked) {
    return {
      strokeColor: { light: '#9ca3af', dark: '#6b7280' },
      strokeWidth: 1,
    }
  }
}

<Tldraw getShapeStyleOverrides={getShapeStyleOverrides} />
```

**5. Arbitrary colors** (e.g. color picker)

Use when you need freeform hex colors instead of predefined tokens. For built-in shapes, store the value on `meta` and resolve it via `getShapeStyleOverrides`:

```tsx
// Store arbitrary color
editor.updateShape({ id: shape.id, meta: { customColor: '#ff6b2a' } })

// Resolve at render time
const getShapeStyleOverrides: TLShapeStyleOverrides = (shape) => {
  if (shape.meta.customColor) {
    return { strokeColor: shape.meta.customColor as string }
  }
}

<Tldraw getShapeStyleOverrides={getShapeStyleOverrides} />
```

For custom shapes, you can use a freeform `color: string` prop directly — `getColorValue()` passes through non-token strings as-is:

```tsx
class CustomShapeUtil extends ShapeUtil<MyShape> {
  override getDefaultStyles(shape: MyShape, ctx: TLStyleContext) {
    return {
      strokeColor: shape.props.color, // any hex value works
    }
  }
}
```

**6. Custom shape style resolution** (`ShapeUtil.getDefaultStyles()`)

Use when building a custom shape that needs to participate in the style system. Maps abstract tokens (color: "blue", size: "m") to concrete CSS values your component consumes via `useShapeStyles()`. Automatically picks up custom tokens and theme changes.

```tsx
class CardShapeUtil extends ShapeUtil<ICardShape> {
  override getDefaultStyles(shape: ICardShape, ctx: TLStyleContext) {
    return {
      borderColor: getColorValue(ctx.theme, shape.props.color, 'solid'),
      borderWidth: ctx.sizes[shape.props.size].stroke,
    }
  }

  component(shape: ICardShape) {
    const styles = useShapeStyles(shape)
    return <div style={{ borderColor: styles.borderColor }}>...</div>
  }
}
```

### Priority (lowest → highest)

When multiple options are used together, later layers override earlier ones:

1. **Built-in defaults** — tldraw's default token values
2. **Global `styles` prop** — overrides defaults for all shapes
3. **Per-shape-type `styles`** via `configure()` — overrides global config for that shape type only
4. **`getShapeStyleOverrides`** — final override, runs last, wins over everything

Example: if the global config sets `m` stroke to 5px, but `GeoShapeUtil.configure()` sets it to 3px, geo shapes get 3px while other shapes get 5px. If `getShapeStyleOverrides` then returns `strokeWidth: 1` for locked shapes, that wins regardless.

### Supported style properties

The `styles` prop supports customizing these token types:

| Property | What it controls | Token keys |
|----------|-----------------|------------|
| `colors` | Shape stroke/fill colors with light/dark variants (`solid`, `fill`, `semi`, `pattern`) | `black`, `grey`, `light-violet`, `violet`, `blue`, `light-blue`, `yellow`, `orange`, `green`, `light-green`, `light-red`, `red`, `white` |
| `sizes` | Stroke widths and font sizes (`stroke`, `font`, `labelFont`, `arrowLabelFont`) | `s`, `m`, `l`, `xl` |
| `fonts` | CSS font-family strings | `draw`, `sans`, `serif`, `mono` |

Not covered by this system (these are enum values, not design tokens):
- **dash** (`draw`, `dashed`, `dotted`, `solid`) — stroke pattern, not a numeric/color value
- **fill** (`none`, `semi`, `solid`, `pattern`) — fill mode selector
- **align/verticalAlign** — text layout enums
- **geo** (`rectangle`, `ellipse`, etc.) — shape type selector
- **opacity** — a direct numeric value, not a token

### Examples

- **Per-shape-type styles** — `GeoShapeUtil.configure({ styles })` to override size tokens for geo shapes only (option 1)
- **Custom styles** — overriding existing size/font values and adding new colors via the `styles` prop (options 2 and 3)
- **Shape style overrides** — `getShapeStyleOverrides` to conditionally mute locked shapes at runtime (option 4)
- **Custom shape styles** — building a custom shape with `getDefaultStyles()` and `useShapeStyles()` (option 6)

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -3253,6 +3253,14 @@ export interface TLArrowPoint {
 // @public (undocumented)
 export interface TLArrowShapeResolvedStyles {
     // (undocumented)
+    fillColor: string;
+    // (undocumented)
+    fillLinedFillColor: string;
+    // (undocumented)
+    fillPatternColor: string;
+    // (undocumented)
+    fillSolidColor: string;
+    // (undocumented)
     fontFamily: string;
     // (undocumented)
     labelColor: string;
@@ -3405,6 +3413,14 @@ export function TldrawShapeIndicators(): JSX.Element;
 
 // @public (undocumented)
 export interface TLDrawShapeResolvedStyles {
+    // (undocumented)
+    fillColor: string;
+    // (undocumented)
+    fillLinedFillColor: string;
+    // (undocumented)
+    fillPatternColor: string;
+    // (undocumented)
+    fillSolidColor: string;
     // (undocumented)
     strokeColor: string;
     // (undocumented)
@@ -3720,6 +3736,14 @@ export interface TLFrameShapeResolvedStyles {
 
 // @public (undocumented)
 export interface TLGeoShapeResolvedStyles {
+    // (undocumented)
+    fillColor: string;
+    // (undocumented)
+    fillLinedFillColor: string;
+    // (undocumented)
+    fillPatternColor: string;
+    // (undocumented)
+    fillSolidColor: string;
     // (undocumented)
     fontFamily: string;
     // (undocumented)
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -199,6 +199,10 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 			labelColor: getColorValue(ctx.theme, shape.props.labelColor, 'solid'),
 			labelFontSize: ctx.sizes[shape.props.size].arrowLabelFont,
 			fontFamily: ctx.fonts[shape.props.font],
+			fillSolidColor: getColorValue(ctx.theme, shape.props.color, 'semi'),
+			fillColor: getColorValue(ctx.theme, shape.props.color, 'fill'),
+			fillPatternColor: getColorValue(ctx.theme, shape.props.color, 'pattern'),
+			fillLinedFillColor: getColorValue(ctx.theme, shape.props.color, 'linedFill'),
 		}
 	}
 
@@ -1141,6 +1145,10 @@ export interface TLArrowShapeResolvedStyles {
 	labelColor: string
 	labelFontSize: number
 	fontFamily: string
+	fillSolidColor: string
+	fillColor: string
+	fillPatternColor: string
+	fillLinedFillColor: string
 }
 
 declare module '@tldraw/editor' {
@@ -1261,7 +1269,7 @@ const ArrowSvg = track(function ArrowSvg({
 					<ShapeFill
 						theme={theme}
 						d={as}
-						color={shape.props.color}
+						fillColors={styles}
 						fill={shape.props.fill}
 						scale={shape.props.scale}
 					/>
@@ -1270,7 +1278,7 @@ const ArrowSvg = track(function ArrowSvg({
 					<ShapeFill
 						theme={theme}
 						d={ae}
-						color={shape.props.color}
+						fillColors={styles}
 						fill={shape.props.fill}
 						scale={shape.props.scale}
 					/>
--- a/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
@@ -89,6 +89,10 @@ export class DrawShapeUtil extends ShapeUtil<TLDrawShape> {
 		return {
 			strokeWidth: ctx.sizes[shape.props.size].stroke + 1,
 			strokeColor: getColorValue(ctx.theme, shape.props.color, 'solid'),
+			fillSolidColor: getColorValue(ctx.theme, shape.props.color, 'semi'),
+			fillColor: getColorValue(ctx.theme, shape.props.color, 'fill'),
+			fillPatternColor: getColorValue(ctx.theme, shape.props.color, 'pattern'),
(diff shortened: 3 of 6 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -820,6 +857,93 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		const util = this.getShapeUtil(shape)
+		const ctx = this.getStyleContext()
+		let styles: object = util.getDefaultStyles(shape, ctx) ?? {}
+
+		if (this._getShapeStyleOverrides) {
+			const overrides = this._getShapeStyleOverrides(shape, ctx)
+			if (overrides) {
+				const resolvedOverrides: Record<string, unknown> = {}
+				for (const [key, value] of Object.entries(overrides)) {
+					if (value && typeof value === 'object' && 'light' in value && 'dark' in value) {
```
mimecuvalo, line 932: i would be careful of this anchoring this in the color heuristics. 
one could imagine having to, say, having to have different viewport breakpoint options (sm, md, lg, etc.) and we should have a good answer for that

apps/examples/src/examples/custom-styles/CustomStylesExample.tsx
```diff
@@ -0,0 +1,190 @@
...
+		forest: true
+	}
+}
+
+// [2]
+const styles: TLStylesConfig = {
+	colors: {
+		coral: {
+			light: {
+				solid: '#ff6b6b',
```
mimecuvalo, line 18: i think my biggest pushback here would be that if we don't have functions to derive a value, that may be fine, but then i would highlight that people should lean into CSS variables, like:
```
solid: 'var(--color-primary)',
```

I would be super concerned if hard coded CSS values (hex colors, pixel values, etc.) start getting propagated into the shape props/meta data and then they start to bitrot as color defaults change over time

MitjaBezensek: But these would not get propagated. This is a static definition. If you need to change it you just change this hex here and everything updates. Shape would still store `coral`, which I think is one of the strengths of this approach.

PR #7932, opened 2026-02-14
fix(arrows): clamp arrow terminals to frame boundary when bound shape is clipped

### LINKED ISSUE #7824: Arrow endpoint shows at shape boundary even when shape is clipped by frame
## Description

When an arrow connects to a shape that is inside a frame, and that shape extends beyond the frame boundary, the arrow endpoint terminates at the shape's boundary rather than at the frame's boundary where the shape is visually clipped.

## Steps to reproduce

1. Create a frame
2. Add a shape (e.g., a note or rectangle) inside the frame
3. Position the shape so it extends beyond the frame boundary (it will be clipped)
4. Draw an arrow from outside the frame, connecting to the shape

## Expected behavior

The arrow should terminate at the frame boundary, since that's where the shape visually ends due to clipping.

## Actual behavior

The arrow terminates at the shape's actual boundary, which is inside the frame but beyond where the shape is visible. This creates a visual disconnect where the arrow appears to point into empty space within the frame.

## Screenshot

Image

## Additional context

This affects the visual coherence of diagrams where frames are used to group and clip content. The current behavior technically follows the shape's geometry, but the user expectation is that arrows should respect the visual appearance of the canvas.

steveruizok:
Fixes #7824. When a shape inside a frame extends beyond the frame boundary, arrows bound to that shape would terminate at the shape's actual boundary rather than at the frame's clip boundary, creating a visual disconnect where the arrow appears to point into empty space.

https://github.com/user-attachments/assets/fcce61fb-20ea-4ef2-abd0-46f1a370ee83

This PR adds a `clampArrowTerminalToMask()` post-processing step that runs after arrow-shape geometry intersection. If the computed terminal point falls outside the bound shape's mask (the frame clip boundary), it casts a ray from the binding anchor through the terminal to find where it crosses the mask polygon, and clamps the terminal to that point.

### Change type

- [x] `bugfix`

### Test plan

1. Create a frame
2. Add a shape (e.g. geo or note) inside the frame
3. Position the shape so it extends beyond the frame boundary (it will be clipped)
4. Draw a straight arrow from outside the frame, connecting to the shape
5. Verify the arrow terminates at the frame boundary, not at the shape's invisible clipped edge
6. Repeat with a curved arrow

- [x] Unit tests

### Release notes

- Fix arrow endpoints terminating at invisible clipped shape boundaries instead of at the frame edge


kaneel:
Not sure how this should behave so feel free to brush me off 

https://github.com/user-attachments/assets/0ebe428f-5be9-4a02-a365-3720cf2c7ed4

steveruizok:
> Not sure how this should behave so feel free to brush me off
> 
>  Screen.Recording.2026-02-24.at.14.29.59.mov

I think we'll have to live with this one: the only alternative I can think of is to extend the arrow past the anchor point, which is usually a one-way ticket to arrow-bug city.

PR #7933, opened 2026-02-14
fix(utils): prevent false positive multiple instances warning in dev mode

### LINKED ISSUE #4614: Multiple instances error after upgrading to v3
### What happened?

I'm using tldraw with Next.js, and I got the "multiple instances" error after I updated it to v3.1.0 (I also tried to do a clean install of modules). The error looks like

```
[tldraw] You have multiple instances of some tldraw libraries active. This can lead to bugs and unexpected behavior. 

This usually means that your bundler is misconfigured, and is importing the same library multiple times - usually once as an ES Module, and once as a CommonJS module.

The following libraries have been imported multiple times:
  • ❌ @tldraw/utils v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/state v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/store v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/validate v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/tlschema v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/sync-core v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/state-react v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/editor v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ tldraw v3.1.0: 
      1. ES Modules
      2. ES Modules
  • ❌ @tldraw/sync v3.1.0: 
      1. ES Modules
      2. ES Modules
```

My `package.json` looks like this (which is used by both the server and front-end, so both sync and sync-core are installed in the same place, removing them didn't remove the error message)

```
"dependencies": {
    "@tldraw/sync": "^3.1.0",
    "@tldraw/sync-core": "^3.1.0",
    "tldraw": "^3.1.0"
  }
```

I can confirm that this is the only place where I install the packages. I didn't modify the config file for Next.js.

I wonder if there are other possibilities for the cause of the issue. Thanks.

Please see above.

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

steveruizok:
In order to fix a false positive "multiple instances" warning that fires in Next.js dev mode, this PR deduplicates registrations in `registerTldrawLibraryVersion()`.

When Next.js re-evaluates modules on page refresh, each tldraw package re-registers itself via `registerTldrawLibraryVersion()`. Since `globalThis.__TLDRAW_LIBRARY_VERSIONS__` persists across module re-evaluations, duplicate entries accumulate and trigger the "multiple instances" warning — even though only one copy of each library is loaded (all entries show the same version and same module type).

The fix skips registration when an identical entry (same name, version, and module type) already exists. Genuine conflicts (different versions or mixed ESM/CJS) are still detected correctly.

Closes #4614

### Change type

- [x] `bugfix`

### Test plan

1. Use tldraw in a Next.js app with dev mode
2. Refresh the page — the false "multiple instances" warning should no longer appear
3. Genuine version mismatches or ESM/CJS conflicts should still produce warnings

- [x] Unit tests

### Release notes

- Fixed false positive "multiple instances" warning in Next.js dev mode caused by module re-evaluation

---

> [!NOTE]
> **Low Risk**
> Small, well-scoped change to warning/dedup logic gated to Next.js dev mode with unit test coverage; main risk is inadvertently suppressing a real duplicate only under those conditions.
> 
> **Overview**
> Prevents false "multiple instances" warnings in Next.js *dev mode* by skipping re-registration when `registerTldrawLibraryVersion` sees an **exact duplicate** entry (same package, version, and module type) during Fast Refresh.
> 
> Adds `isNextjsDev()` gating (checks `NODE_ENV === 'development'` and presence of `__NEXT_DATA__`) so duplicate tracking and warnings still occur in non-Next.js environments, and updates tests to cover both the Next.js dev dedupe behavior and the unchanged non-dev duplicate warning.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cc920c01d365539317e63c0fb542e6fc8d2d91fb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/utils/src/lib/version.ts
+++ b/packages/utils/src/lib/version.ts
@@ -93,10 +93,17 @@ export function registerTldrawLibraryVersion(name?: string, version?: string, mo
 	}
 
 	const info = getLibraryVersions()
-	const isDuplicate = info.versions.some(
-		(v) => v.name === name && v.version === version && v.modules === modules
-	)
-	if (isDuplicate) return
+
+	// In Next.js dev mode, Fast Refresh re-executes module-level code which causes
+	// the same library to register multiple times, producing a false positive warning.
+	// We skip exact duplicates only in Next.js dev to avoid hiding genuine issues.
+	if (isNextjsDev()) {
+		const isDuplicate = info.versions.some(
+			(v) => v.name === name && v.version === version && v.modules === modules
+		)
+		if (isDuplicate) return
+	}
+
 	info.versions.push({ name, version, modules })
 
 	if (!info.scheduledNotice) {
@@ -221,6 +228,14 @@ const formats = {
 } as const
 function format(value: string, formatters: (keyof typeof formats)[] = []) {
 	return `\x1B[${formatters.map((f) => formats[f]).join(';')}m${value}\x1B[m`
+}
+
+function isNextjsDev(): boolean {
+	try {
+		return process.env.NODE_ENV === 'development' && '__NEXT_DATA__' in globalThis
+	} catch {
+		return false
+	}
 }
 
 function entry<K, V>(map: Map<K, V>, key: K, defaultValue: V): V {
--- a/packages/utils/src/lib/version.test.ts
+++ b/packages/utils/src/lib/version.test.ts
@@ -294,10 +294,13 @@ describe('version utilities', () => {
 			expect(mockConsoleLog).not.toHaveBeenCalled()
 		})
 
-		it('should ignore duplicate registrations of the same library', () => {
-			// Re-registering the same library with the same version and module type
-			// (e.g. due to module re-evaluation in Next.js dev mode) should not
-			// trigger a warning
+		it('should ignore duplicate registrations in Next.js dev mode', () => {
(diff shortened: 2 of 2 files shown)


packages/utils/src/lib/version.ts
```diff
@@ -93,6 +93,10 @@ export function registerTldrawLibraryVersion(name?: string, version?: string, mo
 	}
 
 	const info = getLibraryVersions()
+	const isDuplicate = info.versions.some(
+		(v) => v.name === name && v.version === version && v.modules === modules
```
ds300: can we make this also check that it's a nextjs dev env? otherwise i reckon this might produce some false negatives (bad)

PR #7936, opened 2026-02-14
fix(select): allow dragging shapes that have an onClick handler

### LINKED ISSUE #5613: onClick handler on shape prevents PointingShape logic
### What happened?

As noted on Discord:

> [This lovely line of tldraw logic](https://github.com/tldraw/tldraw/blob/7c0dd224d7105102dd2b1d01c6e50c5b02d1b55d/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts#L34) makes it so one can't have an onClick handler on a Shape without interfering with pointerdown+drag to move the shape when it is not selected.  Annoying behavior so probably an unintended side-effect? Fix/workaround?

Without having dug into the reasons here, my cursory reaction would be that we should be calling the shape maybe and seeing if it returns true/false? Not sure.

1. Create a custom shape with an onClick handler. 
2. Make sure the shape is not selected
3. Attempt the drag the shape

### Expected

* Shape is translated

### Actual

* onClick is triggered instead

*No response*

### Where did this happen?

The developer package

### Contact Details

*No response*

### Code of Conduct

steveruizok:
Closes #5613

When a custom shape has an `onClick` handler and is not yet selected, `PointingShape.onEnter` defers selection to pointer up (so that `onClick` can fire). However, if the user drags instead of clicking, the `Translating` state requires a selection to operate and immediately bails back to idle—making it impossible to drag unselected shapes that have an `onClick` handler.

This PR fixes the issue by selecting the shape in `startTranslating` when the shape wasn't selected on enter and there is no existing selection. This ensures that dragging works correctly while preserving the existing behavior for shapes that are already selected or part of a multi-selection.

https://github.com/user-attachments/assets/ff74b3c3-1c57-4810-8b9f-cf39213940bf

### Change type

- [x] `bugfix`

### Test plan

1. Create a custom shape with an `onClick` handler
2. Make sure the shape is not selected
3. Attempt to drag the shape — it should translate as expected
4. Click (without dragging) — the `onClick` handler should still fire
5. Multi-select shapes and drag — existing behavior should be preserved

- [x] Unit tests

### Release notes

- Fix dragging unselected shapes that have an `onClick` handler.

---

> [!NOTE]
> **Medium Risk**
> Touches the `SelectTool` state machine to change when selection is applied during a drag, which could subtly affect selection/translation behavior in edge cases (e.g. grouped or multi-selected shapes). Covered by a new unit test, but still impacts core interaction logic.
> 
> **Overview**
> Fixes a selection/translation edge case where shapes with a `ShapeUtil.onClick` handler could not be dragged while unselected.
> 
> `PointingShape.startTranslating` now ensures the hit shape is selected (only when it wasn’t selected on enter and there’s no existing selection) before transitioning to translating, preserving click-to-run-`onClick` behavior.
> 
> Adds a regression unit test for dragging shapes with `onClick`, and introduces a new examples page (`shape-with-onClick`) demonstrating a clickable custom shape that increments a counter via `ShapeUtil.onClick`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c7f6206cc8261f45ee94a33418e6780f27c599df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


kaneel:
I'm observing something that looks weird, after dragging I can see the handles, trying to resize will "drag along", Is this a regression? 

https://github.com/user-attachments/assets/ff6bd666-6a82-479c-8ba3-7b696c041d7b

mimecuvalo (approved):
LGTM

steveruizok:
> I'm observing something that looks weird, after dragging I can see the handles, trying to resize will "drag along", Is this a regression?
> 
>  dragandresizewithonclick.mov

This is just because the shape doesn't implement an onResize, unrelated to this PR. I'll make it extend baseboxutil though.

kaneel (approved):
My bad here, LGTM

PR #7942, opened 2026-02-15
feat(editor): pass through clicks on transparent image pixels

### LINKED ISSUE #7529: Pass through click events on transparent parts of PNGs
## Problem

When a PNG image with transparent areas is placed on the canvas, clicks on transparent pixels should pass through to objects behind them. Currently, clicking on any part of the image (including transparent areas) selects the image itself, which makes it difficult to interact with layered content.

## Steps to Reproduce

1. Create or import a PNG image with transparent areas
2. Place it on the canvas over another object
3. Try clicking on the transparent part of the PNG
4. Observe that the image gets selected instead of the object behind it

## Expected Behavior

- Clicks on transparent pixels should pass through to the object behind the image
- Only clicks on opaque pixels should select the image
- This would allow for better layering and interaction with overlapped content

## Actual Behavior

- All clicks on the image (transparent or opaque) select the image
- Objects behind transparent areas cannot be clicked

## Technical Details

This would require detecting pixel transparency when handling click events on PNG images. The canvas would need to check the alpha channel of the PNG at the click coordinate and only register a hit if the pixel is opaque (or above a certain opacity threshold).

## Benefits

- Improved UX when working with layered images
- More intuitive interaction model for transparent overlays
- Better alignment with industry-standard design tools

steveruizok:
Closes #7529

In order to allow clicks on transparent parts of PNG/WebP/GIF/AVIF images to pass through to shapes behind them, this PR adds an `ignoreHit()` mechanism to the geometry system and uses it to check image alpha data at the click point.

https://github.com/user-attachments/assets/831ea7e8-fdbf-4211-9b29-3337f117de68

### How it works

1. **`Geometry2d.ignoreHit(point)`** — New method on the base geometry class (default: `false`). Called after a standard hit test succeeds. If it returns `true`, the shape is skipped and the click passes through to shapes behind it. Also wired through `TransformedGeometry2d` via inverse matrix.

2. **`ImageAlphaCache`** — Preloads and caches a downsampled (max 256×256) alpha channel for each image. Uses `Uint32Array` with unsigned right shift for efficient alpha byte extraction. The preload URL and cache key are separated so resolved/CDN URLs work correctly with `asset.props.src` lookups. A shared `OffscreenCanvas` is reused across preloads, with canvas operations kept synchronous after the async `createImageBitmap` to prevent concurrent corruption.

3. **`ImageAlphaGeometry.ts`** — Contains `ImageRectangle2d` (extends `Rectangle2d`) and `ImageEllipse2d` (extends `Ellipse2d`), both with `ignoreHit` and `hitTestPoint` overrides that map shape-space points to normalized image coordinates (accounting for crop, flipX, flipY) and check the alpha cache. Coordinates are clamped to `[0,1]` so edge-margin hits slightly outside bounds map to the nearest edge pixel.

4. **`ImageShapeUtil`** — `getGeometry()` now returns `ImageRectangle2d` for transparent image formats, or `ImageEllipse2d` when `crop.isCircle` is set. The `ImageShape` component preloads alpha data when the resolved URL is available, using `asset.props.src` as the cache key.

5. **`Editor.getShapeAtPoint`** — After a filled shape geometry hit succeeds, calls `geometry.ignoreHit(pointInShapeSpace)`. If `true`, continues to check shapes behind it.

The `hitTestPoint` override on both geometry classes ensures that selection paths using `isPointInShape` / `getSelectedShapeAtPoint` also skip transparent pixels. Before alpha data is loaded, images behave as fully opaque (no change from current behavior).

### Change type

- [x] `feature`

### Test plan

1. `yarn dev` → import a PNG with transparency over another shape
2. Click on a transparent area of the PNG → the shape behind should be selected
3. Click on an opaque area of the PNG → the image should be selected
4. Test with cropped, flipped, and circle-cropped images
5. Test edge clicks near the boundary of transparent images

- [x] Unit tests

### Release notes

- Add click-through on transparent pixels of PNG, WebP, GIF, and AVIF images. Clicking transparent areas now selects shapes behind the image instead of the image itself. Works with cropped, flipped, and circle-cropped images.

### API changes

- Added `Geometry2d.ignoreHit(point)` — allows geometries to reject successful hit tests (e.g. for transparent pixels)

### DIFF
--- a/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
+++ b/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
@@ -24,8 +24,11 @@ export interface ImageAlphaGeometryConfig {
 	flipY: boolean
 }
 
-/** Map a point in shape space to normalized [0,1] image coordinates, accounting for crop and flip. */
-export function mapToImageCoords(
+/**
+ * Map a point in shape space to normalized [0,1] image coordinates, accounting for crop and flip.
+ * @internal
+ */
+function mapToImageCoords(
 	config: ImageAlphaGeometryConfig,
 	point: VecLike,
 	bounds: { minX: number; minY: number; w: number; h: number }
@@ -86,15 +89,23 @@ function extractAlphas(ctx: OffscreenCanvasRenderingContext2D, w: number, h: num
 	return alphas
 }
 
-/** Start loading alpha data for a given image URL. No-op if already loaded or loading. */
-export function preloadAlphaData(src: string): void {
-	if (alphaCache.has(src) || pending.has(src)) return
-	pending.add(src)
+/**
+ * Start loading alpha data for a given image URL. No-op if already loaded or loading.
+ *
+ * @param url - The URL to fetch the image from (may be a resolved/optimized CDN URL).
+ * @param cacheKey - The key to store/lookup the alpha data under. Defaults to `url`.
+ *   Pass `asset.props.src` here so that `getAlphaData(asset.props.src)` in getGeometry
+ *   finds data that was preloaded from a resolved URL.
+ */
+export function preloadAlphaData(url: string, cacheKey?: string): void {
+	const key = cacheKey ?? url
+	if (alphaCache.has(key) || pending.has(key)) return
+	pending.add(key)
 
 	const img = Image()
 	img.crossOrigin = 'anonymous'
 	img.onload = async () => {
-		pending.delete(src)
+		pending.delete(key)
 		const { width: origW, height: origH } = img
 		if (origW === 0 || origH === 0) return
 
@@ -127,12 +138,12 @@ export function preloadAlphaData(src: string): void {
 			ctx.drawImage(img, 0, 0, w, h)
 		}
 
-		alphaCache.set(src, { width: w, height: h, alphas: extractAlphas(ctx, w, h) })
+		alphaCache.set(key, { width: w, height: h, alphas: extractAlphas(ctx, w, h) })
 	}
 	img.onerror = () => {
-		pending.delete(src)
+		pending.delete(key)
 	}
-	img.src = src
+	img.src = url
 }
 
 /** Get cached alpha data for a URL, or null if not yet loaded. */
--- a/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
@@ -342,12 +342,15 @@ const ImageShape = memo(function ImageShape({ shape }: { shape: TLImageShape }) 
 
 	const mimeType = asset && 'mimeType' in asset.props ? asset.props.mimeType : null
 	const supportsTransparency = mimeType != null && TRANSPARENT_IMAGE_MIMETYPES.includes(mimeType)
+	const assetSrc = asset && 'src' in asset.props ? asset.props.src : null
 
 	useEffect(() => {
 		if (url && supportsTransparency) {
-			preloadAlphaData(url)
+			// Cache under asset.props.src so getGeometry (which only has the asset
+			// record) can look up the data even when the resolved URL differs.
+			preloadAlphaData(url, assetSrc ?? undefined)
 		}
-	}, [url, supportsTransparency])
+	}, [url, supportsTransparency, assetSrc])
 
 	const showCropPreview = useValue(
 		'show crop preview',


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5403,6 +5403,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 				// Are we close to the shape's edge?
 				if (distance <= outerMargin || (hitInside && distance <= 0 && distance > -innerMargin)) {
 					if (geometry.isFilled || (isGroup && geometry.children[0].isFilled)) {
+						// If the geometry rejects this hit (e.g. transparent image pixel),
+						// skip this shape and check shapes behind it.
+						if (geometry.rejectHit(pointInShapeSpace)) {
+							continue
+						}
```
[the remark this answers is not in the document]
steveruizok: grouped shapes do ignore hits on transparent areas

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,70 @@
...
+		canvas.width = w
+		canvas.height = h
+		const ctx = canvas.getContext('2d')
+		if (!ctx) return
+
+		ctx.drawImage(img, 0, 0, w, h)
+		const imageData = ctx.getImageData(0, 0, w, h)
+		const alphas = new Uint8Array(w * h)
+		for (let i = 0; i < alphas.length; i++) {
+			alphas[i] = imageData.data[i * 4 + 3]
```
kaneel: Wondering how Claude could extract a known trick for checking alphas and it seems it's right on the money with this: 

```
const imageData = ctx.getImageData(0, 0, w, h);
// this converts from [R, G, B, A, R, G, B, A...] to [RGBA, RGBA, RGBA...]
const pixels = new Uint32Array(imageData.data.buffer);
const alphas = new Uint8Array(w * h);
for (let i = 0; i < alphas.length; i++) {
  // >>> shift! 
  alphas[i] = pixels[i] >>> 24;
}```

Wondering if this would be much faster, or not! Worth measuring?

Comment (kaneel) [reply packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts]: Way to check this work btw:

```ts
// Simulate 2 pixels: [R, G, B, A, R, G, B, A]
const bytes = new Uint8ClampedArray([255, 0, 0, 128, 0, 255, 0, 64]);

const pixels = new Uint32Array(bytes.buffer);

console.log(pixels[0].toString(16)); // 800000ff
// Memory: [ff, 00, 00, 80]
// Little-endian reads lowest address first:
// 0xff | (0x00 << 8) | (0x00 << 16) | (0x80 << 24) = 0x800000ff

console.log(pixels[0] >>> 24); // 128 ✓ (that's our alpha)
console.log(pixels[1] >>> 24); // 64  ✓

// Compare to the manual way:
console.log(bytes[0 * 4 + 3]);  // 128 ✓
console.log(bytes[1 * 4 + 3]);  // 64  ✓
```

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,70 @@
...
+	img.onload = () => {
+		pending.delete(src)
+		const { width: origW, height: origH } = img
+		if (origW === 0 || origH === 0) return
+
+		const scale = Math.min(1, MAX_SIZE / Math.max(origW, origH))
+		const w = Math.max(1, Math.round(origW * scale))
+		const h = Math.max(1, Math.round(origH * scale))
+
+		const canvas = document.createElement('canvas')
```
kaneel: it's 2026 and it should not matter much but we can shave some ms by reusing an offscreen canvas for these manipulations

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -76,14 +79,46 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
 	}
 
 	override getGeometry(shape: TLImageShape): Geometry2d {
+		const asset = shape.props.assetId ? this.editor.getAsset(shape.props.assetId) : null
+		const mimeType = asset && 'mimeType' in asset.props ? asset.props.mimeType : null
+		const supportsTransparency =
+			mimeType != null &&
+			(mimeType.includes('png') || mimeType.includes('webp') || mimeType.includes('gif'))
```
kaneel: Not well versed in all the possibilities yet, do we support `avif`? this mimetype also support transparency.

kaneel: Also, this is doing partial string checks on a three IFs condition, there's a cleaner way to do this and it becomes much more editable when we want to add a mimetype

```ts
[MIME_PNG, MIME_WEBP, MIME_GIF].includes(mimeType);
```

Arguably, we could also do 

```ts
TRANSPARENT_IMAGE_MIMETYPES.includes(mimeType);
```

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
kaneel: (However, I doubt this would be much faster)

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
steveruizok: Want to jump on this one @kaneel? If you get to it before I do, I think these would all be good improvements

kaneel: I could jump on it after my first forays in to the world of mermaid diagrams, yes :)

packages/editor/src/lib/primitives/geometry/Geometry2d.ts
```diff
@@ -420,6 +420,14 @@ export abstract class Geometry2d {
 		return length
 	}
 
+	/**
+	 * Called after a hit test succeeds. Return `true` to reject the hit and allow
+	 * shapes behind this one to be selected instead (e.g. transparent image pixels).
+	 */
+	rejectHit(_point: VecLike): boolean {
```
mimecuvalo: i would rename this - `shouldHitPassthrough` or something. `rejectHit` is harder to reason b/c of the success/reject, sorta inversion logic to think about

packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
```diff
@@ -308,6 +343,17 @@ const ImageShape = memo(function ImageShape({ shape }: { shape: TLImageShape })
 		}
 	}, [editor, isAnimated, prefersReducedMotion, url])
 
+	const mimeType = asset && 'mimeType' in asset.props ? asset.props.mimeType : null
+	const supportsTransparency =
+		mimeType != null &&
+		(mimeType.includes('png') || mimeType.includes('webp') || mimeType.includes('gif'))
```
mimecuvalo: sorta related to @kaneel 's comment above - we should do avif here.

at the same time though, i don't think we should support animated ones - that sounds like it would be too intense.

packages/tldraw/src/lib/shapes/image/ImageRectangle2d.ts
```diff
@@ -0,0 +1,69 @@
+import { Geometry2dOptions, Rectangle2d, VecLike } from '@tldraw/editor'
+import { AlphaData, isPointTransparent } from './ImageAlphaCache'
+
+/** @internal */
+export class ImageRectangle2d extends Rectangle2d {
```
mimecuvalo: feels like this shares a bunch of logic with ImageEllipse2d - prbly there should be a parent class they derive from

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,70 @@
...
+/** Start loading alpha data for a given image URL. No-op if already loaded or loading. */
+export function preloadAlphaData(src: string): void {
+	if (alphaCache.has(src) || pending.has(src)) return
+	pending.add(src)
+
+	const img = Image()
+	img.crossOrigin = 'anonymous'
+	img.onload = () => {
+		pending.delete(src)
+		const { width: origW, height: origH } = img
```
mimecuvalo, line 109: perf questions: 
- if this is a large image will this bottleneck the main thread? is there a "quick-and-dirty" pass to get a rough set of the transparent points?

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
mimecuvalo: 💯 definitely agree, esp for larger images

packages/editor/src/lib/primitives/geometry/Geometry2d.ts
steveruizok: hm, maybe ignoreHit is a better one than reject. Depending on the type of geometry (e.g. a hollow shape) a hit will still pass through, so I think an explicit passthrough would confuse that. We really want to pretend for the sake of the hit test that this shape was never there, or at least was somewhere else and missed by the hit.

steveruizok: forceMiss, ignoreHit, rejectHit

mimecuvalo: ignoreHit is a bit better yeah

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,70 @@
...
+const pending = new Set<string>()
+
+/** Start loading alpha data for a given image URL. No-op if already loaded or loading. */
+export function preloadAlphaData(src: string): void {
+	if (alphaCache.has(src) || pending.has(src)) return
+	pending.add(src)
+
+	const img = Image()
+	img.crossOrigin = 'anonymous'
+	img.onload = () => {
```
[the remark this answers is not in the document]
steveruizok: Not addressing — without CORS headers the browser won't let us read pixel data (canvas is tainted). This is a fundamental browser security constraint. The fallback is correct: images without CORS just behave as opaque (no click-through), which is the same behavior as before this PR.

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface AlphaData {
+	width: number
+	height: number
+	/** Row-major alpha values (0–255) */
+	alphas: Uint8Array
+}
+
+const MAX_SIZE = 256
+
+const alphaCache = new Map<string, AlphaData>()
```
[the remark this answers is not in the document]
steveruizok: Not addressing — the alpha data is small (max 256×256 = 64KB per entry) and keyed by asset.props.src which is bounded by the number of image assets in the document. This won't grow unboundedly in practice.

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,156 @@
...
+		nx = topLeft.x + nx * (bottomRight.x - topLeft.x)
+		ny = topLeft.y + ny * (bottomRight.y - topLeft.y)
+	}
+
+	// Account for flips
+	if (config.flipX) nx = 1 - nx
+	if (config.flipY) ny = 1 - ny
+
+	return { nx, ny }
+}
```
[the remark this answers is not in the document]
steveruizok: Already addressed — mapToImageCoords was unexported in the latest push.

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,167 @@
...
+
+function getOffscreenCanvas(w: number, h: number): OffscreenCanvas {
+	if (!offscreenCanvas) {
+		offscreenCanvas = new OffscreenCanvas(w, h)
+	} else {
+		offscreenCanvas.width = w
+		offscreenCanvas.height = h
+	}
+	return offscreenCanvas
+}
```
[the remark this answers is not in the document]
steveruizok: Not a concern — OffscreenCanvas has been supported in all major browsers since Safari 16.4 (March 2023). And the code already handles the failure case: getContext('2d') returns null and we bail with an early return.

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -0,0 +1,167 @@
...
+		nx = topLeft.x + nx * (bottomRight.x - topLeft.x)
+		ny = topLeft.y + ny * (bottomRight.y - topLeft.y)
+	}
+
+	// Account for flips
+	if (config.flipX) nx = 1 - nx
+	if (config.flipY) ny = 1 - ny
+
+	return { nx, ny }
+}
```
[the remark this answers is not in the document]
steveruizok: Not a real issue — crop values are always normalized to [0,1] by construction, so post-crop coordinates stay in range. The input point is clamped to [0,1] before crop mapping, and isPointTransparent further clamps indices via Math.min(Math.floor(...), width-1).

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
steveruizok: Shouldn't be an issue now — createImageBitmap does the decode+resize off the main thread, and all synchronous work after that operates on a max 256×256 canvas (65K pixels). The MAX_SIZE=256 downsampling is effectively the quick-and-dirty pass.

PR #7949, opened 2026-02-15
perf(resize): batch DOM measurements during multi-shape resize

steveruizok:
In order to eliminate layout thrashing when resizing multiple geo shapes with text labels, this PR batches all label DOM measurements into a single pass per frame.

https://github.com/user-attachments/assets/40f361f0-e619-4c75-9f64-7e302898d158

Previously, each geo shape's `onResize` independently measured its label via `measureHtml`, causing the browser to recalculate layout for every shape. With many labeled shapes selected, this created a costly read-write-read-write cycle.

This PR introduces a three-part optimization:

1. **`TextManager.measureHtmlBatch`** — A new method that measures multiple HTML strings in a single batch using a pool of reusable DOM elements. All style/innerHTML writes happen first, then all measurements are read, triggering only one browser layout reflow. The pool grows on demand and caches innerHTML to skip redundant parsing.

2. **`GeoShapeUtil` measurement helpers** — `measureUnscaledLabelSize` is refactored to extract `getGeoLabelMeasurementRequest` (builds html + opts without measuring) and `getGeoResizeTargetWidth` (computes target width during resize). A module-level batch cache (`setBatchLabelSizeCache`) lets `onResize` and `getGeometry` skip individual measurements when pre-computed sizes are available.

3. **`Resizing.batchMeasureGeoLabels`** — Before each resize loop, collects measurement requests for all eligible geo shapes, batch-measures them via `measureHtmlBatch`, and populates the cache. The cache is cleared on exit.

Original branch: `steveruiz/batch-resize-measurements`

Before:

image

After:

image

### Change type

- [x] `improvement`

### Test plan

1. Create several geo shapes with text labels
2. Select all shapes and resize from a corner handle
3. Verify shapes resize correctly with labels wrapping as expected
4. Verify single-shape resize still works correctly
5. Verify shapes without labels resize normally

### API changes

- Added `TextManager.measureHtmlBatch()` for batched DOM text measurement

### Release notes

- Improve resize performance for multiple geo shapes with text labels by batching DOM measurements

---

> [!NOTE]
> **Medium Risk**
> Changes core resize-path behavior for `geo` shapes by introducing per-frame batched label measurement and caching, which could affect label wrapping/sizing or cleanup if edge cases slip through. Adds new public `TextManager` measurement types/APIs, so downstream consumers may be impacted by signature/type changes.
> 
> **Overview**
> **Improves multi-shape resize performance** by batching `geo` label measurements into a single DOM write-then-read pass per frame, avoiding per-shape layout thrash.
> 
> Adds `TextManager.measureHtmlBatch()` plus new public types (`BatchMeasurementRequest`, `TLMeasuredTextSize`) and updates `measureHtml`/`measureText` to return the shared type. `GeoShapeUtil` now supports building measurement requests, uses a per-editor batch cache (set during resize), and `Resizing` populates/clears this cache around the resize loop so `onResize`/geometry can reuse precomputed label sizes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 68dfd7d3ca41f3a20e4528f4d5650a9d8a4c2162. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -650,39 +658,74 @@ function expandShapeForFirstLabel(
...
 		fontFamily: FONT_FAMILIES[font],
 		fontSize: LABEL_FONT_SIZES[size],
 		minWidth: minWidth,
 		maxWidth: Math.max(
-			// Guard because a DOM nodes can't be less 0
 			0,
-			// A 'w' width that we're setting as the min-width
 			Math.ceil(minWidth + EXTRA_PADDINGS[size]),
-			// The actual text size
 			Math.ceil(w / shape.props.scale - LABEL_PADDING * 2)
```
kaneel, line 737: I liked these comments 🤔 
I think we should keep, just in case of!

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -107,7 +107,94 @@ export class TextManager {
...
+			return this.poolElms[index]
+		}
+		// Create new pool elements on demand
+		while (this.poolElms.length <= index) {
+			const el = document.createElement('div')
+			el.classList.add('tl-text')
+			el.classList.add('tl-text-measure')
+			el.setAttribute('dir', 'auto')
+			el.tabIndex = -1
+			for (const key of objectMapKeys(initialDefaultStyles)) {
```
mimecuvalo: this feels like this will leak styles. below we do `setElementStyles` which i think should be considered here

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -107,7 +107,94 @@ export class TextManager {
...
+	// Pool of temporary measurement elements for batch measurements
+	private poolElms: HTMLDivElement[] = []
+	private poolElmHtml: string[] = []
+
+	private getPoolElm(index: number): HTMLDivElement {
+		if (index < this.poolElms.length) {
+			return this.poolElms[index]
+		}
+		// Create new pool elements on demand
+		while (this.poolElms.length <= index) {
```
mimecuvalo: this pool never gets smaller?

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -107,7 +107,94 @@ export class TextManager {
...
+		requests: Array<{ html: string; opts: TLMeasureTextOpts }>
+	): Array<BoxModel & { scrollWidth: number }> {
+		if (requests.length === 0) return []
+
+		// Write phase: apply styles and innerHTML to each pooled element
+		for (let i = 0; i < requests.length; i++) {
+			const { html, opts } = requests[i]
+			const el = this.getPoolElm(i)
+
+			el.style.setProperty('font-family', opts.fontFamily)
```
mimecuvalo: we should reuse some logic here that already exists in the TextManager

packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts
```diff
@@ -428,6 +443,89 @@ export class Resizing extends StateNode {
 		}
 	}
 
+	/**
+	 * Batch-measure all geo shape labels before the resize loop to avoid layout thrashing.
+	 * For each geo shape with a non-empty label that has compatible rotation, compute the
+	 * measurement request and batch all measurements in a single DOM pass.
+	 */
+	private batchMeasureGeoLabels(
```
mimecuvalo: this introduces a pretty tight coupling with the geo shapes for the select tool.
as geo shapes adapt and change, it feels like these pieces of logic will drift apart from one another.
also, why limit it to just geo shapes? seems pretty limiting to what this is hoping to accomplish.

thoughts?

packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts
```diff
@@ -363,6 +371,13 @@ export class Resizing extends StateNode {
 			})
 		}
 
+		// Pass 1: Batch-measure all geo shape labels to avoid layout thrashing.
+		// We collect measurement requests for all geo shapes with text, measure them
+		// in a single batch (one layout reflow), then set the cache so onResize can
+		// use pre-computed sizes instead of measuring individually.
+		this.batchMeasureGeoLabels(shapeSnapshots, scale, selectionRotation, isAspectRatioLocked)
```
mimecuvalo: should this always do batch? if it's just one item we should do the regular branch of logic?

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -650,39 +658,74 @@ function expandShapeForFirstLabel(
...
+	const minWidth = MIN_WIDTHS[size]
 	const html = renderHtmlFromRichTextForMeasurement(editor, richText)
-	const textSize = editor.textMeasure.measureHtml(html, {
+	const opts: TLMeasureTextOpts = {
 		...TEXT_PROPS,
 		fontFamily: FONT_FAMILIES[font],
 		fontSize: LABEL_FONT_SIZES[size],
 		minWidth: minWidth,
 		maxWidth: Math.max(
-			// Guard because a DOM nodes can't be less 0
```
mimecuvalo, line 679: we can leave these comments in

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -650,39 +658,74 @@ function expandShapeForFirstLabel(
 	return { w, h }
 }
 
+// Module-level batch cache, set by Resizing.ts before the resize loop and cleared after.
+// When set, onResize will use pre-computed label sizes instead of measuring individually.
+let _batchLabelSizeCache: Map<TLShapeId, { w: number; h: number }> | null = null
+
+/** @internal */
+export function setBatchLabelSizeCache(cache: Map<TLShapeId, { w: number; h: number }> | null) {
```
mimecuvalo: hm, just to check, if there's multiple editors on the page this starts getting a bit weird, yeah? since it's module-level

packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts
steveruizok: Yeah, we should address a fix here. But geo shapes are effectively broken right now (multiple text shapes resize by scale, not width) so I'm willing at the worst to ship a fix until we find a general solution.

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -106,8 +106,125 @@ export class TextManager {
 		}
 	}
 
+	/** Apply measurement styles to any element (shared by measureHtml and measureHtmlBatch). */
+	private applyMeasureStyles(el: HTMLElement, opts: TLMeasureTextOpts) {
+		el.style.setProperty('font-family', opts.fontFamily)
```
mimecuvalo: add comment here that any new css property needs to be cleaned up in `resetMeasureStyles`

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -106,8 +106,125 @@ export class TextManager {
...
+			el.style.setProperty(key, initialDefaultStyles[key])
+		}
+		// Reset measurement styles to defaults
+		el.style.removeProperty('font-family')
+		el.style.removeProperty('font-style')
+		el.style.removeProperty('font-weight')
+		el.style.removeProperty('font-size')
+		el.style.removeProperty('line-height')
+		el.style.removeProperty('padding')
+	}
```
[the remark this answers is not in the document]
mimecuvalo: this is still an issue to resolve

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -106,8 +106,125 @@ export class TextManager {
...
+		while (this.poolElms.length <= index) {
+			const el = document.createElement('div')
+			el.classList.add('tl-text')
+			el.classList.add('tl-text-measure')
+			el.setAttribute('dir', 'auto')
+			el.tabIndex = -1
+			for (const key of objectMapKeys(initialDefaultStyles)) {
+				el.style.setProperty(key, initialDefaultStyles[key])
+			}
+			this.editor.getContainer().appendChild(el)
```
mimecuvalo: let's use a `DocumentFragment` here https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment

when batch-adding elements to the DOM it's more efficient to go the fragment route

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -106,8 +106,125 @@ export class TextManager {
...
+				y: 0,
+				w: rect.width,
+				h: rect.height,
+				scrollWidth,
+			})
+		}
+
+		// Reset phase: clean up styles on used elements to avoid leaking
+		for (let i = 0; i < requests.length; i++) {
+			this.resetMeasureStyles(this.getPoolElm(i))
```
mimecuvalo: i think the reset makes more sense to put before line 189. at the moment it's doing:
- apply styles
- measure the dom
- reset the styles

this'll cause two layout recalculations vs. we could collapse that into one layout collection i believe

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -106,8 +106,125 @@ export class TextManager {
...
+		}
+
+		// Trim the pool: remove excess elements beyond what was needed
+		while (this.poolElms.length > requests.length) {
```
mimecuvalo: similar to the prev. comment - prbly better to do this beforehand

mimecuvalo (commented):
@kostyafarber here's a couple more comments to address since you're taking over this work!

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
kostyafarber: thats a great shout

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
kostyafarber: refactoring to do this dynamically instead so we don't have to worry about drifting and updating two places for every new property

kostyafarber:
@mimecuvalo addressed your comments + made cleaned up a bit of code and making it a bit readable, mainly
* made the style setting/resetting dynamic so we don't get drift and have to remember to add code in two places
* added a few types and extracting a few things into helper methods to share code

as discussed this only works for geo shapes, not ideal. however, a decoupling of this is a bigger task and requires more deliberation on the api design/surface so would be better suited in a follow up, this can go in as a tactical fix.

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -280,11 +386,11 @@ export class TextManager {
 			width: `${elementWidth}px`,
 			height: 'min-content',
 			'text-align': textAlignmentsForLtr[opts.textAlign],
-			'overflow-wrap': shouldTruncateToFirstLine ? 'anywhere' : undefined,
-			'word-break': shouldTruncateToFirstLine ? 'break-all' : undefined,
+			'overflow-wrap': shouldTruncateToFirstLine ? 'anywhere' : 'break-word',
+			'word-break': shouldTruncateToFirstLine ? 'break-all' : 'normal',
```
mimecuvalo, line 390: why the explicit settings here now vs. undefined?

kostyafarber: this was related to [this](https://github.com/tldraw/tldraw/pull/7949#discussion_r2932721779), the `setElementStyles` restore function was changed to remove elements that have `undefined` whereas on main we just skipped processing those elements, and relied on the `TextManager` constructor to maintain default values.

currently, this only affects the single element measurement and the measure text spans route, as the batching uses ` resetElementStyles` to reset styles. 

It was added to make setting styles more explicit and not relying on the constructor to maintain default styles in these routes. If we left it as `undefined` then it would remove that style and it would inherit the browsers default style, which is wrong.

to have it behave the same with 'undefined', we could revert back to skipping undefined css values, it was a design choice

mimecuvalo: ah right right, makes sense 👍

MitjaBezensek:
Looks good 👍

PR #7975, opened 2026-02-16
Revert "ci: use node 24 and actions/cache@v5 (#7920)"

kaneel:
There's been a quite recent change in e2e test time + flakyness that seems to be have been brought by #7920, 
Reverting #7920 shows that e2e falls back to acceptable time. We can discuss afterward how to move forward

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`


mimecuvalo:
addressed in https://github.com/tldraw/tldraw/pull/7985

PR #7988, opened 2026-02-17
test(e2e): harden dotcom share link copy in CI

mimecuvalo:
In order to reduce flaky dotcom sharing tests, this PR hardens the Playwright share-link copy helper so tests only proceed once clipboard content is freshly updated and points to the expected file/publish routes.

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn e2e-dotcom`.
2. Confirm `sharing.spec.ts` no longer intermittently picks up stale `/invite/...` links during `copyLink()`.
3. Verify copied URLs in these tests resolve to `/f/...` or `/p/...` before opening a new context.

- [ ] Unit tests
- [x] End to end tests

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/helpers.ts
+++ b/apps/dotcom/client/e2e/fixtures/helpers.ts
@@ -17,15 +17,10 @@ export function sleep(ms: number): Promise<void> {
 
 export async function openNewTab(
 	browser: Browser,
-	opts: {
-		url?: string
-		userProps: UserProps
-		allowClipboard?: boolean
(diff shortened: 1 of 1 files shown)


MitjaBezensek:
Thanks for this!

Strange that these started flaking out, don't remember them being flaky in the past. What do you think caused this?

mimecuvalo:
> Strange that these started flaking out, don't remember them being flaky in the past. What do you think caused this?

so, Node 24 required an upgrade to Playwright and from what i can tell, that upgrade to a newer Playwright caused this regression

PR #8004, opened 2026-02-18
feature(dotcom): Introducing the tldrawing of mermaid diagrams

kaneel:
Still draft-y, needs tests

This implement mermaid diagram parsing into tldraw a meaningful way. Meaning that instead just going with "redrawing" svgs, we want the created shapes to be bound meaningfully, labels be part of the related shapes, etcetc. 

For now, this PR brings support for
1. flowcharts
2. sequence (with notes only for now)
3. state diagram 

This started as something a bit smarter (running dagre ourselves) and devoluted into just a svg -> tldraw shapes exercise!

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

TO BE DEFINED.

### Release notes

- Started implementing parsing of mermaid diagram: you can now copy flowchart, state and sequence mermaid diagram and see them tldraw-ed!

---

> [!NOTE]
> **Medium Risk**
> Introduces a sizable new dependency (`mermaid`) and new external text handling that does DOM/SVG parsing and shape creation, which could affect editor performance and paste behavior if edge cases slip through.
> 
> **Overview**
> Adds first-class Mermaid support by introducing a new workspace package `@tldraw/mermaid` that parses Mermaid text, converts supported diagram types (flowchart, sequence, state) into a blueprint, and renders *editable, bound* tldraw shapes (with grouping, styling, and edge bend/anchor handling).
> 
> Integrates this into the dotcom editors via a new `SneakyMermaidHandler` external text handler that detects Mermaid-like text on paste/drop, dynamically imports the Mermaid package, selects newly created shapes, and falls back to default text handling (including an SVG fallback + warning toast for unsupported diagrams). Also wires up i18n strings and build references, and pulls in Mermaid + its transitive dependencies via lockfile updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2558a5d0d7070fcf8a7c44b67b734314ac33b794. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


steveruizok:
!

steveruizok:
I feel like we should add 10% to the size of shapes as a rule just to help with breaking text. @max-drake any tips from fairies?

kaneel:
> I feel like we should add 10% to the size of shapes as a rule just to help with breaking text. @max-drake any tips from fairies?

We can but mind you I'm still not fully done with all the details. I'm trying to get as close in some cases to what mermaid js auto generate as svg, there's still one or two cases that aint right. Then the next step would be: make it easy to "adjust" and then… tests.

max-drake:
> I feel like we should add 10% to the size of shapes as a rule just to help with breaking text. @max-drake any tips from fairies?

hmm not sure exactly how helpful any of the learnings from fairies are bc most of the ways we handle too-long text is through the system prompt. however @kaneel if you haven't come across these already we used the `TextManager.ts` and `growY` shape prop at different points to measure text and/or detect when text had caused a shape to grow too much

kaneel:
> > I feel like we should add 10% to the size of shapes as a rule just to help with breaking text. @max-drake any tips from fairies?
> 
> hmm not sure exactly how helpful any of the learnings from fairies are bc most of the ways we handle too-long text is through the system prompt. however @kaneel if you haven't come across these already we used the `TextManager.ts` and `growY` shape prop at different points to measure text and/or detect when text had caused a shape to grow too much

I'll look into it thanks, I'm still trying to rearrange the code, move some stuff in utils and constants; once I've cleaned this a bit I'll look into where we have these text issues, if we still have them!

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -0,0 +1,55 @@
...
+				}
+			} catch (e) {
+				if (e instanceof UnsupportedMermaidDiagramError) {
+					addToast({
+						id: 'unsupported-mermaid-diagram',
+						title: unsupportedTitle,
+						description: unsupportedDescription,
+						severity: 'warning',
+					})
+					const { svg } = await mermaid.render(`mermaid-${Date.now()}`, content.text)
```
mimecuvalo: one high-level feedback here, we should be sure to lazy load the `mermaid` package. it's a power-user feature so this shouldn't be part of our default JS bundle

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -0,0 +1,61 @@
+import { useEffect } from 'react'
+import { defaultHandleExternalTextContent, useEditor, useToasts } from 'tldraw'
+import { defineMessages, useMsg } from '../../tla/utils/i18n'
+import { createMermaidDiagram, MermaidDiagramError } from './utils/createMermaidDiagram'
```
mimecuvalo: we should delay-load this code, `await import('./utils/createMermaidDiagram')`
let's only load it after using some heuristic that indicates that we're dealing with a mermaid diagram, say the presence of `-->` and any other relevant indicators

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -0,0 +1,61 @@
...
+						severity: 'warning',
+					})
+
+					if (typeof svgResult === 'string') {
+						editor.putExternalContent({
+							type: 'svg-text',
+							text: svgResult,
+							point: content.point,
+							sources: content.sources,
+						})
```
mimecuvalo: should this return instead of falling thru to `await defaultHandleExternalTextContent`?

packages/mermaid/src/createMermaidDiagram.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	const parsedResult = await mermaid.parse(text, { suppressErrors: true })
+	if (!parsedResult) {
+		throw new MermaidDiagramError('not a mermaid diagram', 'parse')
+	}
+
+	try {
+		const parsedSvg = (await mermaid.render(`mermaid-${Date.now()}`, text)).svg
+		// eslint-disable-next-line @typescript-eslint/no-deprecated
+		const diagramResult = await mermaid.mermaidAPI.getDiagramFromText(text)
```
mimecuvalo, line 76: checking here: it says this `mermaidAPI` is deprecated - is there a newer API or way of doing this?
also, i note that we're importing types by digging deep into the package b/c of this...

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/flowchart.ts
```diff
@@ -0,0 +1,210 @@
...
+// Flowchart-specific ID parsers
+// ---------------------------------------------------------------------------
+
+function flowchartNodeIdParser(domId: string): string {
+	const m = domId.match(/^flowchart-(.+)-\d+$/)
+	return m ? m[1] : domId
+}
+
+function flowchartEdgeIdParser(dataId: string): { start: string; end: string } | null {
+	const m = dataId.match(/^L_(.+)_([^_]+)_\d+$/)
```
mimecuvalo: let's put an example of what this regex is testing against, will make it easier to understand

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/flowchart.ts
```diff
@@ -0,0 +1,210 @@
...
+): { w: number; h: number } {
+	const geo = mapFlowShapeTypeToGeo(v.type)
+
+	if (geo === 'diamond') {
+		return { w: svgNode.width, h: svgNode.height }
+	}
+	if (geo === 'ellipse') {
+		return { w: svgNode.width + 20, h: svgNode.height + 10 }
+	}
+	return { w: svgNode.width + 10, h: svgNode.height + 5 }
```
mimecuvalo: mm, these number seem maybe a little magical:
- can we use a shared number and use it as a base multiplier (say 5 as the base) and then multiply as needed?
- why is the default 10/5? might be worth a comment

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+			return 'rectangle'
+	}
+}
+
+// ---------------------------------------------------------------------------
+// classDef: shared helper to extract fill colors from Mermaid SVG strings
+// ---------------------------------------------------------------------------
+
+const HEX_TO_TLDRAW_COLOR: Record<string, TLDefaultColorStyle> = {
+	'000000': 'black',
```
mimecuvalo: mm, if the color isn't in our mapping, can we use a closest color match? i'd hate for some of the usage of hex here to maybe get dropped on the floor.

packages/mermaid/src/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+ * @param knownClasses - Built-in CSS classes to skip (e.g. "node", "statediagram-state")
+ */
+export function parseClassDefFills(
+	svgString: string,
+	idPrefix: string,
+	knownClasses: Set<string>
+): Map<string, TLDefaultColorStyle> {
+	const result = new Map<string, TLDefaultColorStyle>()
+
+	const styleMatch = svgString.match(/<style[^>]*>([\s\S]*?)<\/style>/i)
```
mimecuvalo, line 89: in general for these regex, we'll need some comments/examples, otherwise, it's too hard to parse in my brain :)

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+			return 'rectangle'
+	}
+}
+
+export function mapLineTypeToArrowProps(type: number): {
+	dash: TLDefaultDashStyle
+	arrowheadEnd: TLArrowShapeArrowheadStyle
+} {
+	switch (type) {
+		case 0: // SOLID
```
mimecuvalo: let's add a comment pointing to mermaid docs/source code of where this is coming from

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+	return type === 33 || type === 34
+}
+
+// ---------------------------------------------------------------------------
+// CSS style parsing (for linkStyle, classDef, style directives)
+// ---------------------------------------------------------------------------
+
+// tldraw light-mode solid hex values (hard-coded to avoid importing the full palette)
+const TLDRAW_PALETTE: [TLDefaultColorStyle, number, number, number][] = [
+	['black', 0x1d, 0x1d, 0x1d],
```
mimecuvalo: sorry, what is this doing?

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
mimecuvalo: oh i see below maybe you're using nearestTldrawColor - how does this play with parseClassDefFills->hexToTldrawColor though - that doesn't use nearestTldrawColor?

packages/mermaid/src/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+]
+
+function parseHexToRgb(hex: string): [number, number, number] | null {
+	const h = hex.replace(/^#/, '')
+	if (h.length === 3) {
+		return [parseInt(h[0] + h[0], 16), parseInt(h[1] + h[1], 16), parseInt(h[2] + h[2], 16)]
+	}
+	if (h.length === 6) {
+		return [parseInt(h.slice(0, 2), 16), parseInt(h.slice(2, 4), 16), parseInt(h.slice(4, 6), 16)]
+	}
```
mimecuvalo, line 260: hex codes can be 8 in length too heh
#RRGGBBAA

packages/mermaid/src/mappings.ts
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * Parse a Mermaid CSS style array (from FlowEdge.style, FlowVertex.styles, etc.)
+ * and return tldraw-compatible overrides. Only returns fields that were found.
+ */
+export function parseCssStyles(styles: string[] | undefined): ParsedCssOverrides {
+	if (!styles || styles.length === 0) return {}
+
+	const props = new Map<string, string>()
+	for (const entry of styles) {
+		for (const part of entry.split(';')) {
```
mimecuvalo, line 307: are these CSS styles we're parsing? or something else?
what's a "Mermaid CSS style array"?

packages/mermaid/src/svgParsing.ts
```diff
@@ -0,0 +1,262 @@
...
+	width: number
+	height: number
+}
+
+export type NodeIdParser = (domId: string) => string
+export type EdgeIdParser = (dataId: string) => { start: string; end: string } | null
+
+function parseTranslate(attr: string | null): Vec2 {
+	if (!attr) return { x: 0, y: 0 }
+	const m = attr.match(/translate\(\s*([\d.e+-]+)[,\s]+([\d.e+-]+)\s*\)/)
```
mimecuvalo, line 26: regex explanation/example pls

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/createMermaidDiagram.ts
```diff
@@ -0,0 +1,75 @@
...
+): Promise<string | undefined> {
+	const mermaid = (await import('mermaid')).default
+
+	const parsedResult = await mermaid.parse(text, { suppressErrors: true })
+	if (!parsedResult) {
+		throw new MermaidDiagramError('not a mermaid diagram', 'parse')
+	}
+
+	try {
+		const parsedSvg = (await mermaid.render(`mermaid-${Date.now()}`, text)).svg
```
mimecuvalo: not sure the order here but looks like mermaid adds this to the DOM which changes the tldraw layout for a split second while pasting :P

Image

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
kaneel: ooooh right, yes I think that at this point we just return indeed

packages/mermaid/src/createMermaidDiagram.ts
kaneel: yeah I need to get through the bottom of this but I'm afraid it means they are just not exporting this in the future versions, eventually… at some point? 
Concerning unless we pint the version.

BTW, excalidraw also relies on this to some degrees.

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
kaneel: Actually I think we could rely on human readable enums for these.

packages/mermaid/src/mappings.ts
kaneel: good 👁️ here! updated simply as it is because we never care of the AA and A values

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+				// Only log unexpected errors; MermaidDiagramError is the normal "not mermaid" path
+				if (!(e instanceof MermaidDiagramError)) {
+					console.error(e)
+				}
+
+				await defaultHandleExternalTextContent(editor, content)
+			}
+		})
+	}, [editor, addToast, unsupportedTitle, unsupportedDescription])
```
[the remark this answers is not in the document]
kaneel: this seems to be automatically deregistering itself

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/mappings.ts
kaneel: Yep, nothing to see here, just an agent that made things complicated 😢

apps/dotcom/client/src/components/SneakyMermaidHandler/utils/createMermaidDiagram.ts
kaneel: wow, you're right, it's especially happening quickly for state diagram and I never registered this thing before! It's now rendering offscreen

packages/mermaid/src/mappings.ts
kaneel: mermaid has its own way to store some styles on edges, we're reading through those. weirdly, there's classdef -> that goes in a <style> tag, but arrows stuff is stores on edges and we read from there. It's a bit messy

steveruizok:
If it helps, @davidkpiano put me onto this project he worked on: https://github.com/statelyai/graph

davidkpiano:
> If it helps, @davidkpiano put me onto this project he worked on: https://github.com/statelyai/graph

Thanks Steve! 

For reference, statelyai/graph provides data structures & algos for working with a common intermediate graph layer that enables converting to/from many different graph formats (including Mermaid) with full round-trip fidelity.

I have a POC I can push up of converting Mermaid to tldraw if you'd find that useful. The library supports *all* Mermaid diagram types that can be converted into a nodes/edges graph (so, most of them)

mimecuvalo:
> I have a POC I can push up of converting Mermaid to tldraw if you'd find that useful. The library supports all Mermaid diagram types that can be converted into a nodes/edges graph (so, most of them)

sure! that POC would be super useful to see!

kaneel:
> I have a POC I can push up of converting Mermaid to tldraw if you'd find that useful. The library supports _all_ Mermaid diagram types that can be converted into a nodes/edges graph (so, most of them)

It would be interesting. 
Looking forward to see how this behaves with all the graphs! I think @angrycaptain19 would love this :)

davidkpiano:
> > I have a POC I can push up of converting Mermaid to tldraw if you'd find that useful. The library supports _all_ Mermaid diagram types that can be converted into a nodes/edges graph (so, most of them)
> 
> It would be interesting. Looking forward to see how this behaves with all the graphs! I think @angrycaptain19 would love this :)

Here you go: https://github.com/davidkpiano/tldraw-mermaid

Also I ensured that the Mermaid tests in `@statelyai/graph` meet OR exceed the coverage in Mermaid etc. for full parity: https://github.com/statelyai/graph/commit/e8c2b99a5a12539e3a2522415271962def6c3035 (and others)

kaneel:
this got a bit mad, spiky, and it's good… we're closing and moving to https://github.com/tldraw/tldraw/pull/8194 :)

PR #8013, opened 2026-02-18
refactor(sync): incremental Pierre persistence with per-record files

ds300:
In order to enable granular git-level diffs in Pierre history, this PR rewrites Pierre persistence to store each tldraw record as an individual file and use the `TLSyncStorage` interface for incremental updates.

**Pierre storage format**: Instead of a single `snapshot.json` blob, each commit now contains `meta.json` (documentClock, schema only) and individual `records/{id}.json` files. Each record file stores only the record state (no lastChangedClock). Tombstone metadata is not stored in Pierre; the file-per-record model represents deletes by removing files. This gives meaningful per-record diffs in Pierre's commit history.

**Incremental persistence**: Uses `storage.getChangesSince(pierreDocClock)` to compute the diff since the last Pierre commit. Only changed records are written per commit. For empty repos, `documentClock=-1` triggers returning all records. On `wipeAll`, stale record files in Pierre are cleaned up via `repo.listFiles()`. Optimistic concurrency via `expectedHeadSha` with retry on CAS conflict.

**getChangesSince(wipeAll)**: When `wipeAll` is true (e.g. empty repo, `sinceClock -1`), storage was still including tombstones in `diff.deletes`, which caused the first Pierre commit to call `deletePath` on non-existent files and could duplicate deletes with the cleanup loop. InMemorySyncStorage and SQLiteSyncStorage now omit `diff.deletes` when `wipeAll`; the durable object applies `diff.deletes` only when `headSha && !wipeAll`.

**Restore fix**: Pierre restore no longer calls `repo.restoreCommit()` (which failed when restoring to HEAD). Instead it reconstructs the snapshot from the Pierre archive via `reconstructSnapshotFromPierre` and loads it into storage with `loadSnapshotIntoStorage`, so the next persist cycle writes the restored state as a new forward commit.

**Snapshot reconstruction**: New `pierreSnapshot.ts` uses `modern-tar`'s `createTarDecoder()` to stream the gzip-compressed tar archive from Pierre. The pipeline is `getArchiveStream` → DecompressionStream('gzip') → createTarDecoder(); entries are consumed with a reader loop. Record files are parsed as state only; reconstructed documents use `lastChangedClock: 0` (loadSnapshotIntoStorage only uses state). History snapshot and restore both call `reconstructSnapshotFromPierre` and return or load the resulting `RoomSnapshot`.

**Other**: Added missing `assert` import in `TLSyncStorage.ts`. `syncPierreState` uses `listCommits({ limit: 1 })` and `getFileStream('meta.json')` to sync HEAD and documentClock. Pierre repo IDs changed from `.../snapshots/...` to `.../files/...`. Upgraded `@pierre/storage` to `^1.1.0` and added `modern-tar`. Client history snapshot page uses `TlaFileError` in the error boundary and drops the unused error prop from the main component.

### Change type

- [x] `improvement`

### Test plan

1. Open a tldraw file with Pierre enabled
2. Make edits and verify Pierre commits contain individual `records/*.json` entries (state only)
3. View Pierre history — verify snapshot loading works
4. Restore from Pierre history — verify the canvas updates and a new Pierre commit is created
5. Verify incremental commits only touch changed records
6. New file (empty Pierre repo): first persist should succeed without deletePath errors

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Internal improvement to Pierre version history storage format for better granularity.

### DIFF
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -1197,11 +1197,17 @@ export class TLDrawDurableObject extends DurableObject {
 					commitBuilder.addFileFromString(`records/${id}.json`, JSON.stringify(state))
 				}
 
-				for (const id of diff.deletes) {
-					commitBuilder.deletePath(`records/${id}.json`)
+				// Only apply diff.deletes when we have a parent commit and we're not in wipeAll.
+				// - Empty repo (no headSha): those paths don't exist in Pierre; deletePath would fail.
+				// - wipeAll with existing repo: the cleanup loop below already deletes any file not in
+				//   putIds, so applying diff.deletes here would duplicate deletePath for the same file.
+				if (headSha && !changes.wipeAll) {
+					for (const id of diff.deletes) {
+						commitBuilder.deletePath(`records/${id}.json`)
+					}
 				}
 
-				// On wipeAll, pruned tombstones won't appear in diff.deletes,
+				// On wipeAll with an existing repo, pruned tombstones may not appear in diff.deletes,
 				// so scan Pierre for stale record files and remove them.
 				if (changes.wipeAll && headSha) {
 					const putIds = new Set(Object.keys(diff.puts))
--- a/packages/sync-core/src/lib/InMemorySyncStorage.ts
+++ b/packages/sync-core/src/lib/InMemorySyncStorage.ts
@@ -376,10 +376,12 @@ class InMemorySyncStorageTransaction<R extends UnknownRecord>
 				diff.puts[doc.state.id] = doc.state as R
 			}
 		}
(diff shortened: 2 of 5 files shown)


apps/dotcom/sync-worker/src/utils/pierreSnapshot.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Streaming tar entry reader backed by a single growable buffer.
+ * Incoming stream chunks are appended in-place; the buffer compacts
+ * (via copyWithin) only when space runs low, and grows only when
+ * compaction isn't enough. File data is decoded to a string before
+ * yielding, so the buffer is safe to reuse immediately — no slice needed.
+ */
+async function* readTarEntries(
```
mimecuvalo: should we be reimplementing .tar logic? 😅 
is there no maintained / native library for this?

PR #8014, opened 2026-02-18
feat(examples): add related examples linking in the examples app

### LINKED ISSUE #7962: Add related examples linking in the examples app
## Problem

There's no way to navigate laterally between related concepts. If a user is reading "Custom shape" they might want "Custom shape with handles" or "Shape with bindings" next, but nothing connects them. The only navigation is the flat category list in the sidebar.

## Proposal

Add a way to link related examples so users can follow learning paths:

1. **Add a `related` field to README frontmatter:**
   ```yaml
   ---
   title: Custom shape
   component: ./CustomShapeExample.tsx
   related: [custom-shape-with-handles, bindings]
   ---
   ```

2. **Render "See also" links** in the description panel or below the example

3. **Auto-suggest related examples** based on shared keywords as a fallback for examples without manual `related` entries

## Use cases

- Users exploring a feature area who want to go deeper
- Learning paths: getting-started → basic customization → advanced patterns
- Discoverability: users find examples they didn't know existed

AniKrisn:
## Summary
- Add a `related` field to each example README frontmatter linking to other examples by slug
- Render a "See also" section with pill-style links in each example's info dialog
- Add category descriptions with hover tooltips in the sidebar
- All 163 examples now have curated related links for lateral navigation

Closes #7962

## Test plan
- [ ] Run `yarn dev` and open the examples app
- [ ] Click the info button on any example and verify "See also" links appear
- [ ] Click a related link and verify it navigates to the correct example
- [ ] Hover over a category name in the sidebar and verify the tooltip appears

### DIFF
--- a/apps/examples/src/examples/collaboration/sync-custom-user/README.md
+++ b/apps/examples/src/examples/collaboration/sync-custom-user/README.md
@@ -15,7 +15,8 @@ keywords:
     presence,
   ]
 multiplayer: true
-related: []
+related:
+  [sync-demo, sync-custom-presence, sync-custom-people-menu, user-presence, sync-private-content]
 ---
 
 Integrate your own user data into tldraw sync.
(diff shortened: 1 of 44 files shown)


steveruizok:
The examples work here feels a little scattershot and undesigned. I'd try to keep the changes pretty small or else think about the examples app's design again from first principles. The related examples being shown in the hard-to-find info screen isn't very useful; the info next to the categories isn't very useful either. This might be a good opportunity to do some horizontal research on other websites like this one, more reading than writing.

Issue #7629, opened 2026-01-06
Consolidate or share setup between Checks workflow jobs


steveruizok:
## Problem

The `checks.yml` workflow has two parallel jobs (`test` and `build`) that each independently run the full setup action including yarn install. This means we're paying the ~1 minute setup cost twice per workflow run.

### Current workflow structure

**checks.yml:**
- `test` job: checkout → setup → constraints, dedupe, check-packages, typecheck, bundle-size, PR template, lint, API check, i18n check, test
- `build` job: checkout → setup → build → pack-tarball

Both jobs run `./.github/actions/setup` which includes:
1. Enable corepack (`npm i -g corepack`)
2. Setup Node.js with yarn cache
3. `yarn install --immutable`

### Affected workflows

- `.github/workflows/checks.yml` - Two parallel jobs duplicating setup
- `.github/workflows/playwright-examples.yml` - Single job (no duplication here)
- `.github/workflows/playwright-dotcom.yml` - Single job (no duplication here)

### Option 1: Merge jobs into a single job
**Pros:** Simplest change, eliminates duplicate setup entirely
**Cons:** Loses parallelism, total wall-clock time may increase

### Option 2: Create a dedicated setup job with artifact upload
**Pros:** Preserves parallelism, single setup cost
**Cons:** Adds complexity, artifact upload/download has its own overhead (~20-30s each), may not be faster for small node_modules

### Option 3: Improve yarn cache hit rates
**Pros:** Minimal changes, benefits all workflows
**Cons:** May not fully solve the problem if cache misses occur

## Suggested evaluation approach

1. Measure current workflow timing to understand the actual cost
2. Compare yarn install time with cache hit vs cache miss
3. Consider if the `test` and `build` jobs have enough parallelizable work to justify separate jobs
4. Evaluate whether the jobs could be merged without significantly increasing total wall-clock time

steveruizok:
## Research findings

### Current workflow timing analysis

Analyzed run [#20753215760](https://github.com/tldraw/tldraw/actions/runs/20753215760) (from merge queue):

**Setup step durations:**
- `Build all projects` job: Setup took **51 seconds** (15:34:34 → 15:35:25)
- `Tests & checks` job: Setup took **61 seconds** (15:35:16 → 15:36:17)

**Total setup overhead: ~112 seconds** (1 min 52 sec) for duplicated work.

**Job execution times:**
[...]

PR #7875, opened 2026-02-10
feat(infra): migrate asset serving to tldrawusercontent.com

MitjaBezensek:
In order to simplify our asset infrastructure, decrease security risks, and improve performance, this PR introduces a new Cloudflare Worker on `tldrawusercontent.com` that unifies asset uploading and serving into a single domain. Previously, three coordinated workers handled user assets: `asset-upload-worker` (uploads), `image-resize-worker` (resizing proxy), and `sync-worker` (authenticated uploads/downloads). The new `tldrawusercontent-worker` replaces all three for asset operations.

### What changed

**New worker: `tldrawusercontent-worker`** (`apps/dotcom/tldrawusercontent-worker/`)
- Handles both `POST /:objectName` (upload to R2) and `GET /:objectName` (serve from R2)
- Lives on `tldrawusercontent.com` (production), `staging.tldrawusercontent.com` (staging), `*.tldraw.workers.dev` (preview)

**Image resizing now uses Cloudflare zone-level Image Transformations** instead of a proxy worker. The client constructs `/cdn-cgi/image/` URLs that Cloudflare's edge intercepts and transforms before they hit the worker. This eliminates an entire network hop (no more proxy worker fetching from the origin).

**Resizing logic:**
- Images >= 1.5 MB are resized based on zoom level, DPR, and network speed
- Smaller images skip resize but still get `format=auto` for format conversion (avif/webp)
- SVGs are routed through `/cdn-cgi/image/` for [Cloudflare's SVG sanitization](https://developers.cloudflare.com/images/transform-images/) (strips scripts, hyperlinks, cross-origin refs) but are not resized
- Animated images and videos bypass transforms entirely
- In dev/preview, `/cdn-cgi/image/` is unavailable (reserved Cloudflare path), so raw file URLs are served

**Auth handling:**
- Anonymous uploads go directly to `tldrawusercontent-worker` with no auth
- App file uploads (associated with a tldraw file) also go to `tldrawusercontent-worker`, but the client passes the Clerk auth token as a `Authorization: Bearer <token>` header (needed because `tldrawusercontent.com` is a different domain, so Clerk cookies aren't sent)
- After a successful upload, `tldrawusercontent-worker` calls `sync-worker` via a Cloudflare service binding (`POST /app/associate-asset/:objectName`) to validate auth, check the file exists, and queue the DB association
- Service bindings are internal worker-to-worker calls — no public network hop

**Lazy URL migration:**
[...]

PR #8015, opened 2026-02-18
perf(ci): parallelize deploy-dotcom staging phases

mimecuvalo:
In order to reduce staging deploy duration, this PR runs independent deploy-dotcom phases in parallel while keeping multiplayer and Zero-dependent work ordered. Relates to #7629.

### Change type

- [x] `other`

### Test plan

1. Trigger a staging deploy and compare the deploy timing breakdown before/after this change.
2. Verify the deploy still completes successfully with multiplayer, non-multiplayer workers, and Vercel deployment all succeeding.
3. Confirm the timing logs show parallelized phases (`cloudflare deploy dry run` tasks and post-multiplayer parallel deploys).

- [ ] Unit tests
- [ ] End to end tests

---

> [!NOTE]
> **Medium Risk**
> Touches CI/deploy orchestration and introduces parallelism plus a new dependency scheduler; failures or mis-specified dependencies could break staging/production deploy ordering even though runtime behavior of services is unchanged.
> 
> **Overview**
> Speeds up `deploy-dotcom.ts` by parallelizing independent staging/preflight work while preserving ordering for Cloudflare workers that depend on `multiplayer`.
> 
> The script now models worker deploys as a dependency graph (`workerDeployments`) and runs them via `deployWorkersInDependencyOrder`, enabling parallel dry-run deploys for ready workers and replacing the previous hard-coded per-worker sequence. It also runs `uploadSourceMaps` and asset cache `coalesceWithPreviousAssets` in parallel after the dotcom build, and enforces “workers first, then Vercel deploy” in the real deploy phase.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1eaaa28080666f8ea14313586243beba16e9ef64. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


internal/scripts/deploy-dotcom.ts
```diff
@@ -264,45 +264,59 @@ async function main() {
...
+					async () =>
+						await Promise.all([
+							withTiming('step: deploy asset uploader worker', () =>
+								discord.step('deploying asset uploader to cloudflare', async () => {
+									await withTiming('worker deploy: asset-upload', () =>
+										deployAssetUploadWorker({ dryRun: false })
+									)
+								})
+							),
+							withTiming('step: deploy tldrawusercontent worker', () =>
```
MitjaBezensek: I'm not sure this will work. When I worked on the tldrawusercontent worker I had to make sure I deployed them in the correct order. Sync worker first, then tldrawuser content worker. 
See this commit https://github.com/tldraw/tldraw/pull/7875/changes/dca69689ddab6e66f238bb7de449d4088186fe96. 

The reason for this is that we use service bindings and I think those require correct ordering. Think the image worker also has a binding. 

So maybe we need to write down the dependencies and then at least paralelize once those are satisfied?

mimecuvalo: makes sense, i figured there was some dependency i was missing here.
i'll pull you aside so we figure out the dep graph

mimecuvalo: PTAL

MitjaBezensek (commented):
I guess discord step numbers might get out of order?

Before we also aborted the run on the first failure, I guess this will continue deploying? Unsure if that can cause some issues? (we didn't handle this gracefully before either) 🤷‍♂️

internal/scripts/deploy-dotcom.ts
```diff
@@ -236,73 +321,48 @@ async function main() {
...
-				return await withTiming('vercel deploy --prebuilt', () => deploySpa())
-			})
+			await Promise.all([
+				withTiming(
+					'step: deploy dotcom app to vercel',
+					async () =>
+						await discord.step('deploying dotcom app to vercel', async () => {
+							return await withTiming('vercel deploy --prebuilt', () => deploySpa())
+						})
+				),
```
ds300: the vercel deploy should happen after the workers have been successfully deployed to avoid version skew issues.

Issue #5859, opened 2025-04-09
Feature: custom assets


SomeHats:
We allow custom shapes, and custom bindings. We should create a custom assets API too.

This work would involve:

1. Defining an "asset".
   1. Custom assets means introducing `AssetUtil`s. There's some research involved in defining the boundaries of an asset and what the responsibility of the asset util would be.
   2. The original idea of assets were generic resources that could be shared by several shapes. They're mostly used as (and all of the custom-asset-like customer requests are for) *files*.
   3. Bookmarks are an outlier to the way most users seem to conceptualise assets - we should consider splitting bookmarks out of assets
2. \[maybe, depending on the outcome of 1\] split bookmarks into a new record type, or inline them into bookmark shapes
3. Adapting the custom record setup we have in tlschema for shapes and bindings to work for assets as well
4. Introducing a new `AssetUtil` base class, plus asset utils for our existing asset types.
5. Make sure that all code that checks `asset.type` now calls out to an asset util instead
6. Add an `assetUtils` prop everywhere that currently accepts a `shapeUtils` and `bindingUtils` prop

mimecuvalo:
#keep

Issue #5235, opened 2025-01-20
[Example Request]: How to handle asset uploads that aren't media


MattCrouch:
### What's the example?

We're looking into ways that we can support sharing files on a shared canvas. For example, someone could drag a PDF on to the canvas, we would upload that file and provide a shape to preview and download it.

I understand we should be using an asset store to catch these files, which would trigger uploads for images and videos, but it's not clear how we would extend this to support other types of files.

I have tried re-implementing [`registerExternalContentHandler`](https://github.com/tldraw/tldraw/blob/80294e641ad673cd53fc946176ef89d459663a67/packages/tldraw/src/lib/defaultExternalContentHandlers.ts#L226) and [`registerExternalAssetHandler`](https://github.com/tldraw/tldraw/blob/80294e641ad673cd53fc946176ef89d459663a67/packages/tldraw/src/lib/defaultExternalContentHandlers.ts#L67) to allow these different types, but looks like there's checks in place to limit assets to `image`, `video` or `bookmark` types and I feel like I'm going about it the wrong way. 

The [PDF Editor](https://tldraw.dev/examples/use-cases/pdf-editor) example is close to what we'd need in the long run, but that example appears to create image assets/shapes from the PDF on the fly. We would need to upload the file and provide a link for other users to download it alongside something like that. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mimecuvalo:
At the moment, we don't support custom assets. It's something that we're thinking about generalizing this year though so it could be more extensible in this manner.

It would be gross, but you could "override" one of the existing assets and use the `meta` prop to add any relevant data 😬 But I'm not saying you should do that for real :P

MattCrouch:
No problem! Our requirements had slightly changed since this request and we found a way to work around it without the need of.

We settled on detecting the drop of the file on the page externally, uploading the file separately and creating a link to that file on the canvas.

It works for what we need for now, but I'll keep a lookout for ways we can improve it in the future. Thanks!

mimecuvalo:
I'm actually gonna reopen this because I think we do want to enable this in general for our community :)
But I'm glad you were able to move on without this!

mimecuvalo:
@darrinm this is the case currently. I think you're referring to the PRs (https://github.com/tldraw/tldraw/pull/4290 and https://github.com/tldraw/tldraw/pull/4413 ) that do both. They introduce the assets and a new Audio shape, but that doesn't have to be the case fwiw.

[...]

PR #8031, opened 2026-02-19
feature: AssetUtil for extensible asset types

mimecuvalo:
fixes https://github.com/tldraw/tldraw/issues/5859
fixes https://github.com/tldraw/tldraw/issues/5235

## Summary

- Adds an `AssetUtil` base class following the `ShapeUtil` / `BindingUtil` patterns, making the asset system extensible
- Previously, assets were a hardcoded union of `TLImageAsset | TLVideoAsset | TLBookmarkAsset` — this PR allows custom asset types to be registered
- Scope: `AssetUtil` handles **type-specific behavior** only (MIME types, file → asset metadata, asset → shape creation). `TLAssetStore` keeps `upload`/`resolve`/`remove` as cross-cutting concerns

## What changed

**`@tldraw/tlschema`** — extensible asset schema
- Per-key prop validators (`imageAssetProps`, `videoAssetProps`, `bookmarkAssetProps`)
- `createAssetRecordType()` + `defaultAssetSchemas`, matching shape/binding patterns
- `assets` parameter on `createTLSchema()`

**`@tldraw/editor`** — `AssetUtil` base class
- `AssetUtil` with `configure()`, `getDefaultProps()`, `getSupportedMimeTypes()`, `getAssetFromFile()`, `createShape()`
- `Editor` registers asset utils; exposes `getAssetUtil()`, `hasAssetUtil()`, `getAssetUtilForMimeType()`
- Wired through `TldrawEditor`, `createTLStore`

**`@tldraw/tldraw`** — default implementations + refactoring
- `ImageAssetUtil`, `VideoAssetUtil`, `BookmarkAssetUtil`
- `Tldraw` component accepts `assetUtils` prop (merged with defaults via `mergeArraysAndReplaceDefaults`)
- `defaultExternalContentHandlers` delegates to asset utils instead of hardcoded switch/MIME-type checks

### Change type

- [x] `feature`

## Test plan

- [x] `yarn typecheck` passes
- [x] `packages/tldraw` tests pass (129 files, 2028 tests)
- [x] `packages/editor` tests pass (41 files, 764 tests)
- [x] `packages/tlschema` tests pass (9 files, 442 tests)
- [ ] Manual: `yarn dev` → drop images/videos, verify they work as before
- [ ] Manual: verify custom asset util registration works

### API changes

- Adds `AssetUtil` and a family of classes/helpers to assist with creating custom assets.
- Breaking: `notifyIfFileNotAllowed` signature changed from (file, options) to (editor, file, options)
- Breaking: `getAssetInfo` signature changed from (file, options, assetId?) to (editor, file, assetId?) and now returns `TLAsset | null` instead of throwing
- Breaking: `getMediaAssetInfoPartial` removed. Use `AssetUtil.getAssetFromFile` instead
- Breaking: `assetValidator` is removed. Use `imageAssetValidator`/`videoAssetValidator`/`bookmarkAssetValidator`
- New: `TLGlobalAssetPropsMap` interface for type-safe custom asset registration (mirrors `TLGlobalShapePropsMap`)
- New: `TLIndexedAssets` type
- New: `TLAsset` now accepts a type parameter for narrowing: `TLAsset<'image'>`

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -94,6 +94,7 @@ import { TLTheme } from '@tldraw/tlschema';
 import { TLThemeColors } from '@tldraw/tlschema';
 import { TLThemeId } from '@tldraw/tlschema';
 import { TLThemes } from '@tldraw/tlschema';
+import { TLUnknownAsset } from '@tldraw/tlschema';
 import { TLUnknownBinding } from '@tldraw/tlschema';
 import { TLUnknownShape } from '@tldraw/tlschema';
 import { TLUser } from '@tldraw/tlschema';
@@ -143,6 +144,26 @@ export class Arc2d extends Geometry2d {
 
 // @public
 export function areAnglesCompatible(a: number, b: number): boolean;
+
+// @public
+export abstract class AssetUtil<Asset extends TLAsset = TLAsset> {
+    constructor(editor: Editor);
+    acceptsMimeType(mimeType: string): boolean;
+    static configure<T extends TLAssetUtilConstructor<any, any>>(this: T, options: T extends new (...args: any[]) => {
+        options: infer Options;
+    } ? Partial<Options> : never): T;
+    // (undocumented)
+    editor: Editor;
+    getAssetFromFile(_file: File, _assetId: TLAssetId): Promise<Asset | null>;
+    abstract getDefaultProps(): Asset['props'];
+    getSupportedMimeTypes(): readonly string[];
+    // (undocumented)
+    static migrations?: LegacyMigrations | MigrationSequence | TLPropsMigrations;
+    options: {};
+    // (undocumented)
+    static props?: RecordProps<TLUnknownAsset>;
+    static type: string;
+}
 
 // @public (undocumented)
 export function average(A: VecLike, B: VecLike): string;
@@ -851,7 +872,7 @@ export class EdgeScrollManager {
 
 // @public (undocumented)
 export class Editor extends EventEmitter<TLEventMap> {
-    constructor({ store, user, shapeUtils, bindingUtils, tools, getContainer, cameraOptions, initialState, autoFocus, options: _options, textOptions: _textOptions, getShapeVisibility, colorScheme, fontAssetUrls, themes, initialTheme }: TLEditorOptions);
+    constructor({ store, user, shapeUtils, bindingUtils, assetUtils: assetUtilConstructors, tools, getContainer, cameraOptions, initialState, autoFocus, options: _options, textOptions: _textOptions, getShapeVisibility, colorScheme, fontAssetUrls, themes, initialTheme }: TLEditorOptions);
     alignShapes(shapes: TLShape[] | TLShapeId[], operation: 'bottom' | 'center-horizontal' | 'center-vertical' | 'left' | 'right' | 'top'): this;
     animateShape(partial: null | TLShapePartial | undefined, opts?: TLCameraMoveOptions): this;
     animateShapes(partials: (null | TLShapePartial | undefined)[], opts?: TLCameraMoveOptions): this;
@@ -862,6 +883,9 @@ export class Editor extends EventEmitter<TLEventMap> {
         tags?: Record<string, boolean | number | string>;
         willCrashApp: boolean;
     }): this;
+    assetUtils: {
+        readonly [K in string]?: AssetUtil<TLAsset>;
+    };
     bail(): this;
     bailToMark(id: string): this;
     bindingUtils: {
@@ -1217,6 +1241,12 @@ export class Editor extends EventEmitter<TLEventMap> {
     getAsset<T extends TLAsset>(asset: T | T['id']): T | undefined;
     getAssetForExternalContent(info: TLExternalAsset): Promise<TLAsset | undefined>;
     getAssets(): (TLBookmarkAsset | TLImageAsset | TLVideoAsset)[];
+    getAssetUtil<S extends TLAsset>(asset: {
+        type: S['type'];
+    } | S): AssetUtil<S>;
+    // (undocumented)
+    getAssetUtil(type: string): AssetUtil;
+    getAssetUtilForMimeType(mimeType: string): AssetUtil | null;
     getAttributionDisplayName(userId: null | string): null | string;
     getAttributionUser(userId: null | string): null | TLUser;
     getAttributionUserId(): null | string;
@@ -1365,6 +1395,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     getShapeUtil<S extends TLShape>(shape: S | S['type'] | TLShapePartial<S>): ShapeUtil<S>;
     // (undocumented)
     getShapeUtil<T extends ShapeUtil>(type: T extends ShapeUtil<infer R> ? R['type'] : string): T;
+    getShapeUtilForAssetType(assetType: string): ShapeUtil | undefined;
     getSharedOpacity(): SharedStyle<number>;
     getSharedStyles(): ReadonlySharedStyleMap;
     // (undocumented)
@@ -1404,6 +1435,9 @@ export class Editor extends EventEmitter<TLEventMap> {
         select: boolean;
     }>): this;
     hasAncestor(shape: TLShape | TLShapeId | undefined, ancestorId: TLShapeId): boolean;
+    hasAssetUtil(arg: {
+        type: string;
+    } | string): boolean;
     // (undocumented)
     hasExternalAssetHandler(type: TLExternalAsset['type']): boolean;
     hasShapeUtil(shape: TLShape | TLShapePartial<TLShape>): boolean;
@@ -2865,6 +2899,7 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
     static configure<T extends TLShapeUtilConstructor<any, any>>(this: T, options: T extends new (...args: any[]) => {
         options: infer Options;
     } ? Partial<Options> : never): T;
+    createShapeForAsset?(asset: TLAsset, position: VecModel): null | TLShapePartial;
     // (undocumented)
     editor: Editor;
     // @internal (undocumented)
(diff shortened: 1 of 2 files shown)


mimecuvalo:
@ds300 and @MitjaBezensek i think this is ready to start looking at.

@ds300 one question here is: I think this needs a better migration for adding this to the store. can you assist with that in the branch?

apps/examples/src/examples/data/assets/custom-asset-type/CustomAssetTypeExample.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+		return {
+			id,
+			type: FILE_ASSET_TYPE,
+			typeName: 'asset',
+			props: {
+				name: file.name,
+				size: file.size,
+				mimeType: file.type,
+				src: '',
```
MitjaBezensek: Is this `src` even used? Do we needed? Feels a bit confusing.

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -121,6 +122,11 @@ export interface TldrawEditorBaseProps {
 	 */
 	bindingUtils?: readonly TLAnyBindingUtilConstructor[]
 
+	/**
+	 * An array of asset utils to use in the editor.
+	 */
+	assetUtils?: readonly TLAnyAssetUtilConstructor[]
```
MitjaBezensek, line 138: I guess we want to allow users not to have any assets?

packages/tldraw/src/lib/assets/ImageAssetUtil.ts
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+
+	override getSupportedMimeTypes(): readonly string[] {
+		return DEFAULT_SUPPORTED_IMAGE_TYPES
+	}
+
+	override async getAssetFromFile(file: File, assetId: TLAssetId): Promise<TLImageAsset | null> {
+		const hash = getHashForBuffer(await file.arrayBuffer())
+		const id = assetId ?? (`asset:${hash}` as TLAssetId)
```
MitjaBezensek: Looks like `assetId` is always provided, do we need to hash? We could also hash inline to make sure we don't do it if we have the `assetId`.

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -948,27 +914,15 @@ export function notifyIfFileNotAllowed(file: File, options: TLDefaultExternalCon
...
-	const assetInfo = await getMediaAssetInfoPartial(
-		file,
-		assetId,
-		isImageType,
-		isVideoType,
-		maxImageDimension
-	)
+
+	const assetUtil = editor.getAssetUtilForMimeType(file.type)
+	assert(assetUtil, `No asset util found for MIME type "${file.type}"`)
```
MitjaBezensek: Seems like we sometimes show toasts and sometimes we just assert. Like would a failure here result in a shown toast? Would be nice to unify the failure cases.

apps/examples/src/examples/data/assets/custom-asset-type/CustomAssetTypeExample.tsx
```diff
@@ -0,0 +1,315 @@
...
+			},
+			meta: {},
+		}
+	}
+
+	// [6]
+	override createShape(asset: TLFileAsset, position: VecModel): TLShapePartial {
```
MitjaBezensek: Unsure how I feel about this. Like the asset has the knowledge on how to create shapes based on that asset. It feels like it should be the other way around. Maybe shape utils instead should have an api like `canCreateShapeForAssetType` and / or `createShapeForAssetType` or something like that?

apps/examples/src/examples/data/assets/custom-asset-type/CustomAssetTypeExample.tsx
mimecuvalo: yeah, it's used below `reader.readAsDataURL(file)` - we store the base64 as the url in this case (instead of using the asset store)

packages/editor/src/lib/TldrawEditor.tsx
mimecuvalo: yeah, i think it can be optional. i dunno, do you think it should be required?

mimecuvalo:
pushed up changes @MitjaBezensek 

i'm gonna take a look at the inverting suggestion you had regarding `canCreateShapeForAssetType` next

packages/tldraw/src/lib/assets/ImageAssetUtil.ts
```diff
@@ -36,18 +39,15 @@ export class ImageAssetUtil extends AssetUtil<TLImageAsset> {
...
 	override getSupportedMimeTypes(): readonly string[] {
-		return DEFAULT_SUPPORTED_IMAGE_TYPES
+		return this.options.supportedMimeTypes ?? DEFAULT_SUPPORTED_IMAGE_TYPES
 	}
 
 	override async getAssetFromFile(file: File, assetId: TLAssetId): Promise<TLImageAsset | null> {
-		const hash = getHashForBuffer(await file.arrayBuffer())
-		const id = assetId ?? (`asset:${hash}` as TLAssetId)
-
 		const size = await MediaHelpers.getImageSize(file)
```
MitjaBezensek: Looks like we never use the `pixelRatio` from `getImageSize`? Looks like a regression?

MitjaBezensek:
Looks good. Worth [checking up this](https://github.com/tldraw/tldraw/pull/8031#discussion_r3020803776), then good to go I think.

packages/tldraw/src/lib/assets/VideoAssetUtil.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	override async getAssetFromFile(file: File, assetId: TLAssetId): Promise<TLVideoAsset | null> {
+		let fileType = file.type
+		if (fileType === 'video/quicktime') {
+			// hack to make .mov videos work
+			fileType = 'video/mp4'
+		}
+
+		const size = await MediaHelpers.getVideoSize(file)
```
MitjaBezensek: We used to pass in `doc` here in the past, but we no longer do. Feels like it might cause some issues in iframe envs?

packages/tlschema/src/assets/TLImageAsset.ts
```diff
@@ -51,18 +51,21 @@ export type TLImageAsset = TLBaseAsset<
  *
  * @public
```
MitjaBezensek: We have double @public in some cases (TLBookmarkAsset.ts, TLImageAsset.ts, TLVideoAsset.ts).

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -891,18 +808,16 @@ async function maybeSanitizeSvgFile(file: File): Promise<File | null> {
...
-		maxAssetSize = DEFAULT_MAX_ASSET_SIZE,
-		toasts,
-		msg,
-	} = options
-	const isImageType = acceptedImageMimeTypes.includes(file.type)
-	const isVideoType = acceptedVideoMimeTypes.includes(file.type)
-
-	if (!isImageType && !isVideoType) {
+export function notifyIfFileNotAllowed(
+	editor: Editor,
```
MitjaBezensek, line 817: Missing editor param in docs.

packages/tlschema/src/assets/TLBaseAsset.ts
```diff
@@ -107,21 +107,24 @@ export const assetIdValidator = idValidator<TLAssetId>('asset')
  *
  * @public
  */
-export function createAssetValidator<Type extends string, Props extends JsonObject>(
+export function createAssetValidator<
```
MitjaBezensek, line 104: docs seem outdated.

packages/tldraw/src/lib/assets/ImageAssetUtil.ts
mimecuvalo: oh man, good catch on this 😓 i added some tests to help w not regressing here

packages/tldraw/src/lib/assets/VideoAssetUtil.ts
mimecuvalo: fixed but makes me think we should make `doc` required 🤔

packages/tlschema/src/index.ts
```diff
@@ -82,11 +87,17 @@ export { scribbleValidator, TL_SCRIBBLE_STATES, type TLScribble } from './misc/T
 export {
 	assetMigrations,
 	AssetRecordType,
-	assetValidator,
```
MitjaBezensek, line 85: We should mention this breaking change? Also saw this mentioned in a couple of tsdocs.

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -160,10 +159,10 @@ export async function defaultHandleExternalFileAsset(
 /** @public */
 export async function defaultHandleExternalFileReplaceContent(
 	editor: Editor,
-	{ file, shapeId, isImage }: TLFileReplaceExternalContent,
```
MitjaBezensek, line 163: Looks like `isImage` is no longer used? Can we deprecate?

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -886,23 +803,32 @@ async function maybeSanitizeSvgFile(file: File): Promise<File | null> {
...
+		acceptedVideoMimeTypes,
 		toasts,
 		msg,
 	} = options
-	const isImageType = acceptedImageMimeTypes.includes(file.type)
-	const isVideoType = acceptedVideoMimeTypes.includes(file.type)
 
-	if (!isImageType && !isVideoType) {
+	// If explicit MIME type lists are provided, use them; otherwise check asset utils
+	const isFileTypeAllowed =
```
MitjaBezensek, line 829: The outer ternary checks acceptedImageMimeTypes || acceptedVideoMimeTypes. If either is provided, it enters the explicit-list branch. But then ?? fills in the default list for whichever one **wasn't** provided. Feels surprising? Like you'd need to provide `acceptedVideoMimeTypes: []` if you don't want to support videos?

mimecuvalo: good shout 👍

PR #8036, opened 2026-02-19
feat(i18n): migrate the SDK to ICU MessageFormat and react-intl

### LINKED ISSUE #7823: Modernize internationalization to use ICU MessageFormat consistently
## Problem

tldraw currently has two separate internationalization systems with different approaches:

1. **SDK (`packages/tldraw`)**: A custom simple key-value system using `{variable}` placeholders with no support for pluralization, gender, or ICU MessageFormat features
2. **Dotcom (`apps/dotcom`)**: Uses `react-intl` with `@formatjs/cli` and proper ICU MessageFormat support

The SDK's custom i18n solution lacks ICU MessageFormat features that are standard in modern i18n:

- **No pluralization**: Can't handle "1 shape" vs "5 shapes" correctly across languages
- **No select/gender**: Can't handle gender-specific translations
- **No ordinals**: Can't handle "1st", "2nd", "3rd" correctly
- **No nested messages**: Can't compose complex messages

### SDK translation format (simple placeholders only)
```json
{
  "a11y.multiple-shapes": "{num} shapes selected",
  "a11y.shape-index": "{num} of {total}"
}
```

These can't properly pluralize across languages - translators can only provide one form.

### Dotcom uses proper ICU (via react-intl)
```json
{
  "translation": "{total, plural, one {Uploading .tldr file…} other {Uploading {uploaded} of {total} .tldr files…}}"
}
```

### Option A: Adopt react-intl for the SDK
Migrate the SDK to use `react-intl` and `@formatjs/cli` like dotcom already does. This would:
- Unify the i18n approach across the codebase
- Provide full ICU MessageFormat support
- Leverage existing tooling (Lokalise integration, formatjs CLI)
- Be a breaking change for SDK users who provide custom translations

### Option B: Adapt custom system to support ICU
Enhance the existing SDK i18n system to parse ICU MessageFormat:
- Keep the simple `useTranslation()` API
- Add ICU message parsing under the hood
- Maintain backwards compatibility with simple `{var}` syntax
- Could use `intl-messageformat` (already a dependency in dotcom)

## Considerations

- **Breaking changes**: SDK users who provide custom translations would need to update their translation files
- **Bundle size**: ICU parsing adds ~10-15KB to the bundle
- **Translation workflow**: Need to update Lokalise export settings and ensure translator tooling supports ICU

## References

- [ICU MessageFormat guide](https://lokalise.com/blog/complete-guide-to-icu-message-format/)
- [FormatJS/react-intl](https://formatjs.io/)
- Current SDK i18n: `packages/tldraw/src/lib/ui/hooks/useTranslation/`
- Current dotcom i18n: `apps/dotcom/client/src/tla/utils/i18n.tsx`

mimecuvalo:
> [!NOTE]
> Draft. Foundation is in and verified; the bulk string migration hasn't started. See **Where this is** below for the exact state.

fixes #7823

Rebased onto `main` and rewritten. The earlier proof of concept is gone — it was 480 commits behind, and its `useTldrawI18n` called `createIntl({ messages: {} })`, so ICU messages only ever rendered their `defaultMessage` and never actually translated.

## What this does

Moves the SDK from a flat key→string dictionary to ICU MessageFormat, then adds react-intl authoring on top so strings can live next to the code that renders them.

**ICU formatting** (`dfdef2e7f1`). `useTranslation`'s `msg` formats through formatjs and takes an optional `values` argument. Interpolation was hand-rolled before — `msg('a11y.multiple-shapes').replace('{num}', …)` — and all 8 such call sites are gone.

`packages/commenting` faked pluralization with five keys and a regex, which is wrong for Arabic (6 plural forms), Polish (4) and Russian (3). Those five collapse into one `comments.reacted`, and `comments.replies-one` folds into `comments.replies`.

**react-intl authoring** (`e76ae95d2a`). `TldrawUiTranslationProvider` now also provides an `IntlProvider`, built from the catalog it already computed — one `IntlShape` behind both APIs, so an app's `overrides.translations` reaches `<F>` and `useMsg` as well as `msg()`. Adds `F`, `useMsg`, `defineMessages`, the two pseudo-locales, and `yarn i18n-extract`.

**Dead keys removed** (`6cca4563e5`). 71 keys had no call site anywhere. Mostly leftovers from when tldraw.com's UI lived in the SDK.

**`@tldraw/editor` made translatable** (`091b9ab168`). It had no way to translate at all, so every string it rendered was English by construction. It gets `TLI18nMessage`, `defineMessages` and `translateMessage` — and no i18n dependency, because `formatjs extract` matches the call by name rather than by import. 16 strings in `DefaultErrorFallback`, the commenting unknown-author fallback, and the embed dialog's "Learn more" are now translated.

## Not breaking

All four extensibility mechanisms still work, which is the reason ids are stable names rather than content hashes:

- `overrides.translations` — merged into react-intl's `messages`
- an app's own keys — an unknown id still comes back as-is
- `assetUrls.translations` — same flat JSON shape, ICU inside
- `useTranslation()` — `values` is an additive second argument

A public scan found this matters: BigBlueButton self-hosts the catalog with 16 keys of its own, and one repo uses `overrides.translations`.

Consumer-supplied strings aren't necessarily valid ICU, so a malformed one falls back to the raw string rather than taking down the UI. There's a test for it.

## Where this is

```
catalog keys         : 482
already migrated     : 23 (5%)
still only in catalog: 459
text drift           : 0
```

`yarn i18n-extract` reports that and fails CI on text that disagrees between code and `main.json`. It refuses to write the catalog while any key is still only in `main.json`, because extraction is destructive: a key nobody moved into code is absent from the output, and dropping it silently would take its 49 translations with it.

Remaining, roughly in order:

1. Enumerate the 6 dynamic-key families (115 keys) — extraction can't see `msg(`geo-style.${geo}`)`, so each needs an explicit `defineMessages` map. This blocks the catalog flip.
2. Migrate the ~344 remaining literal keys, package by package.
3. Flip `main.json` to extraction output.
4. Lokalise: switch the SDK project to ICU plural format, and delete the 71 keys there — a download reintroduces them otherwise.
5. Docs and the `custom-language-translations` example.

## Notes for review

- `label:` arrives as data in ~110 places, so it stays a key — but sourced from a declared descriptor (`label={messages.url.id}`), which keeps it extracted and non-breaking. See `EditLinkDialog`.
- The locale messages for the collapsed plural families were composed from the strings already in each locale, so no language regresses while waiting on Lokalise. They use `=1` where the old code branched on `x === 1`: CLDR's `one` also matches 21 and 31 in Russian, and those strings have a hardcoded "1". Verified with 539 assertions across 52 locales, zero mismatches.
- Descriptors are `TLI18nMessage`, not react-intl's, so which library formats a message stays out of the public contract. The d.ts rollup caught the leak when it wasn't.
- Bundle cost: 15.6 KB gzipped. SDK and dotcom are both on react-intl `^8.1.3`, so they dedupe.
- `packages/editor` has no `<F>` — it has no react-intl — so its strings use `translateMessage`. Happy to add an editor-side `<F>` if the idiom should match.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn api-check` 20/20
- [x] `yarn lint-current` clean
- [x] editor 1239, tldraw 3089, commenting 379 tests passing
- [x] `yarn i18n-extract --check` exits 1 on drift, 0 clean
- [ ] visual check of the error screen and comment reactions in a non-English locale


frolic:
I merged main, which now uses pnpm's isolated linker. `@formatjs/cli` is only linked into `packages/tldraw`, so `i18n-extract` now runs the formatjs command from there.

Sent from my aiPhone

PR #8038, opened 2026-02-19
fix(dotcom): show error page when Clerk auth service is unavailable

steveruizok:
In order to show users a helpful error page instead of a blank white screen when Clerk is down or rate-limited (429), this PR adds error boundaries and a loading timeout to the Clerk authentication flow.

When Clerk is unavailable, `auth.isLoaded` stays `false` indefinitely and tldraw.com renders a blank white page. This PR catches that case with a 10-second timeout and also wraps the Clerk provider tree in error boundaries.

### Change type

- [x] `bugfix`

### Test plan

1. Start `yarn dev-app`
2. Block Clerk's script loading (e.g. via DevTools network blocking for `clerk.tldraw.com`)
3. Reload the page
4. After ~10 seconds, the error page appears with "Unable to connect" and a "Refresh" button
5. Unblock Clerk and click "Refresh" — app loads normally

### Release notes

- Show an error page instead of a blank screen when the authentication service is unavailable

### Code changes

| Apps            | +116 / -31  |
| Automated files | +27 / -0    |


apps/dotcom/client/src/components/ErrorPage/ErrorPage.tsx
```diff
@@ -64,3 +64,27 @@ export function ErrorPage({
...
+	static getDerivedStateFromError() {
+		return { hasError: true }
+	}
+
+	render() {
+		if (this.state.hasError) {
+			return (
+				<ErrorPage
+					messages={this.props.messages}
+					cta={<button onClick={() => window.location.reload()}>Refresh</button>}
```
mimecuvalo: i guess as long as you're passing `messages` you could also pass one for "Refresh"

Issue #7998, opened 2026-02-18
Dynamic size is ignored when dragging shapes from the toolbar


steveruizok:
When `isDynamicSizeMode` is enabled, shapes created by dragging from the toolbar always get `scale: 1` instead of being scaled relative to the current zoom level.

## Steps to reproduce

1. Enable dynamic size mode (user preference `isDynamicSizeMode: true`)
2. Zoom in or out to a non-1x zoom level (e.g. 2x)
3. Drag a shape (geo, arrow, note, text, line, etc.) from the toolbar onto the canvas
4. Observe the shape's `scale` property

## Expected behavior

The shape should be created with `scale: 1 / zoomLevel` (e.g. `scale: 0.5` at 2x zoom), matching the behavior when creating shapes by clicking/dragging on the canvas with a tool.

## Actual behavior

The shape is created with `scale: 1` regardless of the zoom level and dynamic size setting. This means shapes dragged from the toolbar appear at a different visual size than shapes created with the tool directly.

## Context

Each shape tool's `Pointing` state correctly applies the dynamic scale:

```ts
scale: this.editor.user.getIsDynamicResizeMode() ? 1 / this.editor.getZoomLevel() : 1
```

But the toolbar drag `createShape` callbacks in `useTools.tsx` never set the `scale` prop, so it defaults to `1`.

steveruizok:
## Research findings

### Root cause

The toolbar drag-and-drop `createShape` callbacks in `packages/tldraw/src/lib/ui/hooks/useTools.tsx` (lines ~156-276) never set the `scale` prop when creating shapes. Every callback passes hardcoded dimensions without considering dynamic size mode.

For example, the geo toolbar drag callback (line ~158):
```ts
createShape: (id) =>
    editor.createShape({ id, type: 'geo', props: { w: 200, h: 200, geo } }),
```

Compare to the normal geo tool in `packages/tldraw/src/lib/shapes/geo/toolStates/Pointing.ts` (line 37):
```ts
scale: this.editor.user.getIsDynamicResizeMode() ? 1 / this.editor.getZoomLevel() : 1,
```

### Affected toolbar drag callbacks

All shape-creating toolbar drag callbacks are missing the `scale` prop:

| Shape | File | Line |
|-------|------|------|
| geo | `packages/tldraw/src/lib/ui/hooks/useTools.tsx` | ~158 |
| arrow | `packages/tldraw/src/lib/ui/hooks/useTools.tsx` | ~175 |
[...]

PR #8042, opened 2026-02-19
fix(tldraw): apply dynamic sizing when dragging shapes from toolbar

AniKrisn:
In order to make shapes dragged from the toolbar respect dynamic sizing (matching the behavior of click-to-create), this PR applies the same `scale` calculation to each toolbar `onDragStart` callback and centralizes the scale factor logic into a new `editor.getResizeScaleFactor()` method.

When dynamic size mode is enabled, shapes created by clicking with a tool respect the zoom level via the `scale` prop, but shapes dragged out of the toolbar did not. This adds the same scaling to toolbar drag-create for geo, arrow, line, frame, text, and note tools, and refactors all Pointing states to use the shared helper.

Closes https://github.com/tldraw/tldraw/issues/7998

### Change type

- [x] `bugfix`

### Test plan

1. Enable dynamic size mode in user preferences
2. Zoom in/out to a non-1x zoom level
3. Drag each shape type (geo, arrow, line, frame, text, note) out of the toolbar
4. Verify the created shapes appear at a consistent screen size regardless of zoom, matching the behavior of click-to-create

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Added `Editor.getResizeScaleFactor()` — returns `1 / zoomLevel` when dynamic sizing is enabled, `1` otherwise

### Release notes

- Fix shapes dragged from the toolbar not respecting dynamic size mode


mimecuvalo (commented):
this is starting to get repetitive across the codebase. let's create a function called `editor.getResizeScaleFactor()` please! and if you don't mind, let's change the other places you see this in the codebase 🙏

PR #8050, opened 2026-02-20
fix(editor): clear rich text editor when editing shape is deleted

### LINKED ISSUE #8049: Link editor stays open if text shape is deleted while editing link
## Description

In multiplayer, if user A is editing a link in a text shape's rich text toolbar and user B deletes that text shape, the link editor stays open with no way to dismiss it.

https://github.com/user-attachments/assets/820c51f7-c912-4e14-ac42-6c548c84724c

## Steps to reproduce

1. Open the same room in two browser tabs
2. In tab A, create a text shape with some text
3. In tab A, select the text and open the link editor in the rich text toolbar
4. In tab B, delete the text shape
5. In tab A, the link editor remains visible and stuck

## Expected behavior

The link editor should close when the underlying shape is deleted.

## Location

`packages/tldraw/src/lib/ui/components/Toolbar/LinkEditor.tsx`

MitjaBezensek:
Closes #8049

See the issue for how to repro.

When a remote user deletes a shape that's being edited, `cleanupInstancePageState` sets `editingShapeId = null` directly. The `defaultSideEffects` handler then transitions to `select.idle`, which fires `EditingShape.onExit()`. But `onExit()` had an early return (`if (!editingShapeId) return`) that bailed because the id was already cleared — so `setEditingShape(null)` never ran and `_currentRichTextEditor` was never cleared. The rich text toolbar (and link editor) stayed visible.

### Fix

- **Remove the early return** in `EditingShape.onExit()`. The guard originally protected code that used `editingShapeId` after calling `setEditingShape(null)`, but that logic has since moved into `setEditingShape` itself. Now `setEditingShape(null)` always runs on exit, clearing `_currentRichTextEditor`.
- **Remove the `_currentRichTextEditor.set(null)`** added in the initial commit inside `cleanupInstancePageState` — it's no longer needed since the state machine exit path handles it.

### Change type

- [x] `bugfix`

### Test plan

1. Open the same tldraw room in two browser tabs
2. In tab A, create a text shape, enter edit mode, select text, open the link editor
3. In tab B, delete the text shape
4. Verify tab A's link editor disappears

### Release notes

- Fix rich text toolbar staying open when editing shape is deleted by another user

---

> [!NOTE]
> **Low Risk**
> Small, localized change to `EditingShape.onExit()` cleanup logic; primary risk is unintended tool switching when leaving edit mode, but the new `hadEditingShape` guard narrows that behavior.
> 
> **Overview**
> Fixes a state-machine edge case where leaving `select.editing_shape` could skip cleanup if `editingShapeId` had already been cleared externally (e.g. by collaborative deletion).
> 
> `EditingShape.onExit()` now **always** calls `editor.setEditingShape(null)` and only reverts the tool-lock-to-`text` flow when an editing shape actually existed, preventing stale rich-text UI (toolbar/link editor) from remaining visible.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 30daedb304f192bab52a74930c46448b2f3b8e07. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
@@ -49,11 +49,12 @@ export class EditingShape extends StateNode {
 	}
 
 	override onExit() {
+		const hadEditingShape = !!this.editor.getEditingShapeId()
 		this.editor.setEditingShape(null)
 
 		updateHoveredShapeId.cancel()
 
-		if (this.info.isCreatingTextWhileToolLocked) {
+		if (this.info.isCreatingTextWhileToolLocked && hadEditingShape) {
 			this.parent.setCurrentToolIdMask(undefined)
 			this.editor.setCurrentTool('text', {})
 		}


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -447,6 +447,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 			if (prevPageState.editingShapeId && shapesNoLongerInPage.has(prevPageState.editingShapeId)) {
 				if (!nextPageState) nextPageState = { ...prevPageState }
 				nextPageState.editingShapeId = null
+				this._currentRichTextEditor.set(null)
```
mimecuvalo: good catch! we should make this logic change, however, in `setEditingShape`, not in this side effect path.
i think this used to maybe work but we added `if (!this.canEditShape(id)) return this` in setEditingShape so maybe the cleanup code isn't running.
there's some lifecycle stuff there (onEditEnd/onEditStart) that i think might get skipped (potentially) if we don't make the logic check there

MitjaBezensek: Addressed this in the latest commit. To me that looked the right place to do the fix that will trigger the correct clean up, but I may be missing something so I'd love if you could take another look 🙌

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -49,10 +49,6 @@ export class EditingShape extends StateNode {
 	}
 
 	override onExit() {
-		const { editingShapeId } = this.editor.getCurrentPageState()
-		if (!editingShapeId) return
-
-		// Clear the editing shape
 		this.editor.setEditingShape(null)
 
 		updateHoveredShapeId.cancel()
```
MitjaBezensek, line 55: It also felt like this additional cleanup should also run?

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -49,10 +49,6 @@ export class EditingShape extends StateNode {
 	}
 
 	override onExit() {
-		const { editingShapeId } = this.editor.getCurrentPageState()
-		if (!editingShapeId) return
-
-		// Clear the editing shape
```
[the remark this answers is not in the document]
MitjaBezensek: Yeah, think that's correct. We do want to run all the cleanup logic for cases when we are no longer editing a shape.

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
mimecuvalo: agreed

packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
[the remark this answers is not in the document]
mimecuvalo: yeah i think this is good

PR #8054, opened 2026-02-20
fix(eraser): erase shapes inside groups when starting drag from within group bounds

### LINKED ISSUE #7684: Eraser fails to erase shapes when starting drag from inside a group
## Description

When using the eraser tool on a shape that's inside a group, if you start the drag from inside the group's bounds and then drag (transitioning from "pointing" to "erasing" state), the erasing doesn't work - the shapes don't get erased.

## Steps to reproduce

1. Create a group containing one or more shapes
2. Select the eraser tool
3. Click inside the group bounds, on a shape within the group
4. Drag (moving the pointer while holding down) to trigger the "erasing" state
5. Release the pointer

**Expected**: The shape(s) touched by the eraser should be deleted  
**Actual**: Nothing gets erased

## Workaround

- Clicking without dragging (just a point-and-click) works correctly
- Starting the eraser drag from outside the group bounds also works

## Initial analysis

The issue appears to be in `Erasing.ts` where the `excludedShapeIds` set is populated. When the eraser starts inside a group, the group gets added to `excludedShapeIds` (lines 36-44). Later, when shapes are hit-tested, `getOutermostSelectableShape()` returns the group as the outermost selectable shape, but then the group is filtered out because it's in the excluded set (line 157).

The logic intends to prevent accidentally erasing a group/frame when you start erasing from inside it, but it inadvertently prevents erasing the children too since they resolve to the group via `getOutermostSelectableShape()`.

**Relevant files:**
- `packages/tldraw/src/lib/tools/EraserTool/childStates/Erasing.ts`
- `packages/tldraw/src/lib/tools/EraserTool/childStates/Pointing.ts`

AniKrisn:
In order to fix erasing shapes inside groups when the eraser drag starts from within the group bounds, this PR updates the hit-resolution logic in the Erasing state. Previously, `getOutermostSelectableShape` resolved every hit shape to the group, which was then filtered out by the exclusion set — so nothing got erased. Now, when the outermost shape is excluded (i.e. the group we started inside), we re-resolve using a filter that skips excluded shapes, targeting the child shape or nested group instead.

Closes #7684

### Change type

- [x] `bugfix`

### Test plan

1. Create a group with shapes inside it
2. Use the eraser tool, start dragging from inside the group bounds over a child shape
3. Verify the child shape is erased and the group is not
4. Verify erasing from outside the group still erases the entire group
5. Test with nested groups (Group A > Group B > Shape) — erasing from inside Group A should erase Group B or its children, not Group A

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix eraser not erasing shapes when starting drag from inside a group's bounds

---

> [!NOTE]
> **Medium Risk**
> Changes shape hit-resolution in the eraser’s deletion path, which could cause unintended shapes (or groups) to be erased if the exclusion logic misfires. Scope is limited to eraser interactions and is covered by added unit tests.
> 
> **Overview**
> Fixes eraser hit resolution for grouped shapes when the drag starts *inside* a group’s bounds: if the initially-resolved outermost selectable shape is excluded (e.g. the hovered group), the eraser now re-resolves to the nearest non-excluded ancestor/shape so child shapes (or nested groups) can be erased instead of erasing nothing.
> 
> Updates tests to cover the new group behavior, including starting inside a group (erase children), starting outside a group (erase the whole group), and the case where the group is excluded at drag start but the drag later crosses its children.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8ebd32352f793533a2f89f6a06fa64be767d1ee2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/test/EraserTool.test.ts
+++ b/packages/tldraw/src/test/EraserTool.test.ts
@@ -385,10 +385,10 @@ describe('When clicking and dragging', () => {
 		editor.groupShapes([ids.box2, ids.box3], { groupId: ids.group1 })
 		editor.setCurrentTool('eraser')
 		editor.expectToBeIn('eraser.idle')
-		// Start outside the group
-		editor.pointerDown(-100, -100)
+		// Start outside the group (avoiding box1 which is at 0,0)
+		editor.pointerDown(150, -100)
 		// Drag over box2 (child of group)
-		editor.pointerMove(125, 125)
+		editor.pointerMove(150, 125)
 		vi.advanceTimersByTime(16)
 		expect(editor.getInstanceState().scribbles.length).toBe(1)
 		// The whole group should be erased since we started outside


steveruizok:
Hey! This is Claude posting on behalf of Steve.

I made a couple of changes to this PR:

**PR description** — Updated the body to follow our PR template (description paragraph, change type, test plan, release notes section) and removed the AI attribution.

**Tests** — Pushed a commit to update the tests to match the new behavior:

- **Updated the existing "Excludes a group" test** — the old test expected that dragging from inside a group to `(0, 0)` would only erase `box1`, but since the drag path now crosses through `box2` (a child of the excluded group), the new behavior correctly includes `box2` in the erasing set too. The test also now accounts for the group auto-dissolving when it's left with only one child.

- **Added two new tests:**
  - "Erases child shapes when starting drag inside a group and dragging over them" — tests the core fix from this PR (the original bug in #7684)
  - "Erases the whole group when starting drag outside of it" — confirms existing behavior is preserved

packages/tldraw/src/test/EraserTool.test.ts
```diff
@@ -338,21 +338,65 @@ describe('When clicking and dragging', () => {
...
+		editor.setCurrentTool('eraser')
+		editor.expectToBeIn('eraser.idle')
+		// Start outside the group
+		editor.pointerDown(-100, -100)
+		// Drag over box2 (child of group)
+		editor.pointerMove(125, 125)
+		vi.advanceTimersByTime(16)
+		expect(editor.getInstanceState().scribbles.length).toBe(1)
+		// The whole group should be erased since we started outside
+		expect(editor.getErasingShapeIds()).toEqual([ids.group1])
```
kaneel: @steveruizok you're going over box1 here I think so you may want to do something reversed, starting at `100, -100` till box2 only? I would not edit the expectations. 

Or edit the expectations but group box1 and box2 together? Which means changing the distance of the erasing

kaneel (approved):
well LGTM!

PR #8057, opened 2026-02-20
refactor(editor): simplify putContentOntoCurrentPage paste-parent and positioning logic

steveruizok:
In order to make the paste logic in `putContentOntoCurrentPage` easier to reason about and extend, this PR replaces the convoluted paste-parent selection and positioning code with clearer, well-separated logic paths.

### Change type

- [x] `improvement`

### Test plan

1. Copy shapes and paste them — they should appear at the viewport center or in-place if already visible
2. Select a frame, copy shapes, paste — they should land inside the selected frame
3. Right-click inside a frame and paste — shapes should land inside that frame at the cursor position
4. Copy a frame that is the same size as a selected frame and paste — the duplicate should not be reparented into the original (the "duplicating a frame" case)
5. Paste shapes that overlap multiple frames — only shapes whose center is inside a frame should be auto-reparented
6. Copy shapes inside a group, paste — they should land in the correct parent

- [x] Unit tests

### Release notes

- Simplify paste-parent selection to use `canReceiveNewChildrenOfType` instead of frame-specific checks
- Prevent edge-only overlap from auto-reparenting pasted shapes into adjacent frames

---

> [!NOTE]
> **Medium Risk**
> Changes core paste/reparenting and positioning logic, which can subtly affect many copy/paste scenarios across containers and frames despite test coverage.
> 
> **Overview**
> Refactors `Editor.putContentOntoCurrentPage` paste behavior into clearer paths for *paste-at-cursor*, *standard paste*, and *preserve-position* pastes, using `canReceiveNewChildrenOfType` (with optional chaining) to choose a valid container from the selection/ancestors and avoiding pasting content into itself (duplicate-container case).
> 
> Updates positioning rules to center pastes within the chosen parent, keep coordinates for `preservePosition`, otherwise choose between original vs viewport center based on viewport overlap, and skips the `updateShapes` pass when the computed offset is zero.
> 
> Tightens page-level auto-reparenting into frames by filtering out source shapes and requiring the pasted shape’s bounds center to be inside the candidate parent (preventing reparenting on edge-only overlap); updates `paste.test.ts` expectations accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c7e57192837506317a0f117407a267cce0a73353. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -133,7 +133,7 @@ import {
 } from '../utils/deepLinks'
 import { getIncrementedName } from '../utils/getIncrementedName'
 import { getReorderingShapesChanges } from '../utils/reorderShapes'
-import { getDroppedShapesToNewParents } from '../utils/reparenting'
+import { getDroppedShapesToNewParents, kickoutOccludedShapes } from '../utils/reparenting'
 import { TLTextOptions, TiptapEditor } from '../utils/richText'
 import { applyRotationToSnapshotShapes, getRotationSnapshot } from '../utils/rotation'
 import { BindingOnDeleteOptions, BindingUtil } from './bindings/BindingUtil'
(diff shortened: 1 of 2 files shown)


kaneel:
Worth a shot addressing some of the claude comments but as far as I'm concerned, this is working for the "paste at pointer" feature and the test are passing (in comparison with my "fix" which broke a lot of unrelated tests)

steveruizok:
need to fire kick out children when pasting into a paste parent

PR #8059, opened 2026-02-20
perf(editor): experiment with uWrap for canvas-based text measurement

mimecuvalo:
This PR experiments with using [uWrap](https://github.com/leeoniya/uWrap) for canvas-based text measurement instead of DOM-based measurement in the `TextManager`. The goal is to evaluate whether bypassing DOM layout/reflow can speed up text measurement, which is a hot path when creating and resizing text shapes.

**How it works:**
- uWrap uses Canvas2D `measureText()` to build character width tables and kerning pairs, then does word wrapping algorithmically — no DOM layout needed
- `measureText()` now routes to the canvas path by default for plain text
- `measureHtml()` extracts plain text from simple HTML (no formatting tags) and uses the canvas path; falls back to DOM for rich text with bold/italic/etc.
- `measureTextSpans()` still uses DOM since it needs per-character positioning
- uWrap instances and resolved font families are cached per font configuration

**Known limitations:**
- uWrap currently works most accurately with Latin charsets
- Canvas2D text metrics may differ slightly from DOM layout (sub-pixel differences, different shaping)
- Does not handle `\r\n` line breaks (uWrap limitation)
- Height calculation uses `lineCount * fontSize * lineHeight` which may not perfectly match CSS box model

**More notes:**
- there would still be work to be done here - def. cases where you're typing and it doesn't do the right thing, reflow wise
- if we did this for real we would handroll our own bespoke uwrap-like logic. i wouldn't want to rely on uwrap explicitly.
- this is meant to be helping with the hot path of a plain-text shape. this does introduce a separate logic path but it would prbly be worth it if this is speedy enough.
- we would need to more rigorously investigate how well non-Latin characters work w this

### Change type

- [x] `improvement`

### Test plan

1. Open the examples app (`yarn dev`)
2. Create text shapes, note shapes, geo shapes with labels
3. Compare text sizing/wrapping behavior with `main` branch
4. Look for cases where text measurement is visibly different

- [ ] Unit tests
- [ ] End to end tests

### DIFF
--- a/packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
+++ b/packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
@@ -161,12 +161,15 @@ export class TextManager {
 		const resolvedFamily = this.resolveFontFamily(opts.fontFamily)
 		const key = `${opts.fontStyle} ${opts.fontWeight} ${opts.fontSize}px/${opts.lineHeight} ${resolvedFamily}`
 
+		// Keep the shared context in sync with this wrapper's font.
+		// uWrap may use ctx.measureText for chars outside its initial LUT.
+		this.canvasCtx.font = `${opts.fontStyle} ${opts.fontWeight} ${opts.fontSize}px ${resolvedFamily}`
+
 		const cached = this.uwrapCache.get(key)
 		if (cached) return cached
 
 		// Configure the canvas context with the font
 		const ctx = this.canvasCtx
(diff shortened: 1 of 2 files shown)


mimecuvalo:
pushed up fixes to this.
- there would still be work to be done here - def. cases where you're typing and it doesn't do the right thing, reflow wise
- if we did this for real we would handroll our own bespoke uwrap-like logic. i wouldn't want to rely on uwrap explicitly.
- this is meant to be helping with the hot path of a plain-text shape. this does introduce a separate logic path but it would prbly be worth it if this is speedy enough.
- we would need to more rigorously investigate how well non-Latin characters work w this

PR #8070, opened 2026-02-23
feat(sync): support WebSocket hibernation in tlsync

### LINKED ISSUE #7754: Support WebSocket hibernation in sync-cloudflare template
## Problem statement

The current `sync-cloudflare` template and tldraw-sync libraries use the traditional WebSocket pattern with Cloudflare Durable Objects:

```typescript
serverWebSocket.accept()
socket.addEventListener('message', handleSocketMessage)
socket.addEventListener('close', handleSocketClose)
```

This means the Durable Object must remain in memory for the entire duration of all WebSocket connections, even when idle. Cloudflare's [WebSocket hibernation API](https://developers.cloudflare.com/durable-objects/examples/websocket-hibernation-server/) allows DOs to be evicted from memory during inactivity while keeping connections open, significantly reducing duration charges.

## Proposed solution

Update the `sync-cloudflare` template to use the hibernation APIs:

1. Replace `serverWebSocket.accept()` with `this.ctx.acceptWebSocket(serverWebSocket)`
2. Implement the `webSocketMessage(ws, message)` handler method on the DO class
3. Implement the `webSocketClose(ws, code, reason, wasClean)` handler method
4. Use `ws.serializeAttachment()`/`ws.deserializeAttachment()` for session state
5. Recover connections on DO reconstruction via `this.ctx.getWebSockets()`

**Good news:** The `TLSocketRoom` class already supports this pattern. Its `handleSocketMessage`, `handleSocketError`, and `handleSocketClose` methods can be called directly without event listeners, and the `WebSocketMinimal` interface marks `addEventListener` as optional.

### Template (primary target)
- `templates/sync-cloudflare/worker/TldrawDurableObject.ts` - Main changes needed

### Sync-core (already compatible)
- `packages/sync-core/src/lib/TLSocketRoom.ts` - Already supports manual message routing
- `packages/sync-core/src/lib/ServerSocketAdapter.ts` - `WebSocketMinimal` interface is hibernation-ready

### Production (optional follow-up)
- `apps/dotcom/sync-worker/src/TLDrawDurableObject.ts` - Could benefit from hibernation

## Example implementation sketch

```typescript
export class TldrawDurableObject extends DurableObject {
  private room: TLSocketRoom<...>
  
  async handleConnect(request: IRequest) {
    const sessionId = nanoid()
    const { 0: clientWebSocket, 1: serverWebSocket } = new WebSocketPair()
    
    // Use hibernation API instead of accept()
    this.ctx.acceptWebSocket(serverWebSocket)
    serverWebSocket.serializeAttachment({ sessionId })
    
    // Register with TLSocketRoom (no event listeners needed)
    this.room.handleSocketConnect({ sessionId, socket: serverWebSocket })
    
    return new Response(null, { status: 101, webSocket: clientWebSocket })
  }
  
  // Hibernation handler - called when message arrives
  async webSocketMessage(ws: WebSocket, message: string | ArrayBuffer) {
    const { sessionId } = ws.deserializeAttachment()
    this.room.handleSocketMessage(sessionId, message)
  }
  
  // Hibernation handler - called when connection closes
  async webSocketClose(ws: WebSocket, code: number, reason: string, wasClean: boolean) {
    const { sessionId } = ws.deserializeAttachment()
    this.room.handleSocketClose(sessionId)
  }
}
```

## Benefits

- **Reduced costs** - DOs are evicted during idle periods, reducing duration charges
- **Better scalability** - Memory freed up when connections are quiet
- **Automatic ping/pong** - Can use `setWebSocketAutoResponse()` for keep-alive without waking the DO

## Considerations

- Need to handle DO reconstruction and session recovery
- Hibernation only works for server-side WebSockets (which is our use case)
- Should maintain backward compatibility for non-Cloudflare deployments

ds300:
In order to allow Cloudflare Durable Objects to hibernate while maintaining active WebSocket connections, this PR adds session resume/snapshot APIs to `TLSocketRoom`, replaces the periodic session pruning interval with on-demand pruning (with optional interval when a finite idle timeout is set), and updates both the `sync-cloudflare` template and the dotcom sync-worker to use the WebSocket Hibernation API.

Closes #7754

### Changes

**`@tldraw/sync-core`**

- Added `handleSocketResume()` to `TLSocketRoom` — restores a session from a `SessionStateSnapshot` without re-attaching event listeners (hibernation environments deliver events via class methods)
- Added `getSessionSnapshot()` to `TLSocketRoom` — captures a connected session's state (schema, presence, flags) for persistence in WebSocket attachments; strips scribbles/chatMessage/meta from tldraw instance_presence to keep attachments small
- Added optional `onSessionSnapshot` to `TLSocketRoom` — room debounces (5s) after message activity and invokes the callback with the session snapshot for persisting to WebSocket attachments; clears pending snapshot when the session closes
- Exported `SessionStateSnapshot` interface
- Added `handleResumedSession()` to `TLSyncRoom` — lower-level session restoration that reconstructs the session state machine
- Replaced always-on interval pruning with: throttled on-demand pruning on message, one-shot `setTimeout` for `AwaitingRemoval` cleanup, and an optional interval that runs only when `clientTimeout` is finite and > 0 (so `clientTimeout: Infinity` keeps the DO hibernation-friendly)
- Added `clientTimeout` option to `TLSyncRoom` constructor (set to `Infinity` in Cloudflare deployments since the platform handles keep-alive)

**`sync-cloudflare` template**

- Switched to `DurableObject` base class with `webSocketMessage`, `webSocketClose`, `webSocketError` handlers for hibernation
- Added `setWebSocketAutoResponse` to handle ping/pong at the platform layer without waking the DO
- On wake-up, reconstructs `TLSocketRoom` by iterating `ctx.getWebSockets()` and resuming sessions from deserialized attachments
- Stores session snapshots in WebSocket attachments via TLSocketRoom's `onSessionSnapshot` (built-in 5s debounce)
- When the DO is woken by a disconnecting socket, resume the session from the attachment briefly then close so the room can broadcast presence removal to other clients

**`apps/dotcom/sync-worker` (TLDrawDurableObject)**

- WebSocket hibernation: `setWebSocketAutoResponse` for ping/pong, `state.acceptWebSocket()`, `webSocketMessage`/`webSocketClose`/`webSocketError` handlers, attachment with sessionId/meta/isReadonly/snapshot, resume loop on room creation, resume-from-attachment on close for presence cleanup, `onSessionSnapshot` for snapshot storage (built-in debounce)
- Replaced `createPostgresConnectionPool` with `TLPostgresPool` + Kysely so the DO can hibernate (no long-lived pg connections or timers); tear down pool when room becomes empty

### Change type

- [x] `feature`

### Test plan

1. Deploy the `sync-cloudflare` template to Cloudflare
2. Open a document and draw some shapes
3. Wait for the DO to hibernate (no activity for ~10s)
4. Resume drawing — the session should seamlessly continue without reconnection
5. With multiple clients, close one tab — other clients should see that user's presence disappear

- [x] Unit tests

### Release notes

- Add WebSocket hibernation support to `TLSocketRoom` with new `handleSocketResume()` and `getSessionSnapshot()` methods
- Add optional `onSessionSnapshot` to `TLSocketRoom` for debounced snapshot persistence (e.g. WebSocket attachments)
- Replace always-on interval pruning with on-demand pruning and optional interval when `clientTimeout` is finite
- Update `sync-cloudflare` template to use Cloudflare's WebSocket Hibernation API
- Add WebSocket hibernation and `TLPostgresPool` to dotcom sync-worker TLDrawDurableObject

### API changes

- Added `TLSocketRoom.handleSocketResume()` for restoring sessions after hibernation
- Added `TLSocketRoom.getSessionSnapshot()` for capturing session state (presence record may be stripped for size)
- Added `onSessionSnapshot` option to `TLSocketRoom` (callback invoked after 5s idle; room clears pending snapshot on session close)
- Added `SessionStateSnapshot` interface (exported from `@tldraw/sync-core`)
- Added `clientTimeout` option to `TLSyncRoom` constructor
- Changed `TLSyncRoom.pruneSessions` from a plain function to a throttled function (`DebouncedFuncLeading`)


packages/sync-core/src/lib/TLSocketRoom.ts
```diff
@@ -410,9 +479,124 @@ export class TLSocketRoom<R extends UnknownRecord = UnknownRecord, SessionMeta =
...
+			serializedSchema: snapshot.serializedSchema,
+			presenceId: snapshot.presenceId,
+			presenceRecord: snapshot.presenceRecord,
+			requiresLegacyRejection: snapshot.requiresLegacyRejection,
+			supportsStringAppend: snapshot.supportsStringAppend,
+			socket: new ServerSocketAdapter({
+				ws: socket,
+				onBeforeSendMessage: this.opts.onBeforeSendMessage
+					? (message, stringified) =>
+							this.opts.onBeforeSendMessage!({
```
kostyafarber, line 547: NIT: this ternary is a bit confusing, having to use the non-null assertion operator because of the callback, even though we just checked in the ternary that `onBeforeSendMessage` is not null seems a bit circular.

I think `const onBeforeSendMessage = this.opts.onBeforeSendMessage` could make this section easier to read to avoid the `!`

kostyafarber (approved):
I tested this with multiple browsers and windows, checking multiplayer (joining rooms, leaving rooms, user presence) etc in general and with 15s of on activity and all seems OK.

There is a conflict to fix tho.

jezzzm:
this reduced our cloudflare bill by 9x

PR #8081, opened 2026-02-24
feat(shapes): add display values pattern to all shape utils

### LINKED ISSUE #8116: Add resizeToBounds method to Editor
## Problem

There's currently no single method on the Editor to take a set of shapes and transform them so their combined bounding box matches a target bounding box. To achieve this today, an API consumer (or an AI agent) would need to:

1. Compute the common bounds of the shapes
2. Calculate scale factors and translation offsets
3. Call `resizeShape` on each shape individually with the correct `scaleOrigin` and scale values
4. Handle edge cases like rotation, parent transforms, and arrow bindings

This is error-prone and requires deep knowledge of tldraw internals.

## Proposed solution

Add an `editor.resizeToBounds()` method that accepts a set of shapes and a target `Box`, then translates and resizes all the shapes so their common page bounds matches the target bounds afterward.

```ts
editor.resizeToBounds(shapes: TLShapeId[] | TLShape[], bounds: BoxLike): this
```

**Behavior:**
- Compute the current common page bounds of the provided shapes
- Derive scale factors (`targetBounds.w / currentBounds.w`, `targetBounds.h / currentBounds.h`)
- Derive translation offset to move the group origin to `targetBounds.point`
- Apply the transformation to all shapes, using the existing `resizeShape` infrastructure with `scaleOrigin` set to the current common bounds origin
- Respect arrow binding clusters (shapes connected via arrows should move together, matching the pattern in `stretchShapes` and `packShapes`)

**Options to consider:**
- `aspectRatio?: 'lock' | 'free'` — whether to preserve the group's aspect ratio (fitting inside the target bounds) or stretch to fill exactly
- Whether to batch the operation inside a single history entry

## Use cases

- **AI integrations**: An AI agent can lay out or rearrange shapes by specifying target bounding boxes, without needing to compute per-shape transforms
- **Programmatic layouts**: Fitting a set of shapes into a specific region of the canvas (e.g., arranging content into a grid cell or frame)
- **Snap-to-region**: User or plugin selects shapes and snaps them into a predefined area

## Architecture context

The existing multi-shape transformation methods (`stretchShapes`, `packShapes`, `alignShapes`, `distributeShapes`) all follow the same pattern:

1. Normalize input (accept IDs or shapes)
2. Cluster related shapes via arrow bindings
3. Compute common geometry
4. Calculate per-shape transforms
5. Apply via `resizeShape` / `updateShapes`

This new method would fit naturally into that pattern. The `Resizing` state in `SelectTool` already demonstrates applying a uniform scale with a shared `scaleOrigin` across multiple shapes — the logic could be extracted or mirrored.

### Key reference points

- `Editor.resizeShape()` — single shape resize with scale origin support
- `Editor.stretchShapes()` — stretches shapes to fill common bounds along one axis
- `Resizing` child state — handles multi-shape resize during interactive drag
- `collectShapesViaArrowBindings()` — clusters shapes connected by arrows

steveruizok:
Adds the display values pattern to all shape utils, making visual properties overridable via `configure()`. This externalizes hardcoded rendering values (colors, fonts, sizes, opacities) so SDK consumers can customize shape appearance without subclassing.

Also adds `Editor.resizeToBounds()` (closes #8116), a new method that takes a set of shapes and a target bounding box, then translates and resizes all shapes so their combined page bounds matches the target.

### The pattern

Each shape util defines a `DisplayValues` interface describing its visual properties and provides default values via `getDisplayValues()`. Consumers override specific values via `getDisplayValueOverrides()` through `configure()`.

### Defining display values

```tsx
export interface GeoShapeUtilDisplayValues {
  strokeColor: string
  strokeWidth: number
  fillColor: string
  labelColor: string
  labelFontFamily: string
  labelFontSize: number
  // ...
}

export class GeoShapeUtil extends ShapeUtil<TLGeoShape> {
  override options: GeoShapeOptions = {
    getDisplayValues(_editor, shape, isDarkMode) {
      const theme = getDefaultColorTheme({ isDarkMode })
      return {
        strokeColor: getColorValue(theme, shape.props.color, 'solid'),
        strokeWidth: STROKE_SIZES[shape.props.size],
        fillColor: getColorValue(theme, shape.props.color, DEFAULT_FILL_COLOR_NAMES[shape.props.fill]),
        // ...
      }
    },
    getDisplayValueOverrides() {
      return {} // no overrides by default
    },
  }
}
```

### Consuming display values

In React component methods, use `useValue` for reactivity:

```tsx
component(shape: TLGeoShape) {
  const isDarkMode = useIsDarkMode()
  const dv = useValue(
    'geo shape display values',
    () => getDisplayValues(this, shape, isDarkMode),
    [shape, isDarkMode]
  )
  return <rect fill={dv.fillColor} stroke={dv.strokeColor} strokeWidth={dv.strokeWidth} />
}
```

In non-React methods (geometry, indicators), call directly:

```tsx
getGeometry(shape: TLGeoShape) {
  const dv = getDisplayValues(this, shape, false)
  const strokeWidth = dv.strokeWidth * shape.props.scale
  // ...
}
```

For SVG export, pass `ctx.isDarkMode`:

```tsx
toSvg(shape: TLGeoShape, ctx: SvgExportContext) {
  const dv = getDisplayValues(this, shape, ctx.isDarkMode)
  // ...
}
```

### Overriding via configure()

SDK consumers customize shapes without subclassing:

```tsx
const shapeUtils = [
  GeoShapeUtil.configure({
    getDisplayValueOverrides() {
      return {
        strokeColor: 'red',
        fillColor: 'pink',
      }
    },
  }),
]

function MyApp() {
  return <Tldraw shapeUtils={shapeUtils} />
}
```

### Fill pattern

Shapes that support pattern fills use local `*Fill`/`*PatternFill` components (e.g. `GeoFill`, `DrawFill`, `ArrowFill`) that accept a pre-computed `fillColor` instead of the shared `ShapeFill` which computes colors from the raw theme. This ensures fill colors flow through display values.

## Display values changes

**New infrastructure:**
- `ShapeUtilOptionsWithDisplayValues` interface and `getDisplayValues()` helper for merging base values with overrides
- `DEFAULT_FILL_COLOR_NAMES` mapping fill styles to theme color names
- `tlenv.supportsP3ColorSpace` for synchronous P3 color space detection
- Simplified `useColorSpace` hook to use `tlenv` instead of stateful effect-based detection

**Shape utils updated:**
- **GeoShapeUtil** — stroke, fill, label font/color/alignment/padding (16 display values)
- **NoteShapeUtil** — background, border, label font/color/alignment/padding (15 display values)
- **ArrowShapeUtil** — stroke, fill, label font/color/padding/border-radius (9 display values)
- **HighlightShapeUtil** — stroke color (sRGB + P3), width, underlay/overlay opacity (5 display values)
- **DrawShapeUtil** — stroke color, width, fill color (3 display values)
- **LineShapeUtil** — stroke color, width (2 display values)
- **TextShapeUtil** — color, font family/size/weight/style/variant, line height (7 display values)
- **FrameShapeUtil** — stroke, fill, heading colors for both default and showColors modes (8 display values)
- **EmbedShapeUtil** — shadow visibility (1 display value)
- **BookmarkShapeUtil, ImageShapeUtil, VideoShapeUtil** — scaffolded with empty display values

**Measurement consistency:** Text and note shape measurement functions now route through display values, so font overrides affect both rendering and geometry.

## resizeToBounds

Adds `Editor.resizeToBounds(shapes, bounds)` which translates and resizes a set of shapes so their combined page bounds matches the given target bounds. Follows the same pattern as `stretchShapes`: normalizes input IDs, filters to 90-degree-compatible rotations, builds arrow-connected clusters via `collectShapesViaArrowBindings`, computes a common bounds, then applies per-shape translate + resize within a single history entry.

### Change type

- [x] `api`

### Test plan

1. Open the examples app
2. Draw shapes of each type (geo, note, arrow, highlight, draw, line, text, frame)
3. Verify shapes render identically to before (no visual changes)
4. Test the `GeoShapeDisplayValues` example to confirm overrides work via `configure()`

- [x] Unit tests (`resizeToBounds.test.tsx`)
- [ ] End to end tests

### API changes

- Added `ShapeUtilOptionsWithDisplayValues` interface and `getDisplayValues()` helper
- Added display value interfaces for all shape utils: `GeoShapeUtilDisplayValues`, `NoteShapeUtilDisplayValues`, `ArrowShapeUtilDisplayValues`, `HighlightShapeUtilDisplayValues`, `DrawShapeUtilDisplayValues`, `LineShapeUtilDisplayValues`, `TextShapeUtilDisplayValues`, `FrameShapeUtilDisplayValues`, `EmbedShapeUtilDisplayValues`, `BookmarkShapeUtilDisplayValues`
- Added options interfaces: `GeoShapeUtilOptions`, `BookmarkShapeUtilOptions`, `EmbedShapeUtilOptions`
- Changed `ArrowShapeOptions`, `DrawShapeOptions`, `FrameShapeOptions`, `HighlightShapeOptions`, `LineShapeUtilOptions`, `TextShapeOptions` to extend `ShapeUtilOptionsWithDisplayValues`
- Moved `underlayOpacity`/`overlayOpacity` from `HighlightShapeOptions` static options to `HighlightShapeUtilDisplayValues`
- Added `tlenv.supportsP3ColorSpace`
- Added `Editor.resizeToBounds()` for resizing a set of shapes to fit a target bounding box
- Added `'resize_to_bounds'` to `TLShapeUtilCanBeLaidOutOpts` type union

### Release notes

- Add display values API to all shape utils, allowing customization of shape colors, fonts, sizes, and opacities via `configure()`
- Add `Editor.resizeToBounds()` method for fitting shapes into a target bounding box

---

> [!NOTE]
> **High Risk**
> Touches core shape rendering/options APIs and adds a new editor transform operation (`Editor.resizeToBounds`), so regressions could affect visuals, exports, and layout/selection behavior across many default shapes.
> 
> **Overview**
> Introduces a **new “display values” customization layer** for default shape utils, adding `getDisplayValues`/`getDisplayValueOverrides`-style options and exporting the related types/helpers so SDK consumers can override computed visual styling (colors, fonts, stroke widths, opacities) via `ShapeUtil.configure()` without subclassing.
> 
> Adds `Editor.resizeToBounds(shapes, bounds)` and extends `TLShapeUtilCanBeLaidOutOpts` with `resize_to_bounds` so sets of shapes (including arrow-bound clusters) can be translated/resized to fit a target box.
> 
> Docs/tooling updates: replaces link checking with a DB-driven internal/external link checker (new `check-external-links` script, `--fail` mode, build now fails early on broken links), refactors docs sitemap Framer path handling via `fetchFramerPaths`, adds new examples (`display-options`, `arrow-labels`), and updates many docs links/anchors to the new `sdk-features` routes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7ca62bdc92269b3ddad0964f78ea0f59e1267017. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/editor/src/lib/globals/environment.ts
```diff
@@ -16,6 +16,7 @@ const tlenv = {
 	isWebview: false,
 	isDarwin: false,
 	hasCanvasSupport: false,
+	supportsP3ColorSpace: false,
```
kaneel: sneaking a little something extra I can see here :)
joke aside, is this also related to display values? 🤔

packages/tldraw/src/lib/shapes/geo/components/GeoShapeBody.tsx
```diff
@@ -27,14 +36,55 @@ export function GeoShapeBody({
...
+	fill,
+	fillColor,
+}: {
+	d: string
+	fill: TLDefaultFillStyle
+	fillColor: string
+}) {
+	if (fill === 'none') return null
+	if (fill === 'pattern') return <GeoPatternFill d={d} fillColor={fillColor} />
+	return <path fill={fillColor} d={d} />
```
kaneel: I guess claude is somehow right, you could always

```tsx
if (fill === 'none') return null

return (
  <>
    <path fill={fillColor} d={d} />
	{fill === 'pattern' && <GeoPatternFill d={d} fillColor={fillColor} />}
  </>
)
```

packages/tldraw/src/lib/shapes/geo/components/GeoShapeBody.tsx
```diff
@@ -27,14 +36,55 @@ export function GeoShapeBody({
...
+	const teenyTiny = zoomLevel <= 0.18
+
+	return (
+		<>
+			<path fill={fillColor} d={d} />
+			<path
+				fill={
+					svgExport
+						? `url(#${getHashPatternZoomName(1, themeId)})`
+						: teenyTiny
```
kaneel: why not just return early in that case?

mimecuvalo:
Heya, I think this starts to lose the spirit of what we were trying to achieve with this body of work.

Some high-level concerns:
- what's the data model for storing things that `getDisplayValueOverrides` would use? This override function provides the mechanism/hook to give a custom style, but doesn't let me store/access that into the shape, right? We're forcing folks to just put things on `meta` which makes `getDisplayValueOverrides` feel 2nd-class and like an afterthought.
- this pattern of `*UtilDisplayValues` (e.g. `GeoShapeUtilDisplayValues`) means two things: that devs will need to understand the custom surface area of each of our unique shapes, which, generally have a strong commonality. Plus, this will be more of a moving target for each shape. As these start to deviate more, we will be forcing devs to chase shifting/breaking changes, adding to cognitive load between versions. An example might be if we were to add `labelLetterSpacing` for `GeoShapeUtilDisplayValues`. Now, the dev needs to make sure that they account for that in their display values function.
The function `getDisplayValues` works like this:
```
		...util.options.getDisplayValues(util.editor, shape, isDarkMode),
		...util.options.getDisplayValueOverrides(util.editor, shape, isDarkMode),
```

  so, if I've provided a custom `getDisplayValues` function, it will be _missing_ the new `labelLetterSpacing` and my shape will start to look more crufty than the canonical version. The drift will start to become more apparent over time.
- speaking of `getDisplayValues`, this means, that I can't reuse the canonical logic of the shape. Say, if I wanted to just override the `strokeColor`. In this current iteration, one is forced to override _all_ the custom logic for `getDisplayValues` if you override just one of them. Then I have to copy paste logic that I wouldn't have otherwise, or it forces me to use `getDisplayValueOverrides` but then that begs the question of why is `getDisplayValues` an option?
- again, as noted in our discussions, if we're doing this at the `configure()` level, per shape, we're forcing extra cognitive load on developers to iteratively go thru each individual shape. If we had any extra default shapes to the tldraw set, it'll get out of sync. Devs are lazy, and this would force them to be more intimately following every nuance of tldraw's roadmap. I think we're missing the global theming/styling, let's discuss if that should be a fast-follow to any foundational groundwork we do at this lower/non-global level.
- we can't add new colors/sizes/fills with this iteration and being able to add new colors to the palette (again, speaking to the `getDisplayValueOverrides`). I'd suggest trying to extend your example that has `CustomGeoShapeUtil` with a custom color when locked (i.e. a semantic `primary` color, or a `coral` color — alongside a 32-color palette, say). What that would look like might be a good direction to explore. I think back to my first week at tldraw, trying to implement Mario Paint for fun — the palette there is a lot larger than we affordance for. I would have loved to be able to add 10 more colors to bring fidelity to my prototype. Perhaps it would be helpful to have a practical customer/design partner to target here and make sure that their use case is being addressed.


So all that being said, I'd have strong reservations of going this route, so we should probably discuss more as a team.

MitjaBezensek:
My thoughts:
1. Like Mime mentioned `getDisplayValues` is a footgun. By using it you tie yourself directly to our logic and any future changes will break you. Do we need to expose it? 
2. The display values are computed from `DefaultColorThemePalette`, but the override mechanism sits alongside it as a separate API. There's no clear guidance on which to use when. Both can control the same visual property (`frameStroke` on the palette vs `strokeColor` on `FrameShapeUtilDisplayValues`), and they interact in non-obvious ways. If we decide to go this route then I'd at least unify this (even though it would be a breaking change).
3. Some simple tasks feel quite verbose and I need a deep understanding of the shape properties. 
For example, what if I want to make heavier strokes. I'd need to do something like
```ts
const STROKES = { s: 1, m: 3, l: 5, xl: 8 } as const

const shapeUtils = [
  GeoShapeUtil.configure({
    getDisplayValueOverrides(_editor, shape) {
      return { strokeWidth: STROKES[shape.props.size] }
    },
  }),
  DrawShapeUtil.configure({
    getDisplayValueOverrides(_editor, shape) {
      return { strokeWidth: STROKES[shape.props.size] }
    },
  }),
  ArrowShapeUtil.configure({
    getDisplayValueOverrides(_editor, shape) {
      return { strokeWidth: STROKES[shape.props.size] }
    },
  }),
  LineShapeUtil.configure({
    getDisplayValueOverrides(_editor, shape) {
      return { strokeWidth: STROKES[shape.props.size] }
    },
  }),
  HighlightShapeUtil.configure({
    getDisplayValueOverrides(_editor, shape) {
      return { strokeWidth: STROKES[shape.props.size] }
    },
  }),
]

<Tldraw shapeUtils={shapeUtils} />
  ```
I'd need to investigate which shapes use strokes. I'd need to update this if we add any new default shape with a stroke. I need to make sure I use this consistently across all the shapes. Which feels like another footgun.
4. This whole system takes a bit to understand and discover. I need to find three separate APIs (`DefaultColorThemePalette` mutation, `.configure()` on each shape util, shape-specific `DisplayValues` types) to understand the full customization surface.

I guess this is just part 1 of the changes? Would there be a follow up that would allow you to add or remove colors (as an example). Would that be a completely different system with another api? If that's the case then it's a bit hard for me to judge this PR since I'm not sure how it would work as a whole. Also, would people start using this together with custom colors in meta and then have to switch once we implement a type safe way to do that?

**Comment by steveruizok** [comment]
Some notes ahead:

- this PR is scoped to the first part of the problem that we discussed early on, that David also addressed, and which is common and separable from those other decisions: going from abstract values -> literal values. This is stuff we agreed on in our meeting and which is generally the same across all of the designs we've produced so far:  https://github.com/tldraw/tldraw/pull/7683, https://github.com/tldraw/tldraw/pull/8053, https://github.com/tldraw/tldraw/pull/7984, https://github.com/tldraw/tldraw/pull/6602
 

It differs from David/Mitja's designs by where the API is placed. I feel very strongly that we should put the API surface for this problem at the level of the shape options because of the relative fragility of the work here and because it only relates to our custom shapes. I do not think this is or should be a first-class API (at the editor level), but should rather be scoped as close to the shape as possible.

It differs from Alex's designs (at this point, uncontroversially) by not making literal values part of the styles themselves.

We should still add the ability to extend the theme, e.g. by adding other colors. Consider that the next PR in this suite.

We could also extend the theme to include other values, such as a base stroke width or a base line height, and then change our defaults to reference those values (kind of like `rem`). But we'd still need to provide developers the option to change those interpretations, too, so this level of API would be needed either way.

IMO this PR represents some design opinions but is otherwise in line with everything we've talked about so far.

> Heya, I think this starts to lose the spirit of what we were trying to achieve with this body of work.
> 
> Some high-level concerns:
> 
> * what's the data model for storing things that `getDisplayValueOverrides` would use? This override function provides the mechanism/hook to give a custom style, but doesn't let me store/access that into the shape, right? We're forcing folks to just put things on `meta` which makes `getDisplayValueOverrides` feel 2nd-class and like an afterthought.

It's different for every shape. It is a second class thing: it's a way of customizing our default shapes. We could / should still include the logic to extend the theme (in order to add colors, etc), that would be the next PR. 

> * this pattern of `*UtilDisplayValues` (e.g. `GeoShapeUtilDisplayValues`) means two things: that devs will need to understand the custom surface area of each of our unique shapes, which, generally have a strong commonality. Plus, this will be more of a moving target for each shape. As these start to deviate more, we will be forcing devs to chase shifting/breaking changes, adding to cognitive load between versions. An example might be if we were to add `labelLetterSpacing` for `GeoShapeUtilDisplayValues`. Now, the dev needs to make sure that they account for that in their display values function.
>   The function `getDisplayValues` works like this:
> 
> ```
> 		...util.options.getDisplayValues(util.editor, shape, isDarkMode),
> 		...util.options.getDisplayValueOverrides(util.editor, shape, isDarkMode),
> ```
> 
> so, if I've provided a custom `getDisplayValues` function, it will be _missing_ the new `labelLetterSpacing` and my shape will start to look more crufty than the canonical version. The drift will start to become more apparent over time.

Yes, that's right. The display values are different for each of our shapes and the logic for deriving the literal value is different in many cases. IMO this is fine. I haven't seen anything proposed that would address the fact that our changes could be breaking for downstream users—the same criticism would have been valid for earlier designs.

> * speaking of `getDisplayValues`, this means, that I can't reuse the canonical logic of the shape. Say, if I wanted to just override the `strokeColor`. In this current iteration, one is forced to override _all_ the custom logic for `getDisplayValues` if you override just one of them. Then I have to copy paste logic that I wouldn't have otherwise, or it forces me to use `getDisplayValueOverrides` but then that begs the question of why is `getDisplayValues` an option?

No, that's not true, the overrides accepts a partial. If there was complicated logic in `getDisplayValues`, and you want to modify part of that, then yes you'd need to include the complex logic. But I don't see any complex logic in the defaults, with the exception perhaps of the note shape's black = yellow in light mode. Is that what you're talking about?

We could remove `getDisplayValues` as an option and make it a private method of the component. However I would rather leave it customizable because, being reactive, it might end up capturing more than the end user would want. That could mean overriding the method to just return `{}` or doing some code surgery based on a copy and paste from our source code.

> * again, as noted in our discussions, if we're doing this at the `configure()` level, per shape, we're forcing extra cognitive load on developers to iteratively go thru each individual shape. If we had any extra default shapes to the tldraw set, it'll get out of sync. Devs are lazy, and this would force them to be more intimately following every nuance of tldraw's roadmap. I think we're missing the global theming/styling, let's discuss if that should be a fast-follow to any foundational groundwork we do at this lower/non-global level.

Yes and no. We could extend the theme to include certain values as well, such as a base stroke width or a base line height, and then change our defaults to reference those values (kind of like `rem`). But we'd still need to provide developers the option to change those interpretations, too, so this level of API would be needed either way.

> * we can't add new colors/sizes/fills with this iteration and being able to add new colors to the palette (again, speaking to the `getDisplayValueOverrides`). I'd suggest trying to extend your example that has `CustomGeoShapeUtil` with a custom color when locked (i.e. a semantic `primary` color, or a `coral` color — alongside a 32-color palette, say). What that would look like might be a good direction to explore. I think back to my first week at tldraw, trying to implement Mario Paint for fun — the palette there is a lot larger than we affordance for. I would have loved to be able to add 10 more colors to bring fidelity to my prototype. Perhaps it would be helpful to have a practical customer/design partner to target here and make sure that their use case is being addressed.

That's not the scope for this PR, that's a follow.

> So all that being said, I'd have strong reservations of going this route, so we should probably discuss more as a team.

**Comment by steveruizok** [comment]
> 1. Like Mime mentioned `getDisplayValues` is a footgun. By using it you tie yourself directly to our logic and any future changes will break you. Do we need to expose it?

Answered above but yeah, it may be better to make this one private to avoid confusion. It's just that it will still be captured and the developer would have no way of changing that. I'd rather not combine the two, of course. `getDisplayValueOverrides` is the one that should be used.

> 2. The display values are computed from `DefaultColorThemePalette`, but the override mechanism sits alongside it as a separate API. There's no clear guidance on which to use when. Both can control the same visual property (`frameStroke` on the palette vs `strokeColor` on `FrameShapeUtilDisplayValues`), and they interact in non-obvious ways. If we decide to go this route then I'd at least unify this (even though it would be a breaking change).

There were a few places where I felt like we should either remove keys from the theme (e.g. `frameStroke`) and just have a table in the component, or else go the other direction and push a lot of these things into the theme, to avoid accidental mappings of `solid` into ten different properties. But I think the displayOptions pattern here would be the same either way.

> 3. Some simple tasks feel quite verbose and I need a deep understanding of the shape properties.
>    For example, what if I want to make heavier strokes. I'd need to do something like
> 
> ```ts
> const STROKES = { s: 1, m: 3, l: 5, xl: 8 } as const
> 
> const shapeUtils = [
>   GeoShapeUtil.configure({
>     getDisplayValueOverrides(_editor, shape) {
>       return { strokeWidth: STROKES[shape.props.size] }
>     },
>   }),
>   DrawShapeUtil.configure({
>     getDisplayValueOverrides(_editor, shape) {
>       return { strokeWidth: STROKES[shape.props.size] }
>     },
>   }),
>   ArrowShapeUtil.configure({
>     getDisplayValueOverrides(_editor, shape) {
>       return { strokeWidth: STROKES[shape.props.size] }
>     },
>   }),
>   LineShapeUtil.configure({
>     getDisplayValueOverrides(_editor, shape) {
>       return { strokeWidth: STROKES[shape.props.size] }
>     },
>   }),
>   HighlightShapeUtil.configure({
>     getDisplayValueOverrides(_editor, shape) {
>       return { strokeWidth: STROKES[shape.props.size] }
>     },
>   }),
> ]
> 
> <Tldraw shapeUtils={shapeUtils} />
> ```
> 
> I'd need to investigate which shapes use strokes. I'd need to update this if we add any new default shape with a stroke. I need to make sure I use this consistently across all the shapes. Which feels like another footgun. 4. This whole system takes a bit to understand and discover. I need to find three separate APIs (`DefaultColorThemePalette` mutation, `.configure()` on each shape util, shape-specific `DisplayValues` types) to understand the full customization surface.
> 
> I guess this is just part 1 of the changes? Would there be a follow up that would allow you to add or remove colors (as an example). Would that be a completely different system with another api? If that's the case then it's a bit hard for me to judge this PR since I'm not sure how it would work as a whole. Also, would people start using this together with custom colors in meta and then have to switch once we implement a type safe way to do that?

See the last paragraph in my reply above:

> We could also extend the theme to include other values, such as a base stroke width or a base line height, and then change our defaults to reference those values (kind of like rem). But we'd still need to provide developers the option to change those interpretations, too, so this level of API would be needed either way.

File: packages/tldraw/src/lib/shapes/note/toolStates/Pointing.ts [hunk 4]
```diff
@@ -32,12 +34,22 @@ export class Pointing extends StateNode {
 		const id = createShapeId()
 		this.markId = editor.markHistoryStoppingPoint(`creating_note:${id}`)
 
+		// Resolve note dimensions from the util's defaults
+		const noteUtil = editor.getShapeUtil('note') as NoteShapeUtil
+		const dv = getDisplayValues(
+			noteUtil,
+			{ props: noteUtil.getDefaultProps() } as TLNoteShape,
+			false
+		)
```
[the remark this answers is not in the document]
steveruizok: This is intentional. The shape doesn't exist yet at this point — we're in `Pointing.onEnter` and need the default note dimensions to check for adjacent position snapping before creating the shape. The default `getDisplayValues` only accesses `shape.props`, and the default `getDisplayValueOverrides` ignores the shape entirely. If a consumer overrides `getDisplayValueOverrides` and accesses properties beyond `props`, that's on them.

packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
```diff
@@ -346,20 +416,24 @@ function DrawShapeSvg({ shape, zoomOverride }: { shape: TLDrawShape; zoomOverrid
...
-			/>
+			{isDot || !shape.props.isClosed || fill === 'none' ? null : fill === 'pattern' ? (
+				<PatternFill
+					d={solidStrokePath}
+					fillColor={fillColor}
+					patternFillFallbackColor={patternFillFallbackColor}
+				/>
+			) : (
+				<path fill={fillColor} d={solidStrokePath} />
+			)}
```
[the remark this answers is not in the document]
steveruizok: Not a real issue. Dots are single-point strokes with zero area — the fill path from `getDot()` is just a tiny circle that gets fully covered by the stroke-colored fill on the `<path>` below (`fill={isDot ? strokeColor : 'none'}`). Rendering a fill underlay beneath a dot has no visual effect. The old `<ShapeFill>` for dots was wasted work.

PR #8091, opened 2026-02-25
feature spike: add hotkey to create connected shapes

### LINKED ISSUE #6595: Add hotkeys for flowcharting
### What's the feature?

I often use other tools (excalidraw, etc.) to create flow charts. Using these tools, this can be done quickly without taking my hands off the keyboard by using a shortcut that duplicates my object in a particular direction, automatically generates a connecting arrow from the original object to the new object, and puts my focus into the new object.

I'd love to see this functionality brought into tldraw.

https://github.com/user-attachments/assets/ff489209-c993-431e-9c78-470a4f5f563f

The tldraw.com website

### Contact Details

chip@chipjohnson.net

### Code of Conduct

AniKrisn:
## Summary
- Adds a flowcharting keyboard shortcut: **Shift+A+Arrow** creates a new geo shape connected by an arrow in the chosen direction
- Works with dynamic sizing — the new shape accounts for `growY` so text-expanded shapes are duplicated at the correct height
- Arrow is automatically bound to both shapes; selection moves to the new shape

Closes #6595

## Test plan
- [ ] Select a geo shape, press Shift+A+Arrow in each direction — verify a connected shape + arrow is created
- [ ] Verify the new shape inherits style props from the source
- [ ] Verify the arrow is properly bound to both shapes
- [ ] Verify selection moves to the newly created shape
- [ ] Verify undo restores the previous state in one step
- [ ] Verify holding keys down doesn't repeat-fire the action
- [ ] Verify it does nothing when no shape is selected or a non-geo shape is selected
- [ ] Verify it does nothing in readonly mode

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -18,6 +18,7 @@ import { isOverArrowLabel } from '../../../shapes/arrow/arrowLabel'
 import { getHitShapeOnCanvasPointerDown } from '../../selection-logic/getHitShapeOnCanvasPointerDown'
 import { selectOnCanvasPointerUp } from '../../selection-logic/selectOnCanvasPointerUp'
 import { updateHoveredShapeId } from '../../selection-logic/updateHoveredShapeId'
+import { createConnectedShape } from '../createConnectedShape'
 import { hasRichText, startEditingShapeWithRichText } from '../selectHelpers'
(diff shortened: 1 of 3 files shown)


mimecuvalo:
would loooove to have something like this

angrycaptain19:
@max-drake thoughts on this as the broader direction for some of the stuff you were thinking about arrows -> create text shape?

steveruizok:
Closing this one, it would be better as an example.

PR #8101, opened 2026-02-26
fix(editor): wrap debug SVG shapes with ShapeCullingProvider

### LINKED ISSUE #8100: Debug SVG crashes app when shapes exist (useShapeCulling must be used within ShapeCullingProvider)
## Description
Enabling **Debug Mode** + **Debug SVG** crashes/breaks the app when there is at least one shape on the canvas.

Console error:
```useShapeCulling must be used within ShapeCullingProvider```

## Additional notes
- Works fine when the canvas is empty (no shapes).
- Once it happens, the state persists in a way that prevents recovery from within the app (requires restoring data).
- Worked ~a week ago

## Environment
- The most recent version on tldraw.com

### PR DESCRIPTION
**Pull request by MitjaBezensek**
Closes #8100

The `ShapesWithSVGs` component (used when Debug SVG is enabled) renders `<Shape>` without a `<ShapeCullingProvider>` ancestor. Since the `Shape` component calls `useShapeCulling()`, this throws: "useShapeCulling must be used within ShapeCullingProvider". The `ShapesToDisplay` component already had this provider but `ShapesWithSVGs` was missing it.

This wraps `ShapesWithSVGs` content with `<ShapeCullingProvider>`, `<CullingController>`, and the Safari reflow workaround — matching what `ShapesToDisplay` already does.

### Change type

- [x] `bugfix`

### Test plan

1. Open tldraw and create some shapes on the canvas
2. Open the debug menu and enable **Debug mode**
3. Enable **Debug SVG**
4. Verify the app no longer crashes and SVG debug copies appear below each shape

### Release notes

- Fix crash when enabling Debug SVG with shapes on the canvas.

---

> [!NOTE]
> **Low Risk**
> Low risk: localized change to canvas shape rendering that mainly affects debug-only `Debug SVG` mode and reuses existing culling logic.
> 
> **Overview**
> Fixes a crash when enabling **Debug SVG** by ensuring shapes rendered in that mode are wrapped in `ShapeCullingProvider` (so `useShapeCulling()` always has a provider).
> 
> Simplifies the canvas rendering path by replacing the separate `ShapesWithSVGs`/`ShapesToDisplay` branches with a single `ShapesLayer` that toggles debug SVG copies internally while still running `CullingController` and the Safari reflow workaround.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit aba4a4e3b1facb8e9b3a1cc5f63e8da1221cb14a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### CODE REVIEW

File: packages/editor/src/lib/components/default-components/DefaultCanvas.tsx [hunk 1]
```diff
@@ -401,12 +401,18 @@ function ShapesWithSVGs() {
...
+		<ShapeCullingProvider>
+			{renderingShapes.map((result) => (
+				<Fragment key={result.id + '_fragment'}>
+					<Shape {...result} />
+					<DebugSvgCopy id={result.id} mode="iframe" />
+				</Fragment>
+			))}
+			<CullingController />
+			{tlenv.isSafari && <ReflowIfNeeded />}
+		</ShapeCullingProvider>
```
kaneel, line 417: Is there a reason why this provider cannot be at a higher level in the tree? i.e. wrapping all shapes?

MitjaBezensek: Yeah, I guess we could add it here, after we confirm that we want to render shapes.
`
{hideShapes ? null : debugSvg ? <ShapesWithSVGs /> : <ShapesToDisplay />}
`
Probably not further up though, as we don't want to run this if we don't have shapes that need to be culled.

PR #8102, opened 2026-02-26
docs: fix broken example links after examples reorganization

MitjaBezensek:
Closes #8099

PR #7957 reorganized examples from flat `apps/examples/src/examples/{name}` into category subdirectories (`shapes/tools/`, `data/assets/`, `events/`, `ui/`, etc.). All ~115 "Related examples" GitHub links across 40 docs files now 404. This PR updates every link to use the new paths.

Also fixes 2 misnamed example references in `default-shapes.mdx`:
- `custom-styles` → `shapes/tools/shape-with-custom-styles`
- `shape-options` → `configuration/configure-shape-util`

@steveruizok we sometimes point to hosted examples and sometimes to code on github, not sure if intentional or not?

### Change type

- [x] `bugfix`

### Test plan

1. Open any docs page with "Related examples" links (e.g. `/sdk-features/shapes`, `/sdk-features/tools`, `/sdk-features/deep-links`)
2. Click the GitHub example links
3. Verify they resolve to the correct example directory on GitHub

### Release notes

- Fix ~115 broken GitHub example links across SDK documentation

---

> [!NOTE]
> **Low Risk**
> Low risk: documentation-only changes that update URLs and a couple of example name corrections, with no runtime or API impact.
> 
> **Overview**
> Fixes SDK docs and release notes that linked to example directories that were moved in the examples reorg.
> 
> Updates "Related examples" links across many `sdk-features/*.mdx` pages to point at the new categorized paths (e.g. `ui/`, `events/`, `configuration/`, `editor-api/`, `shapes/tools/`, `data/assets/`, `use-cases/`, `layout/`). Also corrects a couple of misnamed example references in `default-shapes.mdx` and updates the v4.3.0 release note link for the pin bindings example.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de7a306e818a040b43905f2e5e76e5182b796e9a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8103, opened 2026-02-27
fix(docs): replace linkinator with static broken link checker and fix 134 broken links

steveruizok:
In order to catch broken internal links before they ship (like the ones fixed in #8102), this PR replaces the `linkinator`-based HTTP crawler with a static checker that validates links against the content database. The old checker required a running server, couldn't fail the build, and missed most broken links. The new checker runs after `refresh-everything` and before `next build`, failing early if any broken links are found.

Along the way, this fixes all 134 broken internal links found across the docs:
- 25 content files with stale `/docs/*` paths (should be `/sdk-features/*`), wrong heading anchors, double-prefixed paths, and renamed example links
- API doc generator skipping `TypeAlias`/`Interface` namespace members (the `// TODO: document these` that caused `T.Validatable` etc. to generate broken links)
- Namespace member path generation producing standalone paths like `T-Validatable` instead of anchor links like `T#Validatable`

The checker uses a `--fail` flag to control behavior: `yarn check-links --fail` (used in `build`) exits non-zero on broken links, while `yarn check-links` (used in dev) only warns. This ensures broken links never block the dev server from starting.

Additionally, this adds a standalone `yarn check-external-links` script for manually checking external (HTTP/HTTPS) links across all docs content. It deduplicates URLs, uses HEAD with GET fallback (30s timeout, 10 concurrent requests), and retries transient failures with backoff. A skip list filters out domains that block automated requests (npmjs.com, shadertoy.com). This replaces the external link checking that was previously handled by linkinator but is too slow/flaky for the build pipeline.

Also refactors the hardcoded Framer path prefixes into a shared `fetchFramerPaths()` utility that fetches the live Framer sitemap. Both the link checker and `app/sitemap.ts` now use this, so the list stays up to date automatically.

Fixes a broken external link to the removed `custom-toolbar` example in the v2.1.0 release notes.

### Change type

- [x] `improvement`

### Test plan

1. `cd apps/docs && yarn refresh-everything && yarn check-links` — should pass with 0 broken links (warnings only)
2. `yarn check-links --fail` — same result but would exit 1 if any broken links existed
3. Manually add a broken link to a content file (e.g. `[test](/nonexistent)`) and run `yarn check-links --fail` — should exit 1 with the broken link reported
4. Run `yarn build` from `apps/docs` — check-links runs with `--fail` before `next build`
5. `yarn check-external-links` — checks all ~630 unique external URLs, reports broken ones

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix 134 broken internal links across SDK docs, starter kits, and API reference
- Add static broken link checker to docs build pipeline (replaces linkinator)
- Add standalone external link checker script (`yarn check-external-links`)
- Document `TypeAlias` and `Interface` members within API reference namespaces

### DIFF
--- a/apps/docs/scripts/lib/checkBrokenLinks.ts
+++ b/apps/docs/scripts/lib/checkBrokenLinks.ts
@@ -1,3 +1,4 @@
+import { fetchFramerPaths } from '@/utils/framer-sitemap'
 import { nicelog } from '@/utils/nicelog'
 import { connect } from './connect'
 
@@ -7,25 +8,6 @@ export interface BrokenLink {
 	url: string
 	reason: string
 }
-
-// Framer rewrite prefixes — any path starting with these is served externally
-const FRAMER_PREFIXES = [
-	'/',
-	'/404',
-	'/accessibility',
-	'/blog',
-	'/careers',
-	'/company',
-	'/events',
-	'/faq',
-	'/features',
-	'/get-a-license',
-	'/legal',
-	'/partner',
-	'/pricing',
-	'/showcase',
-	'/thanks',
-]
 
 // Internal rewrites: source → destination (both are valid paths)
 const INTERNAL_REWRITES: Record<string, string> = {
@@ -124,23 +106,11 @@ function tryRedirect(urlPath: string): string | null {
 	return null
 }
 
-/**
- * Check if a path is covered by a Framer rewrite prefix.
- * The exact prefix '/' only matches exactly '/'.
- */
-function isFramerPath(urlPath: string): boolean {
-	for (const prefix of FRAMER_PREFIXES) {
-		if (prefix === '/') {
-			if (urlPath === '/') return true
-		} else {
-			if (urlPath === prefix || urlPath.startsWith(prefix + '/')) return true
-		}
-	}
-	return false
-}
-
 export async function checkBrokenLinks(): Promise<number> {
 	const db = await connect({ mode: 'readonly' })
+
+	// Fetch framer paths from the live sitemap so we don't maintain a stale list
+	const framerPaths = await fetchFramerPaths()
 
 	// Build set of all valid paths
 	const validPaths = new Set<string>()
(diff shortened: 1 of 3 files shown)


steveruizok:
bot lost external link checking, re-steering

apps/docs/scripts/lib/checkBrokenLinks.ts
```diff
@@ -1,25 +1,311 @@
...
+			// Skip static asset paths (images, fonts, etc.)
+			if (STATIC_ASSET_EXTENSIONS.test(url)) continue
+
+			// Split path and fragment
+			const hashIdx = url.indexOf('#')
+			const urlPath = hashIdx >= 0 ? url.slice(0, hashIdx) : url
+			const fragment = hashIdx >= 0 ? url.slice(hashIdx + 1).toLowerCase() : null
+
+			// Empty path with fragment = same-page anchor
+			if (urlPath === '' && fragment) {
```
MitjaBezensek, line 202: Not sure we can hit this part at all, we filter out fragment only links in `extractInternalLinks`.

apps/docs/scripts/lib/checkBrokenLinks.ts
```diff
@@ -1,25 +1,311 @@
...
-		recurse: true,
+export interface BrokenLink {
+	articlePath: string
+	line: number
+	url: string
+	reason: string
+}
+
+// Framer rewrite prefixes — any path starting with these is served externally
+const FRAMER_PREFIXES = [
```
MitjaBezensek: We already have some logic for getting the framer sitemap here
https://github.com/tldraw/tldraw/blob/ce797ebfac7a235133111db10abc2b6166546dff/apps/docs/app/sitemap.ts#L4

We could use that instead so that we don't have stale values / have to update this when we add / remove framer pages.

steveruizok: done!

PR #8120, opened 2026-03-01
feat(editor): add resizeToBounds method

steveruizok:
In order to support resizing a set of shapes to fit a target bounding box (e.g. for layout operations), this PR adds `Editor.resizeToBounds()`.

The method takes an array of shapes (or shape IDs) and a target `BoxLike`, then scales and repositions the shapes so their combined page bounds match the target. It handles arrow-bound shape clusters, parent transforms, and aspect ratio locking.

This PR also:
- Fixes a bug in `stretchShapes` where the shared `localOffset` vector was being mutated instead of the per-shape clone when applying parent rotation
- Adds a zero-dimension guard to `resizeToBounds` to avoid `NaN`/`Infinity` scales
- Extracts the shared clustering boilerplate from 6 layout methods (`stackShapes`, `packShapes`, `alignShapes`, `distributeShapes`, `stretchShapes`, `resizeToBounds`) into a private `getShapeClusters` helper, removing ~230 lines of duplicated code
- Adds a `resizeToBounds` interactive example
- Updates the shape-transforms docs page with a new "Resizing to bounds" subsection and updates `canBeLaidOut` type comment

| Area | Added | Removed | Net |
|------|------:|--------:|----:|
| Core code | +161 | -270 | -109 |
| Tests | +245 | 0 | +245 |
| Example | +172 | 0 | +172 |
| Docs | +14 | -1 | +13 |
| API report | +2 | -1 | +1 |
| **Total** | **+594** | **-271** | **+323** |

### Change type

- [x] `feature`

### Test plan

1. Call `editor.resizeToBounds([shape1, shape2], { x: 0, y: 0, w: 500, h: 500 })`
2. Verify shapes are resized and repositioned to fill the target bounds
3. Verify undo restores original positions
4. Navigate to the `resize-to-bounds` example and verify buttons resize shapes correctly

- [x] Unit tests

### API changes

- Added `Editor.resizeToBounds(shapes, bounds)` method
- Added `'resize_to_bounds'` to `TLShapeUtilCanBeLaidOutOpts.type`

### Release notes

- Add `Editor.resizeToBounds()` method for resizing shapes to fit target bounds
- Fix `localOffset` mutation bug in `stretchShapes` when shapes have parent transforms

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -158,6 +158,7 @@ import {
 	TLEditStartInfo,
 	TLGeometryOpts,
 	TLResizeMode,
+	TLShapeUtilCanBeLaidOutOpts,
 	TLShapeUtilCanBindOpts,
 } from './shapes/ShapeUtil'
 import { RootState } from './tools/RootState'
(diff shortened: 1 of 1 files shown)


steveruizok:
For additional context: I was watching an agent try to use our Editor methods to move a group of shapes from one place / size to another. (IIRC, something like "make the man bigger and put him next to the tree"). It was very difficult / inefficient because of how many intermediate calculations it needed to make in order to update each of the shapes. As a user, I would have used the selection to operate on the shapes as if they were a single rectangle: select the shapes to create the selection box, then move the selection box, resize the selection box, etc. This method allows the same sort of abstraction at the level of our API.

packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -69,7 +69,7 @@ export interface TLShapeUtilCanBindOpts<Shape extends TLShape = TLShape> {
  */
 export interface TLShapeUtilCanBeLaidOutOpts {
 	/** The type of action causing the layout. */
-	type?: 'align' | 'distribute' | 'pack' | 'stack' | 'flip' | 'stretch'
+	type?: 'align' | 'distribute' | 'pack' | 'stack' | 'flip' | 'stretch' | 'resize_to_bounds'
```
mimecuvalo, line 72: maybe just `scale`?

steveruizok: hm, could be collidey with other scale stuff. I'd like to reserve scale for when/if scale becomes a first class property

PR #8123, opened 2026-03-02
tldraw mcp app

max-drake:
Creates an MCP app that allows users to interact with a tldraw canvas alongside their llm of choice.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/apps/tldraw-mcp-app/wrangler.toml
+++ b/apps/tldraw-mcp-app/wrangler.toml
@@ -11,12 +11,10 @@ compatibility_flags = ["nodejs_compat"]
 "@modelcontextprotocol/sdk/server/mcp.js" = "./node_modules/agents/node_modules/@modelcontextprotocol/sdk/dist/esm/server/mcp.js"
 "@modelcontextprotocol/sdk/types.js" = "./node_modules/agents/node_modules/@modelcontextprotocol/sdk/dist/esm/types.js"
 
-[define]
-"process.env.TLDRAW_MCP_JSON_HEALING_DEBUG" = "undefined"
-
 [assets]
 directory = "./dist"
 binding = "ASSETS"
+run_worker_first = true
 
 [durable_objects]
 bindings = [{ class_name = "TldrawMCP", name = "MCP_OBJECT" }]


apps/tldraw-mcp-app/src/shared/utils.ts
```diff
@@ -0,0 +1,43 @@
+import type { CallToolResult } from '@modelcontextprotocol/sdk/types.js'
+import type { TLShape } from 'tldraw'
+
+export function isPlainObject(value: unknown): value is Record<string, unknown> {
```
kostyafarber, line 4: possibly some of these utilities already exist somewhere else in tldraw? worth a quick check so we don't duplicate code

apps/tldraw-mcp-app/src/widget/debug.ts
```diff
@@ -0,0 +1,10 @@
+export const debugLines: string[] = []
+
+export function log(msg: string) {
+	debugLines.push(`${new Date().toISOString().slice(11, 23)} ${msg}`)
+	const el = document.getElementById('debug')
+	if (el) el.textContent = debugLines.join('\n')
+}
+
+window.addEventListener('error', (e) => log(`ERROR: ${e.message}`))
```
kostyafarber, line 9: will need to double check what we use these for, don't add these on the window in prod

apps/tldraw-mcp-app/src/widget/export-tldr.ts
```diff
@@ -0,0 +1,71 @@
...
+		for (const binding of editor.getBindingsInvolvingShape(shape)) {
+			if (seen.has(binding.id)) continue
+			seen.add(binding.id)
+			bindings.push(structuredClone(binding))
+		}
+	}
+	return bindings
+}
+
+function buildTldrDocument(editor: Editor) {
```
kostyafarber, line 17: are there helpers already somewhere to create tldr files? I'm not sure but worth checking as they might handle more edge cases etc

apps/tldraw-mcp-app/src/widget/export-tldr.ts
```diff
@@ -0,0 +1,71 @@
...
+export function exportTldr(editor: Editor) {
+	const doc = buildTldrDocument(editor)
+	const json = JSON.stringify(doc, null, 2)
+
+	// Copy to clipboard
+	navigator.clipboard.writeText(json).catch(() => {
+		// Clipboard may be unavailable in some contexts.
+	})
+
+	log(json)
```
kostyafarber, line 61: lets not log the whole serialised document (at least in prod)

MCP_APPS_GUIDE.md
```diff
@@ -0,0 +1,1881 @@
+# MCP Apps: Complete Reference Guide
```
kostyafarber, line 1: we can remove these from the commit (keep them locally)

mcp-deployed-server-plan.md
kostyafarber, line 1: same here we can remove this

kostyafarber (commented):
lgtm pending a few comments, can sic claude onto them. 

on another note, we need to decide what to do with the local server/stdio setup and if we still want this in there

Issue #8089, opened 2026-02-25
Add an "open in tldraw" API for creating files from external sources


steveruizok:
## Problem

There's no way for external services to programmatically create a tldraw.com file pre-populated with content. Use cases like "open in tldraw" buttons (similar to "Open in CodeSandbox" or "Open in StackBlitz") aren't possible today.

The existing `/api/app/tldr` endpoint creates files from `.tldr` snapshots but requires authentication and is designed for internal use (file imports, slurping). There's no public-facing API or URL scheme for external tools to send data into tldraw and get back an editable file.

## Proposed solution

Add a public API endpoint and/or URL scheme that accepts tldraw-compatible data and creates a new file, then redirects the user to it. For example:

**POST endpoint:**

```
POST https://tldraw.com/api/open
Content-Type: application/json

{
  "records": [...TLRecord[]],
  "schema": SerializedSchema
}
```

Returns a redirect or URL to the newly created file.

**URL scheme (for simple cases):**

```
https://tldraw.com/open?snapshot=<url-to-json>
```

Where `snapshot` points to a publicly accessible JSON file containing the records and schema.

## Use cases

- **Developer tools**: "Open in tldraw" buttons in documentation, code playgrounds, or design tools that generate tldraw-compatible output
- **AI/LLM integrations**: Tools that generate diagrams or visual content could offer a one-click "view in tldraw" experience
- **Import pipelines**: Convert from other formats (Excalidraw, Mermaid, SVG) and open directly in tldraw
- **Sharing workflows**: Generate a tldraw file from structured data (e.g. a JSON schema visualizer) and share it
- **Education**: Course materials that link directly to pre-built tldraw canvases for students to interact with

## Design considerations

- **Authentication**: Should this require a tldraw.com account? Could create an anonymous/temporary file for unauthenticated users, or prompt sign-in first
- **Rate limiting**: Needs protection against abuse
- **Data size limits**: Large snapshots with many assets need bounded handling
- **Ephemeral vs persistent**: Could create temporary files that expire, or require auth for persistent files
- **CORS**: If supporting cross-origin POST, needs appropriate CORS headers
- **Existing infrastructure**: The `createFile()` flow and `/api/app/tldr` endpoint already handle file creation from snapshots — this would be a public-facing wrapper with appropriate guards

## Related

- #7680 — forking published files (related but UI-only, not an external API)
- The existing `createSource` tracking in `TldrawApp.createFile()` already supports tracking file origins
[...]

PR #8128, opened 2026-03-02
feat(dotcom): add /import route and app file download endpoint

ds300:
In order to support importing a tldraw file from a URL (e.g. \`/import?url=...\`) and to allow downloading app files as .tldr from the sidebar and editor, this PR adds the \`/import\` route, a server-side download endpoint, and wires the Download action to use it for all app files. Relates to #8089.

The import URL is passed through React Router location state so there's no extra session storage key. On import failure the user is shown an error toast; on success they're navigated to the new file. When an upload fails, the Promise resolves immediately via a new \`onUploadError\` callback on \`uploadTldrFiles\`. The duplicate Download item was removed from the top-left editor menu, and Download is now available for all files in the sidebar (not just the active one).

The download endpoint (\`GET /api/app/file/:id/download\`) runs in the Durable Object, streams a \`.tldr\` snapshot, and inlines referenced R2 assets as base64 data URIs. It uses the same auth/group/shared checks and rate limiting as joining the file. The stream's \`start\` callback is wrapped in \`try-catch\` so any R2 failure mid-stream calls \`controller.error()\` — without this, an exception after the 200 header is sent would silently corrupt the file.

The download filename prefers the \`TlaFile.name\` (kept in sync by the app layer) over \`TLDocument.name\` (only updated while the file is open in an editor). Asset URL parsing strips query parameters before extracting the object name, so signed/cache-busted URLs are handled correctly.

### Change type

- [x] \`feature\`

### Test plan

1. **Import from URL (signed in):** Open \`/import?url=<valid-tldraw-json-url>\`. Should redirect to \`/\`, import the file, then navigate to it.
2. **Import from URL (signed out):** Open \`/import?url=<valid-url>\`. Sign-in dialog should appear. After signing in, the import should run and navigate to the new file.
3. **Import from URL (bad URL):** Open \`/import?url=<invalid-url>\`. Should show an "Import failed" toast.
4. **Download (app file, active):** From the file menu or editor on an open app file, use Download. Browser should download a \`.tldr\` file with assets inlined.
5. **Download (app file, inactive):** From the sidebar file menu on a file that is not currently open, use Download. Should download that file's \`.tldr\` with the correct name.
6. **Download (local/unsaved):** In the editor without an app file, Download should still serialize from editor state as before.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Add \`/import?url=...\` route to import a tldraw file from a URL (shows sign-in dialog when logged out).
- Add server-side file download: Download in the file menu and editor now downloads a \`.tldr\` from the server with assets inlined as base64, and is available for all files in the sidebar.

### DIFF
--- a/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
@@ -681,7 +681,9 @@ export class TLFileDurableObject extends DurableObject {
 		const assetObjectNames = new Set(assetRows.map((r) => r.objectName))
 
 		const documentRecord = records.find((r) => r.typeName === 'document') as TLDocument | undefined
-		const rawName = documentRecord?.name?.trim()
+		// Prefer the TlaFile.name (kept in sync by the app layer) over the TLDocument.name
+		// (which is only updated when the document is open in an editor).
+		const rawName = file.name?.trim() || documentRecord?.name?.trim()
 		const sanitized =
 			rawName?.replace(/[^ \w-]/g, '_').slice(0, 200) || `${this.documentInfo.slug}.tldr`
 		const filename = sanitized.endsWith('.tldr') ? sanitized : `${sanitized}.tldr`
@@ -705,7 +707,7 @@ export class TLFileDurableObject extends DurableObject {
 							assetSrc &&
 							!assetSrc.startsWith('data:')
 						) {
-							const objectName = assetSrc.split('/').pop()
+							const objectName = new URL(assetSrc).pathname.split('/').pop()
 							if (objectName && assetObjectNames.has(objectName)) {
 								const blob = await env.UPLOADS.get(objectName)
 								if (blob) {
--- a/apps/dotcom/client/e2e/fixtures/Editor.ts
+++ b/apps/dotcom/client/e2e/fixtures/Editor.ts
@@ -57,7 +57,9 @@ export class Editor {
 	}
 
 	async getCurrentFileName() {
-		return await this.fileName.innerText()
+		const text = await this.fileName.innerText()
+		// The UI replaces spaces with \u00a0 to prevent line-breaking; normalise back.
+		return text.replace(/\u00a0/g, ' ')
 	}
 
 	@step
--- a/apps/dotcom/client/e2e/fixtures/ImportHelper.ts
+++ b/apps/dotcom/client/e2e/fixtures/ImportHelper.ts
@@ -1,0 +1,51 @@
+import type { Page } from '@playwright/test'
+import { step } from './tla-test'
+
+export const IMPORT_URL = 'https://e2e-mock.tldraw.xyz/test-import.tldr'
+
+// Sequences match the current record versions so no migrations run on import,
+// which would otherwise overwrite fields (e.g. AddName sets name='').
+const MINIMAL_TLDR_FILE = {
+	tldrawFileFormatVersion: 1,
+	schema: {
+		schemaVersion: 2,
+		sequences: { 'com.tldraw.document': 2, 'com.tldraw.page': 1 } as Record<string, number>,
+	},
+	records: [
+		{
+			typeName: 'document',
+			id: 'document:document',
+			name: 'e2e import test',
+			gridSize: 10,
+			meta: {},
+		},
+		{
+			typeName: 'page',
+			id: 'page:page',
+			index: 'a1',
+			meta: {},
+			name: 'Page 1',
+		},
+	],
+}
+
+export class ImportHelper {
+	constructor(private readonly page: Page) {}
+
+	@step
+	async mockUrl(url = IMPORT_URL) {
+		await this.page.route(url, (route) =>
+			route.fulfill({
+				status: 200,
+				contentType: 'application/json',
+				headers: { 'Access-Control-Allow-Origin': '*' },
+				body: JSON.stringify(MINIMAL_TLDR_FILE),
+			})
+		)
+	}
+
+	@step
+	async navigate(url = IMPORT_URL) {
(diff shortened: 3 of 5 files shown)


apps/dotcom/client/src/tla/components/TlaEditor/useFileEditorOverrides.ts
```diff
@@ -14,7 +14,18 @@ import { useMaybeApp } from '../../hooks/useAppState'
...
-export async function download(editor: Editor, name: string) {
+/** Triggers a GET to the app file download endpoint; browser downloads via Content-Disposition. */
+export function downloadAppFile(fileId: string) {
+	const url = `/api/app/file/${fileId}/download`
+	const a = document.createElement('a')
+	a.href = url
+	a.download = ''
+	document.body.appendChild(a)
+	a.click()
+	document.body.removeChild(a)
```
kaneel: You can remove append and remove, just click!

apps/dotcom/client/src/tla/pages/import.tsx
```diff
@@ -0,0 +1,43 @@
...
+	useEffect(() => {
+		if (!url || url.trim() === '') {
+			navigate(routes.tlaRoot(), { replace: true })
+			return
+		}
+		if (!auth.isLoaded) return
+
+		if (!auth.isSignedIn) {
+			if (!dialogShown) {
+				setDialogShown(true)
```
kaneel: Curious. What's isLoaded and isSignedIn in that case? Is there a way to simplify this cascade of condition into something that is quickly readable?

apps/dotcom/client/src/tla/pages/local.tsx
```diff
@@ -25,8 +32,46 @@ export function Component() {
...
+				if (isImportRedirect) {
+					setInSessionStorage(
+						SESSION_STORAGE_KEYS.PENDING_IMPORT_URL,
+						redirectUrl.searchParams.get('url')!
+					)
+					// Fall through to pending import handling below
+				} else {
+					navigate(redirectTo, { replace: true })
+					return
+				}
```
kaneel: conditions could be streamlined a little: 


```ts
const redirectUrl = redirectTo.startsWith('/')
  ? new URL(redirectTo, window.location.origin)
  : null

const isImportUrl =
  redirectUrl?.pathname === '/import'
    ? redirectUrl.searchParams.get('url')
    : null

if (!isImportUrl) {
  navigate(redirectTo, { replace: true })
  return
}

setInSessionStorage(SESSION_STORAGE_KEYS.PENDING_IMPORT_URL, pendingImportUrl)
```

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -91,6 +98,25 @@ async function canAccessTestProductionFile(
...
 
+function pruneUnusedAssetsForTldr(records: TLRecord[]): TLRecord[] {
+	const usedAssets = new Set<TLAssetId>()
+	for (const record of records) {
+		if (record.typeName === 'shape' && 'assetId' in record.props && record.props.assetId) {
+			usedAssets.add(record.props.assetId as TLAssetId)
+		}
+	}
+	return records.filter((r) => r.typeName !== 'asset' || usedAssets.has(r.id as TLAssetId))
+}
```
kaneel, line 109: ✂️ 💪

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
```diff
@@ -91,6 +98,25 @@ async function canAccessTestProductionFile(
...
+
+function arrayBufferToBase64(ab: ArrayBuffer): string {
+	const bytes = new Uint8Array(ab)
+	let binary = ''
+	for (let i = 0; i < bytes.length; i++) {
+		binary += String.fromCharCode(bytes[i])
+	}
+	return btoa(binary)
```
kaneel: Do you think we could improve this one speed-wise by going every 1024 bytes? 

```ts
for (let i = 0; i < bytes.length; i += 1024) {
  binary += String.fromCharCode(...bytes.subarray(i, i + 1024))
}
```

apps/dotcom/client/src/tla/components/TlaEditor/useFileEditorOverrides.ts
```diff
@@ -14,7 +14,18 @@ import { useMaybeApp } from '../../hooks/useAppState'
 import { useIntl, useMsg } from '../../utils/i18n'
 import { editorMessages as messages } from './editor-messages'
 
-export async function download(editor: Editor, name: string) {
+/** Triggers a GET to the app file download endpoint; browser downloads via Content-Disposition. */
+export function downloadAppFile(fileId: string) {
+	const url = `/api/app/file/${fileId}/download`
+	const a = document.createElement('a')
+	a.href = url
+	a.download = ''
```
mimecuvalo, line 23: prbly give this a name? feels strange to leave blank i guess

in general, let's take the approach in `downloadFile` - we shouldn't need to attach to the DOM is my understanding

apps/dotcom/sync-worker/src/TLDrawDurableObject.ts
ds300: ugh i should have caught this. thanks. i have switched to UInt8Array.toBase64

apps/dotcom/client/src/tla/components/TlaEditor/useFileEditorOverrides.ts
ds300: i added a comment to explain the empty string here. we set the filename in the content-disposition header on the backend so there's no need to do it here.

ds300:
@mimecuvalo @kaneel thanks for the review, updated

steveruizok:
Can we not block PRs please?

PR #8130, opened 2026-03-02
fix(arrows): clamp boundary anchors to avoid degenerate intersections

### LINKED ISSUE #8125: Arrows disappear when terminals land exactly on shape edges or corners
## Description

https://github.com/user-attachments/assets/17fcdc32-efdb-499a-97af-3adf944cfce4

When arrow anchors place terminals exactly on shape boundaries (e.g., `x === 0`, `y === 0`, or `x === 1`, `y === 1` in normalized coordinates), arrows can flicker, collapse, or disappear entirely. This is most visible with **self-referential arrows** (both terminals bound to the same shape), but may also affect arrows between two separate shapes.

This happens because the segment-intersection math used for arrow endpoint resolution oscillates around a numerical degeneracy at exact edge/corner alignments. Intersections can flip between:

- A valid intersection point
- `null` (coincident/parallel branch — the line lies exactly along the edge)

When intersections return `null`, the computed start/end points collapse or jump. This makes arrow info invalid (zero length or invalid arc), and the render code intentionally returns `null` to hide invalid geometry — causing the arrow to blink out.

## Steps to reproduce

1. Create a shape and an arrow with both start and end terminals bound to the same shape
3. Use `isPrecise: true` with `normalizedAnchor` values at exact boundaries (e.g., `{ x: 0, y: 0 }` for start, `{ x: 0.1, y: 0 }` for end)
4. The arrow at anchor `x: 0` (the exact corner/edge) may flicker or disappear, while arrows at non-boundary anchors render fine
5. Moving or resizing the shape or connected arrows can cause arrows near boundary anchors to oscillate between visible and invisible

## Expected behavior

Arrows should remain stable and visible when terminals are at or near exact shape boundaries. The intersection math should handle these degenerate cases gracefully.

## Actual behavior

Arrows intermittently disappear because:
- `intersectLineSegmentLineSegment()` returns `null` for coincident/parallel lines (precision `1e-10`)
- Without a valid intersection, the arrow endpoint can't be resolved
- Arrow info becomes invalid (`length === 0` or non-finite arc center)
- Render code returns `null` for invalid arrows, hiding them

## Relevant code paths

- **Straight arrow intersection**: `packages/tldraw/src/lib/shapes/arrow/straight-arrow.ts` — `updateArrowheadPointWithBoundShape()` calls `geometry.intersectLineSegment()`
- **Curved arrow intersection**: `packages/tldraw/src/lib/shapes/arrow/curved-arrow.ts` — calls `geometry.intersectCircle()`
- **Line-line intersection**: `packages/editor/src/lib/primitives/intersect.ts` — `intersectLineSegmentLineSegment()` uses `1e-10` precision
- **Validity checks**: `ArrowShapeUtil.tsx` returns `null` at lines ~771, ~820, ~829 when `info.isValid` is false or start/end collapse
- **Epsilon utilities**: `packages/editor/src/lib/primitives/utils.ts` — `approximately()` uses `1e-6`, intersection uses `1e-10`

## Possible approaches

- Add a small perpendicular offset or jitter to intersection rays when they are near-parallel to shape edges
- Use a fallback intersection strategy (e.g., project onto the nearest edge point) when standard intersection returns `null` for bound shapes
- Clamp anchor coordinates to avoid exact 0/1 boundary values (e.g., clamp to `[ε, 1-ε]`)
- Widen the coincidence tolerance in intersection code specifically for arrow endpoint resolution

AniKrisn:
Closes #8125

https://github.com/user-attachments/assets/cf96fc2a-c7f2-4448-8ccf-42e0eaa5ba2b

Arrow endpoints flicker when `normalizedAnchor` coordinates are exactly 0 or 1 (shape edges/corners). At these exact values, the intersection math becomes numerically unstable — small floating-point errors cause intersections to toggle between found and not-found across frames.

**Solution:** if arrow is precise or forceImprecise, clamp each anchor coordinate from [0, 1] to [0.001, 0.999]. This clamping happens inside `getArrowTerminalInArrowSpace`, which runs every time the arrow info is recomputed (i.e. whenever the arrow or its bound shapes change). The binding record in the store is never modified — the nudge only exists ephemerally during computation, so the intersection math never sees an exact boundary value.

### Change type

- [x] `bugfix`

### Test plan

1. Create a shape and programmatically bind a self-referential arrow with anchors at exact boundary positions (e.g. `{x: 0, y: 0}`, `{x: 0.5, y: 0}`)
2. Verify arrows render without flickering
3. Resize and move the shape — arrows should remain stable

- [x] Unit tests

### Release notes

- Fix arrow endpoint flickering when anchors are at exact shape boundaries

---

> [!NOTE]
> **Medium Risk**
> Touches core arrow terminal computation; clamping precise anchors near 0/1 can subtly shift endpoints (up to ~0.1% of shape size) and may affect arrow rendering/attachment in edge cases.
> 
> **Overview**
> Prevents arrow endpoint flicker when a bound terminal’s `normalizedAnchor` lands exactly on a shape edge/corner by clamping precise anchors away from `0`/`1` during `getArrowTerminalInArrowSpace` computations.
> 
> Adds a small epsilon-based `clampNormalizedAnchor` helper and refactors the terminal point calculation to use the clamped anchor without mutating stored binding data.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4ecde2cf1bec7edadca0a0d18ecb7186a86e60f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/shared.ts
+++ b/packages/tldraw/src/lib/shapes/arrow/shared.ts
@@ -89,11 +89,7 @@ export function getArrowTerminalInArrowSpace(
 		// If the parent is the bound shape, then it's always treated as precise.
 		const shouldUsePreciseAnchor = binding.props.isPrecise || forceImprecise
 		const normalizedAnchor = shouldUsePreciseAnchor
-			? // Keep exact bindings untouched; only sanitize imprecise/precise anchors to avoid
-				// edge/corner degeneracies when anchors are exactly 0 or 1.
-				binding.props.isExact
-				? binding.props.normalizedAnchor
-				: clampNormalizedAnchor(binding.props.normalizedAnchor)
+			? clampNormalizedAnchor(binding.props.normalizedAnchor)
 			: { x: 0.5, y: 0.5 }
 
 		const shapePoint = Vec.Add(point, Vec.MulV(normalizedAnchor, size))


packages/tldraw/src/lib/shapes/arrow/shared.ts
```diff
@@ -78,16 +86,17 @@ export function getArrowTerminalInArrowSpace(
...
-				binding.props.isPrecise || forceImprecise
-					? binding.props.normalizedAnchor
-					: { x: 0.5, y: 0.5 },
-				size
-			)
-		)
+		// If the parent is the bound shape, then it's always treated as precise.
+		const shouldUsePreciseAnchor = binding.props.isPrecise || forceImprecise
+		const normalizedAnchor = shouldUsePreciseAnchor
+			? // Keep exact bindings untouched; only sanitize imprecise/precise anchors to avoid
```
steveruizok: the normalized difference here is negligible... unless I'm missing something, I'd normalize everything

mimecuvalo:
best PR title i've seen in a while

packages/tldraw/src/lib/shapes/arrow/shared.ts
AniKrisn: yep sounds good. made the change

steveruizok:
send it

Issue #7587, opened 2026-01-05
Improve customization of default shapes


steveruizok:
## Problem

tldraw's default shapes (geo, draw, text, arrow, etc.) have limited customization options. Developers frequently need to customize the appearance of default shapes to match their application's branding or design requirements, but the current APIs make this difficult or impossible without creating entirely custom shapes.

### What's currently possible

- Modify color hex values directly via `DefaultColorThemePalette.lightMode.black.solid = 'aqua'`
- Override font families through CSS variables (`--tl-font-draw`, `--tl-font-sans`, etc.)
- Create entirely custom shapes with custom style properties
- Override UI components including the style panel

### What's difficult or impossible

1. **Color palette limitations**
   - Cannot add or remove colors from the palette (limited to 13 hardcoded color names)
   - Cannot customize color variants (solid, semi, pattern, fill, linedFill, etc.) independently
   - Modifying `DefaultColorThemePalette` directly has side effects

2. **Size/font/stroke mappings**
   - Size-to-pixel mappings are hardcoded (`FONT_SIZES`, `STROKE_SIZES`, etc.)
   - Cannot customize what "small" or "large" means for different shape types
   - No way to customize line-height, letter-spacing, etc. per font style

3. **Fill patterns**
   - Pattern fills are hardcoded (8x8 hash pattern)
   - No API to customize or provide alternative patterns

4. **Per-shape customization**
   - Shapes access colors directly via `getColorValue()` in their components
   - No way to customize specific shapes without overriding the entire ShapeUtil

## Proposed solution

Provide a comprehensive styling configuration API that allows customization without creating custom shapes:

### 1. Color palette configuration

```tsx
<Tldraw
  colorConfig={{
    // Extend or replace the color palette
    colors: {
      brand: { light: '#4F46E5', dark: '#818CF8' },
      accent: { light: '#059669', dark: '#34D399' },
      // Override existing colors
      red: { light: '#DC2626', dark: '#F87171' },
    },
    // Customize which colors appear in the UI
    enabledColors: ['black', 'brand', 'accent', 'red', 'blue'],
  }}
/>
```

### 2. Size/typography configuration

```tsx
<Tldraw
  styleConfig={{
    fontSizes: { s: 14, m: 18, l: 24, xl: 32 },
    strokeSizes: { s: 1, m: 2, l: 3.5, xl: 5 },
    fonts: {
      draw: { family: 'Shantell Sans', weight: 500, lineHeight: 1.35 },
[...]

PR #8145, opened 2026-03-03
feat(editor): add first-class theme system with ThemeManager

steveruizok:
In order to provide a proper API for customizing and switching color themes at runtime, this PR replaces the mutable global `DefaultColorThemePalette` with a first-class theme system on the editor. Themes are reactive, scoped to the editor instance, and can be customized or switched at runtime. This also includes a display values system that lets shape utils derive visual properties from the active theme.

Relates to #7587

---

### Theme structure

A theme is an object with an `id`, a `colors` palette, and base typography values:

```ts
interface TLTheme {
  id: string
  colors: TLThemeColors
  fontSize: number   // default 16
  lineHeight: number // default 1.35
}
```

Light and dark modes are **separate top-level themes**, not nested inside a single object. The editor ships with two built-in themes: `'light'` and `'dark'`.

```ts
type TLThemes = Record<string, TLTheme>  // e.g. { light: TLTheme, dark: TLTheme }
```

Each color in the palette has variants for different contexts — `solid`, `semi`, `pattern`, `fill`, `linedFill`, frame variants (`frameHeadingStroke`, `frameHeadingFill`, `frameStroke`, `frameFill`, `frameText`), note variants (`noteFill`, `noteText`), and highlight variants (`highlightSrgb`, `highlightP3`) — plus base properties like `text`, `background`, `solid`, `cursor`, and `noteBorder`.

All default shape font sizes are derived by multiplying the theme's base `fontSize` by per-shape-type multipliers (e.g. `theme.fontSize * FONT_SIZES[size]`), so changing the base value scales all text proportionally. Similarly, `lineHeight` is sourced from the theme rather than a hardcoded constant.

---

### Reading the current theme

`editor.getCurrentTheme()` returns the active `TLTheme`. It's a computed value — components that read it re-render when the theme changes.

```ts
const theme = editor.getCurrentTheme()
theme.colors.blue.solid  // '#4263eb'
theme.fontSize           // 16
theme.lineHeight         // 1.35
theme.id                 // 'light'
```

Use `getColorValue()` to resolve a color style name (like `'red'`) to its value for a specific variant:

```ts
import { getColorValue } from 'tldraw'

const color = getColorValue(theme, 'red', 'solid')  // '#e03131'
```

`editor.getCurrentThemeId()` returns just the ID if you don't need the full palette.

---

### Switching themes

By default, the editor uses the `'light'` theme. You can switch explicitly:

```ts
editor.setTheme('dark')   // force dark
editor.setTheme(null)     // revert to automatic (based on user preference)
```

Or declaratively via the `theme` prop — it's reactive, so changing it updates the editor:

```tsx
<Tldraw theme="dark" />
```

The `useCurrentThemeId()` hook provides a reactive theme ID for use in React components:

```tsx
const themeId = useCurrentThemeId()  // re-renders on theme change
```

---

### Customizing colors

`editor.updateThemes()` merges overrides into the registered theme map:

```ts
const lightTheme = editor.getThemes()['light']
editor.updateThemes({
  light: {
    ...lightTheme,
    colors: {
      ...lightTheme.colors,
      black: { ...lightTheme.colors.black, solid: 'navy' },
    },
  },
})
```

You can also pass a full theme map as a prop. This is useful when your theme is driven by external state (e.g. a settings panel):

```tsx
<Tldraw themes={myThemes} />
```

Both the `theme` and `themes` props are reactive and kept in sync with the editor.

---

### Custom named themes

You can register themes beyond `'light'` and `'dark'`:

```ts
editor.updateThemes({
  'my-brand': {
    id: 'my-brand',
    fontSize: 18,
    lineHeight: 1.4,
    colors: { /* ... */ },
  },
})
editor.setTheme('my-brand')
```

---

### Adding custom colors

The `TLThemeColors` interface can be extended via module augmentation to add custom color names. Colors added this way are type-safe and automatically available in the color style system:

```ts
declare module 'tldraw' {
  interface TLThemeColors {
    pink: TLDefaultColor
  }
}
```

Register the new colors at runtime so the style validators accept them:

```ts
registerColors(['pink'])
```

Then provide the color values in your theme:

```ts
editor.updateThemes({
  light: { ...lightTheme, colors: { ...lightTheme.colors, pink: { solid: '#e91e8c', /* ... */ } } },
})
```

---

### Display values system

Shape utils can now derive visual properties (colors, sizes, fonts) from the active theme via `getDisplayValueOverrides`. This replaces per-shape hooks like `useDefaultColorTheme()`.

For **built-in shapes**, use `getDisplayValueOverrides` on a configured shape util:

```ts
const CustomGeoShapeUtil = GeoShapeUtil.configure({
  getDisplayValueOverrides(editor, shape, theme) {
    if (shape.isLocked) {
      return { fillColor: theme.colors.red.solid }
    }
    return {}
  },
})
```

For **custom shapes**, call `this.editor.getCurrentTheme()` directly:

```ts
class MyShapeUtil extends ShapeUtil<MyShape> {
  component(shape: MyShape) {
    const theme = this.editor.getCurrentTheme()
    const color = getColorValue(theme, shape.props.color, 'solid')
    return <HTMLContainer><div style={{ color }}>Hello</div></HTMLContainer>
  }
}
```

---

### Change type

- [x] `api`

### Test plan

1. Create shapes of all types — verify colors render correctly in both light and dark mode
2. Toggle dark mode — shapes should update their colors
3. Use pattern fills — verify hash patterns render at all zoom levels
4. Export shapes as SVG — verify colors are correct in exported SVG
5. Export shapes as SVG with a custom theme active — verify custom theme colors appear in the export
6. Check the style panel — color picker should show correct color swatches
7. Set `theme` prop — verify explicit theme override works
8. Pass custom `themes` prop — verify custom palettes apply
9. Customize `fontSize` on a theme — verify all shape text scales proportionally
10. Customize `lineHeight` on a theme — verify text line height updates across all shapes
11. Add a custom color via `TLThemeColors` augmentation — verify it appears in the color picker

- [x] Unit tests

### Release notes

- Add `ThemeManager` for managing editor color themes via `editor.updateThemes()` and `editor.getCurrentTheme()`
- Add `TLTheme`, `TLThemeColors`, and `TLThemes` types
- Add `editor.getCurrentThemeId()` and `editor.setTheme()` for theme switching
- Add `useCurrentThemeId()` hook
- Add `theme` and `themes` props on `<Tldraw>` and `<TldrawEditor>`
- Add `themeId` to `SvgExportContext` for correct theme resolution during SVG export
- Add display values system for shape utils to derive visual properties from the active theme
- Add `fontSize` and `lineHeight` to `TLTheme` — all default shape font sizes and line heights are now derived from the theme
- Add `EnumStyleProp.removeValues()` for removing values from enum style props at runtime
- 💥 `TLThemeColorPalette` replaced with extensible `TLThemeColors` interface
- 💥 `TLCustomColorNames` removed — extend `TLThemeColors` directly instead
- 💥 `defaultColorNames` is now internal
- 💥 `getColorStyleItems()` now requires a `TLTheme` parameter
- 💥 `FONT_SIZES`, `LABEL_FONT_SIZES`, `ARROW_LABEL_FONT_SIZES` are now multipliers (relative to `theme.fontSize`) instead of absolute pixel values
- 💥 `TEXT_PROPS.lineHeight` removed — line height now comes from `theme.lineHeight`
- 💥 Remove `DefaultColorThemePalette`, `getDefaultColorTheme()`, `useDefaultColorTheme()`, `useIsDarkMode()`, and `TLDefaultColorTheme`
- 💥 Remove `inferDarkMode` prop

### API changes

- Added `Editor.getCurrentTheme(): TLTheme`
- Added `Editor.getCurrentThemeId(): string`
- Added `Editor.getThemes(): TLThemes`
- Added `Editor.updateThemes(themes: Partial<TLThemes>): void`
- Added `Editor.setTheme(themeId: string | null): this`
- Added `ThemeManager` class
- Added `useCurrentThemeId()` hook
- Added `getColorValue(theme, color, variant)` helper
- Added `TLTheme`, `TLThemeColors`, `TLThemes` types
- Added `TLDefaultColor` type (moved from `TLColorStyle` to `TLTheme`)
- Added `TLTheme.fontSize` and `TLTheme.lineHeight` properties
- Added `DEFAULT_LIGHT_THEME`, `DEFAULT_DARK_THEME` constants
- Added `theme` and `themes` props on `TldrawEditorBaseProps`
- Added `SvgExportContext.themeId` for custom theme support in exports
- Added display values system (`getDisplayValueOverrides`, `ShapeOptionsWithDisplayValues`)
- Added `EnumStyleProp.removeValues()` method
- Changed `getColorStyleItems()` to accept `TLTheme` parameter (previously no params)
- Changed `TLDefaultColorStyle` — now derived from `TLThemeColors` interface keys
- Changed `getColorValue()` first param from `TLDefaultColorTheme` to `TLTheme`
- Breaking! `TLThemeColorPalette` replaced with `TLThemeColors` interface
- Breaking! `TLCustomColorNames` removed — extend `TLThemeColors` via module augmentation instead
- Breaking! `defaultColorNames` no longer exported (internal only)
- Breaking! `FONT_SIZES`, `LABEL_FONT_SIZES`, `ARROW_LABEL_FONT_SIZES` values are now multipliers relative to `theme.fontSize` (previously absolute pixel values)
- Breaking! `TEXT_PROPS` no longer includes `lineHeight`
- Removed `TLDefaultColorTheme`, `TLDefaultColorThemeColor`, `DefaultColorThemePalette`, `getDefaultColorTheme()`, `useDefaultColorTheme()`, `useIsDarkMode()`
- Removed `inferDarkMode` prop from `TldrawEditorBaseProps` and `TLEditorOptions`
- Removed `TLCustomColorNames` interface

### Code changes

| Section         | LOC change       |
| --------------- | ---------------- |
| Core code       | +2485 / -1729    |
| Tests           | +24 / -47        |
| Automated files | +391 / -110      |
| Documentation   | +654 / -50       |
| Templates       | +21 / -12        |

### DIFF
--- a/apps/docs/content/sdk-features/themes.mdx
+++ b/apps/docs/content/sdk-features/themes.mdx
@@ -200,7 +200,7 @@ To change the automatic color scheme preference, update user preferences:
 editor.user.updateUserPreferences({ colorScheme: 'dark' })
 ```
 
-See [User preferences](/docs/user-preferences#dark-mode) for more on color scheme settings.
+See [User preferences](/sdk-features/user-preferences) for more on color scheme settings.
 
 The [useCurrentThemeId](?) hook provides a reactive theme ID for use in React components:
 


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -347,6 +355,9 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this.user = new UserPreferencesManager(user ?? createTLUser(), inferDarkMode ?? false)
 		this.disposables.add(() => this.user.dispose())
 
+		this.theme = new ThemeManager(this, themes)
```
mimecuvalo: nit: this should be `this.themeManager` - `this.theme` makes it sound like it's the theme but it's not

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -944,6 +955,51 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+		this.theme.updateThemes(themes)
+	}
+
+	/**
+	 * Get the current theme for the active color mode (light/dark).
+	 *
+	 * @public
+	 */
+	getCurrentTheme(): TLTheme {
+		return this.theme.getCurrentTheme()
```
mimecuvalo: the naming thing of `this.theme` shows up here - it's strange that it says `this.theme.getCurrentTheme()`, a bit confusing

apps/docs/content/sdk-features/themes.mdx
```diff
@@ -0,0 +1,260 @@
...
+```
+
+See [User preferences](/docs/user-preferences#dark-mode) for more on color scheme settings.
+
+## Customizing colors
+
+Use [Editor#updateThemes](?) to override specific colors in a theme:
+
+```typescript
+editor.updateThemes({
```
mimecuvalo, line 220: i think the code and the docs are mismatched maybe with what i was expecting from the PR description. the PR description says:
```
// Customize colors by merging into the theme map
editor.updateThemes({
  default: {
    color: {
      light: { ...existing, black: { ...existing.black, solid: 'aqua' } },
      dark: existingDark,
    }
  }
})
...
editor.setCurrentTheme('my-brand-theme')
```

so i think the code is missing this stuff, for example, setCurrentTheme is missing and the current code is maybe one step behind the higher-level architecture here. as it stands, the code/docs here are missing the "brand" layer.

another mismatch here, the PR says:
`editor.getCurrentTheme()  // → TLThemeColorPalette`
but this should be:
`editor.getCurrentTheme()  // → TLTheme`

apps/docs/content/sdk-features/themes.mdx
```diff
@@ -0,0 +1,260 @@
...
+export default function App() {
+	return (
+		<Tldraw
+			onMount={(editor) => {
+				// Read the current theme
+				const theme = editor.getCurrentTheme()
+				console.log(theme.colors.red.solid) // e.g. '#e03131'
+
+				// Customize the light theme
+				editor.updateThemes({
```
mimecuvalo, line 40: high-level note: let's make sure we can do this declaratively. if people can only set the theme during runtime, then we'll run into the issue of the flash-of-previous-theme-before-the-next-theme-is-loaded-problem 😅

packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
```diff
@@ -125,6 +128,30 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
...
+							: getColorValue(theme, color, DEFAULT_FILL_COLOR_NAMES[fill]),
+				patternFillFallbackColor: getColorValue(theme, color, 'semi'),
+				labelColor: getColorValue(theme, labelColor, 'solid'),
+				labelFontFamily: FONT_FAMILIES[font],
+				labelFontSize: ARROW_LABEL_FONT_SIZES[size],
+				labelLineHeight: TEXT_PROPS.lineHeight,
+				labelPadding: ARROW_LABEL_PADDING,
+				labelBorderRadius: 3.5,
+			}
+		},
```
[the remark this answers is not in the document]
steveruizok: Fixed in 4d8a2b5b5d. All `getDisplayValues` implementations across 8 shape utils (arrow, draw, geo, highlight, line, text, note, frame) now use the `theme` parameter passed by the wrapper function instead of calling `editor.getCurrentTheme()`. This ensures SVG exports with a different dark mode than the editor's current setting produce correct colors.

packages/editor/src/lib/editor/Editor.ts
steveruizok: This has been addressed — the property is now `_themeManager` (see the resolved thread above). So calls read as `this._themeManager.getCurrentTheme()` internally, and the public API is just `editor.getCurrentTheme()`.

apps/docs/content/sdk-features/themes.mdx
steveruizok: This is supported declaratively. Both `<Tldraw>` and `<TldrawEditor>` accept a `theme` prop (the active theme ID) and a `themes` prop (the named color themes map). Setting `theme="dark"` at render time avoids any flash, and custom themes can be passed in on initial render too. The docs cover this at lines 191-195 and 234-253 of themes.mdx.

apps/docs/content/sdk-features/themes.mdx
steveruizok: I reviewed the docs against the code — they're consistent:

- The API is `editor.setTheme()` (not `setCurrentTheme`), and the docs correctly use it at line 184.
- `getCurrentTheme()` returns `TLTheme` (not `TLThemeColorPalette`), and the docs correctly show the full shape (`id`, `colors`, `fontSize`, `lineHeight`).
- The "brand" theme layer from the PR description isn't implemented — the current design uses `'light'` and `'dark'` as theme IDs, with `setTheme()` to switch and `updateThemes()` to customize.

The mismatch was between the PR description and the actual implementation, not between the code and docs. The PR description may need updating to reflect the current design.

Also fixed the `getDisplayValueOverrides` example in the docs to use the `theme` parameter instead of `editor.getCurrentTheme()` (same pattern as the shape util fix).

steveruizok:
Continuing on a _clean copy_ https://github.com/tldraw/tldraw/pull/8216

Issue #7742, opened 2026-01-25
Add user-facing revision history UI


steveruizok:
## Problem statement

tldraw already has robust revision history infrastructure (snapshots stored in R2, Pierre-based versioning, restore endpoints), but the UI at `/f/{id}/history` is currently only intended for debugging. Users would benefit from a polished interface to browse and restore previous versions of their files.

## Proposed solution

Build a user-facing revision history feature with:

### UI/UX requirements

1. **History panel or page** - Browse snapshots with timestamps, grouped by date
2. **Preview mode** - View historical snapshots in a read-only editor before restoring
3. **Restore action** - One-click restore with confirmation dialog
4. **Linear history model** - Restoration copies the historical snapshot as a new head (no branching), keeping history simple and linear

### Access control

Consider gating this feature to higher-tier plans (e.g., Pro Plus / "tldraw extreme"). This provides value for power users while keeping the free tier simple.

### Technical notes

The backend infrastructure already exists:
- History list endpoints: `/api/f/:roomId/history`, `/api/f/:roomId/pierre-history`
- Snapshot retrieval: `/api/f/:roomId/history/:timestamp`
- Restore endpoint: `/app/file/:roomId/restore`
- Current debug UI: `apps/dotcom/client/src/tla/pages/file-history.tsx`

The restore operation already follows the linear model - it loads a snapshot and saves it as the new current state.

## Design considerations

- How to surface revision history (sidebar, menu option, dedicated page?)
- Snapshot preview UX (thumbnail vs full editor view?)
- Granularity of history display (individual saves, hourly/daily groupings?)
- Visual indicators for significant changes between versions
- Mobile experience

steveruizok:
## Codebase research

### Existing infrastructure

The revision history backend is fully implemented with dual storage systems:

**API endpoints** (in `apps/dotcom/sync-worker/src/worker.ts:82-102`):
- `GET /api/f/:roomId/history` - List history with pagination
- `GET /api/f/:roomId/history/:timestamp` - Get snapshot at timestamp
- `GET /api/f/:roomId/pierre-history` - Git-based history (Pierre system)
- `POST /app/file/:roomId/restore` - Restore to previous version

**Storage architecture**:
- **R2 ephemeral bucket** (`ROOMS_HISTORY_EPHEMERAL`): Timestamped snapshots at `app_rooms/{fileId}/{timestamp}`
[...]

PR #8146, opened 2026-03-04
fix(sync-worker): use custom tar reader for Pierre history snapshot to avoid Workers deadlock

ds300:
In order to fix the Pierre history snapshot endpoint hanging in Cloudflare Workers, this PR replaces `modern-tar` with a minimal USTAR stream reader and keeps the response streaming to avoid OOM for large files.

`modern-tar` yields `entry.body` as a nested `ReadableStream` that is fed in a later `transform()` call. In Workers that causes a backpressure deadlock: the pipe does not push more input while the consumer is blocked on `entry.body`, so the body never receives data. The custom reader (`tarStreamReader.ts`) uses a single consumer, buffers as needed, and yields entry bodies as `Uint8Array`s — no nested streams, no deadlock. The route streams the snapshot as JSON via `streamPierreSnapshotAsJson` so the full response is never held in memory.

Relates to #7742.

### Change type

- [x] `bugfix`
- [ ] `improvement` | `feature` | `api` | `other`

### Test plan

1. Deploy sync-worker and open a file that has Pierre history.
2. Request a history snapshot at a valid commit (e.g. from the revision history UI or direct API call).
3. Confirm response is valid JSON with `documentClock`, `schema`, and `documents` array; request should complete (no hang).

- [x] Unit tests (tar stream reader: fixture-based tests in `tarStreamReader.test.ts`)
- [ ] End to end tests

### Release notes

- No user-facing change; fixes sync-worker Pierre history snapshot endpoint in Cloudflare Workers.

---

> [!NOTE]
> **Medium Risk**
> Adds a custom USTAR tar parser and swaps out the archive decoding path for Pierre history snapshots, which could affect snapshot correctness for unusual tar variants. Scope is limited to the Pierre snapshot/archive streaming flow and is covered by new unit tests/fixtures.
> 
> **Overview**
> Fixes Pierre history snapshot responses hanging in Cloudflare Workers by removing `modern-tar` and decoding Pierre’s gzipped tar archives via a new single-consumer USTAR reader (`readTarStream`) that yields file bodies as `Uint8Array`s (no nested `ReadableStream` backpressure).
> 
> Updates `iteratePierreArchive`/`streamPierreSnapshotAsJson` to use the new reader and ensures the JSON stream closes cleanly, and adds `vitest` coverage plus committed tar fixtures + a script to regenerate them. Dependency cleanup removes `modern-tar` from `package.json`/`yarn.lock`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a422f887172411ac0f2933ccd8e01a96ecb4d99b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mimecuvalo:
oh man, wild. @ds300 is this something to consider fixing upstream? perhaps as an option on their end?

ds300:
> oh man, wild. @ds300 is this something to consider fixing upstream? perhaps as an option on their end?

sorry the PR description didn't make it clear that this is a problem with the workers runtime, not with pierre or even modern-tar. but yes it does need fixing. gonna file an issue.

Issue #6115, opened 2025-05-22
Add avatar field for user


yi-boide:
### What's the feature?

I want to add an avatar field in the user. When there is an avatar, I will display the nickname, avatar and background first. Only when there is none will the nickname, color and background be displayed. Add avatar in the TLUserPreferences type? Add avatar: T.string.nullable().optional() to the string | null userTypeValidator validation.

### Where would you like this feature?

The developer package

### Contact Details

454846659@qq.com

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

steveruizok:
This issue was reopened as it was closed by the stale bot automation, but active development is ongoing.

**Related work:**
- PR #6117 (feat: Add the user avatar function) - authored by @yi-boide addressing this exact feature
- PR #7008 (feat: add user avatar sync in share panel) - comprehensive implementation with full API changes including avatar fields in TLUserPreferences, TLPresenceUserInfo, TLInstancePresence, and new UserPreferencesManager.getAvatar() method

The feature is actively being implemented and should not have been closed due to stale automation.

PR #8147, opened 2026-03-04
feat(editor): add shape attribution, TLUserStore, and extensible user records

mimecuvalo:
In order to let developers track who created and last edited shapes — and to display that information in the UI — this PR introduces a shape attribution system built on a reactive `TLUserStore` provider interface, note-specific "first edited by" labeling, and extensible user records via `createTLSchema`.

### What's included

**`TLUserStore` interface** (`@tldraw/tlschema`)
A new optional `users` prop on the editor (and `<Tldraw />`) that connects tldraw to your auth system. Both methods return reactive `Signal`s so the editor automatically tracks changes to user data:
- `getCurrentUser()` — returns a `Signal<TLUser | null>` for the active user (used for both presence broadcasting and shape attribution)
- `resolve(userId)` — returns a `Signal<TLUser | null>` to resolve any user ID to display info (for rendering attribution labels)

When no provider is given, a default implementation falls back to `UserPreferences` + the collaborator presence list.

**Unified `TLUser` record type** (`@tldraw/tlschema`)
Collapsed the previous `TLUser` + `TLUserInfo` types into a single `TLUser` store record (following the `TLAsset` pattern). User records are document-scoped and persist alongside shapes, assets, and pages — surviving across boards, clipboard paste, and `.tldr` files.

**Extensible user metadata via `createTLSchema`**
SDK users can now extend user records with custom validated metadata:
```ts
const schema = createTLSchema({
  user: {
    meta: {
      isAdmin: T.boolean,
      department: T.string,
    },
  },
})
```
Custom meta fields are validated when present but treated as optional, so existing user records remain valid. The new `createUserRecordType(config?)` function builds the record type with extended validators.

**Note shape "first edited by" label**
`NoteShapeUtil` now tracks `textFirstEditedBy` in its props and renders a small attribution label (first name) in the bottom-right corner of the note. The label resolves the display name through the user store and shows a tooltip with the full name. This tracks who first added text to the note — subsequent typo fixes by others don't change the attribution. Includes a note shape migration (`AddFirstEditedBy`).

Screenshot 2026-03-04 at 11 20 58

**New editor methods**
- `editor.getAttributionUser(userId)` — resolves a user record by ID, asking the `TLUserStore` first (the app's source of truth), falling back to the `user:` record in the store
- `editor.getAttributionUserId()` — returns the current user's ID for attribution purposes
- `editor.getAttributionDisplayName(userId)` — resolves a display name for a user ID, with the same lookup order

**New "Users" example category with four examples**
- **Attribution** — demonstrates wiring up `TLUserStore` with a mock user switcher and an attribution inspector panel
- **Attribution timeline** — shows a history scrubber that steps through undo history and displays per-shape attribution metadata
- **Custom user metadata** — shows how to extend `TLUser` records with custom meta fields (isAdmin, department)
- **Multiplayer sync with custom user data** — integrates custom user data into tldraw sync

https://github.com/user-attachments/assets/bd476596-67ca-483b-8d1d-bec3f82e35d5

https://github.com/user-attachments/assets/c31f8db6-c7ae-4e21-b0d5-48e079eb5767

fixes https://github.com/tldraw/tldraw/issues/6115

**Unit tests**
New `attribution.test.ts` covering attribution user stamping, custom user stores, note `textFirstEditedBy` tracking, and display name resolution.

### Change type

- [x] `feature`

### Test plan

1. Open the Attribution example — switch between users and create/edit shapes; verify attribution labels update in the inspector panel
2. Open the Attribution Timeline example — draw shapes, use the scrubber to step through history, verify per-shape metadata display
3. Open the Custom User Metadata example — switch between users and verify custom meta (department, admin badge) displays
4. Create note shapes — type text, verify the first name appears in the bottom-right corner; hover for full name tooltip
5. Copy a note with attribution to a different board — verify the attribution label persists

- [x] Unit tests

### API changes

- Added `TLUserStore` interface (with reactive `Signal`-based methods)
- Added `TLUser` record type (unified from previous `TLUser` + `TLUserInfo`)
- Added `UserRecordType`, `createUserId`, `isUserId`, `userIdValidator`
- Added `createUserRecordType(config?)` for extensible user schemas
- Added `UserSchemaInfo` interface
- Added `user` parameter to `createTLSchema()`
- Added `textFirstEditedBy` prop to `TLNoteShapeProps`
- Added `Editor.getAttributionUser()`
- Added `Editor.getAttributionUserId()`
- Added `Editor.getAttributionDisplayName()`
- Removed `TLUserInfo` type (collapsed into `TLUser`)

### Release notes

- Add shape attribution system with `TLUserStore` for connecting tldraw to your auth system.
- `TLUserStore` methods are reactive (`Signal`-based) so the editor automatically tracks user data changes.
- Add extensible user records: pass custom meta validators to `createTLSchema({ user: { meta: ... } })`.
- Add `createUserRecordType()` for building user record types with custom validation.
- Note shapes now display a small "first edited by" label in the corner.
- New "Users" example category with attribution, attribution timeline, custom user metadata, and sync examples.

### DIFF
--- a/packages/tlschema/src/migrations.test.ts
+++ b/packages/tlschema/src/migrations.test.ts
@@ -2616,25 +2616,6 @@ describe('Adding textFirstEditedBy to note shape', () => {
 	})
 })
 
-describe('Converting note textFirstEditedBy to ID string', () => {
-	const { up, down } = getTestMigration(noteShapeVersions.FirstEditedByToId)
-
-	test('up converts object to ID string', () => {
-		const result = up({ props: { textFirstEditedBy: { id: 'user-1', name: 'Alice' } } })
-		expect(result.props.textFirstEditedBy).toBe('user-1')
-	})
-
-	test('up is a no-op for null', () => {
-		const result = up({ props: { textFirstEditedBy: null } })
-		expect(result.props.textFirstEditedBy).toBeNull()
-	})
-
-	test('down converts ID string back to object', () => {
-		const result = down({ props: { textFirstEditedBy: 'user-1' } })
-		expect(result.props.textFirstEditedBy).toEqual({ id: 'user-1', name: '' })
-	})
-})
-
 describe('TLUser initial migration', () => {
 	const { up } = getTestMigration(userVersions.Initial)
 
--- a/packages/tlschema/src/shapes/TLNoteShape.ts
+++ b/packages/tlschema/src/shapes/TLNoteShape.ts
@@ -147,7 +147,6 @@ const Versions = createShapePropsMigrationIds('note', {
 	AddRichText: 9,
 	AddRichTextAttrs: 10,
 	AddFirstEditedBy: 11,
-	FirstEditedByToId: 12,
 })
 
 /**
@@ -277,19 +276,6 @@ export const noteShapeMigrations = createShapePropsMigrationSequence({
 			},
 			down: (props) => {
 				delete props.textFirstEditedBy
-			},
-		},
-		{
-			id: Versions.FirstEditedByToId,
-			up: (props: any) => {
-				if (props.textFirstEditedBy && typeof props.textFirstEditedBy === 'object') {
-					props.textFirstEditedBy = props.textFirstEditedBy.id
-				}
-			},
-			down: (props: any) => {
-				if (typeof props.textFirstEditedBy === 'string') {
-					props.textFirstEditedBy = { id: props.textFirstEditedBy, name: '' }
-				}
 			},
 		},
 	],


steveruizok:
Thanks for kicking this off Mime. Couple of design questions as I was reading the diff:

- What's the advantage of the `tlmeta` property, vs. putting these properties on the root of the shape? I can understand avoiding collisions with the meta property. We have many places where we isEqual based on props or meta changing; if there's renderable information here (such as the attributed user's name), then we'd have to add this one to that list, too.
- Is the identity system meant to be reactive? IIRC we use a reactive user in user preferences to handle things like live name changes or color changes.
- Why did you choose to make the `createdBy` `editedBy` contain the full user object (i.e. id, name), as opposed to an ID that references a centralized list of users? There's a comment about copy and paste but I think that could be resolved by including attributed users in the copy and paste object. Are there other angles there?

mimecuvalo:
> Thanks for kicking this off Mime. Couple of design questions as I was reading the diff:

yeah, we should discuss when you're back — there's some architecture/design questions to resolve.
- `tlmeta` felt better in the sense that i could easily see us adding a bunch more meta-properties in the future — was trying to predict future direction here. but i'd be also down with putting it at the top-level, i don't feel strongly about it. and, unfortunately, `meta` was already taken heh.
- the TLIdentity piece is a direction/idea we could take — putting it in here to discuss. I could leave out of this PR to be honest, but we need to start paving the path with what we'll do in regards to permissions and all that. 
- "Why did you choose to make the createdBy editedBy contain the full user object (i.e. id, name)"
   - yeah, b/c if you paste content onto another board, say an incognito/logged out board, it should still work/have a fallback. this is some experience gleaned from my Paper work — it can feel broken if pasting something into a new board starts degrading/omitting bits you'd expect, it can feel lossy.
   - furthermore, a centralized list of users changes over time: people leave companies/permissions change — and you don't want that to affect the experience.
   - finally finally, with bots being a possibility on the board, they aren't necessarily part of a centralized list of users — you still want a way to be able to have a display name for something created by "Fairy Twinklebot"

will put time next week for us to sit down and explore concepts introduced here.

steveruizok:
> > Thanks for kicking this off Mime. Couple of design questions as I was reading the diff:
> 
> yeah, we should discuss when you're back — there's some architecture/design questions to resolve.
> 
> * `tlmeta` felt better in the sense that i could easily see us adding a bunch more meta-properties in the future — was trying to predict future direction here. but i'd be also down with putting it at the top-level, i don't feel strongly about it. and, unfortunately, `meta` was already taken heh.
> * the TLIdentity piece is a direction/idea we could take — putting it in here to discuss. I could leave out of this PR to be honest, but we need to start paving the path with what we'll do in regards to permissions and all that.
> * "Why did you choose to make the createdBy editedBy contain the full user object (i.e. id, name)"
>   
>   * yeah, b/c if you paste content onto another board, say an incognito/logged out board, it should still work/have a fallback. this is some experience gleaned from my Paper work — it can feel broken if pasting something into a new board starts degrading/omitting bits you'd expect, it can feel lossy.
>   * furthermore, a centralized list of users changes over time: people leave companies/permissions change — and you don't want that to affect the experience.
>   * finally finally, with bots being a possibility on the board, they aren't necessarily part of a centralized list of users — you still want a way to be able to have a display name for something created by "Fairy Twinklebot"
> 
> will put time next week for us to sit down and explore concepts introduced here.

Right on, I think the questions are:

- do we expect to add more meta properties?
- is it better to create a new property or slide something into meta, like `__tldraw`: {}`.
- how should we handle identities provided externally, if a way that:
  - handles changes to users?
  - handles attributions to people who are no longer discoverable?
  - handles situations where content is brought into a file from somewhere else?

I can think of a way to do the "fallback" for users who are no longer discoverable, or while copy and pasting, by including the name and other user info in a table keyed by id. We'd just never clean up this table, so any user that passes through, or who is mentioned, would be added to the table. Even if it wasn't a reactive thing, we could have a way for the application to update the user table (pushing a change to a username, etc) and a step during mount where users are updated or resolved (pulling fresh usernames, etc).

When copying content, we could include relevant users from that table (e.g. folks who are attributed) and merge that into the pasted-to document's list of known users. Maybe they're strangers who can't be resolved by the host document / app, maybe not. It does make me think that if the user meta is extensible, which it should be, then we may need to tolerate content coming in with properties that the host document / app doesn't know how to use.

mimecuvalo:
all good topics for in-person! since we're using PRs in lieu of design docs, this makes it a little awkward to discuss these broader/finer points over this medium — there's no threading in Github PRs for "concepts introduced by a PR" :-/

apps/dotcom/client/src/utils/csp.ts
```diff
@@ -75,5 +75,10 @@ export const csp = Object.keys(cspDirectives)
 
 export const cspDev = Object.keys(cspDirectives)
 	.filter((key) => key !== 'report-uri')
-	.map((directive) => `${directive} ${cspDirectives[directive].join(' ')}`)
+	.map((directive) => {
+		const values = cspDirectives[directive]
+		// We allow data: urls for frame-src to allow debugging SVG embeds in dev.
+		if (directive === 'frame-src') return `${directive} ${[...values, 'data:'].join(' ')}`
```
mimecuvalo, line 81: drive-by fix, only for dev mode

mimecuvalo:
@ds300 and @steveruizok can you start taking a look at this PR? also happy to chat in person to debate any choices here.
i did go for some breaking changes over trying to have backwards compatibility for 5.0, that might be worth a discussion.

ds300:
**Overall:** This is an ambitious PR that introduces a lot of new concepts (TLUserStore, document-scoped TLUser records, shape attribution metadata, note-level attribution labels, extensible user schemas). The direction is interesting but there are some architectural concerns I'd like to see addressed before this lands, particularly around the always-on nature of stamping, the data model for user records, and the note attribution UX.

### Attribution stamping is always-on — should it be?

`meta.__tldraw` is unconditionally written on every `createShapes` and `_updateShapes` call, even when no user store is configured (all fields become `null` with `Date.now()` timestamps). This means:

1. **Every shape update generates a new `updatedAt`**, even for position/size changes. In multiplayer, this inflates diff payloads with metadata changes that are noise for most use cases.
2. **There's a per-operation cost**: `getAttributionUserId()` → `getCurrentUser()` → `_ensureUserRecord()` (store read + comparison) runs on every single shape create/update.
3. **SDK users who don't care about attribution** still pay these costs and carry the data.

Should this be opt-in? For example, attribution stamping could only activate when a `TLUserStore` (or `users` prop) is explicitly provided. The `Date.now()` timestamps could be gated too — if nobody's consuming them, they're just churn.

### `meta.__tldraw` vs a top-level optional property

Putting system metadata inside `meta` — which is documented as the user-controlled escape hatch — creates an implicit reservation of the `__tldraw` key. A few concerns:

- `meta` is no longer purely user-owned space. Developers could accidentally overwrite `__tldraw` and break attribution.
- The validator has to handle the case where `meta` contains both user-defined keys and the system `__tldraw` key.
- If this were a top-level optional property (e.g. `attribution?: TLShapeAttribution`), it would be type-safe, wouldn't collide with user meta, and wouldn't require an up-migration for existing shapes (only a down-migration). Steve raised this question too — I think a top-level optional property is the cleaner path.

### Document-scoped TLUser records and staleness

User records are written into the document store via `_ensureUserRecord` whenever a shape is created/updated. These records persist indefinitely with no cleanup mechanism. This means:

- **Stale names**: If a user changes their display name, every document they previously touched still has the old name until they create/update another shape there.
- **Deleted users**: Records for users who've been removed from the organization persist forever.
- **Growing user tables**: In high-traffic collaborative documents, the user record list grows monotonically.

`getAttributionDisplayName` checks the store _before_ calling `TLUserStore.resolve()`, so stale store records take priority over fresh data from the application's auth system. Should the lookup order be reversed? Or should there be a mount-time "resolve all known users" step, as Steve suggested?

### TLUserStore doesn't match the presence system's reactivity model

The existing presence system is reactive: `createPresenceStateDerivation` takes a `Signal<TLUser>`, so when user name/color changes, presence updates propagate automatically. `TLUserStore` is a plain synchronous interface — `getCurrentUser()` returns a snapshot, not a signal. These two systems now have mismatched contracts for representing the same user.

In `useSync.ts`, a `computed` wraps `getCurrentUser()`, but since `getCurrentUser()` doesn't read from any `@tldraw/state` signal, the computed has no reactive dependencies from the user store side. The presence system will re-derive when user data changes; the attribution system won't. This means a name change could update your cursor label but leave stale attribution on shapes you're actively editing.

I think we should align `TLUserStore` with the presence system's reactive model. Maybe even combine the APIs somehow.

### Note "last edited by" label — authorship vs last editor

Is 'Last edited by' the best label for this? I think we should consider showing the creator of the shape rather than the last editor, or rather whoever added content to the note first.

mimecuvalo:
> **Overall:** This is an ambitious PR that introduces a lot of new concepts (TLUserStore, document-scoped TLUser records, shape attribution metadata, note-level attribution labels, extensible user schemas). The direction is interesting but there are some architectural concerns I'd like to see addressed before this lands, particularly around the always-on nature of stamping, the data model for user records, and the note attribution UX.

Thanks for the review @ds300 ! I think I forgot that you weren't included in some of the offline conversations we had about things like `__tldraw`. I'm gonna put a sync tomorrow for us to solidify some choices here. I might ask you to possibly spin up a child PR to exhibit some of the proposed changes you'd like to see.

mimecuvalo:
high level takeaways from our discussion:
- we're going to nix `__tlmeta` it's actually not necessary at this time and complicates the PR. if people want `createdAt`/`updatedAt`/etc they can add it themselves. it bloats the shape as well unnecessarily. furthermore, we don't actually even consume it ourselves, we sort of just added it as it felt perfunctory necessary.
- add `Signal` to properties/fns on `TLUserStore`. I'll take a look again also at the presence info and see how that does some things, see if we can utilize some patterns there.
- `getAttributionDisplayName` / `TLUserStore.resolve()` resolve calling order
- `textLastEditedBy` -> `textFirstEditedBy` on `NoteShapeUtil` b/c it felt like it doesn't matter later if someone fixed a typo - it matters who first grabbed the sticky note

@ds300 @steveruizok and @kostyafarber i'm gonna create a child PR to flesh this feedback out. @steveruizok lemme know if you have any high-level thoughts on the points above / have questions

steveruizok:
> ### Attribution stamping is always-on — should it be?
> `meta.__tldraw` is unconditionally written on every `createShapes` and `_updateShapes` call, even when no user store is configured (all fields become `null` with `Date.now()` timestamps). This means:
> 
> 1. **Every shape update generates a new `updatedAt`**, even for position/size changes. In multiplayer, this inflates diff payloads with metadata changes that are noise for most use cases.
> 2. **There's a per-operation cost**: `getAttributionUserId()` → `getCurrentUser()` → `_ensureUserRecord()` (store read + comparison) runs on every single shape create/update.
> 3. **SDK users who don't care about attribution** still pay these costs and carry the data.
> 
> Should this be opt-in? For example, attribution stamping could only activate when a `TLUserStore` (or `users` prop) is explicitly provided. The `Date.now()` timestamps could be gated too — if nobody's consuming them, they're just churn.

The overhead doesn't sound _too_ bad. It would add 4 properties to shapes, assets, bindings. These would also get smooshed, when we compact the messages we send to the server. We could debounce those but then we'd have to send a followup to the server. Even `createdBy: null` would be noise but I think that may be acceptable, too.

> ### `meta.__tldraw` vs a top-level optional property
> Putting system metadata inside `meta` — which is documented as the user-controlled escape hatch — creates an implicit reservation of the `__tldraw` key. A few concerns:
> 
> * `meta` is no longer purely user-owned space. Developers could accidentally overwrite `__tldraw` and break attribution.
> * The validator has to handle the case where `meta` contains both user-defined keys and the system `__tldraw` key.
> * If this were a top-level optional property (e.g. `attribution?: TLShapeAttribution`), it would be type-safe, wouldn't collide with user meta, and wouldn't require an up-migration for existing shapes (only a down-migration). Steve raised this question too — I think a top-level optional property is the cleaner path.

I'm also open to having this be a top-level set of properties. This "put it in meta" was my idea, motivated by how often we check for `props` or `meta` in isEqual checks, and how we would now also be checking for `tlmeta` or whatever. Its true that while we usually spread previous meta into new meta, there's no guarantee that downstream users will do that.

> ### Document-scoped TLUser records and staleness
> User records are written into the document store via `_ensureUserRecord` whenever a shape is created/updated. These records persist indefinitely with no cleanup mechanism. This means:
> 
> * **Stale names**: If a user changes their display name, every document they previously touched still has the old name until they create/update another shape there.
> * **Deleted users**: Records for users who've been removed from the organization persist forever.
> * **Growing user tables**: In high-traffic collaborative documents, the user record list grows monotonically.
> 
> `getAttributionDisplayName` checks the store _before_ calling `TLUserStore.resolve()`, so stale store records take priority over fresh data from the application's auth system. Should the lookup order be reversed? Or should there be a mount-time "resolve all known users" step, as Steve suggested?

IIRC we defer management of this to the application. So when the document loads, we should be asking for clarification on all of the users known to the document, receiving a response from the application, and either updating or deleting records in the document based on the response. Maybe we need to also track / send a "last active" date, allowing the application to (possibly) prune stale records according to its own logic.

> ### TLUserStore doesn't match the presence system's reactivity model
> The existing presence system is reactive: `createPresenceStateDerivation` takes a `Signal<TLUser>`, so when user name/color changes, presence updates propagate automatically. `TLUserStore` is a plain synchronous interface — `getCurrentUser()` returns a snapshot, not a signal. These two systems now have mismatched contracts for representing the same user.
> 
> In `useSync.ts`, a `computed` wraps `getCurrentUser()`, but since `getCurrentUser()` doesn't read from any `@tldraw/state` signal, the computed has no reactive dependencies from the user store side. The presence system will re-derive when user data changes; the attribution system won't. This means a name change could update your cursor label but leave stale attribution on shapes you're actively editing.
> 
> I think we should align `TLUserStore` with the presence system's reactive model. Maybe even combine the APIs somehow.

I had some thoughts on this as well and suggested that the signal was actually less useful than having some callbacks that the application could manage, though open to alternative designs.

IIRC, when an editor mounts, the system would first check for new information (from the application) about all of the application users known to the document, and then the application would set up a subscription to respond when any of those application users changed in a way that mattered to their tldraw identity. That subscription would need to be updated when new users become known to the editor. I guess this could be done by providing a signal and managing the data in it; I was thinking of the record being managed inside of tldraw with callbacks as the API for handling external changes.

> ### Note "last edited by" label — authorship vs last editor
> Is 'Last edited by' the best label for this? I think we should consider showing the creator of the shape rather than the last editor, or rather whoever added content to the note first.

There are some odd edge cases here, like duplicating someone else's note and then editing it. We could do something like "clearing the text in a label resets the attributed user".

steveruizok:
> high level takeaways from our discussion:
> 
> * we're going to nix `__tlmeta` it's actually not necessary at this time and complicates the PR. if people want `createdAt`/`updatedAt`/etc they can add it themselves. it bloats the shape as well unnecessarily. furthermore, we don't actually even consume it ourselves, we sort of just added it as it felt perfunctory necessary.
> * add `Signal` to properties/fns on `TLUserStore`. I'll take a look again also at the presence info and see how that does some things, see if we can utilize some patterns there.
> * `getAttributionDisplayName` / `TLUserStore.resolve()` resolve calling order
> * `textLastEditedBy` -> `textFirstEditedBy` on `NoteShapeUtil` b/c it felt like it doesn't matter later if someone fixed a typo - it matters who first grabbed the sticky note
> 
> @ds300 @steveruizok and @kostyafarber i'm gonna create a child PR to flesh this feedback out. @steveruizok lemme know if you have any high-level thoughts on the points above / have questions

got me while I was writing up

mimecuvalo:
PTAL @kostyafarber @ds300 @MitjaBezensek

kostyafarber (approved):
nice!

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -9246,6 +9370,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 				...Object.fromEntries(
 					content.bindings?.map((bindings) => [bindings.id, bindings] as const) ?? []
 				),
+				...Object.fromEntries(content.users?.map((user) => [user.id, user] as const) ?? []),
```
MitjaBezensek, line 9363: Do we do something with migrated users? Seems like they are just dropped? Doesn't seem an issue now, but if we add a migration the migration might be skipped?

packages/sync/src/useSync.ts
```diff
@@ -170,52 +165,68 @@ export function useSync(opts: UseSyncOptions & TLStoreSchemaOptions): RemoteTLSt
...
 				}
-			}
-		)
+			: {
+					getCurrentUser: defaultUserStore.getCurrentUser,
+					resolve: (userId) => {
+						const current = defaultUserStore.getCurrentUser()
+						if (current && current.id === createUserId(userId)) return current
+						const presences = store.query.records('instance_presence').get()
+						const match = presences.find((p) => p.userId === userId)
```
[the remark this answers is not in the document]
MitjaBezensek: Think this is a valid error.

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -3972,6 +3985,93 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+
+	/**
+	 * Ensure a user record exists in the store for the given user,
+	 * updating it if the data has changed.
+	 *
+	 * @internal
+	 */
+	_ensureUserRecord(user: TLUser): void {
+		const existing = this.store.get(user.id)
+		if (
```
MitjaBezensek, line 4012: Shouldn't we include `meta` here? Otherwise we might miss some data changes. Like what if a user goes from `meta.isAdmin = false` to `meta.isAdmin = true`?

packages/tldraw/setupVitest.js
```diff
@@ -1,5 +1,38 @@
 import { vi } from 'vitest'
 
+// Snapshot serializer that normalizes unstable tlmeta attribution fields.
```
MitjaBezensek: Think this should be removed?

packages/tlschema/src/TLStore.ts
```diff
@@ -200,6 +201,97 @@ export interface TLAssetStore {
...
+ * }
+ * ```
+ */
+export interface TLUserStore {
+	/**
+	 * Return a signal resolving to the currently authenticated user,
+	 * or `null` for anonymous / unknown.
+	 * Called when stamping attribution on shape create/update.
+	 */
+	getCurrentUser(): Signal<TLUser | null>
```
MitjaBezensek: Should we use a property here instead? Could make it a bit cleaner and would avoid the caching requirement. Think we do something similar for `TLCurrentUser .userPreferences`?

PR #8151, opened 2026-03-04
feat(tldraw): add frame selection action

### LINKED ISSUE #6949: Add quick option to frame highlighted objects
### What's the feature?

Should be able to switch between group <> frame like Figma does also.

Image

The developer package

### Contact Details

zb@clickup.com

### Code of Conduct

AniKrisn:
`Cmd+G` toggles grouping: it groups selected shapes, or ungroups if a group is selected. Similarly, this PR makes `Cmd+Shift+F` toggle framing: it wraps selected shapes in a frame, or removes the frame if one is selected. Previously, `Cmd+Shift+F` could only remove frames. Frame padding scales with the zoom level so it works well with dynamic sizing.

Not sure if 'frame-selection' is a clear enough name.

Closes #6949

Edit: following comments, changed the shortcut from `Cmd+Shift+F` to `Opt+Shift+G`.

Edit: added context menu option.

image

### Change type

- [x] `feature`

### Test plan

1. Select 2+ shapes on the canvas
2. Press Cmd+Shift+F — shapes should be wrapped in a frame
3. With the frame selected, press Cmd+Shift+F again — the frame should be removed
4. Right-click → Edit → "Frame selection" should also appear and work
5. Undo/redo should work correctly for both operations

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add "Frame selection" action to wrap selected shapes in a frame (Cmd+Shift+F), toggling with remove frame when a frame is selected.

### API changes

- Added `FrameSelectionMenuItem` component
- Added `frame-selection` action
- Added `frame-selection` event to `TLUiEventMap`
- Added `action.frame-selection` translation key
- Removed keyboard shortcut from `remove-frame` action (now handled by `frame-selection` toggle)


mimecuvalo:
this is a pretty cool idea 💡

steveruizok:
I'd follow Figma here and use Cmd+Shift+G. Take a look through, there might be other things we're doing that don't have kbds.

image

edit - or Cmd + S, since our frames are more like sections... somewhere between them, actually.

steveruizok:
cmd shift g - to ungroup / unframe
opt shift g - frame shapes

AniKrisn:
(Claude):
Did a pass on this. Two angles: other **toggle pairs** (like this one and group/ungroup), and plain **shortcut gaps**.

**Toggles are basically covered.** group⇄ungroup (`cmd+g`), frame⇄unframe (this PR, `alt+shift+g`), and lock⇄unlock (`shift+l`) are all already smart toggles — there's no remaining clean inverse-pair to merge. (`remove-frame` keeps no shortcut now and survives as a menu item only, which is correct.)

**The real gap is the arrange/align family being half-covered.** `align-*` and `distribute-*` have shortcuts, but their siblings in the same menu don't:

- `stretch-horizontal` / `stretch-vertical` — no kbd
- `stack-horizontal` / `stack-vertical` — no kbd
- `pack` — no kbd

These are the cleanest "feature exists, just missing the shortcut" candidates and would round out the family. The only fiddly part is key selection — the `alt+shift+*` range is fairly crowded (`alt+shift+h`/`alt+shift+v` are already distribute), so we'd need to pick non-conflicting combos.

I'd suggest keeping this PR scoped to framing and doing stretch/stack/pack as a small follow-up. Happy to open that if we want it.

steveruizok:
(Sorry, do whatever shortcut Figma has)

PR #8182, opened 2026-03-09
add cmd+shift+a action to select shapes that are only in your viewport

kostyafarber:
I'd like to be able to select all shapes that are currently in my viewport. Often I work on large boards and want to quickly select all shapes with the keyboard - but only the ones I am currently working on in the viewport.

Also thought an editor method is useful here, but could just inline this logic in the action if we don't want to add another method to the Editor api.

I'm not sure if it needs to be added in the context menus etc (i.e exist only as a shortcut), but have decided to add them in this PR.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add a new method to the editor to select all shapes in the viewport

### API changes
- Add a new method to the editor to select all shapes in the viewport

---

> [!NOTE]
> **Medium Risk**
> Adds a new public `Editor` API plus a new keyboard shortcut/menu action, which could subtly change selection behavior and collide with existing shortcuts, but the implementation is small and covered by unit tests.
> 
> **Overview**
> Adds `Editor.selectAllInViewport()` to select **only unlocked shapes currently visible in the viewport** (using viewport bounds + spatial index).
> 
> Wires this into the UI as a new `select-all-in-viewport` action with `cmd+shift+a,ctrl+shift+a`, exposes it in the main menu and context menu, and updates translations / API reports / analytics event typing.
> 
> Adds unit tests covering viewport visibility, camera panning, locked shapes, and a regression case where a shape util’s `canCull()` returns `false`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c431b71c6b43437181ec9de09d31e3b1c3288416. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/assets/translations/main.json
+++ b/assets/translations/main.json
@@ -70,6 +70,7 @@
 	"action.rotate-cw": "Rotate clockwise",
 	"action.save-copy": "Save a copy",
 	"action.select-all": "Select all",
+	"action.select-all-in-viewport": "Select all in viewport",
 	"action.select-none": "Select none",
 	"action.send-backward": "Send backward",
 	"action.send-to-back": "Send to back",
--- a/packages/tldraw/src/lib/ui/hooks/useTranslation/TLUiTranslationKey.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useTranslation/TLUiTranslationKey.ts
@@ -74,6 +74,7 @@ export type TLUiTranslationKey =
 	| 'action.rotate-cw'
 	| 'action.save-copy'
 	| 'action.select-all'
+	| 'action.select-all-in-viewport'
 	| 'action.select-none'
 	| 'action.send-backward'
 	| 'action.send-to-back'
--- a/packages/tldraw/src/lib/ui/hooks/useTranslation/defaultTranslation.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useTranslation/defaultTranslation.ts
@@ -75,6 +75,7 @@ export const DEFAULT_TRANSLATION = {
 	'action.rotate-cw': 'Rotate clockwise',
 	'action.save-copy': 'Save a copy',
 	'action.select-all': 'Select all',
+	'action.select-all-in-viewport': 'Select all in viewport',
 	'action.select-none': 'Select none',
 	'action.send-backward': 'Send backward',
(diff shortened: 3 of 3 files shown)


mimecuvalo:
Neat idea — we should chat about it. I'd be worried about "domain parking" in the sense of Cmd+Shift+A might be useful to have available for another type of quick selection. For example, maybe because I'm in attribution mode, it'd be cool to be able to just quickly select just my shapes on a board vs. including my shapes+everyone elses. 
We should check that a) this is something people reach for maybe when on a board (viewport selection) and b) that we don't have some other type of selection that would benefit from this "prime" real estate :P

steveruizok:
I think this could be useful. I'm stingy as hell about keyboard shortcuts though, especially when there isn't strong consensus about the shortcut between other apps. Here's a quick survey of how this shortcut is used in other apps:

```md
Figma

⌘ + ⇧ + A → Add Auto Layout

Wraps the selected objects in an Auto Layout frame so they stack and space automatically.

Sketch

⌘ + ⇧ + A → Align to artboard (center)

Aligns the selected layers relative to the artboard.

Adobe Illustrator

⌘ + ⇧ + A → Select Same Appearance

Selects all objects that share the same appearance properties.

Adobe Photoshop

⌘ + ⇧ + A → Camera Raw Filter

Opens the Camera Raw editing panel for image adjustments.

General pattern in modern UI design tools

In newer tools (especially Figma-like tools), the shortcut is typically associated with layout automation, e.g.:

Add Auto Layout

Convert selection to layout container

Apply stack / flex-like layout

So in modern interface design workflows, people most often mean:

👉 Cmd + Shift + A = “Add Auto Layout” (Figma)
```

Thinking further:

If we had game-style keyboard customization (something I've always wanted to add in the eventual preference panel) this would be a great action to have that has no keyboard shortcut associated with it by default. We could add it to the library and not assign any UI or kbd to actually engage it, but have it available for users who wanted to assign a shortcut to it.

steveruizok:
closing with care

Issue #8187, opened 2026-03-09
Pasting copied shape into an editing text shape does nothing


AniKrisn:
When you copy a text shape (select it, Cmd+C) and then double-click into another text shape to edit it, pressing Cmd+V does nothing. The paste is silently dropped.

Pasting onto the canvas (not editing a shape) works fine — it creates a new shape. It's specifically pasting while editing a text shape that fails.

By contrast, pasting Excalidraw content while editing a text shape dumps the raw JSON clipboard data into the shape as text.

### Steps to reproduce

**tldraw shape paste:**
1. Create two text shapes
2. Select the first text shape and copy it (Cmd+C)
3. Double-click the second text shape to start editing it
4. Press Cmd+V — nothing happens

**Excalidraw paste:**
1. Copy a text element from Excalidraw
2. Double-click a text shape in tldraw to start editing it
3. Press Cmd+V — raw JSON is pasted into the shape

### Expected behavior

In both cases, the text content should be extracted and pasted as text into the shape being edited.

### Actual behavior

- tldraw shapes: paste is silently dropped
- Excalidraw content: raw JSON string is pasted

### Context

This happens because tldraw's copy puts custom shape data on the clipboard (not standard `text/html` or `text/plain`). When editing a text shape, the global paste handler in `useClipboardEvents.ts` bails out early (expecting TipTap to handle it natively), but TipTap doesn't understand tldraw's custom clipboard format, so the paste is silently dropped.

For Excalidraw, the clipboard contains JSON as `text/plain`, so TipTap pastes it literally as text.

Fixing this would mean detecting when pasting tldraw (or Excalidraw) shape data while editing a text shape, extracting the text content from the copied shape(s), and inserting it into the TipTap editor.

PR #8192, opened 2026-03-10
fix(text): paste plain text instead of blocking paste in editable shapes

steveruizok:
When pasting tldraw clipboard data into an editable text shape (e.g. a geo shape label or note), the paste was silently blocked. This PR fixes it to extract and insert the plain text content from the clipboard instead.

fixes https://github.com/tldraw/tldraw/issues/8187

### Change type

- [x] `bugfix`

### Test plan

1. Create two shapes with text (e.g. geo shapes with labels)
2. Select and copy one shape (Cmd+C)
3. Double-click the other shape to edit its text
4. Paste (Cmd+V) — the plain text from the copied shape should be inserted

### Release notes

- Fix pasting into editable text shapes when clipboard contains tldraw data

---

> [!NOTE]
> **Low Risk**
> Low risk: a small, localized change to paste handling for editable text, with minimal impact outside the specific tldraw-clipboard paste path.
> 
> **Overview**
> Fixes paste into editable text shapes when the clipboard contains tldraw HTML data. Instead of silently blocking the paste, `useEditablePlainText` now prevents the default and inserts the clipboard’s `text/plain` content via `document.execCommand('insertText')` to preserve undo support.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0fa808bd71217aeff4e82fa428b37533fac210fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/useEditablePlainText.ts
+++ b/packages/tldraw/src/lib/shapes/shared/useEditablePlainText.ts
@@ -156,6 +156,7 @@ export function useEditableTextCommon(shapeId: TLShapeId) {
 						const plainText = e.clipboardData.getData('text/plain')
 						preventDefault(e)
 						if (plainText) {
+							// eslint-disable-next-line @typescript-eslint/no-deprecated -- best way to insert text with undo support
 							document.execCommand('insertText', false, plainText)
 						}
 					}


mimecuvalo:
:request changes: :P

steveruizok:
> mm, this is trickier maybe - it loses any rich text if you do this. so, if the other shape has some bolding, for example, it'll be lost in the paste

This is a little more complicated. We'd need to parse the tldraw data to extract the rich text and then paste it in.

steveruizok:
> > mm, this is trickier maybe - it loses any rich text if you do this. so, if the other shape has some bolding, for example, it'll be lost in the paste
> 
> This is a little more complicated. We'd need to parse the tldraw data to extract the rich text and then paste it in.

as a followup

PR #8194, opened 2026-03-10
feat(mermaid): add @tldraw/mermaid package for diagram-to-shape conversion

### LINKED ISSUE #4353: [Feature]: Paste Mermaid diagram as text and automatically draw it with tldraw
### What's the feature?

Sometimes I have Mermaid diagrams but I want to make small visual adjustments for presentations. Or I just want to write small annotations which is hard to do with a Mermaid syntax. 
I would like to be able to paste my Mermaid diagram as text into tldraw and have it render it as boxes, arrows, etc.
I'm not very familiar with Mermaid diagrams and I mostly use it for simple graphs. Not sure if other kinds of diagram like sequence diagrams would be hard ot support.

The tldraw.com website

### Contact Details

_No response_

### Code of Conduct

kaneel:
In order to let users paste Mermaid diagram syntax and have it converted into native tldraw shapes, this PR introduces the `@tldraw/mermaid` package and integrates it into the dotcom app.
Closes #4353

### How it works

The system has three layers: **detection**, **parsing + layout extraction**, and **blueprint rendering**.

**Detection (dotcom integration).** When text is pasted on dotcom, `SneakyMermaidHandler` registers an external text content handler on the editor. Before pulling in the heavy mermaid library, it runs a lightweight regex-based check (`simpleMermaidStringTest`) that strips YAML frontmatter, `%%{...}%%` directives, and `%%` comments, then tests whether the cleaned text starts with a known diagram keyword. This keeps the mermaid library out of the main bundle — it's only `import()`-ed when a paste actually looks like a diagram. If the text doesn't match, it falls through to `defaultHandleExternalTextContent`. If the diagram type is unsupported (e.g. pie charts), the handler falls back to SVG import via `putExternalContent({ type: 'svg-text' })` and shows a toast.

**Parsing + layout extraction (`createMermaidDiagram`).** This is the main entry point. It initializes mermaid with inflated font sizes to compensate for tldraw's hand-drawn font being wider, parses and validates the syntax, renders to an offscreen DOM element for layout extraction via `getBBox()`, then extracts the diagram's semantic data (vertices, edges, actors, states, etc.) and dispatches to the appropriate diagram-specific converter.

**Diagram-specific converters.** Each converter combines the semantic data (from mermaid's DB) with layout data (from the rendered SVG) to produce a `DiagramMermaidBlueprint`:

- *Flowcharts* — Node positions/dimensions from SVG `.node` elements, cluster bounds from `.cluster` elements, edge waypoints from `path[data-points]` attributes. Maps mermaid shape types to tldraw geo types. Builds subgraph hierarchy as frames with `parentId`. Computes arrow bend from waypoint geometry. Parses `classDef` CSS colors and maps to nearest tldraw palette color via Euclidean RGB distance.
- *Sequence diagrams* — Actor layouts from SVG rect elements with fixed spacing. Creates top/bottom actor shapes, vertical lifelines, signal arrows with precise bindings, note shapes, and fragment frames for loop/alt/opt/par/critical blocks.
- *State diagrams* — Recursively flattens the state hierarchy. Creates frames for compound states, handles special types (start, end, fork/join, choice). Reuses the same SVG node/edge parsing strategy as flowcharts.
**Blueprint rendering (`renderBlueprint`).** The blueprint is a diagram-type-agnostic intermediate representation with `nodes`, `edges`, `lines`, and `groups`. The renderer centers the diagram at the paste point, creates shapes in parent-first order with proper coordinate transforms, creates arrows with bindings (precise anchored, self-loop, and standard center-to-center), groups shapes, and applies a final position correction pass.

### Change type

- [x] `feature`
- [x] `api`

### Test plan

1. Paste a flowchart mermaid diagram (e.g. `graph TD; A-->B; B-->C;`) into tldraw.com — shapes should appear
2. Paste a sequence diagram — shapes with lifelines and arrows should appear
3. Paste a state diagram — shapes with transitions should appear
4. Paste an unsupported diagram type (e.g. pie chart) — should fall back to SVG with a warning toast
5. Paste normal text — should behave as before (no mermaid handling)

- [x] Unit tests

### API changes

- Added `createMermaidDiagram(editor, text, options?)` — main entry point to parse and render a Mermaid diagram
- Added `renderBlueprint(editor, blueprint, opts?)` — renders a `DiagramMermaidBlueprint` into tldraw shapes
- Added `MermaidDiagramError` — error class for parse/unsupported diagram errors
- Added types: `DiagramMermaidBlueprint`, `MermaidBlueprintGeoNode`, `MermaidBlueprintEdge`, `MermaidBlueprintLineNode`, `BlueprintRenderingOptions`, `MermaidDiagramOptions`

### Release notes

- Add Mermaid diagram support: paste Mermaid syntax to create native tldraw shapes (flowcharts, sequence diagrams, state diagrams)

### Code changes

| Core code       | +2661 / -0    |
| Tests           | +964 / -0     |
| Automated files | +176 / -0     |
| Documentation   | +1246 / -0    |
| Apps            | +99 / -0      |
| Config/tooling  | +1252 / -43   |

---

> [!NOTE]
> **Medium Risk**
> Adds a new diagram parsing/rendering subsystem and hooks it into editor paste handling, which could impact editor stability/perf and external-content security surface (SVG/text parsing). Scope is large but mostly additive behind lightweight detection and fallback behavior.
> 
> **Overview**
> Adds Mermaid-to-tldraw conversion support via a new `@tldraw/mermaid` package that parses Mermaid text, extracts layout from rendered SVG, and renders a diagram-agnostic “blueprint” into native tldraw shapes (nodes/lines/arrows/groups) for supported diagram types.
> 
> Integrates this into dotcom editors (`LocalEditor` and `TlaEditor`) with a `SneakyMermaidHandler` that detects Mermaid-ish pasted text without bundling Mermaid eagerly, lazily `import()`s the converter on demand, and falls back to importing SVG + showing a warning toast for unsupported diagram types.
> 
> Updates app deps/tsconfig references, adds i18n strings for the new warning toast, and introduces an examples use-case showcasing large-scale Mermaid rendering.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d3b07f184ed02bdccec528a72957d1f2e97b727. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/mermaid/src/sequenceDiagram.ts
+++ b/packages/mermaid/src/sequenceDiagram.ts
@@ -539,7 +539,6 @@ export function sequenceToBlueprint(
 		events.push(msg)
 	}
 
-	const eventCount = events.length
 	const layouts = layout.actorLayouts
 
 	// Pre-compute lifecycle event indices for created/destroyed actors.


apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
```diff
@@ -0,0 +1,23 @@
...
+ * (from mermaid v11.12.2 src/diagram-api/detectType.ts and src/diagram-api/regexes.ts).
+ * Strips YAML frontmatter, %%{...}%% directives, and %% comments, then tests for
+ * a known diagram keyword at the start of the cleaned text.
+ *
+ * This file intentionally has zero imports so it can be loaded statically without
+ * pulling in the heavy mermaid library.
+ */
+const FRONTMATTER_REGEX = /^-{3}\s*[\n\r]([\s\S]*?)[\n\r]-{3}\s*[\n\r]+/
+const DIRECTIVE_REGEX =
+	/%{2}{\s*(?:(\w+)\s*:|(\w+))\s*(?:(\w+)|((?:(?!}%{2}).|\r?\n)*))?\s*(?:}%{2})?/gi
```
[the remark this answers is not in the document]
kaneel: false positive it seems?

packages/mermaid/src/flowchartDiagram.ts
```diff
@@ -0,0 +1,172 @@
...
+			h,
+			geo,
+			parentId: nodeToSubGraph.get(id),
+			label: vertex.text ? sanitizeDiagramText(vertex.text) : undefined,
+			...(fillColor && { fill: 'solid' as const, color: fillColor }),
+			align: 'middle',
+			verticalAlign: 'middle',
+			size: 'm',
+		} satisfies MermaidBlueprintGeoNode)
+	}
```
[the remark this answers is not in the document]
kaneel: hmmm, we could guard that but this actually rings true in tests because of JSDOM, I'll switch to do browser tests for vitest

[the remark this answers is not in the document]
kaneel: hmmm, we could guard that but this actually rings true in tests because of JSDOM, I'll switch to do browser tests for vitest

packages/mermaid/LICENSE.md
```diff
@@ -0,0 +1,9 @@
+MIT License
```
mimecuvalo: i think we just remove this file - we're going to add a reference in package.json to LICENSE.md

apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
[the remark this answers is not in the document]
mimecuvalo: nope! this is def. a bug!

apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Lightweight mermaid detection replicating mermaid's own detectType preprocessing
+ * (from mermaid v11.12.2 src/diagram-api/detectType.ts and src/diagram-api/regexes.ts).
```
mimecuvalo, line 3: can you add a full link to the file in their repo? that'd be helpful

apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Strips YAML frontmatter, %%{...}%% directives, and %% comments, then tests for
```
mimecuvalo, line 8: we should consider adding tests for these regexs (to make sure they pass/fail)
if the Mermaid repo have tests for these regex's then i'm less worried

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -0,0 +1,63 @@
...
+						type: 'svg-text',
+						text: svgString,
+						point: content.point,
+						sources: content.sources,
+					})
+				}
+
+				const result = await createMermaidDiagram(editor, content.text, { onUnsupportedDiagram })
+				selectNewShapes()
+				if (result === null) {
```
mimecuvalo: hm `result` is a `Promise<void>` isn't result always null?

apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx
```diff
@@ -0,0 +1,191 @@
...
+				borderRadius: '0 0 8px 8px',
+				display: 'flex',
+				gap: '8px',
+				zIndex: 1000,
+				opacity: isGenerating ? 0 : 1,
+			}}
+		>
+			<TldrawUiButton type="low" onClick={handleClick}>
+				Click to see a thousand mermaids
+				{count > 0 && <>({count} actually…)</>}
```
mimecuvalo, line 184: very cool. 
do we need to have the button work a 2nd time? i'd consider clearing the canvas and regenerating if that's the intent

apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx
```diff
@@ -0,0 +1,191 @@
...
+			currentY = 0
+
+		try {
+			for (const group of mermaidDefinitions) {
+				currentX = 0
+				let maxRowHeight = 0
+
+				for (const def of group) {
+					const shapesBefore = new Set(editor.getCurrentPageShapeIds())
+					let nativeSize = { w: 0, h: 0 }
```
mimecuvalo, line 95: one tweak here: would love to have it wrap sooner (rather than going far off to the right offscreen)

apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx
```diff
@@ -0,0 +1,191 @@
...
+							editor.setSelectedShapes(nativeIds)
+						}
+						nativeSize = measureSelection(editor)
+						currentX += nativeSize.w + PAIR_GAP
+					} catch (e) {
+						console.warn('[mermaid] blueprint failed:', e, '\n---\n' + def)
+					}
+
+					try {
+						const { svg } = await mermaid.render(
```
mimecuvalo, line 116: doesn't have to be for this PR but still would love to see a subset (10?) of these mermaid diagrams being used in a screenshot test. basically a smoke test to make sure these don't regress

packages/mermaid/src/blueprint.ts
```diff
@@ -0,0 +1,67 @@
...
+	TLDefaultDashStyle,
+	TLDefaultFillStyle,
+	TLDefaultHorizontalAlignStyle,
+	TLDefaultSizeStyle,
+	TLDefaultVerticalAlignStyle,
+	TLGeoShapeGeoStyle,
+} from 'tldraw'
+
+/** @public */
+export interface DiagramMermaidBlueprint {
```
mimecuvalo, line 20: love it. quick comment here would be useful for future maintainer's context

packages/mermaid/src/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+			maxDistance = distance
+		}
+	}
+
+	const bend = maxDistance * BEND_SCALE
+	return Math.max(-MAX_ARROW_BEND, Math.min(MAX_ARROW_BEND, bend))
+}
+
+/** Normalize HTML line breaks to newlines, decode HTML entities, and trim. */
+export function sanitizeDiagramText(text: string): string {
```
mimecuvalo, line 36: huh i'm surprised we don't have this helper function in dom.ts or text.ts
can you put this function in text.ts? and let's be sure to replace the other ones specified in `COMMON_ENTITY_MAP`

packages/mermaid/src/renderBlueprint.ts
```diff
@@ -0,0 +1,389 @@
...
+		const y = parent ? absoluteY - (offsetY + parent.y) : absoluteY
+
+		editor.createShape<TLGeoShape>({
+			id: shapeId,
+			type: 'geo',
+			x,
+			y,
+			parentId: parentShapeId,
+			props: {
+				geo: node.geo,
```
mimecuvalo, line 103: pretty sure we can do just:
```
props: {
   ...omit(node, 'richText'),
   ...(node.label && { richText: toRichText(node.label) }),
}
```

packages/mermaid/src/sequenceDiagram.ts
```diff
@@ -0,0 +1,731 @@
...
+import { parseRgbToTldrawColor } from './colors'
+import { getAccumulatedTranslate } from './svgParsing'
+import { sanitizeDiagramText } from './utils'
+
+type MermaidLinetype = SequenceDB['LINETYPE']
+type MermaidPlacement = SequenceDB['PLACEMENT']
+
+function isSignalMessage(type: number | undefined, lineType: MermaidLinetype): boolean {
+	if (type === undefined) return false
+	const signalTypes: number[] = [
```
mimecuvalo: this should be defined at the top-level

mimecuvalo: we're passing around `db.LINETYPE` but i don't think that should be passed around 🤔

packages/mermaid/src/sequenceDiagram.ts
```diff
@@ -0,0 +1,731 @@
...
+	return type === lineType.AUTONUMBER
+}
+
+/** Returns the fragment keyword (e.g. "loop", "opt") if this message starts a combined fragment, or null. */
+function getFragmentStartKeyword(
+	type: number | undefined,
+	lineType: MermaidLinetype
+): string | null {
+	if (type === undefined) return null
+	if (type === lineType.LOOP_START) return 'loop'
```
mimecuvalo: we should prefer here a map at the top level (for simplicity/less verboseness in the logic)
so:
```
const LINE_TYPE_MAP = {
  [LINETYPE.ALT_START]: 'alt',
  [LINETYPE.OPT_START]: 'opt',
  etc.
}
```

packages/mermaid/src/sequenceDiagram.ts
```diff
@@ -0,0 +1,731 @@
...
+	if (type === lineType.CRITICAL_START) return 'critical'
+	if (type === lineType.BREAK_START) return 'break'
+	if (type === lineType.PAR_OVER_START) return 'par'
+	return null
+}
+
+function isFragmentEnd(type: number | undefined, lineType: MermaidLinetype): boolean {
+	if (type === undefined) return false
+	const endTypes: number[] = [
+		lineType.LOOP_END,
```
mimecuvalo: same here, we shouldn't pass these lineTypes around, so something at the top-level defined as `const END_TYPES = [LINETYPE.ALT_END, etc.]`

packages/mermaid/src/sequenceDiagram.ts
```diff
@@ -0,0 +1,731 @@
...
+		const regularBottoms = bottom.filter((r) => !actorManSet.has(r))
+		const refHeight =
+			regularTops.length > 0 ? Math.max(...regularTops.map((r) => r.h)) : FALLBACK_ACTOR_HEIGHT
+		const refWidth =
+			regularTops.length > 0 ? Math.max(...regularTops.map((r) => r.w)) : FALLBACK_ACTOR_WIDTH
+		const refTopY = regularTops.length > 0 ? Math.min(...regularTops.map((r) => r.y)) : undefined
+		const refBottomEndY =
+			regularBottoms.length > 0 ? Math.max(...regularBottoms.map((r) => r.y + r.h)) : undefined
+
+		for (const r of top) {
```
mimecuvalo: nit: avoid single length vars, hard to follow along

packages/mermaid/src/stateDiagram.ts
```diff
@@ -0,0 +1,469 @@
...
+		states,
+		relations
+	)
+
+	// Collect notes attached to states and add them as synthetic leaf nodes + edges.
+	for (const [id, state] of states) {
+		const note = (state as StateStmt & { note?: { position?: string; text: string } }).note
+		if (!note) continue
+
+		const noteId = `${id}----note`
```
mimecuvalo, line 277: let's make this defined somewhere - hard to tell that this is tied with a check below

packages/mermaid/src/flowchartDiagram.ts
```diff
@@ -0,0 +1,219 @@
...
+	subGraphs?: FlowSubGraph[]
+): DiagramMermaidBlueprint {
+	const nodeFillMap = parseClassDefFills(root.outerHTML, 'flowchart-', FLOW_KNOWN_CLASSES)
+	const svgNodes = parseNodesFromSvg(root, '.node', (domId) => {
+		const match = domId.match(/^flowchart-(.+)-\d+$/)
+		return match ? match[1] : domId
+	})
+	const svgClusters = parseClustersFromSvg(root, '.cluster')
+	const svgEdges = parseAllEdgePointsFromSvg(root, (dataId) => {
+		const match = dataId.match(/^L_(.+)_([^_]+)_\d+$/)
```
mimecuvalo, line 96: nit: example of what this regex is matching in the real world, pls

packages/mermaid/src/colors.ts
```diff
@@ -0,0 +1,233 @@
...
+	knownClasses: Set<string>
+): Map<string, TLDefaultColorStyle> {
+	const result = new Map<string, TLDefaultColorStyle>()
+
+	const styleMatch = svgString.match(/<style[^>]*>([\s\S]*?)<\/style>/i)
+	if (!styleMatch?.[1]) return result
+	const css = styleMatch[1]
+
+	const classColors = new Map<string, { fill?: string; stroke?: string }>()
+	const ruleRegex = /\.([\w-]+)\s*(?:>|[\s])\s*(?:\*|\w+)\s*\{([^}]*)\}/g
```
mimecuvalo: oof these regexes are gnarly - please put some good comments+examples if we really need to use them

mimecuvalo: if possible, please reuse any useful regexs in `parseCss.ts`
you'll note in that file that it references canonical regexes found in MDN

packages/mermaid/src/colors.ts
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	if (classColors.size === 0) return result
+
+	const escapedPrefix = idPrefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+	const idRegex = new RegExp(`\\bid="${escapedPrefix}(.+)-\\d+"`)
+	const tagRegex = /<g\b([^>]*)>/g
+	let tagMatch: RegExpExecArray | null
+	while ((tagMatch = tagRegex.exec(svgString)) !== null) {
```
mimecuvalo: instead of all this gnarly regex stuff. can we explore trying doing something like:
```
const doc = document.implementation.createHTMLDocument('')
doc.documentElement.innerHTML = html.trim()
...
doc.body.querySelector('#someId .className')
etc. etc.
```
parsing HTML via regex is just gonna be a mess 😬

packages/mermaid/src/colors.ts
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	return result
+}
+
+export function parseRgbToTldrawColor(
+	text: string
+): { color: TLDefaultColorStyle; hasAlpha: boolean } | null {
+	const match = text.match(/rgba?\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*(?:,\s*([\d.]+)\s*)?\)/)
```
mimecuvalo: comment

packages/mermaid/src/createMermaidDiagram.ts
```diff
@@ -0,0 +1,136 @@
...
+				break
+			}
+			case 'sequence': {
+				const db = diagramResult.db as SequenceDB
+				const blueprint = sequenceToBlueprint(
+					liveSvg,
+					db.getActors(),
+					db.getActorKeys(),
+					db.getMessages(),
+					db.LINETYPE,
```
mimecuvalo: yeah, please let's not pass these around as parameters (LINETYPE nor PLACEMENT)

packages/mermaid/src/createMermaidDiagram.ts
```diff
@@ -0,0 +1,136 @@
...
+				const db = diagramResult.db as StateDB
+				const states = db.getStates()
+				const relations = db.getRelations()
+				const blueprint = stateToBlueprint(liveSvg, states, relations)
+				renderBlueprint(editor, blueprint, options.blueprintRender)
+				break
+			}
+			default:
+				await options.onUnsupportedDiagram?.(parsedSvg)
+		}
```
mimecuvalo, line 147: to make this clearer:
```
let blueprint
...
if (blueprint) {
  renderBlueprint(editor, blueprint, options.blueprintRender)
}
```
instead of having to repeat that for every case

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
kaneel: Yes, this one is a weird regression I must say.

apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx
kaneel: hard agree, I can still make that happen within the example itself I think!

packages/mermaid/src/sequenceDiagram.ts
kaneel: Agreed, I've let that slid while I was trying to get sequences to yield good results but it's not proper, I need to get those type and make them usable enums for comparison

packages/mermaid/src/flowchartDiagram.ts
kaneel: oops, it seems that after I've asked the bots to reorganise the code they just decided not to keep all the comments we had there, that's DEFINITELY something I need to remember on next prompts.

apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx
kaneel: It does clean and regenerate! sometimes it gets a bit stuck though so I added a timer before regenerating… it could be improved 🤔

apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
kaneel: I can think of some tests, yep!

packages/mermaid/src/colors.ts
```diff
@@ -0,0 +1,219 @@
...
+			bestDistance = distance
+			best = name
+		}
+	}
+	return best
+}
+
+function stripHash(value: string | undefined): string | undefined {
+	return value?.startsWith('#') ? value.slice(1) : undefined
+}
```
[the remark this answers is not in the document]
kaneel: did a full refactor of this

packages/mermaid/src/renderBlueprint.ts
```diff
@@ -0,0 +1,373 @@
...
+					bend: -80,
+				},
+			})
+			editor.createBindings([
+				makeArrowBinding(arrowId, startShapeId, 'start', { x: 0.9, y: 0.5 }, false, false),
+				makeArrowBinding(arrowId, endShapeId, 'end', { x: 0.85, y: 0.8 }, false, false),
+			])
+		})
+		return arrowId
+	}
```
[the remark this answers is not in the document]
kaneel: this is intentional for now but if I throw claude at claude: 
> Summary: The -80 hardcode is a reasonable constant that produces a visually appropriate arc for the specific top-center → right-center self-loop geometry used in tldraw. Using edge.bend would be worse — ranging from invisible (state diagrams) to unpredictable (flowcharts). If anything, a future improvement might compute self-loop bend from the shape dimensions rather than from SVG waypoints, but -80 is a sensible static default.

packages/mermaid/src/flowchartDiagram.ts
```diff
@@ -0,0 +1,231 @@
...
+		return match ? match[1] : domId
+	})
+	const svgClusters = parseClustersFromSvg(root, '.cluster')
+	// Mermaid edge data-id attributes look like "L_startId_endId_0".
+	// Group 1 = start node id, group 2 = end node id (using [^_]+ so it
+	// stops at the trailing _digit suffix).
+	const svgEdges = parseAllEdgePointsFromSvg(root, (dataId) => {
+		const match = dataId.match(/^L_(.+)_([^_]+)_\d+$/)
+		return match ? { start: match[1], end: match[2] } : null
+	})
```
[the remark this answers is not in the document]
kaneel: >  the issue as filed is not a real bug. The regex is technically imprecise, but it has zero effect on output because the code never reads the misparsed values.

packages/mermaid/src/sequenceDiagram.ts
```diff
@@ -0,0 +1,819 @@
...
+	for (let i = 0; i < events.length; i++) {
+		const ev = events[i]
+		if (!isSignalMessage(ev.type)) continue
+
+		if (ev.to && createdActors.has(ev.to) && !creationEventIndex.has(ev.to)) {
+			creationEventIndex.set(ev.to, i)
+		}
+		if (ev.from && destroyedActors.has(ev.from) && !destructionEventIndex.has(ev.from)) {
+			destructionEventIndex.set(ev.from, i)
+		}
```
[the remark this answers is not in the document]
kaneel: Let's focus on those post release

packages/mermaid/src/utils.ts
```diff
@@ -0,0 +1,74 @@
...
+	if (typeof text !== 'string') return ''
+	const el = document.createElement('div')
+	el.innerHTML = text.replace(/<br\s*\/?>/gi, '\n')
```
mimecuvalo: this is a security risk here, either use `new DOMParser().parseFromString()` or we can the pattern in `useClipboardEvents.ts` → `stripHtml`

PR #8196, opened 2026-03-10
refactor(editor): use scoped document/window for cross-window embedding support

ds300:
In order to support embedding tldraw in environments where the global `document` and `window` differ from the ones tldraw is mounted in (iframes, Electron pop-out windows, Obsidian plugins), this PR replaces bare `document` and `window` references with container-aware alternatives throughout `@tldraw/editor`, `@tldraw/tldraw`, and `@tldraw/utils`.

### Core infrastructure

- Add exported `getOwnerDocument()` and `getOwnerWindow()` helpers in `domUtils.ts`
- Add `Editor.getContainerDocument()` and `Editor.getContainerWindow()` convenience methods
- Add `getComputedStyle()` helper that uses the element's owner window

### DOM context fixes (~70 files)

- **Event listeners**: attach to `container.ownerDocument` / `editor.getContainerWindow()` instead of bare `document` / `window`
- **Element creation**: use scoped document for `createElement`, `createElementNS`
- **Font management**: use `editor.getContainerDocument().fonts`
- **Active element checks**: use `editor.getContainerDocument().activeElement`
- **`instanceof` guards**: use window-scoped constructors (`getOwnerWindow().HTMLElement`, etc.)
- **`getComputedStyle`**: use the domUtils helper or call on the correct window
- **`matchMedia`**: use `editor.getContainerWindow().matchMedia()`
- **`getSelection`**: use `editor.getContainerWindow().getSelection()`
- **`devicePixelRatio`**: use `editor.getContainerWindow().devicePixelRatio` or read from instance state
- **`visualViewport`**: use `editor.getContainerWindow().visualViewport`
- **Print**: use `editor.getContainerWindow()` for `print()`, `addEventListener('beforeprint')`
- **`MediaHelpers`**: accept optional `doc` parameter so callers can pass the correct document
- **SVG export**: thread document context through `FontEmbedder`, `StyleEmbedder`, and `getSvgJsx`

### ESLint rules to prevent regressions

Added lint rules for `packages/editor`, `packages/tldraw`, and `packages/utils`:
- `no-restricted-globals`: bare `document`, `getComputedStyle`, `matchMedia`, `getSelection`, `devicePixelRatio`, `innerWidth`, `innerHeight`, `visualViewport`
- `no-restricted-syntax`: `window.addEventListener/removeEventListener/getComputedStyle/getSelection/matchMedia/navigator/print/devicePixelRatio/innerWidth/innerHeight/scrollX/scrollY/visualViewport`, and bare `instanceof` checks against DOM constructors (`HTMLElement`, `SVGElement`, `Element`, `ClipboardItem`, etc.)
- `no-restricted-syntax` turned off for test files (alongside existing `no-restricted-globals` and `no-restricted-properties` overrides)

### Bug fixes

- Fix `areShortcutsDisabled` regression in `useDocumentEvents.ts`: the cross-window refactor changed `activeElementShouldCaptureKeys()` from its default (`includeButtonsAndMenus=true`) to explicitly passing `false`, which caused `<select>` and `<button>` elements to no longer suppress keyboard shortcuts when focused
- Fix `useFixSafariDoubleTapZoomPencilEvents.ts` to use cross-window `instanceof` checks for `PointerEvent` and `Element`
- Fix `FontEmbedder` to use cross-window constructors for `CSSFontFaceRule` and `CSSImportRule` instanceof checks
- Restore missing `elementShouldCaptureKeys` export in `api-report.api.md`
- Fix `api-report.api.md` parameter name (`ignoreButtons` → `includeButtonsAndMenus`) to match implementation
- Update test mocks (`TextManager`, `FocusManager`, `FontManager`) to include `getContainerDocument`
- Remove unused eslint-disable directives in e2e tests and `storage.test.ts`
- Fix Prettier formatting in `Tldraw.tsx`
- Update CI lint job to `actions/checkout@v6` (was v3, matching the test and build jobs)
- Fix ESM build: add `.mjs` extensions to dynamic `import()` expressions in the `addJsExtensions` post-build step (previously only static imports and re-exports were rewritten)

### Change type

- [x] `improvement`

### Test plan

1. Run `yarn dev` and verify the editor loads and works normally
2. Test basic interactions: drawing, selecting, erasing, text editing
3. Test clipboard operations (copy/paste)
4. Test SVG/PNG export
5. Test print functionality
6. Test minimap rendering
7. Test rich text toolbar (text selection positioning)
8. Verify keyboard shortcuts are disabled when `<select>` or `<button>` elements are focused
9. If possible, test in an iframe or cross-window context to verify scoped document/window usage

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improve cross-window embedding support by using scoped `document` and `window` references throughout the SDK, enabling tldraw to work correctly in iframes, Electron pop-out windows, and browser extension contexts.
- Fixed dynamic `import()` expressions in ESM builds missing `.mjs` file extensions.

### API changes

- Added `Editor.getContainerDocument()` for accessing the document where the editor is mounted
- Added `Editor.getContainerWindow()` for accessing the window where the editor is mounted
- Added `getOwnerDocument()` helper for getting the correct document from a DOM node
- Added `getOwnerWindow()` helper for getting the correct window from a DOM node
- Changed `activeElementShouldCaptureKeys()` to accept optional `doc` parameter
- Changed `downloadFile()` to accept optional `doc` parameter
- Changed `preloadFont()` to accept optional `targetDocument` parameter
- Changed `MediaHelpers.loadVideo()` to accept optional `doc` parameter
- Changed `MediaHelpers.getImageAndDimensions()` to accept optional `doc` parameter
- Changed `MediaHelpers.getVideoSize()` to accept optional `doc` parameter
- Changed `MediaHelpers.getImageSize()` to accept optional `doc` parameter
- Changed `getMediaAssetInfoPartial()` to accept optional `doc` parameter

---

> [!NOTE]
> **Medium Risk**
> Touches many DOM- and event-handling call sites across `@tldraw/editor`/`@tldraw/tldraw`/`@tldraw/utils`, which could introduce subtle focus, clipboard, export, or media-loading regressions despite being largely mechanical.
> 
> **Overview**
> Enables **cross-window/iframe embedding** by replacing direct `window`/`document` usage with container-scoped helpers throughout the editor and UI (e.g. `Editor.getContainerDocument()`/`getContainerWindow()`, `getOwnerDocument()`/`getOwnerWindow()`, and `getGlobalDocument()`/`getGlobalWindow()`). This updates event listener attachment, element creation, `instanceof` DOM checks, media/font loading, clipboard handling, printing, and export pipelines (SVG/style/font embedding) to use the correct owning window/document.
> 
> Updates public/internal APIs to accept an optional `Document` in key utilities (e.g. `activeElementShouldCaptureKeys`, `downloadFile`, `preloadFont`, `MediaHelpers.*`, `getMediaAssetInfoPartial`) and adds ESLint restrictions to prevent reintroducing bare globals (with tests exempted). Also fixes the ESM build post-step to rewrite dynamic `import()` specifiers to include `.mjs`, bumps CI `actions/checkout` to `v6`, and removes now-unneeded eslint-disable noise in a few tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7371184d65e2f99ee5391c82424dd9b4d10f5dc3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/config/TLUserPreferences.ts
+++ b/packages/editor/src/lib/config/TLUserPreferences.ts
@@ -2,6 +2,7 @@ import { atom } from '@tldraw/state'
 import { getDefaultTranslationLocale } from '@tldraw/tlschema'
 import { getFromLocalStorage, setInLocalStorage, structuredClone, uniqueId } from '@tldraw/utils'
 import { T } from '@tldraw/validate'
+import { getGlobalWindow } from '../utils/dom'
 
 const USER_DATA_KEY = 'TLDRAW_USER_DATA_v3'
 
@@ -154,10 +155,8 @@ function getRandomColor() {
 
 /** @internal */
 export function userPrefersReducedMotion() {
-	// eslint-disable-next-line no-restricted-syntax
-	if (typeof window !== 'undefined' && window.matchMedia) {
-		// eslint-disable-next-line no-restricted-syntax
-		return window.matchMedia?.('(prefers-reduced-motion: reduce)')?.matches ?? false
+	if (typeof window !== 'undefined' && getGlobalWindow().matchMedia) {
+		return getGlobalWindow().matchMedia?.('(prefers-reduced-motion: reduce)')?.matches ?? false
 	}
 
 	return false
--- a/packages/editor/src/lib/utils/browserCanvasMaxSize.ts
+++ b/packages/editor/src/lib/utils/browserCanvasMaxSize.ts
@@ -1,3 +1,5 @@
+import { getGlobalDocument } from './dom'
+
 /** @internal */
 export interface CanvasMaxSize {
 	maxWidth: number
@@ -94,8 +96,7 @@ const TEST_SIZES = {
  * @returns The maximum size of the canvas for the given dimension.
  */
 function getCanvasSize(dimension: 'width' | 'height' | 'area'): number {
-	// eslint-disable-next-line no-restricted-globals
-	const cropCvs = document.createElement('canvas')
+	const cropCvs = getGlobalDocument().createElement('canvas')
 	cropCvs.width = 1
 	cropCvs.height = 1
 	const cropCtx = cropCvs.getContext('2d')!
@@ -104,8 +105,7 @@ function getCanvasSize(dimension: 'width' | 'height' | 'area'): number {
 		const w = dimension === 'height' ? 1 : size
 		const h = dimension === 'width' ? 1 : size
 
-		// eslint-disable-next-line no-restricted-globals
-		const testCvs = document.createElement('canvas')
+		const testCvs = getGlobalDocument().createElement('canvas')
 		testCvs.width = w
 		testCvs.height = h
 		const testCtx = testCvs.getContext('2d')!
--- a/packages/editor/src/lib/utils/dom.ts
+++ b/packages/editor/src/lib/utils/dom.ts
@@ -112,8 +112,39 @@ export function elementShouldCaptureKeys(el: Element | null, includeButtonsAndMe
 	)
 }
 
+/**
+ * Returns the global `document`. Use this instead of bare `document` to satisfy lint rules.
+ *
+ * When you have a DOM node or editor instance, prefer the scoped versions instead:
+ * - `getOwnerDocument(node)` – the document that owns a specific DOM node
+ * - `editor.getContainerDocument()` – the document where the editor is mounted
+ *
+ * @internal
+ */
+export function getGlobalDocument(): Document {
+	// eslint-disable-next-line no-restricted-globals
+	if (typeof document !== 'undefined') return document
+	return globalThis.document
+}
+
+/**
+ * Returns the global `window`. Use this instead of bare `window` to satisfy lint rules.
+ *
+ * When you have a DOM node or editor instance, prefer the scoped versions instead:
+ * - `getOwnerWindow(node)` – the window that owns a specific DOM node
+ * - `editor.getContainerWindow()` – the window where the editor is mounted
+ *
+ * @internal
+ */
+export function getGlobalWindow(): Window & typeof globalThis {
+	if (typeof window !== 'undefined') return window as Window & typeof globalThis
+	return globalThis as Window & typeof globalThis
+}
+
 /** @internal */
 export function activeElementShouldCaptureKeys(includeButtonsAndMenus = true, doc?: Document) {
-	// eslint-disable-next-line no-restricted-globals
-	return elementShouldCaptureKeys((doc ?? document).activeElement, includeButtonsAndMenus)
+	return elementShouldCaptureKeys(
+		(doc ?? getGlobalDocument()).activeElement,
+		includeButtonsAndMenus
+	)
 }
--- a/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
+++ b/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
@@ -6,6 +6,7 @@ import {
 	TLDefaultFillStyle,
 	TLShapeUtilCanvasSvgDef,
 	debugFlags,
+	getGlobalDocument,
 	last,
 	suffixSafeId,
 	tlenv,
@@ -68,8 +69,7 @@ const generateImage = (dpr: number, currentZoom: number, darkMode: boolean) => {
 	return new Promise<Blob>((resolve, reject) => {
 		const size = TILE_PATTERN_SIZE * currentZoom * dpr
 
-		// eslint-disable-next-line no-restricted-globals
-		const canvasEl = document.createElement('canvas')
+		const canvasEl = getGlobalDocument().createElement('canvas')
 		canvasEl.width = size
 		canvasEl.height = size
 
@@ -112,8 +112,7 @@ const generateImage = (dpr: number, currentZoom: number, darkMode: boolean) => {
 }
 
 const canvasBlob = (size: [number, number], fn: (ctx: CanvasRenderingContext2D) => void) => {
-	// eslint-disable-next-line no-restricted-globals
-	const canvas = document.createElement('canvas')
+	const canvas = getGlobalDocument().createElement('canvas')
 	canvas.width = size[0]
 	canvas.height = size[1]
 	const ctx = canvas.getContext('2d')
--- a/packages/tldraw/src/lib/ui/getLocalFiles.ts
+++ b/packages/tldraw/src/lib/ui/getLocalFiles.ts
@@ -1,11 +1,16 @@
+import { getGlobalDocument } from '@tldraw/editor'
+
 export function getLocalFiles(options?: {
 	allowMultiple?: boolean
 	mimeTypes?: string[] | readonly string[]
 	document?: Document
 }) {
 	return new Promise<File[]>((resolve) => {
-		// eslint-disable-next-line no-restricted-globals
-		const { allowMultiple = true, mimeTypes = [], document: doc = document } = options || {}
+		const {
+			allowMultiple = true,
+			mimeTypes = [],
+			document: doc = getGlobalDocument(),
+		} = options || {}
 
 		const input = doc.createElement('input')
 		input.type = 'file'
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -7,6 +7,7 @@ import {
 	activeElementShouldCaptureKeys,
 	assert,
 	compact,
+	getGlobalDocument,
 	isDefined,
 	preventDefault,
 	uniq,
@@ -39,8 +40,7 @@ const expectedPasteFileMimeTypes = [
  */
 function stripHtml(html: string) {
 	// See <https://github.com/developit/preact-markup/blob/4788b8d61b4e24f83688710746ee36e7464f7bbc/src/parse-markup.js#L60-L69>
-	// eslint-disable-next-line no-restricted-globals
-	const doc = document.implementation.createHTMLDocument('')
+	const doc = getGlobalDocument().implementation.createHTMLDocument('')
 	doc.documentElement.innerHTML = html.trim()
 	return doc.body.textContent || doc.body.innerText || ''
 }
--- a/packages/tldraw/src/lib/utils/export/exportAs.ts
+++ b/packages/tldraw/src/lib/utils/export/exportAs.ts
@@ -1,4 +1,11 @@
-import { Editor, sanitizeId, TLExportType, TLImageExportOptions, TLShapeId } from '@tldraw/editor'
+import {
+	Editor,
+	getGlobalDocument,
+	sanitizeId,
+	TLExportType,
+	TLImageExportOptions,
+	TLShapeId,
+} from '@tldraw/editor'
 
 /** @public */
 export interface ExportAsOptions extends TLImageExportOptions {
@@ -56,9 +63,8 @@ function getTimestamp() {
 }
 
 /** @internal */
-// eslint-disable-next-line no-restricted-globals
-export function downloadFile(file: File, doc: Document = document) {
-	const link = doc.createElement('a')
+export function downloadFile(file: File, doc?: Document) {
+	const link = (doc ?? getGlobalDocument()).createElement('a')
 	const url = URL.createObjectURL(file)
 	link.href = url
 	link.download = file.name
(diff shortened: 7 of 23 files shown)


ds300:
@cursor push 4dd66c7284

packages/editor/src/lib/exports/domUtils.ts
```diff
@@ -22,6 +22,26 @@ export function* getRenderedChildren(node: Element) {
 	}
 }
 
+/** @internal */
+export function getOwnerWindow(
```
steveruizok, line 26: This is a naming thing, but I might go with `getWindow` and `getDocument` here instead. Maybe I have permissions on my mind... but I don't think `Owner` is doing much work here. It would be easier to assume that the window is usually just the window but that this helper will account for edge cases (e.g. where the window is owned by the current process but in a different view).

mimecuvalo: mm, i'd disagree. the thing with this PR is it's dealing with native clients (tauri/electron) and in that context it is specifically the use of `ownerDocument`
this is an internal API so i think it's more of an implementation detail really for us

.github/workflows/checks.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - name: Check out code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v6
```
mimecuvalo, line 25: huh weird straggler here

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -651,7 +654,8 @@ export async function getMediaAssetInfoPartial(
 	assetId: TLAssetId,
 	isImageType: boolean,
 	isVideoType: boolean,
-	maxImageDimension?: number
+	maxImageDimension?: number,
+	doc?: Document
```
mimecuvalo, line 658: annoyingly we should've probably made this have {options} but too late

packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -661,8 +665,8 @@ export async function getMediaAssetInfoPartial(
 	}
 
 	const size = isImageType
-		? await MediaHelpers.getImageSize(file)
-		: await MediaHelpers.getVideoSize(file)
+		? await MediaHelpers.getImageSize(file, doc)
+		: await MediaHelpers.getVideoSize(file, doc)
```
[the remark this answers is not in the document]
mimecuvalo: this looks valid

packages/editor/src/lib/exports/domUtils.ts
steveruizok: Is this just an internal API? I read this as more like our local storage helpers where we recommend/lint for everyone to use those, ie "if you want to reference the window in your code, use the `getWindow` rather than using `window` directly".

In that case this api would be _motivated_ by electron/tauri but the actual "owner" part seems like edge cases handled by a "give me the window", vs. "give me the owner window if there is one".

packages/tldraw/src/lib/utils/export/exportAs.ts
```diff
@@ -56,8 +56,9 @@ function getTimestamp() {
 }
 
 /** @internal */
-export function downloadFile(file: File) {
-	const link = document.createElement('a')
+// eslint-disable-next-line no-restricted-globals
+export function downloadFile(file: File, doc: Document = document) {
```
mimecuvalo: nit: let's keep it consistent across parameters - either we're specifying the default in the params `doc = document` or it's like in preload-font.ts (above) where it's using the null coalescing way)

packages/editor/src/lib/exports/domUtils.ts
steveruizok: ah the `getFromLocalStorage` etc is also internal. Ok, all good from me

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -39,6 +39,7 @@ const expectedPasteFileMimeTypes = [
  */
 function stripHtml(html: string) {
 	// See <https://github.com/developit/preact-markup/blob/4788b8d61b4e24f83688710746ee36e7464f7bbc/src/parse-markup.js#L60-L69>
+	// eslint-disable-next-line no-restricted-globals
 	const doc = document.implementation.createHTMLDocument('')
```
mimecuvalo: mm, i'm not loving these eslint sprinkled everywhere.
suggestion: let's make a `getDocument` method in `dom.ts` and that becomes the central place for retrieving the document the "safe"/canonical way. otherwise, this starts to get a little noisy just to reference `document`

packages/tldraw/src/lib/ui/getLocalFiles.ts
```diff
@@ -1,11 +1,13 @@
 export function getLocalFiles(options?: {
 	allowMultiple?: boolean
 	mimeTypes?: string[] | readonly string[]
+	document?: Document
 }) {
 	return new Promise<File[]>((resolve) => {
-		const { allowMultiple = true, mimeTypes = [] } = options || {}
+		// eslint-disable-next-line no-restricted-globals
```
mimecuvalo: so, like here it might be `doc = getDocument()`

packages/utils/src/lib/media/media.ts
```diff
@@ -141,9 +142,10 @@ export class MediaHelpers {
 	 * ```
 	 * @public
 	 */
-	static loadVideo(src: string): Promise<HTMLVideoElement> {
+	static loadVideo(src: string, doc?: Document): Promise<HTMLVideoElement> {
 		return new Promise((resolve, reject) => {
-			const video = document.createElement('video')
+			// eslint-disable-next-line no-restricted-globals
+			const video = (doc ?? document).createElement('video')
```
mimecuvalo, line 148: or if we're doing null coalescing it might be: `getDocument(doc)` and it does the right thing under the hood

packages/tldraw/src/lib/shapes/shared/useColorSpace.tsx
```diff
@@ -1,19 +1,21 @@
-import { debugFlags, useValue } from '@tldraw/editor'
+import { debugFlags, useMaybeEditor, useValue } from '@tldraw/editor'
 import { useEffect, useState } from 'react'
 
 export function useColorSpace(): 'srgb' | 'p3' {
+	const editor = useMaybeEditor()
 	const [supportsP3, setSupportsP3] = useState(false)
 
 	useEffect(() => {
+		const win = editor?.getContainerWindow() ?? window
```
mimecuvalo, line 9: why does referencing `window` here not trigger the no-restricted-globals? 🤔

packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.tsx
```diff
@@ -68,6 +68,7 @@ const generateImage = (dpr: number, currentZoom: number, darkMode: boolean) => {
 	return new Promise<Blob>((resolve, reject) => {
 		const size = TILE_PATTERN_SIZE * currentZoom * dpr
 
+		// eslint-disable-next-line no-restricted-globals
 		const canvasEl = document.createElement('canvas')
```
mimecuvalo: i'd consider also maybe having a helper method in `dom.ts` → `createCanvasEl()`

packages/editor/src/lib/utils/browserCanvasMaxSize.ts
```diff
@@ -94,6 +94,7 @@ const TEST_SIZES = {
  * @returns The maximum size of the canvas for the given dimension.
  */
 function getCanvasSize(dimension: 'width' | 'height' | 'area'): number {
+	// eslint-disable-next-line no-restricted-globals
```
mimecuvalo: yeah, i think i'd want to pull that into dom.ts after seeing it sprinkled around so much

packages/editor/src/lib/editor/managers/FocusManager/FocusManager.ts
```diff
@@ -69,7 +71,7 @@ export class FocusManager {
 		}
 	}
 
-	private handleMouseDown() {
+	@bind private handleMouseDown() {
```
mimecuvalo, line 74: why the new `@bind`s?

packages/editor/src/lib/config/TLUserPreferences.ts
```diff
@@ -154,7 +154,9 @@ function getRandomColor() {
 
 /** @internal */
 export function userPrefersReducedMotion() {
+	// eslint-disable-next-line no-restricted-syntax
 	if (typeof window !== 'undefined' && window.matchMedia) {
+		// eslint-disable-next-line no-restricted-syntax
```
mimecuvalo: just to make it explicit, we should also have a `getWindow()` in dom.ts

packages/editor/src/lib/components/default-components/CanvasShapeIndicators.tsx
```diff
@@ -223,7 +224,7 @@ export const CanvasShapeIndicators = memo(function CanvasShapeIndicators() {
 				$renderData.get()
 
 			const { w, h } = editor.getViewportScreenBounds()
-			const dpr = window.devicePixelRatio || 1
+			const dpr = editor.getInstanceState().devicePixelRatio
```
mimecuvalo, line 227: no fallback to 1 anymore?

packages/editor/src/lib/exports/domUtils.ts
```diff
@@ -22,6 +22,26 @@ export function* getRenderedChildren(node: Element) {
...
+	if (!nodeOrDocument) return globalThis.document
+	if (isDocument(nodeOrDocument)) return nodeOrDocument
+	return nodeOrDocument.ownerDocument ?? globalThis.document
+}
+
+function isDocument(node: Node | Document): node is Document {
+	return node.nodeType === Node.DOCUMENT_NODE
+}
+
 function getWindow(node: Node) {
```
mimecuvalo, line 45: oh i didn't realize we had a domUtils file (why is this separate from dom.ts??)
anyway, we should use some of the utils here or in dom.ts for the things in the PR

packages/editor/src/lib/exports/domUtils.ts
mimecuvalo: funnily enough, in my latest comments, i think i do want David to use getWindow/getDocument more, but to retrieve the regular window/document. we can centralize the access to these things and then have good comments there on what is going on - why this is all necessary

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
steveruizok: isn't that what `getOwnerDocument` does tho? or is this a place where we actually would want to get the local (?) document in the tauri/electron case, too?

ds300: here (and other places with the eslint-disable) it doesn't matter what document we use.

packages/tldraw/src/lib/shapes/shared/useColorSpace.tsx
ds300: it's only enabled on certain window methods/classes. There's a bunch of window usage that is safe across different contexts. This probably isn't one of them though, will investigate

packages/editor/src/lib/editor/managers/FocusManager/FocusManager.ts
ds300: because previously they were being bound on demand and it was buggy because the event listeners were never being released

packages/editor/src/lib/components/default-components/CanvasShapeIndicators.tsx
ds300: no longer needed

packages/editor/src/lib/utils/dom.ts
```diff
@@ -112,7 +112,37 @@ export function elementShouldCaptureKeys(el: Element | null, includeButtonsAndMe
...
+ *
+ * When you have a DOM node or editor instance, prefer the scoped versions instead:
+ * - `getOwnerDocument(node)` – the document that owns a specific DOM node
+ * - `editor.getContainerDocument()` – the document where the editor is mounted
+ *
+ * @internal
+ */
+export function getGlobalDocument(): Document {
+	// eslint-disable-next-line no-restricted-globals
+	return document
```
mimecuvalo: tyty

Issue #7794, opened 2026-02-02
Pass common shape props down instead of using per-shape hooks


steveruizok:
## Problem

Shape rendering currently uses hooks extensively within `ShapeUtil.component()` methods. Each shape independently calls hooks to access common values:

- `useDefaultColorTheme()` - called in ~7+ shapes/components
- `useEfficientZoomThreshold()` - called per shape for rendering optimization
- `useIsReadyForEditing()` - called in GeoShapeUtil, NoteShapeUtil
- `useValue()` for selection state (`isOnlySelected`) - called per shape
- `useValue()` for dark mode - called in NoteShapeUtil

This creates redundant reactive subscriptions and hook overhead per shape instance.

## Proposed solution

Pass commonly-needed values as props from the `Shape` wrapper component down to `ShapeUtil.component()`, reducing hook calls:

```typescript
interface ShapeRenderContext {
  // Theme/styling
  colorTheme: TLDefaultColorTheme
  isDarkMode: boolean
  
  // Selection/state
  isOnlySelected: boolean
  isReadyForEditing: boolean
  
  // Performance optimization
  zoomLevel: number
}
```

The `Shape` component already has access to the editor and could compute these values once, passing them to `util.component()`.

## Affected files

**Core rendering:**
- `packages/editor/src/lib/components/Shape.tsx` - would pass new props
- ShapeUtil interface would need updated `component()` signature

**Shape utils that would benefit:**
- `packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx` - uses theme, zoom threshold, selection, edit-ready
- `packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx` - uses theme, zoom threshold, dark mode, edit-ready
- `packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx` - uses theme
- `packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx` - uses theme
- `packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx` - uses zoom threshold
- `packages/tldraw/src/lib/shapes/highlight/HighlightShapeUtil.tsx` - uses theme

**Shared components:**
- `packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx`
- `packages/tldraw/src/lib/shapes/shared/PlainTextLabel.tsx`
- `packages/tldraw/src/lib/shapes/shared/ShapeFill.tsx`
[...]

Issue #7712, opened 2026-01-20
Add example for exploring default shape options


steveruizok:
## Summary

Create an interactive example that demonstrates how to explore and configure the options available on tldraw's default shapes. This would help developers understand what can be customized on shapes like frames, geo shapes, arrows, etc.

### Shape utility options

Options that control shape behavior:

- **Frame**: `showColors`, `resizeChildren`
- **Geo**: `showTextOutline`
- Other shape-specific options

### Style props

The built-in style system with values developers can use:

- Colors: black, grey, violet, blue, yellow, orange, green, red, white, etc.
- Dash styles: solid, dashed, dotted, draw
- Fill styles: none, solid, semi, pattern
- Font styles: sans, serif, mono, draw
- Size styles: s, m, l, xl

### Default props

How `getDefaultProps()` works and how defaults can be overridden.

## Suggested approach

Create an interactive example where users can:

1. Select a shape type from a dropdown
2. See all available options/props for that shape
3. Modify values and see shapes update reactively
4. Understand the relationship between options, props, and styles

Could use a side panel with controls that update shapes on the canvas in real-time.

## Why this is useful

- Helps developers understand what's configurable without reading source code
- Demonstrates reactive patterns for shape customization
- Shows the distinction between shape options, props, and styles
- Useful reference when building custom shape tools or overriding defaults

steveruizok:
## Research findings

### Relevant files

**Shape utilities** (with options):
- `packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx` - `{ showTextOutline }`
- `packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx` - Extensive options for snap distances, timing, text outline
- `packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx` - `{ extraArrowHorizontalPadding, showTextOutline }`
- `packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx` - `{ maxPointsPerShape: 600 }`
- `packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx` - `{ resizeMode: 'none' | 'scale' }`
- `packages/tldraw/src/lib/shapes/frame/FrameShapeUtil.tsx` - `{ showColors, resizeChildren }`
- `packages/tldraw/src/lib/shapes/highlight/HighlightShapeUtil.tsx` - `{ maxPointsPerShape, underlayOpacity, overlayOpacity }`

**Style props**:
[...]

Issue #7020, opened 2025-10-28
Add option to hide GeoShape border


iga1y:
## Summary
Currently, GeoShape always renders with a border (stroke/outline), even when `fill` is set. For modern flat design aesthetics, it would be useful to create geometric shapes with fill only and no border. This feature request proposes adding an option to hide the border entirely.

## Current Behavior

GeoShape always renders with both:
1. **Fill** - controlled by the `fill` property (`none`, `solid`, `semi`, `pattern`)
2. **Border** - always visible, controlled by `color`, `dash`, and `size` properties

**Example**:
- User creates a GeoShape with `fill: 'solid'` and `color: 'blue'`
- The shape renders with both blue fill AND blue border
- There is no way to create a shape with fill only (no border)

## Expected Behavior

Extend the existing `dash` property to include a `'none'` value, allowing users to hide the border entirely:

```typescript
export const DefaultDashStyle = StyleProp.defineEnum('tldraw:dash', {
  defaultValue: 'draw',
  values: ['none', 'draw', 'solid', 'dashed', 'dotted'], // add 'none'
})
```

### Where would you like this feature?

The developer package

### Contact Details

rentaro.igari@alma-inc.co.jp

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

MitjaBezensek:
Hey!

Thanks for the suggestion. 

Yes, we definitely want to allow users to customize the existing styling. We already did some exploration about this and have a pretty good understanding of what changes would be needed.

iga1y:
@MitjaBezensek 

Thank you for the quick response! I'm really glad to hear that this is already on your radar.

Style customization is something we're really looking forward to, especially for creating modern flat design compositions. 

If there's anything I can help with (testing, feedback, use cases, etc.), please let me know. Looking forward to seeing this feature in a future release!

zb-builds-things:
Very interested in this as well! We built some custom geoShapes with the borders gone but would love it to be built in too.

PR #8216, opened 2026-03-12
feat(editor): add first-class theme system with display values for shape customization

steveruizok:
In order to give SDK consumers full control over the visual appearance of shapes — colors, font sizes, stroke widths, and more — this PR introduces three interconnected systems: first-class themes, display values, and runtime style extensibility.

Previously, shape styling was spread across hardcoded constants, a boolean `isDarkMode` flag, and a static `DefaultColorThemePalette`. Customizing colors required overriding internal theme objects, and there was no way to adjust font sizes or stroke widths per-theme. This PR replaces all of that with a structured, reactive theme system and a display values pattern that makes every visual property of every built-in shape configurable.

Clean reimplementation of the [`first-class-theme`](https://github.com/tldraw/tldraw/tree/first-class-theme) branch.

Relates to #7587, #7794, #7712, #2515, #3211, #1665, #7020

https://github.com/user-attachments/assets/ad3f6fad-3da6-4690-8bd6-2d28053b3b76

---

## First-class themes

**New types (`@tldraw/tlschema`)**
- `TLTheme` — a theme definition with `id`, `fontSize`, `lineHeight`, `strokeWidth`, and a `colors: TLThemeColors` palette
- `TLThemeColors` — extensible interface containing base UI colors (`text`, `background`, `solid`, `cursor`) and named shape colors (each a `TLDefaultColor` with 5 variants: `solid`, `semi`, `pattern`, `fill`, `linedFill`)
- `TLThemes` — a `Record<string, TLTheme>` map of named themes

**ThemeManager (`@tldraw/editor`)**
- Reactive manager maintaining a map of named themes with automatic light/dark selection based on the user's dark mode preference
- Supports theme overrides via `setCurrentTheme(themeId)` and runtime updates via `updateThemes()`
- `DEFAULT_LIGHT_THEME` and `DEFAULT_DARK_THEME` define the complete built-in color palettes

**Editor API**
- `editor.getCurrentTheme()` — returns the active `TLTheme`
- `editor.getCurrentThemeId()` — returns the active theme ID string
- `editor.setTheme(themeId | null)` — override the active theme (or revert to automatic)
- `editor.getThemes()` — returns all registered themes
- `editor.updateThemes(partial)` — merge additional themes

**Component props**
- `TldrawEditor` accepts `theme` (active theme ID) and `themes` (theme map) props
- Both are reactive — changing them updates the editor and all shapes immediately
- Replaces the old `inferDarkMode` boolean prop

## Display values

**Pattern**
- Each ShapeUtil declares `getDisplayValues(editor, shape, theme)` in its `options`, returning a typed object of every resolved visual property the shape needs to render (e.g. `strokeColor`, `strokeWidth`, `fillColor`, `labelFontSize`, `labelColor`, etc.)
- A companion `getDisplayValueOverrides(editor, shape, theme)` returns a `Partial<DisplayValues>` that is merged on top, letting SDK consumers customize any property without subclassing
- The `getDisplayValues()` helper function merges base + overrides, with results cached via a `WeakMap` keyed on the shape record

**Shape-specific visual properties**
- Frame, note, and highlight colors that were previously part of `TLDefaultColor` are now defined directly in each shape's options (e.g. `noteBorder`, `noteFill`, `noteText` in `NoteShapeOptions`; `frameHeadingStroke`, `frameFill`, etc. in `FrameShapeOptions`; `highlightSrgb`, `highlightP3` in `HighlightShapeOptions`)
- This keeps the core `TLDefaultColor` interface minimal and lets each shape own its color mappings

**Shape util migration**
All 12 built-in shape utils are migrated to use display values:
- **GeoShapeUtil** — stroke, fill, label styling with theme-derived colors and font sizes
- **NoteShapeUtil** — note dimensions, background color, border, label styling; `fontSizeAdjustment` converted from absolute pixels to a nullable ratio
- **ArrowShapeUtil** — stroke, label, arrowhead colors and sizes; arrow info functions accept display values and thread them through geometry computation
- **DrawShapeUtil** — stroke color and width from theme; display values threaded through geometry functions
- **FrameShapeUtil** — frame heading, stroke, fill, and text colors from shape options
- **TextShapeUtil** — text color, font size, line height from theme
- **LineShapeUtil** — stroke color and width from theme
- **HighlightShapeUtil** — highlight color with P3/sRGB selection from shape options
- **EmbedShapeUtil** — border styling from theme
- **ImageShapeUtil** — border styling from theme
- **VideoShapeUtil** — border styling from theme
- **BookmarkShapeUtil** — border styling from theme

**Constants as multipliers**
- `STROKE_SIZES`, `FONT_SIZES`, `LABEL_FONT_SIZES`, `ARROW_LABEL_FONT_SIZES` are now multipliers applied against `theme.fontSize` and `theme.strokeWidth`
- This lets themes globally scale all text and strokes by adjusting two values

## Runtime style extensibility

- `EnumStyleProp.addValues()` / `removeValues()` — mutate style prop values at runtime
- `registerColors(names)` — registers new color names with `DefaultColorStyle` and `DefaultLabelColorStyle`
- Extend `TLThemeColors` via module augmentation to add custom color definitions
- `getColorStyleItems(theme)` dynamically derives style panel color options from the theme palette, ordered by `DefaultColorStyle.values` first, then any additional theme colors

## Other changes

- `useCurrentThemeId()` hook replaces `useIsDarkMode()` — returns the active theme ID string
- P3 color space detection moved from per-component `useColorSpace` hook to a global reactive atom in `environment.ts`
- `SvgExportContext` now includes `themeId` so SVG exports resolve theme-aware colors
- `getColorValue(theme, color, variant)` moved from `@tldraw/tlschema` to `@tldraw/editor` with a new signature accepting `TLTheme`
- `PatternFill` component extracted from deleted `ShapeFill`, using `useCurrentThemeId` for pattern URL resolution
- `TLFrameShape` color prop delegates validation to `DefaultColorStyle` so runtime additions are picked up
- `TLNoteShape.fontSizeAdjustment` changed from `number` to `number | null` (ratio-based, with migration)
- Removed `inferDarkMode` from `UserPreferencesManager` constructor

## New examples

- **Custom theme** — demonstrates custom light/dark themes, runtime theme switching, custom colors via `registerColors` and module augmentation, and reactive theme value sliders
- **Display options** — demonstrates `getDisplayValueOverrides` for customizing shape appearance

## Documentation

- New `themes.mdx` doc covering the theme system, custom themes, custom colors, and display value overrides
- Updated `options.mdx`, `styles.mdx`, and `user-preferences.mdx` to reference new APIs

### Change type

- [x] `feature`

### Test plan

1. Open the examples app and verify all shapes render correctly with default themes
2. Toggle dark mode — shapes should update to dark theme colors
3. Open the "Custom theme" example and switch between themes
4. Open the "Display options" example and verify shape customization works
5. Create shapes of each type and verify colors, strokes, and fonts render correctly
6. Export shapes as SVG and verify theme colors are applied correctly

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add first-class theme system with `TLTheme`, `TLThemeColors`, and `ThemeManager`
- Add `theme` and `themes` props to `TldrawEditor` for custom theme support
- Add display values system for theme-aware shape customization via `getDisplayValueOverrides`
- Add `registerColors()` for runtime color palette extension
- Add `useCurrentThemeId` hook (replaces `useIsDarkMode`)
- Add `getColorStyleItems(theme)` for dynamic style panel color derivation
- Convert shape size constants to theme-relative multipliers
- Move frame/note/highlight-specific color variants from `TLDefaultColor` into shape options
- Remove `inferDarkMode` prop (use `theme` prop instead)
- Remove `getDefaultColorTheme`, `DefaultColorThemePalette`, `useDefaultColorTheme`, `useIsDarkMode`

### API changes

- Added `TLTheme`, `TLThemeColors`, `TLDefaultColor`, `TLThemes` types to `@tldraw/tlschema`
- Added `TLDefaultColorStyle` type (derived from `TLThemeColors` keys) to `@tldraw/tlschema`
- Added `registerColors()` to `@tldraw/tlschema`
- Added `EnumStyleProp.addValues()` and `EnumStyleProp.removeValues()` to `@tldraw/tlschema`
- Added `ThemeManager` class to `@tldraw/editor`
- Added `DEFAULT_LIGHT_THEME`, `DEFAULT_DARK_THEME` constants to `@tldraw/editor`
- Added `Editor.getCurrentTheme()`, `Editor.getCurrentThemeId()`, `Editor.setTheme()`, `Editor.updateThemes()`, `Editor.getThemes()`
- Added `useCurrentThemeId()` hook to `@tldraw/editor`
- Added `getColorValue(theme, color, variant)` to `@tldraw/editor` (moved from tlschema, signature changed to accept `TLTheme`)
- Added `theme` and `themes` props to `TldrawEditorBaseProps`
- Added `themeId` to `SvgExportContext`
- Added `getDisplayValues()` helper and `ShapeOptionsWithDisplayValues` interface to `@tldraw/tldraw`
- Added `getColorStyleItems(theme)` to `@tldraw/tldraw`
- Added `DisplayValues` types for all built-in shape utils (e.g. `GeoShapeUtilDisplayValues`, `NoteShapeUtilDisplayValues`, `ArrowShapeUtilDisplayValues`, etc.)
- Added `Options` types for shape utils (e.g. `GeoShapeOptions`, `NoteShapeOptions`, `ArrowShapeOptions`, etc.)
- Breaking! Removed `inferDarkMode` from `TLEditorOptions` and `TldrawEditorBaseProps`
- Breaking! Removed `useIsDarkMode` hook from `@tldraw/editor`
- Breaking! Removed `useDefaultColorTheme` hook from `@tldraw/tldraw`
- Breaking! Removed `ShapeFill` component from `@tldraw/tldraw`
- Breaking! Removed `getDefaultColorTheme`, `DefaultColorThemePalette`, `TLDefaultColorTheme`, `TLDefaultColorThemeColor`, `defaultColorNames`, `getColorValue` from `@tldraw/tlschema`
- Breaking! Removed `FONT_SIZES`, `STROKE_SIZES`, `LABEL_FONT_SIZES`, `ARROW_LABEL_FONT_SIZES`, `TEXT_PROPS` exports from `@tldraw/tldraw` (now internal)
- Changed `TLNoteShape.fontSizeAdjustment` from `number` to `number | null`
- Changed `TLDefaultColor` from 14 variants to 5 (`solid`, `semi`, `pattern`, `fill`, `linedFill`) — frame/note/highlight colors moved to shape options

### Code changes

| Section         | LOC change     |
| --------------- | -------------- |
| Core code       | +3212 / -2218  |
| Tests           | +440 / -67     |
| Automated files | +447 / -195    |
| Documentation   | +896 / -146    |
| Apps            | +8 / -8        |
| Templates       | +66 / -34      |
| Config/tooling  | +2 / -2        |

### DIFF
--- a/apps/docs/content/sdk-features/themes.mdx
+++ b/apps/docs/content/sdk-features/themes.mdx
@@ -310,7 +310,7 @@ const themes: TLThemes = {
 <Tldraw themes={themes} />
 ```
 
-tldraw validates shape properties (including colors) when data enters the store, for example when loading from IndexedDB or syncing from a server. Passing `themes` tells tldraw about your custom colors *before* data is loaded, so they pass validation. When creating a store yourself, pass `themes` there too, and still pass `themes` to `<Tldraw>` so the editor knows how to render them:
+tldraw validates shape properties (including colors) when data enters the store, for example when loading from IndexedDB or syncing from a server. Passing `themes` tells tldraw about your custom colors _before_ data is loaded, so they pass validation. When creating a store yourself, pass `themes` there too, and still pass `themes` to `<Tldraw>` so the editor knows how to render them:
 
 **Direct store creation:**
 
--- a/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
+++ b/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
@@ -18,9 +18,9 @@ import {
 	toFixed,
 	uniqueId,
 } from '@tldraw/editor'
-import { getDisplayValues } from '../../shared/getDisplayValues'
 import { HighlightShapeUtil } from '../../highlight/HighlightShapeUtil'
 import { STROKE_SIZES } from '../../shared/default-shape-constants'
+import { getDisplayValues } from '../../shared/getDisplayValues'
 import { DrawShapeUtil } from '../DrawShapeUtil'
 
 type DrawableShape = TLDrawShape | TLHighlightShape
--- a/apps/docs/content/sdk-features/default-shapes.mdx
+++ b/apps/docs/content/sdk-features/default-shapes.mdx
@@ -221,9 +221,9 @@ Properties:
 
 Configuration options:
 
-| Option              | Type     | Default | Description                                                                                     |
-| ------------------- | -------- | ------- | ----------------------------------------------------------------------------------------------- |
-| `maxPointsPerShape` | `number` | `600`   | Maximum points before starting a new shape. Same behavior as draw shapes.                       |
+| Option              | Type     | Default | Description                                                               |
+| ------------------- | -------- | ------- | ------------------------------------------------------------------------- |
+| `maxPointsPerShape` | `number` | `600`   | Maximum points before starting a new shape. Same behavior as draw shapes. |
 
 The highlight's `underlayOpacity` (default `0.82`) and `overlayOpacity` (default `0.35`) are display values that can be overridden via `getDisplayValueOverrides`.
 
--- a/packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
@@ -25,8 +25,8 @@ import {
 	embedShapePermissionDefaults,
 } from '../../defaultEmbedDefinitions'
 import { TLEmbedResult, getEmbedInfo } from '../../utils/embeds/embeds'
-import { BookmarkIndicatorComponent, BookmarkShapeComponent } from '../bookmark/BookmarkShapeUtil'
 import { BOOKMARK_JUST_URL_HEIGHT, BOOKMARK_WIDTH } from '../bookmark/bookmarks'
+import { BookmarkIndicatorComponent, BookmarkShapeComponent } from '../bookmark/BookmarkShapeUtil'
 import { ShapeOptionsWithDisplayValues, getDisplayValues } from '../shared/getDisplayValues'
 import { getRotatedBoxShadow } from '../shared/rotated-box-shadow'
 
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
@@ -16,11 +16,11 @@ import {
 	isPageId,
 	kickoutOccludedShapes,
 } from '@tldraw/editor'
-import type { NoteShapeUtil } from '../../../shapes/note/NoteShapeUtil'
 import {
 	NOTE_ADJACENT_POSITION_SNAP_RADIUS,
 	getAvailableNoteAdjacentPositions,
 } from '../../../shapes/note/noteHelpers'
+import type { NoteShapeUtil } from '../../../shapes/note/NoteShapeUtil'
 import { getDisplayValues } from '../../../shapes/shared/getDisplayValues'
 import { DragAndDropManager } from '../DragAndDropManager'
 


packages/editor/src/lib/editor/managers/ThemeManager/ThemeManager.ts
```diff
@@ -0,0 +1,75 @@
...
+		const override = this._themeOverride.get()
+		if (override !== null) return override
+		return this.editor.user.getIsDarkMode() ? 'dark' : 'light'
+	}
+
+	@computed getCurrentTheme(): TLTheme {
+		const themes = this._themes.get()
+		const themeId = this.getCurrentThemeId()
+		return themes[themeId] ?? themes['light']
+	}
```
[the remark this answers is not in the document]
steveruizok: acceptable

packages/editor/src/lib/exports/getSvgJsx.tsx
```diff
@@ -217,12 +220,14 @@ function SvgExport({
...
 	isDarkMode: boolean
+	themeId: string
 	renderingShapes: TLRenderingShape[]
 	onMount(): void
 	waitUntil(promise: Promise<void>): void
 }) {
 	const masksId = useUniqueSafeId()
-	const theme = getDefaultColorTheme({ isDarkMode })
+	const themes = editor.getThemes()
+	const theme = themes[themeId] ?? themes[Object.keys(themes)[0]]
```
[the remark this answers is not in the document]
steveruizok: acceptable

apps/examples/src/examples/shapes/tools/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
```diff
@@ -190,7 +210,7 @@ export class SpeechBubbleUtil extends ShapeUtil<SpeechBubbleShape> {
 		const pathData = 'M' + vertices[0] + 'L' + vertices.slice(1) + 'Z'
 		const isSelected = shape.id === this.editor.getOnlySelectedShapeId()
 		// eslint-disable-next-line react-hooks/rules-of-hooks
-		const theme = useDefaultColorTheme()
+		const theme = useEditor().getCurrentTheme()
```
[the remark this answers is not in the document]
steveruizok: acceptable

packages/editor/src/lib/hooks/useDarkMode.ts
```diff
@@ -2,17 +2,17 @@ import { useValue } from '@tldraw/state-react'
...
 export function useDarkMode() {
 	const editor = useEditor()
 	const container = useContainer()
-	const isDarkMode = useIsDarkMode()
+	const themeId = useCurrentThemeId()
 	const forceSrgb = useValue(debugFlags.forceSrgb)
 
 	React.useEffect(() => {
-		if (isDarkMode) {
+		if (themeId === 'dark') {
```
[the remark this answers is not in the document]
steveruizok: acceptable, deprecated api

packages/editor/src/lib/editor/managers/ThemeManager/defaultThemes.ts
```diff
@@ -0,0 +1,491 @@
...
+	theme: TLTheme,
+	color: TLDefaultColorStyle | string,
+	variant: keyof TLDefaultColor
+): string {
+	if (!(color in theme.colors)) {
+		// If the color is not a key in the theme's colors, assume it's a custom color value and return it directly
+		return color
+	}
+	return theme.colors[color as TLDefaultColorStyle][variant]
+}
```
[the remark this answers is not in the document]
steveruizok: doctor it hurts if i do this *arggh

angrycaptain19:
I tested this out with examples that we've heard recently as requested use cases:
- Custom brand kits
- Creating a colour wheel for styles
- Draw stroke override
- Making black sticky notes black (you could do this by customizing the theme already)

All of these worked very well

packages/tlschema/src/shapes/TLNoteShape.ts
```diff
@@ -265,5 +266,14 @@ export const noteShapeMigrations = createShapePropsMigrationSequence({
 				}
 			},
 		},
+		{
+			id: Versions.MakeFontSizeAdjustmentRatio,
+			up: (props) => {
+				props.fontSizeAdjustment = null
```
mimecuvalo, line 276: why not use the pre-existing number and just calculate the ratio?

packages/tlschema/src/shapes/TLNoteShape.ts
```diff
@@ -265,5 +266,14 @@ export const noteShapeMigrations = createShapePropsMigrationSequence({
...
+			},
+			down: (props) => {
+				props.fontSizeAdjustment = 0
```
mimecuvalo, line 279: mm, not good - this'll make the font 0 if people go backwards 😬

packages/tlschema/src/shapes/TLNoteShape.ts
```diff
@@ -50,8 +50,8 @@ export interface TLNoteShapeProps {
 	size: TLDefaultSizeStyle
 	/** Font family style for the note text */
 	font: TLDefaultFontStyle
-	/** Adjustment to the base font size (positive increases, negative decreases) */
-	fontSizeAdjustment: number
+	/** Ratio to scale the base font size when text needs to shrink to fit. Null means needs recomputation, 0 means no adjustment, and values less than 1 indicate shrinkage. */
```
mimecuvalo: technically, people could make it grow if they wanted to?

apps/examples/e2e/tests/test-text.spec.ts
```diff
@@ -269,7 +269,7 @@ test.describe('text measurement', () => {
 			return editor.getShape(id) as TLNoteShape
 		})
 
-		expect(shape.props.fontSizeAdjustment).toEqual(32)
+		expect(shape.props.fontSizeAdjustment).toEqual(0)
```
mimecuvalo: this feels very strange that 0 is being used - i think we should just use 1.
this is interpreted as a ratio, so we can always special case `1` to be the "needs no adjustment case"

apps/examples/src/examples/configuration/configure-shape-util/README.md
```diff
@@ -20,4 +20,4 @@ Change the behavior of built-in shapes by setting their options.
 
 Some of the builtin tldraw shapes can be customized to behave differently based on your needs. This is done via the `ShapeUtil.configure` function which returns a new version of the shape's util class with custom options specified.
 
-You can see a shape's options by looking at the `options` property of its `ShapeUtil`. For example, the note shape's options are listed at [`NoteShapeOptions`](https://tldraw.dev/reference/tldraw/NoteShapeOptions).
+You can see a shape's options by looking at the `options` property of its `ShapeUtil`. For example, the note shape's options are listed at [`NoteShapeUtilOptions`](https://tldraw.dev/reference/tldraw/NoteShapeUtilOptions).
```
mimecuvalo: would we ever rename NoteShapeUtil → NoteShape :ducks: :)
nit: feels like `NoteShapeUtilOptions` is a mouthful

apps/docs/content/sdk-features/themes.mdx
```diff
@@ -0,0 +1,303 @@
...
+Each color in the palette is a [TLDefaultColor](?) with variants for different contexts:
+
+| Variant              | Purpose                                        |
+| -------------------- | ---------------------------------------------- |
+| `solid`              | Full-opacity color for strokes and solid fills |
+| `semi`               | Semi-transparent color for background fills    |
+| `pattern`            | Color used in pattern/hatch fills              |
+| `fill`               | Explicit fill color (usually same as solid)    |
+| `linedFill`          | Slightly lighter fill for lined patterns       |
+| `frameHeadingStroke` | Stroke color for frame headings                |
```
mimecuvalo, line 130: b/c these are very specific to our canonical shapes, and since your philosophy is to have things closer to where they are used: it feels like these frame* and note* options should actually be configure options on their respective shapes. wdyt?

apps/docs/content/sdk-features/themes.mdx
```diff
@@ -0,0 +1,303 @@
...
+		},
+	},
+})
+```
+
+You can also pass themes as a prop on the [Tldraw](?) or [TldrawEditor](?) component. The prop is reactive: when it changes, the editor's themes update automatically.
+
+```tsx
+const myThemes: TLThemes = {
+	light: {
```
mimecuvalo, line 232: we should make it clear here that you can do more than just 'light' and 'dark' - that it can also be called a custom name like 'heavy-metal-goth' theme.

apps/examples/src/examples/ui/changing-default-colors/ChangingDefaultColorsExample.tsx
```diff
@@ -1,32 +1,47 @@
...
+			<Tldraw
+				persistenceKey="example"
+				onMount={(editor) => {
+					const lightTheme = editor.getThemes()['light']
+					editor.updateThemes({
+						light: {
+							...lightTheme,
+							colors: {
+								...lightTheme.colors,
+								black: {
```
mimecuvalo, line 19: we should show how to add a new custom color here too, like:
```
primary: {
  
},
secondary: {
},
warning: {
},
info: {
},
error: {
},
etc...
```

they should absolutely not be limited to our current limited color count

apps/examples/src/examples/ui/custom-theme/CustomThemeExample.tsx
```diff
@@ -0,0 +1,343 @@
...
+		colors: { ...DEFAULT_DARK_THEME.colors, pink: pinkDark },
+	},
+	'my-brand': {
+		id: 'my-brand',
+		fontSize: DEFAULT_DARK_THEME.fontSize,
+		lineHeight: DEFAULT_DARK_THEME.lineHeight,
+		strokeWidth: DEFAULT_DARK_THEME.strokeWidth,
+		colors: {
+			...DEFAULT_DARK_THEME.colors,
+			pink: pinkDark,
```
mimecuvalo: i think it'd be useful to show that pink can be an object:
```
pink: {
  solid:
  semi:
  etc.
}
```

otherwise it seems limited

apps/examples/src/examples/ui/custom-theme/CustomThemeExample.tsx
```diff
@@ -0,0 +1,343 @@
...
+	}
+}
+
+// There's a guide at the bottom of this file!
+
+// [1]
+// Register a custom "pink" color so it appears in the style panel and
+// passes validation everywhere. Call this before rendering any tldraw
+// components.
+registerColors(['pink'])
```
mimecuvalo: to make this example powerful we should use a label that means something semantically instead of a direct color name like pink. pick something like `primaryBrand`

packages/editor/src/lib/editor/types/SvgExportContext.tsx
```diff
@@ -42,6 +42,11 @@ export interface SvgExportContext {
 	 */
 	readonly isDarkMode: boolean
 
+	/**
+	 * The theme ID to use for this export (e.g. `'light'`, `'dark'`, or a custom theme ID).
```
mimecuvalo: nit: weird double quoting
nit: or a custom theme ID, e.g. "company-brand"

packages/editor/src/lib/TldrawEditor.tsx
```diff
@@ -219,6 +222,11 @@ export interface TldrawEditorBaseProps {
 	 * The URLs for the fonts to use in the editor.
 	 */
 	assetUrls?: { fonts?: { [key: string]: string | undefined } }
+
+	/**
+	 * Named color themes for the editor.
+	 */
+	themes?: TLThemes
```
mimecuvalo: nit: move this closer to `theme`

packages/tldraw/src/lib/shapes/arrow/elbow/getElbowArrowInfo.tsx
```diff
@@ -347,7 +348,9 @@ function getElbowArrowTerminalInfo(
 	binding: TLArrowBinding | undefined,
 	point: VecModel
 ): ElbowArrowTerminal {
-	const arrowStrokeSize = (STROKE_SIZES[arrow.props.size] * arrow.props.scale) / 2
+	const theme = editor.getCurrentTheme()
+	const arrowStrokeSize =
+		(theme.strokeWidth * STROKE_SIZES[arrow.props.size] * arrow.props.scale) / 2
```
mimecuvalo: seems like this needs to be able to be hooked into the runtime overrides where necessary (via `getDisplayValueOverrides()`).

packages/tldraw/src/lib/shapes/arrow/arrowLabel.ts
```diff
@@ -132,12 +137,14 @@ function getArrowLabelSize(editor: Editor, shape: TLArrowShape) {
 	return labelSizeCache.get(editor, shape.id) ?? new Vec(0, 0)
 }
 
-function getLabelToArrowPadding(shape: TLArrowShape) {
-	const strokeWidth = STROKE_SIZES[shape.props.size]
+function getLabelToArrowPadding(shape: TLArrowShape, theme: TLTheme) {
+	const strokeWidth = theme.strokeWidth * STROKE_SIZES[shape.props.size]
```
mimecuvalo, line 141: using `theme` is fine but i think we need to also possibly allow for per-shape overrides?

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -138,15 +128,21 @@ export const RichTextLabel = React.memo(function RichTextLabel({
 				showTextOutline ? 'tl-text__outline' : 'tl-text__no-outline'
 			)}
 			aria-hidden={!isEditing}
-			data-font={font}
```
mimecuvalo, line 145: i think this comment keeps getting lost but these data- properties are tied to the CSS and are necessary

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -138,15 +128,21 @@ export const RichTextLabel = React.memo(function RichTextLabel({
...
 			data-isediting={isEditing}
 			data-textwrap={!!wrap}
 			data-isselected={isSelected}
 			style={{
-				justifyContent: align === 'middle' || legacyAlign ? 'center' : align,
-				alignItems: verticalAlign === 'middle' ? 'center' : verticalAlign,
+				fontFamily,
+				textAlign,
+				justifyContent:
+					textAlign === 'center' ? 'center' : textAlign === 'end' ? 'flex-end' : 'flex-start',
```
mimecuvalo: looks like this comment got lost too in the other PR - this needs to still compare to `middle`

packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
```diff
@@ -217,7 +218,10 @@ export class Drawing extends StateNode {
 				this.pagePointWhereNextSegmentChanged = null
 				const segments = [...shape.props.segments, newSegment]
 
-				if (this.currentLineLength < STROKE_SIZES[shape.props.size] * 4) {
+				if (
+					this.currentLineLength <
+					this.editor.getCurrentTheme().strokeWidth * STROKE_SIZES[shape.props.size] * 4
```
mimecuvalo: does this logic need to take into account dv as well?

packages/tldraw/src/lib/shapes/arrow/straight-arrow.ts
```diff
@@ -150,8 +152,10 @@ export function getStraightArrowInfo(
 			!endShapeInfo.isExact
 		) {
 			strokeOffsetB =
-				STROKE_SIZES[shape.props.size] / 2 +
-				('size' in endShapeInfo.shape.props ? STROKE_SIZES[endShapeInfo.shape.props.size] / 2 : 0)
+				(theme.strokeWidth * STROKE_SIZES[shape.props.size]) / 2 +
+				('size' in endShapeInfo.shape.props
+					? (theme.strokeWidth * STROKE_SIZES[endShapeInfo.shape.props.size]) / 2
```
mimecuvalo, line 159: ditto here about dv vs theme

packages/editor/src/lib/editor/managers/ThemeManager/ThemeManager.ts
```diff
@@ -0,0 +1,75 @@
...
+import { DEFAULT_DARK_THEME, DEFAULT_LIGHT_THEME } from './defaultThemes'
+
+function buildDefaultThemes(): TLThemes {
+	return {
+		light: DEFAULT_LIGHT_THEME,
+		dark: DEFAULT_DARK_THEME,
+	}
+}
+
+/** @public */
```
mimecuvalo: comment here would be good

packages/tlschema/src/index.ts
```diff
@@ -266,15 +266,9 @@ export {
 } from './shapes/TLVideoShape'
 export { EnumStyleProp, StyleProp, type StylePropValue } from './styles/StyleProp'
 export {
-	defaultColorNames,
 	DefaultColorStyle,
-	DefaultColorThemePalette,
-	DefaultLabelColorStyle,
-	getColorValue,
-	getDefaultColorTheme,
+	registerCustomColors as registerColors,
```
mimecuvalo: why not just call this registerColors in general?

packages/tldraw/src/lib/shapes/text/TextShapeUtil.tsx
```diff
@@ -120,26 +150,27 @@ export class TextShapeUtil extends ShapeUtil<TLTextShape> {
...
 
 		const { width, height } = this.getMinDimensions(shape)
 		const isSelected = shape.id === this.editor.getOnlySelectedShapeId()
-		const theme = useDefaultColorTheme()
+		const themeId = useCurrentThemeId()
+		const dv = getDisplayValues(this, shape, themeId)
 		const handleKeyDown = useTextShapeKeydownHandler(id)
 
 		return (
 			<RichTextLabel
```
mimecuvalo, line 164: comment from prev. PR worth discussing:
b/c there's so much duplication here with prop-drilling to RichTextLabel, and since you're advocating for a philosophy of the options being as close/tightly coupled to the component they're targeting, you should consider just having RichTextLabel/PlainTextLabel as having a configure option as well so that folks don't have do the same thing over and over with text options (font, color, line-height, etc.)

packages/tldraw/src/lib/shapes/shared/PlainTextLabel.tsx
mimecuvalo: comment same from previous PR:
you nixed legacyAlign in RichTextLabel (not sure if intentional). let's make that choice consistent here

packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @public
+ */
+export function getDisplayValues<Shape extends TLShape, DisplayValues extends object>(
+	util: { editor: Editor; options: ShapeOptionsWithDisplayValues<Shape, DisplayValues> },
+	shape: Shape,
+	themeId?: string
+): DisplayValues {
+	const theme = themeId ? util.editor.getThemes()[themeId] : util.editor.getCurrentTheme()
+	return {
+		...util.options.getDisplayValues(util.editor, shape, theme),
```
mimecuvalo, line 34: as noted in the previous PR, this getDisplayValues needs some rethinking. the example there was:

> so, if I've provided a custom getDisplayValues function, it will be missing the new labelLetterSpacing in the future, and my shape will start to look more crufty than the canonical version. The drift will start to become more apparent over time.

i think maybe we consider only `getDisplayValueOverrides` being able to be set as an option, overriding `getDisplayValues` ends up being a footgun in the long run

packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -49,18 +48,48 @@ import {
...
 import { useEfficientZoomThreshold } from '../shared/useEfficientZoomThreshold'
-import {
-	CLONE_HANDLE_MARGIN,
-	NOTE_CENTER_OFFSET,
-	NOTE_SIZE,
-	getNoteShapeForAdjacentPosition,
-} from './noteHelpers'
+import { CLONE_HANDLE_MARGIN, getNoteShapeForAdjacentPosition } from './noteHelpers'
+
+const NOTE_SHAPE_HORIZONTAL_ALIGNS = Object.freeze({
```
mimecuvalo, line 59: migrating comments from previous PR:
> why is this recreating DefaultHorizontalAlignStyle?

same w stuff below

packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
```diff
@@ -0,0 +1,27 @@
...
+	getDisplayValues(editor: Editor, shape: Shape, theme: TLTheme): DisplayValues
+	getDisplayValueOverrides(editor: Editor, shape: Shape, theme: TLTheme): Partial<DisplayValues>
+}
+
+/**
+ * Get the resolved display values for a shape, merging the base values with any overrides.
+ *
+ * @public
+ */
+export function getDisplayValues<Shape extends TLShape, DisplayValues extends object>(
```
mimecuvalo, line 19: we should consider some kind of WeakCache for this stuff possibly

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -945,6 +959,65 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	readonly user: UserPreferencesManager
 
+	/**
+	 * A manager for the editor's color themes.
+	 *
+	 * @public
+	 */
+	readonly _themeManager: ThemeManager
```
mimecuvalo, line 979: this is @public but has a `_` prefix?

packages/tlschema/src/styles/TLColorStyle.ts
```diff
@@ -651,94 +50,20 @@ export const DefaultLabelColorStyle = StyleProp.defineEnum('tldraw:labelColor',
...
-	theme: TLDefaultColorTheme,
-	color: TLDefaultColorStyle,
-	variant: keyof TLDefaultColorThemeColor
-): string {
-	if (!isDefaultThemeColor(color)) {
-		return color
-	}
-
-	return theme[color][variant]
+export function registerCustomColors(colorNames: TLDefaultColorStyle[]): void {
```
mimecuvalo: does this mean we're going to need / want a `registerCustomFont` system? 
let's discuss this as i think what we pick here will have knock-on effects of how other custom things are registered

steveruizok: I don't think so, the only reason they're here is as convenience to keep you from having to add the color to colorstyle and labelcolorstyle

packages/editor/src/lib/editor/types/SvgExportContext.tsx
steveruizok: Fixed: updated the JSDoc to show a custom theme ID example (`'company-brand'`) and removed the weird double quoting.

packages/editor/src/lib/TldrawEditor.tsx
steveruizok: Done — moved `themes` prop right after `theme` in `TldrawEditorBaseProps`.

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
steveruizok: Fixed — changed RichTextLabel's `verticalAlign` prop type to `'start' | 'middle' | 'end'` (matching PlainTextLabel and `TLDefaultVerticalAlignStyle`), updated all callers, and propagated through GeoShapeUtil and NoteShapeUtil display value types.

packages/editor/src/lib/editor/managers/ThemeManager/ThemeManager.ts
steveruizok: Added a JSDoc comment to the `ThemeManager` class.

packages/tldraw/src/lib/shapes/shared/PlainTextLabel.tsx
steveruizok: Fixed — added `legacyAlign` handling to `RichTextLabel` (and `RichTextSVG`) matching PlainTextLabel's pattern: imported `isLegacyAlign`, added the check, and updated `justifyContent` logic.

packages/editor/src/lib/editor/Editor.ts
steveruizok: Fixed — changed from `@public` to `@internal`.

apps/docs/content/sdk-features/themes.mdx
steveruizok: Good catch — added a `'company-brand'` example to the themes code snippet in the docs to show that custom names beyond 'light' and 'dark' are supported.

apps/examples/e2e/tests/test-text.spec.ts
steveruizok: Agreed — changed `fontSizeAdjustment` to use `1` as identity (no adjustment) instead of `0`. Added a `FontSizeAdjustmentIdentity` migration to convert existing `0` values to `1` in stored documents. Updated migrations, tests, rendering code (`|| 1` → `?? 1`), templates, and docs.

apps/examples/src/examples/configuration/configure-shape-util/README.md
steveruizok: Renamed all `*ShapeUtilOptions` to `*ShapeOptions` — BookmarkShapeOptions, EmbedShapeOptions, GeoShapeOptions, ImageShapeOptions, LineShapeOptions, NoteShapeOptions. Updated exports and README.

packages/tlschema/src/styles/TLColorStyle.ts
steveruizok: Claude on behalf of Steve, leaving a note for later:

**The problem:** `registerColors` lets you add new color names at runtime (they show up in the style panel and pass validation). There's no equivalent for fonts. `DefaultFontStyle` is a fixed enum (`draw`, `sans`, `serif`, `mono`) and `FONT_FAMILIES` is a hardcoded mapping to CSS variables. You can override what CSS family an existing font name maps to via `getDisplayValueOverrides`, but you can't add new font names that appear in the style panel.

**What a `registerFonts` system would need:**
1. `DefaultFontStyle.addValues(...)` — same pattern as `registerColors`, to add new font names to the enum so they pass validation and appear in the style panel
2. A way to register the CSS font-family mapping for each new name (currently hardcoded in `FONT_FAMILIES` in `default-shape-constants.ts`)
3. Font file loading — new fonts need `@font-face` declarations and actual font file URLs, which currently go through `assetUrls.fonts`. This is the part that makes it more complex than colors.
4. UI for the style panel — the font picker would need to show the new entries

**Verdict:** Out of scope for this PR. Could be a follow-up if there's user demand. The color registration pattern (`registerColors`) establishes the precedent, so the API shape is clear — the extra complexity is just the font loading/CSS side.

packages/tldraw/src/lib/shapes/arrow/straight-arrow.ts
steveruizok: Fixed — arrow geometry functions (`getStraightArrowInfo`, `getCurvedArrowInfo`, `getElbowArrowInfo`) now accept an `arrowStrokeWidth` parameter. The `getArrowInfo` cache resolves it from the util's display values, so `getDisplayValueOverrides` on `ArrowShapeUtil` now affects geometry. Bound shape stroke widths still use theme directly since those are separate shapes.

packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
steveruizok: Fixed — Drawing tool state now uses `getDisplayValues(this.util, shape).strokeWidth` for line length threshold checks instead of reading theme directly.

packages/tldraw/src/lib/shapes/arrow/elbow/getElbowArrowInfo.tsx
steveruizok: Addressed as part of the arrow geometry threading — `getElbowArrowInfo` and `getElbowArrowTerminalInfo` now accept and use `arrowStrokeWidth` from display values.

apps/docs/content/sdk-features/themes.mdx
steveruizok: I think so... I guess the user would have to define them anyway.

packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
steveruizok: we could do this keyed on the shape itself, just for when it is read multiple times in a single frame. I'd like to have this cache on the props/meta properties but I don't know if we actually want to do that (it would limit any root-level reactions)

steveruizok: Added a WeakMap cache keyed on the shape record + theme reference. Same shape in the same frame skips recomputation; naturally invalidates when shape changes (new record) or theme changes (different reference).

packages/tlschema/src/shapes/TLNoteShape.ts
steveruizok: good idea!

apps/docs/content/sdk-features/themes.mdx
steveruizok: Update: went ahead and did this after all. Moved all shape-specific color variants out of the theme:

- `TLDefaultColor` now only has generic variants: `solid`, `semi`, `pattern`, `fill`, `linedFill`
- Frame colors → `FrameShapeUtil.options.frameColors`
- Note colors → `NoteShapeUtil.options.noteColors` / `noteBorder`
- Highlight colors → `HighlightShapeUtil.options.highlightColors`

`getDisplayValues`/`getDisplayValueOverrides` now receive the shape options as a 4th parameter. Custom shapes can define their own color mappings without touching the theme type.

apps/examples/src/examples/ui/changing-default-colors/ChangingDefaultColorsExample.tsx
steveruizok: this is the "changing default colors example", which I think is tight enough

steveruizok: (we have other examples for what you're describing, which is adding new colors)

packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
steveruizok: going for it

apps/examples/src/examples/ui/custom-theme/CustomThemeExample.tsx
steveruizok: usually "primary brand" is a single color rather than a set of slight variations on that color, which we demonstrate here. Something conceptual like "warning" or "danger" would be more appropriate, but that's a similar issue. I just want this example to show people how to do things with the API.

packages/tlschema/src/shapes/TLNoteShape.ts
mimecuvalo: this got resolved but it's still an issue!

packages/tlschema/src/shapes/TLNoteShape.ts
```diff
@@ -265,5 +267,27 @@ export const noteShapeMigrations = createShapePropsMigrationSequence({
...
+			id: Versions.MakeFontSizeAdjustmentRatio,
+			up: (props) => {
+				props.fontSizeAdjustment = null
+			},
+			down: (props) => {
+				props.fontSizeAdjustment = 1
+			},
+		},
+		{
+			id: Versions.FontSizeAdjustmentIdentity,
```
mimecuvalo: let's collapse these two migrations for this PR

packages/tlschema/src/styles/TLColorStyle.ts
mimecuvalo: Yeah, definitely more complex for sure (with dealing with assets and all that). Might be good for us to have a followup PR to deal with it as a bunch of users are indeed asking for this! Def. the demand is there.

packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const dvCache = new WeakMap<TLShape, { theme: TLTheme; values: object }>()
+
+/**
+ * Get the resolved display values for a shape, merging the base values with any overrides.
+ *
+ * @public
+ */
+export function getDisplayValues<Shape extends TLShape, DisplayValues extends object>(
```
MitjaBezensek, line 19: Not sure if that's a valid use case, but if `getDisplayValueOverrides` would read anything outside of theme and shape we wouldn't bust the cache. Like if somebody wanted to style a shape differently when it is selected this would not bust the cache.

We can probably wait to see if somebody needs something like that though.

apps/examples/src/examples/ui/dark-mode/DarkModeExample.tsx
MitjaBezensek, line 1: We should probably rename the folder, it still says `infer-dark-mode`, but that's no longer what we showcase.

packages/tlschema/src/shapes/TLNoteShape.ts
MitjaBezensek: Yeah, would be nice to do that. Also seems like `MakeFontSizeAdjustmentRatio` `down` should use 0 instead of 1.

packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
```diff
@@ -76,6 +117,75 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
...
+			},
+		},
+		noteBorder: {
+			light: 'rgb(144, 144, 144)',
+			dark: 'rgb(20, 20, 20)',
+		},
+		getDisplayValues(_editor, shape, theme, options: NoteShapeOptions): NoteShapeUtilDisplayValues {
+			const { color, labelColor, font, size, align, verticalAlign } = shape.props
+			const noteColor =
+				options.noteColors[theme.id]?.[color] ?? options.noteColors['light']?.[color]
```
MitjaBezensek: Why do we default to `light`?

packages/tlschema/src/styles/TLColorStyle.ts
MitjaBezensek: Yeah, people definitely want this. I added a comment in how we could maybe improve `registerColors`, but `registerFonts` could use the exact same system.

packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
steveruizok: This is an opinion

apps/examples/src/examples/ui/dark-mode/DarkModeExample.tsx
steveruizok: that would break the URL, do we care?

mimecuvalo: meh break away

templates/chat/tsconfig.json
```diff
@@ -13,11 +13,22 @@
...
+			{
+				"name": "next"
+			}
+		],
 		"experimentalDecorators": true,
-		"paths": { "@/*": ["./src/*"] }
+		"paths": {
+			"@/*": ["./src/*"]
+		},
+		"target": "ES2017"
```
mimecuvalo: why was ES2017 added? this shouldn't be necessary? 🤔

packages/tlschema/src/shapes/TLNoteShape.ts
mimecuvalo: still an issue

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
mimecuvalo: still need to fix this one too

apps/examples/src/examples/ui/dark-mode/DarkModeExample.tsx
steveruizok: Already done — the folder was renamed from `infer-dark-mode` to `dark-mode` in commit a38b5a96.

steveruizok:
Automated PR Review — written by Codex (OpenAI Codex CLI)

Summary
- Introduces a first-class theme system to Editor/Tldraw with default light/dark palettes, runtime theme overrides, and dynamic display values for shapes.
- Hooks the UI, SVG export, and cursor rendering into the active theme.
- Adds tests, examples, and docs. Overall direction and integration look solid.

What’s Working Well
- ThemeManager and TLTheme/TLThemes are cleanly designed and extensible.
- EnumStyleProp now supports add/remove values at runtime for custom color names.
- registerColorsFromThemes scans themes and augments DefaultColorStyle/DefaultLabelColorStyle; nice DX for custom palettes.
- Display values pipeline (getDisplayValues) cleanly separates style props from final render values.
- SVG export and canvas pattern fills use theme values consistently.
- Good tests for caching, overrides, and theme switching.

Key Issue Fixed (locally)
- Potential undefined theme when resolving display values if a provided themeId is not registered.
- Patch adds a safe fallback to the current theme.
  - File: packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
  - Effect: prevents failures when SVG export or shape utils pass an unknown themeId.
  - Note: This local change isn’t pushed; I can open a commit or suggestion if you want it included in the PR.

Concerns / Suggestions
- registerColorsFromThemes is invoked during render in TldrawEditor; in dev it can repeat warnings. Consider moving it into a layout effect keyed by `themes` to reduce noise.
  - File: packages/editor/src/lib/TldrawEditor.tsx (early render call)
- Consider a helper like `editor.resolveTheme(themeId)` returning a valid TLTheme with a documented fallback; use it in both export and display-value paths.
- Minor: `Editor.getCurrentThemeId()` returns `keyof TLThemes` while `useCurrentThemeId()` returns `string`. It’s fine but worth explicitly documenting.

API/DX Highlights
- TldrawEditor props: `themes` and `theme` are straightforward for consumers.
- Editor: `getThemes`, `updateThemes`, `getCurrentTheme`, `getCurrentThemeId`, `setTheme` provide clear control points.
- TLTheme/TLThemeColors/TLDefaultColor are clear and extensible via module augmentation.

Tests
- displayValues.test.ts has good coverage: theme resolution, cache invalidation, overrides.
- Nice-to-have additions:
  - Export with a non-existent themeId falls back without error (now safe with the fallback).
  - registerColorsFromThemes across multiple custom themes to confirm dedupe and non-duplicated warnings.

Performance / Reliability
- Pattern generation defers/revokes object URLs properly; Safari refresh workaround included.
- Cursor SVG generation uses theme colors with quick reactors; looks good.
- WeakMap cache in getDisplayValues avoids recompute and is validated by tests.

Docs / Examples
- New themes docs align with API.
- Examples (custom-theme, dark-mode, display-options) demonstrate usage well.

Edge Cases
- Custom themes missing certain palette entries: you warn in dev; rendering of shapes using missing palette keys may be undefined. Consider soft fallbacks or a tooling check.
- Migrations/serialization: schema for shapes remains compatible; migrations tests pass. Keep an eye on any downstream loaders relying on old color names.

Small Follow-Ups
- Move `registerColorsFromThemes` into a layout effect in TldrawEditor keyed by `themes`.
- Add a central `resolveTheme(themeId)` helper in Editor and use it across call sites.

Local Change Made
- Safe fallback when resolving display values:
  - packages/tldraw/src/lib/shapes/shared/getDisplayValues.ts
  - Falls back to `editor.getCurrentTheme()` if `themeId` is missing or unknown.
  - Happy to open a commit/suggestion to include this in the PR if desired.

If you want, I can push the tiny fallback fix and add a small test asserting the behavior when an unknown themeId is provided during export.

steveruizok:
Upon closer review, I am going to drop support for custom themes. Instead, we'll stick to customizing light and dark themes. My reasoning is: we don't currently go far enough in supporting custom themes; the two modes are deeply woven through CSS variables, user preferences, etc. Customizing (and extending) the themes is enough, adding more themes needs separate design and deeper cuts.

PR #8222, opened 2026-03-13
feat(translate): add connector arrows between originals and clones via ctrl+alt

max-drake:
In order to make it easy to create linked duplicates of shapes, this PR adds a new modifier interaction to the clone-while-translating flow. When a user holds Alt to clone shapes and then also holds Ctrl, arrow shapes are automatically created connecting each original shape to its clone. Releasing Ctrl removes the connector arrows; releasing Alt stops cloning entirely. The arrows inherit compatible styles (color, size, dash, etc.) from the source shape.

This branch also removes the previous restriction that filtered out group shapes from connector creation — groups now participate in the clone-source map like any other shape.

### Change type

- [x] `feature`

### Test plan

1. Create one or more shapes on the canvas
2. Select them, start dragging, then hold Alt to clone
3. While cloning, press Ctrl — arrows should appear connecting originals to clones
4. Release Ctrl — arrows should disappear
5. Release Alt — clones should disappear and originals resume translating
6. Repeat with grouped shapes — connector arrows should also be created

- [x] Unit tests

### Release notes

- Add connector arrows between originals and clones when holding Ctrl+Alt during translate

### Code changes

| Core code       | +131 / -3   |
| Tests           | +59 / -0    |

---

> [!NOTE]
> **Medium Risk**
> Adds new modifier-driven shape creation/binding behavior inside the translate/cloning interaction, which can affect history, selection state, and arrow binding correctness. Changes are localized but touch core pointer/keyboard flow.
> 
> **Overview**
> **Adds a new Ctrl+Alt interaction during translate-clone**: while Alt-cloning selected shapes, holding Ctrl now creates temporary connector `arrow` shapes bound from each original to its corresponding clone, and releasing Ctrl removes those arrows.
> 
> Cloning now filters out locked shapes, tracks a clone-to-original map to bind connectors correctly, and copies compatible arrow style props from the source shapes when creating the connector arrows. Unit tests cover connector creation when Ctrl is pressed before/after starting cloning and ensure connectors are not created when cloning grouped selections.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 42fc05cb90a3423aef0e69c7b8ec0fa09df7f223. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Translating.ts
@@ -140,7 +140,6 @@ export class Translating extends StateNode {
 
 		if (this.isCloning && this.editor.inputs.getCtrlKey() && !this.isArrowConnecting) {
 			this.startArrowConnecting()
-			return
 		}
 
 		// need to update in case user pressed a different modifier key
@@ -180,7 +179,6 @@ export class Translating extends StateNode {
 		let unlockedSelectedIds = this.editor
 			.getSelectedShapeIds()
 			.filter((id) => !this.editor.isShapeOrAncestorLocked(id))
-		// const originalIds = Array.from(unlockedSelectedShapes.map((shape) => shape.id))
 
 		// If we can't create the shapes, don't even start cloning
 		if (!this.editor.canCreateShapes(unlockedSelectedIds)) return


steveruizok:
Not sure we can land this one. There are quite a lot of collisions between the modifier effects, which mean that a user needs to start dragging before pressing the ctrl key. This breaks the rules of how our modifiers work (they should be able to be toggled before or during the interaction) and leads to situations where the right click menu opens by accident, or an accidental scribble select instead of alt-drag. 

For this type of feature, an app usually has UI that a user can drag from instead. We could also look at switching tools to be selected on keyup rather than keydown, and then using "a" as a kind of modifier, e.g. "if dragging while holding a, do a clone+arrow".  We could also think of an entirely new tool for this, though I don't know if that would make sense. 

Lots of good code ends up on the cutting room floor, this might be one of those. A closely-related feature that I would love to see explored is including bound arrows when cloning shapes, so that if a geo A has an arrow connecting it to geo B, then cloning B (to make C) will also create an arrow from A to C.

PR #8229, opened 2026-03-14
perf(state): skip isEqual for leaf computed signals

steveruizok:
In order to avoid unnecessary equality checks on computed signals that have no downstream dependents, this PR skips the `isEqual` call for leaf computeds (no children, no history buffer). Since nobody inspects `lastChangedEpoch` on a leaf node, the equality check is wasted work.

The optimization is a single condition added to `Computed.ts`: when `!this.historyBuffer && this.children.isEmpty`, we skip `isEqual` and always update the cached state. The `historyBuffer` guard ensures that computeds with diff history (used by `getDiffSince` consumers like `StoreQueries`) still run equality checks to avoid polluting history with spurious entries.

When a child attaches later (leaf → non-leaf), the child captures the current `lastChangedEpoch` via `maybeCaptureParent`, so there's no epoch mismatch — and `children.isEmpty` becomes false, restoring full `isEqual` behavior.

### Change type

- [x] `improvement`

### Test plan

1. Run `cd packages/state && yarn test run` — all 141 tests pass
2. Run `cd packages/store && yarn test run` — all 446 tests pass (validates `getDiffSince` consumers)
3. Run `yarn typecheck` — passes

- [x] Unit tests

### Release notes

- Improved performance of reactive computed signals by skipping equality checks on leaf nodes with no dependents.

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +4 / -1    |
| Tests      | +108 / -2  |

---

> [!NOTE]
> **Medium Risk**
> Touches core computed-signal caching logic and intentionally changes leaf computed behavior (may break callers relying on referential stability or `lastChangedEpoch` semantics), though the change is gated to nodes with no children and no history buffer.
> 
> **Overview**
> **Optimizes computed evaluation** by skipping `isEqual` checks when a computed is a *leaf* (no children) and has no `historyBuffer`, always updating the cached value in that case.
> 
> **Extends unit coverage** to assert the new behavior: leaf computeds may store new references even when values are equal, while computeds with dependents or with history enabled still run `isEqual` and deduplicate as before (including when children are added/removed at runtime).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4db0f058267f0c0d773a229e12a3ed8563f53b13. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/tldraw/src/test/styles2.test.tsx
+++ b/packages/tldraw/src/test/styles2.test.tsx
@@ -140,7 +140,7 @@ describe('Editor.styles', () => {
 			},
 		])
 
-		expect(editor.getSharedStyles()).toStrictEqual(initialStyles)
+		expect(editor.getSharedStyles()).toBe(initialStyles)
 	})
 })
 


steveruizok:
I did a test to log out every time the `!this.historyBuffer && this.children.isEmpty` path was hit and its very often. So quite a lot of skipped `isEqual` checks.

steveruizok:
Closing this — the optimization is unsound. Skipping `isEqual` for leaf computeds breaks reference stability: derive functions that create new objects (like `getSharedStyles()`) would return a different reference even when the value is semantically equal. Any code relying on `===` across re-derivations would see false negatives.

The `styles2.test.tsx` failure caught a real regression, not a test that needed updating. `isEqual` exists to deduplicate values and maintain reference stability, not just to prevent downstream epoch propagation.

steveruizok:
sad

PR #8238, opened 2026-03-15
perf(editor): use callback refs instead of useTransform hook

### LINKED ISSUE #8234: useTransform hook runs useLayoutEffect on every render
`useTransform` in `packages/editor/src/lib/hooks/useTransform.ts` calls `useLayoutEffect` without a dependency array, causing it to run on every render.

```ts
useLayoutEffect(() => {
  const elm = ref.current
  if (!elm) return
  if (x === undefined) return

  let trans = `translate(${x}px, ${y}px)`
  // ...
  elm.style.transform = trans
})
// ^ no dependency array
```

The effect only reads `ref`, `x`, `y`, `scale`, `rotate`, and `additionalOffset` — it should include these as dependencies so it only re-runs when the transform values actually change.

This hook is used by `DefaultBrush`, `DefaultCursor`, `DefaultCollaboratorHint`, `DefaultSelectionBackground`, `DefaultSelectionForeground`, and `TldrawSelectionForeground`, so unnecessary re-executions could add up.

### Expected behavior

The effect should have a dependency array:

```ts
useLayoutEffect(() => {
  // ...
}, [ref, x, y, scale, rotate, additionalOffset])
```

Or, if the intent is to run on every render as a deliberate choice (e.g. because callers always pass fresh values), this should be documented with a comment explaining why.

steveruizok:
👨‍🦰 It's not totally clear to me whether the overhead of assigning a new ref on each render is greater or lesser than the overhead of busting the useLayoutEffect on every render. My hunch is that it is cheaper overall.

---

Closes #8234

Introduces a new `setTransform` function that directly sets an element's CSS transform. Internal components now use callback refs with `setTransform` instead of the `useTransform` hook, avoiding the overhead of `useRef` + `useLayoutEffect` on hot paths like cursor, brush, and selection overlays.

The `useTransform` hook is preserved (now a thin wrapper around `setTransform`) for backward compatibility.

`setTransform` takes `offsetX`/`offsetY` as separate number parameters instead of a `VecLike` object, avoiding object allocation at call sites.

### Change type

- [x] `improvement`

### Test plan

1. Run `yarn dev` and verify selection foreground/background, brush, cursor, and collaborator hints still render and update correctly when interacting with shapes
2. Verify transforms update when moving/rotating/scaling shapes

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add `setTransform` helper for applying CSS transforms directly to DOM elements.
- Internal overlay components now use callback refs instead of `useTransform` for slightly less overhead on hot paths.

### API changes

- Added `setTransform(elm, x, y, scale, rotate, offsetX, offsetY)` — public helper to set CSS transform on a DOM element

### Code changes

| Core code       | +72 / -63  |
| Documentation   | +2 / -7    |

---

> [!NOTE]
> **Medium Risk**
> Touches transform application for hot-path overlay UI (cursor/brush/selection), so subtle positioning/rotation regressions are possible despite being a localized refactor. Also adds a new public API surface (`setTransform`) that needs to remain stable.
> 
> **Overview**
> Adds a new public `setTransform(elm, x, y, scale, rotate, offsetX, offsetY)` helper and reworks `useTransform` into a thin wrapper around it.
> 
> Updates multiple overlay components (brush, cursor, collaborator hint, selection foreground/background, and the examples app’s custom brush) to use callback refs that call `setTransform` directly, reducing per-render `useRef`/`useLayoutEffect` overhead. Selection foreground/background also adjust to set dimensions/expanded bounds inline with the new ref pattern.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0432b31159fa1663244257cd410f35f5020e20c8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/components/default-components/DefaultSelectionForeground.tsx
+++ b/packages/editor/src/lib/components/default-components/DefaultSelectionForeground.tsx
@@ -21,10 +21,11 @@ export function DefaultSelectionForeground({ bounds, rotation }: TLSelectionFore
 		? editor.getShapeUtil(onlyShape).expandSelectionOutlinePx(onlyShape)
 		: 0
 
-	bounds =
-		expandOutlineBy instanceof Box
+	const expandedBounds = expandOutlineBy
+		? expandOutlineBy instanceof Box
 			? bounds.clone().expand(expandOutlineBy).zeroFix()
 			: bounds.clone().expandBy(expandOutlineBy).zeroFix()
+		: bounds
 
 	return (
 		<svg
@@ -36,8 +37,8 @@ export function DefaultSelectionForeground({ bounds, rotation }: TLSelectionFore
 		>
 			<rect
 				className={classNames('tl-selection__fg__outline')}
-				width={toDomPrecision(bounds.width)}
-				height={toDomPrecision(bounds.height)}
+				width={toDomPrecision(expandedBounds.width)}
+				height={toDomPrecision(expandedBounds.height)}
 			/>
 		</svg>
 	)


steveruizok:
Unclear whether this is an improvement. I think we're heading in the right direction though, which is to remove hook overhead for managing refs. Considering a centralized store of refs that does this type of imperative work using a reactor.

PR #8243, opened 2026-03-15
perf(editor): improve history system performance and fix pending diff loss bug

steveruizok:
In order to reduce allocations on hot paths and fix a data-loss bug in the history system, this PR makes several targeted improvements to the undo/redo infrastructure in `@tldraw/store` and `@tldraw/editor`.

**Bug fix:** `bailToMark()` (and `_undo` with `toMark`) would silently discard the pending diff when the target mark didn't exist in the undo stack. The pending diff was cleared at the start of `_undo` but never restored on early return, causing accumulated changes to be lost. Now the pending diff is restored if the mark is not found.

👨‍🦰 not sure if this is a real problem, it might be adding complexity to protecting against a problem that doesn't really come up in practice

**Allocation reductions:** `isRecordsDiffEmpty` previously allocated 3 arrays per call via `Object.keys().length` — replaced with zero-allocation `for...in` early returns. `PendingDiff.apply()` no longer wraps a single diff in an array to call `squashRecordDiffsMutable` — a new `applyDiffToTarget` function handles single diffs directly. Redo stack clearing now short-circuits when the redo stack is already empty, avoiding closure + object allocation on every user change.

👨‍🦰 good bot

**Undo stack limit:** Added a `MAX_UNDO_STACK_SIZE` of 10,000 entries. When the stack exceeds this limit, oldest entries are trimmed. This prevents unbounded memory growth in long editing sessions.

👨‍🦰 good bot

### Change type

- [x] `improvement`

### Test plan

1. Open a tldraw editor and perform various draw/undo/redo operations
2. Verify undo/redo works correctly after many operations
3. Verify that `bailToMark` with an invalid mark ID does not lose pending changes

- [x] Unit tests

### Release notes

- Fix a bug where calling `bailToMark` with a non-existent mark ID could silently discard pending history changes.
- Reduce allocations on the history system hot path (every user change).
- Add an undo stack size limit (10,000 entries) to prevent unbounded memory growth in long sessions.

### API changes

- Added `applyDiffToTarget()` (`@internal`) for applying a single diff without array wrapping
- Added `MAX_UNDO_STACK_SIZE` (`@internal`) constant export from `HistoryManager`

### Code changes

| Core code       | +100 / -44 |
| Tests           | +53 / -1   |

---

> [!NOTE]
> **Medium Risk**
> Touches core undo/redo logic and diff squashing, so regressions could impact history correctness despite being internal/perf-focused changes.
> 
> **Overview**
> Improves history performance and robustness by reducing allocations in diff handling (`isRecordsDiffEmpty` no longer uses `Object.keys`, and new internal `applyDiffToTarget` replaces single-item `squashRecordDiffsMutable` usage).
> 
> Fixes a bug where `bailToMark`/`_undo` would clear and lose the pending diff when the requested mark didn’t exist (now restored via `PendingDiff.restore`).
> 
> Adds an internal `MAX_UNDO_STACK_SIZE` (10,000) and trims the undo stack after marks/undo/redo to prevent unbounded memory growth, with new unit tests covering trimming and the pending-diff preservation case.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 865e097de79cc9202cddf15db0469346b3051d6b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/editor/src/lib/editor/managers/HistoryManager/HistoryManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/HistoryManager/HistoryManager.test.ts
@@ -1,7 +1,7 @@
 import { BaseRecord, RecordId, Store, StoreSchema, createRecordType } from '@tldraw/store'
 import { vi } from 'vitest'
 import { TLHistoryBatchOptions } from '../../types/history-types'
-import { HistoryManager } from './HistoryManager'
+import { HistoryManager, MAX_UNDO_STACK_SIZE } from './HistoryManager'
 
 interface TestRecord extends BaseRecord<'test', TestRecordId> {
 	value: number | string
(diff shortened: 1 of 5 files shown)


steveruizok:
Not really an improvement, though I'm extracting a PR for the history bug.

Issue #8251, opened 2026-03-17
Context menu flashes and disappears when right-clicking with one already open


max-drake:
## Description

Right-clicking on the canvas while a context menu is already open causes the new context menu to briefly flash and then disappear, along with the original one.

## Steps to reproduce

1. Open tldraw
2. Right-click on the canvas to open a context menu
3. Without closing the first context menu, right-click somewhere else on the canvas

## Expected behavior

A new context menu should open at the new right-click position, replacing the old one.

## Actual behavior

The new context menu flashes briefly and then both menus disappear. The user is left with no context menu open.

max-drake:
## Research findings

### Root cause analysis

The issue is a race condition between the `MenuClickCapture` overlay and Radix UI's context menu handling.

**What happens when you right-click while a context menu is already open:**

1. The `MenuClickCapture` overlay (`packages/editor/src/lib/components/MenuClickCapture.tsx`) is visible because a menu is open
2. The right-click `pointerdown` event lands on the overlay, not on the Radix `ContextMenu.Trigger`
3. `handlePointerDown` (line 34-47) calls `editor.menus.clearOpenMenus()` **unconditionally** on line 45 — regardless of which mouse button was pressed
4. This clears tldraw's menu state, which unmounts the menu content (line 108 of `DefaultContextMenu.tsx`: `{isOpen && ...}`) and also removes the `MenuClickCapture` overlay
5. The `contextmenu` event fires next — now the overlay is gone, so it may reach the Radix trigger, causing a brief flash as a new menu tries to open
6. But the state transitions conflict and the menu closes again immediately

**The core problem**: `MenuClickCapture.handlePointerDown` clears all menus on every pointer down (line 45), but only sets up drag-forwarding for left-clicks (`e.button === 0`, line 36). Right-clicks clear the menus without giving Radix a clean opportunity to reopen a context menu at the new position.

### Relevant files

| File | Lines | Role |
|------|-------|------|
| `packages/editor/src/lib/components/MenuClickCapture.tsx` | 34-47 | Clears all menus on any `pointerdown`, intercepts right-clicks |
| `packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx` | 103-126 | Radix context menu wrapper, conditional render on `isOpen` |
| `packages/tldraw/src/lib/ui/hooks/useMenuIsOpen.ts` | 6-29 | Menu open state hook |
[...]

PR #8261, opened 2026-03-18
fix(editor): Fix context menu flashing when right-clicking with one open

kaneel:
1. The MenuClickCapture overlay's clearOpenMenus() call raced with Radix's contextmenu handling, causing a close→reopen→close sequence. Removing it is safe because Radix's DismissableLayer already handles menu dismissal on outside clicks, and Editor.dispatch clears menus on pointer_down for drag interactions.

2. Make sure that right clicks are correctly handled (with platform quirks - macOS mostly) and dispatched as right click to the editor

Fixes https://github.com/tldraw/tldraw/issues/8251
(Potentially) fixes https://github.com/tldraw/tldraw/issues/8277?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. On canvas, right click to open a context menu
2. Right click again real quick 
3. First context menu is closed and new context menu opens
4. keep moving and right clicking, no blinking and closing of the newly opened context menu should be observed.
4. Left click anywhere else, it closes the context menu

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where right clicking on canvas quickly runs into a race condition and automatically close the newly open context menu resulting in a non-savoury blinking

### DIFF
--- a/packages/editor/src/lib/components/MenuClickCapture.tsx
+++ b/packages/editor/src/lib/components/MenuClickCapture.tsx
@@ -35,6 +35,9 @@ export function MenuClickCapture() {
 
 	const handlePointerDown = useCallback(
 		(e: PointerEvent) => {
+			const isCoarsePointer = editor.getInstanceState().isCoarsePointer || e.pointerType === 'touch'
+			const isPrimaryPointer =
+				e.button === LEFT_MOUSE_BUTTON || (isCoarsePointer && e.button === -1)
 			// On macOS, ctrl+left-click fires as button 0 with ctrlKey but triggers a
 			// contextmenu event just like a real right-click (button 2). We dispatch
 			// right_click directly so the editor updates state (selection, hovered shape)
@@ -49,17 +52,30 @@ export function MenuClickCapture() {
 					name: 'right_click',
 					...getPointerInfo(editor, e),
 				})
-			} else if (e.button === LEFT_MOUSE_BUTTON) {
+			} else if (isPrimaryPointer) {
+				// Dismiss open menus on primary pointer interactions. Keep this out of the
+				// right-click path to avoid racing with Radix contextmenu open handling.
+				editor.menus.clearOpenMenus()
 				setIsPointing(true)
 				rPointerState.current = {
 					isDown: true,
 					isDragging: false,
 					start: new Vec(e.clientX, e.clientY),
 				}
-				rDidAPointerDownAndDragWhileMenuWasOpen.current = false
+				// On coarse pointers, forward pointer_down immediately so long-press can
+				// retarget/open context menus while one is already open.
+				if (isCoarsePointer) {
+					canvasEvents.onPointerDown?.({
+						...e,
+						button: LEFT_MOUSE_BUTTON,
+					})
+					rDidAPointerDownAndDragWhileMenuWasOpen.current = true
+				} else {
+					rDidAPointerDownAndDragWhileMenuWasOpen.current = false
+				}
 			}
 		},
-		[editor]
+		[canvasEvents, editor]
 	)
 
 	const rDidAPointerDownAndDragWhileMenuWasOpen = useRef(false)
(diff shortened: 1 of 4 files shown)


max-drake:
ooo found a regression here: on pointer down the most recently opened context menu is spawned, and on pointer up it switches to the correct one

https://github.com/user-attachments/assets/59410bcd-7cf2-4593-9e2f-a0fc988b9fa1

kaneel:
> ooo found a regression here: on pointer down the most recently opened context menu is spawned, and on pointer up it switches to the correct one
> Screen.Recording.2026-03-18.at.10.12.42.mov

I think we're going in the right direction though, I'll figure the next step thanks a lot for spotting this!

max-drake:
found another regression: if you 
a) have a context menu already open and
b) right click on a locked shape

you get same issue where the old context menu appears on pointerdown, then on pointerup it changes to the correct context menu

https://github.com/user-attachments/assets/30582847-0fe2-41db-a422-dad874263a07

packages/editor/src/lib/hooks/useGlobalMenuIsOpen.ts
```diff
@@ -19,7 +19,6 @@ export function useGlobalMenuIsOpen(
 			} else {
 				tlmenus.deleteOpenMenu(id)
 			}
-
```
kaneel, line 22: nit: let's remove this change

packages/editor/src/lib/components/MenuClickCapture.tsx
```diff
@@ -35,6 +35,9 @@ export function MenuClickCapture() {
 
 	const handlePointerDown = useCallback(
 		(e: PointerEvent) => {
+			const isCoarsePointer = editor.getInstanceState().isCoarsePointer || e.pointerType === 'touch'
+			const isPrimaryPointer =
+				e.button === LEFT_MOUSE_BUTTON || (isCoarsePointer && e.button === -1)
```
kaneel, line 40: I'm looking at the specs here: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#determining_button_states 

I have troubles wrapping my head around coarse + e-button -1 and I've been refreshing my memory a little – so bear with me – but I think for pointer down, this should not be occuring at all actually, for `pointermove`, 100% sure this is occuring.

packages/editor/src/lib/components/MenuClickCapture.tsx
```diff
@@ -49,17 +52,30 @@ export function MenuClickCapture() {
 					name: 'right_click',
 					...getPointerInfo(editor, e),
 				})
-			} else if (e.button === LEFT_MOUSE_BUTTON) {
+			} else if (isPrimaryPointer) {
+				// Dismiss open menus on primary pointer interactions. Keep this out of the
+				// right-click path to avoid racing with Radix contextmenu open handling.
+				editor.menus.clearOpenMenus()
```
kaneel, line 58: this seemed already handled somewhere else, maybe by Radix itself; so I'm not sure bringing it back is right unless there's one reason I'm not aware of (and should be documented in this comment)

packages/editor/src/lib/components/MenuClickCapture.tsx
```diff
@@ -49,17 +52,30 @@ export function MenuClickCapture() {
...
+					canvasEvents.onPointerDown?.({
+						...e,
+						button: LEFT_MOUSE_BUTTON,
+					})
+					rDidAPointerDownAndDragWhileMenuWasOpen.current = true
+				} else {
+					rDidAPointerDownAndDragWhileMenuWasOpen.current = false
+				}
 			}
 		},
```
kaneel, line 77: I'm not sure about all of this, I've removed this change and ran the tests and they still pass though

packages/tldraw/src/test/ui/ContextMenu.test.tsx
```diff
@@ -50,3 +53,89 @@ it('tunnels context menu', async () => {
...
+		/>,
+		{ waitForPatterns: false }
+	)
+	const canvas = await screen.findByTestId('canvas')
+	fireEvent.contextMenu(canvas)
+	await screen.findByTestId('context-menu')
+
+	const capture = await screen.findByTestId('menu-click-capture.content')
+	fireEvent.pointerDown(capture, {
+		button: -1,
```
kaneel, line 131: I need to figure in what case button is -1 because it looks like we're definitely testing the behaviour that is written!

PR #8272, opened 2026-03-18
fix(tldraw): use DOMParser in stripHtml to prevent XSS via innerHTML

kaneel:
The previous implementation used createHTMLDocument + innerHTML which can trigger side effects like  during parsing. DOMParser produces an inert document with no script execution.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Fixed a security issue within our tldraw stripHtml method by relying on DOMParser.


mimecuvalo:
mm, i think this is fine, the parsed DOM isn't mounted into the live document.

kaneel:
Did not think much and this should be safe

PR #8280, opened 2026-03-18
fix(infra): add retry logic for transient R2 and service binding errors

MitjaBezensek:
Sentry shows multiple transient Cloudflare R2 errors (`internal error (10001)`, `connectivity issue`, `network connection lost`, `unspecified error (0)`) and service binding errors (`proxy request failed`) in the tldrawusercontent worker. Found these when investigating user reports of image uploads not working for them.

This PR wraps R2 `head`, `put`, and `get` calls in `retry()` (3 attempts, 500ms wait) with an error matcher for known transient patterns. The upload body is buffered to `ArrayBuffer` before retries since `ReadableStream` is single-use. Service binding calls (`validateUpload`, `confirmUpload`) get the same treatment.

### Change type

- [x] `bugfix`

### Test plan

- Cannot be manually tested (transient infra errors)

### Release notes

- Fix transient R2 upload/download failures by adding retry logic

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +30 / -7   |
| Apps       | +14 / -4   |

### DIFF
--- a/apps/dotcom/sync-worker/src/worker.ts
+++ b/apps/dotcom/sync-worker/src/worker.ts
@@ -318,7 +318,8 @@ export default class Worker extends WorkerEntrypoint<Environment> {
 				await db
 					.insertInto('asset')
 					.values({ objectName, fileId, userId })
-					.executeTakeFirstOrThrow()
+					.onConflict((oc) => oc.column('objectName').doNothing())
+					.execute()
 				message.ack()
 			} catch (_e) {
 				message.retry({


packages/worker-shared/src/userAssetUploads.ts
```diff
@@ -1,6 +1,20 @@
+import { retry } from '@tldraw/utils'
 import { IRequest } from 'itty-router'
 import { notFound } from './errors'
 
+function isTransientWorkerError(error: unknown): boolean {
```
MitjaBezensek, line 5: Should we just retry on all errors?

Issue #8203, opened 2026-03-11
Curved arrow drifts during undo/redo after deleting a bound shape


AniKrisn:
## Description

When a curved (bent) arrow connects two geo shapes and one shape is deleted, repeatedly undoing and redoing causes the arrow to gradually drift from its original position.

https://github.com/user-attachments/assets/c6b973e3-06cd-4239-9799-2e82d3cc2b1a

## Steps to reproduce

1. Create two geo shapes
2. Connect them with a curved (bent) arrow — drag the arrow's middle handle to add a bend
3. Delete one of the geo shapes
4. Press undo then redo repeatedly
5. Observe the arrow shifting position with each undo/redo cycle

## Expected behavior

The arrow should return to its exact original position after undo/redo cycles.

## Actual behavior

The arrow gradually drifts from its original position with each undo/redo cycle.

## Analysis

When a bound shape is deleted, `onBeforeIsolateFromShape` in `ArrowBindingUtil` calls `updateArrowTerminal()`, which recalculates the arrow's terminal position and bend via `getArrowInfo()` and writes it back with `editor.updateShape()`. This creates a separate history entry for the arrow position change.

During undo/redo, the binding restoration and arrow position restoration are decoupled — they are separate history operations. Since `getArrowInfo()` depends on the current editor state (which differs between undo and redo steps), repeated cycles produce slightly different terminal/bend values, causing cumulative drift.

Related to #8176 (which fixes a crash in the same code path) but distinct — this is about position instability, not a crash.

AniKrisn:
## Research findings

### Relevant files

| Component | File | Lines |
|-----------|------|-------|
| Arrow terminal update | `packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts` | 220-285 |
| Isolation callback | `packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts` | 97-107 |
| beforeDelete side effects | `packages/editor/src/lib/editor/Editor.ts` | ~594-632 |
| deleteShapes | `packages/editor/src/lib/editor/Editor.ts` | ~8522-8552 |
| History manager | `packages/editor/src/lib/editor/managers/HistoryManager/HistoryManager.ts` | 116-188 |
| Curved arrow geometry | `packages/tldraw/src/lib/shapes/arrow/curved-arrow.ts` | 28-220+ |
[...]

PR #8283, opened 2026-03-18
fix(arrows): prevent arrow terminal recalculation during undo/redo

AniKrisn:
For [#8203](https://github.com/tldraw/tldraw/issues/8203).

Bound arrow terminals break during undo/redo of shape deletions. This draft PR documents the issue and the two approaches for fixing it.

### Problem

When a bound shape is deleted, `onBeforeIsolateFromShape` converts the arrow terminal from a binding to an absolute page position before the binding is removed. That is correct during a live delete.

During undo/redo, though, history replay restores a diff that already contains the correct terminal values. The store's `beforeDelete` side effect fires again during replay, `onBeforeIsolateFromShape` runs again, and the arrow terminal is recomputed from intermediate replay state instead of preserving the restored value. The result is that the arrow ends up in the wrong place.

The core issue is that `onBeforeIsolateFromShape` cannot currently distinguish a live edit from a history replay.

### Approaches explored

**1. Explicit provenance on isolation callbacks**

Pass replay context to `onBeforeIsolateFromShape` explicitly, e.g. a `source` on the isolation callback options. This gives the hook the information it actually cares about and keeps the decision at the callback boundary.

The downside here is that it requires API and plumbing changes outside the arrow binding util.

**2. `HistoryManager.isReplaying()` (implemented in this branch)**

Add a dedicated `isReplaying()` signal to HistoryManager, set only while replaying diffs in `_undo()` / `redo()`, and have ArrowBindingUtil bail out of `onBeforeIsolateFromShape` when that signal is true.

This is the smallest practical fix and avoids changing semantics elsewhere. It is also more precise than reusing `isPaused()`, because replay and `{ history: 'ignore' }` are different cases.

The downside is that it is more implicit than explicit callback provenance, and ArrowBindingUtil has to reach across an API boundary to ask history whether replay is in progress.

### Change type

- [x] `bugfix`

### Test plan

1. Create two shapes and connect them with an arrow
2. Delete the target shape — arrow terminal should snap to an absolute position
3. Undo — arrow should return to its original bound position
4. Redo — arrow terminal should match the position from step 2

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix arrows jumping to incorrect positions after undoing or redoing shape deletions involving bindings.

### API changes

- Added `HistoryManager.isReplaying()` (internal)

### Code changes

| Core code       | +25 / -10  |
| Tests           | +32 / -0   |


packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
```diff
@@ -94,11 +94,12 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
 		reparentArrow(this.editor, binding.fromId)
 	}
 
-	// when the arrow is isolated we need to update it's x,y positions
+	// when the arrow is isolated we need to update its (x,y) positions
```
kaneel, line 95: yes!

mimecuvalo (commented):
this is an interesting PR @AniKrisn! i think this might be the right approach but just to check: are there other things that might benefit from leaning on this `isReplaying` check? i want to make sure we're not exposing this bit only for the benefit of arrow terminals and thinking how this could help in general. thoughts?

kaneel:
it works but I cannot exactly pinpoint how this feels like we're building a guard against an other problem tht somewhere else.

~~The problem lies within how we recalculate, not only we're not recalculating "very well" when we remove a shape but undo -> redo recalculate it "worst"~~

https://github.com/user-attachments/assets/d9a0a250-bbcd-474a-a73f-1c18813ee8d6

Actually I don't think it's fixable in any way, we'll always end up in a complicated situation when we delete a shape an arrow is bound because the curvature is calculated differently, takes this example into account where I bound an arrow, copied/pasted it and look the curvature… that's what we're basing ourselves on when we recalculate the position of the arrow (it ends around where the second arrow start touching the geo shape)

image

So the real problem is that we're having this recalculation on redo when it should just apply the state

kaneel:
I've tried to run a quick spike and that seems to be very similar a solution: https://github.com/tldraw/tldraw/pull/9079

Actually it's a bit different and more global, @mimecuvalo what do you think? I've chatted with @AniKrisn about it so he can take the inspiration from this draft and go ahead with having the tests written. This would also expose a method so people can reuse it for their own custom shapes

AniKrisn:
thanks for the review! @mimecuvalo 

i guess that the general case that warrants a public editor api rather than this private `HistoryManager` change would be where there are binding side effects for other binding utils that replay on redo and overwrite a correct value in the recorded diff

so i checked the other binding utils we have (connections in workflow and image pipeline + the globs and pin binding examples) - and it seems like arrows are the only case in which we freeze the remaining visual state upon isolation. for the other binding utils, we remove the connection when it gets separated from a node, so this is not a bug that appears there.

i suppose there could be cases for SDK users where they create custom shapes that have these binding isolation side-effects, but it seems fairly narrow and niche to expose a general purpose api for that; not sure.

that being said, the arrow case alone seems nontrivial and worth a fix

perhaps @steveruizok will want to weigh in here?

packages/editor/src/lib/editor/managers/HistoryManager/HistoryManager.ts
```diff
@@ -172,15 +182,15 @@ export class HistoryManager<R extends UnknownRecord> {
...
+				// whoops, we didn't find the mark we were looking for
+				// don't do anything
 				return this
 			}
-
 			this.store.applyDiff(diffToUndo, { ignoreEphemeralKeys: true })
 			this.store.ensureStoreIsUsable()
 			this.stacks.set({ undos, redos })
 		} finally {
+			this._isReplaying = previousIsReplaying
```
mimecuvalo, line 197: hm, if `previousIsReplaying` is set to `true` how does this ever revert back to `false`? 🤔

AniKrisn: claude: "This line restores whatever the flag was when this call started — the snapshot is taken at the top of the method, before we set it to true. For a top-level undo/redo that snapshot is always false, so this is exactly the line that reverts it. It's only true if undo/redo re-enters mid-replay, and in that case the outer call's finally is the one that ends at false (same save/restore pattern as previousState on the next line)." :D

mimecuvalo: I see — thx for checking, I see it now, was hard to wrap my brain around for a sec.

mimecuvalo (approved):
ok! let's do this 🤘

PR #8284, opened 2026-03-19
chore(mermaid): add a README

kaneel:
Absolutely did not notice but… we're missing a README for the mermaid package, which is OK, is new, but come on! README Time!

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Release notes

- Added a README to our mermaid package!

### DIFF
--- a/packages/mermaid/README.md
+++ b/packages/mermaid/README.md
@@ -29,19 +29,27 @@ await createMermaidDiagram(
 )
 ```
 
-Shapes are created at the center of the viewport by default. Pass a position to place them elsewhere:
+By default, shapes are centered on the viewport. You can control placement with `blueprintRender`:
 
 ```tsx
+// Center the diagram on a specific point (default behavior)
 await createMermaidDiagram(editor, diagramText, {
 	blueprintRender: {
-		position: { x: 200, y: 100 },
+		position: { x: 500, y: 300 },
 		centerOnPosition: true,
+	},
+})
+
+// Place the diagram's top-left corner at the given point
+await createMermaidDiagram(editor, diagramText, {
+	blueprintRender: {
+		position: { x: 0, y: 0 },
+		centerOnPosition: false,
 	},
 })
 ```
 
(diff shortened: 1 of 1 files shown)


packages/mermaid/README.md
```diff
@@ -0,0 +1,145 @@
...
+```tsx
+await createMermaidDiagram(editor, diagramText, {
+	blueprintRender: {
+		position: { x: 200, y: 100 },
+		centerOnPosition: true,
+	},
+})
+```
+
+If you wish for the diagram not to be centered on the position, or centered in
```
angrycaptain19: What do we mean by "centered on the position" here?

packages/mermaid/README.md
```diff
@@ -0,0 +1,145 @@
...
+    ExtractBlueprint --> CreateShapes["Create tldraw shapes"]
+```
+
+1. **Parse and render** — Mermaid parses the text and renders an SVG into an offscreen DOM element, giving us accurate layout positions via `getBBox()`.
+2. **Extract blueprint** — a diagram-specific converter reads positions from the SVG and semantic data from Mermaid's internal database, producing a `DiagramMermaidBlueprint`.
+3. **Create shapes** — `renderBlueprint` converts the blueprint into tldraw geo shapes, arrows, lines, frames, and groups.
+
+## Lazy loading
+
+The `mermaid` dependency is roughly 2 MB. If you're integrating this into a paste handler or similar, consider lazy-loading the package so it's only fetched when actually needed:
```
angrycaptain19: Might be good to also include an example of integrating it into a paste handler since I think that will be pretty common behaviour

packages/mermaid/README.md
kaneel: HA! one of the few parts I've written myself and you find it :D
centerOnPosition means that the mermaid diagram will be centered on the point. 
If not centered on position, that is the center of the screen, the mermaid diagram top left corner will be at the center of the screen.

PR #8290, opened 2026-03-19
feat(editor): add clipboard hooks to TldrawOptions

### LINKED ISSUE #7547: Allow disabling native copy/paste behavior
## Problem

Users cannot fully disable copy and paste functionality in tldraw. Even when disabling the copy/paste actions in the actions API, the native clipboard behavior still works because it's handled separately via DOM event listeners.

There are two separate systems handling clipboard operations:

1. **Actions API** (`actions.tsx:942-976`) - Defines `copy`, `paste`, `cut` actions for menu-based interactions
2. **Native clipboard events** (`useNativeClipboardEvents` in `useClipboardEvents.ts:716-835`) - Directly listens to `copy`, `paste`, `cut` DOM events

Users who disable or override the actions only affect menu-triggered operations, not keyboard shortcuts (Cmd/Ctrl+C/V/X) which go directly through the native event listeners.

## Use cases

- Building a view-only or limited-edit experience where users shouldn't copy content
- Implementing custom clipboard behavior that completely replaces the default
- Restricting copy/paste in collaborative environments for content protection
- Educational contexts where copying answers should be prevented

### Option A: Editor option
Add an option like `editor.options.enableNativeClipboard` or `editor.options.clipboard`:
```typescript
const editor = new Editor({
  options: {
    clipboard: {
      copy: false,
      paste: false,
      cut: false,
    }
  }
})
```

### Option B: Expose native clipboard as actions
Add `native-copy`, `native-paste`, `native-cut` actions (or make existing actions cover both paths):
```typescript
// users could then override:
overrides: {
  actions: {
    'native-copy': () => null, // disable
    'native-paste': () => customPasteHandler(),
  }
}
```

### Option C: Hook into existing actions
Make the native clipboard event handlers respect the actions API - if an action is disabled or overridden, the native handler should defer to that.

## Related files

- `packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts` - Native clipboard handling
- `packages/tldraw/src/lib/ui/context/actions.tsx` - Actions API definitions

ds300:
Closes #7547

In order to let SDK consumers intercept and customize clipboard copy, cut, and paste from both keyboard shortcuts and app-invoked clipboard reads, this PR adds `TldrawOptions` hooks and wires them through `@tldraw/tldraw` clipboard handling. Integrators can filter or transform serialized content before it hits the clipboard, filter parsed paste payloads before shapes are created, or handle raw clipboard data before tldraw parses it (`native-event` `ClipboardEvent`/`DataTransfer` vs `clipboard-read` `ClipboardItem[]`). Keyboard paste invokes the raw hook synchronously on the `paste` event so `clipboardData` stays usable.

This was requested by the Replit team, who need to filter certain shapes from clipboard operations and work with platform clipboard APIs. Previously, keyboard clipboard shortcuts could not be intercepted in a structured way.

### Change type

- [x] `api`

### API changes

- Added `TLClipboardWriteInfo` for copy/cut metadata: `operation` (`copy` | `cut`) and `source` (`native` | `menu`)
- Added `TLClipboardPasteRawInfo` as a discriminated union: `source: 'native-event'` (`ClipboardEvent`, `clipboardData`, `point`) vs `source: 'clipboard-read'` (`clipboardItems`, `point`)
- Added `TldrawOptions.onBeforeCopyToClipboard` to filter or transform `TLContent` before clipboard writes; return `false` to cancel and preserve selection on cut
- Added `TldrawOptions.onBeforePasteFromClipboard` to filter or transform parsed `TLExternalContent` before placement; clipboard paste only (not drops); return `false` to cancel; `source` is `native-event` | `clipboard-read`
- Added `TldrawOptions.onClipboardPasteRaw` to intercept raw clipboard data before tldraw parses it; return `false` to cancel tldraw handling for that gesture
- Exported `handleNativeOrMenuCopy` from `@tldraw/tldraw` for custom clipboard write flows

### Test plan

1. Open the `Clipboard events` example at `/clipboard-events`
2. Copy/paste with keyboard shortcuts and verify the event log shows `copy` / `cut` with `native` and paste events with `native-event`
3. Toggle `Block copy/cut` and verify copy/cut are cancelled without deleting the selection on cut
4. Toggle `Block paste`, `Filter red on copy`, `Filter red on paste`, and `Handle raw paste (take over)` and verify the logged behavior matches the toggle state

- [x] Unit tests
- [ ] End to end tests
- [x] Typecheck

### Code changes

| Core code       | +358 / -102 |
| Tests           | +511 / -0 |
| Automated files | +45 / -1 |
| Documentation   | +329 / -0 |

### Release notes

- Add `onBeforeCopyToClipboard` and `onBeforePasteFromClipboard` to `TldrawOptions` for filtering or transforming clipboard content before copy/cut and after paste is parsed.
- Add `onClipboardPasteRaw` for handling raw clipboard data before tldraw's default paste pipeline.
- Add paste source names `native-event` and `clipboard-read` to distinguish DOM paste events from explicit Clipboard API reads.
- Export `handleNativeOrMenuCopy` from the `tldraw` package for custom clipboard flows.

### DIFF
--- a/apps/dotcom/client/src/tla/utils/FeatureFlagPoller.test.ts
+++ b/apps/dotcom/client/src/tla/utils/FeatureFlagPoller.test.ts
@@ -2,9 +2,8 @@ import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
 import type { FeatureFlags } from './FeatureFlagPoller'
 
 const mockFetch = vi.fn()
-vi.mock('tldraw', async () => {
-	const actual = await vi.importActual<typeof import('tldraw')>('tldraw')
-	return { ...actual, fetch: (...args: any[]) => mockFetch(...args) }
+vi.mock('tldraw', () => {
+	return { fetch: (...args: any[]) => mockFetch(...args) }
 })
 
 function makeFlags(overrides: Partial<FeatureFlags> = {}): FeatureFlags {


apps/examples/e2e/tests/test-clipboard-events.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+		const modifier = isMac ? 'Meta' : 'Control'
+		await page.keyboard.down(modifier)
+		await page.keyboard.press('KeyC')
+		await page.keyboard.up(modifier)
+		await sleep(100)
+
+		const log = await page.evaluate(() => (window as any).__tldraw_clipboard_log)
+		expect(log).toHaveLength(1)
+		expect(log[0]).toMatchObject({ action: 'copy', source: 'native', prevented: true })
+	})
```
kaneel: this does test we're logging the right event but I think we could also test that the clipboard is empty

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -731,6 +740,14 @@ export function useNativeClipboardEvents() {
 			}
 
 			preventDefault(e)
+
+			if (editor.options.onClipboardCopy) {
+				if (editor.options.onClipboardCopy({ editor, source: 'native' }) === true) {
```
kaneel: this could be just `if (editor.options.onClipboardCopy) && editor.options.onClipboardCopy({ editor, source: 'native' }) === true)` ?

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -744,6 +761,14 @@ export function useNativeClipboardEvents() {
 				return
 			}
 			preventDefault(e)
+
+			if (editor.options.onClipboardCut) {
+				if (editor.options.onClipboardCut({ editor, source: 'native' }) === true) {
```
kaneel: same here, feels like a one condition opportunity!

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
```diff
@@ -784,6 +809,21 @@ export function useNativeClipboardEvents() {
...
 
+			if (editor.options.onClipboardPaste && e.clipboardData) {
+				if (
+					editor.options.onClipboardPaste({
+						editor,
+						source: 'native',
+						clipboardData: e.clipboardData,
+						point,
+					}) === true
+				) {
```
kaneel: oneline too?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -971,6 +971,10 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					if (!canApplySelectionAction()) return
 					if (mustGoBackToSelectToolFirst()) return
 
+					if (editor.options.onClipboardCut) {
+						if (editor.options.onClipboardCut({ editor, source: 'menu' }) === true) return
+					}
```
kaneel: one line?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -984,6 +988,10 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 					if (!canApplySelectionAction()) return
 					if (mustGoBackToSelectToolFirst()) return
 
+					if (editor.options.onClipboardCopy) {
+						if (editor.options.onClipboardCopy({ editor, source: 'menu' }) === true) return
+					}
```
kaneel: ☝️ one line

packages/tldraw/src/lib/ui/hooks/clipboard/putPastedContent.ts
```diff
@@ -0,0 +1,14 @@
+import { Editor, TLExternalContent } from '@tldraw/editor'
+
+/** @internal */
+export async function putPastedExternalContent(
+	editor: Editor,
+	content: TLExternalContent<unknown>
+) {
+	if (editor.options.onBeforePasteFromClipboard) {
+		const result = editor.options.onBeforePasteFromClipboard({ editor, content })
```
kaneel: should we consider users using asynchronous callbacks and should we await here just in case. Worth flagging as if this occurs, we'd be passing a promise to `putExternalContent` and I'm not sure how that would be handled.

kaneel (commented):
This is a very strong PR. 
I've had few comments but they're all truly nits 

I've had one major question regarding async callbacks and how that could – if it could – become a problem.

apps/examples/e2e/tests/test-clipboard-events.spec.ts
ds300: this was removed during a refactor

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
ds300: this was also removed during a refactor

packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
ds300: this was also removed during a refactor. i don't know why cursor was doing these weird nested ifs

packages/tldraw/src/lib/ui/context/actions.tsx
ds300: same

packages/tldraw/src/lib/ui/hooks/clipboard/putPastedContent.ts
ds300: good shout, i'll do this

steveruizok:
Looks good to me but we'll want to test this on a few different browsers. IIRC any async code in the clipboard API (e.g. right click, paste) will cause the event to fail on Safari, I believe we have some weird code juggling there to avoid these quirks.

Adding the async hooks option to the example so that we can test those too. Seems to all work for me.

PR #8294, opened 2026-03-19
perf(tooling): speed up pre-commit hook

mimecuvalo:
In order to make local commits less painful when only a subset of packages change, this PR narrows what the Husky pre-commit hook runs: `yarn install` only when a `package.json` is staged, `build-api` only for packages with staged changes (via lazy filters), `build-i18n` only when `apps/dotcom/client` files are staged, and `refresh-assets` can be skipped when invoked from that filtered `build-api` path if nothing relevant to assets changed. A few verbose `typecheck` script logs were also trimmed.

Personally I often ended up committing with \`--no-verify\` because the hook could take long enough that skipping checks felt worth the risk. On this machine a typical commit now finishes in about **~7s** instead of sometimes **~25s**.

### Change type

- [x] `other`

### Test plan

1. Stage a change under a single package (e.g. `packages/editor`) and commit; confirm only that package’s `build-api` runs (and API reports re-staged as needed).
2. Stage a change with no `packages/*` API impact; confirm `build-api` is skipped when appropriate.
3. Stage only non-dotcom files; confirm `build-i18n` does not run.
4. Stage a `package.json` change; confirm `yarn install --immutable` runs.

- [ ] Unit tests
- [ ] End to end tests


internal/scripts/refresh-assets.ts
```diff
@@ -50,6 +51,35 @@ const collectedAssetUrls: Record<
 
 const mergedIconFooter = '</svg>'
 
+function shouldSkipRefreshAssets(): boolean {
```
mimecuvalo, line 54: @MitjaBezensek i know you were recently looking at the refresh-assets stuff, this only runs on pre-commit so i think it's fine though

PR #8306, opened 2026-03-20
feat(embeds): support pasting arbitrary iframe embed codes

mimecuvalo:
This PR allows pasting any `<iframe>` embed code onto the canvas to create an embed shape. Previously, only iframe URLs that matched a known embed definition (YouTube, Google Maps, etc.) would work, and the iframe had to be the only element in the pasted content. Now any valid iframe with an HTTP(S) src creates an embed directly.

Screenshot 2026-03-20 at 18 11 27

Examples that now work:
- OpenStreetMap: `<iframe width="425" height="350" src="https://www.openstreetmap.org/export/embed.html?bbox=-0.1258492469787598%2C51.558009663698876%2C-0.09986400604248047%2C51.57165515830595&amp;layer=mapnik" style="border: 1px solid black"></iframe>`
- SoundCloud: `<iframe src="https://w.soundcloud.com/player/?visual=true&url=https%3A%2F%2Fapi.soundcloud.com%2Ftracks%2F2184622839&show_artwork=true" style="top: 0; left: 0; width: 100%; height: 100%; position: absolute; border: 0;" allowfullscreen></iframe>`
- Loom: `<iframe src="https://www.loom.com/embed/e5b8c04bca094dd8a5507925ab887002?hideEmbedTopBar=true&t=20s" frameborder="0" webkitallowfullscreen mozallowfullscreen allowfullscreen style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></iframe>`

(screenshot coming)

### Change type

- [x] `feature`

### Test plan

1. Copy an iframe embed code from any service (OpenStreetMap, SoundCloud, Loom, etc.)
2. Paste onto the tldraw canvas
3. An embed shape should be created with the iframe content rendered inside
4. The embed should be resizable

- [x] Unit tests

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +55 / -39  |
| Tests      | +211 / -0  |

### Release notes

- Add support for pasting arbitrary `<iframe>` embed codes onto the canvas. Any iframe from services like OpenStreetMap, SoundCloud, Loom, and others can now be pasted directly to create an embed shape.


mimecuvalo:
easter egg for @kaneel in the screenshot ;)

steveruizok:
This reminds me, we should make tldraw compatible with iframely so that you can embed it in Linear tickets.

steveruizok:
I am concerned with pasting iframes here because it would open us up to being a vector for sneaky things. I remember Ives from CodeSandbox talking about how much that project was dragged down by fraudulent users sending .csb.com links around. Good that we've locked down the feature to just https here though.

Just to raise it before we land, is this a concern? Are we taking on any new risk?

mimecuvalo:
> Just to raise it before we land, is this a concern? Are we taking on any new risk?

Good question to raise, I think the main thing is that:
- everything is `sandbox`'d in `EmbedShapeUtil`
- and we use CSP on dotcom

With those mitigations in place i think we should be good. I'm not sure about the CodeSandbox example that you mentioned, is there more info/link to the vulnerability they ran into there?

> Good that we've locked down the feature to just https here though.

well, it's locked to http or https - it's preventing `javascript:`-type stuff

PR #8322, opened 2026-03-23
refactor(@tldraw/mermaid): streamline node creation for extensibility

kaneel:
Refactor the code around blueprint creation as well as the rendering by making sure shape types are defined at one level only, the blueprint creation, so we can easily allow extensibility through passing a function that can map each diagram node types to different, preferred shapes, or simply custom shapes.

Also allows to pass a function to take full control of the node creation, if ever needed. 

### Boring bullet points!

- Add MermaidBlueprintNode.render (geo vs custom shape type + props) and diagramKind on the blueprint IR.
- Optional per-diagram mapNodeToRenderSpec when building blueprints; default mapping matches previous geo behavior.
- renderBlueprint creates nodes from node.render; optional blueprintRender.createShape for full override (shapeId contract preserved for arrows).
- Export helpers: defaultMermaidNodeRenderSpec, resolveMermaidNodeRender, defaultCreateMermaidNodeFromBlueprint, MERMAID_MINDMAP_NODE_TYPE.
- Update README, api-report, and tests.

### Example page

Added an example page which show how custom shapes can be used to create "on the fly" CI pipeline workflows with mermaid diagram.

https://github.com/user-attachments/assets/861179c2-7757-4657-b39c-e071a14e8756

### API changes

- createMermaidDiagram options: add flowchart, state, sequence, mindmap with optional mapNodeToRenderSpec each.
- BlueprintRenderingOptions: add optional createShape (MermaidNodeCreateFunction) to override default node creation from the blueprint.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- createMermaidDiagram options: add flowchart, state, sequence, mindmap with optional mapNodeToRenderSpec each.
- BlueprintRenderingOptions: add optional createShape (MermaidNodeCreateFunction) to override default node creation from the blueprint.


angrycaptain19:
Could you add in an example of what this looks like to customize the shapes for a diagram?

apps/examples/src/examples/use-cases/custom-shape-mermaids/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Paste a linear Mermaid flowchart, run a simulated CI-style pipeline on the nodes, and retry failed steps from the canvas.
+
+---
+
+This example shows how **Mermaid** can define a **strictly sequential** `flowchart` (one path, no branches or joins). **Apply workflow** parses the source to get step order, renders the diagram with a custom shape util, and wires each node to a shared pipeline state atom. **Run pipeline** walks the steps with a short delay and a random failure chance (`FAIL_PROBABILITY` in the source). Failed steps show a **Retry** button on the shape; a successful retry continues with the remaining steps.
+
+**Linear only:** branching diagrams still render if Mermaid accepts them, but the built-in parser refuses to run the simulator until the text describes a single path. **Status colors** (pending, running, passed, failed) are **in memory only** for this demo—they are not stored in shape props or persisted in `.tldr` files.
+
+Helpers live in [`pipelineFromMermaid.ts`](./pipelineFromMermaid.ts).
```
angrycaptain19: I'd simplify this description a bit and move the fact that it's built on custom shapes earlier on / explain how the custom shapes work a bit more here.

angrycaptain19:
Love the example itself. Tiny nit - could we make the example diagram load left -> right so it's rendered below the text box? Right now it gets slightly cut off by the page nav controls 
Screenshot 2026-03-24 at 11 02 55

apps/examples/src/examples/use-cases/custom-shape-mermaids/CustomShapeMermaids.tsx
```diff
@@ -0,0 +1,500 @@
...
+			status === 'running'
+				? '2px solid #856404'
+				: status === 'failed'
+					? '2px solid #842029'
+					: '1px solid rgba(0,0,0,0.12)'
+
+		return (
+			<HTMLContainer
+				style={{
+					position: 'relative',
```
mimecuvalo: nit: can you pull most of these styles into a css file pls? (here and elsewhere)

apps/examples/src/examples/use-cases/custom-shape-mermaids/pipelineFromMermaid.ts
```diff
@@ -0,0 +1,123 @@
+/**
+ * Parse a strictly linear Mermaid flowchart into an ordered list of node ids.
```
mimecuvalo: mm, it seems this file is parsing Mermaid diagrams by just string comparisons? 
can't we use some of our new built-in Mermaid logic (or reference Mermaid.js directly)?

packages/mermaid/src/blueprint.ts
```diff
@@ -10,28 +10,70 @@ import type {
...
+	/** Semantic kind for this node in its diagram family. */
+	kind: string
+}) => MermaidBlueprintNodeRenderSpec | undefined
+
+/**
+ * Instructions for creating one blueprint node on the canvas.
+ * Set while building the blueprint; {@link renderBlueprint} consumes this without recomputing Mermaid → tldraw mapping.
+ * @public
+ */
+export type MermaidBlueprintNodeRenderSpec =
```
mimecuvalo, line 39: hah bit of a mouthful :P

packages/mermaid/src/defaultMermaidNodeRenderSpec.ts
```diff
@@ -0,0 +1,128 @@
...
+	switch (type) {
+		case 3:
+			return 'ellipse'
+		case 4:
+			return 'cloud'
+		case 6:
+			return 'hexagon'
+		case 5:
+			return 'star'
+		case 2:
```
mimecuvalo, line 81: nit: i dunno why but feels like we should go 0,1,2 :D

packages/mermaid/src/defaultMermaidNodeRenderSpec.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Uses the optional per-diagram mapper when it returns a value; otherwise {@link defaultMermaidNodeRenderSpec}.
+ * @public
+ */
+export function resolveMermaidNodeRender(
+	diagramKind: MermaidDiagramKind,
+	nodeId: string,
+	kind: string,
+	mapper?: MermaidNodeRenderMapper | undefined
+): MermaidBlueprintNodeRenderSpec {
+	return mapper?.({ nodeId, kind }) ?? defaultMermaidNodeRenderSpec(diagramKind, kind)
```
mimecuvalo: i want to check the intent here:
is the `??` supposed to kick in for both cases?
- when mapper is undefined
- when mapper is defined, called, but returns undefined

is this what we expect, or should it only be when mapper is undefined?

packages/mermaid/src/defaultMermaidNodeRenderSpec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Uses the optional per-diagram mapper when it returns a value; otherwise {@link defaultMermaidNodeRenderSpec}.
+ * @public
+ */
+export function resolveMermaidNodeRender(
+	diagramKind: MermaidDiagramKind,
+	nodeId: string,
+	kind: string,
+	mapper?: MermaidNodeRenderMapper | undefined
```
mimecuvalo, line 125: let's call this also `mapNode` - throws me off that this is called something slightly different

packages/mermaid/src/createMermaidDiagram.ts
```diff
@@ -37,11 +38,26 @@ const MERMAID_CONFIG = {
...
+	 */
+	mapNodeToRenderSpec?: MermaidNodeRenderMapper
+}
+
 /** @public */
 export interface MermaidDiagramOptions {
 	mermaidConfig?: Record<string, any>
 	blueprintRender?: BlueprintRenderingOptions
 	onUnsupportedDiagram?(svg: string): Promise<void>
+	flowchart?: MermaidDiagramKindOptions
```
mimecuvalo: i think instead of splitting these out into separate options we should just have one render similar to `defaultMermaidNodeRenderSpec` - then people can just a `switch` inside, much like we do

packages/mermaid/src/flowchartDiagram.ts
```diff
@@ -127,13 +107,16 @@ export function flowchartToBlueprint(
...
 		nodes.push({
-			id: subGraph.id,
+			id,
+			kind,
 			x: cluster.topLeft.x,
 			y: cluster.topLeft.y - FRAME_TOP_PAD,
 			w: cluster.width,
 			h: cluster.height + FRAME_TOP_PAD,
-			geo: 'rectangle',
+			render: resolveMermaidNodeRender('flowchart', id, kind, mapNode),
```
mimecuvalo: i'm thinking, high-level, that the blueprint generation shouldn't concern itself with `render`. 
i think what i'm noticing is that there's some prop drilling of these render functions but i think what should happen instead is that this resolution of rendering should happen when we call `renderBlueprint` in `createMermaidDiagram.ts` instead.

wdyt? let's discuss!

packages/mermaid/src/defaultMermaidNodeRenderSpec.ts
kaneel: if mapper is undefined then `undefined ?? defaultMapper` -> `defaultMapper` and if the return from the mapper is undefined, it would still call the default one as well

packages/mermaid/src/defaultMermaidNodeRenderSpec.ts
kaneel: no no this is how we count in France I swear :p

packages/mermaid/src/blueprint.ts
kaneel: Yeah this comment is not great, I'll do a pass on all of them!

apps/examples/src/examples/use-cases/custom-shape-mermaids/pipelineFromMermaid.ts
kaneel: totally different thing, it's a quick way to generate the "pipeline flow" but i think it should be generated from the generated shape imo

kaneel:
> nice, i like the direction. for my comments, in particular, let's discuss the last one regarding maybe not passing `render` into the blueprint phase

I've changed the API so we just have one point where we can update the process, one render function that is called at rendering time, if the rendering function returns undefined, it will default to all the previous mappings.

I've also updated the example with some choice 🦞, it builds the pipeline from the graph and not from the flowchart (coz I mean, whyyyyy) and it accepts a bit more complex branching flowchart, just to make the example even better

PR #8329, opened 2026-03-24
fix(arrows): prevent NaN propagation from zero-length labeled arrows

MitjaBezensek:
When arrow endpoints overlap exactly (same position), the arrow has zero length. If that arrow has a label, NaN values propagate through geometry → spatial index → culling, causing all shapes on the canvas to disappear. Easiest to reproduce with "always snap" on, but can happen in any scenario where endpoints coincide.

The root cause is `interpolateAlongEdge` computing `invLerp(0, 0, 0)` → `0/0` → NaN when two vertices are identical. This NaN flows into label positioning, arrow bounds, page bounds, and ultimately corrupts the R-tree spatial index, breaking viewport culling for every shape.

This PR fixes the issue at the source and adds defense-in-depth:
- Guard `interpolateAlongEdge` to return the vertex directly when segment length is 0
- Guard `uninterpolateAlongEdge` to return 0 when total length is 0
- Skip label geometry computation when the arrow is invalid (`isValid === false`), since the arrow component already returns null in that case
- Reject NaN bounds from the spatial index R-tree to prevent tree corruption

### Before

https://github.com/user-attachments/assets/fdac33de-2d06-44f6-9012-edd275d6387f

### After

https://github.com/user-attachments/assets/5d387565-d829-4cd5-93e6-063ce0c9435b

### Change type

- [x] `bugfix`

### Test plan

1. Enable "always snap" in user preferences
2. Create an arrow and add a label to it
3. Drag one endpoint to overlap the other (snap to same grid point)
4. Verify other shapes on canvas remain visible

### Release notes

- Fix all shapes disappearing when a labeled arrow has zero length (e.g. when both endpoints overlap)

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +7 / -4    |

### DIFF
--- a/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
+++ b/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
@@ -74,7 +74,7 @@ export class SpatialIndexManager {
 		// Collect all shape elements for bulk loading
 		for (const shape of shapes) {
 			const bounds = this.editor.getShapePageBounds(shape.id)
-			if (bounds && isFinite(bounds.minX) && isFinite(bounds.minY)) {
+			if (bounds && bounds.isValid()) {
 				elements.push({
 					minX: bounds.minX,
 					minY: bounds.minY,
@@ -101,7 +101,7 @@ export class SpatialIndexManager {
 			for (const shape of objectMapValues(changes.added) as TLShape[]) {
 				if (isShape(shape) && this.editor.getAncestorPageId(shape) === this.lastPageId) {
 					const bounds = this.editor.getShapePageBounds(shape.id)
-					if (bounds && isFinite(bounds.minX) && isFinite(bounds.minY)) {
+					if (bounds && bounds.isValid()) {
 						this.rbush.upsert(shape.id, bounds)
 					}
 					processedShapeIds.add(shape.id)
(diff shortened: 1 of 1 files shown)


packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
MitjaBezensek, line 1: A more generic check to prevent similar issues in the future.

packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
```diff
@@ -145,7 +145,7 @@ export class SpatialIndexManager {
 			const indexedBounds = this.rbush.getBounds(shapeId)
 
 			if (!this.areBoundsEqual(currentBounds, indexedBounds)) {
-				if (currentBounds) {
+				if (currentBounds && isFinite(currentBounds.minX) && isFinite(currentBounds.minY)) {
```
kaneel: areBoundsFinite maybe?

MitjaBezensek: Used `Box.isValid()` 👍

mimecuvalo:
cc @AniKrisn who has been in 0-length arrows land for a bit

MitjaBezensek:
> cc @AniKrisn who has been in 0-length arrows land for a bit

@AniKrisn could you take a look? Seems like a hotfix worthy fix as it does affect everyone (tldraw.com included).

AniKrisn:
@MitjaBezensek Ah I was working on #8130 which had a parallel case of a bad value propagating downstream and breaking things. That was solved by clamping the bad values [0, 1] to [0.001, 0.999]. This is a separate issue, but this guard makes sense to me

MitjaBezensek:
Will go ahead and merge this one. Might hotfix at a later time.

Issue #8328, opened 2026-03-24
"Move focus to canvas" button overlaps "Back to content" and flickers


MitjaBezensek:
## Description

The "Move focus to canvas" skip-to-content button (`SkipToMainContent` / `.tl-skip-to-main-content`) appears in the same spot as the "Back to content" helper button when both are visible. It also flickers as event handlers trigger.

Both buttons show in the top left, seemingly in the same spot.

## Steps to reproduce

1. Open the editor with shapes on the canvas
2. Pan away so all shapes are off-screen (triggers "Back to content" button)
3. Tab into the UI so the skip-to-content button gains focus
4. Both buttons overlap in the top-left area

## Expected behavior

The buttons should not overlap. The skip-to-content button should appear in a position that doesn't conflict with the helper buttons area.

## Actual behavior

"Move focus to canvas" renders on top of "Back to content". Additionally, the button flickers as event handlers fire.

## Video

https://github.com/user-attachments/assets/0e3ae422-fe41-4683-8e49-06e25d0c3262

MitjaBezensek:
## Research

### Relevant files

- **`packages/tldraw/src/lib/ui/components/A11y.tsx:18-55`** — `SkipToMainContent` component renders the "Move focus to canvas" button
- **`packages/tldraw/src/lib/ui/components/HelperButtons/BackToContent.tsx`** — `BackToContent` component renders the "Back to content" button
- **`packages/tldraw/src/lib/ui/TldrawUi.tsx:194`** — `SkipToMainContent` is rendered as the first child of `.tlui-layout`
- **`packages/tldraw/src/lib/ui/TldrawUi.tsx:209-211`** — `HelperButtons` (containing `BackToContent`) is rendered inside `.tlui-layout__top__left`
- **`packages/tldraw/src/lib/ui.css:1012-1022`** — CSS for `.tl-skip-to-main-content`
- **`packages/tldraw/src/lib/ui.css:389-401`** — CSS for `.tlui-helper-buttons`

### Root cause

The skip-to-content button uses `position: fixed; top: 48px; left: -9999px` and moves to `left: 8px` on `:focus`. The helper buttons area (`.tlui-helper-buttons`) sits inside `.tlui-layout__top__left` with `margin: var(--tl-space-2) var(--tl-space-3)`. Both end up at roughly the same top-left position.

[...]

PR #8334, opened 2026-03-24
fix(tldraw): skip to main content to keep its z-index on hover

kaneel:
Just a spike that shows what it takes to get rid of issues #8328 as this is basically a CSS UI caused by an UX/UI problem.

If the button is shown on focus, we should make sure that the following ui helper buttons are not showing.

### Before

Example of problem in issue: https://github.com/tldraw/tldraw/issues/8328

### After

https://github.com/user-attachments/assets/70bd5f89-63b3-4e7e-a5a9-8ce13d09ac62

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. go to canvas, draw some shapes 
2. move canvas until no shapes are visible
3. the "back to content" button should be visible 
4. press `tab` key
5. the "move focus to canvas" button is visible
6. hovering the "move focus to canvas" button does not display the "back to content" button

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed the flickering "back to content" button occuring when both the "back to content" + "move focus" are visible

### DIFF
--- a/packages/tldraw/src/lib/ui.css
+++ b/packages/tldraw/src/lib/ui.css
@@ -1021,6 +1021,10 @@
 	left: 8px;
 }
 
+/* 
+ * focusing skip-to-main-content hide top ui buttons to make sure we avoid
+ * z-index clashes on hover (see issue #8328)
+ */
 .tl-skip-to-main-content:focus + .tlui-layout__top .tlui-helper-buttons {
 	display: none;
 }


packages/tldraw/src/lib/ui.css
```diff
@@ -1015,12 +1015,17 @@
 	left: -9999px;
 	padding: 8px 16px;
 	z-index: var(--tl-layer-toasts);
+
```
mimecuvalo: nit: i think oxfmt will clean this up but you can remove the newline

packages/tldraw/src/lib/ui.css
```diff
@@ -1015,12 +1015,17 @@
...
 }
 
 .tl-skip-to-main-content:focus {
 	left: 8px;
 }
 
+.tl-skip-to-main-content:focus+.tlui-layout__top>.tlui-layout__top__left>.tlui-helper-buttons {
```
mimecuvalo: oxfmt might clean this up too but can you add spaces between ` + ` and ` > `?

mimecuvalo: nit: i think you can simplify to `.tl-skip-to-main-content:focus + .tlui-layout__top .tlui-helper-buttons`

kaneel:
> ach, i don't love it but i don't see another option. can you add a comment next to the new rule? otherwise we'll forget why this is here. pre-approving

Yeah I tried going around, it's not an easy one at all, moving the a11y button closer to the other buttons lose the nice "first button on tab" feature so it's back to those two commits and honestly, the second one makes a bit more sense than the first.

PR #8344, opened 2026-03-25
feat: use plaintext for mermaid parsing when clipboard has html+text

angrycaptain19:
Fixes an issue where mermaid code from VS Code, textedit, and live mermaid editor (among others) wasn't rendering as shapes because the stripped HTML removed the linebreaks so it was parsing as invalid mermaid.

This has been kept inside the mermaid handler to make sure it shouldn't impact any other copy/paste functionality, but still worth testing pasting of:
1) Pasting in formatted text (eg from google docs)
2) Copying non-mermaid code from VS Code
3) Copying in a webpage link and making sure that turns into an embed

I've tested all of those cases locally and they work fine

### DIFF
--- a/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
+++ b/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
@@ -74,7 +74,7 @@ export class SpatialIndexManager {
 		// Collect all shape elements for bulk loading
 		for (const shape of shapes) {
 			const bounds = this.editor.getShapePageBounds(shape.id)
-			if (bounds) {
+			if (bounds && bounds.isValid()) {
 				elements.push({
 					minX: bounds.minX,
 					minY: bounds.minY,
@@ -101,7 +101,7 @@ export class SpatialIndexManager {
 			for (const shape of objectMapValues(changes.added) as TLShape[]) {
 				if (isShape(shape) && this.editor.getAncestorPageId(shape) === this.lastPageId) {
 					const bounds = this.editor.getShapePageBounds(shape.id)
-					if (bounds) {
+					if (bounds && bounds.isValid()) {
 						this.rbush.upsert(shape.id, bounds)
 					}
 					processedShapeIds.add(shape.id)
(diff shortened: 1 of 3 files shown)


apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts
```diff
@@ -15,6 +15,8 @@ const FRONTMATTER_REGEX = /^-{3}\s*[\n\r]([\s\S]*?)[\n\r]-{3}\s*[\n\r]+/
 const DIAGRAM_KEYWORD_REGEX =
 	/^\s*(flowchart|graph|sequenceDiagram|classDiagram|stateDiagram|erDiagram|journey|gantt|pie|gitGraph|mindmap|timeline|sankey|xychart|block|quadrantChart|requirement|C4Context|C4Container|C4Component|C4Dynamic|C4Deployment|packet|kanban|architecture|treemap|radar|info)/
 
+const MARKDOWN_MERMAID_FENCE_REGEX = /^\s*```+\s*mermaid\s*\n([\s\S]*?)\n\s*```+\s*$/
```
kaneel, line 21: though I can read that perfectly, a quick comment would be a m a z i n g!

apps/dotcom/client/src/components/SneakyMermaidHandler/SneakyMermaidHandler.tsx
```diff
@@ -16,10 +16,20 @@ export function SneakyMermaidHandler() {
 
 	useEffect(() => {
 		editor.registerExternalContentHandler('text', async (content) => {
-			if (!simpleMermaidStringTest(content.text)) {
+			// When both text/html and text/plain are on the clipboard, content.text may
+			// be derived from stripHtml() which loses line breaks. Check the original
+			// text/plain source as well for mermaid detection.
+			const plainTextSource = content.sources?.find(
```
kaneel: I had a bit of a time trying to figure the comment, I went and tried myself – some consoling – and understood this as: 
```
when pasting html, the derived text is stripped and loses line breaks which make evaluating mermaid diagrams impossible. We look into the sources and see if there's some plain text alongside the html and if there are, we can use this to evaluate mermaid diagrams.
```

What do you think?

PR #8345, opened 2026-03-25
fix(paste): align pre-paste text measurement with final shape geometry

### LINKED ISSUE #8343: Text paste positioning differs from final shape geometry
When pasting text onto the canvas, the shape's initial position is calculated using a pre-paste measurement that differs from the final shape geometry. This causes the pasted shape to land slightly offset from where it should be centered.

https://github.com/user-attachments/assets/99bd71f0-1c3a-41db-a0e3-7c1d59f6699d

### Root cause

Text paste positioning is computed in two separate measurement passes:

1. **Pre-paste** (`defaultExternalContentHandlers.ts`): Measures raw HTML or plain text with `<br>` substitutions (`text.replace(/\n/g, '<br>')`) to determine width/height, then centers the shape at `p.x - w/2, p.y - h/2`.

2. **Shape geometry** (`TextShapeUtil.tsx`): Re-measures the text after the shape exists using `renderHtmlFromRichTextForMeasurement()`, which wraps content in `<div class="tl-rich-text">` and generates HTML through TipTap. It also adds `+1` to the width.

Since the two passes measure different HTML representations, they can produce different dimensions, causing the shape to be positioned based on one width but rendered with another.

### Steps to reproduce

1. Copy some text to the clipboard
2. Paste with `Cmd+V` — note the position
3. Paste the same text again — compare positions; they should be consistent but the centering is based on the pre-measurement, not the final rendered size

### Expected behavior

The pre-paste measurement should use the same HTML that the shape will ultimately render, so positioning matches the final geometry.

### Possible fix

Use `renderHtmlFromRichTextForMeasurement(editor, richTextToPaste)` in the pre-paste measurement instead of the ad-hoc `html ?? text.replace(/\n/g, '<br>')`. This aligns the two measurement passes.

AniKrisn:
This fixes pasted text positioning by measuring the same rich-text HTML that the final text shape renders. Previously, we positioned pasted text using a different pre-measurement format, which could make it land slightly off-center.

Closes #8343

### Change type

- [x] `bugfix`

### Test plan

1. Copy text from a rich text editor
2. Paste onto the canvas with Cmd+V
3. Copy the same text in plain text
4. Paste onto the canvas with Cmd+V
5. The text should be in the same place, and centred on the page

Note: unit tests won't catch text measurement bugs.

- [x] End to end tests

### Release notes

- Fixed slight positioning drift when pasting text onto the canvas.

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +3 / -5    |


steveruizok:
Looks good, asking the bot to push a test here. Review that it's checking what matters and then land this one

Issue #8288, opened 2026-03-19
Add overridable subcomponents to SharePanel


ds300:
The SharePanel currently only supports full replacement via `components={{ SharePanel: MyComponent }}`, unlike other panels (MenuPanel, NavigationPanel, Toolbar) that expose their subcomponents through `TLUiComponents` for granular override.

## Problem

To customize any part of the people/presence UI (e.g. the people menu items, avatar display, or user presence editor), you must replace the entire SharePanel. Other panels let you override individual pieces — for example, `DefaultMenuPanel` reads `MainMenu`, `QuickActions`, `ActionsMenu`, and `PageMenu` from `useTldrawUiComponents()`, and `DefaultNavigationPanel` reads `ZoomMenu` and `Minimap`.

## Proposed solution

Add SharePanel subcomponents to `TLUiComponents` so they can be individually overridden:

- `PeopleMenu` — the popover with the user list
- `PeopleMenuItem` — individual user row (follow/zoom actions)
- `UserPresenceEditor` — current user name/color editor

`DefaultSharePanel` and `PeopleMenu` would read these from `useTldrawUiComponents()` instead of importing them directly, following the same pattern as `DefaultMenuPanel`.

## Current workaround

Replace the entire SharePanel component, re-implementing the parts you don't want to change. There's an example of this in `SyncCustomPeopleMenuExample.tsx`.

ds300:
## Research

### Current state

**SharePanel files** (`packages/tldraw/src/lib/ui/components/SharePanel/`):
- `DefaultSharePanel.tsx` — wraps `<PeopleMenu />` in a div
- `PeopleMenu.tsx` — popover with user list, follows, and presence editor
- `PeopleMenuItem.tsx` — single user row with follow/zoom-to actions
- `PeopleMenuAvatar.tsx` — avatar circle in the trigger button
- `PeopleMenuMore.tsx` — "+N" overflow indicator
- `UserPresenceEditor.tsx` — name input + color picker for current user
- `UserPresenceColorPicker.tsx` — color swatch picker

**Component override system** (`packages/tldraw/src/lib/ui/context/components.tsx`):
- `TLUiComponents` interface defines all overrideable components (~line 69)
- `SharePanel` is already in `TLUiComponents` (top-level override works)
- Subcomponents are **not** in the interface — they're imported directly

### Pattern to follow

`DefaultMenuPanel.tsx` is the best model:

```tsx
[...]

PR #8346, opened 2026-03-25
feature(tldraw): SharePanel sub components in TldrawUiComponents

kaneel:
This PR to allow users to override SharePanel sub components as suggested in https://github.com/tldraw/tldraw/issues/8288

Also updated the doc in order to list all TLUIcomponents and the props types they rely on (it's more polite)!

### API changes
- Added `PeopleMenu`, `PeopleMenuItem`, `PeopleMenuFacePile` and `UserPresenceEditor` to the list of components users can override via the <Tldraw /> `components` property. This allows users to override these components and customise how tldraw displays the list of users currently editing a canvas.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Release notes

- SharePanel components can now be overridden via the Tldraw component API

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -918,7 +918,9 @@ export function useNativeClipboardEvents() {
 			// input instead; e.g. when pasting text into a text shape's content
 			if (editor.getEditingShapeId() !== null || areShortcutsDisabled(editor)) return
 
-			// Cmd+Shift+V / Ctrl+Shift+V = paste as plain text (no formatting)
+			// Cmd+Shift+V / Ctrl+Shift+V = paste as plain text (no formatting).
+			// If there's no plain text on the clipboard (e.g., a copied PNG), fall
+			// through to the normal paste handler so the file still gets pasted.
 			if (nativeShiftKey) {
 				const text = e.clipboardData?.getData('text/plain')
 				if (text?.trim()) {
@@ -927,10 +929,10 @@ export function useNativeClipboardEvents() {
 						: editor.getViewportPageBounds().center
 					editor.markHistoryStoppingPoint('paste')
 					defaultHandleExternalTextContent(editor, { text, point })
+					preventDefault(e)
+					trackEvent('paste', { source: 'kbd' })
+					return
 				}
-				preventDefault(e)
-				trackEvent('paste', { source: 'kbd' })
-				return
 			}
 
 			// Cmd+V: paste at center by default, or at cursor when the preference is on.


packages/tldraw/src/lib/ui/components/SharePanel/DefaultPeopleMenuAvatar.tsx
```diff
@@ -1,13 +1,18 @@
...
+
+/** @public @react */
+export function PeopleMenuAvatar({ userId }: TLUiPeopleMenuAvatarProps) {
 	const presence = usePresence(userId)
 
 	if (!presence) return null
 	return (
 		<div
 			className="tlui-people-menu__avatar"
-			key={userId}
```
kaneel, line 10: yeah this is a side quest

kaneel:
> Hm, I think this is providing too much granularity as far as components we can swap out. I think the sprit of the issue is that we would have some larger lego blocks that can be swapped in, but I think the smaller bits should just be exported (so that people can use them, or not, in their implementation).
> 
> Does that track with your intent @ds300 ?

Right so you mean being able to replace the PeopleMenu entirely and just providing the building blocks for doing so? That would not be a big change. Waiting for @ds300 answer on that.

packages/tldraw/src/lib/ui/components/SharePanel/PeopleMenu.tsx
```diff
@@ -59,7 +64,9 @@ export function PeopleMenu({ children }: PeopleMenuProps) {
 								{userName?.[0] ?? ''}
 							</div>
 						)}
-						{userIds.length > maxAvatars && <PeopleMenuMore count={userIds.length - maxAvatars} />}
+						{PeopleMenuAvatar && PeopleMenuMore && userIds.length > maxAvatars && (
```
ds300: feels weird to allow people to override just the 'more' component.

i think what probably makes more sense would be to inline the more component, or make it local in this file, and then extract this "tlui-people-menu__avatars" div into a `PeopleMenuFacePile` component that is overrideable. Then people can customise the 'more' behavior at a higher level

kaneel: So still export `PeopleMenuAvatar` and `PeopleMenuMore` so people can override the `PeopleMenuFacePile` while reusing `PeopleMenuAvatar` and/or `PeopleMenuMore`, right? :D

ds300: no don't bother exporting PeopleMenuMore or allowing people to override it. It's just a div with a number in it. I think could even inline it into the PeopleMenuFacePile component i menetioned

kaneel: Ha! yeah alright I see, I can improve that

steveruizok:
I'm going to tidy up some of the export patterns here to better match other components (naming `Default`). I'll preserve existing export names.

The pattern we follow is:

- use `Default` prefix for component / component parts that are overrideable
- use an exported default content so that a user can add components in front of or after the regular content using children

PR #7731, opened 2026-01-22
chore(sync): upgrade Zero to 0.25 and deploy staging to Fly.io

MitjaBezensek:
Upgrades Zero from 0.19 to 0.25 and sets up multi-node zero-cache deployment on Fly.io for staging. Production continues using the polyfill.

### Zero 0.25 upgrade

**API changes:**
- Query pattern: `tx.query.table.where(...).run()` → `tx.run(zql.table.where(...))`
- Imports now from `@rocicorp/zero` directly (removed internal `/out/...` imports)
- `ZERO_PUSH_URL` → `ZERO_MUTATE_URL` + `ZERO_QUERY_URL`
- `definePermissions` replaced with Synced Queries (`queries.ts`)

**Schema changes:**
- Added `schemaWithoutFairies` variant (fairy tables not in zero_data publication)

**Server changes:**
- New `/app/zero/mutate` endpoint using Zero's `PushProcessor`
- New `/app/zero/query` endpoint using Zero's `handleQueryRequest`
- New `/app/:userId/init` endpoint to ensure user exists before Zero queries
- `executeServerQuery()` replaces `ServerQuery` class with AST-to-SQL conversion

**Client changes:**
- Polyfill now implements `z.materialize(query)` and `z.preload(query)` to match Zero API
- Added AST helpers (`evaluateCondition`, `validateAST`) with tests

### Staging deployment (Fly.io multi-node)

- **Replication manager**: Single node handling PostgreSQL change stream, R2 backup
- **View syncers**: 2+ nodes serving client connections, session affinity via cookie
- Both services use Litestream for SQLite replica backup to R2

### Deployment changes

- `main` branch → `flyio-multinode` (staging with proper Zero)
- `production` branch → `sst` (production stays on polyfill)
- PR previews can use `preview-flyio-zero-deploy-please` label

### Other changes

- CSP allows `*.fly.dev` connections
- Deleted `ClientQuery.ts` and `ServerQuery.ts` (replaced by new query pattern)

### Change type

- [x] `improvement`

### Test plan

1. Verify staging (main branch) works with proper Zero on Fly.io
2. Verify production continues working with polyfill
3. Test `window.zero()` toggle in browser to switch modes locally

- [x] Unit tests (ast-helpers)
- [ ] End to end tests

### Release notes

- Internal infrastructure: upgraded Zero sync library to 0.25

---

> [!NOTE]
> **High Risk**
[...]

PR #8358, opened 2026-03-27
test(sync): add mutator permission tests

MitjaBezensek:
In order to verify no permission regressions after the Zero 0.25 migration (PR #7731), this PR adds comprehensive unit tests for all mutator permission logic in `packages/dotcom-shared`.

Relates to #7731

### What's tested

- **User mutations**: self-update, cross-user forbidden, immutable fields, flags mutability
- **File mutations**: group-member update, shared-user write forbidden, immutable fields (ownerId, owningGroupId, isDeleted), non-member forbidden
- **File creation**: group membership check, non-member forbidden
- **File state**: self-update, cross-user forbidden, server-side inaccessible file forbidden
- **onEnterFile**: server access check, no-access forbidden, no duplicate group_file
- **Group CRUD**: create group, owner update/delete, admin forbidden
- **Membership**: owner sets roles, admin forbidden, last owner demotion/leave forbidden, member leave
- **Cross-group file ops**: moveFileToGroup, removeFileFromGroup, addFileLinkToGroup with various access levels
- **Home group shortcut**: userId === groupId bypass for member/owner checks
- **Invite secret**: admin regenerate, non-admin forbidden
- **Immutable column bypass attempts**: ownerId, owningGroupId, isDeleted (including falsy values), firstVisitAt
- **Cross-user isolation**: unrelated users cannot access each other's files

### Scope note

TLDrawDurableObject's room access logic (the code that checks file ownership, group membership, and shared status when a user opens a file in the editor) was not changed by the Zero 0.25 migration — it uses Kysely queries directly, not the mutator layer. File room access should remain the same.

### Bug fix

Fixed `disallowImmutableMutations` truthiness bug: was using `!data[col]` which allowed falsy values like `false`, `0`, `null` to bypass the immutable guard. Now uses `!(col in data)` to check key presence instead.

### Change type

- [x] `improvement`

### Test plan

1. `cd packages/dotcom-shared && yarn test run`

- [x] Unit tests

### Code changes

| Section    | LOC change  |
| ---------- | ----------- |
| Source     | +1 / -1     |
| Tests      | +1100 / -0  |

### DIFF
--- a/packages/dotcom-shared/src/mutators.test.ts
+++ b/packages/dotcom-shared/src/mutators.test.ts
@@ -1,6 +1,5 @@
 import { IndexKey, uniqueId } from '@tldraw/utils'
 import { describe, expect, it } from 'vitest'
-import { MAX_NUMBER_OF_FILES } from './constants'
 import { createMutators, parseFlags, userHasFlag } from './mutators'
 import {
 	TlaFile,
@@ -291,10 +290,6 @@ function expectBadRequest(fn: () => Promise<any>) {
 	return expect(fn()).rejects.toThrow(ZErrorCode.bad_request)
 }
 
-function expectMaxFiles(fn: () => Promise<any>) {
-	return expect(fn()).rejects.toThrow(ZErrorCode.max_files_reached)
-}
-
 // ---- tests ----
 
 describe('parseFlags / userHasFlag', () => {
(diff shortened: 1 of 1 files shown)


packages/dotcom-shared/src/mutators.ts
```diff
@@ -62,7 +62,7 @@ function disallowImmutableMutations<
 	S extends TlaFilePartial | TlaFileStatePartial | TlaUserPartial,
 >(data: S, immutableColumns: Set<keyof S>) {
 	for (const immutableColumn of immutableColumns) {
-		assert(!data[immutableColumn], ZErrorCode.forbidden)
+		assert(!(immutableColumn in data), ZErrorCode.forbidden)
```
MitjaBezensek, line 65: This seems like a proper bug. Falsy values could bypass this check (like `isDeleted: false`).

packages/dotcom-shared/src/mutators.test.ts
```diff
@@ -0,0 +1,1276 @@
...
+			file_state: [],
+			group: [],
+			group_user: [],
+			group_file: [],
+		})
+		const m = createMutators(userId)
+		await expectForbidden(() => m.user.update(tx, { id: userId, email: 'evil@evil.com' }))
+	})
+
+	it('user can change own flags field', async () => {
```
MitjaBezensek, line 361: We should probably disallow this?

Issue #7943, opened 2026-02-15
Batch modification of width and height causes significant frame rate drop


random-yang:
It seems like this issue only appears in the recent versions.

https://github.com/user-attachments/assets/cb839da2-44c2-4d57-9f7d-195e436bb842

steveruizok:
This is due to the Rich text editor, which requires re-measurement of text on each resize frame. We've been slowly trying to improve performance here, but the regression is real.

steveruizok:
## Investigation notes

The performance drop during batch resize is caused by **per-frame DOM text measurement** that scales linearly with the number of selected shapes.

### What happens on each pointer move during resize

1. `Resizing.updateShapes()` ([Resizing.ts:366-383](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/tools/SelectTool/childStates/Resizing.ts#L366-L383)) loops through all selected shapes and calls `editor.resizeShape()` for each one.

2. For shapes containing text, `onResize()` calls `measureUnscaledLabelSize()` which goes through `TextManager.measureHtml()` — this sets `innerHTML` on a DOM element and calls `getBoundingClientRect()`, forcing a **browser layout recalculation**.

3. With N selected text-containing shapes, this means **N DOM layout recalculations per frame**.

### Shape-specific hotspots

- **GeoShapeUtil.onResize()** ([GeoShapeUtil.tsx:330-357](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx#L330-L357)): Directly calls `measureUnscaledLabelSize()` on every resize frame to constrain dimensions to text bounds. The code has a comment acknowledging that this deliberately **bypasses the WeakCache** because the temporary shape dimensions change every frame.

- **NoteShapeUtil** ([NoteShapeUtil.tsx:434-496](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx#L434-L496)): `getNoteLabelSize()` has a **loop of up to 50 iterations** measuring HTML at different font sizes. The WeakCache misses because shape props change each frame.

[...]

PR #8368, opened 2026-03-28
refactor(editor): replace DOM-based text measurement with pretext

steveruizok:
In order to eliminate synchronous layout reflows during text measurement, this PR replaces the TextManager's DOM-based measurement system (`getBoundingClientRect` + `Range.getClientRects()`) with [pretext](https://github.com/chenglou/pretext), a canvas-based text measurement library.

Relates to #7943

### What changed

- **`measureText()`** and **`measureHtml()`** now use pretext's `prepare()` + `layout()` for height and `walkLineRanges()` for width, instead of rendering into a hidden div and reading `getBoundingClientRect()`
- **`measureHtmlBatch()`** maps through `measureHtml()` — DOM batching is no longer needed since pretext avoids reflows entirely
- **`measureTextSpans()`** uses `prepareWithSegments()` + `layoutWithLines()` to reconstruct per-word spans with computed positions, replacing the `Range.getClientRects()` character-walking approach
- **`measureElementTextNodeSpans()`** is unchanged (DOM-based, used for already-rendered elements)
- Pretext v0.0.1 source (MIT) is vendored into `TextManager/pretext/` because the package exports raw `.ts` files incompatible with the monorepo's `moduleResolution: "node"` config
- Newlines are handled by splitting text into paragraphs and measuring each separately (pretext only supports `white-space: normal`)
- HTML content is stripped to plain text before measurement via a lightweight `stripHtmlToText()` helper

The public TextManager API is **fully preserved** — all method signatures and return types are unchanged.

### Change type

- [x] `improvement`

### Test plan

1. Create text, note, geo, and arrow shapes — verify labels size correctly
2. Resize shapes — verify text re-wraps at new widths
3. Test frame headings — verify they measure correctly
4. Copy as SVG — verify exported text positions match canvas
5. Paste text from clipboard — verify sizing and wrapping

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Replace DOM-based text measurement with canvas-based pretext library for faster text layout without layout reflows

### Code changes

| Section         | LOC change     |
| --------------- | -------------- |
| Core code       | +2787 / -116   |
| Tests           | +44 / -132     |

### DIFF
--- a/packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
+++ b/packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
@@ -5,7 +5,6 @@ import {
 	clearCache,
 	layout,
 	layoutWithLines,
-	prepare,
 	prepareWithSegments,
 	walkLineRanges,
 	type PreparedTextWithSegments,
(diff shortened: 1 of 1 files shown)


steveruizok:
This doesn't seem to make things much faster. I think a more efficient use would be deeper than just replacing text measurement.

Issue #5776, opened 2025-03-30
Custom font with default font example


bestony:
### What's the example?

Hi, I want to request a new Example with set a Custom font, and use Default Font.

Custom Font for some non-english language like Chinese, Japanese, Korean.

for example

default draw font family: `'tldraw_draw', sans-serif;`

custom font family: `'custom font', 'tldraw_draw', sans-serif;`

still use default font because I will use Chinese and English at the same time.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kostyafarber:
does this example help? https://tldraw.dev/examples/rich-text-font-extensions

mimecuvalo:
> does this example help? https://tldraw.dev/examples/rich-text-font-extensions

I think the author is asking about the shape's font, not so much the rich text stuff.

PR #8387, opened 2026-03-31
feat(editor): register fonts in themes alongside colors

mimecuvalo:
In order to complete the theming system started in #8216, this PR extends it to support custom font registration, loading, and deregistration — paralleling the existing color theme system. Fonts defined in themes are automatically registered at store creation time, loaded via the CSS Font Loading API, and appear in the style panel with support for custom icons and translation overrides.

Relates to #5776, #7587

### Change type

- [x] `feature`

### Test plan

1. Open the custom-theme example
2. Verify the style panel shows draw, sans, mono, pixel, and cursive fonts (serif should be removed)
3. Verify the pixel font icon shows an "A" in Silkscreen, cursive shows "A" in Comic Neue
4. Hover over the pixel/cursive font buttons and the pink color — tooltips should show "Pixel", "Cursive", and "Pink" (not raw keys)
5. Create shapes using pixel and cursive fonts — text renders correctly
6. Switch between light, dark, and brand themes — fonts persist
7. Adjust sliders — values only commit on mouse release (no mid-drag reloads)

- [x] Unit tests

### Release notes

- Add font registration to the theme system: custom fonts defined in themes are automatically available in the editor and style panel
- Add `TLThemeFont`, `TLThemeFonts`, `TLFullTheme` types for defining fonts in themes
- Add `registerFontsFromThemes()` to register/deregister font styles from theme definitions
- Add `getFontStyleItems()` for dynamic font options in the style panel
- Add `icon` field to `TLThemeFont` for custom style panel icons
- Add `TLThemeUiColorKeys` type for required UI color keys
- Move `TLFontFace`/`TLFontFaceSource` to `@tldraw/tlschema` for broader reuse
- Support deregistration of colors and fonts no longer present in themes
- Update custom-theme example with Silkscreen (local TTF) and Comic Neue (Google Fonts) custom fonts, color/font removal, translation overrides

### API changes

- Added `TLThemeFont` interface with `fontFamily`, optional `faces`, and optional `icon`
- Added `TLThemeFonts` interface (extensible via module augmentation)
- Added `TLFullTheme` type for themes with all defaults present
- Added `TLThemeUiColorKeys` type
- Added `DEFAULT_THEME_FONTS` constant
- Added `registerFontsFromThemes()` function
- Added `getFontStyleItems(theme)` function
- Changed `DEFAULT_LIGHT_THEME` / `DEFAULT_DARK_THEME` type from `TLTheme` to `TLFullTheme`
- Changed `TLDefaultFontStyle` from `T.TypeOf<typeof DefaultFontStyle>` to `keyof TLThemeFonts & string`
- Changed `TLTheme.fonts` to `Partial<TLThemeFonts>` (fonts are optional)
- Moved `TLFontFace` / `TLFontFaceSource` from `@tldraw/editor` to `@tldraw/tlschema` (re-exported for compat)

### Code changes

| Core code       | +473 / -112  |
| Tests           | +12 / -5     |
| Automated files | +85 / -41    |
| Documentation   | +263 / -20   |

### DIFF
--- a/packages/tlschema/api-report.api.md
+++ b/packages/tlschema/api-report.api.md
@@ -663,7 +663,7 @@ export type RecordPropsType<Config extends Record<string, T.Validatable<any>>> =
 export function registerColorsFromThemes(definitions: Record<string, TLTheme> | undefined): void;
 
 // @public
-export function registerFontsFromThemeDefinitions(definitions: Record<string, TLTheme> | undefined): void;
+export function registerFontsFromThemes(definitions: Record<string, TLTheme> | undefined): void;
 
 // @public
 export const richTextValidator: T.ObjectValidator<{


packages/tlschema/src/styles/TLTheme.ts
```diff
@@ -87,9 +155,18 @@ export interface TLTheme {
...
+	fonts: Partial<TLThemeFonts>
 }
 
+/**
+ * A fully-specified theme with all default colors and fonts present.
+ * This is the type of the built-in `DEFAULT_LIGHT_THEME` and `DEFAULT_DARK_THEME`.
+ *
+ * @public
+ */
+export type TLFullTheme = TLTheme & { colors: TLThemeColors; fonts: TLThemeFonts }
```
mimecuvalo: from Claude:
The split exists because we made palette colors and fonts optional on TLTheme to support removing defaults — but some code needs to know that all defaults are present.

Here's the core issue. TLTheme was changed so that custom themes can omit colors and fonts:

The Partial on fonts and the Partial<Omit<...>> on palette colors means that on a TLTheme, theme.colors.red is TLDefaultColor | undefined and theme.fonts.serif is TLThemeFont | undefined. This is what enables the custom-theme example to omit serif and light-violet etc.

But the built-in default themes (DEFAULT_LIGHT_THEME, DEFAULT_DARK_THEME) always have every color and every font. Without TLFullTheme, their type would be TLTheme and you'd get undefined back from theme.colors.red even though it's always there. That forces non-null assertions or unnecessary guards everywhere those defaults are used — in the mermaid color mapping, display values tests, shape utils that spread default color entries, etc.

TLFullTheme simply narrows the type back to "everything is present":
```
type TLFullTheme = TLTheme & { colors: TLThemeColors; fonts: TLThemeFonts }
So the two types serve different audiences:
```

- `TLTheme` — what users pass in (palette colors and fonts are optional, UI colors required)
- `TLFullTheme` — what the built-in defaults are typed as (everything guaranteed present)

That said, you could argue it's not strictly necessary — you could just type the defaults as TLTheme and add ! assertions or as casts at call sites. TLFullTheme is really just a convenience to avoid that noise. If you'd rather keep a single type and handle the optionality differently (e.g. a runtime "resolved theme" that fills in defaults), that would be an alternative design.

packages/tlschema/src/styles/TLColorStyle.ts
```diff
@@ -72,6 +75,12 @@ export function registerColorsFromThemes(themes: TLThemes | undefined): void {
 		DefaultLabelColorStyle.addValues(...colorNames)
 	}
 
+	const toRemove = DefaultColorStyle.values.filter((v) => !colorNames.has(v as TLDefaultColorStyle))
```
MitjaBezensek, line 80: Looks like the logic to remove colors is also new, not sure if we decided to go with this?
One thing we should be careful about if we decide to do this is that we have some hardcoded defaults like black for color and draw for font. If users removed those then creating shapes might fail (unless the users manually change styles for next shape I guess).

Feels like this added quite some complexity to types, not sure if worth it?

packages/tlschema/src/styles/TLTheme.ts
```diff
@@ -87,9 +155,18 @@ export interface TLTheme {
 	lineHeight: number
 	/** Base stroke width in pixels. Shape stroke widths are derived by multiplying this value. @defaultValue 2 */
 	strokeWidth: number
-	colors: TLThemeColors
+	colors: Pick<TLThemeColors, TLThemeUiColorKeys> & Partial<Omit<TLThemeColors, TLThemeUiColorKeys>>
+	fonts: Partial<TLThemeFonts>
```
MitjaBezensek, line 181: I guess we'd have to pass `{}` if we don't want to modify fonts? Would it be better to make this optional then? The same question if we want to modify fonts, but don't want to modify colors.

packages/tldraw/src/lib/shapes/shared/defaultFonts.tsx
```diff
@@ -145,3 +145,18 @@ export const DefaultFontFaces: TLDefaultFonts = {
...
 )
+
+/**
+ * Get the font faces for a given font style from the theme. For built-in fonts, returns
+ * undefined so callers can fall back to the rich-text-aware font scanning. For custom
+ * fonts (defined in the theme but not in DefaultFontFaces), returns the faces directly.
+ *
+ * @internal
+ */
+export function getThemeFontFaces(theme: TLTheme, font: string): TLFontFace[] | undefined {
```
MitjaBezensek, line 156: Maybe we can add a warning log in dev mode if we have a missing font? Think I added something similar for missing colors.

packages/tldraw/src/lib/styles.tsx
```diff
@@ -45,6 +45,54 @@ export function getColorStyleItems(theme: TLTheme): StyleValuesForUi<string> {
...
 }
 
+const defaultFontIcons: Record<string, string> = {
+	draw: 'font-draw',
+	sans: 'font-sans',
+	serif: 'font-serif',
+	mono: 'font-mono',
+}
+
+function isFontEntry(value: unknown): value is TLThemeFont {
```
MitjaBezensek: We have a duplicate of this in `TLFontStyle.ts`, we could export from here and resue in `styles.tsx`

apps/examples/src/examples/shapes/tools/speech-bubble/SpeechBubble/SpeechBubbleUtil.tsx
MitjaBezensek: Think we can get rid of this and call `getFontFamily(theme, shape.props.font)` and can drop the `FONT_FAMILIES` above?  Maybe we should make it public and export it?

packages/tlschema/src/styles/TLTheme.ts
```diff
@@ -74,9 +76,75 @@ export interface TLThemeColors {
...
+	/** CSS font-family declaration, e.g. `"'Inter', sans-serif"` */
+	fontFamily: string
+	/** Font face definitions for loading. Omit for system fonts that don't need loading. */
+	faces?: TLFontFace[]
+	/**
+	 * Icon for the style panel. Accepts a string icon ID (resolved via `assetUrls.icons`)
+	 * or a React element (`TLUiIconJsx`). Defaults to the built-in icon for known fonts,
+	 * or `'font-draw'` for custom fonts.
+	 */
+	icon?: unknown
```
MitjaBezensek, line 107: Another reason why separating editor and UI would make sense. We could define better types if these two were separated.

mimecuvalo: yeah i don't disagree - for a future PR we can explore

packages/tlschema/src/styles/TLColorStyle.ts
mimecuvalo: i definitely think we need this (this was from your PR originally? to be able to remove colors?)
otherwise people would be stuck with colors they didn't need, right?

MitjaBezensek: Yeah, unsure why @steveruizok didn't support this. Was just surprised to see this in the fonts pr (and also the complexity it brings).

packages/tlschema/src/styles/TLTheme.ts
mimecuvalo: this is a good idea actually, pushed up a change 👍

PR #8402, opened 2026-04-01
feat(examples): add audio shape example using a custom asset type

### LINKED ISSUE #3749: Add audio file import support

mimecuvalo:
Adds an audio shape example built on the extensible `AssetUtil` system from #8031, showing how to support audio files on the canvas without any SDK changes. Fixes #3749.

This PR previously added audio as a built-in shape and asset across `tlschema`, `editor`, and `tldraw`. That approach is now dropped (the branch merges `main` back in, removing all SDK changes) in favor of a self-contained example in `apps/examples`, since the custom asset system makes audio support possible in userland.

The example (`data/assets/audio-shape`) demonstrates:

- A custom `audio` asset type registered via `TLGlobalAssetPropsMap`, with an `AudioAssetUtil` that matches audio MIME types and extracts metadata (name, MIME type, duration) from dropped files in `getAssetFromFile`
- A custom `audio-player` shape whose util declares `handledAssetTypes` and `createShapeForAsset`, so the default external content handler creates it automatically when an audio file is dropped
- Resolving asset URLs through `TLAssetStore` via `editor.resolveAssetUrl`, with the asset read reactively so the shape updates when the upload completes
- A native audio player with pointer-event handling so playback controls work while the shape can still be dragged by its header

### Change type

- [x] `other`

### Test plan

1. Open the "Audio shape" example
2. Drop an audio file (mp3, ogg, wav, m4a) onto the canvas — an audio shape appears with the file name and duration
3. Play, seek, and adjust volume with the native controls; drag the shape by its header

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added an audio shape example showing how to support audio files with a custom asset type and shape.


lorenzolewis:
Big +1 from me on this one. I was also looking at how the video shape was implemented and I'm not a fan of a lot of the defaults muted for example). More can be seen here: https://github.com/tldraw/tldraw/blob/1d99b686825ae3b7886544ab6b2c4e988c5595bf/packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx#L212-L217

Ideally I'd love for the respective audio shape to be more flexible if it isn't already. :D

Issue #7586, opened 2026-01-05
Set up in-office Mac Mini for automated e2e tests, performance measurement, and QA


steveruizok:
## Summary

Set up a dedicated in-office Mac Mini to perform automated e2e tests, performance measurements, and QA tasks. This hardware would run comprehensive test suites at least once daily, acting as a fallback safety net that enables faster development workflows.

## Problem statement

Currently, developers face a tradeoff:
- Running the full e2e test suite before merging takes significant time (40-60 minute CI timeouts)
- Skipping comprehensive tests to move faster introduces risk of regressions

We need a system where code can ship to main quickly while comprehensive validation happens asynchronously, with any failures flagged promptly.

### Core functionality

1. **Scheduled comprehensive testing**
   - Run full e2e test suites (examples + dotcom) at least once daily
   - Execute tests that are too slow/comprehensive for every PR
   - Perform visual regression testing with baseline comparisons

2. **Performance measurement**
   - Track FPS, memory usage, and responsiveness metrics
   - Compare against baselines to detect performance regressions
   - Run performance-sensitive tests on consistent, dedicated hardware

3. **Computer use agent QA** (stretch goal)
   - Automated exploratory testing using an AI agent
   - Simulate real user workflows and catch edge cases
   - Generate bug reports for unexpected behavior

### Developer experience

- Developers can still run e2e tests locally for development and testing
- Fast path: quick CI checks enable rapid merging to main
- Safety net: Mac Mini catches issues the fast path might miss
- Clear alerting when the system detects failures or regressions

## Benefits

1. **Faster code → main**: Ship with confidence knowing comprehensive tests will catch issues
2. **More thorough automated testing**: Run longer, more comprehensive test suites that would be impractical in CI
3. **Reliable performance baselines**: Consistent hardware eliminates variability in performance measurements
4. **Expanded test coverage**: Time and resources for cross-browser testing, stress tests, etc.

## Current state

The existing e2e infrastructure uses:
- **Playwright** for browser automation
- **Two test suites**: `apps/examples/e2e/` and `apps/dotcom/client/e2e/`
- **CI runners**: 8-16 core Ubuntu machines with 40-60 minute timeouts
- **Visual snapshots**: Playwright snapshot comparison for visual regression
- **Commands**: `yarn e2e`, `yarn e2e-dotcom`

What's missing:
- Dedicated performance testing infrastructure
- FPS/memory baseline tracking
- Automated daily comprehensive test runs
[...]

PR #8421, opened 2026-04-02
feat(editor): add performance measurement hooks

MitjaBezensek:
In order to let SDK users observe editor performance in their apps (and for tldraw.com to use the same API for real-time monitoring), this PR adds a public `editor.performance` manager with EventEmitter-based subscriptions and zero overhead when no listeners are attached.

Relates to #7586

### Design

- **`PerformanceManager`** on `editor.performance` with `on(event, fn)` returning an unsubscribe function
- **Session-based aggregation**: frame times collected during interactions, percentile stats (avg/median/p95/p99) computed at session end
- **Flat, analytics-friendly events**: all frame time stats are top-level scalar fields (`avgFrameTime`, `medianFrameTime`, `p95FrameTime`, etc.) so events can be forwarded directly to services like Datadog or PostHog with `const { frameTimes, longAnimationFrames, ...metrics } = event`. The raw `frameTimes` and `longAnimationFrames` arrays are available for local analysis but should be excluded when sending to analytics.
- **Long Animation Frames (LoAF)**: automatically observes `long-animation-frame` entries (Chromium 123+) during interactions and camera ops. LoAF data is attached to `interaction-end` and `camera-end` events with script attribution for diagnosing jank sources. Observer is lazily started/stopped with listener lifecycle.
- **Lazy listener management**: editor event handlers only attached when first perf listener subscribes, detached when last unsubscribes
- **Zero overhead when unused**: `_notify*` methods early-return when `listenerCount === 0`
- **Opt-in interaction tracking**: tool states opt in via `static trackPerformance = true` instead of a hardcoded list, so custom tools can participate
- **Shared base type**: `TLPerfFrameTimeStats` extracted as common base for `TLInteractionEndPerfEvent` and `TLCameraEndPerfEvent`
- **Tree-shakeable `PerformanceApiAdapter`**: optional adapter piping events into `performance.mark()`/`performance.measure()` for DevTools, with cross-browser safe `mark()` fallback
- **Coexists with existing `PerformanceTracker`**: both fire from same `enter()`/`exit()` points, gated independently
- **Proper cleanup**: `PerformanceManager.dispose()` registered in editor disposables, clears all timers/listeners/observers

### Events

| Event | When |
|-------|------|
| `interaction-start/end` | Opted-in StateNode enter/exit (translating, resizing, drawing, etc.) |
| `camera-start/end` | Panning/pinch-zooming with debounced session end |
| `shapes-created/updated/deleted` | Shape operations with type breakdowns (lazily attached) |
| `frame` | Every animation frame (only when listener registered) |
| `undo/redo` | After undo/redo with stack depths |

`interaction-end` and `camera-end` events include LoAF data (`longAnimationFrames` field) when available, with per-script attribution for identifying jank sources. Both include `zoomLevel` and `timestamp`. Camera end events also include `visibleShapeCount` and `culledShapeCount`.

### Change type

- [x] `feature`

### Test plan

1. Open the "Performance hooks" example
2. Select and drag/resize shapes — perf panel in bottom-right updates with frame time stats
3. Pinch-zoom — verify camera-start/end fires
4. Delete shapes — verify shapes-deleted includes shape type breakdown
5. Verify no console output or overhead when no listeners registered

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add `editor.performance` API for subscribing to performance events (interaction frame times, camera ops, shape operations, undo/redo)
- Add Long Animation Frame (LoAF) data to interaction and camera end events for jank diagnosis (Chromium 123+)
- Add `PerformanceApiAdapter` for piping perf events into browser Performance API
- Add `static trackPerformance` on `StateNode` for custom tools to opt into interaction tracking

### API changes

- Added `Editor.performance: PerformanceManager`
- Added `PerformanceManager` class with `on()`, `once()`, `dispose()` methods
- Added `PerformanceApiAdapter` class with `dispose()` method
- Added `TLStateNodeConstructor.trackPerformance` and `StateNode.trackPerformance` static field
- Added types: `TLPerfFrameTimeStats`, `TLPerfEventMap`, `TLInteractionStartPerfEvent`, `TLInteractionEndPerfEvent`, `TLCameraStartPerfEvent`, `TLCameraEndPerfEvent`, `TLShapeOperationPerfEvent`, `TLFramePerfEvent`, `TLUndoRedoPerfEvent`, `TLPerfLongAnimationFrame`, `TLPerfLongAnimationFrameScript`
- Removed `TLInteractionEndPerfEvent.updateCount` (was always 0)
- Added `zoomLevel`, `timestamp` to `TLInteractionEndPerfEvent` and `TLCameraEndPerfEvent`
- Added `visibleShapeCount`, `culledShapeCount` to `TLCameraEndPerfEvent`
- Event names use kebab-case (`interaction-end`, `camera-start`, `shapes-created`) consistent with editor event naming

### Code changes

| Core code       | +913 / -18  |
| Tests           | +522 / -0   |
| Automated files | +131 / -0   |
| Documentation   | +188 / -0   |


packages/editor/src/lib/editor/managers/PerformanceManager/PerformanceApiAdapter.ts
```diff
@@ -0,0 +1,82 @@
...
+				})
+				try {
+					performance.measure(`tldraw:interaction:${event.name}`, startMark, endMark)
+				} catch {
+					// start mark may not exist if adapter attached mid-interaction
+				}
+			})
+		)
+
+		this.cleanups.push(
```
mimecuvalo, line 55: does the ApiAdapter need the shapes-* events?

apps/examples/src/examples/editor-api/performance-hooks/PerformanceHooksExample.tsx
```diff
@@ -0,0 +1,138 @@
...
+		// [3]
+		const adapter = new PerformanceApiAdapter(editor.performance)
+
+		return () => {
+			unsub()
+			adapter.dispose()
+		}
+	}, [editor])
+
+	return (
```
mimecuvalo, line 71: can we add camera events in this example too? that'd be handy

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
```diff
@@ -0,0 +1,196 @@
...
+	/** Total shapes on the current page. */
+	shapeCount: number
+	/** Viewport width in screen pixels. */
+	viewportWidth: number
+	/** Viewport height in screen pixels. */
+	viewportHeight: number
+	/** Number of shapes visible (not culled) in the viewport. */
+	visibleShapeCount: number
+	/** Number of shapes culled (off-screen) from rendering. */
+	culledShapeCount: number
```
mimecuvalo, line 90: should this have totalShapes as well (visible + culled)?

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
```diff
@@ -0,0 +1,196 @@
...
+ */
+export interface TLFramePerfEvent {
+	/** Time since the last frame in ms. */
+	elapsed: number
+	/** Total shapes on the current page. */
+	shapeCount: number
+	/** Number of shapes culled (off-screen) from rendering. */
+	culledShapeCount: number
+	/** Number of shapes visible (not culled) in the viewport. */
+	visibleShapeCount: number
```
mimecuvalo, line 128: ditto

packages/editor/src/lib/editor/managers/PerformanceManager/PerformanceApiAdapter.ts
```diff
@@ -0,0 +1,82 @@
+import type { PerformanceManager } from './PerformanceManager'
```
mimecuvalo, line 1: can PerformanceTracker use the new system? i was finding myself reaching for the Measure Performance debug-flag and then expecting to see the camera perf events in the console, but it was the older stuff. thoughts?

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
```diff
@@ -0,0 +1,196 @@
...
+	'shapes-updated': [TLShapeOperationPerfEvent]
+	/** Shapes were deleted. */
+	'shapes-deleted': [TLShapeOperationPerfEvent]
+	/** An animation frame was rendered. Only fires when listeners are registered. */
+	frame: [TLFramePerfEvent]
+	/** An undo operation was performed. */
+	undo: [TLUndoRedoPerfEvent]
+	/** A redo operation was performed. */
+	redo: [TLUndoRedoPerfEvent]
+}
```
mimecuvalo, line 196: i'm wondering if it'd be useful (or too granular) to have a culling event wired in from useShapeCulling/notVisibleShapes. sort of gives a pulse on some of our main DOM manipulation.
but maybe also for v2 to consider

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
```diff
@@ -0,0 +1,196 @@
...
+	/** Viewport height in screen pixels. */
+	viewportHeight: number
+	/** Number of shapes visible (not culled) in the viewport. */
+	visibleShapeCount: number
+	/** Number of shapes culled (off-screen) from rendering. */
+	culledShapeCount: number
+	/** Camera zoom level (`camera.z`) at session end. */
+	zoomLevel: number
+	/** `performance.now()` when the camera session ended. */
+	timestamp: number
```
mimecuvalo, line 98: it might be worth adding a `memory` property.
- the new fanciness is https://developer.mozilla.org/en-US/docs/Web/API/Performance/measureUserAgentSpecificMemory
- (the deprecated one is https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory )

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
```diff
@@ -0,0 +1,196 @@
...
+}
```
mimecuvalo, line 99: i'm wondering how it would be to test for, like, heavy shapes with high-res images/heavy embeds. is there any way to add that in as a property somehow? 🤔

mimecuvalo (approved):
LGTM! pre-approving, you can take any of my comments and just do followups as a v2 if you like or incorporate into this PR, up to you.

packages/editor/src/lib/editor/managers/PerformanceManager/PerformanceApiAdapter.ts
MitjaBezensek: These are just point in time events, not sure if it's worth marking them them for the profiler.

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
MitjaBezensek: `shapeCount` tracks that. Will reogranize a bit so that they are closer together.

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
MitjaBezensek: `shapeCount` tracks that.

packages/editor/src/lib/editor/managers/PerformanceManager/PerformanceApiAdapter.ts
MitjaBezensek: I'd leave it for now, can always address later on.

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
MitjaBezensek: I'm actually exploring this. 

It feels like it's not a sdk thing tbh. Like this measurement should live outside it. For example, the memory leak we had was when switching between files. SDK wouldn't catch those wide spanning operations. It's also just using regular browser api, nothing specific to our app.

I'd also be cautious with doing this for every event, so that we don't hurt the performance. 

We still might decide we want to add some helpers to the users, but at least right now I'd keep this separate.

packages/editor/src/lib/editor/managers/PerformanceManager/perf-types.ts
MitjaBezensek: We do have shape type distribution, so we can start with that. And if that turns out insufficient we can add more granular info.

PR #8439, opened 2026-04-03
fix(editor): fix memory leak from shared throttled updateHoveredShapeId

### LINKED ISSUE #8430: fix(editor): remove window.__tldraw__hardReset global — causes memory retention
Hi folks I have a patch prepared for this memory leak within __tldraw that is causing memory leaks.

Patch: https://github.com/tldraw/tldraw/compare/main...stewartmcgown:tldraw:main

`hardReset.ts` unconditionally assigns the `hardReset` function to `window.__tldraw__hardReset` at module load time:

This is causing TLDraw to leak memory after unmount in our production app.

Confirmed with a heap snapshot:
```
Window.__tldraw__hardReset
  -> hardReset closure
    -> module scope context (3,700+ entries)
      -> debounced closure (updateHoveredShapeId)
        -> getContainer closure context
          -> Editor (retained: 5.9 MB)
```

## Fix

Stop exposing this object to the window.

Perhaps your team may want to keep this reference stored in window for development, in which case we can add an if statement.

MitjaBezensek:
In order to fix an Editor memory leak after unmount (reported in #8430), this PR changes `updateHoveredShapeId` from a single shared throttled function to per-editor throttled instances stored in a `WeakMap`.

Previously, `updateHoveredShapeId` was a module-level `throttle()` call that captured the `Editor` via its closure. Because the throttled function (and its cancel handle) was shared across all editor instances, a disposed editor could be retained indefinitely by the pending throttle timer. The closure chain `throttled → _updateHoveredShapeId → editor` kept the entire `Editor` object alive.

Now each `Editor` gets its own throttled function stored in a `WeakMap<Editor, ...>`, so when the editor is garbage collected the throttled function goes with it. The cleanup function from `registerDefaultSideEffects` also cancels the throttle on teardown, and `_updateHoveredShapeId` bails early if the editor is disposed.

Might be worth hotfixing.

Closes #8430

### Change type

- [x] `bugfix`

### Test plan

1. Mount a tldraw editor, interact with it, then unmount
2. Take a heap snapshot — the Editor should not be retained
3. Previously the Editor was retained via `Window.__tldraw__hardReset → throttled closure → Editor` (5.9 MB per the issue report)

### Release notes

- Fix memory leak where the Editor was retained after unmount via a shared throttled `updateHoveredShapeId` closure.

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +49 / -12  |

### DIFF
--- a/packages/tldraw/src/lib/defaultSideEffects.ts
+++ b/packages/tldraw/src/lib/defaultSideEffects.ts
@@ -1,9 +1,12 @@
 import { Editor } from '@tldraw/editor'
-import { updateHoveredShapeId } from './tools/selection-logic/updateHoveredShapeId'
+import {
+	cancelUpdateHoveredShapeId,
+	updateHoveredShapeId,
+} from './tools/selection-logic/updateHoveredShapeId'
 
 /** @public */
 export function registerDefaultSideEffects(editor: Editor) {
-	return editor.sideEffects.register({
+	const unsub = editor.sideEffects.register({
 		instance: {
 			afterChange: (prev, next) => {
 				if (prev.cameraState !== next.cameraState && next.cameraState === 'idle') {
@@ -62,4 +65,8 @@ export function registerDefaultSideEffects(editor: Editor) {
 			},
 		},
 	})
+	return () => {
+		unsub()
+		cancelUpdateHoveredShapeId(editor)
+	}
 }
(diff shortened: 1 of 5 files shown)


MitjaBezensek:
Had claude look into other similar issues, found quite a few https://github.com/tldraw/tldraw/issues/8440

PR #8458, opened 2026-04-05
feat(dotcom): add UI color themes

steveruizok:
🅯 This PR is strongly influenced by @ds300's work to support themes in the Obsidian plugin. It only affects tldraw.com but is a good pressure test for some theming decisions we've made already and some that are still being made (see #8469).

---

In order to let tldraw.com users personalize the editor's visual appearance beyond light/dark mode, this PR adds a UI color theme system with 13 built-in themes. Closes #8455

Each theme provides both light and dark variants that map to existing CSS custom properties (`--tl-color-*` and `--tla-color-*`). The existing color scheme preference (Light / Dark / System) continues to control which variant is active — themes are orthogonal to the light/dark toggle.

https://github.com/user-attachments/assets/42c2ddc8-e3c4-4933-ae91-8da99729e5dd

**Available themes:** Default, Solarized, Dracula, Nord, Monokai, Gruvbox, Catppuccin, One Dark, Tokyo Night, Everforest, GitHub, Flexoki, Night Owl, High Contrast

**How it works:**
- Theme definitions are CSS variable override maps stored in `apps/dotcom/client/src/tla/themes/ui-themes.ts`
- When a theme is selected, `useUITheme` applies overrides via `element.style.setProperty()` on both the root `.tla-theme-container` and any inner `.tl-container` elements (using a MutationObserver to catch dynamically mounted containers)
- Switching back to "Default" removes all inline overrides so the base CSS classes take effect again
- The selected theme is persisted in `tldrawapp_session_3` localStorage alongside the existing `theme` preference
- Flash prevention (`theme-init.js`) reads a cached background color from localStorage to avoid white flash on page load with custom themes
- Canvas indicator colors (rendered via `<canvas>`, not CSS) use a reactive atom that re-resolves `--tl-color-selected` when the container's style attribute changes

### Change type

- [x] `feature`

### Test plan

1. Run `yarn dev-app` and open tldraw.com locally
2. Open user settings menu (three-dot icon in sidebar)
3. Verify "Color theme" submenu appears below "Theme" (light/dark/system)
4. Select each theme and verify UI colors change across sidebar, panels, and canvas
5. Toggle light/dark mode within each theme — verify the correct variant applies
6. Select shapes and verify the indicator color matches the theme's accent
7. Refresh the page — verify the theme persists with no flash of default colors
8. Switch back to "Default" — verify original colors restore cleanly

- [ ] Unit tests
- [ ] End to end tests

### Release notes

[...]

PR #8469, opened 2026-04-06
feat(editor): introduce canvas OverlayUtil system

steveruizok:
🅯 Hello, this PR is a continuation of both themes and indicators.

It introduces a new primitive, the "overlay" (`OverlayUtil`), which is a long-discussed idea to give arbitrary hit-testable-and-renderable stuff. Alex, in the mists of 2023-24, once proposed "accoutrements". This design provides a way to extend and replace "everything that is rendered on the html canvas", and is a significant change in API surface, though not in places that are commonly used. It fully retires our legacy indicators system. I've made no attempt to avoid breaking changes here.

Because many elements are were previous SVGs are now moved to canvas, many CSS variables have become unnecessary and have moved into our theme system. This is good for the that system. It will make projects like https://github.com/tldraw/tldraw/pull/8458 much easier and better, as fewer properties will need to be duplicated between css and the theme.

---

In order to make tldraw's canvas overlays (brushes, handles, scribbles, snap indicators, selection foreground, collaborator cursors, arrow hints, etc.) extensible, themeable, and uniformly hit-testable, this PR introduces a new `OverlayUtil` system on the editor and migrates all of the built-in overlays to use it.

Previously, each overlay was a one-off React component injected through `TLEditorComponents`, with bespoke rendering, hit-testing, and styling. The new `OverlayUtil` provides a single extensibility surface — analogous to `ShapeUtil` — with a managed lifecycle, hover hit-testing routed through `OverlayManager`, theme-driven colors via `TLTheme`, and an options/configure pattern. Overlays now render in front of indicators in a unified `CanvasOverlays` layer, and the legacy default component implementations and their CSS have been removed.

### Concepts

| Term            | Type         | Meaning                                                                                              |
| --------------- | ------------ | ---------------------------------------------------------------------------------------------------- |
| `OverlayUtil`   | class        | A class that defines a category of canvas overlays — when they're active, what they look like, where they hit-test, and what cursor they show. |
| `TLOverlay`     | interface    | A single overlay instance produced by an `OverlayUtil`, with an `id`, a `type`, and arbitrary `props`. |
| `OverlayManager`| class        | The manager on `Editor.overlays` that registers overlay utils, computes active overlays reactively, performs hover hit-testing, and drives rendering. |
| `defaultOverlayUtils` | array  | The set of built-in overlay utils shipped with `packages/tldraw`. Pass via the `overlayUtils` prop to extend or replace them. |

### Editor API

| Method                                | Description                                                                 |
| ------------------------------------- | --------------------------------------------------------------------------- |
| `Editor.overlays`                     | The `OverlayManager` instance for this editor.                              |
| `OverlayManager.getActiveOverlays()`  | Reactive list of all currently active overlay instances.                    |
| `OverlayManager.getHoveredOverlay()`  | The overlay currently under the pointer, if any.                            |
| `OverlayUtil.isActive()`              | Whether this util's overlays should currently exist.                        |
| `OverlayUtil.getOverlays()`           | Returns the current overlay instances for this util.                        |
| `OverlayUtil.getGeometry(overlay)`    | Hit-test geometry in page space; return `null` for non-interactive overlays. |
| `OverlayUtil.getCursor(overlay)`      | Cursor type to show when hovering this overlay.                             |
| `OverlayUtil.render(ctx, overlays)`   | Renders the overlays into a 2D canvas context already transformed to page space. |
| `OverlayUtil.configure(options)`      | Returns a subclass with options merged in.                                  |

### Component props

| Prop           | Type                              | Description                                                          |
| -------------- | --------------------------------- | -------------------------------------------------------------------- |
| `overlayUtils` | `TLAnyOverlayUtilConstructor[]`   | Overlay util classes to register on `<Tldraw />` / `<TldrawEditor />`. Defaults to `defaultOverlayUtils`. |

### Example

A minimal `OverlayUtil` that draws a circle at every selected shape's center:

```ts
import { OverlayUtil, TLOverlay, Vec } from 'tldraw'

interface MarkerOverlay extends TLOverlay<{ x: number; y: number }> {
	type: 'marker'
}

class MarkerOverlayUtil extends OverlayUtil<MarkerOverlay> {
	static override type = 'marker'
	options = { color: 'tomato', radius: 6 }

	isActive() {
		return this.editor.getSelectedShapeIds().length > 0
	}

	getOverlays(): MarkerOverlay[] {
		return this.editor.getSelectedShapes().map((s) => {
			const c = this.editor.getShapePageBounds(s)!.center
			return { id: `marker:${s.id}`, type: 'marker', props: { x: c.x, y: c.y } }
		})
	}

	render(ctx: CanvasRenderingContext2D, overlays: MarkerOverlay[]) {
		ctx.fillStyle = this.options.color
		for (const o of overlays) {
			ctx.beginPath()
			ctx.arc(o.props.x, o.props.y, this.options.radius, 0, Math.PI * 2)
			ctx.fill()
		}
	}
}
```

A consumer registers it alongside the defaults, optionally configured:

```tsx
import { Tldraw, defaultOverlayUtils } from 'tldraw'

const BigMarker = MarkerOverlayUtil.configure({ color: 'dodgerblue', radius: 12 })

	return (
		<Tldraw
			overlayUtils={[...defaultOverlayUtils, BigMarker]}
		/>
	)
}
```

### FAQ

**How do I customize an existing overlay (e.g. the brush color)?**

Use `OverlayUtil.configure(...)` to produce a subclass with new options, then pass it via `overlayUtils`:

```ts
const MyBrush = BrushOverlayUtil.configure({ fill: 'rgba(0,0,255,0.1)' })
```

**How do I remove a built-in overlay?**

Pass `overlayUtils` without it:

```tsx
<Tldraw overlayUtils={defaultOverlayUtils.filter((u) => u.type !== 'snap_indicator')} />
```

**How do I hit-test against my overlay so it shows a custom cursor?**

Implement `getGeometry()` and `getCursor()`. Returning `null` from `getGeometry()` makes the overlay non-interactive (correct for things like scribbles or snap lines).

**Where do overlay colors come from?**

From `TLTheme`. Read them in `render()` via `this.editor.getColorTheme()` so overlays follow light/dark mode automatically.

**Can I render React inside an overlay?**

No — overlays render into a 2D canvas context. For React-based UI, use a regular `TLEditorComponents` slot or an HTML layer above the canvas.

### New examples

- `editor-api/lasso-select-tool` — now demonstrates implementing a lasso as a custom `OverlayUtil` (`LassoOverlayUtil.ts`) registered alongside the defaults.

### Change type

- [x] `api`

### Test plan

1. Open the examples app and verify brushes, zoom brush, snap indicators, selection handles, crop handles, scribbles, and arrow target hints all render and behave as before.
2. Open a multiplayer room and confirm collaborator cursors, brushes, scribbles, and hints appear correctly.
3. Switch between light and dark themes and confirm overlay colors update.
4. Hover near selection handles and crop handles and confirm cursors update correctly under rotation.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added a new `OverlayUtil` system for canvas overlays. Built-in brushes, handles, scribbles, snap indicators, selection foreground, collaborator cursors, and arrow hints are now implemented as overlay utils and can be customized or extended via `overlayUtils`.
- Removed the legacy default overlay components (`Brush`, `ZoomBrush`, `Scribble`, `SnapIndicator`, `Handle`, `Handles`, `SelectionForeground`, `CollaboratorHint`, `ShapeIndicators`, etc.) from `TLEditorComponents`. Use a custom `OverlayUtil` instead.

### API changes

- Added `OverlayUtil`, `TLOverlay`, `OverlayManager`, and `Editor.overlays` for registering and managing canvas overlays.
- Added `defaultOverlayUtils` and an `overlayUtils` prop on `<Tldraw />` / `<TldrawEditor />`.
- Added overlay-related theme colors to `TLTheme`.
- Breaking! Removed `Brush`, `ZoomBrush`, `Scribble`, `SnapIndicator`, `Handle`, `Handles`, `SelectionForeground`, `SelectionBackground`, `CollaboratorHint`, `ShapeIndicator`, `ShapeIndicators`, and `ShapeIndicatorErrorFallback` from `TLEditorComponents`. Migrate custom implementations to `OverlayUtil`.
- Breaking! Removed the corresponding `Default*` component exports and `LiveCollaborators`.

### Code changes

| Core code       | +2983 / -2852 |
| Tests           | +1943 / -504  |
| Automated files | +464 / -261   |
| Documentation   | +398 / -476   |
| Apps            | +2116 / -103  |
| Templates       | +31 / -15     |
| Config/tooling  | +1 / -0       |


steveruizok:
### Proposed options for built-in overlays

These are *proposed* — the built-in utils don't actually expose them yet (their `render()` methods still read from theme/constants directly). Tracking here so the tables don't bloat the PR description.

**`BrushOverlayUtil`** — selection brush rectangle

| Option       | Default                  | Description                              |
| ------------ | ------------------------ | ---------------------------------------- |
| `fill`       | `theme.brushFill`        | Fill color of the brush rectangle.       |
| `stroke`     | `theme.brushStroke`      | Stroke color of the brush rectangle.     |
| `lineWidth`  | `1`                      | Stroke width in screen pixels.           |

**`ZoomBrushOverlayUtil`** — zoom-tool brush rectangle

| Option       | Default                  | Description                              |
| ------------ | ------------------------ | ---------------------------------------- |
| `fill`       | `theme.brushFill`        | Fill color of the zoom brush.            |
| `stroke`     | `theme.brushStroke`      | Stroke color of the zoom brush.          |
| `lineWidth`  | `1`                      | Stroke width in screen pixels.           |

**`ScribbleOverlayUtil`** — local scribble / laser pointer trail

| Option       | Default | Description                                      |
| ------------ | ------- | ------------------------------------------------ |
| `streamline` | `0.32`  | Path smoothing factor applied to scribble points. |
| `cacheSize`  | `500`   | Max number of cached scribble paths.             |

**`CollaboratorScribbleOverlayUtil`** — remote scribbles / laser pointers

| Option           | Default | Description                                              |
| ---------------- | ------- | -------------------------------------------------------- |
| `laserOpacity`   | `0.5`   | Opacity used for laser-style remote scribbles.           |
| `scribbleOpacity`| `0.1`   | Opacity used for non-laser remote scribbles.             |
| `streamline`     | `0.32`  | Path smoothing factor.                                   |
| `cacheSize`      | `500`   | Max number of cached scribble paths.                     |

**`SnapIndicatorOverlayUtil`** — snap lines, cross marks, and gap ticks

| Option        | Default        | Description                                       |
| ------------- | -------------- | ------------------------------------------------- |
| `stroke`      | `theme.snap`   | Color used for all snap indicators.               |
| `lineWidth`   | `1`            | Stroke width of snap lines in screen pixels.      |
| `pointLength` | `2.5`          | Half-length of the cross marker arms.             |
| `gapLength`   | `3.5`          | Half-length of the gap tick marks.                |

**`ShapeHandleOverlayUtil`** — resize / rotate / vertex handles

| Option           | Default                          | Description                                       |
| ---------------- | -------------------------------- | ------------------------------------------------- |
| `fill`           | `theme.selectedContrast`         | Handle fill color.                                |
| `stroke`         | `theme.selectionStroke`          | Handle stroke color.                              |
| `lineWidth`      | `1.5`                            | Handle stroke width in screen pixels.             |
| `radius`         | `editor.options.handleRadius`    | Default handle radius (fine pointer).             |
| `coarseRadius`   | `editor.options.coarseHandleRadius` | Handle radius for coarse (touch) pointers.     |

**`SelectionForegroundOverlayUtil`** — selection outline + crop handles

| Option                  | Default                  | Description                                              |
| ----------------------- | ------------------------ | -------------------------------------------------------- |
| `stroke`                | `theme.selectionStroke`  | Selection outline color.                                 |
| `lineWidth`             | `1.5`                    | Selection outline stroke width.                          |
| `mobileHandleMultiplier`| `1.75`                   | Touch-target multiplier applied to handles on coarse input. |
| `mobileRotateBgRadius`  | `20`                     | Background radius behind the mobile rotate handle.       |
| `desktopRotateBgRadius` | `14`                     | Background radius behind the desktop rotate handle.      |

**`ArrowHintOverlayUtil`** — arrow target / snap hints

| Option            | Default                   | Description                                       |
| ----------------- | ------------------------- | ------------------------------------------------- |
| `stroke`          | `theme.selectionStroke`   | Hint stroke color.                                |
| `fill`            | `theme.selectionFill`     | Hint fill color.                                  |
| `contrast`        | `theme.selectedContrast`  | Inner contrast color used by handle dots.         |
| `lineWidth`       | `1.5`                     | Stroke width in screen pixels.                    |
| `edgeRadius`      | `8`                       | Snap circle radius for edge targets.              |
| `edgePointRadius` | `12`                      | Snap circle radius for edge-point targets.        |
| `handleRadius`    | `4`                       | Radius of edge handle dots.                       |

**`CollaboratorBrushOverlayUtil`** — remote collaborator brushes

| Option         | Default | Description                                                    |
| -------------- | ------- | -------------------------------------------------------------- |
| `fillOpacity`  | `0.75`  | Alpha applied to the brush fill (over the collaborator color). |
| `strokeOpacity`| `0.1`   | Alpha applied to the brush stroke.                             |
| `lineWidth`    | `1`     | Stroke width in screen pixels.                                 |

**`CollaboratorCursorOverlayUtil`** — remote collaborator cursors, name tags, chat bubbles

| Option         | Default          | Description                                        |
| -------------- | ---------------- | -------------------------------------------------- |
| `fontSize`     | `12`             | Font size of the name tag and chat bubble (px).    |
| `nameMaxWidth` | `120`            | Max width of the name tag before truncation.       |
| `chatMaxWidth` | `200`            | Max width of the chat message bubble.              |
| `cornerRadius` | `4`              | Corner radius of the background pill.              |
| `padding`      | `{ x: 6, y: 3 }` | Inner padding of the pill background.              |
| `shadowOpacity`| `0.2`            | Opacity of the cursor's drop shadow.               |
| `outlineWidth` | `3`              | Width of the cursor outline.                       |
| `textColor`    | `'#ffffff'`      | Foreground text color over the collaborator color. |

**`CollaboratorHintOverlayUtil`** — off-screen collaborator pointer hints

| Option            | Default | Description                                            |
| ----------------- | ------- | ------------------------------------------------------ |
| `lineWidth`       | `3`     | Width of the hint outline stroke.                      |
| `viewportPadding` | `5`     | Distance from the viewport edge in screen pixels.      |

ds300:
Overall: good shit ✅

Minor API design questions:

1. **Hit-test vs render ordering**
   `getOverlayAtPoint` returns the first match by registration order, but visually later overlays paint on top. These can diverge: a user might click what looks like the topmost element but hit something underneath. Should we align both to "topmost wins", or introduce explicit priority?

2. **Custom pointer behavior extensibility**
   Hover/cursor are pluggable, but pointer-down routing in `select.idle` is hardcoded to built-in overlay types (`resize_handle`, `rotate_handle`, `shape_handle`). What's the intended path for custom interactive overlays that need their own drag/click behavior — custom tool states, or a future overlay event API?

3. **React-rendered overlays**
   Should overlay utils support an optional React render path alongside canvas? "It's just HTML" is a core tldraw selling point, and some overlays (tooltips, rich labels, interactive controls) are more natural as DOM elements. One approach: let overlay utils optionally return a React component; contiguous canvas-only utils could be batched onto a single `<canvas>` element (in practice almost all would be). This keeps the default path fast while opening up the full DOM when needed. Maybe html-in-canvas will allow to squash the whole thing at some point, not sure how interactions would be affected there.

4. **Minimap rendering**
   Should `OverlayUtil` have an optional `renderMinimap()` method? Some overlays (brush bounds, collaborator hints) are hard-coded in the minimap rendering code and we could easily move their rendering logic into overlay utils to make the minimap extensible. Could be a follow up obviously, just posing the question.

MitjaBezensek:
Agree with @ds300, looking good!

High level:
1. Similar to what @ds300 has in his first point. We should fix the ordering / render / hit test priorities. For example, replacing an interactive overlay changes hit test priority: `mergeArraysAndReplaceDefaults` moves replacements to the end, so the original order is lost. There are a few gotchas there, like the mentioned inversion between hit order and render order.
It feels like we might want to offer an more explicit way for specifying the order. Array based order also makes it weird to "insert an overlay between two existing overlays`, array slicing is not that nice.
2. We lose on accessibility. Unsure if we can do something to support this better inside `canvas`.
3. We push away from react components to using `canvas`. It dilutes our messaging of being "react rendered canvas" a bit. Probably still worth it for the performance benefits though. Also AI makes all the canvas works quite easy to do, so probably not a major issue. Might be worth exposing some basic canvas primitives though? Draw a box, draw a circle,...?

Low level:
1. Duplicate work across `getOverlays/getGeometry/render`, should be easy to improve.
2. `getGeometry` allocates on every pointer move.
3. `configure()` doesn't actually do anything, I guess this part is just stubbed out, and we'd define all the options for the different overlays before shipping?
4. Think `mobile_rotate` is broken.`getGeometry` is missing `isShapeTooCloseToContextualToolbar` check.

mimecuvalo:
> We lose on accessibility. Unsure if we can do something to support this better inside canvas.

I was going to chime in to say the same thing, but actually, while the indicators do have aria attributes on them, they're not really accessible / reachable by traditional means. We still rely solely on keyboard shortcuts to rotate/translate/manipulate the shapes. So actually, I think we're counter-intuitively fine on this front.

MitjaBezensek:
> I was going to chime in to say the same thing, but actually, while the indicators do have aria attributes on them, they're not really accessible / reachable by traditional means. We still rely solely on keyboard shortcuts to rotate/translate/manipulate the shapes. So actually, I think we're counter-intuitively fine on this front.

👍

It would make supporting that much harder though, we'd have to implement our own in canvas a11y logic.

steveruizok:
This is good feedback guys, thanks.

Accessibility sounds acceptable, nothing seriously lost or gained.

The story about react feels like it will survive. After indicators, it's already a hybrid canvas, which many tools (especially in design) seem to be converging on. Paper, pencil, wonder most recently. I think it's legible as a product. The "on the canvas" and "in front of the canvas" are still good for interactive elements (we have plenty of examples) in case folks want to do things the old way too. To my knowledge no one has ever made interactive overlays like this using our system... because these things haven't really been an API before.

I'll look into the z-index work next. We've used css z-index as the way before, so all handles are technically on the same z-index, which is why we hide them when they get dense to prevent collisions / mess.

packages/tldraw/src/lib/overlays/SnapIndicatorOverlayUtil.ts
```diff
@@ -0,0 +1,182 @@
...
+				horizontal ? gap.endEdge[1].y : gap.endEdge[1].x
+			)
+			if (nextEdgeIntersection) {
+				edgeIntersection = nextEdgeIntersection
+			} else {
+				continue
+			}
+		}
+
+		if (edgeIntersection === null) return
```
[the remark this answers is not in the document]
steveruizok: Good catch — fixed in 46263426ad. Narrowed `edgeIntersection` to `number[]`, dropped the `!` assertions, and removed the unreachable `=== null` guard.

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -95,6 +116,63 @@ export class Idle extends StateNode {
...
+						}
+						case 'resize_handle': {
+							this.onPointerDown({
+								...info,
+								target: 'selection',
+								handle: overlay.props.handle as any,
+							})
+							break
+						}
+						default: {
```
steveruizok, line 173: falls into the default overlay branch here, so the event gets re-dispatched as  without a . In the selection switch below that no longer enters , which means coarse-pointer rotate taps behave like a normal selection press instead of starting rotation.

packages/tldraw/src/lib/tools/SelectTool/childStates/Crop/children/Idle.ts
```diff
@@ -41,6 +48,24 @@ export class Idle extends StateNode {
 
 		switch (info.target) {
 			case 'canvas': {
+				// Check overlays first — if we hit a crop/resize handle, re-dispatch
+				const currentPagePoint = this.editor.inputs.getCurrentPagePoint()
+				const hitOverlay = this.editor.overlays.getOverlayAtPoint(
+					currentPagePoint,
+					this.editor.options.hitTestMargin / this.editor.getZoomLevel()
+				)
+				if (hitOverlay) {
```
steveruizok, line 57: Crop-mode overlay hit-testing only remaps  and . The visible  overlay is skipped here, so tapping the crop rotate control on coarse pointers falls through to normal canvas / shape handling instead of rotating.

packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts
```diff
@@ -0,0 +1,262 @@
...
+ */
+export class CollaboratorCursorOverlayUtil extends OverlayUtil<TLCollaboratorCursorOverlay> {
+	static override type = 'collaborator_cursor'
+	override options = { zIndex: 1100, fontSize: 12, nameMaxWidth: 120, chatMaxWidth: 200 }
+
+	override isActive(): boolean {
+		return this.editor.getCollaboratorsOnCurrentPage().length > 0
+	}
+
+	override getOverlays(): TLCollaboratorCursorOverlay[] {
```
steveruizok, line 59: This visibility logic now depends on  inside , but there is no reactive timer source driving recomputation. As a result, collaborators will not transition from active to idle / inactive as time passes unless some unrelated editor signal changes.

packages/tldraw/src/index.ts
```diff
@@ -59,14 +59,41 @@ export { BookmarkAssetUtil } from './lib/assets/BookmarkAssetUtil'
 export { ImageAssetUtil } from './lib/assets/ImageAssetUtil'
 export { VideoAssetUtil } from './lib/assets/VideoAssetUtil'
 export { ArrowBindingUtil } from './lib/bindings/arrow/ArrowBindingUtil'
-export { TldrawCropHandles, type TldrawCropHandlesProps } from './lib/canvas/TldrawCropHandles'
-export { TldrawHandles } from './lib/canvas/TldrawHandles'
-export { TldrawArrowHints, TldrawOverlays } from './lib/canvas/TldrawOverlays'
-export { TldrawScribble } from './lib/canvas/TldrawScribble'
-export { TldrawSelectionForeground } from './lib/canvas/TldrawSelectionForeground'
-export { TldrawShapeIndicators } from './lib/canvas/TldrawShapeIndicators'
+export { ArrowHintOverlayUtil, type TLArrowHintOverlay } from './lib/overlays/ArrowHintOverlayUtil'
```
steveruizok, line 62: This export-surface change removes previously public  canvas components like , , , , and  with no compatibility shim. That is a source-breaking API change for downstream consumers of the published package.

packages/tlschema/src/styles/TLTheme.ts
```diff
@@ -61,6 +61,20 @@ export interface TLThemeDefaultColors {
 	solid: string
 	cursor: string
 	noteBorder: string
+	/** Snap indicator line color */
+	snap: string
```
steveruizok, line 65: These new fields are now required on , which means any downstream custom theme object typed against the previous public API will stop compiling until it supplies all of them. If that break is intentional it probably needs an explicit migration / release note; otherwise these should stay backwards-compatible.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
```diff
@@ -0,0 +1,613 @@
...
+		const size = 8 / zoom
+
+		const isTinyX = width < size * 2
+		const isTinyY = height < size * 2
+
+		const showSelectionBounds =
+			(onlyShape ? !editor.getShapeUtil(onlyShape).hideSelectionBoundsFg(onlyShape) : true) &&
+			!isChangingStyle
+
+		const shouldDisplayBox =
```
steveruizok: The old implementation kept the selection outline visible while resizing text ( + text shape). That special case is gone here, so text-resize loses its box mid-drag even though the previous behavior intentionally preserved it.

packages/tldraw/src/lib/overlays/CollaboratorHintOverlayUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Overlay util for off-screen collaborator cursor hints.
+ * Shows a small directional arrow at the viewport edge pointing toward the collaborator.
+ *
+ * @public
+ */
+export class CollaboratorHintOverlayUtil extends OverlayUtil<TLCollaboratorHintOverlay> {
+	static override type = 'collaborator_hint'
+	override options = { zIndex: 900, lineWidth: 3, viewportPadding: 5 }
+
+	override isActive(): boolean {
```
steveruizok, line 35: This new hint path no longer applies the old collaborator-visibility gating before deciding whether to render off-screen arrows. Inactive collaborators, and idle collaborators who are following you, can still show edge hints when they should stay hidden.

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -42,7 +43,10 @@ export class Idle extends StateNode {
 
 	override onEnter() {
 		this.parent.setCurrentToolIdMask(undefined)
-		updateHoveredShapeId(this.editor)
+		// Check overlays first, then shapes
+		if (!updateHoveredOverlayId(this.editor)) {
```
steveruizok, line 47: can set an overlay-specific cursor here, but  still unconditionally resets the cursor to default a few lines later. If selection state changes under a stationary pointer, the correct hover cursor is lost until the next pointer move.

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
steveruizok, line 173: `mobile_rotate` falls into the default overlay branch here, so the event gets re-dispatched as `target: selection` without a `handle`. In the selection switch below that no longer enters `pointing_rotate_handle`, which means coarse-pointer rotate taps behave like a normal selection press instead of starting rotation.

packages/tldraw/src/lib/tools/SelectTool/childStates/Crop/children/Idle.ts
steveruizok, line 57: Crop-mode overlay hit-testing only remaps `resize_handle` and `rotate_handle`. The visible `mobile_rotate` overlay is skipped here, so tapping the crop rotate control on coarse pointers falls through to normal canvas / shape handling instead of rotating.

packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts
steveruizok, line 59: This visibility logic now depends on `Date.now()` inside `getOverlays()`, but there is no reactive timer source driving recomputation. As a result, collaborators will not transition from active to idle / inactive as time passes unless some unrelated editor signal changes.

packages/tldraw/src/index.ts
steveruizok, line 62: This export-surface change removes previously public `@tldraw/tldraw` canvas components like `TldrawHandles`, `TldrawSelectionForeground`, `TldrawScribble`, `TldrawOverlays`, and `TldrawShapeIndicators` with no compatibility shim. That is a source-breaking API change for downstream consumers of the published package.

packages/tlschema/src/styles/TLTheme.ts
steveruizok, line 65: These new fields are now required on `TLThemeDefaultColors`, which means any downstream custom theme object typed against the previous public API will stop compiling until it supplies all of them. If that break is intentional it probably needs an explicit migration / release note; otherwise these should stay backwards-compatible.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
steveruizok: The old implementation kept the selection outline visible while resizing text (`select.resizing` + text shape). That special case is gone here, so text-resize loses its box mid-drag even though the previous behavior intentionally preserved it.

packages/tldraw/src/lib/overlays/CollaboratorHintOverlayUtil.ts
steveruizok, line 35: This new hint path no longer applies the old collaborator-visibility gating before deciding whether to render off-screen arrows. Inactive collaborators, and idle collaborators who are following you, can still show edge hints when they should stay hidden.

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
steveruizok, line 47: `updateHoveredOverlayId()` can set an overlay-specific cursor here, but `onEnter` still unconditionally resets the cursor to default a few lines later. If selection state changes under a stationary pointer, the correct hover cursor is lost until the next pointer move.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
```diff
@@ -0,0 +1,613 @@
...
+				'select.crop.pointing_crop',
+				'select.crop.pointing_crop_handle'
+			) && !editor.getIsReadonly()
+
+		const showResizeHandles =
+			!showCropHandles &&
+			(onlyShape
+				? editor.getShapeUtil(onlyShape).canResize(onlyShape) &&
+					!editor.getShapeUtil(onlyShape).hideResizeHandles(onlyShape)
+				: true)
```
steveruizok: **P0 bug — invisible hit targets on locked shapes / non-display states.**

`getOverlays` emits resize-handle overlays based only on `canResize` / `hideResizeHandles`, but `render` (L342-344) gates on `shouldDisplayControls && !isLockedShape` as well. During `select.brushing`, `select.scribble_brushing`, while a style panel is open, and on locked shapes, the resize handles have hit-test geometry and cursors but no visible render — invisible clickable targets that overlap the shape.

Fix: extract the full predicate block into one private helper (`_computeHandleVisibility(onlyShape, …)`) and call it from both `getOverlays` and `render`. That kills the divergence once and also retires the ~60 lines of duplicated predicate logic.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
```diff
@@ -0,0 +1,613 @@
...
+
+		const showCornerRotateHandles =
+			!isCoarsePointer &&
+			!(isTinyX || isTinyY) &&
+			(onlyShape ? !editor.getShapeUtil(onlyShape).hideRotateHandle(onlyShape) : true)
+
+		const showMobileRotateHandle =
+			isCoarsePointer &&
+			(!isSmallX || !isSmallY) &&
+			(onlyShape ? !editor.getShapeUtil(onlyShape).hideRotateHandle(onlyShape) : true)
```
steveruizok: **P0 bug — same `getOverlays` vs `render` divergence for the mobile rotate handle.**

Here `getOverlays` lacks the `shouldDisplayControls && !isLockedShape` gate that `render` applies at L360-363. Result: on coarse pointers, a clickable mobile-rotate hit target sits over the selection during states like `select.pointing_shape`, `select.brushing`, and on locked shapes, with no visible handle drawn.

Roll into the same `_computeHandleVisibility` helper suggested above.

packages/tldraw/src/lib/overlays/SnapIndicatorOverlayUtil.ts
```diff
@@ -0,0 +1,180 @@
...
+			secondX = maxX
+			secondY = maxY
+		} else {
+			firstX = minX
+			firstY = maxY
+			secondX = maxX
+			secondY = minY
+		}
+
+		ctx.strokeStyle = color
```
steveruizok, line 82: **Missing `ctx.save()` / `ctx.restore()` pair — canvas state leaks to the next util.**

This `render` mutates `strokeStyle` + `lineWidth` without saving/restoring. Same pattern in `BrushOverlayUtil.render` (L52-56, L69-72), `ZoomBrushOverlayUtil`, and `ScribbleOverlayUtil`. All current built-ins re-set what they need, so it's latent — but any third-party util that reads ambient state (or any new util that forgets to set `fillStyle`) will see corrupted values from whichever util ran last.

Two options: (a) wrap each util's render body in `ctx.save()` / `ctx.restore()`, or (b) do it once per util in `OverlayManager`'s render loop. (b) is preferable — guarantees isolation for user-written utils without a discipline requirement.

apps/docs/content/sdk-features/overlay-utils.mdx
```diff
@@ -0,0 +1,254 @@
...
+const BlueOverlay = MyOverlayUtil.configure({ color: 'blue' })
+```
+
+## Accessing overlay utils at runtime
+
+Use [Editor#overlays](?) to interact with the overlay system:
+
+```tsx
+// Get a specific overlay util
+const brushUtil = editor.overlays.getUtil<BrushOverlayUtil>('brush')
```
steveruizok: **Docs method name is wrong.**

The example uses `editor.overlays.getUtil<BrushOverlayUtil>('brush')`, but the actual public method is `getOverlayUtil` (`OverlayManager.ts:51`, api-report entry at `packages/editor/api-report.api.md:2588`). Either fix the mdx to `getOverlayUtil`, or rename the method to `getUtil` — the latter reads better under the `editor.overlays.*` namespace and also fixes the 3-verb inconsistency between `getOverlayUtilsInZOrder` / `getActiveOverlayEntries` / `getCurrentOverlays`.

packages/editor/src/lib/editor/overlays/OverlayUtil.ts
```diff
@@ -0,0 +1,138 @@
...
+	 * Use {@link OverlayUtil.configure} to customize existing overlay utils.
+	 *
+	 * `zIndex` controls paint and hit-test order across utils — higher numbers
+	 * paint on top and are hit-tested first. Ties resolve by registration order.
+	 * Defaults to `0`; built-in utils use larger integers (100, 200, …) with
+	 * gaps so custom utils can slot between.
+	 *
+	 * @public
+	 */
+	options: { zIndex?: number } & Record<string, unknown> = {}
```
steveruizok: **`options` base type is too permissive — breaks `configure()` inference.**

`{ zIndex?: number } & Record<string, unknown>` means any string key type-checks in `configure()`: a typo like `BrushOverlayUtil.configure({ fil: 'red' })` compiles clean on the base class. Subclasses that redeclare `options = { fill: 'red' }` fare better, but the base contract is unsafe.

Suggest `options: { zIndex?: number } = {}` and let subclasses extend with their concrete shape.

packages/editor/src/lib/editor/overlays/OverlayManager.ts
```diff
@@ -0,0 +1,177 @@
...
+	private _hoveredOverlayId = atom<string | null>('hoveredOverlayId', null)
+
+	getHoveredOverlayId(): string | null {
+		return this._hoveredOverlayId.get()
+	}
+
+	getHoveredOverlay(): TLOverlay | null {
+		const id = this._hoveredOverlayId.get()
+		if (!id) return null
+		return this.getCurrentOverlays().find((o) => o.id === id) ?? null
```
steveruizok, line 148: **`getHoveredOverlay` keys by bare `id` — cross-util collisions.**

`this.getCurrentOverlays().find((o) => o.id === id)` matches the first overlay with that `id`, regardless of `type`. Today's built-ins happen not to collide, but a third-party util emitting `id: 'brush'` (or any colliding id) would return the wrong overlay.

Fix: key hover state by `(type, id)` or by object identity. Same consideration applies to `setHoveredOverlay`.

packages/editor/src/lib/editor/overlays/OverlayManager.ts
```diff
@@ -0,0 +1,177 @@
...
+	constructor(public readonly editor: Editor) {}
+
+	/** @internal */
+	readonly _overlayUtils = new Map<string, OverlayUtil>()
+
+	/**
+	 * Register an overlay util instance. Called during editor construction.
+	 * @internal
+	 */
+	registerUtil(util: OverlayUtil) {
```
steveruizok, line 30: **`registerUtil` silently accepts missing/empty `static type`.**

`OverlayUtil.ts:44` declares `static type: string` with no default. A subclass that forgets `static override type = 'foo'` gets `undefined`, and `registerUtil` only throws on duplicate registration, not on missing or empty type. The failure shows up much later as a confusing 'no overlay for type undefined' lookup miss.

Throw loudly here when `util.constructor.type` is falsy.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
```diff
@@ -0,0 +1,613 @@
...
+
+		// Mobile rotate handle
+		if (showMobileRotateHandle) {
+			overlays.push(this._makeOverlay('mobile_rotate', 'mobile_rotate'))
+		}
+
+		return overlays
+	}
+
+	override getGeometry(overlay: TLSelectionForegroundOverlay): Geometry2d | null {
```
steveruizok: **Perf: `getGeometry` recomputes per-util constants for every overlay.**

`bounds`, `expandedBounds`, `rotation`, `zoom`, `transform`, `targetSize*` don't depend on the individual overlay — they're per-tick constants. With ~13 overlays per selection this runs ~13× per pointermove until the `WeakMap` cache fills, and the cache misses on every `getOverlays()` re-run (fresh object identities).

Fix: cache the shared values once per reactive tick — either as a `@computed` derived off `getSelectionRotatedPageBounds` + zoom, or stashed as a non-hit-test field on the util that `getGeometry` reads.

packages/tldraw/src/lib/overlays/ScribbleOverlayUtil.ts
```diff
@@ -0,0 +1,136 @@
...
+	len: number
+	lastX: number
+	lastY: number
+	zoom: number
+	size: number
+	taper: boolean
+	state: TLScribble['state']
+	path: Path2D
+}
+const _scribblePathCache = new Map<string, ScribbleCacheEntry>()
```
steveruizok: **Module-level cache shared across editor instances.**

`_scribblePathCache` lives at module scope. If a consumer mounts two `<Tldraw />` editors on the same page, their scribbles will trample each other's entries (and the cache keys only the `scribble.id`, not the editor instance).

Same pattern in `CollaboratorCursorOverlayUtil.ts` (`_truncateCache` at module scope). Move these onto the util instance so each editor has its own cache.

packages/tldraw/src/lib/overlays/ShapeHandleOverlayUtil.ts
```diff
@@ -0,0 +1,147 @@
...
+		const fgColor = themeColors.selectedContrast
+		const strokeColor = themeColors.selectionStroke
+
+		ctx.save()
+		// Apply shape's page transform
+		ctx.transform(transform.a, transform.b, transform.c, transform.d, transform.e, transform.f)
+
+		for (const overlay of overlays) {
+			const { handle } = overlay.props
+			const fr = (handle.type === 'create' && isCoarse ? 3 : 4) / Math.max(zoom, 0.25)
```
steveruizok: **Zoom clamp is asymmetric inside a single function.**

`fr = (...) / Math.max(zoom, 0.25)` clamps zoom on L131, but `ctx.lineWidth = this.options.lineWidth / zoom` on L136 doesn't. At degenerate zoom (0 or near-zero) the radius is bounded but the stroke width blows up. Either clamp both, or neither — pick whichever matches the rest of the codebase's convention. (A quick grep shows other utils don't clamp at all, so removing the clamp here is probably the right move.)

Also on L135: `ctx.strokeStyle = handle.type === 'clone' ? 'none' : strokeColor` — canvas silently ignores `'none'`, so the assignment is dead (stroke is skipped anyway by the L139 guard). Safe, but misleading — drop the ternary.

steveruizok (commented):
Running this through a few review passes surfaced the list below. I tried to land inline comments on the concrete, file:line issues and keep cross-cutting stuff here.

## P0 — real bugs worth fixing before merge

- **Invisible hit targets on the selection** — `getOverlays` in `SelectionForegroundOverlayUtil` emits mobile-rotate and resize handle overlays without the `shouldDisplayControls` / `!isLockedShape` gates that `render` applies. Inline comments on L94 and L106. Extracting a shared predicate helper would also kill the ~60 lines of duplicated logic between `getOverlays` and `render`.
- **Canvas state leaks between utils** — `SnapIndicatorOverlayUtil`, `BrushOverlayUtil`, `ZoomBrushOverlayUtil`, `ScribbleOverlayUtil` mutate `strokeStyle`/`fillStyle`/`lineWidth`/`globalAlpha` without a `ctx.save()` / `ctx.restore()` pair. Today's built-ins re-set what they need so it's latent, but any user-written util that reads ambient state (or forgets to set `fillStyle`) will see corrupted values. Either wrap each util's `render` in save/restore, or do it once in `OverlayManager` between util calls. Inline comment on the representative case in `SnapIndicatorOverlayUtil`.
- **Docs method name is wrong** — `overlay-utils.mdx:238` uses `editor.overlays.getUtil(...)` but the actual public method is `getOverlayUtil`. Inline comment on the mdx.
- **Stale reference docs for removed components** — `apps/docs/content/reference/` still contains `DefaultBrush.mdx`, `DefaultScribble.mdx`, `DefaultHandles.mdx`, `DefaultSelectionForeground.mdx`, `DefaultCollaboratorHint.mdx`, `DefaultShapeIndicator.mdx`/`*s.mdx`, `TLBrushProps.mdx`, `TLScribbleProps.mdx`, `TLHandlesProps.mdx`, `TLSelectionForegroundProps.mdx`, `TLSnapIndicatorProps.mdx`, `TLCollaboratorHintProps.mdx`, `TLShapeIndicatorProps.mdx`/`*sProps.mdx`, `TLShapeIndicatorErrorFallbackComponent.mdx` for APIs this PR removes. The docs build will link to nothing — delete them alongside the code removal.
- **No error boundary around `render()` / `getOverlays()`** — `ShapeIndicatorErrorFallbackComponent` was the old safety net, and this PR removes it without replacement. A throw in any single util (including a user-written one) now takes down the entire canvas overlay layer. Consider per-util try/catch in `OverlayManager.getCurrentOverlays` and in the render loop.

## P1 — should-fix

- Release notes should call out the `LiveCollaborators` → 4 utils migration explicitly (users who replaced `LiveCollaborators` wholesale now need to replace `CollaboratorBrush/Cursor/Hint/Scribble`).
- Naming churn in `OverlayManager`: `getOverlayUtilsInZOrder` / `getActiveOverlayEntries` / `getCurrentOverlays` — three verbs for related things. Renaming `getOverlayUtil` → `getUtil` would also fix the doc example.
- `getHoveredOverlay` / `getHoveredOverlayId` / `setHoveredOverlay` are `(undocumented)` in api-report — add tsdoc.
- `_overlayUtils` is a `readonly` public-with-underscore leaking to api-report. Make it `#overlayUtils` (or rename and mark truly internal).

## P2 — perf / maintainability

- `SelectionForegroundOverlayUtil` has ~60 lines of predicate logic duplicated between `getOverlays` (L58-165) and `render` (L280-509). Same helper that fixes the P0 bug above would pay off here.
- `SnapIndicatorOverlayUtil.ts:62-66` runs four separate `points.reduce` passes for min/max; collapse to one loop.
- `SelectionForegroundOverlayUtil.ts:515-540` allocates two inline `cursorMap` object literals per call — hoist to module scope.
- `SelectionForegroundOverlayUtil.ts:456` has a commented-out line — remove or explain.
- Color-lookup boilerplate `editor.getCurrentTheme().colors[editor.getColorMode()]` is copy-pasted in ~7 utils, as is `lineWidth / zoom`. A helper on `OverlayUtil` base (`this.getColors()`, `this.scale(lineWidth, zoom)`) would be worth it.
- Magic numbers across utils (2.5, 3.5, 1.75, 0.75, 8, 14, 20) unnamed/uncommented.

## P3 — test coverage gaps

- No E2E pixel coverage for overlays (`apps/examples/e2e/tests/test-overlays.spec.ts` doesn't exist). A small canvas-pixel suite would catch regressions the unit tests can't.
- Not visibly covered: rotated selection handle geometry (`getGeometry` with non-zero `getSelectionRotation()`), `isCoarsePointer: true` branches, extreme zoom (0.1 and 10+) exercising `Math.max(zoom, 0.25)` in `ShapeHandleOverlayUtil`, theme switch mid-render, `OverlayUtil.configure()` override flow, `_geometryCache` invalidation when `getOverlays()` emits fresh instances.

## Checked and fine

- `configure()` static options merging (`OverlayUtil.ts:69-78`) — `this` binds to the parent class at class-expression time; spread reads parent's static options correctly.
- `OverlayManager.getOverlayAtPoint` walk-order vs paint-order pairing — matches.
- `SnapIndicatorOverlayUtil._renderGaps` nullability after the earlier fix.
- `ArrowHintOverlayUtil.isActive` `getStateDescendant(...)!` — preceded by `isIn` check, safe.
- No remaining in-repo callers of removed `TLEditorComponents` entries in `apps/examples`, `templates`, or `packages/tldraw/src/lib/canvas`.
- `defaultOverlayUtils` exported as a `readonly` tuple preserving identity — good for filter patterns.
- `overlayUtils?: readonly TLAnyOverlayUtilConstructor[]` prop symmetric across `TLEditorProps` and `TldrawProps`.

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
steveruizok: Already handled — `Idle.ts:157` has a dedicated `case 'mobile_rotate'` that re-dispatches with `target: 'selection'` and the `mobile_rotate` handle, and the `'selection'` switch at L207 enters `pointing_rotate_handle`. Addressed in 7187729e1f.

steveruizok: Already handled — `Idle.ts:157` has a dedicated `case 'mobile_rotate'` that re-dispatches with `target: 'selection'` and the `mobile_rotate` handle, and the `'selection'` switch at L207 enters `pointing_rotate_handle`. Addressed in 7187729e1f.

packages/tldraw/src/lib/tools/SelectTool/childStates/Crop/children/Idle.ts
steveruizok: Already handled — `Crop/children/Idle.ts:62` includes `mobile_rotate` in the remap set, so the crop rotate control re-dispatches as a `selection` handle event into `select.pointing_rotate_handle`. Addressed in 7187729e1f (plus a cropping test to lock it in).

steveruizok: Already handled — `Crop/children/Idle.ts:62` includes `mobile_rotate` in the remap set, so the crop rotate control re-dispatches as a `selection` handle event into `select.pointing_rotate_handle`. Addressed in 7187729e1f (plus a cropping test to lock it in).

packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts
steveruizok: Fixed in 7187729e1f — `getOverlays()` now reads `editor._collaboratorVisibilityClock.get()` before sampling `Date.now()`, so overlays invalidate as presence ages. The Editor runs an interval that ticks the atom, so active → idle → inactive transitions recompute without relying on unrelated editor signals.

steveruizok: Fixed in 7187729e1f — `getOverlays()` now reads `editor._collaboratorVisibilityClock.get()` before sampling `Date.now()`, so overlays invalidate as presence ages. The Editor runs an interval that ticks the atom, so active → idle → inactive transitions recompute without relying on unrelated editor signals.

packages/tldraw/src/lib/overlays/CollaboratorHintOverlayUtil.ts
steveruizok: Fixed in 7187729e1f — `_shouldShow` now funnels through a shared `shouldShowCollaborator` helper that applies the same idle / inactive / `followingUserId` / `highlightedUserIds` gating as the cursor util. Covered by new cases in `CollaboratorHintOverlayUtil.test.ts`.

steveruizok: Fixed in 7187729e1f — `_shouldShow` now funnels through a shared `shouldShowCollaborator` helper that applies the same idle / inactive / `followingUserId` / `highlightedUserIds` gating as the cursor util. Covered by new cases in `CollaboratorHintOverlayUtil.test.ts`.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
steveruizok: Fixed in 29f95556ed — `_computeSelectionState()` is now the single source of truth for both `getOverlays()` and `render()`, and its `shouldDisplayBox` explicitly preserves the old text-resize case: `select.resizing` while the only selected shape is a text shape still renders the selection box (`SelectionForegroundOverlayUtil.ts:533-549`).

steveruizok: Fixed in 29f95556ed — `_computeSelectionState()` is now the single source of truth for both `getOverlays()` and `render()`, and its `shouldDisplayBox` explicitly preserves the old text-resize case: `select.resizing` while the only selected shape is a text shape still renders the selection box (`SelectionForegroundOverlayUtil.ts:533-549`).

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
steveruizok: `onEnter` no longer contains an unconditional `setCursor({ type: 'default' })` — the current version just calls `parent.setCurrentToolIdMask(undefined)`, then `updateHoveredOverlayId`/`updateHoveredShapeId`. Cursor state is now fully owned by `updateHoveredOverlayId`: it sets the overlay's cursor when one is hovered, and only resets to default on the edge where we leave an overlay. No cursor-reset race remains at this site.

steveruizok: `onEnter` no longer contains an unconditional `setCursor({ type: 'default' })` — the current version just calls `parent.setCurrentToolIdMask(undefined)`, then `updateHoveredOverlayId`/`updateHoveredShapeId`. Cursor state is now fully owned by `updateHoveredOverlayId`: it sets the overlay's cursor when one is hovered, and only resets to default on the edge where we leave an overlay. No cursor-reset race remains at this site.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
steveruizok: Fixed in 29f95556ed — `getOverlays()` and `render()` now share `_computeSelectionState()` as the single source of truth. `shouldDisplayControls`, `!isLockedShape`, and `showResizeHandles` are computed once, so hit-test and render geometry can't drift across `brushing`, `scribble_brushing`, style-panel-open, or locked-shape states.

packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
steveruizok: Fixed in 29f95556ed — `_computeSelectionState()` computes `showMobileRotateHandle` with the full `shouldDisplayControls && !isLockedShape && isCoarsePointer` predicate (`SelectionForegroundOverlayUtil.ts:518-523`), and both `getOverlays()` and `render()` read from it. No more invisible mobile-rotate hit target during `pointing_shape` / `brushing` / on locked shapes.

apps/docs/content/sdk-features/overlay-utils.mdx
steveruizok: Fixed in 9b283e70aa — mdx now uses `editor.overlays.getOverlayUtil<BrushOverlayUtil>('brush')` to match the real method name. Keeping the existing name avoids a breaking rename to `.getUtil`; we can revisit the `getOverlayUtilsInZOrder` / `getActiveOverlayEntries` / `getCurrentOverlays` naming separately if it's worth the churn.

packages/editor/src/lib/editor/overlays/OverlayManager.ts
steveruizok: Fixed in 9b283e70aa — `registerUtil` now throws `Overlay util <ClassName> is missing a static 'type' property.` when the constructor's `type` is falsy, so a forgotten `static override type = 'foo'` surfaces at registration time instead of as a later `No overlay util found for type: "undefined"` lookup.

packages/tldraw/src/lib/overlays/SnapIndicatorOverlayUtil.ts
steveruizok: Already done — the render loop wraps each util's `render()` in `ctx.save()` / `ctx.restore()` at the manager level: `CanvasOverlays.tsx:49-53` and `MinimapManager.ts:272-276`. Third-party utils get isolated ambient state without having to remember to save/restore themselves, matching the reviewer's preferred option (b).

packages/tldraw/src/index.ts
steveruizok: Acceptable — this is an intentional, deliberate API-surface change for the overlay system migration; it'll be called out in the release notes rather than shimmed.

steveruizok: Acceptable — this is an intentional, deliberate API-surface change for the overlay system migration; it'll be called out in the release notes rather than shimmed.

packages/tlschema/src/styles/TLTheme.ts
steveruizok: Acceptable — the expanded `TLThemeDefaultColors` shape is intentional and will be covered by the release notes; downstream custom themes should update to supply the new fields.

steveruizok: Acceptable — the expanded `TLThemeDefaultColors` shape is intentional and will be covered by the release notes; downstream custom themes should update to supply the new fields.

packages/editor/src/lib/editor/overlays/OverlayManager.ts
steveruizok: Acceptable — overlay ids should be globally unique across utils, and that's a constraint we should enforce at registration/emission rather than disambiguating lookups by `(type, id)`. Filing a follow-up to add a uniqueness check in `OverlayManager`.

packages/tldraw/src/lib/overlays/ShapeHandleOverlayUtil.ts
steveruizok: Fixed in fdcb855c74 — dropped the `Math.max(zoom, 0.25)` clamp on `fr` to match the rest of the overlay utils (none of them clamp), and removed the dead `strokeStyle = 'none'` ternary since the L139 guard already skips `stroke()` for `clone` handles. Also hoisted `strokeStyle` + `lineWidth` out of the loop while I was in there, since they don't vary per-overlay.

PR #8492, opened 2026-04-08
fix(zero): drop custom litestream config in favor of zero defaults

MitjaBezensek:
In order to simplify our fly.io deployment and rely on zero's built-in litestream configuration instead of maintaining our own, this PR removes the custom `[[files]]` litestream config sections from both the replication manager and view syncer templates.

Zero already supports `ZERO_LITESTREAM_ENDPOINT` for non-AWS S3 endpoints (like Cloudflare R2), and its default config handles the metrics `addr` (port+2 = 4850), replica path, bucket, and auth via `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` fly secrets. The only thing we needed to add was the `ZERO_LITESTREAM_ENDPOINT` env var so zero knows the R2 endpoint URL.

Also enables multinode deployment for previews so the replication manager + view syncer templates are actually tested in preview deploys (previously previews used single-node which doesn't have litestream at all).

### Change type

- [x] `improvement`

### Test plan

1. Deploy preview with `dotcom-preview-please` + `preview-flyio-zero-deploy-please` labels (now multinode)
2. Check `flyctl logs` for litestream errors (endpoint resolution, auth failures)
3. Verify backup files appear in Cloudflare R2 bucket at the expected path
4. Confirm changelog draining: `SELECT pg_size_pretty(pg_total_relation_size('zero."changeLog"'));` stays small
5. Verify backup-monitor logs show successful watermark determination

### Release notes

- Remove custom litestream configuration from zero fly.io deployments in favor of zero's built-in defaults

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +2 / -32   |
| Config/tooling | +5 / -8 |


MitjaBezensek:
I tested out the default zero config and seems to work with r2. Looks like the region issue might have been just with flyio tigris deployment.

PR #8493, opened 2026-04-08
fix(sync): handle invalidated replication slots in TLPostgresReplicator

MitjaBezensek:
In order to let TLPostgresReplicator self-heal from invalidated replication slots, this PR replaces the `WHERE NOT EXISTS` slot creation SQL with an `ensureValidSlot()` method that checks `pg_replication_slots.wal_status`.

During the April 7 incident, PG invalidated tlpr's replication slot during crash recovery. tlpr entered an infinite retry loop for 2+ hours because `WHERE NOT EXISTS` only checks slot *existence*, not *validity*. Zero's replicator self-healed in ~2 minutes.

When `wal_status = 'lost'` (invalidated), `ensureValidSlot()`:
1. Fires a Sentry alert
2. Drops the dead slot
3. Increments a `slotGeneration` counter in SQLite
4. Derives a new slot name from `slotNameBase` + generation suffix (e.g. `tlpr_abc...xy_1`)
5. Updates `meta.slotName` and nulls `meta.lsn`, which triggers a sequence break → all user DOs reboot
6. Creates the new slot

The new slot name changes all `sequenceId`s sent to user DOs, guaranteeing they detect the break and hard reset. This uses the same mechanism as the existing "force reboot by changing slot name" escape hatch described in migration 003.

Called in two places:
- **Constructor's `blockConcurrencyWhile`** — catches invalidation that happened while the DO was down
- **`boot()`** — the critical fix, catches invalidation during runtime (replication errors out → retry → `ensureValidSlot()` recovers)

On normal deploys (generation 0), the slot name is unchanged — no existing DOs are affected.

### Change type

- [x] `bugfix`

### Test plan

1. Next time a slot gets invalidated, tlpr should self-heal within one reboot cycle (~2-4s) instead of looping forever
2. Can verify manually: invalidate a slot, check that tlpr drops it, creates a new one with `_1` suffix, and all user DOs reboot
3. Verify normal deploys (generation 0) don't change slot name or trigger reboots

### Release notes

- Fix TLPostgresReplicator infinite retry loop when PostgreSQL invalidates a replication slot

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +45 / -3   |

### DIFF
--- a/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
+++ b/apps/dotcom/sync-worker/src/TLPostgresReplicator.ts
@@ -94,7 +94,8 @@ export class TLPostgresReplicator extends DurableObject<Environment> {
 	private log
 
 	private readonly replicationService
-	private readonly slotName
+	private readonly slotNameBase
+	private slotName
 	private readonly wal2jsonPlugin = new Wal2JsonPlugin({
 		addTables:
 			'public.user,public.file,public.file_state,public.user_mutation_number,public.replicator_boot_id,public.group,public.group_user,public.group_file',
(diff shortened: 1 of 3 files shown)


ds300 (commented):
this will _probably_ be fine but i think it might be safest to also generate a new slotName before creating the new replication slot? that's what the sequence ids are based on, and changing that is what is guaranteed to cause downstream user DOs to hard reset. Might be a little more involved since at the moment the slotName is readonly and hardcoded, but i'm sure claude can figure it out

MitjaBezensek:
Closing in favor of https://github.com/tldraw/tldraw/pull/8510

PR #7376, opened 2025-12-13
feat(editor): add right-click to drag preference

AniKrisn:
In order to allow users to pan the canvas using the right mouse button (a common interaction pattern in other canvas tools), this PR adds a new `isRightClickToDrag` user preference that enables right-click panning while preserving context menu access on non-drag right-clicks.

Closes #5808

### Change type

- [x] `feature`

### Test plan

1. Open the Preferences menu from the hamburger menu.
2. Enable "Drag on right-click".
3. Right-click and drag on the canvas — it should pan.
4. Right-click without dragging — the context menu should still appear.
5. Disable the preference and verify normal right-click behavior is restored.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add a new "Drag on right-click" preference that allows panning the canvas with the right mouse button.

### API changes

- Added `TLUserPreferences.isRightClickToDrag` preference field
- Added `UserPreferencesManager.getIsRightClickToDrag()` accessor
- Added `ToggleRightClickToDragItem` menu component
- Added `toggle-right-click-to-drag` action

Issue #5808, opened 2025-04-03
Add right-click drag panning and scroll zoom toggle


ColemanDunn:
### What's the feature?

This is an attempt at reviving #1254 and [#1628](https://github.com/tldraw/tldraw/issues/1628) of not only adding right click drag to pan, but also scroll to zoom as a default setting. These these features allow very quick editing/canvas navigation with one hand only! I think it makes a lot of sense for a canvas-centric app to have this behavior as you are typically not just editing up and down, but in all directions (arguably more left and right and up and down honestly). Scrolling to go up in down is makes a lot of sense when reading an article on a regular webpage but not as much when editing on a canvas that can move in any direction. 

Both of the linked issues have comments talking not having seen this behavior or asking about other apps that work like this, so I have made a list of just a couple of very popular canvas apps that have this:

- Lucid Spark
- Figma (both of these features are toggleable in settings)
- Discord whiteboard
- Miro
- Unreal Engine
- draw.io (right click drag only)

It should be easy to try any of those out to see how great it is to navigate. These features allow for some of the following things:

- easy one-handed editing
- no need to switch tools to simply navigate the canvas
- easily zoom in to parts of the canvas with scroll (scrolling up and down doesn't make much sense in a canvas)
- more in-line with other popular canvas apps

I would love to see if this either or both of these ideas are supported and would happily help with or start a PR to get this feature in!

Thanks!

### Where would you like this feature?

The tldraw.com website

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ColemanDunn:
example of Figma settings (These settings are amazing and make figma so much easier to work with):

![Image](https://github.com/user-attachments/assets/e33732be-ebcd-42e6-a8f5-397c3c3b990a)

mimecuvalo:
@AniKrisn is exploring this in this PR https://github.com/tldraw/tldraw/pull/7376

steveruizok:
Scroll zoom toggle is in! #6164

steveruizok:
The scroll zoom toggle shipped in #6164. Closing this since the main ask has been addressed — if right-click drag panning is still wanted, please open a separate issue for it.

PR #8501, opened 2026-04-08
feat(editor): right-click and drag to pan the camera

kostyafarber:
In order to support a common panning interaction from tools like Figma, this PR lets users pan the canvas with right-click + drag while keeping a static right-click as the trigger for the context menu. Enabled by default via the new `rightClickPanning` editor option (mirroring `spacebarPanning`) — set it to `false` to restore the old behavior where right-click opens the context menu on press with no drag-to-pan. Relates to #5808. Clean reimplementation of #7376.

Key pieces:

- Added `rightClickPanning` to `TldrawOptions` (defaults to `true`) and gated the right-click drag logic on it in `Editor`, `useCanvasEvents`, and `MenuClickCapture`.
- `InputsManager` tracks a pre-drag right-click window (`getIsRightPointing()` / `setIsRightPointing()`), and the `Editor` pointer-move handler only transitions into `isPanning` once the pointer crosses the drag threshold.
- `MenuClickCapture` captures right-button pointer events on an open menu and re-dispatches them to the canvas, so drag-then-release pans (instead of selecting) and a second right-click while the menu is open can re-open the menu at the new location.
- `useCanvasEvents` dispatches a synthetic `contextmenu` event on static right-click `pointerup`, so the menu opens on release rather than on press.
- Playwright coverage for the new right-click panning and context menu flows.

### Change type

- [x] `improvement`

### Test plan

1. Right-click on empty canvas — context menu opens at the release position.
2. Right-click + drag — camera pans with a grabbing cursor, no context menu opens on release.
3. With the context menu open, right-click + drag elsewhere — menu closes and the canvas pans.
4. With the context menu open, right-click (no drag) at a different location — the old menu closes and a new one opens at the new position.
5. Right-click + drag over shapes — canvas pans without selecting shapes.
6. Middle-mouse panning still works as before.
7. Construct an editor with `options: { rightClickPanning: false }` — right-click opens the context menu on press and drag does not pan.

- [x] Unit tests
- [x] End to end tests

### Release notes

- Right-click and drag to pan the camera. A static right-click still opens the context menu. The new `rightClickPanning` editor option (default `true`) can be set to `false` to disable.

### API changes

- Added `TldrawOptions.rightClickPanning` (default `true`) — gates the right-click + drag panning behavior.
- Added `InputsManager.getIsRightPointing()` — whether the user is right-click pointing before the drag threshold is crossed.
- Added `InputsManager.setIsRightPointing(isRightPointing)` — internal setter.

### Code changes

| Core code       | +254 / -87  |
| Tests           | +302 / -14  |
| Automated files | +5 / -0     |


on5sl:
Would love to see this! Many of our users are so used to the Miro experience they are completely thrown-off by having to pan differently.

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -47,6 +52,19 @@ export function useCanvasEvents() {
...
+					Vec.Dist2(editor.inputs.getOriginScreenPoint(), editor.inputs.getCurrentScreenPoint()) <
+					editor.options.dragDistanceSquared
+				if (pointerDownNearPointerUp && e.button === 2) {
+					rWasStaticRightClick.current = true
+					const contextMenuEvent = new MouseEvent('contextmenu', {
+						bubbles: true,
+						clientX: e.clientX,
+						clientY: e.clientY,
+					})
+					e.currentTarget.dispatchEvent(contextMenuEvent)
```
MitjaBezensek: Should this depend on `isRightClickToDrag`? Otherwise we there might be event duplication? Might be harmless 🤷‍♂️

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -16,12 +16,17 @@ export function useCanvasEvents() {
...
+
 	const events = useMemo(
 		function canvasEvents() {
 			function onPointerDown(e: React.PointerEvent) {
 				if (editor.wasEventAlreadyHandled(e)) return
+				rWasStaticRightClick.current = false
 
-				if (e.button === RIGHT_MOUSE_BUTTON) {
+				// Handle right-click when right-click-to-drag is disabled
+				// This matches main branch behavior - dispatch right_click and return early
```
MitjaBezensek: Once this is merged the comment doesn't make much sense as it will become main.

MitjaBezensek (commented):
In some cases I can get it to pan, as well as show the context menu. The movement is really minor though:

https://github.com/user-attachments/assets/9847b8c0-f76d-4098-b8d6-6b99f8bcb60d

Also noticed a slight bug if this feature is used together with spacebar panning. As soon as you let go of spacebar the panning stops (even if you still hold the right mouse button). Probably a weird edge case though.

PR #8506, opened 2026-04-09
Add 10% production pierre load test

ds300:
I spoke with @fat from pierre and agreed to a 10% trial of production traffic so we could get a sense of how much space we'd consume as well as what kind of write throughput we'd expect to see.

For throughput I added a workers analytics event to track the number of bytes written in incremental writes. Then we can sum all the events in grafana and multiply by 10 to get a realistic view of write throughput.

For disk space consumption I'll be running some separate tests using staging R2 snapshot data to 'migrate' existing rooms to pierre to get a sense of what kind of compression factor it offers (i think it'll be logarithmic and dependent on usage patterns), and then we can try to apply that to the production R2 bucket to get a good sense of how much space we'd be using.

### Change type

- [x] `other`


MitjaBezensek (approved):
Looks good. We probably could reuse the existing percentage base feature flag? Doesn't matter much though 🤷‍♂️

PR #8532, opened 2026-04-13
move cmd+shift+v paste default shortcut from svg to png

kostyafarber:
it is more common to copy png then it is svg, lets make the shortcut apply to pngs instead of svgs! Note this is a breaking change.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- Changed the default "Copy as" keyboard shortcut from SVG to PNG. 💥 This is a breaking change.


mimecuvalo (approved):
nice. just mark it as a breaking change in the PR desc pls!

PR #8535, opened 2026-04-13
fix(mcp-app): safe analytics when logging exec/search code

max-drake:
In order to trace which code runs in `exec` and `search` without risking tool failures or oversized Analytics Engine payloads, this PR logs tool code via a shared helper that UTF-8–truncates the `code` blob to stay under the 16 KB per-datapoint blob limit and wraps `writeDataPoint` in try/catch so analytics never breaks a tool call.

### Change type

- [x] `bugfix`
- [ ] `improvement` | `feature` | `api` | `other`

### Test plan

1. Call `exec` with a short script and confirm normal behavior.
2. Call `exec` with a script whose UTF-8 length would exceed the remaining blob budget after the fixed `tool_called` / tool name prefixes; confirm the tool still succeeds and analytics receives a truncated payload ending in `...[truncated]` if inspected.
3. Call `search` similarly with a long query script.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Apps           | +67 / -8   |
| Core code      | +0 / -0    |
| Tests          | +0 / -0    |
| Automated files | +0 / -0   |
| Documentation  | +0 / -0    |
| Templates      | +0 / -0    |
| Config/tooling | +0 / -0    |

### DIFF
--- a/apps/mcp-app/src/shared/utils.ts
+++ b/apps/mcp-app/src/shared/utils.ts
@@ -64,7 +64,69 @@ export function resolveMcpAppHostNameFromClientInfo(
 }
 
 const CODE_EDITOR_HOST_NAMES: MCP_APP_HOST_NAMES[] = ['cursor', 'vscode']
+const ANALYTICS_ENGINE_MAX_BLOB_BYTES = 16 * 1024
+const TRUNCATED_ANALYTICS_SUFFIX = '...[truncated]'
 
 export function isHostCodeEditor(hostName: MCP_APP_HOST_NAMES): boolean {
 	return CODE_EDITOR_HOST_NAMES.includes(hostName)
+}
+
+function truncateUtf8String(value: string, maxBytes: number): string {
+	const encoder = new TextEncoder()
+	if (maxBytes <= 0) return ''
+	if (encoder.encode(value).byteLength <= maxBytes) return value
+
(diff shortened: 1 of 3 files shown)


apps/mcp-app/src/shared/utils.ts
```diff
@@ -64,7 +64,69 @@ export function resolveMcpAppHostNameFromClientInfo(
...
+	const maxCodeBytes =
+		ANALYTICS_ENGINE_MAX_BLOB_BYTES -
+		baseBlobs.reduce((total, blob) => total + encoder.encode(blob).byteLength, 0)
+
+	try {
+		analytics.writeDataPoint({
+			blobs: [...baseBlobs, truncateUtf8String(code, maxCodeBytes)],
+		})
+	} catch {
+		// Best-effort analytics should never break a tool call.
```
kostyafarber: if analytics write does fail, we could log something so we can diagnose why instead of it failing silently

kostyafarber (approved):
lgtm modulo the logginc comment

apps/mcp-app/src/shared/utils.ts
max-drake: updated the comment to reflect the fact that `writeDataPoint()` never throws, so sadly this is just best-effort

PR #8543, opened 2026-04-14
feat(shapes): add extensibility API for custom geo shape types

### LINKED ISSUE #8076: Add extensibility API for custom geo shape types
## Problem

The geo shape system currently supports 20 hardcoded shape types (rectangle, ellipse, star, cloud, etc.), but there's no way for SDK consumers to add custom geo types without forking or monkey-patching GeoShapeUtil. All geo type behavior is locked behind switch statements that use `exhaustiveSwitchError()` for type safety, making extension impossible.

Developers who want to add domain-specific shapes (e.g. flowchart symbols, circuit diagram components, architectural shapes) currently have to either create entirely new ShapeUtils from scratch—duplicating all the label, resize, and styling logic that geo shapes already handle—or fork the GeoShapeUtil.

## Proposed solution

Give GeoShapeUtil an option that accepts a map of "geo utils"—small bundles that plug into the existing switch statements. Each geo util would define the shape-specific behavior for a single geo type:

```ts
interface GeoTypeDefinition {
  // Path/geometry generation
  getPath(w: number, h: number, id: string): PathBuilder
  // Snap behavior: 'polygon' snaps to vertices + center, 'blobby' snaps to center only
  snapType: 'polygon' | 'blobby'
  // Default creation size (optional, defaults to 200x200)
  defaultSize?: { w: number; h: number }
  // Icon for the style panel picker
  icon: string
  // Double-click behavior (optional)
  onDoubleClick?: (shape: TLGeoShape) => TLGeoShape | void
}
```

SDK consumers would register custom geo types through GeoShapeUtil options:

```ts
const MyGeoShapeUtil = GeoShapeUtil.configure({
  customGeoTypes: {
    'my-custom-shape': {
      getPath: (w, h) => { /* ... */ },
      snapType: 'polygon',
      icon: 'my-icon',
    },
  },
})
```

## Implementation areas

The following switch statements and integration points would need to be updated to consult the geo type map:

1. **Path generation** — `getGeoShapePath.ts` `_getGeoPath()` — the main 20-case switch
2. **Snap geometry** — `GeoShapeUtil.tsx` `getHandleSnapGeometry()` — polygon vs blobby categorization
3. **Default sizing** — `toolStates/Pointing.ts` `complete()` — creation size per type
4. **Double-click toggle** — `GeoShapeUtil.tsx` `onDoubleClick()` — special interactions
5. **Style panel UI** — `styles.tsx` `STYLES.geo` — icon entries for the picker

## Use cases

- **Domain-specific diagramming**: flowchart shapes, UML components, circuit symbols
- **Branded shape libraries**: company-specific shapes that inherit all geo behavior
- **Community extensions**: shareable geo type packs without forking the SDK

mimecuvalo:
In order to allow SDK consumers to add custom geo types without forking or monkey-patching GeoShapeUtil, this PR adds a `GeoTypeDefinition` interface and a `customGeoTypes` option to `GeoShapeUtil.configure()`. Closes #8076.

Custom geo types plug into the existing switch statements and inherit all standard geo shape behavior — labels, resizing, fill/dash/color styling, SVG export, hyperlink support, and the full editing UX — while providing their own path geometry, snap behavior, creation size, style panel icon, and optional double-click handler.

### Concepts

| Term | Type | Meaning |
|------|------|---------|
| `GeoTypeDefinition` | interface | Defines the behavior bundle for a single custom geo type |
| `customGeoTypes` | `GeoShapeOptions` field | Map of custom type name → definition, passed via `configure()` |

### Example

**Registering a custom geo type:**

```ts
import { GeoShapeUtil, PathBuilder } from 'tldraw'

const MyGeoShapeUtil = GeoShapeUtil.configure({
  customGeoTypes: {
    'rounded-rect': {
      getPath: (w, h, shape, strokeWidth) => {
        const r = Math.min(w, h) * 0.2
        return new PathBuilder()
          .moveTo(r, 0, { geometry: { isFilled: shape.props.fill !== 'none' } })
          .lineTo(w - r, 0)
          .circularArcTo(r, false, true, w, r)
          .lineTo(w, h - r)
          .circularArcTo(r, false, true, w - r, h)
          .lineTo(r, h)
          .circularArcTo(r, false, true, 0, h - r)
          .lineTo(0, r)
          .circularArcTo(r, false, true, r, 0)
          .close()
      },
      snapType: 'polygon',
      icon: 'geo-rectangle',
      defaultSize: { w: 200, h: 150 },
    },
  },
})

// Pass to Tldraw
<Tldraw shapeUtils={[MyGeoShapeUtil]} />
```

### `GeoTypeDefinition` interface

| Field | Type | Description |
|-------|------|-------------|
| `getPath` | `(w, h, shape, strokeWidth) => PathBuilder` | Generate the path geometry for this type |
| `snapType` | `'polygon' \| 'blobby'` | `'polygon'` snaps to vertices + center; `'blobby'` snaps to center only |
| `icon` | `string` | Icon name for the style panel geo picker |
| `defaultSize` | `{ w: number; h: number }` (optional) | Default size when clicking (not dragging). Defaults to 200x200 |
| `onDoubleClick` | `(shape) => { props } \| void` (optional) | Custom double-click handler |

### New examples

- `shapes/custom-geo-types` — Demonstrates registering two custom geo types (rounded-rect and cross) via `GeoShapeUtil.configure()`, showing how custom types inherit all standard geo behavior

### Change type

- [x] `feature`

### Test plan

1. Open the examples app (`yarn dev`) and navigate to Shapes > Custom geo types
2. Verify the two custom shapes (rounded-rect and cross) render correctly with labels and fill
3. Select the geo tool and open the style panel — custom types should appear at the bottom of the geo picker
4. Click to create a shape — should use the custom `defaultSize`
5. Drag to create a shape — should resize normally
6. Verify label editing, fill/dash/color styling, and SVG export work
7. Verify snap points match the configured `snapType`

- [x] Unit tests (existing geo tests pass)

### Release notes

- Add extensibility API for custom geo shape types via `GeoShapeUtil.configure({ customGeoTypes })`. Custom types define their own path geometry, snap behavior, creation size, style panel icon, and double-click handler while inheriting all standard geo shape behavior.

### API changes

- Added `GeoTypeDefinition` interface (public)
- Added `customGeoTypes` optional field to `GeoShapeOptions`
- Added `configure()` override on `GeoShapeUtil` that registers custom types in the geo enum, path system, and style panel

### Code changes

| Core code       | +152 / -10 |
| Automated files | +20 / -18  |
| Documentation   | +141 / -0  |


packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -113,6 +114,25 @@ export interface GeoShapeOptions extends ShapeOptionsWithDisplayValues<
...
+	 *     'my-shape': {
+	 *       getPath: (w, h) => new PathBuilder().moveTo(0, 0).lineTo(w, 0).lineTo(w, h).lineTo(0, h).close(),
+	 *       snapType: 'polygon',
+	 *       icon: 'geo-rectangle',
+	 *     },
+	 *   },
+	 * })
+	 * ```
+	 */
+	customGeoTypes?: Record<string, GeoTypeDefinition>
```
MitjaBezensek: Maybe we can use `customGeoStyles` as it would be more in line with existing `GeoShapeGeoStyle`?

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
```diff
@@ -121,6 +141,24 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 	static override props = geoShapeProps
 	static override migrations = geoShapeMigrations
 
+	static override configure<T extends TLShapeUtilConstructor<any, any>>(
```
MitjaBezensek, line 145: Should this also delegate to `super.configure`? Otherwise we might miss some of the parent steps?

apps/examples/src/examples/shapes/tools/custom-geo-types/CustomGeoTypesExample.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+// [1]
+const CustomGeoShapeUtil = GeoShapeUtil.configure({
+	customGeoTypes: {
+		'rounded-rect': {
+			getPath(w, h, shape) {
+				const isFilled = shape.props.fill !== 'none'
+				const r = Math.min(w, h) * 0.2
+				return new PathBuilder()
+					.moveTo(r, 0, { geometry: { isFilled } })
```
MitjaBezensek, line 26: Maybe we should add a comment why `isFilled` is needed?

MitjaBezensek (commented):
Looking good!

1. Missing tooltip label in the example, worth adding, so users understand how to do it:
Image

2. Should the new geo styles be automatically added to the the toolbar? We do add them to the style panel dropdown, but the toolbar seems to list items one by one and thus misses the new options 🤷‍♂️

3. What if somebody uses an existing option, like `rectangle`? Maybe we should add at least a console.warn in dev mode if that happens (otherwise we just drop it)?

packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
mimecuvalo: good call

mimecuvalo:
> Should the new geo styles be automatically added to the the toolbar? We do add them to the style panel dropdown, but the toolbar seems to list items one by one and thus misses the new options 🤷‍♂️

mm, i don't think so, we don't list all geo shapes in the main toolbar, so i'd rather this piece be something that people have to wire in. could be convinced though.

PR #8378, opened 2026-03-30
fix(shapes): fix shadow artifact and oversized caret on notes/geo at high zoom in dynamic size mode

AniKrisn:
In order to fix visual glitches with sticky notes and geo shapes in dynamic size mode at high zoom levels, this PR clamps the note shadow offset and switches both note and geo shape labels to use CSS `transform: scale()` instead of direct fontSize/padding multiplication.

image
image

Closes #8377

### Change type

- [x] `bugfix`

### Test plan

1. Enable dynamic size mode (View menu → Dynamic size mode)
2. Create a sticky note and a geo shape with a text label
3. Zoom to 800%
4. Verify no shadow appears above the sticky note
5. Double-click to edit text in both shapes — verify the caret is proportionally sized

- [x] Unit tests

### Release notes

- Fix shadow rendering above sticky notes at high zoom in dynamic size mode
- Fix oversized text caret when editing notes and geo shapes at high zoom in dynamic size mode

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +18 / -5   |

PR #8570, opened 2026-04-16
perf(shapes): skip CSS transform when dynamic size is not active

### LINKED ISSUE #8562: Rotation performance regression from dynamic size style transforms
## Description

There's a performance regression in rotation caused by #8378. The PR introduced CSS `transform: scale()` on shape label containers (geo, note) to fix shadow artifacts and oversized carets at high zoom in dynamic size mode. However, this transform is applied **unconditionally** — even when dynamic size is not enabled (i.e., `scale === 1`), which incurs unnecessary compositing cost during rotation.

Image

The graph shows a clear drop in rotation performance starting around April 4 (when #8378 was merged).

## Root cause

In `GeoShapeUtil.tsx` and `NoteShapeUtil.tsx`, the `style` prop with `transform: scale(...)` is always passed to `RichTextLabel`, regardless of whether dynamic sizing is active. During rotation, this creates an extra composite transform layer the browser must recalculate on every frame.

## Proposed fix

Conditionally apply the `style` attribute only when dynamic size is enabled (i.e., when `scale !== 1`). When scale is 1 (the default), skip the transform entirely to avoid the performance hit.

For example, in `GeoShapeUtil.tsx` (~line 370) and `NoteShapeUtil.tsx` (~line 405), only pass the style prop when the shape's scale differs from 1. When scale is 1, use the previous approach of multiplying font size and padding directly.

## Affected files

- `packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx` (lines ~360-375)
- `packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx` (lines ~393-410)
- Possibly `packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx` (lines ~849-862, from follow-up #8451)

mimecuvalo:
In order to fix a rotation performance regression introduced by #8378 and #8451, this PR conditionally applies CSS `transform: scale()` on shape label containers only when dynamic size is active (`scale !== 1`). Closes #8562.

PR #8378 changed geo and note shapes to use CSS `transform: scale()` on label containers instead of multiplying `fontSize` and `padding` directly, fixing shadow artifacts and oversized carets at high zoom in dynamic size mode. However, this transform was applied unconditionally — even when `scale === 1` (the default), which forces an unnecessary compositing layer on every shape label, causing a measurable rotation performance regression.

The fix skips the `style` prop entirely when `scale === 1`, avoiding the extra composite layer during rotation while preserving the shadow/caret fix when dynamic sizing is active.

### Change type

- [x] `bugfix`

### Test plan

1. Create a canvas with many geo and note shapes (no dynamic sizing)
2. Select all and rotate — rotation should be smooth without the compositing overhead
3. Enable dynamic size on some shapes (scale !== 1) and verify labels still render correctly at high zoom (no shadow artifacts, no oversized carets)
4. Add arrow labels with dynamic size and verify positioning and scaling still work

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix rotation performance regression caused by unconditional CSS transforms on shape labels

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +21 / -13  |

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -858,7 +858,7 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 						padding={0}
 						showTextOutline={this.options.showTextOutline}
 						style={{
-							transform: `translate(${labelPosition.box.center.x}px, ${labelPosition.box.center.y}px) scale(${shape.props.scale})`,
+							transform: `translate(${labelPosition.box.center.x}px, ${labelPosition.box.center.y}px)${shape.props.scale !== 1 ? ` scale(${shape.props.scale})` : ''}`,
 						}}
 					/>
 				)}
--- a/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/geo/GeoShapeUtil.tsx
@@ -367,12 +367,16 @@ export class GeoShapeUtil extends BaseBoxShapeUtil<TLGeoShape> {
 							labelColor={dv.labelColor}
 							wrap
 							showTextOutline={this.options.showTextOutline}
-							style={{
-								transform: `scale(${shape.props.scale})`,
-								transformOrigin: 'top left',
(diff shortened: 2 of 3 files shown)


steveruizok:
Nice one!

PR #6276, opened 2025-06-18
image: make sure replace respects mime types, file size, dimensions

mimecuvalo:
for the new image replace functionality, this reworks the code so that it has the proper mime type/file size/dimension checks.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- image: make sure replace respects mime types, file size, dimensions

PR #8579, opened 2026-04-16
fix: export of defaultHandleExternalFileReplaceContent

angrycaptain19:
This function was added in PR #6276 but was never added to the package's public exports, despite being listed in the documentation alongside the other defaultHandleExternal* functions which are all exported.

### API changes
- Adds `defaultHandleExternalFileReplaceContent` to exports


mimecuvalo (approved):
ty!

PR #8585, opened 2026-04-17
docs(agent): update agent starter kit doc to reflect readme

angrycaptain19:
Updates the agent starter kit docs to reflect the readme changes @max-drake made with the fairies backport

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`
- [x] `docs`

### DIFF
--- a/apps/docs/content/starter-kits/agent.mdx
+++ b/apps/docs/content/starter-kits/agent.mdx
@@ -16,8 +16,7 @@ keywords:
 readability: 7
 voice: 6
 completeness: 9
-accuracy: 7
-notes: "Very comprehensive but long. Notion link at line 422 is an error. 'Use cases' should be lowercase. Bolded bullet format. Good code examples."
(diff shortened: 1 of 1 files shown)


max-drake:
is this  supposed to be the exact same text as the templates/agent/readme.md? because aside from the frontmatter, there are some differences across the doc. maybe I'm misunderstanding but I thought what was on dotdev was supposed to be the exact same as the readme

angrycaptain19:
Hmm yeah it looks like Claude tried to balance into the old content too
much. I'll shift the approach

max-drake:
don't we have a script that does this? I don't think this was being done manually before

angrycaptain19:
If we do it hasn't been working for a bit. Will investigate more Monday

max-drake (approved):
lgtm following our convo from earlier

PR #8587, opened 2026-04-17
ci(dotcom): deploy MCP app after production dotcom deploy

max-drake:
In order to keep the public MCP worker (`tldraw-mcp-app`) on the same commit and SDK build as production dotcom, this PR adds a workflow step that builds and deploys `apps/mcp-app` after the main dotcom deploy succeeds. It runs only on pushes to `production` so staging (`main`) does not overwrite the shared worker. The Vite widget receives `VITE_TLDRAW_LICENSE_KEY` from the `MCP_APP_TLDRAW_LICENSE_KEY` repository secret at build time.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Confirm `MCP_APP_TLDRAW_LICENSE_KEY` is set for the `deploy-production` environment (or repository) alongside existing Cloudflare secrets.
2. After merge, trigger or wait for a `production` branch deploy and confirm the **Deploy MCP app** step completes and the worker serves the new bundle.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Config/tooling | +9 / -0    |


mimecuvalo:
Mm, are we sure that we want to tie the MCP release to the dotcom release? 🤔 
I don't think they should be coupled in this way really. If we want to keep a regular cadence, I would agree, but a separate deploy is probably better (in case there are dotcom issues and we need to push a bunch / rollback / etc.)

max-drake:
> Mm, are we sure that we want to tie the MCP release to the dotcom release? 🤔 I don't think they should be couple in this way really. If we want to keep a regular cadence, I would agree, but a separate deploy is probably better (in case there are dotcom issues and we need to push a bunch / rollback / etc.)

do you think we should do something like just making it part of the dotcom deploy ritual? after you press the dotcom deploy button you press the mcp app deploy button? this is what steve recommended we do for the record, not sure what the best path forward is

mimecuvalo (approved):
LG

PR #8594, opened 2026-04-20
fix(ci): skip S3 upload when Playwright report is missing

steveruizok:
In order to stop noisy red ❌ on the `Upload S3` step whenever Playwright bails out before producing a report (build/setup failure, environment error, etc.), this PR gates the `shallwefootball/s3-upload-action@master` step on `hashFiles('…/playwright-report/index.html') != ''` so it's skipped cleanly when the report entrypoint doesn't exist. Previously the action would try to `readdirSync` the missing directory and crash with `ENOENT`.

The `Log report to summary` step is also now gated on a successful S3 upload (`steps.s3.outputs.object_key != ''`) so we no longer emit a broken `https://playwright.tldraw.xyz//index.html` summary URL when the upload was skipped.

Applied to all three Playwright workflows: `playwright-dotcom.yml`, `playwright-examples.yml`, and `playwright-perf.yml`.

### Change type

- [x] `other`

### Test plan

Cannot be manually tested without triggering a failing Playwright run on CI — the fix is observable the next time a run fails before producing a report, where the `Upload S3` and `Log report to summary` steps should both be skipped instead of erroring / printing a broken URL.

### Code changes

| Config/tooling  | +6 / -6    |

### DIFF
--- a/.github/workflows/playwright-dotcom.yml
+++ b/.github/workflows/playwright-dotcom.yml
@@ -110,7 +110,7 @@ jobs:
         # only upload if we have AWS credentials:
         env:
           HAS_AWS_KEY: ${{ secrets.AWS_S3_SECRET_KEY && '1' || '' }}
-        if: always() && env.HAS_AWS_KEY == '1'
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
👍

PR #8595, opened 2026-04-20
cleanup: rm --font-body which is unused

mimecuvalo:
just a little nit i noticed. @steveruizok was this used somewhere before?

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Release notes

- cleanup: rm `--font-body` which is unused


steveruizok:
nope, looks like dead code

PR #8598, opened 2026-04-20
fix(scripts): prevent secret leaks in deploy Discord messages

MitjaBezensek:
In order to prevent secrets from leaking into Discord error notifications when deploy commands fail, this PR adds value-based masking to the Discord helper — replacing any occurrence of known secret values before sending messages, similar to how GitHub Actions masks secrets.

The existing regex-based sanitization (`KEY=VALUE` pattern matching) fails when a value contains spaces. For example, `OTEL_EXPORTER_OTLP_HEADERS=Authorization=Basic ...` gets partially sanitized but the base64 token after the space leaks as an orphaned string in the Discord message. This is a whack-a-mole problem — we've fixed the regex multiple times but keep hitting edge cases.

**How it works:**
- `Discord` constructor accepts `secretValues: string[]` — all `Object.values(env)` from `makeEnv`
- Before sending, `sanitizeVariables` replaces every known secret value (≥8 chars, longest-first) with an internal null-byte placeholder
- Existing regex patterns still run as a fallback for secrets not in the env object
- Placeholder is swapped to `` `***` `` at the end, preventing regex patterns from re-matching masked output

Also fixes the catch blocks in hotfix scripts which previously created Discord instances without any secret masking — the exact leak vector this change is meant to close.

### Change type

- [x] `bugfix`

### Test plan

1. Deploy script errors containing secret values are masked before reaching Discord
2. Normal progress messages may over-redact (acceptable tradeoff vs leaking)
3. Regex fallback still catches patterns not covered by value masking

### Code changes

| Config/tooling | +32 / -5   |

### DIFF
--- a/internal/scripts/deploy-analytics.ts
+++ b/internal/scripts/deploy-analytics.ts
@@ -31,6 +31,7 @@ const discord = new Discord({
 	shouldNotify: env.TLDRAW_ENV === 'production',
 	totalSteps: 2,
 	messagePrefix: '[ANALYTICS]',
+	secretValues: Object.values(env),
 })
 
 const { previewId, sha } = getDeployInfo()
--- a/internal/scripts/deploy-bemo.ts
+++ b/internal/scripts/deploy-bemo.ts
@@ -33,6 +33,7 @@ const discord = new Discord({
 	webhookUrl: env.DISCORD_DEPLOY_WEBHOOK_URL,
 	shouldNotify: env.TLDRAW_ENV === 'production',
 	totalSteps: 3,
+	secretValues: Object.values(env),
 })
 
 const { previewId, sha } = getDeployInfo()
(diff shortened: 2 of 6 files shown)


mimecuvalo (commented):
before i approve, can you help my understand the history here?
there were also these PRs
- https://github.com/tldraw/tldraw/pull/8075
- https://github.com/tldraw/tldraw/pull/8287

these didn't work / help?

MitjaBezensek:
> before i approve, can you help my understand the history here? there were also these PRs
> 
> * #8075
> * #8287

Different CLI have a different way of setting the env variables:

Wrangler (--var KEY:VALUE):
Original code, pre-existing regex.

flyctl (KEY=VALUE KEY2="VALUE WITH SPACES"):
[#8075](https://github.com/tldraw/tldraw/issues/8075) catches single-word values. 

Vercel (--token TOKEN_VALUE):
[#8287 ](https://github.com/tldraw/tldraw/issues/8287) catches --token arguments.

Connection strings (postgres://user:pass@host/db):
[#8075 ](https://github.com/tldraw/tldraw/issues/8075) catches URL-scheme patterns.

This PR:
Catches everything above plus any format we haven't thought of, by masking the literal values regardless of surrounding syntax. So it also catches things like env vars with spaces like (the `Authorization=Basic ...` that is used for `ZERO_OTEL_EXPORTER_OTLP_HEADERS`).

mimecuvalo (approved):

danieljamesross (approved):

PR #8600, opened 2026-04-20
perf(editor): drop intermediate array allocations in hit-test hot paths

mimecuvalo:
In order to cut per-pointer-event allocations in the core hit-test paths, this PR replaces `.filter().find()`, `.filter().reverse().find()`, and `.flatMap(EMPTY_ARRAY)` chains in `Group2d` and `Editor` hit-testing with plain backwards `for` loops that inline the filter predicates.

The motivation came out of a question about whether the iterator-helpers proposal (`Iterator.from(...)`) would help in the core SDK. The short answer: not meaningfully — iterator helpers ship as a runtime feature (not syntax), and with `browserslist: ["defaults"]` and no `core-js` polyfill pipeline, we can't use them yet without a crash risk on older Safari. The _actual_ win in the hot paths was always "stop materializing the filtered array at all." That's what this PR does.

### What changed

- `Group2d.hitTestPoint` / `hitTestLineSegment`: were building a temporary `this.children.filter(...)` array just to `.find()` on it; now a single `for` loop with early-return. This is hot on every pointer event whenever a group is on the page.
- `Group2d.intersectLineSegment` / `intersectCircle` / `intersectPolygon` / `intersectPolyline`: were using `.flatMap` with an `EMPTY_ARRAY` sentinel for excluded children; now a single `for` loop pushing into a shared accumulator. Drops the per-child array and the final `flatMap` allocation. Used during snapping / arrow binding drags.
- `Group2d.getVertices` / `getBoundsVertices`: same pattern, same rewrite.
- `Editor.getSelectedShapeAtPoint`: `.filter().reverse().find()` → backwards `for` loop over the sorted list with inline filter. Drops two intermediate arrays.
- `Editor.getShapeAtPoint`: the body already had a reverse `for` over a pre-filtered array. Inlined the filter predicates into the existing loop as `continue` guards so we don't allocate `shapesToCheck` at all.
- `Editor.getShapesAtPoint`: `.filter().reverse()` → backwards `for` that pushes directly into the result in reverse z-order. One allocation instead of two.
- `deriveShapeIdsInCurrentPage.fromScratch`: `new Set([...shapesIndex.get()].filter(...))` → direct `for…of` into a pre-constructed `Set`. Drops the spread array and the filtered intermediate. Cold path (only runs on page switch / init / history reset) but trivial to fix.

No public API changes. Same return types, same ordering semantics. The `EMPTY_ARRAY` import from `@tldraw/state` was removed from `Group2d.ts` since it's no longer used.

### Change type

- [x] `improvement`

### Test plan

1. Existing hit-test behavior should be unchanged — click/hover/lasso/erase on shapes, groups, and nested groups in the editor and verify selection works as before.
2. Arrow binding and snap behavior with groups should be unchanged.

- [x] Unit tests

Full editor + tldraw test suites pass (800 + 2257 tests, 0 failures).

### Release notes

- Performance: reduced per-pointer-event allocations in `Group2d` and `Editor` hit-testing.


mimecuvalo:
@steveruizok marking you as reviewer

mimecuvalo:
@steveruizok similar vein to https://github.com/tldraw/tldraw/pull/9150 - can you review this one? 🙏

mimecuvalo:
marking @MitjaBezensek as reviewer 👍

Issue #8586, opened 2026-04-17
Crash and memory usage spike on iOS when interacting with shapes while at a low zoom level


zacwood9:
I noticed a reproducible crash in [my company's](https://constructable.ai) iOS app related to zooming on, or moving around, shapes and was able to reproduce this on tldraw.com as well. See attached video. It seems to trigger more the lower the zoom level.

Video taken on iPhone 17, iOS 26.2.1. I was not able to reproduce this on Android. I was able to reproduce as far back as v4.3, but I didn't try any further back than that.

This is making our app borderline unusable because our users are often on very low zoom levels looking at and marking up large construction drawings. Thank you for any help!

https://github.com/user-attachments/assets/93d304c8-cac9-4281-b3cd-1b2c73995c72

Another "smoking gun" as our AI friends would say :)
We can measure memory usage spiking drastically when the shapes are being dragged around, spiking until the process hits the memory limit ~1.5gb. See the yellow line on the graph.

<img width="1199" height="668" alt="Image" src="https://github.com/user-attachments/assets/a595b60f-083c-4660-8b89-61e4952e1a64" />

mimecuvalo:
Thanks for the report, we'll investigate on our end 👍

Issue #8440, opened 2026-04-03
Module-level state causes memory leaks after editor unmount


MitjaBezensek:
Several module-level caches, globals, and debounced/throttled functions retain `Editor`, `Store`, or `TLLocalSyncClient` references after the editor unmounts, preventing garbage collection. This was originally reported in #8430 for the `updateHoveredShapeId` throttle, but the pattern appears in multiple other places.

#8439 fixes the `updateHoveredShapeId` case. This issue tracks the remaining cases.

### 1. `window.tlsync` retains `TLLocalSyncClient` (and `Store`) — HIGH

**File:** `packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts:105`

The constructor unconditionally sets `window.tlsync = this`. The `close()` method (line 261) runs disposables but never clears `window.tlsync`. After unmount, the closed client and its entire retained graph (Store, LocalIndexedDb, BroadcastChannel, subscriptions) cannot be GC'd.

```ts
// constructor
if (typeof window !== 'undefined') {
    ;(window as any).tlsync = this
}
```

**Fix:** Clear `window.tlsync` in `close()`, or guard it behind `process.env.NODE_ENV !== 'production'`.

### 2. `window.tlsync` retains `TLSyncClient` (and `Store`) — HIGH

**File:** `packages/sync-core/src/lib/TLSyncClient.ts:538`

Same pattern as above in the multiplayer sync client. `TLSyncClient.close()` (line 824) does not clear `window.tlsync`.

**Fix:** Same as above.

### 3. Module-level `debounce` in bookmarks retains `Editor` — HIGH

**File:** `packages/tldraw/src/lib/shapes/bookmark/bookmarks.ts:87`

```ts
const createBookmarkAssetOnUrlChange = debounce(async (editor: Editor, shape: TLBookmarkShape) => {
    if (editor.isDisposed) return
    // ...
}, 500)
```

This is the exact same class of bug as the `updateHoveredShapeId` throttle: a module-level `debounce()` that receives `editor` as an argument. The debounce stores `latestArgs` internally, retaining the Editor after unmount until the debounce fires (500ms). Additionally, `cancel()` in the debounce utility does not release retained args (see case 6).

**Fix:** Use per-editor debounced instances via `WeakMap<Editor, ...>`, same pattern as the `updateHoveredShapeId` fix.

### 4. `debounce.cancel()` does not release retained arguments — MEDIUM

[...]

PR #8603, opened 2026-04-21
feat(perf): add memory usage tracking to RUM telemetry

MitjaBezensek:
In order to detect memory leaks in production (#8586, #8440), this PR extends the existing RUM telemetry in dotcom with periodic heap snapshots and richer slicing dimensions on all rum events.

**Memory sampling** (Chrome-only via `performance.memory`, no-ops elsewhere):
- 60s interval + samples on `visibilitychange` (hidden) and page switches
- Tracks heap size, shape/page counts, page change count
- Gated behind existing `rum_enabled` flag

**New properties on all rum events** (interaction, camera, memory):
- `release` — deploy tag from `sentryReleaseName` (`local` in dev, `{env}-{sha}` in prod) for correlating regressions to specific deploys

OS/platform slicing uses PostHog's built-in `$os` property instead of a custom field — it already provides Mac OS X / Windows / Linux / Chrome OS / iOS / Android breakdown.

Relates to #8586, #8440

### Change type

- [x] `improvement`

### Test plan

1. Run `yarn dev-app`, open Chrome
2. Enable `rum_enabled` flag (or temporarily set default to true in FeatureFlagPoller)
3. Verify PostHog captures `rum` events with `type: 'memory'` and `release: 'local'`
4. Verify interaction and camera events also include `release`
5. Switch pages → verify immediate `page_change` sample
6. Background tab → verify `visibility_hidden` sample
7. Open Firefox → verify no errors (graceful no-op)

- [ ] Unit tests
- [ ] End to end tests

### DIFF
--- a/apps/dotcom/client/src/hooks/usePerformanceTracking.ts
+++ b/apps/dotcom/client/src/hooks/usePerformanceTracking.ts
@@ -5,6 +5,7 @@ import {
 	TLInteractionEndPerfEvent,
 	TLPerfFrameTimeStats,
 } from 'tldraw'
+import { sentryReleaseName } from '../../sentry-release-name'
 import { fetchFeatureFlags } from '../tla/utils/FeatureFlagPoller'
 import { trackEvent } from '../utils/analytics'
 
@@ -49,6 +50,7 @@ function commonStats(event: TLPerfFrameTimeStats & { shapeCount: number; zoomLev
 		has_loaf: (loafs?.length ?? 0) > 0,
 		loaf_count: loafs?.length ?? 0,
 		coarse_platform: (coarsePlatform ??= getCoarsePlatform()),
+		release: sentryReleaseName,
 	}
 }
 
@@ -130,6 +132,7 @@ export function usePerformanceTracking() {
 							device_memory_gb: sampleIndex === 0 ? deviceMemoryGb : null,
 							coarse_platform: (coarsePlatform ??= getCoarsePlatform()),
 							ab_indicators: abIndicators,
+							release: sentryReleaseName,
 						}
 						trackEvent('rum', event)
 						sampleIndex++
--- a/apps/dotcom/client/src/hooks/usePerformanceTracking.ts
+++ b/apps/dotcom/client/src/hooks/usePerformanceTracking.ts
@@ -11,14 +11,6 @@ import { trackEvent } from '../utils/analytics'
 
 const r2 = (n: number) => Math.round(n * 100) / 100
 
-function getCoarsePlatform(): 'ios' | 'android' | 'desktop' {
-	if (typeof navigator === 'undefined') return 'desktop'
-	const ua = navigator.userAgent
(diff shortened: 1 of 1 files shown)


apps/dotcom/client/src/hooks/usePerformanceTracking.ts
```diff
@@ -5,11 +5,23 @@ import {
...
 } from 'tldraw'
+import { sentryReleaseName } from '../../sentry-release-name'
 import { fetchFeatureFlags } from '../tla/utils/FeatureFlagPoller'
 import { trackEvent } from '../utils/analytics'
 
 const r2 = (n: number) => Math.round(n * 100) / 100
 
+function getZoomBucket(zoom: number): string {
+	if (zoom <= 0.15) return 'lte_0.15'
+	if (zoom <= 0.35) return '0.15_0.35'
```
mimecuvalo: why these specific numbers? let's make these more semantic in case they change in the future

apps/dotcom/client/src/hooks/usePerformanceTracking.ts
```diff
@@ -56,27 +70,104 @@ export function usePerformanceTracking() {
...
+				unsubInteraction = editor.performance.on('interaction-end', (event) =>
+					handleInteractionEnd(event, abIndicators)
+				)
+				unsubCamera = editor.performance.on('camera-end', (event) =>
+					handleCameraEnd(event, abIndicators)
+				)
+
+				editorMountCount++
+				// Memory sampling (Chrome-only via performance.memory)
+				if ((performance as any).memory) {
```
mimecuvalo, line 86: we sure we want to use this? i think this api might go away completely: https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory

take a look at https://web.dev/articles/monitor-total-page-memory-usage as well, can we use `measureUserAgentSpecificMemory`?

apps/dotcom/client/src/hooks/usePerformanceTracking.ts
MitjaBezensek: Think I'll just remove these, we can defined them in posthog anyway.

apps/dotcom/client/src/hooks/usePerformanceTracking.ts
MitjaBezensek: My understanding is that we'd have to add certain headers for that to work and it might not be fully compatible with our app. You might know more on this though?

>  measureUserAgentSpecificMemory():
>  - Requires Cross-Origin-Opener-Policy: same-origin + Cross-Origin-Embedder-Policy: require-corp
> - Dotcom does not set these headers
>  - Adding them is non-trivial — breaks cross-origin iframes, third-party scripts, and any resource without Cross-Origin-Resource-Policy headers
 
CleanShot 2026-04-21 at 13 43 20@2x

mimecuvalo (approved):
👍

PR #8604, opened 2026-04-21
fix(memory): clean up module-level state leaks after editor unmount

### LINKED ISSUE #8440: Module-level state causes memory leaks after editor unmount

MitjaBezensek:
In order to prevent `Editor`, `Store`, and sync client instances from being retained after unmount, this PR cleans up several module-level caches, globals, and debounced functions that hold references. Particularly impactful for apps that mount/unmount editors frequently (iOS WebViews hit memory limits). Closes #8440.

**`debounce.cancel()`** — sets `state = undefined` so `latestArgs` (and any captured Editor) are released on cancel. Previously only cleared the timeout.

**`window.tlsync`** — cleared in `close()` for both `TLLocalSyncClient` and `TLSyncClient`. Guarded with `=== this` to avoid clearing a newer client's reference.

**Bookmark debounce** — replaced module-level `debounce(editor, shape)` with per-editor instances via `WeakMap<Editor, ...>` (same pattern as the `updateHoveredShapeId` fix in #8439). Debouncer is GC'd with the editor.

**`LruCache`** — new `@tldraw/utils` class (~20 LOC) using `Map` insertion-order for O(1) LRU eviction. `fetchCache` (cap 200) and `alphaCache` (cap 500) now use `LruCache` instead of unbounded `Map`s, preventing unbounded growth while keeping hot entries warm.

### Change type

- [x] `bugfix`

### Test plan

1. Mount/unmount editor multiple times
2. Take heap snapshots, confirm Editor instances are GC'd
3. Verify bookmark URL unfurling still works after editor remount
4. Verify image alpha hit-testing works with many unique images

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed memory leaks from module-level state retaining editor references after unmount
- Added `LruCache` utility class for bounded caches with least-recently-used eviction

### API changes

- Added `LruCache<K, V>` class to `@tldraw/utils` with `get`, `set`, `has`, and `size`

### Code changes

| Core code       | +60 / -8   |
| Tests           | +89 / -0   |
| Automated files | +13 / -0   |

### DIFF
--- a/packages/editor/src/lib/exports/fetchCache.ts
+++ b/packages/editor/src/lib/exports/fetchCache.ts
@@ -1,13 +1,15 @@
 import { FileHelpers, assert, fetch } from '@tldraw/utils'
 
-// TODO(alex): currently, this cache will grow unbounded. we should come up with a better strategy
-// for clearing items from the cache over time.
+// Best-effort memory bound; full clear on overflow trades brief re-fetch cost for simplicity.
+const MAX_FETCH_CACHE_SIZE = 200
+
 export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
 	const cache = new Map<string, Promise<T | null>>()
 
 	return async function fetchCached(url: string): Promise<T | null> {
 		const existing = cache.get(url)
 		if (existing) return existing
+		if (cache.size >= MAX_FETCH_CACHE_SIZE) cache.clear()
 
 		const promise = (async () => {
 			try {
--- a/packages/editor/src/lib/exports/fetchCache.ts
+++ b/packages/editor/src/lib/exports/fetchCache.ts
@@ -1,7 +1,7 @@
 import { FileHelpers, LruCache, assert, fetch } from '@tldraw/utils'
 
 export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
-	const cache = new LruCache<string, Promise<T | null>>(200)
+	const cache = new LruCache<string, Promise<T | null>>(100)
 
 	return async function fetchCached(url: string): Promise<T | null> {
 		const existing = cache.get(url)
--- a/packages/editor/src/lib/exports/fetchCache.ts
+++ b/packages/editor/src/lib/exports/fetchCache.ts
@@ -1,15 +1,11 @@
-import { FileHelpers, assert, fetch } from '@tldraw/utils'
-
-// Best-effort memory bound; full clear on overflow trades brief re-fetch cost for simplicity.
-const MAX_FETCH_CACHE_SIZE = 200
+import { FileHelpers, LruCache, assert, fetch } from '@tldraw/utils'
 
 export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
-	const cache = new Map<string, Promise<T | null>>()
+	const cache = new LruCache<string, Promise<T | null>>(200)
 
 	return async function fetchCached(url: string): Promise<T | null> {
 		const existing = cache.get(url)
 		if (existing) return existing
-		if (cache.size >= MAX_FETCH_CACHE_SIZE) cache.clear()
 
 		const promise = (async () => {
 			try {
--- a/packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts
+++ b/packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts
@@ -262,6 +262,9 @@ export class TLLocalSyncClient {
 		this.debug('closing')
 		this.didDispose = true
 		this.disposables.forEach((d) => d())
+		if (typeof window !== 'undefined' && (window as any).tlsync === this) {
+			delete (window as any).tlsync
+		}
 	}
 
 	private isPersisting = false
--- a/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
+++ b/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
@@ -65,6 +65,9 @@ export function isImagePointTransparent(
 
 const MAX_SIZE = 256
 
+// Best-effort memory bound; full clear on overflow trades brief re-decode cost for simplicity.
+const MAX_ALPHA_CACHE_SIZE = 500
+
 const alphaCache = new Map<string, AlphaData>()
 const pending = new Set<string>()
 let offscreenCanvas: OffscreenCanvas | null = null
@@ -139,6 +142,7 @@ export function preloadAlphaData(url: string, cacheKey?: string): void {
 			ctx.drawImage(img, 0, 0, w, h)
 		}
 
+		if (alphaCache.size >= MAX_ALPHA_CACHE_SIZE) alphaCache.clear()
 		alphaCache.set(key, { width: w, height: h, alphas: extractAlphas(ctx, w, h) })
 	}
 	img.onerror = () => {
--- a/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
+++ b/packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
@@ -1,4 +1,4 @@
-import { Image, VecLike } from '@tldraw/editor'
+import { Image, LruCache, VecLike } from '@tldraw/editor'
 
 /** Mime types of image formats that support transparency / alpha channel. */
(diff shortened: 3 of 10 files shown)


packages/utils/src/lib/LruCache.ts
```diff
@@ -0,0 +1,32 @@
+/** Simple LRU cache backed by a Map's insertion-order iteration. @public */
```
mimecuvalo, line 1: huh i'm surpised we don't have one of these yet

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
```diff
@@ -65,7 +65,7 @@ export function isImagePointTransparent(
 
 const MAX_SIZE = 256
 
-const alphaCache = new Map<string, AlphaData>()
+const alphaCache = new LruCache<string, AlphaData>(500)
```
mimecuvalo: 500 seems high?

packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts
```diff
@@ -262,6 +262,9 @@ export class TLLocalSyncClient {
 		this.debug('closing')
 		this.didDispose = true
 		this.disposables.forEach((d) => d())
+		if (typeof window !== 'undefined' && (window as any).tlsync === this) {
+			delete (window as any).tlsync
```
mimecuvalo, line 266: huh TIL that we set this on the window obj

packages/editor/src/lib/exports/fetchCache.ts
```diff
@@ -1,9 +1,7 @@
-import { FileHelpers, assert, fetch } from '@tldraw/utils'
+import { FileHelpers, LruCache, assert, fetch } from '@tldraw/utils'
 
-// TODO(alex): currently, this cache will grow unbounded. we should come up with a better strategy
-// for clearing items from the cache over time.
 export function fetchCache<T>(cb: (response: Response) => Promise<T>, init?: RequestInit) {
-	const cache = new Map<string, Promise<T | null>>()
+	const cache = new LruCache<string, Promise<T | null>>(200)
```
mimecuvalo: 200 seems maybe high too, i think 100 is prbly fine

MitjaBezensek: Adjusted.

packages/tldraw/src/lib/shapes/image/ImageAlphaCache.ts
MitjaBezensek: Adjusted this as well.

packages/utils/src/lib/LruCache.ts
MitjaBezensek: Yeah 🤷‍♂️

packages/utils/src/lib/LruCache.ts
```diff
@@ -0,0 +1,32 @@
...
+export class LruCache<K, V> {
+	private map = new Map<K, V>()
+	constructor(private maxSize: number) {}
+
+	get(key: K): V | undefined {
+		const value = this.map.get(key)
+		if (value !== undefined) {
```
mimecuvalo: value _could_ be undefined if we wanted that in the cache (since this is a generic cache)


```suggestion
    if (!this.map.has(key)) return undefined
    const value = this.map.get(key)!
    this.map.delete(key)
    this.map.set(key, value)
    return value
```

packages/utils/src/lib/LruCache.ts
```diff
@@ -0,0 +1,32 @@
...
+			// Evict oldest entry
+			this.map.delete(this.map.keys().next().value!)
+		}
+	}
+
+	has(key: K): boolean {
+		return this.map.has(key)
+	}
+
+	getSize(): number {
```
mimecuvalo: AtomMap and AtomSet use `get size() {` - let's do the same here prbly

mimecuvalo (approved):
pre-approved 👍

PR #8607, opened 2026-04-21
docs: adding sections for mermaid and driver packages

angrycaptain19:
5.0 is going to introduce 2 new packages for mermaid and the driver package. This adds in docs sections similar to what we have for sync.


kaneel (approved):
LGTM on the mermaid bits

steveruizok (approved):

PR #7123, opened 2025-11-14
feat(sync): abstract out sync storage and add TLSyncStorage API

ds300:
This PR does a pretty hefty refactor to our sync storage to prepare us to be able to use sqlite as a storage layer (which I'm gonna spike in a stacked PR to make sure it'll work) and also to just make the sync library better. This really cleans up a lot of scalability and customiztion issues.

todo:
- [x] Look into making it async (i have a strong supposition that this will be requested, to work with dbs other than sqlite, and even with sqlite on node the blocking might be a bottleneck for scalability)
- [x] Make the change backwards-compatible. For ease of refactoring a bunch of deprecated things have been simply deleted at time of PR opening, and some things have been changed in backwards-incompatible ways. I need to resolve those and deprecate what needs to be deprecated. Maybe even move all the new stuff into separate classes like TLSocketRoom2 and TLSyncRoom2, especially if async seems doable.
- [x] test

#### Pluggable smart storage layer for `TLSocketRoom`

Before now, `TLSocketRoom`—the server-side class responsible for managing document state and websocket connections for [tldraw sync](https://tldraw.dev/docs/sync)—stored the entire document in memory and only emitted coarse grained 'document changed' events.

The new `TLSyncStorage` API allows servers to make use of _any_ synchronous storage mechanism, primarily the SQLite implementations available in modern JavaScript server environments (Node, Bun, Deno, Cloudflare durable objects). This means lowered server-side memory usage and faster start up time.

<details>
<summary>Migration guide</summary>

The `initialSnapshot` and `onDataChange` options on `TLSocketRoom` are now deprecated. Instead, use the new `storage` option with `InMemorySyncStorage` if you want to keep the same behavior as before.

##### Before

```ts
import { TLSocketRoom } from '@tldraw/sync-core'

const room = new TLSocketRoom({
  initialSnapshot: existingSnapshot,
  onDataChange: () => {
    persistSnapshot(room.getCurrentSnapshot())
  },
})
```

#### After

```ts
import { InMemorySyncStorage, TLSocketRoom } from '@tldraw/sync-core'

// Create storage with an existing snapshot, or `undefined` to create an empty document.
const storage = new InMemorySyncStorage({
  snapshot: existingSnapshot ?? undefined,
})

// Listen for changes before creating the room to capture migration changes
storage.onChange(() => {
  persistSnapshot(storage.getSnapshot())
})

// Pass storage to the room
const room = new TLSocketRoom({ storage })
```

#### Key changes

[...]

PR #8631, opened 2026-04-24
feat(sync): persist active users' presence in room snapshots

steveruizok:
In order to preserve which users were active on a file at persist time, this PR attaches recently-active presence records to the snapshot returned by `TLSocketRoom.getCurrentSnapshot`. The R2 + Pierre persist path in `TLFileDurableObject` now pulls the snapshot through the room so presence rides along. Presence is archival-only — `loadSnapshotIntoStorage` continues to ignore it, and the user-facing `.tldr` export path still uses `storage.getSnapshot()` directly so link visitors don't get cursor/name data.

Filtering: only presence records with a `lastActivityTimestamp` within the last 5 minutes are included (`PRESENCE_SNAPSHOT_ACTIVE_WINDOW_MS`). The check is duck-typed on the field so the sync-core stays schema-generic — records without that field are skipped.

### Change type

- [x] `improvement`

### Test plan

1. Open a shared dotcom file with another client connected.
2. Wait for a persist cycle (or force one) and inspect the R2 snapshot — it should have a `presence` array containing the active clients' presence records.
3. Disconnect all clients, wait > 5 min, re-persist — the `presence` array should be empty/absent.
4. Download the same file via the `.tldr` export endpoint — the exported file should **not** contain a `presence` field.
5. Load an older snapshot (one without `presence`) and confirm the room loads normally.

- [x] Unit tests

### API changes

- Added optional `presence?: UnknownRecord[]` field to the public `RoomSnapshot` interface. Populated by `TLSocketRoom.getCurrentSnapshot` for archival; ignored on load.

### Release notes

- The sync server now records which users were active when a snapshot is taken, so stored snapshots preserve presence for archival/analytics. Presence is not restored when a snapshot is loaded.

### Code changes

| Core code       | +24 / -3   |
| Apps            | +2 / -1    |

### DIFF
--- a/packages/sync-core/src/lib/TLSocketRoom.ts
+++ b/packages/sync-core/src/lib/TLSocketRoom.ts
@@ -15,6 +15,12 @@ import {
 } from './TLSyncStorage'
 
 /**
+ * Presence records seen more recently than this are captured into the room
+ * snapshot for archival. Presence is never restored on load.
+ */
+const PRESENCE_SNAPSHOT_ACTIVE_WINDOW_MS = 5 * 60 * 1000
+
+/**
  * Strip potentially large fields from a tldraw instance_presence record so the
  * snapshot stays small when stored in WebSocket attachments (e.g. for hibernation).
  * Keeps cursor, selection, page, and user identity; clears scribbles, chatMessage, brush.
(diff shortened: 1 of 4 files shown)


packages/sync-core/src/lib/TLSocketRoom.ts
```diff
@@ -697,10 +703,20 @@ export class TLSocketRoom<R extends UnknownRecord = UnknownRecord, SessionMeta =
 	 * ```
 	 */
 	getCurrentSnapshot() {
```
max-drake, line 705: ~this fn is was deprecated by #7123, should we reinstate it if we're going to be using it in this new way?~

addressed in my [other comment](https://github.com/tldraw/tldraw/pull/8631#issuecomment-4633445080)

max-drake:
I think adding this new functionality to the already-deprecated `TLSocketRoom.getCurrentSnapshot()` method is confusing, and muddies the difference between that `SQLiteSyncStorage.getSnapshot()`

I think we should add a new more clearly-documented method to `TLSocketRoom` called like `getSnapshotWithPresence()` and maybe add a new `RoomSnapshotWithPresence` type that it can return in order to make it clear when/how to use this new method

I'm worried about someone accidentally using `TLSocketRoom.getCurrentSnapshot()` in the future without realizing that we really don't want presence in most cases (except for the dotcom snapshots, which is what this pr is for)

PR #8651, opened 2026-04-27
feat: hold accel to temporarily erase from draw / highlight

### LINKED ISSUE #8650: Hold accel to temporarily erase while using the draw tool
## Problem

When drawing, you often want to quickly fix small mistakes without breaking flow. Today this requires switching tools (D ↔ E, or via the toolbar), which is disruptive — especially mid-sketch. There's no quick "scratch out" gesture.

## Proposed solution

While the draw tool is active and **before** the user starts a stroke (i.e. in the draw tool's `Idle` state), holding the accel key (Cmd on macOS, Ctrl elsewhere) should temporarily put the editor into the eraser state. Releasing accel restores the draw tool.

The toolbar / UI should continue to show the draw tool as active for the duration. Use `setCurrentToolIdMask('draw')` so the eraser is technically running but the user perceives "I'm still in draw, just erasing for a moment."

Behavior:

- Active only when the draw tool is in `Idle` (not during an in-progress stroke). Accel during a draw stroke must keep its existing meaning.
- On `accel down` → switch to eraser, mask current tool id as `draw`.
- On `accel up` → switch back to draw, clear the mask.
- Cursor should reflect the eraser while accel is held.
- Should work for the highlight tool too, ideally — same pattern.

## Implementation notes

- `isAccelKey` already exists at `packages/editor/src/lib/utils/keyboard.ts`.
- The closest existing precedent is the spacebar-pan handler in `Editor.ts` (global key handler that swaps modes and restores on key up).
- The masking primitive is `StateNode.setCurrentToolIdMask` — used today by `ZoomTool` to keep the originating tool visible in the UI during a temporary interaction.
- The handler likely belongs in `DrawShapeTool` / its `Idle` state (and a peer in the highlight tool) rather than as a global Editor-level handler, so it stays scoped to the tools that opt in.

## Alternatives considered

- Pressing `E` to switch tools — what users do today, but it changes the active tool in the UI and requires switching back.
- A dedicated "scratch out" gesture (e.g. fast back-and-forth detection) — fuzzier UX, harder to discover.

steveruizok:
In order to let users dip into the eraser without leaving the draw or highlight tool, this PR makes accel-click temporarily erase. While in draw / highlight idle, holding the accel key (Cmd on macOS, Ctrl elsewhere) and clicking goes straight into eraser pointing/dragging. The toolbar still shows the originating tool, and releasing accel returns to it.

Closes #8650.

## How it works

`Idle` (shared by draw and highlight) intercepts `onPointerDown` when accel is held (and shift isn't, since shift is multi-point drawing). It calls `setCurrentTool('eraser.pointing', { ...info, onInteractionEnd: this.parent.id })`, which:

- Enters the eraser tool, masking the toolbar as the originating tool via `setCurrentToolIdMask(onInteractionEnd)`.
- Skips eraser `idle` and goes directly into `pointing` with the original pointer event, so the click counts as the start of the erase interaction.

Returning to the originating tool is driven by the eraser's `idle` child state:

- `Idle.onEnter` returns to the originating tool if accel is no longer held (e.g. user pressed accel, clicked-and-dragged, released accel, then pointer-up brought us back to idle).
- `Idle.onKeyUp` returns immediately when accel is released while idle (the eraser just sits there with no in-flight interaction).

This lets in-flight interactions (`Pointing` → `Erasing`) finish naturally before we switch back, regardless of when the user releases accel.

## Why this approach

The earlier draft pre-switched into the eraser on `keydown` and tracked accel state on the eraser itself. That had two problems: there was no visual cue for the silent swap (both tools use the cross cursor), and accel state had to be hand-rolled because `editor.inputs.getAccelKey()` is debounced. Pointer-down-only is simpler and matches the user's mental model — the swap commits when they actually click.

This also removes the eraser's old "hold accel to erase only the first shape" mode. Accel inside the eraser now means "switch back to the originating tool," and the inner mode added confusion. Erasing now snapshots the shapes at the origin point and erases along the drag path.

## Behavior matrix

| Scenario | Result |
| -- | -- |
| Hold accel in draw idle, no click | Toolbar / cursor unchanged; release accel is a no-op |
| Hold accel + click | Erase as normal; toolbar still shows draw |
| Click + drag with accel held | Drag erases shapes along the path (single-shape mode is gone) |
| Release accel mid-drag | Continue erase; return to draw on pointer up |
| Press accel while drawing a stroke | Existing draw behavior preserved (no tool switch mid-stroke) |
| Eraser opened standalone (E key) | No mask, no transient mode |

### Change type

- [x] `feature`

### Test plan

1. In the draw tool, hold Cmd (or Ctrl) and click a shape — it erases. The toolbar still shows draw.
2. Hold Cmd, drag across multiple shapes — they all erase.
3. Hold Cmd, click-drag, release Cmd mid-drag — drag continues erasing; on pointer up, you're back in draw.
4. Press Cmd while mid-stroke — existing draw behavior (no tool switch) is preserved.
5. Repeat 1–4 with the highlight tool.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Hold accel (Cmd / Ctrl) in the draw or highlight tool to temporarily erase. Click or drag to erase, release accel to return to drawing.

### API changes

- Added public `EraserTool.info: { onInteractionEnd?: string }` field on `EraserTool`.
- Added `EraserTool.maybeReturnToOriginatingTool()`.
- Added `EraserTool.onExit()`.
- Changed `EraserTool.onEnter()` to accept an optional `{ onInteractionEnd?: string }` argument.

### Code changes

| Core code       | +64 / -41  |
| Tests           | +130 / -20 |
| Automated files | +10 / -1   |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,10 +126,10 @@ opencode.json
 # PR walkthrough intermediate files
 .claude/worktrees
 .claude/settings.local.json
-.claude/skills/pr-walkthrough/tmp
-.claude/skills/pr-walkthrough/out
-.claude/skills/pr-walkthrough/video/node_modules
-.claude/skills/pr-walkthrough/video/dist
-.claude/skills/pr-walkthrough/video/public/audio-*
-.claude/skills/pr-walkthrough/video/public/manifest.json
+skills/pr-walkthrough/tmp
+skills/pr-walkthrough/out
+skills/pr-walkthrough/video/node_modules
+skills/pr-walkthrough/video/dist
+skills/pr-walkthrough/video/public/audio-*
+skills/pr-walkthrough/video/public/manifest.json
 todo.md
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -1750,11 +1750,24 @@ export class EraserTool extends StateNode {
     // (undocumented)
     static id: string;
     // (undocumented)
+    info: {
+        onInteractionEnd?: string | undefined;
+    };
+    // (undocumented)
     static initial: string;
     // (undocumented)
     static isLockable: boolean;
+    maybeReturnToOriginatingTool(): void;
     // (undocumented)
-    onEnter(): void;
+    onEnter(info?: {
+        onInteractionEnd?: string;
+    }): void;
+    // (undocumented)
+    onExit(): void;
+    // (undocumented)
+    onKeyDown(info: TLKeyboardEventInfo): void;
+    // (undocumented)
+    onKeyUp(info: TLKeyboardEventInfo): void;
 }
 
 // @public (undocumented)
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -1757,6 +1757,7 @@ export class EraserTool extends StateNode {
     static initial: string;
     // (undocumented)
     static isLockable: boolean;
+    // @internal
     maybeReturnToOriginatingTool(): void;
     // (undocumented)
     onEnter(info?: {
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -1764,10 +1764,6 @@ export class EraserTool extends StateNode {
     }): void;
     // (undocumented)
     onExit(): void;
-    // (undocumented)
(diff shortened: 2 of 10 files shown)


mimecuvalo (commented):
neat idea!

PR #8670, opened 2026-04-28
feat(tldraw): shortcut for adding bound text under arrow terminal while dragging

max-drake:
👩 diagramming annoying sometimes because it requires too many clicks. my most recent take on making this easier is adding the ability to press t while dragging an arrow's end to spawn a bound text shape in editing mode at your pointer

💥 important note: in order to enable this, this PR includes a change to `useKeyboardShortcuts()` that blocks tool switching via keyboard shortcuts unless the user is in an idle state. I _didn't_ include the same logic to block actions in non idle states, because I can imagine there are certain actions that a user or sdk consumer might want to enable while in some non-idle state. 
doing this frees up the tool hotkeys while doing things like dragging, resizing, etc, which opens up some more design space for us to do things like this.

---

In order to label arrow endpoints while adjusting bindings, this PR lets you press **T** while dragging an arrow start or end handle: on key up, tldraw creates a text shape at that terminal, binds it there, and opens it for editing. Page-space terminal helpers in `shared.ts` support correct placement; global tool letter shortcuts are ignored unless the current tool path ends in `.idle`, so an active handle drag is not replaced by switching to the text tool.

### Change type

- [x] `feature`

### Test plan

1. Select an arrow, drag the start or end handle (bound or unbound).
2. Press **T** and release; a new text shape should appear at that terminal and enter edit mode without switching to the text tool first or inserting **t** into the new label.
3. Run `yarn test run src/lib/shapes/arrow/dropShapeOnArrowDrag.test.ts` in `packages/tldraw`.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add **T** while dragging an arrow terminal to create bound, editable text at that terminal.

### Code changes

| Core code       | +180 / -13 |
| Tests           | +58 / -0   |
| Automated files | +0 / -0    |
| Documentation   | +0 / -0    |
| Apps            | +0 / -0    |
| Templates       | +0 / -0    |
| Config/tooling  | +0 / -0    |

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/dropShapeOnArrowDrag.test.ts
+++ b/packages/tldraw/src/lib/shapes/arrow/dropShapeOnArrowDrag.test.ts
@@ -24,11 +24,18 @@ describe('dropBoundTextShapeAtArrowTerminal', () => {
 		editor.pointerMove(500, 500)
 		editor.expectToBeIn('select.dragging_handle')
 
-		const arrow = editor.getCurrentPageShapes().find((s) => s.type === 'arrow') as TLArrowShape
-		expect(arrow).toBeTruthy()
+		const arrowId = editor.getCurrentPageShapes().find((s) => s.type === 'arrow')!
+			.id as TLArrowShape['id']
+		editor.pointerMove(600, 400)
+
+		const arrow = editor.getShape<TLArrowShape>(arrowId)!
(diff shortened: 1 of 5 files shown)


mimecuvalo:
neat idea!

angrycaptain19:
Discussed during design review and the conclusion was that there are too many open questions about how we would generalize this to a broader feature than just this use case. Either for being able to handle this arrow creation behavior for all shape types or to be able to generally have custom post shape creation behavior in different ways

PR #8683, opened 2026-04-29
fix(utils): reject pending promise on debounce.cancel()

mimecuvalo:
In order to stop callers from hanging forever when a debounced function is cancelled mid-flight, this PR makes `debounce(...).cancel()` reject the pending promise with `Error('Debounced function was cancelled')`. Follow-up to a review note on #8604, where `cancel()` was updated to release captured arguments but the pending promise was still left unsettled.

Previously, `cancel()` cleared the pending timeout but never resolved or rejected the in-flight promise — so any code that did `await debouncedFn(...)` and then called `.cancel()` (or had it called externally, e.g. via a hook cleanup) would hang forever. To avoid breaking the common fire-and-forget pattern (`useZoomCss`, `Editor.deepLink`, search debouncers, bookmark debouncer, VS Code change responder, etc.), an internal no-op `.catch` is attached so callers that discard the promise don't trigger unhandled-rejection warnings. Consumers that `await` or chain `.then` / `.catch` on the returned promise still observe the rejection through their own chain.

### Change type

- [x] `bugfix`

### Test plan

- [x] Unit tests

New tests cover:

- `cancel()` rejects all pending callers from the same debounce window with the expected error.
- A fresh call after `cancel()` gets a brand-new promise (and verifies state is reset cleanly).
- `cancel()` is a no-op when nothing is pending; double-cancel is safe.
- No `unhandledRejection` event fires when callers discard the promise and call `cancel()` (verifies the fire-and-forget pattern is unaffected).

### Release notes

- Fixed `debounce(...).cancel()` leaving any in-flight promise unsettled. Calling `cancel()` while a debounced call is pending now rejects the returned promise with `Error('Debounced function was cancelled')`. Code that `await`s a debounced call across a cancel should wrap it in `try/catch`.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +21 / -3   |
| Tests     | +62 / -0   |

### DIFF
--- a/packages/utils/src/lib/debounce.test.ts
+++ b/packages/utils/src/lib/debounce.test.ts
@@ -118,9 +118,7 @@ describe(debounce, () => {
 			debounced()
 			debounced.cancel()
 
-			vi.useRealTimers()
-			await new Promise((resolve) => setTimeout(resolve, 0))
-			vi.useFakeTimers()
+			await new Promise(process.nextTick)
 
 			expect(handler).not.toHaveBeenCalled()
 		} finally {
--- a/packages/utils/src/lib/debounce.test.ts
+++ b/packages/utils/src/lib/debounce.test.ts
@@ -65,4 +65,66 @@ describe(debounce, () => {
 		expect(fn).toHaveBeenCalledTimes(2)
 		expect(await Promise.all([promiseC, promiseD])).toEqual(['gh', 'gh'])
 	})
+
+	it('rejects the pending promise when cancelled', async () => {
+		const fn = vi.fn()
+		const debounced = debounce(fn, 100)
+		const promiseA = debounced()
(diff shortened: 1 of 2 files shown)


mimecuvalo:
@MitjaBezensek lemme know if still worth doing this

MitjaBezensek (approved):

frolic:
Ohh I wonder if we can use an AbortSignal for this (there's a lot of recent tooling heading more and more in the direction of using AbortSignals)

mimecuvalo:
> Ohh I wonder if we can use an AbortSignal for this (there's a lot of recent tooling heading more and more in the direction of using AbortSignals)

Ah good call — I'm down for this, wanna give it a stab?

PR #8691, opened 2026-04-29
fix(toolbar): update geo icon when switching between geo variants

### LINKED ISSUE #8689: Toolbar geo icon does not swap when pressing R then O
## Description

When the geo tool slot in the main toolbar is showing rectangle and you press `O` to switch to oval (or vice versa), the toolbar icon does not update to reflect the new geo variant. The blue \"selected\" indicator disappears (correctly showing the rectangle button is no longer active), but the icon itself stays the same instead of swapping to the ellipse, so the now-active geo tool isn't shown anywhere in the visible toolbar.

The bug only reproduces when switching directly between two geo variants. Going through any non-geo tool first (for example `R`, `V`, `O`) works correctly.

## Where it reproduces

- tldraw.com (https://tldraw.com)
- examples-canary (https://canary.examples.tldraw.com)

It does **not** reproduce on the deployed examples app at https://examples.tldraw.com.

## Steps to reproduce

1. Open https://tldraw.com or examples-canary.
2. Press `R` to select the rectangle tool. The toolbar shows a rectangle icon with the blue selected indicator.
3. Press `O` to select the oval tool.

**Expected:** the toolbar slot swaps to show the ellipse icon, with the blue selected indicator still on it.

**Actual:** the icon stays as rectangle, but the blue selected indicator goes away.

Pressing `R`, `V`, `O` does not show the bug — going through select tool resets the state correctly.

## Likely cause

PR #8574 (`fix(toolbar): avoid hover-triggered overflow recalculations causing ResizeObserver loop warnings`) removed `attributes` and `characterData` from the `MutationObserver` in `packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx`. The PR description noted that React-driven attribute changes already trigger recalculation via the unconditional `useLayoutEffect` that runs after every render of `OverflowingToolbar`.

That assumption holds when the parent re-renders, but switching directly between two geo variants only re-renders the individual `ToolbarItem` children whose `useIsToolSelected` value changed. `OverflowingToolbar` itself does not re-render, so its `useLayoutEffect` does not fire and `lastActiveOverflowItem` stays pinned to the previous geo variant. Going through the select tool re-renders the toolbar parent, which is why `R, V, O` works.

max-drake:
Closes #8689

In order to fix the toolbar geo icon not updating when switching directly between geo variants (e.g. pressing `R` then `O`), this PR subscribes `OverflowingToolbar` to the current tool ID and geo style via `useValue`. When either value changes, the component re-renders, the existing unconditional `useLayoutEffect` fires `onDomUpdate`, and `lastActiveOverflowItem` gets recalculated to promote the correct geo variant.

Previously, switching between geo variants only re-rendered the individual `ToolbarItem` children (whose `useIsToolSelected` hook toggled `aria-pressed`), but `OverflowingToolbar` itself did not re-render. After #8574 removed `attributes` from the `MutationObserver`, there was no remaining mechanism to trigger `onDomUpdate` for attribute-only changes. This fix uses reactive state as the source of truth rather than DOM attribute observation, avoiding the ResizeObserver loop that #8574 was solving.

### Change type

- [x] `bugfix`

### Test plan

1. Open the examples app at localhost:5420
2. Press `R` to select the rectangle tool — the toolbar shows a rectangle icon with the blue selected indicator
3. Press `O` to select the oval tool
4. Verify the toolbar slot swaps to show the ellipse icon with the blue selected indicator
5. Try other direct geo→geo switches (e.g. `R` → `O` → `R`, triangle via overflow → rectangle via `R`)
6. Hover over toolbar items and verify no ResizeObserver loop warnings appear in the console

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix toolbar geo icon not updating when switching directly between geo variants via keyboard shortcuts.

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +14 / -0   |

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/OverflowingToolbar.tsx
@@ -1,4 +1,5 @@
 import {
+	GeoShapeGeoStyle,
 	activeElementShouldCaptureKeys,
 	assert,
 	modulate,
@@ -7,6 +8,7 @@ import {
 	useEditor,
 	useEvent,
 	useUniqueSafeId,
+	useValue,
 } from '@tldraw/editor'
 import classNames from 'classnames'
 import { createContext, useEffect, useLayoutEffect, useRef, useState } from 'react'
@@ -75,6 +77,18 @@ export function OverflowingToolbar({
 	const [overflowTools, setOverflowTools] = useState<HTMLDivElement | null>(null)
 	const [lastActiveOverflowItem, setLastActiveOverflowItem] = useState<string | null>(null)
 	const [shouldShowOverflow, setShouldShowOverflow] = useState(false)
+
+	// Subscribe to tool and geo style changes so this component re-renders (and the
+	// unconditional useLayoutEffect fires onDomUpdate) when switching between geo
+	// variants directly, e.g. rectangle → ellipse.
+	useValue(
+		'current tool and geo style',
+		() => {
+			const toolId = editor.getCurrentToolId()
+			return toolId === 'geo' ? editor.getSharedStyles().getAsKnownValue(GeoShapeGeoStyle) : toolId
+		},
+		[editor]
+	)
 
 	const onDomUpdate = useEvent(() => {
 		if (!mainToolsRef.current) return


max-drake:
closed, duplicated by https://github.com/tldraw/tldraw/pull/8690

PR #8705, opened 2026-04-29
refactor(shapes): rename customGeoStyles option back to customGeoTypes

mimecuvalo:
In order to make the geo extensibility API self-consistent before its first release, this PR renames the `customGeoStyles` option on `GeoShapeUtil.configure()` back to `customGeoTypes`. Follow-up to #8543.

The original PR introduced `customGeoTypes`, then it was renamed to `customGeoStyles` during review feedback. The rename was applied to the option name, but several adjacent surfaces were missed:

- The exported `GeoTypeDefinition` interface (still uses "type")
- The internal `getCustomGeoType` helper and `customType` locals
- The `apps/examples/src/examples/shapes/tools/custom-geo-types/` example folder, README title, and `CustomGeoTypesExample` component
- The release notes in `apps/docs/content/releases/next.mdx`, which still document `customGeoTypes` (so the example as published would not actually compile)

Re-aligning everything on "types" is the smaller change and matches the conceptual model — `'rectangle' | 'ellipse' | 'rounded-rect'` are kinds of geo, not styles. The "style" naming would have collided with the existing `GeoShapeGeoStyle` `StyleProp` and the unrelated style panel terminology.

This API has not shipped yet (it lives in `next.mdx`), so no released consumers are affected.

### Change type

- [x] `api`

### Test plan

1. Open the `custom geo types` example.
2. Verify the rounded rectangle and cross shapes render, drag-create from the toolbar, and appear in the geo style panel picker.

- [x] Unit tests (existing geo + style tests still pass)

### API changes

- Renamed `GeoShapeOptions.customGeoStyles` to `GeoShapeOptions.customGeoTypes`. The option keys this map populates on `GeoShapeGeoStyle` are unaffected.

### Code changes

| Core code       | +27 / -27  |
| Documentation   | +11 / -11  |

### DIFF
--- a/apps/examples/src/examples/shapes/tools/custom-geo-types/CustomGeoTypesExample.tsx
+++ b/apps/examples/src/examples/shapes/tools/custom-geo-types/CustomGeoTypesExample.tsx
@@ -14,7 +14,7 @@ import 'tldraw/tldraw.css'
 
 // [1]
 const CustomGeoShapeUtil = GeoShapeUtil.configure({
-	customGeoStyles: {
+	customGeoTypes: {
 		'rounded-rect': {
 			getPath(w, h, shape) {
 				// `isFilled` is used by the path builder to determine whether the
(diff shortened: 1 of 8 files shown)


MitjaBezensek:
Still feels strange that we call these styles [here](https://github.com/tldraw/tldraw/blob/65b9ee3a8f3ebe19425f0074b4e0e3ebd86c5bb6/packages/tlschema/src/shapes/TLGeoShape.ts#L40), but for extending we use types.

mimecuvalo:
> Still feels strange that we call these styles [here](https://github.com/tldraw/tldraw/blob/65b9ee3a8f3ebe19425f0074b4e0e3ebd86c5bb6/packages/tlschema/src/shapes/TLGeoShape.ts#L40), but for extending we use types.

yeah, Steve and I talked about this but still felt like types was the better path fwd

PR #8710, opened 2026-04-29
fix(tldraw): split collaborator cursor head and tail into separate fills

### LINKED ISSUE #8692: White square visible at base of collaborator cursor
## Description

The collaborator cursor renders with a visible white square at the base of the arrow. The white outline layer is significantly larger than the colored fill that should sit on top of it, so the white path bleeds through where it should be hidden.

collaborator-cursor-white-square

## Steps to reproduce

1. Open a tldraw room with another collaborator (or use the multi-user examples).
2. Observe the other user's cursor.
3. A white rectangle is visible attached to the bottom-left of the cursor arrow.

## Expected behavior

The cursor renders as a single colored arrow with a thin white outline around the entire shape.

## Actual behavior

The "tail" portion of the cursor renders as a white rectangle with only a small sliver of color showing.

## Likely cause

In `packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts` (lines 19-35) the cursor is built from three stacked `Path2D` objects (shadow, white, fill). Each path contains two subpaths (the arrow head and the tail/fletching).

The white path's tail subpath and the fill path's tail subpath have very different sizes. Roughly:

- White tail: ~12 × 4 units
- Fill tail: ~8 × 2 units

So the colored fill cannot cover the white outline beneath it. This mismatch is the white square visible in the screenshot.

The arrow-head subpaths look closer in size, but the tail subpaths are clearly out of sync — most likely a bad path or wrong fill order from the SVG-to-Path2D conversion done in the recent canvas overlay migration (#8633).

## Affected files

- `packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts`

kostyafarber:
In order to remove the white square visible at the base of the collaborator cursor, this PR splits each cursor layer (shadow, white outline, colored fill) into two separate `Path2D` objects — one for the arrowhead, one for the tail — and fills them with two separate `ctx.fill()` calls per layer. Closes #8692.

The previous implementation packed both subpaths into one `Path2D` per layer and filled them in a single call. With the default `nonzero` winding rule, the canvas renderer evaluates both subpaths together as one compound shape; along the seam where the head and tail overlap, this produced sub-pixel rasterisation artefacts that surfaced as a visible white sliver/square at the base of the tail in some DPR + zoom combinations. The original SVG-DOM implementation did not have this problem because each `<path>` was rendered with its own fill operation, with no compound-shape rasterisation across the shared boundary.

The split mirrors the original SVG's two-`<path>`-per-`<g>` layout and matches its rendering behaviour exactly. The white tail's path coordinates were also corrected from a 4×12 parallelogram to a 4×10 one — a 1-unit perpendicular outline of the 2×8 colored fill tail — so the white outline no longer extends past the colored fill.

### Change type

- [x] `bugfix`

### Test plan

1. Open a multiplayer example (e.g. `multiplayer-demo`) with another collaborator.
2. Observe the remote collaborator's cursor at default zoom on a HiDPI display.
3. Confirm the cursor renders as a single colored arrow with a thin white outline; no white rectangle visible at the base of the tail.
4. Zoom in/out and pan; confirm the cursor stays clean at all zoom levels.

- [x] Unit tests

### Release notes

- Fix a white square artefact that appeared at the base of remote collaborator cursors.

### DIFF
--- a/packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts
+++ b/packages/tldraw/src/lib/overlays/CollaboratorCursorOverlayUtil.ts
@@ -12,26 +12,37 @@ export interface TLCollaboratorCursorOverlay extends TLOverlay {
 }
 
 // Lazy-initialized Path2D objects for the cursor arrow shape (deferred to avoid jsdom issues in tests).
-let _shadowPath: Path2D | null = null
-let _whitePath: Path2D | null = null
-let _fillPath: Path2D | null = null
+// Head and tail are separate Path2D objects per layer so each is filled in its own draw call,
+// mirroring the original SVG's two-<path> layout per group.
+let _shadowHead: Path2D | null = null
+let _shadowTail: Path2D | null = null
+
+let _whiteHead: Path2D | null = null
+let _whiteTail: Path2D | null = null
+
+let _fillHead: Path2D | null = null
+let _fillTail: Path2D | null = null
 
 function getCursorPaths() {
-	if (!_shadowPath) {
+	if (!_shadowHead) {
 		// Shadow layer (original SVG translate(-11,-11) pre-applied):
-		_shadowPath = new Path2D(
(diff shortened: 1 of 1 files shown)


max-drake (approved):
lgtm

PR #8761, opened 2026-05-05
feature(docs): process image url through an asset url helper

kaneel:
Following dotdev imminent release, we need to make sure all assets URLs in doc can be prefixed with the right url so we don't have issues when visitors are coming from dotdev urls

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

- [ ] visit staging docs
- [ ] images are working

### DIFF
--- a/apps/docs/utils/asset-url.ts
+++ b/apps/docs/utils/asset-url.ts
@@ -1,0 +1,6 @@
+export function assetUrl(path: string): string {
+	if (!path.startsWith('/')) return path
+	const prefix = process.env.ASSET_PREFIX?.trim()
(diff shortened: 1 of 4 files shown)


apps/docs/utils/asset-url.ts
```diff
@@ -0,0 +1,6 @@
+export function assetUrl(path: string): string {
```
kostyafarber, line 1: maybe a small test to ensure this works proprely

kostyafarber (approved):
approved modulo test

PR #8768, opened 2026-05-05
feature(docs): improve logic for asset origin fix

kaneel:
### Problem

Docs and marketing are two separate Next apps on Vercel. When marketing rewrites a path (e.g. `tldraw.dev/quick-start`) and returns docs HTML, the browser’s address bar is still the marketing origin, but the HTML referenced `/_next/static/...` as root-relative paths. Those requests went to marketing’s `/_next` (wrong app / wrong hashes) instead of docs’ `/_next`.

Setting `ASSET_PREFIX` only in some environments (or only on Production) was easy to get wrong: Preview builds often had no prefix, and Production could pin the wrong hostname (e.g. a `*.vercel.app` alias) while the live site was `tldraw.dev`, which caused 404s, cache skew (old HTML vs new static files), and confusion.

### What we changed

`assetPrefix` (and the same base for `assetUrl()` on `public/` paths) now resolves to:

1. `ASSET_PREFIX` when set (explicit override), else
2. `https://${VERCEL_URL}` on Vercel (the current deployment URL), else
3. unset locally → root-relative assets as before.

### Why it helps

- Every Vercel build automatically points JS/CSS (and prefixed `public/` URLs) at that deployment’s static host, without maintaining Preview-specific `ASSET_PREFIX` for every branch.
- `ASSET_PREFIX` remains an escape hatch when you need a fixed canonical origin (e.g. force the apex in prod).
- Docs HTML served through rewrites still loads `/_next` from the docs deployment that produced the page, which removes the “wrong `/_next` host” class of bugs.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/apps/docs/next.config.js
+++ b/apps/docs/next.config.js
@@ -3,9 +3,19 @@
 // Configurable domain for rewrites
 const REWRITE_DOMAIN = 'tldrawdotdev.framer.website'
 
+function resolveAssetOrigin() {
+	const explicit = process.env.ASSET_PREFIX?.trim()
+	if (explicit) return explicit.replace(/\/$/, '')
+
+	const vercelUrl = process.env.VERCEL_URL?.trim()
+	if (vercelUrl) return `https://${vercelUrl}`
+
+	return undefined
+}
+
 const nextConfig = {
 	reactStrictMode: true,
-	assetPrefix: process.env.ASSET_PREFIX || undefined,
+	assetPrefix: resolveAssetOrigin(),
 	experimental: {
 		scrollRestoration: true,
 	},
--- a/apps/docs/utils/asset-url.test.ts
+++ b/apps/docs/utils/asset-url.test.ts
@@ -3,53 +3,96 @@ import { assetUrl } from './asset-url'
 
 describe('assetUrl', () => {
 	it('returns the original path when ASSET_PREFIX is unset', () => {
-		const prev = process.env.ASSET_PREFIX
+		const prevPrefix = process.env.ASSET_PREFIX
+		const prevVercel = process.env.VERCEL_URL
 		try {
 			delete process.env.ASSET_PREFIX
+			delete process.env.VERCEL_URL
 			expect(assetUrl('/favicon.svg')).toBe('/favicon.svg')
 		} finally {
-			process.env.ASSET_PREFIX = prev
+			process.env.ASSET_PREFIX = prevPrefix
+			process.env.VERCEL_URL = prevVercel
 		}
 	})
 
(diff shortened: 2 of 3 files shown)


kostyafarber:
🧓 i asked claude to see if there was less hacky solution and this is what it said. worth exploring as a follow up later, (as in after we launch if this pr fixes the issues), not sure if it'll work, but this was just one prompt

---

The long-term fix here is the official [Next.js Multi-Zones](https://nextjs.org/docs/pages/guides/multi-zones) pattern, which removes the cross-origin asset problem entirely instead of working around it.

  **The idea:** keep everything same-origin under `tldraw.dev` by giving docs a unique path prefix, and have marketing rewrite both pages *and* the asset path through to the docs deployment.

  **On docs (the child zone):**

  ```js
  // apps/docs/next.config.js
  const nextConfig = {
    basePath: '/docs-static',
    // (assetPrefix is implied by basePath in Next 15+)
  }
  ```

  **On marketing (the parent zone):**

  ```js
  async rewrites() {
    return [
      // existing page rewrites
      { source: '/quick-start', destination: 'https://docs-deploy.vercel.app/quick-start' },

      // NEW: forward the prefixed asset path to the docs deployment
      { source: '/docs-static/:path+', destination: 'https://docs-deploy.vercel.app/docs-static/:path+' },
    ]
  }
  ```

  **What this gives us:**

  - Docs HTML now references `/docs-static/_next/...`. The browser requests it from `tldraw.dev`, marketing rewrites it to the docs deployment, the docs deployment serves its own hashed chunks. HTML and chunks always come from the same deployment, so cache skew
  goes away.
  - Everything stays on `tldraw.dev` — no cross-origin, no CORS preflights for fonts/workers, no `*.vercel.app` URLs in DevTools.
  - **Preview deployments work without exception.** The `VERCEL_URL` approach in this PR breaks on previews where Deployment Protection is enabled, because assets get blocked by Vercel's SSO wall. With multi-zones, the asset request goes through marketing's
  origin and never hits the protected `*.vercel.app` directly.
  - `ASSET_PREFIX` env juggling per environment goes away — the prefix is a static config value.

  **Two caveats to plan for:**

  1. `assetPrefix` / `basePath` doesn't cover `public/` files — keep the `assetUrl()` helper for those, or move them under `app/(static)` so they ride `/_next`.
  2. `/_next/data` has its own [known limitations](https://github.com/vercel/next.js/discussions/25681) with prefixes — worth checking if any SSG pages here use it.

  So: ship this PR as the stopgap, and track moving to the prefix + marketing-side rewrite as the real fix.

kostyafarber (approved):

PR #8780, opened 2026-05-06
dev: add editor to window object when the debug menu is opened

mimecuvalo:
i keep reaching for editor but it's never there when i need it

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### DIFF
--- a/packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
+++ b/packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
@@ -1,4 +1,5 @@
-import { ReactNode } from 'react'
+import { useEditor } from '@tldraw/editor'
+import { ReactNode, useEffect } from 'react'
 import { TldrawUiButton } from '../primitives/Button/TldrawUiButton'
 import { TldrawUiButtonIcon } from '../primitives/Button/TldrawUiButtonIcon'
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui/components/DebugMenu/DefaultDebugMenu.tsx
```diff
@@ -16,8 +17,22 @@ export interface TLUiDebugMenuProps {
...
 	const content = children ?? <DefaultDebugMenuContent />
 
+	// While the debug menu is mounted, expose the editor on `window.editor` for
+	// console-driven debugging. We remove it on unmount so the editor isn't
+	// retained when debug mode is turned off.
+	useEffect(() => {
+		const win = window as any
+		win.editor = editor
+		return () => {
+			if (win.editor === editor) {
```
kostyafarber, line 30: my memory leak spidey senses are firing, how do these objects get compared under the hood? might be a silly question :P

mimecuvalo: naw, this'll be fine, the `===` will do the right thing here

mimecuvalo: good to check though 👍

Issue #7831, opened 2026-02-04
Centralize shape culling display updates to reduce per-shape subscriptions


steveruizok:
## Problem

Each `Shape` component in `packages/editor/src/lib/components/Shape.tsx` has its own `useQuickReactor` that subscribes to `editor.getCulledShapes()` to update its display property:

```tsx
useQuickReactor(
    'set display',
    () => {
        const shape = editor.getShape(id)
        if (!shape) return

        const culledShapes = editor.getCulledShapes()
        const isCulled = culledShapes.has(id)
        if (isCulled !== memoizedStuffRef.current.isCulled) {
            setStyleProperty(containerRef.current, 'display', isCulled ? 'none' : 'block')
            setStyleProperty(bgContainerRef.current, 'display', isCulled ? 'none' : 'block')
            memoizedStuffRef.current.isCulled = isCulled
        }
    },
    [editor]
)
```

With N shapes on the canvas, this creates N separate subscriptions all watching the same computed value. When culled shapes change, all N reactors run even though most shapes' culling state hasn't changed.

## Proposed solution

Create a centralized registry of shape containers that allows a single reactor to imperatively update display properties:

1. **Registry**: A `Map<TLShapeId, { container, bgContainer }>` that tracks shape container refs
2. **Registration**: Shape components register their refs on mount, unregister on unmount
3. **Single reactor**: One centralized subscription that diffs the culled shapes set and only updates containers whose state changed

This changes the complexity from O(N) subscriptions to O(1) subscription, with O(changed shapes) DOM updates.

## Implementation sketch

```typescript
// Registry (could live on Editor or as a separate manager)
class ShapeContainerRegistry {
  private containers = new Map<TLShapeId, {
    container: HTMLDivElement | null
    bgContainer: HTMLDivElement | null
  }>()

  register(id, container, bgContainer) { ... }
  unregister(id) { ... }
  
  updateCulling(culledShapes: Set<TLShapeId>, prevCulled: Set<TLShapeId>) {
    // Only update shapes whose culling state changed
    for (const id of culledShapes) {
      if (!prevCulled.has(id)) {
        // Newly culled - hide
      }
    }
    for (const id of prevCulled) {
      if (!culledShapes.has(id)) {
        // Newly visible - show
      }
    }
  }
}

// Single reactor in DefaultCanvas/ShapesToDisplay
function CullingController() {
  const editor = useEditor()
  const prevCulledRef = useRef<Set<TLShapeId>>(new Set())

[...]

PR #8788, opened 2026-05-07
perf(editor): skip notVisibleShapes recompute on prop-only changes

MitjaBezensek:
While drawing, every pointer move updates the active draw shape's `segments` prop. Previously this caused the `notVisibleShapes` derivation to recompute for every prop change (it subscribed to `getCurrentPageShapes()`, which invalidates on any shape prop change), and the spatial index would scan every shape on the page looking for transitive bounds changes even when no shape's bounds had actually moved.

This PR makes two narrow changes:

1. `notVisibleShapes` now subscribes to `getCurrentPageShapeIds()` (incremental, only invalidates on add/remove/reparent) and reads each shape via `store.unsafeGetWithoutCapture()` so per-shape prop changes do not invalidate the derivation. Visibility itself is driven by `getShapeIdsInsideBounds()`, which is gated on a new bounds epoch.
2. `SpatialIndexManager` returns a `_boundsEpoch` that only ticks when the rbush actually changes (a bounds was added, removed, or moved). On incremental updates we now compare the new bounds to the indexed bounds and only `upsert` when they differ. When the diff produced no real bounds changes, we also skip the all-shapes sweep — transitive bounds invalidations require at least one direct input bounds change, so there is nothing to find.

Net effect: while actively drawing on a populated page, `notVisibleShapes` no longer recomputes per pointer move, and the spatial index avoids redundant rbush work and the O(n) sweep.

Relates to #7831

### Change type

- [x] `improvement`

### Test plan

1. Open a page with a few hundred shapes.
2. Start drawing a freehand stroke and move the pointer continuously.
3. Confirm no perf regression in interaction; existing culling behavior (shapes outside the viewport hidden) is unchanged when panning/zooming or when shapes are added, moved, or deleted.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve drawing performance on pages with many shapes by skipping viewport culling recomputation when only shape props change.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +76 / -38  |

### DIFF
--- a/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
+++ b/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
@@ -17,6 +17,11 @@ export function notVisibleShapes(editor: Editor) {
 		//  - getViewportPageBounds: invalidates on pan/zoom
 		//  - getShapeIdsInsideBounds: subscribes to the spatial index epoch,
 		//    which only ticks when a shape's bounds actually change
+		//  - editor.getShape(id) in the slow path below: per-shape subscription,
+		//    only for offscreen shapes — needed so canCull() flips driven by
+		//    prop changes (e.g. a shape with `preventCulling`) re-run this
+		//    derivation. The active draw shape is onscreen and short-circuited
+		//    out before we read it, so drawing pointer-moves don't re-subscribe.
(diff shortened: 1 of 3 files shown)


MitjaBezensek:
Will make more targeted changes.

PR #8792, opened 2026-05-07
chore: remove unnecessary useCallback in apps/docs and apps/dotcom/client

AniKrisn:
In order to remove bookkeeping that has no downstream consumer, this PR drops `useCallback` wrappers from event handlers in 8 components across `apps/docs` and `apps/dotcom/client`. In every case the consumer component is not wrapped in `React.memo`, the callback is not used as a dependency in any other hook, and there is no other reason for the cached reference to matter — so the `useCallback` was pure overhead.

Opening as a draft because I am unsure about whether there is something I cannot see!

### Motivation

While reviewing `apps/docs/components/content/copy-button.tsx`, the `useCallback` on `handleClick` looked unnecessary. The consumer (`Button` in `apps/docs/components/common/button.tsx`) is not wrapped in `React.memo`, the callback was never used as a hook dep, and `Button` itself creates a fresh inline arrow function each render and forwards _that_ to the real DOM `<button>` — so any reference stability from the parent's `useCallback` was discarded one component down. Investigating the same pattern across `apps/docs` and `apps/dotcom/client` surfaced 7 more files.

Related: https://react.dev/learn/you-might-not-need-an-effect

### What changed

Replaced `useCallback` wrappers with plain function declarations (and dropped the now-unused `useCallback` imports):

| File | useCallbacks removed |
| ---- | -------------------- |
| `apps/docs/components/content/copy-button.tsx` | 1 |
| `apps/docs/components/docs/copy-markdown-button.tsx` | 1 |
| `apps/docs/components/docs/docs-feedback-widget.tsx` | 3 |
| `apps/dotcom/client/.../TlaSidebarCreateFileButton.tsx` | 1 |
| `apps/dotcom/client/.../TlaFileShareMenu/Tabs/TlaAnonCopyLinkTab.tsx` | 1 |
| `apps/dotcom/client/.../TlaFileShareMenu/Tabs/TlaExportTab.tsx` | 6 |
| `apps/dotcom/client/.../dialogs/TlaDeleteFileDialog.tsx` | 1 |
| `apps/dotcom/client/.../TlaSidebar/components/TlaSidebarFileLink.tsx` | 1 |

**Skipped:** `apps/docs/components/search/FullPageSearch.tsx`. Its `handleKeyDown` is passed to a third-party Ariakit `Combobox`. I did not want to remove memoization without verifying Ariakit's internals.

### Reasoning

For each file I verified three things:

1. The consumer component is not wrapped in `React.memo`. Confirmed for `Button`, `TldrawUiButton` (`forwardRef` only, not `memo`), `TlaShareMenuCopyButton`, `TlaMenuSwitch`, `TlaMenuSelect`, and `TlaButton` (`forwardRef` only, not `memo`).
2. The callback is not referenced as a dependency in any other hook in the same file.
3. Every handler is a local `const` inside its component body — no cross-file consumption.

Without a memoized child or another hook depending on the function reference, `useCallback` adds bookkeeping cost (deps comparison, cached-function retention) without any downstream benefit. The deps arrays exist because `eslint-plugin-react-hooks/exhaustive-deps` requires them once `useCallback` is in use — not because the callback relies on stale-capture semantics. Every closure here reads values that should always be fresh, so removing `useCallback` is, if anything, slightly safer (no risk of a missing dep producing a stale closure later).

### Verification

- `yarn typecheck` from repo root: passes.
- `yarn lint-current` on changed files: 0 warnings, 0 errors.
- `yarn test run` in `apps/dotcom/client`: 133 tests pass.
- `yarn test run` in `apps/docs`: 10 tests pass.
- Manual smoke testing of affected UI surfaces: not done.

### Open questions for reviewers

1. Is there a team convention I'm missing where `useCallback` is preferred prophylactically (e.g. in case a child is later memoized)? If so, this PR contradicts that convention.
2. Are any of the affected dotcom components (`TlaButton`, `TldrawUiButton`, etc.) intended to be wrapped in `React.memo` in the future? If so, leaving the `useCallback`s in might be the right call.
3. Should `FullPageSearch.tsx` be cleaned up too? I left it because the consumer is third-party.
4. The cleanup is consistent with React's official guidance ("you don't need useCallback unless..."), but I want to confirm that matches this codebase's preference before extending the audit.

### Proposition

If the reasoning holds, I'd like to extend a similar audit/cleanup pass to other parts of the repo where the pattern recurs. I'd want sign-off on the philosophy first.

### Change type

- [x] `other`

### Test plan

1. In docs, navigate to a page with a code block, click the copy button — text should copy and the icon should briefly change to a checkmark.
2. In docs, on any docs page click the feedback widget thumbs up/down, then submit feedback in the textarea — the success state should appear.
3. In tldraw.com, sign in and click the sidebar "Create file" button — it should create a new file and open inline rename.
4. In tldraw.com, open the share menu on a file, switch to the Export tab, toggle padding/background, change theme/format, click "Export image" — should produce an image download.
5. In tldraw.com (signed out), open the share menu and click "Copy link" — clipboard should contain a deep link.
6. In tldraw.com, right-click a file in the sidebar and choose Delete — the confirm dialog should fire and the file should be deleted/forgotten.
7. In tldraw.com, hover the guest badge on a shared file in the sidebar — clicking the tooltip should navigate to the file.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Documentation   | +29 / -32  |
| Apps            | +41 / -55  |

### DIFF
--- a/apps/docs/components/content/copy-button.tsx
+++ b/apps/docs/components/content/copy-button.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { RefObject, useCallback, useState } from 'react'
+import { RefObject, useState } from 'react'
 import { Button } from '../common/button'
 
 export function CopyButton({
@@ -11,7 +11,7 @@ export function CopyButton({
 	className?: string
 }) {
 	const [copied, setCopied] = useState<boolean>(false)
-	const handleClick = useCallback(() => {
+	const handleClick = () => {
 		const code: string = typeof copy === 'string' ? copy : (copy.current?.innerText ?? '')
 
 		navigator.clipboard.writeText(code)
@@ -25,7 +25,7 @@ export function CopyButton({
 
 		setCopied(true)
 		setTimeout(() => setCopied(false), 1500)
-	}, [copy])
+	}
 
 	return (
 		<Button
--- a/apps/docs/components/docs/copy-markdown-button.tsx
+++ b/apps/docs/components/docs/copy-markdown-button.tsx
@@ -1,6 +1,6 @@
 'use client'
 
-import { useCallback, useState } from 'react'
+import { useState } from 'react'
 import { Button } from '../common/button'
 
 export function CopyMarkdownButton({
@@ -12,7 +12,7 @@ export function CopyMarkdownButton({
 }) {
 	const [copied, setCopied] = useState(false)
 
-	const handleClick = useCallback(() => {
+	const handleClick = () => {
 		navigator.clipboard.writeText(markdown)
 
 		if (window.tlanalytics?.trackCopyCode) {
@@ -24,7 +24,7 @@ export function CopyMarkdownButton({
 
 		setCopied(true)
 		setTimeout(() => setCopied(false), 1500)
-	}, [markdown])
+	}
 
 	return (
 		<Button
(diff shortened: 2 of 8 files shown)


mimecuvalo (approved):
👍

PR #8794, opened 2026-05-07
fix(ui): use TldrawUiToolbarButton for all video toolbar buttons

### LINKED ISSUE #8787: Inconsistent button types on video import
In `DefaultVideoToolbarContent.tsx` there seems to be an inconsistency with the buttons used. For the first two options we make use of `<TldrawUiButton>` and for the last button we make use of `<TldrawUiToolbarButton>`.

The above mentioned difference leads to a case where the tooltips shown on the buttons are not the same. This is not observed for the `DefaultImageToolbarContent.tsx` as they all use the same type of button which is `<TldrawUiToolbarButton>`

Here is a video demonstrating this.

https://github.com/user-attachments/assets/7fd6dc77-c16a-4602-b823-0b4dcad9b392

MitjaBezensek:
In order to make tooltip behavior consistent across the video toolbar, this PR switches the replace and download buttons in `DefaultVideoToolbarContent` from `TldrawUiButton` to `TldrawUiToolbarButton`. The alt-text button already used `TldrawUiToolbarButton`, so the buttons now match each other and the image toolbar.

`TldrawUiToolbarButton` wraps the button in a `TldrawUiTooltip`, while `TldrawUiButton` falls back to the native `title` attribute. Mixing the two in the same toolbar produced inconsistent hover tooltips, as shown in the issue.

### Before

https://github.com/user-attachments/assets/55cb00ea-a801-4fe3-bd44-3d871477dea0

### After

https://github.com/user-attachments/assets/6d98885c-eefe-4eb1-a010-456a56d27d88

Closes #8787

### Change type

- [x] `bugfix`

### Test plan

1. Run \`yarn dev\` and open the examples app.
2. Drop a video onto the canvas and select it.
3. Hover each button on the video toolbar (replace, download, alt text).
4. Confirm all three show the same styled tooltip with the same delay.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix inconsistent tooltip behavior on the video toolbar buttons.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Toolbar/DefaultVideoToolbarContent.tsx
+++ b/packages/tldraw/src/lib/ui/components/Toolbar/DefaultVideoToolbarContent.tsx
@@ -3,7 +3,6 @@ import { useCallback } from 'react'
 import { useActions } from '../../context/actions'
 import { useUiEvents } from '../../context/events'
 import { useTranslation } from '../../hooks/useTranslation/useTranslation'
-import { TldrawUiButton } from '../primitives/Button/TldrawUiButton'
 import { TldrawUiButtonIcon } from '../primitives/Button/TldrawUiButtonIcon'
 import { TldrawUiToolbarButton } from '../primitives/TldrawUiToolbar'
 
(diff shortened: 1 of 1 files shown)


danieljamesross (approved):

mimecuvalo:
thx!

PR #8799, opened 2026-05-08
perf(editor): skip rbush mutation on prop-only shape diffs

MitjaBezensek:
In order to avoid redundant work in the spatial index when shape props change without affecting bounds, this PR makes `SpatialIndexManager` skip rbush mutation on prop-only diffs and exposes a private `_boundsEpoch` that only ticks when the index actually changes.

While drawing on a populated page, every pointer move emits one updated shape (the active draw shape with new `segments`) through the store diff. Previously the index would unconditionally `upsert` the shape — an O(log N) tree rebalance — and bump its return value, invalidating downstream consumers like `getShapeIdsInsideBounds`. It also walked every indexed shape allocating a `Box` per shape during the transitive-bounds sweep.

Three changes:

1. `processIncrementalUpdate` compares incoming bounds to the indexed bounds and skips `upsert` when they're equal, removing the rebalance for prop-only updates.
2. The computed returns a private `_boundsEpoch` that only ticks when an add/remove/upsert actually happened, so consumers like `notVisibleShapes` no longer invalidate when nothing in the rbush moved. Bookkeeping is centralised in a small `rebuildAndBumpEpoch` helper.
3. The transitive-bounds sweep iterates the rbush's element map directly and compares against the raw `SpatialElement` instead of allocating a `Box` per shape and a fresh shape-id array. Existence checks use `RBushIndex.has(id)` instead of `getBounds(id) !== undefined`.

The transitive sweep itself still runs unconditionally — it has to, because some diffs (e.g. a geo's `rectangle`→`ellipse` flip at the same width/height) leave the source shape's axis-aligned bounds unchanged but shift a bound arrow's intersection points. There's a regression test covering exactly that case.

### Change type

- [x] \`improvement\`

### Test plan

1. Open a page with a few hundred shapes.
2. Start drawing a freehand stroke and move the pointer continuously.
3. Confirm no perf regression in interaction; existing culling behavior (shapes outside the viewport hidden) is unchanged when panning/zooming or when shapes are added, moved, or deleted.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improve drawing performance on pages with many shapes by skipping spatial index updates when only shape props change.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +100 / -28 |
| Tests     | +240 / -0  |

### DIFF
--- a/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
+++ b/packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
@@ -11,12 +11,16 @@ import { RBushIndex, type SpatialElement } from './RBushIndex'
  * Manages spatial indexing for efficient shape location queries.
  *
  * Uses an R-tree (via RBush) to enable O(log n) spatial queries instead of O(n) iteration.
- * Handles shapes with computed bounds (arrows, groups, custom shapes) by checking all shapes'
- * bounds on each update using the reactive bounds cache.
+ * Handles shapes with computed bounds (arrows, groups, custom shapes) by re-checking
+ * indexed shapes' bounds on every incremental update.
  *
  * Key features:
  * - Incremental updates using filterHistory pattern
- * - Leverages existing bounds cache reactivity for dependency tracking
+ * - Exposes a `_boundsEpoch` that ticks only when the rbush actually changes,
+ *   so downstream computeds (e.g. notVisibleShapes) can skip work on prop-only
+ *   shape changes that don't move any bounds
+ * - Step-1 upserts compare against the indexed bounds and skip rbush mutation
+ *   when bounds are unchanged, so prop-only updates are cheap
  * - Works with any custom shape type with computed bounds
  * - Per-page index (rebuilds on page change)
  * - Optimized for viewport culling queries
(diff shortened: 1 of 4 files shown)


steveruizok:
Hm, I can't find any differences here in the flame charts when working between this PR and canary. Could you double check the steps to test?

MitjaBezensek:
> Hm, I can't find any differences here in the flame charts when working between this PR and canary. Could you double check the steps to test?

Yeah, on its own this one won't move the flame chart much, the real benefit is together with #8804. They need to land together.

notVisibleShapes invalidates via two paths:

1. getCurrentPageShapes() per-shape subs (fixed in #8804)
2. spatialIndex bumping on every diff (fixed here)

Either path alone keeps not visible shape recalculation on every prop change. One way to see this PR's effect in isolation: subscribe to the spatial index signal and see how often updates. It should not update on prop only changes.

```ts
react('spatial index updates', () => {
    editor.getShapeIdsInsideBounds(editor.getViewportPageBounds())
    console.log("spatial index updated")
})
  ```

File: packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts [hunk 1]
```diff
@@ -126,35 +133,52 @@ export class SpatialIndexManager {
...
+				this.rbush.remove(shapeId)
 			}
+			changed = true
 		}
+
+		return changed
 	}
 
-	private areBoundsEqual(a: Box | undefined, b: Box | undefined): boolean {
+	private areBoundsEqualToElement(a: Box | undefined, b: SpatialElement | undefined): boolean {
```
mimecuvalo: tiniest nit, you can ignore: Element slightly throws me off b/c of the DOM-connotation. maybe consider areBoundsEqualToSpatialElement, but meh it's also fine as is.

mimecuvalo (approved):
LG, nice one @MitjaBezensek

PR #8800, opened 2026-05-08
fix(release): compare against latest published version in patch-release diff check

mimecuvalo:
In order to stop a half-finished previous patch release from permanently wedging a release branch's CI, this PR fixes `getAnyPackageDiff` so it compares each local package against the version that's actually published on npm rather than against `pkg.version` from the local `package.json`.

The patch-release workflow on `v4.0.x` and `v4.5.x` is currently broken in exactly this way (e.g. [run 25551498755](https://github.com/tldraw/tldraw/actions/runs/25551498755/job/75000238224) and [run 25551566270](https://github.com/tldraw/tldraw/actions/runs/25551566270/job/75000520586), both dying with `Package not found at url https://registry.npmjs.org/@tldraw/assets/-/assets-X.Y.Z.tgz: 404`). In both branches a previous run pushed the `v.X.Y.Z [skip ci]` version-bump commit but never finished `npm publish`, so the local `package.json`s are ahead of npm. The diff check then asks npm for the local-but-unpublished version, gets a 404, and crashes — and every subsequent push to that branch hits the same wall.

The fix has three parts:

- `getAnyPackageDiff(publishedVersion)` now takes a required version string and forwards it to `getPackageFirstDiff`, which uses it to build the npm tarball URL instead of `pkg.version`.
- A 404 from npm is no longer fatal — it's interpreted as "this package isn't published at this version yet" and returned as an `added` diff. This also handles the brand-new-monorepo-package case the previous code couldn't.
- Both callers (`publish-patch.ts`, `trigger-sdk-hotfix.ts`) pass in the latest npm version they already look up themselves (`latestVersionInBranch.format()` and `version.format()` respectively).

This unblocks the diff check in the broken-branch case. (Recovering the actual `v4.0.x` / `v4.5.x` releases will still need a one-off cleanup of the orphaned `[skip ci]` bump commits/tags — that's a separate follow-up.)

### Change type

- [x] `bugfix`

### Test plan

1. Verify the next patch-release workflow run on `v4.0.x` / `v4.5.x` gets past the diff check (it'll fetch `@tldraw/assets@4.0.3` / `@tldraw/assets@4.5.10` instead of the missing `4.0.4` / `4.5.11`).
2. Confirm normal (non-broken) release branches still publish unchanged — the comparison version is the same as `pkg.version` in steady state, so behaviour is identical.

### Code changes

| Config/tooling  | +30 / -9   |

### DIFF
--- a/internal/scripts/lib/didAnyPackageChange.ts
+++ b/internal/scripts/lib/didAnyPackageChange.ts
@@ -5,6 +5,7 @@ import kleur from 'kleur'
 import * as tar from 'tar'
 import tmp from 'tmp'
 import { exec } from './exec'
+import { nicelog } from './nicelog'
 import { PackageDetails, getAllPackageDetails } from './publishing'
 
 async function getPackageFirstDiff(
@@ -25,6 +26,9 @@ async function getPackageFirstDiff(
 			// commit landed but before npm publish completed. In either case
 			// treat it as an added package so the release proceeds instead of
 			// failing the entire workflow.
+			nicelog(
+				`No published tarball for ${pkg.name}@${publishedVersion} (${url}); treating as added.`
+			)
 			return { type: 'added', packageName: pkg.name, filePath: pkg.name }
 		}
 		if (res.status >= 400) {
--- a/internal/scripts/lib/didAnyPackageChange.ts
+++ b/internal/scripts/lib/didAnyPackageChange.ts
@@ -7,18 +7,28 @@ import tmp from 'tmp'
 import { exec } from './exec'
 import { PackageDetails, getAllPackageDetails } from './publishing'
 
-async function getPackageFirstDiff(pkg: PackageDetails): Promise<Diff | null> {
+async function getPackageFirstDiff(
+	pkg: PackageDetails,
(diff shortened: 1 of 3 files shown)


internal/scripts/lib/didAnyPackageChange.ts
```diff
@@ -7,18 +7,28 @@ import tmp from 'tmp'
...
 	const dir = tmp.dirSync({ unsafeCleanup: true })
 	const dirPath = dir.name
 	try {
-		const version = pkg.version
-
 		const unscopedName = pkg.name.replace('@tldraw/', '')
-		const url = `https://registry.npmjs.org/${pkg.name}/-/${unscopedName}-${version}.tgz`
+		const url = `https://registry.npmjs.org/${pkg.name}/-/${unscopedName}-${publishedVersion}.tgz`
 		const res = await fetch(url)
+		if (res.status === 404) {
```
MitjaBezensek, line 22: Should we add a log here, just for possible debugging purposes.

MitjaBezensek (approved):

PR #8804, opened 2026-05-08
perf(editor): tier notVisibleShapes per-shape subscriptions

MitjaBezensek:
In order to stop the canvas culling derivation from invalidating on every shape prop change, this PR makes `notVisibleShapes` skip per-shape subscriptions for shapes whose `ShapeUtil` uses the default `canCull` (which is always-true).

Previously the derivation read `editor.getCurrentPageShapes()`, which calls `store.get(id)` for every shape on the page. Each shape record became a parent of the derivation, so any prop change anywhere — even an onscreen color flip during a draw stroke — invalidated `notVisibleShapes` and cascaded to `CullingController`, `Shape` consumers, and `PerformanceManager`.

The new flow:

1. Subscribe to `getCurrentPageShapeIds()` (a Set; only changes on add/remove/reparent) instead of `getCurrentPageShapes()`.
2. For each offscreen id, peek the record via `store.unsafeGetWithoutCapture` to get `type` without subscribing.
3. If `util.canCull === ShapeUtil.prototype.canCull` (the default), push the id with no per-shape subscription.
4. Only when a shape's util overrides `canCull` do we call `editor.getShape(id)` (which subscribes) and invoke `canCull(shape)`.

Across the repo only two real overrides of `canCull` exist (`ConditionalCullingExample` and `SizeFromDomExample`), plus test-only shapes — so >99% of shapes in the wild now contribute zero per-shape subscriptions to this derivation.

### Findings: interaction with #8799

`notVisibleShapes` had two propagation paths that fired on any shape change:

1. `getCurrentPageShapes()` → `notVisibleShapes` (5000 per-shape `store.get` subs)
2. `spatialIndex` → `notVisibleShapes` (via `getShapeIdsInsideBounds` — bumped its computed value on every shape diff)

Either path alone is sufficient to invalidate the derivation. Benchmark on a 5000-shape page (500 onscreen, 4500 offscreen, 60 onscreen prop flips), counting `notVisibleShapes` body executions:

| Configuration | body runs |
| --- | --- |
| main | 60 |
| this PR alone | 60 (spatial-index path still propagates) |
| #8799 alone | 60 (`getCurrentPageShapes` path still propagates) |
| this PR + #8799 | **0** |

This PR and #8799 are complementary. Together they drop body runs to zero on prop-only changes; either alone leaves the other path firing. Read-time average for `getNotVisibleShapes()` over 60 iterations on the combined config is ~620ms vs ~687ms on main (~10% faster); standalone this PR is a smaller ~3-7% win driven by removing per-shape capture overhead.

### Change type

- [x] `improvement`

### Test plan

1. Open the conditional culling example, scroll a `preventCulling` shape offscreen, toggle the prop, confirm it remains visible/culled correctly.

- [x] Unit tests

### Release notes

- Reduce `notVisibleShapes` recomputation cost by skipping per-shape subscriptions for shapes with default `canCull`.

### Code changes

| Core code       | +30 / -21  |
| Tests           | +39 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
+++ b/packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
@@ -1,6 +1,7 @@
 import { computed, isUninitialized } from '@tldraw/state'
 import { TLShape, TLShapeId } from '@tldraw/tlschema'
 import type { Editor } from '../Editor'
+import { ShapeUtil } from '../shapes/ShapeUtil'
 
 /**
  * Non visible shapes are shapes outside of the viewport page bounds.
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -10,44 +11,51 @@ import type { Editor } from '../Editor'
...
+
+			// Peek at the shape without subscribing — we only need its type to look up the util.
+			// Type is treated as immutable for a given id, so this is safe.
+			const peek = editor.store.unsafeGetWithoutCapture(id) as TLShape | undefined
+			if (!peek) continue
+			const util = editor.getShapeUtil(peek.type)
+
+			// If canCull is the default (always-true), skip per-shape subscription entirely.
+			// >99% of shapes hit this path in practice.
+			if (util.canCull === defaultCanCull) {
```
MitjaBezensek, line 41: Unsure if this is a good idea 🤷‍♂️

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -10,44 +11,51 @@ import type { Editor } from '../Editor'
...
 
-		// First run: compute from scratch
-		if (isUninitialized(prevValue)) {
-			const nextValue = new Set<TLShapeId>()
-			for (let i = 0; i < allShapes.length; i++) {
-				shape = allShapes[i]
-				if (visibleIds.has(shape.id)) continue
-				if (!editor.getShapeUtil(shape.type).canCull(shape)) continue
-				nextValue.add(shape.id)
+		const notVisibleIds: TLShapeId[] = []
```
mimecuvalo: you're doing a `Set` down below - why not just do a `Set` here? esp since visibleIds is a Set

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
```diff
@@ -10,44 +11,51 @@ import type { Editor } from '../Editor'
...
+		for (const id of allShapeIds) {
+			if (visibleIds.has(id)) continue
+
+			// Peek at the shape without subscribing — we only need its type to look up the util.
+			// Type is treated as immutable for a given id, so this is safe.
+			const peek = editor.store.unsafeGetWithoutCapture(id) as TLShape | undefined
```
mimecuvalo: unnecessary type cast

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
mimecuvalo: i think it make sense

mimecuvalo (approved):
LGTM - couple minor notes

packages/editor/src/lib/editor/derivations/notVisibleShapes.ts
MitjaBezensek: Yeah, nice catch 👍

PR #8819, opened 2026-05-11
fix(dotcom): unblock preview deploys (DO migrations + vercel --yes)

MitjaBezensek:
This PR fixes two distinct breakages in dotcom preview deploys.

## 1. Sync worker durable-object migrations

In order to make preview builds of the sync worker succeed, this PR overrides the durable-object migration list at the `env.preview` level so preview workers skip the v10 `delete_classes` migration for `TLDrawDurableObject`.

Background:
- v1 created `TLDrawDurableObject`. v10 deletes it.
- Fresh preview workers (`pr-XXXX-tldraw-multiplayer`) have no prior script version, so Cloudflare rejects v10 with code 10074: `Cannot apply delete-class migration to class 'TLDrawDurableObject' which was not exported in the previous version of the script`.
- Staging and production already have v10 applied; they keep the top-level migration list unchanged.

#8638 attempted to drop v10 outright, which would have violated the append-only invariant (and would orphan the migration tag staging/prod already recorded). This PR instead adds an `[[env.preview.migrations]]` block: per Cloudflare's docs, env-level migrations fully replace the top-level list, so this is a preview-only override.

The preview block also skips v1, since the only point of creating `TLDrawDurableObject` was to delete it in v10. The leftover `export class TLDrawDurableObject {}` stub in `worker.ts` is an unbound class export with no migration record on preview, which CF treats as a no-op (verified by this PR's own preview deploy).

Relates to #8638 (closed).

## 2. Vercel CLI `--yes` flag

After fixing (1), the deploy job still failed at the `vercel alias set` step with `Error: unknown or unexpected option: --yes` ([failed run](https://github.com/tldraw/tldraw/actions/runs/25655757937/job/75303692969)).

Cause: `internal/scripts/deploy-dotcom.ts` had a `vercelCli` helper that unconditionally injected `--yes` into every subcommand. Per [Vercel's docs](https://vercel.com/docs/cli/alias), `--yes` is per-subcommand, not a global flag — valid for `vercel link`, `vercel deploy`, and `vercel alias rm`, but never for `vercel alias set` / `alias ls`. The injection was silently tolerated for a long time because the Vercel CLI parsed `alias` permissively; [vercel/vercel#14881](https://github.com/vercel/vercel/pull/14881) removed `permissive: true` from leaf commands so unknown flags now hard-error.

Fix: drop `--yes` from the generic helper and pass it explicitly only at the `link` and `deploy` call sites.

### Change type

- [x] `bugfix`

### Test plan

1. Confirm preview deploy of this PR succeeds end-to-end: sync worker uploads cleanly, and `vercel alias set` no longer rejects `--yes`.
2. Confirm staging/production deploys still apply v1-v10 from the top-level migration list and still skip prompts via the explicit `--yes` on `vercel link` and `vercel deploy`.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Apps           | +50 / -4   |
| Config/tooling | +9 / -3    |

### DIFF
--- a/apps/dotcom/sync-worker/wrangler.toml
+++ b/apps/dotcom/sync-worker/wrangler.toml
@@ -8,9 +8,11 @@ port = 8787
 ip = "0.0.0.0"
 
 # these migrations are append-only. you can't change them. if you do need to change something, do so
-# by creating new migrations
+# by creating new migrations.
+# IMPORTANT: when adding a new migration here, also add it to the [[env.preview.migrations]] block below
+# env-level migrations fully replace the top-level list, so preview won't inherit new entries.
 [[migrations]]
-tag = "v1" # Should be unique for each entry
+tag = "v1"                            # Should be unique for each entry
 new_classes = ["TLDrawDurableObject"]
 
 [[migrations]]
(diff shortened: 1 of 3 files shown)


apps/dotcom/sync-worker/wrangler.toml
```diff
@@ -8,9 +8,11 @@ port = 8787
 ip = "0.0.0.0"
 
 # these migrations are append-only. you can't change them. if you do need to change something, do so
-# by creating new migrations
+# by creating new migrations.
+# IMPORTANT: when adding a new migration here, also add it to the [[env.preview.migrations]] block below
+# env-level migrations fully replace the top-level list, so preview won't inherit new entries.
 [[migrations]]
-tag = "v1" # Should be unique for each entry
+tag = "v1"                            # Should be unique for each entry
```
MitjaBezensek, line 15: I guess formatting made these chages.

internal/scripts/deploy-dotcom.ts
```diff
@@ -670,7 +677,6 @@ async function vercelCli(command: string, args: string[], opts?: ExecOpts) {
 			env.VERCEL_TOKEN,
 			'--scope',
 			env.VERCEL_ORG_ID,
-			'--yes',
```
MitjaBezensek, line 673: Looks like vercel is more strict in the newer vercel cli versions - we bumped it a while back since the one we were using was no longer supported. It rejected some of the cli commands that used --yes, but that don't really support that option.

kostyafarber (approved):

PR #8824, opened 2026-05-11
fix(tldraw): catch image.decode() errors in asset urls provider

danieljamesross:
In order to silence noisy unhandled promise warnings from the icon preload effect, this PR catches rejections from `image.decode()` in `AssetUrlsProvider`. The preload is best-effort — any real load failure will still surface where the icon is actually used.

`image.decode()` returns a promise that can reject (e.g. with `EncodingError` when the browser cannot decode the image, or `AbortError` when the source changes). Without a `.catch()` handler, these rejections become uncaught promise errors in the console.

### Change type

- [x] `bugfix`

### Test plan

1. Open the editor in a browser
2. Confirm no `Uncaught (in promise) EncodingError` warnings appear in the console from the asset URLs provider

### Release notes

- Silently catch `image.decode()` rejections from the icon preload effect.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +7 / -2    |

### DIFF
--- a/packages/tldraw/src/lib/ui/context/asset-urls.tsx
+++ b/packages/tldraw/src/lib/ui/context/asset-urls.tsx
@@ -1,4 +1,4 @@
-import { Image } from '@tldraw/editor'
+import { Image, noop } from '@tldraw/editor'
 import { createContext, useContext, useEffect } from 'react'
 import { TLUiAssetUrls } from '../assetUrls'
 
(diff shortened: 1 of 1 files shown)


packages/tldraw/src/lib/ui/context/asset-urls.tsx
```diff
@@ -7,6 +7,11 @@ type UiAssetUrlsContextType = TLUiAssetUrls | null
 
 const AssetUrlsContext = createContext<UiAssetUrlsContextType>(null)
 
+function noop() {
```
mimecuvalo: funny enough we have a `noop` function already in the library :P (utils/function.ts)

mimecuvalo (approved):
pre-approving module one cleanup

packages/tldraw/src/lib/ui/context/asset-urls.tsx
danieljamesross: Well well! If only Clause was better at this!

danieljamesross: this is done

PR #8827, opened 2026-05-11
feat(tldraw): redesign page menu

### LINKED ISSUE #8814: Remove edit mode from the page menu
## Problem

The page menu has a pencil-icon toggle that switches the menu into a separate "edit mode". Drag handles for reordering and the inline rename input are only visible while edit mode is active. This causes several UX problems:

1. **Drag-to-reorder is undiscoverable.** Users have to find and press the pencil button before they ever see a drag handle. Most won't.
2. **Edit mode is mostly redundant.** The per-row submenu already exposes rename, move-up, move-down, duplicate, and delete. The only behavior unique to edit mode is drag reordering and inline rename.
3. **Renaming a non-current page secretly navigates to it.** When you rename via the submenu, the code switches the current page to the one being renamed (`DefaultPageMenu.tsx:486-489`). This exists because `PageItemInput` is only rendered for the current row in edit mode. Decoupling rename from edit mode removes the need for the navigation side effect.
4. **The header is cramped.** "Pages" title + edit toggle + add button compete for space.

## Proposal

Remove the edit-mode toggle entirely.

- Always show the drag handle (on hover, or always visible) so reordering is discoverable.
- Allow inline rename in place via double-click on a row, or Enter on a focused row, with no mode switch.
- Drop the pencil button from the header — it becomes "Pages" + add button only.
- Remove the rename navigation side effect — renaming a page should never change the current page.

## Files involved

- `packages/tldraw/src/lib/ui/components/PageMenu/DefaultPageMenu.tsx` — owns `isEditing` state and the conditional rendering split.
- `packages/tldraw/src/lib/ui/components/PageMenu/PageItemSubmenu.tsx` — the `onRename` callback that triggers navigation.
- `packages/tldraw/src/lib/ui/components/PageMenu/PageItemInput.tsx` — currently only rendered in edit mode.

## Notes

- Scoped to the page menu component. No public API changes expected.
- The `shouldUseWindowPrompt` iOS Safari fallback for renames may need to be revisited as part of the inline-rename redesign.
- Consider as a follow-up: keyboard reorder support, since the existing pointer-based drag has no a11y story.

steveruizok:
The page menu used to require flipping into an explicit edit mode to rename pages or reorder them. This PR removes that mode entirely. Reordering happens by dragging the row itself, renaming happens inline (double-click, Enter, or the row submenu), and the popover gains a draggable resize handle so the list height can be adjusted and persisted across sessions.

Closes #8814.

### Interactions

| Interaction        | Before                                     | After                                       |
| ------------------ | ------------------------------------------ | ------------------------------------------- |
| Reorder a page     | Toggle Edit, drag the dedicated handle     | Drag the row itself                         |
| Rename a page      | Toggle Edit, click the field               | Double-click the row, press Enter, or use the kebab submenu |
| Current page mark  | Checkmark icon                             | Subtle background pill (matches dotcom sidebar) |
| Submenu trigger    | Always visible                             | Hidden until row hover (and hidden while renaming or dragging) |
| List height        | Fixed                                      | Drag the resize handle to adjust; double-click it to reset to default. Persisted to localStorage. |
| Create page button | Header                                     | Pinned to the bottom of the popover         |

Renaming a non-current page no longer silently navigates to it. The rename input matches the row's vertical metrics so the label doesn't shift when editing, and blur commits the change.

### Change type

- [x] `improvement`

### Test plan

1. Open the page menu with multiple pages.
2. Drag a row by its body — it should reorder without entering any mode.
3. Double-click a row's label (or press Enter / use the kebab menu) — it should rename inline. Blur should commit.
4. Confirm the current page shows the highlight pill and renaming a non-current page does not navigate.
5. Drag the resize handle at the bottom of the popover up and down; reopen the menu and confirm the height is remembered.
6. Double-click the resize handle — the list should snap back to its default height and forget the saved preference.
7. Open the page menu when the current page is far down the list — it should scroll into view.

- [x] End to end tests

### Release notes

- Redesigned the page menu: reorder rows by dragging them directly, rename inline by double-clicking or pressing Enter, and resize the page list to remember your preferred height (double-click the resize handle to reset it). The explicit edit mode has been removed.

### API changes

- Added `page-menu.resize` translation key for the resize handle's accessible label.
- Removed unused `page-menu.max-page-count-reached` translation key.

### Code changes

| Core code       | +499 / -445  |
| Tests           | +8 / -23     |
| Automated files | +1 / -1      |
| Apps            | +1 / -1      |


steveruizok:
Superseded by #8836 — a clean-history reimplementation of this branch as three narrative commits. Diff is byte-identical; please review there.

PR #8837, opened 2026-05-12
feat: add soft clipping region example

angrycaptain19:
Adds a use-cases example where a regular geo shape acts as a "soft" clipping region. Everything outside of the shape is dimmed.


mimecuvalo (approved):
very cool. sry for the delay - LGTM!

PR #8841, opened 2026-05-13
explore: lightweight read-only editor for grid previews

MitjaBezensek:
Exploration on a lightweight tldraw editor for rendering many read-only document previews on one page.

Adds a standalone `TldrawViewer` (no UI, tools, events, watermark, or theme/camera sync) and a `TldrawPreview` wrapper that pre-wires the default shape, binding, and asset utils. Backs a `/grid` page that fetches a snapshot per file and renders a tile, with a `/grid-original` baseline for comparison.

Also includes a server-side bounds filter on `/app/file/:slug/download?bounds=…` that drops shapes outside a per-document preview viewport (with arrow / binding-target handling for arc and elbow arrows), and a `BulkShapeCreator` button + per-doc `meta.previewViewport` for setting the bounds.

Drafted as exploration — not aimed at landing as-is.

Comparison of loading the readonly grid vs the original (full editor)

https://github.com/user-attachments/assets/86a0aa93-c4d1-483e-820e-333c6aab04b9

### DIFF
--- a/apps/dotcom/client/src/routeDefs.ts
+++ b/apps/dotcom/client/src/routeDefs.ts
@@ -4,6 +4,8 @@ export const ROUTES = {
 	tlaOptIn: '/preview',
 
 	tlaRoot: `/`,
+	tlaGrid: `/grid`,
+	tlaGridOriginal: `/grid-original`,
 	tlaNew: `/new`,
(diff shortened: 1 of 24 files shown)


MitjaBezensek:
Part of the hackaton, closing for now.

PR #8852, opened 2026-05-15
feat(tldraw): resizable color grid in style panel

steveruizok:
In order to support customizing the color palette beyond the built-in 12 colors, this PR adds a vertically scrollable, resizable color grid to the style panel. When the palette contains more than 12 colors, the grid clamps to a default height with hidden scrollbars and a separator below the opacity picker becomes a resize handle — drag to grow the grid, double-click to reset. The chosen height persists per browser via `localStorage`.

The default theme is unchanged. Consumers opt in by registering extra colors via the `themes` prop (see the new example).

### New examples

- `apps/examples/src/examples/ui/color-palette/` — adds 12 extra colors via the `themes` prop and `TLThemeDefaultColors` augmentation; demonstrates the scroll/resize behaviour.

### Change type

- [x] `feature`

### Test plan

1. Run `yarn dev` and open the "Color palette" example.
2. Select a shape and confirm the style panel shows 24 colors with the grid clamped to its default height.
3. Drag the separator below the opacity slider downward and release — the grid grows.
4. Reload the page and confirm the height persists.
5. Double-click the separator and confirm the grid returns to its default height.
6. Open another example (e.g. the basic one); the separator should not appear because the palette has only the 12 default colors.
7. Hover the color grid and scroll the wheel — the grid scrolls, the page does not.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add a resizable, scrollable color grid to the style panel for palettes with more than 12 colors.

### API changes

- Added the `style-panel.resize` translation key.

### Code changes

| Core code       | +160 / -4  |
| Automated files | +3 / -1    |
| Documentation   | +135 / -0  |
| Config/tooling  | +1 / -0    |


steveruizok:
I'm hesitant to add lots of new colors to anything with the current "hardcode all variants" strategy, though I think we could do it on tldraw.com. I'd also like to look at creating those variants based on some kind of logic that we pull from our hardcoded examples.

mimecuvalo:
@steveruizok i think i'm really thrown off that the resize divider is the grey line _below_ the opacity slider. 
i'd expect it to be the one above the opacity slider. i think that speaks to a resizable divider needing a different UX treatment, otherwise it's super hidden, prbly also an a11y footgun.

mimecuvalo:
adding #needs-design-review

PR #8860, opened 2026-05-18
feat(editor): add selectLockedShapes option

### LINKED ISSUE #8548: Add option to allow selecting locked shapes with left click
## Problem

Currently, locked shapes can only be selected by right-clicking them. Left-clicking a locked shape is treated as clicking the canvas — the shape is completely invisible to the hit test. This makes it difficult to inspect or interact with locked shapes without using the context menu.

## Current behavior

- **Left click**: `getHitShapeOnCanvasPointerDown` calls `getShapeAtPoint` without `hitLocked: true`, so locked shapes are not found. Even if a locked shape is hit via `info.target === 'shape'`, the `Idle` state checks `isShapeOrAncestorLocked` and transitions to `pointing_canvas`.
- **Right click**: `getShapeAtPoint` is called with `hitLocked: true`, so locked shapes are found and the context menu opens.
- **Select All**: Explicitly filters out locked shapes via `_getUnlockedShapeIds`.

## Proposed solution

Add an editor-level option (e.g. on `TldrawOptions`) that controls whether locked shapes can be selected via left click. When enabled:

1. Pass `hitLocked: true` in `getHitShapeOnCanvasPointerDown`
2. Remove or conditionalize the `isShapeOrAncestorLocked` guard in `Idle.onPointerDown`
3. Possibly still prevent moving/resizing locked shapes while allowing selection + inspection

This would let SDK consumers configure the locked shape behavior for their use case.

## Alternatives

- Override the `SelectTool` with a custom `Idle` state — works but requires duplicating significant logic
- Use right-click only — current behavior, but not discoverable

## Use cases

- Canvas viewers where locked shapes serve as background/template elements that users still need to inspect
- Apps where "locked" means "not movable" rather than "not selectable"
- Workflows where users need to read properties of locked shapes without unlocking them

steveruizok:
Adds a new `selectLockedShapes` option to `TldrawOptions` that, when enabled, lets users select locked shapes with a left click (and via brush or scribble selection). The editor's existing lock guards still prevent moving, resizing, editing, or deleting locked shapes, so the option enables inspection without unlocking.

When `selectLockedShapes: true`:
- `getHitShapeOnCanvasPointerDown` passes `hitLocked: true`
- `Idle.onPointerDown` no longer routes left clicks on locked shapes to `pointing_canvas`
- `Brushing` and `ScribbleBrushing` include locked shapes
- Hover detection (`updateHoveredShapeId`) treats locked shapes as hoverable
- `selectOnCanvasPointerUp` allows locked shapes through

When `selectLockedShapes` is `false` (default), behavior is unchanged.

Closes #8548

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Set `<Tldraw options={{ selectLockedShapes: true }} />` in an example app.
2. Create a shape, lock it, and click it — it should become selected and show selection bounds.
3. Try to drag or resize it — nothing should move (lock still applies).
4. Brush over the locked shape with the select tool — it should be included in the brush selection.
5. Alt-drag (scribble select) over the locked shape — it should also be included.
6. With the option set to `false` (default), confirm left-clicking a locked shape still falls through to the canvas, matching the previous behavior.

- [x] Unit tests
- [ ] End to end tests

 ### API changes

  - Add a new `selectLockedShapes` option to `TldrawOptions` (and to `defaultTldrawOptions`). When `false` (default), left-clicking a locked shape is treated as a click on the canvas; only right-click selects it. When `true`, locked shapes can be selected via left-click and included in brush and scribble selections, but the editor's lock guards still prevent moving, resizing, editing, or deleting them.

### Release notes

- Added a `selectLockedShapes` option to `TldrawOptions`. When enabled, locked shapes can be selected with left click and brush/scribble selection while remaining protected from edits, moves, and deletes.

### DIFF
--- a/apps/examples/src/examples/editor-api/locked-shapes/LockedShapesExample.tsx
+++ b/apps/examples/src/examples/editor-api/locked-shapes/LockedShapesExample.tsx
@@ -1,3 +1,4 @@
+import { useState } from 'react'
 import { createShapeId, Tldraw, TldrawUiButton, TLShapeId, toRichText, useEditor } from 'tldraw'
 import 'tldraw/tldraw.css'
 
@@ -13,7 +14,20 @@ const TEMPLATE_IDS: TLShapeId[] = [
 function ControlPanel() {
 	const editor = useEditor()
 
-	// [3] Update locked shapes using ignoreShapeLock option
+	// [3] Local state mirrors editor.options.selectLockedShapes for the UI.
+	// The option is readonly at the type level but the editor stores a copy
+	// internally that the SelectTool reads live on every interaction — so
+	// mutating the underlying field flips behaviour immediately, without a
+	// remount.
+	const [selectLocked, setSelectLocked] = useState(editor.options.selectLockedShapes)
+
+	const toggleSelectLocked = () => {
+		const next = !selectLocked
+		;(editor.options as { selectLockedShapes: boolean }).selectLockedShapes = next
+		setSelectLocked(next)
+	}
+
+	// [4] Update locked shapes using ignoreShapeLock option
 	// Without ignoreShapeLock: true, these updates would be blocked
 	const handleScatter = () => {
 		editor.run(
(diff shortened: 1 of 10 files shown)


mimecuvalo:
psst, i think we should make this the default!

PR #8867, opened 2026-05-19
chore(dotcom): increase OTEL metric export interval for zero-cache

MitjaBezensek:
In order to reduce the volume of OTEL metric exports from the zero-cache Fly.io services, this PR sets `OTEL_METRIC_EXPORT_INTERVAL` to `120000` (2 minutes) in the three zero-cache Fly templates (single, replication-manager, view-syncer). Default is 60s; this halves the export frequency.

### Change type

- [x] `other`

### Test plan

1. Deploy zero-cache to staging.
2. Confirm metrics still arrive in the OTEL backend, just at the new 2-minute cadence.

### Code changes

| Apps           | +3 / -0    |


MitjaBezensek:
Grafana costs down quite a bit already, will close this as not needed. It's nice to have more granular data, also means we don't need to figure out alerting changes.

PR #8897, opened 2026-05-20
feat(editor): simplify multi-click handling to double-click

steveruizok:
This simplifies tldraw's multi-click handling. Previously `ClickManager` tracked double, triple, and quadruple clicks, each emitting its own event with `down`, `up`, and `settle` phases. In practice only double-click and the hand tool's zoom shortcuts used the higher counts, and the `settle` phase couldn't tell a multi-click that finished with the pointer held apart from one that finished with it released.

This PR:

- Removes `triple_click` and `quadruple_click` entirely. `TLCLickEventName` is now just `'double_click'`, and the `onTripleClick` / `onQuadrupleClick` handlers are gone from `TLEventHandlers` and `StateNode`.
- Splits the `settle` phase into `settle-down` and `settle-up`. `ClickManager` tracks whether the pointer is pressed when the click timer fires and emits the matching phase.
- Fires double-click-to-edit on the `down` phase. Select tool child states (`Idle`, `PointingShape`, `PointingCanvas`, `PointingHandle`, `PointingSelection`, and the crop states) now act on the second pointer-down instead of waiting for pointer-up, so editing and handle behavior trigger immediately.
- Updates `HandTool` to keep only double-click zoom-in (on `settle-up`); triple-click zoom-out and quadruple-click reset/zoom-to-fit are removed.

### API changes

- Breaking! `TLCLickEventName` no longer includes `'triple_click'` or `'quadruple_click'`; only `'double_click'` remains.
- Breaking! `onTripleClick` and `onQuadrupleClick` are removed from `TLEventHandlers` and `StateNode`.
- Breaking! `TLClickEventInfo.phase` replaces `'settle'` with `'settle-down'` and `'settle-up'`. Consumers that checked `phase === 'settle'` should switch to `'settle-up'` (or branch on both if they care whether the pointer is still pressed).
- `TLClickState` drops `'pendingTriple'` and `'pendingQuadruple'`.

### Change type

- [x] `api`

### Test plan

1. With the hand tool, double-click the canvas and release — should zoom in.
2. With the select tool, double-click a shape — should enter editing immediately on the second press.
3. Confirm triple- and quadruple-clicks no longer zoom out or reset zoom with the hand tool.

- [x] Unit tests
- [x] End to end tests

### Release notes

- Simplified multi-click handling: triple- and quadruple-click events are removed, and the `settle` click phase is split into `settle-down` and `settle-up` so tools can tell whether a multi-click finished with the pointer held or released.

### Code changes

| Section          | LOC change   |
| ---------------- | ------------ |
| Core code        | +157 / -112  |
| API report       | +3 / -15     |
| Tests and infra  | +107 / -141  |
| Documentation    | +55 / -63    |


packages/editor/src/lib/editor/managers/ClickManager/ClickManager.ts
```diff
@@ -219,5 +226,6 @@ export class ClickManager {
 	cancelDoubleClickTimeout() {
 		this._clickTimeout = clearTimeout(this._clickTimeout)
 		this._clickState = 'idle'
+		this._isPressingWhilePending = false // maybe clear this!
```
max-drake: keep this line in (but remove the comment)
we don't want `_isPressingWhilePending` to be true if a double click is canceled for any reason, even if the user is still pressing, because there is no longer any pending next click

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -261,7 +261,7 @@ export class Idle extends StateNode {
 	}
 
 	override onDoubleClick(info: TLClickEventInfo) {
-		if (this.editor.inputs.getShiftKey() || info.phase !== 'up') return
+		if (this.editor.inputs.getShiftKey() || info.phase !== 'settle-up') return
```
max-drake: this change is causing a bunch of tests to fail because double clicking no longer happens immediately, so we should either adjust the tests to take into account the waiting-for-settle-up timing for select tool double clicks, or rethink this specific line

steveruizok: It's a weakness in our inputs system that handlers aren't set dynamically. If we knew that there was a triple click handler, then we would have to wait for the interaction to settle, to disambiguate the double- and triple-clicks, but since we don't know that then the system has to act as if there was a triple etc click handler.

We could just remove triple / quadruple clicks entirely, that would solve this. Most double click behaviors should on the second pointer down.

steveruizok:
Alright, let's forget about triple clicks and quadruple clicks. We were only using them for the map interactions which should soon be better served by @max-drake's click and hold. I'm not sure anyone ever used that part of our API anyway.

apps/examples/vercel.json
```diff
@@ -1,4 +1,7 @@
 {
+	"installCommand": "yarn workspaces focus examples.tldraw.com @tldraw/monorepo config",
+	"buildCommand": "cd ../../packages/tldraw && yarn prebuild && cd ../../apps/examples && yarn build",
+	"outputDirectory": "dist",
```
max-drake, line 4: is this in here intentionally?

packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
```diff
@@ -261,7 +261,7 @@ export class Idle extends StateNode {
 	}
 
 	override onDoubleClick(info: TLClickEventInfo) {
-		if (this.editor.inputs.getShiftKey() || info.phase !== 'up') return
+		if (this.editor.inputs.getShiftKey() || info.phase !== 'down') return
 
 		// We don't want to double click while toggling shapes
 		if (info.ctrlKey || info.shiftKey) return
```
max-drake, line 267: consolidate conditionals

max-drake:
Good faith approving in case I'm missing something, but I don't know how great I feel about the new requirement for leaf nodes whose `idle` state parents have double click handlers, where the leaves have to manually transition back to `idle` and call the parent's `handleEvent()` method in their own double click handlers.

idk if that description makes sense but it's not immediately obvious why all these

```ts
override onDoubleClick(info: TLClickEventInfo) {
      if (
	      this.editor.inputs.getShiftKey() ||
	      info.phase !== 'down' ||
	      info.ctrlKey ||
	      info.shiftKey
      ) {
	      return
      }
      
      this.parent.transition('idle')
      this.parent.getCurrent()?.handleEvent(info)
}
```

blocks exist. 

At the very least I think we should document this behavior a little better.

max-drake (approved):

apps/docs/content/sdk-features/events.mdx
MitjaBezensek, line 226: Outdated.

MitjaBezensek (approved):

PR #8904, opened 2026-05-20
chore(utils): replace deprecated lodash.isequal with lodash subpath imports

### LINKED ISSUE #8729: lodash.isequal is deprecated
As noted in the title, `lodash.isequal` is deprecated and throws a warning when the library is installed.
Consider using alternatives.

https://github.com/tldraw/tldraw/blob/main/packages/utils/package.json#L47

Image

Image

mimecuvalo:
The `lodash.isequal` package was deprecated by the lodash maintainers and now emits an install-time warning. This switches `@tldraw/utils` to import `isEqual` and `isEqualWith` from the main `lodash` package's subpath entries (`lodash/isEqual.js` and `lodash/isEqualWith.js`), which are not deprecated. The functions behave identically — same code, same lodash version.

I also dropped the now-unneeded `lodash.isequalwith` standalone subpackage since we're already pulling in the main `lodash` package.

Closes #8729

### Change type

- [x] `other` (chore: dep cleanup, no behavior change)

### Test plan

- [x] Unit tests — `packages/utils` tests pass (288/288), `packages/sync-core` tests pass (608/608)
- [x] Typecheck — `yarn typecheck` passes
- [x] API check — `yarn api-check` passes; `api-report.api.md` regenerated to reflect the new import paths

### Release notes

- Internal: removed deprecated `lodash.isequal` dependency from `@tldraw/utils`.

### DIFF
--- a/internal/scripts/api-check.ts
+++ b/internal/scripts/api-check.ts
@@ -6,10 +6,9 @@ import { nicelog } from './lib/nicelog'
 import { getAllWorkspacePackages } from './lib/workspace'
 
 const packagesOurTypesCanDependOn = [
+	'@types/lodash',
 	'@types/lodash.throttle',
 	'@types/lodash.uniq',
-	'@types/lodash.isequal',
-	'@types/lodash.isequalwith',
 	'@types/react',
 	'@types/react-dom',
 	'eventemitter3',
--- a/packages/utils/api-report.api.md
+++ b/packages/utils/api-report.api.md
@@ -4,8 +4,8 @@
 
 ```ts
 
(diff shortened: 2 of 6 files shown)


kaneel:
have we looked into possible alternative to lodash? such as es-toolkit's isEqual?

mimecuvalo:
Closing this — after looking more carefully, nothing is actually broken. `lodash.isequal` still works correctly in browsers; the deprecation is an install-time warning, and npm's suggested replacement (`node:util.isDeepStrictEqual`) doesn't apply to browser code.

PR #8905, opened 2026-05-20
fix(examples): update custom-theme at runtime via editor.updateTheme

mimecuvalo:
Adjusting a theme slider in the `custom-theme` example passed a brand-new `themes` object to `<Tldraw>` on every change. `useLocalStore` treats `themes` as part of the store options, so each new reference re-ran `createTLStore` and reloaded from persistence — the whole board visibly flashed on every slider drag.

This reworks the example to apply theme changes through the programmatic API instead of the prop:

- The `themes` prop is now a stable, module-level object passed once to `<Tldraw>`. It registers the custom `pink` color and custom fonts at store creation and seeds the starting values. Because its reference never changes, the store is never recreated.
- Slider adjustments are applied at runtime with `editor.updateTheme({ ...editor.getTheme('default')!, fontSize, lineHeight, strokeWidth })`, spreading the current theme so custom colors and fonts are preserved. Shapes re-render reactively without reloading the canvas.

No SDK changes — the fix is entirely in the example.

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn dev` and open http://localhost:5420/custom-theme.
2. Adjust any of the sliders (font size, line height, stroke width).
3. The shapes re-render with the new theme values without the canvas flashing or reloading.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Update the `custom-theme` example to apply runtime theme changes via `editor.updateTheme()` so adjusting theme display values no longer reloads the canvas.

### DIFF
--- a/apps/examples/src/examples/ui/custom-theme/CustomThemeExample.tsx
+++ b/apps/examples/src/examples/ui/custom-theme/CustomThemeExample.tsx
@@ -1,6 +1,7 @@
-import { useCallback, useMemo, useRef, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import {
 	DEFAULT_THEME,
+	Editor,
 	TLDefaultColor,
 	TLTheme,
 	TLThemeFont,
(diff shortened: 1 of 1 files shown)


steveruizok:
Themes can be changed at runtime through an editor API, `editor.updateTheme` or `editor.updateThemes`. We should have an example that shows this as the suggested route (if that's not how the example works, then that would be a good thing to change). IIRC we've guarded against these things for configuration that isn't possible via the editor API, but if theme changing via props is something that is unavoidable or a common problem, then this would work.

PR #8909, opened 2026-05-20
fix(tldraw): export TldrawUiTranslationProvider and dedupe missing-translation warning

### LINKED ISSUE #6236: [Bug]: Unable to use `TldrawSelectionForeground` without `TldrawUiContextProvider`
### What happened?

We use `TldrawSelectionForeground`, but not most of tldraw's other UI. With the upgrade to v3.13, wrapping our app in `TldrawUiContextProvider` became necessary in order to continue using `TldrawSelectionForeground`.

We'd prefer to avoid that because we don't need most of what `TldrawUiContextProvider` does. It seems to be pre-fetching icons that our application doesn't ever use.

From a quick look through the code, it seems like one option for us would be to use `TldrawUiTranslationProvider`, but that's not currently exported from the tldraw package.

_No response_

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

mimecuvalo:
In order to let consumers render components like `TldrawSelectionForeground` without pulling in the full `TldrawUiContextProvider` (and its icon preloading), this PR promotes `TldrawUiTranslationProvider`, `AssetUrlsProvider`, and `useAssetUrls` from `@internal` to `@public`. It also deduplicates the "No translation messages found" warning so it only logs once per session instead of once per component that calls `useTranslation`, and rewords the message to point developers at `TldrawUiTranslationProvider` as an alternative.

Closes #6236

Consumers can now do:

```tsx
<AssetUrlsProvider assetUrls={...}>
  <TldrawUiTranslationProvider locale="en">
    <TldrawSelectionForeground ... />
  </TldrawUiTranslationProvider>
</AssetUrlsProvider>
```

### Change type

- [x] `bugfix`

### Test plan

1. Render a component that uses `useTranslation` (e.g. `TldrawSelectionForeground`) outside of `TldrawUiContextProvider`.
2. Confirm the "No translation messages found" warning prints **once**, not once per consumer.
3. Wrap the same component in `<AssetUrlsProvider><TldrawUiTranslationProvider locale="en">` and confirm the warning is silenced and translations resolve.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Export `TldrawUiTranslationProvider`, `AssetUrlsProvider`, and `useAssetUrls` as public API so `TldrawSelectionForeground` and similar components can be used without `TldrawUiContextProvider`.
- Only log the missing-translation warning once per session instead of once per `useTranslation` consumer.

### API changes

- Changed `TldrawUiTranslationProvider` from `@internal` to `@public @react`.
- Changed `AssetUrlsProvider` from `@internal` to `@public @react`.
- Changed `useAssetUrls` from `@internal` to `@public`.

### Code changes

| Core code       | +18 / -6   |
| Automated files | +3 / -3    |

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
+++ b/packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
@@ -36,9 +36,11 @@ export function useCurrentTranslation() {
 }
 
 /**
- * Provides a translation context to the editor.
+ * Provides a translation context to the editor. Wrap this around components that use
+ * `useTranslation` (such as `TldrawSelectionForeground`) when you don't want to use the
+ * full `TldrawUiContextProvider`. Must be rendered inside an `AssetUrlsProvider`.
  *
- * @internal
+ * @public @react
  */
 export function TldrawUiTranslationProvider({
 	overrides,
@@ -97,6 +99,8 @@ export function TldrawUiTranslationProvider({
 	)
 }
 
+let hasWarnedAboutMissingTranslations = false
+
 /**
  * Returns a function to translate a translation key into a string based on the current translation.
  *
@@ -114,8 +118,11 @@ export function useTranslation() {
 	const messages = translation?.messages ?? DEFAULT_TRANSLATION
 
 	React.useEffect(() => {
-		if (!translation?.messages) {
-			console.warn('No translation messages found, falling back to default translation.')
+		if (!translation?.messages && !hasWarnedAboutMissingTranslations) {
+			hasWarnedAboutMissingTranslations = true
+			console.warn(
+				'No translation messages found, falling back to default translation. Wrap your app in <TldrawUiContextProvider> or <TldrawUiTranslationProvider> to provide translations.'
+			)
 		}
 	}, [translation?.messages])
 
--- a/packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
+++ b/packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
@@ -1,3 +1,4 @@
+import { warnOnce } from '@tldraw/editor'
 import * as React from 'react'
 import { useAssetUrls } from '../../context/asset-urls'
 import { DEFAULT_TRANSLATION } from './defaultTranslation'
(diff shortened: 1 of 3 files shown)


packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
```diff
@@ -114,8 +118,11 @@ export function useTranslation() {
 	const messages = translation?.messages ?? DEFAULT_TRANSLATION
 
 	React.useEffect(() => {
-		if (!translation?.messages) {
-			console.warn('No translation messages found, falling back to default translation.')
+		if (!translation?.messages && !hasWarnedAboutMissingTranslations) {
+			hasWarnedAboutMissingTranslations = true
+			console.warn(
+				'No translation messages found, falling back to default translation. Wrap your app in <TldrawUiContextProvider> or <TldrawUiTranslationProvider> to provide translations.'
+			)
```
MitjaBezensek, line 123: Should we also mention that we need `AssetUrlsProvider`?

packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
```diff
@@ -114,8 +118,11 @@ export function useTranslation() {
 	const messages = translation?.messages ?? DEFAULT_TRANSLATION
 
 	React.useEffect(() => {
-		if (!translation?.messages) {
-			console.warn('No translation messages found, falling back to default translation.')
+		if (!translation?.messages && !hasWarnedAboutMissingTranslations) {
```
MitjaBezensek: Think we also have a `warnOnce` util for this.

MitjaBezensek (approved):

packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
mimecuvalo: Good call — switched to `warnOnce` from `@tldraw/editor` in 814ef59.

packages/tldraw/src/lib/ui/hooks/useTranslation/useTranslation.tsx
mimecuvalo: Updated the warning message to mention `AssetUrlsProvider` in 814ef59 — it now reads: "Wrap your app in <TldrawUiContextProvider>, or in both <AssetUrlsProvider> and <TldrawUiTranslationProvider>, to provide translations."

PR #8911, opened 2026-05-20
fix(tldraw): show context menu in non-select tools

### LINKED ISSUE #8828: Right-click does not show the context menu outside the select tool
I noticed that when we right click on the canvas while having any other tool besides the `select` tool chosen then the cursor changes to the `select` tool pointer but the context menu is not shown.

Here is a video demonstrating it:

https://github.com/user-attachments/assets/40b25608-2c71-4022-b015-4efae3131118

Reading through the code, I realized this was happening due to a specific guard in `DefaultContextMenuContent.tsx` 

`if (!selectToolActive) return null`

The Context menu content is empty when it is opened from any other tool. I am curious to know if this is intentional or a bug? Would appreciate any insights. Thanks!

mimecuvalo:
In order to fix #8828, this PR removes the blanket `if (!selectToolActive) return null` guard at the top of `DefaultContextMenuContent`. Right-clicking the canvas while any non-select tool (draw, eraser, hand, etc.) was active opened an empty context menu. Each submenu inside the context menu already self-hides when not applicable — `EditMenuSubmenu`, `ArrangeMenuSubmenu`, and `ReorderMenuSubmenu` short-circuit when there is no selection; `ConversionsMenuGroup` short-circuits when there are no shapes on the page; `SelectAllMenuItem` disables itself when the page is empty. With the outer guard removed, the menu now renders in any tool and useful items like Select All, Paste, and the export/conversion actions are reachable without first switching to the select tool.

### Change type

- [x] `bugfix`

### Test plan

1. Run the examples app and switch to any non-select tool (draw, eraser, hand, etc.).
2. Right-click on the canvas.
3. The context menu should open with the items that apply to the current state (e.g. Select All, Paste, and any export/conversion actions for shapes on the page).
4. Add a shape, switch back to a non-select tool, right-click on the shape, and confirm the appropriate per-selection items also appear.

- [x] Unit tests

### Release notes

- Fixed the context menu being empty when right-clicking the canvas while any non-select tool was active.

### API changes

- `DefaultContextMenuContent` return type narrowed from `JSX.Element | null` to `JSX.Element` (it no longer returns `null`).

### Code changes

| Core code       | +0 / -7    |
| Tests           | +17 / -0   |

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -1126,7 +1126,7 @@ export { DefaultContextMenu as ContextMenu }
 export { DefaultContextMenu }
 
 // @public (undocumented)
-export function DefaultContextMenuContent(): JSX.Element | null;
+export function DefaultContextMenuContent(): JSX.Element;
 
 // @public (undocumented)
 export function DefaultDebugMenu({ children }: TLUiDebugMenuProps): JSX.Element;
--- a/packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenuContent.tsx
+++ b/packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenuContent.tsx
@@ -17,16 +17,9 @@ export function DefaultContextMenuContent() {
 	const editor = useEditor()
 	const showCollaborationUi = useShowCollaborationUi()
 
-	const selectToolActive = useValue(
-		'isSelectToolActive',
-		() => editor.getCurrentToolId() === 'select',
-		[editor]
-	)
 	const isSinglePageMode = useValue('isSinglePageMode', () => editor.options.maxPages <= 1, [
 		editor,
 	])
-
-	if (!selectToolActive) return null
 
 	return (
 		<>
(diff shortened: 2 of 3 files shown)


mimecuvalo:
@steveruizok was there any good reason we did this historically?

steveruizok:
Hold on this one, there are reasons why we did this. We don't show selected shapes while in other tools, and the correct behavior is usually to return to the select tool on right click, but this is a place where the tool should decide.

kaneel:
so is the issue that we don't open the context menu, or that we don't return to the select shape so it opens the context menu? Because right now I don't see the "return to context menu" behaviour

steveruizok:
The desired behavior is for the tool to handle this interaction and return to the select tool. I'm not sure if the tool is getting this event. Fixing this may break right clicks on touch screen devices, which is where this bug may have originated.

angrycaptain19:
Discussion in the design meeting came to a conclusion that we should have right click switch tools to the select tool when bringing up the context menu. 

For long pressing on mobile there were some concerns raised about accidental context menus being opened, but that's just a general concern around long pressing being false positives and we could look to have different timings on that. 

Steve is going to test out #8948  's approach to mobile before we make a decision there

mimecuvalo:
@jsscclr will explore more in #8948

PR #8914, opened 2026-05-21
feature(examples): example page for handling pasting of mermaid diagrams

### LINKED ISSUE #8795: Add mermaid paste handler to mermaid examples
The two mermaid examples in the examples app each take mermaid source through their own UI (a button that loads a built-in set, or a textarea + Apply button) and call \`createMermaidDiagram\` directly. Neither registers an external content handler, so pasting mermaid source onto the canvas falls through to the default text/HTML behavior.

It would be a much more natural demo if pasting mermaid source onto the canvas produced shapes — that is also how a real app integrating \`@tldraw/mermaid\` would most likely wire it up.

## Proposal

Register an external content handler in both examples that detects mermaid source on paste and routes it through \`createMermaidDiagram\`, falling back to the default text handler otherwise.

Affected examples:

- \`apps/examples/src/examples/use-cases/hundred-mermaids/MermaidDiagramsExample.tsx\` — supports flowchart, state, mindmap, sequence.
- \`apps/examples/src/examples/use-cases/custom-shape-mermaids/CustomShapeMermaids.tsx\` — flowchart/graph only, with the existing \`getFlowchartSourceError\` validation and \`mapNodeToRenderSpec\` already in place.

The pattern in \`apps/examples/src/examples/data/assets/external-content-sources/ExternalContentSourcesExample.tsx\` is a good reference: \`editor.registerExternalContentHandler('text', ...)\` with \`defaultHandleExternalTextContent\` as the fallback.

## Notes

- The custom-shape example should keep its existing \`flowchart\`/\`graph\`-only constraint and surface the same parse errors it does today via the textarea path.
- Detection can be a simple prefix check on \`content.sources\` for \`flowchart\`, \`graph\`, \`stateDiagram\`, \`mindmap\`, or \`sequenceDiagram\`.
- Both READMEs should mention the paste behavior alongside the existing button/textarea workflow.

kaneel:
Title says it all!

Closes https://github.com/tldraw/tldraw/issues/8795

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature
- [ ] `api`
- [ ] `other`

### Test plan

1. Find and copy a solid 5/7 mermaid diagram (that we support)
2. paste it into the example page 
3. ???
4. profit

### Release notes

- Added an example page that shows how to support pasting of mermaid diagrams into a canvas

### DIFF
--- a/apps/examples/src/examples/use-cases/hundred-mermaids/README.md
+++ b/apps/examples/src/examples/use-cases/hundred-mermaids/README.md
@@ -1,5 +1,5 @@
 ---
-title: Mermaid diagrams
+title: Hundreds of Mermaid Diagrams
 component: ./MermaidDiagramsExample.tsx
 priority: 10
 keywords: [mermaid, diagram]
--- a/apps/examples/src/examples/use-cases/mermaid-pasting/MermaidPasting.tsx
+++ b/apps/examples/src/examples/use-cases/mermaid-pasting/MermaidPasting.tsx
@@ -1,0 +1,122 @@
+import { useEffect } from 'react'
+import { Tldraw } from 'tldraw'
+import { defaultHandleExternalTextContent, useEditor, useToasts } from 'tldraw'
+
+export default function MermaidDiagrams() {
+	return (
+		<div className="tldraw__editor">
+			<Tldraw>
+				<SneakyMermaidHandler />
+			</Tldraw>
+		</div>
+	)
+}
+
+export function SneakyMermaidHandler() {
+	const editor = useEditor()
+	const { addToast } = useToasts()
+
+	useEffect(() => {
+		editor.registerExternalContentHandler('text', async (content) => {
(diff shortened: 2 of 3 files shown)


apps/examples/src/examples/use-cases/hundred-mermaids/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Mermaid diagrams
+title: Hundreds of Mermaid Diagrams
```
mimecuvalo, line 2: https://en.wikipedia.org/wiki/Hundreds_of_Beavers ? :P

mimecuvalo (approved):
nice 👍

apps/examples/src/examples/use-cases/hundred-mermaids/README.md
kaneel: I did not have that reference!

PR #8917, opened 2026-05-21
feature(tldraw): use q while hovering to use the hovered styles for n…

### LINKED ISSUE #8820: Match current tool and styles to hovered shape with Q shortcut
## Problem statement

When iterating on a canvas with multiple shape types, continuing to draw more shapes that match an existing one is fiddly. You have to switch to the right tool manually, then re-apply each style (color, fill, dash, size, etc.) through the style panel.

There is no fast way to say "give me what that shape is" and keep drawing.

## Proposed solution

Press `Q` while hovering a shape to:

1. Switch the current tool to the tool associated with the hovered shape's type — e.g. geo tool for a rectangle, draw tool for a freehand stroke, arrow tool for an arrow, text tool for text.
2. Set the style-for-next-shapes to match the hovered shape's styles (color, fill, dash, size, opacity, font, geo, etc.).

No click, no mode, no menu. Hover, press `Q`, keep working.

For tools whose shape type is parameterised (geo, arrow type), the style sync should also set the relevant style — e.g. hovering a rectangle should set `GeoShapeGeoStyle` to `rectangle` before activating the geo tool, the same way the existing geo tool buttons do.

## Alternatives considered

- #7659 — a fuller eyedropper tool with a click-based workflow that applies styles to the current selection. This proposal is complementary: hover-based, single key, focused on the *next* shape rather than the selection.
- Selecting a similar shape and reading styles by eye, then applying through the style panel. Slow.
- Style presets. Heavier feature, doesn't help with one-off matching.

## Use cases

- Drawing more shapes that match one already on the canvas without rebuilding the style by hand.
- Resuming work on a canvas where the active tool and styles have drifted from what you were using earlier.
- Quickly aligning new annotations or arrows to an existing visual language on the canvas.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8820

https://github.com/user-attachments/assets/b98baf26-5b85-4d58-8873-b059e7e1c1d0

https://github.com/user-attachments/assets/197ac9a9-5842-4abc-970c-00a9618290a1

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### API changes

- adds a shortcut for the key `q` that copy the styles of hovered shape and apply them for the next shape

### Test plan

1. Create a shape with a yellow background
2. Create an other shape with a purple background
3. Select a drawing tool
4. Hover the yellow shape
5. Press q and start drawing

(it should be yellow)

1. Create a shape with a yellow background
2. Create an other shape with a purple background
3. Select the yellow background
4. Hover purple shape and press q 

(the purple shape should become yellow)

### Release notes

- Allow user to easily copy styles from hovered shape and apply them to the next shape


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1848,6 +1849,33 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+
+					if (!hovered) return
+
+					const tool = editor.getCurrentTool()
+
+					for (const style of editor.styleProps[hovered.type].keys()) {
+						const value = editor.getShapeStyleIfExists(hovered, style)
+						if (value === undefined) continue
+
+						if (tool instanceof SelectTool) {
```
max-drake: I think `editor.getCurrentToolId() === 'select'` is the more idiomatic way to do this

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1848,6 +1848,31 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+			{
+				id: 'copy-hovered-styles',
+				label: 'action.copy-hovered-styles',
+				kbd: 'q',
+				async onSelect() {
+					const hovered = editor.getShapeAtPoint(editor.inputs.getCurrentPagePoint(), {
+						hitInside: true,
+						renderingOnly: true,
+						hitLocked: true,
+					})
```
max-drake, line 1862: lets add an `margin: editor.options.hitTestMargin / editor.getZoomLevel()` like in `updateHoveredShapeIds.ts` to bring the behavior in line with when selection indicators are rendered, and to make it work at low zoom levels

max-drake:
one more thing!
this is a design opinion, but I think we should omit the geo type from styles that get selected in this

https://github.com/user-attachments/assets/99f91708-3158-4750-829b-29e49dc2b8b3

I feel like this just doesn't make sense behavior-wise. like the geo type is technically a prop, but it doesn't _feel_ like a prop to the user in the same way color does, i suspect

packages/tldraw/src/lib/ui/context/actions.tsx
steveruizok: Hm, you'd think that would be handled internally! I wonder where the exception is that made me keep the zoom-equal margin in caller land.

packages/tldraw/src/lib/ui/context/actions.tsx
steveruizok: Actually `editor.isIn('select')` is the really idiomatic way

steveruizok: You might want to use more specificity, e.g. `editor.isIn('select.idle')` or `editor.isIn('select.idle', 'select.cropping.idle')` or whatever, depending on which states you actually want this interaction to be possible in

steveruizok:
> one more thing! this is a design opinion, but I think we should omit the geo type from styles that get selected in this

Agreed!

packages/tldraw/src/lib/ui/context/actions.tsx
max-drake: ah!

max-drake: yeah maybe we'd only ever want to do this in an idle state, in which case you'd use `editor.getStateDescendant()` to check if we're in an idle state I believe?

kaneel: select.idle seems good enough for that case imo but happy to try anything different :)

max-drake (approved):
loooooove it

kaneel:
I'm merging, @steveruizok get ready to strap the laptop on for this one!

PR #8918, opened 2026-05-22
fix(tldraw):  fix dbl clicking over grouped shapes

### LINKED ISSUE #8898: Double-clicking a shape inside an unselected group enters editing immediately instead of selecting the shape
## Description

When a group is not selected (or not the focused group), double-clicking an editable shape inside the group (e.g. a text shape, note, or geo shape with a label) immediately puts the shape into editing mode.

The expected behavior — and the behavior you get if you instead click once to select the group, wait past the double-click window, then click again — is that the first double-click should select the shape (entering the group as its focus layer) and only a subsequent double-click should start editing it.

## Steps to reproduce

1. Create two text/note/geo shapes and group them together.
2. Deselect everything (or click on the canvas to clear the focused group).
3. Double-click on one of the shapes inside the group.

## Expected behavior

The first double-click selects the shape inside the group and sets it as the focused group layer. A second double-click then enters editing mode on the shape.

## Actual behavior

The shape immediately enters editing mode on the first double-click.

## Notes

The `onDoubleClick` handler for the `canvas` target in `packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts` already has the logic for "the hit shape is the child of a group that isn't the focused group — select the group first and bail." The same guard does not appear to exist on the `shape` target branch, so when the double-click arrives with `target: 'shape'` (e.g. when the pointer lands directly on the shape rather than the canvas margin), the code falls straight through to `startEditingShape`.

Related tests: `packages/tldraw/src/test/selection-omnibus.test.ts` — `describe('Selects inside of groups')`.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8898 

I had to fix it in case of a "shape" and a "selection" because for shapes with no background, the only way to select it is to hover the selection area (and doing that also ended up activating the "edit" mode)

https://github.com/user-attachments/assets/4aa867c2-9bfc-4d67-8a81-41821e494778

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape
2. Copy and paste this shape
3. Group the two shapes
4. try double clicking one of the shape 

The dbl clicked shape should _not_ enter its edit mode but be selected so it can be moved around in the group.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed double clicking on grouped shapes by making it select the shape instead of activating the shape's edit mode.

### DIFF
--- a/packages/tldraw/src/lib/shapes/note/noteCloning.test.ts
+++ b/packages/tldraw/src/lib/shapes/note/noteCloning.test.ts
@@ -335,6 +335,10 @@ it('Puts the new shape into a frame based on its center', () => {
 })
 
 function testNoteShapeFrameRotations(sourceRotation: number, rotation: number) {
+	// This helper runs several times per test; reset the click state so the
+	// next handle pointer-down starts a fresh click sequence instead of
+	// coalescing with the previous call into a double-click.
+	editor.cancelDoubleClick()
 	editor.createShape({ type: 'frame', x: 1220, y: 1000, rotation: rotation })
 	const frameA = editor.getLastCreatedShape()!
 	// top left won't be in the frame, but the center will (barely but yes)
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -367,9 +367,9 @@ export class Idle extends StateNode {
 						// The shape is a child of a group that the user hasn't
 						// drilled into yet (it's outside the focused group and
 						// not an ancestor of it). The user's double-click is a
-						// drill action, not an edit action: select the right
-						// shape via `selectOnCanvasPointerUp` and stop.
-						selectOnCanvasPointerUp(this.editor, info)
+						// drill action, not an edit action: drill one level and
+						// stop.
+						this.drillIntoGroupOnDoubleClick(hitShape)
 						return
 					}
 
@@ -404,6 +404,10 @@ export class Idle extends StateNode {
 					this.editor.isShapeOfType(parent, 'group') &&
 					!this.isGroupInPreviousDrillContext(parent.id)
 				) {
+					// The shape is a child of a group the user hasn't drilled into
+					// yet. Treat the double-click as a drill action — drill one
+					// level — but don't start editing.
+					this.drillIntoGroupOnDoubleClick(onlySelectedShape)
 					break
 				}
 
@@ -480,6 +484,10 @@ export class Idle extends StateNode {
 					this.editor.isShapeOfType(parent, 'group') &&
 					!this.isGroupInPreviousDrillContext(parent.id)
 				) {
+					// The shape is a child of a group the user hasn't drilled into
+					// yet. Treat the double-click as a drill action — drill one
+					// level — but don't start editing.
+					this.drillIntoGroupOnDoubleClick(shape)
 					break
 				}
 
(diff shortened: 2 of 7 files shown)


steveruizok:
Fix here isn't complete, and may introduce some regressions by the early return without the only selected shape. Adding a failing test here for one that I found.

kaneel:
~~Ah, I just realised as I read you that it could have been a feature to let the case falls down with an early return. A comment may help in that case~~ rechecked what we had before, the content of the case was wrapped and the `break` is right after it, seems very much like the early return is performing the same

max-drake:
in my opinion, if a user has a drilled selection (eg, they have a single shape selected that is part of a group), then double clicking a shape that is at or above the nesting level of the group they've drilled into (anything besides drilling down further) should enter editing mode on that single shape. this pr changes that behavior and makes it so that shape will just be selected

my understanding is that selection behavior should be consistent given relative levels of grouping, if you compare scenarios 2 and 3 in the image below, we get inconsistent behavior when relative group levels are the same. it might be more nuanced than that, given that in scenario 2, the mutual parent is the canvas instead of a group, so @steveruizok would love some clarification here on that 

shapes at 26-05-29 10 46 35

I've also noticed that some of the behavior is inconsistent btw images, filled, and hollow shapes, which is shown in the video below.

https://github.com/user-attachments/assets/29681e10-8d15-4367-bc03-212feb734fba

on main, all of the scenarios below result in entering editing mode on the shape, which imo is the correct behavior here

max-drake:
hey this is great! the drill context as a concept feels super clean. and behavior-wise, after copious amounts of fuzz testing, this matches everything that I'd expect. would probably be good to get another pair of eyes on it regardless, but I think this is pretty close, if not there already

kaneel:
removing this one from the design review as there seems to be issues with tests

steveruizok:
Needs to handle the second click (e.g. selecting a nested group when clicking on a group within a group).

max-drake:
some weird behavior when the user is clicks moderately slow during selection. if the user clicks super slow, the drilling and selection and editing works as intended. if they click really fast, it also works as intended. but if they click moderately slow, one of the clicks seems to be swallowed somehow. (ignore the macos click indicators not disappearting in the video, I tihnk there's a bug with macos screen recrodsing, you'll have to to trudst I'm clicking 4 times in each example)

https://github.com/user-attachments/assets/889f63f3-fa41-4100-88bd-4fe0a1100f29

I think the the root of the issue has to do with `focusedGroupIdBeforePointerDown` not being updated consistently. you'll see what i mean if you add console logs to the `onPointerDown` handler in `Idle.ts` and click through some nested groups at various speeds. eg, if you click through really fast, `focusedGroupIdBeforePointerDown` remains set to page:page even if you drill through, which is strange.

```ts
if (this.getClickManager().clickState === 'pendingDouble') {
	this.focusedGroupIdBeforePointerDown = this.editor.getFocusedGroupId()
	console.log('updated focusedGroupIdBeforePointerDown')
}
console.log('focusedGroupIdBeforePointerDown', this.focusedGroupIdBeforePointerDown)
```

I wonder if this has to do with the double click timeout? perhaps one pair of clicks (for example the 2nd and 3rd one) trigger the double click timeout, while the others don't trigger it, and this does causes some problem? tried to illustrate it below, not sure if it makes sense

image

steveruizok:
Closed in favor of #9488 . I think we were overthinking it

PR #8925, opened 2026-05-22
fix(tldraw):  fix reparenting in groups after a deletion

### LINKED ISSUE #8857: Deleting a shape from a group changes the z-order of other shapes
## Description

Deleting a shape that is inside a group can change the z-order of other shapes on the page. After the deletion, surviving children of the group end up in different stacking positions relative to shapes outside the group.

## Reproduction

1. Open https://tldraw.com/f/QI05IUEYZBT4GToj6q_sZ?d=v0.0.2620.1503.page
2. Double-click into the group and select the grey box
3. Delete the grey box

## Expected

The remaining shapes keep their original z-order relative to every other shape on the page.

## Actual

The white box is moved to the front of the patterned box, even though no reorder operation was performed.

## Notes

This appears to happen when a deletion reduces a group to a single child and the group is dissolved. The surviving child is reparented to the group's parent and is assigned a new fractional index based on its new siblings, which discards its prior stacking position relative to shapes outside the group.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8857

Deleting a shape in a group made the reparenting to repop the group at a different z-index, this should not happen and the group should keep the same index it previously owned

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

Described in https://github.com/tldraw/tldraw/issues/8857

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue with deleting shapes in a group that caused a reparenting of the group at a different z-index

### DIFF
--- a/packages/editor/src/lib/editor/shapes/group/GroupShapeUtil.tsx
+++ b/packages/editor/src/lib/editor/shapes/group/GroupShapeUtil.tsx
@@ -133,7 +133,7 @@ export class GroupShapeUtil extends ShapeUtil<TLGroupShape> {
 			if (this.editor.getCurrentPageState().focusedGroupId === group.id) {
 				this.editor.popFocusedGroupId()
 			}
-			this.editor.reparentShapes(children, group.parentId)
+			this.editor.reparentShapes(children, group.parentId, group.index)
 			this.editor.deleteShapes([group.id])
 			return
 		}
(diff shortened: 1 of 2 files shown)


steveruizok:
Good fix!

PR #8928, opened 2026-05-25
fix(tldraw): selected shapes win hover when stacked under others

### LINKED ISSUE #8822: Hovered shape can be set to a shape stacked on top of the selection
## Description

When a shape is selected and another shape is stacked on top of it, hovering over the overlap sets the hovered shape to the top (unselected) shape instead of the selected one. The selection should always win in hit testing: if a shape is selected and the pointer is over it, it should be hit before any other shape.

## Steps to reproduce

1. Draw shape A.
2. Draw shape B on top of A so they overlap.
3. Select shape A.
4. Move the pointer into the overlap area.

## Expected

Shape A (the selected shape) stays hovered, since hovering over the selection should win.

## Actual

Shape B becomes the hovered shape, even though A is selected and the pointer is over A.

## Notes

`getShapeToHover` in `packages/tldraw/src/lib/tools/selection-logic/updateHoveredShapeId.ts` uses `editor.getShapeAtPoint(...)` and does not give priority to selected shapes. The pointer-down equivalent in `getHitShapeOnCanvasPointerDown.ts` falls back to `editor.getSelectedShapeAtPoint(...)`, but only when nothing else is hit — selection still doesn't "win" when another shape is on top.

MitjaBezensek:
In order to make selection win in hover hit-testing, this PR updates `getShapeToHover` to prefer a selected shape under the pointer over any non-selected shape on top of it. Previously, hovering the overlap between a selected shape A and a non-selected shape B stacked above it set the hovered shape to B; now A stays hovered while it's selected and under the pointer. Closes #8822.

The fix runs a selected-only hit test (with the same hover hit options) before falling back to the existing `getShapeAtPoint`. The focused-group / outermost-selectable escalation logic is preserved. The pointer-down path (`getHitShapeOnCanvasPointerDown`) is intentionally not changed — clicking the top shape should still pick the top shape.

### Change type

- [x] `bugfix`

### Test plan

1. Draw shape A.
2. Draw shape B overlapping A.
3. Select A.
4. Move the pointer into the overlap area.
5. Verify A stays the hovered shape, not B.

- [x] Unit tests

### Release notes

- Fix: selecting a shape now keeps it hovered when the pointer is over it, even if another shape is stacked on top.

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/Idle.ts
@@ -1,5 +1,4 @@
 import {
-	Editor,
 	StateNode,
 	TLAdjacentDirection,
 	TLClickEventInfo,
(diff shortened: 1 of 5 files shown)


steveruizok:
I'm going to slam some code on this one, the issue is a little more complex (that rotated page bounds should essentially block hovered shapes).

MitjaBezensek:
@steveruizok this still relevant or should we close it?

angrycaptain19:
@max-drake were we looking at this same behavior yesterday and saw that it was fixed elsewhere?

MitjaBezensek:
Will go ahead and close.

PR #6974, opened 2025-10-21
feat(dotcom): introduce groups-based sidebar UI and workflows

ds300:
This PR introduces a groups-based sidebar UI with file/group drag-and-drop, group management, and invites, plus server/client mutations and routes to support group workflows.

### Change type

- [x] `other`

### Test plan

1. Create a group from the sidebar.
2. Drag and drop files between groups.
3. Invite a user to a group using the invite link.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Added groups-based sidebar UI with file/group management and invites.

---

> [!NOTE]
> Introduce groups-based sidebar UI with file/group DnD, group management and invites, plus server/client mutations and routes to support group workflows.
>
> - **UI/Sidebar (groups frontend)**:
>   - New grouped sidebar with "My files" and per-group sections, inline renaming, pin indicators, and context menus.
>   - Drag-and-drop: reorder pinned files, move files between groups, and reorder groups (visual drop cursor).
>   - New dialogs: `CreateGroupDialog`, `GroupSettingsDialog` (invite link, member roles, leave/delete), `AddFileLinkDialog`.
>   - File menu: “Move to” group, pin/unpin per-group, duplicate-to-group, delete/forget with group context.
> - **Invites & Routing**:
>   - Add route `GET /invite/:token` (client + worker) and accept flow `POST /invite/:token/accept`.
>   - `TlaInviteDialog` and handling on `/invite/:token` and in file/local pages; copy invite links from sidebar/menu.
> - **Client App/State**:
>   - `TldrawApp` enhancements: group memberships, stable ordering, ensure-visible helpers, upload/import to specific group, new drag state, toasts, navigation, and `hasFlag` utilities.
>   - Hooks/utilities for flags, drag tracking, isDragging; updated ownership check (`useHasFileAdminRights`).
> - **Server/Mutators**:
>   - New group mutations: create/update/delete, leave, set member role, regenerate invite secret, move file to group, add file link, handle file DnD, group reordering; max groups limit.
>   - Add `MAX_NUMBER_OF_GROUPS`, `ZErrorCode.max_groups_reached`, flag helpers (`parseFlags`, `userHasFlag`).
>   - Worker endpoints and DO/replicator tweaks to support groups and invites.
> - **Assets/i18n**:
>   - Add folder icons (`icon-folder`, `-open`, `-new`) and extensive locale strings for groups/invites.
> - **Misc**:
>   - Menu item supports right-side icon; CTA/button tweaks; tests snapshot updated.
[...]

PR #8929, opened 2026-05-25
fix(dotcom): unify sidebar group hover color with file/user rows

### LINKED ISSUE #8811: Sidebar group items hover color is inconsistent with other sidebar items
## Description

On tldraw.com, the hover background color used for sidebar group items (e.g. "Today", "Yesterday", "This week", and named groups) is different from the hover color used for file list items and the user settings item at the bottom of the sidebar.

The group items use the drop-zone color variable for their hover state, while the file links and user settings trigger use the standard hover color variable. The visible difference is subtle in light mode but more pronounced in dark mode, where the group item hover ends up nearly invisible because the drop-zone color is a dark overlay rather than a light one.

## Steps to reproduce

1. Open a tldraw.com workspace with files grouped under date headings.
2. Hover over a date group header (e.g. "Today").
3. Hover over a file under that group.
4. Hover over the user account button at the bottom of the sidebar.

## Expected

All three should share the same hover background color, or at least follow the same scale of hover tints.

## Actual

The group header hover is visibly lighter/different from the file link and user settings hovers. In dark mode the group hover is barely perceptible because it uses a dark overlay instead of a light one.

## Notes

- Affects: `apps/dotcom/client`
- Sidebar styles: `apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css`

MitjaBezensek:
In order to fix inconsistent sidebar hover colors on tldraw.com, this PR changes the hover background for sidebar group headers (e.g. "Today", "Yesterday", named groups) to use the same `--tla-color-hover-1` variable that file rows and the user settings row already use. Closes #8811.

Previously the group header `::after` shared a single rule with the drop-zone/menu-open/dragging visuals and painted `--tla-color-drop-zone` for all of them. In dark mode `--tla-color-drop-zone` is a dark overlay while `--tla-color-hover-1` is a white overlay, so hovering a group header darkened the row instead of lightening it — the opposite of every other hover state. In light mode the colors were close but still inconsistent (2.5% vs 3% dark overlay).

The fix splits the rule:

- Hover → `--tla-color-hover-1` (matches `.hoverable`)
- Menu open → `--tla-color-hover-2` (matches `.hoverable`'s open/active state)
- Empty first group hint and dragging → unchanged, still `--tla-color-drop-zone`

A second, smaller fix: `.sidebarFileListItem` has `opacity: 0.999` (a Chrome drag-image hack from #6974) which puts file rows into a separate compositing layer. That made the *composited* hover color render ~2/256 units darker on file rows than on group items and the user settings row, even though all three referenced the same CSS variable. To keep all three rows compositing the same way, the same `opacity: 0.999` is applied to `.sidebarGroupItem` and `.sidebarUserSettingsTrigger`.

### Change type

- [x] `bugfix`

### Test plan

1. Open a tldraw.com workspace with files grouped under date headings.
2. In **dark mode**, hover a group header — the row should lighten, matching the hover for file rows and the user settings row at the bottom of the sidebar.
3. In **light mode**, hover the same three items — all should share the same hover tint.
4. Sample the hover color across all three rows with a color picker — they should match.
5. Right-click a group header (or open its kebab menu) — the open-menu background should match the open-menu background on file rows.
6. Drag a file between groups — the drop-zone highlight on the target group should be unchanged, and the file drag image should still render with a transparent background.
7. Confirm the empty-first-group hint state (when the first date group has no files) still uses the drop-zone tint.
8. Confirm the `.dropping::after` blue drop-target border is unchanged.

### Release notes

- Fix inconsistent hover background color for sidebar group items on tldraw.com, especially in dark mode where the group hover was nearly invisible.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css
@@ -863,6 +863,14 @@
 	visibility: visible;
 }
 
+.sidebarGroupItem:has(.sidebarGroupItemHeader:hover)::after {
+	background-color: var(--tla-color-hover-1);
+}
+
+.sidebarGroupItem[data-menu-open='true']::after {
+	background-color: var(--tla-color-hover-2);
+}
+
 .sidebarGroupItem[data-is-dragging='true'] {
 	opacity: 0.5;
 }
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css
@@ -354,6 +354,8 @@
 	gap: 0px;
 	text-align: left;
 	cursor: pointer;
+	/* Match .sidebarFileListItem so the hover overlay composites identically. */
+	opacity: 0.999;
 }
 
 .sidebarUserSettingsTrigger:focus-visible {
@@ -811,6 +813,8 @@
 	margin-bottom: 0px;
 	transition: margin-bottom 0.15s cubic-bezier(0.4, 0, 0.2, 1);
 	--tla-sidebar-group-item-padding-left: 8px;
+	/* Match .sidebarFileListItem so the hover overlay composites identically. */
+	opacity: 0.999;
 }
 .sidebarGroupItem .sidebarFileListItemContent {
 	padding-left: var(--tla-sidebar-group-item-padding-left);


steveruizok:
> A second, smaller fix: .sidebarFileListItem has opacity: 0.999 (a Chrome drag-image hack from https://github.com/tldraw/tldraw/pull/6974) which puts file rows into a separate compositing layer. That made the composited hover color render ~2/256 units darker on file rows than on group items and the user settings row, even though all three referenced the same CSS variable. To keep all three rows compositing the same way, the same opacity: 0.999 is applied to .sidebarGroupItem and .sidebarUserSettingsTrigger.

Is this needed? This seems like we should be using a specific color for the hovers. The original issue possibly misunderstood the problem (known slop issue with `/issue`). The hover color for the groups is different than the hover color for anything else. This is just a color issue, this shouldn't be a complex fix.

MitjaBezensek:
> Is this needed? This seems like we should be using a specific color for the hovers. The original issue possibly misunderstood the problem (known slop issue with `/issue`). The hover color for the groups is different than the hover color for anything else. This is just a color issue, this shouldn't be a complex fix.

It is needed if we want to have the exact same colors. Looks like the file menu item is using a [hack for chrome](https://github.com/tldraw/tldraw/blob/86a871306da764abfd36bc0215fa1aa513e80211/apps/dotcom/client/src/tla/components/TlaSidebar/sidebar.module.css#L479). This changes the color just slightly (#161616 vs #181818). Maybe there's a different hack that does the same but doesn't influence the color.

MitjaBezensek:
Shoud be fixed with #9130

PR #8936, opened 2026-05-25
chore(infra): consolidate npm publishing into a single workflow

### LINKED ISSUE #8935: Consolidate npm publishing into a single reusable workflow for OIDC trusted publishing
## Problem

After switching npm publishing to OIDC trusted publishers (#8913), publishing can fail with:

```
npm error code ENEEDAUTH
npm error need auth This command requires you to be logged in to https://registry.npmjs.org/
```

`ENEEDAUTH` means npm never engaged the OIDC trusted-publishing flow and fell back to looking for a token, which no longer exists (`NPM_TOKEN` was intentionally removed in `internal/scripts/lib/publishing.ts`).

The root cause is a mismatch between how npm trusted publishing works and how this repo publishes:

- A npm trusted-publisher config pins **exactly one workflow filename per package**.
- This repo publishes the same packages to npm from **five different workflows**, all using the `npm deploy` environment:

| Workflow | Script |
| --- | --- |
| `publish-canary.yml` | `publish-prerelease.ts` (canary/next) |
| `publish-branch.yml` | `publish-prerelease.ts internal` |
| `publish-patch.yml` | `publish-patch.ts` |
| `publish-new.yml` | `publish-new.ts` |
| `publish-manual.yml` | `publish-manual.ts` |

Because only one workflow filename can be registered as the trusted publisher per package, any publish that runs from a different workflow than the registered one fails with `ENEEDAUTH`.

All 15 publishable packages are affected:

```
@tldraw/assets      @tldraw/state        @tldraw/sync-core
create-tldraw       @tldraw/state-react  @tldraw/tlschema
@tldraw/driver      @tldraw/store        @tldraw/utils
@tldraw/editor      @tldraw/mermaid      @tldraw/validate
@tldraw/tldraw      tldraw               @tldraw/sync
```

## Proposed solution

Consolidate all npm publishing into a single reusable workflow (e.g. `publish-npm.yml` with `on: workflow_call`) that performs the actual `npm publish`. The five existing workflows then call it via `uses: ./.github/workflows/publish-npm.yml`.

Because the publish step always runs inside that one reusable workflow file, a single trusted-publisher filename can be registered for each package, and every publish path authenticates.

Steps:

1. Create a reusable `publish-npm.yml` workflow with `on: workflow_call`, declaring `permissions: id-token: write` and `environment: npm deploy` on the publishing job.
2. Refactor `publish-canary.yml`, `publish-branch.yml`, `publish-patch.yml`, `publish-new.yml`, and `publish-manual.yml` to call the reusable workflow instead of running `npm publish` directly.
3. Register the reusable workflow filename as the trusted publisher for all 15 packages on npmjs.com (org `tldraw`, repo `tldraw`, environment `npm deploy`).

## Notes to verify

- For `workflow_call`, confirm npm matches the OIDC claim against the **called** (reusable) workflow's filename, and that `id-token: write` + the environment are declared on the reusable workflow's job. Validate against npm's current trusted-publishing docs before finalizing.
- Trusted publishing cannot bootstrap a brand-new package name (the package must already exist on npm). All 15 packages already exist, so this only matters for any future new package.

steveruizok:
In order to make npm OIDC trusted publishing work across all of our publish flows, this PR consolidates every npm publish (canary, branch/internal, patch, new, manual, and the `production` "next" channel) into a single `publish.yml` workflow. Closes #8935.

npm trusted publishing matches the OIDC `job_workflow_ref` claim against a single workflow filename configured per package. Because we previously published the same packages from several different workflows, only one of them could authenticate and the rest failed with `ENEEDAUTH`. Routing every flow through one workflow file means they all authenticate against the same trusted publisher.

### What changed

- Added `.github/workflows/publish.yml`, a single workflow that:
  - Triggers on push to `main` (canary), `production` (next), and `v*.*.x` branches (patch), on the `publish-packages` PR label (internal), and on `workflow_dispatch` for `manual` and `new` releases.
  - Determines the publish mode and per-flow settings (`bemo_url`, whether to use the huppy app token) in a single "Determine publish mode" step, with input validation for the manual and new flows.
  - Runs the appropriate publish script (`publish-prerelease.ts`, `publish-patch.ts`, `publish-manual.ts`, `publish-new.ts`) based on the resolved mode.
  - Declares `environment: npm deploy` so the environment-scoped secrets (`R2_*`, `HUPPY_*`, `DISCORD_*`) resolve, and `id-token: write` for OIDC.
  - Preserves the per-flow orchestration that lived in the old workflows: PR comment + label removal for internal publishes, the tldraw-desktop dispatch for `next`, Discord failure notification, and the `publish-templates.yml` follow-up job.
- Removed the per-flow workflows (`publish-branch.yml`, `publish-canary.yml`, `publish-manual.yml`, `publish-new.yml`, `publish-patch.yml`) and the reusable `publish-npm.yml`.
- Updated `RELEASES.md` to document the new single-workflow process (publishing new releases from `production` via `publish.yml` with the `new` publish type).

### Follow-up required (not in this PR)

- [x] On npmjs.com, each publishable package must have its trusted publisher set to org `tldraw`, repo `tldraw`, workflow filename `publish.yml`, environment `npm deploy`. Until that is done, publishing continues to fail with `ENEEDAUTH`.

### Change type

- [x] `other` (CI/infra)

### Test plan

These workflows can only be fully exercised by a real publish, so verification is staged:

1. Confirm the workflow parses (actionlint) — actionlint is not installed in the local environment, so this should be checked in CI.
2. After merge, register the trusted publishers on npm against `publish.yml`.
3. Verify a canary publish on push to `main` succeeds end to end.
4. Verify a branch/internal publish (via the `publish-packages` label) succeeds and comments the version.

### Code changes

| Section        | LOC change  |
| -------------- | ----------- |
| Config/tooling | +261 / -371 |


.github/workflows/publish.yml
```diff
@@ -0,0 +1,280 @@
...
+  remove-label:
+    name: Remove the publish-packages label
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.label.name == 'publish-packages' &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
```
MitjaBezensek, line 55: Think we might also need `pull-requests: write` here or removing the label from pull requests won't work? We can try it and see if it fails 🤷‍♂️

.github/workflows/publish.yml
```diff
@@ -0,0 +1,280 @@
...
+
+  comment:
+    name: Tell user released version
+    if: needs.publish.outputs.mode == 'internal'
+    needs: publish
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      issues: write
```
MitjaBezensek, line 262: `pull-requests: write` here as well?

MitjaBezensek (approved):

.github/workflows/publish.yml
steveruizok: 🤖 Handled in cc81a0d116. I split label removal out of the OIDC publish job and gave that lightweight job `issues: write` instead. Pull request labels are issue labels in the GitHub API, so this should cover removing `publish-packages` without giving the npm publish job PR write access.

.github/workflows/publish.yml
steveruizok: 🤖 Handled in cc81a0d116. I moved the PR comment into its own post-publish job and gave that job `issues: write`; the publish job itself now only has `contents: read` and `id-token: write`.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,280 @@
...
+        id: generate_desktop_token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ vars.HUPPY_APP_ID }}
+          private-key: ${{ secrets.HUPPY_PRIVATE_KEY }}
+          owner: tldraw
+          repositories: tldraw-desktop
+
+      - name: Notify tldraw-desktop
+        if: steps.publish_mode.outputs.mode == 'next'
```
danieljamesross, line 240: `needs: generate_desktop_token`

.github/workflows/publish.yml
steveruizok: ^^ good bot??

.github/workflows/publish.yml
steveruizok: 🤖 This one is a same-job step dependency, so `steps.generate_desktop_token.outputs.token` is the right context here. `needs` only applies to outputs from other jobs, and `generate_desktop_token` is the immediately preceding step rather than a job.

PR #8937, opened 2026-05-26
feat(tldraw): export addRoundedRectPath utility

### LINKED ISSUE #8907: Export addRoundedRectPath utility for SDK consumers
## Problem

`addRoundedRectPath()` is a generic Path2D helper that draws a rounded rectangle (optionally counter-clockwise, for use as a clip-path cutout). It currently lives as a private function inside `packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx`, so SDK consumers cannot reuse it when building custom shape utils.

The function has no arrow-specific behavior. It is useful for any custom shape that needs to:

- Build a label-cutout clip path (the arrow's own use case).
- Render a rounded-rect indicator path.
- Generally produce a rounded-rect `Path2D` without manually writing the arc math.

## Proposed solution

Move `addRoundedRectPath` to a shared location alongside the other shape helpers (for example `packages/tldraw/src/lib/shapes/shared/`, next to `PathBuilder`), and re-export it from the `tldraw` package entry point so consumers can import it the same way they import `PathBuilder` today.

Also, change the oxlint plugin linter hint to properly reference this helper, instead of telling the dev to make the roundrect manually

## Current location

`packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx:91`

frolic:
In order to let SDK consumers draw rounded rectangle subpaths into a `Path2D` without copy/pasting from `ArrowShapeUtil`, this PR moves the existing `addRoundedRectPath` helper out of `ArrowShapeUtil.tsx` into its own file under `packages/tldraw/src/lib/shapes/shared/`, re-exports it from the package, and adds typedoc.

Closes #8907.

### Change type

- [x] `api`

### Test plan

- Arrow labels still render with a rounded background and clip out the arrow body cleanly behind the label.

### Release notes

- Export `addRoundedRectPath`, a helper for appending a rounded-rectangle subpath to a `Path2D`.

### API changes

- Added `addRoundedRectPath(path, bounds, radius, counterClockwise?)` for drawing rounded rectangle subpaths into a `Path2D`.

### Code changes

| Core code       | +65 / -34  |

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -186,6 +186,9 @@ export interface ActionsProviderProps {
     overrides?(editor: Editor, actions: TLUiActionsContextType, helpers: TLUiOverrideHelpers): TLUiActionsContextType;
 }
 
+// @public
+export function addRoundedRectPath(path: Path2D, bounds: Box, radius: number, counterClockwise?: boolean): void;
+
 // @public (undocumented)
 export type AlertSeverity = 'error' | 'info' | 'success' | 'warning';
 
--- a/packages/tldraw/src/index.ts
+++ b/packages/tldraw/src/index.ts
@@ -240,6 +240,7 @@ export {
 	type NoteShapeOptions,
 	type NoteShapeUtilDisplayValues,
 } from './lib/shapes/note/NoteShapeUtil'
+export { addRoundedRectPath } from './lib/shapes/shared/addRoundedRectPath'
 export {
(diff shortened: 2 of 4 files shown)


frolic:

frolic:
The underlying issue was fixed in Chrome and should be released next week (See https://github.com/tldraw/tldraw/issues/8907#issuecomment-4546261657)

PR #8943, opened 2026-05-27
feat(tldraw): add allowVideoAutoplay option to TldrawOptions

### LINKED ISSUE #8915: Add an option to TldrawOptions to control whether videos autoplay
## Problem

Today, video shapes autoplay by default. The behavior is configurable in two places:

- Per shape, via `TLVideoShape['props'].autoplay` (set on the shape record itself).
- As a default for newly created video shapes, via `VideoShapeOptions.autoplay` on `VideoShapeUtil` (see `packages/tldraw/src/lib/shapes/video/VideoShapeUtil.tsx`).

The shape-util option was added in #5716, but it can only be set by subclassing `VideoShapeUtil`. There is no way to disable video autoplay at the `<Tldraw />` component level without writing a custom shape util — which is a lot of ceremony for a single boolean.

## Proposed solution

Add a video autoplay option to `TldrawOptions` (`packages/editor/src/lib/options.ts`) so SDK consumers can do:

```tsx
<Tldraw options={{ videoAutoplay: false }} />
```

The default should be `true` so current behavior is preserved. `VideoShapeUtil` should read this value when computing the autoplay default for new video shapes (or when deciding whether to set `autoPlay` on the rendered `<video>` element — the issue is which behavior we want; see below).

Naming suggestion: `videoAutoplay`, to match the existing `shape.props.autoplay` and `VideoShapeOptions.autoplay` and to leave room for a future `audioAutoplay` if needed.

## Open questions

- Should this option control the **default `autoplay` prop on new video shapes**, or **whether any video plays automatically regardless of the shape's prop**? The first matches the existing `VideoShapeOptions.autoplay` semantics; the second is a stronger guarantee that a host app can disable autoplay across the board, including for pasted/restored shapes.
- Does it need to interact with the existing `prefers-reduced-motion` check at `VideoShapeUtil.tsx:213`? Probably not — that should keep behaving as it does today.

## Alternatives considered

- Subclassing `VideoShapeUtil` and overriding `options.autoplay` — works but is overkill for a single flag.
- Setting `shape.props.autoplay = false` on every video at creation time — requires hooking into shape creation and doesn't help for pasted/loaded shapes.

frolic:
In order to let SDK consumers disable video autoplay across the board without subclassing `VideoShapeUtil`, this PR adds a top-level `allowVideoAutoplay` option to `TldrawOptions`. The host opt-out acts as a ceiling, not a floor: when `false`, no video autoplays regardless of its per-shape `autoplay` prop; when `true` (the default, preserving existing behavior), each shape's per-shape `autoplay` prop is respected. The `prefers-reduced-motion` check at the render site continues to apply independently.

```tsx
<Tldraw options={{ allowVideoAutoplay: false }} />
```

### Behavior matrix

| `TldrawOptions.allowVideoAutoplay` | `shape.props.autoplay` | Result          |
| ---------------------------------- | ---------------------- | --------------- |
| `true` (default)                   | `true`                 | Autoplays       |
| `true` (default)                   | `false`                | Does not play   |
| `false`                            | `true`                 | Does not play   |
| `false`                            | `false`                | Does not play   |

So `allowVideoAutoplay: false` is a global opt-out for the host app — it never forces every video to play. The per-shape prop remains the way users opt individual shapes in or out when autoplay is allowed at the host level.

Closes [#8915](https://github.com/tldraw/tldraw/issues/8915).

### Change type

- [x] \`feature\`

### Test plan

1. Create a tldraw editor with \`<Tldraw options={{ allowVideoAutoplay: false }} />\`.
2. Drop a video file onto the canvas — confirm it does not start playing automatically.
3. Click into the video to edit, press play — confirm the user can still play it manually.
4. With the default option (or \`allowVideoAutoplay: true\`), drop a video and confirm it still autoplays as before.
5. With the default option, set \`shape.props.autoplay = false\` on an individual video — confirm only that shape stops autoplaying while others still autoplay.

### Release notes

- Add an \`allowVideoAutoplay\` option to \`TldrawOptions\`. When set to \`false\`, no video shape autoplays regardless of its per-shape \`autoplay\` prop — including pasted and restored shapes. Defaults to \`true\`.

### API changes

- Added \`TldrawOptions.allowVideoAutoplay: boolean\` (default \`true\`).

### Code changes

| Core code       | +18 / -1   |
| Documentation   | +7 / -0    |

### DIFF
--- a/apps/docs/content/sdk-features/default-shapes.mdx
+++ b/apps/docs/content/sdk-features/default-shapes.mdx
@@ -324,6 +324,12 @@ const ConfiguredVideoUtil = VideoShapeUtil.configure({
 })
 ```
 
+To disable autoplay across the board — including for pasted or restored shapes — set the editor-level `videoAutoplay` option to `false` instead:
+
+```tsx
+<Tldraw options={{ videoAutoplay: false }} />
+```
+
 ### Bookmark
 
 The bookmark shape displays a URL as a card with metadata including title, description, and preview image. Bookmarks are created when you paste URLs onto the canvas. The editor fetches metadata from the URL and stores it in an associated asset record. Bookmark shapes have fixed dimensions and can't be resized.
--- a/apps/docs/content/sdk-features/default-shapes.mdx
+++ b/apps/docs/content/sdk-features/default-shapes.mdx
@@ -324,10 +324,10 @@ const ConfiguredVideoUtil = VideoShapeUtil.configure({
 })
 ```
 
-To disable autoplay across the board — including for pasted or restored shapes — set the editor-level `videoAutoplay` option to `false` instead:
+To disable autoplay across the board — including for pasted or restored shapes — set the editor-level `allowVideoAutoplay` option to `false` instead:
 
 ```tsx
-<Tldraw options={{ videoAutoplay: false }} />
+<Tldraw options={{ allowVideoAutoplay: false }} />
 ```
 
 ### Bookmark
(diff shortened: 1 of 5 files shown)


frolic:
I wonder if we should call this `allowVideoAutoplay` or `disableVideoAutoplay` instead of just `videoAutoplay` to clarify that `true` doesn't mean "autoplay all videos"

max-drake (approved):
lgtm :)

PR #8944, opened 2026-05-27
fix(tldraw): block paste while pointer is down

### LINKED ISSUE #8305: Pasting while dragging an arrow terminal interrupts binding
## Description

When dragging an arrow terminal into a geo shape, pasting (Cmd+V / Ctrl+V) interrupts the drag operation. The dotted binding indicator disappears, and on pointer-up the newly pasted shape gets selected instead of completing the arrow binding.

Similarly when pasting during a resize or moving a geoshape, the pasted shape gets selected rather than the one the user was resizing/moving. 

https://github.com/user-attachments/assets/8f672d34-0b81-48e7-b3bf-da7ea52bd997

## Steps to reproduce

1. Create a geo shape and an arrow
2. Start dragging one of the arrow's terminals toward the geo shape
3. While still holding the mouse button, press Cmd+V (with something on the clipboard)
4. Observe: the dotted binding indicator disappears
5. Release the mouse button
6. The pasted shape is selected instead of the arrow terminal completing its binding

## Expected behavior

Paste should be suppressed (or deferred) while actively dragging an arrow handle, similar to how editing a shape's text suppresses paste.

## Actual behavior

The paste goes through, creates new shapes, changes the selection, and disrupts the in-progress arrow handle drag.

## Analysis

The paste handler in `useClipboardEvents.ts` checks `editor.getEditingShapeId() !== null` to avoid pasting during text editing. However, arrow terminal dragging (`DraggingHandle` state) never sets `editingShapeId`, so paste events are not blocked during this interaction.

The `DraggingHandle` state does have `onCancel()` handling, but paste doesn't trigger it — it just changes selection and creates shapes underneath the active drag, disrupting the operation.

frolic:
In order to fix #8305, this PR adds an `editor.inputs.getIsPointing()` guard to the three paste handlers in `useClipboardEvents.ts`. Without it, pressing Cmd+V while dragging an arrow handle (or resizing/translating a shape) would create new shapes underneath the active drag and steal selection from it, so the in-progress interaction ended in a broken state — the dotted arrow-binding indicator disappeared, and on pointer-up the newly pasted shape became the selection instead of the arrow completing its binding.

The clipboard handler already short-circuited when `editor.getEditingShapeId() !== null` (the text-editing case). The fix extends the same guard pattern to cover pointer-driven interactions, which use `editor.inputs.isPointing` rather than `editingShapeId`. The state machine itself was unchanged.

### Change type

- [x] `bugfix`

### Test plan

1. Create a geo shape and an arrow on a tldraw canvas.
2. Press and hold on the arrow's end handle, drag it toward the geo shape so the dotted binding indicator appears.
3. While the pointer is still down, press Cmd+V (with anything on the clipboard).
4. Confirm the binding indicator stays, no new shape is pasted, and releasing the pointer completes the arrow binding.
5. Repeat with a shape resize and a shape translate — pasting mid-drag should be a no-op in both cases.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix paste interrupting an in-progress arrow handle drag, shape resize, or shape translate.

### API changes

- Promoted `handlePasteFromEventClipboardData` from an unexported helper to an `@internal` export (used by the new regression test). No public surface change — `api-report.api.md` is unaffected.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +20 / -7   |
| Tests     | +65 / -1   |

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -2397,7 +2397,7 @@ export function GroupMenuItem(): JSX.Element | null;
 export function GroupOrUngroupMenuItem(): JSX.Element;
 
 // @public
-export const handleNativeOrMenuCopy: (editor: Editor, context?: TLClipboardWriteInfo) => Promise<boolean>;
+export function handleNativeOrMenuCopy(editor: Editor, context?: TLClipboardWriteInfo): Promise<boolean>;
 
 // @public (undocumented)
 export class HandTool extends StateNode {
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -2397,7 +2397,7 @@ export function GroupMenuItem(): JSX.Element | null;
 export function GroupOrUngroupMenuItem(): JSX.Element;
 
 // @public
-export function handleNativeOrMenuCopy(editor: Editor, context?: TLClipboardWriteInfo): Promise<boolean>;
+export const handleNativeOrMenuCopy: (editor: Editor, context?: TLClipboardWriteInfo) => Promise<boolean>;
 
 // @public (undocumented)
 export class HandTool extends StateNode {
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -236,13 +236,16 @@ type ClipboardThing =
  * @param point - The point to paste at
  * @internal
  */
-const handlePasteFromEventClipboardData = async (
+export const handlePasteFromEventClipboardData = async (
 	editor: Editor,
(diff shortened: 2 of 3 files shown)


max-drake:
I think this isn't the ideal path here. (fwiw I think the ai research on https://github.com/tldraw/tldraw/issues/8305 itself was misleading, see [my comment](https://github.com/tldraw/tldraw/issues/8305#issuecomment-4557858917)). I don't see a good reason for blocking paste while dragging, and I think the ideal fix would be to make sure the shape being dragged/resized is selected when the drag/resize completes. this would require digging into the `StateNode`s that constitute our tools, which may not be a first week type task, but if you're interested feel free! our tool system is genuinely really cool

frolic:
Cool, will see about revisiting in a separate PR

PR #8946, opened 2026-05-27
fix(tldraw): cancel pending shape creation on touch long press

### LINKED ISSUE #8277: Long press in coarse pointer mode creates shapes and doesn't open context menu for non-select tools
## Description

When using coarse pointer mode (touch devices), several issues occur with long press behavior across different tools:

1. **Geo tool**: Long pressing spawns a geo shape when the context menu opens, because the shape is created on `pointerUp` (which fires when the long press completes)
2. **Note tool**: The note shape is created immediately on `pointerDown` (in `onEnter`), then the context menu opens after — leaving a stray shape behind
3. **Other tools**: No tools besides Select open the context menu on long press at all

## Expected behavior

- In coarse pointer mode, long pressing should open the context menu without creating shapes
- Shape creation tools should defer shape creation so that a long press can cancel it
- All tools (not just Select) should support long press to open the context menu in coarse pointer mode

## Actual behavior

- Geo tool creates a default-sized shape when the long press fires (via `pointerUp` → `complete()`)
- Note tool creates the shape immediately in `onEnter` of the `Pointing` state, so it's already on canvas before the context menu opens
- Non-select tools don't handle `long_press` events or trigger the context menu

## Steps to reproduce

1. Open tldraw on a touch device (or enable coarse pointer mode)
2. Select the Geo shape tool
3. Long press on the canvas
4. Observe: a geo shape is created and the context menu may open

Repeat with the Note tool — the note appears immediately on touch, then the context menu opens.

Try with any other tool (draw, line, arrow, etc.) — long press does nothing, no context menu appears.

## Relevant code

- **Geo Pointing state**: `packages/tldraw/src/lib/shapes/geo/toolStates/Pointing.ts` — creates shape in `complete()` on `pointerUp`
- **Note Pointing state**: `packages/tldraw/src/lib/shapes/note/toolStates/Pointing.ts` — creates shape in `onEnter()`
- **Long press dispatch**: `packages/editor/src/lib/editor/Editor.ts` ~line 10585 — dispatches `long_press` after 500ms timeout
- **Context menu**: `packages/tldraw/src/lib/ui/components/ContextMenu/DefaultContextMenu.tsx` — handles coarse pointer open/close
- **Select tool handlers**: Only the Select tool's child states implement `onLongPress` handlers

jsscclr:
In order to fix [#8277](https://github.com/tldraw/tldraw/issues/8277), this PR makes every shape-creation tool handle `onLongPress` in coarse pointer mode by canceling any pending creation and returning to idle. The browser-fired `contextmenu` then opens the context menu cleanly with no stray shape left behind, matching the behavior the Select tool already had.

Before this change, long-pressing the canvas on a touch device with a shape-creation tool active produced a mess: the geo and text tools created a default-sized shape on the eventual `pointerUp` (which fires when the long-press releases), while the note and line tools created a shape immediately in `onEnter` and left it behind once the menu opened. Other tools (draw, arrow, frame) simply ignored long-press entirely.

The cancel is gated on `editor.getInstanceState().isCoarsePointer` so desktop behavior is preserved. The "pause before drag to create an arrow with a precise start point" gesture, which relies on the 500ms long-press timer firing during a fine-pointer hold, still works on mouse devices.

Closes #8277

### Change type

- [x] `bugfix`

### Test plan

1. Open the app on a touch device, or enable touch + coarse pointer emulation in Chrome DevTools (⌘⇧M, pick a touch device).
2. For each tool (Rectangle, Ellipse, other geo variants, Note, Line, Arrow, Text, Frame, Draw): long-press the canvas without moving.
3. Confirm: the context menu opens and **no shape is left on the canvas**.
4. Turn touch emulation off. Press R for the Rectangle tool and click-drag — a shape should still be created normally.
5. Press the Arrow tool, mouse-down on a shape, pause for over 500ms, then drag — the arrow should still create with a precise start point (existing fine-pointer behavior).

This behavior is also covered by `packages/tldraw/src/test/long-press-create.test.ts`, which drives a long press through each tool in both coarse- and fine-pointer modes and asserts the coarse case returns to idle with no shape left behind while the fine case stays mid-creation.

- [x] Unit tests

### Release notes

- Fix touch long-press on shape-creation tools so the context menu opens without leaving a stray shape behind.

### Code changes

| Core code       | +32 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
+++ b/packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts
@@ -77,6 +77,10 @@ export class Pointing extends StateNode {
 		this.cancel()
 	}
 
+	override onLongPress() {
+		if (this.editor.getInstanceState().isCoarsePointer) this.cancel()
+	}
+
 	complete() {
 		const originPagePoint = this.editor.inputs.getOriginPagePoint()
 
--- a/packages/tldraw/src/lib/shapes/arrow/toolStates/Pointing.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/toolStates/Pointing.tsx
@@ -81,6 +81,10 @@ export class Pointing extends StateNode {
 		this.cancel()
 	}
 
+	override onLongPress() {
+		if (this.editor.getInstanceState().isCoarsePointer) this.cancel()
+	}
+
 	cancel() {
 		if (this.shape) {
 			// the arrow might not have been created yet!
--- a/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
+++ b/packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
@@ -756,6 +756,14 @@ export class Drawing extends StateNode {
 		this.cancel()
 	}
 
+	override onLongPress() {
+		if (!this.editor.getInstanceState().isCoarsePointer) return
+		if (this.markId) {
+			this.editor.bailToMark(this.markId)
+		}
+		this.cancel()
+	}
+
 	complete() {
 		const { initialShape } = this
 		if (!initialShape) return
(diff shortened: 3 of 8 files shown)


jsscclr:
This just stops the behaviour for long press shape creation for coarse pointers - makes no judgement on whether it should activate selection tool.

kostyafarber (approved):

kostyafarber:
lgtm just adding @steveruizok because touching the root pointing states etc

PR #8258, opened 2026-03-17
chore(tooling): migrate lint and formatting to oxlint and oxfmt

mimecuvalo:
Closes #7572.

In order to remove slow dual-linting and simplify the toolchain, this PR fully migrates the repo from ESLint + Prettier to Oxlint + Oxfmt, with workspace lint scripts executing Oxlint from the repo root for consistent config resolution.

This removes the ESLint config and in-repo ESLint plugin setup, adds an Oxlint JS plugin for tldraw-specific rules, switches formatter commands and config to Oxfmt, and updates package/workspace scripts and CI to use the new stack end-to-end.

### After:
Oxlint: ~6s
Oxfmt: ~162ms

### Before:
Eslint: ~1min 15s
Prettier: ~4s

### Change type

- [x] `improvement`

### Test plan

1. Run `yarn lint` from repo root.
2. Run `yarn lint` from a workspace (for example `packages/tldraw` and `apps/dotcom/client`) and verify it passes.
3. Run `yarn check-packages` and verify lint script expectations pass.
4. Run `yarn format-current` and `yarn lint-current` on changed files.

- [ ] Unit tests
- [ ] End to end tests

Issue #7572, opened 2026-01-02
Migrate from eslint to oxlint


steveruizok:
## Summary

Evaluate and migrate the codebase linting from ESLint to [Oxlint](https://oxc.rs/), a Rust-based JavaScript/TypeScript linter that offers significant performance improvements.

## Problem statement

ESLint is currently the linter used across the monorepo. While ESLint is mature and feature-rich, it can be slow on large codebases. The tldraw monorepo contains dozens of packages and apps, making lint times a potential bottleneck in developer workflows and CI pipelines.

## Proposed solution

Migrate to Oxlint, which offers:

- **50-100x faster** than ESLint in benchmarks
- Written in Rust with near C/C++ performance
- Memory efficient (92MB vs ESLint's 1.4GB in some benchmarks)
- [Oxlint 1.0 reached stable](https://voidzero.dev/posts/announcing-oxlint-1-stable) in June 2025 with 520+ supported ESLint rules
- Real-world adoption: Mercedes-Benz saw 71% decrease in lint time, Airbnb lint 126k+ files in 7 seconds

### Current ESLint setup
The codebase uses ESLint 9 with flat config (`eslint.config.mjs`) and several plugins:
- `@typescript-eslint` for TypeScript rules
- `eslint-plugin-react` and `eslint-plugin-react-hooks`
- `eslint-plugin-import`
- `eslint-plugin-formatjs`
- `eslint-plugin-no-only-tests`
- `@next/eslint-plugin-next`
- Custom local rules plugin

### Custom rules to address
The codebase has 10 custom ESLint rules in `internal/scripts/eslint/eslint-plugin.mjs`:
- `local/no-export-star` - Enforces named exports over `export *`
- `local/no-internal-imports` - Prevents importing from internal tldraw paths
- `local/tagged-components` - Ensures React components are tagged with `@react`
- `local/prefer-class-methods` - Prefers class methods over arrow function properties
- `local/tsdoc-param-matching` - Validates TSDoc @param tags match parameters
- `local/no-whilst` - Enforces "while" over "whilst" spelling
- `local/no-tiptap-default-import` - Prevents default imports from @tiptap
- `local/no-at-internal` - Prevents using `@internal` tagged APIs in public code
- `local/no-fairy-imports` - Prevents direct imports from fairy directory

[...]

PR #8947, opened 2026-05-27
refactor: enforce tldraw/no-exported-arrow-const repo-wide

frolic:
In order to actually enforce `tldraw/no-exported-arrow-const` everywhere it was originally intended, this PR removes the three `.oxlintrc.json` overrides that turned it off for `packages/editor`, `packages/tldraw`, `packages/utils`, `apps/examples`, and `templates`, and converts the 50 violations those overrides were hiding.

The pattern is mechanical: `export const fn = (args) => {...}` becomes `export function fn(args) {...}`. Async functions get `export async function`. For React component aliases like `const X: TLErrorFallbackComponent = (...) =>` the form becomes `export const X: TLErrorFallbackComponent = function X(...) {...}` — a named function expression keeps the type alias visible to api-extractor (the rule only fires on arrows, not function expressions). No behavior changes.

### Affected directories

| Directory | Files | Violations |
| --- | ---: | ---: |
| `packages/tldraw` | 14 | 19 |
| `apps/examples` | 8 | 16 |
| `packages/editor` | 10 | 11 |
| `packages/utils` | 4 | 4 |
| **Total** | **36** | **50** |

### Background

Relates to #7572. The rule was added in #8258 along with the JS plugin, but with these overrides in place it has never fired in the packages it most needed to apply to. The plugin itself works correctly — see commit history for the exploration that confirmed that.

### API changes

The shift to function declarations changes how api-extractor renders the affected exports — `const X: (args) => Y` becomes `function X(args): Y`. No call-site changes required; the public signatures are equivalent.

Specifically affected `@public` exports:

- `@tldraw/editor`: `DefaultSvgDefs`, `isSafeFloat`, `stopEventPropagation`
- `@tldraw/tldraw`: `handleNativeOrMenuCopy`, `TldrawUiContextualToolbar`, `TldrawUiToolbarToggleGroup`, `TldrawUiToolbarToggleItem`, `truncateStringWithEllipsis`, `useSelectedShapesAnnouncer`, `interpolateSegments`
- `@tldraw/utils`: `Image`, `noop`, `safeParseUrl`, `isAvifAnimated`

`DefaultErrorFallback` keeps its `TLErrorFallbackComponent` alias in the api-report — typed as `export const DefaultErrorFallback: TLErrorFallbackComponent = function DefaultErrorFallback(...) {...}` so the alias survives.

### Change type

- [x] `improvement`

### Test plan

- `yarn oxlint .` — 0 errors.
- `yarn typecheck` — passes.
- `yarn test run` in `packages/editor`, `packages/tldraw`, `packages/utils` — all green.
- `yarn api-check` — passes (api reports are regenerated).

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Core code       | +97 / -88  |
| Automated files | +12 / -12  |
| Documentation   | +24 / -18  |
| Config/tooling  | +0 / -3    |

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -261,7 +261,6 @@
 		{
 			"files": ["packages/editor/**/*", "packages/tldraw/**/*", "packages/utils/**/*"],
 			"rules": {
-				"tldraw/no-exported-arrow-const": "off",
 				"no-restricted-globals": [
 					"error",
 					{
@@ -417,7 +416,6 @@
 			"rules": {
 				"tldraw/no-setter-getter": "off",
 				"tldraw/no-direct-storage": "off",
-				"tldraw/no-exported-arrow-const": "off",
 				"tldraw/img-referrer-policy": "off"
 			}
 		},
@@ -427,7 +425,6 @@
 				"tldraw/prefer-class-methods": "off",
 				"tldraw/no-setter-getter": "off",
 				"tldraw/no-direct-storage": "off",
-				"tldraw/no-exported-arrow-const": "off",
 				"tldraw/img-referrer-policy": "off",
 				"no-console": "off",
 				"tldraw/method-signature-style": "off"
--- a/apps/examples/src/examples/events/derived-view/DerivedViewExample.tsx
+++ b/apps/examples/src/examples/events/derived-view/DerivedViewExample.tsx
@@ -31,7 +31,7 @@ function ShowNumberOfDrawShapesOnPage() {
 	)
 }
 
-export const deriveNumberOfDrawShapesInDocument = (editor: Editor) => {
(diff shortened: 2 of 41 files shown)


mimecuvalo:
huh i didn't know we even had this lint rule. i don't mind the arrow const exports? any reason we should not be using them?

frolic:
A few reasons I was happy to see the linter and tend to prefer `function` def over arrows, at least for exports
- function is hoisted so its usable anywhere in the scope, not just after the declaration
- you can overload the function signature
- something something function display names across debugging tools

mimecuvalo:
fair enough, yeah i'm ambivalent on it — i'd be happy to see this goes out 👍

mimecuvalo (approved):
👍

PR #8948, opened 2026-05-27
feat(tldraw): context menu via right-click in any tool, long-press in select

jsscclr:
In order to make the context menu reachable from every tool without leaving a stray shape (#8277) or showing an empty menu (#8828), this PR splits the two gestures that open it: a **right-click** opens the menu in any tool, and a touch **long-press** opens it only in the **select** tool.

The context menu is a selection surface — every item acts on the current selection. It only makes sense where there's a stable selection to act on, which is why a long-press is scoped to the select tool, while a right-click (which the editor routes through the select tool) works anywhere.

## How it works

- **Right-click** — handled in the editor's event dispatch: a static right-click switches to the select tool before the event reaches the state chart, so the menu reflects the shape under the pointer regardless of the active tool.
- **Long-press** — opens the menu **only in the select tool**. In a discrete shape-creation tool (geo, note, line, text, arrow, frame) a long-press instead **cancels any pending creation** so nothing is left behind (#8277). Continuous-gesture tools (draw, highlight, eraser, laser) keep their gesture.
- **Pointer-type gate** — the menu's Radix trigger opens when `!isCoarsePointer || isIn('select')`. The right-click case keys off the *pointer type* (stable across the gesture), not the tool, so the synchronous switch-to-select can't race the React render and suppress the menu. The select-tool long-press case gates on the tool, which is safe because a long-press never switches tools.

### API changes

- `@tldraw/editor`: new `@internal` export `cancelShapeCreationOnLongPress(editor, cancelPendingCreation)` — cancels a tool's pending creation on a coarse-pointer long-press.
- `tldraw`: `DefaultContextMenuContent` return type narrowed from `JSX.Element | null` to `JSX.Element` (non-breaking).

### Change type

- [x] `bugfix`
- [x] `feature`
- [x] `api`

### Test plan

1. **Right-click** (desktop) in any tool → the context menu opens.
2. Touch **long-press in the select tool** → the menu opens.
3. Touch **long-press in another tool** (geo, note, eraser, draw…) → no menu, no stray shape.

- [x] Unit tests

### Release notes

- Open the context menu with a right-click in any tool, and with a touch long-press in the select tool. A long-press in another tool no longer leaves a stray shape behind.

### Code changes

| Core code       | +156 / -48  |
| Tests           | +285 / -22  |
| Automated files | +4 / -1     |


steveruizok:
Quick followup: I think this is the right direction (to go back to the select tool and to show context menu). I want to make sure that long pressing when a tool is selected doesn't do this accidentally; if there is a collision there, then I'd want the context menu to lose to the long-press-to-drag behavior.

steveruizok:
Specific concern: because long press triggers a right click on mobile, we may need to disable this on mobile in order to avoid accidentally opening the context menu on mobile. But we shall test!

packages/editor/src/lib/editor/tools/cancelShapeCreationOnLongPress.ts
```diff
@@ -0,0 +1,31 @@
+import type { Editor } from '../Editor'
+
+/**
```
mimecuvalo, line 3: nit: for some reason this file feels out of place — i guess let's put it in lib/utils?

mimecuvalo (approved):
approving: small comment not blocking

packages/editor/src/lib/editor/tools/cancelShapeCreationOnLongPress.ts
jsscclr: will merge and address in follow up PR!

PR #8949, opened 2026-05-27
feat(sync): share a single WebSocket across same-user tabs

### LINKED ISSUE #8920: Share a single WebSocket between same-user tabs in a multiplayer room
## Problem

When the same user opens a multiplayer room in multiple tabs, each tab opens its own WebSocket to the sync server and round-trips every change (and every presence update) through the network — even though the tabs are on the same device and could gossip locally.

This has two costs:

1. **Server load and bandwidth.** N tabs of the same user = N WebSocket connections, each sending presence at the collaborative rate.
2. **The solo-mode optimization doesn't help.** `useSync` (`packages/sync/src/useSync.ts:322`) computes `presenceMode` by checking for presence records from *other userIds*. But each tab connects as its own session, so two tabs of the same user appear as two presence rows — though both are filtered out by `userId: { neq: currentUser.get().id }`, so the per-user check itself still says "solo". The real waste is that all N tabs are independently maintaining sockets and pushing presence at `COLLABORATIVE_MODE_FPS` (30) to a server that is just echoing them back to siblings on the same machine.

The local-only path already does the right thing: `TLLocalSyncClient` (`packages/editor/src/lib/utils/sync/TLLocalSyncClient.ts:102`) opens a `BroadcastChannel('tldraw-tab-sync-${persistenceKey}')` and propagates diffs between tabs instantly without touching the network. The multiplayer path has no equivalent.

## Proposed solution

Add a cross-tab coordination layer to `useSync` so that, for a given room, tabs on the same device elect a leader tab that owns the WebSocket. Follower tabs subscribe to changes via `BroadcastChannel` and send their own changes to the leader for forwarding.

Sketch:

- Keyed channel per room, e.g. `BroadcastChannel('tldraw-room-sync-${roomId}')`.
- Leader election via `navigator.locks.request(name, { mode: 'exclusive' }, ...)` — the holder is the leader; release on tab close hands the lock off automatically.
- Leader owns the `TLSyncClient` / WebSocket. Followers run a thin client that reads diffs from the channel and writes its own diffs/presence onto the channel.
- On leader handoff (tab close, navigation), a follower acquires the lock and opens a new socket; in-flight state is reconciled through the existing sync protocol.

This naturally makes the solo-mode optimization cover the multi-tab case: the leader is the only socket, and as long as no other *user* is in the room, it stays at `SOLO_MODE_FPS`.

## Alternatives considered

- **Status quo:** acceptable but wasteful, especially in apps where users routinely keep the same room open in multiple windows.
- **SharedWorker as the single socket owner:** cleaner ownership story than leader election, but `SharedWorker` has weaker browser support (notably gaps on mobile Safari historically) and is harder to debug.
- **Server-side dedupe of same-user sessions:** reduces fan-out but doesn't reduce the number of sockets or the local presence chatter.

## Use cases

- Anyone with the same room open in multiple tabs/windows (very common during development and on tldraw.com).
- Self-hosted sync deployments where halving same-user connection counts is meaningful.
- Making the existing `SOLO_MODE_FPS` optimization actually pay off when a single user has the doc open in several tabs.

frolic:
Closes #8920.

In order to stop opening N WebSockets to the sync server when the same user has a multiplayer room open in N tabs, this PR introduces a cross-tab coordination layer. Tabs on the same device elect a leader tab that owns the only WebSocket; follower tabs send their changes through a `BroadcastChannel` and receive server events the same way. The local-only path (`TLLocalSyncClient`) already does this kind of cross-tab gossip; this PR is the multiplayer equivalent.

### Seam

The cleanest insertion point is `TLPersistentClientSocket` — the interface `TLSyncClient` already takes as its transport. We don't change `TLSyncClient` at all. We add a new `CrossTabSocket` that implements `TLPersistentClientSocket` and flips between two modes over its lifetime:

- **leader mode** — wraps an internal `ClientWebSocketAdapter`, behaves like today
- **follower mode** — `BroadcastChannel` is the transport; the (single) leader tab forwards client-sent events to the real socket and broadcasts server-sent events back

`useSync` swaps `new ClientWebSocketAdapter(...)` for `createCrossTabSocket(...)`. Every tab still runs its own `TLSyncClient` + `TLStore` — optimistic updates, rebase, undo all keep working as-is. The only thing the cross-tab layer does is multiplex N tabs onto 1 socket.

Channel name: `tldraw-room-${userId}-${roomIdOrUriHash}-${permissionHash}` — scoping by user prevents cross-user leakage in shared browser profiles; scoping by permission prevents a readonly tab from piggybacking on a readwrite leader's socket.

### Two roles, not one

We separate two concerns that the original issue conflated:

| Role | Driven by | Why |
|---|---|---|
| **Leader** — owns the WebSocket | `visibilitychange` (sticky among visible tabs) | Visibility is what controls timer throttling and tab freezing — a visible-but-unfocused tab is still healthy, but a hidden tab has its `setInterval` clamped, so pings slow down and the server closes the connection |
| **Presenter** — pushes presence | `focus` (most-recently-focused) | Only the focused tab receives pointer events, so only it has live cursor data; brief blurs (clicking the OS taskbar) shouldn't silence presence, so it sticks until another tab claims |

In the common case (one tab focused at a time) the two roles overlap on the same tab and migrate together. The split matters in two specific scenarios:

- **Side-by-side windows.** Two tldraw tabs visible, one focused. Both stay healthy; only the focused tab pushes its cursor. Clicking between them migrates the presenter (cheap) but not the leader (would otherwise mean a socket close+reopen + a peer-visible cursor leave+rejoin flicker).
- **Focused tab becomes occluded.** Most-recently-focused tab is now hidden; another tab is visible. The leader hands off to the visible tab so the socket stays unthrottled.

### Election

Leader: `navigator.locks.request(...)`. The browser handles mutex, queueing, and automatic release on tab close — no homegrown heartbeat protocol, no split-brain risk. Tabs request the lock when they become visible and want to be leader; the holder releases when it becomes hidden and another tab is visible.

Presenter: cheap last-write-wins gossip on the same channel. On `focus`, post `{ type: 'presenter-claim', tabId, claim: ++counter }`. On receive with a higher `(claim, tabId)` pair, drop `presenceMode` to `'solo'`. The existing `presenceMode` signal in `TLSyncClient.ts` already gates presence pushes — we're just adding a new driver for it.

### Sharp edges

- **`clientClock` remapping.** Every tab's `TLSyncClient` numbers its own push requests independently. The leader's socket sees push requests from itself + N followers all numbered from 0. The leader maintains `(fromTabId, theirClientClock) → ourClientClock` so `push_result` from the server gets routed back to the right tab. This is the only really protocol-y bit and where most of the test effort goes.
- **Sibling patch synthesis.** The server suppresses self-broadcasts for the originating session, so when the leader pushes on behalf of follower B, the *other* followers never see B's change. The leader synthesizes a `patch` for everyone-except-B from the `push_result`'s committed/rebased diff.
- **Handoff gap.** When the leader closes, followers see a brief `offline` until the new leader's socket connects. `TLSyncClient`'s existing reconnect/rebase via `lastServerClock` handles the resync.
- **Fallback when `navigator.locks` is unavailable.** Some embedded webviews don't have it. Detect on startup; if missing, skip the cross-tab layer entirely and open per-tab sockets (today's behavior). Don't try to roll a homegrown election as a fallback — that's where the subtle bugs would live.
- **Auth refresh.** The `uri` resolver can be async (token fetch). Each tab has its own auth context. Only the leader's `uri` is called; that's fine — when leadership migrates, the new leader uses its own resolver, picking up that tab's fresh token.

## Design notes captured here, not in #8920

This part is for reviewers — the original issue framed the work as plain leader election + one socket. Discussion since then surfaced a few things worth flagging:

### 1. Visibility vs focus

Browser throttling hooks on `visibilityState === 'hidden'`, not on focus. A visible-but-unfocused tab is fully unthrottled — pings fire on time, timers are accurate. Only hidden tabs (background tab strips, minimized windows, occluded windows under certain conditions, switched-to-another-desktop) get clamped.

This matters because pointer events go to the focused tab, not just any visible tab. So the signal that's right for "where is my cursor" (focus) is different from the signal that's right for "is this tab healthy enough to own a socket" (visibility). They overlap most of the time but diverge in the side-by-side case.

### 2. Background socket / timer behavior

- WebSockets themselves aren't directly throttled by background-tab status, but the pings around them are. `TLSyncClient` pings every 5s; the server gives up after ~10s. Once timer throttling clamps to 1-minute resolution (after ~5min in background), the server hangs up.
- Tab freeze / discard pauses JS entirely and can kill the tab process; the Web Lock releases cleanly when this happens.
- iOS Safari is aggressive — backgrounded tabs can lose sockets within seconds under memory pressure.
- OS/NAT idle drops happen independently, especially on mobile data.

Net: a hidden leader is a bad leader. The leader-migration rule "hand off to a visible tab if the current leader becomes hidden" is what keeps the shared socket alive.

### 3. Alternatives we considered and rejected

We explored several other architectures — collapsing the two roles into one, a socket-overlap handoff, a store-mirror design where only the leader talks to the server, a `SharedWorker`-owned socket, and a leader-as-relay-server — and intentionally didn't take them. The full write-up, with the reasoning and where each one breaks, lives in [a separate comment](https://github.com/tldraw/tldraw/pull/8949#issuecomment-4591792394) to keep this description focused on the chosen design.

### Three-tab walkthrough (with two roles)

Setup: tabs A, B, C. A is leader+presenter (visible+focused). B, C are followers, hidden.

- **Move cursor in A.** A's `TLSyncClient` derives presence, pushes via its own socket. B and C are followers; their `TLSyncClient` is in solo mode; they receive the server echo back through the channel.
- **Focus B (now both A and B visible, B focused).** Presenter claim sent on channel → A drops to solo, B goes to full. No socket touch. A keeps the lock.
- **Minimize A while B is still visible.** Visibility-change handler on A sees other tabs are visible, releases the lock. B's pending lock request resolves → B becomes leader. New socket from B to server; `lastServerClock` handles resync.
- **Close all tabs except C.** C's pending lock request resolves → C becomes leader+presenter. Standard handoff.

## Module layout

The cross-tab layer lives in `packages/sync-core/src/lib/cross-tab/`, split into small composable pieces rather than one class:

| File | Responsibility |
|---|---|
| `createCrossTabSocket.ts` | Top-level orchestrator; implements `TLPersistentClientSocket`, owns mode switching (leader / follower / fallback) |
| `createLeader.ts` | Web Lock acquisition + visibility-driven want-lock rule |
| `createLeaderRouter.ts` | Leader-side: owns the real WebSocket + routing tables, dispatches server messages |
| `createPushRouting.ts` | `clientClock` + `connectRequestId` routing tables |
| `routeServerMessage.ts` | Stateless server-message dispatch (routing + sibling-patch synthesis) |
| `createPresenter.ts` | Focus-driven `$isPresenter` signal |
| `createFollowerReceiver.ts` | Follower-side channel dispatch |
| `createCrossTabChannel.ts` | Thin typed `BroadcastChannel` wrapper |
| `leaderStatus.ts` | Connection-status ↔ channel-message conversion |
| `defaultEnvironment.ts` | DOM/global resolution + `navigator.locks` feature detection |
| `crossTabLog.ts` | `console.debug` role-transition logging |
| `types.ts` | Shared interfaces / channel-message types |

`useSync.ts` wires it in (new `crossTab` option); `TLSyncClient.ts` only changes to stop throttling document sync in solo mode (the scheduler is `requestAnimationFrame`-driven, so hidden tabs are already throttled by the browser — no need to also drop the target FPS ourselves).

### Change type

- [x] `feature`

### Test plan

1. Open the same room in three tabs of the same browser session as the same user. Verify only one WebSocket is open in the network panel.
2. Move the cursor in each tab; verify only the focused tab's cursor is broadcast to other peers (open the same room as a different user to check).
3. Close the focused tab; verify another tab takes over without other peers seeing the user disappear+reappear.
4. Open the room in two side-by-side browser windows. Click between them; verify no socket churn and that other peers see a single continuous session.
5. Minimize the leader tab while another is visible; verify leadership migrates.
6. Open in a browser that lacks `navigator.locks` (or stub it out); verify graceful fallback to per-tab sockets.
7. Open the same room as two different users in separate browser profiles; verify they don't share a leader (channel scoping by userId).
8. Open the same room readonly in one tab and readwrite in another; verify they don't share a socket (channel scoping by permission).

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Multiple tabs of the same user in the same multiplayer room now share a single WebSocket connection to the sync server, reducing server load and bandwidth.

### API changes

- `@tldraw/sync`: added an optional `crossTab` option to `useSync` (`{ roomKey: string } | boolean`) to control cross-tab WebSocket sharing.
- `@tldraw/sync-core`: added `createCrossTabSocket` and supporting types (`CrossTabSocket`, `CrossTabSocketOptions`, `BroadcastChannelLike`, `BrowserContext`, `CrossTabLockManager`, `UnderlyingSocket`). All `@internal`.

### Code changes

| Core code       | +1,498 / -19 |
| Tests           | +1,671 / -0  |
| Automated files | +56 / -0     |

**This is ~850 lines of actual production code.** The rest of "core code" is comments and blank lines:

| Added lines             | Code   | Comments | Blank | Total |
| ----------------------- | ------ | -------- | ----- | ----- |
| Core code               | ~857   | ~540     | ~101  | 1,498 |
| Tests (`*.test.ts`)     | ~1,163 | ~55      | ~216  | 1,434 |
| Test support (`mocks.ts`) | ~177 | ~40      | ~20   | 237   |

- The high comment share in core code (~540 lines, ~36% of non-blank lines) is deliberate — the cross-tab modules carry a lot of "why" rationale (browser throttling behavior, protocol invariants, race-condition reasoning) rather than restating the code.
- **Tests** = +1,434 (`*.test.ts`) + 237 (`mocks.ts`, test-only support imported solely by the test files).
- **Automated files** = the two `api-report.api.md` reports.
- The 19 core deletions are the removed solo-mode FPS-throttling path in `TLSyncClient.ts`.


frolic:
## Alternatives considered (and why we didn't take them)

A paper trail of architectures we explored for sharing one socket across same-user tabs, and why the current design (per-tab `TLSyncClient` + leader/presenter roles + `clientClock` multiplexing) won out. None of these are "wrong" — they're deliberate trade-offs.

The constraint that shapes all of it: **every tab runs its own optimistic `TLSyncClient`** — instant local edits, per-tab undo, per-tab rebase against the server. The ~290 lines of `clientClock` remapping + sibling-patch synthesis exist precisely to preserve that independence while multiplexing N tabs onto one socket. Every alternative below either keeps that cost or pays for removing it somewhere worse.

### 1. Collapse leader + presenter into a single role

Drive one role off most-recently-focused, on the theory that "if a tab is focused it's also visible, so focus covers both."

Slightly simpler — one Web Lock, one event listener, one channel message type — but loses on two fronts:

- **Side-by-side socket churn.** Every cross-window click causes a leader handoff: old leader closes its socket, new leader opens a fresh one and the server re-hydrates. Other peers see the user leave + rejoin (presence row torn down + recreated), so their cursor flickers.
- **Occlusion edge case.** If the most-recently-focused tab is occluded by another OS window without losing focus, the leader is hidden and timer throttling kicks in. Handling that requires a `visibilitychange` listener anyway — at which point the simplicity win is mostly gone.

A robust one-role design ends up with both a focus listener and a visibilitychange listener doing release-on-hidden — basically the leader half of the two-role design wedged into one role. The delta to add the second role on top is roughly one channel message type, one atom per tab, and ~10 lines of handler code. In exchange, focus changes (the fast-moving signal) become cheap channel-only events instead of full socket handoffs. The trade favors keeping them separate.

### 2. Single-role with socket overlap during handoff

Within the single-role design, soften the handoff cost: keep the old leader's WebSocket open *during* a handoff while the new leader opens its socket in parallel.

1. Tab B claims leadership; A doesn't immediately release.
2. B opens its socket alongside A's, sends `connect`, receives initial state.
3. Once B is operational, A drains pending push results, then closes.
4. B is now sole leader.

This helps your *own* experience — incoming server events keep flowing on A's socket until B is ready, so there's zero read gap. But it does **not** fix the things that actually matter:

- **Two concurrent sessions during overlap.** Each socket is its own server-side session with its own `instance_presence` row, so other peers briefly see two of your cursors. Suppressing B's presence until A closes just trades that for a brief gap where A's presence is removed before B's turns on — the flicker gets shorter, not eliminated.
- **Server protocol mismatch.** Transferring a session in-place (presence row + last-server-clock + auth) without a leave+rejoin would need a new "I'm taking over from sessionId X" form of `connect` — a server-side protocol change with downlevel handling.

Net: ~30–50 lines of overlap-coordination logic that turns a ~100–300ms read gap into ~0ms and shortens (but doesn't remove) the cursor flicker. The two-role design avoids the entire handoff dance on focus changes for free, so this stays on the shelf as the next-best fallback if we ever drop the role split.

### 3. Store-mirror: only the leader talks to the server; tabs share state via local cross-tab sync

The tempting big simplification, and the one we spiked most seriously. Idea: only the leader runs a server-facing `TLSyncClient`; all tabs mirror their *store* to each other over a `TLLocalSyncClient`-style `BroadcastChannel` (the local-only path tldraw already ships). Followers never talk to the server. This would delete the entire ~290-line multiplexing block — roughly 850 → 400 lines.

It doesn't work, for two reasons found by reading the code:

- **Source-flag collision.** Both sync layers emit to their transport on `{ source: 'user', scope: 'document' }` and apply incoming changes via `mergeRemoteChanges` (which tags them `source: 'remote'`), and both ignore `'remote'` to avoid re-broadcasting their own echo. Compose them on one store and propagation breaks **both ways**: a follower's edit arrives at the leader as `'remote'`, so `TLSyncClient` never pushes it to the server; and a server change arrives as `'remote'`, so the local-sync layer never broadcasts it to the other tabs. Bridging this needs origin-tagging + re-emission — rebuilding a relay at the store-diff layer rather than deleting it.
- **Followers have no rebase machinery (the fatal one).** The server *rebases* or *discards* pushes (`push_result` → commit / `rebaseWithDiff` / discard). Handling that means rolling back the optimistic local change, applying the authoritative version, and replaying pending work — exactly what `TLSyncClient`'s `speculativeChanges` / `pendingPushRequests` / `rebase` do. A follower with only a store-mirror can't: an incoming "remote" diff lands on top of its un-rolled-back optimistic edit → divergence/corruption, and the edit is stuck in its undo stack. Giving followers that machinery means giving them a `TLSyncClient` — which puts us right back at clock multiplexing.

Plus presence (scope `'presence'`) isn't mirrored by local sync at all, so each follower's cursor still needs a leader relay. So the store-mirror isn't actually smaller once bridged — it's a similar amount of *subtler, harder-to-test* code with new correctness risk.

### 4. SharedWorker owns the socket

A SharedWorker holds the single WebSocket; no leader election needed.

- If the worker runs one `TLSyncClient` and tabs mirror its store → the same optimistic-rebase problem as #3.
- If tabs keep their own `TLSyncClient` and the worker just multiplexes the socket → we keep all ~290 multiplexing lines (now inside the worker), shed only election (~107), and add worker bundling/lifecycle complexity plus worse browser support (no SharedWorker on Chrome for Android, shaky in some webviews) — so we'd still ship the `navigator.locks` fallback anyway.

Small net win, higher operational cost. `navigator.locks` was the deliberate choice.

### 5. Leader as a relay-server

Followers are full `TLSyncClient`s that connect to the leader as if *it* were the server.

The leader would have to reimplement the server half of the protocol (serverClock assignment, rebase, fan-out) for the local pool — strictly more code than forwarding raw client messages and remapping clocks, which is what we do now.

### Where this leaves us

The current design is close to the minimal multiplexer that preserves per-tab optimistic editing. If the line count is a concern, the safe lever is trimming *within* it — drop the presenter role for v1 (accept the side-by-side cursor flicker) and fold the smallest files — not a rewrite. The ~290 multiplexing lines are load-bearing: they're the direct cost of every tab being an independent optimistic client.

mimecuvalo:
👀 ❗

frolic:
I am a bit scared of the size of this PR for what feels like a fairly limited use case. Chatted with @steveruizok and both decided there's probably an easier path with detecting multiple tabs and allowing it as a multiplayer use case and not throttling updates to 1s.

frolic:
closing in favor of https://github.com/tldraw/tldraw/pull/8988

PR #8952, opened 2026-05-28
fix(docs): canonical URLs for SEO — metadata, sitemap, and footer links

kaneel:
Emit rel=canonical on every article via alias→canonical mapping (quick-start, installation, releases, examples, starter-kits). Filter sitemap to canonical paths only. Point footer, category menu, and 404 nav at canonical hrefs instead of URLs that 301.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### DIFF
--- a/apps/docs/utils/sitemap-canonical-paths.ts
+++ b/apps/docs/utils/sitemap-canonical-paths.ts
@@ -1,0 +1,36 @@
+/** Public marketing apex — canonical `<link>` and sitemap `<loc>` URLs. */
+export const PRODUCTION_SITE_ORIGIN = 'https://tldraw.dev'
+
+/**
+ * Paths that 301 or rewrite to a canonical URL. Sitemap and `<link rel="canonical">`
+ * must use the canonical path so alias URLs do not dilute link equity.
+ */
+const DOCS_SITEMAP_PATH_ALIASES: Record<string, string> = {
+	'/getting-started/quick-start': '/quick-start',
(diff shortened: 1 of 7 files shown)


apps/docs/utils/sitemap-canonical-paths.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Paths that 301 or rewrite to a canonical URL. Sitemap and `<link rel="canonical">`
+ * must use the canonical path so alias URLs do not dilute link equity.
+ */
+const DOCS_SITEMAP_PATH_ALIASES: Record<string, string> = {
+	'/getting-started/quick-start': '/quick-start',
+	'/getting-started/installation': '/installation',
+	'/getting-started/releases': '/releases',
+	'/releases-versioning': '/releases',
+	'/examples': '/examples/basic',
+	'/starter-kits': '/starter-kits/overview',
```
angrycaptain19, line 14: Why are we going /getting-started/quick-start -> /quick-start but /starter-kits -> /starter-kits/overview ?

Feels slightly inconsistent here?

angrycaptain19 (approved):

apps/docs/utils/sitemap-canonical-paths.ts
kaneel: Unfortunately, this requires more changes as the inconsistency here predates this PR.

PR #8913, opened 2026-05-21
chore(infra): switch npm publishing to OIDC trusted publishers

mimecuvalo:
In order to remove the long-lived `NPM_TOKEN` secret from our release pipeline, this PR migrates all npm publishing workflows to npm's [trusted publisher](https://docs.npmjs.com/trusted-publishers) OIDC flow.

Closes #7562.

Why this needs both workflow and script changes: the publish call in `internal/scripts/lib/publishing.ts` used `yarn npm publish`, which doesn't speak npm's OIDC trusted-publisher flow. The publish step now shells out to `npm publish` directly so npm CLI (>= 11.5.1, bundled with our Node 24.13.1 runtime) can exchange the GitHub-issued OIDC token for a short-lived publish token. Provenance attestations are generated automatically under trusted publishing, so no `--provenance` flag is passed.

Workflows updated (the six listed in #7562):

- `publish-new.yml`, `publish-canary.yml`, `publish-patch.yml`, `publish-branch.yml`, `publish-manual.yml` — added `permissions: { contents: read, id-token: write }` to the publish job, removed `NPM_TOKEN` env.
- `bump-versions.yml` — removed unused `NPM_TOKEN` env (`npm show` is unauthenticated; no `id-token` permission needed since this workflow doesn't publish).

Script changes (`internal/scripts/lib/publishing.ts`):

- Dropped the `process.env.NPM_TOKEN` check and the `yarn config set npmAuthToken` / `npmRegistryServer` calls.
- Switched the publish invocation from `yarn npm publish ... --tolerate-republish` to `npm publish ...`. The "package already published" fallback still works — the server-side 403 message we match on is unchanged.

### Prerequisites before merging

1. Trusted publishers need to be configured on npmjs.com for every public package in `packages/`. The current public package list is **15** packages (#7562 listed 13; `@tldraw/driver` and `@tldraw/mermaid` have been added since):
   - `@tldraw/assets`, `@tldraw/driver`, `@tldraw/editor`, `@tldraw/mermaid`, `@tldraw/state`, `@tldraw/state-react`, `@tldraw/store`, `@tldraw/sync`, `@tldraw/sync-core`, `@tldraw/tldraw`, `@tldraw/tlschema`, `@tldraw/utils`, `@tldraw/validate`, `create-tldraw`, `tldraw`
[...]

PR #8954, opened 2026-05-28
fix(infra): restore workspace:* rewrite in published tarballs

kostyafarber:
## Summary

The latest few `tldraw@next` versions on npm ship with `"workspace:*"` literals in their `package.json`, which breaks installs for any consumer outside this monorepo:

```
$ yarn up tldraw@next
Error: @tldraw/driver@workspace:*: Workspace not found (@tldraw/driver@workspace:*)
```

Inspecting the broken tarball confirms the source `package.json` (tab-indented, `workspace:*` deps) is shipped untouched:

```
$ npm pack tldraw@5.1.0-next.d83788c14b6f
$ tar -xzOf tldraw-5.1.0-next.d83788c14b6f.tgz package/package.json | grep '@tldraw/'
        "@tldraw/driver": "workspace:*",
        "@tldraw/editor": "workspace:*",
        "@tldraw/store":  "workspace:*",
```

Comparing to a prior good publish (`5.1.0-next.d7c83ba698ae`) shows the expected output — 2-space-indented `package.json` with rewritten npm versions. The registry metadata also tells the story:

| | Good (`d7c83ba698ae`) | Bad (`d83788c14b6f`) |
|---|---|---|
| `_npmUser` | `tldraw-personal <steve@tldraw.com>` | `GitHub Actions <npm-oidc-no-reply@github.com>` |
| `_npmVersion` | (yarn) | `11.8.0` (npm CLI) |
| provenance | none | OIDC attestation |

## Root cause

PR #8913 swapped the publish call in `internal/scripts/lib/publishing.ts` from `yarn npm publish` to `npm publish`. The PR motivation — switching to npm's OIDC trusted-publisher flow — was correct, but the swap had a silent side effect: `yarn npm publish` rewrites `workspace:*` dependency specifiers into concrete sibling versions during pack, and `npm publish` doesn't (the workspace protocol is a yarn-specific extension that npm has no concept of).

The swap turns out to have been unnecessary. **Yarn 4.10** ([yarnpkg/berry#6898](https://github.com/yarnpkg/berry/pull/6898), Oct 2025) added native OIDC trusted-publishing support. When `yarn npm publish` runs in GitHub Actions with `permissions: id-token: write` granted, yarn requests the GitHub-issued OIDC token, exchanges it with the npm registry, and publishes — exactly like the npm CLI does. Provenance attestations are produced automatically. This repo is already on yarn 4.12 via the `packageManager` field, so OIDC was supported the whole time #8913 was being written.

## Change

This PR reverts only the publish-call change from #8913 and updates the surrounding comment to reflect the yarn-driven OIDC path. The workflow-side changes from #8913 (`permissions: id-token: write` on each publish job, removal of `NPM_TOKEN` env, trusted-publisher configuration on npmjs.com) are unchanged and remain required.

`--tolerate-republish` is restored alongside the yarn invocation since it's a yarn-specific flag; the substring-match catch block (added in #8913 to compensate for npm CLI not having `--tolerate-republish`) is kept as a belt-and-braces defense, because the original code comment notes `--tolerate-republish` is unreliable for canary versions.

## Test plan

- [ ] Merge to `main` and watch `publish.yml` run `canary` mode end-to-end on the next push to `main`.
- [ ] After publish, `npm pack <new-canary-version>` and confirm dependency specifiers in the tarball's `package.json` show concrete versions (e.g. `"@tldraw/editor": "5.1.0-canary.<sha>"`), not `"workspace:*"`.
- [ ] Confirm the new canary version still shows a provenance attestation badge on npmjs.com (yarn 4.10+ produces these automatically under OIDC).
- [ ] In a consumer repo, run `yarn up tldraw@canary` and verify it resolves without the "Workspace not found" error.

After this lands, the `5.1.0-next.*` versions published since #8913 merged are still broken on npm; the registry doesn't allow unpublishing after 72 hours so they'll need to be left as-is, with the next good `next` publish (post-merge) being the recommended upgrade target.


internal/scripts/lib/publishing.ts
```diff
@@ -136,11 +136,17 @@ function topologicalSortPackages(packages: Record<string, PackageDetails>) {
 
 export async function publish(distTag?: string) {
 	// Authentication uses npm's trusted publisher OIDC flow. The publish job in
-	// CI must grant `permissions: id-token: write` so npm CLI (>= 11.5.1, bundled
-	// with our Node version) can exchange the GitHub-issued OIDC token for a
+	// CI must grant `permissions: id-token: write` so yarn (>= 4.10, which we are
+	// on via `packageManager`) can exchange the GitHub-issued OIDC token for a
 	// short-lived publish token automatically. Provenance attestations are
 	// generated by trusted publishing automatically, so we don't pass
 	// `--provenance` explicitly. No NPM_TOKEN is required.
```
MitjaBezensek: Think we might still need provenance flag for yarn? At least until this PR gets merged:
https://github.com/yarnpkg/berry/pull/7018

kostyafarber: I'll investigate, it wasn't there before this PR

kostyafarber: yeah guess when you use npm directly to publish does provenance by default, so we did not need to specify the flag. i'll add it as a flag here when using yarn

kostyafarber: done used the flag as mentioned here https://yarnpkg.com/cli/npm/publish#options-provenance

internal/scripts/lib/publishing.ts
```diff
@@ -160,17 +166,30 @@ export async function publish(distTag?: string) {
...
+						`yarn`,
+						[
+							'npm',
+							'publish',
+							'--tag',
+							String(tag),
+							'--tolerate-republish',
+							'--access',
+							'--provenance',
+							'public',
```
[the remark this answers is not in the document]
kostyafarber: will fix

internal/scripts/lib/publishing.ts
MitjaBezensek: Let's also update the comment, it says "so we don't pass	// `--provenance` explicitly. No NPM_TOKEN is required.`

kostyafarber:
updated the comment @MitjaBezensek

MitjaBezensek (approved):

mimecuvalo (commented):
nice save and find! ty ty

PR #8957, opened 2026-05-29
fix(editor): keep focal point fixed when clamping zoom to min/max

### LINKED ISSUE #8942: Viewport shifts slightly when zooming out to minimum zoom
## Description

When zooming out to the minimum zoom level (`0.05` by default), the viewport position shifts slightly. Unclear whether this is intentional or a bug.

## Steps to reproduce

1. Open the editor at any zoom level.
2. Zoom out repeatedly (mouse wheel, pinch, zoom-out button, or `-` shortcut) until the zoom is clamped at the minimum.
3. Observe the viewport shifting position slightly when the zoom hits the limit.

## Expected

Once zoom is clamped at the minimum, the camera position should stay put — further zoom-out gestures shouldn't translate the viewport.

## Actual

The viewport translates slightly at the moment the zoom clamps to the minimum.

## Video

https://github.com/user-attachments/assets/ea3d92ec-1fab-42e4-8f79-0204bd6afe15

MitjaBezensek:
Thinks this works better than preserving the center, wdyt @steveruizok?

In order to stop the viewport from shifting when a zoom gesture overshoots the zoom limits, this PR makes `getConstrainedCamera` preserve the caller's focal point (e.g. the cursor) when it clamps the zoom, instead of snapping back to the viewport center. Closes #8942.

When you wheel/pinch zoom out past the minimum (or in past the maximum), the gesture computes a camera that keeps the point under the cursor fixed at the *requested* zoom. The old code then clamped the zoom and re-derived `x`/`y` to keep the viewport *center* fixed, discarding the cursor anchor — so the canvas visibly jumped. Now the clamped camera keeps the same focal point fixed. When that focal point happens to be the viewport center (e.g. the zoom-out button), the math reduces to the previous center-preserving behaviour, so nothing else changes. `zoomIn`/`zoomOut` are unaffected because they snap exactly onto zoom steps and never overshoot.

### Before

https://github.com/user-attachments/assets/0cc13c03-2ea5-4642-bcf4-2c5f62e57e4d

### After

https://github.com/user-attachments/assets/b800a301-b101-4e98-8303-3e5801b7bbeb

### Change type

- [x] `bugfix`

### Test plan

1. Open the editor at any zoom level with the pointer away from the screen center.
2. Wheel- or pinch-zoom out until the zoom clamps at the minimum.
3. The point under the cursor should stay put as the zoom hits the limit; further zoom-out gestures should not move the canvas.
4. Repeat zooming in past the maximum.

- [x] Unit tests

### Release notes

- Fix the viewport shifting slightly when zooming past the minimum or maximum zoom level.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +20 / -14  |
| Tests     | +97 / -1   |

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -3336,6 +3336,15 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		let { x, y, z = currentCamera.z } = point
 
+		// `requested` kept the caller's focal point (e.g. the cursor) fixed at
+		// zoom `rz`. When `rz` gets clamped, keep that same focal point fixed at
+		// the clamped zoom `z` rather than snapping to the viewport center.
+		const preserveFocalPoint = (current: number, requested: number, rz: number, z: number) => {
+			const cz = currentCamera.z
+			if (rz === cz) return current
+			return current + ((requested - current) * (1 / z - 1 / cz)) / (1 / rz - 1 / cz)
+		}
+
 		// If force is true, then we'll set the camera to the point regardless of
 		// the camera options, so that we can handle gestures that permit elasticity
 		// or decay, or animations that occur while the camera is locked.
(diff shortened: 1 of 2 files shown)


steveruizok:
Much better! What if I'm using the + - buttons? Those normally zoom toward the center. Will the system switch behaviors?

MitjaBezensek:
> Much better! What if I'm using the + - buttons? Those normally zoom toward the center. Will the system switch behaviors?

Buttons only allow zooming in / out along the predefined steps and can't go past min / max zoom. So this clamp branch is never reached there.

MitjaBezensek:
> Much better! What if I'm using the + - buttons? Those normally zoom toward the center. Will the system switch behaviors?

Took another look. I don't think this should change the behaviour. I changed the functionality in `getConstrainedCamera`, which accepts a `point` param. And if you want to zoom towards center you would just pass in center as that point. The clamping logic would preserve that point.

steveruizok (approved):

Issue #8774, opened 2026-05-06
Research adding passkey support via Clerk on dotcom


mimecuvalo:
Research what it would take to add passkey (WebAuthn) authentication to tldraw.com via Clerk, alongside the existing Google OAuth and email-code sign-in methods.

The dotcom client currently uses `@clerk/clerk-react` and `@clerk/elements` with a custom-built sign-in dialog (`apps/dotcom/client/src/tla/components/dialogs/TlaSignInDialog.tsx`) and `@clerk/backend` in the sync worker. There is no existing passkey, WebAuthn, or MFA code path anywhere in the repo.

Clerk supports passkeys as a first-class authentication factor, so this is mostly an integration and UX question rather than a from-scratch implementation. The goal of this issue is to produce enough information for us to scope the work and decide whether to ship it.

### Feasibility and scope

- What Clerk dashboard configuration is required to enable passkeys (instance settings, allowed authentication strategies, allowed origins, RP ID).
- How passkeys interact with our existing strategies (Google OAuth + email code) on the same Clerk instance, and whether existing users keep working unchanged.
- Whether passkeys can be added as an additional factor for existing users without forcing a migration, and what the enrollment UX looks like.
- Browser, OS, and device coverage we should support, and what graceful fallback looks like for unsupported clients.

### Implementation in our custom sign-in dialog

- What the migration looks like for our hand-rolled `TlaSignInDialog` (which uses `useSignIn`, `useClerk`, and `@clerk/elements`) to add a "Sign in with passkey" path and a passkey enrollment path.
- Where in the dialog flow the passkey option should appear (start step, after email entry, account settings only, etc.) and how it should compose with Google + email code.
- Whether we should expose passkey management in account settings or rely on Clerk's hosted UI for management.
- API surface from `@clerk/clerk-react` and `@clerk/elements` we'd actually use, and whether `@clerk/elements` covers passkeys today or if we need imperative APIs.

### Backend impact

- Whether anything in `apps/dotcom/sync-worker` (`getAuth.ts`, `TLUserDurableObject`, admin routes) needs changes; passkey auth should produce the same Clerk session token, but worth confirming.
- Whether passkey-related events need new handling (e.g. webhook or audit considerations).

### UX, accessibility, and edge cases

- New-device enrollment (user signs in on a new device with no passkey yet) and recovery flows.
- Account recovery if a user loses their passkey and only has passkey enrolled (interaction with email-code as a recovery path).
[...]

PR #8960, opened 2026-05-29
feat(dotcom): global Clerk CAPTCHA mount point

### LINKED ISSUE #8951: Add a global Clerk CAPTCHA mount point for non-sign-in flows
## Problem

Clerk's bot protection requires a `<div id="clerk-captcha">` element to be present in the DOM whenever Clerk client methods that may trigger a CAPTCHA challenge run (`signUp.create`, `signIn.create`, `prepareEmailAddressVerification`, `attemptFirstFactor`, etc.). If the element is not mounted, those calls can fail with a CAPTCHA-related error in production.

We added that element in #7105, but only inside `TlaSignInDialog`:

- `apps/dotcom/client/src/tla/components/dialogs/TlaSignInDialog.tsx` renders `<div id="clerk-captcha" className={styles.clerkCaptcha} />` inside `TldrawUiDialogBody`.

That covers the email sign-in / sign-up flow that runs while the dialog is open. It does not cover any other Clerk call that could happen elsewhere in the app — for example resends triggered after the dialog has closed, future flows like adding a new email address, account recovery, password setup, or any background Clerk call from a hook or page that is not the sign-in dialog.

As we extend authentication functionality (passkeys research is already tracked in #8774, and we have other Clerk surfaces like `TlaLegalAcceptance` and `LoginRedirectPage`), we will keep hitting this gap.

## Proposed solution

Hoist the Clerk CAPTCHA mount point to a place that is always rendered while the app is mounted, so any Clerk call anywhere in the app can attach its widget.

Likely the cleanest location is `TlaRootProviders` (or directly under `ClerkProvider` in `apps/dotcom/client/src/main.tsx`), so the element exists for the lifetime of the page rather than only while the sign-in dialog is open.

Concretely:

- Add a single, always-mounted `<div id="clerk-captcha">` at the root.
- Remove the duplicate one inside `TlaSignInDialog` (or keep it as the visible target inside the dialog and have the root one as a fallback — Clerk only uses the first matching element, so we should pick one).
- Confirm styling so an invisible challenge stays hidden and a visible challenge is positioned reasonably (the dialog currently uses `.clerkCaptcha:not(:empty)` styling — we want similar handling globally).

## Use cases

- Resend code after the sign-in dialog is closed but a Clerk call is still in flight.
- Any future authenticated flow that calls Clerk client methods (adding a verified email, MFA setup, passkey enrollment from #8774, password set/reset, etc.).
- Defensive coverage for any Clerk SDK upgrade that starts requiring CAPTCHA on additional methods.

## Alternatives considered

- Adding a per-dialog/per-flow CAPTCHA element each time we introduce a new Clerk flow. This is what we do today and it scales poorly; every new auth surface has to remember to render the element.
- Using Clerk's hosted UI for everything. We've intentionally built a custom flow for the sign-in dialog and want to keep that flexibility.

## References

- Clerk docs: https://clerk.com/docs/guides/development/custom-flows/authentication/bot-sign-up-protection
- Original CAPTCHA work: #7105
- Related future work: #8774 (passkeys research)

frolic:
In order to stop every new auth surface from having to render its own CAPTCHA target, this PR hoists Clerk's `#clerk-captcha` element to the always-mounted app root (`TlaRootProviders`) so any Clerk client call anywhere in the app can attach its challenge. Closes #8951.

Previously the element only existed inside `TlaSignInDialog` (added in #7105), so any Clerk call outside that dialog — a resend after the dialog closes, or future flows like adding an email, MFA, passkeys (#8774), or password reset — could fail with a CAPTCHA error in production. The element now lives inside the themed `.tla` container for the lifetime of the page, the dialog-scoped duplicate is removed (Clerk only uses the first matching element), and global styling keeps an invisible challenge hidden while centering a visible one over the app.

### Change type

- [x] `improvement`

### Test plan

1. Sign in / sign up with email and confirm the verification flow still works, including the invisible CAPTCHA path.
2. Trigger a resend after closing the sign-in dialog and confirm the Clerk call no longer errors for a missing CAPTCHA target.
3. Inspect the DOM and confirm a single `#clerk-captcha` is present at the app root and that the empty element is invisible / takes no space.

### Release notes

- Mount the Clerk CAPTCHA target globally so authentication flows outside the sign-in dialog no longer fail with a CAPTCHA error.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +21 / -8   |


apps/dotcom/client/src/tla/styles/tla.css
```diff
@@ -127,6 +127,25 @@
...
+   so invisible) for the common invisible challenge. When Clerk injects a
+   visible challenge we center it over the app so the user can complete it. */
+#clerk-captcha:not(:empty) {
+	position: fixed;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%, -50%);
+	/* Same layer as the app-level refresh overlay (MaybeForceUserRefresh) so a
+	   visible challenge sits above dialogs and the editor UI. */
+	z-index: 10000;
```
mimecuvalo: oh hm — i think this is borrowing from that MaybeForceUserRefresh erroneously. let's have it tied to one of our constants. the dialogs use `var(--tl-layer-canvas-overlays)` so probably something above that i imagine

mimecuvalo (approved):
pre-approving modulo one comment! thx

apps/dotcom/client/src/tla/styles/tla.css
frolic: switched to `var(--tl-layer-canvas-blocker)` (renders above overlays) since this is a ~blocker

mimecuvalo (approved):
👍
as long as you're looking at this — i'm realizing that maybe we don't do this logic in computer/fairies/teach/etc. 😬 do you wanna point Claude at this PR + those other projects to make sure they have this support?

frolic:
> 👍 as long as you're looking at this — i'm realizing that maybe we don't do this logic in computer/fairies/teach/etc. 😬 do you wanna point Claude at this PR + those other projects to make sure they have this support?

Yep! Wasn't @max-drake working on consolidating those into a single place? Should we wait for that or go ahead anyway? I don't have context on the consolidation or how long it might take to complete.

mimecuvalo:
> Yep! Wasn't @max-drake working on consolidating those into a single place? Should we wait for that or go ahead anyway? I don't have context on the consolidation or how long it might take to complete.

oh i don't have context on that bit as far as consolidation, but to be clear, you can land this PR — just saying let's create a followup issue to cleanup the other places we use Clerk 👍

frolic:
from Claude
> they all rely on Clerk's prebuilt UI, which handles the CAPTCHA widget automatically. The global-mount fix is specific to the custom hook-based flow on dotcom. The only thing that would change this is if we ever build a custom sign-in/sign-up flow on computer/fairies/teach

frolic:
Also for context, I wasn't able to really reproduce and test this out as the captcha issue that motivated this went away on its own before I got this PR wired up. So this is a spray-and-pray PR 😅

mimecuvalo:
> from Claude
> 
> > they all rely on Clerk's prebuilt UI, which handles the CAPTCHA widget automatically. The global-mount fix is specific to the custom hook-based flow on dotcom. The only thing that would change this is if we ever build a custom sign-in/sign-up flow on computer/fairies/teach

ah right right, cool 👍

PR #8963, opened 2026-05-29
perf(tldraw): skip pressure data on draw shapes when device doesn't report pressure

### LINKED ISSUE #8879: Skip pressure data on draw shapes when device doesn't report pressure
## Problem statement

When a draw shape is created with a device that doesn't report pressure (mouse, trackpad, finger on a non-pressure touchscreen), every point in the shape stores a constant `0.5` for pressure. Browsers report `e.pressure === 0.5` as a default for non-pressure devices, and the drawing tool also explicitly forces `0.5` for non-pen/stylus input in `packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts` (lines 202, 324).

That `0.5` is then encoded as the `z` component of every point in the delta-encoded base64 path stored on each segment (`packages/tlschema/src/misc/b64Vecs.ts`). The first point spends 4 bytes on it (Float32) and every subsequent delta point spends 2 bytes (Float16) — roughly a third of the per-point data — to record a value we already know is constant and meaningless.

The shape itself already records this case via `props.isPen` (set to `this.isPenOrStylus` at creation time in `Drawing.startShape`), so consumers can already tell whether pressure is meaningful.

## Proposed solution

When the draw shape is being created from a non-pressure device (`isPen === false`), omit pressure from the encoded path entirely. Concretely:

- Encode points as 2D (x, y) instead of 3D (x, y, z) for non-pen draw shapes, and store that fact on the segment or shape.
- On decode, supply `z: 0.5` for shapes flagged as non-pressure so downstream freehand/stroke code doesn't need a special case.
- Existing pressure shapes (pen/stylus) keep the current 3D encoding unchanged.

This is roughly a 33% reduction in the size of segment path data for the common no-pressure case (mouse, trackpad, basic touch).

## Alternatives considered

- **Leave it as is.** The waste is small per point but adds up on long shapes and large documents; it also bloats sync payloads and persistence storage.
- **Quantize or compress z when constant.** Possible but adds branching to every consumer of segment paths and doesn't help when the constant just happens to be `0.5`.
- **Drop z at the point-array level (`{ x, y }` only).** Cleaner type-wise but more invasive — `VecModel` has `z?: number` and many consumers already tolerate it being absent. The encoder is the cheapest place to make the change.

## Use cases

- Smaller draw shapes in saved documents for users on mouse/trackpad/finger — the majority case for tldraw.com.
- Smaller multiplayer sync messages when many users are drawing without pressure devices.
- No behavior change for users on pen/stylus.

## Notes

- A new segment encoding variant + schema migration would likely be needed so the decoder knows whether to expect 2D or 3D points.
- `packages/tlschema/src/misc/b64Vecs.ts` is the encode/decode site; `packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts` is the producer site.
- Adjacent but separate from #8877 (user-level pressure preferences) — that issue is about remapping reported pressure; this one is about not storing pressure values that were never reported.

jsscclr:
Closes #8879. In order to shrink the on-the-wire and on-disk size of draw shapes created without a pressure-aware device, this PR encodes their segment paths as 2D (x, y) instead of 3D (x, y, z), skipping the constant `0.5` pressure value that browsers report for mouse, trackpad, and basic touch input.

For non-pressure shapes that's roughly a third of the per-point bytes (2 Float16 deltas instead of 3), which compounds over long strokes, multi-shape drawings, and multiplayer sync messages. Pen/stylus input is unchanged.

### How it works

- A new optional `dim?: 2 | 3` field on `TLDrawShapeSegment` records the encoding format. Undefined means 3D, preserving full backward compatibility.
- `b64Vecs` gains 2D variants (`encodePoints2D`, `decodePoints2D`, `decodeFirstPoint2D`, `decodeLastPoint2D`) and the existing methods take an optional `dim` parameter that routes through them. Callers just pass `segment.dim` through.
- `Drawing.ts` decides the encoding once in `startShape` from the input device (`segmentDim = 2` when `!isPenOrStylus`) and creates new segments through a small `makeSegment` helper. The decision is device-driven rather than content-driven so the encoded bytes are deterministic and chosen in O(1) at stroke start.
- The fallback base64 encoder/decoder now handles non-multiple-of-3 byte lengths with `=` padding (the 2D byte layout — 8 bytes for the first point + 4 bytes per delta — requires it).
- A new `OmitNonPressureZ` migration on both draw and highlight shapes converts any 2D segments back to 3D in its `down` step so older clients can still read newer data.
- Dot detection (`getIsDot` on draw and highlight) now calls `b64Vecs.isSinglePoint(path, dim)` instead of comparing the raw `path.length` to a hard-coded `24`. A single point is 16 base64 chars in 3D but only 12 in 2D, and a real two-point 2D stroke is 16 chars — the same length as a 3D dot — so a bare length threshold cannot tell them apart. Keeping the single-point length in `b64Vecs`, next to the byte layout, makes the check correct for every encoding and impossible to drift when a new format is added. This resolves the review feedback about the hardcoded threshold.

### Change type

- [x] `improvement`

### Test plan

1. Draw a freehand shape with a mouse or trackpad; confirm rendering is unchanged.
2. Inspect the resulting shape's segments — `dim` should be `2`, `path` should be ~33% shorter than before.
3. Draw with a pen/stylus or simulate pressure input; segments should have no `dim` field (3D encoding, unchanged behavior).
4. Open a document saved before this change; existing draw shapes still render correctly.
5. Multiplayer: a 2D-encoded shape from a new client should still load on an older client (down migration converts back to 3D).
6. Draw a very short two-point stroke with a mouse or trackpad; it renders as a stroke (not a dot) and keeps its resize/rotate handles.

- [x] Unit tests

### Release notes

- Shrink stored draw shape data for mouse / trackpad / finger input by ~33% by omitting the redundant constant-pressure value from the encoded path.

### API changes

- Added `b64Vecs.encodePoints2D`, `b64Vecs.decodePoints2D`, `b64Vecs.decodeFirstPoint2D`, `b64Vecs.decodeLastPoint2D` for the 2D path format.
- Added an optional `dim?: 2 | 3` parameter to `b64Vecs.encodePoints`, `b64Vecs.decodePoints`, `b64Vecs.decodeFirstPoint`, `b64Vecs.decodeLastPoint`; when `2`, they route through the 2D variant.
- Added `b64Vecs.isSinglePoint(path, dim)` for length-based single-point ("dot") detection that is correct across the 2D and 3D encodings.
- Added an optional `dim?: 2 | 3` field on `TLDrawShapeSegment`.
- Added `drawShapeVersions.OmitNonPressureZ` (v5) and `highlightShapeVersions.OmitNonPressureZ` (v4).

### Code changes

| Core code       | +319 / -72 |
| Tests           | +212 / -7  |
| Automated files | +11 / -6   |


packages/tlschema/src/shapes/TLDrawShape.ts
```diff
@@ -19,9 +19,16 @@ export interface TLDrawShapeSegment {
...
+	 * 3D format (default): first point Float32 x/y/z (12 bytes), subsequent points Float16 dx/dy/dz (6 bytes each).
+	 * 2D format: first point Float32 x/y (8 bytes), subsequent points Float16 dx/dy (4 bytes each); decode supplies z = 0.5.
 	 */
 	path: string
+	/**
+	 * Encoding dimension of `path`. `2` is used when the shape was drawn with a device that
+	 * does not report pressure (mouse, trackpad, finger) — z is implicitly 0.5 and not stored.
+	 * `3` or absent means x, y, and z are all stored.
+	 */
+	dim?: 2 | 3
```
mimecuvalo, line 30: let's make `2` and `3` both a constant (the way `DEFAULT_PRESSURE` is) — it'll be hard to tell what the `2`s around the code are doing

packages/tldraw/src/lib/shapes/draw/toolStates/Drawing.ts
```diff
@@ -180,6 +186,11 @@ export class Drawing extends StateNode {
 		)
 	}
 
+	private makeSegment(type: 'free' | 'straight', points: VecModel[] | Vec[]): TLDrawShapeSegment {
+		const path = b64Vecs.encodePoints(points as VecModel[], this.segmentDim)
+		return this.segmentDim === 2 ? { type, path, dim: 2 } : { type, path }
```
mimecuvalo: like here and everywhere below

packages/tlschema/src/misc/b64Vecs.ts
```diff
@@ -67,11 +72,21 @@ function nativeBase64ToUint8Array(base64: string): Uint8Array {
 
 /** @internal */
 export function fallbackBase64ToUint8Array(base64: string): Uint8Array {
-	const numBytes = Math.floor((base64.length * 3) / 4)
+	// Strip up to 2 '=' padding characters from the end to determine actual byte count
+	const paddedLength = base64.length
+	let padding = 0
+	if (paddedLength > 0 && base64.charCodeAt(paddedLength - 1) === 61) {
```
mimecuvalo: 61 is a bit magical - let's define that as a `const` so that we know it's referring to `=` - even better if we just use `=`.charCodeAt(0) somewhere once to make it explicit

packages/tlschema/src/misc/b64Vecs.ts
```diff
@@ -110,6 +141,20 @@ export function fallbackUint8ArrayToBase64(uint8Array: Uint8Array): string {
 			BASE64_CHARS[bitmap & 63]
 	}
 
+	// Handle trailing 1 or 2 bytes with '=' padding
+	const remaining = len - fullGroups
+	if (remaining === 1) {
+		const bitmap = uint8Array[fullGroups] << 16
+		result += BASE64_CHARS[(bitmap >> 18) & 63] + BASE64_CHARS[(bitmap >> 12) & 63] + '=='
```
mimecuvalo: a comment and/or const for what 63 is doing would also be helpful

packages/tlschema/src/misc/b64Vecs.ts
```diff
@@ -110,6 +141,20 @@ export function fallbackUint8ArrayToBase64(uint8Array: Uint8Array): string {
 			BASE64_CHARS[bitmap & 63]
 	}
 
+	// Handle trailing 1 or 2 bytes with '=' padding
+	const remaining = len - fullGroups
+	if (remaining === 1) {
```
mimecuvalo, line 167: how was this working before without this logic? we didn't add the trailing `=`s i think?
my fuzzy recollection of base64 parsing was that i think the `=` aren't strictly necessary — can you check?

mimecuvalo (commented):
neat! i imagine @steveruizok might have some opinions here possibly since his heart is close to the draw code

packages/tlschema/src/misc/b64Vecs.ts
jsscclr: We didn't need any padding before, because a 3D path gives you 3 bytes mapping to 4 chars. With 2D, sometimes you get a multiple of 4, sometimes not.

Made a little playground to make it easier to grok: https://hsmsds.csb.app/

packages/tlschema/src/shapes/TLDrawShape.ts
jsscclr: Updated - I've left the actual type declaration as `2 | 3` (seems to be the convention whilst also avoiding type golf) but any runtime values point to the constants.

packages/tlschema/src/misc/b64Vecs.ts
mimecuvalo: neat!

PR #8968, opened 2026-05-29
fix(tldraw): keep dragged shape selected when pasting mid-interaction

### LINKED ISSUE #8305: Pasting while dragging an arrow terminal interrupts binding

frolic:
In order to fix #8305, this PR makes two related changes so that pasting (Cmd+V / Ctrl+V) while dragging an arrow terminal, resizing, or translating a shape no longer disrupts the interaction. Paste itself is intentionally left to go through (it's only suppressed while editing text), per the discussion on the issue.

**1. Keep the dragged shape selected.** Pasting mid-interaction creates and selects the pasted shape. Previously that selection persisted, so when the interaction ended the pasted shape stayed selected instead of the shape the user was actually dragging/resizing. The fix lives in the `SelectTool` state machine rather than the clipboard handler: `DraggingHandle`, `Resizing`, and `Translating` now restore the selection to the interacted shape(s) in their `complete()` handlers, just before transitioning back to idle. `setSelectedShapes` short-circuits when the selection already matches, so this is a no-op in the normal case and only takes effect when something (like a paste) changed the selection mid-interaction.

**2. Keep the arrow binding hint visible while dragging.** The dashed binding hint (`ArrowBindingHintOverlayUtil`) was gated on the only-selected shape being the arrow, so when a paste stole the arrow's selection mid-drag the hint disappeared until the drag ended. It now resolves the hinted arrow from the handle-drag/point interaction state (`select.dragging_handle` / `select.pointing_handle`) and only falls back to the current selection otherwise, so the hint stays visible throughout the drag.

### Change type

- [x] `bugfix`

### Test plan

1. Create a geo shape and an arrow on a tldraw canvas, and copy any shape to the clipboard.
2. Press and hold on the arrow's end handle and drag it toward the geo shape — the dashed binding hint appears.
3. While the pointer is still down, press Cmd+V. The binding hint stays visible (it no longer disappears until the drag ends).
4. Release the pointer — the arrow (not the pasted shape) is selected.
5. Repeat with a shape resize and a shape translate — in both cases the shape you were manipulating stays selected after pasting mid-drag.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix paste while dragging an arrow handle, resizing, or moving a shape stealing the selection from the shape being manipulated, and the arrow binding hint disappearing mid-drag after a paste.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +40 / -5   |
| Tests     | +143 / -0  |

### DIFF
--- a/packages/tldraw/src/lib/overlays/ArrowBindingHintOverlayUtil.ts
+++ b/packages/tldraw/src/lib/overlays/ArrowBindingHintOverlayUtil.ts
@@ -10,6 +10,8 @@ import {
 import { TLArrowInfo } from '../shapes/arrow/arrow-types'
 import { getArrowInfo } from '../shapes/arrow/getArrowInfo'
 import { getArrowBindings } from '../shapes/arrow/shared'
+import { DraggingHandle } from '../tools/SelectTool/childStates/DraggingHandle'
+import { PointingHandle } from '../tools/SelectTool/childStates/PointingHandle'
 
 /** @public */
 export interface TLArrowBindingHintOverlay extends TLOverlay {
@@ -41,10 +43,6 @@ export class ArrowBindingHintOverlayUtil extends OverlayUtil<TLArrowBindingHintO
 	override isActive(): boolean {
 		const editor = this.editor
 		if (editor.getIsReadonly()) return false
-		const id = editor.getOnlySelectedShapeId()
-		if (!id) return false
-		const shape = editor.getShape(id)
-		if (!shape || shape.type !== 'arrow') return false
 		if (
 			!editor.isInAny(
 				'select.idle',
@@ -56,12 +54,16 @@ export class ArrowBindingHintOverlayUtil extends OverlayUtil<TLArrowBindingHintO
 		) {
 			return false
 		}
+		const id = this.getActiveArrowId()
+		if (!id) return false
+		const shape = editor.getShape(id)
+		if (!shape || shape.type !== 'arrow') return false
 		const bindings = getArrowBindings(editor, shape as TLArrowShape)
 		return Boolean(bindings.start || bindings.end)
 	}
 
 	override getOverlays(): TLArrowBindingHintOverlay[] {
-		const id = this.editor.getOnlySelectedShapeId()
+		const id = this.getActiveArrowId()
 		if (!id) return []
 		return [
 			{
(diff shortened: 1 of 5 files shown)


max-drake:
digging into this, I think my comment on #8305 might have been partially misleading (sorry!). especially looking at the fix necessary for the arrow binding hint, which feels a little hackier than we generally want. I realized digging around there might be nice entry point for changing this behavior in `defaultHandleExternalTldrawContent()`? I tried setting `select` false in the opts for `putContentOntoCurrentPage()` if we were in any states where we care about this behavior, and that seemed to do this trick. 

however! this creates some weird behavior with undoing (perhaps related to #8969) where cmd+z removes the pasted shape and resets the translated/rotated/resized shape back to the state it was in when the other shape was pasted, which obviously isnt ideal! I'm not sure how thorny that fix will be, but I tihnk it's worth it for you to investigate, so I'm marking this as draft for now, and once you have a draft pr open for the other approach we can pair and compare them.

sorry again for misleading you haha. the tldraw repo is vast and there are many parts I have yet to touch

frolic:
Makes sense! I tossed up the alternative here, which does feel lighter/cleaner: https://github.com/tldraw/tldraw/pull/8976

I think we can probably tackle the paste history issue (#8969) separately as its a bit hairy.

frolic:
closing in favor of https://github.com/tldraw/tldraw/pull/8976

PR #8973, opened 2026-06-01
fix(tldraw): fix the shortcut associated to hoovering style

### LINKED ISSUE #8962: q is bound to both toggle tool lock and copy hovered styles
🧑: in the examples app develop route, q is used to toggle tool lock AND to vacuum shape styles

---

## Description

Pressing `q` fires two actions at once because two default actions are registered to the same keyboard shortcut:

- `toggle-tool-lock` — `kbd: 'q'` ([`actions.tsx:1407`](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/ui/context/actions.tsx#L1407))
- `copy-hovered-styles` — `kbd: 'q'` ([`actions.tsx:1855`](https://github.com/tldraw/tldraw/blob/main/packages/tldraw/src/lib/ui/context/actions.tsx#L1855), added in #8917)

The keyboard shortcut handler in `useKeyboardShortcuts.ts` iterates the full registry on each keydown and invokes **every** registration whose parsed key matches the event — there is no first-match-wins. So a single `q` press toggles tool lock *and* (when hovering a shape with the select tool idle) copies the hovered shape's styles onto the selection / next shapes.

This affects the default SDK UI everywhere, not only the examples `develop` route — that is just where it was noticed.

## Steps to reproduce

1. Open any tldraw editor (e.g. the examples app `develop` route at `localhost:5420`).
2. Hover a shape that has styles.
3. Press `q`.

## Expected behavior

`q` performs a single, well-defined action.

## Actual behavior

`q` both toggles tool lock and copies the hovered shape's styles. The keyboard shortcuts dialog also lists `q` for two different actions.

## Notes

The two actions should not share a binding. Either `copy-hovered-styles` or `toggle-tool-lock` needs a different shortcut, or the behaviors should be reconciled into one intentional binding.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8962 

Following the release of the "hoovering style" feature ( https://github.com/tldraw/tldraw/pull/8917 ), it's been found `q` has been used twice for shortcuts.

I'm moving the shortcut to Q (shift+q) for hoovering styles (we can discuss this) 

+ adding a test that make sure we're not reusing the same shortcuts 
+ adding a missing record of the action, right now undoing goes two steps before because we're missing this.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape.
2. Create an other shape with different styles
3. Select one of the shape
4. hover the other and press shift+q
5. the selected shape hoovered the style of the hovered shape (bit complicated to read this one)
6. press ctrl+z to under 
7. the shape should go back to the previous styles

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Added an ui event for `copy-hovered-styles`

### Release notes

-Fixed a keyboard shortcut conflict where q both toggled tool lock and copied hovered styles; copy hovered styles is now shift+q.
- Added a step record of the copy style features when the editor is not in select mode

### DIFF
--- a/apps/examples/e2e/tests/test-kbds.spec.ts
+++ b/apps/examples/e2e/tests/test-kbds.spec.ts
@@ -202,6 +202,14 @@ test.describe('Keyboard Shortcuts', () => {
 			data: { source: 'kbd' },
 		})
 	})
+
+	test('Copy hovered styles', async () => {
+		await page.keyboard.press('Shift+q')
+		expect(await page.evaluate(() => __tldraw_ui_event)).toMatchObject({
+			name: 'copy-hovered-styles',
+			data: { source: 'kbd' },
+		})
+	})
 })
 
 test.describe('Actions on shapes', () => {
--- a/packages/tldraw/src/lib/ui/context/actions.tsx
+++ b/packages/tldraw/src/lib/ui/context/actions.tsx
@@ -1852,7 +1852,7 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 			{
 				id: 'copy-hovered-styles',
 				label: 'action.copy-hovered-styles',
-				kbd: 'q',
+				kbd: 'shift+q',
 				async onSelect() {
 					const hovered = editor.getShapeAtPoint(editor.inputs.getCurrentPagePoint(), {
 						hitInside: true,
@@ -1866,16 +1866,26 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 
 					if (!hovered || !isIdle) return
 
-					for (const style of editor.styleProps[hovered.type].keys()) {
-						const value = editor.getShapeStyleIfExists(hovered, style)
-						if (value === undefined || style === GeoShapeGeoStyle) continue
-
-						if (editor.isIn('select')) {
-							editor.setStyleForSelectedShapes(style, value)
-						} else {
-							editor.setStyleForNextShapes(style, value)
-						}
+					// Only applying styles to selected shapes modifies the document, so only that
+					// case needs a history stopping point. Setting styles for the next shape is
+					// instance state and shouldn't be undoable.
+					const isInSelect = editor.isIn('select')
+					if (isInSelect && editor.getSelectedShapeIds().length > 0) {
+						editor.markHistoryStoppingPoint('copy-hovered-styles')
 					}
+
+					editor.run(() => {
+						for (const style of editor.styleProps[hovered.type].keys()) {
+							const value = editor.getShapeStyleIfExists(hovered, style)
+							if (value === undefined || style === GeoShapeGeoStyle) continue
+
+							if (isInSelect) {
+								editor.setStyleForSelectedShapes(style, value)
+							} else {
+								editor.setStyleForNextShapes(style, value)
+							}
+						}
+					})
 				},
 			},
 		]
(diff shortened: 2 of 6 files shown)


packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
```diff
@@ -328,6 +328,26 @@ function parseKbd(kbd: string): ParsedKbd[] {
 	return out
 }
 
+function serializeParsedKbd(parsed: ParsedKbd): string {
```
mimecuvalo: these fns are only for tests so let's not ship this code with the SDK - just put this in the test if needed

packages/tldraw/src/test/ui/keyboardShortcuts.test.tsx
```diff
@@ -0,0 +1,70 @@
+import { useEffect } from 'react'
```
mimecuvalo, line 1: can you also augment test-kbds.spec.ts? i think might be useful to test there possibly

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1852,7 +1852,7 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
 			{
 				id: 'copy-hovered-styles',
 				label: 'action.copy-hovered-styles',
-				kbd: 'q',
+				kbd: 'shift+q',
```
mimecuvalo, line 1855: i don't know about shift+q @steveruizok prbly has opinions here for sure

kaneel: yes, I mentioned in the PR that I'm happy to discuss about figuring a better one, fwiw it's not that hard to reach for shift+q

packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
kaneel: good point!

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1866,16 +1866,26 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
-					for (const style of editor.styleProps[hovered.type].keys()) {
-						const value = editor.getShapeStyleIfExists(hovered, style)
-						if (value === undefined || style === GeoShapeGeoStyle) continue
+					// Only applying styles to selected shapes modifies the document, so only that
+					// case needs a history stopping point. Setting styles for the next shape is
+					// instance state and shouldn't be undoable.
+					const isInSelect = editor.isIn('select')
+					if (isInSelect && editor.getSelectedShapeIds().length > 0) {
+						editor.markHistoryStoppingPoint('copy-hovered-styles')
+					}
```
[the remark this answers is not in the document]
kaneel: that's because we don't want to record history mark points for just changing the current style.

mimecuvalo:
@steveruizok to review kbd shortcut choice here

apps/examples/e2e/tests/test-kbds.spec.ts
```diff
@@ -202,6 +202,14 @@ test.describe('Keyboard Shortcuts', () => {
...
 		})
 	})
+
+	test('Copy hovered styles', async () => {
+		await page.keyboard.press('Shift+q')
+		expect(await page.evaluate(() => __tldraw_ui_event)).toMatchObject({
+			name: 'copy-hovered-styles',
+			data: { source: 'kbd' },
+		})
+	})
```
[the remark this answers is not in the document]
kaneel: well it does pass so 🤷

kaneel:
Closing this in favour of a future PR solving every problems (touching wood) this feature ever had

Issue #8969, opened 2026-05-29
Pasting while dragging merges the paste and drag into one undo step


frolic:
### Description

When you paste (Cmd/Ctrl+V) while dragging an arrow terminal — or resizing/translating a shape — the paste and the in-progress drag land in the same undo entry, so a single Cmd/Ctrl+Z undoes both the paste and (part of) the drag instead of just the paste.

This is the remaining sub-item from #8305. The selection-stealing and binding-hint regressions from that issue were fixed in #8968; the undo grouping was left as a separate concern.

### Steps to reproduce

1. Create a shape and an arrow, and copy any shape to the clipboard.
2. Press and hold the arrow's end handle and start dragging it.
3. While still dragging, press Cmd/Ctrl+V.
4. Release the pointer.
5. Press Cmd/Ctrl+Z once.

### Expected

Undo removes only the pasted shape; a second undo reverts the arrow drag. The paste and the drag are independent undo steps.

### Actual

A single undo removes the pasted shape together with the portion of the drag that happened after the paste. The drag is effectively split around the paste moment, so it feels like one undo reverts "both".

Note: native paste is async, so this only reproduces when the clipboard read resolves while the pointer is still down. If it resolves after pointer-up, the two operations already separate cleanly.

### Technical notes

`markHistoryStoppingPoint` flushes the current `pendingDiff` into the undo stack and pushes a `stop` marker (`HistoryManager._mark`). A plain `undo()` clears `pendingDiff`, then walks down undoing diffs until the first `stop` (`HistoryManager._undo`).

Pasting mid-drag injects a `stop 'paste'` marker into the middle of the live drag. The drag never re-marks, so the pasted shape and the rest of the drag accumulate into one shared `pendingDiff` above that marker — hence one undo removes both.

The drag and paste overlap in time, so cleanly splitting them isn't free. Options considered:

- **Defer the paste until the interaction ends** (drain on return to `select.idle`). Cleanest result — the drag is one entry, the paste a separate entry on top — at the cost of the pasted content not appearing until pointer-up. No `HistoryManager` changes.
- **Re-mark around the paste** (pre-drag / paste / post-drag). Simplest code, but splits the drag into two entries, so undo order is confusing.
- **Splice the paste diff beneath the drag's mark.** Two clean entries in intuitive order, but requires `HistoryManager` surgery and history reordering (sketchy w.r.t. the redo stack and sync).

kaneel:
[...]

PR #8976, opened 2026-06-01
fix(tldraw): don't steal selection when pasting mid-interaction

frolic:
## Overview

In order to keep an in-progress interaction intact when content is pasted mid-gesture, this PR stops `defaultHandleExternalTldrawContent` from selecting the pasted shapes while the user is dragging a handle, translating, resizing, or rotating.

This is an **alternative** to #8968 for the same underlying problem (#8305). Where #8968 keeps the binding-hint overlay alive by gating the overlay's `isActive`/`getOverlays` on the interaction state, this PR instead avoids the root cause for the paste case: the pasted content no longer steals selection from the shape being manipulated. Opened as a draft so the two approaches can be compared side by side.

When the user pastes tldraw content while mid-interaction — for example dragging an arrow's endpoint handle — the default handler used to `select` the newly pasted shapes. That stole selection from the shape under manipulation and interrupted the gesture (the arrow's in-progress binding hint, for instance, disappeared until the drag ended). With this change the selection is left untouched while mid-interaction, so the gesture continues uninterrupted. When not mid-interaction, paste still selects the pasted content as before.

Because we no longer reselect mid-interaction, the selection bounds before and after paste are identical, which would make the paste "puff" overlap check always pass. The puff is meant to signal that the newly-selected pasted content landed on the old selection, so it's now skipped while mid-interaction.

## Scope note

The known undo-grouping quirk tracked in #8969 (paste landing in the same undo entry as the drag) is **intentionally left out of scope** here, to keep this change focused and comparable to #8968.

## Change type

- [x] `bugfix`

## Test plan

1. Create a shape and an arrow bound to it.
2. Start dragging the arrow's free endpoint handle.
3. While still dragging, paste some tldraw content.
4. The arrow stays selected and its binding hint remains visible; the pasted content is added without stealing selection.
5. Paste while idle — the pasted content is selected as normal.

- [x] Unit tests added in `SelectTool.test.ts` covering dragging a handle, translating, resizing, the no-puff case, and the idle (still-selects) case.

## Release notes

- Pasting content while dragging, translating, resizing, or rotating no longer steals selection from the shape being manipulated.

## Code changes

| Area | Files | +/- |
| --- | --- | --- |
| Core | `defaultExternalContentHandlers.ts` | +19/-1 |
| Tests | `SelectTool.test.ts` | +134/-0 |

Relates to #8305. Alternative to #8968.

### DIFF
--- a/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
+++ b/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
@@ -644,9 +644,10 @@ export async function defaultHandleExternalTldrawContent(
 		if (
 			// When mid-interaction we don't select the pasted content, so the
 			// selection is unchanged and the before/after bounds are identical —
-			// the overlap check would always pass. The puff signals that the
-			// newly-selected pasted content landed on the old selection, which
-			// is meaningless when we didn't change the selection, so skip it.
+			// the overlap check would always pass. The selection flash below
+			// signals that the newly-selected pasted content landed on the old
+			// selection, which is meaningless when we didn't change the
+			// selection, so skip it.
 			!isMidInteraction &&
 			selectionBoundsBefore &&
 			selectedBoundsAfter &&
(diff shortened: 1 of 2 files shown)


max-drake (approved):
looks great! I agree with your comment re: the undo stack issue is a little haunted and I think we can solve it in a different pr

PR #8981, opened 2026-06-01
fix(editor): resolve line-height to whole pixels for cross-browser text

### LINKED ISSUE #8970: Cross-browser drift: draw strokes misalign with text in multiplayer (Chrome vs Safari), worsens per line
Image

### Summary

When two users collaborate in a multiplayer room, **freehand draw strokes** (`draw` shapes) that are visually placed under **text** lines appear correctly for the drawer, but appear with an **increasing vertical offset** for viewers on a different browser engine. The offset **accumulates line-by-line** (roughly from line ~10 onward in our tests).

This reproduces on **tldraw.com** (official infrastructure), on the **tldraw SDK** with `useSyncDemo` / `useSync`, and is **not** caused by application-specific code (confirmed with a vanilla `<Tldraw store={...} />` page with zero custom logic).

### Steps to reproduce

1. Open a shared tldraw.com board (Google login → Share → copy link).
2. Open the **same share URL** in **Google Chrome** (Client A) and **Safari** (Client B) on macOS.
3. In Client A, use the **Text** tool and create a vertical column of many text lines (we used repeated single characters / short words, ~15–20 lines).
4. In Client A, use the **Draw** tool and draw **horizontal strokes** under each text line (intended as visual underlines).
5. Compare both clients side by side.

### Expected behavior

Draw stroke positions should match on all clients, since shape coordinates are synced.

### Actual behavior

- Client A (drawer, Chrome): strokes align with text as intended.
- Client B (Safari): strokes are vertically shifted relative to the text; the shift **grows with each additional line** further down the column.
- Reversing roles (Safari draws, Chrome views) reverses which client sees the offset.

### Additional observations

| Scenario | Result |
|----------|--------|
| Chrome + Chrome | OK |
| Safari + Safari (Mac + iPad) | OK |
| Chrome + Safari | **Drift** |
| iPad “Chrome” (WebKit) + Safari | OK |

This strongly suggests **cross-browser text layout / font metrics differ**, while draw shape page coordinates sync correctly. Draw strokes and text are independent shapes (not anchored), so visual alignment depends on each client rendering text at the same vertical positions.

### Environments tested

- **tldraw.com** — official product, Google login, shared room (confirmed reproducible)
- **tldraw SDK v4.1.1** — vanilla `<Tldraw />` + `@tldraw/sync` self-hosted (same behavior)
- macOS: Chrome + Safari
- Also observed earlier on Feroclass Board (custom app); ruled out after vanilla SDK + tldraw.com repro

Image

### Possible related areas

- Text measurement / line height differences between WebKit and Chromium
- Prior Safari text fixes: #1980, #3417
- Export text offset: #5651

### Impact

Users collaborating across Chrome and Safari (common in education / tutoring) see misaligned annotations when using draw strokes to underline or mark up text — even though multiplayer sync itself works (cursors, shape creation, etc.).

---

**Reporter context:** Independent tutoring platform (Feroclass) integrating tldraw SDK. We spent time isolating this before reporting. Happy to provide more screenshots, `.tldr` exports, or join a debug session.

frolic:
In order to stop draw strokes and text labels drifting apart vertically between browsers in multiplayer, this PR resolves tldraw's unitless `line-height` to a whole-pixel value before it reaches the DOM. Fixes #8970.

The theme default `line-height` is unitless (`1.35`), so the per-line advance is fractional — e.g. at size "m" the font is `24px` and the line advance is `24 * 1.35 = 32.4px`. WebKit snaps each wrapped line box to a whole pixel (legacy integer line layout) while Blink keeps the fraction, so on Chrome a label advances `32.4px` per line and on Safari `32px`. The gap accumulates with each wrapped line, so a multi-line label sits at a different vertical position depending on the engine. In multiplayer this shows up as text and freehand strokes that line up for one user and are offset for another. Both behaviours are spec-legal (line-box pixel snapping is implementation-defined), so this isn't a browser bug we can wait out.

The fix computes `Math.round(fontSize * lineHeight)` in JS — identically on every client — and applies that pixel value everywhere line-height is used: text measurement (`TextManager`), on-canvas render (`PlainTextLabel`, `RichTextLabel`), and SVG export (`RichTextSVG`). Because the integer is derived from the resolved font size, it still scales with font size; it only loses the sub-pixel fraction that the two engines disagree on. Zoom and per-shape `scale` are applied as CSS transforms on top, so they remain smooth and don't reintroduce divergence.

### Change type

- [x] `bugfix`

### Test plan

1. Open a multi-line text shape (or geo/arrow label) in Chrome and in Safari at the same zoom; confirm wrapped lines sit at the same vertical positions in both.
2. In a multiplayer room, join from Chrome and Safari and confirm text labels and freehand strokes stay aligned across clients.
3. Change font size (S/M/L/XL) and confirm line spacing still scales proportionally.

- [ ] Unit tests
- [ ] End to end tests

Note: this changes the rendered line-height of all text by up to a sub-pixel (e.g. `32.4px` → `32px`), so visual/geometry snapshots that contain text may need regenerating.

### Release notes

- Fix text labels drifting vertically between browsers (e.g. Chrome vs Safari) in multiplayer by snapping line-height to whole pixels.

### API changes

- Added `getLineHeightPx(fontSize, lineHeight)` to `@tldraw/editor` (`@internal`).

### Code changes

| Core code       | +29 / -8   |


frolic:
### Approaches considered

The drift comes from the theme's unitless `line-height` (`1.35`), which leaves the per-line advance fractional (e.g. at size "m", `24 * 1.35 = 32.4px`). WebKit snaps each wrapped line box to a whole pixel while Blink keeps the fraction, so the engines disagree and the gap accumulates per line. Both behaviours are spec-legal (line-box pixel snapping is implementation-defined), so it can't be waited out.

| Approach | Result |
| --- | --- |
| Keep the unitless multiplier (`1.35`) | The fraction is exactly what the engines disagree on — this is the source of the drift. |
| Set an explicit *fractional* px value (`32.4px`) | Doesn't help — WebKit still snaps the resulting line box to a whole pixel, so the engines still diverge. Verified in the repro. |
| **Round to a whole px in JS, identically on every client** (this PR) | Removes the fraction before it reaches the DOM, so there's nothing left for the engines to disagree about. The only option that aligned both engines in testing. |

### Trade-offs of the chosen approach

- Line spacing loses up to ~0.5px of sub-pixel precision — a cosmetic tightening.
- The effective line-height ratio wobbles slightly between font sizes, since each size rounds independently.
- Existing documents reflow once (line positions shift by a sub-pixel).
- Glyphs that overflow the line box (tall scripts, emoji) can still differ between engines — this fixes the box advance, not glyph overflow.
- The value is still derived from the resolved font size, so it scales with font size as before; only the fraction is lost. Zoom and per-shape `scale` are CSS transforms layered on top, so they stay smooth and don't reintroduce divergence.

frolic:
FYI, this caused a small diff in most screenshots because of inconsistent line height rounding across browsers, which is now made consistent by this change.

frolic:
### Follow-up (post-merge): scaled arrow label line-height

Capturing the Bugbot finding here as a follow-up rather than a separate issue, since it only becomes relevant once this lands.

Arrow labels measure line-height at the **scaled** font size but render on-canvas at the **unscaled** size with a CSS `scale()` transform. Because `getLineHeightPx` rounds to whole px and rounding doesn't commute with scaling, the measured box and rendered text can disagree by ≤1px when `shape.props.scale !== 1`:

- Measured / SVG-export: `round(labelFontSize × scale × lineHeight)`
- On-canvas: `round(labelFontSize × lineHeight) × scale`
- e.g. `18`, `1.35`, `scale 2` → `round(48.6)=49` vs `round(24.3)×2=48` → 1px gap.

**Where:** `arrowLabel.ts:77`/`:123` (measure at scaled), `ArrowShapeUtil.tsx:862`+`:875` (render unscaled + CSS scale), `ArrowShapeUtil.tsx:1018` (SVG export at scaled).

**Why it's not a blocker:** only at `scale !== 1` (arrows don't reach this through normal editing), ≤1px, mostly absorbed by `verticalAlign: middle`. It's not introduced by this PR either — the same asymmetry existed via engine-dependent line-box snapping; this change just makes it deterministic and identical across browsers (the goal), so multiplayer stays consistent.

**Note for the follow-up:** the arrow is the only shape with this asymmetry. Text (`TextShapeUtil.tsx:170`/`:181`/`:355`/`:124`) and geo (`GeoShapeUtil.tsx:406`/`:419`/`:829`/`:310-313`) already measure at the unscaled font size and scale the resulting box, which commutes with the rounding. The fix is to make arrow measurement match that pattern (measure unscaled, scale the box; adjust the squish thresholds that compare against page-space `bodyBounds`/`margin` by `/ scale`), and align the SVG export path so it doesn't swap the on-canvas mismatch for an export-vs-box one. Touches arrow label positioning, so it needs a visual check + arrow/export e2e snapshot regen.

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -161,8 +162,8 @@ export const RichTextLabel = React.memo(function RichTextLabel({
 				className={`${cssPrefix}-label__inner tl-text-content__wrapper`}
 				style={{
 					fontSize,
-					lineHeight: lineHeight.toString(),
-					minHeight: Math.floor(fontSize * lineHeight) + 'px',
+					lineHeight: `${getLineHeightPx(fontSize, lineHeight)}px`,
+					minHeight: `${getLineHeightPx(fontSize, lineHeight)}px`,
```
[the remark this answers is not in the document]
frolic: (same as above)

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -2,6 +2,25 @@ import { BoxModel, TLDefaultHorizontalAlignStyle } from '@tldraw/tlschema'
...
+ * A unitless `line-height` leaves the per-line advance fractional (e.g. 24 * 1.35 =
+ * 32.4px). WebKit snaps each line box to a whole pixel while Blink keeps the fraction,
+ * so wrapped lines drift apart across engines and the offset accumulates. Resolving the
+ * value to an integer here — in JS, identically on every client — removes the fraction
+ * before layout, so both engines agree. See https://github.com/tldraw/tldraw/issues/8970.
+ *
+ * This must be used everywhere line-height is applied (measurement, on-canvas render, and
+ * export) so geometry and rendering stay in agreement.
+ *
+ * @internal
```
mimecuvalo: actually, let's make this public (for convenience downstream)

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -2,6 +2,25 @@ import { BoxModel, TLDefaultHorizontalAlignStyle } from '@tldraw/tlschema'
...
+ */
+export function getLineHeightPx(fontSize: number, lineHeight: number): number {
+	return Math.round(fontSize * lineHeight)
```
mimecuvalo, line 17: `round` might be the right choice here, but to check would `floor` / `ceil` would work potentially? we tend to reach for `Math.ceil` for some things with width/height measurement (see PlainTextLabel/RichTextLabel)

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
```diff
@@ -2,6 +2,25 @@ import { BoxModel, TLDefaultHorizontalAlignStyle } from '@tldraw/tlschema'
...
+}
```
mimecuvalo, line 18: actually, this reconfiguration of the line-height might technically be a breaking change (sorta) for WebKit - boards that have calculated one thing might end up slightly different going forward. this is prbly ok but just something to keep aware of

mimecuvalo (commented):
i think there are some more places that reference `lineHeight` that might need updating in the `packages/tldraw` — i grepped around and see some other potential candidates — mind taking a look?

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
frolic: from Claude:
- floor = exact match to current Safari/WebKit (zero change for existing Safari boards), but biases all spacing down vs the 1.35 design and vs Chrome (up to ~1px tighter per line)
- round = closest to the designed ratio and to Chromium's current rendering (the majority engine); shifts Safari boards by +1px/line at the ≥.5 sizes
- ceil = loosest; diverges most from both engines' current rendering → worst for board churn 

thoughts given this?

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
frolic: (see comment above)

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
mimecuvalo: i think that sounds right, yeah just sanity checking here. `round` is prbly the way to go

packages/editor/src/lib/editor/managers/TextManager/TextManager.ts
frolic: done! also renamed this for clarity

frolic:
> i think there are some more places that reference `lineHeight` that might need updating in the `packages/tldraw` — i grepped around and see some other potential candidates — mind taking a look?

done! found a couple more spots and fixed em

frolic:
CI failure is not actually a failure, but a bad exit code when no snapshot changes, fixed in https://github.com/tldraw/tldraw/pull/9070

mimecuvalo:
thx! just to check (again, i'm just grepping here for "lineHeight:") — do these other spots need the same treatment? if not, why so?
```
src/lib/defaultExternalContentHandlers.ts
519:		lineHeight: theme.lineHeight,
533:			lineHeight: theme.lineHeight,

src/lib/shapes/geo/GeoShapeUtil.tsx
211:				labelLineHeight: theme.lineHeight,
831:			lineHeight: dv.labelLineHeight,
890:		lineHeight: theme.lineHeight,

src/lib/shapes/text/TextShapeUtil.tsx
81:				lineHeight: theme.lineHeight,
356:		lineHeight: dv.lineHeight,

src/lib/shapes/arrow/ArrowShapeUtil.tsx
182:				labelLineHeight: theme.lineHeight,

src/lib/shapes/arrow/arrowLabel.ts
82:			lineHeight: dv.labelLineHeight,
113:				lineHeight: dv.labelLineHeight,

src/lib/shapes/note/NoteShapeUtil.tsx
130:				labelLineHeight: theme.lineHeight,
624:				lineHeight: dv.labelLineHeight,
641:					lineHeight: dv.labelLineHeight,
```

frolic:
Good question — none of them need it, and here's the reasoning.

The browser-rounding only happens at the point where line-height becomes **CSS the browser actually lays out**. For canvas text there are exactly three such points, and all three already emit `resolveLineHeightPx(...)px`:

- **measurement** — `TextManager.getMeasureStyles`
- **on-canvas render** — `RichTextLabel` / `PlainTextLabel`
- **SVG export** — `RichTextSVG`

So the unitless multiplier never reaches a browser layout un-resolved.

Everything in your grep is *upstream* of those three points — it's the multiplier being transported to them, in one of two forms:

- **`measureHtml` opts** (`lineHeight: dv.labelLineHeight` / `theme.lineHeight`) — `measureHtml` feeds these into `getMeasureStyles`, which resolves them. (`TLMeasureTextOpts.lineHeight` is even documented as "must be a number, e.g. 1.35, not a pixel value.")
- **display-value assignments** (`labelLineHeight: theme.lineHeight`) — just stores the multiplier on the shape's display values; it's resolved by whichever of the three boundaries consumes it.

We deliberately keep these as the unitless multiplier rather than pre-resolving, because resolving needs a concrete font size — and that isn't fixed upstream (note `fontSizeAdjustment`, arrow/geo `scale`). `theme.lineHeight` is the font-size-independent `1.35`, so it has to stay a multiplier until an actual font size is in hand, i.e. at the boundary.

The only spots that needed routing through the helper were ones doing the `fontSize * lineHeight` multiply *themselves*, outside those boundaries — the geo/note min-height calcs, now fixed. A grep for `* lineHeight` confirms those were the only two.

mimecuvalo (approved):
thank you for the due dilligence 👍 LGTM

PR #8984, opened 2026-06-01
chore: ban bare lodash imports

frolic:
In order to keep `import { x } from 'lodash'` from silently pulling in the full ~72 KB lodash library, this PR adds an oxlint `no-restricted-imports` rule for the bare `lodash` package and migrates the existing usages to `lodash/X` deep imports.

Lodash's main package is CommonJS — its `lodash.js` is a UMD wrapper that `module.exports = _`, with no `module` or `exports` field. Neither Rollup (Vite) nor esbuild can tree-shake that. Bundling a one-line entry confirmed this on the version actually installed (`lodash@4.18.1`):

| Import shape                             | Vite (Rollup) | esbuild |
| ---------------------------------------- | ------------: | ------: |
| `import { throttle } from 'lodash'`      |       72.1 KB | 73.8 KB |
| `import _ from 'lodash'`                 |       72.6 KB | 73.8 KB |
| `import throttle from 'lodash/throttle'` |        4.2 KB |  3.7 KB |
| `import isEqual from 'lodash/isEqual'`   |       15.5 KB | 16.0 KB |

Deep imports (`lodash/X`) work because each method lives in its own file in the main package and `require`s only the helpers it actually needs. They tree-shake naturally under both bundlers, and we only need one `lodash` dep + one `@types/lodash` per workspace instead of one per method.

Affected:

- `apps/vscode/extension/src/WebViewMessageHandler.ts` was importing `{ isEqual }` from `lodash`, shipping the full ~72 KB library in the extension bundle. Now uses `lodash/isEqual` (~16 KB).
- `apps/examples/*` had four files using bare `lodash`. Migrated `throttle`, `isEqual`, `reduce`, and `capitalize` to `lodash/X` deep imports. `lodash.round(x)` → `Math.round(x)` in `GlobShapeUtil.tsx` because every call site is single-argument and behaves identically without the precision parameter.
- `apps/dotcom/client/scripts/build.ts` was using `_.chunk` — replaced with a 3-line inline `for` loop. The build script is Node-only (no bundle impact) and chunking has no edge cases here, so it didn't seem worth adding `lodash` + `@types/lodash` to `apps/dotcom/client` for one call site.
- `packages/state`'s fuzz test was using `lodash.times` — now `lodash/times`.
- Dropped truly-unused `lodash` / `lodash.throttle` deps from `packages/state-react` and `templates/sync-cloudflare`.

The new lint rule restricts only the bare `'lodash'` specifier, so `lodash/X` deep imports, `lodash-es`, and the existing `lodash.<method>` per-method packages (`packages/utils`, `apps/dotcom/sync-worker`, `apps/dotcom/client`'s `lodash.pick`) all still pass.

### Change type

- [x] `improvement`

### Test plan

1. `npx oxlint` from the repo root → 0 errors across the codebase.
2. With a planted `import { capitalize } from 'lodash'`, the rule fires with a message pointing to `lodash/X` deep imports, `lodash-es`, and per-method packages.
3. `yarn typecheck` → clean.
4. `packages/state` fuzz tests → 21/21 pass.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Shrink the VS Code extension bundle by ~56 KB by switching `isEqual` off of the full lodash package.

### Code changes

| Core code      | +0 / -2    |
| Tests          | +1 / -1    |
| Documentation  | +14 / -14  |
| Apps           | +6 / -3    |
| Templates      | +0 / -2    |
| Config/tooling | +17 / -4   |


mimecuvalo (approved):
👍

PR #8986, opened 2026-06-01
fix(text): expand text shape bounds to fit cursive and RTL glyph ink

### LINKED ISSUE #8802: Exported PNG/SVG cuts off RTL text

Image
Image

### LINKED ISSUE #8803: Arabic text renders outside the text shape bounding box
### What happened?

When entering Arabic (RTL) text into a text shape, the rendered glyphs extend well outside the shape's bounding box on both sides. The selection rectangle is much narrower than the visible text, so the box no longer represents where the text actually is on the canvas.

This appears related to but distinct from #8802 (export clipping) and #7497 (RTL list/punctuation handling) — here the on-canvas geometry itself does not match the rendered text.

Image

1. Open tldraw (tldraw.com or the examples app).
2. Create a text shape.
3. Paste a few lines of Arabic text, for example:
   ```
   ألم نشأت بالفطرة
   لتكتشف حقيقة بشرية
   ، فلا أحد يرفض
   ، ولكن بفضل
   ```
4. Observe that the selection bounding box is significantly narrower than the rendered glyphs — text extends past both the left and right edges of the box.

_No response_

### Where did this happen?

The tldraw.com website

### Code of Conduct

frolic:
In order to stop cursive, RTL, and italic glyphs from being painted outside their text shape's bounding box (#8803) and clipped on SVG export (#8802), this PR pads the text shape's geometry by the actual glyph ink overflow.

Text is measured for pen-advance width via the DOM, which ignores glyph side bearings, so the trailing stroke of an Arabic/Hebrew word (or the slant of an italic) ends up outside the box. The fix adds `TextManager.measureTextInkBounds(element)`, which reads the **real ink bounds** back from a laid-out element: wrapping, bidi, alignment, and per-run bold/italic all come from the browser's layout, and only the ink delta is measured with the Canvas `TextMetrics` API. `getGeometry` uses this to grow the shape's bounds per side. The text content and `props.w` are left unchanged, so layout, alignment, and resizing are unaffected — only the reported bounds (indicator, selection, hit-test, snapping, and export viewport) grow to contain the ink. The result is cached and recomputes when a web font finishes loading.

The advance and ink are measured in a **single layout pass** (`measureHtmlBounds` renders the measurement element once and returns both), and `getMinDimensions`/`getGeometry` share one cache — so the ink work adds no extra DOM render over the original advance measurement. Reading everything back from the real layout also means the shape no longer needs hand-rolled RTL detection, alignment math, mark scanning, a soft-wrap guard, or a vertical baseline model.

Handled cases:

- RTL/cursive and italic horizontal side bearings (the #8803 case).
- Bold/italic applied via rich-text marks (each run measured in its own style).
- Vertical overflow (tall diacritics, italic descenders) via top/bottom padding.
- Wrapped lines in fixed-width shapes (measured from the real line layout, so the box neither clips nor balloons to the unwrapped width).
- Plain Latin text, which fits its advance box — overflow is zero and the bounds are unchanged.

### Editor API

`measureTextInkBounds(element)` is reusable beyond this shape — e.g. for selection bounds of custom text content, or trimming/sizing image exports.

### Change type

- [x] `bugfix`

### Test plan

1. Create a text shape with the default (Draw) font and type Arabic or Hebrew, e.g. `שלום עולם`.
2. Select it: the selection box and hover indicator enclose the glyphs instead of clipping the trailing cursive stroke.
3. Export the shape to SVG: the glyphs are no longer clipped at the box edge.
4. Confirm Latin text is unchanged (no extra padding, box origin unmoved) and fixed-width wrapped text keeps its width.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fix text shapes clipping cursive, RTL (e.g. Arabic, Hebrew), and italic glyphs at the edge of their bounding box, both on the canvas and in SVG exports.

### API changes

- Added `TextManager.measureTextInkBounds(element)` for measuring the true ink bounds of text rendered in a DOM element.

### Code changes

| Core code       | +206 / -38 |
| Tests           | +132 / -0  |
| Automated files | +6 / -0    |

Closes #8803
Closes #8802

### DIFF
--- a/apps/examples/e2e/tests/test-text-ink-bounds.spec.ts
+++ b/apps/examples/e2e/tests/test-text-ink-bounds.spec.ts
@@ -10,12 +10,17 @@ test.describe('text shape ink bounds', () => {
 	// Creates an autosized text shape, waits for its webfont to load — so the canvas ink
 	// measurement uses real glyph metrics and the geometry recomputes — then returns the
 	// advance-box width alongside the (possibly ink-padded) geometry bounds.
-	async function measure(page: Page, text: string, font: 'draw' | 'sans') {
+	async function measure(
+		page: Page,
+		text: string,
+		font: 'draw' | 'sans' | 'serif',
+		italic = false
+	) {
 		// Plain string (not a branded TLShapeId): the template-literal `shape:${string}` type
 		// trips TS2589 in page.evaluate's Serializable check on the argument.
 		const id = 'shape:textInkBounds'
 		return await page.evaluate(
-			async ({ id, text, font }) => {
+			async ({ id, text, font, italic }) => {
 				// Drop into `any` for the editor calls: fully typing a text-shape partial and
 				// reaching TextShapeUtil.getMinDimensions inside page.evaluate trips TS2589 (the
 				// shape-util union is too deep). The other e2e tests lean on `as any` similarly.
@@ -29,7 +34,14 @@ test.describe('text shape ink bounds', () => {
 					props: {
 						richText: {
 							type: 'doc',
-							content: [{ type: 'paragraph', content: [{ type: 'text', text }] }],
+							content: [
+								{
+									type: 'paragraph',
+									content: [
+										{ type: 'text', text, marks: italic ? [{ type: 'italic' }] : undefined },
+									],
+								},
+							],
 						},
 						font,
 						size: 'xl',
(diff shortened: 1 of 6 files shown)


angrycaptain19:
Screenshot 2026-06-03 at 09 38 13

I just tested in the example preview link and am seeing a tiny bit of overflow both in selection and in export

mimecuvalo (commented):
haven't reviewed this whole PR in depth but couple broad comments to keep in mind:
- text measurement is one of our biggest bottlenecks for board rendering. we wanna keep it super tight for perf reasons
- in that vein, it'd be great to have a hotpath/short-circuit to speed things up. maybe this involves a quick regex check to see if the text contains RTL (or any other relevant) characters? and/or if it has things like italic that would cause things like this? i'm not sure what the correct heuristic would be. (it also might be a footgun to have a narrow heuristic like "only when italic is present", not sure)
- final note: i see this is looking at TextShapeUtil but if we go in this direction, prbly worth  making this work with all shapes that include text

might take a while to review this, this kind of change is pretty sensitive.

frolic:
taking a fresh stab in https://github.com/tldraw/tldraw/pull/9387

Issue #6827, opened 2025-09-23
Consider dependency cleanup


outslept:
Would you be open to changing a couple of dependencies?

1. `fs-extra`

This could be replaced with node.fs - https://nodejs.org/api/fs.html

2. `lodash.*`

Use of per method packages is discouraged (https://lodash.com/per-method-packages) and we should be able to use native JavaScript to replace each call.

3. `lodash`

Like 2nd, we can replace it with modern JS

4. `chokidar-cli`

A bit questionable, from what I understand we can use `chokidar` v4 + [`glob-parent`](https://www.npmjs.com/package/glob-parent) + [`picomatch`](https://www.npmjs.com/package/picomatch). The problem I have with this package - it pulls `lodash.*` per method packages we're trying migrate from so we'll end up with those packages in our lockfile. Also chokidar cli uses v3 chokidar under the hood which pulls 13 dependencies vs chokidar v4 which uses a single dependency.

I didn't check for unused dependencies. But from a quick look `is-plain-object` is unused now. More that happy to dive deeper if you find this useful

PR #8987, opened 2026-06-01
chore: add knip CI check for undeclared and unused dependencies

frolic:
In order to catch a class of latent dependency bugs that Yarn workspace hoisting hides — a workspace's source `import`s a package its own `package.json` doesn't declare, but the build happens to pass because some _other_ workspace declares it — this PR adds [knip](https://knip.dev) as a CI check.

A concrete instance is the bare `lodash` import in `apps/dotcom/client/scripts/build.ts` while only `lodash.pick` is declared — that one is fixed in [#8984](https://github.com/tldraw/tldraw/pull/8984), so this PR doesn't duplicate the fix; it just temporarily suppresses the finding in `knip.config.ts` with a comment pointing at #8984.

`oxlint` doesn't currently support `import/no-extraneous-dependencies` (verified with oxlint 1.66), so this can't be solved by a lint rule today. Knip complements the existing `yarn check-packages` (which keeps `tsconfig.json` references in sync) and `yarn check-circular-deps`.

Relates to [#6827](https://github.com/tldraw/tldraw/issues/6827).

### What knip checks (and what it doesn't)

The `include` list is scoped to `dependencies`, `devDependencies`, and `unlisted`:

- **Unused** dependencies — declared in `package.json` but not imported anywhere.
- **Unlisted** dependencies — imported from source but not declared in the workspace's `package.json`.

Knip's noisier checks (unused files, unused exports, unresolved imports) are intentionally **not** enabled here — separate decision.

### Real findings fixed in this PR

**Undeclared imports added to the right workspace:**

- `apps/dotcom/client`: added `@clerk/types`, `@peculiar/webcrypto`, `zod` (used by source / setup files / vite shim).
- `internal/scripts`: added `octokit`, `pg`, `@types/pg`, `nanoid`, `strip-ansi` (used by multiple scripts under `internal/scripts/`).
- Root `package.json`: added `glob` (used by `vitest.config.ts`).
- `packages/tldraw`: added `fake-indexeddb` and `jest-diff` (used by `setupVitest.js` and `src/test/testutils/pretty.ts`).
- `packages/sync-core`: added `@peculiar/webcrypto`, `fake-indexeddb`, `resize-observer-polyfill` (used by `setupVitest.js`).

**Unused declarations removed:**

- `apps/dotcom/client`: removed `react-markdown`.

### Pending #8984

[#8984](https://github.com/tldraw/tldraw/pull/8984) cleans up bare-`lodash` imports and unused lodash declarations. This PR suppresses those findings (`lodash` in dotcom client scripts, `lodash` + `@types/lodash` in state-react, `lodash.throttle` + `@types/lodash.throttle` in sync-cloudflare) via `ignoreDependencies` entries that each carry a comment pointing at #8984. Once #8984 merges and this branch rebases, those entries become orphaned and can be deleted (knip will emit hints flagging them as no-longer-needed).

### What's suppressed as a longer-term baseline

Knip flagged a long tail of suspected-unused deps across workspaces that weren't worth chasing one-by-one in this PR. They're listed per-workspace in `knip.config.ts` under `ignoreDependencies`, each with a short comment, so each one is visible in the config rather than hidden. The intent is to trim that list over time as workspaces are audited.

The repo-wide `ignoreDependencies` list covers infra binaries (`lazyrepo`, `wrangler`, `esbuild`, ...) invoked from package.json scripts that knip's binary detection can't reach through `yarn run -T tsx <script>` indirection, plus ambient `@types/*` packages and runtime helpers.

### Follow-up candidates (deferred)

Real unused-looking deps still suppressed in `knip.config.ts` and worth a closer audit — almost certainly removable but each needs eyes:

- `apps/bemo-worker`: `@tldraw/dotcom-shared`, `@tldraw/store`
- `apps/docs`: `@tldraw/assets`, `google-auth-library`, `js-cookie`, `query-string`, `rehype-slug`
- `apps/dotcom/image-resize-worker`: `itty-router`
- `apps/dotcom/sync-worker`: `@supabase/auth-helpers-remix`, `@tldraw/validate`, `jose`
- `apps/examples`: `@tldraw/driver`, `classnames`
- `apps/vscode/extension`: `node-fetch`, `@tldraw/editor`, `assert`, `process`
- `apps/dotcom/client`: `ws` (devDep)
- `internal/scripts`: `gray-matter`, `ignore`
- `packages/assets`: `@tldraw/utils`
- `packages/create-tldraw`: `@tldraw/utils`, `ansi-regex`
- `packages/dotcom-shared`: `@tldraw/store`, `@tldraw/tlschema`, `@tldraw/validate`
- `packages/editor`: `is-plain-object` (called out in [#6827](https://github.com/tldraw/tldraw/issues/6827)), `benchmark`
- `packages/mermaid`: `@tldraw/tlschema`
- `packages/store`: `raf`
- `packages/sync`: `@tldraw/state-react`, `nanoevents`, `ws`, `uuid-by-string`, `uuid-readable`
- `packages/tldraw`: `@tiptap/extension-list`, `idb`
- `packages/tlschema`: `kleur`
- `templates/*`: various

Each of these could be removed in a small follow-up PR; the suppression keeps the check green meanwhile.

### Change type

- [x] `other`

### Test plan

1. `yarn knip` from the repo root exits 0 cleanly.
2. To verify the check actually fires: temporarily add e.g. `import _ from 'lodash'` to `apps/dotcom/client/src/main.tsx`, run `yarn knip` — it should report `lodash` as an unlisted dependency.
3. The CI workflow runs the same `yarn knip` step as part of the `Tests & checks` job.

### Code changes

| Apps            | +0 / -1      |
| Config/tooling  | +931 / -34   |

(yarn.lock changes account for most of the config/tooling delta.)


apps/dotcom/client/package.json
```diff
@@ -54,14 +54,16 @@
 		"react-ga4": "^2.1.0",
 		"react-helmet-async": "^1.3.0",
 		"react-intl": "^8.1.3",
-		"react-markdown": "^10.1.0",
 		"react-router-dom": "^6.28.2",
-		"tldraw": "workspace:*"
+		"tldraw": "workspace:*",
+		"zod": "^4.1.8"
```
mimecuvalo, line 59: i _think_ `zod` is a dev dep (in `scripts`)

internal/scripts/package.json
```diff
@@ -54,13 +54,18 @@
...
 		"@types/minimist": "^1.2.5",
+		"@types/pg": "^8.11.11",
 		"@types/proper-lockfile": "^4.1.4",
 		"@types/tmp": "^0.2.6",
 		"diff": "^9.0.0",
 		"ignore": "^5.3.2",
 		"jszip": "^3.10.1",
 		"minimist": "^1.2.8",
+		"nanoid": "^5.1.2",
+		"octokit": "^3.2.1",
```
mimecuvalo, line 65: i see we have `@octokit/rest` above — possible to just use that?

packages/sync-core/package.json
```diff
@@ -43,11 +43,14 @@
 		"pack-tarball": "yarn pack"
 	},
 	"devDependencies": {
+		"@peculiar/webcrypto": "^1.5.0",
```
mimecuvalo, line 46: i wonder if the use of webcrypto in sync-core setupVitest is copy-pasta - can you take a look at maybe just removing this dep?

packages/sync-core/package.json
```diff
@@ -43,11 +43,14 @@
...
 		"@types/react": "^19.2.7",
 		"@types/react-dom": "^19.2.3",
 		"@types/uuid-readable": "^0.0.3",
+		"fake-indexeddb": "^4.0.2",
```
mimecuvalo, line 50: maybe similar for this one and below

packages/tldraw/package.json
```diff
@@ -80,6 +80,8 @@
 		"@types/react": "^19.2.7",
 		"@types/react-dom": "^19.2.3",
 		"chokidar-cli": "^3.0.0",
+		"fake-indexeddb": "^4.0.2",
```
mimecuvalo, line 83: i wonder if fake-indexeddb is also not needed here

knip.config.ts
```diff
@@ -0,0 +1,274 @@
...
+// which respects the `include` list below.
+//
+// The `ignoreDependencies` lists below establish a baseline: knip flagged a
+// long tail of suspected-unused deps in existing workspaces that we don't
+// have the bandwidth to investigate one-by-one in this PR. They're listed
+// here per-workspace so each one is visible, with the intent of trimming the
+// list over time as workspaces are audited. New unused/undeclared deps will
+// still be caught.
+
+const config: KnipConfig = {
```
mimecuvalo, line 20: this file's long list makes me pause (maybe this is why it's a draft PR), i'd maybe worry about maintaining this as added cognitive load

mimecuvalo (commented):
cc @kostyafarber who's done some knip'ing in the past

frolic:
I don't think we wanna maintain this knip config so closing this out for now

kostyafarber:
i think knip doesn't work well with interface heavy codebases, it generates a lot of false positives that end up bloating the config, I think a lightweight version of this could look for at least unused files maybe or functions, or be used as a cli for periodic agent type sweeps as a lowpass filter for the agent to then go ahead and see whether to prune things or not.

i imagine a similar sweep could be done once a month in ci for our docs

PR #9000, opened 2026-06-02
feat(utils): add { signal } option to debounce

frolic:
In order to let callers cancel a debounced call with a standard web primitive, this PR adds an `AbortSignal`-based `{ signal }` option to the `debounce` util: `debounce(fn, wait, { signal })`. Aborting the signal rejects the pending promise with the signal's `reason` (an `AbortError` by default), matching how modern libraries (TanStack, es-toolkit) thread cancellation. The existing `.cancel()` method is kept as-is. Relates to #8683.

Along the way it refactors `debounce`'s internals to a single `cancel(reason)` closure and `promiseWithResolve` (dropping a hand-rolled deferred), migrates `Editor.scheduleEffect` to the new option, and removes a dead debounce call in `useZoomCss`.

### Change type

- [x] `api`

### Test plan

- [x] Unit tests

### API changes

- Added optional `opts?: { signal?: AbortSignal }` parameter to `debounce()`. Aborting the signal rejects the pending promise with `signal.reason`.
- `debounce(...).cancel()` is unchanged. The api-report now renders it as `cancel(): void` instead of `cancel: () => void` because the function gained an explicit return type — no behavioral change.

### Release notes

- Add an `AbortSignal` `{ signal }` option to the `debounce` utility for canceling a pending call.

### Code changes

| Core code       | +78 / -47  |
| Tests           | +71 / -0   |
| Automated files | +4 / -2    |

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -10642,7 +10642,10 @@ export class Editor extends EventEmitter<TLEventMap> {
 				window.history.replaceState({}, this.getContainerDocument().title, url.toString())
 			})
 
-		const scheduleEffect = debounce((execute: () => void) => execute(), opts?.debounceMs ?? 500)
+		const controller = new AbortController()
+		const scheduleEffect = debounce((execute: () => void) => execute(), opts?.debounceMs ?? 500, {
+			signal: controller.signal,
+		})
 
 		const unlisten = react(
 			'update url on state change',
@@ -10652,7 +10655,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 
 		return () => {
 			unlisten()
-			scheduleEffect.cancel()
+			controller.abort()
 		}
 	}
 
--- a/packages/editor/src/lib/hooks/useZoomCss.ts
+++ b/packages/editor/src/lib/hooks/useZoomCss.ts
@@ -1,5 +1,4 @@
 import { EffectScheduler } from '@tldraw/state'
-import { debounce } from '@tldraw/utils'
 import * as React from 'react'
 import { useContainer } from './useContainer'
 import { useEditor } from './useEditor'
@@ -10,7 +9,6 @@ export function useZoomCss() {
 
 	React.useEffect(() => {
 		const setScale = (s: number) => container.style.setProperty('--tl-zoom', s.toString())
-		const setScaleDebounced = debounce(setScale, 100)
 
 		const scheduler = new EffectScheduler('useZoomCss', () =>
 			setScale(editor.getEfficientZoomLevel())
@@ -21,7 +19,6 @@ export function useZoomCss() {
 
 		return () => {
 			scheduler.detach()
-			setScaleDebounced.cancel()
 		}
 	}, [editor, container])
 }
--- a/packages/utils/api-report.api.md
+++ b/packages/utils/api-report.api.md
@@ -44,9 +44,11 @@ export function clearSessionStorage(): void;
 export function compact<T>(arr: T[]): NonNullable<T>[];
 
 // @public
-export function debounce<T extends unknown[], U>(callback: (...args: T) => Awaitable<U>, wait: number): {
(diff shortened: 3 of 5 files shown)


frolic:
### Potential follow-ups: thread `AbortSignal` through more of the stack

While working on this, a few other spots stood out as good candidates for the same web-standard `{ signal }` cancellation, so a top-level abort could cascade through the stack (the way TanStack/Remix thread signals everywhere). Noting them here rather than expanding this PR:

- **`retry()`** — `packages/utils/src/lib/retry.ts:60`. Already takes an `abortSignal`, but it's the canonical thread-through util and currently half-baked: named `abortSignal` (vs the `signal` convention this PR sets), only polls `aborted` *between* attempts (can't interrupt an in-flight `fn()` or the `await sleep(waitDuration)` wait), and throws `new Error('aborted')` instead of `signal.reason`. Highest leverage since everything that retries flows through it.

- **`TLAssetStore.upload(asset, file, abortSignal?)`** — `packages/tlschema/src/TLStore.ts:182` (+ `packages/editor/src/lib/editor/Editor.ts:5230`). Already threads a signal end-to-end, but it's a positional `abortSignal` (we generally prefer a named `{ signal }` option), the name is non-standard, and it's undocumented in the JSDoc. Public API consumers implement this, and uploads are long + very abort-worthy (e.g. deleting an image mid-upload).

- **`TLSyncClient`** — `packages/sync-core/src/lib/TLSyncClient.ts` (`didCancel?()` polled at lines 555, 563, 576, 591, 603). Textbook bespoke cancellation: a `didCancel: () => boolean` callback checked at the top of every handler, alongside a separate `disposables` array. A single `signal` could unify both, and sync connect/bootstrap is exactly the long-lived async stack you'd want to abort from the top.

- **`throttle`** — `packages/utils/src/lib/throttle.ts`. Add a `{ signal }` option purely for symmetry with `debounce`, so the sibling utils match. Lowest urgency — a consistency win.

If we pursue these, the first two (`retry` and `upload`) also double as the chance to converge everything on `{ signal }` + reject-with-`reason`.

mimecuvalo (commented):
heh this is cool, but at this point i'm starting to wonder whether we should be rolling our own `debounce` anymore 🙃 
@MitjaBezensek do you remember why we did this in the first place?

MitjaBezensek:
> heh this is cool, but at this point i'm starting to wonder whether we should be rolling our own `debounce` anymore 🙃 @MitjaBezensek do you remember why we did this in the first place?

Think lodash does not support promises? Maybe some other lib does though.

frolic:
+1 to leaning on an external lib! but, then again, supply chain attacks have been on the rise lately 🤷 

p-debounce is probably the closest (I've historically had good experience leaning on p-retry, p-queue, etc)

frolic:
apparently p-debounce is esm-only but we're not yet?

I am fine to close this since there's not a clear use case for using signal here yet, was just a fly by suggestion in another PR

mimecuvalo:
> +1 to leaning on an external lib! but, then again, supply chain attacks have been on the rise lately 🤷
> 
> p-debounce is probably the closest (I've historically had good experience leaning on p-retry, p-queue, etc)

ah yeah, similarly i've had good success with `p-lazy`. i'd be down to go that route @frolic !

frolic:
We can't lean on p-debounce (ESM-only) without changing our build pipeline to better support ESM-only modules in a CJS context. I noticed in https://github.com/tldraw/tldraw/pull/7905 we're still actively supporting CJS.

Gonna close this out for now since it's more nice-to-have and separately explore what a build adjustment might look like to give us the ability to pull in ESM-only modules but still a CJS-compatible build.

mimecuvalo:
> We can't lean on p-debounce (ESM-only) without changing our build pipeline to better support ESM-only modules in a CJS context. I noticed in #7905 we're still actively supporting CJS.
> 
> Gonna close this out for now since it's more nice-to-have and separately explore what a build adjustment might look like to give us the ability to pull in ESM-only modules but still a CJS-compatible build.

couple months out of date but worth keeping in mind: https://github.com/wooorm/npm-esm-vs-cjs

PR #9001, opened 2026-06-02
fix(editor): correct nested clip masks with Sutherland–Hodgman intersection

kaneel:
/!\ Just a quick spike with Claude as one of our user founds some artifacts when dealing with nested clipped shapes

Fixes a rendering artifact where nested clipped shapes (e.g. an image under multiple getClipPath ancestors) could show a stray triangular wedge on top of the content. The bug came from intersectPolygonPolygon rebuilding clip polygons by sorting intersection vertices by angle around their centroid, which breaks for concave or multi-level intersections and produces self-intersecting clip-path: polygon(...) values.

Changes: 
- Replaced centroid-angle reconstruction in intersectPolygonPolygon with Sutherland–Hodgman clipping (convex clip window)
- Added dual-clip fallback and zero-area filtering so degenerate touches return no mask
- Added unit tests for concave subject clipping and an integration test for circle-clip + frame nesting

I've added an example taken from their own custom shapes they sent us 

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

Waiting from user info to figure how we can reproduce this effectively 100% of the time.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixes a rendering artifact with deeply nested clipped shapes


steveruizok:
How did we even find this one btw?

MitjaBezensek:
> How did we even find this one btw?

Customer reported it.

apps/examples/src/examples/editor-api/nested-clip-bug-repro/applyClipPathOverride.ts
MitjaBezensek, line 1: Thanks for the example, was much easier to see the fix 👏

Let's remove it before merging 👍

packages/editor/src/lib/primitives/intersect.ts
```diff
@@ -239,65 +238,171 @@ export function linesIntersect(A: VecLike, B: VecLike, C: VecLike, D: VecLike) {
...
+	return deduped
+}
 
-	// Order all points in the result counter-clockwise.
-	return orderClockwise([...result.values()])
+/**
+ * Clip a polygon by a convex clip window (Sutherland-Hodgman).
+ *
+ * `polygonA` may be concave. `polygonB` must be a convex clip window with
+ * consistent winding (the same winding as tldraw shape geometry vertices).
```
MitjaBezensek: Is this always true? I guess it could break some existing users if they might be using non-convex clip or not the same winding? Might be fine 🤷‍♂️

packages/editor/src/lib/primitives/intersect.ts
```diff
@@ -239,65 +238,171 @@ export function linesIntersect(A: VecLike, B: VecLike, C: VecLike, D: VecLike) {
 	return ccw(A, C, D) !== ccw(B, C, D) && ccw(A, B, C) !== ccw(A, B, D)
 }
 
+const CLIP_EPSILON = 1e-10
```
MitjaBezensek: Is it ok to use the same value for the different checks? Like we use this for distance, area, and also in dimensionless scenario (for a ratio).

packages/editor/src/lib/primitives/intersect.ts
```diff
@@ -239,65 +238,171 @@ export function linesIntersect(A: VecLike, B: VecLike, C: VecLike, D: VecLike) {
...
+ * Clip a polygon by a convex clip window (Sutherland-Hodgman).
+ *
+ * `polygonA` may be concave. `polygonB` must be a convex clip window with
+ * consistent winding (the same winding as tldraw shape geometry vertices).
+ *
+ * @param polygonA - Subject polygon to clip.
+ * @param polygonB - Convex clip window polygon.
+ * @public
+ */
+export function intersectPolygonPolygon(
```
MitjaBezensek, line 435: Return type changed a bit for certain cases, might be worth pointing out in release notes.

kaneel: 👍

packages/editor/src/lib/primitives/intersect.test.ts
```diff
@@ -943,4 +964,65 @@ describe('intersectPolygonPolygon', () => {
 		const result = intersectPolygonPolygon(polyA, polyB)
 		expect(result).toBeNull()
 	})
+
+	it('should return a simple polygon when clipping a concave subject by a convex window', () => {
```
MitjaBezensek, line 968: Would this catch a regression? Think adding `expect(result!.length).toBe(8)` (or something like that) would help to make sure we do.

packages/tldraw/src/test/nested-frames-clip-legacy.test.ts
MitjaBezensek, line 1: I guess this guards a bug that we never shipped? Do we still need this file?

MitjaBezensek (approved):
Looks good, left two small comments.

steveruizok:
Reviewed this. Summary of where I land:

**The fix itself is correct.** Swapping the centroid-angle sort in `intersectPolygonPolygon` for Sutherland–Hodgman is the right algorithm. The old approach genuinely produces self-intersecting polygons for concave or multi-level clips, which is the source of the wedge artifact. Existing frame/group/clip tests all still pass, so there's no regression in first-party behavior.

**On the winding requirement the new code introduces:** I checked whether we ever feed reversed-wound points into this function, and we don't. `getShapeLocalTransform` is `translate().rotate()` only (no scale), so page transforms are always orientation-preserving, and flips bake into shape props with consistently-wound geometry. Frame clip paths are just `geometry.vertices`. The only way to get reversed winding in is a custom `getClipPath` that hand-returns points against our convention — and the shape that triggered this report (the circle clip) winds correctly. So normalizing winding would be dead code; the doc comment covering the requirement is enough.

**Why I'm not merging as-is:** this is a spike against a core geometry primitive, and we still don't have a reproduction we can trigger reliably (per the test plan). I don't want to change a shared primitive on an approximated repro. We're not going to add code here without a strong reproduction. Closing until we have a solid, deterministic repro of the user's artifact — at which point this branch is a good starting point to pick back up.

Minor notes if we revisit:

- `nested-frames-clip-legacy.test.ts` compares against a segment-based Sutherland–Hodgman "legacy" implementation that never actually shipped (the real old code was the centroid sort), so its framing is misleading.
- `custom-clipping.test.ts` is mostly boilerplate around the test shape; only the nested-mask test exercises the fix.

angrycaptain19:
Looks like #10622 is this popping up again/similarly

PR #9002, opened 2026-06-02
fix: type collaborator user ids as TLUserId

### LINKED ISSUE #8509: My own cursor shows up as a different person with a delay
Image

This is my own cursor, which shows up on the screen alongside my actual cursor with a delay. 

Which, is pretty annoying

Also this up here shows multiple users, which is just me. 

Image

Browser: 
Helium 0.7.10.1 (Chromium 143.0.7499.192)
tldraw.com

frolic:
In order to stop the local user's own presence and follow-state from being mis-compared against the unprefixed id returned by `editor.user.getId()`, this PR types the collaborator user-id fields as `TLUserId` (the prefixed `user:…` record id) instead of plain `string`.

Closes #8509. Discovered while working on #8988.

`TLInstancePresence.userId` / `followingUserId` and `TLInstance.followingUserId` / `highlightedUserIds` were all typed `string`, while every value flowing into them at runtime is a prefixed `TLUserId`. Because `TLUserId` is just a branded `string`, comparisons like `presence.userId === editor.user.getId()` compiled fine but were always false — `editor.user.getId()` returns the raw, unprefixed preferences id. Tightening the field types turns these latent mismatches into compile errors and forces a `createUserId(...)` at each boundary.

This fixes three bugs that were all silent for the same reason:

- **Phantom self-cursor**: `CollaboratorsManager` never filtered out the local user's own other sessions, so a user open in two tabs rendered a cursor for themselves.
- **Idle following peers never hidden**: the "hide idle collaborators who are following you" rule in `getVisibleCollaborators` never fired.
- **"Following you" indicator broken**: the people-menu item compared a prefixed id against the raw id, so the leading/following state never showed.

Runtime behaviour is otherwise unchanged; this is a breaking *type* change for SDK consumers.

### Change type

- [x] `api`

### Test plan

1. Open the same document in two browser tabs as the same user; confirm neither tab shows a cursor for itself.
2. With two different users, have peer B follow user A, then let B go idle — A should stop seeing B's cursor.
3. In the people menu, when a peer is following you, confirm the "leading" indicator shows.

- [x] Unit tests
- [ ] End to end tests

### API changes

- Breaking! `TLInstancePresence.userId` / `followingUserId` are now `TLUserId` / `TLUserId | null` (were `string` / `string | null`)
- Breaking! `TLInstance.followingUserId` is now `TLUserId | null` and `highlightedUserIds` is now `TLUserId[]`
- Breaking! `Editor.startFollowingUser(userId)` and `Editor.zoomToUser(userId)` now take `TLUserId`
- Breaking! `usePresence(userId)` now takes `TLUserId`; `usePeerIds()` now returns `TLUserId[]`
- Breaking! People-menu prop types (`TLUiPeopleMenuItemProps`, `TLUiPeopleMenuAvatarProps`, `TLUiPeopleMenuFacePileProps`, `DefaultPeopleMenuContentProps`) now use `TLUserId`

### Code changes

| Core code       | +34 / -26  |
| Tests           | +87 / -64  |
| Automated files | +15 / -13  |
| Documentation   | +2 / -2    |

### Release notes

- Fix collaborators sometimes seeing a phantom cursor for their own other browser sessions.
- Fix idle peers who are following you not being hidden, and the "following you" indicator in the people menu not appearing.
- Type the presence/instance user-id fields and the related editor, hook, and component APIs as `TLUserId` instead of `string`.


frolic:
behavior before (same user, two browser windows):

https://github.com/user-attachments/assets/15e16d39-c6ab-4287-aae0-8a2b1cf9bfe9

behavior after (left two is same user, two browser windows, and right is an new user in incognito)

https://github.com/user-attachments/assets/0941812f-0eac-41a8-852a-799f6e21ffa5

packages/tldraw/src/lib/ui/components/SharePanel/DefaultPeopleMenuItem.tsx
```diff
@@ -32,7 +32,7 @@ export const DefaultPeopleMenuItem = track(function DefaultPeopleMenuItem({
 		}
 	}, [editor, userId, trackEvent])
 
-	const theyAreFollowingYou = presence?.followingUserId === editor.user.getId()
+	const theyAreFollowingYou = presence?.followingUserId === createUserId(editor.user.getId())
```
mimecuvalo: i wonder if `TLUserPreferences` → `id` should be `TLUserId` 🤔 feels like this might be a footgun in the future if another person forgets to use `createUserId` in places. would have to migrate the id i think.

cc @MitjaBezensek for thoughts here as well, i very well might be overthinking it here

mimecuvalo (commented):
nice sleuthing, see my main comment for further discussion

packages/tldraw/src/lib/ui/components/SharePanel/DefaultPeopleMenuItem.tsx
frolic: the inconsistent user IDs in various locations is a bit worrying to me, which I think is kinda what you're getting at here

frolic: from Claude
> TLUserPreferences.id is consumer-facing. Apps (including tldraw.com) construct it from their own auth system — a UUID, email, etc. Typing it TLUserId either forces every consumer to pass user:-prefixed ids (breaking) or requires us to normalize on input. getUserPreferences() / getId() return types change too. 

is this accurate and a real concern?

steveruizok: Yeah the user's id might be coming in from app land, in which case it shouldn't be branded

mimecuvalo: a suggestion of what could be done here:
```
  @computed getRecordId(): TLUserId {
        return createUserId(this.getId())
  }
```

and referencing that elsewhere. i'd worry about having to remember to wrap things every time

frolic: but then I guess the question becomes when do I use `getRecordId` vs `getId`? the naming itself doesn't make this clear/intuitive to me

frolic: talking to claude, it sounds like `getId` throughout already uses prefixed IDs but this is the one place that we don't do that? so I would think a better route would be


`getId()` -> `user:someone` (`TLUserId`)
`getExternalId()` -> `someone` (`string`)

but that would prob be a subtly breaking change that we don't want

mimecuvalo: (sry for delay) oof, yeah 😬 let's not do the breaking change.
one suggestion here would be to mark `getId()` as `@deprecated` and rename it to `getExternalId` (i.e. add an additional method) — then at least we have two methods that are more explicit in what they're doing.

frolic: updated!

`getId()` -> deprecated, points to `getExternalId()`
`getExternalId()` -> arbitrary, raw, app-provided ID
`getRecordId()` -> our internal ID, prefixed with `user:` and branded as `TLUserId`

all call sites updated

mimecuvalo (approved):
ty! LGTM

PR #9003, opened 2026-06-02
refactor(editor): decouple FocusManager from select tool state

### LINKED ISSUE #8888: FocusManager references the tldraw-package tool state 'select.editing_shape'
## Problem

`FocusManager` lives in `packages/editor` but reads `this.editor.isIn('select.editing_shape')`. The state path `'select.editing_shape'` is defined in `packages/tldraw` (the select tool's child state), not in the editor package.

This is an upward dependency from the editor package on the tldraw package's tool tree, held together only by a string. If the select tool is swapped or its child state renamed, focus-ring suppression silently breaks.

## Where

- `packages/editor/src/lib/editor/managers/FocusManager/FocusManager.ts:66` — `if (this.editor.isIn('select.editing_shape') && ...)`
- `packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts` — definition of the referenced state path

## Suggested fix

Inject `isEditingShape: () => boolean` into `FocusManager`. The default in `Editor` can supply the current check, and the tldraw package can override or compose its own predicate.

Separately: `editor.blur({ blurContainer: false })` bypasses `FocusManager` and calls `editor.complete()` itself — the blur contract should be centralized on the manager.

## Context

Surfaced during an internal modularity audit of the editor manager classes.

frolic:
In order to remove an upward dependency from the editor package on the tldraw package's tool tree, this PR changes `FocusManager` to stop checking the select tool's `'select.editing_shape'` state path. As reported in #8888, the editor package referenced a tldraw-package tool state held together only by a string, which would break silently if the select tool were renamed, restructured, or swapped out.

Instead, `FocusManager` now uses the editor's own first-class, tool-agnostic editing state via `editor.getEditingShapeId()`. This is the same state that `select.editing_shape` sets on enter and clears on exit, so the focus-ring behavior is equivalent — but it no longer couples the editor to a specific tool, and it correctly suppresses the focus ring whenever any tool is editing a shape.

The blur point from #8888 ("`editor.blur({ blurContainer: false })` bypasses the manager") was already resolved separately by #8895, which centralized the blur contract on `FocusManager`.

### Change type

- [x] `improvement`

### Test plan

1. Double-click a shape to start editing its text label.
2. Press Tab or an arrow key — the container focus ring should stay suppressed while editing.
3. Tab into the contextual toolbar — the focus ring should still be allowed there.

- [x] Unit tests

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +1 / -2    |
| Tests     | +4 / -4    |

Closes #8888


packages/editor/src/lib/editor/managers/FocusManager/FocusManager.ts
```diff
@@ -63,8 +63,7 @@ export class FocusManager {
 		const activeEl = container.ownerDocument.activeElement
 		// Edit mode should remove the focus ring, however if the active element's
 		// parent is the contextual toolbar, then allow it.
-		if (this.editor.isIn('select.editing_shape') && !activeEl?.closest('.tlui-contextual-toolbar'))
-			return
+		if (this.editor.getEditingShapeId() && !activeEl?.closest('.tlui-contextual-toolbar')) return
```
frolic, line 66: @mimecuvalo Steve said you'd be best to sign off on this change, since you understand the intent best

mimecuvalo (approved):
good idea ty ty

PR #9004, opened 2026-06-02
feat(editor): add a user preference for text outlines

### LINKED ISSUE #8808: Add a preference for text outlines
## Problem

Text shapes in tldraw render with a subtle outline (a multi-layer `text-shadow`) that makes labels readable against any canvas background. The outline is currently always on for text shapes (hardcoded `showTextOutline: true` in `TextShapeUtil`), with no way for users to turn it off.

Some users prefer cleaner, outline-free text — for design work, exports, or aesthetic reasons. Today the only way to disable it is to override `TextShapeUtil` options at the SDK level, which isn't accessible from tldraw.com or for typical end users.

## Proposed solution

Add a user preference that controls whether text outlines are rendered:

- Add `isTextOutlineEnabled` (or similar) to `TLUserPreferences` in `packages/editor/src/lib/config/TLUserPreferences.ts`.
- Default to `true` to preserve existing behavior.
- Wire it through `UserPreferencesManager` with a getter.
- Read it in `TextShapeUtil.options.showTextOutline` so the preference flows into `RichTextLabel` / `RichTextSVG` automatically.
- Surface it in the tldraw.com preferences menu alongside existing toggles like snap mode and wrap mode.

## Alternatives considered

- **Per-shape style** (a style picker control on text shapes): heavier, adds a new style to the schema, and most users likely want a global choice rather than a per-shape one.
- **Theme-only / CSS override**: works for SDK consumers but doesn't help dotcom users.
- **Leave as SDK option**: the current state — not discoverable for end users.

A user preference is the lightest change that makes the behavior controllable everywhere without expanding the shape schema.

## Use cases

- Users who find the outline visually noisy on light backgrounds.
- Users exporting clean text for use in other documents.
- Accessibility / readability preferences where users want a simpler text rendering.

frolic:
In order to let people turn off the readability outline on text, this PR adds an `isTextOutlineEnabled` user preference (default `true`, preserving current behavior). The outline was previously hardcoded on for text shapes (`TextShapeUtil`'s `showTextOutline: true`) with no way for end users — especially on tldraw.com — to disable it.

The outline is rendered via the `--tl-text-outline` CSS variable, which `DefaultCanvas` already flips to `none` for Safari and when zoomed far out (a level-of-detail optimization). This PR makes the new preference ride that same global switch, so a single toggle covers every label at once — text, geo, and arrow labels, plus any future label — while notes (which opt out with `showTextOutline={false}`) stay unaffected. SVG export is a separate render, so it reads the preference directly in `RichTextSVG`. The toggle is exposed in both the SDK Preferences menu and the tldraw.com Preferences menu, and persists for signed-in dotcom users via a new column on the user record.

Closes #8808

### Change type

- [x] `feature`

### Test plan

1. Create a text shape, plus a geo or arrow shape with a label.
2. Open Menu → Preferences and toggle "Text outline" off — outlines disappear from all labels at once (notes are unchanged); toggle back on to restore them.
3. Export a selection as SVG with the toggle on vs off — the exported text loses its `text-shadow` outline when disabled.
4. Reload — the preference is remembered.

- [x] Unit tests

### Release notes

- Add a "Text outline" preference (Menu → Preferences) to turn the readability outline on text labels on or off.

### API changes

- Added `isTextOutlineEnabled` to the `TLUserPreferences` interface.
- Added `UserPreferencesManager.getIsTextOutlineEnabled()`.
- Added the `ToggleTextOutlineItem` component export from `@tldraw/tldraw`.
- Added the `toggle-text-outline` action and UI event.

### Code changes

| Core code       | +63 / -8    |
| Tests           | +111 / -0   |
| Automated files | +157 / -133 |
| Apps            | +9 / -0     |
| Config/tooling  | +2 / -0     |


packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
```diff
@@ -260,7 +261,7 @@ export function RichTextSVG({
 		wordWrap: 'break-word' as const,
 		overflowWrap: 'break-word' as const,
 		whiteSpace: 'pre-wrap',
-		textShadow: showTextOutline ? 'var(--tl-text-outline)' : 'none',
+		textShadow: showOutline ? 'var(--tl-text-outline)' : 'none',
```
mimecuvalo, line 264: we'd need to do this in PlainTextLabel too

mimecuvalo (commented):
god the amount of files needed to add a preference 😆 / 😢 

i think the idea of this pref is prbly a good one. however, i think it shouldn't sit along the top-level prefs — it feels at odds with the other tool-level options there. i wonder if it would make sense to put in the Theme or Accessibility grouping maybe? or, maybe this is grouped together with some kind of larger "remove shadows/outlines — minimal mode" or something like that.

frolic:
I feel like inside Accessibility for now makes sense to me until we have more options that can justify recategorizing these things

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
frolic: Good catch on the original approach. Rather than duplicate the gate into `PlainTextLabel`, I changed tack: the canvas now disables outlines centrally in `DefaultCanvas` by setting `--tl-text-outline: none` (the same global switch already used for Safari and the zoom LOD), which `PlainTextLabel` picks up automatically via the shared `.tl-text__outline` class — so there is nothing to duplicate per-component, and any future label is covered too.

The only explicit per-render gate left is `RichTextSVG`, because SVG export renders into a separate DOM that the canvas variable does not reach. There is no plain-text SVG export path (no shape exports via `PlainTextLabel`), so nothing to mirror there. Done in 160376d.

frolic:
Moved the toggle into the **Accessibility** submenu (per @frolic's call) in 7619fc4 — it now sits alongside reduce motion / keyboard shortcuts / enhanced a11y in both the SDK and dotcom preference menus, rather than next to the tool-level toggles at the top level. Happy to fold it into a broader "minimal mode" later if that grows.

packages/tldraw/src/lib/shapes/shared/RichTextLabel.tsx
frolic: ^ sorry that was a claude response

This brings up an interesting question though in that why do we need this logic in here at all if `var(--tl-text-outline)` is already gated behind `getIsTextOutlineEnabled()`, will investigate

frolic: ah okay, so this was a bit unclear in the diff view, but I see now that this logic is applied to `RichTextSVG` which doesn't receive the CSS vars in the same way

so this particular code is for better SVG support whereas the other areas (e.g. plain text label) already inherit the CSS var from the DefaultCanvas conditional

mimecuvalo:
> Moved the toggle into the **Accessibility** submenu (per @frolic's call) in [7619fc4](https://github.com/tldraw/tldraw/commit/7619fc474494fede8e09a6734ee7f782a9f87085) — it now sits alongside reduce motion / keyboard shortcuts / enhanced a11y in both the SDK and dotcom preference menus, rather than next to the tool-level toggles at the top level. Happy to fold it into a broader "minimal mode" later if that grows.

i think this feels better placement, albeit still a bit awkward. i think @steveruizok's call on whether go / no-go. i'm still a little dubious on whether we need this pref. in first place ¯\\_(ツ)_/¯

frolic:
Closing for now as its unclear if we actually want this as a preference, we might just remove text outlines in general

Issue #6907, opened 2025-10-09
Safari pinch zoom resets selection to previous shapes


whiddershins:
### What happened?

Safari pinch-zoom resets selection to the previous shapes

Environment: Safari (desktop, trackpad pinch). Chrome and other browsers keep the active selection intact.
Steps
Open any doc with two shapes; select Shape A.
Shift-click Shape B so both are selected.
Click Shape C so only C is selected.
Pinch zoom on Safari.
Expected: Shape C remains selected.
Actual: Selection jumps back to A/B (whatever was selected before the last pointer interaction).
Root cause

pinch_start copies _selectedShapeIdsAtPointerDown only when that array is empty (packages/editor/src/lib/editor/Editor.ts:10245). We populate the stash on every pointer_down before the statechart updates selection, so it still holds the previous selection. When pinch_end fires (Safari delivers zoom as gesture events), both the immediate and deferred restores reapply that stale list. Chrome uses wheel events, so the pinch path—and the problem—never triggers.

Likely fix

Always copy the live selection when the pinch actually begins:

case 'pinch_start': {
  if (inputs.isPinching) return
  if (!inputs.isEditing) {
    this._pinchStart = this.getCamera().z
    this._selectedShapeIdsAtPointerDown = [...pageState.selectedShapeIds]
    this._didPinch = true
    inputs.isPinching = true
    this.interrupt()
  }
  return
}
After pointer interactions finish, clear the stash so the next pinch captures fresh state:

case 'pointer_up': {
  // existing logic…
  this._selectedShapeIdsAtPointerDown = []
  break
}
These two changes keep the existing deselection-protection logic but ensure Safari doesn’t restore outdated selections. Local testing on Safari confirms the selection now stays put after a pinch gesture.

### How can we reproduce the bug?

https://github.com/user-attachments/assets/d2311a2b-f8a0-4952-9af0-d147a0a82b54

### What browsers are you seeing the problem on?

Safari

### Where did this happen?

The tldraw.com website

### Contact Details

sarth@thirdwallstudio.com

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

steveruizok:
Oh good catch! I can reproduce.

PR #9006, opened 2026-06-02
fix(editor): restore pre-pinch selection on touch devices

### LINKED ISSUE #8397: Pinch gesture unintentionally changes selection on touch devices
When starting a pinch gesture on a touch device, the first finger down may change the current selection (e.g., tapping on a different shape or the canvas). If a pinch gesture immediately follows, the selection change is unintentional — the user was trying to zoom/pan, not select.

## Current behavior

1. User has shapes selected
2. User begins a two-finger pinch (first touch lands on a different shape or the canvas)
3. The first `pointer_down` changes the selection
4. Pinch gesture starts, but selection has already changed
5. When pinch ends, the changed selection persists

## Expected behavior

When a pinch gesture ends, the selection should be restored to what it was _before_ the first touch of the pinch. This way, incidental selection changes caused by the first finger of a two-finger gesture are rolled back.

## Context

This is a regression from the recent refactor replacing `@use-gesture/react` with custom gesture handling (#8391). The previous gesture library may have handled this case differently.

There is already some selection save/restore logic in the pinch handling code (`_selectedShapeIdsAtPointerDown` in `Editor.ts`), but the issue is that the selection snapshot is taken at `pinch_start` — by which point the first `pointer_down` has already changed the selection. The snapshot needs to capture the selection state from _before_ the first touch that initiated the pinch.

frolic:
In order to stop a two-finger pinch from leaving the selection changed on touch devices, this PR captures the pre-gesture selection on the first `pointer_down` and rolls back any incidental change when the pinch starts. Fixes #8397.

On touch, the first finger of a pinch fires a `pointer_down` that can change the selection (for example by landing on a different shape) before the second finger starts the gesture. The snapshot used to roll this back was taken at `pinch_start`, by which point the change had already happened, so the incidental selection persisted after the pinch. Now the snapshot is taken on the first `pointer_down` only, and `pinch_start` keeps that snapshot rather than overwriting it. When no `pointer_down` precedes the pinch — Safari delivers trackpad pinches as gesture events — `pinch_start` still captures the live selection, so the Safari fix from #6907 / #7777 is preserved.

The rollback happens at `pinch_start` (as soon as we know it's a pinch), not at `pinch_end`, so the pre-gesture selection is what's shown *during* the pinch rather than the wrong shape appearing selected for the whole gesture and snapping back on release.

The regression dates to #7777 (which made the `pinch_start` snapshot unconditional to fix a Safari case) and was surfaced by the gesture-handling rewrite in #8392. A deeper follow-up — unifying pinch detection and selection onto a single pointer event stream — is tracked separately.

### Change type

- [x] `bugfix`

### Test plan

Three layers of coverage, each of which fails on the pre-fix behavior and passes with the fix:

- `packages/tldraw/src/test/pinch.test.ts` — unit tests via `TestEditor`: rollback at pinch start (not just end), empty selection, consecutive pinches, the Safari/no-pointer-down path, and the editing guard.
- `packages/tldraw/src/test/pinch-dom.test.tsx` — integration tests that drive the real canvas handlers (`useCanvasEvents`, `useGestureEvents`) with native `pointerdown` and `touchstart`/`touchmove`/`touchend` events, including the case where the second finger's `pointer_down` is processed before `pinch_start`.
- `apps/examples/e2e/tests/test-camera.spec.ts` — a Mobile Chrome e2e that dispatches a real two-finger touch pinch via CDP `Input.dispatchTouchEvent`, asserting the selection is rolled back both during and after the pinch.

Manual check on a touch device: select a shape, start a two-finger pinch with the first finger landing on a different shape, zoom, lift both fingers — the originally selected shape stays selected throughout.

- [x] Unit tests
- [x] End to end tests

### Release notes

- Fix a two-finger pinch unintentionally changing the selection on touch devices.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +39 / -6   |
| Tests     | +376 / -0  |

### DIFF
--- a/apps/examples/e2e/tests/test-camera.spec.ts
+++ b/apps/examples/e2e/tests/test-camera.spec.ts
@@ -179,6 +179,70 @@ test.describe('camera', () => {
 		expect(await page.evaluate(() => editor.getZoomLevel())).toBe(8)
 	})
 
+	test('two-finger pinch does not change the selection', async ({ page, isMobile }) => {
+		test.skip(!isMobile)
+
+		client = await page.context().newCDPSession(page)
+
+		// Finger 1 lands here; finger 2 above it. Both are kept inside the mobile
+		// viewport so the gesture registers.
+		const f1 = { x: 196, y: 470, id: 0 }
+		const f2 = { x: 196, y: 300, id: 1 }
+
+		// Shape "a" (selected, off to the side) plus shape "b" placed directly under
+		// finger 1, so the first finger's pointer_down selects it.
+		await page.evaluate((f) => {
+			const p = editor.screenToPage({ x: f.x, y: f.y })
+			editor.createShapes([
+				{
+					id: 'shape:e2eA',
+					type: 'geo',
+					x: -1000,
+					y: -1000,
+					props: { w: 100, h: 100, fill: 'solid' },
+				},
+				{
+					id: 'shape:e2eB',
+					type: 'geo',
+					x: p.x - 60,
+					y: p.y - 60,
+					props: { w: 120, h: 120, fill: 'solid' },
+				},
+			] as any)
+			editor.setSelectedShapes(['shape:e2eA'] as any)
+		}, f1)
+
+		expect(await page.evaluate(() => editor.getSelectedShapeIds())).toEqual(['shape:e2eA'])
+
+		// Finger 1 down on shape b — through the real touch/pointer pipeline this
+		// changes the selection to b.
+		await client.send('Input.dispatchTouchEvent', { type: 'touchStart', touchPoints: [f1] })
+		await sleep(30)
+		expect(await page.evaluate(() => editor.getSelectedShapeIds())).toEqual(['shape:e2eB'])
+
+		// Finger 2 joins, then both fingers move apart to pinch-zoom, then lift.
+		await client.send('Input.dispatchTouchEvent', { type: 'touchStart', touchPoints: [f1, f2] })
+		await sleep(30)
+		const steps = 12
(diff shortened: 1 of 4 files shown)


frolic:
There's a bit of a UX decision to make here:

1. restore original selection on "pinch end", original issue's suggestion but it felt a bit strange to me
2. restore original selection on "pinch start", which feels like the best trade-off
3. defer selection on "up" so selection doesn't change until the pointer/touch is released, which feels most clear to me, but makes tap-and-drag interaction much less clear and probably not worth the trade-off

Below is a demo of 2 (my favorite option) and what this PR implements. It starts with the bottom-right shape selected, then starts the first finger of the pinch on the top-left shape, which briefly changes the selection until the second finger starts the pinch gesture, which reverts the selection.

A potential forth option is removing the selection during the pinch gesture, then restoring the original selection on pinch end.

https://github.com/user-attachments/assets/59ebc8e6-acc8-4d15-bd17-d33b04283ac0

max-drake:
@frolic 
I agree restoring on pinch end isn't right, and deferring selection to "up" will change a lot of behavior that I don't think should changed. I think restoring on pinch up makes the most sense

the jump in selection between the first finger coming down and starting the pinch _does_ feel a little jarring, but I think that's maybe fine for now?

the 4th option you mentioned is interesting, but I think it might also be weird to have"
1. shape A selected
2. 1 finger down changes selection to shape B
3. 2nd finger down + pinch starts hides selection
4. pinch end and selection is restored to shape A
5. 
this feels also jarring

max-drake (approved):
lgtm

PR #9008, opened 2026-06-03
fix(editor): coordinate touch pinch and selection through an explicit gesture state

### LINKED ISSUE #8397: Pinch gesture unintentionally changes selection on touch devices

frolic:
This builds on #9006 (which restores the pre-gesture selection after a pinch). In order to stop the _second_ finger of a two-finger pinch from changing the selection, this PR makes the pointer stream — the same stream selection runs on — the single authority for the multi-touch gesture lifecycle, so that decision no longer races the touch-stream `pinch_start`. Together with #9006 this closes #8397.

### Background

Touch pinch is detected on the **touch** event stream (`useGestureEvents`), while selection runs on the **pointer** event stream (`useCanvasEvents`). The two streams describe the same fingers with no guaranteed ordering, so whether the second finger's `pointer_down` is processed before or after `pinch_start` is browser-dependent — the root cause behind #8397. #9006 rolls back the _first_ finger's incidental selection change (intrinsic: a pinch isn't detectable until the second finger lands). This PR additionally makes the _second_ finger deterministic, and does so without moving pinch geometry off the touch stream, so the sub-pixel pinch accuracy from #8392 is preserved.

### What changed

- The editor tracks touch pointers on the pointer stream. The moment a second touch pointer lands, the interaction is treated as a multi-touch gesture and its pointer events are no longer routed to the active tool, so the second finger can't change the selection. This is deterministic because it rides the same `pointerdown` the selection handler sees — it doesn't depend on when `pinch_start` arrives.
- Added a `pointercancel` handler. The browser can cancel a pointer (gesture takeover, palm rejection) with no following `pointer_up`, which previously left `isPointing` / `capturedPointerId` stuck on. `Editor.cancelPointer()` tears the interaction down and interrupts the active tool without committing a click.
- The multi-touch / pinch lifecycle is now one explicit state on `InputsManager` (`idle → multi-touch → pinching → idle`). Both streams feed it — the pointer stream counts fingers, the touch/gesture stream confirms the pinch — and `getIsPinching()` is derived from it. The pointer dispatch gate collapses to a single `getGesturePhase() !== 'idle'` check.
- **Input-flag consolidation (two commits).** The overlapping `isPointing` / `isDragging` / `isRightPointing` / `isPanning` / `isSpacebarPanning` booleans are replaced by a single explicit `interaction` state — `idle | pointing { dragging, rightUndecided } | panning { via, pointerDown }`. All five boolean getters are now _derived_ from it (so impossible combinations like dragging-without-pointing are unrepresentable), and the Editor dispatch drives it through named transitions (`beginPointing` / `setDragging` / `beginPanning` / `endPanning` / …) instead of toggling independent flags. The real overlaps are modelled honestly: a middle/right-drag pan is `panning { pointerDown: true }`, which is why `isPointing` and `isPanning` are both true then; a pan sustained by spacebar + middle still ends only when the sustaining input releases.

### Commits

1. `fix(editor)`: gate selection during multi-touch + `pointercancel` recovery
2. `refactor(editor)`: pinch gesture lifecycle as an explicit state
3. `test(examples)`: e2e for the multi-touch selection gate
4. `refactor(editor)`: model pointer interaction as one explicit state (pointing/dragging)
5. `refactor(editor)`: fold pan + right-pointing into the interaction state
6. `test(tldraw)`: cover the interaction state truth table
7. `refactor(editor)`: drop the redundant internal input setters

Commits 4–7 are the input-flag consolidation and are cleanly separable from the #8397 fix (1–3) if we'd rather land them apart.

### Deliberately out of scope

- Pinch geometry stays on the touch stream, preserving the drift fix from #8392. Moving it to the pointer stream would re-introduce ~0.5px pan drift; the gesture state here doesn't preclude that as a future change.
- `isEditing` and `isPen` are left as independent flags — they are orthogonal modes (you can be editing or using a pen while pointing/panning), not part of the pointer/pan interaction state, so folding them in would be wrong rather than just risky.

### Concepts

| Term | Type | Meaning |
| --- | --- | --- |
| `TLGesturePhase` | `'idle' \| 'multi-touch' \| 'pinching'` | The multi-touch / pinch gesture lifecycle phase |
| `TLInteractionState` | `idle \| pointing \| panning` (discriminated union) | The pointer / pan interaction; the `isPointing` / `isDragging` / `isRightPointing` / `isPanning` / `isSpacebarPanning` getters are derived from it |
| `InputsManager.getGesturePhase()` | method | Read the current gesture phase (`getIsPinching()` is derived from it) |
| `InputsManager.getInteraction()` | method | Read the current pointer / pan interaction state |

### API changes

- Added `TLGesturePhase`, `TLInteractionState`, `InputsManager.getGesturePhase()`, and `InputsManager.getInteraction()`.
- The `isPointing` / `isDragging` / `isRightPointing` / `isPanning` / `isSpacebarPanning` getters and setters are unchanged in signature; they now derive from / map onto `getInteraction()`.
- Added `pointerType` to `TLPointerEventInfo` and to `getPointerInfo()`'s return value.
- Added `Editor.cancelPointer()` (internal).

### Change type

- [x] `bugfix`

### Test plan

1. On a touch device, select shape A, then begin a two-finger pinch with the first finger landing on shape B: B should not stay selected, the second finger must never select whatever it lands on, and after the pinch ends shape A is selected again.
2. Two-finger pan and pinch-to-zoom behave as before.
3. Desktop Safari trackpad pinch and wheel zoom/pan are unchanged.

- [x] Unit tests
- [x] End to end tests

Unit and jsdom-integration tests (driving the real `useCanvasEvents` / `useGestureEvents` hooks) cover the gate, the `pointercancel` recovery, and the gesture phases; the e2e (Mobile Chrome, CDP `Input.dispatchTouchEvent`) covers real multi-touch. True iOS Safari behaviour — native `pointercancel` on multitouch and the native touch-vs-pointer event ordering — can't be reproduced in jsdom or Chromium emulation and should be confirmed with on-device QA.

### Release notes

- Fix touch pinch so the second finger of a two-finger gesture can no longer change the selection, and recover cleanly when the browser cancels a pointer mid-interaction.

### Code changes

| Core code       | +372 / -102 |
| Tests           | +302 / -0   |
| Automated files | +46 / -8    |

### DIFF
--- a/apps/examples/e2e/tests/test-camera.spec.ts
+++ b/apps/examples/e2e/tests/test-camera.spec.ts
@@ -243,6 +243,57 @@ test.describe('camera', () => {
 		expect(await page.evaluate(() => editor.getZoomLevel())).not.toBe(1)
 	})
 
+	test('a second finger does not select the shape it lands on', async ({ page, isMobile }) => {
+		test.skip(!isMobile)
+
+		client = await page.context().newCDPSession(page)
+
+		// Finger 1 lands low in the viewport, finger 2 above it — both kept inside
+		// the mobile viewport so the gesture registers.
+		const f1 = { x: 196, y: 470, id: 0 }
+		const f2 = { x: 196, y: 250, id: 1 }
+
+		// A shape under each finger, nothing selected to start.
+		await page.evaluate(
+			(fingers) => {
+				const pb = editor.screenToPage({ x: fingers.f1.x, y: fingers.f1.y })
+				const pc = editor.screenToPage({ x: fingers.f2.x, y: fingers.f2.y })
+				editor.createShapes([
+					{
+						id: 'shape:e2eB',
+						type: 'geo',
+						x: pb.x - 60,
+						y: pb.y - 60,
+						props: { w: 120, h: 120, fill: 'solid' },
+					},
+					{
+						id: 'shape:e2eC',
+						type: 'geo',
+						x: pc.x - 60,
+						y: pc.y - 60,
+						props: { w: 120, h: 120, fill: 'solid' },
+					},
+				] as any)
+				editor.selectNone()
+			},
+			{ f1, f2 }
+		)
+
+		// Finger 1 down on shape b selects it.
+		await client.send('Input.dispatchTouchEvent', { type: 'touchStart', touchPoints: [f1] })
+		await sleep(30)
+		expect(await page.evaluate(() => editor.getSelectedShapeIds())).toEqual(['shape:e2eB'])
+
+		// Finger 2 lands on shape c. Because two touch pointers are now down, the
+		// multi-touch gate must stop the second finger from selecting c.
+		await client.send('Input.dispatchTouchEvent', { type: 'touchStart', touchPoints: [f1, f2] })
+		await sleep(30)
+		expect(await page.evaluate(() => editor.getSelectedShapeIds())).toEqual(['shape:e2eB'])
+
+		await client.send('Input.dispatchTouchEvent', { type: 'touchEnd', touchPoints: [] })
+		await sleep(50)
+	})
+
 	test.fixme('minimap', async () => {
 		// todo
 	})
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -951,6 +951,7 @@ export class Editor extends EventEmitter<TLEventMap> {
                 currentPagePoint: VecModel;
                 currentScreenPoint: VecModel;
                 gesturePhase: TLGesturePhase;
+                interaction: TLInteractionState;
                 isDragging: boolean;
                 isEditing: boolean;
                 isPanning: boolean;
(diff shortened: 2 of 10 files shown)


frolic:
closing this out, just wanted to see what this would look like

PR #9011, opened 2026-06-03
other(tldraw): revert press q to copy hovered styles (#8917)

steveruizok:
In order to ship the upcoming release without the in-progress "press q to copy hovered styles" feature, this PR reverts #8917.

The original feature added a `copy-hovered-styles` action that let users press `q` while hovering a shape to copy its styles onto the next-created shape. We want to release without it, so this revert removes the action, its keyboard shortcut entry, and the associated translation key.

The only non-trivial part of the revert was a conflict in the generated `api-report.api.md`: a later PR added the `page-menu.max-pages-reached` translation key in the same union type. That key is preserved; only `action.copy-hovered-styles` was removed.

### Change type

- [x] `other` (revert)

### Test plan

1. Hover a shape, press `q`, then draw a new shape.
2. Confirm the hovered shape's styles are no longer copied (the keyboard shortcut is gone).

### Release notes

- Remove the "press q to copy hovered styles" action.

### API changes

- Removed `'action.copy-hovered-styles'` from the `TLUiTranslationKey` union.

### Code changes

| Core code       | +0 / -33   |
| Config/tooling  | +0 / -1    |


mimecuvalo (approved):
👍

PR #9018, opened 2026-06-03
refactor(editor): move ShapeIndicatorOverlayUtil to the tldraw package

max-drake:
👩: this also has the nice effect of removing a hardcoded reference to the select tool from `editor` and moving it up into `tldraw`

---

In order to keep `@tldraw/editor` free of concrete default overlay utils, this PR moves `ShapeIndicatorOverlayUtil` (and its `TLShapeIndicatorOverlay` type) out of the editor package and into `tldraw`, where every other built-in overlay util already lives (`BrushOverlayUtil`, `ScribbleOverlayUtil`, `ShapeHandleOverlayUtil`, and the rest). `ShapeIndicatorOverlayUtil` was the lone concrete overlay util still defined in the editor package; after this change `defaultOverlayUtils` is assembled entirely from tldraw-local utils, and the editor package keeps only the abstractions: the `OverlayUtil` base class and the `strokeShapeIndicators` helper.

The `strokeShapeIndicators` helper stays in `@tldraw/editor` — it depends only on `ShapeUtil.getIndicatorPath` and is shared by other indicator overlays (e.g. `CollaboratorShapeIndicatorOverlayUtil`). It's been extracted into its own `strokeShapeIndicators.ts` module so the moved class can import it from the editor.

For consumers importing from `'tldraw'`, nothing changes — `ShapeIndicatorOverlayUtil` and `TLShapeIndicatorOverlay` are still exported from there. Only code importing those two symbols directly from `'@tldraw/editor'` needs to update its import path.

The `defaultOverlayUtils` array is also reordered alphabetically. This is cosmetic: registration order only breaks ties between utils with equal `zIndex`, and every overlay util has a distinct `zIndex`, so paint and hit-test order are unchanged.

This PR also tidies the `indicators-logic` example: the four geo shapes are given a clearer layout (an ellipse and a text shape are added), the two bottom-left shapes are grouped on mount to show indicators on a group, and the README explains what shape indicators are.

### Change type

- [x] `api`

### Test plan

1. Open the `indicators-logic` example; confirm indicators are drawn around every shape, including the grouped pair in the bottom left and the group itself.
2. In any default tldraw editor, confirm selection and hover indicators still render and hit-test correctly (selection, brushing, handles).
3. Confirm `import { ShapeIndicatorOverlayUtil, TLShapeIndicatorOverlay } from 'tldraw'` resolves.

- [x] Unit tests

### Release notes

- Move `ShapeIndicatorOverlayUtil` and `TLShapeIndicatorOverlay` from `@tldraw/editor` to `tldraw`. They remain exported from `tldraw`; update any direct `@tldraw/editor` imports of these two symbols.

### API changes

`@tldraw/editor`:

- Breaking! Removed `ShapeIndicatorOverlayUtil` (moved to `tldraw`).
- Breaking! Removed `TLShapeIndicatorOverlay` (moved to `tldraw`).
- `strokeShapeIndicators` is unchanged and still exported (now from its own module).

`tldraw`:

- Added `ShapeIndicatorOverlayUtil` (moved from `@tldraw/editor`).
- Added `TLShapeIndicatorOverlay` (moved from `@tldraw/editor`).
- `defaultOverlayUtils` now references the tldraw-local `ShapeIndicatorOverlayUtil`; its entries are reordered alphabetically (behavior-preserving, since all z-indices are distinct).

### Code changes

| Core code       | +108 / -107 |
| Automated files | +28 / -29  |
| Documentation   | +19 / -6   |

### DIFF
--- a/packages/tldraw/src/lib/overlays/ShapeIndicatorOverlayUtil.ts
+++ b/packages/tldraw/src/lib/overlays/ShapeIndicatorOverlayUtil.ts
@@ -1,14 +1,6 @@
+import { OverlayUtil, strokeShapeIndicators, TLOverlay } from '@tldraw/editor'
 import { computed } from '@tldraw/state'
-import { createComputedCache } from '@tldraw/store'
-import { TLShape, TLShapeId } from '@tldraw/tlschema'
-import type { Editor } from '../Editor'
-import { OverlayUtil, TLOverlay } from './OverlayUtil'
-
-interface RelevantInstanceFlags {
-	isChangingStyle: boolean
-	isHoveringCanvas: boolean | null
-	isCoarsePointer: boolean
-}
+import { TLShapeId } from '@tldraw/tlschema'
 
 /** @public */
 export interface TLShapeIndicatorOverlay extends TLOverlay {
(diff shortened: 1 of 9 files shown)


packages/tldraw/src/lib/overlays/ShapeIndicatorOverlayUtil.ts
```diff
@@ -18,87 +10,10 @@ export interface TLShapeIndicatorOverlay extends TLOverlay {
...
-		ctx.restore()
-		if (additionalPaths) {
-			for (const p of additionalPaths) ctx.stroke(p)
-		}
-		ctx.restore()
-	}
-
-	ctx.stroke(batched)
+interface RelevantInstanceFlags {
+	isChangingStyle: boolean
```
mimecuvalo, line 14: i can't tell why this file is still around - just for RelevantInstanceFlags?

max-drake: i don't understand, `ShapeIndicatorOverlayUtil` is used in `defaultOverlayUtils.ts` ?

mimecuvalo: oh sorry ignore — i had misread — thought that the file had mostly moved/renamed but had a lingering artifact. i see now it has a lot more content. nvm!

mimecuvalo (approved):
👍

PR #9021, opened 2026-06-03
chore: auto `--host` templates and allow arbitrary args to be forwarded to `dev-template` script

max-drake:
In order to test templates from another device on the local network, this PR lets `yarn dev-template <name>` forward any extra arguments to the template's vite dev server. For example, `yarn dev-template vite --host` exposes the dev server on the LAN.

lazy applies passthrough args (`--`) to *every* matched task, so forwarding `--host` through the combined `lazy run dev` would also hand it to the `packages/tldraw` (chokidar) and `apps/bemo-worker` (tsx) dev scripts, which would break. To scope the args correctly, the shared deps keep running through lazy while vite runs directly in the template directory with the extra args.

Because lazy installs no signal handlers of its own, the deps run in their own process group (`set -m`) and cleanup signals the whole group on exit, so lazy and the watchers it spawns are torn down instead of orphaned. With no extra args, the original single `lazy run dev` fast path is unchanged.

### Change type

- [x] `other`

### Test plan

1. Run `yarn dev-template vite` — behaves as before (template, tldraw, and bemo-worker all start).
2. Run `yarn dev-template vite --host` — vite prints a Network URL; open it from another device on the LAN.
3. Quit with Ctrl-C and confirm no orphaned `vite`, `chokidar`, or `tsx` processes remain (`ps ax | grep -E 'chokidar|tsx'`).

### Code changes

| Config/tooling | +23 / -1   |


internal/scripts/dev-template.sh
```diff
@@ -15,4 +17,24 @@ if [ ! -d "templates/$template_name" ]; then
...
+lazy_pid=$!
+disown
+cleanup() {
+    # negative pid signals the whole process group
+    kill -- "-$lazy_pid" 2>/dev/null
+}
+trap cleanup EXIT INT TERM
+
+cd "$workspace_root/templates/$template_name" || exit 1
+yarn dev "${vite_args[@]}"
```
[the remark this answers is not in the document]
MitjaBezensek: This does seem like a real issue. We might skip any dependencies if we run `yarn dev` directly?

internal/scripts/dev-template.sh
```diff
@@ -15,4 +17,24 @@ if [ ! -d "templates/$template_name" ]; then
     exit 1
 fi
 
-LAZYREPO_PRETTY_OUTPUT=0 lazy run dev --filter="templates/$template_name" --filter='packages/tldraw' --filter='apps/bemo-worker'
+if [ ${#vite_args[@]} -eq 0 ]; then
+    LAZYREPO_PRETTY_OUTPUT=0 lazy run dev --filter="templates/$template_name" --filter='packages/tldraw' --filter='apps/bemo-worker'
+    exit $?
+fi
+
+# lazy does not forward extra args to package scripts, so run vite directly in the template.
```
mimecuvalo: i don't think we need this part - i see the comment about lazy passing parameters but i don't think we need to support that

max-drake: @MitjaBezensek mentioned this might be nice to have, if  we don't want it, I can revert this whole file and this pr can just be adding the --host flags to the templates dev script

MitjaBezensek: I've wanted something like this in the past. Like you can run e2e tests from the repo root but you can't filter the tests. You have to go into the folder where tests live and run from there. Similar with things like run playwright in UI mode,...

mimecuvalo (approved):

Issue #8481, opened 2026-04-07
Mermaid detection is too aggressive for common words like "timeline" and "kanban-board"


angrycaptain19:
## Description

Pasting common English words or phrases that happen to start with a mermaid diagram keyword incorrectly triggers the mermaid diagram renderer on tldraw.com. The text is treated as mermaid syntax and rendered as a (broken) diagram instead of being inserted as a plain text shape.

## Examples

- Pasting **"timeline"** → incorrectly rendered as mermaid
- Pasting **"kanban-board"** → incorrectly rendered as mermaid
- Likely also affects other common words: "graph", "pie", "journey", "block", "info"

## Steps to reproduce

1. Go to tldraw.com
2. Copy the word "timeline" to your clipboard
3. Paste it onto the canvas

## Expected behavior

Common words like "timeline" should be pasted as regular text shapes.

## Actual behavior

The text is detected as a mermaid diagram and rendered incorrectly.

## Root cause

The mermaid detection regex in `simpleMermaidStringTest.ts` matches diagram keywords as prefixes without any further validation:

```typescript
const DIAGRAM_KEYWORD_REGEX =
  /^\s*(flowchart|graph|sequenceDiagram|classDiagram|...|kanban|...)/
```

Two issues:

1. **No word boundary** — "kanban-board" matches because "kanban" is a prefix
2. **No structural validation** — a single word like "timeline" matches even though a real mermaid diagram would have content after the keyword (e.g., `timeline\n  title My Timeline\n  ...`)

The detection needs to require some evidence of actual mermaid diagram structure beyond just the keyword.

angrycaptain19:
## Research

### Affected file

**`apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.ts:15-16`**

The `DIAGRAM_KEYWORD_REGEX` matches mermaid diagram keywords as simple prefixes:

```typescript
const DIAGRAM_KEYWORD_REGEX =
  /^\s*(flowchart|graph|sequenceDiagram|classDiagram|...|kanban|architecture|...)/
```

`kanban` is a valid mermaid diagram type, but the regex has no word boundary after the keyword group. So "kanban-board", "gantt-chart", "timeline-view", etc. all falsely match.

### Paste flow

1. User pastes text → `useClipboardEvents.ts` processes clipboard
2. `SneakyMermaidHandler.tsx:18-35` registers a `text` external content handler
[...]

PR #9022, opened 2026-06-03
fix(dotcom): tighten mermaid detection for common words

mimecuvalo:
In order to stop common English words from being rendered as broken mermaid diagrams on tldraw.com, this PR tightens the mermaid detection heuristic in `simpleMermaidStringTest`. Pasting words like "timeline", "kanban-board", or "info" previously matched a mermaid diagram keyword as a bare prefix and was sent to the mermaid renderer instead of being inserted as a plain text shape (#8481).

The detection regex matched diagram keywords with no word boundary and no structural validation, so:

1. **No word boundary** — "kanban-board" matched because "kanban" is a prefix.
2. **No structural validation** — a single word like "timeline" matched even though a real diagram has content after the keyword.

This PR addresses both:

- Adds a trailing `(?![\w-])` word-boundary lookahead so keywords glued to more text no longer match: "kanban-board", "gantt-chart", "information", "graphql".
- Requires evidence of an actual diagram beyond the bare keyword — either a recognized variant suffix (e.g. `sankey-beta`, `stateDiagram-v2`) or additional diagram content after the keyword.
- Keeps explicit ` ```mermaid ` fences permissive: a bare keyword inside a fence the user has labelled as mermaid is still detected, since intent is clear.

### Change type

- [x] `bugfix`

### Test plan

1. Go to tldraw.com.
2. Copy the word "timeline" (or "kanban-board", "info") and paste it onto the canvas — it should appear as a plain text shape, not a diagram.
3. Paste a real diagram (e.g. `timeline\n  title History` or a ` ```mermaid ` fenced diagram) — it should still render as a mermaid diagram.

- [x] Unit tests

### Release notes

- Fix mermaid detection on tldraw.com so common words like "timeline", "kanban-board", and "info" paste as plain text instead of being rendered as broken diagrams.

### DIFF
--- a/apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.test.ts
+++ b/apps/dotcom/client/src/components/SneakyMermaidHandler/simpleMermaidStringTest.test.ts
@@ -1,7 +1,7 @@
 import { simpleMermaidStringTest, stripMarkdownMermaidFence } from './simpleMermaidStringTest'
 
 describe('simpleMermaidStringTest', () => {
-	describe('bare keywords', () => {
+	describe('diagram keywords with content', () => {
 		const keywords = [
 			['flowchart', 'flowchart TD\n  A --> B'],
 			['graph', 'graph LR\n  A --> B'],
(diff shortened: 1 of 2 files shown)


mimecuvalo:
bad bot closing

Issue #9026, opened 2026-06-04
Make custom themes work with sync automatically and document the server-side requirement


MitjaBezensek:
🧑: 1. we should improve the dev experience when using custom themes. currently the server side experience is quite janky, they need to understand the logic for registering themes, colors, fonts. we should allow an option to pass in the themes and the registration happens automatically (similar to what we already do on the client)
2. we should update our docs to let people be aware that server side configuration is necessary when adding new colors, fonts,...

---

## Problem

Custom theme colors and fonts work locally but crash over `tldraw-sync` with a `sync error INVALID_RECORD`. This is a recurring source of confusion (see linked support thread).

The root cause is that the **client and the sync server are separate processes with separate schemas**, and custom style values must be registered on both:

- Custom colors/fonts are not a special record format. They are added to a style-prop **enum** at runtime via `registerColorsFromThemes()` / `registerFontsFromThemes()`, which mutate the global `DefaultColorStyle` / `DefaultLabelColorStyle` / `DefaultFontStyle` enums.
- Shapes store the style **name** (e.g. `color: 'pink'`, `font: 'pixel'`), not the values. The sync server re-validates every incoming record against its own schema's enums.
- On the **client**, `createTLStore` and `useSync` call `registerColorsFromThemes` / `registerFontsFromThemes` automatically when you pass `themes`. So the client just works.
- On the **server**, `TLSocketRoom` builds its schema with a bare `createTLSchema()` (the default when no `schema` is passed). Nothing registers the custom names, so the server enum only knows the built-in values. A synced shape with `color: 'pink'` fails validation → `INVALID_RECORD` → the socket closes → the page crashes.

### Why this is a bad DX today

To make custom themes work over sync, a developer currently has to understand all of the following and wire it up by hand on the server:

1. That registration is required on the server at all (not discoverable; not documented).
2. That `registerColorsFromThemes` / `registerFontsFromThemes` must run **before** `createTLSchema()` because they mutate global enums (ordering footgun).
3. That removals matter too — these functions also *remove* names absent from the themes, so client and server must use the exact same `themes` object or the enums drift.
4. That the resulting `schema` must be passed explicitly into `TLSocketRoom` / `TLSyncRoom`, not left to default.

This is a lot of internal mechanism to expose for what is conceptually "I added a color."

### 1. Accept `themes` on the server room API (SDK)

[...]

PR #9027, opened 2026-06-04
feat(dotcom): add pink to the color palette

MitjaBezensek:
In order to add a custom "pink" color to tldraw.com, this PR registers the new color on both the client and the sync worker so that shapes using it pass validation everywhere. Relates to #9026.

Custom colors are added to a runtime style-prop enum, not a special record format — shapes store only the color *name* (`pink`), and the sync worker re-validates every incoming record against its own schema. The client already registers theme colors automatically when you pass `themes` to `useSync`/`<Tldraw>`, but the worker builds its schema independently, so a synced `pink` shape would otherwise be rejected with `INVALID_RECORD`. This PR registers the color on both sides through one shared module.

### Concepts

| Term | Where | Meaning |
| --- | --- | --- |
| `PINK_LIGHT` / `PINK_DARK` | `dotcom-shared` | The pink color values for the light/dark palettes (client rendering only) |
| `registerDotcomColors()` | `dotcom-shared` | Registers dotcom's custom color names into the global color and label-color enums |
| `createDotcomTLSchema()` | `dotcom-shared` | `createTLSchema()` with dotcom's custom colors registered first; used everywhere dotcom built a schema directly |
| `themes` | client | Full theme (defaults + pink), passed to `useSync` (registration + store schema) and `<Tldraw>` (rendering + palette) |

### Change type

- [x] `feature`

### Test plan

1. Open a file on tldraw.com.
2. Select the pink color and add a sticky note (and a geo shape).
3. Confirm the shapes render pink and there is no `sync error INVALID_RECORD` / page crash.
4. Reload the page and confirm the pink shapes persist.
5. Open the same file in a second tab/client and confirm the pink shapes sync.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Area | Net LOC |
| --- | --- |
| `packages/dotcom-shared` (shared color module + export) | +91 |
| `apps/dotcom/sync-worker` (register on server schemas + room) | +15 |
| `apps/dotcom/client` (themes, palette label, import schemas) | +30 |
| Generated locales | +9 |

### Release notes

- Add a pink color to the tldraw.com palette.

### DIFF
--- a/apps/dotcom/client/public/tla/locales-compiled/en.json
+++ b/apps/dotcom/client/public/tla/locales-compiled/en.json
@@ -737,6 +737,12 @@
       "value": "Check your email for a verification code."
     }
   ],
+  "8dc5344bc0": [
+    {
+      "type": 0,
+      "value": "Pink"
+    }
+  ],
   "8eb2c00e96": [
     {
       "type": 0,
--- a/apps/dotcom/client/public/tla/locales/en.json
+++ b/apps/dotcom/client/public/tla/locales/en.json
@@ -329,6 +329,9 @@
   "8c2d3d730c": {
     "translation": "Check your email for a verification code."
   },
+  "8dc5344bc0": {
+    "translation": "Pink"
+  },
   "8eb2c00e96": {
     "translation": "Shared with you"
   },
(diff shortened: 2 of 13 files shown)


mimecuvalo:
wait are we actually adding this color or is this just an example PR?

MitjaBezensek:
> wait are we actually adding this color or is this just an example PR?

No. Was just trying to understand what you need to do on the server in order to support new colors since one of the customers asked about it / had trouble implementing it.

Looks like our dx is not the best for this, we also don't mention it in the docs either. Opened an issue for this here https://github.com/tldraw/tldraw/issues/9026

Will close this PR.

PR #9028, opened 2026-06-04
feature(tldraw): apply/copy styles between shapes

kaneel:
Use shift + Q as a shortcut to select the style form a shape and have it be applied to the next shape.

Have the same base behavior that https://github.com/tldraw/tldraw/pull/8917 initially introduced to allow you to capture the style of a shape on hover, but rather than apply it directly to an existing shape if you have one selected it should just apply the style to the next shape.

### API changes

- added a `shift+q` shortcut to copy styles form a hovered shape in order to apply it for the next shape

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

<breathing intensifies>

1. Create a shape using any colour
2. Create an other shape using any other colour
3. Hover first shape
4. press `shift+q`
5. create an other shape

The shape created should now have the same style than the first shape

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Added new shortcuts in order to copy styles from a selected shape. `shift+q` will copy the styles of a hovered shape and set it for the next shape to be created


max-drake:
serious: I feel like "mouse has styles copied that will be applied when clicking a shape" is, as a state,  too potentially disruptive to not have any kind of hint to the user about it, so I think if we add this
potential counter argument to this I could see is that this is a power user feature, and so they'll be pros enough to know about it? but still idk

joke: we should do this to hint to the user you're in this state
https://github.com/user-attachments/assets/fadf6c78-d800-4fd9-a96b-9c712e58f308

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1848,6 +1851,58 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				label: 'action.copy-hovered-styles',
+				kbd: 'shift+q',
+				async onSelect(source) {
+					// Use the editor's hovered shape so this matches the shape the user sees
+					// highlighted, rather than re-running a hit test with different options.
+					const hovered = editor.getHoveredShape()
+
+					// Only run from the main select idle state. Checking the leaf node's id would
+					// also match nested idle states like `select.crop.idle`, where this shouldn't run.
+					if (!hovered || !editor.isIn('select.idle')) return
```
max-drake: what's the thinking behind having this (and the `copy-styles` one) only working in select, or select.idle, as is the case here?

max-drake: the more I test this, the more I think that this should work in all modes (or at least all idle modes). further to that, it should not use `getHoveredShape()` as the shape to check, because that is only updated by the `select` tool. instead we should use `editor.getShapeAtPoint(...)`, and match the `margin` we use for normal hit testing (which is `editor.options.hitTestMargin / editor.getZoomLevel()`). you can find a reference here at `getHoveredShapeId.ts`

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -1848,6 +1851,58 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+					}
+					if (styles.length === 0) return
+
+					const selectTool = editor.getStateDescendant<SelectTool>('select')
+					if (!selectTool) return
+					selectTool.stylesToPaste = styles
+
+					trackEvent('copy-styles', { source })
+				},
+			},
```
max-drake: I think for now, we should axe this part of the PR. the interaction of "shape styles will be pasted on pointer down" does not work I think (for a couple reasons)
a couple alternatives that I think would be more focused
1. remove this action entirely
2. remove this action, and let the `"shift+q"` action grab selected styles if you have a selection and no hovered shapes, merging the two behaviors but decreasing the interaction surface area
3. keep this action, but store copied styles on `TLInstance`, away from select tool, and paste them into a selection using `"cmd+alt+v"` (instead of clicking on a shape)

i think for the scope of this pr, I prefer 1 or 2, and I'd like to see 3 in a followup

kaneel: yeah that's definitely not working for me either, gave this as a try but it feels weird to use. 
for this to work, a visual interaction should occur, a bubble or something that follow the pointer for example… but anyway I don't think we have anything working like that on click so that feels a bit… estranged?

packages/tldraw/src/lib/ui/context/actions.tsx
kaneel: when i first implemented these shortcuts I got them to work while selecting selected, or drawing I think even, I figured this would be better to early return when non select.idle state 🤔

max-drake: I would prefer for this to happen in _any_ idle state. 

imagine if a user is in `"geo.idle"` or `"text.idle"` if we only allowed vacuuming styles while in `"select"`, then a user would have to swap back to `"select"` from `"geo"`, etc, to vacuum styles, which kind of defeats the purpose of having a nice fast keyboard shortcut

kaneel: Ah yes, did not think of that!

max-drake (approved):
not sure if you're still battling tests but approved pending that if so :)

kaneel:
> not sure if you're still battling tests but approved pending that if so :)

not, we all good here

PR #9029, opened 2026-06-04
refactor(dotcom): rename group admin role to member

### LINKED ISSUE #9014: Rename group Admin role to Member
🧑: groups: rename Admin to Member

---

In the groups feature on tldraw.com, the non-owner group role is currently labeled **Admin**. This is misleading, since the role does not confer admin-level capabilities — it's really a standard group member. Rename **Admin** to **Member** so the role label matches what it actually represents.

### Where it appears

The role label shows up in the group settings dialog (`GroupSettingsDialog.tsx`):

- The members list, next to each member's name
- The role select dropdown (Owner / Admin)

### Scope to confirm

Two possible levels of change:

1. **Label only** — rename the user-facing string `Admin` → `Member`. Low risk, no data changes.
2. **Label + role value** — also rename the stored role value `'admin'` → `'member'`. This is more invasive: it requires a DB migration to update the `CHECK` constraint and existing rows, plus updates to backend and client code that reference `'admin'`.

The simplest fix is label-only. We should decide whether we also want the underlying role value renamed for consistency.

frolic:
In order to make the group role label match what it actually represents, this PR renames the non-owner group role from `admin` to `member`. The role confers no admin-level capabilities — it's a standard group member — so the old "Admin" label was misleading. Closes #9014.

This builds on the capabilities abstraction from #9067: authorization asks `can(role, capability)` and never branches on the role name, so the rename is purely a relabel plus a data migration — no authorization logic changes.

What changes:

- **Role definition** (`roles.ts`) — the `admin` key in the `roles` capability table becomes `member`. `Role` is derived from these keys, so the `group_user.role` enum in `tlaSchema.ts` and every `Role`-typed call site update for free.
- **Sync worker** — accepting a group invite now assigns `role: 'member'`.
- **Client UI** — the members list and role dropdown show "Member" (the dropdown is data-driven from the role → label map, so only the label and the map key change).
- **Database** — migration `033` drops the `group_user` role check constraint, rewrites existing `admin` rows to `member`, then re-adds the constraint allowing `('member', 'owner')`.

Because the new check constraint rejects `'admin'`, the migration should run as part of the deploy that ships this change. In the `deploy-dotcom` flow migrations run before the sync worker is deployed, so existing rows are rewritten and the constraint swapped before any new code writes `'member'`. Groups is behind the `groups_backend` / `groups_frontend` flags and was added recently, so little data has accumulated.

### Change type

- [x] `improvement`

### Test plan

1. Open a group's settings dialog as the owner.
2. Confirm the members list shows "Member" (not "Admin") next to non-owner members.
3. Confirm the role dropdown offers "Owner" and "Member".
4. Change a member's role to Owner and back; confirm it persists.
5. Accept a group invite as a new user; confirm they join as a member.

- [x] Unit tests
- [ ] End to end tests

`roles.test.ts` covers the role → capability matrix and `isRole` under the new name; `mutators.test.ts` group tests run against `member` fixtures.

### Code changes

| Core code       | +2 / -8    |
| Tests           | +35 / -35  |
| Automated files | +9 / -9    |
| Apps            | +17 / -4   |

### DIFF
--- a/apps/dotcom/client/public/tla/locales-compiled/en.json
+++ b/apps/dotcom/client/public/tla/locales-compiled/en.json
@@ -699,6 +699,12 @@
       "value": "."
     }
   ],
+  "858ba4765e": [
+    {
+      "type": 0,
+      "value": "Member"
+    }
+  ],
   "85a082de1b": [
     {
       "type": 0,
@@ -1275,12 +1281,6 @@
     {
       "type": 0,
       "value": "Legal summary"
-    }
-  ],
-  "e3afed0047": [
-    {
-      "type": 0,
-      "value": "Admin"
     }
   ],
   "e404559826": [
--- a/apps/dotcom/client/public/tla/locales/en.json
+++ b/apps/dotcom/client/public/tla/locales/en.json
@@ -314,6 +314,9 @@
   "815e116a2e": {
     "translation": "Have a bug, issue, or idea for tldraw? Let us know! Fill out this form and we will follow up over email if needed. You can also <discord>chat with us on Discord</discord> or <github>submit an issue on GitHub</github>."
   },
+  "858ba4765e": {
+    "translation": "Member"
+  },
   "85a082de1b": {
     "translation": "Please refresh the page to get the latest version of tldraw."
   },
@@ -548,9 +551,6 @@
   },
   "e2ae80a510": {
     "translation": "Legal summary"
-  },
-  "e3afed0047": {
-    "translation": "Admin"
   },
   "e404559826": {
     "translation": "Terms of service"
--- a/apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
@@ -33,7 +33,7 @@ const messages = defineMessages({
 	copyInviteLink: { defaultMessage: 'Copy invite link' },
 	members: { defaultMessage: 'Members' },
 	owner: { defaultMessage: 'Owner' },
-	admin: { defaultMessage: 'Admin' },
+	member: { defaultMessage: 'Member' },
 	you: { defaultMessage: 'you' },
 	dangerZone: { defaultMessage: 'Danger zone' },
 	leaveGroup: { defaultMessage: 'Leave group…' },
@@ -62,7 +62,7 @@ export function GroupSettingsDialog({ groupId, onClose }: GroupSettingsDialogPro
 
 	const namePlaceholderMsg = useMsg(messages.namePlaceholder)
 	const ownerMsg = useMsg(messages.owner)
-	const adminMsg = useMsg(messages.admin)
+	const memberMsg = useMsg(messages.member)
 	const youMsg = useMsg(messages.you)
 
 	// Get group data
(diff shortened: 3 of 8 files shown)


packages/dotcom-shared/src/mutators.ts
```diff
@@ -123,7 +123,7 @@ async function assertUserIsGroupMember(
 	assert(groupUser, ZErrorCode.forbidden)
 }
 
-async function assertUserIsGroupAdminOrOwner(
+async function assertUserIsGroupMemberOrOwner(
```
frolic: thoughts on `assertUserIsGroupMemberOrOwner` or `assertUserIsGroupOwnerOrMember` or maybe just `assertUserIsGroupMember` (an owner is a member, just with extra permissions)?

steveruizok: This may be guarding admin-like behaviors. IIRC admins do have "admin-like" permissions, so perhaps we're doing to the wrong thing here by just renaming members to admins. Maybe the answer here is to preserve the admin role but add the member role (with the same set of permissions to start with?) and make the member role default, and migrate all admins to members? And then decide on the different permissions for members? I'm sure there's a truth table somewhere but we may have only implemented owners / admins for our own internal convenience.

mimecuvalo: a driving principle here is high-trust w.r.t. sharing models. otherwise you get into byzantine UI that gets in people's way.

to that end, we should avoid adding a _third_ role type if at all possible. part of the problem is that this can start to bleed into permissions which we don't have in general. (can edit / can view). (like, can you have an owner that's view-only? weird)

so, in my head, it's more like edit access or view access w.r.t. to folders ("groups"). we're not setting access levels for people's status here in an org. this boils down to file access and whether you can read / write. the only thing that the Owner has power over is whether the group is deleted or not.

finally, keep the surface area here tight in the first version that's shipped, it's hard to claw back something once it's out. if there's a need for a third-level of access we can add that after feedback / iteration once this is in the wild.

## in sum:
- `owner`: can invite anyone | can delete the group
- `member`("admin"): can invite anyone | _cannot_ delete the group (can forget the group)
- _separate lever_: we can add `edit` | `view` permissions on an individual basis. whether this goes into V1 or V2, i'm ambivalent

MitjaBezensek: Few other things to think about:
* Who can rename groups?
* Who can move files out of the group (example that @jsscclr shared)?
* Can owners remove members? Currently we don't have a way of doing that, you can only leave a group. What if the group invite link gets shared / leaks and some unwanted people join?

angrycaptain19: I think this is where we landed on our conversations in discord re:high trust

- Anyone can rename a group
- Anyone can move files out of a group

I'd lean towards no one can remove members for V1, but if we think this is critical then anyone could remove anyone except removing the owner. 

I also think this PR is probably good to get merged in? If we're saying there's no functional shift between Admin -> Member

angrycaptain19: @frolic Per discussion this afternoon keep Owner & Member

That can be the first piece of this PR and then we can have a separate PR for ability to kick members

frolic:
converting back to draft until #9067 and #9033 land

MitjaBezensek (approved):

mimecuvalo (approved):

mimecuvalo:
heh double-approved 👍 👍

MitjaBezensek:
Let's not merge just yet, let's do it after the dotcom release.

PR #9033, opened 2026-06-05
fix(dotcom): notify Zero of schema changes to avoid full resets on Supabase

### LINKED ISSUE #9031: Use Zero schema change hooks to avoid full resets on Supabase DDL changes
## Problem

tldraw.com runs [Zero](https://zero.rocicorp.dev) with its upstream Postgres on Supabase. Zero relies on Postgres **event triggers** to detect DDL/schema changes efficiently and migrate its replica in place.

Two issues bite us here (per [Zero's docs](https://zero.rocicorp.dev/docs/connecting-to-postgres#schema-change-hooks)):

- Some hosted providers restrict event triggers. Supabase specifically **does not fire DDL event triggers for `ALTER PUBLICATION`**.
- When Zero can't detect a schema change via an event trigger, **any schema change triggers a full reset of all server-side and client-side state**. The replica is rebuilt from scratch and every connected client re-downloads everything. This gets more expensive as the database grows.

We already ship migrations that hit exactly this case. For example `apps/dotcom/zero-cache/migrations/023_groups.sql`:

```sql
ALTER PUBLICATION zero_data ADD TABLE public."group", public."group_user", public."group_file";
```

plus the many `ALTER TABLE ... ADD COLUMN` migrations. On Supabase each of these risks a full Zero reset rather than an incremental schema update.

## Proposed solution

Adopt Zero's manual **schema change hook** so DDL is detected explicitly instead of via event triggers:

1. Enable manual DDL detection once on the shard:

   ```sql
   UPDATE zero_0."shardConfig" SET "ddlDetection" = true;
   ```

2. After DDL statements (and before any dependent data changes), notify Zero in the same transaction:

   ```sql
   SELECT zero_0.update_schemas();
   ```

   Zero's docs show wrapping DDL + `update_schemas()` in a single transaction:

   ```sql
   BEGIN;
     ALTER TABLE foo ADD COLUMN bar TEXT;
     SELECT zero_0.update_schemas();
   COMMIT;
   ```

   For publication-only changes, `COMMENT ON PUBLICATION ...` also works, but `update_schemas()` is preferred because it catches all DDL, not just publication changes.

## How this fits our migration runner

`apps/dotcom/zero-cache/migrate.ts` already runs all pending migrations inside a **single** Kysely transaction, and migration files are forbidden from containing `BEGIN`/`COMMIT` (the runner wraps them). That makes the "DDL + `update_schemas()` in one transaction" requirement easy to satisfy — we just need to call `update_schemas()` at the right point.

Open design questions are below.

## Risks / unknowns

- Confirm the exact shard schema name. We don't set `ZERO_APP_ID`, so this is likely `zero_0`, but the real metadata schema (`<appID>_<shardNum>`) should be verified against the deployed shard before hardcoding.
- `ddlDetection = true` is a behavior switch — once enabled, Zero stops relying on event triggers, so a missed `update_schemas()` call after a future DDL migration would silently leave the replica schema stale. We need a convention (and ideally a guard) so every DDL migration calls it.
- Needs validation against a realistic dataset to confirm it avoids the full reset, and that timing (DDL before data changes) holds for our migration ordering.

## Unresolved questions

- Call `update_schemas()` once at end of each migration run, or per-migration after DDL?
- Hardcode `zero_0` or derive shard schema from config/env?
- Backfill the `ddlDetection` flag via a migration, or one-time manual op on each environment?
- Need a lint/guard so future DDL migrations don't forget the hook?

MitjaBezensek:
In order to stop Zero from doing a full reset of its server replica and every connected client on every Supabase schema change, this PR adopts Zero's manual schema-change hook.

tldraw.com runs Zero with upstream Postgres on Supabase. Zero detects schema changes via Postgres event triggers, but Supabase doesn't fire them for `ALTER PUBLICATION`. An undetected schema change makes Zero do a full reset of its server replica and every connected client — increasingly expensive as the DB grows.

- `migrate.ts` calls `zero_0.update_schemas()` once after all migrations apply, in the same transaction, so schema changes update the replica in place. Guarded so it's a no-op before Zero has booted (fresh DB). Uses `to_regprocedure` to resolve the no-arg overload (`zero_0` has two `update_schemas` overloads; `to_regproc` returns NULL for the ambiguous name and would silently skip the call).
- A drift check fails the migration loudly if the shard schema is ever not `zero_0`, so the hardcoded name can't silently stop matching.
- `032_enable_ddl_detection.sql` sets `ddlDetection = true` (guarded) as a safety-net; it's already `true` on staging.

Verified on staging: shard schema is `zero_0`, the no-arg overload exists, and `ddlDetection` is already `true`.

This migration is numbered `032` (the next free number on `main`). #9029 also currently uses `032` for `032_rename_group_admin_role_to_member.sql`; whichever PR lands second will need to bump its migration number to avoid the collision.

Closes #9031

### Change type

- [x] `other`

### Test plan

1. Deploy to staging and apply a migration that contains DDL.
2. Confirm Zero updates its replica in place — no full reset, clients don't re-download all state (check Zero logs).
3. Confirm `SELECT "ddlDetection" FROM zero_0."shardConfig";` is `true` and `update_schemas()` ran without error.


mimecuvalo (approved):
👍

PR #9034, opened 2026-06-05
chore(skills): rework /issue into a confidence-scored Q&A flow

steveruizok:
In order to make `/issue` useful for quick, low-context reports without sending downstream agents past missing intent, this PR creates issues immediately and then matures them through a confidence-scored follow-up flow.

The old flow front-loaded codebase research before creating the issue. The new flow inverts that: create the issue from the user's initial description, read back the problem in product terms, score how complete the captured intent is, and ask the questions that would fill the gaps — updating the issue in place as answers arrive.

### What changes

- **Create first, mature later.** The skill creates the issue right away so the user can track it, rather than researching the codebase to death first.
- **Standard body shape.** The body starts with the user's verbatim description, then a short product-facing **readback** of the problem, a numbered list of **Open questions**, and a plain-text **Confidence** status line (e.g. `Confidence: 84%, ready to get started.`).
- **Light investigation.** Codebase searching is kept to just enough to ground the confidence score and ask sharp questions. Anything answerable from the code should not become an open question, and implementation breadcrumbs (file paths, function names, fix recipes) stay out of the body unless the user asks for them.
- **Q&A loop.** As the user answers, the issue is updated via `gh issue edit`: answers are written beneath each question, the readback is revised, the confidence line is recomputed, and the issue framing (title, type, labels) can be adjusted if an answer changes what the issue is.
- **Critical vs. deferred questions.** Questions prefixed `Critical:` block readiness and cannot be skipped. Non-critical questions the user chooses not to answer are marked `_Deferred by user; not blocking implementation._` and do not block.
- **Downstream guardrails.** The `take` skill now treats unresolved `Critical:` questions, `_Awaiting answer._` placeholders, and the `More Info Needed` label as blockers before implementation.
- **Issue type helper.** `skills/issue/scripts/set-issue-type.sh` sets GitHub issue types via GraphQL because `gh issue create --type` is unreliable across versions.

### Change type

- [x] `other`

### Test plan

1. Read the `/issue` workflow and confirm follow-up answers retrigger the skill and update the issue body in place.
2. Read the `write-issue` standards and confirm the readback, open questions, and deferred-question conventions are documented.
3. Read the `take` workflow and confirm unresolved agent-drafted questions block implementation.
4. Run `git diff --check`.

### Code changes

| Config/tooling | +110 / -23 |

### DIFF
--- a/skills/issue/SKILL.md
+++ b/skills/issue/SKILL.md
@@ -5,17 +5,36 @@ description: Create and research a GitHub issue in the tldraw repository from a 
 
 # Issue
 
-Create a GitHub issue on `tldraw/tldraw` from a user description, then research it.
+Create a GitHub issue on `tldraw/tldraw` from a user description, then interrogate the user to capture enough of their intent for the issue to be worked on.
 
 Use `../write-issue/SKILL.md` as the standards reference for issue titles, bodies, types, labels, and triage conventions.
 
-At the top of the issue, include the user's original description verbatim. Annotate it with the human emoji, separated from your work by a horizontal rule. Example:
+The goal is not to research the codebase to death. It is to capture the user's full intent and context in the issue, so that whoever (or whatever) picks it up later has what they need. You do this by creating the issue immediately, scoring your confidence in it, and then asking the user the questions that would raise that confidence.
+
+## Issue body shape
+
+The body always starts with the user's original description, verbatim, annotated with the human emoji and separated from the rest by a horizontal rule:
 
 ```
 🧑: {user_description}
 
 ---
+
+## Confidence: {n}%
+
+{one or two sentences on what is still missing or uncertain about the user's intent}
+
+## Open questions
+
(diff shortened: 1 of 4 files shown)


mimecuvalo:
omg

frolic:
> omg

Cannot determine sentiment. 🤖

frolic:
@steveruizok Saw you added some confidence notation, curious what you think of this: https://github.com/tldraw/tldraw/pull/9038

steveruizok:
First try here: 
image

steveruizok:
And after follow up questions: 
image

steveruizok:
Ok, I've taken this for a few turns and am happy with it. It would be good to have a bot that updates the PR in response to replies or something, so that we can continue to work on the issue without firing up claude again to pick up the conversation.

steveruizok:
Ok, final form achieved. This is good and light.

image

PR #9038, opened 2026-06-05
chore: reduce /issue slop

frolic:
In order to stop the `issue` skill from filling underspecified one-liner issues with confident-but-wrong inferences, this PR adds a confidence gate to the issue-creation workflow. The original text is always preserved verbatim; the agent only expands an issue into a single problem statement and solution when it's ≥95% confident of the intent. Below that bar it captures open questions and lays out the possible approaches with their tradeoffs, so the original creator can come back and specify — rather than sending a downstream implementer with less context down the wrong path.

It also splits ownership cleanly between the paired skills, so the same guidance no longer lives in two places:

- `issue` (workflow) owns the *judgment* — when an issue is underspecified and how confident to be before expanding it — and structures its research comment so verified codebase facts are distinguishable from inferred intent.
- `write-issue` (standards) owns only the *body format* — the open-questions / possible-approaches sections and the `More Info Needed` label.
- `take` now detects an underspecified issue (the label, an open-questions section, vague text) and surfaces those questions to confirm intent before implementing, instead of guessing past them.

### Change type

- [x] `improvement`

### Test plan

These are agent skill instructions (markdown), so there is nothing to execute. Verified by reading:

- `issue` no longer commits to a single interpretation below the confidence bar; it records open questions plus approaches and adds `More Info Needed`.
- `take` pauses on an underspecified issue and confirms intent before implementing.
- No skill refers to itself; cross-references stay skill → other-skill, matching the existing `pr` / `write-pr` pattern.

### Code changes

| Config/tooling | +39 / -9   |

### DIFF
--- a/skills/issue/SKILL.md
+++ b/skills/issue/SKILL.md
@@ -17,6 +17,22 @@ At the top of the issue, include the user's original description verbatim. Annot
 ---
 ```
 
+## Underspecified issues
+
+Many issues arrive as a one-liner, and the original text may be everything the creator has in their head. Your job is to add useful context, not to invent a specification they never gave. A wrong inference is worse than an open question: it sends an implementer with less context down the wrong path while reading like settled scope.
+
+Throughout investigation and research, keep two things separate:
+
+- **Codebase facts** — a file does X, this code path runs on Y. You confirm these by reading; state them plainly.
+- **Intent and scope** — what the creator wants, which behavior is correct, what the fix should be. This is inference; never present it as fact.
+
+Gate how far you expand on your confidence in the intent:
+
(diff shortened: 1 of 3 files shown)


steveruizok:
I like the incorporation of /take here. There's definitely a use for having something thrown up for me to respond to, i.e. a lazy "take this issue and try to come up with a solution, any solution, as a starting point for me to think about it". But it would be good to have the bot surface missing context before it begins, e.g. "hey there's missing context here, you might want to ping Kevin about x y z before we begin". It really depends on how good or bad these follow up questions are. If they're shitty or universal, like "How important should we consider end to end testing on this change?" then I wouldn't want that to be blocking missed context.

frolic:
closing in favor of https://github.com/tldraw/tldraw/pull/9034

PR #9042, opened 2026-06-05
fix(dotcom): require source read access when creating a file from another file

jsscclr:


packages/dotcom-shared/src/mutators.ts
```diff
@@ -209,6 +209,28 @@ async function assertUserCanUpdateFile(tx: Transaction<TlaSchema>, userId: strin
...
+ * This was previously the `file.insertWithFileState` mutator. It was demoted to
+ * a private helper so it can only be reached through `createFile`, which gates
+ * the `createSource` for read access. As a public mutator it let a client
+ * insert an arbitrary file row — including a `createSource` pointing at another
+ * user's file — which would copy that file's content into the new room,
+ * bypassing the `createFile` gate.
+ */
+async function insertFileWithState(tx: Tx, userId: string, file: TlaFile, fileState: TlaFileState) {
+	// User must be the owner for legacy file creation
+	assert(file.ownerId === userId, ZErrorCode.forbidden)
```
MitjaBezensek: Think we allowed duplication of shared files (so the ones you didn't own as long as they were shared). But this is legacy so either way is fine.

MitjaBezensek (approved):

PR #9043, opened 2026-06-05
improvement(tldraw): add tooltip descriptions for preference menu items

### LINKED ISSUE #9009: Add tooltip descriptions for preference menu items
🧑: some user preference names are a bit unclear in isolation and as a new user ("dynamic sizing", "select on wrap"), would be nice to have a hover tooltip description of each

---

## Problem

The preferences submenu (main menu → Preferences) lists several toggles whose labels are hard to understand in isolation, especially for new users:

- **Select on wrap** — does this select shapes when you wrap a marquee around them? When text wraps?
- **Dynamic size** — dynamic relative to what?
- Others like **Edge scrolling**, **Tool lock**, **Always snap**, and **Paste at cursor** are also non-obvious without prior context.

Today each item only has a native `title` attribute that repeats the label text, so hovering provides no extra information.

## Proposed solution

Add a short description for each preference toggle, surfaced as a hover tooltip (and ideally exposed to assistive tech). For example:

- **Select on wrap** — "Select shapes fully enclosed by a selection box, instead of any shape it touches."
- **Dynamic size** — "Keep stroke and font sizes consistent on screen as you zoom."

Concretely:

1. Add optional description translation keys (e.g. `action.toggle-wrap-mode.description`) alongside the existing `.menu` labels in the default translation.
2. Thread an optional `description`/`title` through the menu checkbox primitives so the tooltip shows the description rather than echoing the label.
3. Populate descriptions for the preference toggles.

## Use cases

- New users discovering the SDK's default UI can understand what each preference does without leaving the menu.
- Improves clarity and accessibility of the default UI surfaced by `tldraw`.

## Notes

This affects the default UI in `packages/tldraw`. The preference toggles are defined in `DefaultMainMenuContent.tsx` and rendered via the `TldrawUiMenuCheckboxItem` primitive; labels live in `defaultTranslation.ts`.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9009 

https://github.com/user-attachments/assets/ca2688c0-edb9-4c4a-b7bd-1bc114a9bb70

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Hover preferences 
2. See tooltips?

Check with different screen sizes

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Added tooltip with descriptions for preference menu items

### Release notes

- Added tooltips to preference menu items


frolic:
niiice! any concerns with the length of that tooltip? should we max-width it? and/or put it under the menu item?

kaneel:
> niiice! any concerns with the length of that tooltip? should we max-width it? and/or put it under the menu item?

Hmmm, good questions. Maybe the tooltip will look better with a max width. As with placing under the menu item, I don't know radix much but I'd be afraid it would keep the tooltip visible by occluding the next item.

Scratch that, it's probably best under it and adapt much better with smaller screen… but also these descriptions are too long I'll need someone to jump in and figure good ones.

image

frolic:
oh yeah, good call on not hiding the next menu item! but yeah, prob need to set a max width so they don't run too far horizontal and instead break into multiple lines (I think the description length is ~fine)

kaneel:
> oh yeah, good call on not hiding the next menu item! but yeah, prob need to set a max width so they don't run too far horizontal and instead break into multiple lines (I think the description length is ~fine)

Not looking too bad 🤔 
image

mimecuvalo:
mm, under is tricky b/c there's a race condition there - you can accidentally hover over the tooltip and then the interaction becomes a bit finicky.

i worry that tooltips for menu items aren't really a design convention that i've seen in the wild — is there an example of somebody doing this well? alternate way of thinking about this is just making the wording on the preferences less terse, example for Figma:
Screenshot 2026-06-08 at 16 52 59

kaneel:
@mimecuvalo that's what I also thought at first but appears it works well and we're already doing this in the palette box (for fills) but what could be good enough for this would be to add an (i) log next to these complicated items as well 🤔

mimecuvalo:
hm, ok, maybe — i'm willing to be convinced but i'm not sold on this yet. i hear you as it almost working with being below. but i'm going to push back a bit again:
- are there menus in other apps that have tooltips like this combined with menus? it doesn't seem like a real convention around this vs. just having a less terse description?
- "log next to these complicated items as well" - can you expand on this? not sure what you meant there
- in any case the tooltip gets "stuck" sometimes when in the menu, but doesn't get stuck in the style panel, see video:
Kapture 2026-06-09 at 15 51 35

frolic:
I wouldn't mind just a better `title` attribute as a first-pass here

kaneel:
> hm, ok, maybe — i'm willing to be convinced but i'm not sold on this yet. i hear you as it almost working with being below. but i'm going to push back a bit again:
> 
>     * are there menus in other apps that have tooltips like this combined with menus? it doesn't seem like a real convention around this vs. just having a less terse description?
> 
>     * "log next to these complicated items as well" - can you expand on this? not sure what you meant there
> 
>     * in any case the tooltip gets "stuck" sometimes when in the menu, but doesn't get stuck in the style panel, see video:
> 
> 
> Kapture 2026-06-09 at 15 51 35

sorry, "an (i) icon… next to the item", the tooltip would appear only when hovering the icon

steveruizok:
> I wouldn't mind just a better `title` attribute as a first-pass here

So I think we're going to go with no titles, more descriptive text.

mimecuvalo:
rm'd titles on https://github.com/tldraw/tldraw/pull/9315 — longer descriptions for a separate PR by someone

kaneel:
> longer descriptions for a separate PR by someone

Should we sit down and think about the copy?

kaneel:
Btw I'm closing this

mimecuvalo:
> > longer descriptions for a separate PR by someone
> 
> Should we sit down and think about the copy?

i'm happy for anybody to run with this! we can sit down after there's a PR as a starting point methinks

PR #9049, opened 2026-06-05
fix(dotcom): clean up guest file_state rows when unsharing group-owned files

### LINKED ISSUE #9037: Unsharing a group-owned file does not clean up guest file_state rows
🧑: Bug: unsharing a group-owned file does not clean up guest file_state rows (delete_file_states trigger keys on ownerId which is NULL for group-owned files)

---

## Summary

When the owner of a **group-owned** file (groups backend, `file.owningGroupId` set, `file.ownerId` NULL) sets sharing to "no access" (`shared` true→false), guests' `file_state` rows are **not** deleted. The cleanup trigger `delete_file_states` keys on `ownerId`, which is NULL under the groups model, so its predicate never matches and zero rows are deleted.

The equivalent legacy (single-owner, `ownerId` set) path works correctly — this gap is specific to group-owned files.

## Where it is

`delete_file_states` in `apps/dotcom/zero-cache/migrations/000_seed.sql:59-72`:

```sql
CREATE OR REPLACE FUNCTION delete_file_states() RETURNS TRIGGER AS $$
BEGIN
  IF OLD.shared = TRUE AND NEW.shared = FALSE  THEN
    DELETE FROM file_state WHERE "fileId" = OLD.id AND OLD."ownerId" != "userId";
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER file_shared_update
AFTER UPDATE OF shared ON file
FOR EACH ROW
EXECUTE FUNCTION delete_file_states();
```

This is the only unshare-cleanup path. It is never `CREATE OR REPLACE`'d, dropped, or altered in any later migration (verified across all migrations `000`–`031`).

## Root cause

`apps/dotcom/zero-cache/migrations/023_groups.sql:96` enforces an XOR invariant: a file has either `ownerId` or `owningGroupId`, never both. So a group-owned file has `ownerId = NULL`.

The trigger predicate `OLD."ownerId" != "userId"` then evaluates as `NULL != '<userId>'` → `NULL`, which is falsy in a `WHERE` clause. **No rows are deleted.**

## Why the delete/group paths are fine (for contrast)

The other state-change triggers key on `fileId`/`owningGroupId`, not `ownerId`, so they correctly cover group-owned files:

| Event | Trigger | Covers group files? |
| --- | --- | --- |
| File deleted (`isDeleted`→true) | `cleanup_deleted_file` — `WHERE "fileId" = NEW."id"` (`023_groups.sql:154`) | ✅ |
| Group deleted | `cleanup_deleted_group` → marks files deleted → cascades to above (`023_groups.sql:186`) | ✅ |
| User leaves a group | `cleanup_group_user_file_states` — deletes leaver's states for `owningGroupId` files where `shared=false` (`023_groups.sql:214`) | ✅ (different event) |
| **File unshared (`shared` true→false)** | `delete_file_states` (`000_seed.sql:62`) | ❌ **this issue** |

## Impact

This is a correctness/staleness gap, **not** an edit-access or room-access hole:

1. **The file row keeps syncing to the ex-guest.** The `fileStates` synced query is `file_state.where(userId).related('file')` (`packages/dotcom-shared/src/queries.ts:27`) with no permission gate on the related file. While the stale `file_state` exists, the now-private file row — including its `name` — continues streaming to the ex-guest's client. The legacy path does not have this staleness.
2. **Editing/opening is still blocked.** `assertUserCanAccessFileInternal` grants guest access only `if (allowGuestAccess && file.shared)` (`packages/dotcom-shared/src/mutators.ts`), so once `shared=false` the guest cannot mutate or open the room.
3. **Visibility in recents depends on the viewer's migration state:**
   - **Non-migrated guest:** `getMyFiles` iterates `file_state`s (`apps/dotcom/client/src/tla/app/TldrawApp.ts:561`); the stale entry passes all filters (file exists, not `isDeleted`, no matching group membership) → **the file visibly lingers in recents** with a "Guest" badge.
   - **Migrated guest:** `getMyFiles` short-circuits to `getGroupFilesSorted(homeGroupId)` (`TldrawApp.ts:546-548`), which never includes another group's shared file → not shown in recents, but the dangling row + file-row sync persist.

## Steps to reproduce

1. As a migrated user, create a file in your group and enable sharing (link access).
2. As a different user (guest), visit the shared link — this creates a guest `file_state` row.
3. As the owner, set the file to "no access" (`setFileSharedLinkType(fileId, 'no-access')` → `file.update({ shared: false })`).
4. Inspect the DB: the guest's `file_state` row still exists. The file row (with `name`) continues to sync to the guest's client; for a non-migrated guest it also still appears in their recents list.

**Expected:** the guest's `file_state` is removed on unshare (as it is for legacy `ownerId`-owned files), the file row stops syncing to them, and it disappears from their recents.

## Suggested fix

Make `delete_file_states` ownership-model-agnostic — NULL-safe owner check plus a group-membership check so group members keep their states:

```sql
DELETE FROM file_state fs
WHERE fs."fileId" = OLD.id
  AND OLD."ownerId" IS DISTINCT FROM fs."userId"          -- NULL-safe owner check
  AND NOT EXISTS (
    SELECT 1 FROM group_user gu
    WHERE gu."groupId" = OLD."owningGroupId" AND gu."userId" = fs."userId"
  );
```

`IS DISTINCT FROM` fixes the immediate NULL bug; the `NOT EXISTS` clause keeps current group members' file states intact under the groups model. Ship as a new migration that `CREATE OR REPLACE`s the function, with a focused test covering both legacy and group-owned unshare.

jsscclr:
In order to stop a file lingering in an ex-guest's recents (and continuing to sync to them) after its owner revokes access, this PR fixes the `delete_file_states` trigger so it cleans up both guest records when a **group-owned** file is unshared. Fixes #9037.

Two records make a shared file visible to a guest, and neither was cleaned up on unshare:

- **`file_state`** (sync/visit state). The trigger (from `000_seed.sql`) decides which rows belong to "guests" with `OLD."ownerId" != "userId"`. The groups backend (`023_groups.sql`) introduced group-owned files where `ownerId` is `NULL` (per the `ownerId` XOR `owningGroupId` constraint). For those files `NULL != "userId"` evaluates to `NULL`, so the `DELETE` matched nothing and guest `file_state` rows survived — leaving the now-private file syncing to ex-guests. Legacy (user-owned) files were unaffected, which is why this only bites group files.
- **`group_file`** (the file link created when a visitor opens a shared file, in their home group — home group id == user id). For groups-migrated users the sidebar file list is built from these links (`TldrawApp.getMyFiles`), so even with `file_state` cleaned up the revoked file's name kept showing in an ex-guest's recent files. Nothing deleted these links on unshare for any ownership model.

Migration `034` replaces the function so it cleans up both, for both ownership models:

- `file_state`: delete states for anyone who isn't the legacy owner (`IS DISTINCT FROM`, so a `NULL` `ownerId` no longer accidentally protects a guest) or a current member of the owning group (`NOT EXISTS` against `group_user`, a no-op for legacy files);
- `group_file`: delete the file's links except the owning group's own row, the legacy owner's home-group link, and links of current owning-group members — the same access rule, with the same NULL-safety.

The trigger binding (`file_shared_update`) is unchanged. Note: viewing access itself is enforced at read time (and duplication needs source read access since #9042) — this PR is about cleaning up the stale per-user records that kept revoked files syncing and listed.

> [!NOTE]
> This migration started life as `033` and was amended in place, then renamed to `034` after #9029's `033_rename_group_admin_role_to_member` merged. The persistent preview DB has the stale `033_fix_unshare_group_file_cleanup` in its ledger, so the `reset-preview-db` label flow needs to run before re-testing on the preview deploy.

### Change type

- [x] `bugfix`

### Test plan

This bug is only visible on **group-owned** files, so you need the groups flag and two accounts (an owner and a guest). The cleanest way to see the fix is to run the identical flow on current tldraw.com (the bug) and on this PR's preview deploy (the fix), side by side.

**Setup (both accounts)**

1. Open the preview deploy at https://pr-9049-preview-deploy.tldraw.com/ and sign in.
2. Turn on the groups flag for each account: go to https://pr-9049-preview-deploy.tldraw.com/admin, search for the email, and enrol it in groups.

**Current tldraw.com behaviour (the bug)**

Run this flow on production https://tldraw.com/ (where the fix isn't deployed):

1. As the owner, create a group, create a file **in that group**, and enable sharing.
2. As the guest, open the share link — the file appears in their recent files (creates a guest `file_state` and a `group_file` link).
3. As the owner, set sharing to "no access" (`shared = false`).
4. ❌ The file **stays** in the guest's recent files and keeps syncing to them (its name is still visible). This is the regression.

**With this PR (the fix)**

Repeat the exact same flow on https://pr-9049-preview-deploy.tldraw.com/:

1. Owner creates a group, creates a file in it, enables sharing.
2. Guest opens the share link — file appears in their recent files.
3. Owner sets sharing to "no access".
4. ✅ Both guest records are removed: the file stops syncing to the guest **and** disappears from their recent files. Group members and the owner keep their access and their recents entries.

(Sanity check, optional: a **legacy** personal file that's unshared already cleaned up its `file_state` before this PR, but its `group_file` links from migrated guests also linger — so unsharing a personal shared file now also clears the guest's recents entry.)

Verified end to end on the local dev stack (two accounts, share → visit → unshare): the guest's `file_state` and `group_file` rows are deleted by the trigger, the guest gets "Invite only" on the file URL, and the file disappears from their recent files; the owner keeps access and their recents entry.

- [x] Unit tests

`apps/dotcom/zero-cache/delete_file_states.test.ts` is a focused postgres integration test that loads the real shipped function body from migration `034` (rather than a hand-copied duplicate) and covers group-owned unshare, legacy unshare, and a still-shared control — for both `file_state` and `group_file` cleanup. It also runs the original buggy function body as a regression guard, proving the old function left the group guest's state and link behind and the new one removes them. The trigger is `plpgsql`, so it needs a real postgres — the suite is opt-in via `ZERO_CACHE_TEST_POSTGRES_URL` (a **direct** connection string; the suite is also hardened against transaction-pooled connections by schema-qualifying everything it owns and running verbatim SQL under `SET LOCAL search_path` in transactions) and skips cleanly (exit 0) when unset, so CI stays green. `zero-cache` had no vitest project, so this PR also adds `apps/dotcom/zero-cache/vitest.config.ts`. Verified locally: 7/7 pass against the dev postgres; skips with exit 0 without a DB; `yarn typecheck` passes.

### Release notes

- Fix tldraw.com files lingering in a guest's recent files (and continuing to sync to them) after the owner revokes access to a group-owned file.

### Code changes

| Tests          | +250 / -0  |
| Apps           | +56 / -0   |
| Config/tooling | +15 / -0   |


apps/dotcom/zero-cache/delete_file_states.test.ts
```diff
@@ -0,0 +1,179 @@
...
+// See migration 033_fix_unshare_group_file_cleanup.sql.
+//
+// This talks to a real postgres (the trigger is plpgsql, so SQLite/PGlite-free
+// fakes can't exercise it). It is opt-in: set ZERO_CACHE_TEST_POSTGRES_URL (or rely
+// on BOTCOM_POSTGRES_POOLED_CONNECTION_STRING from a running `yarn dev` stack) to
+// run it. Without a connection string the suite is skipped so CI stays green.
+
+const DIRNAME = dirname(fileURLToPath(import.meta.url))
+const CONNECTION_STRING =
+	process.env.ZERO_CACHE_TEST_POSTGRES_URL ?? process.env.BOTCOM_POSTGRES_POOLED_CONNECTION_STRING
```
MitjaBezensek: Maybe we should always qualify the queries instead of setting them in a `beforeAll`. After exploring this with claude it seems like `SET search_path TO "${schemaName}` might not stick between the queries, which means we'd start executing queries against `public`.

> We should drop the `BOTCOM_POSTGRES_POOLED_CONNECTION_STRING` fallback. We checked this with Claude: on a transaction-mode pooled connection (which is what that var points to locally — pgbouncer on 6432), `SET search_path TO "${schemaName}"` doesn't persist between queries (verified empirically — it was lost on 29/30 statements even with no other load). The unqualified statements then run against `public` — and since the setup includes `DROP TABLE ... CASCADE` and `CREATE OR REPLACE FUNCTION delete_file_states()`, pointing this at a real database would drop the real tables and overwrite the real trigger function, while the tests still pass.
Requiring an explicit `ZERO_CACHE_TEST_POSTGRES_URL` helps, but even a direct connection can silently reconnect (pg.Pool reaps idle clients after 10s) and reset `search_path`. So we should also make the SQL safe by construction: schema-qualify the table names (or `SET LOCAL search_path` inside a transaction), and/or assert `current_schema()` is the test schema before running any DDL.

MitjaBezensek (approved):

PR #9050, opened 2026-06-05
build: make CommonJS builds safe against ESM-only deps, upgrade rbush to v4

frolic:
In order to safely consume modern ESM-only leaf dependencies without breaking CommonJS consumers, this PR adds selective CJS bundling to the package build, uses it to upgrade `rbush` from v3 to v4 in `@tldraw/editor`, adds a CI smoke test that catches ESM-only dependency leaks in the published CJS builds, and fixes every leak it found.

tldraw dual-publishes CJS + ESM, and the package build (`internal/scripts/build-package.ts`) runs esbuild with `bundle: false`, so dependencies stay external as `require("x")` in the CJS output. When `x` is ESM-only, that leaks a `require(<esm>)` into the published CJS build, which throws `ERR_REQUIRE_ESM` on Node 20.0–20.18 (the repo floor is `^20.0.0`; `require(esm)` only landed in 20.19 / 22.12) and breaks Jest and ts-node. This already bit us once: [#7905](https://github.com/tldraw/tldraw/pull/7905) downgraded `rbush` v4 → v3 because v4 went ESM-only. This PR makes v4 safe for CJS, effectively reversing that downgrade.

## How it works

A narrow allowlist (`ESM_ONLY` — `rbush`, `jittered-fractional-indexing`, `nanoevents`) controls which ESM-only deps get inlined into the CommonJS build.

- **Asymmetric externals.** The ESM build is unchanged (`bundle: false`), so allowlisted deps stay external and ESM consumers resolve them natively, keeping dedup and tree-shaking. Only the CJS build inlines them.
- **Gated per package.** `buildPackage` computes which allowlisted deps are in the package's own `dependencies`. Bundling turns on only for the CJS pass, and only when that list is non-empty. Every other package keeps the exact `bundle: false` path — verified that an unaffected package's `dist-cjs` is byte-for-byte identical before and after this change.
- **The esbuild plugin.** When bundling the CJS pass, an `onResolve` plugin leaves entry points to esbuild, inlines allowlisted bare specifiers and everything reachable from them (any importer inside `node_modules` is bundled), keeps relative imports external so the per-file source mirror is preserved instead of collapsed into each entry, and keeps every other bare dep external.
- **Transitive bundling.** An inlined dep's own deps are inlined too: `rbush` pulls in `quickselect` and `jittered-fractional-indexing` pulls in `fractional-indexing` — both also ESM-only. Inlining only the top dep while leaving `require("quickselect")`/`require("fractional-indexing")` external would just move the bug down a level, so the plugin bundles the whole reachable subtree. License notices for the inlined deps are preserved via esbuild `legalComments: 'linked'` in `dist-cjs`.

## CJS smoke test (CI)

`yarn check-cjs` (`internal/scripts/check-cjs.ts`) runs in the CI build job after `build-package`. It scans every `packages/*/dist-cjs` build for the external deps it `require()`s and strict-requires each under `node --no-experimental-require-module`, which reinstates the `ERR_REQUIRE_ESM` throw even on modern Node (where `require(esm)` is otherwise allowed by default). It fails on any ESM-only dep that leaks into a CJS build — the general version of #7905's smoke check, with no hardcoded dep list.

The check found four leaks (rbush plus three latent, pre-existing ones). **All are fixed in this PR**, so the check guards a fully clean tree:

| Package | ESM-only dep | Fix |
| --- | --- | --- |
| `@tldraw/editor` | `rbush@4` (+ `quickselect`) | inlined (`ESM_ONLY`) |
| `@tldraw/utils` | `jittered-fractional-indexing` (+ `fractional-indexing`) | inlined (`ESM_ONLY`) |
| `@tldraw/sync-core` | `nanoevents` | inlined (`ESM_ONLY`) |
| `@tldraw/mermaid` | `mermaid@11` | dynamic `import()` (see below) |

## Inline vs dynamic import

The three small, sync-used deps (`rbush`, `jittered-fractional-indexing`, `nanoevents`) are **inlined** — they're used synchronously (rbush is even a base class), so they must be present at module-eval time, and their trees are tiny pure JS that bundle cleanly under `platform: 'neutral'`.

`mermaid` is **not** inlined. Inlining it fails outright (78 esbuild errors — its tree, e.g. chevrotain/lodash-es, can't resolve under `platform: 'neutral'`), and it's huge and lazily used. Instead `createMermaidDiagram()` (already `async`) loads it with a dynamic `import('mermaid')`. `import()` works from CommonJS, and esbuild keeps it as a real dynamic import (it doesn't downlevel to `require`), so mermaid stays external — no multi-MB blob, dedup preserved — and only loads when a diagram is actually created (so `require('@tldraw/mermaid')` no longer pulls mermaid in at all).

## Follow-up before allowlisting more deps

Each dep inlined here is a single import site, so adding it to `ESM_ONLY` was safe directly. For any **future** dep imported across many files, we should first route it through a single "vendor shim" import site and guard it with a lint rule — the same pattern as the existing `no-tiptap-default-import` rule — so an ESM-only dep can't be imported somewhere that bypasses the CJS inlining.

### Change type

- [x] `improvement`

### Test plan

1. `yarn build-package`.
2. `yarn check-cjs` passes (no ESM-only `require` leaks in any `dist-cjs`), and correctly fails if a new ESM-only leak is introduced.
3. Allowlisted deps stay external in ESM: e.g. `rbush` appears as `import RBush from "rbush"` in `packages/editor/dist-esm`, not inlined.
4. CJS load smoke test (mirrors #7905): requiring the built `RBushIndex` from a CommonJS context constructs and queries the index with no `ERR_REQUIRE_ESM`.
5. `@tldraw/mermaid`: `dist-cjs/createMermaidDiagram.js` contains `await import("mermaid")` and no `require("mermaid")`; the 53 mermaid tests pass.
6. Regression: an unaffected package's `dist-cjs` output is byte-for-byte identical before and after the build-script change.
7. `yarn typecheck`, `yarn api-check`, `yarn lint-current`, and the spatial index tests (`packages/tldraw/src/test/spatialIndex.test.ts`, `getCulledShapes.test.tsx`) all pass.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Core code      | +6 / -1    |
| Config/tooling | +308 / -19 |

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -98,3 +98,8 @@ jobs:
         # the sed pipe makes sure that github annotations come through without
         # turbo's prefix
         run: "yarn build-package | sed -E 's/^.*? ::/::/'"
+
+      - name: Smoke test CJS output
+        # fail if a published dist-cjs build require()s an ESM-only dependency
+        # that isn't inlined (would throw ERR_REQUIRE_ESM for CJS consumers)
+        run: yarn check-cjs
--- a/internal/scripts/build-package.ts
+++ b/internal/scripts/build-package.ts
@@ -1,12 +1,23 @@
 import { copyFileSync, existsSync } from 'fs'
 import path from 'path'
 import { pathToFileURL } from 'url'
-import { build } from 'esbuild'
+import { build, type BuildOptions, type Plugin } from 'esbuild'
 import { glob } from 'glob'
 import kleur from 'kleur'
 import rimraf from 'rimraf'
 import { addJsExtensions } from './lib/add-extensions'
 import { readJsonIfExists } from './lib/file'
+
+/**
+ * Allowlist of ESM-only dependencies to inline into the CommonJS build output.
+ * We dual-publish CJS + ESM with `bundle: false`, so a dependency normally stays
+ * external as a `require("x")` in the CJS output. That breaks CJS consumers when
+ * `x` is ESM-only (`require` of an ES module throws on Node <20.19, and trips up
+ * Jest/ts-node). Bundling these deps into the CJS output keeps the CJS path
+ * working. The ESM build still leaves them external so ESM consumers resolve
+ * them natively and keep dedup/tree-shaking.
+ */
+const ESM_ONLY = new Set(['rbush', 'jittered-fractional-indexing', 'nanoevents'])
 
 interface LibraryInfo {
 	name: string
@@ -44,10 +55,16 @@ async function buildPackage({ sourcePackageDir }: { sourcePackageDir: string }) 
 	const packageVersion = packageJson.version
 	if (typeof packageVersion !== 'string') throw new Error('package.json version is not a string')
 
+	// allowlisted ESM-only deps that this package actually depends on. Only these
+	// get inlined into the CJS build; every other package keeps the plain
+	// `bundle: false` path with byte-for-byte identical output.
+	const inlineDeps = Object.keys(packageJson.dependencies ?? {}).filter((dep) => ESM_ONLY.has(dep))
+
 	// build CommonJS files to /dist-cjs
 	await buildLibrary({
 		sourceFiles,
 		sourcePackageDir,
+		inlineDeps,
 		info: {
 			name: packageName,
 			version: packageVersion,
@@ -59,6 +76,7 @@ async function buildPackage({ sourcePackageDir }: { sourcePackageDir: string }) 
 	await buildLibrary({
 		sourceFiles,
 		sourcePackageDir,
+		inlineDeps,
 		info: {
 			name: packageName,
 			version: packageVersion,
@@ -82,10 +100,12 @@ async function buildLibrary({
 	sourceFiles,
 	sourcePackageDir,
 	info,
+	inlineDeps,
 }: {
 	sourceFiles: string[]
 	sourcePackageDir: string
 	info: LibraryInfo
+	inlineDeps: string[]
 }) {
 	const dirName = `dist-${info.moduleSystem}`
 	const outdir = path.join(sourcePackageDir, dirName)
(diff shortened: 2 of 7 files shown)


frolic:
@mimecuvalo This would be a prerequisite to moving to `p-debounce` since it's ESM-only and uses a past dep upgrade (rbush -> v4) as an example. Not sure our story around CJS, but an alternative is to ditch CJS support (which I don't think we want to do yet?).

frolic:
Interesting that Claude uncovered some other ESM-only deps that may not have surfaced as a bug report yet like we saw in #7905. Worth deciding if we want to continue with CJS support (original bug report noted its fine if so, just wasn't clear we dropped CJS support).

image

mimecuvalo:
ah commented on the wrong PR but meant to post this here, to keep in mind stats-wise: https://github.com/wooorm/npm-esm-vs-cjs

my esm-vs-cjs-fu isn't the strongest though — i'm open to discussions / opinions here though

frolic:
To clarify, this preserves the CJS build while allowing us to use ESM-only dependencies by bundling them in sync contexts, dynamic importing in async contexts

mimecuvalo:
> To clarify, this preserves the CJS build while allowing us to use ESM-only dependencies by bundling them in sync contexts, dynamic importing in async contexts

ah ok, i misread / quickly glanced. lemme take a look more deeply tomorrow.

(another thing to consider here, in a separate PR, is that Node 20 just hit EOL a month ago or so, so related to this change, we could also explicitly make some node engine specifications here to tighten this up. looks like Node >= 20.19 (backport) and >= 22.12 help in a different way)

frolic:
> (another thing to consider here, in a separate PR, is that Node 20 just hit EOL a month ago or so, so related to this change, we could also explicitly make some node engine specifications here to tighten this up. looks like Node >= 20.19 (backport) and >= 22.12 help in a different way)

oh I much rather just bump the min node versions than this approach! nice! PR up: https://github.com/tldraw/tldraw/pull/9098

.github/workflows/checks.yml
```diff
@@ -98,3 +98,8 @@ jobs:
         # the sed pipe makes sure that github annotations come through without
         # turbo's prefix
         run: "yarn build-package | sed -E 's/^.*? ::/::/'"
+
+      - name: Smoke test CJS output
+        # fail if a published dist-cjs build require()s an ESM-only dependency
+        # that isn't inlined (would throw ERR_REQUIRE_ESM for CJS consumers)
+        run: yarn check-cjs
```
mimecuvalo, line 105: how fast is this? i'd be wary of adding more time to our CI for this, esp if deps haven't changed.

frolic:
closing in favor of node bump + async mermaid import

PR #9063, opened 2026-06-08
feat(dotcom): admin action to enroll a user in groups

### LINKED ISSUE #9062: Add admin action to enroll a single user in groups
## Problem

There's no usable admin control to enroll a *specific* user into the groups *experience*. The `/admin` page has a single-user "Migrate to Groups" control, but:

- it's hidden unless you've searched for a user **and** that user lacks `groups_backend` — and since new users get `groups_backend` by default, it almost never shows;
- it only grants `groups_backend` (the data model), never `groups_frontend` (the flag that actually shows the groups UI).

`groups_frontend` is otherwise only granted when a user accepts a group invite, so an admin currently has no straightforward way to turn the groups experience on for a given user (e.g. a customer or a teammate on staging). The only always-visible groups control is the batch "Migrate all users to groups backend".

## Proposed solution

Add an always-visible (per found user) "Groups enrollment" control in the admin User Data section that:

- shows the user's current `groups_backend` / `groups_frontend` status, and
- has an "Enroll in groups" action that ensures **both** flags — migrating the data model if needed (idempotent) and granting the groups UI — then reboots the user so it takes effect.

Backed by a new admin endpoint (`POST /app/admin/user/enroll_groups`) behind the existing `@tldraw.com` admin gate. Idempotent / safe to re-run.

## Notes

- The existing batch migration and `POST /app/admin/user/migrate` endpoint are unaffected.
- Surfaced while testing on staging: "Migrate all users to groups backend" was the only visible groups control.

jsscclr:
In order to let an admin turn the groups experience on for a specific user, this PR adds an "Enroll in groups" action to the `/admin` page that grants both groups flags. Closes #9062.

The page previously only had a single-user "Migrate to Groups" control that was hidden unless you'd searched a user lacking `groups_backend`, and it only granted `groups_backend` — never `groups_frontend`, the flag that actually shows the groups UI (otherwise only granted on accepting a group invite). So there was no usable way to enroll a specific user into the groups experience.

This replaces that control with an always-visible (per found user) "Groups enrollment" block that shows the user's current `groups_backend` / `groups_frontend` status and an "Enroll in groups" button. The button calls a new admin endpoint `POST /app/admin/user/enroll_groups` (behind the existing `@tldraw.com` admin gate), which runs `admin_enrollInGroups`: migrate the data model if needed (idempotent `migrate_user_to_groups`) and add `groups_frontend`, then reboot the user so it takes effect. It reports what changed and is safe to re-run. The existing batch migration and `POST /app/admin/user/migrate` are unchanged.

### Change type

- [x] `improvement`

### Test plan

1. Sign in to a local/staging dotcom with an `@tldraw.com` account and open `/admin`.
2. Enter a user's email or id → **Find User**.
3. In **User Data**, the **Groups enrollment** block shows `groups_backend ✓/✗ · groups_frontend ✓/✗`.
4. For a user lacking `groups_frontend` (e.g. a freshly created account — new users get `groups_backend` by default but not `groups_frontend`), click **Enroll in groups** → confirm. The status flips to both ✓, and the groups UI appears for that user on their next load.
5. Re-running on an already-enrolled user reports "already fully enrolled" (idempotent).

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +101 / -21 |

### DIFF
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -716,7 +716,7 @@ function EnrollUserInGroups({
 
 			const result = await res.json()
 			const parts: string[] = []
-			if (result.backendMigrated) parts.push(`migrated data (files: ${result.files_migrated})`)
+			if (result.backendMigrated) parts.push('migrated to groups backend')
 			if (result.frontendGranted) parts.push('granted groups UI')
 			onSuccessMessage(
 				parts.length ? `Enrolled in groups: ${parts.join(', ')}` : 'User was already fully enrolled'
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -6,7 +6,7 @@ import {
 	userHasFlag,
 	ZStoreData,
 } from '@tldraw/dotcom-shared'
-import { RefObject, useCallback, useEffect, useRef, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { Navigate } from 'react-router-dom'
 import { fetch } from 'tldraw'
 import { sentryReleaseName } from '../../sentry-release-name'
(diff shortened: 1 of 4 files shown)


MitjaBezensek (approved):
👍

PR #9066, opened 2026-06-08
docs(releases): back-port 5.1 release notes and nav ordering to v5.1.x

angrycaptain19:
Moves the changes from #9052 into v.5.1.x so that we avoid the merge conflicts that have cropped up with the release notes automations since we released it. 

### Change type

- [x] `other`

### DIFF
--- a/apps/docs/content/getting-started/releases.mdx
+++ b/apps/docs/content/getting-started/releases.mdx
@@ -33,6 +33,7 @@ We ship an experimental `tldraw-migrate` agent skill to let an agent help you up
 
 ## v5.x
 
+- [v5.1](/releases/v5.1.0) - Page menu redesign, copy-styles shortcut, selectable locked shapes, public translation APIs, and performance improvements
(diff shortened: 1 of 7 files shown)


frolic:
ooc how are release notes generated?

angrycaptain19:
> ooc how are release notes generated?

Normally we have a release notes automation that goes in and updates things into the Next section and then when we release a minor we should have a skill that moves those from Next -> version. But that hasn't been working for a little bit. 

We didn't move things over when we released 5.1 but the next automation has run since then which leads to this weird issue with trying to use the docs hotfix please label on main leading to a merge conflict. 

Ideally we wouldn't be merging directly to the release branch but we also had to do this pre 5.0 and for 5.0

frolic:
but who writes the entries? do they get written as part of each user-facing PR? do we do a summary pass before release?

angrycaptain19:
> but who writes the entries? do they get written as part of each user-facing PR? do we do a summary pass before release?

Ah sorry! They're automatically generated via claude skill based on commist/PRs. Occasionally we update and write them manually

PR #9067, opened 2026-06-08
refactor(dotcom): express group permissions as named capabilities

frolic:
In order to make group roles easy to read and safe to adjust, this PR refactors the dotcom groups prototype so authorization is a single predicate — `can(role, 'deleteGroup')` — instead of branching on role literals (`role === 'owner'` / `role === 'admin'`) or checking group membership existence directly. Each role is defined as a list of named capabilities in one `roles` table, so what a role can do reads off one place, and roles can be added or changed without hunting down scattered checks.

This came out of renaming `admin` → `member`: it was hard to tell what that role actually grants. With capabilities, a role's powers are a list you can read, and the rename becomes a relabel rather than a behavioral question.

The change is behavior-preserving, with one intentional exception: the group-name input is now `disabled` for members without the `editGroup` capability (previously they could type into it and the mutation silently failed server-side).

### Concepts

| Term | Type | Meaning |
| --- | --- | --- |
| `capabilities` | `as const` array | The list of every capability, enumerable at runtime (`capabilities.ts`) |
| `Capability` | `(typeof capabilities)[number]` | A single thing a member can do — derived from the list |
| `roles` | table | The single source of truth: each role → its capability list (`roles.ts`) |
| `Role` | `keyof typeof roles` | The role string stored in `group_user.role` (`'admin' \| 'owner'`) — derived from the table keys |
| `can(role, capability)` | function | The one check. `role` is loose (`string \| null \| undefined`); unknown/null → `false` |
| `isRole(role)` | guard | Whether a string is a known role (used for validating client input) |
| `getRole(db, …)` | sync-worker helper | Resolves the role from the DB (mirrors the mutators' `getRole`); compose with `can` |

The role stays a plain string in the DB; capabilities are never persisted — they're derived from that string in code.

### Usage

```ts
// client — role comes straight from the synced membership
const role = currentUser?.role
can(role, 'editGroup')        // boolean, no wrapper, no narrowing

// mutators — resolve the role, then check it
const role = await getRole(tx, userId, id)
assert(can(role, 'deleteGroup'), ZErrorCode.forbidden)

// sync-worker (Kysely) — getRole resolves from the DB
const role = await getRole(db, userId, groupId)
if (can(role, 'accessFiles')) { /* allow */ }
```

### Where it's enforced

Every group authorization check now goes through `can`, so they all consult the same table:

- **Mutators** (`@tldraw/dotcom-shared`) — `getRole` + `can(role, …)` on each group mutation.
- **Sync-worker** — `getRole` + `can(role, 'accessFiles')` for file access in `getAuth` (`requireWriteAccessToFile`), `worker.ts`, and the three `TLFileDurableObject` socket/snapshot/prune paths that previously checked membership existence directly.
- **Client** — `TldrawApp.canUpdateFile` and `useHasFileAdminRights` use `can(role, 'accessFiles')` instead of bare membership checks.

The "a group must always keep at least one owner" rule (last owner can't be demoted or leave) is a group **invariant**, not a capability, so it stays as explicit, labeled logic — it's now the only place that compares against a role literal.

### Capability → role matrix

These map 1:1 onto current behavior — no role gains or loses anything.

| Capability | Operation | admin | owner |
| --- | --- | :---: | :---: |
| `accessFiles` | open/edit the group's files, reorder own membership | ✓ | ✓ |
| `addFiles` | link/move a file into the group, drag | ✓ | ✓ |
| `removeFiles` | remove a file from the group, move-out | ✓ | ✓ |
| `manageInvites` | regenerate the invite link | ✓ | ✓ |
| `editGroup` | rename the group | | ✓ |
| `editMembers` | change other members' roles (more member actions to come) | | ✓ |
| `deleteGroup` | delete the group and its files | | ✓ |

### On the rename

Stored role values stay `'admin' | 'owner'` here to keep this separate from the rename migration. Because no logic branches on the literal anymore, renaming `admin` → `member` later is just relabeling the key in the `roles` table (plus a data migration and flipping the hardcoded `role: 'admin'` in `acceptInvite.ts`).

### Change type

- [x] `improvement`

### Test plan

1. As a group **owner**: rename the group, regenerate the invite, change a member's role, remove a file, delete the group — all succeed.
2. As a group **admin/member**: regenerate the invite and add/remove files succeed; the name input is disabled and the role dropdown / delete group are hidden.
3. Confirm the last owner still can't be demoted or leave.
4. Confirm a non-member is denied file access (socket connect, download) on a group-owned file.

- [x] Unit tests
- [ ] End to end tests

`roles.test.ts` covers `can`/`isRole` and the role → capability matrix; existing `mutators.test.ts` group tests cover the mutator gates.

### Code changes

| Section        | LOC change  |
| -------------- | ----------- |
| Core code      | +139 / -67  |
| Tests          | +69 / -0    |
| Apps           | +83 / -92   |


frolic:
Okay, cleaned up the API and I think this is ready to go.

This change was motivated by the comments in https://github.com/tldraw/tldraw/pull/9029/changes#r3356960773 where it was a bit unclear what the capabilities we're expressing in various method names.

Can follow up, or add here, the rename from admin to member.

jsscclr:
awesome! I think `can()` or `check()` are the way to go

packages/dotcom-shared/src/capabilities.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Every capability — a single thing someone can do in a group. Exported as a
+ * runtime list so all capabilities can be enumerated; the {@link Capability}
+ * type is derived from it. Roles grant subsets of these — see `roles.ts`.
+ */
+export const capabilities = [
```
angrycaptain19, line 6: renameGroup maybe?

MitjaBezensek, line 6: I wonder if this should be an exhaustive list or not. Like should we include `rotateInviteLink` even though everyone can do it right now. I'd say probably not, we can add those once we need them.

packages/dotcom-shared/src/mutators.ts
```diff
@@ -180,7 +162,7 @@ async function assertUserCanAccessFileInternal(
 		assert(file.ownerId === userId, ZErrorCode.forbidden)
 	} else if (file.owningGroupId) {
 		// New model: user must be a member of the owning group
-		await assertUserIsGroupMember(tx, userId, file.owningGroupId)
+		assert(can(await getRole(tx, userId, file.owningGroupId), 'accessFiles'), ZErrorCode.forbidden)
```
MitjaBezensek: How about if we moved `getRole` inside of `can` (or have a wrapper for doing it, something like `canUser`). This seems repetitive.

packages/dotcom-shared/src/capabilities.ts
frolic: these are just a mapping of current behaviors, and all these capabilities are internal, so they're easy to add/adjust! figured we'd start here and expand/refine over time

packages/dotcom-shared/src/mutators.ts
frolic: `getRole` is context dependent (here we fetch it in a DB tx, in the frontend we have it as part of the group membership object) whereas `can` is more general/pure/can be used in any context

if it's clearer, I can lift out the role like

```
const role = await getRole(...)
assert(can(role, 'accessFiles', ZErrorCode.forbidden)
```

frolic: fwiw I tried this in an earlier iteration and I thought it was harder to read
`can(tx, userId, groupId, 'accessFiles')`

"can transaction with user ID with group ID access files" doesn't read so clear as "can this role access files"

MitjaBezensek: FE and BE could define their own wrappers of can?

Happy to leave it as is if you think it reads better 👍

packages/dotcom-shared/src/capabilities.ts
frolic: these are easy to adjust as we need more specific capabilities, but in general I think an "edit group" encompasses most of what the intent is and reflects the current mutator naming (`updateGroup`)

packages/dotcom-shared/src/mutators.ts
frolic: I tried a few designs and this felt most readable to me! open to ideas if you can provide some code snippets that you find easier to read 

easy to wrap/replace `getRole` in mutators if we find a clearer way to represent this

MitjaBezensek (approved):

packages/dotcom-shared/src/roles.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Group authorization, expressed as capabilities rather than role names.
+ *
+ * Authorization asks `can(role, capability)` — never `role === 'owner'` — so the
+ * meaning of a role lives in exactly one place: the `roles` table below. Read a
+ * role's list to see what it can do; edit the list, or add a role, to change it.
+ *
+ * The role is stored in the DB as a plain string (`group_user.role`);
+ * capabilities are never persisted — they're derived from that string here.
+ *
+ * NOTE: `'admin'` is in the process of being renamed to `'member'`. Because no
```
MitjaBezensek: We still have some branching, see `setGroupMembershpRole` and a few other cases.

packages/dotcom-shared/src/mutators.ts
```diff
@@ -595,7 +599,8 @@ export function createMutators(userId: string) {
 			{ groupId, targetUserId, role }: { groupId: string; targetUserId: string; role: Role }
 		) => {
 			await assertUserHasFlag(tx, userId, 'groups_backend')
-			assert(can(await getRole(tx, userId, groupId), 'editMembers'), ZErrorCode.forbidden)
+			const actingRole = await getRole(tx, userId, groupId)
+			assert(can(actingRole, 'editMembers'), ZErrorCode.forbidden)
```
frolic: these should prob move under the other assertions (assert the group + user ID + role before fetching+checking role capability)

packages/dotcom-shared/src/mutators.ts
```diff
@@ -595,7 +599,8 @@ export function createMutators(userId: string) {
 			{ groupId, targetUserId, role }: { groupId: string; targetUserId: string; role: Role }
 		) => {
 			await assertUserHasFlag(tx, userId, 'groups_backend')
-			assert(can(await getRole(tx, userId, groupId), 'editMembers'), ZErrorCode.forbidden)
+			const actingRole = await getRole(tx, userId, groupId)
```
frolic: how about we alias the `role` arg to `targerRole`
then we can use `role` here (instead of actingRole) to represent user's role

packages/dotcom-shared/src/mutators.ts
frolic: > ```
> const role = await getRole(...)
> assert(can(role, 'accessFiles', ZErrorCode.forbidden)
> ```

moved to this approach for clarity

MitjaBezensek (approved):

PR #9068, opened 2026-06-08
improvement(tldraw): embed config

### LINKED ISSUE #9024: Make the default Google Maps embed read its API key from an option instead of NEXT_PUBLIC_GC_API_KEY
🧑: Make the default map embed definitions use an option for the GC API key, rather than "NEXT_PUBLIC". The key is public anyway, so it doesn't really matter if it's an environment variable or if it's leaked; the key itself should have protections on it.

---

## Problem

The default Google Maps embed definition in `packages/tldraw` reads its API key directly from a Next.js-specific environment variable:

```ts
// packages/tldraw/src/lib/defaultEmbedDefinitions.ts:129
result = `https://${host}/maps/embed/v1/view?key=${process.env.NEXT_PUBLIC_GC_API_KEY}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
```

This is a leaky abstraction in the SDK:

- `packages/tldraw` is a framework-agnostic library, but `NEXT_PUBLIC_GC_API_KEY` bakes in an assumption about a Next.js build (`process.env` inlining via the `NEXT_PUBLIC_` prefix). Consumers using Vite, Webpack, Bun, etc. have no way to populate this key, so the default Google Maps embed silently produces a URL with `key=undefined` for them.
- The env var name is specific to tldraw.com's own deployment (it's wired up in `internal/scripts/deploy-dotcom.ts` from `GC_MAPS_API_KEY`), so it only works for our own app.

The key is public regardless (it ships to the browser), so the security posture should come from restrictions on the key itself (HTTP referrer restrictions, API quotas), not from how it's injected. There's no reason to tie it to an env var read inside the SDK.

## Proposed solution

Expose the Google Maps API key as a configurable option rather than reading `process.env.NEXT_PUBLIC_GC_API_KEY` inside the default embed definition. Consumers should be able to provide their own key through the existing embed configuration surface, e.g.:

```ts
EmbedShapeUtil.configure({ embedDefinitions: [...] /* or a dedicated key option */ })
```

For tldraw.com, the key would be passed in at app setup time rather than read from the bundled SDK.

### Considerations

- `EmbedShapeUtil` already has an `options` object (`EmbedShapeOptions`) and a `configure(...)` path — a `googleMapsApiKey` (or more general per-embed config) option fits naturally there.
- The `toEmbedUrl` function for `google_maps` would need access to that option. Today the definition is a plain object with no access to the shape util's options, so the API will need to thread the key through (e.g. options-aware definition factory, or reading from the util at call time).
- Keep backward compatibility: tldraw.com currently relies on `NEXT_PUBLIC_GC_API_KEY` being present. The dotcom deploy wiring (`internal/scripts/deploy-dotcom.ts`) and the client setup should be updated to pass the key through the new option.
- Tests in `packages/tldraw/src/lib/utils/embeds/embeds.test.ts` reference `process.env.NEXT_PUBLIC_GC_API_KEY` directly and will need updating to match the new mechanism.

## Affected files

- `packages/tldraw/src/lib/defaultEmbedDefinitions.ts` (google_maps `toEmbedUrl`, line ~129)
- `packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx` (`EmbedShapeOptions`, `configure`)
- `packages/tldraw/src/lib/utils/embeds/embeds.test.ts` (test fixtures)
- `internal/scripts/deploy-dotcom.ts` and tldraw.com client setup (wire the key through the new option)

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9024

In order to stop the default Google Maps embed from reading `process.env.NEXT_PUBLIC_GC_API_KEY` inside the SDK (#9024), this PR lets consumers pass per-embed config through `EmbedShapeUtil.configure`. Config is keyed by embed type, so each definition only receives its own values and no key leaks across embeds.

```ts
EmbedShapeUtil.configure({
	embedConfig: { google_maps: { apiKey: '...' } },
})
```

tldraw.com now configures the embed util explicitly (reading the vite-inlined key in app code) instead of relying on the SDK reading the env var.

### Change type

- [x] `api`

### Test plan

1. Configure `EmbedShapeUtil` with a `google_maps` api key and paste a Google Maps URL — the embed renders with the key.
2. Without config, the maps embed URL has no key (no env fallback in the SDK).

- [x] Unit tests

### API changes

- Added `embedConfig` option to `EmbedShapeOptions`.
- Added `GoogleMapsEmbedConfig` and `DefaultEmbedConfig` types.
- `EmbedDefinition`/`CustomEmbedDefinition` are now generic over a `Config`; `toEmbedUrl(url, config?)` takes an optional per-embed config.
- `getEmbedInfo`/`matchUrl` accept an optional `embedConfig` argument.

### Release notes

- Add `embedConfig` to `EmbedShapeUtil` so embeds like Google Maps can be given an API key via `EmbedShapeUtil.configure({ embedConfig: { google_maps: { apiKey } } })`. The default Google Maps embed no longer reads `process.env.NEXT_PUBLIC_GC_API_KEY`.

### DIFF
--- a/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
+++ b/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
@@ -126,8 +126,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 							? zoomOrMeters
 							: -Math.log2(parseInt(zoomOrMeters) / 14772321) / 0.8
 					const host = new URL(url).host.replace('www.', '')
-					const apiKey = config?.apiKey ?? process.env.NEXT_PUBLIC_GC_API_KEY
-					result = `https://${host}/maps/embed/v1/view?key=${apiKey}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
+					result = `https://${host}/maps/embed/v1/view?key=${config?.apiKey}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
 				} else {
 					result = ''
 				}
--- a/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
+++ b/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
@@ -111,7 +111,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 		overridePermissions: {
 			'allow-presentation': true,
 		},
-		toEmbedUrl: (url) => {
+		toEmbedUrl: (url, config?: GoogleMapsEmbedConfig) => {
 			if (url.includes('/maps/embed?')) {
 				return url
 			} else if (url.includes('/maps/')) {
@@ -126,7 +126,8 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 							? zoomOrMeters
 							: -Math.log2(parseInt(zoomOrMeters) / 14772321) / 0.8
 					const host = new URL(url).host.replace('www.', '')
-					result = `https://${host}/maps/embed/v1/view?key=${process.env.NEXT_PUBLIC_GC_API_KEY}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
+					const apiKey = config?.apiKey ?? process.env.NEXT_PUBLIC_GC_API_KEY
+					result = `https://${host}/maps/embed/v1/view?key=${apiKey}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
 				} else {
 					result = ''
 				}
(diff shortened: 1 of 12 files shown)


packages/tldraw/src/lib/utils/embeds/embeds.test.ts
```diff
@@ -629,7 +632,7 @@ const MATCH_EMBED_TEST_URLS: (MatchEmbedTestMatchDef | MatchEmbedTestNoMatchDef)
 	},
 	{
 		embedUrl:
-			'https://google.com/maps/embed?key=${process.env.NEXT_PUBLIC_GC_API_KEY}&center=52.2449313,0.0813192&zoom=14',
+			'https://google.com/maps/embed?key=${GOOGLE_MAPS_API_KEY}&center=52.2449313,0.0813192&zoom=14',
```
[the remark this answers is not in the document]
kaneel: Pretty sure we're doing this for tests

mimecuvalo (commented):
i think the direction makes sense 👍 lemme know when ready for review!

packages/tldraw/src/lib/defaultEmbedDefinitions.ts
```diff
@@ -126,7 +126,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 							? zoomOrMeters
 							: -Math.log2(parseInt(zoomOrMeters) / 14772321) / 0.8
 					const host = new URL(url).host.replace('www.', '')
-					result = `https://${host}/maps/embed/v1/view?key=${process.env.NEXT_PUBLIC_GC_API_KEY}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
+					result = `https://${host}/maps/embed/v1/view?key=${config?.apiKey}&center=${lat},${lng}&zoom=${z}&maptype=${mapType}`
```
mimecuvalo: Cursor left a comment about this and I'd agree: to avoid breaking existing folks, let's have a fallback to `process.env.NEXT_PUBLIC_GC_API_KEY` and leave a note that that's a legacy fallback b/f th embedConfig existed.

mimecuvalo (approved):
pre-approved modulo one comment! great change

packages/tldraw/src/lib/defaultEmbedDefinitions.ts
kaneel: lessgo

PR #9069, opened 2026-06-08
test(examples): run-on-request device e2e (iOS + Android)

frolic:
In order to catch device-specific regressions — starting with multi-touch pinch, and eventually exports, clipboard, and shortcuts — that our Playwright suite structurally can't reach, this PR adds a run-on-request "Device e2e" lane that drives **real Mobile Safari** (iOS Simulator) and **real Chrome** (Android emulator) through the OS input pipeline via WebdriverIO + Appium. Relates to #7586.

Our existing Playwright tests inject synthetic touch via Chrome DevTools Protocol, so they only ever run Blink and never see the browser's own gesture recogniser. The regressions we actually hit (a pinch cancelling pointers / incidentally stealing the selection) live in that recogniser. This lane performs gestures as real OS touches in the native context and asserts against the live `editor` in the web context, so it exercises the part synthetic DOM events can't.

It runs **on demand only** — manually from the Actions tab (`workflow_dispatch`, choose platform) or by adding the `e2e-device-tests` label to a PR — on free GitHub-hosted runners (`macos-26` for iOS, `ubuntu-latest` for Android). It adds no per-PR cost, and since it's device automation a little flakiness is expected: re-run the job. The harness is deliberately named broadly (`e2e-device`, not "gesture"/"mobile") so it can grow to more behaviours and platforms.

### How it works

- `apps/examples/e2e-device/` — WebdriverIO + Appium suite. A shared base config plus per-platform capability files (`wdio.ios.conf.ts` → XCUITest/Safari, `wdio.android.conf.ts` → UiAutomator2/Chrome).
- `helpers/pinch.ts` — measures the canvas in CSS pixels (web context), translates to native device coordinates, then runs two simultaneous touch pointers in the native context. This native-context-with-coordinate-translation path is the only way multi-touch works on both iOS and Android.
- `tests/pinch.e2e.ts` — initial coverage mirroring the intent of the Playwright camera tests: pinch zooms in / out, and a two-finger pinch does not change the selection.
- `tests/draw-pressure-encoding.e2e.ts` — guards the engine contract behind #8963's 2D draw encoding: a real finger stroke must classify as non-pressure input (`dim: 2`). Unit tests choose the pressure values they inject, so only a device test can catch an engine starting to report varying pressure for plain touches (as 3D Touch iPhones once did), which would silently revert the encoding. Feature-detects `dim` and skips until #8963 merges; validated green against the merged feature on both platforms.
- The iOS job auto-picks an available iPhone simulator, pre-boots it (`simctl bootstatus`), and attaches Appium by UDID; the WebDriverAgent build is cached (`derivedDataPath` keyed on Xcode version + `usePrebuiltWDA` on a hit) so reruns skip the ~3.5 min build.

### Status — both lanes verified green

- **iOS** (`macos-26`): 3 pinch tests pass on real Mobile Safari (iPhone 17 Pro / iOS 26). Cache-hit run ~3–5 min, dominated by simulator boot and Safari's web-inspector session bring-up (inherently variable on the simulator) rather than our code.
- **Android** (`ubuntu-latest` + KVM): 3 pinch tests pass on real Chrome on the emulator. ~4 min total, with the tests themselves ~7s — no WDA build and no Safari handshake, so it's notably faster and lighter than iOS.

The default coordinate translation (`DEVICE_TOP_INSET=0`) worked on both, so no per-device calibration was needed.

### Change type

- [x] `other`

### Test plan

1. From the Actions tab, run the **Device e2e** workflow and choose a platform (`ios`, `android`, or `both`).
2. Alternatively, add the `e2e-device-tests` label to a PR that touches device behaviour.

- [ ] Unit tests
- [x] End to end tests

### Code changes

| Tests          | +346 / -0    |
| Config/tooling | +2174 / -76  |


mimecuvalo:
👀

apps/examples/e2e-device/wdio.shared.conf.ts
```diff
@@ -0,0 +1,56 @@
...
+	logLevel: 'warn',
+	waitforTimeout: 10_000,
+	// Generous per-request timeout: the first session creation has to build and
+	// launch WebDriverAgent on a cold simulator, which can take several minutes.
+	connectionRetryTimeout: 600_000,
+	// Do NOT retry session creation: each retry boots another simulator, and
+	// several simulators booting at once on a small CI runner causes them all to
+	// time out. One attempt with a long timeout; the spec-level retry handles
+	// genuine flakes.
+	connectionRetryCount: 1,
```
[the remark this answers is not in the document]
frolic: Good catch on the contradiction — the comment was stale, not the value. `1` (one retry, two attempts) is intentional in the current design: CI pre-boots a single simulator and Appium attaches by `appium:udid`, so a session retry reuses the same simulator and the running WebDriverAgent rather than booting another one (the stacking problem the old comment described only existed when Appium owned the simulator lifecycle). The retry also absorbs a real iOS flake: the first session attempt can fail on Safari's slow web-inspector attach, and the second attempt recovers against the warm Safari. Setting `0` would push that flake out to whole-spec-file retries. Updated the comment in 996722b7a3.

apps/examples/e2e-device/wdio.ios.conf.ts
```diff
@@ -0,0 +1,48 @@
...
+			browserName: 'Safari',
+			'appium:safariInitialUrl': `${shared.baseUrl}/end-to-end`,
+			'appium:newCommandTimeout': 120,
+			// Let Appium own the simulator lifecycle (create/boot the one it uses).
+			// Don't pre-boot in CI — a second simulator just causes boot contention.
+			'appium:noReset': true,
+			// First run builds WebDriverAgent from source, which is slow on a cold
+			// CI simulator. Give the build/launch generous headroom (no signing is
+			// needed on the simulator, so this is just time, not config).
+			'appium:wdaLaunchTimeout': 300_000,
```
mimecuvalo, line 37: small nit here (i noted in another PR that we have 840000 for a timeout limit which is nuts) - lets lower the limit from 5min to, i dunno, 1 min? if the server's not up by then, i think something else in the CI is up - instead of burning thru CI minutes on GH

.github/workflows/e2e-device.yml
```diff
@@ -0,0 +1,194 @@
...
+          api-level: 34
+          arch: x86_64
+          target: google_apis
+          # The emulator's `localhost` is the emulator itself, so forward the host
+          # dev server into it before running the tests.
+          script: |
+            adb reverse tcp:${{ env.EXAMPLES_PORT }} tcp:${{ env.EXAMPLES_PORT }}
+            cd apps/examples && yarn e2e-device-android
+
+      - uses: actions/upload-artifact@v6
```
mimecuvalo: are we actually uploading something?

mimecuvalo (approved):
great idea! let's gooooo

apps/examples/e2e-device/wdio.ios.conf.ts
frolic: fwiw this specific timeout came from a _lot_ of trial and error in claude, can probe the convo/context for exactly why this was picked

frolic: Dug up the reasoning: this is `wdaLaunchTimeout` (how long XCUITest waits for WebDriverAgent to build + launch on the sim), separate from the dev-server wait — that one's already 2min (`wait-on -t 120000` in the workflow). On a cache miss (first run, or after an Xcode/image bump busts the WDA cache) WebDriverAgent is compiled from source by xcodebuild, which measured ~3.5min in testing; an earlier 2min setting failed in exactly this spot. The 5min is just buffer over that measured ~3.5min cold build.

Worth noting it's a ceiling, not a per-run cost: a healthy run proceeds the instant WDA is up (~40s on a cache hit, ~3.5min cold), so it never burns the full 5min on green runs — it's only failure-detection latency when something's genuinely stuck. Keeping it at 5min for the cold-build headroom.

Sent from my aiPhone

.github/workflows/e2e-device.yml
frolic: Nope — good catch, they were uploading nothing. That `artifacts` dir was leftover from a temporary Appium-log diagnostic I added while debugging and then reverted, so both upload steps were just emitting "no files found." Removed them in a690785890. If we want real debugging output for flaky device runs later, the right move is to point the Appium service's `logPath` (and a failure screenshot) at that dir — but no point uploading an empty folder in the meantime.

Sent from my aiPhone

PR #9071, opened 2026-06-08
fix(dotcom): auto-scroll sidebar while dragging a file

### LINKED ISSUE #8972: Groups: Sidebar doesn't scroll while dragging a file
### What happened?

Native scrolling on the sidebar breaks while dragging a file. I should be able to scroll and for the list to scroll automatically when the item is at the top and bottom. Consider re-using the scrolling logic exactly from the page menu and align these two pieces of the UI.

_No response_

_No response_

### Where did this happen?

tldraw.com

jsscclr:
In order to let users scroll the sidebar file list while dragging a file, this PR adds edge auto-scroll to the sidebar's drag tracking. The sidebar uses the native HTML5 drag-and-drop API, which suppresses normal scrolling, so when a dragged file reached the top or bottom of an overflowing list the list stayed put and you couldn't reach the rest of it. The fix reuses the page menu's auto-scroll approach — same edge zone, ramp distance, and min/max speed as `DefaultPageMenu` — inside the drag tracking's existing `requestAnimationFrame` loop, aligning the two pieces of UI as requested. Closes #8972.

Because the loop re-detects drop targets from fresh element rects every frame and the browser's native drag image follows the cursor on its own, the only addition needed was a guarded `tickAutoScroll(container, mousePosition)` call plus a `data-sidebar-scroll-container` marker so the drag hook can resolve the scroll element (consistent with the drag subsystem's existing `data-*` contract).

### Preview

Test it on the live preview deploy: **https://pr-9071-preview-deploy.tldraw.com/**

(This URL is stable across commits — it always points at the latest build of this PR.)

### Change type

- [x] `bugfix`

### Test plan

1. Open the preview deploy at https://pr-9071-preview-deploy.tldraw.com/ and sign in.
2. Turn on the groups flag for your account: go to https://pr-9071-preview-deploy.tldraw.com/admin, search for your email, and enrol yourself in groups.
3. Create a number of files to fill up the sidebar so the file list overflows.
4. Start dragging a file and hold it near the top or bottom edge of the list.
5. The list auto-scrolls, with speed ramping up the closer the pointer is to the edge, matching the page menu's behavior.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix the sidebar not scrolling while dragging a file; the file list now auto-scrolls when a dragged file reaches the top or bottom edge.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +46 / -1   |

### DIFF
--- a/apps/dotcom/client/src/tla/hooks/useDragTracking.ts
+++ b/apps/dotcom/client/src/tla/hooks/useDragTracking.ts
@@ -60,6 +60,44 @@ function findHoveredWorkspaceId(
 	)
 }
 
+// Auto-scroll the sidebar while dragging when the pointer nears the top or
+// bottom edge of the scroll container. These mirror the page menu's drag
+// scrolling (see DefaultPageMenu) so the two pieces of UI behave the same.
+// AUTO_SCROLL_ZONE is how far from the edge the auto-scroll begins;
+// RAMP_DISTANCE is how far past that point it takes to reach max speed.
+const AUTO_SCROLL_ZONE = 16
+const AUTO_SCROLL_RAMP_DISTANCE = 48
+const MIN_AUTO_SCROLL_SPEED = 1
+const MAX_AUTO_SCROLL_SPEED = 6
+
+// Scrolls the sidebar container by one frame's worth of auto-scroll when the
+// pointer is within AUTO_SCROLL_ZONE of an edge, ramping the speed up the
+// further past the edge the pointer goes. Reorder/move targets are recomputed
+// on the next frame from fresh element rects, so no offset bookkeeping is
+// needed here — the native drag image follows the cursor on its own.
+function tickAutoScroll(container: HTMLElement, mousePosition: { y: number }) {
+	const rect = container.getBoundingClientRect()
+	const fromTop = mousePosition.y - rect.top
+	const fromBottom = rect.bottom - mousePosition.y
+	const maxScroll = container.scrollHeight - container.clientHeight
+
+	const overshootTop = AUTO_SCROLL_ZONE - fromTop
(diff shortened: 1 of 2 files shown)


apps/dotcom/client/src/tla/hooks/useDragTracking.ts
```diff
@@ -107,6 +107,44 @@ function findHoveredGroupId(
...
+const AUTO_SCROLL_RAMP_DISTANCE = 48
+const MIN_AUTO_SCROLL_SPEED = 1
+const MAX_AUTO_SCROLL_SPEED = 6
+
+// Scrolls the sidebar container by one frame's worth of auto-scroll when the
+// pointer is within AUTO_SCROLL_ZONE of an edge, ramping the speed up the
+// further past the edge the pointer goes. Reorder/move targets are recomputed
+// on the next frame from fresh element rects, so no offset bookkeeping is
+// needed here — the native drag image follows the cursor on its own.
+function tickAutoScroll(container: HTMLElement, mousePosition: { y: number }) {
```
frolic, line 78: woah is there really no browser API for this 😭

jsscclr: IIUC there's default behaviour surrounding dragging when the page is longer the available client height, but not for arbitrary divs. For scrolling to arbitrary points of random elements, there are scrolling APIs, but generally for traversing a large distance and/or at a single point in time. There's a high level `scrollTo` animation behaviour setting (`smooth`) but has no concept over control over velocity!

jsscclr:
@frolic pretty isolated code without too much domain knowledge required so think you would be a good candidate for review :D

frolic (approved):

PR #9074, opened 2026-06-09
feat(editor): align shapes on both axes

angrycaptain19:
In order to let SDK consumers center shapes on both axes with one command, this PR adds the `center` operation to `Editor.alignShapes`. The operation composes the existing horizontal and vertical center alignments, preserving their layout behavior without adding a default tldraw UI action. The shape transforms docs and align/distribute SDK example cover the new option.

### Change type

- [x] `api`

### Test plan

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add `Editor.alignShapes(..., 'center')` to center shapes on both axes.

### API changes

- Added `'center'` as an `Editor.alignShapes` operation. It applies the existing `'center-horizontal'` and `'center-vertical'` operations together.

### Code changes

| Section | LOC change |
| --- | --- |
| Core code | +11 / -1 |
| Tests | +19 / -10 |
| Automated files | +1 / -1 |
| Documentation | +8 / -4 |


mimecuvalo:
is this what you'd expect as the outcome?
Kapture 2026-06-09 at 15 57 46

angrycaptain19:
@mimecuvalo 

> is this what you'd expect as the outcome?

Yup exactly. One of the use cases a user described is often wanting to make a check icon where a circle has a checkmark inside it and they wanted to be able to easily align both the shapes center over each other

jsscclr:
https://github.com/user-attachments/assets/1159219d-0b45-479c-ade3-bea6252b22a4

Very much think we need this (super useful), but I think we need to separate horizontal axis and vertical axis

mimecuvalo:
agreed on the needs design review label 👍 added @steveruizok as a reviewer

angrycaptain19:
> Screen.Recording.2026-06-11.at.13.36.56.mov 
> Very much think we need this (super useful), but I think we need to separate horizontal axis and vertical axis

We talked about in person, but posting here for context. We already have both horizontal and vertical align - this is just adding the full align center to simplify that behavior from two commands -> one

steveruizok:
Not really needed from a UX level, clicking both is fine. But AI might benefit from having the explicit intent... I think we would only do the code path (the editor.align) rather than any extra buttons etc. Maybe we should call it center-center ?

angrycaptain19:
> Not really needed from a UX level, clicking both is fine. But AI might benefit from having the explicit intent... I think we would only do the code path (the editor.align) rather than any extra buttons etc. Maybe we should call it center-center ?

Yeah fair about not necessarily needing it at a UX level. We had it requested from a user and it's something I find myself wanting to do fairly often, but not terribly tricky to click align vert and align horizontal. Could see it being particularly useful for the AI stuff

PR #9080, opened 2026-06-09
fix(dotcom): show an empty state when a group invite link has expired

### LINKED ISSUE #9077: Show an empty state when a group invite link has expired
🗣️: When a group invite link is shared that can be used by an unlimited number of users for an unlimited amount of time until the link is changed. But as soon as the link in a group settings is changed the initial link no longer works and any new user trying to use it to join the group will experience a silent failure to join the group. We should add in a toast notifying the user that the link they were using has expired.

---

A group invite link stays valid for any number of users and any length of time until the group owner regenerates it from group settings. Regenerating is effectively a re-roll: the current UI only exposes a "refresh" action that overwrites the old link, so there is a single failure mode — the old link no longer resolves. Today a new user who follows that stale link gets no feedback at all; the join simply doesn't happen. We want that case to land the user on a dedicated empty-state (mirroring the existing "you don't have access to this file" state) so the failure is no longer silent. This is part of the groups work tracked under #8859, and touches the same invite-link join path as the previously fixed #8035.

## Open questions

1. **What exactly should the message say, and what is the right surface?**
   The message should read along the lines of "This invite link has expired" / "Ask the group owner for a new one." The surface is a dedicated empty-state, mirroring the existing "you don't have access to this file" state. (A toast was the original framing, but in practice a toast fired during the post-invite navigation is dismissed before the user can read it; the empty-state is the surface that reliably reaches them.)
2. **Should an expired link be distinguished from an otherwise invalid or revoked one?**
   No. There is no separate concept of expired vs. invalid/revoked — regenerating the link is just a re-roll (the UI only offers "refresh"), so a single "this link has expired" message covers it.
3. **Where does the user land after the failed join?**
   A dedicated empty-state, similar to the existing "you don't have access to this file" state.

jsscclr:
In order to stop a regenerated group invite link from failing silently, this PR shows a dedicated empty-state when someone follows an expired link. Closes #9077.

A group invite link works for any number of users until the owner regenerates ("refreshes") it from group settings, which is effectively a re-roll: the old link's secret no longer matches any group. Previously, a new user who followed the stale link hit a silent failure — `GroupInviteHandler` fetched the invite info, got a `404 { error: true }`, and `return`ed without telling the user anything. Genuine network failures in the same handler were also swallowed in the `.catch`.

Now an expired/invalid link shows a `TlaInviteExpiredDialog` empty-state (sad-face panel mirroring the existing "you don't have access to this file" state: "This invite link has expired" / "Ask the group owner for a new one."), and a real network/parse failure surfaces an error toast instead of being ignored.

A note on why the empty-state is a dialog rather than a toast (the issue's original framing): the handler runs while the app is mid-navigation (`/invite/:token → / → /f/:file`), and a toast added during that transition is dismissed at its default 4s duration before the user can read it. The dialog isn't on a duration timer, so it reliably reaches the user. The `.catch` toast uses `keepOpen: true` for the same reason.

### Change type

- [x] `bugfix`

### Test plan

1. Open the preview deploy at https://pr-9080-preview-deploy.tldraw.com/ and sign in.
2. Turn on the groups flag for your account: go to https://pr-9080-preview-deploy.tldraw.com/admin, search for your email, and enrol yourself in groups.
3. Create a group, open group settings, and copy the invite link.
4. Click "refresh" in group settings to regenerate the link (this invalidates the link you just copied).
5. Open an incognito window and paste the **original** (now stale) invite link.
6. Confirm an empty-state dialog appears: "This invite link has expired" / "Ask the group owner for a new one." (Previously: nothing happened.)

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Show an empty state when a user follows an expired group invite link, instead of silently failing to join.

### Code changes

| Apps            | +111 / -3  |
| Automated files | +36 / -0   |

Screenshot 2026-06-09 at 12 14 46

### DIFF
--- a/apps/dotcom/client/public/tla/locales-compiled/en.json
+++ b/apps/dotcom/client/public/tla/locales-compiled/en.json
@@ -69,6 +69,12 @@
       "value": "File menu"
     }
   ],
+  "0c4383617a": [
+    {
+      "type": 0,
+      "value": "Something went wrong loading this invite. Please try again."
+    }
+  ],
   "0d1f2bb523": [
     {
       "type": 0,
@@ -679,6 +685,12 @@
       "value": "You are already a member of this workspace"
     }
   ],
+  "88bf6ab42c": [
+    {
+      "type": 0,
+      "value": "This invite link has expired"
+    }
+  ],
   "8ad4303b83": [
     {
       "type": 0,
@@ -1177,6 +1189,12 @@
       "value": "App debug flags"
     }
   ],
+  "e0aa021e21": [
+    {
+      "type": 0,
+      "value": "OK"
+    }
+  ],
   "e0c88179b3": [
     {
       "type": 0,
@@ -1229,6 +1247,12 @@
     {
       "type": 0,
       "value": "Cancel"
+    }
+  ],
+  "eb1e2b0542": [
+    {
+      "type": 0,
+      "value": "Ask the workspace owner for a new one."
     }
   ],
   "ebfe9ce86e": [
--- a/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
@@ -6,6 +6,7 @@ import {
 } from 'tldraw'
 import { sadFaceIcon } from '../../../components/ErrorPage/ErrorPage'
 import { F } from '../../utils/i18n'
+import { TlaButton } from '../TlaButton/TlaButton'
 import styles from './TlaInviteExpiredDialog.module.css'
 
 export function TlaInviteExpiredDialog({ onClose }: { onClose(): void }) {
@@ -22,9 +23,9 @@ export function TlaInviteExpiredDialog({ onClose }: { onClose(): void }) {
 				<p className={styles.message}>
 					<F defaultMessage="Ask the workspace owner for a new one." />
 				</p>
-				<button className={styles.okButton} onClick={onClose}>
+				<TlaButton variant="cta" className={styles.okButton} onClick={onClose}>
 					<F defaultMessage="OK" />
-				</button>
+				</TlaButton>
 			</TldrawUiDialogBody>
 		</>
 	)
--- a/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
@@ -12,16 +12,13 @@ export function TlaInviteExpiredDialog({ onClose }: { onClose(): void }) {
 	return (
 		<>
 			<TldrawUiDialogHeader className={styles.dialogHeader}>
-				<TldrawUiDialogTitle>
-					<span />
-				</TldrawUiDialogTitle>
 				<TldrawUiDialogCloseButton />
 			</TldrawUiDialogHeader>
 			<TldrawUiDialogBody className={styles.dialogBody}>
 				{sadFaceIcon}
-				<h1 className={styles.header}>
+				<TldrawUiDialogTitle className={styles.header}>
 					<F defaultMessage="This invite link has expired" />
-				</h1>
+				</TldrawUiDialogTitle>
 				<p className={styles.message}>
 					<F defaultMessage="Ask the workspace owner for a new one." />
 				</p>
--- a/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
@@ -1,0 +1,34 @@
+import {
+	TldrawUiDialogBody,
+	TldrawUiDialogCloseButton,
+	TldrawUiDialogHeader,
+	TldrawUiDialogTitle,
+} from 'tldraw'
+import { sadFaceIcon } from '../../../components/ErrorPage/ErrorPage'
+import { F } from '../../utils/i18n'
+import styles from './TlaInviteExpiredDialog.module.css'
+
+export function TlaInviteExpiredDialog({ onClose }: { onClose(): void }) {
+	return (
+		<>
+			<TldrawUiDialogHeader className={styles.dialogHeader}>
+				<TldrawUiDialogTitle>
+					<span />
+				</TldrawUiDialogTitle>
+				<TldrawUiDialogCloseButton />
+			</TldrawUiDialogHeader>
+			<TldrawUiDialogBody className={styles.dialogBody}>
+				{sadFaceIcon}
+				<h1 className={styles.header}>
+					<F defaultMessage="This invite link has expired" />
+				</h1>
+				<p className={styles.message}>
+					<F defaultMessage="Ask the workspace owner for a new one." />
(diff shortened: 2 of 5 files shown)


apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { sadFaceIcon } from '../../../components/ErrorPage/ErrorPage'
+import { F } from '../../utils/i18n'
+import styles from './TlaInviteExpiredDialog.module.css'
+
+export function TlaInviteExpiredDialog({ onClose }: { onClose(): void }) {
+	return (
+		<>
+			<TldrawUiDialogHeader className={styles.dialogHeader}>
+				<TldrawUiDialogTitle>
+					<span />
```
jsscclr: 🤔

jsscclr:
@angrycaptain19 FYI - the original issue suggested using a toast, but I ended up going with a dialog instead - toast felt too subtle.
Screenshot 2026-06-09 at 12 11 21

angrycaptain19:
> @angrycaptain19 FYI - the original issue suggested using a toast, but I ended up going with a dialog instead - toast felt too subtle. 

Makes sense and looks good to me!

apps/dotcom/client/public/tla/locales-compiled/en.json
```diff
@@ -1259,6 +1277,12 @@
       "value": "App debug flags"
     }
   ],
+  "e0aa021e21": [
+    {
+      "type": 0,
+      "value": "OK"
```
mimecuvalo, line 1195: funny this is the 1st time we have the word OK in the translations heh

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
```diff
@@ -0,0 +1,31 @@
...
+			</TldrawUiDialogHeader>
+			<TldrawUiDialogBody className={styles.dialogBody}>
+				{sadFaceIcon}
+				<TldrawUiDialogTitle className={styles.header}>
+					<F defaultMessage="This invite link has expired" />
+				</TldrawUiDialogTitle>
+				<p className={styles.message}>
+					<F defaultMessage="Ask the group owner for a new one." />
+				</p>
+				<button className={styles.okButton} onClick={onClose}>
```
mimecuvalo: let's use one of the defined buttons in our instead of the custom styling

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
```diff
@@ -0,0 +1,31 @@
...
+					<F defaultMessage="OK" />
```
mimecuvalo, line 27: hm, i guess they can just close the dialog if they want to ignore this but since this auto-popups it feels like they're forced to hit OK?
i'm actually wondering, since there's a new toast in this PR anyway, we could just add it as part of the toast which has an affordance to add a primary action + close/ignore.

example:

Image

wdyt?

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
jsscclr: Can do - for some context, it was mostly a dumber version of the existing dialog. But then I hated `<TldrawUiDialogTitle><span /></TldrawUiDialogTitle>` anyway so I half started cleaning it up. 

I was umming and ahhing about just doing a clean-up as I go - as a little offshoot, as a general rule for PRs, are we more "leave better than we left it" or "focused only"? In the first case would probably look like filing issues as we go (have noticed very spotty / mixed button usage, ideally just a single clean-up pass, or more targeted Dialog clean-up, etc), second case I'd do a little bit of clean-up unification. In an ideal world first is better but second is more realistic in terms of actually fixing things up (in my experience anyway!)

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
angrycaptain19: I'm ok with them being forced to hit ok in this scenario. It's an error state where we really want them to know it's failed and they need to re-ask for the invite in this scenario

jsscclr: So we're on the same page:

- I added the toast for the 'Network failure' edge case of when something's gone wrong during this process
- The dialog case, where you're forced to hit OK, feels right to me, because there was a strong intent to invite the group (I pasted the URL).

Are you saying that you don't agree with the use of the Dialog in this case, since it's too much friction, and that showing a Toast would be preferable in this case? In which case I would push back, I think friction is useful here (somewhat relevant: https://news.ycombinator.com/item?id=46196831).

apps/dotcom/client/public/tla/locales-compiled/en.json
jsscclr: quite interesting! I'm actually more of an 'Okay' person, but none of those either AFAICT - and I felt like "OK" was more tldraw (although maybe "ok" is peak tldraw tone of voice????). was toying around with a 'tone of voice' doc to make these kinds of things super obvious and avoid (me) thinking too hard about these things :D

angrycaptain19:
Tested on mobile and tablet and looks good. How do you see this working alongside your PR to show the invite message before you log in?

Would they then see the expired link modal and then a blank canvas (or whatever canvas they were working on)? I assume that and not see the error and then be prompted to log in?

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
mimecuvalo: makes sense — def. feel your pain here :)

apps/dotcom/client/src/tla/components/dialogs/TlaInviteExpiredDialog.tsx
mimecuvalo: fair enough — i think i might still slightly disagree but don't feel super strongly about it.
feels like if the link is busted, that might be our bad vs. the owner having to regenerate one again. (which then might break for others w the older link?) i'd be wary of some loop of it being busted and just trying to bother the owner a lot. but totally may be overthinking it here.

jsscclr: Wouldn't that be an issue in either case? If the link 404ing is our fault or the result of a re-roll, the end user is still going to be unable to join the workspace, and reaching out to whoever sent you the link would be first port of call?

apps/dotcom/client/public/tla/locales-compiled/en.json
mimecuvalo: haha yeah so with you on your line of thinking there and back-and-forth :D

mimecuvalo: which is to say, i don't have a strong pref here, except maybe it should be "Ok" - might be more "tldraw-y"

mimecuvalo (approved):
👍

Issue #9082, opened 2026-06-09
Show invite message before the log-in screen to logged-out users


jsscclr:
🗣️: When a logged-out user clicks an invite link, it doesn't tell them anything about the invite — it just expects them to log in. We should show the invite message (e.g. which group they've been invited to) before/alongside the log-in screen.

---

When a logged-out user opens a group invite link, they're taken straight to the sign-in screen with no context about the invite — no group name, nothing indicating they were invited to anything. The invite details only surface after they've signed in. We want logged-out users to see which group they've been invited to before/alongside the sign-in prompt, so the link feels intentional and they understand why they're being asked to log in.

## Open questions

1. **Does this cover only group invite links, or also other invite/share links (e.g. shared-file links)?**
   Group invite links only. `/invite/:token` is group-specific; shared-file links are a separate access mechanism with no accept/decline invite flow, so there's no other "invite link" to cover.
2. **What should the message include — just the group name, or also who invited them (inviter name/avatar)?**
   Just the group name for now. (Inviter attribution would need a backend change and is out of scope.)
3. **Should the copy/CTA differentiate sign-up (new user) from sign-in (returning user) for an invited person, or is the existing dialog wording fine?**
   Existing dialog wording is fine.

Confidence: 90%, ready to get started.

PR #9085, opened 2026-06-09
feat(dotcom): show invite message before sign-in for logged-out users

jsscclr:
In order to make workspace invite links feel intentional to people who aren't signed in, this PR shows the invite message before sign-in (#9082) — and rather than presenting that dialog on a blank page, the invite route now resolves to the normal root experience with the dialog overlaid on top.

For a logged-out visitor, `/invite/:token` stores the token, redirects to `/`, and the anonymous editor renders behind the sign-in dialog, which reads "You have been invited to join **{workspace name}**". The invite info comes from the auth-free `/api/app/invite/:token` endpoint; invalid or expired tokens fall back to the plain sign-in dialog, so nothing hangs if the lookup fails. For a signed-in visitor nothing changes: they land on their most recent file with the accept-invite dialog on top, as before.

Mechanically, `WorkspaceInviteHandler` (mounted at the root) now owns both halves of the invite flow. The invite route redirects with an `?invite=true` marker; the handler shows the sign-in dialog only when arriving with that marker and strips it once the dialog is up. This matters because the invite token intentionally stays in session storage after the dialog is dismissed (so a later sign-in still completes the join) — without the marker, every reload of tldraw.com in the same tab would re-show the dialog. Dismissing the dialog simply leaves the visitor on the usable anonymous editor.

### Change type

- [x] `improvement`

### Test plan

You need a workspace invite link. Create it while signed in, then test the logged-out experience in an incognito window.

**Setup**

1. Open the preview deploy at https://pr-9085-preview-deploy.tldraw.com/ and sign in.
2. Turn on the workspaces UI for your account: go to https://pr-9085-preview-deploy.tldraw.com/admin, search for your email, and enrol yourself in groups.
3. Create a workspace, open workspace settings, and copy the invite link.

**Logged-out invite (the change)**

4. Open an incognito window and paste the invite link (`/invite/:token`).
5. ✅ You land on the anonymous editor (not a blank page) with the sign-in dialog on top reading "You have been invited to join **{workspace name}**".
6. Close the dialog → you stay on the anonymous editor and can draw.
7. Reload tldraw.com in the same tab → the invite dialog does not reappear.
8. Open a second workspace's invite link → the dialog shows the new workspace's name.
9. Sign in through the dialog → the join completes (accept-invite dialog, or the already-a-member toast) as before.

**Edge cases / no regression**

10. In an incognito window, open an invalid or expired invite link → the plain sign-in dialog shows over the editor with no invite line (doesn't hang).
11. While signed in, open a valid invite link → still lands on your most recent file and shows the accept-invite dialog.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Show the invited workspace's name on the sign-in dialog when a logged-out user opens an invite link, with the anonymous editor visible behind instead of a blank page.

### Code changes

| Apps            | +143 / -75 |
| Tests           | +93 / -1   |
| Automated files | +19 / -0   |


angrycaptain19:
We might want to rephrase the invite message so it's a bit more obvious what you're invited to. 

Maybe "You've been invited to join the {workspace name} tldraw workspace"?

Tested on mobile and ipad and all looks good and worked correctly

jsscclr:
The original copy is "You have been invited to join group: <group name>"

I actually made a conscious decision to drop the mention of group (which would now be workspace), since again I'm coming from the angle of a user is coming in explicitly via a workspace invite link. 

Thinking of e.g. "You've been invited to join the *Class of 2025* tldraw workspace" -> I think `tldraw` is pretty redundant. I also think workspace is actually pretty low information density in the sense that an end user who is joining might now know that they're joining a workspace, but without any information about workspaces. We have a little spiel explaining tldraw, I think explaining workspaces as well would also be a bit too wordy.

I'm not too attached, but explaining my reasoning since it was intentional on my part to drop the terminology! Here's an example of joining Discord, for reference: 

Screenshot 2026-06-12 at 10 24 05

jsscclr:
@angrycaptain19 Your questions about how this relates to #9085 made me realise it doesn't really make sense at all to not have this dialog as an overlay on the anonymous editor - feels very wrong now that I've realised!

angrycaptain19:
> I think tldraw is pretty redundant.

Why do you think it's redundant in this case? I guess if someone already knows tldraw it is fairly redundant, but thinking about the case where someone on a team is invited to a workspace for the first time and has no idea what this link they were just sent is.

But that's also true for the discord example you shared and I do quite like that flow they have

jsscclr:
To clarify, I think the 'tldraw' is redundant because there's a logo just above saying tldraw, so for us to have

```
tldraw

You've been invited to join the {workspace name} tldraw workspace
```

The second tldraw doesn't feel like it's doing a lot of heavy lifting IMO!

frolic:
Is this the screen the PR refers to?

image

jsscclr:
Good feedback from @frolic: currently if you refresh, the invite modal disappears, since the `?invite=true` param is cleared on the redirect to the home page. Don't think there's really any negatives to persisting the query params anyhow so will keep them so it makes refresh behaviour more predictable.

frolic:
also ideally the invite link is a persistent URL where dismissing is a `push`, and only consider a `replace` if you're already a member/signed in

jsscclr:
@frolic updated - uses consistent query param `invite` + dismissing the invite pushes a new url onto the history stack

apps/dotcom/client/src/tla/components/WorkspaceInviteHandler.tsx
```diff
@@ -14,82 +15,158 @@ import { defineMessages, useMsg } from '../utils/i18n'
...
-						addToast({
-							id: 'workspace-invite-already-member',
-							title: alreadyMemberMsg,
-						})
-						app.navigateToWorkspaceFiles(data.workspaceId)
-						return
-					}
-
-					// Show invite dialog
+		if (!auth.userId) return
```
mimecuvalo, line 110: can you check something for me here? these 3 gates were at the top before but now they're after this new bit of invite logic. that might not worry me by itself but because the new logic does `return () => controller.abort()` combined with `useEffect` dep array checking for these, makes me worry that the `controller.abort()` might be firing if the dep array is changing while the requests are going out (i.e. multiple requests being fired in quick succession) can you check that isn't happening? thx!

mimecuvalo (commented):
LG! one comment to check before landing por favor

apps/dotcom/client/src/tla/components/WorkspaceInviteHandler.tsx
jsscclr: DId some stress testing locally and holds up fine for me. I split out the logic for signed-in and signed-out into 2 separate effects - hopefully that makes it a bit more readable / predictable! Made a little flowchart as well: https://www.tldraw.com/f/85p1Ofq5avypYT-hbRV7T?d=v-1570.-3324.8818.6071.page

mimecuvalo: hah we should start putting mermaid diagrams in the code, i'm only half-kidding

mimecuvalo (approved):
nice! thanks for the latest rework 🤘

PR #9093, opened 2026-06-10
perf(editor): recheck only affected bounds in the spatial index

### LINKED ISSUE #8882: SpatialIndexManager re-walks every indexed shape on each diff
## Problem

`SpatialIndexManager.processIncrementalUpdate` walks every shape currently in the R-tree on every diff (SpatialIndexManager.ts:139-154), comparing current bounds against indexed bounds. This catches shapes whose bounds changed transitively — arrows whose bound shapes moved, groups whose children moved — but it makes every shape mutation O(n) in the total indexed shape count, defeating the O(log n) point of the spatial index.

## Where

- `packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts:137-154`

## Impact

Performance regression at scale. On a page with thousands of shapes, every shape mutation triggers an O(n) walk plus n bounds-cache lookups. Invisible at small sizes but becomes a CPU hotspot for large canvases.

## Why the fallback exists

This is the load-bearing part: arrows and groups can have their **bounds** change without their **record** changing. An arrow bound to a moved shape doesn't get a new arrow record in the diff — only `getShapePageBounds(arrowId)` returns something new on next read, because the bounds cache invalidated. Same for a group with moved children. The diff-only path can't see these.

So this isn't simply "fix the diff completeness" — the diff is complete for what it represents (record-level changes), but the spatial index cares about a derived view (bounds) that doesn't have its own record stream.

## Possible approaches

Roughly ordered by scope:

1. **Restrict fallback to known dependent shapes.** Maintain an inverse index: for each shape in the diff, look up arrows bound to it (via binding records) and groups containing it, and only check those shapes' bounds. Avoids the full walk but requires keeping the inverse mapping fresh.
2. **React to the bounds cache directly.** Subscribe to bounds-cache invalidations rather than the shape diff. Catches everything, but probably means going through the shape page-bounds computed cache's reactivity — see #8883 for the related implicit-dep issue.
3. **Bound the fallback by ancestor.** Only check shapes whose ancestors appear in the diff. Cheap but misses arrow-binding cases.

Whatever the path, add a regression test: arrow anchor moves and group child moves must keep the index in sync without scanning every shape.

## Context

Surfaced during an internal modularity audit of the editor manager classes. Related: #8883 (the same module's implicit dependency on the bounds cache).

MitjaBezensek:
In order to make viewport culling scale to large pages, this PR replaces the spatial index's full bounds recheck with a structural dirty set. Previously every shape diff (i.e. every pointer move during a drag) iterated all indexed shapes and pulled their page bounds to catch derived-bounds changes — O(page) work per update. The dirty-set approach expands each record diff into the set of shapes whose bounds can actually change and rechecks only those — O(affected) per update. Closes #8882.

This is a performance change, not a correctness fix. The binding-only correctness fix — arrows staying hit-testable after `createBindings`/`updateBinding` with no shape change — already shipped in #9183, which routed binding diffs through that same full O(page) sweep. This PR keeps that guarantee (binding diffs still invalidate dependent arrow bounds) while removing the per-update full scan, so it is layered on top of #9183 rather than replacing its behavior. #9183's note that "any future optimization of the step-2 sweep must keep binding diffs invalidating dependent arrow bounds" is exactly what this honors; its two binding-only regression tests are retained, and a third covering binding deletion in its own transaction is added.

The expansion covers the three channels through which bounds change without a record change:

- descendants of a moved shape (their page transforms move with the parent)
- ancestors of a changed shape (group bounds derive from children)
- binding partners, to a fixpoint (arrows follow bound shapes, including arrows bound to groups whose children moved)

### Behavior change

Replacing the full sweep with structural expansion narrows one previously-incidental behavior: a custom shape whose `getGeometry` reads other shapes outside parent/child and binding relationships is no longer rechecked when those other shapes change, so its indexed bounds can go stale. The old full sweep rechecked every shape on every update and so covered this case by accident. Core shapes only depend on other shapes via bindings (arrows) and parent/child relationships (groups), both of which are covered. This is documented as a known limitation in the class doc and called out in the release notes below.

### Benchmark

10k-shape page, jsdom, per pointer-move step including the store update itself. The benchmark was a throwaway micro-benchmark and is removed from the branch; the [source is preserved in this commit](https://github.com/tldraw/tldraw/commit/22749ecdd4c2c1ba964b705e20d4a4e2ecd82362) so reviewers can see how the numbers were produced.

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| Drag 1 shape | 23.1ms | 1.0ms | 23x |
| Drag shape with bound arrow (200 arrows on page) | 28.6ms | 0.9ms | 31x |
| Drag group with 1k children | 27.5ms | 14.1ms | 2x |
| Move 1k shapes per step | 35.3ms | 20.2ms | 1.7x |

The remaining ~1ms/step in the single-shape drag is the `updateShapes` transaction itself; the index update is near-zero. The group and bulk scenarios are dominated by bounds that genuinely change.

### Change type

- [x] `improvement`

### Test plan

1. Open a document with many shapes, groups, and bound arrows
2. Drag shapes, groups, and arrow-bound shapes; pan and zoom
3. Verify shapes cull correctly at viewport edges, arrows follow their targets, and grouped shapes follow their group

- [x] Unit tests

### Release notes

- Improve viewport culling performance on large pages: the spatial index now rechecks only shapes affected by a change instead of every shape on the page.
- Custom shapes whose geometry depends on other shapes outside parent/child and binding relationships are no longer automatically rechecked for culling when those other shapes change. Shapes that follow the standard parent/child and binding patterns (including all built-in shapes) are unaffected.


packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
```diff
@@ -52,27 +66,28 @@ export class SpatialIndexManager {
 		const bindingHistory = this.editor.store.query.filterHistory('binding')
 
 		return computed<number>('spatialIndex', (_prevValue, lastComputedEpoch) => {
-			if (isUninitialized(_prevValue)) {
-				return this.rebuildAndBumpEpoch()
-			}
-
+			// These three reads are the computed's only tracked dependencies.
```
steveruizok, line 69: Could this still leave stale spatial index entries for built-in shapes whose geometry depends on non-shape records? Bookmark geometry can depend on a resolved bookmark asset via `getBookmarkShapeHeight`. I reproduced this locally by indexing a placeholder bookmark at 320px height, creating the resolved no-image bookmark asset so `getShapePageBounds(bookmarkId).h` shrinks, then moving an unrelated geo shape. A query over the old-only lower region still returned the bookmark because asset changes never dirty it, and the later unrelated shape diff only rechecked that shape. The previous full sweep would have repaired this on the next shape/binding update. I think asset/theme/font/zoom-derived geometry needs an invalidation path, or this assumption needs narrowing.

packages/editor/src/lib/editor/managers/SpatialIndexManager/SpatialIndexManager.ts
```diff
@@ -101,88 +116,160 @@ export class SpatialIndexManager {
...
-							changed = true
-						}
-					} else if (this.rbush.has(to.id)) {
-						this.rbush.remove(to.id)
-						changed = true
-					}
-				} else if (this.rbush.has(to.id)) {
-					this.rbush.remove(to.id)
-					changed = true
+				transformChanged.add(to.id)
```
steveruizok, line 164: This treats every shape update as a transform change, so prop-only updates to containers still walk and recheck the entire descendant subtree. For example, renaming/restyling/resizing a frame or group with many children will mark every child dirty even though their page transforms cannot move. That seems to give back a lot of the performance win on common container edits. Could this only add to `transformChanged` when `x`, `y`, `rotation`, or `parentId` changes, while still marking the updated shape itself dirty for bounds recheck?

mimecuvalo (commented):
async review: lookin' good! let's check perf benchmarks (synthetic+RUM) later

PR #9094, opened 2026-06-10
perf(draw): optimize the freehand ink algorithm

steveruizok:
In order to make the freehand ink algorithm faster and its rendered output smaller without changing how strokes look, this PR replaces the implementation in `packages/tldraw/src/lib/shapes/shared/freehand` with an optimized one. The optimization was developed against a frozen copy of the previous algorithm using a benchmark and comparison harness, with each round validated for bit-identical output over a stroke corpus against a frozen baseline of the previous round. The harness, the baseline, and the temporary `@tldraw/freehand` package the work was developed in were all removed once the results were validated, and remain available in this branch's history.

### Layout

| Part | What it is |
| --- | --- |
| `packages/tldraw/src/lib/shapes/shared/freehand` | The optimized implementation, in the same folder the previous one lived in. New files: `core.ts` (the struct-of-arrays pipeline) and `fmt.ts` (the byte-buffer svg path writer) |
| `packages/tldraw/src/test/freehand` | The svg well-formedness and svg path tests, plus the corpus that drives them: 20 deterministic simulated strokes (dots to 2,400-point strokes; mouse, stylus, solid, highlighter) using the option presets tldraw passes in production, and real recorded strokes |

### Results (current implementation vs original, full corpus)

Direct head-to-head benchmark of the final implementation against the original algorithm (the frozen pre-optimization baseline, restored from this branch's history). Interleaved ops/sec for the full render pipeline (input points -> svg path data), best of 5 rounds per implementation:

| Bucket | Cases | Speedup (geomean) |
| --- | --- | --- |
| Short (<50 input points) | 8 | 1.59x |
| Medium (50-500) | 15 | 1.91x |
| Long (500+) | 5 | 2.67x |
| **Overall** | **28** | **1.92x** |

- Svg path data is 57% smaller overall
- No case is slower; long draw strokes gain the most (up to 3.15x on a dense 2,400-point stroke), while dots and taps are bounded by fixed overhead (~1.45x at 500k-2M ops/sec)
- Output parity vs the original, measured during round 1: 38% fewer outline points, max shape deviation 0.26px (5% of the stroke size, the simplifier's hard bound), mean 0.018px. Rounds 2 and 3 were bit-identical

<details>
<summary>Per-case results</summary>

| Case | Input | Original | Current | Speedup | Svg size |
| --- | --- | --- | --- | --- | --- |
| dot-single-point | 1 | 1,497,754/s | 2,150,809/s | 1.44x | 56% smaller |
| dot-jittered-tap | 4 | 541,821/s | 794,778/s | 1.47x | 35% smaller |
| dot-pen-tap | 5 | 1,271,355/s | 1,819,704/s | 1.43x | 62% smaller |
| short-flick | 9 | 343,420/s | 700,319/s | 2.04x | 68% smaller |
| short-tick | 14 | 298,555/s | 562,381/s | 1.88x | 58% smaller |
| short-pen-comma | 18 | 256,245/s | 444,495/s | 1.73x | 53% smaller |
| short-solid-line | 12 | 1,386,493/s | 1,967,561/s | 1.42x | 36% smaller |
| medium-cursive-word | 220 | 22,898/s | 55,583/s | 2.43x | 62% smaller |
| medium-pen-signature | 260 | 17,995/s | 41,477/s | 2.30x | 54% smaller |
| medium-zigzag-sharp | 180 | 28,340/s | 71,038/s | 2.51x | 79% smaller |
| medium-spiral | 240 | 19,808/s | 46,779/s | 2.36x | 45% smaller |
| medium-scribble-fast | 150 | 30,953/s | 72,524/s | 2.34x | 60% smaller |
| medium-highlight-wave | 160 | 101,865/s | 203,487/s | 2.00x | 40% smaller |
| medium-solid-curve | 200 | 78,965/s | 163,909/s | 2.08x | 42% smaller |
| medium-in-progress | 180 | 27,281/s | 66,753/s | 2.45x | 61% smaller |
| long-handwriting-line | 1200 | 2,979/s | 7,607/s | 2.55x | 63% smaller |
| long-pen-paragraph | 1600 | 1,845/s | 5,366/s | 2.91x | 59% smaller |
| long-scribble-fill | 2000 | 1,510/s | 4,431/s | 2.93x | 63% smaller |
| long-spiral-dense | 2400 | 1,269/s | 4,000/s | 3.15x | 57% smaller |
| long-highlight-strikes | 1000 | 13,309/s | 26,400/s | 1.98x | 44% smaller |
| real-1 | 61 | 76,337/s | 121,402/s | 1.59x | 55% smaller |
| real-2 | 81 | 58,823/s | 90,494/s | 1.54x | 40% smaller |
| real-3 | 57 | 82,299/s | 127,313/s | 1.55x | 43% smaller |
| real-4 | 95 | 47,290/s | 69,515/s | 1.47x | 39% smaller |
| real-5 | 5 | 519,285/s | 752,912/s | 1.45x | 34% smaller |
| real-6 | 132 | 35,270/s | 56,530/s | 1.60x | 36% smaller |
| real-7 | 99 | 55,004/s | 82,522/s | 1.50x | 36% smaller |
| real-8 | 125 | 35,913/s | 55,013/s | 1.53x | 43% smaller |

</details>

The three optimization rounds were each measured against a frozen baseline of the previous round during development: 1.33x for the initial rewrite, ~1.5x for removing the stored per-point vector, and 1.23x on `svgInk` (1.13x `getStroke`, 1.05x centerline) for the struct-of-arrays core. Their compounded estimate (~2.5x) overstates the directly measured 1.92x; the head-to-head above is the number to trust.

### How the optimized implementation gets there

- Svg path data is written into a shared byte buffer with integer math and decoded once per path. Doubles rounded to 2 decimals routinely miss V8's fast float-to-string path, so coordinates are tracked as integer hundredths of a pixel.
- Paths use relative commands (`t`/`q`/`a`/`l`): deltas between consecutive points are small numbers that write fast and keep the data compact, and integer deltas sum exactly so rounding never drifts.
- Outline tracks are simplified with a tolerance of 5% of the stroke size, and cap arcs are tessellated adaptively by chord error instead of fixed step counts.
- The elbow partitioner and outline-track loops use scalar math instead of allocating temporary vectors per point, and `getStrokePoints` strips leading points by index instead of repeated splices.
- Stroke points no longer store a per-point unit vector: consumers derive each point's vector from its predecessor with scalar math, eliminating a `Vec` allocation and normalization per input point. `svgInk`'s elbow partitions carry a "vector anchor" so points following an elbow or a cleanup splice keep the direction they had in the uncut stroke; output stays bit-identical across the corpus.
- The internal pipeline runs entirely in one module-level set of reusable struct-of-arrays buffers (`Float64Array`s for coordinates, pressure, distance, running length, and radius; non-reentrant, like the path writer's byte buffer). Raw input is ingested and streamlined straight into the arrays with no per-point objects, radii are computed over the arrays in a single pass with the taper pass folded in, and the outline tracks are built into flat arrays with an in-place simplifier. `svgInk`'s elbow partitions are index ranges plus two skip counters (cleanup only ever removes points adjacent to the partition ends), with elbow boundaries reading the raw input coordinates at the boundary index.
- The public functions are thin adapters over the array core with unchanged signatures and behavior: `getStrokePoints` materializes `StrokePoint` objects from the arrays, `setStrokePointRadii` keeps an in-place object form (in lockstep with the array core's radius pass), and `svgInk`/`getStroke` never materialize stroke points at all.

### SDK integration

Nothing changes for consumers: the same stroke functions and types are exported from the `tldraw` package with the same names, paths, and call signatures, and `DrawShapeUtil`, `HighlightShapeUtil`, `getPath`, and the scribble overlays import the algorithm from the same local module as before.

The editor's `Vec` no longer imports `EASINGS`: the two easings it used as defaults (`easeInQuad`, `easeInOutQuad`) are inlined as local functions, so the vector primitive no longer depends on the easing table.

### Change type

- [x] `improvement`

### Test plan

1. `yarn workspace tldraw test run src/test/freehand` - corpus-driven svg well-formedness checks and svg path unit tests
2. Draw and highlight in the examples app with solid and pattern fills, thin to extra-thick sizes, and a stylus if available; strokes should look identical to production

- [x] Unit tests

### Release notes

- Improve freehand ink performance: draw and highlighter strokes compute roughly 2x faster (up to 3x for long strokes) and produce 57% smaller svg path data, with no visible change to how strokes look.

### API changes

- Breaking! Removed `StrokePoint.vector`. The vector is derivable from consecutive points (`uni(prev.point - point)`); nothing in the SDK consumed it outside the stroke pipeline itself.
- No other public API changes: `tldraw` exports the same stroke functions and types with the same signatures, and the point types are still the editor's `Vec`.

### Code changes

| Core code       | +1512 / -596 |
| Tests           | +541 / -0    |
| Automated files | +0 / -5      |


steveruizok:
Closed in favor of https://github.com/tldraw/tldraw/pull/9154

PR #9098, opened 2026-06-10
build: require Node >=22.12.0 (drop EOL Node 20)

frolic:
In order to let our published packages depend on modern ESM-only modules without breaking CommonJS consumers, this PR drops support for the now-EOL Node 20 and requires Node `>=22.12.0` — a version where `require()` of an ES module works natively.

Calling `require()` on an ES module (instead of throwing `ERR_REQUIRE_ESM`) is unflagged as of Node 22.12.0. Node 20 reached EOL in April 2026, so rather than carry the awkward `^20.19.0 || >=22.12.0` range we just require `>=22.12.0`. With that floor, an ESM-only dependency in a `dist-cjs` build just works for consumers on a supported Node.

## Changes

- Set the root `engines.node` to `>=22.12.0`.
- Add a yarn constraint (`enforceNodeEngineOnPackages` in `yarn.config.cjs`) that applies the same range to every `packages/*` workspace, applied with `yarn constraints --fix`. (CI's `yarn constraints` keeps it consistent going forward.)
- Update the AGENTS.md setup note to match.

## Why per-package, not just the root

The root `package.json` is `private` and is **never published** — its `engines` only constrains the monorepo dev environment; a consumer never sees it. When someone runs `npm install @tldraw/editor`, npm reads **that package's own** published `engines` and warns on a mismatch. There's no "monorepo root" in what they install (just a single package tarball), and npm has no notion of inheriting `engines` from a workspace root.

| Where `engines` lives | Who it affects |
| --- | --- |
| root `package.json` (private) | tldraw developers building the monorepo |
| each published package | consumers who install that package |

So to actually signal the requirement to consumers, the field has to be on the packages they install. Each is independently installable (someone might pull in just `@tldraw/store`), so each declares its own; the yarn constraint keeps them in sync. Note this is a new pattern for the repo — until now `engines` lived only at the root.

## Notes

- **`engines` is advisory** — npm only *warns* on an unmet engine, it doesn't hard-fail the install. So this won't break anyone's build; it signals the supported range.
- Alternative/complement: #9050 solves the same ESM-only-in-CJS problem at the build level (inlining ESM-only deps into the CJS output). The `@tldraw/mermaid` lazy-import that previously rode along here is now its own PR: #9100.

### Change type

- [x] `improvement`

### Test plan

1. `yarn constraints` and `yarn check-packages` pass (engines applied consistently to every `packages/*`).
2. Published packages now declare `engines.node: ">=22.12.0"`.

### Release notes

- Drop support for the EOL Node 20. The minimum supported Node is now `>=22.12.0` — the version where `require()` of an ES module is supported natively, which lets tldraw safely depend on modern ESM-only packages.

### Code changes

| Documentation  | +1 / -1    |
| Config/tooling | +71 / -2   |


frolic:
lifted out the mermaid change here: https://github.com/tldraw/tldraw/pull/9100

mimecuvalo (commented):
actually, let's just drop Node 20 support — it got to EOL in April so we don't need to support it. we can just specify `>=22.12.0`. 
looking into, this won't break builds by default, it'll just warn that the engine isn't supported

frolic:
Good call — dropped Node 20 entirely. Now just `>=22.12.0` on the root and every published package (and the AGENTS.md note), and agreed it's advisory so it only warns on a mismatch rather than failing installs.

frolic:
do we want to wait for [the security advisory](https://nodejs.org/en/blog/vulnerability/june-2026-security-releases) and set a min node version across LTS releases?

mimecuvalo (approved):

mimecuvalo:
> do we want to wait for [the security advisory](https://nodejs.org/en/blog/vulnerability/june-2026-security-releases) and set a min node version across LTS releases?

agreed — pre-approved nonetheless

mimecuvalo:
ok the latest advisory look like they don't affect us directly. happy to do >= 22.23 though

frolic:
Re: the [June 2026 Node security releases](https://nodejs.org/en/blog/vulnerability/june-2026-security-releases) — decided to **keep `engines` at `>=22.12.0`** rather than ratchet the floor up to the patched versions (22.23.1 / 24.17.1).

`engines` is advisory (it only warns), and it expresses a feature/compat floor — `>=22.12.0` is "where `require(esm)` is unflagged," not a security policy. Pinning it to security patches would be perpetual churn (Node patches roughly monthly) and wouldn't actually enforce anything.

Still **bumped the CI/build Node to 24.17.1** — the runtime we actually control — in a separate PR: #9311. That's the right lever for staying patched.

(For what it's worth, we don't think tldraw is exposed to these CVEs anyway — they're in Node's TLS/HTTP/WebCrypto/permission-model subsystems; the SDK is browser code and tldraw.com runs on workerd. And 20.x got no patch since it's EOL, which only reinforces dropping it here.)

Sent from my aiPhone

PR #9099, opened 2026-06-10
fix(tldraw): don't trigger adjacent shortcut when releasing a modifier mid-hold

### LINKED ISSUE #9095: Releasing a modifier while holding a shortcut key triggers the adjacent non-modifier shortcut
### What happened?

1. Hit Shift+Q for the new copy-on-hover style command
2. Let go of Shift first but keep holding Q
3. The Toggle Lock action will be changing constantly now.

If you selected one action, you shouldn't fallback to a similarly keyed action when the modifier is let go.

_No response_

_No response_

### Where did this happen?

mimecuvalo:
In order to fix #9095, this PR stops a held shortcut key from falling back to an adjacent non-modifier shortcut when a modifier is released mid-hold.

Previously, pressing `shift+q` (copy hovered styles) and then releasing shift while still holding `q` would let the auto-repeat keydown events start firing the plain `q` shortcut (toggle tool lock) over and over. The keyboard shortcut hook now tracks which registration each physically-held key first triggered (keyed by `event.code`) and won't switch to a different shortcut while that key stays down. The same shortcut is still allowed to repeat, so behaviors like holding `=` to keep zooming are unaffected.

### Change type

- [x] `bugfix`

### Test plan

1. Select a shape and hover another shape with a style.
2. Press `shift+q` to copy the hovered styles.
3. Release shift but keep holding `q`.
4. Confirm the tool lock toggle does not start flickering on/off.

- [x] Unit tests

### Release notes

- Fix releasing a modifier while still holding a shortcut key triggering the adjacent non-modifier shortcut.


packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
```diff
@@ -141,12 +141,25 @@ export function useKeyboardShortcuts() {
...
 			for (const reg of registry) {
 				if (!reg.onKeyDown) continue
 				for (const p of reg.parsed) {
 					if (matchesEvent(e, p)) {
+						const prev = code ? heldKeyRegistrations.get(code) : undefined
+						// The held key already triggered a different shortcut; don't fall back to
+						// this one just because a modifier was released.
+						if (prev && prev !== reg) break
+						if (code) heldKeyRegistrations.set(code, reg)
 						reg.onKeyDown(e)
```
frolic, line 163: ooc would it be a bad idea to `event.preventDefault()` in each key handler and then wrap the `break` in a check on `event.isDefaultPrevented`?

mimecuvalo: mmm, i'd be wary of conflating the two mechanisms. i think there's valid paths where want preventDefault semantics to stay "pure" and in my head adding this added layer to it might be non-obvious to folks reasoning about this.

packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
```diff
@@ -156,6 +169,7 @@ export function useKeyboardShortcuts() {
 
 		const handleKeyUp = (e: KeyboardEvent) => {
 			if (shouldSkipEvent(e)) return
+			if (e.code) heldKeyRegistrations.delete(e.code)
```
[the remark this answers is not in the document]
mimecuvalo: Good catch — fixed in 127c4eb. The held-key entry is now cleared at the top of `handleKeyUp` before the `shouldSkipEvent` gate, so a key released while focus is inside a text input still releases the tracking. Added a regression test ("releases the held key even when the keyup lands on a text input").

packages/tldraw/src/lib/ui/hooks/useKeyboardShortcuts.ts
frolic: I guess what I am thinking is some indication that the keyboard shortcut was "consumed"

Issue #7372, opened 2025-12-13
Copy and paste keyboard shortcuts do not work on iPad


kaznaan:
### What happened?

I'm using Tldraw on my iPad Pro with Magic Keyboard but copy (Cmd+C) and paste (Cmd+D) doesn't work on canvas.  Some other keyboard shortcuts such as 'Cmd+/' works.

### How can we reproduce the bug?

1. Open tldraw.com with iPad with keyboard
2. Create shapes
3. Try copy and paste them using keyboard shortcuts

### What browsers are you seeing the problem on?

Safari

### Where did this happen?

The tldraw.com website

### Contact Details

kaz@naan.net

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

steveruizok:
Just looked into this, seems to be quite a few issues right now with the iPad keyboard. We might need to do a bigger roundup of these.

PR #9101, opened 2026-06-10
test(examples): device e2e for copy/paste keyboard shortcuts (#7372)

frolic:
## Status: closed — not reliably testable on the current simulator setup

This PR added a characterization test for #7372 (Cmd+C / Cmd+V doing nothing on iPad Safari with a hardware keyboard). The minimal test actually went **green** on the iOS simulator, but we closed this route because the simulator observably does not match the real-device behavior the issue describes — so the test would guard the wrong layer. Notes below so this is quick to pick back up.

### What we established empirically (iOS 26 simulator, Appium/XCUITest)

- WebdriverIO's standard `browser.keys()` **never reaches Safari web content** — zero keydowns arrive at the page, even plain letters into a focused, visible `<input>`. The W3C key-action API errors out in WebDriverAgent ("Key Down action must have a closing Key Up successor").
- Appium's native **`mobile: keys`** (XCUITest hardware-keyboard injection; `modifierFlags` bitmask Shift=`1<<1`, Ctrl=`1<<2`, Cmd=`1<<4`) **does** deliver real keyboard events to the page. This is the only working way to drive the keyboard on iOS Safari from the harness.
- With the normal, non-editable canvas container focused: plain keys and Shift+key arrive fine, but **Cmd- and Ctrl-modified chords are swallowed entirely** (no keydown, no copy/paste events, nothing pastes).
- The same Cmd chord with a focused **visible `<input>`**: arrives with `metaKey: true`. Two foot-guns that cost debugging runs: an `opacity: 0` input receives nothing, and enumerating Appium contexts blurs the focused element.

### Why we closed instead of landing the green test

- The simulator swallows **all** Cmd/Ctrl chords on the canvas, but the original report says some chords (e.g. Cmd+/) **do** work on a real iPad while Cmd+C/V don't. The simulator is blunter than real hardware, so a green/red flip of this test tracks *what the simulator does*, not *what an iPad with a Magic/Bluetooth keyboard does*.
- Real-hardware keyboard quirks here are documented and iOS-version-dependent, which only a real device can capture: [WebKit #149054](https://bugs.webkit.org/show_bug.cgi?id=149054) (external-keyboard events not firing; fixed iOS 13), [WebKit #137072](https://bugs.webkit.org/show_bug.cgi?id=137072) (CapsLock sets `ctrlKey`), [WebKit #273681](https://bugs.webkit.org/show_bug.cgi?id=273681) (Shift state changes without events), and the iPadOS 14.5 Magic Keyboard Shift regression.

### What would make this testable

- A **real-device lane** (physical iPad + hardware keyboard; Appium supports real devices but that's hosted hardware), or
- a one-time validation on a real iPad that the simulator's swallow behavior matches the device — which would restore this sim test's value as a proxy.

### Salvageable from this branch

- `e2e-device/helpers/keyboard.ts` — the `mobile: keys` helper (only working keyboard path on iOS Safari).
- `e2e-device/tests/copy-paste-shortcut.e2e.ts` — the minimal symptom repro (green as of this close).

The underlying bug #7372 stays open. The fix direction (hidden contenteditable keyboard sink so iOS delivers Cmd chords) is being prototyped separately.

---

*Original description follows.*

Builds on #9069 (the device e2e harness) — **base branch is `frolic/device-e2e`, not `main`**.

In order to guard against tldraw/tldraw#7372 — Cmd+C / Cmd+V not working on iPad Safari with a hardware keyboard (while other shortcuts like Cmd+/ do) — this PR adds a device e2e test that selects a shape, presses the copy and paste shortcuts via real hardware-keyboard events, and asserts the current (buggy) behavior: nothing is pasted.

### DIFF
--- a/apps/examples/e2e-device/helpers/app.ts
+++ b/apps/examples/e2e-device/helpers/app.ts
@@ -22,7 +22,14 @@ export async function openEditor() {
 		editor.selectAll().deleteShapes(editor.getSelectedShapeIds())
 		editor.setCurrentTool('select')
 		editor.resetZoom()
+		// Focus the canvas container so keyboard shortcuts reach the editor.
+		;(document.querySelector('.tl-container') as HTMLElement | null)?.focus()
 	})
+}
+
+/** Number of shapes on the current page. */
+export function getShapeCount(): Promise<number> {
+	return browser.execute(() => editor.getCurrentPageShapeIds().size)
 }
 
 /** Read the current zoom level from the live editor. */
--- a/apps/examples/e2e-device/helpers/keyboard.ts
+++ b/apps/examples/e2e-device/helpers/keyboard.ts
@@ -1,0 +1,12 @@
+import { Key } from 'webdriverio'
+
+/**
+ * Press the platform "accelerator" chord plus a key, mirroring tldraw's
+ * `isAccelKey` (Cmd on Apple platforms, Ctrl elsewhere). Passing the keys as an
+ * array makes WebdriverIO hold the modifier down for the chord and release both
+ * at the end — i.e. a real Cmd+C, not "Cmd" then "C" typed separately.
+ */
+export async function pressAccel(key: string) {
+	const accel = browser.isIOS ? Key.Command : Key.Ctrl
+	await browser.keys([accel, key])
+}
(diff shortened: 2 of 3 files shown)


frolic:
Closing — the simulator can't faithfully represent the real-iPad keyboard behavior this test targets (full post-mortem in the description). #7372 stays open; revisit with a real-device lane or after validating sim fidelity on one physical iPad.

PR #9102, opened 2026-06-10
feat(frame): hint shapes that will be enclosed while drag-creating a frame

### LINKED ISSUE #8846: Frame tool should indicate shapes that will be enclosed while drawing
## Problem

When drawing a frame with the frame tool, shapes whose bounds fall inside the drag rectangle automatically become children of the new frame on pointer up. Today there is no visual feedback during the drag to communicate which shapes that will be — users only find out after they release the pointer.

## Proposed solution

While the user is dragging out a new frame, render selection-style indicators on the shapes that would be enclosed by the frame at the current cursor position. The indicators should update live as the drag rectangle changes, so users can see exactly which shapes will be parented into the frame before committing.

This mirrors the feedback the brush selection already provides for the select tool and would make the frame tool's containment behavior discoverable.

## Implementation sketch

- The frame tool inherits from `BaseBoxShapeTool` and uses its `Pointing` child state for drag-to-create (`packages/editor/src/lib/editor/tools/BaseBoxShapeTool/children/Pointing.ts`).
- Containment logic for frame children lives in `FrameShapeTool.onCreate` and the `canEnclose` helper (`packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts`).
- Indicator rendering can reuse the existing hinting mechanism (`editor.setHintingShapes()` / `getHintingShapeIds()`), which the indicator overlay already styles distinctly.

The feedback should clear out when the drag ends or is cancelled.

## Alternatives considered

- A translucent fill on shapes to be enclosed — heavier and less consistent with existing tldraw feedback patterns.
- No live feedback, document the behavior — leaves the discoverability problem in place.

AniKrisn:
In order to make the frame tool's containment behavior discoverable, this PR shows live selection-style indicators on the shapes that will become children of a new frame while it is being dragged out. Previously, shapes silently became children of the frame on pointer up with no feedback during the drag. Closes #8846.

Because the frame tool hands the live drag off to `select.resizing` (via `BaseBoxShapeTool`'s `Pointing` state), the hinting runs in the select tool's `Resizing` state: when resizing a shape that is being created and is a frame, it computes the sibling shapes that would be enclosed at the current bounds and feeds them to `editor.setHintingShapes()`, which the indicator overlay already renders with a thicker stroke. The enclosure predicate is extracted from `FrameShapeTool.onCreate` into a shared internal helper, `getEnclosedShapeIds`, so the live hints and the final reparenting always agree. Hints update in both directions as the drag rectangle grows and shrinks, skip locked shapes and partially-overlapped shapes, and clear on pointer up, cancel, and between successive creations in tool-locked mode.

https://github.com/user-attachments/assets/1077057b-c9a8-4ae9-97e0-c7ebec7d1d63

### Change type

- [x] `improvement`

### Test plan

1. Create a few shapes on the canvas
2. Select the frame tool and drag out a frame over them
3. Shapes fully enclosed by the drag rectangle show a thicker indicator outline, updating live as the rectangle grows and shrinks
4. On pointer up, the indicators clear and exactly the hinted shapes become children of the frame
5. Pressing Escape mid-drag clears the indicators
6. Locked shapes and partially-overlapped shapes are never hinted

- [x] Unit tests

### Release notes

- Frame tool now shows live indicators on the shapes that will be enclosed by a frame while it is being drawn.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +60 / -24  |
| Tests     | +72 / -0   |

### DIFF
--- a/packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts
+++ b/packages/tldraw/src/lib/shapes/frame/FrameShapeTool.ts
@@ -1,4 +1,4 @@
-import { BaseBoxShapeTool, TLShape, TLShapeId } from '@tldraw/editor'
+import { BaseBoxShapeTool, Editor, TLShape, TLShapeId } from '@tldraw/editor'
 
 /** @public */
 export class FrameShapeTool extends BaseBoxShapeTool {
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
i think this is a great idea 🤘

PR #9106, opened 2026-06-10
feat(dotcom): improve member management in workspace settings

### LINKED ISSUE #9131: Unify dropdown styling (role dropdowns differ from style panel; Firefox renders them unstyled)
🗣️: the role dropdown's are not correctly styled in firefox

---

The workspace settings member role dropdowns (e.g. "Owner") have two related styling problems. First, even in Chrome they don't match the other dropdowns/selects used across the app — they look noticeably different from the controls in the share panel and from the style panel's dropdowns. Second, in Firefox they fall back to the browser's default native `select` appearance entirely (a grey native box with the browser's own chevron), making the inconsistency much worse.

The desired direction is to unify the role dropdown styling with the app's existing select styling — specifically the style panel dropdowns as the reference — so all of these controls look consistent. That unification should also resolve the Firefox-specific native-rendering bug, since the style panel selects already render correctly in Firefox.

Visual context (from a Firefox screenshot the reporter shared): two stacked role dropdowns labeled "Owner", each shown as a light-grey rounded rectangle with a native chevron, looking out of place against the surrounding UI.

## Open questions

1. **Which control should be the canonical reference to unify all of these dropdowns on?**
   The style panel dropdowns. The share panel selectors also render correctly but look different again; the style panel ones are the intended target.
2. **Which browsers render the role dropdown incorrectly?**
   Firefox falls back to native styling. Safari and Chrome render the role dropdown without the native fallback, but it still looks different from the style/share panel dropdowns — so the inconsistency is cross-browser, and the native fallback is a Firefox-only additional problem.
3. **What OS and Firefox version shows the native-fallback bug?**
   macOS, Firefox 151.0.4 (aarch64).
4. **Should the unification cover every mismatched select in the app, or just the workspace member role dropdown for now?**
   _Awaiting answer._

frolic:
In order to improve member management in workspace settings, this PR reworks the members list and the role picker. Closes #9131.

**Members list**

- Sorted by role (owners first, then members), with the current user pinned to the top of their own role group. For example, a member sees:

  ```
  Owner 1
  Owner 2
  Member 1 (you)
  Member 2
  Member 3
  ```

- Every member's role is shown to everyone as a read-only label; changing a role still requires the `editMembers` capability.

**Role picker**

- Migrated from a bespoke native `<select>` to the shared Radix `TlaMenuSelect` used elsewhere in the app, for consistency and accessibility. This fixes the role dropdown rendering as an unstyled native `select` in Firefox and brings it in line with the app's other selects.
- Members with `editMembers` get a separate, destructive section in the picker to **remove another member** from the workspace; selecting it asks for confirmation. Hidden on your own row — use "Leave workspace" there.
- Opening the picker no longer dismisses the settings modal.

**Cleanup**

- Dropped the redundant client-side last-owner guards on the picker; that invariant is enforced server-side in the mutators.

### Change type

- [x] `improvement`

### Test plan

1. As a **member**: owners are listed first, then members, with you pinned to the top of the members.
2. As an **owner**: you appear at the top of the owners; the role picker is the standard app dropdown (keyboard accessible, check on the current role, and properly styled in Firefox).
3. Every member's role label is visible; only members with `editMembers` get the editable picker.
4. As an **owner**, open another member's picker → "Remove from workspace" (in its own section) → confirm → the member is removed. Confirm the last owner can't be removed.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Workspace settings: members are grouped by role (owners first) with you pinned to the top of your role; the role picker is now consistent and keyboard-accessible (and no longer falls back to a native dropdown in Firefox); members who can edit members can remove others from the workspace.

### Code changes

| Core code       | +28 / -0   |
| Tests           | +56 / -0   |
| Automated files | +35 / -0   |
| Apps            | +140 / -108 |

An an owner:
image

As a member:
image

### DIFF
--- a/apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
+++ b/apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
@@ -105,6 +105,7 @@ export function TlaMenuSelect<T extends string>({
 	disabled,
 	onChange,
 	options,
+	actions,
 	usePortal,
 	'data-testid': dataTestId,
 }: {
@@ -114,6 +115,9 @@ export function TlaMenuSelect<T extends string>({
 	disabled?: boolean
 	onChange(value: T): void
 	options: { value: T; label: ReactNode }[]
+	// Extra actions shown in their own section below the options (e.g. a
+	// destructive "remove"). Selecting one runs its onSelect instead of onChange.
+	actions?: { id: string; label: ReactNode; onSelect(): void; destructive?: boolean }[]
 	// When set, render the dropdown in a portal (popper-positioned) so it isn't
 	// clipped by / doesn't overflow a constrained container like a modal dialog.
 	usePortal?: boolean
@@ -123,9 +127,14 @@ export function TlaMenuSelect<T extends string>({
 	const container = useContainer()
 	const handleChange = useCallback(
 		(value: string) => {
+			const action = actions?.find((a) => a.id === value)
+			if (action) {
+				action.onSelect()
+				return
+			}
 			onChange(value as T)
 		},
-		[onChange]
+		[actions, onChange]
 	)
 
 	const handleOpenChange = (open: boolean) => {
@@ -198,6 +207,23 @@ export function TlaMenuSelect<T extends string>({
 									<_Select.ItemText>{option.label}</_Select.ItemText>
 								</_Select.Item>
 							))}
+							{actions && actions.length > 0 && (
+								<>
+									<_Select.Separator className={styles.menuSelectSeparator} />
+									{actions.map((action) => (
+										<_Select.Item
+											key={action.id}
+											className={classNames(
+												styles.menuSelectOption,
+												action.destructive && styles.menuSelectOptionDestructive
+											)}
+											value={action.id}
+										>
+											<_Select.ItemText>{action.label}</_Select.ItemText>
+										</_Select.Item>
+									))}
+								</>
+							)}
 						</_Select.Viewport>
 					</_Select.Content>
 				</TlaSelectPortal>
--- a/apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
+++ b/apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
@@ -10,7 +10,7 @@ import {
 	useEffect,
 	useState,
 } from 'react'
-import { TldrawUiButton, TldrawUiIcon, TldrawUiTooltip } from 'tldraw'
+import { TldrawUiButton, TldrawUiIcon, TldrawUiTooltip, useContainer } from 'tldraw'
 import { defineMessages, useMsg } from '../../utils/i18n'
 import { TlaIcon } from '../TlaIcon/TlaIcon'
 import styles from './menu.module.css'
@@ -105,6 +105,7 @@ export function TlaMenuSelect<T extends string>({
 	disabled,
 	onChange,
 	options,
+	usePortal,
 	'data-testid': dataTestId,
 }: {
 	id: string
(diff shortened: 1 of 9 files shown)


apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
```diff
@@ -258,45 +259,53 @@ export function GroupSettingsDialog({ groupId, onClose }: GroupSettingsDialogPro
...
+											{member.userName.charAt(0).toUpperCase()}
+										</div>
+										<span className={styles.memberName}>
+											{member.userName}
+											{member.userId === app.getUser().id ? ` (${youMsg})` : ''}
+										</span>
+										{canEditThisRole ? (
+											<MemberRoleSelect
+												value={member.role}
+												disabled={member.role === 'owner' && ownersCount <= 1}
```
frolic: is this applicable now that `canEditThisRole` has a similar check?

apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
```diff
@@ -258,45 +259,53 @@ export function GroupSettingsDialog({ groupId, onClose }: GroupSettingsDialogPro
...
+												labels={roleLabels}
+												onChange={async (value) => {
+													if (value === member.role) return
+													if (member.role === 'owner' && value !== 'owner' && ownersCount <= 1)
```
frolic: same as above

apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
frolic: Right — it's unreachable, so I removed it. The dropdown only renders when `canEditThisRole` is true, which requires the viewer to be an owner (only owners have `editMembers`). For an owner's row to also satisfy `ownersCount <= 1` that owner would have to be the *sole* owner — but with one owner `canEditThisRole` collapses to `member.userId !== viewerId`, so the sole owner's own row never renders a dropdown, and no other row can be an owner. So `member.role === 'owner' && ownersCount <= 1` is always false here. The last-owner invariant is also enforced server-side in `setGroupMemberRole` (asserts `owners.length > 1`), so nothing is lost. Dropped the now-orphaned `disabled` prop from `MemberRoleSelect` too. (1599e48)

apps/dotcom/client/src/tla/components/dialogs/GroupSettingsDialog.tsx
frolic: Same reasoning — removed. This guard's condition is identical (`member.role === 'owner' && … && ownersCount <= 1`), so it's unreachable for the same reason the `disabled` prop was: `canEditThisRole` already prevents the sole owner's editable row from rendering, and the server enforces the invariant in `setGroupMemberRole`. Kept the `if (value === member.role) return` no-op check since that's unrelated. (1599e48)

mimecuvalo:
mind adding a screenshot(s) in the desc. so i can see how the two look side-by-side? i'm curious to see if this looks noisy from the owner's POV

frolic:
> mind adding a screenshot(s) in the desc. so i can see how the two look side-by-side? i'm curious to see if this looks noisy from the owner's POV

This doesn't change the owner's view, only reduces noise for members.

jsscclr:
Screenshot 2026-06-11 at 12 05 12
Screenshot 2026-06-11 at 12 08 49
Screenshot 2026-06-11 at 12 09 35
Screenshot 2026-06-11 at 12 13 13
Screenshot 2026-06-11 at 12 15 02

FWIW I think it looks a little weird for end users with a lone 'Owner' floating, and I would prefer busy over less coherent. Would be less of an issue I think if owners were always pinned on top (which may make sense anyway). 

current state, which I prefer:
Screenshot 2026-06-11 at 12 18 51

frolic:
so to clarify, is this your proposal:
- always show role labels
- pin owners to the top
- your own user is always pinned above all others?

(I am in favor of this)

frolic:
An an owner:
image

As a member:
image

apps/dotcom/client/src/tla/components/tla-menu/tla-menu.tsx
```diff
@@ -113,14 +115,26 @@ export function TlaMenuSelect<T extends string>({
 	disabled?: boolean
 	onChange(value: T): void
 	options: { value: T; label: ReactNode }[]
+	// Extra actions shown in their own section below the options (e.g. a
```
MitjaBezensek, line 118: I wonder if we should introduce a proper section concept instead of this special "destructive actions" case?

frolic: yeah probably, will rethink this menu in a follow up

PR #9086, opened 2026-06-09
feat(dotcom): rework sidebar into a workspace switcher

MitjaBezensek:
Behind the `groups_frontend` flag. The sidebar UI matches the Figma design; opening for UX review + a dotcom preview.

> [!IMPORTANT]
> **Some changes affect users *without* the flag too.** The legacy sidebar shares the sidebar shell and file-row CSS, so these visual tweaks apply to everyone: section titles ("Today", "Older") go bold → medium, idle file names go `text-2` → `text-1` (darker), the footer username goes `text-1` → `text-3` (lighter), file rows lose the per-row pin icon styling space (right padding 8 → 4px), the top-row indentation changed slightly, and the sidebar shows its right border on mobile. The legacy recents list structure and behaviour are otherwise untouched.
>
> Also: the preview-deploy auto-enable of `groups_frontend` was removed — testing the new UI on a preview now requires the flag on your user.

In order to simplify how users navigate their files and groups, this PR reworks the tldraw.com sidebar from the previous "all groups visible with inline expand/collapse" model into a single-selection workspace switcher.

- A fixed top region lists the spaces the user can switch between — `My files` (home group) and their workspaces — plus a `Create a workspace +` button.
- Selecting a space opens its most recent file. The active space is **derived from the open file** (its `owningGroupId`), so there is no new persistence — the highlighted space always matches the current file's group. (Whether clicking a workspace should open its recent file or only swap the file list is still an open product question; e2e tests deliberately only assert on file-list contents.)
- The lower region scrolls and shows the active space's files, pinned first and then grouped by recency (Today / Older / …). Per-item pin icons were removed in favour of the `Pinned` section label.
- Files move between spaces by dragging a file onto a space, or via the file `Move to` menu. The drag follows the file so its new space becomes active. `Move to` no longer shows checkboxes and hides the space the file is already in (a file lives in exactly one group).
- Selecting an empty workspace (or creating a new one) creates a file in it, so a space always lands you on a file. The `New file` button creates in whichever space is active.
- Removed the **"Add file link"** option entirely (dialog, sidebar menu item, app method, Zero mutator + tests, analytics event). It only inserted a `group_file` row for a file you don't own; no schema/migration change, existing linked files still render as guests.
[...]

PR #9107, opened 2026-06-10
feat(dotcom): rework sidebar into a workspace switcher

MitjaBezensek:
Clean-history copy of #9086 (branch `mitja/rework-groups-sidebar`) — same final tree, restructured into reviewable commits so renames are separate from behavior changes:

1. `refactor(dotcom): rename group to workspace outside the db schema` — purely mechanical, no behavior change
2. `feat(dotcom): rework sidebar into a workspace switcher` — the actual feature
3. `refactor(dotcom): remove the add file link option`
4. `fix(dotcom): pin ordering and silent workspace failures`
5. `test(dotcom): rewrite groups e2e suite for the workspace sidebar`
6. `fix(dotcom): gate the sidebar restyle behind the workspaces flag`
7. `refactor(dotcom): rename workspace capabilities` — review feedback
8. `refactor(dotcom): rename hasWorkspaces flag check to workspacesEnabled` — review feedback

Behind the `groups_frontend` flag. The sidebar UI matches the Figma design; opening for UX review + a dotcom preview.

In order to simplify how users navigate their files and groups, this PR reworks the tldraw.com sidebar from the previous "all groups visible with inline expand/collapse" model into a single-selection workspace switcher.

- A fixed top region lists the spaces the user can switch between — `My files` (home group) and their workspaces — plus a `Create a workspace +` button.
- Selecting a space opens its most recent file. The active space is **derived from the open file** (its `owningGroupId`), so there is no new persistence — the highlighted space always matches the current file's group. (Whether clicking a workspace should open its recent file or only swap the file list is still an open product question; e2e tests deliberately only assert on file-list contents.)
- The lower region scrolls and shows the active space's files, pinned first and then grouped by recency (Today / Older / …). Per-item pin icons were removed in favour of the `Pinned` section label.
- Files move between spaces by dragging a file onto a space, or via the file `Move to` menu. The drag follows the file so its new space becomes active. `Move to` no longer shows checkboxes and hides the space the file is already in (a file lives in exactly one group).
- Selecting an empty workspace (or creating a new one) creates a file in it, so a space always lands you on a file. The `New file` button creates in whichever space is active.
- Removed the **"Add file link"** option entirely (dialog, sidebar menu item, app method, Zero mutator + tests, analytics event). It only inserted a `group_file` row for a file you don't own; no schema/migration change, existing linked files still render as guests.
- Styling pass to match the Figma: tokenised typography (row labels Inter Medium 14/22/1%, section subtitles 12/19.2, bold active-space title), `radius-1` (4px) hover pills, `--tl-color-low` sidebar border, full-width `--tl-color-divider`, and per-region padding so the sidebar owns no padding itself.
- The restyle is fully gated: the sidebar root carries `data-workspaces` (from `groups_frontend`), the shared base CSS keeps its previous values byte-for-byte, and the redesign's values apply only under `[data-workspaces='true']` — users without the flag see the sidebar exactly as on main. The preview-deploy auto-enable of `groups_frontend` was also removed, so previews need the flag on your user.

### Group → workspace rename

Everything is renamed to "workspace" except the DB schema vocabulary:

- User-facing copy (dialogs, invites, menus), component/file names (`WorkspaceSettingsDialog`, `CreateWorkspaceDialog`, `WorkspaceInviteHandler`, `TlaSidebarWorkspaceMenu`, …), hooks (`useActiveWorkspaceId`), CSS classes, menu/test ids, session-storage key, analytics event.
- `TldrawApp` methods (`getWorkspaceMemberships`, `getWorkspaceFilesSorted`, `getHomeWorkspaceId`, `copyWorkspaceInvite`, …).
- Capability names (`editWorkspace`, `deleteWorkspace`) and the capabilities/roles docs — code-only abstractions, never persisted; the role string stored in `group_user.role` is unchanged.
- Zero mutator names and payload keys (`createWorkspace`, `moveFileToWorkspace({ fileId, workspaceId })`, …), `queries.workspaceMemberships`, invite API response fields (`workspaceId`/`workspaceName`), `ZErrorCode.max_workspaces_reached`, `MAX_NUMBER_OF_WORKSPACES`.
- **Unchanged (schema vocabulary):** table names (`group`, `group_user`, `group_file`), column names (`groupId`, `owningGroupId`), schema types and relationship aliases (`TlaGroup*`, `.groupFiles`, `.groupMembers`), and the `groups_backend` / `groups_frontend` flags. Column keys inside table writes map explicitly, e.g. `group_file.insert({ groupId: workspaceId, … })`.

### Removed

- `TlaSidebarGroupItem` (inline-expandable group) and all expand/collapse state (`expandedGroups`, `ensureSidebarGroupExpanded`, `ensureFileVisibleInSidebar` — scrolling the active file into view is handled by the existing active-link effect), the group-drag/reorder machinery in `useDragTracking`, `pinIcon` asset, `AddFileLinkDialog`, dead sidebar CSS, and the old empty-group state copy.

### E2E tests

`groups.spec.ts` (written against the old expandable sidebar) is replaced by `workspaces.spec.ts` — 18 tests covering create/rename/delete workspace, switching, create-file-in-active-workspace, move via menu + drag, the `Move to` menu contents, pin/unpin, delete/duplicate, delete-active-file-stays-in-workspace, and the invite/sharing flows. The `Sidebar` fixture was rewritten for the new layout.

Also fixes a test-infra bug: `__test__prepareForTest` now deletes the user DO snapshot before rebooting, so the first page load after a database reset no longer resurrects deleted files (this was causing local e2e flakiness).

### Change type

- [x] `improvement`

### Test plan

1. Enable the `groups_frontend` flag for your user.
2. Confirm the sidebar shows `My files`, your workspaces, and `Create a workspace +` in a fixed top region, with the active space's files scrolling below.
3. Click a workspace and confirm it opens that workspace's most recent file and highlights it.
4. Drag a file onto another space → it moves there and the sidebar follows it. Dragging within the current space (incl. unpinning a pinned file) does **not** highlight the current space.
5. Open a file's `Move to` menu and confirm there are no checkboxes and the current space is hidden.
6. Create a workspace, open it while empty, and confirm a file is created.
7. Open a workspace's `…` menu / right-click menu and confirm it shows Copy invite link, Settings, Import file… (no "Add file link").
8. Confirm the group → workspace copy across the create / settings / invite / sign-in dialogs.

- [x] Unit tests
- [x] End to end tests


apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
```diff
@@ -106,16 +110,24 @@ export function FileItems({
 	const trackEvent = useTldrawAppUiEvents()
 	const copiedMsg = useMsg(messages.copied)
 	const hasAdminRights = useHasFileAdminRights(fileId)
-	const isPinned = useIsFilePinned(fileId, groupId ?? '')
-	const hasGroups = useHasFlag('groups_frontend')
+	const isPinned = useIsFilePinned(fileId, workspaceId ?? '')
+	const hasWorkspaces = useHasFlag('groups_frontend')
```
frolic: do we wanna rename the flag too? or is this flag likely to be short-lived?

apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarWorkspaceList.tsx
```diff
@@ -0,0 +1,203 @@
+import { ZErrorCode } from '@tldraw/dotcom-shared'
+import classNames from 'classnames'
+import { ContextMenu as _ContextMenu } from 'radix-ui'
```
frolic, line 3: ooc why is this aliased

apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
MitjaBezensek: I'd say we should migrate everyone then completely remove the legacy logic together with all the flags.

packages/dotcom-shared/src/mutators.ts
```diff
@@ -402,7 +405,7 @@ export function createMutators(userId: string) {
 			})
 			await tx.mutate.group_file.insert({
 				fileId,
-				groupId,
+				groupId: workspaceId,
```
frolic, line 408: do `group_file` and `groupId` need a rename?

edit: just saw the PR description says DB-shape stuff isn't done yet, what's the plan for that?

packages/dotcom-shared/src/mutators.ts
```diff
@@ -625,38 +630,42 @@ export function createMutators(userId: string) {
...
-					zql.group_user.where('groupId', '=', groupId).where('role', '=', 'owner')
+					zql.group_user.where('groupId', '=', workspaceId).where('role', '=', 'owner')
 				)
 				assert(owners.length > 1, ZErrorCode.forbidden)
 			}
 
-			await tx.mutate.group_user.update({ userId: targetUserId, groupId, role: targetRole })
+			await tx.mutate.group_user.update({
+				userId: targetUserId,
+				groupId: workspaceId,
```
frolic, line 640: same here, `group_user` -> `workspace_user`? and `groupId` -> `workspaceId`?

edit: just saw the PR description says DB-shape stuff isn't done yet, what's the plan for that?

frolic:
`capabilities.ts` and `roles.ts` may need adjustments to reflect new naming (these are all internal/only code abstractions so should be easy to migrate here)

packages/dotcom-shared/src/mutators.ts
MitjaBezensek: These are tied to our schema and table / column names. I made a ticket in notion for this, we should probably rename, but let's wait for to make sure this is the final name and also let's make sure zero will handle this change on our production db correctly.

apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
frolic: maybe a nitpicky thing but `hasWorkspaces` to me reads like "has 1+ records" rather than "is flag enabled", wondering about `enableWorkspaces` or `workspacesEnabled` or something?

apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarWorkspaceList.tsx
MitjaBezensek: We kinda have a convention that we use this for radix. It's then easier to distinguish our own context menu and other components (which we export from tldraw) vs the radix one.

apps/dotcom/client/src/tla/components/TlaFileMenu/TlaFileMenu.tsx
MitjaBezensek: Good shout, updating.

frolic (approved):

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -1188,8 +1192,18 @@ export class TldrawApp {
...
+			if (attempts++ > 100) {
+				this.toasts?.addToast({
+					severity: 'error',
+					title: 'Error accepting invite',
+					description: 'Please try again.',
+				})
+				this.navigate(routes.tlaRoot())
+				return
+			}
 			await sleep(50)
```
mimecuvalo: 50 seems on the small side - i might consider bumping to 100

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
```diff
@@ -785,6 +785,9 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 				.execute()
 		})
 
+		// Drop the stale user snapshot so the reboot pulls fresh post-reset state
+		// instead of resurrecting deleted files and groups.
+		await this.env.USER_DO_SNAPSHOTS.delete(getUserDoSnapshotKey(this.env, userId))
```
mimecuvalo, line 790: i want to check with this particular line change — this seems sensitive to add to this PR potentially.
can you explain this a bit?

mimecuvalo (approved):
pre-approving to not block, but i do have one comment i'd like to understand better

apps/dotcom/sync-worker/src/TLUserDurableObject.ts
MitjaBezensek: This is only used for tests, the test routes are gated behind `isDebugLogging(env)` and 
```ts
if (this.env.IS_LOCAL !== 'true') {
			return
}
```

We don't have proper e2e test isolation, so before each test what we rest the db for that user (`database.reset()`). This resets all the user's data (like delete all the files, groups the user has,..). But this snapshot still might get loaded and it contained stale data, so now we also delete the snapshot as part of this db reset.

Issue #8859, opened 2026-05-18
Ship user groups on tldraw.com


steveruizok:
## Overview

User groups (teams) on tldraw.com are substantially built behind two feature flags: `groups_backend` (data model and mutations) and `groups_frontend` (UI). The schema, sync-worker routes, migration path, sidebar UI, invite flow, file-move actions, and E2E coverage are all in place. This issue tracks the work to take groups from "behind flags" to "shipped to all users", with a soft gate during rollout.

## Goal

Ship user groups on tldraw.com to all users, gated to 5 members per group, with a contact path for larger groups.

## Current state

What's there:

- Schema: `group`, `group_user`, `group_file` tables and `file.owningGroupId` (`packages/dotcom-shared/src/tlaSchema.ts`)
- Migration: per-user and batch migration paths into the new model, including a "home group" with `id === userId` (`apps/dotcom/zero-cache/migrations/023_groups.sql`, `apps/dotcom/sync-worker/src/adminRoutes.ts`)
- Mutators: create/update/delete group, membership, file move/pin, all asserting `groups_backend` (`packages/dotcom-shared/src/mutators.ts`)
- Routes: `GET /app/invite/:token` and `POST /app/invite/:token/accept` (`apps/dotcom/sync-worker/src/routes/tla/`)
- UI: sidebar group section, group settings, invite dialog, drag-to-group, file menu actions behind `groups_frontend` (`apps/dotcom/client/src/tla/`)
- E2E: groups suite at `apps/dotcom/client/e2e/tests/groups.spec.ts` (~60 cases)
- Unit tests: mutator coverage at `packages/dotcom-shared/src/mutators.test.ts`
- Limits: `MAX_NUMBER_OF_GROUPS = 20` groups per user (`packages/dotcom-shared/src/constants.ts`)

What's missing for ship:

- Per-group member soft gate (5 members; contact path for larger)
- Resolution of known TODOs in `deleteGroup` and `UserDataSyncer` group iteration
- A full implementation + UX review pass
- A test pass beyond the existing E2E suite
- A rollout plan that removes the flags (or graduates them into a permissions check)

### Implementation review

- [ ] Review backend mutators and assertions in `packages/dotcom-shared/src/mutators.ts`
- [ ] Review sync-worker routes and durable-object methods for groups (`apps/dotcom/sync-worker/src/`)
- [ ] Review migration correctness for users with shared files, pinned files, and edge cases (`023_groups.sql`)
[...]

PR #9123, opened 2026-06-10
feat(dotcom): let owners remove group members

frolic:
In order to let group owners manage who's in their group, this PR adds the ability to remove a member from a group. It's part of the broader groups effort (#8859).

Members with the `editMembers` capability (currently owners) get an × button next to each other member's role in the group settings dialog. Clicking it opens a danger confirmation; confirming calls a new `removeGroupMember` mutator. The mutator enforces the `editMembers` capability and the "a group always keeps at least one owner" invariant, so the last owner can't be removed (the group must be deleted instead). You can't remove yourself this way — that path stays `leaveGroup`.

Removing a member who is currently viewing one of the group's files would otherwise leave them stranded on a file they can no longer access until they reconnect. To handle that, a small reactive guard (`useGroupFileAccessGuard`) watches the file the user is viewing and navigates them back to their files with a toast the moment their access disappears. This also covers the existing case of a group being deleted out from under other members.

### Concepts

| Term | Type | Meaning |
| --- | --- | --- |
| `removeGroupMember` | Zero mutator | Deletes a `group_user` row for a target member; requires `editMembers` and won't remove the last owner |
| `useGroupFileAccessGuard` | React hook | Navigates the viewer away (with a toast) when they lose access to the group file they're on |

### Change type

- [x] `feature`

### Test plan

1. As an owner of a group with at least one other member, open **Group settings**.
2. Confirm an × button appears next to each other member (but not next to yourself).
3. Click ×, confirm the dialog — the member is removed from the list.
4. As a non-owner (member), confirm no × button is shown.
5. Confirm the last owner cannot be removed.
6. As a member viewing one of the group's files, have an owner remove you — confirm you're navigated back to your files with an "Access removed" toast.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Group owners can now remove members from a group in group settings.

### Code changes

| Core code       | +31 / -0   |
| Tests           | +88 / -0   |
| Automated files | +44 / -0   |
| Apps            | +109 / -0  |

### DIFF
--- a/apps/dotcom/client/public/tla/locales-compiled/en.json
+++ b/apps/dotcom/client/public/tla/locales-compiled/en.json
@@ -189,10 +189,22 @@
       "value": "Please try refreshing the page."
     }
   ],
+  "33d8b74c4b": [
+    {
+      "type": 0,
+      "value": "You no longer have access to this file."
+    }
+  ],
   "344a7f427f": [
     {
       "type": 0,
       "value": "Editor"
+    }
+  ],
+  "375396d6e5": [
+    {
+      "type": 0,
+      "value": "Remove member"
     }
   ],
   "3bc7d6ea2f": [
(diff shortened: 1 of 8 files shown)


frolic:
going with a different strategy

PR #9126, opened 2026-06-10
fix(editor): prototype keyboard sink for iPad hardware-keyboard shortcuts

frolic:
In order to make Cmd/Ctrl keyboard shortcuts work on iPad Safari with a hardware keyboard ([#7372](https://github.com/tldraw/tldraw/issues/7372)), this PR prototypes a hidden contenteditable "keyboard sink" that holds DOM focus whenever the canvas is in its idle (non-text-editing) state. iOS Safari only delivers Cmd/Ctrl-modified key events to the page when an editable element (input, textarea, or contenteditable) holds focus — plain keys reach the focused `.tl-container` div, but Cmd chords are silently swallowed (verified with Appium `mobile: keys` probes on the iOS simulator; characterization test on #9101). The same applies to native `copy`/`cut`/`paste` clipboard events, which is how tldraw implements those shortcuts.

> **Stacked on #9069** (device e2e harness). This branch carries a device e2e test (adapted from #9101) that drives a real Cmd+C / Cmd+V through the OS keyboard pipeline and asserts a copy of the selected shape gets pasted. **Both device lanes pass**: [iOS Safari + Android Chrome run](https://github.com/tldraw/tldraw/actions/runs/27357166930). Re-trigger via the `e2e-device-tests` label or workflow dispatch.

### How it works

The sink is a visually hidden contenteditable div created by `FocusManager` as a child of `.tl-container`. Because it's a child, its key events bubble through the container and to the document, so both existing keyboard pipelines (`useDocumentEvents` on the container, `useKeyboardShortcuts` and clipboard handlers on the document) fire unchanged — only the focus target moves, in one place:

- `FocusManager.focus()` now focuses the sink instead of the container, and a `focusin` listener forwards the container's own focus to the sink, so every existing `getContainer().focus()` call site (escape-refocus, edit-exit, menus, style panel, skip-to-content) lands on the sink with no changes at those sites.
- A `pointerdown` listener hands focus to the sink on canvas interactions: desktop browsers focus the container natively on mousedown, but iOS Safari does not focus non-editable elements on tap.
- With `autoFocus`, the editor now takes real DOM focus on mount (previously it only set `instanceState.isFocused`); keyboard listeners on the document fire either way on desktop, but on iOS the sink must actually hold focus.

Two carve-outs keep the sink from being treated as a real text input, while real text inputs (shape text editing) keep suppressing shortcuts exactly as today:

- `elementShouldCaptureKeys` in `packages/editor` exempts elements marked `data-tl-keyboard-sink` (drives `areShortcutsDisabled`, clipboard gating, and `EditingShape.isTextInputFocused`).
- `shouldSkipEvent` in `useKeyboardShortcuts` exempts the sink before its contenteditable check.

The sink never takes focus while a shape is being edited: because it's contenteditable, focusing it would capture the DOM selection and route typed characters into the sink instead of the shape's text input. (The browser natively moves focus to the container on every mousedown, including the second mousedown of the double click that starts a text edit, so this guard matters in practice.)

Two findings from the device lanes shaped the sink itself:

- The sink keeps a **selected zero-width-space sentinel** inside it: iOS only fires the native `copy`/`cut` events (which tldraw's clipboard shortcuts ride) when the focused editable has a non-collapsed selection. The sentinel never reaches the clipboard — the sink suppresses its default copy/cut, and tldraw's handlers write the real payload via the async clipboard API.
- The sink is hidden with the clip-based "sr-only" recipe, deliberately **not `opacity: 0`**: iOS suppresses hardware-key delivery to editables it considers invisible (an `opacity: 0` input receives zero key events — established in #9101).

Sink hygiene: `inputmode="none"` so the iOS virtual keyboard doesn't appear for touch users, `beforeinput` is canceled (plus an `input` fallback that clears content) so typed characters never accumulate, `pointer-events: none` so it's never a pointer target, focus uses `preventScroll`, and it mirrors the container's `role="application"` and `aria-label` rather than being `aria-hidden` (focused elements must not be aria-hidden).

### Testing on iPad

**Preview:** **https://examples-git-frolic-keyboard-sink-prototype.tldraw.club** (Vercel examples preview for this PR)

On an iPad with a hardware keyboard:

1. Open the preview URL in Safari.
2. Draw a shape, then tap it to select it.
3. Cmd+C then Cmd+V should paste a copy; Cmd+A should select all; Cmd+Z should undo.
4. Double-tap a shape and type — text editing should work normally (the sink must not steal input).
5. As a touch-only user, tap around the canvas — the virtual keyboard should not pop up.

### Device e2e results

The diagnostics in the test trace every stage (key delivery → copy event → clipboard write → paste event → clipboard read). What they established:

- iOS Safari (simulator): with the sink focused, Cmd+C / Cmd+V keydowns and the native `paste` event now reach the page — they are silently swallowed without an editable focused, which is #7372's root cause. `copy` only fires once the sink holds a non-collapsed selection (hence the sentinel). With both in place the full copy → paste round-trip passes.
- Android Chrome (emulator): key delivery and clipboard events all work; the one hitch is Chrome's native permission dialog for the page's first `navigator.clipboard.read()`, which the test answers like a real user would. (Worth noting: tldraw's paste handler prefers `clipboard.read()` and only falls back to the paste event's `clipboardData` on rejection — a read that hangs on an unanswered permission prompt currently loses the paste even though the event data is present.)

### Known gaps and open questions

- The simulator is not a real iPad: #9101 documented that the sim swallows all Cmd chords on non-editables while real iPads deliver some — so the green lane is strong evidence for the mechanism, but a manual pass on a physical iPad with a hardware keyboard (preview URL above) is still the final word.
- While idle, the document selection lives inside the sink (the selected sentinel). The rich text toolbar only reads the selection during shape editing (when the sink never has focus), but other selection consumers should keep this in mind.
- The sink is active on all platforms rather than gated to iOS (`tlenv.isIos`), so desktop testing exercises the same code path that runs on iPad. Gating is a one-line change if we'd rather not touch desktop at all.
- `autoFocus` taking real DOM focus on mount is a behavior change for embedders (previously the container never received DOM focus until clicked).
- Screen reader behavior with the role/label-mirroring sink needs a proper a11y review.
- Pre-existing (also on `main`): keystrokes typed in the brief window between starting a text edit and the text input taking focus are dropped; the editing guard restores exact baseline behavior here but doesn't fix that window.

### Change type

- [x] `bugfix`

### Test plan

1. Desktop Chrome: draw, select, Cmd+C/Cmd+V/Cmd+A/Cmd+Z/tool shortcuts, text editing, style panel and menus — unchanged.
2. iPad Safari with hardware keyboard: steps above.
3. Device e2e: `e2e-device/tests/copy-paste-shortcut.e2e.ts` selects a shape and sends a real Cmd+C / Cmd+V chord through the OS keyboard pipeline (XCUITest `mobile: keys` / UiAutomator2 `mobile: pressKey`); passing on both lanes.

- [x] Unit tests
- [x] End to end tests

`packages/editor` (813) and `packages/tldraw` (2492) unit suites pass. Examples e2e suite passes locally except pixel-snapshot and text-measurement specs that fail identically on `main` on this machine (CI-recorded snapshots vs local font rendering), and were verified by running the same specs against a stashed baseline build.

### Release notes

- Fix Cmd/Ctrl keyboard shortcuts and clipboard actions on iPad Safari with a hardware keyboard.

### Code changes

| Section | Net LOC |
| --- | --- |
| Core code | +187 |
| Tests | +244 |


steveruizok:
Where did this one end up?

PR #9129, opened 2026-06-11
fix(dotcom): lock down the personal home workspace

MitjaBezensek:
In order to keep a user's home/personal workspace private, this PR enforces its special permissions server-side instead of only hiding them in the UI. The home workspace is the `group` row whose `id` equals the owning user's id; previously the only special-casing was `getRole()` treating the user as `owner`, which granted capabilities without restricting anything. Because the mutators re-run on the server as the authority, a crafted mutation could regenerate an invite secret, add members via an invite link, leave, delete, rename, or change roles in the home workspace — bypassing the sidebar UI that hid those actions.

This adds an `assertNotHomeWorkspace` guard (a home workspace is any group whose id matches an existing user id) to the `regenerateWorkspaceInviteSecret`, `updateWorkspace`, `deleteWorkspace`, `leaveWorkspace`, and `setWorkspaceMemberRole` mutators, and rejects home workspaces in the `acceptInvite` and `getInviteInfo` routes so an invite link can never resolve to one. Client-side defense-in-depth: `WorkspaceSettingsDialog` renders nothing for the home workspace and `copyWorkspaceInvite` refuses it.

### Change type

- [x] `bugfix`

### Test plan

1. Open the home workspace in the sidebar — confirm it has no workspace menu, settings, invite link, rename, leave, or delete.
2. Attempt each home-workspace mutation directly (regenerate invite, rename, delete, leave, change member role) — all are rejected by the server.
3. Visit an `/invite/<secret>` link that resolves to a home workspace — it is treated as invalid.

- [x] Unit tests

### Release notes

- Lock down the personal/home workspace: it can no longer be shared via an invite link, renamed, left, deleted, or have its members managed.

### Code changes

| Section    | LOC change |
| ---------- | ---------- |
| Core code  | +15 / -0   |
| Tests      | +52 / -10  |
| Apps       | +36 / -0   |

### DIFF
--- a/apps/dotcom/sync-worker/src/routes/tla/acceptInvite.ts
+++ b/apps/dotcom/sync-worker/src/routes/tla/acceptInvite.ts
@@ -5,7 +5,7 @@ import { sql } from 'kysely'
 import { createPostgresConnectionPool } from '../../postgres'
 import { Environment } from '../../types'
 import { requireAuth } from '../../utils/tla/getAuth'
-import { getJoinableGroupFromInvite } from '../../utils/tla/getJoinableGroupFromInvite'
+import { getJoinableWorkspaceFromInvite } from '../../utils/tla/getJoinableWorkspaceFromInvite'
 
 export async function acceptInvite(request: IRequest, env: Environment): Promise<Response> {
 	const { token } = request.params
@@ -21,10 +21,10 @@ export async function acceptInvite(request: IRequest, env: Environment): Promise
 
 	try {
 		return await db.transaction().execute(async (tx) => {
-			// First, validate the invite token and get group info
-			const group = await getJoinableGroupFromInvite(tx, token)
+			// First, validate the invite token and get workspace info
+			const workspace = await getJoinableWorkspaceFromInvite(tx, token)
 
-			if (!group) {
+			if (!workspace) {
 				return Response.json(
 					{
 						error: true,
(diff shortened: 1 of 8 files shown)


apps/dotcom/sync-worker/src/routes/tla/acceptInvite.ts
```diff
@@ -21,13 +22,9 @@ export async function acceptInvite(request: IRequest, env: Environment): Promise
 	try {
 		return await db.transaction().execute(async (tx) => {
 			// First, validate the invite token and get group info
-			const group = await tx
-				.selectFrom('group')
-				.select(['id', 'name', 'isDeleted'])
-				.where('inviteSecret', '=', token)
-				.executeTakeFirst()
+			const group = await getJoinableGroupFromInvite(tx, token)
```
frolic: should this be workspace?

packages/dotcom-shared/src/mutators.test.ts
```diff
@@ -1035,20 +1035,62 @@ describe('home workspace special case', () => {
...
+		}
+		return {
+			user,
+			file: [] as TlaFile[],
+			file_state: [] as TlaFileState[],
 			group: [makeGroup({ id: userId })],
-			group_user: [],
-			group_file: [],
+			group_user,
+			group_file: [] as TlaGroupFile[],
```
frolic: does `satisfies` work here instead of `as`? I find it generally ~bad to use `as` except in very rare cases where you're sure you want to force the type to a particular shape

(not super critical here but more the pattern, to let AI learn good patterns)

frolic (approved):
comment above isn't blocking, this looks good to me!

packages/dotcom-shared/src/mutators.test.ts
MitjaBezensek: Definitely better yeah.

PR #7076, opened 2025-11-07
feat(dotcom): incremental migration auto batching

ds300:
We hit subrequest limits on this so let's do it in smaller batches that the client initiates

### Change type

- [x] `other`

### Test plan

1. Run user migration from the admin panel and verify it batches correctly.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improved user migration performance by implementing client-initiated auto-batching to avoid subrequest limits.

---

> [!NOTE]
> Switch user migration to server-limited batches (50/users per run) and auto-chain batches on the client via SSE, with stop support and hasMore signaling.
> 
> - **Admin UI (`apps/dotcom/client/src/pages/admin.tsx`)**:
>   - Batch migration now runs in incremental batches: wraps SSE flow in `startBatch()` Promise and auto-starts next batch when `data.hasMore` is true.
>   - Adds `shouldContinueRef` to allow stopping between batches; updates stop logic and error handling.
>   - Maintains capped progress log and updates stats from SSE `details`.
> - **Sync Worker (`apps/dotcom/sync-worker/src/adminRoutes.ts`)**:
>   - `GET /app/admin/migrate_users_batch`: includes `hasMore` in completion events.
>   - `startUserMigration(...)`: processes up to `batchSize = 50` users per invocation, returns `boolean` indicating remaining work; adds `processedCount` loop and final remaining-users check.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 264c093edaf5e4a5755823a1ebec7bac83bc6a6e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7052, opened 2025-10-31
chore(dotcom): use a single sql connection for incremental migration

ds300:
after trying this in staging yesterday we noticed a spike in sql connections.

I had the connection hoisted to the parent process and cleaned up the migration script a little.

### Change type

- [x] `other`

### Test plan

1. Run the migration script in a staging environment and monitor SQL connections.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Optimize database connection usage during incremental user migration.

---

> [!NOTE]
> Switches batch user migration to a single-connection, sequential flow with per-user sleep, new stats/progress, and matching UI/API changes.
> 
> - **Backend (sync-worker)**:
>   - Replace batch migration with sequential per-user loop using one PG pool; add `sleepMs` param only.
>   - Introduce `startUserMigration`, `getNextUnmigratedUser`, `getNumUnmigratedUsers`, `getTotalUsers` helpers.
>   - Use SQL function `migrate_user_to_groups(...)` and force user DO reboot; send richer stats (`totalUsers`, `usersToMigrate`, `successCount`, `failureCount`, `progress`).
>   - Update SSE endpoint `/app/admin/migrate_users_batch` to emit new progress schema; `/app/admin/unmigrated_users_count` now returns DB count directly.
> - **Frontend (admin.tsx)**:
>   - Simplify Batch Migration UI: remove `batchSize` and `maxUsers`; add `sleepMs` only.
>   - Update confirmation text, help copy, and stats display to show `usersToMigrate` and percentage `progress`.
>   - Stream handling now directly sets `stats` from SSE `details`; updated controls to start/stop accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d86ab5ff7476aaceb15ba484d925c2858d8815be. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #9135, opened 2026-06-11
feat(dotcom): percentage-based workspaces UI rollout in admin panel

MitjaBezensek:
In order to roll out the workspaces UI to existing users, this PR reworks the admin panel's batch migration tool (previously used for the now-complete groups_backend data migration) into a percentage-based rollout of the groups_frontend flag (the workspaces UI).

The admin sets a target percentage and the batch converges to it from either side: raising the target enrolls users, lowering it unenrolls them (including users who got the flag by accepting a workspace invite — the confirm dialog warns about this). Reruns at the current target are no-ops.

Enrollment and unenrollment are plain SQL on the shared connection pool: chunks of 200 users per UPDATE, up to 2000 per SSE request, with the client chaining requests. No per-user durable object work is needed since clients pick flag changes up live through Zero (the same way acceptInvite grants the flag), which also keeps the process clear of the worker subrequest and connection limits the backend migration ran into (#7052, #7076). Unenrollment rebuilds the flag list via array operations, handling both comma and space separators.

The panel loads the current enrollment (X/Y users, share) from the server on mount and prefills the percentage input with the current share, so the default action is a no-op rather than a mass unenroll. The worker defaults to 0% when no percentage is passed. A failure stops the rollout and is surfaced in the admin UI instead of the success banner.

### Change type

- [x] `other`

### Test plan

1. Open the admin panel and verify the current enrollment count and prefilled percentage.
2. Raise the target and verify enrollment stops at the target; verify an enrolled user sees the workspaces UI without reloading.
3. Rerun with the same percentage and verify nothing changes.
4. Lower the target and verify users are unenrolled down to the target.

- [x] Unit tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Tests   | +109 / -0  |
| Apps    | +278 / -103 |

### DIFF
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -811,6 +811,8 @@ function RolloutGroupsFrontend() {
 		}
 	)
 	const [unmigratedCount, setUnmigratedCount] = useState<number | null>(null)
+	const [totalUsers, setTotalUsers] = useState<number | null>(null)
+	const percentageTouchedRef = useRef(false)
 	const [isLoadingCount, setIsLoadingCount] = useState(false)
 	const [eventSource, setEventSource] = useState<EventSource | null>(null)
 	const [sleepMs, setSleepMs] = useState(100)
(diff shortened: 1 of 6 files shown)


mimecuvalo (approved):
👍

PR #9137, opened 2026-06-11
feat(dotcom): default new users to the groups_frontend flag

MitjaBezensek:
> [!WARNING]
> Do not merge until we want new users to be fully on the workspaces experience — merging this puts every new signup on the groups UI from day one.

In order to show the groups UI to new signups by default, this PR adds the `groups_frontend` flag to the default flags for newly created users, copying the approach from #7093 which did the same for `groups_backend`. Both new-user creation paths are updated: the user durable object insert and the deprecated `init` mutator.

The e2e reset helper (`__test__prepareForTest`) intentionally keeps `groups_backend` only, since e2e tests grant `groups_frontend` explicitly via the database fixture.

An alternative to this PR would be a proper cleanup: remove the flags entirely, remove all branching based on them, and make the workspaces experience the only code path. This PR is the minimal stopgap; the cleanup is the eventual end state once all existing users are migrated.

### Change type

- [x] `improvement`

### Test plan

1. Sign up as a new user
2. Verify the user row has `groups_backend,groups_frontend` flags and the groups UI (workspace sidebar) is shown

### Release notes

- New tldraw.com users now get the groups UI by default.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +1 / -1    |
| Apps      | +1 / -1    |


mimecuvalo:
closing in favor of https://github.com/tldraw/tldraw/pull/9348

PR #9140, opened 2026-06-11
fix(editor): keep zoom level when yeeting camera after a right-click pan

### LINKED ISSUE #9138: Fast right-click pan zooms the camera to 800% on release
🗣️: there's a bug where on right-click and move (i.e. panning), if you move fast, when you release, it zooms all the way to 800%. it doesn't happen when it's slow movement then release. any idea why?

---

When panning the canvas with a right-click drag, releasing the mouse after a fast movement causes the camera to instantly zoom in to the maximum zoom level (800%) instead of gliding to a stop at the current zoom. Slow drags release normally because no inertia slide kicks in. Expected behavior: a fast right-click pan release should fling the camera in the pan direction with the zoom level unchanged, matching how left-button (spacebar) and middle-button pans behave. This is core editor camera behavior in the SDK, reproducible in any tldraw app.

## Open questions

1. **On release of a fast right-click pan, should the camera still glide with inertia (just without the zoom jump), or did you expect it to stop dead?**
   Keep the glide. The reporter hadn't consciously noticed the inertia effect before (only the zoom jump), which suggests it feels natural — and it matches the existing behavior of spacebar and middle-button pans.

Confidence: 95%, ready to get started. Fix in #9140.

jsscclr:
In order to stop fast right-click pans from zooming the camera to 800% on release, this PR passes a sanitized direction (`z: 0`) to `slideCamera` in the right-mouse-button pointer-up path. The raw pointer-velocity `Vec` was being passed directly, and `Vec.z` defaults to 1 because it doubles as pen pressure; since #7858, `slideCamera` interprets `direction.z` as a zoom direction, so the inertia fling multiplied the zoom every tick until it hit the max-zoom clamp. The left and middle button paths already sanitize the direction this way; this brings the right-click path in line. Closes #9138. The underlying hazard of `Vec.z` carrying different meanings across APIs is tracked separately in #9146.

### Change type

- [x] `bugfix`

### Test plan

1. Open any tldraw canvas.
2. Right-click and drag quickly to pan, then release while still moving.
3. The camera glides to a stop at the same zoom level instead of jumping to 800%.
4. Slow right-click pans and left/middle button (spacebar) pans behave as before.

- [x] Unit tests

### Release notes

- Fixed a bug where releasing a fast right-click pan zoomed the camera to the maximum zoom level.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +4 / -1    |
| Tests     | +11 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -11341,7 +11341,10 @@ export class Editor extends EventEmitter<TLEventMap> {
 									// Don't pass right-click panning events to the state chart
 									// as it causes unintended shape selection on release
 									if (slideSpeed > 0) {
-										this.slideCamera({ speed: slideSpeed, direction: slideDirection })
+										this.slideCamera({
+											speed: slideSpeed,
+											direction: { x: slideDirection.x, y: slideDirection.y, z: 0 },
+										})
 									}
 									this._selectedShapeIdsAtPointerDown = []
 									this._didCaptureSelectionAtPointerDown = false
--- a/packages/tldraw/src/test/rightClickPanning.test.ts
+++ b/packages/tldraw/src/test/rightClickPanning.test.ts
@@ -1,3 +1,4 @@
+import { Vec } from '@tldraw/editor'
 import { TestEditor } from './TestEditor'
 
 describe('with rightClickPanning enabled (default)', () => {
@@ -15,6 +16,16 @@ describe('with rightClickPanning enabled (default)', () => {
 		editor.expectCameraToBe(100, 100, 1)
 		editor.pointerUp(200, 200, { button: 2 })
 		expect(editor.inputs.getIsPanning()).toBe(false)
+	})
+
+	it('does not zoom when momentum panning on release', () => {
+		editor.pointerDown(100, 100, { button: 2 })
+		editor.pointerMove(200, 200)
+		expect(editor.inputs.getIsPanning()).toBe(true)
+		editor.inputs.setPointerVelocity(new Vec(1, 1))
+		editor.pointerUp(200, 200, { button: 2 }).forceTick()
+
+		expect(editor.getCamera().z).toBe(1)
 	})
 
 	it('does not pan on a static right-click', () => {


jsscclr:
@kostyafarber as our resident person who likes to pan using right-click

jsscclr:
~just noticed the inertia sliding isn't there so feels pretty gross compared to space + drag at the moment~

maybe not, just that right-click + pan on trackpad feels a little weird

kostyafarber (approved):

frolic:
the PR title in https://github.com/tldraw/tldraw/pull/7858 notes one finger zoom and touch devices, but this is happening on desktop, so I wonder if this points at a deeper issue where this gesture is being interpreted on a non-touch device?

frolic:
any other `slideCamera` usages to audit?

kostyafarber:
> any other `slideCamera` usages to audit?

there isn't many I checked, most just pass `slideDirection` with has the z component in it, for instance the

```
getPointerVelocity() {
	return this._pointerVelocity.get()
}
```

create a plain `new Vec` which defaults z to 1

jsscclr:
> the PR title in #7858 notes one finger zoom and touch devices, but this is happening on desktop, so I wonder if this points at a deeper issue where this gesture is being interpreted on a non-touch device?

The TLDR: `7858 widened slideCamera's parameter contract, and one pre-existing caller's vector carried a meaningful-looking z into it.`

- The one-finger zoom gesture is hard-gated to touch. It lives as a child state of the hand tool and is only entered from a double-tap-and-hold (settle-down) when isCoarsePointer is true (packages/tldraw/src/lib/tools/HandTool/HandTool.ts:21-22). A desktop right-click pan runs entirely in Editor.ts's raw pointer handling and never enters the hand tool's state machine — the gesture recognizer was never involved.
- What #7858 actually did to desktop: to implement the gesture, it widened the generic Editor.slideCamera() API so that direction.z means "zoom while sliding". That zoom branch is gesture-agnostic — it runs for any caller passing a nonzero z, on any device.
- The bug was a stale caller leaking into that widened contract. The right-click release path passed the raw pointer-velocity Vec, whose z defaults to 1 (it doubles as pen pressure). So the fling was unintentionally requesting zoom at full rate every tick. Notably, #7858's own call sites were sanitized from day one (Dragging.ts passes z: 0, OneFingerZooming.ts passes z: ±0.01), as was the left/middle-button path — only the right-click path passed an unsanitized Vec.

PR #9141, opened 2026-06-11
docs(examples): add page panel example with live page thumbnails

angrycaptain19:
Adds an example showing how to replace the built-in page menu with a custom page sidebar rendered in external DOM. Demonstrates the reactive editor.getPages() API, page management (switch, create, rename, delete), and rendering live thumbnails of non-current pages via getSvgString, since the canvas only renders the current page.

### DIFF
--- a/apps/examples/src/examples/ui/page-panel/PagePanelExample.tsx
+++ b/apps/examples/src/examples/ui/page-panel/PagePanelExample.tsx
@@ -1,0 +1,200 @@
+import { useEffect, useState } from 'react'
+import {
+	Editor,
+	EditorProvider,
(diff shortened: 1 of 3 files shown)


kaneel (approved):
works like a charm

mimecuvalo:
@angrycaptain19 i have some prior art here if interested https://github.com/tldraw/tldraw/pull/4181
(this was Google Jamboard's treatment which i liked)

angrycaptain19:
> #4181

Oh interesting! Goal here was to have a core example showing how to use the page menu after a user request for a simple example. Might just leave here as is and we could potentially extend with a follow on example?

PR #9142, opened 2026-06-11
feat(dotcom): seed new workspaces with a welcome file

### LINKED ISSUE #9143: Seed new workspaces with a welcome file
🗣️: When creating a new workspace there should be a single file called "Welcome to tldraw". The file should have a simple canvas that explains how to invite people or move files etc. It should be easily renameable like any other file. The messaging should be shapes on the canvas — you should be able to edit them, delete them, etc.

---

A new workspace should greet its members with a ready-made first file instead of a blank canvas, introducing workspaces (inviting your team, moving files in, shared visibility) without an onboarding modal. The welcome content is ordinary canvas shapes and the file is an ordinary file: renameable, editable, deletable. Scope was clarified during implementation: the file is created at the moment a workspace's first file comes into existence — when someone clicks into an empty workspace — not at workspace-creation time, and the file name settled on "Welcome to your workspace". The personal "My files" space keeps its blank first file. Implemented in #9142 (draft). A workspace emptied of all files currently gets re-seeded on next select; that is a known interim behavior, not the desired end state, and the eventual solution is being worked on outside that PR.

## Open questions

1. **Is the canvas content (copy and layout) right?**
   _The current draft canvas is in #9142 — this is the main piece still being designed._
2. **Should a workspace that is emptied of all files get a fresh welcome file the next time it's selected, or should the welcome file be seeded only once per workspace?**
   _Re-seeding every time is the current behavior but not the desired end state. The eventual solution (seeding only the genuinely-first file) is being worked on separately and is deliberately not answered definitively in #9142._

jsscclr:
In order to introduce workspaces without an onboarding modal, this PR seeds a new workspace's first file with a welcome canvas instead of a blank file. Creating a workspace (behind the `groups_frontend` flag) immediately creates its first file and opens it; the file is seeded from a template baked into the sync worker: a three-panel comic in the brand's illustration style, reusing the Downy/Cloudy/Penta characters and hand-drawn UI mockups from the workspaces presentation — what a workspace is, how to invite your team (including revoking a shared link), and how to move files in. The file is an ordinary file — it can be renamed, edited, and deleted, and "My files" keeps its blank first file with inline rename.

Closes #9143.

> [!NOTE]
> Draft: the mechanism is settled; the canvas content (copy, art, layout) is the piece under design review. It is easy to iterate — the template is regenerated by designing on a canvas and re-exporting (instructions in `templates/newWorkspace.ts`).

How it works:

- A new `createSource` kind, `template/<templateId>` (`TEMPLATE_PREFIX` in `dotcom-shared`), resolved by `TLFileDurableObject.handleFileCreateFromSource` from a worker-side registry. Unlike the other `createSource` kinds it is not backed by another room — the snapshot ships in the worker bundle.
- The `new-workspace` template was designed on a real canvas and exported (`editor.store.serialize('document')` plus the serialized schema captured at export time, so records migrate on load like any `.tldr` file). Exports are scrubbed of the authoring user's identity (`user` records, `textFirstEditedBy`), and a test fails if a future re-export forgets, or if the records stop migrating/validating under the current schema.
- A `TldrawApp.createWorkspace` method creates the workspace plus its seeded file and is shared by both creation flows (the sidebar dialog, which navigates straight to the file, and a file's 'move to new workspace' menu). Clicking into an empty workspace remains as a fallback that seeds the same file (guarded against double-activation), skipping the inline rename since the file arrives named.
- An unknown template id (deploy skew, a retired template — ids live forever in `createSource`) degrades to an empty room with a reported error instead of permanently failing the room's first load, and the registry lookup ignores prototype-chain keys.
- The canvas is sized to fit the default camera on a typical editor viewport, since fresh files open without a saved camera.

A workspace emptied of all its files gets a fresh welcome file the next time it is selected (the trigger is "first file of an empty workspace", not "first file ever"). This is known interim behavior rather than the desired end state — the once-only solution is being worked on separately and is deliberately not addressed in this PR (see #9143).

### Change type

- [x] `feature`

### Test plan

Preview deploy: https://pr-9142-preview-deploy.tldraw.com/

**Setup**

1. Open https://pr-9142-preview-deploy.tldraw.com/ and sign in.
2. Turn on the workspaces UI for your account: go to https://pr-9142-preview-deploy.tldraw.com/admin, search for your email, and enrol yourself in groups (the `groups_frontend` flag).

**The change**

3. Create a workspace from the sidebar dialog.
4. ✅ You land directly on a file named "Welcome to your workspace" with the three-panel welcome canvas, no inline-rename prompt, and the new workspace active.
5. ✅ The file is ordinary: rename it, edit and delete its shapes, delete the file.
6. With the workspace now empty, click it in the sidebar: a fresh welcome file is created (the fallback path; known interim behavior, see #9143).
7. Open a file's menu → 'Move to' → create a new workspace: the file moves there and the workspace also contains a welcome file (you stay on the moved file).

**No regression**

8. Empty "My files" and click it: it still creates a blank date-named file with the inline rename open.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Core code       | +11 / -0    |
| Tests           | +66 / -0    |
| Apps            | +2287 / -6  |

### Release notes

- Add a welcome file to new workspaces: clicking into an empty workspace now creates a "Welcome to your workspace" canvas introducing workspaces, invites, and moving files, instead of a blank file.

Screenshot 2026-06-12 at 14 28 41


jsscclr:
@mimecuvalo @MitjaBezensek curious about your thoughts on this approach since AFAICT it's a bit of a new concept to start with a seeded file?

apps/dotcom/sync-worker/src/templates/templates.test.ts
```diff
@@ -0,0 +1,105 @@
...
+		for (const doc of snapshot.documents) {
+			expect(doc).toMatchObject({ lastChangedClock: 0 })
+			expect(doc.state.id).toBeDefined()
+		}
+	})
+
+	// The template's records were exported under the schema version current at export time.
+	// They must keep migrating and validating under whatever schema the repo is on now,
+	// since that is exactly what happens when the snapshot is loaded into a room.
+	it('migrates and validates under the current schema', () => {
```
mimecuvalo, line 38: i would suggest that this snapshot is always up-to-date instead of can-it-be-updated:
```
import { readFileSync } from 'node:fs'
import { fileURLToPath } from 'node:url'

const TEMPLATE_FILE = fileURLToPath(new URL('./newWorkspace.ts', import.meta.url))

it('newWorkspace.ts is baked at the current schema (run `yarn test -u` to re-bake)', async () => {
  const schema = createTLSchema()

  // migrate every record up to head + revalidate, then re-anchor the stored schema to head
  const documents = snapshot.documents.map(({ state, lastChangedClock }) => {
    const r = schema.migratePersistedRecord(state, snapshot.schema!, 'up')
    if (r.type !== 'success') throw new Error(`record ${state.id} failed to migrate`)
    schema.types[r.value.typeName].validate(r.value)
    return { state: r.value, lastChangedClock }
  })
  const rebaked = { ...snapshot, schema: schema.serialize(), documents }

  // embed the pretty-printed JSON back into the template literal, escaped for backticks/${ }/\
  const json = JSON.stringify(rebaked, null, '\t')
    .replace(/\\/g, '\\\\').replace(/`/g, '\\`').replace(/\$\{/g, '\\${')

  // replace only the literal; keep the file's header/imports as they are on disk
  const regenerated = readFileSync(TEMPLATE_FILE, 'utf8').replace(
    /export const newWorkspaceTemplateJson = `[\s\S]*`\n?$/,
    'export const newWorkspaceTemplateJson = `' + json + '`\n'
  )

  await expect(regenerated).toMatchFileSnapshot(TEMPLATE_FILE)
})
```

running `yarn test -u` would help with updating

mimecuvalo (approved):
pre-approving modulo one comment — thx!

MitjaBezensek:
> @mimecuvalo @MitjaBezensek curious about your thoughts on this approach since AFAICT it's a bit of a new concept to start with a seeded file?

I like it!❤️

I think it opens up a new way to give the app more character. Makes me think we could do something similar for new users: we generate a new file for them, that one could also contain something about how to interact with files, how to share them,..

The only thing that gives me pause is that this might be more of a one off thing and not a general way we could let people know about what's new. In this case it's easy to slot it into the workspace flow, but probably wouldn't apply to many other situations - which might be fine. Just something to think off, like would we want some other system for other types of changes, how do they interact?

Maybe we could have a permanent what's new file, which has this kind of content? This would make it more general. The users would only see it once and it only pops up again when we make a change to it so that it's not too intrusive.

jsscclr:
> The only thing that gives me pause is that this might be more of a one off thing and not a general way we could let people know about what's new. In this case it's easy to slot it into the workspace flow, but probably wouldn't apply to many other situations - which might be fine. Just something to think off, like would we want some other system for other types of changes, how do they interact?

It's a good point! Right now it's a very permanent solution to only a single use case... 

I was thinking a short term / hacky solution would be a duplication-driven flow. Would be more brittle, but we'd copy from an existing document that exists. Avoids having to build something permanent at this point in time.

PR #9148, opened 2026-06-11
perf: improve performance when resizing many shapes

steveruizok:
In order to keep the canvas responsive when resizing thousands of shapes at once, this PR removes the per-tick bottlenecks across the resize interaction, the history system, the store's change pipeline, validation, and sync. Relates to #7943.

Profiling a 2000-shape resize showed most of the frame budget going to O(N) work and allocations that ran on every pointer tick:

- The `Resizing` tool committed each shape's resize as its own `updateShapes` call. It now collects partials via a new internal `Editor.getResizeShapePartial` and commits them in a single batch per tick. Rotated-out-of-alignment shapes keep the old immediate path.
- `HistoryManager`'s pending diff recomputed emptiness from the full accumulated diff on every store change. Since updates can never cancel existing entries during a squash, it now derives emptiness incrementally and only does the full check when a diff adds or removes records. It also skips clearing the redo stack when the stack is already empty.
- `squashRecordDiffsMutable` allocated per entry per tick (`Object.entries` pair arrays, a fresh `[from, to]` tuple per updated record, unconditional deletes against empty collections). The hot loop is now allocation-free: `for…in` iteration, in-place tuple mutation (the target copies tuples on first insert so it exclusively owns them), and emptiness-guarded deletes. In a micro-benchmark of a 2000-shape, 60-tick gesture this is 3.5x faster on average and removes a 10ms worst-case GC tick.
- `isRecordsDiffEmpty` and `Store.filterChangesByScope` used `Object.keys(...).length === 0`, which materializes every key to compare against zero. Both now use an early-exit `for…in` check.
- `squashHistoryEntries` squash-copied every same-source chunk on each frame flush even when the chunk held a single entry; single-entry chunks now pass through untouched.
- `TLSyncClient.push` deep-cloned the first diff of each send window via `structuredClone` just to get an owned accumulator. It now squashes into a fresh empty diff: owned containers and tuples, shared record references. Records are immutable by construction (`devFreeze` on every store write), and every consumer of the unsent diff (`getNetworkDiff`, the rebase replay via `applyObjectDiff`) is read-only or copy-on-write.
- `ObjectValidator` walked its config object with a `hasOwnProperty` check per key per validation and allocated a closure per property for error prefixing. It now iterates a cached entries array and uses a plain try/catch with a shared rethrow helper.

### Change type

- [x] `improvement`

### Test plan

1. Create a large number of shapes (e.g. 2000).
2. Select all and resize via a corner or edge handle; the interaction should stay smooth.
3. Undo and redo the resize; verify the document returns to the correct states.
4. Repeat in a multiplayer room and verify changes sync correctly, including resizing while offline and reconnecting.

- [x] Unit tests

### Release notes

- Improve performance when resizing many shapes at once.

### API changes

- Added internal `Editor.getResizeShapePartial()`, which returns the update for a resized shape without committing it, so interactions can batch many resizes into one store transaction.

### Code changes

| Core code       | +396 / -301 |

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1348,6 +1348,8 @@ export class Editor extends EventEmitter<TLEventMap> {
     _getReferencedUserIds(shapes: TLShape[]): Set<string>;
     getRenderingShapes(): TLRenderingShape[];
     getResizeScaleFactor(): number;
+    // @internal
+    getResizeShapePartial(shape: TLShape | TLShapeId, scale: VecLike, opts?: TLResizeShapeOptions): null | TLShapePartial;
     getRichTextEditor(): null | TiptapEditor;
     getSelectedShapeAtPoint(point: VecLike): TLShape | undefined;
     getSelectedShapeIds(): TLShapeId[];
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -8102,32 +8102,53 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 * @public
 	 */
 	resizeShape(shape: TLShapeId | TLShape, scale: VecLike, opts: TLResizeShapeOptions = {}): this {
+		const partial = this.getResizeShapePartial(shape, scale, opts)
+		if (partial) this.updateShapes([partial])
+		return this
+	}
+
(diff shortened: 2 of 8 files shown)


steveruizok:
Closed in favor of smaller PRs

PR #9151, opened 2026-06-11
perf(store): squash record diffs without per-entry allocations

steveruizok:
🙇‍♂️ This shook out of sending Fable down a "make resizing faster" perf hunt. The perf claims here are hard for me to validate; I'm at the limit of my knowledge about how JS engines optimize. I'm relying on the complexity of our test suite to catch any problems.

---

In order to keep per-tick store bookkeeping cheap when many records change at once (e.g. resizing thousands of shapes), this PR removes the allocations and O(N) work in the store's diff-squashing pipeline and in the sync client's unsent-changes accumulator. Split out from #9148; relates to #7943.

- `squashRecordDiffsMutable` allocated per entry per call: `Object.entries` pair arrays, a fresh `[from, to]` tuple per updated record, and unconditional deletes against collections that were usually empty. The hot loop is now allocation-free: `for...in` iteration, in-place tuple mutation, and emptiness-guarded deletes. In-place mutation is safe because the target copies tuples on first insert, so it exclusively owns them and source diffs are never mutated. In a micro-benchmark of a 2000-shape, 60-tick gesture this is 3.5x faster on average and removes a 10ms worst-case GC tick.
- `isRecordsDiffEmpty` and `Store.filterChangesByScope` used `Object.keys(...).length === 0`, which materializes every key just to compare the count against zero. Both now use an early-exit `for...in` check (`hasAnyKey`).
- `squashHistoryEntries` squash-copied every same-source chunk on each flush even when the chunk held a single entry; single-entry chunks now pass through untouched.
- `TLSyncClient.push` deep-cloned the first diff of each send window via `structuredClone` just to get an owned accumulator. It now squashes into a fresh empty diff: owned containers and tuples, shared record references. Records are immutable by construction (`devFreeze` on every store write), and every consumer of the unsent diff (`getNetworkDiff`, the rebase replay via `applyObjectDiff`) is read-only or copy-on-write. This change depends on the tuple-copy behavior above, which is why both are in this PR.

### Change type

- [x] `improvement`

### Test plan

1. Make many changes in a multiplayer room and verify they sync correctly, including while offline and after reconnecting.

- [x] Unit tests

### Release notes

- Improve performance of change tracking and sync when many shapes change at once.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +52 / -29  |

### DIFF
--- a/packages/sync-core/src/lib/TLSyncClient.ts
+++ b/packages/sync-core/src/lib/TLSyncClient.ts
@@ -12,7 +12,6 @@ import {
 	exhaustiveSwitchError,
 	isEqual,
 	objectMapEntries,
-	structuredClone,
 	uniqueId,
 } from '@tldraw/utils'
 import {
@@ -853,10 +852,11 @@ export class TLSyncClient<R extends UnknownRecord, S extends Store<R> = Store<R>
 		// in offline mode, we only accumulate in speculativeChanges
 		if (!this.isConnectedToRoom) return
 		if (!this.unsentChanges.nextDiff) {
-			this.unsentChanges.nextDiff = structuredClone(change)
-		} else {
-			squashRecordDiffsMutable(this.unsentChanges.nextDiff, [change])
+			this.unsentChanges.nextDiff = { added: {} as any, updated: {} as any, removed: {} as any }
 		}
+		// records are immutable, so sharing their references with `change` is fine — the
+		// squash gives nextDiff its own containers and tuples without deep-cloning records
+		squashRecordDiffsMutable(this.unsentChanges.nextDiff, [change])
 		this.sendUnsentChanges()
 	}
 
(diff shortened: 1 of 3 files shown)


packages/sync-core/src/lib/TLSyncClient.ts
```diff
@@ -853,10 +852,11 @@ export class TLSyncClient<R extends UnknownRecord, S extends Store<R> = Store<R>
 		// in offline mode, we only accumulate in speculativeChanges
 		if (!this.isConnectedToRoom) return
 		if (!this.unsentChanges.nextDiff) {
-			this.unsentChanges.nextDiff = structuredClone(change)
-		} else {
-			squashRecordDiffsMutable(this.unsentChanges.nextDiff, [change])
+			this.unsentChanges.nextDiff = { added: {} as any, updated: {} as any, removed: {} as any }
```
steveruizok, line 855: Don't worry about the `as any`, we have this pattern in many other places in this file.

PR #9154, opened 2026-06-11
perf(draw): optimize the freehand ink algorithm

steveruizok:
In order to make the freehand ink algorithm faster and its rendered output smaller without changing how strokes look, this PR replaces the implementation in `packages/tldraw/src/lib/shapes/shared/freehand` with an optimized one.

https://github.com/user-attachments/assets/9bbe9265-09c2-47ff-8b3d-00f8e78713a8

This is a clean-history reimplementation of [`ink-performance-improvements`](https://github.com/tldraw/tldraw/tree/ink-performance-improvements) (#9094): the final state is identical to that branch, rebuilt as a sequence of self-contained commits for review. The optimization was developed there against a frozen copy of the previous algorithm, with each round validated for bit-identical output over a stroke corpus; the benchmark harness and baselines remain available in the original branch's history.

### How the optimized implementation gets there

- The internal pipeline runs entirely in one module-level set of reusable struct-of-arrays buffers (`core.ts`): `Float64Array`s for coordinates, pressure, distance, running length, and radius. Raw input is ingested and streamlined straight into the arrays with no per-point objects, radii are computed in a single pass with the taper pass folded in, and outline tracks are built into flat arrays with an in-place simplifier.
- Svg path data is written into a shared buffer (`fmt.ts`) with integer math: coordinates are tracked as integer hundredths of a pixel, and paths use relative commands (`t`/`q`/`a`/`l`) whose small integer deltas write fast, stay compact, and sum exactly so rounding never drifts.
- Stroke points no longer store a per-point unit vector: consumers derive each point's vector from its predecessor with scalar math, eliminating a `Vec` allocation and normalization per input point. `svgInk`'s elbow partitions are index ranges into the pipeline buffers and carry a "vector anchor" so points following an elbow keep the direction they had in the uncut stroke.
- Outline tracks are simplified with a tolerance of 5% of the stroke size, and cap arcs are tessellated adaptively by chord error instead of fixed step counts.
- The public functions are thin adapters over the array core with unchanged names and signatures; `DrawShapeUtil`, `HighlightShapeUtil`, `getPath`, and the scribble overlays import the algorithm from the same local module as before.
- The editor's `Vec` no longer imports `EASINGS`: the two easings it used as defaults are inlined as local functions.

### Results

Measured head-to-head against the original algorithm over the full corpus during development (interleaved ops/sec for the full render pipeline, best of 5 rounds; see #9094 for per-case results):

| Bucket | Cases | Speedup (geomean) |
| --- | --- | --- |
| Short (<50 input points) | 8 | 1.59x |
| Medium (50-500) | 15 | 1.91x |
| Long (500+) | 5 | 2.67x |
| **Overall** | **28** | **1.92x** |

Svg path data is 57% smaller overall, no case is slower, and shape deviation vs the original is bounded at 0.26px (5% of the stroke size, the simplifier's hard bound) with a mean of 0.018px.

### Change type

- [x] `improvement`

### Test plan

1. `yarn workspace tldraw test run src/test/freehand` - corpus-driven svg well-formedness checks and svg path unit tests
2. Draw and highlight in the examples app with solid and pattern fills, thin to extra-thick sizes, and a stylus if available; strokes should look identical to production

- [x] Unit tests

### Release notes

- Improve freehand ink performance: draw and highlighter strokes compute roughly 2x faster (up to 3x for long strokes) and produce 57% smaller svg path data, with no visible change to how strokes look.

### API changes

- Breaking! Removed `StrokePoint.vector`. The vector is derivable from consecutive points; nothing in the SDK consumed it outside the stroke pipeline itself.
- Breaking! Removed `setStrokePointRadii`. Radius and taper computation now happens inside the pipeline; `getStroke` and `svgInk` handle it, and the highlight shape (the last external caller) now uses `getStroke` directly.
- No other public API changes: `tldraw` exports the same stroke functions and types with the same signatures, and the point types are still the editor's `Vec`.

### Code changes

| Core code       | +1512 / -596 |
| Tests           | +541 / -0    |
| Automated files | +0 / -5      |


steveruizok:
The LOC change here is a little misleading. It's an add for sure, but freehand ink is both the most expensive "common" interaction in tldraw and the place where dropped frames are most consequential, so I think it's a good idea to squeeze this code as much as we can (even at the expense of some powerful AI cleverness).

packages/editor/src/lib/primitives/Vec.ts
```diff
@@ -1,5 +1,8 @@
 import { VecModel } from '@tldraw/tlschema'
-import { EASINGS } from './easings'
+
+const easeInQuad = (t: number) => t * t
+const easeInOutQuad = (t: number) => (t < 0.5 ? 2 * t * t : -1 + (4 - 2 * t) * t)
```
mimecuvalo: hm, why copy out the fns from `easings.ts`? doesn't seem useful to duplicate?

mimecuvalo (commented):
@steveruizok how can we help out with reviewing this and/or rolling it out? Looks like you had some parity/harness tests around this to ensure it didn't regress much. Is there anything else we need to specifically to check to make sure this doesn't degrade the experience going forward?

packages/editor/src/lib/primitives/Vec.ts
steveruizok: At one point this was in a separate package and these were the only dependencies between the editor and the freehand code, so I inlined it. Not really needed anymore, I'll revert.

steveruizok:
This is pretty well tested and everything seems to be good across browsers. Did a quick test of the ink (deleting 1200 draw shapes and then undoing). On my machine, the frame went from 170ms to 90ms. So the overhead is real. I imagine on less powerful computers this will have a more serious impact.

PR #9155, opened 2026-06-11
perf: replace lodash micro-packages and remove dead code to shrink bundle

steveruizok:
🙇‍♂️ Spike on reducing bundle size. Any wins from here that we like can be moved into other PRs.

---

In order to reduce the size of the bundled SDK, this PR replaces the vendored lodash micro-packages with compact local implementations (relates to #6827), deletes dead code, generates repetitive data tables, and consolidates repeated patterns across the editor, tldraw, and validate packages. Behavior is unchanged: the full test suite (6,795 tests), typecheck, and api-check all pass. It also documents a vetted catalogue of further opportunities for follow-up PRs (see "Future opportunities" below).

Measured by bundling `tldraw`'s entry with esbuild (minify, `react`/`react-dom` external) and gzipping at level 9:

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| Minified | 1,709,232 B (1669.2 KB) | 1,674,650 B (1635.4 KB) | **−33.8 KB (−2.0%)** |
| Minified + gzip | 518,549 B (506.4 KB) | 511,346 B (499.4 KB) | **−7.0 KB (−1.4%)** |

Net source LOC: `packages/editor` −578, `packages/tldraw` −412, `packages/validate` −20, `packages/utils` +367 (the local lodash replacements) — core source down 643 lines net, plus 379 lines of new unit tests.

### What changed

**Replace lodash micro-packages with local implementations (−25.6 KB minified).** `@tldraw/utils` re-exported `lodash.isequal` (10.2 KB), `lodash.isequalwith` (9.1 KB), `lodash.throttle` (1.9 KB), and `lodash.uniq` (4.4 KB) — the first two being near-duplicate copies of the same deep-equality code. They are now implemented locally in ~340 lines total, sharing one recursive comparison engine, with the same semantics at every call site (NaN equality, unordered Map/Set comparison, cyclic structures, leading/trailing throttle edges, `cancel()`/`flush()`). 379 lines of new unit tests cover them, and the existing validation fuzz tests exercise `isEqual` against randomized JSON. `apps/dotcom/sync-worker` imported `lodash.throttle` directly and now uses the `@tldraw/utils` version, dropping its own lodash dependency too.

**Generate the default theme data (−430 lines, −4.4 KB minified).** `defaultThemes.ts` spelled out 4 font families × 4 faces and 2 modes × 13 colors × 14 variants as object literals, so each of the 14 variant key names shipped 26 times in the bundle. The fonts are now generated from a 4-line pattern and each color is one `c(...)` row with the variant keys written once. The rows were generated programmatically from the original object and verified deep-equal (the only delta: two `draw` font faces gain an explicit `style: 'normal'` where it was previously omitted — identical CSS either way).

**Delete dead code (−~110 lines).** Removed functions and components that nothing imports and that are not part of the public API: `getArrowLength`, `getArrowHandlePath`, `getLegacyOffsetX`, `clampToRange`, `doRangesOverlap`, `rangeCenter`, `perpDistanceToLine`, `isActiveTLUiToolItem`, `PentagonToolbarItem`, `TldrawUiButtonSpinner`, `ToggleDarkModeItem`, and the editor's `uniq` wrapper. A mechanical scan of all non-public exports across the 8 SDK packages (editor, tldraw, tlschema, store, validate, state, sync-core, utils) found no further unused symbols — symbol-level dead code is now exhausted.

**Consolidate repeated patterns (−~640 lines).**
- `Editor.ts`: the 3-line shapes-to-ids coercion was copy-pasted 22 times, now a `toShapeIds` helper; the horizontal/vertical axis-key mapping appeared 3 times across `stackShapes`/`distributeShapes`/`stretchShapes`, now an `AXIS_MAPPINGS` table.
- `actions.tsx`: the 15 align/distribute/stretch/flip/pack/stack actions shared an identical guard → track → mark → run-and-kickout body, now one `selectionShapesAction` helper; the export/copy-as fallback logic, the four zoom actions, and the triplicated paste-error toast are similarly factored out.
- `menu-items.tsx`: the 13 simple toggle checkbox items render through a single internal `ToggleCheckboxItem` component; each public component keeps its exact `export function X()` signature.
- `validation.ts`: the 4-line ValidationError path-wrapping block was inlined 8 times in production catch handlers (plus once in `prefixError`); all now call one `rethrowWithPath` helper. The intentionally-inlined try/catch structure (which avoids closure allocation in hot loops) is preserved — only the cold error path is shared.
- `buildFromV1Document.ts`: the 36-line "apply v1 label then grow square geo shapes" block was byte-identical across the Rectangle/Triangle/Ellipse cases, now one helper.
- `ArrowShapeUtil.tsx`: the four-branch scale-flip ladder in `onResize` reduces exactly to two flip flags; `curved-arrow.ts`: the start/end intersection filter-and-sort blocks (identical except a 0.25/0.75 fraction) now share `pickIntersectionInArc`; `getElbowArrowInfo.tsx`: the four identical usable-edge object literals now share `getUsableEdges`.
- `StyleEmbedder.ts`: the two 95%-identical computed-style extraction loops (StylePropertyMap vs CSSStyleDeclaration) now share one loop parameterized by accessor.
- `TldrawUiMenuCheckboxItem.tsx`: the dropdown and context-menu branches rendered identical JSX through different Radix components, now one branch with the component selected by menu type.
- `BoundsSnaps.ts`: `getResizeSnapPointsForHandle`'s four copy-pasted switches became a corner table loop; `useClipboardEvents.ts`: the copy/cut guard triple became one predicate.

### Future opportunities (vetted, not changed here)

Larger refactors that look worthwhile but deserve their own PRs:

- **SelectTool child states (~380–520 lines).** Ten `childStates` files implement identical `onCancel`/`onComplete`/`onInterrupt` triplets; four duplicate a 12-line `onInteractionEnd` handler in both `complete()` and `cancel()`; seven repeat the tool-id-mask setup/teardown; three (`Translating`, `Resizing`, `DraggingHandle`) share a 15-line mark-id initialization block and a shape-callback iteration loop. A shared base class (e.g. `PointingStateNode`) would absorb most of this, but changing the inheritance hierarchy of the core interaction code deserves isolated review.
- **tlschema per-shape boilerplate (~300 lines).** All 14 shape files repeat the same interface + props-validator + version-ids + migration-sequence scaffolding; a `defineShape()`-style factory could collapse it while keeping every existing export. Migrations themselves must stay byte-identical (append-only).
- **`InputsManager` accessor facade (~200 lines).** 28 atom-backed getter/setter pairs plus deprecated property facades. Mechanically generable, but the explicit declarations are what produce the typed public API surface, so any change shows up heavily in the API report.
- **`TLInstance` parallel key maps (~60 lines).** `shouldKeyBePreservedBetweenSessions`, `ephemeralKeys`, and the defaults in `pluckPreservingValues` maintain the same ~30 keys in three places; a single source-of-truth table would prevent drift.
- **`HistoryManager` undo/redo symmetry (~40 lines).** `_undo()` and `redo()` share ~70% of their stack-walking logic; a parameterized replay is possible but mark-hunting semantics need careful tests.
- **`DefaultKeyboardShortcutsDialogContent` (~40 lines).** A 250-line static JSX list that could be data-driven.
- **Excalidraw color table (~30 lines)** in `putExcalidrawContent.ts`, and the **elbow-arrow route table** (where `routeRightToBottom` is provably `FlipY(routeRightToTop)`) could shed dedicated route functions.
- **Lazy-loading the rich text stack.** `@tiptap/*` + prosemirror is ~250 KB minified (15% of the bundle) and the nine `@radix-ui` packages are ~70 KB more — by far the biggest remaining wins, but they change loading behavior, so they're out of scope for a behavior-identical PR.

Things that look like wins but aren't:

- `collectGapSnaps` in `BoundsSnaps.ts` looks like a horizontal/vertical copy-paste (~135 lines each), but the halves are asymmetric: the vertical branch `continue`s after a center match while the horizontal one falls through to the duplication checks, and they use `rangeIntersection` vs `rangesOverlap` for the breadth-overlap test. Either this is intentional or one side has a latent bug — worth investigating, but consolidation is not behavior-preserving-by-construction.
- `NoteShapeUtil` and `GeoShapeUtil` label measurement is similar in spirit but algorithmically different (notes shrink the font iteratively, geo shapes grow), so no shared helper was extracted.
- The 33 tiny `XToolbarItem` components and the menu-item wrapper trio (`TldrawUiMenuActionItem` etc.) could be factory-generated, but each is an individually-exported public component — generating them would churn the API report and TS declarations for ~1 line saved each.
- `PentagonToolbarItem` was tagged `@public` but never exported from the package index (unlike `RhombusToolbarItem`, `HeartToolbarItem`, etc.). This PR removes it as dead code, but exporting it may have been the original intent.
- `watermarks.ts` (11 KB) is generated license-enforcement data; `defaultTranslation.ts` (19.6 KB) is the required English fallback; `defaultEmbedDefinitions.ts` derives public literal types from its array shape — all off limits.

### Change type

- [x] `improvement`

### Test plan

1. Select shapes and use align/distribute/stretch/flip/pack/stack, reorder, and zoom actions; verify identical behavior and history marks.
2. Export/copy as SVG/PNG/JSON with and without a selection.
3. Toggle preference menu items (snap mode, grid, focus mode, etc.) in both the main menu and context menu, and verify checkbox state tracks the preference.
4. Verify light/dark theme colors and fonts render identically (the default theme object was verified deep-equal to the previous literal).
5. Resize an arrow shape across both axes (flip horizontally, vertically, and both); draw curved arrows bound to closed and open shapes; drag elbow arrows between shapes.
6. Import a v1 `.tldr` file containing labeled rectangles, triangles, and ellipses.
7. Hover shapes (throttled hover updates) and resize the window (throttled viewport bounds) to exercise the new throttle.

- [x] Unit tests

### Release notes

- Reduce the minified bundle size by ~34 KB by replacing the `lodash.isequal`, `lodash.isequalwith`, `lodash.throttle`, and `lodash.uniq` dependencies with compact local implementations, generating repetitive theme data, and removing dead code.

### API changes

- `isEqual`, `isEqualWith`, `throttle`, and `uniq` in `@tldraw/utils` are now local implementations with explicit signatures instead of lodash re-exports; added `EqualityCustomizer` and `ThrottledFunction` types. `ThrottledFunction` replaces lodash's `DebouncedFunc` in the types of the internal `pruneTombstones`/`pruneSessions` members in `@tldraw/sync-core`.
- `uniq` in `@tldraw/editor` is now re-exported from `@tldraw/utils` (identical signature).

### Code changes

| Core code       | +988 / −1631 |
| Tests           | +379 / −1    |
| Automated files | +29 / −20    |
| Apps            | +1 / −2      |
| Config/tooling  | +1 / −75     |


mimecuvalo:
ach, i'd have great hesitation about rolling our own versions of this — i already wanna maybe consider nixing our own homerolled async throttle function. it's not much win for a cognitive load on our part going fwd

PR #9160, opened 2026-06-12
refactor(dotcom): rename groups to workspaces in the backend

frolic:
In order to finish the group → workspace rename started in the UI layer (#9107), this PR renames the dotcom backend data model: the Postgres tables `group`/`group_user`/`group_file` become `workspace`/`workspace_user`/`workspace_file`, the `groupId` and `file.owningGroupId` columns become `workspaceId` and `owningWorkspaceId`, and everything named after them follows — zero schema tables/relationships, `TlaGroup*` types, replicator topics, SQL functions, triggers, indexes, and constraints.

### Migrations

**Postgres (`035_rename_groups_to_workspaces.sql`)** renames the tables, columns, indexes, constraints, and triggers, and re-creates every plpgsql function whose body references a renamed table (function bodies are stored as text, so renames don't propagate into them). Table renames keep their OIDs, so the `zero_data` publication membership, FKs, and trigger bindings follow automatically. `migrate_user_to_groups` is renamed to `migrate_user_to_workspaces` with a thin compatibility wrapper kept under the old name, because the previous sync-worker deployment keeps calling it during the few minutes between the migration running and the new worker shipping (the deploy script runs migrations → zero → sync-worker → client, in that order).

**Replicator (sqlite migration `008_rename_group_topics_to_workspace`)** wipes `history` and `active_user` and nulls the stored lsn. History rows store serialized changes under the old table/column names and `group:` topic prefixes, which new code can't replay — so instead every user DO is forced to re-register and refetch from postgres through the existing sequence-break machinery. This also repairs any updates silently dropped during the deploy window, when the old worker's wal2json table filter no longer matches the renamed tables. Existing `topic_subscription` rows are rewritten from `group:` to `workspace:` so edges for users who don't reconnect promptly don't linger with a dead prefix.

**Stale sessions** are handled by bumping `Z_PROTOCOL_VERSION`/`MIN_Z_PROTOCOL_VERSION` to 4 (old clients get `CLIENT_TOO_OLD` and the refresh prompt) and bumping the user DO snapshot `stateVersion` to 5 (v4 R2 snapshots are discarded rather than migrated, since the forced refetch rebuilds them from postgres anyway).

### Intentionally not renamed

- The `user.flags` values `groups_backend`/`groups_frontend` — persisted per-user rollout markers, written from several code paths. Rewriting them races against the not-yet-deployed worker and buys nothing structurally.
- The admin enroll/unenroll routes and DO methods (`admin_enrollInGroups` etc.) that manage those flags.
- Historical migration files (023, 031, 033, 034) — applied history stays untouched.

### Verification

The full migration chain (000 → 035) was run against a scratch Postgres, then smoke-tested: `migrate_user_to_workspaces` and the `migrate_user_to_groups` wrapper migrate a legacy user correctly; the home-workspace name triggers, the unshare guest-cleanup trigger, and the workspace soft-delete cascade all fire correctly against the renamed tables. The opt-in `delete_file_states` suite (which now extracts the function definition from 035) passes against real Postgres.

### Change type

- [x] `other`

### Test plan

1. Existing coverage: dotcom-shared (77), sync-worker (123), dotcom client (133), zero-cache (14, incl. the real-Postgres trigger suite) all pass; `yarn typecheck` is clean.
2. `fetchEverythingSql.snap.ts` regenerated via `UPDATE_SNAPSHOTS=1`.

- [x] Unit tests

### Code changes

| Core code       | +183 / -157 |
| Tests           | +469 / -404 |
| Automated files | +33 / -33   |
| Apps            | +668 / -229 |
| Config/tooling  | +3 / -3     |

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/Database.ts
+++ b/apps/dotcom/client/e2e/fixtures/Database.ts
@@ -61,7 +61,7 @@ export class Database {
 		if (!id) throw new Error('User not found')
 
 		// Call the migration function
-		await sql`SELECT * FROM migrate_user_to_groups(${id}, ${inviteSecret})`.execute(db)
+		await sql`SELECT * FROM migrate_user_to_workspaces(${id}, ${inviteSecret})`.execute(db)
 	}
 
 	/**
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -13,9 +13,9 @@ import {
 	TlaFileState,
 	TlaFileStatePartial,
 	TlaFlags,
-	TlaGroup,
-	TlaGroupFile,
-	TlaGroupUser,
+	TlaWorkspace,
+	TlaWorkspaceFile,
+	TlaWorkspaceUser,
 	TlaMutators,
 	TlaSchema,
 	TlaUser,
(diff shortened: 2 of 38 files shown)


frolic:
Just wanted to see what was involved in making this change, but gonna tackle this another day when we can do this more thoughtfully.

PR #9162, opened 2026-06-12
fix(editor): forgets previous touch interactions when entering pen mode

### LINKED ISSUE #9145: Stray line connects palm touch point to stylus when switching into pen mode on iPad
🗣️: on ipad when using a stylus before you're in pen mode there is a problem where you can touch the tablet with you hand and the pen and then as it switches to pen mode it creates a line connecting where your hand was touching to the stylus

---

On a tablet, before the editor has switched into pen mode, touching the canvas with your hand and then the stylus produces an unwanted line: as the editor detects the stylus and switches into pen mode, the draw stroke begins at the point where the hand was touching and connects to the stylus position, so the palm contact becomes the first point of the same stroke the stylus goes on to draw. The expected behavior is that the hand touch is treated as a palm and ignored, so the stroke begins only where the stylus is. Reported on tldraw.com on iPad, and SDK users have reported the same on both iPad and Android, so the fix belongs in the SDK rather than the app.

## Open questions

1. **Where are you seeing this — tldraw.com on iPad Safari, the examples app, or another SDK app?**
   On tldraw.com on iPad, but we have had bug reports from SDK users on Android and iPad as well.
2. **Does the gesture order matter — does it only happen when your hand touches before the stylus, or also when they land at roughly the same time?**
   Hand first — the hand/palm touches the canvas before the stylus does.
3. **Is the stray segment part of the same draw stroke you go on to make, or does it appear as a separate line?**
   Same stroke — the line from the palm point is connected to the stroke the stylus then draws.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9145

"On a tablet, before the editor has switched into pen mode, touching the canvas with your hand and then the stylus produces an unwanted line: as the editor detects the stylus and switches into pen mode, the draw stroke begins at the point where the hand was touching and connects to the stylus position, so the palm contact becomes the first point of the same stroke the stylus goes on to draw. The expected behavior is that the hand touch is treated as a palm and ignored, so the stroke begins only where the stylus is. Reported on tldraw.com on iPad, and SDK users have reported the same on both iPad and Android, so the fix belongs in the SDK rather than the app."

We simply cancel previous touches when entering pen mode.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Put a finger on touch screen
2. Bring the pen on the canvas
3. Start drawing

The drawing should start where the pen first hit the canvas, not from the previous finger touch.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug when user enter pen mode while a finger/hand is on the touch screen device. We cancel previous touches and the line starts where the pen first touched the screen.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -11191,7 +11191,12 @@ export class Editor extends EventEmitter<TLEventMap> {
 						inputs.setIsDragging(false)
 
 						// If pen mode is off but we're not already in pen mode, turn that on
-						if (!isPenMode && isPen) this.updateInstanceState({ isPenMode: true })
+						if (!isPenMode && isPen) {
+							this.updateInstanceState({ isPenMode: true })
+							// Once pen mode is on, touch input is ignored, so we discard the
+							// in-progress touch interaction .
+							this.interrupt()
+						}
 
 						// On devices with erasers (like the Surface Pen or Wacom Pen), button 5 is the eraser
 						if (info.button === STYLUS_ERASER_BUTTON) {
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
👍

PR #9163, opened 2026-06-12
fix(dotcom): bound concurrency when copying file assets to R2

### LINKED ISSUE #9109: maybeAssociateFileAssets exceeds Cloudflare's connection limit when copying file assets
The sync worker's `maybeAssociateFileAssets` background job copies a file's assets to new object names by running an R2 get + put per asset inside a single `Promise.allSettled`, with no bound on concurrency. On files with many assets this exceeds Cloudflare's per-request simultaneous connection limit and the copies fail with `Error: Response closed due to connection limit`. The job should stop hitting the limit, e.g. by bounding its concurrency. (The Sentry grouping problem originally tracked here has been removed from this issue's scope.)

## Sentry context

[TLDRAW-SYNC-Q0](https://tldraw.sentry.io/issues/TLDRAW-SYNC-Q0): ~1.14M events from ~750 users since December 2025, still ongoing at roughly 20k events per day (~50 users/day). All events are marked handled — the rejections are swallowed by `Promise.allSettled` — so the visible symptom is asset copies silently failing inside the job rather than user-facing errors. The stack points at the per-asset `env.UPLOADS.put(asset.newObjectName, currentAsset.body, …)` fired from the `Promise.allSettled` in `TLFileDurableObject.maybeAssociateFileAssets`.

frolic:
In order to stop the sync worker's `maybeAssociateFileAssets` job from exceeding Cloudflare's simultaneous connection limit (closes #9109), this PR runs the per-asset R2 get + put copies through a `p-queue` bounded at three copies at a time instead of firing them all at once inside `Promise.allSettled`. On files with many assets the unbounded version fails with `Response closed due to connection limit` (Sentry issue TLDRAW-SYNC-Q0, ~1.14M events), with the rejections swallowed by `Promise.allSettled` so the copies just silently fail.

Two details worth calling out:

- The concurrency is 2 rather than the suggested 4–6 because each copy holds two connections at once — the R2 `get` body streaming into the `put` — and Workers allow six simultaneous open connections. Two copies use four slots and leave two free for everything else the durable object does while a pass runs (snapshot persistence to R2, Pierre pushes, Postgres queries), all of which share the same budget since the DO is a single I/O context. Saturating all six would put the persistence path in contention with the pass, and a saturated pool is the regime where the runtime reclaims in-use connections — the failure mode behind the Sentry error.
- `maybeAssociateFileAssets` now runs at most one pass at a time, via an `ExecutionQueue` that is skipped when busy. It's fired without awaiting from room start, restore, and every persist tick (every 8 seconds), and a bounded queue on a large file can outlast the persist interval — without the guard, each tick would re-scan and re-enqueue copies for every asset still waiting in the queue, stacking overlapping passes and defeating the bound. Assets that show up while a pass is running are picked up by the next persist tick, same as before.

`p-queue` is ESM-only, which is fine here: `@tldraw/dotcom-worker` is a private `type: module` app bundled by esbuild, not a published package. The worker bundle stays well under the size limit (1.36 MB of 1.5 MB).

### Change type

- [x] `bugfix`

### Test plan

1. In a dev/staging environment, duplicate a file containing many image assets (enough that unbounded copies would previously exceed the connection limit).
2. Verify every asset in the duplicate gets a new object name and `meta.fileId`, and that no `Response closed due to connection limit` errors are reported.

### Release notes

- Fix file asset copies silently failing on files with many assets.

### Code changes

| Apps           | +44 / -22  |
| Config/tooling | +25 / -0   |


apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -1068,12 +1076,22 @@ export class TLFileDurableObject extends DurableObject {
...
 	// We use this to make sure that all of the assets in a tldraw app file are associated with that file.
 	// This is needed for a few cases like duplicating a file, copy pasting images between files, slurping legacy files.
 	// Also migrates old-format asset URLs to tldrawusercontent.com.
+	// Only one pass runs at a time: copying many assets can take longer than the persist interval,
+	// and each pass rescans the whole store, so overlapping passes would re-copy the same assets.
+	// Calls made while a pass is running are dropped; the next persist tick picks up anything
+	// that arrived in the meantime.
 	async maybeAssociateFileAssets() {
 		if (!this.documentInfo.isApp) return
+		if (!this.associateAssetsQueue.isEmpty()) return
```
MitjaBezensek, line 1198: I guess if we bail out we might catch some left over assets on the next run?

MitjaBezensek (approved):

mimecuvalo (changes requested):
blocking for a sec, i'm gonna add some more changes here to address the related `_uploadSnapshotToBucketMultipart` issue

mimecuvalo:
i'm excited for this PR btw - let's park it until next week - i don't want the workspace rollout to be complicated by making sure this is working correctly after deploy 👍

mimecuvalo (approved):
unblocking and approving modulo one last gut check from @MitjaBezensek 
marking as dotcom-hotfix-pls as well

MitjaBezensek (approved):

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
frolic: Good instinct — it usually would get caught on the next run, but there's a gap: a pass only scans the store once up front, and a persist tick can advance `_lastPersistedClock` past assets that arrive *mid-pass*. So if the file then goes idle, those stragglers wouldn't get re-triggered until the next document change or a room restart.

Pushed 87278ea to close it: instead of dropping the call while a pass is running, it sets a pending flag and re-runs the pass once it finishes, picking up anything that showed up mid-pass. It coalesces, so it's at most one extra pass no matter how many ticks fire during a long copy.

Sent from my aiPhone

frolic:
sorry @mimecuvalo, I didn't love how the queue depth stuff was sprinkled around so I refactored it a bit as a closure

still not super happy with it but feels a bit better/more encapsulated

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -1150,6 +1176,15 @@ export class TLFileDurableObject extends DurableObject {
 		}
 	})()
 
+	// Submits an R2 operation to the shared connection budget (see r2Queue). Snapshot uploads run at a
+	// higher priority than asset copies so persistence isn't stuck behind a large association pass's
+	// backlog of copies — durability shouldn't wait on background asset work.
+	private addR2Operation<T>(type: R2OperationType, task: () => Promise<T>) {
+		return this.r2Queue.add(this.trackQueuedTask(type, task), {
+			priority: type === 'snapshot_upload' ? 1 : 0,
```
mimecuvalo, line 1188: @frolic and @MitjaBezensek added this to make sure snapshots take priority over assets

steveruizok:
Is this one still in play?

mimecuvalo:
> Is this one still in play?

yes, we've been jamming on it. prbly going out this week

mimecuvalo:
@MitjaBezensek + @frolic 
I think let's try landing this tomorrow (Thu) and give it a go. it's been baking for quite a while

mimecuvalo:
i take back my prev. comment — too busy this week, we'll tackle next week during deploy

driev (approved):
I think this will resolve the issue, but the new thumbnails/MCP endpoints and queue/R2 operations might cause contention again... some global helper functions for all outbound connections might be required at some stage?

mimecuvalo:
> I think this will resolve the issue, but the new thumbnails/MCP endpoints and queue/R2 operations might cause contention again... some global helper functions for all outbound connections might be required at some stage?

good shoutout 👍 @driev can you take on investigating that as a further improvement?

PR #9166, opened 2026-06-12
Remember user last used styles

### LINKED ISSUE #8932: Persist last-used styles across sessions on tldraw.com
## Problem

On tldraw.com, the next-shape styles (size, color, fill, dash, font, etc.) reset to their defaults every time a whiteboard is loaded. A user who prefers drawing at size `S` has to change the size every session — same for any other preferred style.

Reported by a user:

> Could you please add a setting for a default pen size when opening a whiteboard? I prefer S however every time i load a whiteboard it automatically resets to M

This is the same underlying ask as the closed/stale #1634 (text alignment).

## Scope

**Dotcom feature, not an SDK change.** The SDK already exposes everything needed (`setStyleForNextShapes`, `getStyleForNextShape`, `stylesForNextShape` on `TLInstance`). The missing piece is dotcom-side persistence / a user preference UI. SDK consumers who want different default-style behavior can already wire it up themselves.

## Why it resets today

`stylesForNextShape` lives on `TLInstance` and is explicitly marked as non-preserved between sessions:

```ts
// packages/tlschema/src/records/TLInstance.ts
stylesForNextShape: false, // does not preserve because it's a temporary state
```

So on every reload it falls back to each style's `defaultValue` (e.g. `'m'` for `DefaultSizeStyle`).

## Proposed solution

Persist the user's last-used next-shape styles across sessions on tldraw.com. Options to consider:

1. **Auto-remember** — dotcom snapshots/restores `stylesForNextShape` (and `opacityForNextShape`) per user, so whatever the user last picked is what they get on next load. Lowest friction, no UI needed.
2. **Explicit preference** — add a "default styles" section in the tldraw.com user preferences menu where users pick their preferred starting size / color / fill / etc.
3. **Hybrid** — auto-remember by default, with a "reset to defaults" action in preferences.

Scope should not be limited to pen size — apply to all per-tool style props (size, color, fill, dash, font, opacity, geo, arrowhead, spline, etc.) so the same fix covers future similar requests.

## Considerations

- Decide whether memory is per-document or per-user (per-user matches the reporter's expectation).
- Should not override styles when loading a snapshot that explicitly sets them.
- Keep the SDK default behavior unchanged so embedders aren't surprised.

## Related

- #1634 — Remember text alignment (closed, stale) — same pattern for a single style
- #8877 — Add pressure preferences to user preferences — similar 'make this a real user preference' framing

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8932 

Adds an user preference for always remembering the last used styles (and opacity) so refreshing the page, or visiting from a different machine will always remember what's the last used colours and styles.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [x] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape with a funky colour and some fills. 
2. refresh browser
3. drag drop a new geo and see if styles match

Do that again moving to a different machine (and being logged). 
And try this while being unlogged.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- The draw can now remember your last used styles when visiting from different machines or simply after a refresh


steveruizok:
Are these universal or scoped to each project? It would be easy I think to persist on device the user's last tool / styles for next shape. I wouldn't make this an option, I think we should try to figure out what the right UX is.

kaneel:
> Are these universal or scoped to each project? It would be easy I think to persist on device the user's last tool / styles for next shape. I wouldn't make this an option

it's stored for the user. 

> I think we should try to figure out what the right UX is.

That's why I kept it as a draft, I'm not even sure this is the right way, maybe a simple store into localStorage is enough and indeed, make it with no option.

mimecuvalo:
I think @MitjaBezensek was championing this b/f btw https://github.com/tldraw/tldraw/issues/8932
maybe not a user pref, just a general dotcom improvment

steveruizok:
Probably not this one, lets look at defaults when we work on app-wide user preferences.

PR #9168, opened 2026-06-12
feat(dotcom): rework the manage workspace dialog

### LINKED ISSUE #9131: Unify dropdown styling (role dropdowns differ from style panel; Firefox renders them unstyled)

frolic:
In order to give workspace owners a clearer place to manage their workspace, this PR reworks the "Workspace settings" dialog into a tabbed "Manage workspace" dialog and adds an enable/disable toggle for the invite link. Follow-up to #9106; closes #9131.

The dialog now has a shared header (name + invite link) above two tabs:

- **Members** — each member shows a role dropdown; owners can change roles or remove members. Your own row's action is **Leave**; it (and removing the last owner) is disabled with a tooltip when it would leave the workspace without an owner.
- **Settings** (owners only) — enable/disable the invite link, regenerate it, leave, or delete the workspace.

Other changes:

- **Capabilities** — the per-action capabilities (`manageInvites`, `editMembers`, `editWorkspace`, `deleteWorkspace`) are consolidated into a single owner-only `manageWorkspace`. Members can no longer manage the invite link, roles, name, or deletion.
- **Invite link toggle** — like file sharing, turning the link off *keeps* it (so it can be turned back on) rather than clearing it. Backed by a `group.inviteLinkEnabled` column (migration 035), a mutator, and gating on the invite-accept path; the user durable-object state version is bumped so clients refetch.
- **SDK dialog fix** — `TldrawUiDialog` now closes on a backdrop click only when the press both starts and ends on the overlay, and no longer auto-dismisses when a portaled popover opened inside it (e.g. a Select) is used. Without this, opening/dismissing the role dropdown tore down the whole dialog.

https://github.com/user-attachments/assets/a9571f8d-546f-42b6-add7-eee1cfdca956

### Change type

- [x] `feature`

### Test plan

1. As an **owner**, open Manage workspace → **Members**: every member has a role dropdown; your own row's dropdown has "Leave" (disabled with a tooltip if you're the only owner), other members have "Remove".
2. **Settings** tab: toggle "Enable invite link" off → the copy button becomes a disabled "Invites are disabled"; toggle on → the same link returns. Regenerate / Leave / Delete each open a confirm dialog (descriptive title, short action button).
3. As a **member**: roles are read-only, your own row offers "Leave", and there's no settings management.
4. Open a role dropdown, then click elsewhere in the dialog or on the backdrop: the dropdown closes; the dialog only closes on a deliberate backdrop click (and Escape).

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix a dialog unexpectedly closing when a popover (e.g. a select) opened inside it is dismissed.

### Code changes

| Core code       | +64 / -39   |
| Tests           | +90 / -15   |
| Automated files | +102 / -75  |
| Apps            | +373 / -302 |


jsscclr:
Some comments, don't think they're blocking but might be good to revisit soon

- This is currently a no-op, whilst 'leave' is disabled 
Screenshot 2026-06-16 at 15 02 22


- Could we add `font-variant-numeric: tabular-nums;` for nicer numbers here: Screenshot 2026-06-16 at 15 07 36
Screenshot 2026-06-16 at 15 08 31


- Visual spacing between tabs is a little off - too much on the second tab in particular:
Screenshot 2026-06-16 at 15 10 57
Screenshot 2026-06-16 at 15 11 08
Screenshot 2026-06-16 at 15 13 06
Screenshot 2026-06-16 at 15 13 20

The static height between dialogs feels wasteful, especially in the non-owner case - lots of remaining whitespace without much use

apps/dotcom/client/src/tla/components/dialogs/WorkspaceSettingsDialog.tsx
```diff
@@ -156,13 +185,38 @@ export function WorkspaceSettingsDialog({ workspaceId, onClose }: WorkspaceSetti
...
+			console.error('Error removing member:', error)
+			app.showMutationRejectionToast((error as Error).message as ZErrorCode)
+		}
+	}
+
+	const openRegenerateConfirmDialog = () => {
+		addDialog({
+			component: ({ onClose }) => (
+				<ConfirmDialog
+					title={<F {...messages.regenerateInviteLink} />}
```
mimecuvalo: fwiw, you can just do `<F defaultMessage="Text" />` — actually my preference would be to not put everything into a `messages` grabbag whenever possible so people don't have to jump back & forth a lot

mimecuvalo (approved):
pre-approved!

PR #9176, opened 2026-06-12
test(sync-core): rebuild unit tests around a behavior specification

steveruizok:
In order to make the sync-core package's unit tests an expression of the library's actual rules — repeating what #9158 did for the state package and #9172 did for the store package — this PR adds a behavior specification at `packages/sync-core/SPEC.md` (183 numbered rules across 29 sections, derived from the implementation) and rebuilds the test suite around it. Each test name cites the rule ID it expresses (for example `[SS5]`, `[RP7]`), so the mapping between rules and tests is greppable in both directions: two tests citing the same rule with the same scenario are redundant, and a rule with no citing test is a coverage gap. Every rule in the spec currently has at least one citing test.

**Fixes: bugs surfaced while writing the spec** (one commit each)

- **An error thrown while handling a client message now rejects the session.** `TLSocketRoom.handleSocketMessage` wraps message handling in a try/catch that rejects the session with `UNKNOWN_ERROR`, but `TLSyncRoom.handleMessage` is async and was called without awaiting, so a non-`TLSyncError` thrown while handling a message (e.g. a message with an unknown type) escaped the catch as an unhandled promise rejection and the session stayed connected. The promise now carries a rejection handler that does what the catch block always intended. (The catch path previously only worked for synchronous throws, like the chunk assembler's JSON parse errors.)
- **The flushed data-message array is handed to the socket instead of truncated in place.** `TLSyncRoom._flushDataMessages` sent the live `outstandingDataMessages` array and then cleared it with `length = 0`, so any `TLRoomSocket` implementation that defers serialization observed an empty `data` array — both in-repo test sockets had to `structuredClone` to work around it. The flush now hands the buffer over and starts a fresh one.

The `@tldraw/sync-core`, `@tldraw/sync`, dotcom sync-worker, and bemo-worker suites all pass with these fixes.

**Storage asymmetries documented as contract**

Writing the spec surfaced that the two storage backends disagree at construction time, now specced explicitly rather than papered over: `InMemorySyncStorage` clamps the document clock up to the max `lastChangedClock`/tombstone clock in the snapshot, clamps `tombstoneHistoryStartsAtClock` down to the document clock, and discards tombstones when the history window is empty (IM1–IM3); `SQLiteSyncStorage` takes all three verbatim (SQ3). Behavior after construction is identical and is covered by a single shared contract suite.

**Tests: merge the parallel suites**

The package had parallel per-backend storage suites (`InMemorySyncStorage.test.ts` and `SQLiteSyncStorage.test.ts` were ~80% the same tests, copy-pasted). They merge into one storage contract suite (`src/test/storageContractSuite.ts`, the SS rules) registered by both backend test files, with implementation-specific rules tested separately. Also merged or moved:

- `NodeSqliteSyncWrapper.test.ts` + `NodeSqliteSyncWrapper.integration.test.ts` → `NodeSqliteWrapper.test.ts` (named after its source file; the integration file's storage-level coverage was redundant with the contract suite)
- `computeTombstonePruning.test.ts` folds into `InMemorySyncStorage.test.ts` (the function lives in that source file); `diff.test.ts` and `chunk.test.ts` move from `src/test/` to sit alongside their sources
- the `validation.test.ts`, `customMessages.test.ts`, and `presenceMode.test.ts` scenario files fold into the room and client suites
- `RoomSession.test.ts` (asserted constant ordering and built object literals) and `server-types.test.ts` (JSON round-tripped a plain object for a type-only module) are deleted

New coverage for previously untested behavior includes:

- `DurableObjectSqliteSyncWrapper` and `recordDiff.ts` (neither had any tests)
- the storage transaction contract run against SQLite: rollback of documents/tombstones/clock on throw, `emitChanges`, iterator invalidation after transaction end
- the room connect response shape: own presence excluded from the connect diff, `wipe_all` vs `wipe_presence` selection, presence-id reuse across reconnects
- the push-result trio: `commit` on verbatim application, `discard` on no-op, `rebaseWithDiff` on partial application — including the storage-embellishment path (`emitChanges: 'when-different'`) via a custom delegating storage
- data-message debouncing and flush ordering, including that `pong` skips the flush
- the session prune lifecycle with fake timers, the legacy (protocol ≤ 6) rejection mapping, and broadcast-side down-migration failure rejecting only the affected session
- client push coalescing details: clientClock increments, presence put-then-patch, presence re-put after reconnect, solo-mode fps drop
- `chunk` boundary semantics (a message exactly at the size limit is chunked) and the assembler's synchronous throw on invalid `{`-prefixed JSON
- `updateStore` isolation details, including delete-then-equal-put resurrecting a record
- top-level string fields getting append ops (not just `props`), and `legacyAppendMode` leaving array appends untouched

Several existing tests turned out to be vacuous and were replaced with working versions: the v5 protocol special-case test was dead code (early-returned since protocol v7), the v1-deletion migration test pushed from the v2 client, the websocket chunking test never exceeded the chunking threshold, and four `it.fails`-marked backoff tests are now real deterministic tests.

The fuzz suite is untouched. The suite grows from 608 to 628 tests.

### Change type

- [x] `bugfix`

### Test plan

1. Read `packages/sync-core/SPEC.md` and check the rules against your understanding of the library.
2. Run `yarn test run` in `packages/sync-core` (628 tests).

- [x] Unit tests

### Release notes

- Fixed an error thrown while handling a sync client message leaving the session connected instead of rejecting it with `UNKNOWN_ERROR`.
- Fixed `TLSyncRoom` clearing the batched data-message array after handing it to the socket, which broke custom `TLRoomSocket` implementations that serialize messages lazily.

### Code changes

| Section       | LOC change    |
| ------------- | ------------- |
| Core code     | +11 / -3      |
| Tests         | +7728 / -9223 |
| Documentation | +286 / -0     |


steveruizok:
Follow-up notes from the spec rebuild — latent issues left unfixed, quirks for future test writers, and things worth a second look in review.

### Latent issues, deliberately not fixed

- `TLSyncClient.resetConnection(hard: true)` sets `lastServerClock` to 0 rather than the −1 "never connected" sentinel. The only observable difference is which clock a hard reset asks the server for; left as is.
- `NodeSqliteWrapper.transaction` only rolls back when the callback throws — a failing `COMMIT` is not rolled back. Relatedly, `SQLiteSyncStorage` queues its `onChange` microtask before `COMMIT` runs, so a failing commit would still notify listeners. Both are unreachable with in-memory databases but worth knowing for durable deployments.
- If a `wipe_all` connect diff leaves a client store without tldraw's required document/page records, the store's integrity checker recreates them as user-source changes, which the client then pushes back to the server. This is a tlschema/store interaction rather than a sync-core bug, but it silently polluted the old client tests' push queues.
- Two sharp edges are now documented as contract rather than changed: passing `log: undefined` explicitly to `TLSocketRoom` disables logging entirely (SR3, the code checks `'log' in opts`), and registering the same listener function twice on `MicrotaskNotifier` deduplicates — either unsubscribe removes both registrations (MN3).
- The two storage backends genuinely disagree at construction: `InMemorySyncStorage` clamps the document clock up to the max `lastChangedClock`/tombstone clock, clamps `tombstoneHistoryStartsAtClock` down to the document clock, and discards tombstones when the history window is empty (IM1–IM3); `SQLiteSyncStorage` takes all three verbatim (SQ3). Specced separately rather than unified, since either behavior could be intentional.
- The comment in `TLSyncRoom.handleConnectRequest` saying the v5/v6 coercion "can be deleted when the protocol gets to v7" is stale — the behavior is still load-bearing at v8, only the comment is wrong.

### Quirks for future test writers

- Storage-contract tests (the SS rules) belong in `src/test/storageContractSuite.ts`, not the per-backend files — they run against both backends automatically.
- `node:sqlite` returns cross-realm `Uint8Array`s under jsdom; use `ArrayBuffer.isView`, never `instanceof Uint8Array`.
- `warnOnce` from `@tldraw/utils` dedupes per process, so the CW5 test partial-mocks it; incidental 1006 closes from other tests' teardown would otherwise consume the one-time warning nondeterministically.
- Mock `TLRoomSocket`s should still clone received messages — broadcast diff contents are shared across sessions even after the flush fix in this PR.

### Worth a second look in review

- The RC5 (force-reconnect on wipe) test pokes `InMemorySyncStorage`'s internal `tombstoneHistoryStartsAtClock` atom to provoke the wipe condition — the only deterministic way without waiting on the 1s-throttled tombstone pruner.
- The CP6 fps-drop assertion reads the client's internal `fpsScheduler.targetFps`; there is no public observable for it.
- Two test files were deleted outright as valueless: `RoomSession.test.ts` (asserted constant ordering and built object literals) and `server-types.test.ts` (JSON round-tripped a plain object for a type-only module).
- Four previously vacuous tests now assert real behavior and are worth checking against expectations: the v5 protocol special-case test was dead code (an early return made its body unreachable once the protocol passed v7), the v1-deletion migration test pushed from the v2 client, the websocket chunking test never exceeded the chunking threshold, and the `it.fails`-marked backoff tests are now real deterministic tests.

PR #9201, opened 2026-06-15
fix(editor): let InputsManager own its frame subscription

### LINKED ISSUE #8892: TickManager listener lifecycle is the root cause of the strict-mode camera-state workaround
## Problem

The "reset camera state on dispose" block in `Editor.ts:411–416` papers over a deeper issue: `'tick'` event listeners are added by `Editor` after `TickManager` starts, but `TickManager.dispose` doesn't track or remove them — they're owned by the `EventEmitter` superclass. On React Strict Mode double-mount, listeners may re-bind to stale closure scopes.

## Where

- `packages/editor/src/lib/editor/managers/TickManager/TickManager.ts:49–55` — `dispose` only cancels RAF, doesn't unregister listeners
- `packages/editor/src/lib/editor/Editor.ts:411–416` — the workaround comment plus `off('tick', ...)` and `_setCameraState('idle')` reset
- Multiple `'tick'` listener call sites across `Editor.ts`

## Suggested fix

Let `TickManager` own the listener registry: `onTick(cb)` returns an unsubscribe, and `TickManager.dispose` unregisters all of them. Demote `editor.inputs.updatePointerVelocity` from a privileged hard-coded call inside `tick()` to an ordinary frame listener registered by `InputsManager`. Once that's symmetric, the workaround in `Editor.ts:411–416` can be removed.

## Context

Surfaced during an internal modularity audit of the editor manager classes.

kaneel:
Closes #8892 (partial — focused fix; no `EditorManager` base class or full manager sweep).

I've had some more ideas for a suitable way to keep having the Editor being the event emitter bus with every "managers" owning their events, it has been spiked in https://github.com/tldraw/tldraw/pull/9204 

## Summary

- `InputsManager` now subscribes to the editor's `frame` event and tears down its own listener in `dispose()`.
- `TickManager` no longer hard-codes `editor.inputs.updatePointerVelocity()`; it only drives the clock and emits `frame` / `tick`.
- `Editor.dispose()` calls `removeAllListeners()` after emitting `dispose`, and the strict-mode camera workaround is simplified to `_setCameraState('idle')` only.

## Why

Issue #8892 flagged asymmetric lifecycle: `TickManager` drove updates but didn't own the listeners they triggered. This moves pointer velocity into a normal frame subscription owned by `InputsManager`, and centralizes dispose-time listener cleanup on the editor.

The load-bearing strict-mode fix is resetting persisted `cameraState` to `'idle'` when the editor is disposed — leftover tick listeners on a dead editor instance were never the real bug.

## Test plan

- [x] `cd packages/editor && yarn test run TickManager InputsManager`
- [x] `cd packages/tldraw && yarn test run HandTool Editor setCamera`
- [x] `yarn workspace @tldraw/editor build-api` (no public API change on `InputsManager`)

### API changes

- No public API changes for SDK consumers.
- Added `@internal` `InputsManager.dispose()` to the API report. This is an editor lifecycle hook; do not call it from app code — use `editor.dispose()` instead.
- `Editor.dispose()` now removes all event listeners after emitting `dispose`. Custom `editor.on(...)` handlers are cleared when the editor is torn down (expected for a disposed instance).

### Release notes

- Fix camera state getting stuck at `moving` when the editor is disposed during camera movement (e.g. React strict mode with a shared store).

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -409,9 +409,6 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this._tickManager = new TickManager(this)
 		this.disposables.add(() => this._tickManager.dispose())
 		this.disposables.add(() => {
-			// Reset camera state to 'idle' so the store isn't left stuck at 'moving'
-			// when tick events stop (e.g. React strict mode disposes while camera is moving)
-			this.off('tick', this._decayCameraStateTimeout)
 			this._setCameraState('idle')
 		})
 
@@ -419,6 +416,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this.disposables.add(() => this.fonts.dispose())
 
 		this.inputs = new InputsManager(this)
+		this.disposables.add(() => this.inputs.dispose())
 		this.performance = new PerformanceManager(this)
 		this.disposables.add(() => this.performance.dispose())
 		this.collaborators = new CollaboratorsManager(this)
@@ -1179,6 +1177,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this.store.dispose()
 		this.isDisposed = true
 		this.emit('dispose')
+		this.removeAllListeners()
 	}
 
 	/* ------------------ Themes (shadowing the theme manager) ------------------ */
--- a/packages/tldraw/src/test/translating.test.ts
+++ b/packages/tldraw/src/test/translating.test.ts
@@ -783,9 +783,6 @@ describe('snapping with multiple shapes', () => {
 })
 
 describe('Snap-between behavior', () => {
-	beforeEach(() => {
-		editor?.dispose()
-	})
 	it('snaps a shape horizontally between two others', () => {
 		// ┌─────┐               ┌─────┐
 		// │     │               │     │
@@ -1160,9 +1157,6 @@ describe('Snap-between behavior', () => {
 })
 
 describe('Snap-next-to behavior', () => {
-	beforeEach(() => {
-		editor?.dispose()
-	})
 	it('snaps a shape to the left of two others, matching the gap size', () => {
 		// ┌───┐
 		// │ X │
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -2198,6 +2198,8 @@ export class InputsManager {
     get currentPagePoint(): Vec;
     // @deprecated (undocumented)
     get currentScreenPoint(): Vec;
+    // @internal (undocumented)
+    dispose(): void;
     getAccelKey(): boolean;
     getAltKey(): boolean;
     getCtrlKey(): boolean;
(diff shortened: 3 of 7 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -409,16 +409,14 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this._tickManager = new TickManager(this)
 		this.disposables.add(() => this._tickManager.dispose())
 		this.disposables.add(() => {
-			// Reset camera state to 'idle' so the store isn't left stuck at 'moving'
-			// when tick events stop (e.g. React strict mode disposes while camera is moving)
-			this.off('tick', this._decayCameraStateTimeout)
```
mimecuvalo, line 414: why was this removed?

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -409,16 +409,14 @@ export class Editor extends EventEmitter<TLEventMap> {
...
 			this._setCameraState('idle')
 		})
 
 		this.fonts = new FontManager(this, fontAssetUrls)
 		this.disposables.add(() => this.fonts.dispose())
 
 		this.inputs = new InputsManager(this)
+		this.disposables.add(() => this.inputs.dispose())
```
mimecuvalo, line 419: huh good catch

packages/editor/src/lib/editor/Editor.ts
```diff
@@ -1179,6 +1177,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this.store.dispose()
 		this.isDisposed = true
 		this.emit('dispose')
+		this.removeAllListeners()
```
mimecuvalo, line 1180: why was this added?

packages/tldraw/src/test/translating.test.ts
```diff
@@ -783,9 +783,6 @@ describe('snapping with multiple shapes', () => {
 })
 
 describe('Snap-between behavior', () => {
-	beforeEach(() => {
-		editor?.dispose()
```
mimecuvalo, line 787: not sure if we should drop these `dispose` calls 🤔

mimecuvalo (commented):
nice cleanup! not sure about some of the other stuff that snuck in though

packages/editor/src/lib/editor/Editor.ts
kaneel: This is the simpler fix here, it makes sure we just remove all existing listener after disposing.

packages/tldraw/src/test/translating.test.ts
kaneel: the changes exposed a problem in the tests, now that we clearly dispose on the first time, I'll take a second look if you want

he outer `beforeEach` (line 44) creates a fresh editor before every test and the outer `afterEach` (line 22) disposes after every test, so per-test isolation is already handled.

THe previous dispose being weaker, we never realised these `beforeEach` were breaking the tests (or should have)

packages/editor/src/lib/editor/Editor.ts
kaneel: it's now covered in the clearAllListener() call

max-drake (approved):
my concerns were addressed by the responses to @mimecuvalo's comments, so looks good to me

PR #9203, opened 2026-06-15
feat(dotcom): seed new workspaces with a welcome file forked from a template

jsscclr:
In order to give new workspaces a welcoming first file instead of a blank canvas — and to make that content editable without a deploy — this PR seeds a new workspace's first file by forking a "welcome template". When a workspace is created (behind the `groups_frontend` flag), its first file is seeded by passing a fixed `createSource: 'welcome'`, which the sync worker resolves: it forks the published snapshot of the file currently marked as the welcome template, falling back to a committed default snapshot when none is set or it can't be read (fresh env, deploy skew, unpublished). An admin marks any published file as the template. The file is ordinary — renameable, editable, deletable — and "My files" keeps its blank first file.

Relates to #9143.

> [!NOTE]
> Draft + alternative approach. This is the fork/publish alternative to #9142, which seeds the same welcome content from a `RoomSnapshot` baked into the sync worker. The trade-off: #9142 ships the content in the worker bundle (identical everywhere, no data setup, but editing means re-exporting and re-deploying); this PR makes the welcome content a real published file an admin can edit and re-publish live, with a committed default so it still works everywhere out of the box. Opening both for comparison — only one should land.

How it works:

- **`createSource: 'welcome'`** (`WELCOME_CREATE_SOURCE` in `dotcom-shared`) is a fixed, slug-less marker — the client never knows which file is the template, so an admin can retarget it without a client change. `TLFileDurableObject.handleFileCreateFromSource` special-cases it before the prefix/id split.
- **Resolution** (`resolveWelcomeSnapshot`): read the `welcome_template` pointer, fork that file's published snapshot via the existing `getPublishedRoomSnapshot`, else the committed default. A template that is *set but unreadable* is reported (Sentry) rather than silently serving the default; a genuinely-absent template is silent.
- **`welcome_template`** (migration `035`) is a singleton Postgres table, deliberately outside the Zero publication — worker-side config, never replicated to clients.
- **Admin**: `/app/admin/welcome-template` GET/POST/clear (admin-gated) + a "Welcome template" section in the admin panel. Setting requires the file be published; the panel warns if the current template is no longer live.
- **Seeding** is single-flighted per workspace, so a creation flow and a workspace switch can't create duplicates and all callers get the result. The committed default reuses a self-rebaking snapshot test that keeps it migrated to the current schema.

### Change type

- [x] `feature`

### Test plan

Preview deploy: https://pr-9203-preview-deploy.tldraw.com/

**Setup**

1. Open the preview and sign in; enrol yourself in groups via `/admin` (the `groups_frontend` flag).

**Default (no template set)**

2. Create a workspace and open it → its first file is "Welcome to your workspace" with the default welcome canvas; no inline-rename prompt.
3. Empty "My files" and open it → still a blank, date-named file with the rename prompt.

**Admin template + fork**

4. Create a file, give it distinct content, and **publish** it.
5. In `/admin` → Welcome template, set it by file ID.
6. Create another workspace → its welcome file is a fork of that published file's content.
7. Unpublish/delete that file → `/admin` shows the template is no longer live; new workspaces fall back to the default.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Core code       | +22 / -0    |
| Tests           | +209 / -0   |
| Apps            | +2784 / -44 |

### Release notes

- Add a welcome file to new workspaces: creating a workspace opens a "Welcome to your workspace" canvas (forked from an admin-set template, or a built-in default). Deleting it, or opening an already-empty workspace, gives a blank file as before.

### DIFF
--- a/apps/dotcom/sync-worker/src/welcome/defaultWelcomeSnapshot.ts
+++ b/apps/dotcom/sync-worker/src/welcome/defaultWelcomeSnapshot.ts
@@ -1,0 +1,2441 @@
+// The committed *default* welcome document — the fallback content a new workspace's first
+// file is seeded with when no file is marked as the welcome template (fresh dev/preview/prod,
+// before an admin has set one). Once a welcome template file is marked, its published
+// snapshot is used instead and this default is no longer consulted. Its canvas is a
+// three-panel comic in the brand's illustration style (the Downy/Cloudy/Penta characters)
+// covering shared visibility, inviting people, and moving files in.
+//
+// Stored pre-serialized as the JSON of a RoomSnapshot: the worker writes the string to R2
+// verbatim and parses a fresh snapshot per seed, so worker isolates don't pay object-graph
+// evaluation at cold start and no mutable snapshot object is shared between rooms.
+//
+// Schema migrations are handled for you: welcome.test.ts keeps this file baked at the
+// current schema, so when a migration touches these records the test fails and `yarn test -u`
+// re-bakes the literal below. Only redesigning the default canvas needs a manual re-export:
+// open a file in the app, build the content, run this in the console, and rebuild the
+// RoomSnapshot JSON between the backticks below ({ documentClock: 0,
+// tombstoneHistoryStartsAtClock: 0, schema, documents: records.map((state) => ({ state,
+// lastChangedClock: 0 })) }):
+//
+//   JSON.stringify({
+//     schema: editor.store.schema.serialize(),
+//     records: Object.values(editor.store.serialize('document')),
+//   })
+//
+// Keep only document/page/shape (and asset/binding) records: drop any `user` records and
+// reset any `textFirstEditedBy` props to null, since those carry the authoring user's
+// identity.
+//
+// Keep the content near the origin and inside roughly 1200×800 page units: new files open
+// at the default camera (no zoom-to-fit on first visit), so the canvas must fit a typical
+// editor viewport. welcome.test.ts asserts a coarse bounds envelope for this.
+export const defaultWelcomeSnapshotJson = `{
+	"documentClock": 0,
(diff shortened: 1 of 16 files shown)


apps/dotcom/sync-worker/src/welcome/defaultWelcomeSnapshot.ts
```diff
@@ -0,0 +1,2441 @@
...
+//   })
+//
+// Keep only document/page/shape (and asset/binding) records: drop any `user` records and
+// reset any `textFirstEditedBy` props to null, since those carry the authoring user's
+// identity.
+//
+// Keep the content near the origin and inside roughly 1200×800 page units: new files open
+// at the default camera (no zoom-to-fit on first visit), so the canvas must fit a typical
+// editor viewport. welcome.test.ts asserts a coarse bounds envelope for this.
+export const defaultWelcomeSnapshotJson = `{
```
mimecuvalo, line 32: random comment, makes me wonder if these snapshots should be translated? 😬 :ducks:
for the purposes of this PR though I wouldn't overarchitect on that - buuuut it could speak to possibly generating on the frontend in the future...not sure

jsscclr: Agreed! Do think we should be tackling this sooner rather than later - created an issue, will see what's feasible!

mimecuvalo: yeah i don't wanna block on tomorrow's release so i think it's fine for now. it'd be a bigger push to setup i18n wiring on the sync-worker. let's not worry about it for now 👍

mimecuvalo (approved):
approving — let's goooo

PR #9204, opened 2026-06-15
spike(editor): introduce EditorManager base class for subscription lifecycle

kaneel:
This spike establishes a consistent vocabulary for cleanup in editor-attached managers, addressing the lifecycle inconsistency behind [#8892](https://github.com/tldraw/tldraw/issues/8892): listeners and resources were set up in one place and torn down in another (or not at all), with every manager hand-rolling its own `dispose()`.

It adds a `@public` `EditorManager` base class and adopts it in the managers that actually own a subscription or resource. The goal is that engineers know what to reach for so anything they create sees itself out automatically.

## Concepts

| Term | Type | Meaning |
|------|------|---------|
| `EditorManager` | `@public` abstract class | Base for managers that hold an `Editor` ref and need lifecycle cleanup |
| `_register(fn)` | protected method | Register any teardown fn; runs on `dispose()` (VS Code `Disposable._register`) |
| `addEditorEvent(event, fn)` | protected method | Subscribe to a `TLEventMap` event via `editor.on` and auto-register the matching `off` |

## What to use to register cleanup

Register the "undo" next to the "do". The base picks up where the editor's existing registries leave off:

| You create | Register via | Cleaned up on |
|------------|--------------|---------------|
| editor bus listener (`tick`/`frame`/`change`…) | `this.addEditorEvent(event, fn)` | manager `dispose()` |
| store side effect (`editor.sideEffects.register…`) | `this._register(disposer)` | manager `dispose()` |
| reaction (`react(...)`) | `this._register(react(...))` | manager `dispose()` |
| DOM listener | `this._register(() => el.removeEventListener(...))` | manager `dispose()` |
| timeout / interval / frame | `editor.timers.setTimeout(ctx, …)` | context / editor dispose |
| other resource (cache, index, child) | `this._register(() => x.dispose())` | manager `dispose()` |
| cleanup on the editor itself | `editor.disposables.add(fn)` | `editor.dispose()` |

`addEditorEvent` is sugar over `_register` for the most footgun-prone case — bus subscriptions, the exact asymmetry behind #8892.

## Managers adopted

| Manager | Uses | Why |
|---------|------|-----|
| `InputsManager` | `addEditorEvent('frame', …)` | owns its pointer-velocity frame subscription |
| `FocusManager` | `_register` | side-effect handler + DOM keydown/mousedown listeners |
| `TextManager` | `_register` | measurement DOM nodes created in its constructor |
| `TickManager` | `dispose()` override + `super.dispose()` | needs ordered teardown (pause loop before cancelling RAF) |

## Deliberately not adopted

The pattern is opt-in by need, not mandated by category. Managers with no teardown (`ThemeManager`) or fully self-contained, well-named teardown (`SpatialIndexManager`) stay as plain classes — extending the base would add an inheritance hop for no cleanup benefit.

## Documentation

The disposable vocabulary is documented where engineers will look for it: a decision-guide doc comment on `EditorManager`, and a "Managers" note in `AGENTS.md` under Architecture notes.

### Change type

- [x] `improvement`

### Test plan

- [x] Unit tests: `EditorManager`, `InputsManager`, `TickManager`, `FocusManager`, `TextManager`
- [x] Full `packages/editor` suite
- [x] `yarn typecheck`
- [x] `yarn api-check`

### API changes

- Adds `EditorManager` as a `@public` abstract class (required so the `@public` `InputsManager` can extend it; `_register` / `addEditorEvent` are `protected`, not consumer-facing).
- `InputsManager`, `TickManager`, `FocusManager`, and `TextManager` now extend `EditorManager`.
- `TextManager` no longer exposes its own public `dispose()` or `editor` field (now inherited; `dispose()` is `@internal`) — a minor public-surface reduction.

### Release notes

- Internal: introduce an `EditorManager` base class so editor managers share one teardown contract; no change to documented SDK behavior.

### DIFF
--- a/packages/editor/src/lib/editor/managers/EditorManager.ts
+++ b/packages/editor/src/lib/editor/managers/EditorManager.ts
@@ -12,10 +12,10 @@ import type { TLEventMap } from '../types/emit-types'
  * Register cleanup by what you're creating:
  *
  * - Editor bus event (`tick`, `frame`, `change`, …): {@link EditorManager.addEditorEvent}
- * - Store side effect (`editor.sideEffects.register…`): `this._register(disposer)`
- * - Reaction (`react(...)` from `@tldraw/state`): `this._register(react(...))`
- * - DOM listener: `this._register(() => el.removeEventListener(...))`
- * - Other resource (cache, index, child manager): `this._register(() => x.dispose())`
+ * - Store side effect (`editor.sideEffects.register…`): `this.register(disposer)`
+ * - Reaction (`react(...)` from `@tldraw/state`): `this.register(react(...))`
+ * - DOM listener: `this.register(() => el.removeEventListener(...))`
+ * - Other resource (cache, index, child manager): `this.register(() => x.dispose())`
  *
  * For timeouts, intervals, and animation frames prefer `editor.timers` (context-grouped
  * and auto-disposed) over raw `setTimeout`. For cleanup on the editor itself rather than a
@@ -35,7 +35,7 @@ export abstract class EditorManager {
 	 * Register a teardown function to run on `dispose()`. The universal way for a manager to
 	 * make a resource "see itself out". Returns the same function for convenience.
 	 */
-	protected _register(dispose: () => void): () => void {
+	protected register(dispose: () => void): () => void {
 		this.disposables.add(dispose)
 		return dispose
 	}
@@ -49,7 +49,7 @@ export abstract class EditorManager {
 		fn: (...args: TLEventMap[E]) => void
 	): void {
 		this.editor.on(event, fn as any)
-		this._register(() => this.editor.off(event, fn as any))
+		this.register(() => this.editor.off(event, fn as any))
 	}
 
 	/** @internal */
--- a/packages/editor/src/lib/editor/managers/EditorManager.ts
+++ b/packages/editor/src/lib/editor/managers/EditorManager.ts
@@ -1,0 +1,60 @@
+import type { Editor } from '../Editor'
+import type { TLEventMap } from '../types/emit-types'
+
+/**
+ * Base class for editor-attached managers that need to clean up after themselves.
+ *
+ * Extending `EditorManager` gives a manager one consistent teardown contract: register
+ * anything that needs undoing, and it runs automatically on `dispose()`. This keeps
+ * subscription lifecycle symmetric — whatever sets a thing up is what tears it down — so
+ * it can't be forgotten (the failure mode behind the strict-mode camera bug, #8892).
+ *
+ * Register cleanup by what you're creating:
+ *
+ * - Editor bus event (`tick`, `frame`, `change`, …): {@link EditorManager.addEditorEvent}
+ * - Store side effect (`editor.sideEffects.register…`): `this._register(disposer)`
+ * - Reaction (`react(...)` from `@tldraw/state`): `this._register(react(...))`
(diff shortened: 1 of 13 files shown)


packages/editor/src/lib/editor/managers/InputsManager/InputsManager.ts
```diff
@@ -472,7 +472,7 @@ export class InputsManager {
 	private _velocityPrevPoint = new Vec()
 
 	/**
-	 * Update the pointer velocity based on elapsed time. Called each frame.
+	 * Update the pointer velocity based on elapsed time. Called by the tick manager.
```
mimecuvalo, line 472: Was InputsManager meant to use the new stuff?

packages/editor/src/lib/editor/managers/EditorManager.ts
```diff
@@ -0,0 +1,28 @@
...
+	protected readonly disposables = new Set<() => void>()
+
+	constructor(protected readonly editor: Editor) {}
+
+	protected _register(dispose: () => void): () => void {
+		this.disposables.add(dispose)
+		return dispose
+	}
+
+	protected onEditor<E extends keyof TLEventMap>(
```
mimecuvalo: TickManager isn't really using any of this stuff heh
let's do InputsManager to make this work tangible

kaneel: @mimecuvalo it was a spike to demonstrate the usage, do you want me to apply to all other managers just for kicks?

mimecuvalo: yeah i get that it was a spike — the tricky thing is that TickManager isn't actually consuming anything — yes, please at least apply to InputsManager as it will better demonstrate this spike 🙏

packages/editor/src/lib/editor/managers/TickManager/TickManager.ts
```diff
@@ -45,11 +46,10 @@ export class TickManager {
 		this.cancelRaf = throttleToNextFrame(this.tick)
 	}
 
-	// Clear the listener
 	@bind
```
mimecuvalo: you don't need the `@bind` anymore i think

packages/editor/src/lib/editor/managers/EditorManager.ts
```diff
@@ -0,0 +1,60 @@
...
+export abstract class EditorManager {
+	protected readonly disposables = new Set<() => void>()
+
+	constructor(protected readonly editor: Editor) {}
+
+	/**
+	 * Register a teardown function to run on `dispose()`. The universal way for a manager to
+	 * make a resource "see itself out". Returns the same function for convenience.
+	 */
+	protected _register(dispose: () => void): () => void {
```
mimecuvalo: let's just call this `register` - `_register` makes it sound like it's `private` but it's actually `protected`.
let's keep the naming consistent with how `addEditorEvent` below which is also protected isn't prefixed by `_`

mimecuvalo (commented):
cool, couple nits to address but otherwise good

mimecuvalo (approved):
pre-approving modulo comments

packages/editor/src/lib/editor/managers/TickManager/TickManager.ts
kaneel: I think you may be right

Issue #9189, opened 2026-06-15
Design-system consistency in the dotcom client


jsscclr:
The dotcom client (`apps/dotcom/client`) has a design system — shared components (`TlaButton`, `TldrawUiInput`, `TlaIcon`, `ExternalLink`, the `tla-menu` family, `TldrawUiDialog*`) and tokens (`--tla-color-*`, `--tl-space-*`, `--tla-radius-*`) — but a large share of the UI bypasses it with raw HTML elements, inline styles, hardcoded values, and duplicated CSS. This makes theming, accessibility, and visual consistency harder to maintain: a styling change has to be made in many places instead of one, and ad-hoc elements miss the focus, keyboard, and ARIA behavior the shared components provide.

One measure of the drift: the app renders `<TlaButton>` in ~20 places and a raw `<button>` in ~20 user-facing places — roughly half of all buttons skip the component. Spacing tokens (`--tl-space-*`) exist but are used in only a handful of files; typography has no token layer at all.

## Scope

`apps/dotcom/client/src`, including the admin and dev pages (lower priority, but in scope). SDK packages (`packages/*`) are out of scope — their design system is separate. Raw elements inside the primitives themselves (`TlaButton.tsx`, `tla-menu.tsx`) are correct and excluded.

## Sub-issues

Mechanical sweeps — swap an ad-hoc pattern for an existing component/token:

- [ ] #9190 Replace raw `<button>` with `TlaButton` / `TldrawUiButton`
- [ ] #9191 Replace raw `<input>` / `<textarea>` with shared form components
- [ ] #9192 Replace inline `style` props with CSS classes and design tokens
- [ ] #9193 Consolidate duplicated button CSS into `TlaButton` variants
- [ ] #9194 Fold `TlaCtaButton` into `TlaButton` (`variant="cta"`)
- [ ] #9197 Replace hardcoded spacing with `--tl-space-*` tokens
- [x] #9198 Use `ExternalLink` consistently instead of raw `<a>` tags
- [x] #9199 Miscellaneous cleanups (color holes, inline SVGs, custom modal, dead CSS)

Audit-first — needs review to separate intentional design from drift:

- [ ] #9200 Audit and unify the border-radius scale (two colliding token scales; some radii are intentional pills / concentric pairs)

Foundational — a canonical mechanism must be decided before the sweep:

- [ ] #9196 Establish a typography system (no `--tla-text-*` tokens or text component exist today)

## Notes

[...]

PR #9205, opened 2026-06-15
refactor(dotcom): fold TlaCtaButton into TlaButton cta variant

### LINKED ISSUE #9194: Fold TlaCtaButton into TlaButton (variant cta) in the dotcom client
Part of #9189.

`TlaCtaButton` is a separate component that duplicates `TlaButton`, which already exposes a `cta` variant. Maintaining two button components means CTA styling and behavior can drift apart.

## Proposal

Migrate `TlaCtaButton` call sites to `TlaButton` with `variant="cta"` (preserving the `canvas` prop behavior if needed), then delete `TlaCtaButton.tsx` and `cta-button.module.css`.

## Call sites (~7 across 5 files)

- `tla/components/dialogs/TlaSignInDialog.tsx:212, 255`
- `tla/components/dialogs/TlaLegalAcceptance.tsx:114`
- `tla/components/TlaEditor/TlaHistorySnapshotEditor.tsx:75`
- `tla/components/TlaFileError/TlaFileError.tsx:107`
- `tla/components/TlaEditor/TlaEditorTopRightPanel.tsx:53, 76, 149` (incl. a `canvas` variant)

## Notes

- Check whether `TlaCtaButton`'s `canvas` prop and any CTA-specific styling are fully covered by `TlaButton`'s `cta` variant; extend the variant if there is a gap rather than keeping the component.

## Acceptance

- All `TlaCtaButton` usages migrated to `TlaButton variant="cta"`.
- `TlaCtaButton.tsx` and `cta-button.module.css` deleted.
- No CTA visual or behavior regressions (sign-in, legal acceptance, file error, editor top-right import/share).

jsscclr:
In order to remove a duplicate button component, this PR folds `TlaCtaButton` into `TlaButton`'s `cta` variant. The two components both rendered call-to-action buttons, but `TlaButton`'s existing `cta` variant was a near-unused stub (one call site) that did not match `TlaCtaButton`'s actual styling. This promotes `TlaCtaButton`'s styling to be the canonical `cta` variant (background drawn by a `::before` pseudo-element, the `canvas` ring, and the muted `secondary` treatment), adds `canvas` and `ctaSecondary` props plus the `type="button"` / `draggable={false}` defaults `TlaCtaButton` hardcoded, migrates all 8 call sites, and deletes the component. Closes #9194 (part of the design-system consistency tracking issue #9189).

The CTA buttons should look and behave **identically** to before, with one intended exception: the expired-invite "OK" button previously used the unmatched stub and now uses the real CTA styling.

### Change type

- [x] `improvement`

### Test plan

A `dotcom-preview-please` preview is requested. Every CTA button in the app changed component, so the review is "do they all still look and behave the same". Here is exactly where to find each one on the preview:

**Easy to reach**

1. **Sign-in dialog — two CTAs.** While signed out, open the sign-in dialog (e.g. the "Sign in" button, top-right of any board).
   - "Sign in with Google" — filled blue CTA with the Google icon and label, **with a gap between icon and text** (this was a regression risk — confirm they are not touching).
   - "Continue with email" (below the "or" divider, under the email field) — the muted **secondary** CTA. Same height/shape as the Google button, lighter fill.
2. **Editor "Share" button.** Open one of your own files. Top-right of the canvas: the blue "Share" CTA. It floats over the canvas, so confirm the subtle **ring** around it (a background-coloured halo separating it from canvas content) is present.

**Needs a specific state**

3. **Anonymous "Sign in to share" CTA.** Open a shared file while signed out. Top-right shows a canvas-style sign-in CTA (same ring as Share).
4. **"Copy to my files" import button.** Open a legacy/published room. Top-right canvas CTA.
5. **File-error "Sign in" CTA.** While signed out, open a file URL that requires auth — the error page shows a "Sign in" CTA.
6. **"Restore version" CTA.** Open a file → version history → pick a past snapshot. The share panel shows a canvas-style "Restore version" CTA.
7. **Expired-invite "OK" button** ⚠️ *intended change*. Open an expired invite link. The "OK" button is the **one** button whose appearance changes (stub → real CTA styling) — confirm it now looks like a proper CTA.
8. **Legal-acceptance "Accept and continue"** (hardest — brand-new account only). Shown on first sign-in before accepting terms.

**What "correct" means for all of them:** 36px height, rounded corners, no layout shift versus production, and on hover the fill darkens to the CTA hover colour (not the primary-button hover — the stub had this wrong).

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +114 / -141 |


apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
```diff
@@ -12,6 +12,10 @@ export const TlaButton = forwardRef<
 		iconRightClassName?: string
 		ghost?: boolean
 		variant?: 'primary' | 'secondary' | 'warning' | 'cta'
+		/** Adds a background-coloured ring so a `cta` button reads correctly floating over canvas content. */
+		canvas?: boolean
```
kaneel, line 16: `canvas` does not communicate the intent, could we think of a better name?

apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
```diff
@@ -12,6 +12,10 @@ export const TlaButton = forwardRef<
...
+		/** Renders a `cta` button with the muted secondary treatment instead of the filled call-to-action colour. */
+		ctaSecondary?: boolean
```
kaneel, line 18: We should probably think of this as an other variant.

kaneel (commented):
I like the premise but not entirely convinced by the two booleans added to `TlaButton`

We could either make them all variant but if the "canvas" modifier is meant to be applied on all button then we keep it but rename it to something a bit clearer.

`<TlaButton canvas />` does not necesarily tell me it's over the canvas, `<TlaButton overCanvas />` maybe a bit more. If it does only add a ring, why not `<TlaButton withRing />` ?

apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
```diff
@@ -23,22 +27,33 @@ export const TlaButton = forwardRef<
...
+		canvas = false,
+		ctaSecondary = false,
+		type,
+		draggable,
 		onClick,
 		...props
 	},
 	ref
 ) {
+	const isCta = variant === 'cta'
```
mimecuvalo, line 39: drive-by comment here: i'm not sure why historically we added CTA but CTA in my head should just be `primary`. Any reason those two can't just be one and the same?

jsscclr:
Eep, forgot to add draft to this one - this was a claude drive-by, and just reading this now it's some of the worst code I've seen and embarrassed to have it attached to my name 😂

jsscclr:
Closing this - will tackle with fresh eyes!

PR #9208, opened 2026-06-15
fix(dotcom): apply the warning class for TlaButton variant="warning"

### LINKED ISSUE #9207: TlaButton variant="warning" renders without warning styling
`TlaButton` accepts a `variant="warning"`, but the variant renders nothing — its `classNames` block never applies the `warning` class, so the prop is a silent no-op.

## Where

- `apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx` — `variant` is typed `'primary' | 'secondary' | 'warning' | 'cta'`, but the `classNames` object only sets `cta` / `primary` / `secondary` / `ghost`. There is no `[styles.warning]: variant === 'warning'` entry.
- `apps/dotcom/client/src/tla/components/TlaButton/button.module.css` — `.warning` (and `.warning.ghost`) are defined but, because the class is never applied, unreachable.

## Impact

Latent, and admin-only today. All four `variant="warning"` call sites live in `apps/dotcom/client/src/pages/admin.tsx`, and each passes its own `className`:

- Three (`admin.tsx:561`, `:1082`, `:1194`) use `styles.deleteButton`, which is `border: 2px solid var(--tla-color-warning) !important`. They show a warning *border*, but not the variant's intended treatment (`background-color: secondary` + `color: warning`).
- One (`admin.tsx:220`, the reboot button) uses `styles.userActionButton`, which has no warning colour at all — so that button currently renders as a plain base button with no warning styling.

No user-facing breakage today (it is the internal admin panel, and the delete buttons are partially masked by their own CSS), but the `warning` variant is effectively dead and one admin button is unstyled.

## Expected

`variant="warning"` should apply the `.warning` class so the button gets the warning treatment, consistent with the other variants.

## Fix

Add `[styles.warning]: variant === 'warning'` to the `classNames` block in `TlaButton.tsx`. (`.warning.ghost` already composes with the existing `ghost` handling.)

## Notes

Found while folding `TlaCtaButton` into `TlaButton` (#9194), which required a close read of the `classNames` block. Relates to the design-system consistency effort (#9189). This is also a prerequisite for part of #9193: `admin.tsx`'s `.deleteButton` / `.userActionButton` are exactly the kind of ad-hoc button CSS #9193 wants to fold into variants, but those sites can only drop their `className` workarounds once the `warning` variant actually works — and once the canonical "warning button" look is decided (the current border hack and the variant's secondary-bg/warning-text differ).

jsscclr:
In order to fix a button variant that silently did nothing, this PR makes `TlaButton`'s `classNames` apply `styles.warning` when `variant="warning"`. The variant was typed and used (four call sites in the admin panel) but never applied its class, so it rendered with no warning treatment. Closes #9207 (found while folding `TlaCtaButton`, part of the design-system consistency effort #9189).

### Change type

- [x] `bugfix`

### Test plan

A `dotcom-preview-please` preview is requested. The affected buttons are in the admin panel (`/admin`), so a reviewer with admin access can check that the warning-variant buttons now render with the warning treatment (secondary background + warning-coloured text):

1. The user "Reboot" action button (previously rendered as a plain button — now styled as a warning).
2. The "Delete (cannot be undone)" and related delete buttons (previously only had their own warning border via `className`; now they also get the variant's warning fill).

This is admin-only internal tooling, so there is no public user-facing impact.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +1 / -0    |

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx
@@ -43,6 +43,7 @@ export const TlaButton = forwardRef<
 					[styles.cta]: variant === 'cta',
 					[styles.primary]: variant === 'primary',
 					[styles.secondary]: variant === 'secondary',
+					[styles.warning]: variant === 'warning',
 					[styles.ghost]: ghost,
 					[styles.big]: big,
 				},


kaneel (approved):
👏

jsscclr:
Closing as superseded. The `warning` variant this wires up was removed entirely as dead code in #9261, which also dropped its admin call sites (Reboot / Delete / Migrate now use the default variant). With no `variant="warning"` left in the codebase, there's nothing to apply.

Closing #9207 alongside.

Issue #9193, opened 2026-06-15
Consolidate duplicated button CSS into TlaButton variants in the dotcom client


jsscclr:
Part of #9189.

Several CSS modules re-implement button styling that `TlaButton` variants already provide. Once the raw `<button>` sweep (#9189) replaces the markup, these ad-hoc classes should be deleted so button styling lives in one place.

## Canonical component

`apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx` (variants `primary`, `secondary`, `warning`, `cta`).

## Duplicated classes

- `tla/components/dialogs/TlaInviteDialog.module.css` — `.acceptButton`, `.declineButton`
- `tla/components/dialogs/TlaInviteExpiredDialog.module.css` — `.okButton`
- `tla/components/dialogs/dialogs.module.css` — `.cookieButton`, `.inlineButton`
- `tla/components/MaybeForceUserRefresh/MaybeForceUserRefresh.module.css` — `.modalButton`
- `tla/components/TlaSidebar/sidebar.module.css` — `.sidebarActionButton`, `.sidebarCreateFileButton`, `.sidebarCreateWorkspaceButton`, `.sidebarFeedbackButton`, `.sidebarWorkspaceButton`, `.sidebarFileListItemButton`, `.sidebarDotDevDismissButton` and related modifiers

## Notes

- Depends on the raw `<button>` sweep — coordinate so markup and CSS move together rather than leaving orphaned classes.
- Some sidebar buttons have bespoke layout (icon + label rows, hover affordances) that may need a new `TlaButton` variant or a documented exception rather than a forced fit. Capture those as they come up.

## Acceptance

- Ad-hoc button classes removed where a `TlaButton` variant covers them.
- Any remaining bespoke button styling is justified in a comment or a new shared variant.

Issue #9190, opened 2026-06-15
Replace raw <button> elements with TlaButton in the dotcom client


jsscclr:
Part of #9189.

The dotcom client renders raw `<button>` elements in ~20 user-facing spots that should use `TlaButton` (or `TldrawUiButton` where a tldraw-UI button fits). Raw buttons miss the focus ring, disabled handling, variant styling, and consistent sizing that the shared component provides, and they force per-site CSS to re-create that behavior.

For contrast, the app already uses `<TlaButton>` in ~20 places — so about half of all buttons bypass the component.

## Canonical component

- `apps/dotcom/client/src/tla/components/TlaButton/TlaButton.tsx` — variants `primary`, `secondary`, `warning`, `cta`
- `packages/tldraw/src/lib/ui/components/primitives/Button/TldrawUiButton.tsx` — for tldraw-UI contexts

## Violation sites

- `components/BoardHistoryLog/BoardHistoryLog.tsx:72`
- `components/StoreErrorScreen.tsx:27`
- `components/ErrorPage/ErrorPage.tsx:84`
- `tla/providers/TlaRootProviders.tsx:289`
- `tla/components/TlaSidebar/TlaSidebar.tsx:53`
- `tla/components/TlaSidebar/components/TlaSidebarFeedbackButton.tsx:16`
- `tla/components/TlaSidebar/components/TlaSidebarWorkspaceSwitcher.tsx:73, 135, 263`
- `tla/components/dialogs/WorkspaceSettingsDialog.tsx:382, 387`
- `tla/components/dialogs/TlaInviteDialog.tsx:48, 62`
- `tla/components/dialogs/TlaCookieConsent.tsx:112, 119, 125`
- `tla/components/TlaEditor/TlaEditorTopLeftPanel.tsx:128, 359`
- `tla/components/dialogs/TlaSignInDialog.tsx:470`
- `tla/pages/local-file-index.tsx:46`

## Out of scope

The raw `<button>` inside the primitives themselves is correct and should be left alone: `TlaButton.tsx`, `TlaCtaButton.tsx` (see #9189), and `tla-menu.tsx`. That is where the abstraction bottoms out.

## Notes

- Some sites also carry ad-hoc button CSS classes (`.acceptButton`, `.declineButton`, `.cookieButton`, `.inlineButton`) — coordinate with the CSS consolidation sweep so the markup and styles move together.
[...]

PR #9209, opened 2026-06-15
refactor(dotcom): use TlaButton for error-page action buttons

jsscclr:
In order to start replacing raw `<button>` elements with the shared `TlaButton` (part of the design-system consistency effort), this PR converts the reload/refresh action buttons on the shared `ErrorPage` from unstyled raw `<button>`s to `TlaButton variant="primary"`. These were the cleanest conversions in the audit for #9190: they have no existing styling, and `TlaButton` depends only on its bundled CSS and global tokens (no React providers), so it is safe in `ErrorPage`'s intentionally provider-independent context (offline, sync failure, Clerk timeout). Part of #9190.

The remaining raw buttons in #9190 turned out to carry bespoke CSS (cookie consent, sidebar, workspace-settings inline buttons, etc.) that needs reconciling with `TlaButton` variants — that is really #9193's consolidation work — or are intentionally raw (the mobile sidebar overlay, the file-name width-setters). Those are out of scope here.

### Change type

- [x] `improvement`

### Test plan

A `dotcom-preview-please` preview is requested. The affected buttons only appear on error screens, so they take some effort to reach. In each case the reload/refresh button should now render as a styled primary button instead of a browser-default `<button>`:

1. **`StoreErrorScreen`** — shown on certain sync errors (e.g. a client-too-old close reason renders the "Refresh" button).
2. **`RefreshErrorBoundary` (`ErrorPage`)** — shown when a React error boundary catches an error; the refresh CTA uses this.
3. **Clerk-unavailable fallback (`TlaRootProviders`)** — shown if Clerk fails to load within the timeout; the refresh button uses this.

Confirm the button is styled (primary fill, correct hover) and still reloads the page on click.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +15 / -4   |

### DIFF
--- a/apps/dotcom/client/src/components/ErrorPage/ErrorPage.tsx
+++ b/apps/dotcom/client/src/components/ErrorPage/ErrorPage.tsx
@@ -1,6 +1,7 @@
 import { Component, ReactNode } from 'react'
 import { Link } from 'react-router-dom'
 import translationsEnJson from '../../../public/tla/locales-compiled/en.json'
+import { TlaButton } from '../../tla/components/TlaButton/TlaButton'
 import { F, IntlProvider } from '../../tla/utils/i18n'
 import { isInIframe } from '../../utils/iFrame'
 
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
you're a star — ty!

Issue #9197, opened 2026-06-15
Resolve off-scale spacing values (6/10/24px) in the dotcom client


jsscclr:
Part of #9189.

The mechanical part of spacing-token adoption — single-value `padding` / `margin` / `gap` that exactly matches a `--tl-space-*` rung — shipped in #9211 (with the `admin.module.css` follow-up). What remains is not a sweep but a scale decision.

## The remaining question

A few hardcoded spacing values match no token: `6px` (~4), `10px` (~3), `24px` (~3). The scale skips 6/10 and jumps `20 → 28`. Decide whether to:

- add rungs to the spacing scale for these values, or
- snap each site to the nearest existing token.

Don't invent tokens ad hoc per site — this is one scale decision, like the border-radius scale (#9200) and typography (#9196).

## Out of scope

- Drift (values that already match a rung) — done in #9211.
- Bespoke multi-value paddings such as `8px 40px 36px 40px` (auth / dialog card layout) — deliberately off-scale, left as-is.

PR #9211, opened 2026-06-15
refactor(dotcom): use --tl-space tokens for matching spacing values

jsscclr:
In order to route spacing through the shared scale instead of hardcoded pixels, this PR replaces single-value `padding` / `margin` / `gap` declarations that exactly match a `--tl-space-*` rung with the corresponding token, across the dotcom client CSS modules (52 declarations in 10 files). Part of #9197 (design-system consistency, #9189).

Because each token resolves to the same pixel value it replaced (e.g. `8px` = `var(--tl-space-3)`), there is **no visual change** — this is a pure refactor.

Deliberately left for follow-up:
- **Off-scale values** (`6px`, `10px`, `24px`) — no matching rung; they need the scale decision discussed in #9197.
- **Multi-value shorthands** (e.g. the bespoke auth/dialog card paddings) — left as-is.
- **`button.module.css` / `cta-button.module.css`** — skipped to avoid conflicts with the in-flight button PR #9205.

### Change type

- [x] `improvement`

### Test plan

No visual change is expected — every substituted token equals its previous pixel value. A reviewer can confirm there is no layout difference versus production in the affected areas (sidebar, dialogs, menus, invite/auth dialogs, admin panel).

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +52 / -52  |

### DIFF
--- a/apps/dotcom/client/src/pages/admin.module.css
+++ b/apps/dotcom/client/src/pages/admin.module.css
@@ -40,19 +40,19 @@
 	padding: 24px 32px;
 	display: flex;
 	flex-direction: column;
-	gap: 32px;
+	gap: var(--tl-space-8);
 	overflow: auto;
 }
 
 .adminSection {
 	display: flex;
 	flex-direction: column;
-	gap: 16px;
+	gap: var(--tl-space-5);
 }
 
 .searchContainer {
 	display: flex;
-	gap: 12px;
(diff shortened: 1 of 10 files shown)


mimecuvalo (approved):
ty 🙏

PR #9212, opened 2026-06-15
fix(sync): report SERVER_TOO_OLD when client runs a newer schema

### LINKED ISSUE #6169: CLIENT_TOO_OLD thrown instead of SERVER_TOO_OLD
### What happened?

We were updating tldraw to newer version both on front-end and back-end and encountered that when versions mismatch we always get `CLIENT_TOO_OLD` error, even when it's actually server that uses older version.

- Use newer version of `tldraw` on client (for example: `3.12`) and older version of `tldraw-sync` on server (for example: `3.7`)
- Observer the error on client

ER: `SERVER_TOO_OLD`
AR: `CLIENT_TOO_OLD`

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

max-drake:
In order to give developers an accurate error when their tldraw client and sync server are on mismatched versions, this PR makes the server report `SERVER_TOO_OLD` (instead of always `CLIENT_TOO_OLD`) when the client is actually running a newer schema than the server. Closes #6169.

Previously, when the server couldn't reconcile a client's schema (`getMigrationsSince` returned an error), it always rejected the session with `CLIENT_TOO_OLD` — even when the server was the one running the older SDK. The connect handshake now inspects the schemas: if the client's `schemaVersion`, or any of its sequence versions, is ahead of the server's, the session is rejected with `SERVER_TOO_OLD`; otherwise it stays `CLIENT_TOO_OLD`.

The pre-existing "older client we can't migrate down to" case (a migration isn't record-scoped or has no `down` migration) still reports `CLIENT_TOO_OLD`.

### Change type

- [x] `bugfix`

### Test plan

1. Run a newer `tldraw` client (e.g. 3.12) against an older `@tldraw/sync` server (e.g. 3.7).
2. Observe the client now receives `SERVER_TOO_OLD` instead of `CLIENT_TOO_OLD`.

- [x] Unit tests

### Release notes

- Fix sync version-mismatch handshake reporting `CLIENT_TOO_OLD` when the server is actually the one running an older version; it now correctly reports `SERVER_TOO_OLD`.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +29 / -2   |
| Tests     | +42 / -0   |

### DIFF
--- a/packages/sync-core/src/lib/TLSyncRoom.ts
+++ b/packages/sync-core/src/lib/TLSyncRoom.ts
@@ -838,6 +838,28 @@ export class TLSyncRoom<R extends UnknownRecord, SessionMeta> {
 		this.broadcastPatch(diff)
 	}
 
+	/**
+	 * Work out whether a client we can't reconcile schemas with is running a newer or older SDK
+	 * than us.
+	 */
+	private getVersionMismatchReason(theirSchema: SerializedSchema) {
+		const ourSchema = this.serializedSchema
+
+		if (theirSchema.schemaVersion > ourSchema.schemaVersion) {
+			return TLSyncErrorCloseEventReason.SERVER_TOO_OLD
+		}
+
+		if (theirSchema.schemaVersion === 2 && ourSchema.schemaVersion === 2) {
+			for (const [sequenceId, theirVersion] of Object.entries(theirSchema.sequences)) {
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
nice 👍

PR #9215, opened 2026-06-15
fix(tldraw): preserve external changes during shape gestures

### LINKED ISSUE #6191: Rotating shape while holding it with mouse translates it too
### What happened?

When I rotate a shape while also holding it and translating it, I get weird behavior (see recording).

https://github.com/user-attachments/assets/797d9c84-0a1d-46ef-98ae-ee1fd477cc11

Notice when I rotate the shapes without holding them, they rotate as expected. Only when I pick them up and am still holding them and rotate do I get awkward additional translation behavior. Also, this happens consistently on Safari. In Firefox, it also happens, but if you keep the mouse still during rotation it's fine. Once you start translating again in Firefox, you see the jump.

Minimal reproduction at https://stackblitz.com/edit/vitejs-vite-ztlg36aq?file=src%2FApp.tsx

```typescript
import { Tldraw, TLUiOverrides } from 'tldraw';
import { JustForSandbox } from './JustForSandbox';
import 'tldraw/tldraw.css';
import './styles.css';

const overrides: TLUiOverrides = {
  actions: (editor, actions) => {
    return {
      ...actions,
      rotateCW: {
        id: 'rotate-cw-hover',
        label: 'Rotate clockwise',
        kbd: 'w',
        onSelect: () => {
          editor.rotateShapesBy(editor.getSelectedShapeIds(), Math.PI / 4);
        },
      },
    };
  },
};

  return (
    <div style={{ position: 'fixed', inset: 0 }}>
      <Tldraw autoFocus={false} overrides={overrides}>
        <JustForSandbox />
      </Tldraw>
    </div>
  );
}
```

Also, for what it's worth, even if I implement the rotation manually and bypass `rotateShapesBy` by calculating the new top left corner position with vector math and apply the `updateShape` by setting x, y, and rotation I get the same behavior: works for a stationary shape, but translates while holding and rotating.

Safari, Firefox

### Where did this happen?

The developer package

### Contact Details

visnup@gmail.com

### Code of Conduct

AniKrisn:
This PR fixes a bug where a change made to a shape mid-gesture (for example pressing the rotate shortcut while dragging it) was overwritten by the next pointer move. Closes #6191.

### Before

Translate, resize, rotate, and crop all work from a snapshot of the shapes taken when the gesture starts. Every update recomputes the shapes as `snapshot + gesture delta` and writes the result, reading nothing from the live store. So when a keyboard shortcut, a remote update, or any other command changed a manipulated shape during the gesture, the next update reapplied the stale snapshot and the change was lost.

The reported case is rotating a shape while translating it, but resize, rotate, and crop share the same pattern and the same failure. Here the rotate shortcut is pressed six times mid-drag and the shape slides away from the pointer:

https://github.com/user-attachments/assets/a7d1392b-9e73-4932-963d-c6f5c05f8f63

### After

Each gesture notices when one of its shapes is changed by something other than itself and re-anchors its snapshot on the current shapes before applying the next delta. The external change survives and the gesture continues from the new state without a jump. Same interaction as above:

https://github.com/user-attachments/assets/1ef98f2d-0747-4426-8ed0-3c83fe4b802e

### Implementation notes

A small shared `GestureShapeChangeTracker` does the detection. It registers a `shape` after-change side effect only while a gesture is active, wraps the gesture's own writes in `ignoreChanges` so they are not mistaken for external ones, and exposes a flag the gesture checks at the top of each update. Gesture-owned writes cost one early return per changed shape, other shape updates cost a set lookup, and the O(N) snapshot rebuild only runs when an external change actually hit a tracked shape.

Re-anchoring differs by gesture:

- Translating folds only the position into the existing snapshot, offset by the last applied drag delta so `pagePoint + offset` still equals the live position. Translation never writes rotation, so an external rotation is already preserved.
- Resizing, rotating, and cropping rebuild their snapshot from the current shapes and reset the gesture origin to the current pointer, so the delta resolves to identity at that point.

Bailing to the mark (for example when snapping toggles or cloning starts) clears the tracker, since the restored shapes are not an external change to react to.

### Change type

- [x] `bugfix`

### Test plan

1. Start translating, resizing, rotating, or cropping a shape.
2. Change that shape through an editor command (for example the rotate or nudge shortcut) before releasing the pointer.
3. Keep moving the pointer and confirm the external change is preserved and the shape does not jump.

- [x] Unit tests — the new tests in `translating`, `resizing`, `rotating`, and `cropping` fail on `main` and pass with this change

### Release notes

- Fix shapes losing changes made by other commands while being translated, resized, rotated, or cropped.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +353 / -96 |
| Tests     | +206 / -0  |


steveruizok:
This is too much work here. Make the snapshot include the initial rotation and have that clobber any changes that were made during the interaction.

AniKrisn:
hm @steveruizok not sure if you mean discard the rotation/make it a no-op while dragging? That would go against the reason for the issue (#6191), where the user wants to add 'rotate shape while dragging' behaviour to their app

steveruizok:
> hm @steveruizok not sure if you mean discard the rotation/make it a no-op while dragging? That would go against the reason for the issue (#6191), where the user wants to add 'rotate shape while dragging' behaviour to their app

I think we could solve the current weirdness with a clobber, then look at how to handle it better. We're already doing a lot while translating, atm this diff looks like this it would be an expensive way to check for changes that are happening on every frame during the interaction, which is very rare.

kaneel:
Taking this one over the line.
 
 clobbering is fine but maybe we can extract the logic and apply it to other changes without doing the real fix that would be making this happening just once at a different level.

Issue #9199, opened 2026-06-15
Miscellaneous design-system cleanups in the dotcom client


jsscclr:
Part of #9189.

A bucket of small, low-priority design-system cleanups that don't each warrant their own issue.

## Hardcoded colors in CSS modules (5)

Replace with `var(--tla-color-*)` / `var(--tl-color-*)` tokens:
- `tla/components/dialogs/dialogs.module.css:232` — `color: #666`
- `tla/components/dialogs/dialogs.module.css:239` — `color: #ccc`
- `tla/components/dialogs/dialogs.module.css:256` — `border-top: 1px solid #e0e0e0`
- `tla/components/dialogs/dialogs.module.css:319` — `background-color: rgba(0, 0, 0, 0.5)`
- `tla/components/TlaSidebar/sidebar.module.css:1071` — `background-color: rgba(0, 122, 255, 0.7)`

(The `var(--tl-color-low, rgba(...))` fallbacks at `sidebar.module.css:22, 737` are fine — leave them.)

## Inline `<svg>` icons (2)

Use `TlaIcon` (sprite system) instead of inline SVG:
- `components/IFrameProtector.tsx:67`
- `components/ErrorPage/ErrorPage.tsx:18`

## Custom modal not on the shared dialog primitive (1)

`tla/components/MaybeForceUserRefresh/MaybeForceUserRefresh.tsx` builds its own `position: fixed` modal instead of using `TldrawUiDialog*` like every other dialog. Migrate it, or document why it can't (it shows when the client is outdated, which may have constraints).

## Dead modal CSS

`tla/components/dialogs/dialogs.module.css:313-342` defines `.modalOverlay` / `.modalContent` / `.modalMessage` / `.modalActions` that appear unused. Remove if confirmed dead.

## Acceptance

- The 5 color holes use tokens; the 2 inline SVGs use `TlaIcon`.
- `MaybeForceUserRefresh` uses `TldrawUiDialog` or has a documented exception.
- Dead modal CSS removed.

jsscclr:
Status update — almost all of this has been resolved, much of it independently of this effort:

**Hardcoded colours (5)** — all resolved:
- `#666` (`.helpIcon`), `#ccc` (`.disabledIcon`), and the dead `.modalOverlay` `rgba(0, 0, 0, 0.5)` were removed as dead code in #9261.
- `.divider`'s `#e0e0e0` is now tokenized as `var(--tl-color-divider)` (#9326).
[...]

PR #9217, opened 2026-06-16
refactor(dotcom): use TlaIcon for the iframe-warning external-link icon

jsscclr:
In order to stop hand-maintaining an inline copy of the external-link icon, this PR replaces `IFrameProtector`'s inline `<svg>` with the shared `<TlaIcon icon="external" inline />`. The sprite already has that icon, and `TlaIcon` has no provider dependencies, so it is safe inside that low-dependency iframe guard.

It also documents why `MaybeForceUserRefresh` keeps a custom `position: fixed` overlay rather than `TldrawUiDialog`: it renders at the app-provider root to dim the whole app when the client is outdated, a context where the Radix portal and UI-context hooks (`useTranslation` / `useDirection`) are not available.

### Scope note

This PR originally also tokenized `.disabledIcon` (`#ccc`) and added a `--tla-color-drop-cursor` token for `.dropCursorLine`, but both have since been removed from `main` independently — the `.disabledIcon` block is gone, and `.dropCursorLine` was deleted in #9254 — so only the icon swap and the doc comment remain. The remaining colour and CSS items from this effort were resolved by #9326 (SDK token reuse) and #9261 (dead-code removal).

Part of the design-system consistency effort #9189 / #9199.

### Change type

- [x] `improvement`

### Test plan

A `dotcom-preview-please` preview is requested: https://pr-9217-preview-deploy.tldraw.com/

On the preview, embed a non-allowed tldraw room in an iframe so `IFrameProtector` shows its warning, and confirm the "Visit this page on tldraw.com" link still shows the external-link icon (now rendered via `TlaIcon`).

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +5 / -14   |

### DIFF
--- a/apps/dotcom/client/src/components/IFrameProtector.tsx
+++ b/apps/dotcom/client/src/components/IFrameProtector.tsx
@@ -1,5 +1,6 @@
 import { ReactNode, useEffect } from 'react'
 import { useUrl } from '../hooks/useUrl'
+import { TlaIcon } from '../tla/components/TlaIcon/TlaIcon'
 import { trackEvent } from '../utils/analytics'
 import { getParentOrigin, isInIframe } from '../utils/iFrame'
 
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
🤘

PR #9224, opened 2026-06-16
fix(tldraw): check active selection before tldraw selections

### LINKED ISSUE #9127: Copying text from outside the canvas pastes the selected shape instead
### What happened?

If you try to copy+paste text from outside the canvas with a shape selected, you instead copy the shape itself. This breaks workflows when the canvas is embedded within a larger application, as it is common to copy/paste text from elsewhere on the page.

I noticed it in our own implementation and the reproduced on the `https://examples.tldraw.com/` page:

https://github.com/user-attachments/assets/72cb6e88-d92a-4c33-bed7-c2aa5215470a

1. Select a shape on the canvas.
2. Without deselecting it, select and copy some text elsewhere on the page (outside the canvas).
3. Paste somewhere 

**Expected**: the text I copied is pasted.
**Actual**: the previously-selected canvas shape is pasted — my copied text is gone.

Chrome

### Where did this happen?

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9127 where an active selections in the interface outside of the canvas is not copied because we do copy the selection shapes.

The easy fix for now seems to be checking wether or not we have some active text selections.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. go to examples
2. create a shape
3. select it
4. try and select some text in the examples left sidebar
5. copy 
6. paste it on the canvas

The text you see should be what you selected in the left sidebar, not the json of the currently selected shape 

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where active text selections cannot be copied/cut if we have selected shapes on a canvas

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
+++ b/packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
@@ -145,6 +145,16 @@ function areShortcutsDisabled(editor: Editor) {
 	)
 }
 
+/**
+ * Check if there's an active selection
+ *
+ * @internal
+ */
+function hasActiveTextSelection(editor: Editor) {
+	const selection = editor.getContainerWindow().getSelection()
+	return !!selection && !selection.isCollapsed && selection.toString().trim().length > 0
+}
+
 import { putPastedExternalContent } from './clipboard/putPastedContent'
 export { putPastedExternalContent } from './clipboard/putPastedContent'
 
@@ -869,7 +879,8 @@ export function useNativeClipboardEvents() {
 			if (
 				editor.getSelectedShapeIds().length === 0 ||
 				editor.getEditingShapeId() !== null ||
-				areShortcutsDisabled(editor)
+				areShortcutsDisabled(editor) ||
+				hasActiveTextSelection(editor)
 			) {
 				return
 			}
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/ui/hooks/useClipboardEvents.ts
steveruizok, line 877: This line (which should be renamed `editorIsFocused`) the safest guard against accidental canvas interactions, e.g. with keyboard shortcuts or modifiers). Managing the editor focus is semi-complicated though but most applications will need to do it eventually to prevent canvas kbds from firing while the user's in keyboard-navigating an application-level menu, for example.

steveruizok: Not blocking on this PR, we should still land it, but for if we're getting bug reports like this then we should probably surface editor focus better.

kaneel: ~~We could provide an example page that demonstrate how to handle focus in/out of the canvas.~~ 

we have one example for this https://examples.tldraw.com/editor-focus

kaneel:
@steveruizok thanks for making me doubt, after a bit more testing I figured this was not a good idea after all because one can keep interacting with the canvas while having an active selection and it makes the canvas looks even more buggy than planned.

https://github.com/user-attachments/assets/028e29a8-7c09-42a4-b8e3-094c842468e9

PR #9225, opened 2026-06-16
fix(tldraw): prevent crash when dragging a toolbar tool in read-only mode

### LINKED ISSUE #9116: Dragging a tool from the toolbar crashes with "Shape not found" when the shape can't be created
Dragging a tool from the toolbar to create a shape can crash the app with `Error: Shape not found`. The drag-to-create helper creates a shape and then asserts it exists in the store — if the creation didn't take effect, the assertion throws as an unhandled error and crashes the app (`willCrashApp: true`). The captured event shows it happening in a **read-only room** (`isReadonly: true`), where shape creation is a no-op, on mobile. The creation path should tolerate failure (and/or toolbar dragging should be disabled when the user can't create shapes).

## Sentry context

- Sentry: [LITE-5SS](https://tldraw.sentry.io/issues/LITE-5SS) — **3,722 events, 665 users**, first seen 2025-08-14, still occurring
- Unhandled, `willCrashApp: true`
- Culprit: `onDragFromToolbarToCreateShape` in `packages/tldraw/src/lib/ui/hooks/useTools.tsx:389`
- Sample event: Chrome Mobile 149 / Android 10, read-only multiplayer file, `activeStateNode: root.select.idle`, entered via dotcom's `useExtraToolDragIcons` → `TldrawUiMenuItem` onDragStart

```ts
const id = createShapeId()
opts.createShape(id)
const shape = assertExists(editor.getShape(id), 'Shape not found')  // ← throws
```

## Open questions

1. **Should toolbar drag-to-create be disabled/hidden in read-only contexts, or just no-op gracefully when creation fails?**
   _Awaiting answer._

AniKrisn:
In order to stop the app from crashing when a user drags a tool out of the toolbar in a read-only room, this PR makes `onDragFromToolbarToCreateShape` tolerate the case where the shape isn't created. Closes #9116.

In read-only mode `editor.createShape` is a no-op, so the helper created a shape id, asked for it back, and then `assertExists(editor.getShape(id), 'Shape not found')` threw an unhandled error and crashed the app (`willCrashApp: true`). This surfaced on tldraw.com, where `useExtraToolDragIcons` attaches `onDragStart` to the select/hand/draw tools — and those tools stay visible in read-only mode because they're `readonlyOk`.

The fix:

- Bail out early when the editor is read-only, before any history mark or tool change, so a read-only drag does nothing.
- Replace the `assertExists` invariant with a graceful rollback: if shape creation doesn't take effect for any other reason, bail to the history mark, return to `select.idle`, and exit instead of throwing.

### Change type

- [x] `bugfix`

### Test plan

1. Open a read-only editor that wires `onDragStart` onto a `readonlyOk` tool (e.g. the select tool), as tldraw.com does.
2. Drag that tool out of the toolbar onto the canvas.
3. The app no longer crashes; nothing is created and the editor stays in `select.idle`.

- [x] Unit tests

### Release notes

- Fix a crash ("Shape not found") when dragging a tool from the toolbar in a read-only room.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +11 / -2   |
| Tests     | +84 / -0   |


packages/tldraw/src/lib/ui/hooks/useTools.tsx
```diff
@@ -379,14 +378,24 @@ export function onDragFromToolbarToCreateShape(
...
 	editor.setCurrentTool('select.translating')
 
 	const id = createShapeId()
 	opts.createShape(id)
-	const shape = assertExists(editor.getShape(id), 'Shape not found')
+	const shape = editor.getShape(id)
+	if (!shape) {
+		// Shape creation didn't take effect (for example it was blocked by a side
+		// effect). Bail out gracefully instead of crashing the app.
+		editor.bailToMark(stoppingPoint)
```
steveruizok: Nothing would have changed yet, no need to bail. You might instead move the `select.translating` transition down to after this block.

kaneel:
taking this over the line.

my very – dumb –  confounded face when 
image

mimecuvalo (approved):
👍

PR #9227, opened 2026-06-16
fix(arrows): delete frame-bound arrows when the frame is deleted

### LINKED ISSUE #8978: Arrow bound on both ends to a frame is not deleted with the frame
🧑: Bug: Arrow bound on both ends to a frame itself is not deleted when the frame is deleted. When an arrow is drawn with both terminals bound to the frame shape itself (not to children inside it), the arrow gets parented to the page instead of the frame, so deleting the frame leaves the arrow orphaned. Root cause is in reparentArrow() in packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts:121 which calls findCommonAncestor([startShape, endShape]) where both are the same frame shape; findCommonAncestor starts from the frame's parent and walks up, returning undefined, so the arrow falls back to the page parent.

---

## Description

When an arrow has **both** of its terminals bound to a frame *itself* (rather than to shapes inside the frame), deleting the frame leaves the arrow behind. The arrow is parented to the page instead of the frame, so it is not included in the frame's descendant deletion and survives as an orphaned shape.

This is inconsistent with the case where an arrow is bound to two shapes *inside* a frame — there the arrow is correctly parented to the frame and is deleted along with it.

## Steps to reproduce

1. Create a frame.
2. Select the arrow tool.
3. Draw an arrow from one empty area of the frame to another empty area, so both terminals bind to the frame shape itself (not to any child).
4. Delete the frame.

## Expected behavior

The arrow is deleted together with the frame, since both of its endpoints depend on the frame.

## Actual behavior

The frame is deleted, but the arrow remains on the page (unbound, with its terminals converted to absolute points).

## Root cause

In `reparentArrow()` (`packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts:121`):

```ts
if (startShape && endShape) {
    nextParentId = editor.findCommonAncestor([startShape, endShape]) ?? parentPageId
}
```

When both ends bind to the frame, `startShape` and `endShape` are the **same** shape (the frame). `findCommonAncestor` then takes its multi-shape branch (`packages/editor/src/lib/editor/Editor.ts:5640`), starts the search from the frame's **parent**, and walks upward — so it can never return the frame itself. It returns `undefined`, the `?? parentPageId` fallback applies, and the arrow is parented to the page.

Because `Editor.deleteShapes` only removes the frame's descendants (`Editor.ts:9059`, via `visitDescendants`), a page-parented arrow is never in the deletion set. The binding-isolation hook `onBeforeIsolateFromShape` (`ArrowBindingUtil.ts:96`) fires and unbinds the arrow but does not delete it.

## Suggested fix

Special-case the same-shape binding in `reparentArrow` so the arrow is parented to that shape:

```ts
if (startShape && endShape) {
    nextParentId =
        startShape === endShape
            ? startShape.id
            : editor.findCommonAncestor([startShape, endShape]) ?? parentPageId
}
```

This makes the arrow a child of the frame, so it cascade-deletes naturally — consistent with the two-children-inside case. (An alternative — making `findCommonAncestor` return the shape itself when all inputs are identical — has broader blast radius since it is also used by grouping, so the localized fix is safer.)

## Notes

Reproduced with a unit test in `packages/tldraw`: after drawing an arrow with both ends bound to a frame, `arrow.parentId` is `page:page` rather than the frame id, and the arrow still exists after `deleteShapes([frameId])`.

AniKrisn:
In order to stop a frame deletion from leaving an orphaned arrow behind, this PR deletes an arrow that is bound to a frame on both ends and sits inside that frame when the frame is deleted. Closes #8978.

When an arrow is drawn across empty frame space, both terminals bind to the frame shape itself (not to any child). `reparentArrow` parents such an arrow to the page (`findCommonAncestor` of the same shape returns `undefined`, so the `?? parentPageId` fallback applies), and `deleteShapes` only removes a frame's descendants — so the page-parented arrow was never deleted and survived as an orphan. This is inconsistent with how everything else inside a frame behaves: a shape drawn inside a frame becomes a child and is deleted with it.

The fix adds `ArrowBindingUtil.onBeforeDeleteToShape`. When a frame is deleted, an arrow bound to it on both handles is deleted too — but only if the arrow sits inside the frame's bounds. Such an arrow is part of the frame's contents. An arrow bound to the frame but routed outside it is a connector that happens to use the frame, so it is isolated and survives. The arrow stays parented to the page either way (it is never reparented into the frame), so it is never clipped to the frame's bounds. Arrows bound to the frame on only one end, or bound to a non-frame shape, are isolated as before.

### Change type

- [x] `bugfix`

### Test plan

1. Create a frame.
2. With the arrow tool, draw an arrow from one empty area inside the frame to another empty area, so both ends bind to the frame.
3. Delete the frame — the arrow is deleted with it. Undo restores both.
4. Repeat, then bend the arrow so it routes outside the frame: deleting the frame now leaves the arrow.
5. Draw an arrow from inside the frame to a point outside it (one end bound): deleting the frame leaves the arrow.
6. Repeat step 2 with a rectangle instead of a frame (self-loop arrow): deleting the rectangle leaves the arrow.

- [x] Unit tests
- [ ] End to end tests

### API changes

- Added `ArrowBindingUtil.onBeforeDeleteToShape()`, which deletes an arrow bound to a frame on both handles and contained within it when the frame is deleted.

### Release notes

- Fix an arrow with both ends bound to a frame being left behind when the frame is deleted.

### Code changes

| Core code       | +24 / -0   |
| Tests           | +122 / -0  |

### DIFF
--- a/packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
+++ b/packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
@@ -2,6 +2,7 @@ import {
 	BindingOnChangeOptions,
 	BindingOnCreateOptions,
 	BindingOnShapeChangeOptions,
+	BindingOnShapeDeleteOptions,
 	BindingOnShapeIsolateOptions,
 	BindingUtil,
 	Editor,
@@ -103,6 +104,29 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
 			arrow,
 			terminal: binding.props.terminal,
 		})
+	}
+
+	// When a frame is deleted, delete an arrow bound to it on both ends *only if the arrow
+	// sits inside the frame*. Such an arrow is part of the frame's contents, like any shape
+	// drawn inside it, so it goes with the frame. An arrow that is bound to the frame but
+	// routes outside it is a connector that happens to use the frame, so it is isolated and
+	// survives instead. The arrow stays parented to the page either way, so it is never
+	// clipped to the frame's bounds.
+	override onBeforeDeleteToShape({
+		binding,
+		shape,
+	}: BindingOnShapeDeleteOptions<TLArrowBinding>): void {
+		if (!this.editor.isShapeOfType(shape, 'frame')) return
+		const arrow = this.editor.getShape<TLArrowShape>(binding.fromId)
+		if (!arrow) return
+		const { start, end } = getArrowBindings(this.editor, arrow)
+		if (start?.toId !== shape.id || end?.toId !== shape.id) return
+
+		const frameBounds = this.editor.getShapePageBounds(shape)
+		const arrowBounds = this.editor.getShapePageBounds(arrow)
+		if (frameBounds && arrowBounds && frameBounds.contains(arrowBounds)) {
+			this.editor.deleteShape(arrow.id)
+		}
 	}
 }
 
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -12,6 +12,7 @@ import { BaseFrameLikeShapeUtil } from '@tldraw/editor';
 import { BindingOnChangeOptions } from '@tldraw/editor';
 import { BindingOnCreateOptions } from '@tldraw/editor';
 import { BindingOnShapeChangeOptions } from '@tldraw/editor';
+import { BindingOnShapeDeleteOptions } from '@tldraw/editor';
 import { BindingOnShapeIsolateOptions } from '@tldraw/editor';
 import { BindingUtil } from '@tldraw/editor';
 import { Box } from '@tldraw/editor';
@@ -234,6 +235,8 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
     onAfterChangeToShape({ binding, shapeBefore, shapeAfter, reason }: BindingOnShapeChangeOptions<TLArrowBinding>): void;
     // (undocumented)
     onAfterCreate({ binding }: BindingOnCreateOptions<TLArrowBinding>): void;
+    // (undocumented)
+    onBeforeDeleteToShape({ binding, shape }: BindingOnShapeDeleteOptions<TLArrowBinding>): void;
     // (undocumented)
     onBeforeIsolateFromShape({ binding }: BindingOnShapeIsolateOptions<TLArrowBinding>): void;
     // (undocumented)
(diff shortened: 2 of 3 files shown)


packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
```diff
@@ -119,8 +119,21 @@ function reparentArrow(editor: Editor, arrowId: TLShapeId) {
...
+			// Both ends are bound to the same shape. findCommonAncestor walks up from
+			// the shape's parent and can never return the shape itself, so it would fall
+			// back to the page and orphan the arrow. Parent the arrow to the shape only
+			// if it can contain children (a frame), so it cascade-deletes with the frame.
+			// Otherwise keep it on the page, as an arrow bound to a single shape would be.
+			nextParentId = editor
+				.getShapeUtil(startShape)
+				.canReceiveNewChildrenOfType(startShape, arrow.type)
+				? startShape.id
+				: parentPageId
```
steveruizok: This is unrelated, we don't want to do this. If the arrow is parented to the frame then self-connecting arrows like this don't appear. 

Image

steveruizok: Try creating a side effect that deletes arrows that are bound at both handles to the frame when the frame is deleted.

AniKrisn:
Note on the inside/outside check in `onBeforeDeleteToShape`: it's a bounding-box test, which runs a bit bigger than the shapes themselves — so an arrow that pokes outside the frame by a small margin still gets deleted with it. Fine for now; if we want it exact later, test the arrow's geometry vertices against the frame geometry. Comment in the code.

kaneel:
The behaviour is exactly the same with a box:

https://github.com/user-attachments/assets/fc38df5c-751f-4b96-9965-4cd526464098

And with a frame: 

https://github.com/user-attachments/assets/a8adf539-fc93-45e1-85f6-713bbed8f131

I'm rereading the issue and people are indeed mentioning this is not a problem of frame deletion but a problem of parenting arrow to the frame.

kaneel (commented):
Please see my comment on this, I don't this is the right fix

steveruizok:
> Please see my comment on this, I don't this is the right fix

I believe the issue is that when an arrow is double bound to a frame and the frame is deleted then the arrow is left behind. Also when frames are duplicated without also selecting the arrow, the arrow would not be duplicated. Its confusing because a double bound arrow will move and behave like a child in other ways. Since frames clip their children, parenting the arrow is not the only fix: we'd also have to exclude the arrow from the clipping.

kaneel:
@AniKrisn OK if we close this PR in favour of https://github.com/tldraw/tldraw/pull/9429 ?

PR #9231, opened 2026-06-16
fix(dotcom): keep workspace switcher open when reopened mid-switch

### LINKED ISSUE #9228: Workspace switcher dropdown auto-dismisses when reopened during a workspace switch
### Problem

On tldraw.com, selecting a different workspace from the sidebar workspace switcher updates the URL immediately, then loads the canvas for that workspace's file asynchronously. If you reopen the workspace switcher quickly — before that new canvas has finished loading — the dropdown auto-dismisses on its own.

### Steps to reproduce

1. Be in a workspace with at least one other workspace available in the switcher.
2. Open the workspace switcher and pick a different workspace.
3. The URL updates right away; the new canvas is still loading.
4. Immediately reopen the workspace switcher.
5. Once the new canvas finishes loading (the load that appends the `?d=` deep-link param), the dropdown closes by itself.

### Expected

Reopening the switcher keeps it open until the user dismisses it.

### Actual

The dropdown dismisses itself when the incoming canvas finishes loading.

### Notes

The switcher stores its open/closed state in the editor's global menu registry, and that key is scoped to the currently mounted file editor's context. The sidebar receives the active file's editor via context, and that editor instance is replaced on every workspace/file switch. When the outgoing editor is disposed as the new canvas loads, it clears the menu state for its context — which is the same key the reopened switcher is using — so the dropdown closes.

jsscclr:
In order to fix #9228, this PR stops the sidebar workspace switcher from dismissing itself when reopened during a workspace switch.

Selecting a workspace navigates immediately, but the target file's canvas loads asynchronously and (deep links are on for the file route) rewrites the URL with a `?d=` param once its editor mounts. The switcher stored its open/closed state via `useMenuIsOpen`, which scopes the menu id to the active file editor's `contextId`. The sidebar receives that editor through `globalEditor`/`EditorContext`, and the editor instance is replaced on every workspace/file switch (`TlaEditor` remounts on `fileSlug`). So when the outgoing editor was disposed as the incoming canvas loaded, its `clearOpenMenus()` cleared the menu state for its context — the same key the just-reopened switcher was using — and, separately, the `globalEditor` flip changed the menu id out from under the open dropdown. Either way the dropdown closed on its own.

The switcher is a sidebar-level control whose lifetime spans editor remounts, so its open state should not be tied to any per-file editor. This switches both the switcher and the layout that reads it to `useGlobalMenuIsOpen` with a stable, editor-independent id. Outside-click dismissal is unchanged: the dropdown's Radix `modal` dismissable layer still closes it on any outside pointer-down, including canvas clicks.

This also addresses the root cause behind several retry-loop workarounds in the dotcom e2e `Sidebar` page object (added to tolerate "a settling re-render dismissed the menu"); those are left in place here and can be simplified separately.

### Change type

- [x] `bugfix`

### Test plan

1. On tldraw.com, be in a workspace with at least one other workspace available in the switcher.
2. Open the workspace switcher and select a different workspace.
3. While the new canvas is still loading, immediately reopen the workspace switcher.
4. Confirm it stays open (previously it dismissed itself once the canvas finished loading).

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fix the tldraw.com sidebar workspace switcher dismissing itself when reopened right after switching workspaces.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Tests   | +35 / -0   |
| Apps    | +10 / -4   |

### DIFF
--- a/apps/dotcom/client/src/tla/layouts/TlaSidebarLayout/TlaSidebarLayout.tsx
+++ b/apps/dotcom/client/src/tla/layouts/TlaSidebarLayout/TlaSidebarLayout.tsx
@@ -1,5 +1,5 @@
 import React, { ReactNode, useCallback, useLayoutEffect, useRef } from 'react'
-import { clamp, tltime, useMenuIsOpen, useQuickReactor } from 'tldraw'
+import { clamp, tltime, useGlobalMenuIsOpen, useQuickReactor } from 'tldraw'
 import { TlaSidebarToggle } from '../../components/TlaSidebar/components/TlaSidebarToggle'
 import { TlaSidebarToggleMobile } from '../../components/TlaSidebar/components/TlaSidebarToggleMobile'
 import { TlaSidebar } from '../../components/TlaSidebar/TlaSidebar'
@@ -28,7 +28,8 @@ export function TlaSidebarLayout({
 }) {
 	const isSidebarOpen = useIsSidebarOpen()
 	const isSidebarOpenMobile = useIsSidebarOpenMobile()
-	const [isWorkspaceSwitcherOpen] = useMenuIsOpen('sidebar-workspace-switcher')
+	// Must match the stable id used by TlaSidebarWorkspaceSwitcher (see note there).
+	const [isWorkspaceSwitcherOpen] = useGlobalMenuIsOpen('sidebar-workspace-switcher')
 
 	usePreventAccidentalDrops()
 
--- a/apps/dotcom/client/e2e/tests/smoke/workspaces.spec.ts
+++ b/apps/dotcom/client/e2e/tests/smoke/workspaces.spec.ts
@@ -33,6 +33,41 @@ test.describe('workspaces', () => {
 			await page.keyboard.press('Escape')
 		})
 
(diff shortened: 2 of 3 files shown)


frolic (commented):
nice! easy

frolic (approved):

apps/dotcom/client/src/tla/layouts/TlaSidebarLayout/TlaSidebarLayout.tsx
```diff
@@ -28,7 +28,8 @@ export function TlaSidebarLayout({
 }) {
 	const isSidebarOpen = useIsSidebarOpen()
 	const isSidebarOpenMobile = useIsSidebarOpenMobile()
-	const [isWorkspaceSwitcherOpen] = useMenuIsOpen('sidebar-workspace-switcher')
+	// Must match the stable id used by TlaSidebarWorkspaceSwitcher (see note there).
```
kostyafarber, line 31: i would of made this a constant somewhere, if this has to match cross file, instead of the magic string

PR #9233, opened 2026-06-16
fix(examples): blur canvas when pointing into sidebar

kaneel:
This is a quick spike to fix the example pages when user click inside the sidebar and the editor isnot blurred.
This was raised in https://github.com/tldraw/tldraw/issues/9127 and it's not really a bug but it raises suspicions this is one when the responsibility of focusing / blurring the editor is left to the consumer of the SDK

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Go to any example
2. Add a shape
3. keep that shape selected
4. select text in the sidebar
5. copy it 
6. paste it into the canvas

The pasted text should be what was selected in the sidebar, not the JSON of a shape or the shape itself

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed examples focusing/blurring the canvas when interacting with the sidebar

### DIFF
--- a/apps/examples/src/exampleTldraw.tsx
+++ b/apps/examples/src/exampleTldraw.tsx
@@ -1,0 +1,24 @@
+import type { Editor } from '@tldraw/editor'
+import { useCallback } from 'react'
+import { Tldraw as TldrawBase, type TldrawProps } from '../../../packages/tldraw/src/lib/Tldraw'
+import { isExampleSiteSidebarRoute, wireExampleSiteFocus } from './wireExampleSiteFocus'
+
+export function Tldraw(props: TldrawProps) {
+	const { onMount, autoFocus, ...rest } = props
+	const manageFocus = isExampleSiteSidebarRoute()
+
+	const handleMount = useCallback(
+		(editor: Editor) => {
+			const teardownFocus = manageFocus ? wireExampleSiteFocus(editor) : undefined
+			const teardownOnMount = onMount?.(editor)
+
+			return () => {
+				teardownFocus?.()
+				teardownOnMount?.()
+			}
+		},
+		[onMount, manageFocus]
+	)
+
+	return <TldrawBase {...rest} autoFocus={manageFocus ? false : autoFocus} onMount={handleMount} />
+}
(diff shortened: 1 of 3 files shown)


steveruizok:
I think you could do this with an `onPointerEnter` on both elements (sidebar, canvas) if that's what you're after?

mimecuvalo:
I think the intent of that particular issues is less the examples app but maybe any instance of tldraw where it's used in a larger application. As a comparsion, does this same behavior happen on dotcom (w.r.t. the sidebar on dotcom)?

kaneel:
> I think the intent of that particular issues is less the examples app but maybe any instance of tldraw where it's used in a larger application. As a comparsion, does this same behavior happen on dotcom (w.r.t. the sidebar on dotcom)?

we cannot select things in the sidebars on dotcom. the intent of this PR is to avoid having users suggesting there's a "bug" and file an issue because they cannot select things and paste them within their own application usage because "it also happens in the examples"

max-drake:
@mimecuvalo theres no selectable text in the dotcom sidebar but this happens in the agent starter I know

mimecuvalo:
> > I think the intent of that particular issues is less the examples app but maybe any instance of tldraw where it's used in a larger application. As a comparsion, does this same behavior happen on dotcom (w.r.t. the sidebar on dotcom)?
> 
> we cannot select things in the sidebars on dotcom. the intent of this PR is to avoid having users suggesting there's a "bug" and file an issue because they cannot select things and paste them within their own application usage because "it also happens in the examples"

yeah, we shouldn't target just fixing examples, this would be good to fix in general. people use tldraw in their applications and also have their own sidebars. it would be good to establish a pattern here of opting in/out to tldraw grabbing copy/paste behavior across the whole document. or, probably the better fix is that tldraw should check the DOM element being interacted with and if it's outside of tldraw it should relinquish control to the parent app.

kaneel:
@mimecuvalo much easier to let people figure their rules for this than trying to relying on just the DOM for this. I think it could potentially clash with the logic that's centered on the SDK… but it's worth a shot!

kaneel:
Closing this

PR #9234, opened 2026-06-16
fix(editor): treat ctrl+click as right-click on macOS

### LINKED ISSUE #8217: Ctrl+click does not open context menu on locked shapes
## Description

When right-clicking a locked shape on the canvas, the full context menu opens as expected. However, when using **Ctrl+Left click**, the shape context menu does not appear.

Edit: this bug also occurs when trying to right click locked shapes using coarse pointer long press on android tablets and ipad, but only when long pressing the shape's edge
Edit: this also occurs with locked frames 

## Steps to reproduce

1. Lock any shape on the canvas
2. Right-click the shape → context menu appears ✅
3. Hold Ctrl and left-click the same shape → context menu does **not** appear ❌

## Environment

- macOS (where Ctrl+click = right-click convention)
- Affects all browsers on macOS

max-drake:
👩: this is a broad fix to #8217, and instead of targeting it specifically, makes the ctrl left click path on mac get treated as a right click everywhere
this also fixes behavior like right clicking on locked shapes as mentioned in #8217, but also right click to pan, which didn't work with ctrl+left, and also ctrl+left clicking while a context menu was already open, which would previously close the context menu instead of opening a new one in the place of the mouse pointer

---

In order to make ctrl+click behave like a real right-click on macOS, this PR normalizes ctrl+left-click (button `0` with `ctrlKey` and no `metaKey`) to button `2` throughout the editor's pointer handling. Previously this combination was treated as a plain left click, so ctrl+clicking a locked shape didn't select it or open the context menu the way a right-click does. Closes #8217.

A shared `getRightClickLikeButton` / `isRightClickLikeEvent` helper (`packages/editor/src/lib/utils/pointer.ts`) centralizes the macOS check, and it's applied consistently across the canvas pointer events, the menu click-capture component, the pointer-info builder, and the native context-menu suppression path. The test `Driver` applies the same normalization so drivers reproduce macOS behavior. The fix is scoped to the macOS ctrl+click path; touch long-press handling is unchanged.

### Change type

- [x] `bugfix`

### Test plan

1. On macOS, lock a shape on the canvas.
2. Hold Ctrl and left-click the locked shape.
3. The shape is selected and the context menu opens, matching a real right-click.
4. Repeat with an unlocked shape, and with right-click panning both enabled and disabled, to confirm no regressions.

- [x] Unit tests

### Release notes

- Fix ctrl+click on macOS not selecting locked shapes or opening the context menu; it now behaves like a right-click.

### Code changes

| Core code      | +47 / -18  |
| Tests          | +13 / -0   |

### DIFF
--- a/packages/editor/src/lib/utils/pointer.ts
+++ b/packages/editor/src/lib/utils/pointer.ts
@@ -1,11 +1,18 @@
 import { tlenv } from '../globals/environment'
 
 /** @internal */
-export function isRightClickLikeEvent(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
+interface PointerLike {
+	button: number
+	ctrlKey: boolean
+	metaKey: boolean
+}
+
+/** @internal */
+export function isSecondaryClickEvent(e: PointerLike) {
 	return e.button === 2 || (tlenv.isDarwin && e.button === 0 && e.ctrlKey && !e.metaKey)
 }
 
 /** @internal */
-export function getRightClickLikeButton(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
-	return isRightClickLikeEvent(e) ? 2 : e.button
+export function getPointerEventButton(e: PointerLike) {
+	return isSecondaryClickEvent(e) ? 2 : e.button
 }
--- a/packages/editor/src/lib/utils/pointer.ts
+++ b/packages/editor/src/lib/utils/pointer.ts
@@ -1,0 +1,11 @@
+import { tlenv } from '../globals/environment'
+
+/** @internal */
+export function isRightClickLikeEvent(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
+	return e.button === 2 || (tlenv.isDarwin && e.button === 0 && e.ctrlKey && !e.metaKey)
+}
+
+/** @internal */
+export function getRightClickLikeButton(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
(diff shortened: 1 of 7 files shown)


packages/editor/src/lib/utils/pointer.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/** @internal */
+export function isRightClickLikeEvent(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
+	return e.button === 2 || (tlenv.isDarwin && e.button === 0 && e.ctrlKey && !e.metaKey)
+}
+
+/** @internal */
+export function getRightClickLikeButton(e: { button: number; ctrlKey: boolean; metaKey: boolean }) {
+	return isRightClickLikeEvent(e) ? 2 : e.button
+}
```
kaneel, line 18: `getRightClickLikeButton` seems a bit wrong as it returns the "button" value, getPointerEventButton may be better? and `e` should be PointerEvent as well

kaneel: Actually, you could just declare the `PointerLike` type once, and use it in both.

kaneel: I'd pay attention to naming here and call it a "secondary click", because left handed mouse users are real too!

kaneel (approved):
👍 some comments but should be good to go!

max-drake:
@kaneel added some tests and merging

PR #9237, opened 2026-06-16
feat(dotcom): localize the welcome file via a build step

jsscclr:
Reworks the new-workspace welcome file's localization (originally spiked in this PR) to resolve translations at **build time** rather than at runtime, the same way dotcom already compiles its UI strings.

## Why

The original spike localized the welcome copy in the sync worker per request — fetching the lokalise catalogs over HTTP and filling the copy shapes, plus a queue job that generated per-locale R2 variants when an admin set a welcome template. The translations are static, so all of that runtime machinery (catalog fetch, `getAppOrigin` cross-origin coupling, queue, R2 variants) was resolving something a build step can resolve once.

It also turned out the spike localized **nothing** in practice: its message bindings used stable shape ids and English strings that don't exist in the committed default snapshot, so at runtime they matched zero shapes and silently fell back to English.

## How it works

- **Manifest** (`welcomeCopy.ts`) — the welcome strings, derived from the actual snapshot (instructional captions, depicted UI labels, example file names). Each carries the exact English (with `<strong>` bold / `<mark>` highlight runs), which doubles as the key used to locate its shape(s).
- **1:1 binding** — depicted app-UI labels are *shared*: they carry no `welcome.*` id and instead resolve to the app's existing lokalise message id by matching English, so the illustration shows the app's official translation in every locale with no welcome-specific translation (and the build fails if a depicted app string disappears). Owned strings (captions, file names, UI labels with no app equivalent) keep a `welcome.*` id. The illustration's "Workspace settings" was renamed to "Manage" to match the real sidebar item.
- **Build step** (`i18n:welcome`, part of `build-i18n`) — writes the owned welcome English into `en.json` for translators, then bakes each locale's localized message per shape into a committed `localizedWelcomeCopy.json`. It **asserts every manifest string matches a snapshot shape**, so reworded art fails the build instead of shipping stale English.
- **Seed time** — the client tags the welcome `createSource` with the creator's locale; the worker resolves the snapshot with a pure synchronous lookup over the baked variants, rebuilding each shape's richText from the baked message. No DB, catalog fetch, queue, or R2.
- The welcome copy is never rendered through react-intl (it lives in a document snapshot), so the build *is* the translation.

Also removes the admin-set welcome-template feature (table, schema, routes, UI) — new workspaces always seed from the committed default, now localized at build time. Migration 038 drops `welcome_template`.

## Status

The depicted app-UI labels now localize in **all 48 locales** automatically, by reusing the app's existing translations (the 1:1 binding). The welcome **captions** carry no translations yet, so they seed in English until lokalise translations land. The worker rebuilds each shape's richText from the baked message at seed time (the artifact stores messages, not full docs, to stay within the worker bundle budget). Covered by unit tests (codec round-trip, injection/rebuild, the resolver's locale fallback, and the manifest↔snapshot drift guard).

## Preview testing

Preview: https://pr-9237-preview-deploy.tldraw.com/

1. Sign in and create a new workspace from the workspace menu in the sidebar — its first file is the welcome file, seeded from the committed default (renders in English).
2. Open `/admin` — the "Welcome template" section is gone (feature removed).
3. To see localization: set your language (user menu) to **German** or **Spanish**, then create a new workspace. The welcome illustration's UI labels render localized — "Mein Arbeitsbereich", "Entfernen", "Suchen…" — pulled from the app's own translations; the captions stay English (no caption translations yet).


mimecuvalo (commented):
thx for tackling this!

steveruizok:
Think we'll do this one? I think we could get away with not doing it.

mimecuvalo:
> Think we'll do this one? I think we could get away with not doing it.

I think it's still worth doing, esp. since it'll be a crucial onboarding for folks. happy to work with @jsscclr to take it over the line.

jsscclr:
Managed to get something working- it means removing the current escape hatch of where an admin user can override the default template with their own design, but means we can leverage the existing build system. The main task at the moment is just the design treatment to handle different locales (make things wider basically). So think we can get this shipped without too much more effort!

mimecuvalo (approved):
let's goooo

mimecuvalo:
closing up some PRs that are eating up some CF usage 😬 , reopen if needed later!

PR #9240, opened 2026-06-16
docs(examples): add multiplayer platformer example

frolic:
In order to demonstrate real-time multiplayer, a custom shape, and a per-frame game loop together, this PR adds a "Multiplayer platformer" example. The canvas becomes a shared 2D platformer level where each collaborator controls a player with WASD/arrow keys and space to jump. The floor, platforms, crates, and the players themselves are all ordinary tldraw shapes, so anyone can select, drag, resize, and delete them while the game runs.

Each client runs gravity and axis-aligned collision detection for its own player only and writes the result to the synced store; every other player's position arrives over `useSyncDemo`. Position updates use `history: 'ignore'` so the physics don't pollute the undo stack, and a resting player stops emitting updates so an idle game produces no sync traffic. Game keys are claimed with a window capture-phase listener that stops propagation, so space doesn't pan, the arrows don't nudge, and W/A/S/D don't switch tools. When a collaborator leaves, their player is despawned by reconciling player shapes against `editor.getCollaborators()`, with a short grace period so brief disconnects don't remove a player that's about to reappear.

### New examples

- `collaboration/multiplayer-platformer` — a shared 2D platformer with per-player gravity, AABB collision, keyboard controls, and despawn-on-leave.

### Change type

- [x] `other`

### Test plan

1. Run `yarn dev` and open `localhost:5420/multiplayer-platformer`.
2. Click the canvas, then move with A/D or ←/→ and jump with space — the player should fall under gravity, land on platforms, and be blocked by walls.
3. Drag the floor, platforms, and players around; physics should keep working against the moved shapes.
4. Open the room in a second window (Copy link) — each window controls its own player. Close one window and its player should disappear from the other after a few seconds.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add a multiplayer platformer example showing a custom shape, a game loop with gravity and collision, keyboard input, and tldraw sync.

### Code changes

| Documentation | +822 / -0  |

### DIFF
--- a/apps/examples/src/examples/collaboration/multiplayer-platformer/MultiplayerPlatformerExample.tsx
+++ b/apps/examples/src/examples/collaboration/multiplayer-platformer/MultiplayerPlatformerExample.tsx
@@ -1,0 +1,378 @@
+import { useSyncDemo } from '@tldraw/sync'
+import { useEffect, useRef } from 'react'
+import { Box, Editor, TLComponents, Tldraw, TldrawUiButton, useEditor } from 'tldraw'
+import 'tldraw/tldraw.css'
+import { type Box as Collider, type PlayerMotion, stepPlayer } from './physics'
+import './multiplayer-platformer.css'
+import { PLAYER_TYPE, PlayerShape, PlayerShapeUtil } from './PlayerShape'
+
+// [1]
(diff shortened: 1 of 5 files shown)


mimecuvalo:
would love to see a demo later :)

PR #9252, opened 2026-06-17
feat(tldraw): add showAttribution option to note shape util

steveruizok:
In order to let SDK consumers turn off the note shape's authorship badge, this PR adds a `showAttribution` option to `NoteShapeUtil`. The note shape renders the display name of the user who first edited its text in the corner (and in SVG exports); previously this was always on. The new option defaults to `true`, so existing behavior is unchanged, and it can be disabled via `NoteShapeUtil.configure({ showAttribution: false })`.

### Component props

| Option | Type | Description |
| --- | --- | --- |
| `showAttribution` | `boolean` | Whether to show the attribution (display name of the user who first edited the note's text) in the corner of the note. Defaults to `true`. |

### Example

```ts
import { NoteShapeUtil, Tldraw } from 'tldraw'

const shapeUtils = [NoteShapeUtil.configure({ showAttribution: false })]

	return <Tldraw shapeUtils={shapeUtils} />
}
```

The existing "Shape options" example (`apps/examples/src/examples/configuration/configure-shape-util`) is extended to demonstrate the option.

### Change type

- [x] `api`

### Test plan

1. Add a note shape and edit its text in a multiplayer or signed-in context so attribution appears.
2. Confirm the attribution name shows in the note corner and in SVG/PNG exports by default.
3. Configure `NoteShapeUtil.configure({ showAttribution: false })` and confirm the attribution no longer renders on canvas or in exports.

- [x] Unit tests

### Release notes

- Add a `showAttribution` option to `NoteShapeUtil` to hide or show the note's text attribution.

### API changes

- Added `showAttribution` to `NoteShapeOptions`.

### Code changes

| Core code       | +11 / -3   |
| Documentation   | +15 / -6   |

### DIFF
--- a/apps/docs/content/sdk-features/attribution.mdx
+++ b/apps/docs/content/sdk-features/attribution.mdx
@@ -91,6 +91,14 @@ Attribution is opt-in per shape type. Each shape util decides what to track and 
 
 The built-in note shape tracks who first added text. When a user types in an empty note, `NoteShapeUtil` sets the `textFirstEditedBy` prop to the current user's ID via `editor.getAttributionUserId()`. Subsequent edits by other users don't change this value. Clearing the text resets it to `null`. The note renders the first editor's name as a small label in the corner.
 
+To hide this label, configure the note shape util with `showAttribution: false`:
+
+```tsx
+import { NoteShapeUtil } from 'tldraw'
+
+const NoteUtil = NoteShapeUtil.configure({ showAttribution: false })
+```
+
 ## Reading attribution
 
 The [Editor](?) provides three methods for working with attribution. These methods check the `TLUserStore` first, then fall back to `user:` records in the store:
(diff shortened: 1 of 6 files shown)


kaneel (approved):
LGTM, maybe worth a discussion on wether we want this true by default for SDK users, otherwise, good to go!

mimecuvalo (changes requested):
I know you don't like 'request changes' @steveruizok but I want to make sure we have a API design discussion about this before landing — thx!

steveruizok:
@mimecuvalo to put up a PR trying out an alternative, e.g. components={{ NoteShapeAttribution: <MyThing> }}

tedyoung:
I could really use this, I was pretty unhappy when attribution appeared with no obvious way to turn it off in tldraw-offline!

PR #9260, opened 2026-06-17
fix(tlschema): make editingShapeId ephemeral so undo/redo can't restore a deleted editing shape

### LINKED ISSUE #9113: Undo can crash with "Entered editing state without an editing shape"
Pressing undo on tldraw.com can crash the editor with the unhandled error `Entered editing state without an editing shape`. The undo restores a snapshot in which a shape was being edited, which makes the editor switch into the text-editing tool state — but by the time the transition happens there is no editing shape to enter (the recorded event shows `editingShapeId: null` and an image shape selected when the crash fired on Ctrl+Z). This is SDK behavior in the select tool's editing state, so it affects SDK consumers as well as tldraw.com. It is unhandled (potential app crash) and has produced ~860 events from 60 users since February 2024. The goal is for undo/redo to never enter the editing state without a valid editing shape.

Sentry: [LITE-3E5](https://tldraw.sentry.io/issues/LITE-3E5)

## Open questions

1. **When undo restores an editing snapshot whose shape can't be edited, should the editor fall back to the idle select state, or try to preserve/restore editing when possible?**
   _Awaiting answer._

kaneel:
Fixes https://github.com/tldraw/tldraw/issues/9113

In order to stop undo/redo from crashing the editor, this PR makes the select tool refuse to enter its editing state without a valid editing shape.

Setting the editing shape is not recorded in history (`setEditingShape` runs with `history: 'ignore'`), but the cleanup that clears `editingShapeId` when a shape is deleted *is* recorded. When a shape is created, edited, and deleted within a single history entry, the shape's add/remove cancel out during squashing while the `editingShapeId` change survives. Undoing then restores `editingShapeId` pointing at a shape that no longer exists. The `instance_page_state.afterChange` side effect reacted to that by transitioning into `select.editing_shape`, and `EditingShape.onEnter` threw `Entered editing state without an editing shape`, crashing the app. This is SDK behavior in the select tool, so it affected SDK consumers as well as tldraw.com (~860 events from 60 users since February 2024).

This PR fixes it at the source: in `defaultSideEffects`, before entering the editing state in response to an `editingShapeId` change, it resolves the editing shape. If the shape doesn't exist, it clears the stale `editingShapeId` and stays put instead of transitioning into `select.editing_shape`. This covers every store-driven path (undo/redo, snapshot load, remote sync, programmatic writes), which are the only ways a stale id can appear. `EditingShape.onEnter` keeps its `throw` as an invariant assertion — with this guard it can no longer be reached without a valid editing shape.

### Change type

- [x] `bugfix`

### Test plan

1. Create a shape, begin editing it, and delete it within a single undo step.
2. Press undo, then redo.
3. The editor stays in `select.idle` and does not crash.

- [x] Unit tests

### Release notes

- Fix a crash that could occur when undo or redo restored an editing state for a shape that no longer exists.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +16 / -4   |
| Tests     | +28 / -0   |

### DIFF
--- a/packages/tldraw/src/test/SelectTool.test.ts
+++ b/packages/tldraw/src/test/SelectTool.test.ts
@@ -517,6 +517,34 @@ describe('When pressing enter on a selected shape', () => {
 	})
 })
 
+describe('When undo/redo restores an invalid editing shape', () => {
+	// Regression: https://github.com/tldraw/tldraw/issues/9113
+	// Setting the editing shape is not recorded in history, but clearing it on delete is.
+	// When create + edit + delete of the same shape collapse into a single history entry,
+	// the shape's add/remove cancel out while the editingShapeId update survives. Undoing
+	// then restores editingShapeId pointing at a shape that no longer exists, which used to
+	// crash with "Entered editing state without an editing shape". The editor must never
+	// enter the editing state without a valid editing shape.
+	it('does not crash when undo restores editingShapeId for a deleted shape', () => {
+		const id = createShapeId()
+		editor.markHistoryStoppingPoint('start')
+		editor.createShape({ id, type: 'geo', x: 0, y: 0, props: { w: 100, h: 100 } })
+		editor.setEditingShape(id)
+		editor.deleteShapes([id])
+
+		expect(() => editor.undo()).not.toThrow()
+
+		editor.expectToBeIn('select.idle')
+		expect(editor.getEditingShapeId()).toBe(null)
+		expect(editor.getShape(id)).toBeUndefined()
+
+		// Redoing back through the same history entry must also stay safe.
+		expect(() => editor.redo()).not.toThrow()
+		editor.expectToBeIn('select.idle')
+		expect(editor.getEditingShapeId()).toBe(null)
+	})
+})
+
 // it('selects the child of a group', () => {
 //   const id1 = createShapeId()
 //   const id2 = createShapeId()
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/tools/SelectTool/childStates/EditingShape.ts
```diff
@@ -34,7 +34,14 @@ export class EditingShape extends StateNode {
...
 		const editingShape = this.editor.getEditingShape()
-		if (!editingShape) throw Error('Entered editing state without an editing shape')
+		if (!editingShape) {
+			// We should never enter the editing state without a valid editing shape, but if
+			// we somehow do (e.g. undo/redo restoring a stale editingShapeId), don't crash —
+			// clear the stale id and fall back to idle.
+			this.editor.setEditingShape(null)
+			this.parent.transition('idle', info)
+			return
+		}
```
kaneel: On second thought this is super daft, I still believe the error was here for a reason and we should not be in that state at all, trying to patch all potential errors < trying to prevent these errors from even occuring

packages/tldraw/src/lib/defaultSideEffects.ts
```diff
@@ -31,15 +31,20 @@ export function registerDefaultSideEffects(editor: Editor) {
 
 				if (prev.editingShapeId !== next.editingShapeId) {
 					if (!prev.editingShapeId && next.editingShapeId) {
-						if (!editor.isIn('select.editing_shape')) {
+						const shape = editor.getEditingShape()
+						if (!shape) {
```
mimecuvalo: let's set this to `const shape = editor.getShape(next.editingShapeId)` to make this explicit the intention here. currently it's unclear why setEditingShape to null would do anything (when it's kinda returning null/undefined)

kaneel: we're setting the editing shape back to null because there are side effects to that which will do some cleaning with recording in the history.

mimecuvalo (approved):
👍 nice

PR #9262, opened 2026-06-17
fix(editor): handle text spans with no layout rects

### LINKED ISSUE #9112: Text measurement throws "Cannot read properties of undefined (reading 'top')" when measuring frame titles
🗣️: Sentry shows text measurement intermittently crashing with "Cannot read properties of undefined (reading 'top')" in TextManager while measuring frame titles (LITE-58J plus duplicate LITE-6FY, ~680 events / ~480 users since March 2025) — let's create an issue for it with the Sentry context.

---

The SDK's text measurement intermittently throws `TypeError: Cannot read properties of undefined (reading 'top')` while measuring text spans, surfacing on tldraw.com when frame title sizes are computed during shape geometry calculation (e.g. while hovering shapes). The crash happens when a measured character produces no layout rectangles in the browser, which the span measurement doesn't tolerate. It has produced ~680 events from ~480 users since March 2025 across two duplicate Sentry issues, on stock browsers (Chrome, Edge). The bug lives in the editor package, so it affects SDK consumers as well as tldraw.com. The goal is for text span measurement to handle this case without throwing.

Sentry: [LITE-58J](https://tldraw.sentry.io/issues/LITE-58J), [LITE-6FY](https://tldraw.sentry.io/issues/LITE-6FY)

## Open questions

1. **Is a defensive fix (skip characters that report no layout rects) acceptable, or do you want the underlying cause of the empty measurements investigated first?**
   _Awaiting answer._

Confidence: 80%, ready to get started.

kaneel:
In order to stop text measurement from intermittently crashing, this PR makes span measurement tolerate characters that report no layout rectangles. While measuring text spans (e.g. frame title sizes computed during shape geometry calculation), `Range.getClientRects()` can return an empty list for certain characters in some browsers. The previous code read `rects[rects.length - 1]` (i.e. `rects[-1]` → `undefined`) and then threw `TypeError: Cannot read properties of undefined (reading 'top')`. This surfaced on stock browsers (Chrome, Edge) as ~680 Sentry events across ~480 users since March 2025. Now we skip those characters while still advancing the text-node index so later characters stay aligned.

Closes #9112

### Change type

- [x] `bugfix`

### Test plan

1. Measure text spans where a character produces no layout rectangles (covered by the new unit test).

- [x] Unit tests

### Release notes

- Fix an intermittent crash in text measurement when a measured character produced no layout rectangles.

### DIFF
--- a/packages/editor/src/lib/editor/managers/TextManager/TextManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/TextManager/TextManager.test.ts
@@ -247,6 +247,38 @@ describe('TextManager', () => {
 			expect(result.spans).toHaveLength(0)
 		})
 
+		it('should skip characters that produce no layout rectangles', () => {
+			const RangeMock = global.Range as unknown as ReturnType<typeof vi.fn>
+			RangeMock.mockImplementationOnce(function () {
+				return {
+					setStart: vi.fn(),
+					setEnd: vi.fn(),
+					// simulate a browser returning no rects for a measured character,
+					// which previously crashed with "Cannot read properties of
+					// undefined (reading 'top')". See issue #9112.
+					getClientRects: vi.fn(() => []),
+				}
+			})
+
+			const mockTextNode = {
+				nodeType: 3, // TEXT_NODE
+				textContent: 'Hello',
+			}
+
+			const mockElementWithText = {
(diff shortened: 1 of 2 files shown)


kaneel:
not ready, absolutely not the fix :D

PR #9266, opened 2026-06-17
fix(tldraw): rely on grapheme segmenter to iterate over textContent

### LINKED ISSUE #9112: Text measurement throws "Cannot read properties of undefined (reading 'top')" when measuring frame titles

kaneel:
Fixes https://github.com/tldraw/tldraw/issues/9112

Error occurs when iterating over text elements that have returns no bounding boxes, this could be a case of some zero width character and/or some emojis that are composed of two characters in order to compose a grapheme.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`
### Test plan

1. Add a frame and give it a name ending in an emoji or accented character (e.g. "Hello 👨‍👩‍👧"). Hover the frame and confirm no console error and the title sizes correctly.
2. Add a text shape containing an emoji ZWJ sequence, a flag (🇫🇷), and an accented letter, then export the selection as SVG. Confirm the exported text keeps all glyphs intact (no dropped accents or broken emoji).
3. Confirm existing text wrapping, truncation, and multiline behavior is unchanged.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improve text measurement by iterating over grapheme clusters instead of individual characters, fixing an occasional crash and keeping emoji, flags, and accented characters intact.

### DIFF
--- a/apps/examples/e2e/baselines/fps-baselines.json
+++ b/apps/examples/e2e/baselines/fps-baselines.json
@@ -1,135 +1,146 @@
 {
-	"baselines": {
-		"linux-1280x720": {
-			"rotate_shapes": {
-				"avgFps": 25,
-				"minFps": 8,
-				"maxFps": 58,
-				"timestamp": "2025-12-06T22:18:16.000Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"drag_shapes": {
-				"avgFps": 54,
-				"minFps": 26,
-				"maxFps": 69,
-				"timestamp": "2025-07-21T14:28:04.232Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"resize_shapes": {
-				"avgFps": 22,
-				"minFps": 18,
-				"maxFps": 67,
-				"timestamp": "2025-07-21T14:28:23.924Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"canvas_panning": {
-				"avgFps": 38,
-				"minFps": 28,
-				"maxFps": 70,
-				"timestamp": "2025-07-21T14:28:13.302Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"canvas_zooming": {
-				"avgFps": 41,
-				"minFps": 19,
-				"maxFps": 67,
-				"timestamp": "2025-07-21T14:28:23.924Z",
-				"environment": {
-					"platform": "linux",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			}
-		},
-		"darwin-1280x720": {
-			"rotate_shapes": {
-				"avgFps": 60,
-				"minFps": 47,
-				"maxFps": 65,
-				"timestamp": "2025-11-26T22:53:39.487Z",
-				"environment": {
-					"platform": "darwin",
-					"viewport": "1280x720",
-					"browser": "chromium"
-				}
-			},
-			"drag_shapes": {
-				"avgFps": 60,
-				"minFps": 48,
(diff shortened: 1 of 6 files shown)


mimecuvalo (commented):
interesting, @kaneel can you check the performance on this change — this is some crucial code to keep performant

kaneel:
> interesting, @kaneel can you check the performance on this change — this is some crucial code to keep performant

Yeah I did not think about that but apparently, `.segment(str)` is **lazy + O(n)** and is all well optimised. 
The only slow thing is to construct the segmenter and it's done once at to of the scope.

Do we have a way to measure, some stress tests?

mimecuvalo:
> Do we have a way to measure, some stress tests?

not yet! can you add something in performance-perf?

kaneel:
> can you add something in performance-perf?

First little benchmark, it's not significant, let me claude that:

- The iteration step itself (Intl.Segmenter vs the old for...of character loop) is slower in isolation, but in absolute terms it's tiny — sub-1ms even for a very large (~1800 char) text shape, and microseconds for normal labels.
- That step was never the bottleneck. measureElementTextNodeSpans is dominated by range.getClientRects(), which runs once per glyph and forces a DOM layout. That cost dwarfs the segmentation pass and hasn't changed.
- For emoji/flag/accented text it can actually be faster, since one grapheme = one loop iteration (and one getClientRects call) instead of 2–7.
- The segmenter is created once at module scope, so there's no per-call construction overhead.

I can check how to claude some good benchmark

frolic (commented):
A couple of follow-ups on the perf discussion:

**A direct microbenchmark would close the loop better than the FPS scenario.** The `create_text_shapes` test is a good smoke test, but it can't really resolve what we're worried about here: the segmentation delta is sub-millisecond, which sits well below the noise floor of an end-to-end FPS metric. It also mostly exercises text shapes, whereas the changed code (`measureElementTextNodeSpans` → `measureTextSpans`) is only reached by frame titles (`FrameShapeUtil`, `frameHelpers`) — regular text shapes render rich text via the DOM and never hit the spans path. The new baseline is also darwin-only, so on CI's linux runners there may be nothing to compare against.

A tighter check would be a micro-benchmark of `measureTextSpans` (old vs new) on a few representative strings — a short ASCII label, a long paragraph, and an emoji/flag/accent-heavy title — reporting per-call time. That isolates the segmenter cost from the `getClientRects()` layout reads and would actually demonstrate the sub-ms claim. The expected result is neutral-to-faster, since iteration count drops from code points to graphemes, so emoji-heavy text makes fewer `getClientRects()` calls than before.

Also noticed the formatting changed on `fps-baselines.json` — can we either `yarn format` to update or ignore the file in `.prettierignore`?

Sent from my aiPhone

mimecuvalo (approved):
pre-approving — looking good in general.
you can followup with a PR if you'd like to improve the perf testing here — that'd be grand (and yeah the formatting is a minor nit)

PR #9268, opened 2026-06-17
fix(dotcom): restore and polish the manage workspace dialog

frolic:
In order to fix the empty "Manage workspace" dialog shown for the home workspace ("My workspace") that #9168 introduced — and to bring the dialog back in line with its pre-#9168 design — this PR restores and polishes the manage workspace dialog. Opening "Settings" for the home workspace returned `null` (a dimmed page with no content) while the sidebar still opened it. The home workspace is private (its group id equals the owner's user id): it can be renamed, but not shared, joined, deleted, or have its members managed.

Changes:

- **Home workspace** — shows a rename-only view (the name field plus a disabled invite control and the note "This is your private workspace. Create a new workspace to invite teammates.") instead of an empty dialog; the Members and Settings tabs are hidden. Resolves the Cursor Bugbot "Home workspace settings empty" flag on #9168.
- **Members tab** — dropped the "({count})" suffix; the rework added it, but the pre-#9168 label was just "Members".
- **Settings tab** — Regenerate / Leave / Delete are now left-aligned text-link buttons, and the negative-margin/padding alignment hacks are gone, so their labels line up with the toggle and the section headings. The shared `.menuControlRow` 4px nudge is disabled (it threw off that alignment; it's also used by the file share menu and the cookie dialog).
- **Invite buttons** — apply the `big` `TlaButton` variant (added on main in #9261) to all three invite buttons so they keep a consistent size across enabled/disabled states, and use the secondary-disabled treatment for the "invites off" state, matching the home workspace's private invite button.

Merged with `main` (#9261), which independently added the `big` variant and cleaned up the dialog CSS; this branch uses main's `big` rather than its own.

### Change type

- [x] `bugfix`

### Test plan

Covered by e2e regression tests in `ui.scenario.spec.ts`:

1. Home workspace settings render the private rename view (editable name, the private-workspace note, a disabled "Copy invite link", and no Members/Settings tabs) rather than an empty dialog.
2. The home workspace can be renamed from the dialog.
3. A shared workspace still shows the invite link and the Members/Settings tabs (existing coverage, updated for the countless "Members" label).

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fix the empty settings dialog shown when opening settings for your personal "My workspace"; it now shows the workspace name (editable) and a note that it's private.
- Polish the manage workspace dialog: a simpler "Members" label, link-style settings actions, and consistently-sized invite buttons.

### Code changes

| Apps            | +234 / -184 |
| Tests           | +56 / -1    |
| Automated files | +18 / -17   |


jsscclr (approved):
it's restored!

jsscclr (approved):
even better

Issue #9134, opened 2026-06-11
Dismissing a select menu inside a popout also dismisses the popout


frolic:
🗣️: clicking out of a radix select menu that lives within a menu/popout dismisses both the select menu AND the containing menu/popout

---

In the file share menu and the export menu, opening a select dropdown and then clicking outside it — but still inside the containing menu/popout, e.g. on the select's own label — closes both the select and the surrounding menu/popout at once. Pressing Escape does the same. Expected: that first action should close only the select and leave the menu/popout open, so it takes a second, separate dismissal to close the menu itself.

## Open questions

1. **Where are you seeing this?** Which specific menu and select — e.g. the workspace settings role picker, the file share menu's link-type select, or somewhere else?
   The share menu and the export menu.
2. **When you click outside the select but still inside the containing menu/popout, what should happen?** Close only the select and keep the menu open, with the menu closing only on a click fully outside everything?
   Confirmed — this is the bug. Clicking outside the select but inside the menu/popout (e.g. on the select label) currently closes both; it should close only the select and leave the menu/popout open.
3. **Does it also happen via the keyboard (pressing Escape), or only with a pointer click outside?**
   Both — it happens with Escape and with a pointer click outside.

Confidence: 88%, ready to get started.

frolic:
https://github.com/user-attachments/assets/b75001d2-96a4-49cd-ad6f-eae10bab8ef8

frolic:
might be related to https://github.com/radix-ui/primitives/issues/3461

PR #9269, opened 2026-06-17
fix(tldraw): make dialog and menu dismissal layer-aware via Radix

frolic:
Two nested-dismissable-layer bugs (#9264, #9134) where a popup opened inside another dismissable container dismissed the container too on the first outside click — both from hand-rolled dismiss logic that bypassed Radix's layer-aware "light dismiss". Fixing them turned into a broader cleanup of dotcom's selects and menus so they lean on Radix (layer-aware dismiss + portals) instead of carrying hand-rolled workarounds.

### SDK dialogs (#9264)

The dialog dismissed via an `onClick` on the overlay that closed on any backdrop click, ignoring Radix's layer stack — so a select opened inside a modal closed the whole modal on the first outside click. That handler dates to #2497 (Jan 2024), where Radix's native dismiss wasn't firing for reasons never found and flagged "revisit later"; on `radix-ui` 1.4.2 it works, so it's gone. The content now sits beside the overlay (Radix's standard sibling structure) with a positioner preserving the centring/scroll/padding, and dismissal is layer-aware (an open select/popover consumes the first outside press; a second closes the dialog). Escape still closes; drag-out (press inside, release outside) no longer dismisses, since Radix is pointer-down based. Stacked dialogs stay modal so a backdrop press / Escape closes one layer at a time. `preventBackgroundClose` is kept only on the legacy read-only modal.

### dotcom selects & menus (#9134 + cleanup)

- **`TlaMenuSelect`**: removed the capture-phase `document.body` pointerdown listener (the cause of #9134 — it closed the select synchronously and de-registered its Radix layer before the popover's own dismiss handler ran, taking the popover down too) and the `stopPropagation` guard. Made the open state **uncontrolled** (it was only controlled to feed that listener). Now **always portals** with Radix popper positioning, dropping the inline-vs-portal `usePortal` split and the hand-rolled CSS positioning override.
- **Share menu**: removed the `<div onPointerDown={preventDefault}>` wrapper — vestigial since the menu moved from a Radix DropdownMenu (#5459's context) to a Popover.
- **Sidebar**: portaled the file context menu, the workspace switcher, and the user-settings menu into the editor container via `useContainer()`, consistent with the SDK's own dialogs/menus.

### Sidebar menu dismissal + positioning

- The workspace switcher and the sidebar "…" / user-settings menus stay **non-modal** (Radix's default), so they cross-dismiss natively — opening one closes the other. The switcher → file-menu **flash** (opening a second menu used to dismiss it, because the modal switcher restored focus to its trigger and that focus shift tore down the just-opened non-modal menu) is fixed with a guarded `onCloseAutoFocus`: it skips restoring focus to the trigger *only when another menu is already opening*, while a plain Escape / outside-click close still restores focus for keyboard users.
- **Note vs #9330:** after coordinating with Steve, this PR deliberately keeps the **non-modal** switcher rather than #9330's `modal` + click-swallow overlay. When this lands after #9330 it replaces that overlay with the non-modal + `onCloseAutoFocus` approach. (No-click-through is intentionally not pursued here — for these controlled menus it needs the overlay or an uncontrolled refactor, both out of scope.)
- Menu offsets are centralised in a shared `TLA_MENU_POSITION` constant (`sideOffset` / `alignOffset` / `collisionPadding`), spread onto the select and share popover. The sidebar dropdowns override `alignOffset` so they hang ~8px over the sidebar's right edge.
- Dropdown **width**: the portaled select and switcher menus are sized off the popper wrapper's `--radix-popper-anchor-width` with `!important` (Radix sets an inline `min-width: max-content` on the wrapper, so the size must be forced there), because `--tla-sidebar-width` is declared on the sidebar layout and doesn't reach the portaled menu.
- `tlmenus.clearOpenMenus()` is called when the mobile sidebar closes, so portaled menus don't linger behind the CSS-hidden sidebar.

### Test plan

Covered by automated tests:

- `packages/tldraw/src/test/ui/Dialogs.test.tsx` (jsdom) and `apps/examples/e2e/tests/test-dialogs.spec.ts` (browser) — dialog dismiss + stacked-escape one layer at a time
- `apps/dotcom/client/e2e/tests/smoke/sharing.spec.ts` — share menu select dismiss
- `apps/dotcom/client/e2e/tests/smoke/workspaces.spec.ts` — switcher (incl. the focus-steal regression and clicking a file with it open) and the file context menu (move / delete / rename)

Also reproduced and verified by hand against a local dotcom + the toasts/dialogs example.

- [x] Unit tests
- [x] End to end tests

### API changes

- Added optional `collisionPadding?: number` to `TldrawUiPopoverContent` (`TLUiPopoverContentProps`) — forwarded to the Radix popover content; defaults to Radix's `0`.
- Added optional `collisionPadding?: number` to `TldrawUiDropdownMenuContent` (`TLUiDropdownMenuContentProps`) — defaults to `4` (the value it previously hardcoded), so no change for existing consumers.

### Change type

- [x] `bugfix`

### Release notes

- Fix modal dialogs being dismissed when you closed a select or popover opened inside them; the inner popup now closes on its own first.

### Code changes

| Core code       | +43 / -40   |
| Tests           | +444 / -71  |
| Automated files | +5 / -2     |
| Documentation   | +139 / -5   |
| Apps            | +195 / -234 |

### Follow-ups (not in this PR)

- Converge the select and dropdown/menu **styling** onto a single shared base (the switcher already builds on `tlui-menu`; `TlaMenuSelect` carries its own duplicate panel/row CSS). Keep them as separate primitives (Select vs DropdownMenu), share the look.
- Consider whether dotcom's `TlaMenuSelect` needs to exist at all, or can lean on the SDK's existing `TldrawUiSelect` (same portaled-select-into-container pattern).

### DIFF
--- a/apps/dotcom/client/e2e/tests/smoke/sharing.spec.ts
+++ b/apps/dotcom/client/e2e/tests/smoke/sharing.spec.ts
@@ -42,6 +42,32 @@ test.describe('signed in user on own file', () => {
 		expect(await shareMenu.publishTabPage.isVisible()).toBe(true)
 	})
 
+	test('a select inside the menu closes without closing the menu', async ({ page, shareMenu }) => {
+		await shareMenu.open()
+		await shareMenu.exportTabButton.click()
+		await expect(shareMenu.exportTabPage).toBeVisible()
+
+		const themeSelect = page.locator('#tla-export-theme-select')
+
+		// Selecting an option closes the select but leaves the share menu open.
+		await themeSelect.click()
+		await expect(page.getByRole('listbox')).toBeVisible()
+		await page.getByRole('option').last().click()
+		await expect(page.getByRole('listbox')).toHaveCount(0)
+		await expect(shareMenu.exportTabPage).toBeVisible()
+
+		// Re-clicking the open select's trigger also closes only the select. The trigger
+		// is non-interactive while the select is open (Radix disables outside pointer
+		// events), so click its position directly rather than via the locator.
+		await themeSelect.click()
+		await expect(page.getByRole('listbox')).toBeVisible()
+		const box = await themeSelect.boundingBox()
+		if (!box) throw new Error('theme select not found')
+		await page.mouse.click(box.x + box.width / 2, box.y + box.height / 2)
+		await expect(page.getByRole('listbox')).toHaveCount(0)
+		await expect(shareMenu.exportTabPage).toBeVisible()
+	})
+
 	test('can unshare and reshare', async ({ page, browser, shareMenu }) => {
 		// Copy the link to the current file
 		await shareMenu.open()
--- a/apps/dotcom/client/e2e/tests/smoke/workspaces.spec.ts
+++ b/apps/dotcom/client/e2e/tests/smoke/workspaces.spec.ts
@@ -63,7 +63,10 @@ test.describe('workspaces', () => {
 			await expect(homeItem).toBeVisible()
 		})
 
-		test('dismissing the switcher does not click a file underneath', async ({ page, sidebar }) => {
+		test('clicking a file while the switcher is open closes it and opens that file', async ({
+			page,
+			sidebar,
+		}) => {
 			const file1 = getRandomName()
 			const file2 = getRandomName()
 			const file3 = getRandomName()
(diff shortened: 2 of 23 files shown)


mimecuvalo:
not sure how fast the base ui stuff is landing but might be worth holding for now

frolic:
> not sure how fast the base ui stuff is landing but might be worth holding for now

Based on my chats with Claude, it sounds like this issue is self-inflicted rather than something in Radix itself, and that a refactor to Base UI is a little more hairy. I'm still letting it cook so stay tuned.

steveruizok (approved):

PR #9270, opened 2026-06-17
fix(dotcom): keep the home workspace file list correct

jsscclr:
## What

Fixes a cluster of "My files" bugs in the workspaces model that all trace to one root cause.

A shared file you open as a guest is mirrored into your home group — this is the intended **guest file** feature (the file shows in "My files" with a guest badge so you can find it again). But if that file is owned by a workspace you're a **member of**, it would show in "My files" yet open into that workspace — because the active workspace is derived from the file's `owningGroupId`. So clicking it (or switching to "My files", which opens the top file) bounced you into the workspace, and you could never get back to your home.

This happens when:

- you open a workspace's shared file as a guest, then later **join** that workspace, or
- a workspace's **welcome file** is mirrored into home during the create-workspace race.

It explains the reported symptoms: shared files and a "Welcome to your workspace" file showing in "My files", and clicking them teleporting the user into another workspace with no way back.

## Preview

https://pr-9270-preview-deploy.tldraw.com/

## Fixes

1. **Teleport / "can't get back to My files" — read-side filter (`getWorkspaceFilesSorted`).** For the home workspace, the sidebar now lists only files it owns, legacy files, and genuine guest files (owned by workspaces you're **not** a member of). A file owned by a workspace you **are** a member of is no longer listed in home, so it can't bounce you out. Guest-file mirroring (`onEnterFile`) is unchanged — guest files still appear as before. Filtering on read also fixes already-affected users on deploy, before any data cleanup.

2. **File-count limit consistency (`canCreateNewFile`).** It now counts only files the workspace actually owns (`owningGroupId === workspaceId`), not guest files or hidden/mislinked rows — matching the server's `assertNotMaxFiles`. Prevents "max files reached" firing on files the user can't see or remove.

3. **Empty home stays switchable (`createFile`).** Authorizes the home workspace via `getRole`/`can('addFiles')` instead of a raw `group_user` lookup. The home workspace (`id === userId`) is implicitly owned and may have no `group_user` row (e.g. before it syncs), so creating a file there could be rejected — leaving an empty home impossible to switch back to. This matches how every other workspace op already authorizes.

## How to verify

On the preview (https://pr-9270-preview-deploy.tldraw.com/), with two accounts:

1. **Account A:** create a workspace, add a file inside it, then copy that **file's share link** and the **workspace invite link**.
2. **Account B (incognito):** open the **file share link** → the file appears in "My files" with a guest badge.
3. **Account B:** open the **workspace invite link** and accept it (you're now a member of the workspace).
4. **Account B:** click **"My files"** → you stay in "My files".

Before the fix, step 4 bounced you into the workspace (you couldn't get back to "My files").

## Testing

- New scenario test — `opening a guest file does not trap the member outside their home after they join its workspace` (`sharing-live.scenario.spec.ts`). It fails on the unfixed code (the member is teleported into the workspace) and passes here.
- Unit tests in `mutators.test.ts` cover `createFile` home authorization and `onEnterFile` guest / member / legacy mirroring.
- `yarn typecheck`, lint, and the full dotcom e2e suite pass.

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -434,8 +434,25 @@ export class TldrawApp {
 		const membership = this.getWorkspaceMembership(workspaceId)
 		if (!membership) return []
 
-		const pinned = membership.groupFiles.filter((f) => f.index !== null)
-		const unpinned = membership.groupFiles.filter((f) => f.index === null)
+		// Decide which of the membership's group_file rows actually belong in this list.
+		// A file owned by this workspace always belongs. A non-home workspace lists only the
+		// files it owns. The home workspace additionally lists legacy files (no owning group)
+		// and "guest files" — shared files the user opened that are owned by a workspace they
+		// are NOT a member of. It must NOT list a file owned by a workspace the user IS a
+		// member of: that's a mislinked row (a guest file whose workspace was later joined, or
+		// a workspace's welcome file mirrored during the create-workspace race), and opening it
+		// from home would bounce the user into that workspace. Guarding here keeps such rows out
+		// of the list even before they're cleaned up.
+		const homeWorkspaceId = this.getHomeWorkspaceId()
+		const groupFiles = membership.groupFiles.filter((f) => {
+			const owningGroupId = f.file.owningGroupId
+			if (owningGroupId === workspaceId) return true
+			if (workspaceId !== homeWorkspaceId) return false
+			return owningGroupId == null || !this.getWorkspaceMembership(owningGroupId)
+		})
+
+		const pinned = groupFiles.filter((f) => f.index !== null)
+		const unpinned = groupFiles.filter((f) => f.index === null)
 
 		const lastOrdering = this.lastWorkspaceFileOrderings.get(workspaceId)
 		const retainedOrdering =
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -629,8 +629,9 @@ export class TldrawApp {
 		if (this.isWorkspacesMigrated()) {
 			// Count only files the workspace actually owns — not guest files (shared files the
 			// user opened) or mislinked rows that getWorkspaceFilesSorted hides. Counting those
-			// would let the limit fire on files the user can neither see nor remove, and would
-			// disagree with the server's own limit check (assertNotMaxFiles counts owningGroupId).
+			// would let the limit fire on files the user can neither see nor remove. For migrated
+			// users createFile runs no server-side file-limit check, so this is the only gate;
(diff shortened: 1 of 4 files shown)


apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -434,8 +434,25 @@ export class TldrawApp {
...
-		const unpinned = membership.groupFiles.filter((f) => f.index === null)
+		// Decide which of the membership's group_file rows actually belong in this list.
+		// A file owned by this workspace always belongs. A non-home workspace lists only the
+		// files it owns. The home workspace additionally lists legacy files (no owning group)
+		// and "guest files" — shared files the user opened that are owned by a workspace they
+		// are NOT a member of. It must NOT list a file owned by a workspace the user IS a
+		// member of: that's a mislinked row (a guest file whose workspace was later joined, or
+		// a workspace's welcome file mirrored during the create-workspace race), and opening it
+		// from home would bounce the user into that workspace. Guarding here keeps such rows out
+		// of the list even before they're cleaned up.
```
frolic, line 445: so to rephrase
- workspaces only list their own files
- home excludes workspace files where I am already a member

is this accurate?

mimecuvalo (approved):
👍  only tweak is a comment change but good to go:
"While verifying, I found the new comment in TldrawApp.ts:633 (canCreateNewFile) overstates the server side:

▎ "would disagree with the server's own limit check (assertNotMaxFiles counts owningGroupId)"

Two problems:
1. assertNotMaxFiles is only called in the non-migrated branch of createFile (line 355). The migrated branch never calls it — so for migrated users (the exact path that comment lives in) the server runs no file-limit check at all; canCreateNewFile on the client is the only gate.
2. If it were called, its server query counts ownerId = $1 OR owningGroupId = $1 — not just owningGroupId."

jsscclr:
The mutator code covers the case where a user could theoretically not have a group_user row at all, because before I figured out the actual issue, I could replicate it with manually deleting the group_user row. But theoretically that path is impossible since the groups rollout. It would be code that would be deleted once we clean up all the forked behaviour. The actual fix is the client side part - the other is just theoretical 'hardening'. Can just drop it if it's confusing / feels riskier?

apps/dotcom/client/src/tla/app/TldrawApp.ts
jsscclr: Correct!

frolic: should we simplify the comment? it was bit of a brain teaser
or maybe the comment is for future AI not future us

frolic (approved):
tested this in preview deploy

- Brave: made a workspace
- Brave: added a file
- Chrome: opened file link, saw it added to my files
- Chrome: joined workspace, opening file now pops me into workspace, can no longer see file in my files

PR #9273, opened 2026-06-17
chore(dotcom): allow parallel dev stacks per worktree

### LINKED ISSUE #9061: Dotcom dev stack can't run in parallel worktrees due to fixed and default wrangler ports
🧑: from an infra + dx perspective, can we file an issue about this? potentially exposing an easy way to swap out ports (potentially automatically) it makes tackling multiple features in different worktrees a bit difficult - maybe we need a skill, etc

---

## Summary

Running the dotcom dev stack (`yarn dev-app`) from a second git worktree while another worktree's stack is already running fails — and fails **silently**. This makes working on multiple features across worktrees painful.

## What happens

- The dotcom workers use **fixed** wrangler inspector ports: `image-resize-worker` 9339, `asset-upload-worker` 9449, `tldrawusercontent-worker` 9450 (see each `dev.sh` / the `--inspector-port` args). A second stack collides on these immediately (`✘ [ERROR] Address already in use (127.0.0.1:9339)`).
- The **sync-worker** is the worst case: its `apps/dotcom/sync-worker/dev.sh` does **not** pass `--inspector-port`, so it uses wrangler's default **9229**. When 9229 is already held by another worktree's stack, the worker **never binds its serving port (8787) and logs no error** — the output just stops after `worker is roughly NNNNkb (minified)`. There is no signal that anything is wrong; you only notice the app doesn't work.
- The serving ports (client 3000, sync-worker 8787, zero-cache 4848) are also fixed, so two stacks can't coexist regardless.
- The **postgres container is shared and torn down on exit**: `zero-cache`'s `dev` runs `docker-up` attached, so stopping one worktree's stack stops the `6543` postgres for everyone.
- Separately, the gitignored secrets (`apps/dotcom/client/.env.local`, `apps/dotcom/sync-worker/.dev.vars`) are **not shared across worktrees**, so every new worktree needs them hand-copied before `dev-app` works at all.

## Asks

1. **Fail loudly, not silently.** `internal/scripts/workers/dev.ts` already has `ensurePortFreeOrExit` for the serving port — extend it to probe the inspector port too and exit with a clear message (e.g. "inspector port 9229 in use — is another dev stack running?"). This alone turns a long diagnosis into one line.
2. **Make ports swappable / auto-allocated.** Either pass `--inspector-port 0` (let wrangler pick a free port) for all dotcom workers, or honor a single `WORKTREE_PORT_OFFSET` (or derive one by hashing the worktree path) that shifts the serving + inspector + zero ports (3000/8787/4848/9229/9339/9449/9450) together. At minimum, give the sync-worker an explicit, configurable inspector port like the others.
3. **Scope the docker project per worktree** (or document that the DB is intentionally shared and stop tearing it down on a single stack's exit) so one worktree's stack doesn't kill `6543` for another.
4. **Reduce the secrets papercut** — share/symlink `client/.env.local` and `sync-worker/.dev.vars` into new worktrees automatically, or document the copy step.

## Possible additional DX win

A small skill / script (e.g. `run-dotcom-worktree`) that encodes the workaround: detect a running stack, free or offset ports, copy the gitignored secrets into the worktree, bring the stack up, and verify 3000/8787/4848. This is the near-term ergonomics stopgap while the dev-script fixes land.

## Repro

1. `yarn dev-app` in worktree A (works).
2. `git worktree add` worktree B; copy the two secret files in.
3. `yarn dev-app` in worktree B → image/asset/usercontent workers error on their fixed inspector ports; sync-worker silently never binds 8787; the app is broken with no clear error.

frolic:
In order to work on multiple dotcom features at once (e.g. several Claude sessions across git worktrees), this PR lets each worktree run its own `yarn dev-app` stack side by side. Closes #9061.

Today the dotcom dev stack is a singleton: every service binds a fixed port, so a second `yarn dev-app` collides — and the sync worker fails *silently* (it never binds its serving port, with no error in the log). This PR gives each dev instance its own contiguous block of ports plus its own Docker, Zero, and Wrangler state, so stacks don't fight over ports, database data, or service bindings.

### How it works

- **Ports**: each instance gets a 100-port block — a service's port is `3000 + instance * 100 + slot`, with a fixed slot per service (see `DOTCOM_DEV_PORT_SLOTS`). So instance 0 = 3000–3014, instance 1 = 3100–3114, etc. The client sits at slot 0, so **instance 0's client stays on `localhost:3000`** and existing e2e/Clerk/bookmarks are unaffected. Slots 3–4 are left free for Zero's change-streamer/litestream, which derive from `ZERO_PORT`.
- **Instance assignment**: stable per worktree. The first time a worktree runs `yarn dev-app` it claims the lowest free instance and records it in `~/.tldraw/dotcom-dev-instances.json` (a small locked registry), so its ports and database stay put across restarts. `DOTCOM_DEV_INSTANCE=<n>` overrides.
- **Isolated state**: per-instance Docker compose project + Postgres volume, Zero replica file, Wrangler `--persist-to` dir, and — importantly — a per-instance `WRANGLER_REGISTRY_PATH` so a second stack's workers don't re-register the shared worker names and cross-wire service bindings between instances.
- **Fail loud**: the shared worker runner now probes the inspector port too, turning the silent sync-worker hang into a clear "port in use" error. Non-dotcom workers (examples, bemo) are untouched.

### Change type

- [x] `other`

### Test plan

1. `yarn dev-app` in worktree A → instance 0, client at http://localhost:3000.
2. `yarn dev-app` in worktree B (with secrets copied in) → instance 1, client at http://localhost:3100. Both serve at once.
3. `yarn dev-app:doctor` in each prints its instance + ports; `yarn dev-app:clean` cleans only that instance.
4. `yarn e2e-dotcom` still targets `localhost:3000` (pinned to instance 0).

- [x] Unit tests
- [ ] End to end tests

> Note: I couldn't run a live dual-stack here (needs Docker + Clerk keys + a second worktree), so the end-to-end "two stacks serving simultaneously" still wants a local smoke test.

### Code changes

| Tests          | +86 / -7   |
| Apps           | +468 / -85 |
| Config/tooling  | +38 / -16  |

### DIFF
--- a/apps/dotcom/asset-upload-worker/package.json
+++ b/apps/dotcom/asset-upload-worker/package.json
@@ -9,7 +9,7 @@
 	},
 	"main": "src/index.ts",
 	"scripts": {
-		"dev": "yarn run -T tsx ../../../internal/scripts/workers/dev.ts --persist-to tmp-assets --inspector-port 9449",
+		"dev": "yarn run -T tsx ../zero-cache/run-dotcom-worker.ts assetUploadWorker --persist-to tmp-assets",
 		"test-ci": "yarn run -T vitest run --passWithNoTests",
 		"test": "yarn run -T vitest --passWithNoTests",
 		"test-coverage": "yarn run -T vitest run --coverage --passWithNoTests",
--- a/apps/dotcom/client/e2e/fixtures/Database.ts
+++ b/apps/dotcom/client/e2e/fixtures/Database.ts
@@ -3,13 +3,16 @@ import { Page } from '@playwright/test'
 import { DB, userHasFlag } from '@tldraw/dotcom-shared'
 import { Kysely, PostgresDialect, sql } from 'kysely'
 import pg from 'pg'
+import { getDotcomDevPorts } from '../../../zero-cache/dev-env'
 import { OTHER_USERS, USERS } from '../consts'
 import { getStorageStateFileName } from './helpers'
 
+// e2e pins the dev stack to instance 0 (see playwright.config.ts), so its pgbouncer is on instance
+// 0's port.
 const db = new Kysely<DB>({
 	dialect: new PostgresDialect({
 		pool: new pg.Pool({
-			connectionString: 'postgresql://user:password@127.0.0.1:6432/postgres',
+			connectionString: `postgresql://user:password@127.0.0.1:${getDotcomDevPorts(0).pgbouncer}/postgres`,
 			application_name: 'migrate',
 			idleTimeoutMillis: 10_000,
 			max: 10,
--- a/apps/dotcom/client/e2e/fixtures/Database.ts
+++ b/apps/dotcom/client/e2e/fixtures/Database.ts
@@ -3,16 +3,17 @@ import { Page } from '@playwright/test'
 import { DB, userHasFlag } from '@tldraw/dotcom-shared'
 import { Kysely, PostgresDialect, sql } from 'kysely'
 import pg from 'pg'
-import { getDotcomDevPorts } from '../../../zero-cache/dev-env'
 import { OTHER_USERS, USERS } from '../consts'
 import { getStorageStateFileName } from './helpers'
 
-// e2e pins the dev stack to instance 0 (see playwright.config.ts), so its pgbouncer is on instance
-// 0's port.
+// e2e pins the dev stack to instance 0 (see playwright.config.ts), so pgbouncer is on instance 0's
+// port: DOTCOM_DEV_PORT_BASE (3000) + pgbouncer slot (6) = 3006. Keep in sync with
+// zero-cache/dev-env.ts (not imported here: that module uses import.meta and breaks Playwright's
+// CommonJS test transpile).
 const db = new Kysely<DB>({
 	dialect: new PostgresDialect({
 		pool: new pg.Pool({
-			connectionString: `postgresql://user:password@127.0.0.1:${getDotcomDevPorts(0).pgbouncer}/postgres`,
+			connectionString: 'postgresql://user:password@127.0.0.1:3006/postgres',
 			application_name: 'migrate',
 			idleTimeoutMillis: 10_000,
 			max: 10,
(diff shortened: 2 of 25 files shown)


mimecuvalo:
thx for digging into this 👍

frolic:
### Considered: containerizing the whole stack

We explored putting the entire per-instance stack in Docker — every service (client, the four workers, zero-cache, postgres) running inside containers, each instance its own compose project, and only the ports we actually hit exposed/remapped to the host from the container. The appeal is that container networking handles the wiring for free: services reach each other by name on fixed internal ports, so most of this PR's port-offsetting and URL rewriting would disappear and you'd only publish the client port.

We decided against it here because, for local dev, the costs outweigh that:

- **macOS perf** — Docker runs in a Linux VM, so rebuilding Vite + four `wrangler dev` workers + zero-cache through a bind mount is much slower than native, and file-watching/HMR over the mount is flaky. That's a permanent iteration-speed tax.
- **node_modules can't be shared** — several deps ship platform-specific native binaries (esbuild, workerd/wrangler, `@rocicorp/zero`), so the host's macOS `node_modules` can't be mounted into a Linux container; you'd need a separate in-container install per instance.
- **`wrangler dev` in Docker is finicky** — workerd, inspector/debugger ports, and source maps all get more awkward across the container boundary.
- **Scope** — it changes how everyone runs dotcom locally (Dockerfiles, mount strategy, in-container installs), and you'd still need per-instance scoping (distinct compose projects + published host ports), so the instance numbering doesn't fully go away.

So this PR keeps the services host-native (matching how the repo already runs) and isolates by port block + per-instance state, with Postgres/pgbouncer staying in Docker where we already do the inner→outer port remap. Full containerization is still a reasonable future initiative for reproducible/onboarding-friendly environments — it's just better pursued on its own merits than as the path to parallel stacks.

Sent from my aiPhone

frolic:
closing in favor of https://github.com/tldraw/tldraw/pull/9296 OR https://github.com/tldraw/tldraw/pull/9301 + https://github.com/tldraw/tldraw/pull/9302

PR #9281, opened 2026-06-17
fix(arrows): re-snap binding when bound shape's geo type changes

### LINKED ISSUE #9030: Arrow binding floats outside geo shape after changing its geo type
🧑: when an arrow is bound to a geo shape, and the geotype is changed to a type that would cause the arrow binding to no longer overlap the shape, the arrow remains bound to the shape but now does not visibly overlap it. imo we should move the binding to the closest point on the new geoshape, maybe with an inset

---

## Description

When an arrow is bound to a geo shape with a **precise** binding, the binding stores a fixed `normalizedAnchor` (a 0–1 coordinate within the shape's bounding box). Changing the geo type (e.g. rectangle → ellipse, triangle, star, diamond) keeps the bounding box but replaces the geometry inside it. The stored anchor is never revalidated, so it can land in a region of the bounds that the new geometry no longer fills. The arrow stays bound but its endpoint floats in empty space outside the visible shape.

## Steps to reproduce

1. Draw a rectangle.
2. Draw an arrow and bind one end precisely near a corner of the rectangle (drag the endpoint onto the corner so it sticks).
3. Select the geo shape and change its type to ellipse, triangle, diamond, or star.
4. Observe the arrow endpoint.

## Expected behavior

The arrow endpoint stays visibly attached to the new geometry — ideally re-anchored to the closest point on the new shape's outline, optionally with a small inset.

## Actual behavior

The arrow remains bound but the endpoint no longer overlaps the visible shape — it floats in the area of the old bounding box that the new geometry doesn't cover (e.g. the corners of an ellipse or the outer points of a star/triangle).

## Suggested fix

When a geo shape's `geo` prop changes, re-point any precise arrow bindings to the closest point on the new geometry (with an optional inset), rather than leaving the stale normalized anchor in place.

max-drake:
👩: worried about perf here, might not be worth it, or need more aggressive checks on what props are getting changed?

---

Closes #9030.

In order to fix arrows that float off geo shapes when their type changes, this PR re-snaps a precise arrow binding's anchor to the nearest point on the bound shape's new geometry whenever the anchor ends up outside it.

A precise arrow binding stores its anchor as a normalized point (0–1) within the bound shape's bounding box. Changing a geo shape's type (e.g. rectangle → ellipse, triangle, diamond, star) keeps the bounding box but replaces the geometry inside it, so the stored anchor can land in a region the new outline no longer fills — the arrow stays bound but its endpoint floats in empty space. `ArrowBindingUtil.onAfterChangeToShape` now hit-tests the anchor against the new geometry and, if it no longer lands on or inside the shape, moves it to the closest point on the new outline.

The re-snap only runs for precise bindings (imprecise ones target the center and intersect the edge, so they never float), only while the editor is idle (not mid-drag/resize), and never during history replay (the diff already holds the correct anchor).

### Change type

- [x] `bugfix`

### Test plan

1. Draw a rectangle.
2. Draw an arrow and bind one end precisely near a corner of the rectangle (drag the endpoint onto the corner so it sticks).
3. Select the geo shape and change its type to ellipse, triangle, diamond, or star.
4. The arrow endpoint should stay attached to the new shape's outline rather than floating in empty space.

- [x] Unit tests

### Release notes

- Fix arrow bindings floating outside a geo shape after changing its geo type. Precise bindings now re-snap to the nearest point on the new shape.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +55 / -0   |
| Tests     | +82 / -0   |


kaneel (approved):
Testing in preview and worked much better than the previous behaviour so ship it!

packages/tldraw/src/lib/bindings/arrow/ArrowBindingUtil.ts
```diff
@@ -82,6 +83,20 @@ export class ArrowBindingUtil extends BindingUtil<TLArrowBinding> {
...
+		// only do it while idle (not mid-interaction), and never during history replay (the diff
+		// already holds the correct anchor).
+		if (
+			binding.props.isPrecise &&
+			this.editor.isIn('select.idle') &&
+			!this.editor.isReplayingHistory()
+		) {
+			updateBindingAnchorIfOutsideShape(this.editor, binding, shapeAfter)
+		}
+
```
steveruizok: I might make this a side effect when the geo shape's geo property changes? Doesn't seem like something we would want to check on every change.

max-drake: do you mean move this to `GeoShapeUtil.tsx`'s `onBeforeUpdate()` method? or just add a check for the geo type changing here?
edit: regardless, i think it makes more sense to keep this logic within the arrowbindingutil and so the binding can update itself if needed, instead of geo shapes updating the bindings, which I feel could get hairy

mimecuvalo: the logic in `onAfterChangeToShape` has quick short-circuits (geo-type change check and isPrecise check), this should be fine w.r.t. perf

mimecuvalo (approved):
👍

PR #7117, opened 2025-11-13
fix(dotcom): handle redirect properly for Google OAuth

mimecuvalo:
@MitjaBezensek noticed that for Google sign-in we weren't redirecting properly. We have to go closer to the metal to make that work, took a while to figure out this magic incantation ✨ 🪄 

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [x] `other`

### Test plan

1. Open the sign-in dialog
2. Click the Google sign-in button
3. Verify that you are redirected to the Google OAuth page and back to the app correctly

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed an issue where Google sign-in would not redirect correctly.

---

> [!NOTE]
> Replaces Clerk Elements Google sign-in with a custom button that initiates Clerk OAuth and redirects via externalVerificationRedirectURL.
> 
> - **Auth (TlaSignInDialog.tsx)**:
>   - **Google OAuth**:
>     - Add `handleGoogleSignIn` to initiate `signIn.create({ strategy: 'oauth_google', redirectUrl })` and redirect using `firstFactorVerification.externalVerificationRedirectURL`.
>     - Replace Clerk Elements UI (`SignIn.Root/Step`, `Clerk.Connection`) with a custom `TlaCtaButton` and static Google icon.
>     - Remove unused Clerk Elements imports.
>   - **Email sign-in**: Existing email flow retained; no functional changes shown.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit becd5df2044a897da5ce60b4c297d8619ad0620c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #9286, opened 2026-06-17
fix(dotcom): reset Google sign-in after browser back from OAuth

### LINKED ISSUE #9103: Sign in with Google button stops responding after returning from Google OAuth via browser back
🗣️: just noticed an issue with group share links but might be a general auth issue

to reproduce (I did this in a private session, so logged out of everything):

* get an invite link
* click "sign in with google" and get redirected to google oauth page
* click back
* click "sign in with google" again but notice the button does ~nothing

---

After starting Google sign-in from a group share/invite link while logged out, hitting the browser back button to return from Google's OAuth page, then clicking "Sign in with Google" again, the button stops responding — the second click does nothing. Expected behavior is that the sign-in flow restarts cleanly so the user can still sign in. It was first noticed on group share invite links, but it may be a general problem with the logged-out sign-in flow rather than something specific to group links.

Visual context: the sign-in modal in question is the standard tldraw.com dialog ("tldraw is a free online whiteboard…") with a "Sign in with Google" button, an email field, and a "Continue with email" button.

## Open questions

1. **Does this affect the general sign-in flow, or only group share / invite links?** Can you reproduce the same dead-button behavior on the normal tldraw.com sign-in modal (not via an invite link)?
   _Awaiting answer._
2. **Does only the Google button break, or does "Continue with email" also stop working after going back?** If email still works, that narrows it to the OAuth path.
   _Awaiting answer._
3. **What browser and OS, and does it reproduce outside a private/incognito session?** You mentioned a private session — does a normal logged-out session behave the same way?
   _Awaiting answer._
4. **Anything in the browser console or network tab when the dead click happens?** If you can grab an error or a screenshot of the console at that moment, it would help.
   _Awaiting answer._

Confidence: 70%, ready to get started; a couple of scope questions open.

mimecuvalo:
In order to fix the dead "Sign in with Google" button reported in #9103, this PR resets Clerk's sign-in state machine when the page is restored from the browser's back/forward cache (bfcache).

When a logged-out user clicks "Sign in with Google", we redirect to Google's OAuth page (a full top-level navigation). Pressing the browser back button to return restores the tldraw.com page from the bfcache, which restores the Clerk Elements (`SignIn.Root`) state machine frozen mid-redirect. In that frozen state the Google button no longer responds to clicks. This was first noticed on group share / invite links but affects the logged-out sign-in flow generally.

The fix listens for the `pageshow` event and, when `event.persisted` is true (the page was restored from the bfcache), bumps a `key` on `SignIn.Root` so it remounts with a fresh state machine. The "Sign in with Google" button then works again. The email flow is unaffected since it uses local React state. This avoids reintroducing the manual OAuth approach from #7117, which was reverted in #7189 because it broke new signups.

Closes #9103

### Change type

- [x] `bugfix`

### Test plan

1. In a logged-out/private session, open a group share / invite link (or the normal sign-in modal).
2. Click "Sign in with Google" and get redirected to Google's OAuth page.
3. Press the browser back button to return to tldraw.com.
4. Click "Sign in with Google" again — the button should restart the flow and redirect to Google instead of doing nothing.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fix the "Sign in with Google" button becoming unresponsive after returning to tldraw.com via the browser back button from Google's OAuth page.


apps/dotcom/client/src/tla/components/dialogs/TlaSignInDialog.tsx
```diff
@@ -111,6 +107,22 @@ function TlaEnterEmailStep({
...
+	const [signInResetKey, setSignInResetKey] = useState(0)
+	useEffect(() => {
+		function handlePageShow(e: PageTransitionEvent) {
+			if (e.persisted) {
+				setSignInResetKey((k) => k + 1)
+			}
+		}
+		window.addEventListener('pageshow', handlePageShow)
+		return () => window.removeEventListener('pageshow', handlePageShow)
+	}, [])
```
frolic: feels workaround-y but I guess its fine

frolic (approved):

mimecuvalo:
this isn't working as intended — sry for the noise, must go deeper

PR #9288, opened 2026-06-17
fix(tldraw): render note attribution as HTML in SVG export

### LINKED ISSUE #8941: Sticky note attribution text degrades when copied and pasted as PNG
## Description

When you copy a sticky note (note shape) as PNG and paste the resulting PNG back onto the canvas, the author **attribution text** on the note degrades in quality — it renders slightly larger and blurrier than on the original note. The main note text looks fine; the issue is specific to the small attribution name shown on the note.

## Steps to reproduce

1. Open a context where notes show author attribution (e.g. a collaborative room on tldraw.com), and create/edit a sticky note so its author attribution name is displayed.
2. Select the note and copy it as PNG (Edit → Copy as → PNG, or right-click → Copy as → PNG).
3. Paste the PNG back onto the canvas.
4. Compare the attribution text in the pasted PNG against the original note.

## Expected

The attribution text in the exported PNG matches the original note — same size and crispness as the rest of the exported note.

## Actual

The attribution text in the PNG is slightly bigger and blurrier than on the original note.

## Notes

In `NoteShapeUtil.toSvg` (`packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx`), the main note text is exported via `RichTextSVG`, which renders inside an SVG `<foreignObject>` (real HTML/CSS), so it matches the on-canvas appearance closely. The author attribution, however, is exported as a native SVG `<text>` element, which the browser lays out and rasterizes with different font metrics/antialiasing than HTML. That difference is the likely reason only the attribution looks bigger and blurrier in the PNG.

mimecuvalo:
In order to fix sticky note attribution text rendering larger and blurrier in PNG/SVG exports than on the live canvas, this PR renders the attribution as HTML inside a `<foreignObject>` instead of as a native SVG `<text>` element. Closes #8941.

The note body text was already exported through `RichTextSVG`, which wraps HTML in a `<foreignObject>` so the rasterized output uses the exact same layout and font metrics as the live DOM. The attribution was the only element exported as a native SVG `<text>`, which the browser's SVG text engine lays out and antialiases with different metrics — that mismatch was the degradation. Rendering it through a `foreignObject` with the existing `tl-note__attribution` class makes the export match the DOM.

Because the `foreignObject` now applies the real `.tl-note__attribution` CSS (`max-width: 60%; overflow: hidden; text-overflow: ellipsis`), ellipsis truncation happens natively and identically to the canvas, so the manual `truncateAttributionForSvg` text-measurement helper is no longer needed and has been removed.

### Change type

- [x] `bugfix`

### Test plan

1. Open a sticky note and type some text so the attribution (first name) appears in the bottom-right.
2. Copy the note as PNG (or export to PNG/SVG) and compare the pasted result against the original.
3. Confirm the attribution text is crisp and the same size/weight as on canvas, at both `scale = 1` and a resized note (`scale ≠ 1`).
4. Use a long attribution name and confirm it truncates with an ellipsis the same way as on the canvas.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +21 / -38  |

### Release notes

- Fix sticky note attribution text rendering larger and blurrier than the canvas when a note is exported or copied as a PNG.

### DIFF
--- a/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/note/NoteShapeUtil.tsx
@@ -1,9 +1,7 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 import {
 	Box,
-	DefaultFontFamilies,
 	EMPTY_ARRAY,
-	Editor,
 	Group2d,
 	IndexKey,
 	Rectangle2d,
@@ -461,13 +459,10 @@ export class NoteShapeUtil extends ShapeUtil<TLNoteShape> {
 		})
 
 		const { textFirstEditedBy } = shape.props
-		const attributionFirstName =
+		const attributionName =
 			textFirstEditedBy && !isEmptyRichText(shape.props.richText)
-				? this.editor.getAttributionDisplayName(textFirstEditedBy)?.split(' ')[0]
+				? (this.editor.getAttributionDisplayName(textFirstEditedBy)?.split(' ')[0] ?? null)
 				: null
-		const attributionName = attributionFirstName
-			? truncateAttributionForSvg(this.editor, attributionFirstName, dv.noteWidth)
-			: null
 
 		return (
 			<>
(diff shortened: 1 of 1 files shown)


frolic:
do foreignObjects render correctly when viewed as a normal SVG in your OS? thought there were some constraints there, or maybe this was only constraints around dynamic stuff like iframes in foreignObjects

max-drake (approved):
:)!

mimecuvalo:
> do foreignObjects render correctly when viewed as a normal SVG in your OS? thought there were some constraints there, or maybe this was only constraints around dynamic stuff like iframes in foreignObjects

fwiw, we use foreignObjects in general for our text — they're necessary for all the rich text rendering (can't get around it afaik).

PR #9290, opened 2026-06-17
feat(tldraw): toggle bold/italic on selected shapes via keyboard shortcuts

### LINKED ISSUE #9108: Apply text formatting (bold/italic/underline) to selected shapes via keyboard shortcuts
### What's the feature?

I'd like to format the text of a selected shape (text/geo/note/arrow label) using keyboard shortcuts like cmd+B, cmd+I, cmd+U. As in: select the shape(s) and toggle on/off bold/italic/underline on all its text, without first double-clicking into text-edit mode. This is a common expectation from users coming from other tools (Figma, PowerPoint, Miro, etc). 

Currently, formatting one shape means double-clicking in, selecting all the text, then formatting. There's no way to do it across a multi-selection at all, so bulk formatting is especially painful.

We can roll our own solution in the meantime by editing the rich text JSON directly, but ideally this functionality would be built into the tool by default :) 

mimecuvalo:
Adds keyboard shortcuts to toggle text formatting on selected shapes without entering edit mode, a common expectation from tools like Figma, PowerPoint, and Miro.

## What changed

- `cmd+B` toggles **bold** on the text of selected shapes (text, geo, note, and arrow labels).
- `cmd+I` toggles _italic_. To free up `cmd+I`, the `insert-embed` shortcut moved to `cmd+E` ("embed" rather than "insert"), so there's no overloaded/ambiguous binding.
- `cmd+U` toggles underline. To free up `cmd+U`, `insert-media` moved to `cmd+shift+U`.
- `cmd+shift+K` opens a link dialog that adds, edits, or removes a `link` mark on the text of the selected shapes.
- All of these work across multi-selections. A mark is only removed when **every** selected shape with text is already fully marked; otherwise it is added to all of them. Shapes with an empty label (no text content) are skipped so they don't block toggling off.
- These only fire when shapes are selected but not being edited — in-editor `cmd+B`/`cmd+I`/`cmd+U` while typing is unchanged.

## Implementation

- New public helpers `richTextHasMarkEverywhere` and `setMarkOnRichText` in `packages/tldraw/src/lib/utils/text/richText.ts` that read/transform rich text mark state at the JSON level (including link `href` attrs). These are also useful for anyone who wants to "roll their own" formatting, as the issue mentions.
- New `format-bold` (`cmd+B`), `format-italic` (`cmd+I`), `format-underline` (`cmd+U`), and `format-link` (`cmd+shift+K`) actions plus a `toggleRichTextMarkOnSelection` helper in `actions.tsx`. `insert-embed` is now `cmd+E` and `insert-media` is `cmd+shift+U`.
- The link dialog reuses the existing `EditLinkDialog` UI: the url input/save/clear/cancel form is extracted into a shared inner component in `EditLinkDialog.tsx`, with one wrapper writing a shape's `url` prop (existing behavior) and a new wrapper applying `link` marks to the selection's rich text. No duplicate dialog component.
- Shapes whose label is empty are filtered via the shared `isEmptyRichText` helper (hardened in #9295 to cover non-canonical empty encodings like `content: []`), so empty labels never block toggling a mark off.
- All new actions are surfaced in the keyboard shortcuts dialog.
- Locked shapes are skipped.

### API changes

- Added public `richTextHasMarkEverywhere(richText, markName)` — returns whether every text node in a rich text document has the given mark.
- Added public `setMarkOnRichText(richText, markName, enabled, attrs?)` — returns a copy of the rich text with the given mark added to or removed from every text node; `attrs` supports link `href`s.
- Added `format-bold`, `format-italic`, `format-underline`, and `format-link` UI actions.
- Changed the default `insert-embed` keyboard shortcut from `cmd+I` to `cmd+E`, and `insert-media` from `cmd+U` to `cmd+shift+U`.

## Test plan

- Unit tests for the mark helpers (`richText.test.ts`), including underline and link rendering through the default TipTap extensions.
- Action tests in `formatShortcuts.test.tsx`: toggling bold/italic/underline on single and multi-selections, the link dialog flow, and locked-shape handling.
- Keyboard shortcut tests in `keyboardShortcuts.test.tsx` exercising the real keydown path, including regression tests that empty labels (both the canonical empty-paragraph form and the non-canonical `content: []` form) don't block a fully-marked selection from toggling off. The existing keyboard-shortcut collision test still passes.
- `yarn typecheck`, `yarn lint`, and `yarn api-check` pass.

Manual: add text/geo/note shapes and an arrow with a label, select one or several without entering edit mode, press `cmd+B` / `cmd+I` / `cmd+U` / `cmd+shift+K`, and confirm marks toggle and render. Confirm `cmd+E` opens the embed dialog and `cmd+shift+U` opens the media picker.

Closes #9108


packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -152,6 +153,42 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+		 * marked, the mark is removed from all of them; otherwise it is added to all of them.
+		 * Returns `true` if there was at least one formattable shape to act on, so callers can fall
+		 * back to other behavior when nothing was formatted.
+		 */
+		function toggleRichTextMarkOnSelection(markName: 'bold' | 'italic', source: TLUiEventSource) {
+			if (!editor.isIn('select')) return false
+
+			const shapes = editor
+				.getSelectedShapes()
+				.filter((shape) => 'richText' in shape.props && !editor.isShapeOrAncestorLocked(shape.id))
```
frolic: is there no other way to narrow this type and avoid the `as any` below?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -166,11 +203,33 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				onSelect(source) {
+					toggleRichTextMarkOnSelection('bold', source)
+				},
+			},
+			{
+				id: 'format-italic',
+				label: 'tool.rich-text-italic',
+				icon: 'italic',
+				// No kbd: cmd+i is owned by `insert-embed`, which dispatches to italic when a
+				// text-bearing shape is selected. This action stays available for menus/overrides.
```
frolic: confused by this, sounds like we have an overlapping shortcut?

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -166,11 +203,33 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				onSelect(source) {
+					toggleRichTextMarkOnSelection('italic', source)
+				},
+			},
 			{
 				id: 'insert-embed',
 				label: 'action.insert-embed',
 				kbd: 'cmd+i,ctrl+i',
```
frolic: what about changing this to cmd+e? I would think "insert embed" shortcut centers around "embed" not "insert" (can insert lots of things)

frolic (approved):
looks fine but I don't have a ton of context here, and I worry a bit about the overlapping/overloaded shortcut

if we're not precious about the "insert embed" living where it does, could change it to cmd+e and then free up cmd+i for italic

mimecuvalo:
adding #needs-design-review

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: Done — narrowed with a type guard (`shape is TLShape & { props: { richText: TLRichText } }`) in the filter, so the `as any` reads of `.richText` are gone.

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: Yeah — that overload is gone now. `cmd+i` maps directly to italic and `insert-embed` moved to `cmd+e`, so there's no more smart-dispatch fallback.

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: Done — `insert-embed` is now `cmd+e` and `cmd+i` is bound directly to `format-italic`. This also drops the overloaded dispatch.

packages/tldraw/src/lib/ui/context/actions.tsx
```diff
@@ -152,6 +153,42 @@ export function ActionsProvider({ overrides, children }: ActionsProviderProps) {
...
+				.filter((shape) => 'richText' in shape.props && !editor.isShapeOrAncestorLocked(shape.id))
+
+			if (shapes.length === 0) return false
+
+			trackEvent('rich-text', { operation: markName, source })
+			editor.markHistoryStoppingPoint(`format ${markName}`)
+			editor.run(() => {
+				const allMarked = shapes.every((shape) =>
+					richTextHasMarkEverywhere((shape.props as any).richText, markName)
+				)
```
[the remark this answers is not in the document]
mimecuvalo: Fixed — `toggleRichTextMarkOnSelection` now skips shapes whose rich text has no text content (via `isEmptyRichText`), so an empty geo/arrow label no longer keeps the "is everything marked?" check from being true. Added a regression test in `keyboardShortcuts.test.tsx`. (The embed-fallback half of this is moot now that `cmd+i` no longer routes through `insert-embed`.)

packages/tldraw/src/lib/ui/context/actions.tsx
mimecuvalo: sry, AI comments this time, but i agree - i think we can consider changing this shortcut around. 
marked this PR as #needs-design-review

angrycaptain19:
From design review - agreed we'd shift the embed shortcut away from cmd + I -> something else and let cmd + I be used for italics.

Also support underline & adding a link.

packages/tldraw/src/lib/utils/text/richText.ts
```diff
@@ -154,6 +155,107 @@ export function renderRichTextFromHTML(editor: Editor, html: string): TLRichText
...
+	editor: Editor
+): (TLShape & { props: { richText: TLRichText } })[] {
+	return editor
+		.getSelectedShapes()
+		.filter(
+			(shape): shape is TLShape & { props: { richText: TLRichText } } =>
+				'richText' in shape.props &&
+				!editor.isShapeOrAncestorLocked(shape.id) &&
+				!isEmptyRichText(shape.props.richText)
+		)
```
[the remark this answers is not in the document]
mimecuvalo: Resolved by merging main — #9295 hardened the shared `isEmptyRichText` to treat `content: []` (and the empty-paragraph forms) as empty, and `getFormattableSelectedShapes` now uses that shared helper instead of a local one. Added a regression test in `keyboardShortcuts.test.tsx` covering the non-canonical `content: []` case.

PR #9292, opened 2026-06-17
fix(dotcom): name the file/workspace in delete confirmation dialogs

mimecuvalo:
In order to make file and workspace deletion less ambiguous, this PR includes the name of the item being deleted in the confirmation dialogs. Previously the dialogs read generic copy like "Are you sure you want to delete this file?", which gave no confirmation of which file or workspace was about to be deleted.

The file delete/forget dialog now reads "Are you sure you want to delete {fileName}?" (and the "forget" variant for non-owners), and the delete-workspace confirmation reads "Are you sure you want to delete {workspaceName}? This action cannot be undone." Unnamed workspaces fall back to the existing name placeholder. This matches the existing pattern already used when removing a workspace member.

### Change type

- [x] `improvement`

### Test plan

1. Open a file's menu and click Delete (or Forget as a non-owner); confirm the dialog names the file.
2. Open a workspace's settings and click Delete workspace; confirm the dialog names the workspace.

### Release notes

- Show the file or workspace name in delete confirmation dialogs on tldraw.com.

### DIFF
--- a/apps/dotcom/client/src/tla/components/dialogs/TlaDeleteFileDialog.tsx
+++ b/apps/dotcom/client/src/tla/components/dialogs/TlaDeleteFileDialog.tsx
@@ -67,9 +67,15 @@ export function TlaDeleteFileDialog({
 			<TldrawUiDialogBody style={{ maxWidth: 350 }}>
 				<>
 					{isOwner ? (
-						<F defaultMessage="Are you sure you want to delete {fileName}?" values={{ fileName }} />
+						<F
+							defaultMessage="Are you sure you want to delete <strong>{fileName}</strong>?"
+							values={{ fileName, strong: (chunks) => <strong>{chunks}</strong> }}
+						/>
 					) : (
-						<F defaultMessage="Are you sure you want to forget {fileName}?" values={{ fileName }} />
+						<F
+							defaultMessage="Are you sure you want to forget <strong>{fileName}</strong>?"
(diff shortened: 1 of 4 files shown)


frolic (approved):

apps/dotcom/client/src/tla/components/dialogs/TlaDeleteFileDialog.tsx
```diff
@@ -65,9 +67,9 @@ export function TlaDeleteFileDialog({
 			<TldrawUiDialogBody style={{ maxWidth: 350 }}>
 				<>
 					{isOwner ? (
-						<F defaultMessage="Are you sure you want to delete this file?" />
+						<F defaultMessage="Are you sure you want to delete {fileName}?" values={{ fileName }} />
 					) : (
-						<F defaultMessage="Are you sure you want to forget this file?" />
+						<F defaultMessage="Are you sure you want to forget {fileName}?" values={{ fileName }} />
```
frolic: small design nit but I wonder if we should emphasize the name in some way, so its distinct from the text and easy to see 

> Are you sure you want to forget scratch?

vs

> Are you sure you want to forget **scratch**?

max-drake (approved):

apps/dotcom/client/src/tla/components/dialogs/TlaDeleteFileDialog.tsx
mimecuvalo: good call 👍

Issue #7484, opened 2025-12-25
Performance options example


steveruizok:
## Summary

Expand the existing `custom-options` example or create a new example specifically for performance-related editor options.

## Background

The `TldrawOptions` interface includes several performance-related settings that aren't currently demonstrated:

- `maxShapesPerPage` (default: 4000) - Maximum shapes allowed per page
- `debouncedZoom` (default: true) - Whether to use debounced zoom for optimizations
- `debouncedZoomThreshold` (default: 500) - Shape count threshold for debounced zoom

The current `custom-options` example only shows `maxPages` and `animationMediumMs`.

## Suggested implementation

Either expand `apps/examples/src/examples/custom-options/` or create a new example in `apps/examples/src/examples/performance-options/` that demonstrates:

1. Setting `maxShapesPerPage` and handling the `max-shapes` event
2. Configuring `debouncedZoom` and `debouncedZoomThreshold`
3. Visual feedback when limits are reached

## Related code

- [`packages/editor/src/lib/options.ts`](https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/options.ts) - TldrawOptions interface and defaults
- [`packages/editor/src/lib/editor/Editor.ts`](https://github.com/tldraw/tldraw/blob/main/packages/editor/src/lib/editor/Editor.ts) - `max-shapes` event handling

## Category

`configuration`

## Keywords

`performance`, `options`, `maxShapesPerPage`, `debouncedZoom`, `limits`, `optimization`

mimecuvalo:
actually @angrycaptain19 did this in https://github.com/tldraw/tldraw/pull/9297

PR #9297, opened 2026-06-18
add performance options example

angrycaptain19:
Adds a performance example covering #7484


mimecuvalo:
mm, i think the options can be clearer here / have higher defaults. not sure if this is showcasing performance at higher shape count? capped pretty low

mimecuvalo:
oh, i think i was misreading this PRs intent - i thought maybe this would show off the new perf hooks?
like, `editor.performance.on(event, cb)` etc? and showing off camera FPS, etc. might be good to include in this example or another — thoughts?

mimecuvalo:
claude sez: "Keep #9297 focused on options, but measure honestly — swap the useState re-render counter for an editor.performance.on('camera-end', …) readout showing real FPS / p95 frame time. Same teaching point, real numbers, and it reuses the PerfPanel pattern already written in the performance-hooks example."

mimecuvalo:
sorry, i'm gonna sound like a nag here :P
the FPS tick loop is a great add but, specifically, it'd be interesting if we can also showcase `editor.performance.on(event, cb)` somehow (whether in this example or elsewhere). thoughts?
the tick / FPS counter is a long standing technique in our tldraw codebase. i guess what i'm intending in my comments is the new perf hooks we added in 5.0 would be great to show off.

angrycaptain19:
> sorry, i'm gonna sound like a nag here :P the FPS tick loop is a great add but, specifically, it'd be interesting if we can also showcase `editor.performance.on(event, cb)` somehow (whether in this example or elsewhere). thoughts? the tick / FPS counter is a long standing technique in our tldraw codebase. i guess what i'm intending in my comments is the new perf hooks we added in 5.0 would be great to show off.

Yeah I'm rethinking this one from scratch a little bit to see if it makes senes to have that here or have a full freestanding example for the perf hooks

angrycaptain19:
@mimecuvalo we already have https://examples.tldraw.com/performance-hooks for the perf hooks but just added in using camera end here to capture zoom perf

angrycaptain19:
Going to close this one for now. I might revisit and rethink this one from scratch in a few weeks

PR #9299, opened 2026-06-18
fix: decouple line handle ids from point indices to prevent a drag crash

### LINKED ISSUE #9267: Dragging a handle on a line whose point keys differ from their indices crashes the editor
🗣️: dragging a handle on certain lines crashes the editor with `Error: a2 >= a2`

A line's `props.points` is a map. By convention each entry's key equals the point's `index`, but the schema (`points: T.dict(T.string, …)`) permits any key and never enforces it. On a line where a key and its `index` differ, dragging a vertex handle crashes the editor.

`onHandleDrag` writes the moved point back keyed by `handle.id` — which is the point's `index`, not its actual map key. When those differ, it inserts a second entry under the index instead of updating the existing point, so two points end up sharing an index. The next `getHandles()` calls `getIndexBetween(sameIndex, sameIndex)`, which throws `aN >= aN` (fractional indices must be strictly increasing) and takes down the editor via its error boundary. The same throw can surface on render alone — e.g. the overlay system recomputing handles — for any line that already has duplicate indices.

## Not a regression

This dates to #2890 (Feb 2024), when line points became a fractional-indexed map, and has been latent ever since: no built-in editor path produces a line whose keys differ from their indices — only programmatically or externally authored data does. The earlier "resizing a line inside a group" framing and the resize-batching regression hypothesis were both incorrect; resize (direct, grouped, or interactive) does not reproduce the crash.

## Reproduction

Create a line whose point keys differ from their indices, then drag an endpoint handle:

```js
editor.createShape({
  type: 'line',
  x: 100, y: 100,
  props: { points: {
    a: { id: 'a', index: 'a1', x: 0,   y: 0 },
    b: { id: 'b', index: 'a2', x: 100, y: 0 },
  }},
})
// select the line, then drag either endpoint handle → Error: a2 >= a2
```

jsscclr:
Fixes #9267.

A line's `props.points` is a map of `{ id, index, x, y }`. By convention each entry's key, its `id`, and its `index` are equal, but the code conflated them: `getHandles` published every vertex handle with `id: point.index`, so `onHandleDrag` keyed the moved point back by that index. On a line whose keys differ from their indices (authored via the SDK, or produced by interpolation), this inserted a **new** point under the index, giving two points the same index — and the next `getHandles` called `getIndexBetween(sameIndex, sameIndex)`, which throws `a2 >= a2` and blanks the editor.

### Changes

- **Decouple the handle id from the index.** A vertex handle's `id` is now the point's `id` (its map key), not its `index`, and `onHandleDrag` updates the point in place by `handle.id`. The index is used only for ordering (`sortByIndex`) and for computing create-handle positions. This stops the editor from ever minting a duplicate index on drag.
- **Tolerate a pre-existing duplicate index on read.** `linePointsToArray` dedupes points that share an index, so any already-malformed line (external data, or a transient interpolation frame) renders without crashing. This matches how the rest of the `getIndexBetween` crash family is handled (e.g. #8448).

### Change type

- [x] `bugfix`

### Test plan

1. Create a line whose point keys differ from their indices and drag a vertex — it updates in place, no duplicate index introduced.
2. Create a line with two points sharing an index — `getHandles` does not throw.
3. Animate a line to a different point count — `getHandles` does not throw at any tick.

- [x] Unit tests

### Release notes

- Fixed a crash when dragging a handle on a line whose point keys differed from their indices.

### Related

- #9397 — line interpolation can produce duplicate point indices during animation. This PR's read-path tolerance prevents the resulting crash, but the underlying producer (interpolation emitting duplicates) remains tracked there.

### DIFF
--- a/packages/tldraw/src/lib/shapes/line/LineShapeUtil.test.tsx
+++ b/packages/tldraw/src/lib/shapes/line/LineShapeUtil.test.tsx
@@ -385,3 +385,56 @@ describe('Misc', () => {
 		expect(ids[0]).toEqual(id)
 	})
 })
+
+describe('Points keyed differently from their indices', () => {
+	// The points map allows any string keys, and they need not match each point's
+	// `index`. Dragging a vertex handle on such a line used to insert a new point
+	// keyed by the handle id (the index), producing two points that share an index
+	// and crashing getHandles() via fractional indexing ("a2 >= a2"). See #9267.
+	const mismatchedId = createShapeId('line-mismatched')
+
+	it('updates the dragged point in place instead of duplicating its index', () => {
+		editor.createShapes([
+			{
+				id: mismatchedId,
+				type: 'line',
+				x: 100,
+				y: 100,
+				props: {
+					points: {
+						a: { id: 'a', index: 'a1' as IndexKey, x: 0, y: 0 },
+						b: { id: 'b', index: 'a2' as IndexKey, x: 100, y: 0 },
+					},
+				},
+			},
+		])
+		editor.select(mismatchedId)
+
+		// the endpoint handle's id is its index ('a2'), which differs from its key ('b')
+		const endHandle = getHandlesFor(mismatchedId).find((h) => h.id === 'a2')!
+		const inParent = editor.getShapePageTransform(mismatchedId)!.applyToPoint(endHandle)
+
+		editor
+			.pointerMove(inParent.x, inParent.y)
+			.pointerDown(inParent.x, inParent.y, {
+				target: 'handle',
+				shape: editor.getShape(mismatchedId),
+				handle: endHandle,
+			})
+			.pointerMove(inParent.x + 100, inParent.y)
+			.pointerUp()
+
+		// still exactly two points, keyed as before — no duplicate index introduced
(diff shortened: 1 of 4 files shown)


steveruizok:
This data format is one of my least favorite parts of the API. Very fragile. We just need the keys to remain static; we could even use uuids here instead of relying on the indices to be the same as the keys.

steveruizok:
We usually _don't_ try to accommodate malformed data like this. We'd instead write a migration that fixes the data, and then a validator that crashes the app if the data is invalid.

steveruizok:
Let's try:
- the id and the key should match (it's an "id mapped object")
- the id and the key should NOT necessarily be indices, the `index` property should be the only index that's ever treated as an index, used for sorting, etc. Right now I think we're also expecting the key/ids to be indices.

mimecuvalo:
> We usually _don't_ try to accommodate malformed data like this. We'd instead write a migration that fixes the data, and then a validator that crashes the app if the data is invalid.

if this is the case, then maybe we should rethink https://github.com/tldraw/tldraw/pull/9295 
wdyt @jsscclr and @steveruizok

steveruizok:
Landing! Seems like a very small one that we can't validate and that shouldn't really have been a requirement in the first place (ids being indices).

PR #9296, opened 2026-06-18
chore(dotcom): spike full Docker dev stack

frolic:
Spike: runs the **entire** dotcom dev stack in Docker — client, all four workers, zero-cache, postgres, pgbouncer — one service per container on conventional ports. Under `apps/dotcom/docker-dev/`. Full trade-off writeup + recommendation in [`docker-dev/README.md`](https://github.com/tldraw/tldraw/blob/frolic/dotcom-docker-spike/apps/dotcom/docker-dev/README.md). Alternative to #9273.

**The branch goes all-in so the end state is concrete to evaluate.** Three commits show the progression:
1. The spike stack alongside the host flow (touches nothing).
2. Docker becomes the default `yarn dev-app`; host stack kept as `dev-app:host` for CI.
3. **All-in**: Docker is the only path (CI e2e included); the host orchestration is **deleted** (~1,450 lines) and #9273 becomes moot.

> The recommendation in the README is unchanged (**not** worth it as the daily Mac dev default, because of HMR latency; **is** worth it for onboarding / CI-parity / reproducibility). Commit 3 just makes the all-in version real to look at.

## Verified empirically (per tier)

- **workerd / `wrangler dev`** in a Linux container (native arm64), incl. the real sync-worker (5 durable objects, `Ready on 8787`, routed 404).
- **zero-cache end-to-end** vs fresh containerized postgres: migrate → initial sync → `Replicating` (+ fixed a migrate→Zero ordering race).
- **Postgres logical replication** + **pgbouncer** pooling in containers.
- **Linux install volume** (~2.6 GB, ~59 s warm) + `predev`/`build-i18n` in-container.

## What the end state deletes (~1,450 lines)

`zero-cache/dev.ts` (the orchestrator), `dev-env.ts`(+test), `wait-for-dev-readiness.ts`, `dev-clean.ts`/`dev-doctor.ts`/`docker-compose.ts`/`clean.sh`, `zero-cache/docker/` (host postgres compose), `sync-worker/dev.ts`, and the `package.json` dev/clean/doctor scripts. The worst code goes (PID-tree reaping, port reconciliation, lockfile serialization). **Kept**: the shared worker runner (bemo/analytics/examples use it), `migrate`/`migrations`/`bundle-schema`, and the fly.io deploy artifacts.

## Known rough edges / least-verified

[...]

PR #9301, opened 2026-06-18
chore(dotcom): run dev with process-compose

frolic:
Brings up the dotcom dev stack — client, the four Cloudflare workers, zero-cache, and postgres — with [process-compose](https://github.com/F1bonacc1/process-compose) from [`apps/dotcom/process-compose.yaml`](https://github.com/tldraw/tldraw/blob/frolic/dotcom-process-compose-spike/apps/dotcom/process-compose.yaml), replacing the hand-rolled supervisor in `zero-cache/dev.ts` (and friends, ~960 lines). process-compose is "docker-compose for processes": the same declarative `depends_on` + readiness model, but the services run as **host processes** (native HMR, shared `node_modules`, working worker↔worker bindings); only postgres stays in a container.

## Why

- **Cleaner management of many processes, with real dependencies.** Each service declares what it waits for (`depends_on` + a readiness probe) instead of the imperative `waitForPostgres`/`waitForHttpOk`/process-tree-reaping the old orchestrator open-coded. Adding or reordering a service is a few lines of YAML.
- **A real control surface for automation (incl. AI agents).** A running stack is a client/server with a REST API + scriptable CLI (`process-compose process list` / `logs <name>` / `restart <name>`, JSON output). You can boot it headless, introspect per-process status/logs, and restart a single service — none of which the old stdout-blob supervisor allowed. (`--tui=false` gives the interleaved-logs feel of `concurrently`; the default TUI gives a per-process view.)
- **A clean path to docker-compose** if full containerization is ever worth it — same `depends_on`/readiness shape, and postgres already runs via `docker compose`. (#9296 is that exploration; host/metal is the current preference for the inner loop.)
- **A clean path to parallel stacks** — commands/ports are env-parameterizable, so multiple worktrees / parallel sessions can run their own copy of dotcom side by side via per-stack port blocks (#9302).

## No install step

process-compose has no npm package, so `yarn dev-app` calls the `process-compose` bin of `@tldraw/scripts` ([`process-compose-bin.cjs`](https://github.com/tldraw/tldraw/blob/frolic/dotcom-process-compose-spike/internal/scripts/process-compose-bin.cjs)), which downloads the pinned binary into `node_modules/.cache` on first use — lazy, since Yarn doesn't run a workspace's postinstall, and re-fetched when node_modules is rebuilt, like any bundled-binary dep. Clone → `yarn install` → `yarn dev-app`, nothing else.

## Code changes

| Area | LOC |
| --- | --- |
| `apps/dotcom/process-compose.yaml` — the declarative stack | +105 |
| `internal/scripts/process-compose-bin.cjs` + `@tldraw/scripts` bin wiring | +91 |
| **Deleted** `zero-cache/dev.ts`, `dev-env.ts`(+test), `wait-for-dev-readiness.ts`, `dev-clean.ts`, `dev-doctor.ts`, `docker-compose.ts`, `clean.sh`, `sync-worker/dev.ts` + their scripts | **−957** |
| Wiring + docs — `apps/dotcom/README.md`, root/zero-cache/client/sync-worker `package.json`, playwright config + teardown, CI workflow, `.oxlintrc.json` | small |
| **Net** | **+238 / −986** |

**Kept:** `zero-cache/docker/` (process-compose runs it for postgres), `migrate.ts` + `migrations/` + `bundle-schema` (schema/deploy), and the shared worker runner `internal/scripts/workers/dev.ts` (bemo/analytics/examples use it too).

## Not yet addressed (follow-ups)

- **Fixed ports — one stack at a time.** Every service binds a fixed host port (client 3000, sync-worker 8787, zero-cache 4848, postgres 6543 / pgbouncer 6432, …). So this PR does **not** support running two stacks at once: a second worktree / parallel session — or a postgres container already running in Docker and holding 6543/6432 (e.g. a leftover from a previous run that wasn't shut down cleanly) — collides on those ports and the conflicting service fails to start. Per-worktree port blocks are the follow-up in **#9302**. In the meantime: quit with `F10` so the postgres container is torn down cleanly; `yarn dev-app:clean` clears a leftover one; `lsof -nP -iTCP:<port> -sTCP:LISTEN` finds a stray host process holding a port.
- **Slim or remove `apps/dotcom/client/scripts/dev-app.ts`** (pre-existing on main, out of scope here). It wraps `vite` for the `client` process to (a) fail fast if port 3000 is taken and (b) reap vite's esbuild children on shutdown via `killProcessTree` so they don't orphan. Now that process-compose owns process lifecycle, the kill-tree may be redundant — if process-compose reliably tree-kills the `client` process on stop, `client` could just run `vite` directly, deleting the script + `kill-tree` util + tests. Behavior-sensitive (orphaned esbuild holding :3000 is the failure mode), so verify by stopping the stack and confirming nothing still holds the port before removing.

CI is green (Build, Tests & checks, dotcom e2e all passing). See [`apps/dotcom/README.md`](https://github.com/tldraw/tldraw/blob/frolic/dotcom-process-compose-spike/apps/dotcom/README.md) for how to run and drive the stack.


mimecuvalo:
🍿

apps/dotcom/process-compose.yaml
```diff
@@ -0,0 +1,105 @@
...
+version: '0.5'
+
+# Readiness = TCP connect (a portable node one-liner; wrangler's `/` 404s, so http_get won't do).
+
+processes:
+  # postgres — the one container (wal2json / logical replication)
+  postgres:
+    command: docker compose -f apps/dotcom/zero-cache/docker/docker-compose.yml up
+    shutdown:
+      command: docker compose -f apps/dotcom/zero-cache/docker/docker-compose.yml down
```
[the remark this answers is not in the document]
frolic: Good catch — fixed in 2fe1890535 by pinning the Compose project to `tldraw_dotcom_dev` via a top-level `name:` in `docker-compose.yml`, so `docker compose up`, the process-compose shutdown hook, and `yarn dev-app:clean` all act on the same project/containers/volumes (matching the name the old orchestrator used, so nothing is orphaned holding 6543/6432). The `--env-file` isn't needed here — the compose file has no `${...}` interpolation.

Sent from my aiPhone

apps/dotcom/process-compose.yaml
```diff
@@ -0,0 +1,105 @@
...
+    readiness_probe:
+      exec:
+        command: node -e "require('net').connect(6543,'127.0.0.1').on('connect',function(){this.end();process.exit(0)}).on('error',()=>process.exit(1))"
+      initial_delay_seconds: 2
+      period_seconds: 2
+      failure_threshold: 30
```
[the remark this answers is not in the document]
frolic: Fixed in 2fe1890535 — raised the postgres readiness `failure_threshold` to 90 (~180s budget, matching the old `DOTCOM_DEV_READINESS_TIMEOUT_MS = 180_000`), so a slow first-time image pull no longer trips it and blocks `migrate` and the rest of the stack.

Sent from my aiPhone

package.json
```diff
@@ -41,10 +41,9 @@
 		"refresh-assets": "lazy refresh-assets",
 		"dev": "LAZYREPO_PRETTY_OUTPUT=0 lazy run dev --filter='apps/examples' --filter='packages/tldraw' --filter='apps/bemo-worker' --filter='apps/dotcom/image-resize-worker'",
 		"dev-vscode": "code ./apps/vscode/extension && lazy run dev --filter='apps/vscode/{extension,editor}'",
-		"dev-app": "LAZYREPO_PRETTY_OUTPUT=0 lazy run dev --filter='apps/dotcom/*' --filter='packages/tldraw'",
-		"dev-app:clean": "yarn workspace @tldraw/zero-cache dev:clean",
-		"dev-app:clean:all": "yarn workspace @tldraw/zero-cache dev:clean:all",
-		"dev-app:doctor": "yarn workspace @tldraw/zero-cache dev:doctor",
+		"dev-app": "process-compose -f apps/dotcom/process-compose.yaml up",
+		"dev-app:clean": "docker compose -f apps/dotcom/zero-cache/docker/docker-compose.yml down --volumes && rm -f /tmp/tldraw-dotcom-zero-dev.db /tmp/tldraw-dotcom-zero-dev.db-shm /tmp/tldraw-dotcom-zero-dev.db-wal && rm -rf apps/dotcom/sync-worker/.wrangler/state-dev apps/dotcom/zero-cache/.schema.js",
```
MitjaBezensek, line 45: Docker failing would prevent other steps from running, different to what we had now. Probably still fine though 🤷‍♂️

internal/scripts/process-compose-bin.cjs
```diff
@@ -0,0 +1,88 @@
...
+
+function releaseAsset() {
+	const os = osPlatform()
+	if (os !== 'darwin' && os !== 'linux') {
+		throw new Error(
+			`process-compose auto-install only supports macOS and Linux (got ${os}). ` +
+				`Install it manually: https://f1bonacc1.github.io/process-compose/installation/`
+		)
+	}
+	const arch = osArch() === 'arm64' ? 'arm64' : 'amd64' // map node's "x64" -> release "amd64"
```
MitjaBezensek: I guess it's every non `arm64` -> `amd64`?

apps/dotcom/process-compose.yaml
```diff
@@ -0,0 +1,106 @@
...
+      # ~180s budget: the first run pulls the postgres image, which can take minutes.
+      failure_threshold: 90
+
+  # zero needs .schema.js: build it once, then watch for changes
+  bundle-schema:
+    command: yarn workspace @tldraw/zero-cache bundle-schema
+    availability:
+      restart: 'no' # one-shot; completes
+
+  schema-watch:
```
MitjaBezensek: Should we gate this behind zero-cache readiness? Otherwise we might immediately restart zero-cache if the initial schema changes after zero-cache starts?

apps/dotcom/process-compose.yaml
```diff
@@ -0,0 +1,106 @@
+# Dotcom dev stack, brought up by `yarn dev-app`. Usage notes in apps/dotcom/README.md.
+
+version: '0.5'
+
+# Readiness = TCP connect (a portable node one-liner; wrangler's `/` 404s, so http_get won't do).
+
+processes:
+  # postgres — the one container (wal2json / logical replication)
```
MitjaBezensek: Think its actually two behind the scenes (`pgbouncer` and `zstart_postgres`).

MitjaBezensek (commented):
Looks good 👍 

Much easier to see what's working correctly and what's failing. It's a bit harder to get the logs / grep them, but I guess we'll get used to doing something like `process-compose process logs sync-worker -f`

frolic:
converting to draft until I can come back to this (hopefully next week)

MitjaBezensek:
Heads-up from #9561: the bundle-schema / schema-watch / nodemon / zero-cache-dev chain turned out to be vestigial — it existed to deploy Zero's legacy permissions (`zero-cache-dev -p`), which we stopped using in the Zero 0.25 upgrade, and nothing reads `.schema.js` anymore (query transformation lives in sync-worker via `ZERO_QUERY_URL`/`ZERO_MUTATE_URL`, DDL is handled by the `update_schemas()` hook). Verified live: mid-session `tlaSchema.ts` changes work with zero-cache never restarting, dotcom e2e green. Once #9561 lands you can drop the `bundle-schema`, `schema-watch`, and nodemon wrapping from process-compose.yaml and run `zero-cache` directly.

apps/dotcom/process-compose.yaml
frolic: You're right — it's two (`zstart_postgres` + `pgbouncer`). Updated the comment (and the README) to say "postgres + pgbouncer" in b9fe75592.

Sent from my aiPhone

internal/scripts/process-compose-bin.cjs
frolic: Yeah — and I tightened it in b9fe75592 so it's explicit: `x64 → amd64`, `arm64 → arm64`, and anything else throws (like the OS guard just above), instead of silently mapping an exotic arch to amd64.

Sent from my aiPhone

apps/dotcom/process-compose.yaml
frolic: Good instinct, but it's moot now — I dropped `bundle-schema` and `schema-watch` entirely in b9fe75592. Per #9561 that whole chain is vestigial (nothing reads `.schema.js` after the Zero 0.25 upgrade), so zero-cache runs directly with no schema-rebuild/restart watch. The race you're describing can't happen anymore.

Sent from my aiPhone

package.json
frolic: Yeah, this is a deliberate behavior change: process-compose's `depends_on` means if postgres/Docker fails the whole stack (including the client) refuses to start, whereas the old setup would still boot the client in a degraded state. I've kept the strict cascade — a client with no backend is mostly just confusing to debug — but if you'd rather it boot degraded, loosening the client's `depends_on` is a one-liner. Happy to change it if you feel strongly.

Sent from my aiPhone

frolic:
Thanks for the review 🙏 Addressed the inline comments in b9fe75592, and — per your #9561 heads-up — merged latest main and aligned the config with it: dropped the now-vestigial `bundle-schema` / `schema-watch` / nodemon chain, so `zero-cache` runs directly (`yarn exec zero-cache`) with `NODE_ENV` + conn-limit env matching the old orchestrator. That takes it from 10 processes to 8.

On grepping logs: `PC_DISABLE_TUI=1 yarn dev-app` runs it headless with concurrently-style interleaved logs you can pipe/grep, and `yarn exec process-compose process logs <name> -f` follows a single service (both are in the README now).

Left the "Docker failing blocks the rest" thread open — that one's a genuine design call (strict cascade vs. boot-degraded), curious what you think.

Sent from my aiPhone

mimecuvalo (approved):
let's do iiiit.
can you wait to land this until tomorrow when we're in office? just wanna make sure if there are issues that we can debug it in real time with the team :)

frolic:
I'll be wfh tomorrow so will save until wednesday!

jsscclr (approved):

PR #9307, opened 2026-06-18
fix(tldraw): drag a selected hollow shape that overlaps a filled shape behind it

### LINKED ISSUE #8733: Can't drag a selected hollow shape that overlaps a filled shape behind it
## Description

When a hollow shape (no fill) sits in front of a filled shape and overlaps it, you can't translate the hollow shape by dragging — even when it's the only selected shape. The drag goes to the filled shape behind instead, and the selection switches.

## Steps to reproduce

1. Draw a large rectangle and set its fill to solid.
2. Draw a smaller rectangle in front of it with fill set to none, positioned so its interior is over the filled rectangle.
3. Click the hollow rectangle to select it.
4. Press and drag from inside the hollow rectangle (over the filled one).

## Expected

The hollow rectangle starts translating, since it's the selected shape and the pointer is inside its bounds.

## Actual

Pointer-down hits the filled shape behind. The hollow shape gets deselected and the filled shape is selected/translated instead.

## Notes

The right-click path in `SelectTool/childStates/Idle.ts` already handles this case — it checks the selection bounds before falling back to `getShapeAtPoint`, with a comment noting "preserves it, even when a filled shape sits behind it." The same selection-bounds-first check is missing from the canvas `onPointerDown` path, which still calls `getHitShapeOnCanvasPointerDown` first. That helper uses `getShapeAtPoint` (which returns the filled shape behind) and only falls back to `getSelectedShapeAtPoint` when nothing is hit at all — so the fallback never triggers in this scenario.

kostyafarber:
## What
2026-06-18 at 14 54 27 - Cyan Partridge

2026-06-18 at 14 52 34 - Plum Guanaco

Fixes a case where an already-selected hollow (no-fill) shape couldn't be dragged from its interior when it overlapped a filled shape sitting behind it. Pointing down inside the hollow shape's interior resolved to the filled shape, deselecting the hollow one.

Closes #8733

## Why it happened

`getHitShapeOnCanvasPointerDown` returned:

```ts
editor.getShapeAtPoint(point, { hitInside: false, ... }) ?? editor.getSelectedShapeAtPoint(point)
```

`getShapeAtPoint({ hitInside: false })` skips a hollow shape's interior, so a point inside the hollow shape resolves to the filled shape behind it. Because of the `??`, the `getSelectedShapeAtPoint` fallback only fired when *nothing* was hit — so it never rescued the hollow-over-filled case. The wrong shape then flowed into `PointingShape`, which committed the selection to the filled shape.

## The fix

Prefer a *selected* shape under the pointer (tested with `hitInside`, so a hollow interior counts) over a shape behind it — but only when the selected shape is in front of, or there is no, other hit shape (z-order check via `getCurrentPageShapesSorted`). That keeps a selected hollow shape grabbable from its interior while still letting a shape stacked *on top* of the selection win a click.

Fixing it in the helper (rather than in `PointingShape`) corrects the input for all of its callers — `Idle`, `EditingShape`, and `Crop` — at the point where "what's under the pointer" is actually decided.

## Tests

Added a `selection-omnibus` describe block covering:
- pointer down inside a selected hollow shape keeps it selected (doesn't fall through to the filled shape behind)
- dragging from the interior translates the hollow shape
- a filled shape stacked *in front* of the selection still wins a click (guards against over-capture)

All 158 selection-omnibus tests pass.

### DIFF
--- a/packages/tldraw/src/lib/tools/selection-logic/getHitShapeOnCanvasPointerDown.ts
+++ b/packages/tldraw/src/lib/tools/selection-logic/getHitShapeOnCanvasPointerDown.ts
@@ -8,16 +8,30 @@ export function getHitShapeOnCanvasPointerDown(
 	const zoomLevel = editor.getZoomLevel()
 	const currentPagePoint = editor.inputs.getCurrentPagePoint()
 
-	return (
-		// hovered shape at point
-		editor.getShapeAtPoint(currentPagePoint, {
-			hitInside: false,
-			hitLabels,
-			hitLocked: editor.options.selectLockedShapes,
-			margin: editor.options.hitTestMargin / zoomLevel,
-			renderingOnly: true,
-		}) ??
-		// selected shape at point
-		editor.getSelectedShapeAtPoint(currentPagePoint)
-	)
+	// The shape whose geometry is under the pointer. For hollow shapes this skips
+	// the interior, so a point inside a hollow shape can resolve to a filled shape
+	// sitting behind it.
+	const hitShape = editor.getShapeAtPoint(currentPagePoint, {
+		hitInside: false,
+		hitLabels,
+		hitLocked: editor.options.selectLockedShapes,
+		margin: editor.options.hitTestMargin / zoomLevel,
+		renderingOnly: true,
+	})
+
+	// A selected shape under the pointer (tested with hitInside, so a hollow
+	// selected shape's interior counts) should win over a shape behind it. This
+	// keeps a selected hollow shape grabbable even when a filled shape sits behind
+	// it. We only prefer it when it's in front of (or there is no) other hit shape,
+	// so clicking a shape stacked on top of the selection still selects that shape.
+	const selectedShapeAtPoint = editor.getSelectedShapeAtPoint(currentPagePoint)
+	if (selectedShapeAtPoint) {
+		if (!hitShape) return selectedShapeAtPoint
+		const sorted = editor.getCurrentPageShapesSorted()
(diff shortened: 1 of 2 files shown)


steveruizok:
Capturing here what we discussed in person: I think there's some switch on whether the hollow shape is "visually containing" the filled shape, or vis versa. The user's cursor _usually_ should interact with whatever is under it, but what's under the cursor is a complicated answer. (I did a [conference talk](https://gitnation.com/contents/whats-under-the-cursor) on this recently, I could present it in person).

I think that logic is still incomplete though.

From a UX perspective, hollow shapes and draw shapes are effectively on the same z-index, since the z-index is impossible to know just by looking at the shapes (unless there are multiple colors). We shouldn't expect the user to know which shape is on top.

image

In arrow binding, if the hit test under the cursor has reached the canvas (ie, has not hit any filled shape or edge on the way down), then we select the smallest shape that the user has clicked into regardless of stacking order.

https://github.com/user-attachments/assets/d74acc45-a5be-4ade-9a39-638e83e42c9f

Maybe we can extend this to do something similar for clicks and drags.

image

This might also let us drop the "label clicking" which is complex and unintuitive.

But z-index would matter with filled shapes; we'd have to at least stop descending to the canvas once we've hit a filled shape.

image

Previously, the logic was: closest edge first, then first filled shape, then smallest enclosing shape whose label we hit.

Perhaps the priority should be: closest-edge-first, then smallest enclosing shape, stopping when we hit a filled shape or frame?

kostyafarber:
will close for now, but an interesting discussion here for future direction

PR #9309, opened 2026-06-18
fix(dotcom): guard zero-cache migrations against blocking column backfills

jsscclr:
In order to prevent the deploy-time outage we hit with `group.inviteLinkEnabled`, this PR adds a test that rejects new zero-cache migrations using `ADD COLUMN ... DEFAULT`, and documents the expand/contract pattern to use instead. Relates to #9031 (the Supabase DDL schema-change hooks this builds on).

### Background

On the current Zero version (`@rocicorp/zero` 1.6.1), adding a column **with a default** is not a metadata-only change for Zero: the replication-manager re-copies the new column across every existing row into its replica before the column becomes visible to the view-syncers. While that backfill runs it holds the Postgres replication slot open, so WAL accumulates.

In the `inviteLinkEnabled` incident, the client bundle (which expected the column) shipped before the backfill finished, so the view-syncers were still serving the old schema and clients got `SchemaVersionNotSupported` → the "this version of tldraw is no longer supported / please reload" dialog. It self-healed once the backfill completed (replication lag fell from ~355k back to baseline), but during the window it also drove the replication-lag/WAL alerts.

### What this does

- Adds `validateMigrationContents` and a unit test that fails when a migration uses `ADD COLUMN ... DEFAULT`, pointing authors to the expand/contract pattern: add the column nullable with no default (visible immediately, so client code can ship right away), backfill values in batches, then set the default / NOT NULL in a later migration.
- The check inspects migration **contents** statically rather than running inside the migration runner, so it never blocks historical migrations replaying on a fresh database (local dev, new preview environments). Existing `ADD COLUMN ... DEFAULT` migrations are grandfathered by number, SQL comments are stripped before matching, and a column on a table Zero does not replicate can opt out with a `-- zero-cache:allow-add-column-default <reason>` marker.
- Documents the convention in `AGENTS.md`.

The check is a version-gated stopgap: Rocicorp confirmed a newer Zero applies scalar defaults instantly with no backfill. Once we upgrade past 1.6.1, the check (and the grandfather list) can be removed — the code comment notes this.

### Change type

- [x] `other`

### Test plan

1. `cd apps/dotcom/zero-cache && yarn test run validateMigrationContents.test.ts` — passes (includes a check that all migrations currently in the package are clean).
2. The test covers: a new `ADD COLUMN ... DEFAULT` is rejected; the expand step (nullable, no default) and contract steps (`ALTER COLUMN ... SET DEFAULT` / `SET NOT NULL`) pass; the pattern inside a comment is ignored; the opt-out marker is honored; grandfathered migrations are allowed.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Apps           | +60 / -0   |
| Tests          | +82 / -0   |
| Config/tooling | +1 / -0    |

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,7 +118,7 @@ Store and schema:
 
 - Store changes should respect migrations, validators, and schema versioning.
 - Schema-affecting changes usually need updates in `packages/tlschema` and focused migration tests.
-- When adding a column to a Zero-replicated table (`apps/dotcom/zero-cache/migrations`), add it nullable with no default, backfill values in batches, then set the default / NOT NULL in a later migration. A single `ADD COLUMN ... NOT NULL DEFAULT` forces a blocking replica backfill that hides the column from view-syncers and holds the Postgres replication slot open until it finishes. `validateMigrationContents` (a unit test) enforces this for new migrations; a column on a table Zero does not replicate can opt out with a `-- zero-cache:allow-add-column-default <reason>` comment.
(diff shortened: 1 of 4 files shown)


jsscclr:
Not sure we need this 🤔

PR #9317, opened 2026-06-18
fix(dotcom): make TlaMenuSelect open on iOS

steveruizok:
In order to make the workspace member-role selects (and other `TlaMenuSelect` menus) usable on iOS, this PR fixes the wrapper that was swallowing the trigger click. Radix Select opens its dropdown on `click` for touch and pen input, but only on `pointerdown` for mouse. The `TlaMenuSelect` wrapper called `stopPropagation()` in the capture phase, so the click was stopped before it could descend to the trigger — harmless on desktop (already opened on `pointerdown`) but on iOS it meant the menu never opened and the control looked dead. Moving the guard to the bubble phase lets the trigger handle the click first, then still stops the click from reaching ancestor handlers (e.g. a background dismiss).

This also tightens the menu select trigger height from 36px to 24px.

### Change type

- [x] `bugfix`

### Test plan

1. On an iOS device or Simulator, open the workspace settings dialog.
2. Tap a member's role select.
3. Confirm the dropdown opens and options can be selected.
4. Confirm desktop mouse behaviour is unchanged.

### Release notes

- Fix workspace settings selects not opening on iOS.

### Code changes

| Apps           | +7 / -2    |

### DIFF
--- a/apps/dotcom/client/src/tla/components/tla-menu/menu.module.css
+++ b/apps/dotcom/client/src/tla/components/tla-menu/menu.module.css
@@ -108,7 +108,7 @@
 	position: relative;
 	display: flex;
 	align-items: center;
-	height: 36px;
+	height: 24px;
 	gap: 4px;
 	justify-content: flex-end;
 	cursor: pointer;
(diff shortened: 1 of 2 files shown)


apps/dotcom/client/src/tla/components/tla-menu/menu.module.css
```diff
@@ -108,7 +108,7 @@
 	position: relative;
 	display: flex;
 	align-items: center;
-	height: 36px;
+	height: 24px;
```
steveruizok, line 111: driveby

PR #8410, opened 2026-04-01
feat(themes): add first-class theme system with display values for customizing default shapes

steveruizok:
## Summary

This PR introduces a first-class theme system, replacing the previous approach where colors were hardcoded and resolved inline. Themes are now named, registered objects that shape utils consume via a structured display values pipeline.

https://github.com/user-attachments/assets/a536084f-b999-41e7-981b-a365b1f5b6c3

### Concepts

| Term | Type | Meaning |
|------|------|---------|
| **Theme** | `TLTheme` | A named set of colors and typographic values (`fontSize`, `lineHeight`, `strokeWidth`) for both light and dark modes. Register custom themes via `TLThemes` module augmentation for type-safe IDs. |
| **Theme ID** | `TLThemeId` | A type-safe key into `TLThemes` (e.g. `'default'`, `'corporate'`). Derived as `keyof TLThemes & string`. |
| **Theme default colors** | `TLThemeDefaultColors` | The full color interface for one mode. Augmentable to add custom named colors. |
| **Theme colors** | `TLThemeColors` | The resolved palette type used by `TLTheme`. Equals `TLThemeDefaultColors` minus any keys in `TLRemovedDefaultThemeColors`. |
| **Color mode** | `'light' \| 'dark'` | The resolved appearance, derived from the user's color scheme preference. |
| **Color scheme** | `'light' \| 'dark' \| 'system'` | The user's preference. `'system'` resolves to a color mode based on OS settings. |
| **Display values** | `getDefaultDisplayValues` / `getCustomDisplayValues` | Shape util callbacks that produce cached, computed visual properties (colors, stroke widths, font sizes) from the theme and color mode. |
| **ThemeManager** | `ThemeManager` | Editor manager that stores themes, tracks the active theme, and exposes the resolved color mode. |

### Module augmentation

```ts
// Register a new theme ID
declare module '@tldraw/tlschema' {
  interface TLThemes {
    corporate: TLTheme
  }
}

// Register a new color name
declare module '@tldraw/tlschema' {
  interface TLThemeDefaultColors {
    pink: TLDefaultColor
  }
}

// Remove built-in colors from the palette
declare module '@tldraw/tlschema' {
  interface TLRemovedDefaultThemeColors {
    'light-violet': true
    'light-blue': true
  }
}
```

### Removing colors

[...]

PR #1636, opened 2023-06-22
Styles API follow-ups

SomeHats:
tldraw-zero themed follow-ups to the styles API added in #1580.

- Removed style related helpers from `ShapeUtil`
- `editor.css` no longer includes the tldraw default color palette. Instead, a global `DefaultColorPalette` is defined as part of the color style. If developers wish to cusomise the colors, they can mutate that global.
- `ShapeUtil.toSvg` no longer takes font/color. Instead, it takes an "svg export context" that can be used to add `<defs>` to the exported SVG element. Converting e.g. fonts to inlined data urls is now the responsibility of the shapes that use them rather than the Editor.
- `usePattern` is not longer a core part of the editor. Instead, `ShapeUtil` has a `getCanvasSvgDefs` method for returning react components representing anything a shape needs included in `<defs>` for the canvas.
- The shape-specific cleanup logic in `setStyle` has been deleted. It turned out that none of that logic has been running anyway, and instead the relevant logic lives in shape `onBeforeChange` callbacks already.

### Change Type
- [x] `minor` — New feature

### Test Plan

- [x] Unit Tests
- [x] End to end tests

### Release Notes
 --

PR #9320, opened 2026-06-18
perf(tldraw): cache fill-pattern tiles across editor instances

jsscclr:
**Draft.** Real-performance fix found by profiling a mobile file-open on tldraw.com; verify on the preview/simulator.

## Problem
Opening a file on tldraw.com pauses for seconds on mobile. A Safari sampling profile of one file tap (iOS Simulator, screenshots disabled) attributes **~66% of the entire interaction to `canvas.toBlob`** — ~2.3s. The caller is `usePattern` in `defaultStyleDefs.tsx`: it generates the hatch fill-pattern tiles in a mount effect, rasterizing a canvas and encoding each tile with `toBlob`, once per zoom LOD and theme, with the largest LODs being big retina-sized canvases.

The tiles lived in component state and the object URLs were revoked on unmount, so every editor that mounts regenerates the whole set. tldraw.com remounts the editor on every file switch (`key={fileSlug}`), so this cost is paid on every open — and on mobile Safari `toBlob` is slow enough to block the main thread for over a second.

## Fix
The tiles are a pure function of `(dpr, zoom LOD, solid color)`, so cache the generated object URLs in a module-level map and stop revoking them. Each tile is encoded at most once per page; remounting the editor (or a theme/dpr that's been seen before) reuses the cached tiles. The cache key includes the solid color so custom themes stay correct.

## Expected effect
- First editor mount in a session: unchanged (tiles still generate once).
- Every subsequent file open / editor remount: `toBlob` time drops to ~0 — removing the dominant cost of the tap.

## Verification
- Re-trace a file open on the preview/simulator: the `toBlob` self-time should be present on the first open and gone on the second.
- `getSvgString` pattern export test passes; typecheck and API report unchanged (cache helpers are module-private).

## Follow-ups (not in this PR)
- The first open still pays the full generation. Generating only the LODs actually needed (rather than all up to `maxEffectiveZoom`) would cut that too.
- Reusing the editor instance across file switches would remove the remaining mount costs (font-face setup, text measurement) and the document-hydration stall.

## Why it was this way (and why dropping the revoke is safe)

This isn't a case of caching being rejected — it was never needed under the original assumptions, which then drifted:

- **#1636 (Jun 2023)** introduced the tiles as a *fixed* set of 6 small zoom levels, generated once per mount, with **no `revokeObjectURL`**. For the normal "one editor for the life of the page" case that's perfectly reasonable — nothing to cache.
- **#3801 (May 2024) "fix pattern fill lods"** made the LOD set *dynamic* (camera options broke the hardcoded 1–6 range) and put `maxZoom` in the effect deps, so the effect could now regenerate at runtime. The `revokeObjectURL` cleanup was added **here**, as memory hygiene for that regeneration case — not as a stance against caching.
- **#8441** later scaled the effective zoom to the worst-case `maxZoom/minZoom` ratio, which is what pushed the top LODs up to 2048²–4096² canvases — the expensive tiles arrived in a separate change.
- Remounting the editor per document is an app-level decision (tldraw.com keys the editor by file), invisible to the SDK.

So the cost emerged from the *composition* of three individually-reasonable changes, not a single mistake.

**On removing the revoke:** its original job was cleaning up orphaned URLs when the effect regenerated on a `dpr`/`maxZoom`/theme change. With the cache those URLs are no longer orphaned — they're retained keyed by `(dpr, zoom, solid)` and reused if those inputs recur (e.g. toggling dark mode back). So this is intentional retention, not a leak, and the growth is bounded to the small set of distinct `(dpr, zoom, solid)` combinations seen in a session (a handful of small encoded PNG blobs).

## Why generate at runtime (not at build time)?

The tiles can't be baked at build time because all three of their inputs only exist at runtime — the cache key is literally `(dpr, zoom, solid)`: device pixel ratio (the user's display), the theme `solid` colors (runtime-customizable since the theme system in #8410), and the LOD count (derived from configurable camera `zoomSteps`, per #3801). As an SDK embedded by arbitrary apps, none of those are known at the SDK's build time, and the tiles are transient in-memory object URLs (never persisted), so baking them would freeze customizable assumptions and add bundle weight for something cheaply derivable on demand. The remaining levers are this cache (kills repeat cost) and lazy/capped LODs (first-open cost) — tracked as follow-ups, not in this PR.

### DIFF
--- a/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.test.ts
+++ b/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.test.ts
@@ -1,0 +1,41 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { clearPatternImageUrlCacheForTests, getOrCreatePatternImageUrl } from './defaultStyleDefs'
+
+describe('getOrCreatePatternImageUrl cache', () => {
+	beforeEach(() => {
+		clearPatternImageUrlCacheForTests()
+		// jsdom may not implement createObjectURL; stub it so the cache can resolve.
+		;(URL as any).createObjectURL = vi.fn(() => 'blob:mock')
+	})
+
+	const makeGenerator = () => vi.fn(() => Promise.resolve(new Blob()))
+
+	it('generates each (dpr, zoom, solid) tile at most once', async () => {
+		const generate = makeGenerator()
+		await getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		await getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		expect(generate).toHaveBeenCalledTimes(1)
+	})
+
+	it('returns the same promise for a repeated key', () => {
+		const generate = makeGenerator()
+		const first = getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		const second = getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		expect(second).toBe(first)
+	})
+
+	it('generates separately for different solid colors (theme correctness)', async () => {
+		const generate = makeGenerator()
+		await getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		await getOrCreatePatternImageUrl(2, 4, '#000000', generate)
+		expect(generate).toHaveBeenCalledTimes(2)
+	})
+
+	it('generates separately for different dpr and zoom', async () => {
+		const generate = makeGenerator()
+		await getOrCreatePatternImageUrl(1, 4, '#ffffff', generate)
+		await getOrCreatePatternImageUrl(2, 4, '#ffffff', generate)
+		await getOrCreatePatternImageUrl(2, 8, '#ffffff', generate)
+		expect(generate).toHaveBeenCalledTimes(3)
+	})
+})
--- a/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.test.ts
+++ b/packages/tldraw/src/lib/shapes/shared/defaultStyleDefs.test.ts
@@ -1,11 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import {
-	clearPatternImageUrlCacheForTests,
-	getOrCreatePatternImageUrl,
-	PATTERN_IMAGE_CACHE_MAX,
-} from './defaultStyleDefs'
-
-const flushMicrotasks = () => new Promise((resolve) => setTimeout(resolve, 0))
+import { clearPatternImageUrlCacheForTests, getOrCreatePatternImageUrl } from './defaultStyleDefs'
 
 describe('getOrCreatePatternImageUrl cache', () => {
 	beforeEach(() => {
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
i _think_ the cache won't grow unbounded? can you check before landing @jsscclr 
pre-approving tho

jsscclr:
@mimecuvalo good call to check — it's bounded in practice. The cache key is `(dpr, zoom-LOD, theme-solid)`:

- **zoom** — only ever a small set of power-of-two LODs (`getPatternLodsToGenerate`), bounded by the camera's zoom range.
- **dpr** — a handful of device / browser-zoom values.
- **solid** — the *theme's* solid color, **not** the shape's. Two values for the default light/dark theme.

Worth stressing: **shape colors aren't part of the key.** The hatch tile is monochrome (the theme solid); a shape's own color is applied as a separate `<path fill={fillColor}>` under the shared `url(#hash_pattern…)` overlay in `PatternFill`. So a canvas with thousands of distinct shape colors adds **zero** tiles.

The only way it can grow without bound is an app that animates the *theme's* solid color at runtime. I did try capping that with an LRU + revoke, but Bugbot rightly caught (the Medium comment) that revoking shared object URLs can pull a tile out from under a still-mounted editor — `usePattern` only touches the cache when its effect runs, so an on-screen tile can look "least recently used" and get revoked, breaking the fill. LRU tells you what's least *useful to keep*, not what's *safe to destroy*; that needs reference counting. So I reverted the cap (`0cc8760ef`) and documented the bound instead — the leak in that exotic theme-animation case is slow and bounded-rate, and far safer than revoking live tiles. If we ever need hard bounding, ref-counting the URLs is the right fix (and pairs with the editor-reuse work).

jsscclr:
> @mimecuvalo good call to check — it's bounded in practice. The cache key is `(dpr, zoom-LOD, theme-solid)`:
> 
> * **zoom** — only ever a small set of power-of-two LODs (`getPatternLodsToGenerate`), bounded by the camera's zoom range.
> * **dpr** — a handful of device / browser-zoom values.
> * **solid** — the _theme's_ solid color, **not** the shape's. Two values for the default light/dark theme.
> 
> Worth stressing: **shape colors aren't part of the key.** The hatch tile is monochrome (the theme solid); a shape's own color is applied as a separate `<path fill={fillColor}>` under the shared `url(#hash_pattern…)` overlay in `PatternFill`. So a canvas with thousands of distinct shape colors adds **zero** tiles.
> 
> The only way it can grow without bound is an app that animates the _theme's_ solid color at runtime. I did try capping that with an LRU + revoke, but Bugbot rightly caught (the Medium comment) that revoking shared object URLs can pull a tile out from under a still-mounted editor — `usePattern` only touches the cache when its effect runs, so an on-screen tile can look "least recently used" and get revoked, breaking the fill. LRU tells you what's least _useful to keep_, not what's _safe to destroy_; that needs reference counting. So I reverted the cap (`0cc8760ef`) and documented the bound instead — the leak in that exotic theme-animation case is slow and bounded-rate, and far safer than revoking live tiles. If we ever need hard bounding, ref-counting the URLs is the right fix (and pairs with the editor-reuse work).

this is claude wtf

steveruizok:
Missed this and sniped it with #9345 :x

PR #9324, opened 2026-06-19
perf(editor): cache resolved font faces across editor instances

jsscclr:
## Problem
After the pattern-tile cache (#9320), profiling a tldraw.com file open on the iOS Simulator showed the next-biggest cost was `findOrCreateFontFace` — ~385 ms, attributed to `next@native` iterating the document's `FontFaceSet`. Crucially this was on a *warm* open (the pattern `toBlob` was already ~0), so it's a per-mount cost, not first-load.

Root cause: `findOrCreateFontFace` deduplicates font faces with an O(n) scan of `document.fonts` on every lookup. It runs for each font on every editor mount, and a fresh editor (e.g. switching documents on tldraw.com, which remounts the editor) gets a fresh `FontManager` with no memory of the previous one — so the scan re-runs every mount.

The trace ruled out recreation: the `FontFace` constructor was ~11 ms, `load()` ~0 ms, and there were zero font-file network requests — faces are *reused*, just found via a slow scan.

## Fix
Cache resolved font faces in a per-document `Map` (keyed by family + descriptors), shared across `FontManager` instances, so repeated lookups and remounts are O(1). On a cache miss the existing `document.fonts` scan still runs once, so faces added outside the manager (preloaded fonts) are still reused rather than duplicated. Keyed per document for cross-window embedding; faces are never removed from `document.fonts`, so the cache stays valid for the document's lifetime.

## Tests
`FontManager.test.ts`: a fresh manager on the same document reuses the existing face instead of recreating it (the remount regression), and fonts with different descriptors get distinct faces. The cross-document cache has a test-only reset. All 22 tests pass; typecheck clean; no public-API change (the reset helper is `@internal` and not re-exported).

## Verification
Re-trace a file open on the simulator: `findOrCreateFontFace`'s scan cost should be present on the first open and ~0 on subsequent opens.

## Context
This is a sibling to #9320 (pattern-tile cache) — same symptom (per-mount cost on tldraw.com file switches), different root cause: patterns had the wrong lifetime (per-instance state); fonts had the wrong data structure (O(n) scan vs O(1) lookup).

### DIFF
--- a/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
@@ -9,7 +9,7 @@ import {
 import { IndexKey } from '@tldraw/utils'
 import { Mock, Mocked, vi } from 'vitest'
 import { Editor } from '../../Editor'
-import { FontManager } from './FontManager'
+import { clearFontFaceCacheForTests, FontManager } from './FontManager'
 
 // Mock the Editor class
 vi.mock('../../Editor')
@@ -73,6 +73,7 @@ describe('FontManager', () => {
 
 	beforeEach(() => {
 		vi.clearAllMocks()
+		clearFontFaceCacheForTests()
 
 		mockAssetUrls = {
 			'test-font.woff2': 'https://example.com/fonts/test-font.woff2',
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
nice one

PR #9325, opened 2026-06-19
impovement(tldraw): toggle editing when actions have the completeEditing flag

kaneel:
Some actions such as duplicate leave the users still in editing mode, when pressing undo they end up removing the text they were actually editing.

This PR is a follow-up to https://github.com/tldraw/tldraw/pull/9260 and implements a boolean flag for actions to automatically buck out of editing mode. 

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a geo shape
2. Enter editing mode
3. type some text
4. press on "duplicate"
5. do an undo

=> the undo should remove the newly created shape 

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Implements "completeEditing" flag for actions that end the editing of a shape.

### DIFF
--- a/apps/docs/content/releases/next.mdx
+++ b/apps/docs/content/releases/next.mdx
@@ -24,6 +24,7 @@ This release simplifies multi-click handling down to a single double-click event
 
 ## Improvements
 
+- Actions that act on a shape as an object — duplicate, delete, group, align, distribute, rotate, flip, reorder, move to a new page, and similar — now end text editing before they run. Actions that change a shape's content or appearance, such as style changes, leave you editing. ([#0000](https://github.com/tldraw/tldraw/pull/0000))
 - Add a `shift+q` shortcut that copies the styles of the hovered shape and applies them to the next shapes you create. ([#9028](https://github.com/tldraw/tldraw/pull/9028))
 - Keep sync at full speed when the same multiplayer room is open in multiple tabs, windows, or devices — previously a room with no other users throttled network sync to once per second. ([#8988](https://github.com/tldraw/tldraw/pull/8988))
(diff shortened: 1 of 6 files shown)


steveruizok:
Could we modify the `mustGoBackToSelectToolFirst` and be more specific about being in / going to `select.idle`? Then we could re-use that. It wouldn't affect any of the current places where `mustGoBackToSelectToolFirst` is used.

kaneel:
> `mustGoBackToSelectToolFirst`

it does not seem to do exactly the same thing and is not called in the same place, if anything we could clean this helper and use the same "declarative" logic though.

If we really really want to keep the help and all the call point, we _could_ pass some option argument to it. 

So it's down to what's the preferred API

steveruizok:
> > `mustGoBackToSelectToolFirst`
> 
> it does not seem to do exactly the same thing and is not called in the same place, if anything we could clean this helper and use the same "declarative" logic though.
> 
> If we really really want to keep the help and all the call point, we _could_ pass some option argument to it.
> 
> So it's down to what's the preferred API

I think in most cases we should disable these actions while in the editing state. We disable the duplicate action while in other tools, but we enable them in the select tool. Maybe we should only enable them when in select.idle? We shouldn't be duplicating while editing a shape, for example. Are there any that definitely should be possible while editing?

kaneel:
> I think in most cases we should disable these actions while in the editing state. 

Ha, I see, totally different work then! After our chat yesterday I was not under the impression this would be about disabling some actions but that will indeed prevent "the thing we try to prevent to happen" to happen

> Are there any that definitely should be possible while editing

Well there are some actual editing actions that definitely can stay here.

PR #9327, opened 2026-06-19
fix(dotcom): serve source maps instead of stripping them from the deploy

### LINKED ISSUE #9322: tldraw.com strips deployed source maps but keeps the sourceMappingURL comments, causing 404s and no devtools symbolication
🗣️: tldraw.com strips JS source maps from the deployed bundle but leaves the //# sourceMappingURL comments pointing at them, so the browser 404s on every .map file and devtools can't symbolicate — even though the source is public at tldraw/tldraw.

---

The tldraw.com client build generates JavaScript source maps, uploads them to Sentry, and then deletes the `.map` files from the deployed assets — but the shipped `.js` files still contain `//# sourceMappingURL=…` comments pointing at those now-missing files. The result is that every page load produces 404s for the absent maps, and anyone debugging a production or preview build in browser devtools sees only minified names with no symbolication. This is the standard "hidden source maps" recipe, whose usual justification is keeping source private — but that rationale does not apply here, since the client source is already public in this repo.

The desired outcome is to **serve the source maps publicly** rather than stripping them: stop deleting the `.map` files from the deploy so that the existing `sourceMappingURL` comments resolve, the 404s disappear, and anyone (team or community) debugging a production or preview build gets real names and lines directly in devtools — no Sentry round-trip needed. The Sentry upload should continue to work as before. This affects both preview and production deploys, which use the same build and upload path.

## Open questions

1. **Which outcome do we want: stop serving maps cleanly, or actually serve them?**
   Serve them. Since the client source is already public in this repo, there is no source-protection reason to withhold the maps, and serving them makes devtools symbolication work for everyone debugging prod/preview without going through Sentry. (Keep the Sentry upload intact.)
2. **Are there deploy-size, asset-count, or cost constraints that would argue against serving maps publicly?** For example the Vercel route/cache-header limits the build already works around — serving the `.map` files means they now need to be included in the deployed assets (and likely the cache-header batching that currently filters out `*.js.map`).
   _Awaiting answer._
3. **Should this also cover the dotcom workers, or just the client bundle?** This report is scoped to the client app; the workers may have their own source-map handling.
   _Deferred by user; not blocking implementation._

jsscclr:
In order to make deployed tldraw.com builds debuggable in browser devtools and stop the source-map 404s, this PR stops stripping `.js.map` files from the client deploy and serves them publicly. Closes #9322.

The client build generated source maps, uploaded them to Sentry, then deleted the `.js.map` files from the deployed assets — while leaving the `//# sourceMappingURL` comments in the shipped JS. That produced a 404 for every map and left browser devtools unable to symbolicate deployed builds. The usual reason to hide source maps is to keep source private, which does not apply here since the client source is public in this repo. Sentry upload is unaffected: it reads from `dist/assets` before the strip step.

### Change type

- [x] `bugfix`

### Test plan

1. From the preview deploy, open devtools → Network, reload, and confirm `*.js.map` requests return 200 rather than 404.
2. Open a minified stack frame in devtools and confirm it symbolicates to real names and lines.

### Code changes

| Config/tooling | +7 / -3    |

### DIFF
--- a/apps/dotcom/client/scripts/build.ts
+++ b/apps/dotcom/client/scripts/build.ts
@@ -1,7 +1,6 @@
 import { mkdirSync, readFileSync, readdirSync, writeFileSync } from 'fs'
 import { T } from '@tldraw/validate'
 import { config } from 'dotenv'
-import { glob } from 'fast-glob'
 import json5 from 'json5'
 import regexgen from 'regexgen'
 import { exec } from '../../../../internal/scripts/lib/exec'
@@ -53,7 +52,10 @@ async function build() {
 	await exec('rm', ['-rf', '.vercel/output'])
 	mkdirSync('.vercel/output', { recursive: true })
 	await exec('cp', ['-r', 'dist', '.vercel/output/static'])
-	await exec('rm', ['-rf', ...glob.sync('.vercel/output/static/**/*.js.map')])
+	// We serve the .js.map files publicly. The client source is open at tldraw/tldraw, so
+	// there's nothing to hide, and serving the maps lets anyone debugging a deployed build get
+	// real names and lines in devtools without going through Sentry. Sentry still gets its own
+	// copy via the upload step, which reads from dist/assets before this point.
 
 	// Add fonts to preload into index.html
 	const assetsList = readdirSync('dist/assets')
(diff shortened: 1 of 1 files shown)


frolic (approved):
lgtm, but I'd want @mimecuvalo or someone to double check

mimecuvalo (approved):
great shout

PR #9328, opened 2026-06-19
fix(dotcom): guard against group_file rows with no synced file

jsscclr:
### The problem

The workspace sidebar can crash the whole app on the proper-Zero backend:

```
TypeError: Cannot read properties of undefined (reading 'owningGroupId')
  at getWorkspaceFilesSorted (TldrawApp.ts)
```

`getWorkspaceFilesSorted` reads `f.file.owningGroupId` for every `group_file` row, assuming the `file` is always there. The `workspaceMemberships` query joins `file` as a Zero `.one()` relationship, which resolves to `undefined` when the referenced `file` row isn't in the client's synced data. A `group_file` row can be present without its `file` when:

- the file was deleted, leaving a dangling join row,
- the join row synced before its file row, or
- the file is filtered out server-side because the user can no longer read it.

Any of these throws inside the sidebar's reactive derive and takes down the component tree. The declared type `groupFiles: Array<TlaGroupFile & { file: TlaFile }>` asserted `file` was always present, so the compiler never flagged the access.

This only affects the proper-Zero path. The legacy `ZeroPolyfill` already drops orphan rows at the source (`compact()` + `if (!file) return null`), so the guard was effectively lost when proper Zero took over the same shape. Introduced in #9270.

### The change

- Make the type honest: `file: TlaFile | undefined`.
- Drop orphan rows in `getWorkspaceFilesSorted` via a type-guard predicate, so the non-undefined narrowing flows into `pinned`/`unpinned` and the rest of the function with no scattered `!` assertions.
- Guard the `canCreateNewFile` file-limit count the same way.

### Testing

Added `getWorkspaceFilesSorted.test.ts`, which drives the real method against a membership whose first row is an orphan (`file: undefined`). Without the guard it reproduces the exact production crash; with it, the orphan is excluded and the owned file is returned. `yarn typecheck` passes (the honest type surfaced two more access sites, fixed by the type guard).

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -126,7 +126,7 @@ export class TldrawApp {
 	private readonly workspaceMemberships$: Signal<
 		(TlaGroupUser & {
 			group: TlaGroup
-			groupFiles: Array<TlaGroupFile & { file: TlaFile }>
+			groupFiles: Array<TlaGroupFile & { file: TlaFile | undefined }>
 			groupMembers: Array<TlaGroupUser>
 		})[]
 	>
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
👍

apps/dotcom/client/src/tla/app/TldrawApp.ts
```diff
@@ -126,7 +126,7 @@ export class TldrawApp {
 	private readonly workspaceMemberships$: Signal<
 		(TlaGroupUser & {
 			group: TlaGroup
-			groupFiles: Array<TlaGroupFile & { file: TlaFile }>
+			groupFiles: Array<TlaGroupFile & { file: TlaFile | undefined }>
 			groupMembers: Array<TlaGroupUser>
 		})[]
```
frolic, line 131: Is there not some query type or something somewhere that we can lift out instead of defining this here? Feels like we're having to keep this aligned with a query result type that probably lives somewhere?

frolic (approved):
down to land this as a fix but feels like this is a smell and that we should be able get the real result type from the query itself?

apps/dotcom/client/src/tla/app/TldrawApp.ts
jsscclr: agreed! will look into next

PR #9331, opened 2026-06-19
refactor(dotcom): derive membership signal types from the Zero queries

jsscclr:
Follow-up to #9328, addressing @frolic's review: *"feels like this is a smell and that we should be able get the real result type from the query itself."*

### The smell

`fileStates$` and `workspaceMemberships$` had **hand-written** type annotations that mirrored the shape of their Zero queries (`packages/dotcom-shared/src/queries.ts`). That mirror is a second source of truth, and it had already drifted from reality: it declared `.one()` joins as always-present, which is exactly what hid the production crash fixed in #9328 (a `group_file` row whose `file` join resolved to `undefined`).

### The change

Derive both signal types straight from the query definitions:

```ts
private readonly fileStates$: Signal<QueryResultType<typeof queries.fileStates>>
private readonly workspaceMemberships$: Signal<QueryResultType<typeof queries.workspaceMemberships>>
```

Now the shape — and the nullability of every `.one()` join — stays in sync with the queries automatically. This whole class of bug becomes a compile error instead of a runtime crash.

### What it caught

Making the types honest surfaced **9 latent nullability errors across 3 files** — and none were about `file`. They were all the `group` join (also a `.one()`), which the hand-written type also falsely claimed was always present. So the original crash wasn't a one-off; the hand-written mirror was concealing the same bug for a second relationship.

Fixes per site:
- Drop group-less rows from the two workspace lists (workspace switcher, "move to" menu) via a type-guard filter — same pattern as the orphan-`file` fix, so downstream `.map` accesses narrow to non-null.
- Optional-chain the display reads (workspace names, invite secret).
- Remove three type imports (`TlaFileState`, `TlaGroup`, `TlaGroupUser`) that existed only to prop up the hand-written shapes.

### Testing

- `yarn typecheck` passes (the derivation resolves; the `getWorkspaceFilesSorted` type-guard from #9328 still composes with the derived element type).
- The `getWorkspaceFilesSorted` regression test from #9328 still passes.
- No behavior change for the common case; the only runtime difference is that a membership with an unsynced `group` is omitted from the workspace lists rather than crashing.

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -1,4 +1,4 @@
-import { Zero } from '@rocicorp/zero'
+import { QueryResultType, Zero } from '@rocicorp/zero'
 import { captureException } from '@sentry/react'
 import {
 	AcceptInviteResponseBody,
@@ -10,12 +10,9 @@ import {
 	ROOM_PREFIX,
 	TlaFile,
 	WELCOME_CREATE_SOURCE,
-	TlaFileState,
 	TlaFileStatePartial,
 	TlaFlags,
-	TlaGroup,
 	TlaGroupFile,
-	TlaGroupUser,
 	TlaMutators,
 	TlaSchema,
 	TlaUser,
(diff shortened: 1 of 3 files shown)


frolic (approved):
lovely!

are there other areas where it makes sense to migrate to this pattern too? (doesn't need to block this PR)

steveruizok:
Nice!

PR #9338, opened 2026-06-19
fix(dotcom): normalize create workspace dialog title and button

frolic:
In order to keep the workspace dialogs consistent, this PR normalizes the create workspace dialog so its title is a descriptive action ("Create workspace") and its confirm button is a bare verb ("Create"). Previously the title had a stray article ("Create a workspace") and the button repeated the noun ("Create workspace"). Every other workspace dialog already follows this convention (Manage workspace, Leave workspace / Leave, Delete workspace / Delete, Regenerate invite link / Regenerate, Remove member / Remove).

The i18n source and compiled message catalogs were regenerated with `yarn workspace dotcom build-i18n` (the message IDs are content-hashed, so changing the copy shifts them), and the e2e fixture and user manual were updated to match the new button label.

### Change type

- [x] `improvement`

### Test plan

1. Open the workspace switcher in the sidebar and choose to create a workspace.
2. Confirm the dialog title reads "Create workspace" and the primary button reads "Create".

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Normalize the create workspace dialog title and button to match the other workspace dialogs.

### Code changes

| Tests           | +1 / -1    |
| Automated files | +9 / -9    |
| Apps            | +3 / -3    |

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/Sidebar.ts
+++ b/apps/dotcom/client/e2e/fixtures/Sidebar.ts
@@ -327,7 +327,7 @@ export class Sidebar {
 		await expect(input).toBeVisible()
 		await input.fill(name)
 
-		await this.page.getByRole('button', { name: 'Create workspace' }).click()
+		await this.page.getByRole('button', { name: 'Create', exact: true }).click()
 		await this.mutationResolution()
 
 		// Creating a workspace switches to it and opens its seeded welcome file. That file
--- a/apps/dotcom/client/public/tla/locales-compiled/en.json
+++ b/apps/dotcom/client/public/tla/locales-compiled/en.json
@@ -425,12 +425,6 @@
       "value": "PNG"
     }
   ],
-  "5638176be5": [
-    {
-      "type": 0,
-      "value": "Create a workspace"
-    }
-  ],
   "5677e381d2": [
     {
       "type": 0,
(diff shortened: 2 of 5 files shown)


jsscclr (approved):
I think this is a good consolidation - but I do think we need to also do a separate accessibility audit since any button that just says `"Create"` will not provide enough context to screen reader users ("Create what"?).

I'm going to file an issue now on this, but think this is a good step for now!

frolic:
ah good call, maybe all these buttons get an `aria-label` with a more descriptive action?

steveruizok:
> I think this is a good consolidation - but I do think we need to also do a separate accessibility audit since any button that just says `"Create"` will not provide enough context to screen reader users ("Create what"?).
> 
> I'm going to file an issue now on this, but think this is a good step for now!

It should be `Create a workspace` and `Create workspace`. We're following Figma's copy here.

steveruizok:
I don't think we should land this. I would actually go the other way with it and add `Leave workspace` (etc) as the confirmation language.

image

image

image

image

image

Not perfectly consistent (see the inconsistent `?` in the titles) but its a good example of a strong pattern I've seen elsewhere, too.

image

image

image

image

image

image

frolic:
> I don't think we should land this. I would actually go the other way with it and add `Leave workspace` (etc) as the confirmation language.

That's fine too, I just want consistency here as we're doing inconsistent things right now. It felt to me that in smaller dialogs where the title and button were fairly close together, the copy was a bit duplicative, but I am fine with either as long as we pick a direction and stick with it.

PR #9341, opened 2026-06-19
fix(ci): include runner arch in node_modules cache key

frolic:
In order to stop x64 e2e jobs from restoring an ARM-built `node_modules`, this PR adds `${{ runner.arch }}` to the shared setup action's cache key.

The `node_modules` cache in `.github/actions/setup/action.yml` was keyed on `${{ runner.os }}` only, which resolves to `Linux` on both the ARM runners (`Checks`, `Deploy .com` — `ubuntu-latest-*-arm-open`) and the x64 runners (the e2e suites). The two architectures therefore shared a single cache.

This is a latent bug that surfaced after the Node 24.17.0 bump (#9311). That bump changed the cache key, invalidating every cache at once. In the fresh fill-race the fast 16-core ARM `Checks/build` job won and saved an ARM-built `node_modules` under the arch-blind key. From then on, x64 e2e jobs restored it; because install-state is marked `built`, `yarn install --immutable` skips the native rebuild, so the x64 `@rocicorp/zero-sqlite3` binary (`build/Release/better_sqlite3.node`) is absent and `zero-cache` crashes with `ERR_DLOPEN_FAILED`. Only the **dotcom** e2e suite boots `zero-cache`, so it was the only suite that failed — examples and perf (also x64) never load that binary and stayed green.

The bump commit's own dotcom run passed because it had a cache *miss* and compiled its own x64 binary; every run after it got a cache *hit* on the poisoned ARM cache and failed.

Adding the architecture to the key keeps ARM and x64 caches separate, and as a side effect busts the currently-poisoned cache so the next run rebuilds cleanly.

While here, the second commit derives the Node version segment of the key from `${{ steps.setup-node.outputs.node-version }}` rather than repeating the literal `24.17.0`, so the version is pinned in a single place (the `node-version` input). Future Node bumps now also bust the cache automatically — that manual key edit was a step #9311 had to do by hand.

### Change type

- [x] `other`

### Test plan

- CI on this PR runs the dotcom e2e suite (x64) and should now restore an x64 `node_modules` with a working native binary. Green dotcom e2e confirms the fix.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Config/tooling | +3 / -2    |


kaneel (approved):

mimecuvalo (commented):
nice catch

PR #9343, opened 2026-06-19
perf(dotcom): cache auth tokens for file navigation

steveruizok:
🙇‍♂️ I noticed that every navigation was paying 150ms for an auth round trip to Clerk. This PR is an attempt to prevent that by holding on to the auth token.

---

In order to avoid making signed-in file navigation wait on repeated Clerk token retrieval, this PR adds a shared auth token cache for the dotcom app and uses it for file sync and asset token access. It also normalizes WebSocket `accessToken` query params before worker-side Clerk authentication so the common file-join path does one authenticate pass, with a cookie fallback only when the query-token attempt fails.

### Change type

- [x] `improvement`

### Test plan

1. `yarn workspace dotcom lint`
2. `yarn workspace @tldraw/dotcom-worker lint`
3. `yarn typecheck`

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve signed-in file navigation by reusing recently refreshed auth tokens.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps | +117 / -28 |

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -73,6 +73,7 @@ import { MULTIPLAYER_SERVER, ZERO_SERVER } from '../../utils/config'
 import { multiplayerAssetStore } from '../../utils/multiplayerAssetStore'
 import { getScratchPersistenceKey } from '../../utils/scratch-persistence-key'
 import { TLAppUiContextType } from '../utils/app-ui-events'
+import type { TlaGetToken } from '../utils/authTokenCache'
 import { getDateFormat } from '../utils/dates'
 import { FeatureFlags } from '../utils/FeatureFlagPoller'
 import { createIntl, defineMessages, setupCreateIntl } from '../utils/i18n'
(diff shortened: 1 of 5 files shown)


MitjaBezensek:
My understanding is that Clerk already does much of this behind the scenes. Clerk caches and [auto refreshes](https://clerk.com/docs/guides/how-clerk-works/overview#token-refresh-mechanism) the tokens every 50s behind the scenes. I confirmed this by:
* keeping a tab open and seeing a fetch go out to clerk about every 50s, even without navigation
* quickly navigating between files does not trigger this, i.e. you won't get a separate fetch to clerk for each file navigation.

One thing we could do to reduce the frequent refreshing of the tokens is to make them longer lived. We'd need to use a [JWT template](https://clerk.com/docs/guides/sessions/jwt-templates#create-a-jwt-template) as this is not a simple config. That said I'm not sure if we should go this route as it does make things less secure.

Preferring the query param token in getAuth so we skip an extra Clerk verification on connect is worth keeping though!

steveruizok:
Ok, closing this one. We can come back to the JWT if we see it showing up more.

PR #9348, opened 2026-06-19
feat(dotcom): enroll new users in groups frontend by default

steveruizok:
In order to enable the groups UI for everyone, this PR makes new users enrolled in the groups frontend by default. New users were already created with the `groups_backend` flag; they now also get `groups_frontend`, so the groups UI (gated by `useHasFlag('groups_frontend')`) is on for every newly created user.

This only affects newly created users — existing users keep their stored flags.

### Change type

- [x] `feature`

### Test plan

1. Create a new user account.
2. Confirm the groups UI (workspace switcher / groups in the sidebar) is available without manual admin enrollment.

### Release notes

- Enroll new users in the groups frontend by default.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +1 / -1    |
| Core code | +1 / -1  |

### DIFF
--- a/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLUserDurableObject.ts
@@ -146,7 +146,7 @@ export class TLUserDurableObject extends DurableObject<Environment> {
 								exportPadding: true,
 								createdAt: now,
 								updatedAt: now,
-								flags: 'groups_backend',
+								flags: 'groups_backend,groups_frontend',
 							})
 							.execute()
 						await tx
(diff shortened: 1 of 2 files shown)


steveruizok:
...however, when I sign in with a new account I already get into the groups UI. Are we ignoring these feature flags now? Should they be removed?

MitjaBezensek:
> ...however, when I sign in with a new account I already get into the groups UI. Are we ignoring these feature flags now? Should they be removed?

I created a [similar PR](https://github.com/tldraw/tldraw/pull/9137) here and mentioned that we need to do this in the [rollout task](https://app.notion.com/p/tldraw/Prepare-the-rollout-37b3e4c324c080e79f02eb5feeac6c2b?source=copy_link#37d3e4c324c080ba8ff1e43c092c3cec) I prepared. Unsure about the current state and why my PR did not get merged when doing the rollout. Would be worth checking with (probably) @mimecuvalo.

It does seem like not all accounts are rolled in, 273 are still missing the flag, so I think we still need to do this. The number seems to be increasing as I checked it over a few mins, went 270 -> 273.
CleanShot 2026-06-20 at 08 49 30@2x

mimecuvalo (approved):
👍

frolic:
when do we consider workspaces are here to stay and remove the flag and old sidebar treatment?

PR #9350, opened 2026-06-19
perf(scripts): upload CDN assets in parallel during publish

steveruizok:
In order to speed up the SDK publish workflow, this PR uploads static assets to the CDN with bounded concurrency instead of one file at a time.

Previously `uploadStaticAssets` walked the assets tree and `await`ed each R2 `PutObjectCommand` sequentially, so every upload waited for the previous one to finish even though they're fully independent. With many asset files this dominated the publish step. The upload now collects every file up front and processes them through a small worker pool (16 concurrent uploads), so wall time drops to roughly the previous time divided by the concurrency limit while staying well-behaved against R2.

### Change type

- [x] `improvement`

### Test plan

This runs only during the package publish workflow (`internal/scripts/lib/upload-static-assets.ts`) and is exercised the next time SDK packages are published. No behavior changes beyond upload ordering and concurrency.

### Code changes

| Config/tooling | +33 / -11  |

### DIFF
--- a/internal/scripts/lib/upload-static-assets.ts
+++ b/internal/scripts/lib/upload-static-assets.ts
@@ -11,6 +11,9 @@ const R2_URL = 'https://c34edc4e76350954b63adebde86d5eb1.r2.cloudflarestorage.co
 const R2_BUCKET = 'cdn'
 const ASSETS_FOLDER = './packages/assets'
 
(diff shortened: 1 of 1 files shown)


steveruizok:
No way to test this but to land it and see how canaries go.

PR #9354, opened 2026-06-19
test(examples): stabilise flaky rich-text and a11y panel e2e tests

steveruizok:
In order to stop several examples e2e tests from flaking the suite (and ejecting unrelated PRs from the merge queue), this PR makes their editing- and focus-readiness waits deterministic. The export-pipeline font-load race that surfaces as the `Export snapshots › Text rendering` flake is handled separately in #9332, so it has been dropped from here — this PR is test-only.

> Note: the export-snapshot flake lives on `main`, so it can still eject this PR from the merge queue until #9332 lands. This PR removes the a11y and rich-text flakes that are this branch's responsibility.

### What changed

- **`test-rich-text`** — the rich text editor focuses on a deferred timeout after mounting (an existing workaround in `RichTextArea`), so waiting for the contenteditable to be *visible* before typing dropped the first keystroke (received `"d like…"` instead of `"id like…"`). Now waits for the contenteditable to be the active element — matching the pattern already used in `test-focus` — before typing, and awaits the `toHaveText` assertions that were previously fire-and-forget.
- **`test-a11y-style-panel-navigation`** — the spec checked focus (`document.activeElement`) synchronously right after keyboard presses, but focus transitions settle asynchronously, so the checks raced and intermittently failed. Three distinct causes are addressed:
  - Synchronous focus/editing reads after a keypress now use `expect.poll` so they wait for the transition to settle.
  - Returning focus to the canvas relied on a single Escape keypress, which can be dropped under CI load (the Escape handler in `DefaultStylePanel` is synchronous, so a timed-out poll means the key was lost, not slow). A `pressUntilFocusedOnCanvas` helper re-sends Escape until focus settles — and only while focus is still inside the panel, where the handler's `stopPropagation()` means a repeat can never deselect the shape.
  - Selecting a colour re-renders the style panel, and its roving-tabindex toolbar bounces focus back onto the selected radio *asynchronously*, after Escape already returned focus to the canvas — so the follow-up Enter landed on the radio and editing never started. A `startEditingSelectedShapeViaKeyboard` helper drives editing through whatever focus state it lands in (Escape if focus bounced into the panel, otherwise Enter), and the re-typed label now waits for the rich-text editor to be focused so leading keystrokes aren't dropped.
  - The `cmd+enter returns focus to the canvas` test, which was marked `test.fail` and flaked via intermittent passes, is skipped until its event dispatch can be made deterministic.

### Change type

- [x] `other`

### Test plan

1. Run `cd apps/examples && yarn e2e -g "Double click from select tool" --project=chromium` repeatedly; the first keystroke is no longer dropped.
2. Run `cd apps/examples && yarn e2e -g "A11y Style Panel Navigation" --repeat-each=10 --project=chromium`; focus transitions are driven to a settled state instead of racing. Verified locally: the full describe block passes 10x, and the long navigation test passes 25x on chromium.

- [ ] Unit tests
- [x] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Tests   | +148 / -69 |

### DIFF
--- a/apps/examples/e2e/tests/test-a11y-style-panel-navigation.spec.ts
+++ b/apps/examples/e2e/tests/test-a11y-style-panel-navigation.spec.ts
@@ -102,10 +102,12 @@ test.describe('A11y Style Panel Navigation', () => {
 		expect(finalText).toBe('Updated Label')
 	})
 
-	// Failing! This is failing because the keyboard event isn't being picked up while editing the rich
-	// text. In practice the behavior does work, so this is a problem with the test or the way that we're
-	// dispatching the event.
-	test.fail('in a geo shape, cmd+enter returns focus to the canvas', async ({ page, isMobile }) => {
+	// Skipped: the cmd+enter keyboard event isn't reliably picked up while editing rich text in the
+	// test harness, so the body passes intermittently. Because it was marked `test.fail`, those
+	// intermittent passes were reported as failures and flaked the suite. The behavior works in
+	// practice; re-enable once the event dispatch can be made deterministic (a reliable way to send
+	// the shortcut to the focused contenteditable).
+	test.skip('in a geo shape, cmd+enter returns focus to the canvas', async ({ page, isMobile }) => {
 		if (isMobile) {
 			test.skip()
 		}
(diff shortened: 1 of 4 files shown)


steveruizok:
Noticed overlap with https://github.com/tldraw/tldraw/pull/9332, I think we're chasing the same thing here?

PR #9356, opened 2026-06-21
fix(sync): drop malformed websocket messages

### LINKED ISSUE #9119: Sync client crashes on malformed WebSocket message JSON
The sync client's WebSocket message handler crashes with an unhandled `SyntaxError` when a message arrives that isn't valid JSON (e.g. truncated mid-transfer). The parse happens without a try/catch, so one bad frame takes down the client instead of being dropped or triggering a reconnect. Mostly seen on mobile Safari, which suggests flaky connections truncating frames.

Note: this Sentry issue is currently archived-until-escalating, so it won't show in the default unresolved view, but it's still occurring (last event yesterday).

## Sentry context

- Sentry: [LITE-483](https://tldraw.sentry.io/issues/LITE-483) — **764 events / 15 users**, since 2024-08-12, status: archived until escalating
- Unhandled; culprit `ClientWebSocketAdapter.onmessage` (`packages/sync-core/src/lib/ClientWebSocketAdapter.ts:215`)
- Sample event: Mobile Safari 26.5 / iOS 18.7

```ts
const parsed = JSON.parse(ev.data.toString())  // ← throws on malformed frame
this.messageListeners.forEach((cb) => cb(parsed))
```

steveruizok:
In order to prevent malformed WebSocket frames from taking down sync clients, this PR catches JSON parse failures in `ClientWebSocketAdapter` and drops the bad message. Valid later messages continue to be delivered normally. Closes #9119.

### Change type

- [x] `bugfix`

### Test plan

1. Send malformed JSON to a `ClientWebSocketAdapter` socket.
2. Confirm the adapter stays online and does not notify message listeners.
3. Send a valid JSON message afterwards and confirm it is delivered.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix sync clients crashing when a malformed WebSocket message is received.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +7 / -1    |
| Tests     | +19 / -0   |

### DIFF
--- a/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
+++ b/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
@@ -410,6 +410,25 @@ describe('ClientWebSocketAdapter', () => {
 			expect(onMessage).toHaveBeenCalledWith({ type: 'message', data: 'hello' })
 		})
 
+		it('[CW7] drops malformed JSON messages without closing the socket', async () => {
+			const warnOnceMock = vi.mocked(warnOnce)
+			const onMessage = vi.fn()
+			adapter.onReceiveMessage(onMessage)
+
+			await waitFor(() => adapter._ws?.readyState === WebSocket.OPEN)
+			warnOnceMock.mockClear()
+
+			connectedServerSocket.send('{ "type": "message",')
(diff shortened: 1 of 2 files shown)


mimecuvalo:
going to do a followup PR for this

PR #9361, opened 2026-06-21
perf(editor): avoid repeated useTransform layout effects

### LINKED ISSUE #8234: useTransform hook runs useLayoutEffect on every render

steveruizok:
In order to avoid applying DOM transforms after every render, this PR adds a dependency array to `useTransform` based on the actual transform values. The optional offset is split into scalar dependencies so fresh offset object literals do not force the effect to rerun unnecessarily. Closes #8234.

### Change type

- [x] `improvement`

### Test plan

1. Render a component using `useTransform`.
2. Confirm the transform style is still applied from `x`, `y`, `scale`, `rotate`, and `additionalOffset`.
3. Confirm the effect only reruns when those values change.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Improve `useTransform` so it only applies styles when transform values change.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +6 / -3    |

### DIFF
--- a/packages/editor/src/lib/hooks/useTransform.ts
+++ b/packages/editor/src/lib/hooks/useTransform.ts
@@ -10,6 +10,9 @@ export function useTransform(
 	rotate?: number,
 	additionalOffset?: VecLike
 ) {
+	const additionalOffsetX = additionalOffset?.x
+	const additionalOffsetY = additionalOffset?.y
+
 	useLayoutEffect(() => {
 		const elm = ref.current
 		if (!elm) return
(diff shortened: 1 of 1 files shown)


steveruizok:
Closing as superseded by #9362, which removes the obsolete SelectionBackground component path and deletes useTransform entirely.

PR #9374, opened 2026-06-22
fix(tldraw): improve rotate hit area

steveruizok:
In order to make the selection foreground overlay easier to maintain and the corner rotate handles easier to grab, this PR renames the opaque `SelectionState` fields and squares off the rotate-handle hit area.

The `SelectionState` fields `size`, `targetSize`, `targetSizeX`, `targetSizeY`, `expandDx`, and `expandDy` are renamed to the more descriptive `handleSize`, `hitTargetSize`, `hitTargetSizeX`, `hitTargetSizeY`, `expandOffsetX`, and `expandOffsetY`, with doc comments explaining what each one represents. There is no behavior change from the rename.

The corner rotate handle's hit area is also changed from a circle (`Circle2d`) to a square (`Rectangle2d`) and its half-size multiplier is widened from `1.5` to `1.75`, so the diagonal rotate targets at each corner are a little easier to hit.

### Change type

- [x] `improvement`

### Test plan

1. Select a shape and hover near each corner just outside the bounds; confirm the rotate cursor appears across the squared hit area.
2. Resize and crop small and large selections; confirm handles render and hit-test as before.

- [x] Unit tests

### Release notes

- Make the selection corner rotate handles easier to grab.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +86 / -71  |
| Tests     | +1 / -1    |

### DIFF
--- a/packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
+++ b/packages/tldraw/src/lib/overlays/SelectionForegroundOverlayUtil.ts
@@ -1,11 +1,11 @@
 import {
 	Box,
-	Circle2d,
 	Geometry2d,
 	HALF_PI,
 	Mat,
 	OverlayUtil,
 	Polygon2d,
+	Rectangle2d,
 	RotateCorner,
 	SelectionCorner,
 	SelectionEdge,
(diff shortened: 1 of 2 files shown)


mimecuvalo:
ok 👍

PR #9376, opened 2026-06-22
fix(tldraw): prevent crash when a pointed-at shape is deleted mid-click

### LINKED ISSUE #8558: PointingShape crashes when shape is deleted during pointer interaction
Getting a bug that's crashing the whiteboard:

`Cannot read properties of undefined (reading 'id')`

Image

## Steps to reproduce

1. Click on a shape (enters `PointingShape` state)
2. Shape gets deleted before pointer up (e.g., remote collaborator deletes it, or undo)
3. Release mouse — crash

## Root cause

`getOutermostSelectableShape` in `Editor.ts` uses a non-null assertion (`this.getShape(id)!`) but `getShape` returns `undefined` for deleted shapes. The function returns `undefined` despite its `TLShape` return type. `PointingShape.onPointerUp` then accesses `.id` on the undefined result.

## Stack trace

```
childStates/PointingShape.ts:123 — outermostSelectableShape.id
tools/StateNode.ts:182 — handleEvent
tools/StateNode.ts:188 — enter
lib/editor/Editor.ts:10656
lib/editor/Editor.ts:10170
lib/transactions.ts:304
lib/transactions.ts:358
HistoryManager/HistoryManager.ts:100
lib/editor/Editor.ts:1213
```

## Affected code

- `packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts` — multiple `.id` accesses without null checks
- `packages/editor/src/lib/editor/Editor.ts:6464` — `getOutermostSelectableShape` uses `!` non-null assertion on `getShape` result

kaneel:
When a shape is deleted between pointer down and pointer up (e.g. by a remote user or an undo), the select tool dereferenced the now-missing shape and threw "Cannot read properties of undefined (reading 'id')" in PointingShape.onPointerUp.

getOutermostSelectableShape used getShape(id)! and could return undefined despite its TLShape return type. It now returns TLShape | undefined, and PointingShape bails to idle when the shape no longer exists. Updated the remaining callers and the API report to match.

Fixes #8558

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

This one requires a friend. 

- Click on a shape (enters PointingShape state)
- Shape gets deleted before pointer up (e.g., remote collaborator deletes it, or undo)
- Release mouse — crash

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix a crash when a shape is deleted mid-click (e.g. by a remote user or an undo) while clicking on it.

### API changes

- `Editor.getOutermostSelectableShape()` now returns `TLShape | undefined`; it could already return `undefined` for a missing shape despite its old `TLShape` type.

### DIFF
--- a/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
@@ -75,6 +75,14 @@ export class PointingShape extends StateNode {
 				renderingOnly: true,
 			}) ?? this.hitShape
 
+		// The hit shape may have been deleted between pointer down and pointer up
+		// (e.g. by a remote user or an undo), in which case the fallback hitShape
+		// is stale. Bail to idle so we don't dereference a missing shape.
+		if (!this.editor.getShape(hitShape.id)) {
+			this.parent.transition('idle', info)
+			return
+		}
+
 		const selectingShape = hitShape
 			? this.editor.getOutermostSelectableShape(hitShape)
 			: this.hitShapeForPointerUp
--- a/packages/tldraw/src/test/SelectTool.test.ts
+++ b/packages/tldraw/src/test/SelectTool.test.ts
@@ -122,6 +122,44 @@ describe.skip('Edit on type', () => {
 	})
 })
 
+describe('TLSelectTool.PointingShape when the shape is deleted mid-click', () => {
+	// Reproduces https://github.com/tldraw/tldraw/issues/8558: a remote user,
+	// undo, or other actor can delete the pointed-at shape between pointer down
+	// and pointer up. The tool should bail to idle instead of crashing.
+
+	it('does not crash on pointer up when an already-selected shape is deleted', () => {
+		// pre-selecting the shape means didSelectOnEnter is false, so pointer up
+		// runs the selection logic that dereferences the (now deleted) shape
+		editor.select(ids.box1)
+		const shape = editor.getShape(ids.box1)!
+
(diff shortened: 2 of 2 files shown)


steveruizok (approved):

steveruizok:
This looks good btw!

kaneel:
> This looks good btw!

thanks but I decided to go against the first solution because of the surface it covered and I'd rather manually check if a shape exists before accessing it… rather than having all methods of the SDK returns "shape | undefined"

packages/tldraw/src/lib/tools/SelectTool/childStates/PointingShape.ts
```diff
@@ -75,6 +75,14 @@ export class PointingShape extends StateNode {
 				renderingOnly: true,
 			}) ?? this.hitShape
 
+		// The hit shape may have been deleted between pointer down and pointer up
+		// (e.g. by a remote user or an undo), in which case the fallback hitShape
+		// is stale. Bail to idle so we don't dereference a missing shape.
+		if (!this.editor.getShape(hitShape.id)) {
+			this.parent.transition('idle', info)
```
steveruizok, line 82:
```suggestion
			this.cancel()
```

Often the tools use a shared `cancel` to make sure that we always clean up in the same way. This would fit better with the rest of the tool design!

kaneel: good to know, thanks for sharing!

PR #9379, opened 2026-06-22
fix(dotcom): pressing outside an open menu only dismisses it

frolic:
In order to make clicking outside an open menu behave consistently on tldraw.com — only dismissing the menu, never also acting on whatever is underneath — this PR makes every "outside press" dismiss-only across the app, without breaking click-to-draw on the canvas.

Two cooperating pieces, both keyed only off the `tlmenus` registry (so they're independent of Radix and survive the planned Base UI refactor):

- **Canvas** — the SDK's existing `MenuClickCapture` overlay already dismisses on a click and forwards a drag to the canvas (so you can dismiss and start drawing in one gesture). Untouched.
- **Everything else (chrome)** — a new `useMenuClickCapture()` hook installs capture-phase `document` listeners once at the app root (`TlaRootProviders`). While a menu is open it inspects the element under each press (`event.target`): a press on the canvas is left to the SDK overlay; a press inside an open menu/dialog/popover/select is left alone; anything else (sidebar, editor top panels, buttons) is swallowed and only dismisses (its follow-up `click` is cancelled too).

It's a listener rather than an overlay because no single app-level z-index sits above all chrome yet below all menus (the editor nests its panels and menus in one stacking context, so an overlay high enough to cover the top bar also covers the open menu). Classifying by the real hit target also makes touch behave exactly like mouse with no cursor tracking, and mounting at the app root covers editor pages that have top-bar menus but no sidebar (published files, snapshots, embeds).

`apps/dotcom/client/src/tla/DISMISSAL_INVARIANTS.md` documents the full behaviour contract — per surface, click vs click-drag, modality, and the z-index model — as a source of truth for the upcoming Base UI refactor. The SDK is untouched.

### Change type

- [x] `bugfix`

### Test plan

With the workspaces UI enabled and a file open, open any menu (the workspace switcher, the top-left `…` menu, a file's `…` menu, the share menu), then:

1. Click a sidebar file → the menu dismisses; you do **not** navigate.
2. Click another menu's trigger (sidebar or top bar) → the menu dismisses; the second menu does **not** open (press again to open it).
3. Click a top-bar button (e.g. Share) → the menu dismisses; the button does **not** fire.
4. Click the canvas → the menu dismisses; **no** shape is created.
5. Click-drag the canvas → the menu dismisses **and** a draw starts from the press location.
6. Open a select inside the share menu, then press the canvas → the whole stack dismisses, no shape drawn.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- On tldraw.com, clicking outside an open menu now only dismisses the menu, instead of also navigating to a file, opening another menu, or activating a button underneath. Clicking and dragging on the canvas still dismisses and starts the drag.

### Code changes

| Tests         | +126 / -4  |
| Apps          | +107 / -19 |
| Documentation | +216 / -0  |


apps/dotcom/client/src/tla/hooks/useMenuClickCapture.ts
```diff
@@ -0,0 +1,77 @@
+import { useEffect } from 'react'
+import { tlmenus } from 'tldraw'
+
+// A press that lands on the canvas is left alone: the SDK's own `MenuClickCapture` covers the canvas
+// while a menu is open and handles dismiss + drag-forward (so click-to-draw keeps working). A press
+// inside an open dismissable must keep working too. Everything else is chrome.
+const CANVAS_SELECTOR = '.tl-canvas, .tlui-menu-click-capture'
+const DISMISSABLE_SELECTOR =
+	'[data-radix-popper-content-wrapper], [role="menu"], [role="dialog"], [role="listbox"]'
```
frolic: why do we need `[data-radix-popper-content-wrapper]`? wouldn't that be captured as a child of one of the other `role` selectors?

steveruizok:
Is this one blocked? Worth landing or waiting on baseui?

mimecuvalo:
> Is this one blocked? Worth landing or waiting on baseui?

waiting for Base UI to maybe progress

frolic:
this was really hairy and hard to get right so I kinda moved on to other things 🙈

mimecuvalo:
closing up some PRs that are eating up some CF usage 😬 , reopen if needed later (and rebase with `main`)

PR #9380, opened 2026-06-22
fix(sync): restart connection on malformed websocket message

mimecuvalo:
Follow-up to #9356. That PR fixed the crash from malformed WebSocket frames (#9119) by catching the `JSON.parse` failure and dropping the message. This tightens the recovery so a dropped message can't silently desync the client.

### The problem with dropping

Server→client messages are not chunked (`ServerSocketAdapter.sendMessage` sends whole JSON), so a malformed frame is genuinely corrupt data, not a chunk fragment. Most message types recover on their own if dropped:

- `push_result` — the next one fails the `clientClock` ordering check in `rebase`, which resets the connection.
- `connect` response — `isConnectedToRoom` never becomes true, so the client stays offline and never applies server data. It can't silently desync; at worst it's visibly stuck offline until the next connection attempt.

But a dropped `patch` does not. The patch path applies network diffs incrementally and does not check `serverClock` for gaps, so a missed partial patch is applied as if it never happened. The result is a silently stale field that only heals on the next reconnect.

### The fix

On a malformed message, restart the connection instead of dropping it. Reconnecting sends a `connect` with the last known `serverClock`, and the server responds with a re-hydration diff that brings the client fully back in sync. The crash is still gone, and these frames are rare and indicate a flaky connection anyway (the original Sentry issue was mostly mobile Safari), so the cost of a reconnect round-trip is negligible.

### Change type

- [x] `bugfix`

### Test plan

1. Send a malformed JSON frame to a `ClientWebSocketAdapter` socket.
2. Confirm the message is not forwarded to listeners.
3. Confirm the adapter restarts: it goes offline then back online with a new socket.
4. Confirm the restarted connection still delivers well-formed messages.

- [x] Unit tests

### Release notes

- Sync clients now restart the connection when they receive a malformed WebSocket message, so a corrupt frame can no longer silently leave the client out of sync.

### DIFF
--- a/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
+++ b/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
@@ -410,23 +410,29 @@ describe('ClientWebSocketAdapter', () => {
 			expect(onMessage).toHaveBeenCalledWith({ type: 'message', data: 'hello' })
 		})
 
-		it('[CW7] drops malformed JSON messages without closing the socket', async () => {
+		it('[CW7] restarts the connection on a malformed JSON message instead of forwarding it', async () => {
 			const warnOnceMock = vi.mocked(warnOnce)
 			const onMessage = vi.fn()
+			const onStatusChange = vi.fn()
 			adapter.onReceiveMessage(onMessage)
+			adapter.onStatusChange(onStatusChange)
 
 			await waitFor(() => adapter._ws?.readyState === WebSocket.OPEN)
+			const prevWs = adapter._ws
 			warnOnceMock.mockClear()
 
 			connectedServerSocket.send('{ "type": "message",')
-			connectedServerSocket.send('{ "type": "message", "data": "hello" }')
 
-			await waitFor(() => onMessage.mock.calls.length === 1)
-			expect(onMessage).toHaveBeenCalledWith({ type: 'message', data: 'hello' })
-			expect(adapter.connectionStatus).toBe('online')
+			// the malformed message is never forwarded to listeners, and the connection restarts so
+			// the client re-hydrates from the last known server clock rather than silently desyncing
+			await waitFor(() => adapter._ws !== prevWs && adapter._ws?.readyState === WebSocket.OPEN)
+			expect(onMessage).not.toHaveBeenCalled()
 			expect(warnOnceMock).toHaveBeenCalledWith(
-				'Received malformed WebSocket message. Dropping message.'
+				'Received malformed WebSocket message. Restarting the connection.'
 			)
+			expect(onStatusChange).toHaveBeenCalledWith({ status: 'offline' })
+			expect(onStatusChange).toHaveBeenCalledWith({ status: 'online' })
+			expect(adapter.connectionStatus).toBe('online')
 		})
 
 		it('[CW7] stops delivering messages after unsubscribe', async () => {
--- a/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
+++ b/packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
@@ -433,6 +433,12 @@ describe('ClientWebSocketAdapter', () => {
 			expect(onStatusChange).toHaveBeenCalledWith({ status: 'offline' })
 			expect(onStatusChange).toHaveBeenCalledWith({ status: 'online' })
 			expect(adapter.connectionStatus).toBe('online')
+
+			// the restarted connection is fully functional: a well-formed message on the new socket
+			// is delivered to listeners as usual
+			connectedServerSocket.send('{ "type": "message", "data": "hello" }')
+			await waitFor(() => onMessage.mock.calls.length === 1)
+			expect(onMessage).toHaveBeenCalledWith({ type: 'message', data: 'hello' })
 		})
 
 		it('[CW7] stops delivering messages after unsubscribe', async () => {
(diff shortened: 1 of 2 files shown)


packages/sync-core/src/lib/ClientWebSocketAdapter.ts
```diff
@@ -216,7 +216,13 @@ export class ClientWebSocketAdapter implements TLPersistentClientSocket<
...
 				parsed = JSON.parse(ev.data.toString())
 			} catch {
-				warnOnce('Received malformed WebSocket message. Dropping message.')
+				// A malformed message means the connection delivered corrupt data, so we can't trust
+				// that we're still in sync. Dropping the message can silently desync the client (e.g. a
+				// missed 'patch' is applied as if it never happened), so instead we restart the
+				// connection. Reconnecting re-hydrates the store from the last known server clock and
+				// brings us fully back in sync.
+				warnOnce('Received malformed WebSocket message. Restarting the connection.')
+				this.restart()
```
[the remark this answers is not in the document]
mimecuvalo: False positive — the assert can't be reached on that path.

`restart()` → `_closeSocket()` calls `ws.close()`, which synchronously sets `readyState` to `CLOSING`. Per the HTML spec's "a WebSocket message has been received" algorithm, the queued task's first step is "if `readyState` is not OPEN, return" — so any frame that arrived back-to-back with the malformed one (even if its task was already queued before we called `close()`) sees `CLOSING` when it runs and returns before firing `onmessage`. The handler, and its assert, never runs.

This is also the existing invariant rather than something this change introduces: `onmessage`/`onopen` use `assert` precisely because an orphaned (CLOSING/CLOSED) socket can't receive messages or open, while `onclose`/`onerror` — which *can* fire post-close — use a `this._ws === ws` guard. `restart()`/`_closeSocket()` already orphan sockets at every other call site (window-offline, reconnect detection); this just adds one more caller on the same path. The CW10 test documents the same invariant.

steveruizok:
Looks good to me but I don't think I can tell either way. Perhaps @frolic or @jsscclr ?

frolic (approved):
Seems reasonable! Is there any risk of a loop happening where one malformed message restarts the connection and receives the same malformed message and repeating the disconnection? Not really sure how/why we get malformed messages.

mimecuvalo:
no rush to land, i'd like one more set of 👀 from @MitjaBezensek when he's back as a sanity check

packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
```diff
@@ -410,23 +410,29 @@ describe('ClientWebSocketAdapter', () => {
...
+			// the client re-hydrates from the last known server clock rather than silently desyncing
+			await waitFor(() => adapter._ws !== prevWs && adapter._ws?.readyState === WebSocket.OPEN)
+			expect(onMessage).not.toHaveBeenCalled()
 			expect(warnOnceMock).toHaveBeenCalledWith(
-				'Received malformed WebSocket message. Dropping message.'
+				'Received malformed WebSocket message. Restarting the connection.'
 			)
+			expect(onStatusChange).toHaveBeenCalledWith({ status: 'offline' })
+			expect(onStatusChange).toHaveBeenCalledWith({ status: 'online' })
+			expect(adapter.connectionStatus).toBe('online')
```
MitjaBezensek, line 435: Should we also check that we can correctly receive messages after the reconnect?

MitjaBezensek (approved):
> connect response — isConnectedToRoom stays false and the health-check loop resets the connection.

Don't think this description is correct since both ping and health check have an early return `if (!this.isConnectedToRoom) return`. The logic is correct, just this bullet is a bit misleading.

packages/sync-core/src/lib/ClientWebSocketAdapter.test.ts
mimecuvalo: Good call — added in 80e708ca7. The CW7 test now sends a well-formed message on the new socket after the restart and asserts it's delivered to listeners.

mimecuvalo:
@frolic no loop risk in practice: a malformed frame is transport-level corruption (truncated/garbled bytes), not a deterministic server output, so there's nothing stored that would re-serialize to the same bad bytes on reconnect — the server re-hydrates fresh state from the last known `serverClock`. If the underlying connection is genuinely flaky, reconnects are already the existing behavior (the health-check loop resets on silence too), and `ReconnectManager`'s backoff throttles the churn. So worst case is the same reconnect backoff we already have, not a tight loop on the same message.

@MitjaBezensek good catch — you're right, both the ping and health-check loops early-return on `!isConnectedToRoom`, so that bullet was wrong. I've updated the PR description: a dropped `connect` response just leaves `isConnectedToRoom` false, so the client never enters the connected state and can't apply stale server data — it stays visibly offline rather than silently desyncing, which is the property that matters for motivating why `patch` is the uniquely dangerous case.

steveruizok:
Ready to go?

steveruizok:
Land it if it is @mimecuvalo !

mimecuvalo:
> @frolic no loop risk in practice: a malformed frame is transport-level corruption (truncated/garbled bytes), not a deterministic server output, so there's nothing stored that would re-serialize to the same bad bytes on reconnect — the server re-hydrates fresh state from the last known `serverClock`. If the underlying connection is genuinely flaky, reconnects are already the existing behavior (the health-check loop resets on silence too), and `ReconnectManager`'s backoff throttles the churn. So worst case is the same reconnect backoff we already have, not a tight loop on the same message.
> 
> @MitjaBezensek good catch — you're right, both the ping and health-check loops early-return on `!isConnectedToRoom`, so that bullet was wrong. I've updated the PR description: a dropped `connect` response just leaves `isConnectedToRoom` false, so the client never enters the connected state and can't apply stale server data — it stays visibly offline rather than silently desyncing, which is the property that matters for motivating why `patch` is the uniquely dangerous case.

ugh sry about the AI comment :-/

PR #9382, opened 2026-06-23
fix(tldraw): refresh minimap screen bounds on pointer down

### LINKED ISSUE #8383: Minimap click has incorrect pointer offset
Clicking on the minimap to navigate the canvas has an offset from the actual mouse pointer position. The viewport doesn't center where you'd expect based on where you clicked. I could not reliably reproduce this, but I did have it happen two times in a short period when doing local development.

The current viewport element (the grey box on the minimap) also seems quite narrow compared to the session when it works correctly.

https://github.com/user-attachments/assets/e9b0a97f-9eb4-4e24-a62e-fa34f415ba64

## Steps to reproduce

1. Open tldraw with some shapes on the canvas
2. Click on the minimap to navigate
3. Observe that the viewport jumps to a position offset from where you clicked
4. The browser window was on the right half on my non retina screen when this happened. Might be a contributing factor.

## Expected behavior

The viewport should center on the point where you clicked in the minimap.

After refresh it started to work. Also compare the width of the viewport element.

https://github.com/user-attachments/assets/cc5c4322-25d7-473b-a56f-4e9d0cf7ac37

## Actual behavior

The viewport centers on a position that's offset from the click point. The offset is not consistent and the issue is difficult to reproduce reliably.

kaneel:
## What

Fixes #8383

Fixes the minimap click offset where clicking to navigate jumps to a point offset from the cursor.

## Why

The minimap converts a click to a page point using its cached on-screen rect (`canvasBoundingClientRect`). That cache was only refreshed by a `ResizeObserver`, which fires on size changes but **not** when the element moves on screen without resizing — page scroll, toggling panels, layout reflow, or an embedded page shifting around. When that happened, the cached position went stale and every click was offset by however much the minimap had moved.

## How

- Refresh the cached rect on **pointer down**, before mapping the click to a page point. This covers every navigation path (click, drag, double-click), since a drag always starts with a pointer down and a double-click is always preceded by one.
- The `canvasBoundingClientRect` atom now uses `Box` value-equality (`isEqual`), so re-measuring to the same rect is a no-op and never triggers a re-render.
- `render()` keeps reading the cached value, so the per-frame render path stays free of layout reads.

## Performance

The fix needs a fresh `getBoundingClientRect()` measurement at the right moment. The table shows how many measurements each approach makes (a forced layout read), per scenario:

| Scenario | Cache only (before, buggy) | `getBoundingClientRect` in `render()` | Refresh in `getMinimapPagePoint` | Refresh on pointer down (this PR) |
|---|---|---|---|---|
| Idle | 0 | 0 | 0 | 0 |
| Main-canvas pan/zoom (not on minimap) | 0 | **1 per frame** | 0 | 0 |
| Hover over minimap | 0 | 0 | 2 per `pointermove` | 0 |
| Click to navigate | 0 | 1 per frame while easing | 1 | **1** |
| Drag on minimap | 0 | 1 per frame | 2 per `pointermove` | **1 (at drag start)** |
| Correct after the minimap moves without resizing | ❌ | ✅ | ✅ | ✅ |

Pointer-down refresh fixes the bug with a single measurement per interaction, while keeping the render loop and normal canvas panning free of layout reads. The only path that still uses the last-known rect is plain hover, which drives hover feedback only — never navigation.

## Test plan

- [ ] Open tldraw with shapes, click around the minimap — viewport centers on the click point.
- [ ] Scroll the page (or toggle panels) so the minimap moves without resizing, then click the minimap — still centers correctly.
- [ ] Drag on the minimap — viewport follows smoothly.
- [ ] Double-click the minimap — centers and zooms to the point.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
+++ b/packages/tldraw/src/lib/ui/components/Minimap/DefaultMinimap.tsx
@@ -76,6 +76,8 @@ export function DefaultMinimap() {
 			if (!minimap) return
 			if (e.button !== 0) return
 
+			minimap.updateCanvasScreenBounds()
+
 			const elm = e.currentTarget
 			setPointerCapture(elm, e)
 			if (!editor.getCurrentPageShapeIds().size) return
--- a/packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
+++ b/packages/tldraw/src/lib/ui/components/Minimap/MinimapManager.ts
@@ -83,16 +83,21 @@ export class MinimapManager {
 		return new Box(x, y, width, height)
 	}
 
-	private readonly canvasBoundingClientRect = atom('canvasBoundingClientRect', new Box())
+	private readonly canvasBoundingClientRect = atom('canvasBoundingClientRect', new Box(), {
+		isEqual: (a, b) => a.equals(b),
+	})
 
 	getCanvasScreenBounds() {
 		return this.canvasBoundingClientRect.get()
 	}
 
+	updateCanvasScreenBounds() {
+		this.canvasBoundingClientRect.set(this._getCanvasBoundingRect())
+	}
+
 	private _listenForCanvasResize() {
 		const observer = new ResizeObserver(() => {
-			const rect = this._getCanvasBoundingRect()
-			this.canvasBoundingClientRect.set(rect)
+			this.updateCanvasScreenBounds()
 		})
 		observer.observe(this.elem)
 		observer.observe(this.container)


steveruizok:
Were you able to reproduce this?

kaneel:
> Were you able to reproduce this?

I've reproduced something a bit different, but it's not exactly as it was described, I ran an example page where the canvas is not embedded into the full available width and height and in that very case, it's clear that slightly moving around the page will mess up the cached bounds.

However, it seems to have occured for Mitja on dotcom, it would be interesting to ask him more about how he figured this issue on dotcom.  To be fair it's intriguing that it's never been reported by our users but it's not impossible some may know about it and just refresh the browser.

I can park this one under "lessons learned" and we can discuss later.

max-drake:
@kaneel I have the [inset editor example](https://examples.tldraw.com/inset) open and I'm struggling to repro the new issue you described, can you give a slightly more detailed description or a video?

kaneel:
maybe not worth it

PR #9384, opened 2026-06-23
fix(editor): treat all loopback hosts as development environment

mimecuvalo:
In order to support local development served over loopback addresses other than `localhost`, this PR makes the license manager treat all loopback hosts as a development environment.

Previously `LicenseManager.getIsDevelopment()` only recognized `localhost`. When an app was served over `https://127.0.0.1` or `https://[::1]` (for example, SiYuan's plugin host opens pages on `https://127.0.0.1`), tldraw treated it as a production deployment and validated the license against the domain, surfacing license errors in what is really a local dev environment.

This change recognizes the full IPv4 loopback range (`127.0.0.0/8`), the IPv6 loopback (`::1`, with or without brackets), and `localhost` as development hosts.

### Change type

- [x] `bugfix`

### Test plan

1. Serve a tldraw app with a non-`localhost` loopback host (e.g. `https://127.0.0.1` or `https://[::1]`) in a production build.
2. Confirm it is treated as a development environment (no production license validation/error).

- [x] Unit tests

### Release notes

- Fix the license manager treating loopback hosts other than `localhost` (such as `127.0.0.1` and `::1`) as production environments.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Core code | +8 / -2 |
| Tests   | +19 / -14 |

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.test.ts
+++ b/packages/editor/src/lib/license/LicenseManager.test.ts
@@ -62,26 +62,29 @@ describe('LicenseManager', () => {
 
 		it('Signals that it is development mode for loopback hosts', async () => {
 			process.env.NODE_ENV = 'production'
-			const schemes = ['http', 'https']
-			const hosts = ['localhost', '127.0.0.1', '127.0.0.2', '[::1]']
-			for (const scheme of schemes) {
-				for (const host of hosts) {
-					// @ts-ignore
-					delete window.location
-					// @ts-ignore
-					window.location = new URL(`${scheme}://${host}:3000`)
+			try {
+				const schemes = ['http', 'https']
+				const hosts = ['localhost', '127.0.0.1', '127.0.0.2', '[::1]']
+				for (const scheme of schemes) {
+					for (const host of hosts) {
+						// @ts-ignore
+						delete window.location
+						// @ts-ignore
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/license/LicenseManager.test.ts
```diff
@@ -60,23 +60,28 @@ describe('LicenseManager', () => {
...
+					const licenseKey = await generateLicenseKey(STANDARD_LICENSE_INFO, keyPair)
+					const result = await testEnvLicenseManager.getLicenseFromKey(licenseKey)
+					expect(result).toMatchObject({
+						isLicenseParseable: true,
+						isDomainValid: false,
+						isDevelopment: true,
+					})
+				}
 			}
+			process.env.NODE_ENV = 'test'
```
MitjaBezensek: Should we do this in a `finally` block so that any potential exceptions in licensing code don't leak the `production` option.

MitjaBezensek (approved):
Looks good. Left a small comment about making the test a bit more isolated.

packages/editor/src/lib/license/LicenseManager.test.ts
mimecuvalo: Good call — wrapped the test body in a try/finally so the NODE_ENV reset always runs even if the licensing code throws.

Issue #7720, opened 2026-01-21
Exported PNG mis-aligned in RTL


mustafamohsen:
When adding Arabic (RTL) text to the canvas, text alignment gets reversed upon export. In the following images, the dashboard screenshot is properly aligned, followed by the misaligned exported image

Here's a screenshot from the canvas:
<img width="1252" height="804" alt="Image" src="https://github.com/user-attachments/assets/0ca090fc-2952-4d0b-977b-4afa0ef94b14" />

and here's the exported PNG:

<img width="606" height="330" alt="Image" src="https://github.com/user-attachments/assets/0226a27c-bcf6-428b-81f3-b48f3aa87a09" />

## Steps to reproduce
1. Add a text elemtn
2. Add multiple lines Arabic text (e.g. "مرحباً بكم")
3. Right-align the text
4. Create a copy of the text element, and left-align it
5. Export to PNG and compare

Potentially relevant to #7497

steveruizok:
Thanks for the issue @mustafamohsen !

steveruizok:
@claude Can you try to find a root cause for this? Maybe something in Tiptap?

MitjaBezensek:
Possibly related to https://github.com/tldraw/tldraw/issues/8802

angrycaptain19:
See #8986 for starting to look at how we're truncating RTL exports. We'll need to solve the truncation problem for selection and export of RTL text

PR #9387, opened 2026-06-23
fix(tldraw): expand text shape bounds to fit cursive and RTL glyph ink

### LINKED ISSUE #8802: Exported PNG/SVG cuts off RTL text

### LINKED ISSUE #8803: Arabic text renders outside the text shape bounding box

frolic:
In order to stop cursive, RTL (e.g. Arabic, Hebrew), and italic glyphs from being clipped at the edge of their text shape — both in the on-canvas selection box (#8803) and on PNG/SVG export (#8802) — this measures the real glyph **ink** and grows the shape's geometry (and the export's `<foreignObject>`) to contain it.

Text is measured for pen-advance width via the DOM, which ignores glyph side bearings, so the trailing stroke of an Arabic/Hebrew word, the slant of an italic, or a tall diacritic ends up outside the box. The fix measures how far the ink spills past the advance box on each side and grows the shape's bounds by it. The text content and `props.w` are unchanged, so layout, alignment, and resizing are unaffected — only the reported bounds (indicator, selection, hit-test, snapping, and export viewport) grow to contain the ink.

**The ink overflow is measured per word and cached.** A word shapes independently of its neighbours (whitespace breaks cursive joining) and lines only ever break at word boundaries, so a word's ink bleed is the same wherever it lands in a wrap. `TextManager.measureWordsInkOverflow(words, opts)` measures each word's bleed once with the canvas `TextMetrics` API (no layout/reflow), caches it by font + word (shared across every shape), and returns the max over the text's words. So resizing/reflowing re-measures only the advance box — the same DOM reflow done before this PR — while the ink padding is a cached lookup that doesn't depend on the wrap. Plain Latin text bleeds ~0, so its bounds are unchanged.

The export needed a step the earlier attempt (#8986) missed: a shape's `<foreignObject>` **clips its content to its own bounds during rasterization**, so growing only the geometry/outer viewport still clipped the export. `toSvg` now also grows the foreignObject by the ink overflow and pads the content back into place, so the text lays out identically and only the clip region grows.

Closes #8802
Closes #8803

### Performance

Resizing many text shapes re-measures their geometry every frame, but the only work this PR adds is the per-word ink overflow — a cached, wrap-invariant lookup (a word's bleed doesn't depend on how the text wraps) on top of the advance-box reflow resizing already does. The per-word path is nested (`font -> word -> bleed`, so the lookup needs no string key), builds the font shorthand once per run, and memoizes the rich-text word split by the doc object — so a width drag re-measures only the advance.

To isolate the added cost from machine drift, the two paths are timed **interleaved within each frame** (advance-only vs advance + per-word ink) and the **paired per-frame difference** is taken, which cancels drift. Re-measuring N wrapped text shapes per resize frame (dev build):

| shapes | frames | added per frame (paired ± SE) |
| --- | --- | --- |
| 1,000 | 40 | within noise (-0.3 ± 0.9 ms) |
| 2,000 | 30 | +3.4 ± 1.1 ms (~+1.5%) |

So the ink overflow adds ~1.5% at most to a mass-resize frame; the rest is the (unchanged) advance reflow. For contrast, a per-layout approach that re-measured ink from the wrapped layout every frame added ~75-150% at this scale. Absolutes are dev-build inflated; the paired delta is the signal. Reproduce with `yarn e2e-perf`.

### Editor API

| Method | Description |
| --- | --- |
| `TextManager.measureWordsInkOverflow(words, opts)` | The ink overflow (side bearings, italic slant, tall diacritics) of wrapped text, derived from its words' cached per-word bleeds — wrap-invariant, so it isn't re-measured as the text reflows or scales. |

### Testing

- `test-text-ink-bounds.spec.ts` — deterministic, independent oracle. Selection: pixel-scans the real glyph ink and asserts `getShapeGeometry` contains it. Export: compares the autosize export's painted glyph bounds to an oversized fixed-width box (which can't clip). Both fail on a pre-fix build and pass here.
- `text-measurement-perf.spec.ts` + `text-resize-profile.spec.ts` (`yarn e2e-perf`) — a measurement microbenchmark and an end-to-end CPU profile of a real resize-drag over a fleet of text shapes.
- `test-rtl-text-alignment.spec.ts` (#7720) — its flush-alignment tolerance is widened: now that cursive ink isn't clipped, word-final trailing strokes legitimately vary per line (the lines are still layout-flush). It previously passed only because main clips those strokes.
- `export-snapshots.spec.tsx` — the RTL cases in the `Text alignment` golden grow to enclose the recovered ink (previously clipped).

### Change type

- [x] `bugfix`

### Test plan

1. Create a text shape with the default (Draw) font and type Arabic or Hebrew, e.g. `مرحباً بكم`.
2. Select it: the selection box / hover indicator enclose the glyphs instead of clipping the trailing cursive stroke.
3. Export to PNG/SVG: glyphs are no longer clipped at the box edge (including tall diacritics' top/bottom).
4. Confirm Latin text is unchanged (no extra padding, box origin unmoved) and fixed-width wrapped text keeps its width.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fix text shapes clipping cursive, RTL (e.g. Arabic, Hebrew), and italic glyphs at the edge of their bounding box, both on the canvas (selection box) and in PNG/SVG exports.

### API changes

- Added `TextManager.measureWordsInkOverflow(words, opts)` for the wrap-invariant ink overflow of text.

### Code changes

| Section         | LOC change                                      |
| --------------- | ----------------------------------------------- |
| Core code       | +230 / -41                                      |
| Tests           | +762 / -3                                       |
| Automated files | +6 / -0 (api report) + snapshot PNGs (binary)   |


mimecuvalo:
similar comment to https://github.com/tldraw/tldraw/pull/9266 - would be good to get some perf metrics on this - that's the main thing i'm concerned with really

mimecuvalo:
oh hah, aaand i totally missed that this has perf — nevermind!

frolic:
closing in favor of #9414 and #9417 (clean copy of new tests/bench -> fix tests)

PR #9388, opened 2026-06-23
test(tldraw): add RTL text export alignment regression coverage (#7720)

### LINKED ISSUE #7720: Exported PNG mis-aligned in RTL

frolic:
Adds the missing regression coverage for #7720 (RTL text alignment reversed in PNG/SVG exports). I found while building #9387 that #7720 is **already fixed on `main`** — but nothing guarded it and the issue was never closed.

### When/how it was fixed

The bug was hardcoded *physical* alignment in `editor.css`:

```css
.tl-text-wrapper[data-align='start'] { text-align: left; }
.tl-text-wrapper[data-align='end']   { text-align: right; }
```

For RTL, `start` should map right and `end` left, so these reversed the export. Those rules were removed in **#8410 (the theme-system refactor, merged 2026-04-02)**, which moved alignment to inline *logical* `text-align`. Combined with the pre-existing `unicode-bidi: plaintext` (#6256) and `dir="auto"`, RTL export alignment now matches the canvas. So #7720 was fixed incidentally ~2.5 months after filing, never linked back to the issue.

### Coverage added

**Automated** (`test-rtl-text-alignment.spec.ts`): exports a two-line shape (long line over short line) and pixel-scans per line-band to check which edge the short line flushes to. Asserts `start`→right / `end`→left for RTL, with LTR as the mirror control. Self-relative, so it doesn't depend on the fallback font. Verified it's a real guard: temporarily restoring the physical mapping fails the RTL cases while LTR stays green — exactly the #7720 signature.

**Visual** (`export-snapshots.spec.tsx`): a **dedicated `Text alignment` snapshot group** (its own image, so it doesn't grow the shared `Text rendering` golden) showing RTL `start`/`end` and the LTR mirror. Confirmed visually: RTL `start` flushes lines right, `end` left.

No production changes — `main` already renders this correctly; this locks it in and gives #7720 a clean resolution.

Closes #7720

### Change type

- [x] `test`

### Test plan

1. Review the `Text alignment` snapshot: RTL `start` flush-right, `end` flush-left; LTR mirror.
2. `cd apps/examples && yarn e2e tests/test-rtl-text-alignment.spec.ts` passes.

- [ ] Unit tests
- [x] End to end tests

### Code changes

| Section         | LOC change                          |
| --------------- | ----------------------------------- |
| Tests           | +205 / -0                           |
| Automated files | 4 new snapshot PNGs (binary)        |


mimecuvalo:
good idea 👍

PR #9395, opened 2026-06-23
fix(tldraw): mirror crop when flipping a grouped image by drag resize

### LINKED ISSUE #9392: Flipping group containing cropped image by drag resize shifts image position
### What happened?

This one's much easier to show than tell (please see below) - but in essence, if you have a cropped image within a group and flip/invert the group orientation by dragging the group's corner handles, the portion of the cropped image that is visible in the crop mask gets shifted/mirrored incorrectly.

- Create/add an asymmterical image
- Crop to a corner of the image
- Add a shape off to the side (e.g. above and to the left) if the image
- Group the image and shape together
- Drag the group corner across to flip the group

See video demo below:

https://github.com/user-attachments/assets/d41b67d2-e3d8-4464-8251-a1d4682b3620

Chrome

### Where did this happen?

m31-galaxy:
In order to fix flipping a group that contains a cropped image, this PR mirrors the crop whenever the resize scale becomes negative on an axis, regardless of resize mode. Closes #9392.

When you flip a group by dragging a corner handle, the group resizes its children with mode `scale_shape` and an arbitrary negative scale (rarely exactly `-1`, since the drag also resizes). `ImageShapeUtil.onResize` only mirrored the crop when the scale was exactly `-1` (the flip command) or in `resize_bounds` mode (a lone selected image). So a grouped cropped image had its `flipX`/`flipY` toggled but its crop left un-mirrored, and the visible window ended up showing the wrong part of the image — the "shifted" appearance in the report.

Both of the previous conditions reduce algebraically to "the scale sign flipped on that axis", so this replaces them with a direct `scaleX < 0` / `scaleY < 0` check. The `flipShapes` command (which passes exactly `-1`) and lone-image drag flips are unchanged; group drag flips now mirror the crop the same way.

### Change type

- [x] `bugfix`

### Test plan

1. Add an asymmetrical image to the canvas and crop it to a corner.
2. Add a second shape offset from the image and group the two.
3. Drag a corner handle of the group past the opposite edge to flip it.
4. The cropped image's visible region mirrors correctly and stays aligned, matching the result of the flip command.

- [x] Unit tests

### Release notes

- Fix the crop of a grouped, cropped image shifting when the group is flipped by dragging a resize handle.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +7 / -11   |
| Tests     | +88 / -0   |

### DIFF
--- a/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx
@@ -170,7 +170,7 @@ export class ImageShapeUtil extends BaseBoxShapeUtil<TLImageShape> {
 	override onResize(shape: TLImageShape, info: TLResizeInfo<TLImageShape>) {
 		let resized: TLImageShape = resizeBox(shape, info)
 		const { flipX, flipY } = info.initialShape.props
-		const { scaleX, scaleY, mode } = info
+		const { scaleX, scaleY } = info
 
 		resized = {
 			...resized,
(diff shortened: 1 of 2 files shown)


m31-galaxy:
Manual testing seems like the fix worked as intended, ready for review :>

kaneel (approved):
LGTM (thanks for the very good description on how to test this :)

PR #9396, opened 2026-06-23
feat(tldraw): replace a cropping image by pasting a new image

### LINKED ISSUE #9334: Replace an image by pasting while cropping, fitting the new image to the frame centered
🗣️: Replace an image with a different image by pasting while the first image is cropping. Make the scale work so that the image fits to the bounds of the new image, centered.

---

When an image shape is in crop mode, pasting a new image replaces the original image in place (it does not create a separate shape). The shape's on-canvas frame stays fixed, and the crop is recomputed from scratch for the new image: the new image is scaled with "cover" so it fills the frame (edges cropped as needed) and is centered, regardless of any aspect-ratio difference. After the paste, the editor stays in crop mode on the new image so the user can keep adjusting. This behavior applies to paste only.

## Open questions

1. **When the new image has a different aspect ratio, "fit to bounds, centered" — contain or cover?**
   Cover: the new image fills the frame and is centered, with overflowing edges cropped.
2. **Should the existing crop frame stay fixed while the new image fits into it, or should the crop reset to show the whole new image?**
   Reset: the crop is recomputed fresh for the new image (centered cover into the fixed frame), not carried over from the original image.
3. **After the paste replaces the image, should the editor stay in crop mode or exit?**
   Stay in crop mode on the new image.
4. **Should this be limited to paste, or also apply to drag-and-drop / file picker while cropping?**
   Paste only.

m31-galaxy:
In order to let people swap a cropped image without leaving crop mode (closes #9334), this PR makes pasting a single image while an image is in crop mode replace that image in place, matching the existing **"replace image"** menu action — the crop transform is preserved across the swap. The editor stays in crop mode. This applies to **paste only**: drag-and-drop and the file picker still create a new shape.

It also fixes a focus issue that made the paste unreliable: the image crop zoom slider holds keyboard focus while cropping (and re-grabs it whenever the crop toolbar remounts after a drag/resize), which previously suppressed clipboard paste until the user clicked the image again. Paste is now allowed to proceed while cropping; open menus and text editing still take precedence.

### Change type

- [x] `feature`

### Test plan

1. Add an image to the canvas and double-click it to enter crop mode.
2. Copy a different image to the clipboard.
3. Paste (Cmd/Ctrl+V) — the cropped image is replaced in place, the crop transform is preserved, and you stay in crop mode.
4. Repeat the paste immediately after each of: dragging the image within the crop, dragging a crop border handle, and dragging the zoom slider — paste works without a preceding click.
5. Confirm drag-and-drop and file-picker insert still create a new shape, and that an open aspect-ratio dropdown still blocks paste.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- When cropping an image, paste an image to replace it in place, matching the replace image action.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +29 / -4   |
| Tests     | +65 / -0   |

### DIFF
--- a/packages/tldraw/src/lib/ui/hooks/clipboard/pasteFiles.ts
+++ b/packages/tldraw/src/lib/ui/hooks/clipboard/pasteFiles.ts
@@ -20,6 +20,24 @@ export async function pasteFiles(
 		blob instanceof File ? blob : new File([blob], 'tldrawFile', { type: blob.type })
 	)
 
+	// When a single image is pasted while an image is being cropped, replace the
+	// cropped image in place instead of creating a new shape. This matches the
+	// "replace image" action, preserving the crop transform across the swap.
+	const croppingShapeId = editor.getCroppingShapeId()
+	if (croppingShapeId && files.length === 1 && files[0].type.startsWith('image/')) {
+		const croppingShape = editor.getShape(croppingShapeId)
+		if (croppingShape?.type === 'image') {
+			editor.markHistoryStoppingPoint('paste')
+			await editor.replaceExternalContent({
+				type: 'file-replace',
+				file: files[0],
+				shapeId: croppingShapeId,
+				isImage: true,
+			})
+			return
+		}
+	}
+
 	editor.markHistoryStoppingPoint('paste')
 
 	await putPastedExternalContent(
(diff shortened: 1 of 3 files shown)


m31-galaxy:
@steveruizok looking at #6323 it seems the design decision was made in the past to preserve image transform when using the "Replace media..." button. Should the same be done here for parity/consistency, instead of the behaviour requested in #9334?

m31-galaxy:
Discussed with @steveruizok in-person, the decision was made to preserve image transform here as well.

m31-galaxy:
Ready for review!

PR #9398, opened 2026-06-23
fix(editor): mirror modifier key state from events instead of debouncing

### LINKED ISSUE #7981: Meta key reactive state not cleared on keyup until pointer moves (macOS)
## Description

The reactive input state for the Meta (Cmd) key — and by extension `getAccelKey()` on macOS — does not properly react to the key being released. When subscribing to `editor.inputs.getMetaKey()`, the signal fires `true` on keydown, but does **not** fire `false` on keyup. It only clears after a subsequent pointer event (e.g. moving the mouse).

This behavior is **not** the same for Shift, Alt, or Ctrl, which all properly react to both keydown and keyup.

## Steps to reproduce

1. Open the reactive inputs example (or any setup that reactively subscribes to `editor.inputs.getMetaKey()`)
2. Press the Meta (Cmd) key — the value reactively changes to `true`
3. Release the Meta key — the value stays `true`
4. Move the mouse — the value finally changes to `false`

Compare with Shift: pressing and releasing Shift correctly toggles between `true` and `false` without requiring pointer movement.

## Expected behavior

`getMetaKey()` should reactively change to `false` immediately when the Meta key is released, matching the behavior of `getShiftKey()`, `getAltKey()`, and `getCtrlKey()`.

## Actual behavior

`getMetaKey()` stays `true` after Meta keyup until a pointer event dispatches with `metaKey: false`, which triggers the 150ms timeout to clear it.

## Root cause

On macOS, the native `KeyboardEvent` for the Meta key's own `keyup` event still reports `e.metaKey === true` (a browser/OS quirk). The modifier state update logic in `Editor.dispatch()` checks `info.metaKey` first and re-sets the atom to `true`, so the 150ms timeout branch (`!info.metaKey && inputs.getMetaKey()`) is never reached:

```ts
// This fires for Meta keyup on macOS because e.metaKey is still true
if (info.metaKey) {
    clearTimeout(this._metaKeyTimeout)
    this._metaKeyTimeout = -1
    inputs.setMetaKey(true)  // re-set to true even during keyup!
}
```

Since `getAccelKey()` derives from `getMetaKey()` on macOS, it is also affected.

## Environment

- macOS (the `e.metaKey` quirk is macOS-specific)
- Tested across browsers on macOS
- Both the new `getMetaKey()` API and the deprecated `editor.inputs.metaKey` property are affected

## Related

- #7713 — adds modifier keys to the reactive inputs example (makes this bug visible)

kaneel:
Fixes #7981.

## The bug

On macOS, releasing Cmd left `editor.inputs.getMetaKey()` (and `getAccelKey()`)
stuck `true` until the next pointer event. Shift/Alt/Ctrl were unaffected.

## Root cause

Modifier state was cleared by a 150ms debounce: on release, a timer re-dispatched
a synthetic `key_up` carrying the *other* modifiers' current flags. Because
`useDocumentEvents` normalizes `ctrlKey: e.metaKey || e.ctrlKey`, pressing Cmd sets
*both* meta and ctrl. On release both timers were armed; the ctrl timer fired first
and re-dispatched a keyup with `metaKey` still `true`, which hit the "set true"
branch, re-asserted meta, and cancelled meta's own clear. Meta then only cleared
when a later pointer event reported `metaKey: false`.

Note: the issue guessed the cause was macOS reporting `metaKey: true` during Meta's
own keyup. That's not it — `e.metaKey` is `false` on keyup. The resurrection above
is the actual cause.

## The change

- Editor: modifier reactive state now simply mirrors the flags on the most recent
  event. Every event carries the live held-modifier set, so a modifier clears the
  instant any event reports it released. Removes the debounce timers and the
  synthetic keyup re-dispatch entirely (~110 lines).
- Driver (test harness): keyboard events now carry the full set of held modifiers,
  not just the key being pressed. This matched neither real DOM events nor the
  Driver's own keyup logic — an inconsistency that masked the issue in tests.

## Approaches tried

1. Minimal `&& !isKeyUp` guard on the "set true" branches (keeps the debounce).
   Works, but keeps the timer machinery and the footguns below.
2. Two-pass: set-on, then authoritative immediate clear on keyup. Works, more code.
3. Mirror-every-event (this PR). Simplest, and removes the debounce, which we
   believe is a misfeature (see below).

## Footguns we think we removed — please sanity check

The old 150ms debounce gave a "grace period" after releasing a modifier. Wondering if 
these were unintended side effects, not features, but they are behavior changes:

- Arrow binding: releasing Ctrl now re-binds immediately (was delayed ~150ms).
- Clone (alt-drag translate): releasing Alt ends the clone immediately.
- Scribble select: releasing Alt switches to box brushing immediately.
- Clipboard: a documented race is gone — `getShiftKey()` could read a stale `true`
  for 150ms after release, risking a plain-text paste on cmd+v
  (clipboardCallbacks.test.ts literally called this out).

## One capability intentionally dropped

The debounce could clear a modifier when its keyup never arrives (focus loss,
cmd+tab) with no follow-up event. Now recovery happens on the next event (e.g. a
pointer move) instead of a timer. We think this is fine, but it's the one thing lost.

## Test plan

- Full editor (826) and tldraw (2566) suites pass.
- Updated the tests that asserted the old debounce grace (modifiers, scribble/shift
  brushing, alt-clone, arrow ctrl-binding, clipboard delay) to the new immediate
  behavior.
  
  ### API changes

No public API surface changes — no exported methods, types, or parameters were added, removed, or changed, so `api-report.api.md` is unaffected.

Behavioral change (no signature change):

- `editor.inputs.getShiftKey()`, `getAltKey()`, `getCtrlKey()`, `getMetaKey()`, and `getAccelKey()` now reflect the modifier flags of the most recent event immediately, instead of clearing after a ~150ms debounce. Reactive subscribers see modifier releases instantly, and `getMetaKey()`/`getAccelKey()` no longer stay `true` on macOS until a pointer move (the bug this PR fixes).

Internal-only removals (not part of the public API; `@internal`, absent from `api-report.api.md`):

- Removed `Editor._setShiftKeyTimeout`, `_setAltKeyTimeout`, `_setCtrlKeyTimeout`, `_setMetaKeyTimeout` and the backing `_shiftKeyTimeout` / `_altKeyTimeout` / `_ctrlKeyTimeout` / `_metaKeyTimeout` fields.

Breaking changes:

- None to the public API. The change is behavioral: any code depending on a modifier staying briefly "held" after its keyup (a ~150ms grace) will now see it cleared immediately.

New/removed parameters:

- None.

### DIFF
--- a/packages/driver/src/lib/Driver.ts
+++ b/packages/driver/src/lib/Driver.ts
@@ -255,13 +255,20 @@ export class Driver {
 		name: TLKeyboardEventInfo['name'],
 		options = {} as Partial<Omit<TLKeyboardEventInfo, 'point'>>
 	): TLKeyboardEventInfo {
-		return {
-			shiftKey: key === 'Shift',
-			ctrlKey: key === 'Control' || key === 'Meta',
-			altKey: key === 'Alt',
-			metaKey: key === 'Meta',
-			accelKey: tlenv.isDarwin ? key === 'Meta' : key === 'Control' || key === 'Meta',
+		// Like a real DOM keyboard event, the flags reflect every modifier currently held — not
+		// just the key being pressed. We OR in the editor's current modifier state so that, e.g.,
+		// pressing Shift while Control is held still reports ctrlKey: true. (keyUp passes explicit
+		// flag overrides for its release semantics, so those win over these defaults.)
+		const flags = {
+			shiftKey: key === 'Shift' || this.editor.inputs.getShiftKey(),
+			ctrlKey: key === 'Control' || key === 'Meta' || this.editor.inputs.getCtrlKey(),
+			altKey: key === 'Alt' || this.editor.inputs.getAltKey(),
+			metaKey: key === 'Meta' || this.editor.inputs.getMetaKey(),
 			...options,
+		}
+		return {
+			...flags,
+			accelKey: flags.accelKey ?? (tlenv.isDarwin ? flags.metaKey : flags.ctrlKey || flags.metaKey),
 			name,
 			code:
 				key === 'Shift'
--- a/packages/driver/src/lib/Driver.ts
+++ b/packages/driver/src/lib/Driver.ts
@@ -255,20 +255,13 @@ export class Driver {
 		name: TLKeyboardEventInfo['name'],
 		options = {} as Partial<Omit<TLKeyboardEventInfo, 'point'>>
 	): TLKeyboardEventInfo {
-		// Like a real DOM keyboard event, the flags reflect every modifier currently held — not
-		// just the key being pressed. We OR in the editor's current modifier state so that, e.g.,
-		// pressing Shift while Control is held still reports ctrlKey: true. (keyUp passes explicit
-		// flag overrides for its release semantics, so those win over these defaults.)
-		const flags = {
-			shiftKey: key === 'Shift' || this.editor.inputs.getShiftKey(),
-			ctrlKey: key === 'Control' || key === 'Meta' || this.editor.inputs.getCtrlKey(),
-			altKey: key === 'Alt' || this.editor.inputs.getAltKey(),
-			metaKey: key === 'Meta' || this.editor.inputs.getMetaKey(),
-			...options,
-		}
 		return {
-			...flags,
-			accelKey: flags.accelKey ?? (tlenv.isDarwin ? flags.metaKey : flags.ctrlKey || flags.metaKey),
+			shiftKey: key === 'Shift',
+			ctrlKey: key === 'Control' || key === 'Meta',
+			altKey: key === 'Alt',
+			metaKey: key === 'Meta',
+			accelKey: tlenv.isDarwin ? key === 'Meta' : key === 'Control' || key === 'Meta',
+			...options,
 			name,
 			code:
 				key === 'Shift'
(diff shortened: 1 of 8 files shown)


steveruizok:
This PR is a regression—we do those timeouts on purpose to avoid a kind of race condition when releasing a click and a modifier key at the same time, which could produce an unintended outcome if the modifier key is released first.

kaneel:
> This PR is a regression—we do those timeouts on purpose to avoid a kind of race condition when releasing a click and a modifier key at the same time, which could produce an unintended outcome if the modifier key is released first.

Can you provide me a series of known behaviour triggering the issue? 

The more I looked at it the more some things did not add up? I'm open to a bit of push back but I want to make sure I did not find a smell here, the issue first did not make much sense because browsers are air-tight on keyboard events, there's no "stuck key" on the browser side, only in our code… which mean our code aint "righty right". 

My problem is that some tests seems to actually expect the timer in order to work, all of that with a forgetful Driver, these smells made me wonder if we had a problem somewhere else and I went poking a little bit too far maybe.

steveruizok:
A good case for the timers is alt+dragging to copy. Users will release the alt and the drag at the "same time" but one of those inputs will arrive first. If the alt release arrives first, the alt+drag will cancel and the interaction will be treated as a drag-to-translate instead. Many design tools do this modifier delay (see Figma) in order to prevent this bug.

Update, found the reproduction steps in the issue. I doubt this is the timers, but it may be!

steveruizok:
The fact that it's _only_ the meta key that produces this behavior, vs. the rest of the modifiers, is interesting.

steveruizok:
Found the fix, new PR on the way

steveruizok:
https://github.com/tldraw/tldraw/pull/9404

kaneel:
> Found the fix, new PR on the way

I was editing my PR to go back to the first commits that worked on the fix

PR #9399, opened 2026-06-24
ci: drop postgres client install from dotcom deploy

MitjaBezensek:
In order to stop installing an unused dependency on every dotcom deploy, this PR removes the `Install PostgreSQL client` step from the deploy workflow. `psql` was only used by the old `deployPermissionsToFlyIo()` step (running `permissions.sql` from `definePermissions`), which was removed in #7731 when Zero moved authorization to Synced Queries. No `psql` usage remains anywhere in the repo.

### Change type

- [x] `other`

### Test plan

- CI-only change; verified by confirming no `psql` references remain in the repo and that the deploy script no longer shells out to it.

### Code changes

| Config/tooling | +0 / -6    |


frolic (approved):

mimecuvalo:
🎉

PR #9402, opened 2026-06-24
feat(dotcom): land on most recent file on root and workspace switch

### LINKED ISSUE #9364: Visiting tldraw.com lands on your workspace instead of your most recent workspace or file
🗣️: visiting tldraw.com brings you back to your workspace, not your most recent workspace or most recent file

---

When you navigate to the bare tldraw.com URL, you land on your default workspace rather than the file you last had open. Returning to tldraw.com should restore your most recent file (and whatever workspace that file belongs to), so you can pick up where you left off instead of re-navigating from the workspace landing each time. This "most recent" state should follow you across devices and sessions, not just within one browser. If the most recent file is no longer available — deleted, moved, or you've lost access — fall back to the most recent file in the user's own workspace.

## Open questions

1. **What is the ideal landing target — your most recent file, or just your most recent workspace (then its default/last view)?**
   Most recent file, and the app should open into whatever workspace that file belongs to.
2. **Should this apply to the bare `tldraw.com` URL only, or also to other entry points (e.g. opening the app from a bookmark, returning after sign-in)?**
   The bare `tldraw.com` URL.
3. **Should this persist across devices/sessions (server-side last-visited state), or only within the same browser?**
   Across devices and sessions, so it needs server-side last-visited state.
4. **When the most recent file is unavailable (deleted, moved, or access lost), what should it fall back to?**
   The most recent file in the user's own workspace.

MitjaBezensek:
In order to let users pick up where they left off, this PR makes tldraw.com restore your most recently opened file instead of always dropping you on your home workspace. Closes #9364.

Visiting the bare `tldraw.com` URL now navigates to the file you most recently had open across **all** of your workspaces, opening into whatever workspace that file belongs to. Recency comes from `file_state`, which is synced per user, so the "most recent" file follows you across devices and sessions. Files you can no longer access (deleted, moved away, or revoked) are skipped, so the next most recent available file wins; if you have no visited files at all, it falls back to the most recent file in your home workspace.

The same logic now applies when switching workspaces from the sidebar switcher: it opens the file you most recently had open in that workspace, rather than the top of the pinned-first list. If you've never visited a file in that workspace, it falls back to that workspace's top file (and the empty-workspace seed/create behavior is unchanged).

Recency lives in two small helpers so it's computed consistently: `getFileVisitDate(state)` returns when the user last opened a file (`lastVisitAt → lastEditAt → firstVisitAt`, or `undefined` if never opened), and `getFileRecencyDate(state, file)` adds a `createdAt` fallback for ordering files in lists. Most-recent navigation ranks by `getFileVisitDate` only — a created-but-never-opened file (its `file_state` row has null visit timestamps) must not hijack the landing on its `createdAt`; it defers to the fallback. The recent-files list and workspace sorting use `getFileRecencyDate`. Note: this brings `lastVisitAt` into the sidebar recent-file ordering, which previously only considered `lastEditAt`/`firstVisitAt`.

### Change type

- [x] `feature`

### Test plan

Note: `lastVisitAt` is written on a 10s throttle (and flushed on file exit), so after opening a file either navigate away or wait ~10s before testing, so the visit timestamp has synced.

1. Sign in and open a file in a non-home workspace, then navigate to the bare `tldraw.com` URL — you should land back on that file, in its workspace.
2. Open files across multiple workspaces, then return to `tldraw.com` — you should land on the single most recently opened one.
3. Switch to another workspace via the sidebar switcher — you should land on the file you most recently had open there, not necessarily its top pinned file.
4. Delete/lose access to your most recent file, then return to `tldraw.com` — you should land on the next most recent available file (home workspace's most recent as the final fallback).
5. New user with no files visiting `tldraw.com` — a file is created and opened as before.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +56 / -11  |
| Tests   | +165 / -0  |

### Release notes

- tldraw.com now reopens your most recently opened file when you visit the site, and switching workspaces opens the file you most recently had open there.

### DIFF
--- a/apps/dotcom/client/src/tla/app/TldrawApp.ts
+++ b/apps/dotcom/client/src/tla/app/TldrawApp.ts
@@ -628,6 +628,46 @@ export class TldrawApp {
 		return sortedFiles
 	}
 
+	/**
+	 * The id of the user's most recently visited file, or null if they have none.
+	 *
+	 * Used to land the user on the file they last had open: across all workspaces when returning to
+	 * the root URL, or within a single workspace when switching to it. Recency comes from
+	 * `file_state`, which is synced per user, so the result follows the user across devices and
+	 * sessions. Files the user can no longer access (deleted, moved away, or revoked) drop out of
+	 * `file_state` sync or come back without a `file` relation, so they're skipped and the next most
+	 * recent available file wins.
+	 *
+	 * When the user has no visited files in scope, fall back to the most recent file in their home
+	 * workspace (root URL) or the top of the workspace's file list (workspace scope) — this covers
+	 * files the user has but has never opened on this account.
+	 *
+	 * @param workspaceId - When provided, only files visible in that workspace are considered.
+	 */
+	getMostRecentFileId(workspaceId?: string): string | null {
+		// Files in scope, ordered pinned-first then by recency. Used both to scope the recency
+		// search and as the fallback when the user has visited none of these files.
+		const scopedFiles = workspaceId ? this.getWorkspaceFilesSorted(workspaceId) : this.getMyFiles()
+		const fileIdsInScope = workspaceId ? new Set(scopedFiles.map((f) => f.fileId)) : null
+
+		let bestFileId: string | null = null
+		let bestDate = -Infinity
+		for (const state of this.getUserFileStates()) {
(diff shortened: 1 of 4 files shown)


mimecuvalo (approved):
LG

PR #9404, opened 2026-06-24
fix(editor): clear meta key reactive state on keyup (macOS)

### LINKED ISSUE #7981: Meta key reactive state not cleared on keyup until pointer moves (macOS)

steveruizok:
🗣️ Fixes a bug where we were assuming that metaKey would be false on meta key keyups, but to my surprise it was not, leading to sticky meta state.

---

Fixes the reactive Meta (Cmd) key input state not clearing on keyup until a pointer event fired.

On macOS, the native `KeyboardEvent` for the Meta key's own `keyup` event still reports `e.metaKey === true` (a browser/OS quirk). The modifier update logic in `Editor.dispatch()` checks `info.metaKey` first and re-sets the atom to `true`, so the branch that schedules the clearing timeout was never reached. The result: `editor.inputs.getMetaKey()` (and `getAccelKey()` on macOS) stayed `true` after release until a pointer event dispatched with `metaKey: false`. Shift, Alt, and Ctrl were unaffected because their native flags clear correctly on keyup.

The fix guards the re-set so it ignores `key_up` events, letting the existing clear-timeout branch run as it does for the other modifiers.

### Change type

- [x] `bugfix`

### Test plan

1. Open the reactive inputs example and subscribe to `editor.inputs.getMetaKey()`.
2. Press the Meta (Cmd) key — value reactively changes to `true`.
3. Release the Meta key — value now changes to `false` immediately, without moving the pointer.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug where the reactive Meta (Cmd) key state was not cleared on keyup until a pointer event, affecting `getMetaKey()` and `getAccelKey()` on macOS.

Closes #7981

### DIFF
--- a/apps/examples/src/examples/editor-api/reactive-inputs/ReactiveInputsExample.tsx
+++ b/apps/examples/src/examples/editor-api/reactive-inputs/ReactiveInputsExample.tsx
@@ -67,7 +67,7 @@ function ReactiveInputsPanel() {
 	const accelKey = useValue('accel key', () => editor.inputs.getAccelKey(), [editor])
 
 	return (
-		<div className="reactive-inputs-panel">
+		<div className="tlui-menu reactive-inputs-panel">
 			<div className="reactive-inputs-content">
 				{/* [5] */}
 				<div className="input-group">
--- a/apps/examples/src/examples/editor-api/reactive-inputs/reactive-inputs.css
+++ b/apps/examples/src/examples/editor-api/reactive-inputs/reactive-inputs.css
@@ -1,35 +1,35 @@
+/* tlui-menu provides the native panel background, radius, and shadow. */
 .reactive-inputs-panel {
-	padding: 6px;
-	border: 1px solid #ccc;
-	box-shadow: none;
+	margin: 8px;
+	padding: var(--tl-space-2);
 	pointer-events: all;
 }
 
 .reactive-inputs-content {
 	display: flex;
 	flex-wrap: wrap;
-	gap: 6px;
+	gap: var(--tl-space-2);
 	align-items: center;
 }
 
 .input-group {
 	display: flex;
 	align-items: baseline;
-	gap: 4px;
-	padding: 2px 8px;
-	background: #fff;
-	border: 1px solid #ccc;
+	gap: var(--tl-space-1);
+	padding: var(--tl-space-1) var(--tl-space-3);
+	background: var(--tl-color-muted-2);
+	border-radius: var(--tl-radius-2);
 }
 
 .input-label {
 	font-size: 12px;
-	color: #333;
+	color: var(--tl-color-text);
 	white-space: nowrap;
 }
 
 .input-value {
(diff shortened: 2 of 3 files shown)


kaneel:
I'm wondering if the meta key staying sticky is only a result of the fact the ctrlKey is actually thrown at the same time as the metaKey.

I only pressed meta key and I received these events when commenting after the whole conditions block
image

However, this fix works as demonstrated in this screenshot

steveruizok:
> I'm wondering if the meta key staying sticky is only a result of the fact the ctrlKey is actually thrown at the same time as the metaKey.
> 
> I only pressed meta key and I received these events when commenting after the whole conditions block image
> 
> However, this fix works as demonstrated in this screenshot

chase it!

PR #9406, opened 2026-06-24
test(driver): carry held modifiers on simulated keyboard events

kaneel:
In order to make the test Driver behave like real DOM input, this PR makes simulated keyboard events carry every currently-held modifier instead of just the key being pressed.

Previously `getKeyboardEventInfo` built `keydown`/`keyrepeat` flags from only the pressed key, so pressing Shift while Control was held would report `ctrlKey: false`. That was inconsistent with two things:

- `keyUp`, which already derives its flags from the editor's held modifier state.
- Real DOM keyboard events, where every event reflects the full set of currently-held modifiers.

That gap meant a test holding two modifiers (e.g. `keyDown('Control')` then `keyDown('Shift')`) didn't actually model "both held", which is easy to trip over when reasoning about modifier behaviour.

We OR in the editor's current modifier state for the four modifier flags (and derive `accelKey` from the merged result). `keyUp` passes explicit flag overrides, so its release semantics are unchanged.

### Change type

- [x] `other` (test infrastructure)

### Test plan

- [x] Unit tests

Full editor (830) and tldraw (2648) suites pass unchanged.

### DIFF
--- a/packages/driver/src/lib/Driver.ts
+++ b/packages/driver/src/lib/Driver.ts
@@ -255,13 +255,20 @@ export class Driver {
 		name: TLKeyboardEventInfo['name'],
 		options = {} as Partial<Omit<TLKeyboardEventInfo, 'point'>>
 	): TLKeyboardEventInfo {
-		return {
-			shiftKey: key === 'Shift',
-			ctrlKey: key === 'Control' || key === 'Meta',
-			altKey: key === 'Alt',
-			metaKey: key === 'Meta',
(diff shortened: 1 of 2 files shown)


kaneel:
I'm putting this in draft, feel free to discuss; it's not really bound to any issues, this is something I found during a research on some keyboard stuff, I find that intriguing that we rely on this Driver that is not entirely correct in behaviour and may in the future leave us writing some tests that are wrong

PR #9407, opened 2026-06-24
feat(tldraw): add NoteShapeAttribution component override

mimecuvalo:
Spike exploring an alternative to #9252. Instead of a `showAttribution` boolean on `NoteShapeUtil`, this exposes the note's "last edited by" attribution badge as an overridable component through a `NoteShapeUtil` option.

```tsx
import { Tldraw, NoteShapeUtil } from 'tldraw'

// hide the badge (equivalent to showAttribution: false)
<Tldraw shapeUtils={[NoteShapeUtil.configure({ AttributionComponent: null })]} />

// or render your own
<Tldraw shapeUtils={[NoteShapeUtil.configure({ AttributionComponent: (props) => <MyBadge {...props} /> })]} />
```

This is strictly more powerful than a boolean: `null` hides it, and a custom component fully restyles or replaces the badge without subclassing the util. The override applies both on the canvas and in image exports.

### How it works

- Adds an `AttributionComponent` option to `NoteShapeOptions`, defaulting to a new `DefaultNoteShapeAttribution` component (with `TLNoteShapeAttributionProps`). The default renders the existing badge.
- `NoteShapeUtil.component()` (canvas) and `toSvg()` (export) both read `this.options.AttributionComponent`.
- Because options live on the util instance, canvas and export share the same source — no bridge is needed between them.

### Why an option instead of a `TLUiComponents` slot

An earlier iteration put the override in `TLUiComponents`. That worked on the canvas but needed an extra mechanism for exports: `toSvg()` renders in a separate React root with no access to the UI-components context, so the resolved override had to be mirrored onto the editor (a per-editor `WeakMap` + a `useLayoutEffect` in `Tldraw.tsx`) and read imperatively.

Moving the override to a `NoteShapeUtil` option removes that entirely. `this.options` is available identically in `component()` and `toSvg()`, so the WeakMap and the mirror effect are gone. Attribution is note-shape rendering (it's drawn into the exported SVG, not DOM chrome), so the util option is also the more natural home — it sits alongside `resizeMode` and matches how other shape/asset utils are configured (`EmbedShapeUtil.configure`, `ImageAssetUtil.configure`, …).

The one thing the option gives up is live runtime swapping via React context; `.configure` is set once at mount, which is the right shape for a hide/restyle config.

### API changes

New public API from `tldraw`:

- `NoteShapeOptions.AttributionComponent: TLNoteShapeAttributionComponent` — new option on the note shape util (`null` hides the badge; defaults to `DefaultNoteShapeAttribution`). Configure with `NoteShapeUtil.configure({ AttributionComponent })`.
- `DefaultNoteShapeAttribution(props: TLNoteShapeAttributionProps): JSX.Element` — the default badge renderer.
- `TLNoteShapeAttributionProps` — props passed to the component:
  ```ts
  interface TLNoteShapeAttributionProps {
  	shape: TLNoteShape
  	name: string
  	firstName: string
  	color: string
  	scale: number
  	variant: 'canvas' | 'export'
  }
  ```
- `TLNoteShapeAttributionComponent` — `ComponentType<TLNoteShapeAttributionProps> | null` type alias for the option.

No change to `TLUiComponents`. No breaking changes; all additions are opt-in.

### Change type

- [x] `api`

### Test plan

1. Add a note and type into it while signed in / in multiplayer so attribution appears.
2. By default the badge shows in the corner and in SVG/PNG exports.
3. `NoteShapeUtil.configure({ AttributionComponent: null })` hides it on canvas and in exports.
4. A custom `AttributionComponent` renders in its place on canvas and in exports.

- [x] Unit tests (canvas render + SVG export, default / custom / null)

### Release notes

- Add a `NoteShapeUtil` `AttributionComponent` option to hide or replace the note shape's text attribution badge.


mimecuvalo:
had an offline chat, ready to go

PR #9411, opened 2026-06-24
refactor(dotcom): remove the groups_frontend flag

jsscclr:
In order to finish the workspaces cleanup, this PR completely removes the `groups_frontend` feature flag and its rollout machinery. Workspaces is fully shipped: `groups_backend` (the data migration) is at 100% and granted to every user on signup, and `groups_frontend` (the UI flag) recently reached 100%. With #9405 removing the last UI fork that read the flag, nothing depends on it anymore. Relates to #8859.

**Preview deploy:** https://pr-9411-preview-deploy.tldraw.com/ — verify the admin page (groups-enrollment section + rollout slider gone, other sections intact) and that signup / invite-acceptance still work. No flag is needed; every account already has `groups_backend`.

**Stacked on #9405** — that PR's branch is this one's base, so the diff below is only the flag removal. Merge #9405 first; keeping this as a draft until then.

- Drop the `groups_frontend` member from the `TlaFlags` enum.
- Delete the workspaces UI rollout utility (`workspacesUiRollout.ts`) and its tests.
- Remove the admin groups-enrollment section and the rollout-percentage slider, plus the `enroll_groups` / `unenroll_groups` / rollout-SSE routes and the `admin_enrollInGroups` / `admin_unenrollFromGroups` durable-object methods.
- Stop granting `groups_frontend` on user creation and invite acceptance.
- Update the flag-parsing tests and the e2e fixture to drop the retired flag.

The backend migration tooling is untouched: `groups_backend` stays in the enum, and the `admin_migrateToGroups` route/method remain for manually migrating a user's data if needed.

No data migration is required. The `user.flags` column is free-form text (not enum-validated), so leftover `groups_frontend` tokens on existing rows are simply ignored once nothing reads them.

### Change type

- [x] `other` (internal refactor, no user-facing change)

### Test plan

1. On the preview deploy, sign in as a normal user — confirm the app loads and the workspaces sidebar works (file list, switcher, "Move to", drag).
2. Accept a workspace invite — confirm it still works (the `groups_backend` check remains; only the redundant `groups_frontend` grant is gone).
3. Open the admin page — confirm it loads with the groups-enrollment section and rollout slider gone, and the remaining sections (feature flags, welcome template, delete user) intact.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +2 / -2    |
| Tests     | +4 / -152  |
| Apps      | +4 / -929  |

PR #9418, opened 2026-06-25
refactor(dotcom): remove the groups_backend flag from production

jsscclr:
This is the next step in retiring the workspaces feature flags. It removes `groups_backend` from production entirely — the logic that read it, the value written to new users, and the now-dead migration tooling. `groups_backend` is fully rolled out (every user is migrated and born with a home workspace), so the legacy (`file_state`/`ownerId`-based) branches it gated are dead, and new users no longer need it. Follows #9405 and #9411 (the `groups_frontend` removal), both merged. Relates to #8859.

**Preview deploy:** https://pr-9418-preview-deploy.tldraw.com/ — sign in and exercise file/workspace flows (create, pin, open, remove, move, accept invite) and the admin page. No flag setup needed; every account is migrated.

### What changed

- Removed `isWorkspacesMigrated()` and collapsed every branch on it to the migrated path: `mutators.ts` (`createFile`, `pinFile`/`unpinFile`, `removeFileFromWorkspace`, `onEnterFile`) and `TldrawApp.ts` (`getMyFiles`, `canCreateNewFile`, `isPinned`, `getFile`).
- Dropped the `assertUserHasFlag('groups_backend')` gates on the workspace mutators.
- Removed the admin migrate-to-groups tooling (`/app/admin/user/migrate` route + the `admin_migrateToGroups` durable-object method) and the `acceptInvite` migration check.
- Stopped writing `groups_backend` at the three user-creation sites; new users get `flags: ''`.
- Removed the redundant e2e groups-migration helpers (`migrateUser*`, `isUserMigrated*`, `ensureGroupsReadyByEmail`) and their call sites. Test users are born migrated via signup, so these were no-ops; the `ensureGroupsReady` scenario helper keeps its navigation/sidebar setup, only its migration call was dropped.
- Demoted `TlaFlags` to a `string` stub, keeping `parseFlags`/`userHasFlag`/`hasFlag` so the next feature flag has scaffolding.

### What's kept (deliberately)

- The `flags` column, the `TlaFlags` string stub, and `parseFlags`/`userHasFlag`/`hasFlag` — retained so the next feature flag has scaffolding (no schema migration needed).
- The `migrate_user_to_groups` SQL function. It lives in an applied (immutable) migration and is now uncalled, so it stays as frozen history.
- Dropping the `flags` column itself (a Postgres + Zero schema change) is intentionally out of scope — we're keeping the column for future flags.

### Decisions worth a reviewer's eye

- **Access control is intentionally untouched.** `assertUserCanAccessFile` / `canUpdateFile` branch on the *file's* shape (`ownerId` vs `owningGroupId`), not on the flag, so they're kept as defensive fallbacks for any legacy `ownerId` files. We stop *producing* legacy files but keep *tolerating* them.
- **New users get `flags: ''`.** Existing users' stored `groups_backend` value is harmless — nothing reads it after this PR.
- **`assertNotMaxFiles` removed.** This server-side `createFile` limit only ran on the legacy path and was already inert for migrated users; the client-side limit in `canCreateNewFile` remains. If a server-enforced per-workspace cap is wanted, that's a separate feature.

### Change type

- [x] `other` (internal refactor, no user-facing change)

### Test plan

1. On the preview deploy, sign in and confirm normal file/workspace flows work: create, pin/unpin, open, remove, move between workspaces, accept an invite.
2. Confirm the admin page loads with the "migrate user" action gone (other sections intact).

- [x] Unit tests
- [x] End to end tests

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +45 / -171 |
| Tests     | +6 / -81   |
| Apps      | +22 / -174 |

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/scenario-test.ts
+++ b/apps/dotcom/client/e2e/fixtures/scenario-test.ts
@@ -508,8 +508,8 @@ class DotcomScenario {
 			.slice(0, MAX_WORKSPACE_NAME_LENGTH)
 		const fileName = opts.fileName ?? this.name('workspace file')
 
-		await this.ensureGroupsReady(opts.owner)
-		await this.ensureGroupsReady(opts.member)
+		await this.goToAndOpenSidebar(opts.owner)
+		await this.goToAndOpenSidebar(opts.member)
 
 		if (!opts.member.email) throw new Error('Workspace member actor is not signed in')
 		const memberUserId = await this.database.getUserIdByEmail(opts.member.email)
@@ -596,7 +596,7 @@ class DotcomScenario {
 		)
 	}
 
-	async ensureGroupsReady(actor: DotcomActor) {
+	async goToAndOpenSidebar(actor: DotcomActor) {
 		if (!actor.email) throw new Error(`Actor ${actor.name} is not signed in`)
 
 		await actor.goto()
--- a/apps/dotcom/client/e2e/fixtures/scenario-test.ts
+++ b/apps/dotcom/client/e2e/fixtures/scenario-test.ts
@@ -599,7 +599,6 @@ class DotcomScenario {
 	async ensureGroupsReady(actor: DotcomActor) {
 		if (!actor.email) throw new Error(`Actor ${actor.name} is not signed in`)
 
-		await this.database.ensureGroupsReadyByEmail(actor.email)
 		await actor.goto()
 		await actor.editor.ensureSidebarOpen()
 	}
--- a/packages/dotcom-shared/src/mutators.test.ts
+++ b/packages/dotcom-shared/src/mutators.test.ts
@@ -1512,9 +1512,9 @@ describe('createFile from source (duplicate) access control', () => {
 		await expectBadRequest(() => duplicate(m, tx, `${FILE_PREFIX}/${sourceId}`))
 	})
 
-	it('non-migrated user cannot duplicate an inaccessible legacy file', async () => {
+	it('cannot duplicate an inaccessible legacy (ownerId-owned) file', async () => {
 		const s = {
-			user: [makeUser({ id: userId, flags: '' })],
+			user: [makeUser({ id: userId, flags: 'groups_backend' })],
 			file: [makeFile({ id: sourceId, ownerId: 'user_someoneElse1234', shared: false })],
 			file_state: [],
 			group: [],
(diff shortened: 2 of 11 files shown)


packages/dotcom-shared/src/mutators.test.ts
```diff
@@ -1512,9 +1512,9 @@ describe('createFile from source (duplicate) access control', () => {
 		await expectBadRequest(() => duplicate(m, tx, `${FILE_PREFIX}/${sourceId}`))
 	})
 
-	it('non-migrated user cannot duplicate an inaccessible legacy file', async () => {
+	it('cannot duplicate an inaccessible legacy (ownerId-owned) file', async () => {
 		const s = {
-			user: [makeUser({ id: userId, flags: '' })],
+			user: [makeUser({ id: userId, flags: 'groups_backend' })],
```
frolic: ooc why is the `groups_backend` flag here still needed?

frolic (commented):
lgtm but will let mitja sign off

packages/dotcom-shared/src/mutators.test.ts
jsscclr: Was being overly cautious, but think we can rip the bandaid off - watch this space (branch)

packages/dotcom-shared/src/mutators.test.ts
MitjaBezensek, line 1: We might clean up some of the `groups_backend` here?

apps/dotcom/client/e2e/fixtures/scenario-test.ts
```diff
@@ -599,7 +599,6 @@ class DotcomScenario {
 	async ensureGroupsReady(actor: DotcomActor) {
```
MitjaBezensek: Should probably rename this, seems like a simple navigation + opening the sidebar now.

MitjaBezensek (approved):
Looks good, left a few comments about some additional clean up 💅

I guess we'll keep `userHasFlag`, `parseFlags`, `getUserFlags`, `hasFlag` around just in case we need them in the future?

We also have some stale ts docs like `getHomeWorkspaceId`, think we can remove the `For migrated user` part.

jsscclr:
Actioned! And yep, I intentionally left the flags infrastructure for now. Guess it depends on if we see flagging features common or uncommon in the future - usually I wouldn't want to keep code around 'just in case', but I think the flags feature has worked well, and if we remove it then we'd have a bit of friction to adopt flags again.

PR #9420, opened 2026-06-25
refactor(editor): inline coarse-pointer long-press guard, drop helper

jsscclr:
In order to keep `@tldraw/editor`'s internal surface honest, this PR removes the `cancelShapeCreationOnLongPress` helper introduced in #8948 and inlines its guard at each call site, following up on review feedback.

The helper's entire payload was one guard — `if (isCoarsePointer) cancelPendingCreation()`. Because each shape-creation tool already owns a `cancel()` method that does the tool-specific cleanup (`bailToMark`, transition to idle), the helper couldn't do the work itself: it had to take a callback and trampoline back into the tool. So the call site `cancelShapeCreationOnLongPress(this.editor, () => this.cancel())` was longer and more indirect than the inlined guard, and a guard-plus-callback didn't earn an `@internal` export on the package's API surface.

Each of the six discrete shape-creation tools (geo, note, line, text, arrow, and the base box tool) now reads:

```ts
override onLongPress() {
	// On a touch (coarse pointer) long-press, cancel the pending shape so it leaves nothing behind.
	if (this.editor.getInstanceState().isCoarsePointer) this.cancel()
}
```

The coarse-pointer rationale — touch is where the browser fires `contextmenu` at the long-press mark, which is the bug behind #8277 — is preserved as a short comment at each site. Behavior is unchanged; this is a pure cleanup.

### Change type

- [x] `improvement`

### API changes

- Removed the `@internal` `cancelShapeCreationOnLongPress(editor, cancelPendingCreation)` export from `@tldraw/editor`.

### Test plan

- Pure refactor with no behavior change.
- `yarn typecheck` and `yarn api-check` pass (the report drops the removed `@internal` symbol).
- Existing `long-press-create` and `right-click-context-menu` suites pass (28/28), covering the inlined guard at every call site.

- [x] Unit tests

### Code changes

| Core code       | +13 / -50  |
| Automated files | +0 / -3    |


steveruizok:
I don't think this should be a helper at all tbh, it's just a guard around isCoarsePointer?

steveruizok:
I'd remove `cancelShapeCreationOnLongPress` entirely and just inline those. A guard + a callback doesn't an API feel like methinks

jsscclr:
@steveruizok inlined!

PR #9425, opened 2026-06-25
feat(editor): spike - add way to find and delete unused assets that are lingering  after soft deletion

angrycaptain19:
Had some recent customer feedback about the risk of boards where assets have been deleted growing very large and wanting a way to properly delete the lingering assets to keep board shape small. 

I tested and confirmed this by creating boards with dozens of images and then deleting them - this left boards with no visible images that were still 10+ mb.

This spike adds two public Editor methods:

- getUnusedAssetIds(): gets the asset ids not referenced by any shape on any page.
- pruneUnusedAssets(): delete those assets (also calling an external asset store's remove() to free the underlying file) and return the removed ids.

Also adds in a prune unused assets example to show how this would work in practice

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5120,6 +5120,11 @@ export class Editor extends EventEmitter<TLEventMap> {
 		return this.store.query.records('asset')
 	}
 
+	/** @internal */
+	@computed private _getAllShapesQuery() {
+		return this.store.query.records('shape')
+	}
+
 	/**
 	 * Get all assets in the editor.
 	 *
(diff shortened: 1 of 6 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5208,6 +5213,48 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+	 * This is not undoable, so only prune where a removed asset can't be needed
+	 * again — typically just after loading or before saving, and not while an
+	 * upload is in flight, a deletion is still undoable, or peers share the
+	 * document.
+	 *
+	 * @returns The ids of the removed assets.
+	 *
+	 * @public
+	 */
+	pruneUnusedAssets(): TLAssetId[] {
```
mimecuvalo, line 5251: i'm wondering if we actually need this one or if just pointing people to combine getUnusedAssetIds+deleteAssets is enough ¯\\_(ツ)_/¯

angrycaptain19: Yeah I think combined is probably easier?

mimecuvalo (approved):
LGTM

PR #9429, opened 2026-06-26
fix(arrows): keep frame-bound arrows in the frame and grabbable when clipped

kaneel:
In order to stop arrows being orphaned when their frame is deleted (#8978) and to let you move an arrow where it bows outside its frame, this PR fixes how arrows bound to a frame are parented and hit-tested.

Two issues:

1. When an arrow's endpoints both resolve to a frame (both ends on the frame itself, or one end on the frame and the other on a child inside it), the arrow was parented to the page instead of the frame, so deleting the frame left the arrow behind. `reparentArrow` now parents the arrow into the bound shape when one binding target is an ancestor-or-self of the other (and can contain the arrow), via a new `getCommonBindingContainer` helper. The existing `findCommonAncestor` path (distinct children, groups) is unchanged.

2. A curved arrow that's a child of a frame is clipped at the frame's edge, but its selection indicator is drawn unclipped — so you could see the part sticking out but not grab it. `getSelectedShapeAtPoint` now hit-tests geometry directly (ignoring the clip mask) using the standard hit-test margin, so a selected shape is grabbable across its whole geometry. Pointing down outside the frame on a selected arrow now starts a move instead of a brush.

### Change type

- [x] `bugfix`

### Test plan

1. Draw a frame, then draw an arrow between two empty areas inside it (both ends bind to the frame).
2. Delete the frame — the arrow is deleted with it.
3. Draw another such arrow and bend it so it bows outside the frame. Select it, then drag from the part outside the frame — it moves.

- [x] Unit tests

### API changes

- ShapeUtil.shouldClipChild now receives a second, optional self argument — the clipping ancestor being asked.

### Release notes

- Fix arrows bound to a frame being left behind when the frame is deleted.
- Fix being unable to select and move an arrow where it extends outside its frame.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +58 / -3   |
| Tests     | +69 / -4   |**

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -3103,7 +3103,7 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
     static props?: RecordProps<TLUnknownShape>;
     // @internal
     providesBackgroundForChildren(shape: Shape): boolean;
-    shouldClipChild?(child: TLShape): boolean;
+    shouldClipChild?(child: TLShape, self?: Shape): boolean;
     toBackgroundSvg?(shape: Shape, ctx: SvgExportContext): null | Promise<null | ReactElement> | ReactElement;
     toSvg?(shape: Shape, ctx: SvgExportContext): null | Promise<null | ReactElement> | ReactElement;
     static type: string;
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5505,7 +5505,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 				const util = this.getShapeUtil(ancestor)
 				const clipPath = util.getClipPath?.(ancestor)
 				if (!clipPath) continue
-				if (util.shouldClipChild?.(shape) === false) continue
+				if (util.shouldClipChild?.(shape, ancestor) === false) continue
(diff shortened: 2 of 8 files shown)


steveruizok:
Before and after screenshots etc would be super useful here!

kaneel:
The issue occurs right now when drawing an arrow on top of a frame, the arrow will not be parented to the frame and this will cause a bit of a kerfuffle when we delete a frame and see the arrow is still here:

https://github.com/user-attachments/assets/006a08b7-5551-4a8c-a4b4-48288349facb

This PR addresses this issue by parenting the arrow to the frame:

https://github.com/user-attachments/assets/c976ba1c-d8a1-44ba-a188-ab6b8e3d27d9

This cause a change in how we display the arrow, if it's parented to the shape then it's not selectable as it's clipped, this PR addresses the issue too

steveruizok:
A frame should not clip an arrow that is bound to it on both sides. We'll need to have a special case for this, not sure whether to add it to the frame, to the arrow, or where. This is something new that we haven't designed for before.

image

kaneel:
I think I got a solution for this 
image

steveruizok:
Closing this out because the changes have been split out and landed in the newer arrow PRs: the frame-bound arrow parenting and frame unclipping behavior went in with #9470, and the remaining selected clipped-arrow dragging fix went in with #9479.

PR #9435, opened 2026-06-26
feature(examples): midi sequencer

kaneel:
Some fun, don't take seriously, it's missing "few stuff" in order to be perfect

1. fire garageband, or any virtual synth
2. find the midi port for it (you can rescan if it did not pop up)
3. create a sequence (toolbar has it)
4. draw note
5. press spacebar

BUT ALSO

chains! chains are frame-like groups for sequences that can have basic follow up actions

AND

you can send to different midi channels! FIRE MORE SYNTHS

ALSO

scroll over a note to set the "rachet"
shift+scroll to change the probability for the note

LAST BUT NOT LEAST

STEPPER MODE
Listen to note on/off from a chain -> step your sequence

https://github.com/user-attachments/assets/29d923c9-4d08-4dc7-834c-d734ff7707e0

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape...
7.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug with…

### DIFF
--- a/apps/examples/src/examples/use-cases/midi-sequencer/MidiSequencerExample.tsx
+++ b/apps/examples/src/examples/use-cases/midi-sequencer/MidiSequencerExample.tsx
@@ -1,0 +1,205 @@
+import { ReactNode, useRef, useState } from 'react'
+import { createPortal } from 'react-dom'
+import {
+	DefaultToolbar,
+	DefaultToolbarContent,
+	TLComponents,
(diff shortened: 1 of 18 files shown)


mimecuvalo:
this is super cool 😎

PR #9442, opened 2026-06-28
experiment(perf): move freehand ink generation to Rust/WASM

steveruizok:
In order to take freehand ink generation off the JS hot path, this PR reimplements the freehand algorithm in Rust compiled to WebAssembly and makes it the sole implementation. The draw, highlight, and scribble render and hit-test paths now generate their geometry and SVG paths in WASM. Output is byte-for-byte identical to the previous JS implementation (verified against it as a test oracle), and it runs ~1.8–2x faster in Node/V8.

Ink generation is one of the heaviest non-React CPU costs in a stroke-heavy document: re-rendering every visible stroke when a zoom crosses the `forceSolid` threshold, on load, and on export. A 2,000-stroke document spends ~20ms regenerating all draw ink in JS — over a frame budget — before any DOM work.

### What moved to WASM

The Rust crate lives in `packages/editor/freehand-wasm` (it has to sit in `@tldraw/editor` so the editor-level `getSvgPathFromPoints` can use it). It ships as a `freehand.wasm` asset and is loaded at runtime via `new URL('./freehand.wasm', import.meta.url)` — fetched in the browser, read from disk in Node. The TypeScript algorithm is removed from the shipped SDK; a verbatim copy is kept under `freehand/oracle/` purely as a test oracle (not exported, not bundled).

| Function | Notes |
| --- | --- |
| `svgInk` | draw shape `dash: 'draw'` ink render |
| `getStrokePoints` / `getSvgPathFromStrokePoints` | solid / fill centerline |
| `getStroke` / `getStrokeOutlinePoints` | outline polygon, with taper + caps |
| `getSvgPathFromPoints` | scribble path formatting |

Public signatures are unchanged (non-breaking). `computeRadii` supports taper, and custom (non-built-in) easings are sampled into a lookup table so the public API stays fully functional.

### Async loading

Freehand generation runs synchronously during render, but a `.wasm` asset loads asynchronously. The load is kicked off eagerly at import, and a reactive readiness signal lets ink and geometry re-render once the module is ready — the same way fonts swap in when they finish loading. In Node/SSR the asset is read synchronously from disk.

### Change type

- [x] `improvement`

### Test plan

1. Draw freehand strokes with the draw tool; confirm ink renders identically (round caps, pressure, tapering).
2. Switch to the highlighter and draw; confirm highlight strokes render and hit-test.
3. Draw with a stylus/pen (real pressure) and with a mouse (simulated pressure).
4. Zoom out past the point where strokes become solid; confirm they re-render.
5. Select a stroke and confirm the selection indicator and bounds are correct.
6. Export a drawing to SVG/PNG; confirm ink is present and correct.
7. Trigger a scribble (e.g. eraser or selection scribble) and confirm it renders.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Freehand ink generation (draw, highlight, and scribble) is now implemented in Rust/WebAssembly for better performance on stroke-heavy documents.

### API changes

- Added `loadFreehandWasm()` and `instantiateFreehandWasm()` to `@tldraw/editor` (internal)

### Code changes

| Core code       | +734 / -848   |
| Rust crate      | +1865 / -0    |
| Tests           | +1407 / -1    |
| Automated files | +79 / -0      |
| Config/tooling  | +31 / -1      |


kostyafarber:
if we are experimenting with wasm off of hot paths i'd also want to explore if we could get text layout out of the js path too

steveruizok:
This one isn't materially faster. :( I think we're fast enough with our current ink.

PR #9443, opened 2026-06-29
perf(state): cache parent change checks

steveruizok:
In order to reduce repeated reactive graph validation during large-board interactions, this PR caches parent-change checks inside `@tldraw/state` for the current global epoch.

### What we found

Moving a large selection on a large board was spending a meaningful amount of time re-validating the same reactive parent relationships. In the traces we looked at, `haveParentsChanged` repeatedly walked parent lists and called `__unsafe__getWithoutCapture(true)` even when another check in the same global epoch had already established the answer for that child.

On the seeded 960-shape board, the scripted browser comparison showed the same pattern. The branch reduced move sync work from 200.7 ms total on main to 65.4 ms total, and reduced total sync work across select, move, create, and delete from 240.0 ms to 76.3 ms. In the windowed DevTools trace, `haveParentsChanged` dropped from 127.2 ms to 30.7 ms, and `__unsafe__getWithoutCapture` dropped from 135.5 ms to 33.8 ms.

### What we changed

`haveParentsChanged` now accepts the current global epoch. When a child has already checked its parents during that epoch, it reuses the cached boolean result instead of walking the same parents again.

The helper also checks each parent's `lastChangedEpoch` before forcing an uncaptured parent read. If the epoch already differs from the child's recorded parent epoch, the parent is definitely changed, so we can return without forcing `__unsafe__getWithoutCapture(true)`.

`Computed` and `EffectScheduler` now pass a stable `globalEpoch` into `haveParentsChanged` for their current pass. The capture lifecycle clears the cached parent-check state whenever a child starts or stops recapturing dependencies, so dependency changes do not inherit stale parent-check results.

The state helper tests cover both new correctness paths: early exit when a parent's epoch already proves it changed, and same-epoch memoization for repeated parent checks.

### What you need to know

This is intended to be an internal performance change. It should not change public state semantics: parent checks are cached only per child, only for the current global epoch, and are cleared when dependencies are recaptured.

The main tradeoff is two optional bookkeeping fields on `Child`: `lastParentCheckEpoch` and `lastParentCheckResult`. They show up in the API report because `Child` is part of the package surface, but they are internal implementation details for parent-check memoization.

The browser comparison restored both seeded tabs afterward: 960 shapes, 160 selected shapes, no temporary test shapes left behind, and no console errors.

### Change type

- [x] `improvement`

### Test plan

1. `cd packages/state && yarn test run`
2. `cd packages/state && yarn lint`
3. `cd packages/state && yarn build && yarn build-api`
4. Select and move shapes on a large board, then compare trace hot paths for `haveParentsChanged` and `__unsafe__getWithoutCapture`.

- [x] Unit tests
- [ ] End to end tests

### API changes

- Added internal `Child.lastParentCheckEpoch` and `Child.lastParentCheckResult` fields for parent-check memoization.

### Code changes

| Core code       | +50 / -9   |
| Tests           | +31 / -0   |

### Release notes

- Improve large-board interaction performance by reducing repeated reactive parent validation.

### DIFF
--- a/packages/state/api-report.api.md
+++ b/packages/state/api-report.api.md
@@ -39,6 +39,8 @@ export interface AtomOptions<Value, Diff> {
 export interface Child {
     __debug_ancestor_epochs__: Map<Signal<any, any>, number> | null;
     isActivelyListening: boolean;
+    lastParentCheckEpoch?: number;
+    lastParentCheckResult?: boolean;
     lastTraversedEpoch: number;
     readonly name: string;
     readonly parentEpochs: number[];
--- a/packages/state/src/lib/Computed.ts
+++ b/packages/state/src/lib/Computed.ts
@@ -275,7 +275,7 @@ class __UNSAFE__Computed<Value, Diff = unknown> implements Computed<Value, Diff>
 				(this.isActivelyListening &&
 					getIsReacting() &&
 					this.lastTraversedEpoch < getReactionEpoch()) ||
-				!haveParentsChanged(this))
+				!haveParentsChanged(this, globalEpoch))
 		) {
 			this.lastCheckedEpoch = globalEpoch
 			if (this.error) {
--- a/packages/state/src/lib/EffectScheduler.ts
+++ b/packages/state/src/lib/EffectScheduler.ts
@@ -92,12 +92,13 @@ class __EffectScheduler__<Result> implements EffectScheduler<Result> {
 	maybeScheduleEffect() {
 		// bail out if we have been cancelled by another effect
 		if (!this._isActivelyListening) return
+		const globalEpoch = getGlobalEpoch()
 		// bail out if no atoms have changed since the last time we ran this effect
-		if (this.lastReactedEpoch === getGlobalEpoch()) return
+		if (this.lastReactedEpoch === globalEpoch) return
 
 		// bail out if we have parents and they have not changed since last time
-		if (this.parents.length && !haveParentsChanged(this)) {
-			this.lastReactedEpoch = getGlobalEpoch()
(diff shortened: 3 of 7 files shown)


steveruizok:
### Manual browser comparison

I compared seeded `develop` tabs with the same 960-shape board: canary/main at `examples-canary.tldraw.com` and this branch at `examples-git-perf-state-parent-check-cache.tldraw.club`.

Scripted operation sequence:

1. Select 160 existing shapes.
2. Move those shapes 6 times.
3. Create 60 temporary geo shapes.
4. Delete the temporary geo shapes.
5. Restore the original selection and positions with `history: 'ignore'`.

Both tabs restored cleanly afterward: 960 shapes, 160 selected shapes, 0 temporary `shape:codex_perf_*` shapes, and no console errors.

| Metric | Main | This branch |
| --- | ---: | ---: |
| Move sync avg | 33.4 ms | 10.9 ms |
| Move sync total | 200.7 ms | 65.4 ms |
| Create sync | 33.5 ms | 6.8 ms |
| Delete sync | 5.7 ms | 4.0 ms |
| Total sync work | 240.0 ms | 76.3 ms |

Windowed DevTools trace over the scripted interaction:

| Hotspot | Main total | This branch total |
| --- | ---: | ---: |
| `haveParentsChanged` | 127.2 ms | 30.7 ms |
| `__unsafe__getWithoutCapture` | 135.5 ms | 33.8 ms |

The trace still has browser/GC noise, but the direct in-page timings and sampled hotspot movement agree: the branch cuts the intended state parent-check overhead substantially during large-board interactions.

steveruizok:
A good test for this is:

- on a page with many, many shapes, drag a single shape

This is a small gain, not sure it's worth landing.

Issue #9196, opened 2026-06-15
Establish a typography system for the dotcom client


jsscclr:
Part of #9189.

Typography is the weakest part of the dotcom design system. A type scale exists, but **only as CSS utility classes** in `tla/styles/tla.css` — `.tla-text_ui__tiny|small|medium|regular|section|title|big` — with **no `--tla-text-*` tokens and no text component**. The only font token is `--tla-font-ui` (the family).

The consequence is structural: styling done inside a `.module.css` file cannot reference the scale, so it hardcodes `font-size`/`font-weight`/`line-height` by hand (~94 declarations).

## Decision needed first

The canonical mechanism doesn't fully exist yet. Decide one of:

1. Add `--tla-text-*` (size/weight) tokens so CSS modules can reference the scale, then migrate.
2. Build a `TlaText` component that applies the existing classes, and route JSX text through it.
3. A mix: tokens for CSS-module use, component for JSX use.

Because this defines a foundational part of the design system (unlike the mechanical sibling sweeps), it likely warrants a quick design / front-end-lead sign-off on the mechanism before migration.

## Audit of the current state

The hardcoded sizes sort into three buckets (sizes confirmed against the scale: `small=11`, `medium/section=12`, `regular/title=14`, `big=24`):

1. **Drift — a scale rung already matches.** `12px` (~21 uses), `14px` (~15), `11px` (~6) all map onto existing entries. ~42 of ~48 hardcoded sizes. Mechanical once a CSS-referenceable mechanism exists.
2. **Off-scale — gap in the scale.** `16px` (~5 uses) and `20px` (~1) match no rung; the scale jumps `14px → 24px`. Decide whether to add a rung or snap. Same story for weights: `600` (~11 uses) isn't in the scale's weight set `{500, 700, 800}`, and `bold` (~2) is the keyword spelling of `700` — pick one expression.
3. **Dead.** `.tla-text_ui__tiny` (6px) is defined but used in **zero** files. Delete it (6px is effectively illegible).

Note the near-duplicate rungs: `medium`/`section` are both 12px and `regular`/`title` both 14px, differing only by weight/color. That may be fine as semantic aliases, but confirm it's intentional when defining the token names.

Full list: `grep -rnE "(font-size|font-weight|line-height):" apps/dotcom/client/src --include="*.module.css"`

## Acceptance

- A canonical typography mechanism is decided and documented (tokens, component, or both).
[...]

PR #9449, opened 2026-06-29
spike(dotcom): design-system gallery at /dev/components (lay of the land)

jsscclr:
## What this is

A spike to get a **lay of the land** of the dotcom design system — *not* for merge. It adds a dev-only gallery at `/dev/components` that renders (almost) every shared UI component the app uses, **live**, so we can see what exists and where it diverges before deciding what a real design system should be. Exploratory groundwork for #9189 — happy to keep it, fold it into something durable, or close it.

## How to view

Preview: <https://pr-9449-preview-deploy.tldraw.com/dev/components/buttons> — use the nav at the top of each page to walk the families. (Gated to dev + preview deploys; it does **not** appear in production.)

## What's in it

16 live pages — **families** (buttons, inputs, typography, dialogs, menus, icons, links, overlays, form controls, presence, logo, states) and two **consumption maps** (sidebar, editor: which design-system primitives a feature reuses vs. where it goes bespoke). Every specimen renders the **real** component — editor/context-coupled ones (menus, dialogs, toasts, presence) run in a minimal in-memory `<Tldraw>` or with faked state (fake presence records, forced-open dropdowns), so you see the resolved UI with no live session.

## What it surfaced

- The sidebar alone reaches for **9 bespoke button/trigger classes** instead of the SDK button (#9193 / #9194)
- Typography: **47 size declarations collapse to 6 sizes**; `12px` is hardcoded **28×** outside the classes (#9196)
- Icons: only **23 of 66** sprite icons are used (~65% dead)
- A duplicate `@keyframes tla-spin`, no shared `<ErrorState>`, and two parallel i18n systems (SDK vs dotcom react-intl)

## Testing

Open the preview, visit `/dev/components/buttons`, and walk the nav. Nothing ships to production — routes are gated on `import.meta.env.DEV` or a `*-preview-deploy` hostname.

### DIFF
--- a/apps/dotcom/client/src/__snapshots__/routes.test.tsx.snap
+++ b/apps/dotcom/client/src/__snapshots__/routes.test.tsx.snap
@@ -79,6 +79,10 @@ exports[`the_routes 1`] = `
     "vercelRouterPattern": "^/dev/components/typography/?$",
   },
   {
+    "reactRouterPattern": "/dev/components/z-index",
+    "vercelRouterPattern": "^/dev/components/z-index/?$",
+  },
+  {
     "reactRouterPattern": "/dev/reset-local-state",
     "vercelRouterPattern": "^/dev/reset-local-state/?$",
   },
--- a/apps/dotcom/client/src/__snapshots__/routes.test.tsx.snap
+++ b/apps/dotcom/client/src/__snapshots__/routes.test.tsx.snap
@@ -63,6 +63,10 @@ exports[`the_routes 1`] = `
     "vercelRouterPattern": "^/dev/components/presence/?$",
   },
   {
+    "reactRouterPattern": "/dev/components/shadows",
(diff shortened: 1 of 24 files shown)


mimecuvalo:
neat! this seems like it'd be useful eventually for people on our docs site to see our UI library.

that being said, why not just use Storybook or something akin to that? not sure if it's necessarily worth creating a custom Storybook but could be convinced.

jsscclr:
@mimecuvalo you're not wrong! The intent of this is a different purpose - this is purely a branch _just_ to see everything, i.e. I don't want to ever merge it. The idea is just to make it easy to see every e.g. button usage side by side. Storybook makes sense when we're actually showcasing an 'end product', so to speak. So this is a reference for when we would actually use Storybook (or similar) to develop + showcase a `<Button>` component. Hope that makes sense!

PR #9455, opened 2026-06-30
fix(editor): fix firing of onChildrenChange on shape creation

### LINKED ISSUE #8078: onChildrenChange does not fire with createShape
Creating a shape with a parentId ` editor.createShape({ ...shape, parentId })` does not cause the parent's `onChildrenChange` to fire. The following snippet works as expected though
```
editor.createShape(shape);
editor.reparentShapes([shape.id], parentId);
```` 

Same issue occurs with `editor.duplicateShapes`

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8078

We're missing the firing of `onChildrenChange` when creating a shape passing the parent id.
After a shape is created, it was not gathered for side effects at all and this is why it was not firing the callback.

This fix was already proposed in https://github.com/tldraw/tldraw/pull/8172 but @max-drake figured it was now firing twice… but in reality it was because we were firing once for creating the shape and a second time by resizing the shape, this is a different issue and in some ways I believe this is correct behaviour. Should we batch the changes in order to fire once is a follow-up question to this PR

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

Reproduction case in issue

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed missing `onChildrenChange` call when directly creating a shape within a parent shape.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -563,6 +563,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		let deletedBindings = new Map<TLBindingId, BindingOnDeleteOptions<any>>()
 		const deletedShapeIds = new Set<TLShapeId>()
 		const invalidParents = new Set<TLShapeId>()
+		const createdShapes = new Set<TLShapeId>()
 		let invalidBindingTypes = new Set<TLBinding['type']>()
 
 		this.disposables.add(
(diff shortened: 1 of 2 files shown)


mimecuvalo (approved):
LGTM

mimecuvalo:
be aware that you might collide w https://github.com/tldraw/tldraw/pull/9456

kaneel:
> be aware that you might collide w #9456

thanks for the headsup, I think it's fine are we're not touching similar regions of the editor!

PR #9456, opened 2026-06-30
fix(editor): batch page state cleanup after shape deletes

steveruizok:
In order to avoid redundant `store.put` calls when deleting many shapes at once, this PR moves instance page state cleanup for deleted shape IDs into the operation-complete handler so bulk deletions update page state once per operation instead of once per shape.

### Change type

- [x] `improvement`

### Test plan

1. Select many shapes and delete them
2. Confirm selection clears and page state updates only once
3. Undo and confirm shapes and selection are restored

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Improve performance of deleting many shapes at once by batching page state cleanup

### Code changes

| Core code       | +13 / -11  |
| Tests           | +38 / -0   |


mimecuvalo:
be aware that you might collide w https://github.com/tldraw/tldraw/pull/9455

mimecuvalo (approved):
LGTM

PR #9458, opened 2026-07-01
fix(tldraw): don't set prevDraggingOverShape to the initial one 

### LINKED ISSUE #9457: Dragging a shape from the toolbar directly into a frame does not reparent it to the frame
🗣️: when creating a shape over a frame the shape is reparented to this frame but when dragging a new shape from the toolbox to a frame does not reparent to the frame if there is no gap from the toolbox to the frame, dragging fast will just never register to the canvas page first and then we're not parenting to the frame, we have to go away from the frame and back into it if we want to be in a state where the frame will encapsulate the shape

---

Creating a shape directly over a frame correctly reparents that shape into the frame. But dragging a new shape out of the toolbar straight onto a frame does not: if the drag path goes from the toolbar into the frame without ever passing over the empty page area first, the shape stays parented to the page instead of the frame. This happens when the frame sits directly under the toolbar — either under the toolbar row itself or under the opened "more shapes" popup — so the pointer crosses from UI straight into the frame with no empty page in between. The reparent only registers once the pointer "visits" the empty page before entering the frame, so introducing a gap (panning the canvas so there's space between the toolbar and the frame, or dragging out of the frame and back in) makes it work. It reproduces with any shape (e.g. geo shapes) and on tldraw.com, so it's an SDK-level problem that affects everywhere frames are used. The expected behavior is that the shape ends up parented to the frame regardless of the drag path.

https://github.com/user-attachments/assets/0ec3c8fd-1f6e-4068-bcb7-1731a69124a2

## Open questions

1. **Which app(s) did you see this in — the examples app, tldraw.com, or both?**
   Seen on tldraw.com. Because it reproduces there, it's an SDK-level bug expected to occur everywhere frames are used.
2. **When it fails, what is the shape's end state?**
   The shape ends up parented to the page, overlapping the frame but not encapsulated by it. Reparenting only happens after the pointer first passes over the empty page, then re-enters the frame.
3. **Which shape/tool from the toolbar reproduces it, and does it happen for all draggable-from-toolbar items or a specific one?**
   Any shape reproduces it (e.g. geo shapes) — it's general to the drag-from-toolbar interaction. The trigger is spatial: the frame sits somewhat under the toolbar, either under the toolbar row itself or under the opened "more shapes" popup, so the drag goes from UI straight into the frame with no empty page in between.

Confidence: 88%, ready to get started.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9457

This happens because we're initialising the dragging with a previous drag set to the same initial drag, which means there no change in location and so we end up not parenting the shape.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a frame and move it under the toolbox
2. Drag a new shape from the toolbox to the frame
3. The frame should indicates the new shape will be parented to it

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed the dragging of a shape from the toolbox to a frame when the frame is under the toolbox, the shape is now properly parented to the frame

### DIFF
--- a/packages/tldraw/src/test/frames.test.ts
+++ b/packages/tldraw/src/test/frames.test.ts
@@ -486,8 +486,9 @@ describe('frame shapes', () => {
 		// box A should still be beneath box B
 		expect(editor.getShape(box1)!.index.localeCompare(editor.getShape(box2)!.index)).toBe(-1)
 
-		// We don't highlight the frame until dragged out and back in
-		expect(editor.getHintingShapeIds()).toHaveLength(0)
+		// The frame is highlighted as a drop target from the first update, even though the shape
+		// started over it (it's still the shape's parent, so nothing is reparented)
+		expect(editor.getHintingShapeIds()).toHaveLength(1)
 
 		expect(editor.getOnlySelectedShape()!.parentId).toBe(frame.id)
 
--- a/packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
+++ b/packages/tldraw/src/lib/tools/SelectTool/DragAndDropManager.ts
@@ -82,7 +82,11 @@ export class DragAndDropManager {
 			.sort((a, b) => allShapes.indexOf(a) - allShapes.indexOf(b))
 
 		this.initialDraggingOverShape = editor.getDraggingOverShape(point, this.shapesToActuallyMove)
-		this.prevDraggingOverShape = this.initialDraggingOverShape
+
+		// The drag hasn't visited any target yet, so start with no previous target (as if over the
+		// page). This makes the first update fire onDragShapesIn for whatever we're already over,
+		// instead of treating the origin's target as already-entered and skipping it.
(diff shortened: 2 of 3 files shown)


packages/tldraw/src/test/frames.test.ts
```diff
@@ -486,8 +486,9 @@ describe('frame shapes', () => {
 		// box A should still be beneath box B
 		expect(editor.getShape(box1)!.index.localeCompare(editor.getShape(box2)!.index)).toBe(-1)
 
-		// We don't highlight the frame until dragged out and back in
-		expect(editor.getHintingShapeIds()).toHaveLength(0)
```
kaneel, line 490: this was a funny one this line was actually testing against us :D

mimecuvalo (approved):
LGTM

PR #9460, opened 2026-07-01
docs(examples): add cursor maze multiplayer example

frolic:
In order to show off a handful of tldraw capabilities in one playable demo, this PR adds a **Cursor maze** example under `use-cases`: a multiplayer maze where your cursor _is_ the player. You can't give the OS cursor collision, so the game inverts the usual relationship between the cursor and the canvas — the player is pinned under the pointer and the camera moves to resolve collisions, so pushing into a wall scrolls the maze instead of letting you through.

It exercises a nice spread of the SDK:

- **Camera takeover** — the camera is locked (`isLocked: true`) and driven entirely by the game via `setCamera(..., { force: true })`; the whole "you can't pass through walls" effect is a camera pin re-computed every tick, so the camera only ever moves as a consequence of a collision.
- **World-space rendering** — the maze and the exit are drawn through the `OnTheCanvas` component; the players are tldraw's own cursors, so there's no avatar to render or sync.
- **Multiplayer with `useSyncDemo`** — the maze is seeded from the room id so everyone carves an identical one with no shape data synced. Because the player is pinned under the cursor, each player's synced cursor _is_ their collision-resolved maze position, so everyone else shows up as their own cursor threading the maze — with an off-screen edge hint pointing toward anyone outside your view. No avatar to render, no presence to wire up.
- **Cursor trails** — every cursor leaves a laser-style trail in that player's color, drawn to a plain canvas from the cursor positions the game already has (your own plus every collaborator's). It's a single filled ribbon that tapers to a point at the tail, so even a self-crossing loop stays solid with no per-segment alpha stacking — a pure local effect that adds nothing to what's synced.
- **Persistent scoreboard, no custom backend** — completion times are written into the synced document's `meta` (`editor.updateDocumentSettings`), which the sync server persists, so scores outlive disconnects and load for anyone who joins the room.

### New examples

- `use-cases/cursor-maze` — a multiplayer maze where your cursor is the player; push into walls to scroll the maze, race to the 🏁, and land your time on a persistent high-score board.

### Change type

- [x] `other`

### Test plan

1. `yarn dev`, then open `http://localhost:5420/cursor-maze/full`.
2. Click **Start**, steer your cursor toward the 🏁 — you can't pass through walls, and pushing into one scrolls the maze under your pointer. A trail follows your cursor and fades out behind it.
3. Reach the exit and confirm the timer freezes and your run lands on the high-score board.
4. Click **Copy link**, open it in a second tab, and confirm the other player shows up as a live cursor (with its own colored trail) threading the maze, plus an edge arrow when they're off-screen.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add a multiplayer "Cursor maze" example.

### Code changes

| Documentation | +839 / -0  |


mimecuvalo:
fun! hard to move the window though — i had to push in the opposite direction?

frolic:
> fun! hard to move the window though — i had to push in the opposite direction?

this is sort of intentional! I wanted to play with the that the cursor _is_ the player rather than the player being something that you move via cursor or keyboard, since we already have cursor presence indicators

PR #9463, opened 2026-07-01
fix(tldraw): stop PNG/SVG exports clipping text glyph ink

### LINKED ISSUE #8802: Exported PNG/SVG cuts off RTL text

frolic:
In order to stop exported PNGs and SVGs from cutting off italic, right-to-left, and cursive glyph ink at a text shape's edge (#8802), this PR pads the exported `<foreignObject>` in `TextShapeUtil.toSvg` by a fraction of the font size and insets the text by the same amount. Glyph ink that spills past the advance box — italic side bearings, slanted ascenders, cursive tails — now renders into that margin instead of being clipped at the foreignObject boundary, and the export's existing content-trim (`getSvgAsImage`) crops the slack back to the real pixels, so the image stays tight. The change is export-only: the shape's geometry — selection bounds, hit-testing, and the text-edit cursor — still uses the advance box, so nothing on-canvas moves.

Before

image

After

image

### Change type

- [x] `bugfix`

### Test plan

1. Create an italic (or Arabic / cursive) text shape whose glyphs lean past the box.
2. Export it as a PNG, then as an SVG, from the export menu.
3. The exported image encloses the full glyphs, with no clipped edges.

- [ ] Unit tests
- [x] End to end tests

### Release notes

- Fix exported PNG and SVG images clipping the edges of italic, right-to-left, and cursive text.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +14 / -2   |
| Tests     | +107 / -0  |

Closes #8802


steveruizok:
Excellent work @frolic! 

- Added before and after images
- Added hotfix-sdk-please, I think this is worth getting out to SDK users asap

steveruizok (approved):

PR #9472, opened 2026-07-02
feat(editor): add Editor.isMounted

frolic:
In order to let code that only holds the `editor` object check whether the editor has mounted, this PR adds an `Editor.isMounted` flag — mirroring the existing `Editor.isDisposed`.

The editor already exposes both a `dispose` event and an `isDisposed` flag, but for mount there was only the `mount` event and no flag. Because `mount` is fire-once, code that obtained the editor after it had already mounted (the common case) had no way to check whether mount had happened. `isMounted` closes that gap:

```ts
if (editor.isMounted) runNow()
else editor.once('mount', runNow)
```

This is aimed at making it easier to script against a live editor from outside (e.g. posting scripts to an exec endpoint), where you need a synchronous "has it mounted yet?" check rather than racing the fire-once event. Reacting to the transitions is still done via the `mount` / `dispose` events, matching how `isDisposed` works.

### API changes

- Added `Editor.isMounted`, set to `true` once the editor's `mount` event has fired. Mirrors `Editor.isDisposed`.

### Change type

- [x] `api`

### Test plan

1. Mount a `<Tldraw>` / `TldrawEditor` and confirm `editor.isMounted` is `true` in `onMount` and after.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add `Editor.isMounted` to check whether the editor has mounted, mirroring `Editor.isDisposed`.

### Code changes

| Core code       | +7 / -0    |

### DIFF
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1342,6 +1342,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     getInstanceState(): TLInstance;
     // (undocumented)
     getIsFocused(): boolean;
+    getIsMounted(): boolean;
     // (undocumented)
     getIsReadonly(): boolean;
     // @internal
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1342,7 +1342,6 @@ export class Editor extends EventEmitter<TLEventMap> {
     getInstanceState(): TLInstance;
     // (undocumented)
     getIsFocused(): boolean;
-    getIsMounted(): boolean;
     // (undocumented)
     getIsReadonly(): boolean;
     // @internal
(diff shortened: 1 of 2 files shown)


steveruizok (approved):
makes sense!

mimecuvalo:
followup to the this — i think we prbly want to be able to `isMount = false`.

~~the issue i see is:~~
~~- TldrawEditor.tsx has `useOnMount` → `Layout` → `teardownCallback` → which ...calls `onMount` weirdly enough~~
~~- b/c we call `onMount` during teardown I think we need to make sure that we set `isMount` to the correct state.~~
~~- not sure if this requires a separate event to be emitted — can you take a look?~~

I think I'm worried that possibly `TldrawEditorWithReadyStore` unmounts the `<Canvas>`, and `Editor` is still around and `isMounted = true`

mimecuvalo:
rewrote the above — I wrote that badly

frolic:
started on that here https://github.com/tldraw/tldraw/pull/9483

PR #9473, opened 2026-07-02
fix(dotcom): keep shared files visible in the sidebar despite orphan group_file links

mimecuvalo:
In order to fix shared files that are invisible in the tldraw.com sidebar, this PR makes `onEnterFile` create a user's home "guest file" link based on file *ownership*, and cleans up the stale `group_file` rows that caused the problem.

### The bug

A file owned by one workspace could have a leftover `group_file` row pointing at a *different* workspace that doesn't own it. These orphan rows were written by the old drag-to-link feature (`handleFileDragOperation`, #9107), which inserted a `group_file` link without changing `owningGroupId`. #9254 removed that feature but left the rows behind.

For any user who is a member of the workspace an orphan row points at, the file then became invisible everywhere:

- `getWorkspaceFilesSorted` only lists a non-home workspace's file when `owningGroupId === workspaceId`, so the orphan workspace never showed it.
- `onEnterFile`'s guard treated "a `group_file` row exists in any of my groups" as "the file is already visible to me", so it skipped adding the home guest-link — even though the orphan row showed the file nowhere.

The two code paths disagreed about whether that row counted, so the file fell through the gap.

### The fix

- `onEnterFile` now only skips the home guest-link when the file is owned by a workspace the user belongs to (or a home link already exists). A stale/mislinked row no longer suppresses the link. This also makes the mutator robust against any future orphan rows.
- Migration `038` re-links affected files into affected users' home groups (so the files reappear without needing to re-open them) and deletes the orphan `group_file` rows. It's scoped to genuine orphans (a `group_file` group that neither owns the file nor is a user's home group) and is idempotent.

### Change type

- [x] `bugfix`

### Test plan

1. As user A, create and share a file from your personal workspace.
2. As user B (a member of a shared workspace W), open A's file, then drag/link it into W under the old model — or seed a `group_file` row for `(file, W)` directly.
3. Re-open the file as user B and confirm it now appears in B's home "recent files".
4. Run migration `038` (dry-run first) and confirm orphan `group_file` rows are removed and affected users get home links.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix shared files that could be missing from the sidebar for members of a workspace the file had been linked into.

### Code changes

| Core code      | +22 / -11  |
| Tests          | +25 / -0   |
| Apps           | +65 / -0   |

### DIFF
--- a/packages/dotcom-shared/src/mutators.ts
+++ b/packages/dotcom-shared/src/mutators.ts
@@ -19,7 +19,6 @@ import {
 	TlaFileStatePartial,
 	TlaFlags,
 	TlaGroupFile,
-	TlaGroupUser,
 	TlaSchema,
 	TlaUser,
 	TlaUserPartial,
@@ -419,8 +418,6 @@ export function createMutators(userId: string) {
 			// If we get here, the user has legitimate access to the file
 			await tx.mutate.file_state.upsert({ fileId, userId, firstVisitAt: time })
 
-			const workspaceFileRows = await tx.run(zql.group_file.where('fileId', '=', fileId))
-			const userWorkspaceMemberships = await tx.run(zql.group_user.where('userId', '=', userId))
 			// Add a visited file to the user's home group so it shows as a "guest file" in the
 			// sidebar — unless it's already visible to the user somewhere else. A file is listed
 			// in a workspace only when that workspace actually OWNS it (getWorkspaceFilesSorted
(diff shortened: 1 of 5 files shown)


mimecuvalo:
@MitjaBezensek can you take a look at this? i honestly don't know if it's the right fix (for my missing document in the sidebar)

packages/dotcom-shared/src/mutators.ts
```diff
@@ -420,16 +421,26 @@ export function createMutators(userId: string) {
...
+			// lists a non-home workspace's file only when owningGroupId === workspaceId), so the
+			// only thing that should suppress the home link is the file being owned by a
+			// workspace the user belongs to.
+			//
+			// We deliberately do NOT key off "any group_file row in one of my groups": a
+			// stale/mislinked row (e.g. a leftover from the removed drag-to-link feature,
+			// #9107/#9254, or a create-workspace-race mirror) points at a workspace that does
+			// not own the file, so it shows the file nowhere. Counting it here would skip the
+			// home link and make the file invisible in the sidebar entirely.
+			const owningGroupId = file?.owningGroupId
```
MitjaBezensek: Maybe something like this might be easier to understand:
```ts
const alreadyLinkedInHome = await tx.run(
      zql.group_file.where('fileId', '=', fileId).where('groupId', '=', userId).one()
)
if (!alreadyLinkedInHome) {
      const ownedByOneOfMyWorkspaces =
              (await getRole(tx, userId, file?.owningGroupId)) !== null
      if (!ownedByOneOfMyWorkspaces) {
              await tx.mutate.group_file.insert({ fileId, groupId: userId, createdAt: time, updatedAt: time, index: null })
      }
}
```

MitjaBezensek (commented):
Looks good 👍 I ran some queries against the production and they confirm the issue, seems like 49 files are affected and the mutation should solve it.

1.  workspaces-feature.md still documents the old logic (where you can link files from a different work space), let's update it.
2. Should we also tighten up `validate_group_file_association` as it also seems to still support the old model?

packages/dotcom-shared/src/mutators.ts
mimecuvalo: Good call — done in 281d6bf. Switched to a single scoped `.where(...).one()` for the home-link check and resolved ownership via `getRole(tx, userId, file?.owningGroupId)` (null when I'm not a member, so no need to scan all memberships).

MitjaBezensek (approved):

PR #9475, opened 2026-07-02
fix(editor): treat license expiry as end of named day in UTC

### LINKED ISSUE #9431: License expiry is compared inconsistently across timezones, expiring up to a day early
🗣️: License expiry dates are compared inconsistently across timezones, causing licenses to expire up to a full day early for users west of UTC.

The license `expiryDate` is stored as a date-only string (e.g. `2026-06-26`), which gets parsed as UTC midnight. The expiry checks then rebuild that date using local-time getters and compare with `now >= expiration`, so the license expires at the *start* of the locally-derived day. For users west of UTC the UTC-midnight instant lands on the previous calendar day in local time, so a license dated `2026-06-26` expires at the start of June 25 local for them. Even in UTC and east, the license stops at the very start of the named day rather than the end of it.

The shift is fully exposed on evaluation licenses, which have no grace period. The same shift exists for annual and perpetual licenses but is normally masked by the 30-day grace period.

---

Reported behavior: a license with an expiry date of June 26 stopped working for users who were still on June 25 in their local timezone. The cause is that the named expiry date is interpreted at the *start* of the day rather than the end, and the date is read back in a different timezone frame than it was parsed in, so users west of UTC lose an additional day. Expected behavior is that the named expiry date is the last day the license is fully usable, applied consistently regardless of the user's timezone. The agreed direction is to treat the named date as valid through the end of that day in UTC, giving everyone a single predictable cutoff. Scope is limited to license expiry date handling in the editor's license validation; no change to how keys are minted is assumed.

Note: license validation runs client-side in the customer's installed SDK bundle, so any code fix only takes effect for customers after they upgrade. Already-issued keys on older bundles keep the current behavior, so newly minted keys should continue to pad the expiry date by a day or two until customers are on a fixed version.

## Open questions

1. **Should the named expiry date be the last fully-usable day (license valid through the end of that day), rather than expiring at its start?**
   Yes. "Expires June 26" should mean the license is usable through the end of June 26. The check should expire at the day boundary after the named date rather than at the start of it.
2. **Which timezone frame should define the expiry boundary — UTC (uniform cutoff for everyone) or each user's local day?**
   UTC, uniform. Define the boundary as the end of the expiry day in UTC (e.g. valid until `2026-06-27T00:00:00Z` for a `2026-06-26` key). This is identical for every user and fixes both the start-of-day and the western-hemisphere day-loss at once. A more lenient variant could anchor to the westmost timezone, but uniform UTC is the clean rule and a strict improvement over current behavior.
3. **How urgent is this relative to the workaround of padding minted expiry dates by a day or two? Should it target a specific release?**
   Medium. The padding workaround fully mitigates newly minted keys operationally, so this is not an emergency, but it is a real correctness bug affecting all customers (masked on annual/perpetual licenses by the 30-day grace period). Target a near-term minor release, and keep the date-padding convention for minting until customers upgrade to the fixed version.

mimecuvalo:
Closes #9431

## Problem

License `expiryDate` values are stored as date-only strings (e.g. `2026-06-26`), which `new Date(...)` parses as UTC midnight. The expiry checks then rebuilt that date with **local-time** getters (`getFullYear`/`getMonth`/`getDate`) and compared `now >= expiration`, so:

- The license expired at the **start** of the named day rather than being usable through the end of it.
- Because the date was read back in a different timezone frame than it was parsed in, users **west of UTC** lost an additional calendar day — a license dated June 26 stopped working while they were still on June 25 locally.

The shift is fully exposed on evaluation licenses (no grace period); on annual/perpetual licenses it was normally masked by the 30-day grace period.

## Fix

Compute the expiration boundary in **UTC** and treat the named date as the **last fully-usable day**. `getExpirationDateWithoutGracePeriod` and `getExpirationDateWithGracePeriod` now use `Date.UTC` with `getUTC*` getters, and the no-grace variant adds one day so the license is valid through the end of the named day.

A license dated `2026-06-26` is now valid until `2026-06-27T00:00:00Z` — a single predictable cutoff, identical for every user regardless of timezone.

Scope is limited to client-side expiry validation; key minting is unchanged. Because validation runs in the customer's installed bundle, minted keys should keep padding the expiry date by a day or two until customers upgrade to a fixed version.

## Change type

- [x] `other`

## Test plan

- Added regression tests in `LicenseManager.test.ts` using a date-only expiry string and a pinned clock (`vi.setSystemTime`):
  - valid at `2026-06-26T23:59:59Z` (end of the named day)
  - expired at `2026-06-27T00:00:01Z` (start of the next day)
  - not expired mid-day on June 26 (guards against the west-of-UTC early expiry)
- `yarn test run LicenseManager` — 58 passed
- `yarn typecheck`, `yarn lint`, `yarn api-check` — pass

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.test.ts
+++ b/packages/editor/src/lib/license/LicenseManager.test.ts
@@ -594,38 +594,54 @@ describe('LicenseManager', () => {
 		})
 
 		it('Handles grace period correctly - 20 days expired should still be within grace period', async () => {
-			const expiredLicenseInfo = JSON.parse(STANDARD_LICENSE_INFO)
-			const expiredDate = new Date(now.getFullYear(), now.getMonth(), now.getDate() - 20) // 20 days ago
-			expiredLicenseInfo[PROPERTIES.EXPIRY_DATE] = expiredDate.toISOString()
+			// Pin the clock and use a date-only UTC expiry so the calculation is timezone
+			// independent. The named day is fully usable, so an expiry 20 calendar days before
+			// "now" has been past its usable window for 19 full days.
+			vi.useFakeTimers()
+			vi.setSystemTime(new Date('2026-06-26T12:00:00.000Z'))
+			try {
+				const expiredLicenseInfo = JSON.parse(STANDARD_LICENSE_INFO)
+				expiredLicenseInfo[PROPERTIES.EXPIRY_DATE] = '2026-06-06' // 20 days before now
 
-			const expiredLicenseKey = await generateLicenseKey(
-				JSON.stringify(expiredLicenseInfo),
-				keyPair
-			)
+				const expiredLicenseKey = await generateLicenseKey(
+					JSON.stringify(expiredLicenseInfo),
+					keyPair
+				)
 
-			// Test the getLicenseFromKey method to verify grace period calculation
-			const result = (await licenseManager.getLicenseFromKey(
-				expiredLicenseKey
-			)) as ValidLicenseKeyResult
-			expect(result.isAnnualLicense).toBe(true)
-			expect(result.isAnnualLicenseExpired).toBe(false) // Within 30-day grace period
-			expect(result.daysSinceExpiry).toBe(20)
+				// Test the getLicenseFromKey method to verify grace period calculation
+				const result = (await licenseManager.getLicenseFromKey(
+					expiredLicenseKey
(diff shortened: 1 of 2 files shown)


mimecuvalo:
@MitjaBezensek can you take a look?

MitjaBezensek (approved):

PR #9479, opened 2026-07-02
fix(arrows): allow selected clipped arrows to be dragged

steveruizok:
In order to preserve the clipped-arrow selection fix from #9429 after #9470 moved frame-bound arrows out of clipping, this PR lets already-selected shapes be hit-tested against their geometry using the standard hit-test margin when deciding whether a drag should start. This keeps selected arrows grabbable at the same places as their selection indicator, including for custom clipping containers that still clip arrow children.

### Why this change is needed

Normal shape picking still goes through `getShapeAtPoint`, which uses the editor's standard hit-test margin and respects clipping masks. That is the right behavior for finding the top-most visible shape under the pointer.

The selected-shape path is different. `getSelectedShapeAtPoint` runs after a shape is already selected, and it decides whether a pointer down should start dragging that selected shape. Before this PR, it used `isPointInShape` with `margin: 0`. That worked for many selected clipped shapes because closed shapes have an inside area and this path passed `hitInside: true`.

Arrows are different: they are open, thin geometry, so `margin: 0` required an exact hit on the arrow body. For a curved arrow that is already selected and visually represented by its selection indicator, pointer down near the selected arrow could miss the selected shape and start another canvas interaction instead.

This PR keeps `isPointInShape` as the clipped/visible-shape test used by normal picking, and makes `getSelectedShapeAtPoint` use the selected shape's own geometry with the standard hit-test margin. The result matches the interaction model: normal picking still uses clipped shape hit-testing, while already-selected arrows remain draggable with the same forgiving margin as regular arrow hits.

### Change type

- [x] `bugfix`

### Test plan

1. Use a custom clipping parent with a curved arrow child that extends outside the clip path.
2. Confirm a normal shape hit-test misses the clipped point, but the selected arrow can be dragged from that point.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix selected clipped arrows so they can be dragged where their geometry extends outside a clipping parent.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +19 / -4   |
| Tests     | +49 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -5805,10 +5805,17 @@ export class Editor extends EventEmitter<TLEventMap> {
 	 */
 	getSelectedShapeAtPoint(point: VecLike): TLShape | undefined {
 		const selectedShapeIds = this.getSelectedShapeIds()
+		const margin = this.options.hitTestMargin / this.getZoomLevel()
 		return this.getCurrentPageShapesSorted()
 			.filter((shape) => shape.type !== 'group' && selectedShapeIds.includes(shape.id))
 			.reverse() // find last
-			.find((shape) => this.isPointInShape(shape, point, { hitInside: true, margin: 0 }))
+			.find((shape) =>
+				this.getShapeGeometry(shape).hitTestPoint(
+					this.getPointInShapeSpace(shape, point),
+					margin,
+					true
+				)
+			)
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


packages/editor/src/lib/editor/Editor.ts
```diff
@@ -5805,10 +5805,25 @@ export class Editor extends EventEmitter<TLEventMap> {
...
+			if (
+				this.getShapeGeometry(shape).hitTestPoint(
+					this.getPointInShapeSpace(shape, point),
+					margin,
+					true
+				)
+			) {
+				return shape
+			}
+		}
```
steveruizok, line 5824: fewer allocations (ram prices are high)

kaneel (approved):
always upvote a reversed loop

PR #9493, opened 2026-07-04
feat(tldraw): hold cmd/ctrl to select the exact shape under the pointer

steveruizok:
In order to let people reach into groups and stacked shapes without changing focus, this PR repurposes the accel key (cmd on mac, ctrl elsewhere) in the select tool. Previously accel acted as a second additive-selection key alongside shift. Now holding accel drills straight to the exact shape under the pointer — bypassing groups and focus layers — and replaces the current selection instead of adding to it. Hover updates live as accel is pressed and released so you can preview what you're about to select. Shift keeps its existing additive-selection behavior.

This PR also refines the selection overlay's `isChangingStyle` flag (which hides the overlay while you nudge or restyle a shape) so it only clears when the pointer actually leaves the selection bounds, rather than on any pointer move. It now also clears when you press shift, change the selection, select all, or undo/redo to a state where the selection is no longer under the pointer, and is set while translating multiple shapes.

### Change type

- [x] `improvement`

### Test plan

1. Create a group of shapes, or stack a small shape on top of a larger filled one.
2. Hold cmd/ctrl and hover a child shape — the exact shape under the pointer highlights.
3. Click while holding cmd/ctrl — only that shape is selected, and the group stays focused.
4. Release cmd/ctrl — hover returns to the outermost selectable shape.
5. Nudge a shape with the arrow keys, then move the pointer inside the selection (overlay stays hidden) and then outside it (overlay reappears).

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Section   | LOC change  |
| --------- | ----------- |
| Core code | +122 / -46  |
| Tests     | +303 / -129 |

### Release notes

- Hold cmd (or ctrl) in the select tool to select the exact shape under the pointer, drilling into groups and past overlapping shapes without changing focus.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -1490,6 +1490,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this._flushEventsForTick(0)
 		this.complete()
 		this.history.undo()
+		this._clearIsChangingStyleIfPointerIsOutsideSelection()
 		this.performance._notifyUndoRedo('undo', this.history.getNumUndos(), this.history.getNumRedos())
 		return this
 	}
@@ -1521,6 +1522,7 @@ export class Editor extends EventEmitter<TLEventMap> {
 		this._flushEventsForTick(0)
 		this.complete()
 		this.history.redo()
+		this._clearIsChangingStyleIfPointerIsOutsideSelection()
 		this.performance._notifyUndoRedo('redo', this.history.getNumUndos(), this.history.getNumRedos())
 		return this
 	}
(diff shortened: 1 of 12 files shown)


mimecuvalo:
Does this help with this issue that we ran into the other day? Checking as long as you're exploring this space.

https://github.com/user-attachments/assets/688a4ac3-dd2f-4b54-8739-b0c1f133631b

steveruizok:
> Does this help with this issue that we ran into the other day? Checking as long as you're exploring this space.
> 
>  Kapture.2026-07-06.at.16.29.22.mp4

Different PR for that!

PR #9494, opened 2026-07-05
fix(editor): ignore stale font load callbacks

### LINKED ISSUE #9114: Unhandled "AtomMap: key not found" crash when a font fails to load
🗣️: Sentry shows an unhandled "AtomMap: key [object Object] not found" crash when a font fails to load (LITE-718, ~1,050 events / ~940 users since May 27, 2026) — let's create an issue for it with the Sentry context and the date it started happening.

---

Since **May 27, 2026**, tldraw.com users hit an unhandled error `AtomMap: key [object Object] not found` — about 1,050 events from 940 users in the first two weeks, making it one of the most widespread client errors right now. It fires when a font fails to load and the editor tries to record the failure for a font it no longer tracks, so the error-handling path itself throws as an unhandled rejection. The error lives in SDK code (editor font handling on top of the store), so it can affect SDK consumers too. The start date suggests it shipped in (or was surfaced by) a release around May 27. The goal is for a failed font load to never throw.

Sentry: [LITE-718](https://tldraw.sentry.io/issues/LITE-718)

## Open questions

1. **Do you know of any font- or store-related change that shipped around May 27 that could be the trigger?**
   _Awaiting answer._

steveruizok:
In order to prevent failed font loads from crashing after an editor is disposed, this PR makes `FontManager` ignore stale font load callbacks when their tracked font state has already been cleared. Closes #9114.

Stale callbacks are detected by identity, not presence: a callback only updates the exact state entry it was created for, so a font re-requested after `dispose()` (which acts as a reset) can never be touched by a leftover callback from before the reset. Interrupted loads are logged at debug level instead of silently swallowed, so a genuine font failure in a short-lived editor is still visible in verbose console output without contributing error noise.

### Change type

- [x] `bugfix`

### Test plan

1. Navigate away from a file while a custom font load is pending, then let the load fail.
2. Confirm the stale callback does not throw `AtomMap: key [object Object] not found`.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix a crash that could happen when a font failed to load after an editor was disposed.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +14 / -0   |
| Tests     | +66 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
@@ -160,15 +160,53 @@ describe('FontManager', () => {
 					),
 				}
 			})
-			const consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+			const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+			const debugSpy = vi.spyOn(console, 'debug').mockImplementation(() => {})
 
 			const promise = fontManager.ensureFontIsLoaded(font)
 			fontManager.dispose()
 			rejectLoad(error)
 
 			await expect(promise).resolves.toBeUndefined()
-			expect(consoleSpy).not.toHaveBeenCalled()
-			consoleSpy.mockRestore()
+			expect(errorSpy).not.toHaveBeenCalled()
+			expect(debugSpy).toHaveBeenCalledWith(
+				`Font "${font.family}" load interrupted by editor dispose`,
+				error
+			)
+			errorSpy.mockRestore()
+			debugSpy.mockRestore()
+		})
+
+		it('ignores font loads that finish after disposal', async () => {
+			const font = createMockFont()
+			let resolveLoad: () => void = () => {}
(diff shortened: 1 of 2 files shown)


frolic:
Pushed 49941c8 with a few additions (this came up independently via the desktop app's reopen error storm, tldraw/tldraw-internal#674 — I'd opened the duplicate #9506 before finding this PR):

- **Resolve-path coverage** — a font that *succeeds* after dispose hit the same `AtomMap.update` throw, so there's now a test for that path too, including asserting no post-dispose `document.fonts.add`.
- **Debug log instead of silent swallow** — stale callbacks now `console.debug("Font "X" load interrupted by editor dispose")` (with the error on the failure path), so genuine font failures in short-lived editors stay visible in verbose output without adding error noise.
- **Dropped the redundant check** — the staleness condition ran twice (inline and inside `updateState`); it's now a single `isStale()` identity check with a comment on why identity rather than presence.

The identity check was the right call over a plain `has(font)` — it stays correct even if a stale callback races a *new* FontFace instance (e.g. if the per-document font face cache ever evicts between dispose and re-request).

Sent from my aiPhone

mimecuvalo (approved):

PR #9496, opened 2026-07-05
fix(tldraw): prevent frame heading crash when text spans can't be measured

steveruizok:
In order to stop a `TypeError: Cannot read properties of undefined (reading 'box')` crash when computing frame geometry, this PR makes `getFrameHeadingSize` tolerate `measureTextSpans` returning an empty array. That happens when the text measurement element can't be laid out — for example during the initial editor mount, where the spatial index is built synchronously inside `new Editor()` before the measurement element is attached/visible, or when every grapheme lacks a layout rect (the case the #9112 fix started skipping). Previously the code did `last(spans)!` and immediately dereferenced `.box`, crashing whenever `spans` was empty.

The fix guards the empty-spans case in `getFrameHeadingSize` (returning a zero-width heading, uncached, so it's remeasured once layout is possible) and hardens the identical latent dereference in `TextManager`'s `truncate-ellipsis` path.

Relates to #9112.

### Change type

- [x] `bugfix`

### Test plan

Because `getFrameHeadingSize` short-circuits when `NODE_ENV === 'test'` and never calls `measureTextSpans`, this crash can't be reproduced through frame geometry in unit tests, so regression coverage lives at the `TextManager` level (the shared root cause).

1. Add a frame to a canvas and confirm the heading renders normally.
2. Confirm no crash occurs when an editor mounts with an existing frame present.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +18 / -3   |
| Tests     | +42 / -0   |

### Release notes

- Fix a crash when computing a frame's geometry if its heading text couldn't be measured yet.

### DIFF
--- a/packages/editor/src/lib/editor/managers/TextManager/TextManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/TextManager/TextManager.test.ts
@@ -420,6 +420,48 @@ describe('TextManager', () => {
 			expect(Array.isArray(result)).toBe(true)
 		})
 
+		it('should not throw when the truncated remeasure yields no spans', () => {
+			// The text truncates on the first pass, but once we narrow the width
+			// to make room for the ellipsis, nothing measurable remains (e.g. the
+			// measurement element can't be laid out). This must not crash.
+			vi.spyOn(textManager, 'measureElementTextNodeSpans')
+				.mockReturnValueOnce({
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
👍

Issue #9502, opened 2026-07-05
Add Cloudflare Browser Run thumbnail generation for tldraw.com files


steveruizok:
tldraw.com should generate server-side thumbnails for files using Cloudflare Browser Run rather than a bespoke long-running Playwright service. The first version should render only tldraw-owned thumbnail pages, not arbitrary URLs, and should use existing tldraw.com auth, permission checks, snapshot storage, and R2 caching. This issue is the focused implementation slice for the broader server-side thumbnail generation work in #7551.

## Proposed scope

- Add a tldraw.com thumbnail generation path backed by Cloudflare Browser Run screenshot rendering.
- Render an internal, signed tldraw-owned page that loads a file snapshot and marks itself ready when the canvas is stable.
- Cache generated thumbnails in R2 using a key based on file identity, page or viewport options, dimensions, theme, and document version.
- Return cached thumbnails immediately when available.
- Add request limits, render timeouts, output dimension limits, and telemetry for cache hits, render duration, failures, and request source.
- Avoid arbitrary URL screenshots in the production thumbnail endpoint.

## Prior art

- #7551 is the umbrella issue and original server-side thumbnail generation spec.
- #7866 is a closed prototype PR that added a Playwright-driven thumbnail generator app.
- #4602 removed the older dotcom thumbnail logic and server, which may be useful for historical integration context.
- #9497 is related future work for shared board screenshots and MCP access.
- Cloudflare Browser Run screenshot endpoint: https://developers.cloudflare.com/browser-run/quick-actions/screenshot-endpoint/

## Open questions

1. **Should v1 support private user files, or only published/shared files?**
   Start with the smallest useful surface. Published/shared files are likely enough for v1 unless product requirements need private dashboard thumbnails immediately.
2. **Should thumbnail generation be synchronous, asynchronous, or both?**
   Prefer cached synchronous reads with asynchronous generation on cache miss or file update. A blocking fallback is acceptable only if it has strict timeouts and is not used by hot file-list paths.

Issue #9497, opened 2026-07-05
Add a tldraw.com MCP server for shared board screenshots


steveruizok:
🗣️: Ok, lets add an issue that would implement an MCP server that would allow an agent to access information about a board. For simplicity sake, we can hold off on authentication at first and just expect that the board is shared. We'd need telemetry to be sure that this isn't being abused, and perhaps add a sanity check on individual boards. We might need to do the thumbnail generation first, as this service would also need to produce images from the canvas. The place where this feature would exist is tldraw.com.

---

Add a tldraw.com MCP server that lets agents request screenshots of shared boards without requiring logged-in user authentication in the first version. The first MCP surface should be image-only: it does not need to expose board metadata, document structure, or other board information beyond rendered images. Arbitrary viewport screenshots should be the initial target, with current viewport and selected-shape screenshots treated as closely related targets once the rendering path exists. This likely depends on or should be coordinated with server-side thumbnail generation, since both features need reliable canvas image rendering. The first version should include telemetry and a per-IP sanity check, starting around more than 20 requests per minute from a given IP, so the team can detect abuse and limit excessive access.

## Open questions

1. **What board information should the first MCP version expose besides rendered images?**
   Nothing; the first version should only expose rendered images.
2. **Should the first version include all screenshot targets, or start with a narrower target such as fit-to-page or arbitrary viewport only?**
   Start with arbitrary viewport screenshots. Current viewport and selected-shape screenshots should be similar work once any screenshot target is implemented.
3. **What kind of board-level sanity check should block or throttle access in the first version?**
   Start with an IP-based threshold, such as more than 20 requests per minute from a given IP.

Confidence: 88%, ready to get started.

PR #9423, opened 2026-06-25
feat(dotcom): include board name in a board's social preview

steveruizok:
In order to make shared tldraw.com board links recognizable, this PR includes the board's name in a board's social preview. Closes #9365.

tldraw.com is a static single-page app, so the only metadata social crawlers (Slack, Discord, Twitter, Facebook, etc.) ever saw was the generic `tldraw` Open Graph tags baked into `index.html` — the runtime title updates never reach a crawler that doesn't run JavaScript. This PR routes those crawlers (by user-agent) to the multiplayer worker, which renders a small HTML document whose preview title is the board's name.

The preview title is `<board name> • tldraw.com`, falling back to `tldraw.com` for boards with no name. The existing description and social image are unchanged. Search-engine crawlers (Googlebot, bingbot) render JavaScript and are intentionally not routed here, so they keep indexing the real app.

Coverage by link type:

- `/f/` live multiplayer files — name from the `file` table, only for shared files so private board names aren't leaked
- `/p/` published files — name from the published file's `file` record
- `/s/` snapshots — name from the snapshot's `document` record
- `/r/`, `/ro/`, `/v/` legacy rooms — name from the persisted room snapshot's `document` record

### Change type

- [x] `feature`

### Test plan

1. Build the client (`yarn build-app`) and confirm `.vercel/output/config.json` contains the social-preview rewrite routes for `/f/`, `/p/`, `/s/`, `/r/`, `/ro/`, `/v/`, each gated on a social-crawler `user-agent`.
2. Against a deployed/preview environment, request a shared board link with a crawler user-agent, e.g. `curl -A 'Twitterbot' https://<host>/p/<slug>`, and confirm the returned HTML contains `<meta property="og:title" content="<board name> • tldraw.com" />`.
3. Confirm a board with no name returns `tldraw.com`, and that an unshared `/f/` file does not expose its name.
4. Confirm a normal browser request to the same URL still loads the app.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- tldraw.com board links now include the board's name in their social/link-unfurl preview.

### Code changes

| Section        | LOC change |
| -------------- | ---------- |
| Tests          | +77 / -0   |
| Apps           | +210 / -0  |
| Config/tooling | +43 / -0   |

PR #9503, opened 2026-07-06
feat(dotcom): production Browser Run thumbnails and shared board screenshot MCP

steveruizok:
In order to give tldraw.com boards real preview images — both a synchronous MCP tool that returns a PNG of a published board (#9497) and rich link-unfurl OG images for shared and published boards (#9502) — this PR adds a production Browser Run screenshot pipeline to the dotcom sync worker and client. It replaces the earlier spike's fixture rendering: boards are resolved from real published/shared records, rendered through a signed internal render job, cached in R2, and rate limited by real Cloudflare bindings.

The pipeline never hands Browser Run a user-provided URL. The worker validates and resolves the board itself, mints a short-lived HMAC-signed render token, and Browser Run only ever visits the tldraw-owned render page carrying that token. Both surfaces stay image-only: no board metadata beyond page names, no document structure, shape listing, arbitrary selectors, arbitrary URLs, or private-file access.

There are two consumers of the same render-and-cache core:

- **Synchronous MCP tool** (`POST /api/app/mcp`) — returns the PNG in-band, so it renders inline on a cache miss.
- **Asynchronous OG images** (`GET /api/app/social-preview/:prefix/:slug/image`) — never blocks a request on Browser Run. It serves whatever is cached (fresh or stale) or redirects to the default OG image, and enqueues an out-of-band render job. This is what makes it safe on high-traffic link-unfurl paths.

### MCP surface

The MCP server exposes two read-only tools:

- `get_board_info` — returns a board's name, page count, and each page's name, 0-based index, and `hasContent` flag. Callers use this to discover pages first.
- `get_shared_board_screenshot` — returns a content-fit PNG of a single page, preceded by the page name. Inputs are a `boardId`, a `page` index (default 0), and a `theme` (`light`/`dark`).

`boardId` is a board **id** (a slug), not a URL, so there is no host to validate: it is tried as an anonymously-shared file id (`tldraw.com/f/:slug`) first and a published-board slug (`tldraw.com/p/:slug`) second. Shared files must currently be shared for anonymous viewing; private, invite-only, deleted, or unknown boards fail before any Browser Run spend.

### How it works

1. The worker resolves the board — published boards through `SNAPSHOT_SLUG_TO_PARENT_SLUG` and `getPublishedFileInfo`, shared files through `getSharedFileInfo` gated on `isFileAnonymouslyViewable`. Unknown or unviewable boards fail early.
2. Screenshots are cached in a new `THUMBNAILS` R2 bucket, keyed per page on board kind, slug, `lastPublished`/version, dimensions, theme, and page ordinal, so republishing or editing rotates every page's key. Hits return without invoking Browser Run.
3. On a miss, the worker mints a 5-minute HMAC render token (`renderTokens.ts`) and calls Browser Run's `/screenshot` Quick Action against `{MCP_SCREENSHOT_RENDER_ORIGIN}/__thumbnail-render?token=...`.
4. The production render page (`thumbnail-render.tsx`, registered in production routing and the generated Vercel rewrites) exchanges the token for snapshot data at the token-gated `GET /api/app/thumbnail-render/snapshot` endpoint, renders a hidden-UI read-only editor, and sets `data-thumbnail-ready` only after fonts settle, image assets load, and the editor's `` elements are stable — closing the review finding that raster content could be captured before it painted.
5. For OG images, the `queue` consumer (`ogImageQueue.ts`) performs the capture out of band and refreshes the cache the route reads. A pending marker suppresses duplicate enqueues, stale-but-recent images are served without re-rendering (one board spends Browser Run capacity at most about once an hour), and render attempts are bounded independently of capacity backpressure. `getSocialPreview` now emits `og:image`/`twitter:image` tags pointing at the board's own thumbnail with a `summary_large_image` card for board kinds that support it, falling back to the static site images otherwise.

Rate limits are enforced by Cloudflare `ratelimit` bindings declared in wrangler.toml for every environment: ~2/min per IP, ~2/min per board (cache misses only), and a global ~6/min cap on Browser Run invocations. An isolate-local fallback covers local dev and tests only.

Telemetry (`mcp_shared_board_screenshot`, plus OG-image datapoints) records cache hit/miss status and which rate-limit scope blocked a request, alongside hashed IP/board, Browser Run duration, `X-Browser-Ms-Used`, dimensions, and failure reason. `internal/scripts/fetch-screenshot-metrics.ts` pulls these back for inspection.

The dev-only fixture page (`/dev/browser-run-thumbnail`) and the `browser-run-thumbnail` script remain for credential-free local iteration and share the render page's ready-signal implementation. `apps/dotcom/browser-run-thumbnails.md` documents the architecture, configuration, and remaining follow-ups.

### One-time ops setup before deploy

- Create the R2 buckets: `thumbnails-preview` and `thumbnails`.
- Add the `BROWSER_RENDERING_API_TOKEN` (Browser Rendering – Edit permission only) and `MCP_SCREENSHOT_TOKEN_SECRET` GitHub secrets. Until they exist, deploys still succeed and the MCP tool returns a configuration error.

### Change type

- [x] `feature`

### Test plan

1. From the repo root, run `yarn dev-app`, publish a file locally, then call `POST /app/mcp` on the local sync worker with `tools/call` — `get_board_info` to list pages, then `get_shared_board_screenshot` for a page. The render URL it uses can be opened in a browser while the token is valid.
2. Local render-behavior check without credentials: `yarn workspace dotcom browser-run-thumbnail --mode local --base-url http://127.0.0.1:3000 --fixture snapshot-example --output tmp/browser-run-thumbnail/local-thumbnail.png`.
3. After the ops setup lands, verify on staging: call `POST https://staging.tldraw.com/api/app/mcp` with a published staging board and confirm a PNG result, an R2 cache hit on the second call, and rate-limit errors past the per-IP limit. Separately, request `GET /api/app/social-preview/p/:slug/image` for a published board and confirm the first hit redirects/serves default while the queued render lands, then subsequent hits serve the board thumbnail.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- tldraw.com: add an image-only MCP endpoint (`/api/app/mcp`) that returns PNG screenshots of published and shared tldraw.com boards.
- tldraw.com: shared and published board links now unfurl with a live preview image of the board.

### Code changes

| Core code       | +54 / -0     |
| Tests           | +1340 / -1   |
| Automated files | +4 / -0      |
| Apps            | +2834 / -41  |
| Config/tooling  | +439 / -5    |

### Related issues

- Relates to #9502
- Relates to #9497


steveruizok:
Ok, this works. You can visit the MCP tester page on the teamldraw app to try it out.

driev:
Screenshot 2026-07-16 at 17 10 48

... getting there.

apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -178,6 +182,10 @@ const router = createRouter<Environment>()
 	})
 	.post('/app/submit-feedback', submitFeedback)
 	.get('/app/feature-flags', getFeatureFlags)
+	.post('/app/mcp', sharedBoardScreenshotMcp)
+	.get('/app/og-html/:kind/:slug', getOgHtml)
+	.get('/app/og-image/:kind/:slug', getOgImage)
```
driev: arrr... `/app/og-html/:kind/:slug` is a duplicate of `/app/social-preview/:prefix/:slug` which was merged in https://github.com/tldraw/tldraw/pull/9423

It might make sense to end up with something like:
```
- GET /app/social-preview/:prefix/:slug → the HTML (unchanged)
- GET /app/social-preview/:prefix/:slug/image → the PNG (renamed from og-image)
```

Any preferences?

mimecuvalo: i think your proposal works 👍

max-drake:
cool its mostly looking good! 
the main thing is that the mcp server still seems to be using the original input/output as opposed to what we tdiscussed in the notion
- taking board id instead of url as input
- not allowing w/h or bounds as input (we should screenshot everything on the board (maybe with some nice margins)
- return the page name and screenshot for each page with content, instead of just a screenshot of one page 

also lets make the light/dark arg optional, default to light

apps/dotcom/browser-run-thumbnails.md
```diff
@@ -0,0 +1,190 @@
...
+3. Otherwise the worker enqueues an `og-image-render` job on the existing sync-worker queue (guarded by a per-board rate limit and a two-minute pending marker in R2 that dedupes concurrent enqueues) and serves the previous image marked stale with `max-age=300`, or the default-image redirect with `max-age=60` if the board has never been rendered. Scrapers pick up the fresh render on their next visit.
+4. The queue consumer (`ogImageQueue.ts`, dispatched from the worker's `queue()` handler) re-resolves the board at render time: a board un-shared while queued is dropped and its cached OG image deleted, and the version is re-read so bursts of enqueues coalesce into one capture of the newest content. It checks the shared global Browser Run rate limit (requeueing when capacity is busy), mints a render token with `camera: 'content'`, captures through Browser Run, and writes the PNG back to the cache key the route reads. Transient failures retry up to three times with backoff, then drop.
+
+### Request limits
+
+- Per IP: ~20 tool calls per minute (`MCP_SCREENSHOT_RATE_LIMITER`).
```
mimecuvalo: i'll sleep better at night if this 10x less :P

let's do ~2 tool calls per minute / ~2 browser run captures per minute

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
```diff
@@ -0,0 +1,360 @@
...
+
+		writeOgImageTelemetry(env, {
+			source: 'queue',
+			boardHash,
+			cacheStatus: 'miss',
+			browserRunDurationMs: render.durationMs,
+			browserMsUsed: null,
+		})
+		message.ack()
+	} catch (error) {
```
MitjaBezensek, line 260: Let's log these to sentry?

apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
```diff
@@ -0,0 +1,174 @@
...
+const STALE_IMAGE_MAX_AGE_SECONDS = 5 * 60
+const REDIRECT_MAX_AGE_SECONDS = 60
+
+export async function getOgImage(request: IRequest, env: Environment): Promise<Response> {
+	// Crawlers probe this URL with HEAD before (or instead of) GET, so the route is registered with
+	// .all and HEAD must still return the cache/redirect headers. But a HEAD must not spend Browser
+	// Run: only a real GET reads the R2 body and enqueues a render. Any non-GET method is treated
+	// like a probe (headers only, no enqueue).
+	const wantsBody = request.method === 'GET'
+	const board = await resolveOgBoard(request, env).catch(() => null)
```
MitjaBezensek: Let's log these to sentry?

apps/dotcom/sync-worker/src/routes/tla/getThumbnailSnapshot.ts
```diff
@@ -0,0 +1,68 @@
...
+		return json({ error: true, message: 'Invalid or expired render token' }, 403)
+	}
+
+	// Shared files re-check their share status here (not just when the token was minted), so that a
+	// board un-shared during a token's 5 minute window stops resolving.
+	const snapshot = await (
+		job.kind === 'published'
+			? getPublishedRoomSnapshot(env, job.slug)
+			: getSharedFileRoomSnapshot(env, job.slug)
+	).catch(() => undefined)
```
MitjaBezensek: Let's log these to sentry?

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
```diff
@@ -0,0 +1,802 @@
...
+			return toolError('This board has no saved content yet.')
+		}
+		const pages = enumerateBoardPages(snapshot)
+		return toolJsonResult({
+			name: getDocumentNameFromSnapshot(snapshot),
+			pageCount: pages.length,
+			pages: pages.map((p) => ({ index: p.index, name: p.name, hasContent: p.hasContent })),
+		})
+	} catch (error) {
+		return toolError(
```
MitjaBezensek, line 380: Let's log these to sentry?

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
```diff
@@ -0,0 +1,802 @@
...
+			return jsonRpcError(rpcRequest.id, -32601, `Method not found: ${rpcRequest.method}`)
+	}
+}
+
+export function parseSharedBoardScreenshotInput(input: unknown): SharedBoardScreenshotInput {
+	const value = requireArgumentsObject(input)
+	return {
+		boardId: parseBoardId(value.boardId),
+		page: parsePageOrdinal(value.page),
+		theme: value.theme === 'dark' ? 'dark' : 'light',
```
MitjaBezensek: Should we enforce `light`? This would default to light even for non existent themes like `blue`?

apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -320,10 +331,24 @@ export default class Worker extends WorkerEntrypoint<Environment> {
 	}
 
 	override async queue(batch: MessageBatch<QueueMessage>): Promise<void> {
-		const db = createPostgresConnectionPool(this.env, 'sync-worker-queue')
+		// The pool is only needed for asset-upload messages, so create it lazily: OG image render
+		// batches should not open database connections they never use.
+		let db: ReturnType<typeof createPostgresConnectionPool> | undefined
 		for (const message of batch.messages) {
+			if (message.body.type === 'og-image-render') {
```
MitjaBezensek: Should we do an exhaustive switch based on the type here?

It also got me thinking if we should use a separate queue for this? Not really sure 🤷‍♂️

apps/dotcom/browser-run-thumbnails.md
MitjaBezensek, line 1: Looks like we have some stale docs here (like `returned render URL` - we don't seem to return the URL any longer).

MitjaBezensek (commented):
Looking good in general, left a few small comments.

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
driev: Ooh, I didn't realise that needed to be explicit.

apps/dotcom/sync-worker/src/worker.ts
driev: I can add a separate queue in a follow up PR.

apps/dotcom/browser-run-thumbnails.md
driev: Refreshed.

MitjaBezensek (approved):

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
```diff
@@ -0,0 +1,873 @@
...
+			board: {
+				kind: 'published',
+				slug: boardId,
+				version: published.lastPublished,
+			},
+		}
+	}
+
+	return { ok: false, reason: 'not_found' }
+}
```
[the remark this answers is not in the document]
driev: This is not worth fixing... the error is not actionable either way.

driev:
This works... but I ran into a weird error when I was testing. I deployed the new endpoints today, but had been testing with a board that had links to itself via the old endpoints (`GET /api/app/og-image/...`), which no longer existed. The Browser Run task then hangs until all content is loaded and eventually times out after 45 seconds.

max-drake:
mcp surface looks good to me!

PR #9506, opened 2026-07-06
fix(editor): don't update font states after FontManager is disposed

### LINKED ISSUE #9114: Unhandled "AtomMap: key not found" crash when a font fails to load

frolic:
In order to stop unhandled `AtomMap: key [object Object] not found` rejections when an editor is disposed while fonts are still loading, this PR guards the `FontFace.load()` callbacks in `FontManager.ensureFontIsLoaded` with a `fontStates.has(font)` check. `dispose()` clears the `fontStates` map, so any font that resolved or rejected after disposal tried to `update()` a missing key, and `AtomMap.update` throws on missing keys — the error-handling path itself became an unhandled rejection. Closes #9114.

The per-font `has` check (rather than a manager-wide disposed flag) keeps the existing dispose-then-reuse behavior working: `dispose()` acts as a reset, and fonts re-requested afterwards get a fresh state entry that late callbacks may still legitimately update. On the success path the guard also skips the post-load `document.fonts.add`, so a disposed manager no longer touches the document's font set; the failure path still logs the load error before bailing.

Root-caused in tldraw/tldraw-internal#674 while fixing the desktop app's reopen error storm; the desktop app worked around it, but the SDK race remained for anyone unmounting an editor during font loading.

### Change type

- [x] `bugfix`

### Test plan

1. Delay `FontFace.prototype.load` (e.g. via devtools override or a mock)
2. Mount an editor with text shapes, then unmount it before the font loads resolve
3. Verify no unhandled `AtomMap: key ... not found` rejection is thrown

- [x] Unit tests

### Release notes

- Fix an unhandled `AtomMap: key not found` error when a font finishes loading or fails to load after the editor has been disposed.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +5 / -0    |
| Tests     | +43 / -0   |

### DIFF
--- a/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
+++ b/packages/editor/src/lib/editor/managers/FontManager/FontManager.test.ts
@@ -144,6 +144,49 @@ describe('FontManager', () => {
 		})
 	})
 
+	describe('dispose during font loading', () => {
+		it('does not throw when a font finishes loading after dispose', async () => {
+			const font = createMockFont()
+			let resolveLoad: () => void
+			;(global.FontFace as Mock).mockImplementationOnce(function (family: any) {
+				return {
+					family,
+					load: vi.fn(() => new Promise<void>((resolve) => (resolveLoad = resolve))),
+				}
+			})
+
+			const loadingPromise = fontManager.ensureFontIsLoaded(font)
+			fontManager.dispose()
+			resolveLoad!()
+
+			await expect(loadingPromise).resolves.toBeUndefined()
+			// only the creation-time add from findOrCreateFontFace, not a second
+			// post-load add after dispose
+			expect(document.fonts.add).toHaveBeenCalledTimes(1)
+		})
+
+		it('does not throw when a font fails to load after dispose', async () => {
+			const font = createMockFont()
+			const error = new Error('Font load failed')
+			let rejectLoad: (err: Error) => void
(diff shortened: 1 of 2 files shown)


frolic:
Duplicate of #9494, which takes the stronger identity-check approach — test coverage and edge fixes from this PR have been folded into that branch.

Sent from my aiPhone

PR #9509, opened 2026-07-06
fix(tldraw): don't highlight a fill dropdown item when the current fi…

### LINKED ISSUE #9339: More fills button in style panel doesn't select correctly
### What happened?

When clicking on the rightmost Fill button in the style panel, a side panel opens with more fill options. This has a bunch of confusing small UI bugs:
- By default the leftmost icon (Fill -- Pattern) in the side panel is selected, despite not being clicked by the user yet. This appears selected at the same time as the currently selected fill button.
- After selecting a side panel icon, the rightmost button shows no indication of being selected.
- The rightmost button can't be "scrubbed" onto.
- The rightmost button has no preview when being "scrubbed" off of, until the mouse pointer is released.

Semi-relatedly, the different fill styles have very strange naming - might be worth making a separate issue for that.

https://github.com/user-attachments/assets/60b36ce1-d10e-4c58-baca-2b1f8480b8a7

Chrome

### Where did this happen?

meg-an31:
Fixes how the style panel highlights the active fill when the current fill lives in the overflow dropdown.

Two related changes:

- **Don't highlight a dropdown item when the current value isn't in the dropdown.** Previously, hovering the rightmost fill could pre-select a fill option before one was chosen. Items now only show as active when the current value actually belongs to that dropdown (`valueInItems`).
- **Highlight the overflow dropdown trigger when its active item is selected.** The overflow dropdown's trigger button now shows as active when the current value is one of the items inside it, via a new `isOverflow` prop on `StylePanelDropdownPicker`.

Closes #9339

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Create a shape and select a light fill (2nd or 3rd in the list).
2. Click the 4th fill icon.
3. None of the previous 3 fills should stay selected — all should be blank.
4. Select a fill that lives in the overflow dropdown and confirm the overflow trigger shows as active.

### API changes

- Added optional `isOverflow` prop to `StylePanelDropdownPickerProps` — marks a dropdown as the overflow of another radio group so its trigger shows active when the group's active item is inside it.

### DIFF
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -3670,6 +3670,7 @@ export const StylePanelDropdownPickerInline: <T extends string>(props: StylePane
 export interface StylePanelDropdownPickerProps<T extends string> {
     // (undocumented)
     id: string;
+    isOverflow?: boolean;
     // (undocumented)
     items: StyleValuesForUi<T>;
     // (undocumented)
--- a/packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
+++ b/packages/tldraw/src/lib/ui/components/StylePanel/DefaultStylePanelContent.tsx
@@ -173,6 +173,7 @@ export function StylePanelFillPicker() {
 					items={STYLES.fillExtra}
 					value={fill}
 					sideOffset={116}
+					isOverflow
 				/>
 			</TldrawUiToolbar>
 		</>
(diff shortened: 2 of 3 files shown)


meg-an31:

steveruizok (approved):
Good one!

Issue #9441, opened 2026-06-28
MCP app iframe CSP blocks cdn.tldraw.com` — exec tool always times out after first use


boglim1984:
Summary
The tldraw MCP app embedded in Claude.ai times out on nearly every exec call with Exec failed: Callback timed out after 30000ms for channel "exec". After extensive console debugging, the root cause is clear: the MCP app iframe's Content Security Policy blocks cdn.tldraw.com, which the tldraw editor needs to initialize.
Environment

Chrome (latest)
Claude.ai with tldraw MCP connector enabled
MCP URL: https://tldraw-mcp-app.tldraw.workers.dev/mcp

Console errors
Connecting to 'https://cdn.tldraw.com/5.1.1/translations/en.json' violates the 
following Content Security Policy directive: 
"connect-src 'self' https://tldraw-mcp-app.tldraw.workers.dev"

Fetch API cannot load https://cdn.tldraw.com/5.1.1/translations/en.json. 
Refused to connect because it violates the document's Content Security Policy.

Connecting to 'https://cdn.tldraw.com/5.1.1/watermarks/watermark-track.svg?
version=5.1.1&license_type=evaluation...' violates the following Content Security 
Policy directive: "connect-src 'self' https://tldraw-mcp-app.tldraw.workers.dev"
What's happening
The iframe at fef96fbdfefca0b5f5ede7521201f887.claudemcpcontent.com/mcp_apps has a connect-src that only allows 'self' and https://tldraw-mcp-app.tldraw.workers.dev. But the tldraw editor inside the iframe tries to fetch:

https://cdn.tldraw.com/5.1.1/translations/en.json (i18n)
https://cdn.tldraw.com/5.1.1/watermarks/watermark-track.svg (evaluation license watermark)

Both are blocked. The editor never fully initializes, the oncalltool handler gets replaced mid-flight (visible as [MCP Apps] oncalltool handler replaced. Previous handler will no longer be called.), and the exec callback is never delivered.
Reproduction

Connect tldraw MCP in Claude.ai
Ask Claude to run any exec call
Observe 30s timeout
Open Chrome DevTools on the Claude.ai tab — CSP violations are visible immediately when the exec fires

Expected behavior
exec calls complete successfully and shapes appear on the canvas.
Observed behavior
Consistent 30s timeout. The tool worked exactly once (likely when cdn.tldraw.com assets were warm in browser cache from a prior session).
Fix
[...]

PR #9513, opened 2026-07-06
fix(mcp-app): make exec resilient to host MCP session routing

max-drake:
In order to stop `exec` calls from hanging for 30 seconds and reporting failure on canvases that had actually rendered (relates to #9441), this PR makes the widget→server exec result handoff independent of how the host routes MCP sessions.

The old design blocked the `exec` tool on an in-memory promise keyed by MCP session, assuming the widget's `_exec_callback` would arrive on the same session. Instrumented testing showed hosts don't guarantee that: Cursor uses one session for everything (worked), while Claude gives the model a fresh session per tool call and routes all widget-initiated calls over a separate long-lived session — so the callback always landed on a different Durable Object and was dropped, guaranteeing the timeout.

### How it works now

- **Rendezvous keyed by the code itself.** The server and the widget independently derive `execKey = sha256(code)` — the one value both sides share regardless of session routing — which addresses a rendezvous Durable Object (`CanvasStore`, `exec:<key>` instances). A callback landing on the wrong session DO forwards the result there; the blocked exec races its local pending promise against the rendezvous and takes whichever answers first. Early-arriving results are stashed, so ordering doesn't matter.
- **Bounded, graceful wait.** The 30s hang becomes a 4s window (8s for code-editor hosts, whose cold-start callbacks measured ~4.3s). On timeout, exec returns a non-error "your code is executing" result instead of failing — the widget independently pushes canvas state to the model context, so a missed handoff degrades instead of erroring.
- **canvasId travels in the tool result.** New canvases learned their server-assigned id only from the callback response — the exact channel that misroutes. It now rides `structuredContent` on every exec result, which hosts deliver to the widget reliably, keeping multi-turn "edit the same canvas" working.
- **Widget fixes found along the way.** Claude's event ordering made every exec run its code twice (`ontoolresult` reset the run-once guard while a debounced run was still queued); models passing arrays to variadic proxy methods (`editor.select([...])`) crashed with `Cannot read properties of undefined (reading 'id')`. Both fixed.
- **Cached-URI compatibility.** Hosts cache the widget resource URI durably (across connector re-adds), so a catch-all resource template serves the widget for any `ui://show-canvas/…` variant instead of returning `-32602`.
- **Dev tunnel.** `dev-tunnel.sh` now uses a named cloudflared tunnel with a stable per-user hostname, so hosted MCP clients only need to be registered once.

`apps/mcp-app/hardening-explainer.html` is a self-contained interactive walkthrough of the bug and the fix (open it in a browser).

Verified against a dev tunnel with instrumented session logging: Cursor resolves inline (~1.1s warm), Claude resolves via the rendezvous (~1.6s), runtime errors reach the model as proper `isError` results, and canvases restore correctly across calls on both hosts.

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn workspace mcp-app dev:tunnel` and register the tunnel URL as a connector in Claude and Cursor.
2. In each host, ask for a drawing: the server log should show `exec succeeded via callback after <N>ms` (not the 30s timeout), and the widget dev log exactly one `Exec succeeded` per call.
3. Ask for an edit to the same drawing: the widget should restore the existing shapes before executing.
4. Ask for code that throws (for example, referencing a shape that doesn't exist): the model should receive the runtime error and retry, instead of reporting success.

### Release notes

- Fix tldraw MCP canvas `exec` calls timing out after 30 seconds in Claude; results (including runtime errors) now reach the model in ~2 seconds.

### Code changes

| Section | LOC change   |
| ------- | ------------ |
| Apps    | +1405 / -161 |


max-drake:
note: I disconnected the worker git watch thing because that was setup to deploy the mcp app from main, instead of prod, and we have that set up in our github CI already. 

the failing build here is because this pr includes a DO migration and our upload step in the cf dash runs (ran, now) `wrangler versions upload` instead of a full `wrangler deploy`. upon investigating this error I realized we didn't need the git watcher at all, and in fact it was causing the mcp app to get out of sync with dotcom

apps/mcp-app/src/shared/exec-key.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Rendezvous key for handing the widget's exec result back to the server.
+ *
+ * The server (from the tool args) and the widget (from the code it executes)
+ * derive the same key independently — the exec code itself is the only value
+ * both sides are guaranteed to share regardless of how the host routes MCP
+ * sessions. Used to address the `exec:<key>` CanvasStore rendezvous DO.
+ *
+ * Runs in both the Cloudflare worker and the widget (browser).
+ */
+export async function computeExecKey(code: string): Promise<string> {
```
kostyafarber: this can cause collisions two separate users produce the same exec code, it'll hash to the same key. i.e identical code from different users can collide, steal results etc. not sure if there is a better way to get a true invocation id here but worth exploring a bit further

kostyafarber (approved):
gonna approve modulo my exec key comment, otherwise rest looks good.

Issue #9444, opened 2026-06-29
Cancel any modifier timeouts on pointer down


steveruizok:
🗣️: Cancel any modifier timeouts on pointer down

---

The editor debounces modifier-key releases: when shift, alt, ctrl, or meta is released, the editor waits a short delay before actually clearing that modifier's state (to smooth over OS-level key flicker and key repeat). This leaves a window where a just-released modifier is still treated as held. A concrete symptom: cmd-select something, release cmd, then start dragging — the drag snaps for a moment because the meta modifier is still considered active during the timeout window. On any pointer down, the pending modifier-release timeouts should be cancelled and the affected modifiers treated as immediately released, so the interaction starts with the correct modifier state.

## Open questions

1. **When a pointer down cancels a pending modifier timeout, should the modifier be treated as immediately released (state cleared now), or kept active for the duration of the interaction?**
   Immediately released — clear the modifier state right away on pointer down.
2. **Is this prompted by a specific bug you've hit, or is it a preventative cleanup?**
   Specific bug: cmd-select then drag snaps for a moment, because the released cmd/meta modifier is still treated as held during the debounce window.
3. **Should this apply to all pointer-down events, or only to ones that start a real interaction?**
   Any pointer down.

Confidence: 85%, ready to get started.

PR #9525, opened 2026-07-07
spike(editor): modifier timer free latch

kaneel:
Spike (not for merge): explore removing the 150ms modifier-release debounce timers in favour of a timer-free "latch during gesture" model.

Idea I got while working on https://github.com/tldraw/tldraw/pull/9523 

Would be amazing to consider this and also put all this behaviour into an editor manager.

- flag on → modifier on immediately
- flag off while pointing → latch (defer release to gesture end), so a modifier released just before pointer up still applies to the finishing gesture — the release race is protected without a timer
- flag off while idle → release immediately, so nothing lingers into the next interaction (covers the cmd-release-then-drag snap from #9444)

Race safety is proven by the new release-race test. This is a spike to inform a potental future `ModifierManager`; it intentionally leaves some tests red because they encode the old timer behaviour (`advanceTimersByTime` as the resolution trigger) and one genuine behaviour change (mid-gesture release now latches until the gesture ends instead of clearing after 150ms).

### Change type

- [x] `other` (spike / exploration)

### Test plan

- [x] Unit tests (expected red — see below)

Intentionally failing, documenting the fork:
- `modifiers` Shift/Alt/Ctrl + pointer-down tests — encoded the old timer grace; would move to an event-driven contract.
- `selection-omnibus` mid-gesture tests — mid-gesture release now latches until gesture end.

### Release notes

- Internal spike, no user-facing change.

### DIFF
--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -10905,33 +10905,37 @@ export class Editor extends EventEmitter<TLEventMap> {
 		})
 	}
 
+	/** @internal EXPERIMENT: pending-release flags for the timer-free latch model. */
+	private _shiftKeyPendingRelease = false
+	/** @internal */
+	private _altKeyPendingRelease = false
+	/** @internal */
+	private _ctrlKeyPendingRelease = false
+	/** @internal */
+	private _metaKeyPendingRelease = false
+
 	/**
-	 * Flush any modifier still sitting in its 150ms release-debounce window, so a new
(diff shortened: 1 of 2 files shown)


steveruizok:
This isn't the right behavior for this interaction. If I am dragging something and then press and release shift, it stays in the shift mode; that should not happen. It's more important that we follow conventions among canvases / design tools here. If there's some other way that the interactions are getting a release triggered, then we can chase that; otherwise I think the timer is the right move.

kaneel:
I'll get this right at some point but first, I want to move this behaviour somewhere else to make it more readable

Issue #1714, opened 2023-07-05
Add undo/redo history snapshot


LacombeJ:
### What's the feature?

Functionality in the tldraw API that allows one to get a snapshot of the undo/redo history to persist it. This could be used in the case of switching between tabs and different tldraw contexts.

A proposed usage could be like the following (similar to store snapshots):
```js
const historySnapshot = editor.history.getSnapshot();
editor.history.loadSnapshot(historySnapshot);
```

Currently I am using the below method to manually set the `_redo` and `_undo` properties of the history, but this is a hack because the `stack` function isn't public (I am copying the code from tldraw source to call it).
```js
React.useEffect(() => {
  if (editor) {
    const history = props.restore();
    if (history) {
      editor.history.clear();
      editor.history._redos.set(stack(history.redos) as any);
      editor.history._undos.set(stack(history.undos) as any);
    }

    return () => {
      const undos = editor.history._undos.value.toArray();
      const redos = editor.history._redos.value.toArray();
      props.store({ undos, redos });
    };
  }
}, [props.loaded, editor]);
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #9526, opened 2026-07-07
feat(editor): add undo/redo history snapshots

frolic:
**The motivation.** We want to be able to reload the editor without losing the user's place — the same way hot-module reload swaps out component code without resetting your app state. A host sometimes has to rebuild the `Editor` on the same document: the shape utils, tools, bindings, and UI components an editor is built with are fixed when it's constructed, so the only way to change any of them at runtime is to construct a new editor. When that happens, the document is carried across with `getSnapshot`/`loadSnapshot` and the camera, selection, and current page with the session snapshot — but the undo/redo history isn't. It lives in memory inside the per-editor `HistoryManager` with no capture/restore path, so today rebuilding the editor silently wipes everything the user could have undone. That's the one piece of state that turns an otherwise-seamless reload back into a hard reset.

**What this adds.** `editor.getHistorySnapshot()` and `editor.loadHistorySnapshot(snapshot)` — the history counterpart to the existing document and session snapshots. Capture the history off the old editor, load it into the new one, and undo/redo continue as if nothing happened. The concrete case driving this is a document that carries its own custom shape and tool definitions as code: editing that code has to reconstruct the editor to pick up the new definitions, and this keeps the undo stack intact across that reload. The same applies to toggling a plugin, switching editor modes, or #1714's request to carry history between tabs and editor contexts (people were reaching into `editor.history._undos` by hand to fake it).

**How it works.** A `TLHistorySnapshot` is plain JSON: a format `version`, the store's serialized `schema`, and the `undos`/`redos` stacks as arrays ordered oldest to newest. Capturing is side-effect free — changes not yet committed to the undo stack are included as a final undo entry without flushing them in the live editor. Loading replaces both stacks atomically and discards any uncommitted changes. Mark ids are preserved, so `squashToMark` and `bailToMark` keep working with ids taken before the rebuild.

**When history is dropped.** History entries hold full record values and are never migrated, so a snapshot is only valid against a store with the schema it was captured under. Loading validates the snapshot's structure and requires its schema to match the current store's schema exactly; on any mismatch it warns and leaves the history untouched. This is deliberate: if the editor is rebuilt with a genuinely different set of shapes or tools — a different schema — replaying old diffs against records it no longer understands would be unsound, so the old history is dropped rather than restored. Behavior-only reloads keep history; schema changes drop it.

### Change type

- [x] `feature`

### Test plan

1. Create and move some shapes, then undo a step or two so both stacks are non-empty.
2. Capture with `editor.getHistorySnapshot()`, rebuild the editor on the same document (`getSnapshot`/`loadSnapshot`), then call `editor.loadHistorySnapshot(history)` on the new editor.
3. Undo and redo continue seamlessly across the rebuild; a mark captured before the rebuild still works with `bailToMark`.

- [x] Unit tests

### Release notes

- Add `editor.getHistorySnapshot()` and `editor.loadHistorySnapshot()` for carrying the undo/redo history across editor instances working on the same document.

### API changes

- Added `Editor.getHistorySnapshot()` and `Editor.loadHistorySnapshot(snapshot)`
- Added `HistoryManager.getSnapshot()` and `HistoryManager.loadSnapshot(snapshot)`
- Added the `TLHistorySnapshot` type

### Code changes

| Core code       | +217 / -4  |
| Tests           | +247 / -0  |
| Automated files | +12 / -0   |
| Documentation   | +14 / -0   |

### DIFF
--- a/apps/docs/content/sdk-features/persistence.mdx
+++ b/apps/docs/content/sdk-features/persistence.mdx
@@ -140,6 +140,20 @@ export default function App() {
 }
 ```
 
+### History snapshots
+
+Document and session snapshots don't include the editor's undo/redo history. If you rebuild the editor on the same document — for example, after changing which shape utils or tools are passed to it — you can carry the history across with [Editor#getHistorySnapshot](?) and [Editor#loadHistorySnapshot](?):
+
+```ts
+const history = editor.getHistorySnapshot()
+
+// ...rebuild the editor on the same document...
+
+newEditor.loadHistorySnapshot(history)
+```
+
+History entries contain full record values, so a history snapshot can only be loaded into an editor whose store schema matches the schema it was captured under. If the schemas differ — for example, a shape type was added, removed, or migrated — the snapshot is ignored with a warning. Unlike document snapshots, history snapshots are never migrated: discard them when your schema changes.
+
 ## Custom persistence with store
 
 For more control, create your own store and pass it to the editor. This lets you load data before mounting and implement custom sync logic.
--- a/packages/editor/api-report.api.md
+++ b/packages/editor/api-report.api.md
@@ -1335,6 +1335,7 @@ export class Editor extends EventEmitter<TLEventMap> {
     getHighestIndexForParent(parent: TLPage | TLParentId | TLShape): IndexKey;
     getHintingShape(): NonNullable<TLShape | undefined>[];
     getHintingShapeIds(): TLShapeId[];
+    getHistorySnapshot(): TLHistorySnapshot;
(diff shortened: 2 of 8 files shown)


kaneel (approved):
This could be very useful for SDK consumers as well!

Maybe @mimecuvalo can pitch in if he catches something I would not have thought of but to me, this is a LGTM

frolic:
@steveruizok is thinking maybe we move history into snapshots so they can pick up the migration logic/behavior too, so moving this to draft while we spec it out

PR #9530, opened 2026-07-08
feat(tldraw): flip geo shapes with flipX/flipY like the image shape

### LINKED ISSUE #9373: Flip directional geo shapes to their corresponding shape (rhombus ↔ rhombus-2, arrows)
🗣️: Make certain geo shapes flip when they have corresponding shapes (e.g. rhombus -> rhombus 2, or the arrow geo shapes)

---

Some geo shapes are directional and have a mirror-image counterpart in the geo style set. When one of these is flipped, instead of just mirroring its geometry within its bounding box, it should swap to its corresponding geo type along the matching axis. So a flipped rhombus becomes a rhombus-2, a horizontally flipped right arrow becomes a left arrow, and so on. The type swap alone should produce the correct geometry — no separate mirroring needed on top of it. This should apply both when flipping a single shape and when flipping a multi-shape selection or group. The image shape already models flip state with `flipX` / `flipY` props, and a similar mechanism may be what's needed here so that axis flips are also handled correctly when resizing a shape past its opposite edge, not only via the flip command.

## Open questions

1. **Which geo pairs should swap, and along which axis?**
   Inferred: `rhombus` ↔ `rhombus-2`, `arrow-left` ↔ `arrow-right` (horizontal flip), `arrow-up` ↔ `arrow-down` (vertical flip). Shapes without a counterpart (rectangle, ellipse, trapezoid, etc.) keep today's mirror-in-bounds behavior. Please correct if the rhombus pair should be vertical, or if other shapes have counterparts I missed.
2. **Should the swap apply when flipping a single shape only, or also within a multi-shape / group flip?**
   Both — single-shape flips and multi-shape / group flips should swap directional shapes to their counterparts.
3. **When a shape swaps to its counterpart, should its geometry stay mirrored as today, or should swapping the type replace the need to mirror?**
   Swapping the type alone produces the right geometry; no additional mirroring should be applied.
4. **Should the type swap also happen when a flip occurs during resize (dragging a resize handle past the opposite edge), not just via the flip command?**
   Yes — the resize-flip path should swap directional shapes too. This likely needs the geo shape to carry flip state (similar to image `flipX` / `flipY`) so axis flips during resize are handled consistently.

Confidence: 82%, ready to get started.

kaneel:
TLDR: geo shapes did not flip, it was pretty awkward to resize an arrow until flipping point and seeing it reappear unchanged on the other side of the mirror. 

https://github.com/user-attachments/assets/f1beab5f-7620-4e13-a09e-1090eb556dd0

Closes https://github.com/tldraw/tldraw/issues/9373

### What we did

- **New `flipX` / `flipY` props on the geo shape** (`TLGeoShape`) with validators and an `AddFlipProps` migration (v12) defaulting both to `false`, modeled on the image shape's flip props.
- **Mirror the geometry at the source.** `getGeoShapePath` mirrors the generated `PathBuilder` around the shape's centre when `flipX`/`flipY` is set. Because geometry, hit-testing, the selection indicator, the canvas render (`GeoShapeBody`), and SVG export all flow through this one path, they stay consistent from a single change. The label lives in its own container and is deliberately left unmirrored, so text stays upright and readable.
- **Toggle the props on flip.** `GeoShapeUtil.onResize` sets the new props using a shared helper (`scaleX < 0 !== flipX`), so flips register from both the flip command and any negative-scale resize (group/handle drags), not just an exact `-1`.
- **Extracted the shared behaviour into a util** (`shared/flip.ts`): `getFlipForResize` (the toggle logic, now used by both the image and geo shapes) and `getFlipMatrix` (the mirror matrix). `ImageShapeUtil.onResize` was refactored to use it, removing the duplicated toggle.
- **Added a reusable `PathBuilder.transform(mat)`** that returns a transformed copy of a path (points + bezier control points), preserving fill/dash/close metadata. It uses `Mat.applyToXY` to avoid allocating a `Vec` per point so it stays cheap when many shapes are flipped at once.

### Why this approach

- **No type-swapping "magic".** An earlier idea was to swap `rhombus ↔ rhombus-2` / `arrow-left ↔ arrow-right` on flip. Mirroring the geometry with flip props keeps the shape's identity intact, matches how the image shape works, and covers every geo type (including custom ones) uniformly.
- **Single source of truth.** Mirroring the `PathBuilder` once means we don't maintain four separate flip transforms across geometry/indicator/render/export.
- **Performance-neutral.** The mirrored path is memoized by `getGeoShapePath`'s `WeakCache` and again by the editor's geometry cache, so it's computed once per edit, never per frame; non-flipped shapes pay only a boolean check.

Prior art: this mirrors the fix in #9395, which established that "scale exactly `-1`" and "`resize_bounds` mode" both reduce to "the scale sign flipped on this axis".

### Change type

- [x] `feature`

### Test plan

1. Add an `arrow-left` (or rhombus) geo shape; flip it horizontally (Shift+H) — it visually points the other way and its selection outline/hit area match.
2. Flip a multi-shape selection and a group containing a geo shape; the geo shapes mirror within the selection.
3. Drag a group/selection handle past the opposite edge; directional geo shapes mirror the same as the flip command.
4. Export a flipped geo shape to SVG/PNG; the export matches the canvas.

- [x] Unit tests

### API changes

- Added `flipX` / `flipY` to `TLGeoShapeProps` (with an `AddFlipProps` migration).
- Added `PathBuilder.transform(mat)`.

### Release notes

- Flip geo shapes (rhombus, arrows, trapezoid, etc.) correctly by mirroring their geometry, matching how image shapes flip. Geo shapes now have `flipX` / `flipY` props.

### Code changes

| Core code       | +144 / -5   |
| Tests           | +111 / -2   |
| Automated files | +173 / -39  |
| Apps            | +2 / -0     |
| Templates       | +2 / -0     |


mimecuvalo:
neat idea! added #needs-design-review so that we discuss for next time

steveruizok:
This was my issue so it's blessed in terms of UX. 🇻🇦

mimecuvalo:
nice, lookin' good. two notes:
- shall we enable the Arrange menu? similar to the ctx menu on images:
Screenshot 2026-07-15 at 20 17 33

- let's add some screenshot tests for these shapes (e2e geo shapes)

kaneel:
> nice, lookin' good. two notes:
> 
>     * shall we enable the Arrange menu? similar to the ctx menu on images:
> 
> 
> Screenshot 2026-07-15 at 20 17 33
> 
>     * let's add some screenshot tests for these shapes (e2e geo shapes)

1. yes, I think we should
2. also yes if 1. is yes because the actions make it even easier, let me do this

mimecuvalo (approved):
🤘

PR #9535, opened 2026-07-08
fix(i18n): use forget wording for ja/ko forget-file confirmation

mimecuvalo:
Follow-up to #9521 addressing a late review comment.

The non-owner forget-file confirmation dialog (`81666c6e4a`) used permanent-delete wording in Japanese (削除) and Korean (삭제), while the button and menu label (`Forget` / `Forget file`) use ignore wording (無視 / 무시). Users could believe they were destroying the shared file rather than removing it from their list.

This aligns the confirmation with the forget/ignore verb already used for the button in those locales, matching how fr/de/es keep the confirmation and button consistent.

### Change type
- [x] `bugfix` (translation copy)

### Test plan
- Verified source (`locales/`) and compiled (`locales-compiled/`) files stay consistent; `yarn i18n:compile` produces no further diff.


mimecuvalo:
bad bot

PR #9539, opened 2026-07-08
feat(ci): post weekly dotcom release crew to Discord

mimecuvalo:
Automates a weekly heads-up in `#development` naming the people whose critical fixes / significant features are in the upcoming tldraw.com release — so the right folks are around for the release, without pinging the whole team.

### What's here

- **`skills/dotcom-release-crew/`** — a skill that examines the `production...main` commit range, selects contributors of *critical* changes (features and meaningful fixes in dotcom/editor/sync/store/etc.; excludes docs, chores, tests, version bumps, and trivia), and posts a concise plain-name summary to Discord.
- **`.github/workflows/post-release-crew.yml`** — runs the skill via Claude Code every **Wednesday at 09:00 London time**. GitHub cron is UTC-only, so it triggers at 08:00 and 09:00 UTC and a gate step runs the job only when it's actually 09:00 in London (covers BST/GMT). `workflow_dispatch` allows manual runs, skipping the time gate.

### Before this works

Add a repo secret `DISCORD_RELEASE_WEBHOOK_URL` pointing at the `#development` webhook. The URL is intentionally **not** committed since this repo is public. `ANTHROPIC_API_KEY` is already configured (used by other Claude workflows).

### Notes

- Contributors are listed as plain names (no Discord @-mentions / notifications).
- To change the day, edit the day-of-week field (`3` = Wednesday) in both cron lines.
- Running via `workflow_dispatch` posts to the real `#development` channel — point the secret at a test channel first if you want to dry-run.

### Change type

- [x] `other` (CI / tooling)

### Test plan

- [x] Add the `DISCORD_RELEASE_WEBHOOK_URL` secret, then trigger via `workflow_dispatch` and confirm the message posts to `#development`.

### DIFF
--- a/.github/workflows/post-release-crew.yml
+++ b/.github/workflows/post-release-crew.yml
@@ -34,11 +34,20 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Fail fast if the Discord webhook is not configured
+        run: |
+          if [ -z "$DISCORD_RELEASE_WEBHOOK_URL" ]; then
+            echo "::error::DISCORD_RELEASE_WEBHOOK_URL secret is not set — the release crew message cannot be posted."
+            exit 1
+          fi
+        env:
+          DISCORD_RELEASE_WEBHOOK_URL: ${{ secrets.DISCORD_RELEASE_WEBHOOK_URL }}
+
       - name: Install Claude Code
-        run: npm install -g @anthropic-ai/claude-code
+        run: npm install -g @anthropic-ai/claude-code@2.1.206
 
       - name: Run dotcom-release-crew skill
-        run: claude --skip-permissions --print "/dotcom-release-crew"
+        run: claude --dangerously-skip-permissions --print "/dotcom-release-crew"
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/post-release-crew.yml
+++ b/.github/workflows/post-release-crew.yml
@@ -1,15 +1,15 @@
 name: Post release crew to Discord
 
-# Every Monday at 09:30 UTC, post to #development naming the people whose critical
+# Every Wednesday at 09:30 UTC, post to #development naming the people whose critical
 # fixes / significant features are in this week's dotcom release.
 # GitHub cron is UTC-only. We use a single fixed UTC time (no BST/GMT adjustment) so
 # the job never fires twice: 09:30 UTC is 09:30 in London during winter (GMT) and
 # 10:30 during summer (BST). To change the day, edit the day-of-week field
-# (`1` = Monday) in the cron line.
+# (`3` = Wednesday) in the cron line.
 
 on:
   schedule:
-    - cron: '30 9 * * 1' # Monday 09:30 UTC
+    - cron: '30 9 * * 3' # Wednesday 09:30 UTC
   workflow_dispatch: # allow manual runs
 
 concurrency:
--- a/.github/workflows/post-release-crew.yml
+++ b/.github/workflows/post-release-crew.yml
@@ -1,0 +1,59 @@
+name: Post release crew to Discord
+
+# Every Monday at 09:00 London time, post to #development naming the people whose
+# critical fixes / significant features are in this week's dotcom release.
+# GitHub cron is UTC-only, so we trigger at both 08:00 and 09:00 UTC (covering BST
+# and GMT) and the gate step below runs the job only when it is actually 09:00 in London.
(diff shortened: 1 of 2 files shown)


steveruizok:
Could be on `tldraw-internal`?

.github/workflows/post-release-crew.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Check out code
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Install Claude Code
+        run: npm install -g @anthropic-ai/claude-code
+
+      - name: Run dotcom-release-crew skill
+        run: claude --skip-permissions --print "/dotcom-release-crew"
```
MitjaBezensek: Think this is the right flag.

```suggestion
        run: claude --dangerously-skip-permissions --print "/dotcom-release-crew"
```

skills/dotcom-release-crew/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Dotcom release crew
+
+Identify who contributed **critical fixes or significant features** to this week's dotcom release (the commits on `main` that are not yet on `production`) and post a short summary to the `#development` Discord channel.
+
+The goal is a very tight list of people to involve in the release — **2–3 people at most**, not a changelog and not the whole engineering team. Only exceed 3 in extraordinary weeks where more than three people each own a genuinely critical, release-risky change.
+
+## Inputs
+
+- `DISCORD_RELEASE_WEBHOOK_URL` — environment variable holding the Discord webhook that posts to `#development`. Required. If it is unset, stop and tell the user to set it (do not hardcode a webhook URL — this repo is public).
```
MitjaBezensek, line 14: I guess this will result in a green run if the webhook url is not set? Should we actually fail the action?

.github/workflows/post-release-crew.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Install Claude Code
+        run: npm install -g @anthropic-ai/claude-code
```
MitjaBezensek: Should we pin the version? I guess there isn't a big vector of attack here, but still?

MitjaBezensek (commented):
Seems like the PR description says Wed and the chron is set to Mon, worth updating.

Looking forward to this!

.github/workflows/post-release-crew.yml
mimecuvalo: Applied, thanks — switched to `--dangerously-skip-permissions` in fc15b0d.

skills/dotcom-release-crew/SKILL.md
mimecuvalo: Good catch. Added a fail-fast step that checks `DISCORD_RELEASE_WEBHOOK_URL` before installing/running Claude, so a missing secret now fails the action (red run) instead of silently passing.

.github/workflows/post-release-crew.yml
mimecuvalo: Pinned to `@anthropic-ai/claude-code@2.1.206`.

skills/dotcom-release-crew/SKILL.md
```diff
@@ -0,0 +1,87 @@
...
+```bash
+gh api repos/tldraw/tldraw/compare/production...main --paginate \
+  --jq '.commits[] | [ (.author.login // .commit.author.name), (.commit.author.name), (.commit.message | split("\n")[0]) ] | @tsv'
+```
+
+Each row is: `login`\t`display name`\t`subject`. tldraw squash-merges PRs, so the commit author is the PR author and the subject is the PR title (usually a conventional commit like `feat(editor): ...` with a trailing `(#1234)`).
+
+Also capture the human-readable diff URL for the message: `https://github.com/tldraw/tldraw/compare/production...main`.
+
+If there are **zero** commits, post a brief note that there are no changes to release this week and stop.
```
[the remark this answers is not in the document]
mimecuvalo: Intentional. During normal weeks the dotcom release *is* the `main → production` promotion, so `production...main` is exactly what's about to ship. The freeze-week edge case (production-only hotfixes + main moving ahead) is real but only bites during SDK release weeks, which are all-hands anyway — so a perfectly tailored crew list matters less then. Documented the assumption + caveat in the skill's Notes (c8178f1) rather than adding logic for it.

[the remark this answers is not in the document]
mimecuvalo: this is fine

PR #9543, opened 2026-07-08
fix(editor): give rich text headings proportional line-height

mimecuvalo:
Headings (`h1`–`h6`) inside rich text rendered with their lines crammed together — an H1 label overlapped itself vertically.

This is a regression from #8981 (*resolve line-height to whole pixels for cross-browser text*). That change replaced tldraw's unitless line-height multiplier (`1.35`) with a resolved whole-pixel value set on the `.tl-rich-text` container. Body text is unaffected, but headings render larger than the base (the browser's default `h1 { font-size: 2em }`), so inheriting a pixel line-height derived from the base font size leaves the taller lines too close together.

The fix gives headings a line-height that scales with their own font-size, using tldraw's unitless line-height multiplier. The multiplier is exposed as a `--tl-rich-text-heading-line-height` CSS variable in all three paths that render or measure rich text — on-canvas render (`RichTextLabel`), SVG export (`RichTextSVG`), and text measurement (`TextManager`) — so geometry stays consistent across render, export, and measurement. Body text keeps the whole-pixel line-height from #8981, preserving that cross-browser fix.

before:

Screenshot 2026-07-08 at 17 13 43

after:

Screenshot 2026-07-08 at 17 49 32

### Change type

- [x] `bugfix`

### Test plan

1. Create a geo or text shape and enter rich text edit mode.
2. Apply the H1 (heading) style to a few lines of text.
3. Confirm the heading lines are spaced comfortably and no longer overlap, both while editing and when rendered statically.
4. Export the shape to SVG and confirm heading spacing matches the canvas.

- [x] Unit tests (updated `getSvgString` snapshot; existing rich text / `TextManager` / `TextShape` tests pass)
- [ ] End to end tests

### Release notes

- Fixed rich text headings (H1–H6) rendering with line-height too tight, causing overlapping lines.


steveruizok (approved):

frolic:
do we have/can we add snapshot/screenshots for the multi line headings? it wasn't clear from the snapshots in the diff that these got fixed

just wanna capture this for future so we don't regress again

mimecuvalo:
> do we have/can we add snapshot/screenshots for the multi line headings? it wasn't clear from the snapshots in the diff that these got fixed
> 
> just wanna capture this for future so we don't regress again

yup! indeed — i had to tweak the e2e tests b/c this wasn't captured before. the new `Export-snapshots-Text` screens will help with regressions

PR #9546, opened 2026-07-09
fix(tldraw): fix toast title and content for unbreakable text

### LINKED ISSUE #9536: Long toast content is not truncated and overflows
### What happened?

Toast content (title and description) is not truncated, so a long message causes the toast to grow and overflow instead of clamping to a reasonable width.

Image

Show a toast with a long title or description, e.g. `editor.addToast({ title: "<very long string>" })`.

_No response_

### Where did this happen?

kaneel:
Fixes https://github.com/tldraw/tldraw/issues/9536

Unbreakable text does not display well, we need the css to account for potential long urls / text and/or filepath to be correctly broken

Screenshot 2026-07-09 at 10 38 33
image

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

Open a toast with some url into it and see if it display nicely

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Fixed long toast and dialog content overflowing instead of wrapping. Long text and unbroken strings like URLs now wrap within the toast or dialog.


mimecuvalo (approved):
LG 👍

PR #9550, opened 2026-07-09
feat(commenting): persist comments in Postgres instead of R2

MitjaBezensek:
In order to have a single durable store for dotcom canvas comments, this PR removes the R2 `${key}/comments` object lane and makes Postgres (the DB zero replicates from) the sole persistence for comment records. Comments still sync as records over the room websocket — guests, latency, and the commenting toolkit are unchanged; only where the file's Durable Object loads and stores them changes.

How it works:

- **Write path:** each committed diff appends touched comment/thread ids to a durable outbox in the DO's SQLite; a serialized drain pushes rows to Postgres with clock-guarded upserts (`WHERE lastChangedClock < excluded`), so replays are WAL-silent and zero sees no redundant mutations. Delivery is at-least-once: outbox entries clear only after Postgres acks, failed ids stay queued.
- **Read path:** on cold room open the DO merges the R2 document snapshot with comment records rebuilt from Postgres rows (the two fetches overlap, so PG adds no serial latency), clamping the snapshot clock up to the highest comment clock and declaring lost sync history so reconnecting clients full-resync instead of diverging.
- **Storage shape:** typed columns are the canonical store (no serialized record blob) — `anchor` jsonb, `resolvedAt`/`resolvedBy`, `editedAt`, `meta`, `lastChangedClock` — so record-shape changes have one migration surface. Persistence columns are absent from the zero schema and never replicate to clients. `comment.shapeId` is gone; the `/comments` inbox joins the new `comment_thread` table via a zero relationship instead (no cross-row staleness).
- **Restore:** restoring a document version now drops the file's comments (lane wipe + scoped outbox clear + Postgres delete, ordered so comments created mid-restore still persist).
- **User deletion:** when a drain hits the dead `authorId` FK after a user-row cascade, the comment is pruned from the live room, mirroring the cascade.

Migration 038 is edited in place (it never reached main).

> [!IMPORTANT]
> Deploy note: any database that ran the previous version of migration 038 (the commenting preview env, local dev stacks) must drop/recreate the comment tables — the old table shape makes every app room open fail. Locals: `yarn dev:clean` in zero-cache.

### Change type

- [x] `improvement`

### Test plan

1. Add comments/threads in a file → rows appear in `comment_thread`/`comment` with correct denormalized columns
2. Kill the dev worker (or open the file cold) → comments rehydrate from Postgres, edits to them persist
3. Restore an older document version → comments disappear from the canvas and from Postgres; comments added during/after restore survive
4. `/comments` inbox still lists comments with working shape deep-links (via the thread relation)

- [x] Unit tests

### API changes

- Added `isCommentId()` and `isCommentThreadId()` type guards to `@tldraw/tlschema`, symmetric with `createCommentId()`/`createCommentThreadId()`

### Code changes

| Core code       | +94 / -11   |
| Tests           | +346 / -0   |
| Automated files | +6 / -0     |
| Apps            | +537 / -136 |

### DIFF
--- a/apps/dotcom/sync-worker/src/TLFileDurableObject.ts
+++ b/apps/dotcom/sync-worker/src/TLFileDurableObject.ts
@@ -1065,6 +1065,14 @@ export class TLFileDurableObject extends DurableObject {
 		try {
 			const key = getR2KeyForRoom({ slug, isApp: this.documentInfo.isApp })
 
+			// Kick off the Postgres comments load now so it overlaps with the R2 fetch below;
+			// only the merge points further down actually await it.
+			const commentsPromise = this.documentInfo.isApp ? this.loadCommentsFromPostgres() : null
+			// Prevent an unhandled rejection if we exit via a path that never merges (e.g.
+			// ROOM_NOT_FOUND). Merge points still await commentsPromise itself, so a Postgres
+			// failure there still fails the room open.
+			commentsPromise?.catch(() => {})
+
 			// when loading, prefer to fetch documents from the bucket
 			const r2FetchTimer = this.timer()
 			const roomFromBucket = await this.r2.rooms.get(key)
@@ -1078,8 +1086,8 @@ export class TLFileDurableObject extends DurableObject {
 				// seed the room and hydrate to clients on connect. A Postgres failure here fails the
 				// room open (bubbling like an R2 failure) — silently opening without comments would
 				// let the next persist treat them as deleted.
-				if (this.documentInfo.isApp) {
-					mergeCommentDocumentsIntoSnapshot(snapshot, await this.loadCommentsFromPostgres())
+				if (commentsPromise) {
+					mergeCommentDocumentsIntoSnapshot(snapshot, await commentsPromise)
 				}
 
 				loadTimer.report('db_load_total')
@@ -1114,7 +1122,7 @@ export class TLFileDurableObject extends DurableObject {
 				// here too. Clone the shared DEFAULT_INITIAL_SNAPSHOT constant — the merge reassigns
 				// `documents` and clamps clocks, and must not mutate the module-level object.
 				const snapshot: RoomSnapshot = { ...DEFAULT_INITIAL_SNAPSHOT }
-				mergeCommentDocumentsIntoSnapshot(snapshot, await this.loadCommentsFromPostgres())
+				mergeCommentDocumentsIntoSnapshot(snapshot, await assertExists(commentsPromise))
 
 				return {
 					snapshot,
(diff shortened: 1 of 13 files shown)


apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -554,21 +564,51 @@ export class TLFileDurableObject extends DurableObject {
 
 			await this.r2.rooms.put(roomKey, dataText)
 
-			// Version snapshots only contain the drawing data, so we have to restore both the
-			// document and the current comments — loadSnapshotIntoStorage deletes anything not
-			// present in the snapshot, and deleting comments here would leave their projected
-			// Postgres rows behind (storage transactions don't fire onCommittedChanges).
+			// Version snapshots only contain the drawing data. Restoring drops the file's comments
+			// (product decision): loading the bare snapshot wipes the object lane, and the Postgres
```
MitjaBezensek, line 578: Me and Mime discussed that this is probably the way to go for now.

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -1038,6 +1094,14 @@ export class TLFileDurableObject extends DurableObject {
 		try {
 			const key = getR2KeyForRoom({ slug, isApp: this.documentInfo.isApp })
 
+			// Kick off the Postgres comments load now so it overlaps with the R2 fetch below;
+			// only the merge points further down actually await it.
+			const commentsPromise = this.documentInfo.isApp ? this.loadCommentsFromPostgres() : null
```
MitjaBezensek, line 1099: I think this should be fast enough, but definitely worth checking that it doesn't delay the document load too much.

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -1517,117 +1606,244 @@ export class TLFileDurableObject extends DurableObject {
...
+				const threadResult = await runBatchWithFallback(threadUpserts, insertThreadRows)
+				for (const id of threadResult.failedIds) {
+					failedIds.add(id)
+				}
+				// A comment upsert failing the authorId FK means the author's user account was
+				// deleted: Postgres cascaded their comment rows away (ON DELETE CASCADE) while this
+				// warm room still holds the records. Retrying can never succeed, so mirror the
+				// cascade into the room instead: prune the records below and let their outbox
+				// entries clear normally (the Postgres row being absent is already the desired end
+				// state). Threads are untouched — comment_thread has no user FK by design.
```
mimecuvalo, line 1788: checking here: so threads don't get pruned ever?

mimecuvalo (approved):
pre-approved, let's land this bad boy

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
MitjaBezensek: Hard deleting a file would cascade delete the thread.

PR #9551, opened 2026-07-09
improvement: stop DefaultDebugMenu from setting window.editor

frolic:
In order to let host apps own `window.editor` without the SDK deleting it out from under them, this removes the `window.editor` effect from `DefaultDebugMenu` entirely — publishing a global is a host-app concern, not the SDK's.

`DefaultDebugMenu` set `window.editor = editor` on mount and deleted it on unmount (added in #8780 as a console-debugging convenience). But the debug menu only mounts in debug mode, so its unmount fires every time a user toggles debug mode **off**. Host apps commonly set `window.editor` themselves in `onMount` — tldraw.com does, and so does the tldraw desktop app — and the menu's cleanup deleted that host-owned global too (it deleted whenever `window.editor === editor`). Turn debug mode off and `window.editor` vanished until the next editor mount. In the desktop app this broke an e2e test when the global disappeared mid-suite, and more generally it means any app or agent relying on `window.editor` loses it after a debug toggle.

### Approach

The SDK now doesn't touch `window.editor` at all. Exposing a global is a host-app responsibility, and every place inside this repo that wants it already sets it itself in `onMount` (`develop.tsx`, `end-to-end.tsx`, and the e2e fixtures), so nothing internal regresses.

I considered keeping the effect but making it non-destructive (only claim `window.editor` when unset, only clean up what it set). That fixes the clobbering, but it keeps the SDK in the business of owning a global that really belongs to the host — so removal is the cleaner call.

### Backfilling the old behavior

If you relied on the debug menu populating `window.editor`, set it yourself in `onMount`. There's no new API to learn: `isDebugMode` is already public, reactive instance state — it's the same flag the SDK reads to decide whether to show the debug panel, so it doubles as an "is the debug menu open?" signal.

```ts
// Simplest: always expose it.
<Tldraw onMount={(editor) => { (window as any).editor = editor }} />
```

```ts
// Or reproduce the exact old debug-only behavior with the reactive isDebugMode flag:
<Tldraw
	onMount={(editor) =>
		react('expose editor on window', () => {
			if (editor.getInstanceState().isDebugMode) {
			} else if ((window as any).editor === editor) {
				delete (window as any).editor
			}
		})
	}
/>
```

`react()` re-runs whenever the state it reads changes, and returning its disposer from `onMount` ties the reaction's lifetime to the editor.

### Follow-up

Once this lands and releases, the tldraw desktop app can drop its `DesktopDebugMenu` override (a copy of `DefaultDebugMenu` with the `window.editor` effect removed) and go back to the default component.

### Change type

- [x] `improvement`

### Test plan

1. In an app that sets `window.editor` in `onMount`, toggle debug mode off and confirm the reference survives.
2. In an app that doesn't set it, confirm `window.editor` is never populated by the debug menu.

### Release notes

- `DefaultDebugMenu` no longer sets `window.editor`. Exposing the editor as a global is now a host-app responsibility — set it yourself in `onMount` (gate on `editor.getInstanceState().isDebugMode` if you want the old debug-mode-only behavior). This also fixes the debug menu deleting a host-set `window.editor` when debug mode was toggled off.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +1 / -16   |


mimecuvalo:
heya, i added this in #8780 just for my own ease of debugging on production.
i'd like to keep this around at least for dotcom if possible?

frolic:
> heya, i added this in #8780 just for my own ease of debugging on production. i'd like to keep this around at least for dotcom if possible?

Ah, I don't think we should do this at the SDK level though. Polluting/mutating globals should be user land IMO.

It looked like we were already putting this on the window global in dotcom:
https://github.com/tldraw/tldraw/blob/b7898239508e834517af03cbb3b5bb83534b0411/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx#L134

Was that not the case when you added it in #8780? Perhaps it's missing from other properties like examples, etc.?

mimecuvalo:
> > heya, i added this in #8780 just for my own ease of debugging on production. i'd like to keep this around at least for dotcom if possible?
> 
> Ah, I don't think we should do this at the SDK level though. Polluting/mutating globals should be user land IMO.
> 
> It looked like we were already putting this on the window global in dotcom:
> 
> https://github.com/tldraw/tldraw/blob/b7898239508e834517af03cbb3b5bb83534b0411/apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx#L134
> 
> Was that not the case when you added it in #8780? Perhaps it's missing from other properties like examples, etc.?

oh yeah, misremembered here, yeah i guess dotcom is a separate entry point. i think the pain point here was testing people's PRs. we don't auto-deploy dotcom for PRs (we're limited on Cloudflare's side) so when examining example preview deploys it was useful to have this there.

def. i could see it feeling kinda icky in the SDK but it since it's tied to usage of the Debug menu, it feels like that's an ok side effect. we can debate it more in the design review though 👍

frolic:
decision:
- add a tldraw/editor option to `enableDebugEditorGlobal` that is default on (to preserve current behavior) and a way to opt out for users like tldraw offline
- later consider adding a global tldraw registry for all editors and other things to make debugging/introspection easier

PR #9552, opened 2026-07-09
feat(editor): add FEAT_COMMENTING and FEAT_COLLABORATION license flags

mimecuvalo:
In order to gate the new commenting feature behind the license, this PR adds the first two _feature_ flags to `LicenseManager` — `FEAT_COLLABORATION` and `FEAT_COMMENTING` — and uses them to gate the commenting UI in `@tldraw/commenting`. Every existing flag describes the license _type_ (annual, perpetual, internal, watermark, evaluation, native); these are the first flags that gate product _features_.

`FEAT_COLLABORATION` is an umbrella flag for a family of collaboration features, and `FEAT_COMMENTING` is its first sub-feature. The umbrella grants all of its sub-features, so a license carrying `FEAT_COLLABORATION` enables commenting even without the `FEAT_COMMENTING` bit (`commenting = FEAT_COMMENTING || FEAT_COLLABORATION`).

Consumers gate on features through a reactive check that folds in license state: features are enabled in development so SDK developers can build against them, and in production require a valid, active license (grace-period `licensed` and `licensed-with-watermark` both count; unlicensed/expired/pending do not). The gate fails closed while validation is pending.

The commenting toolkit consumes this through a `useCommentingEnabled()` hook and hides the whole feature when unlicensed: no comment tool button, no canvas overlay/pins, and no sidebar, so nothing can be created. Following the pattern from the earlier plugin-gating exploration (#9533), record and tool _registration_ are deliberately left untouched — the schema must exist so synced comment documents load safely; only the rendered UI is gated.

> [!NOTE]
> Stacked on top of #9471 (`commenting`). The base branch is `commenting`, not `main`.

### Concepts

| Term | Type | Meaning |
| --- | --- | --- |
| `FLAGS.FEAT_COLLABORATION` | license bit (`1 << 6`) | Umbrella flag; grants all collaboration sub-features |
| `FLAGS.FEAT_COMMENTING` | license bit (`1 << 7`) | Commenting sub-feature |
| `LicenseFeatureName` | `'collaboration' \| 'commenting'` | Names of the licensable features |
| `useCommentingEnabled()` | `() => boolean` | Reactive gate the commenting components (and custom UI) read |

### API changes

- Added `LicenseManager.getFeatureFlags()` and `LicenseManager.isFeatureEnabled(feature)` (reactive).
- Added the `useLicenseFeatureFlag(licenseManager, feature)` hook and exported `useLicenseContext()` so consumers can reach the manager.
- Added the `LicenseFeatureName` type.
- Added `isCollaborationEnabled` / `isCommentingEnabled` to `ValidLicenseKeyResult`.
- Added `useCommentingEnabled()` to `@tldraw/commenting/canvas`.

(All of the above are `@internal`.)

### Change type

- [x] `api`

### Test plan

1. Mint a test key with a feature flag via `internal/scripts/generate-test-licenses.ts` (now includes commenting and collaboration entries).
2. In production, load the editor with an entitled key and confirm the comment tool button, pins, and sidebar appear and comments can be created.
3. Load with an unentitled key (or no key) and confirm the comment tool button, overlay, and sidebar are all hidden.
4. Confirm everything works in local development regardless of key (features are dev-enabled).

- [x] Unit tests

### Code changes

| Core code       | +128 / -5  |
| Tests           | +103 / -0  |
| Automated files | +17 / -0   |
| Config/tooling  | +20 / -0   |


packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -117,13 +137,30 @@ export class LicenseManager {
 				this.maybeTrack(result, licenseState)
 
 				this.state.set(licenseState)
+				this.featureFlags.set(getEnabledFeatures(result, licenseState, this.isDevelopment))
```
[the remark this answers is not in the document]
mimecuvalo: Good catch — valid for development. `featureFlags` started fail-closed and was only updated in the async validation `.then`, so in dev `isFeatureEnabled` read `false` during the validation window (and stayed `false` if validation rejected, since `.catch` never set it). Fixed in ecd8e0bb8d by setting the dev-enabled feature set eagerly in the constructor (dev enablement doesn't depend on the async result); production keeps the fail-closed default until validation resolves. Added a test asserting features are enabled synchronously in development.

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -22,6 +22,12 @@ export const FLAGS = {
 	// Native means the license is for native apps which switches
 	// on special-case logic.
 	NATIVE_LICENSE: 1 << 5,
+
+	// -- FEATURE FLAGS --
+	// Collaboration is the umbrella flag for collaboration features; it grants all sub-features.
+	FEAT_COLLABORATION: 1 << 6,
```
MitjaBezensek, line 28: In wonder if we need this as a separate flag or if we just handle collaboration at license generation (so we just set all collaboration flags to true when we generate).

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -117,13 +153,30 @@ export class LicenseManager {
...
 				console.error('License validation failed:', error)
 				this.state.set('unlicensed')
 			})
 	}
 
+	/**
+	 * Returns the current map of licensable features to whether they're enabled. Reactive: reading
+	 * this inside a signal recomputes when license validation resolves.
+	 */
+	getFeatureFlags(): Record<LicenseFeatureName, boolean> {
```
MitjaBezensek: Should this be exposed? People might try to mutate things under the hood?

packages/editor/src/lib/license/LicenseManager.ts
```diff
@@ -117,13 +153,30 @@ export class LicenseManager {
 				this.maybeTrack(result, licenseState)
 
 				this.state.set(licenseState)
```
MitjaBezensek: Should we use `transact` around these two set operations? Just so that they are atomic?

packages/editor/src/lib/license/LicenseManager.test.ts
```diff
@@ -508,6 +509,63 @@ describe('LicenseManager', () => {
...
+			// No await: the feature flags should already reflect development (all enabled) rather than
+			// the fail-closed default, which only lifts once async validation resolves.
+			const devLicenseManager = new LicenseManager('', keyPair.publicKey)
+			expect(devLicenseManager.isFeatureEnabled('commenting')).toBe(true)
+			expect(devLicenseManager.getFeatureFlags()).toEqual({
+				collaboration: true,
+				commenting: true,
+			})
+
+			process.env.NODE_ENV = 'test'
```
MitjaBezensek: Put this in finally not to leak it if something goes wrong?

MitjaBezensek (commented):
We remove the comment tool from the toolbar, but if it is registered with a keyboard shortcut I think that shortcut might still work? Worth checking.

packages/editor/src/lib/license/LicenseManager.ts
mimecuvalo: I think I like the idea of being opted into features that may launch within the terms of the license. So if you've obtained the "collab" package you get access to any upcoming features (instead of having to re-roll the license which would be annoying)

MitjaBezensek: Yeah, good point 👍

PR #9557, opened 2026-07-10
chore: share the worker esbuild externals list between dev and CI scripts

MitjaBezensek:
In order to fix the sync-worker dev size reporter failing with `Could not resolve "util/types"`, this PR extracts the esbuild `--external:` list shared by the dev bundle-size watcher (`internal/scripts/workers/dev.ts`) and the CI bundle-size check (`internal/scripts/check-worker-bundle.ts`) into a single `WORKER_EXTERNAL_DEPS` module.

The two scripts each hard-coded the same list and they drifted: `check-worker-bundle.ts` gained `util` and `util/types` (needed since `pg` 8.22 does a bare `require('util/types')`), but the dev watcher's copy never got the same update, so `yarn dev` printed a resolve error instead of the worker size. Sharing one list fixes the error and prevents the drift from recurring. The worker itself was unaffected — wrangler with `nodejs_compat` resolves these builtins at runtime.

### Change type

- [x] `other`

### Test plan

1. `yarn check-bundle-size` in `apps/dotcom/sync-worker` — passes.
2. `yarn dev` — the "worker is roughly Xkb" report appears again with no `util/types` error.

### Code changes

| Config/tooling | +32 / -41  |

### DIFF
--- a/internal/scripts/check-worker-bundle.ts
+++ b/internal/scripts/check-worker-bundle.ts
@@ -3,6 +3,7 @@ import { resolve } from 'path'
 import kleur from 'kleur'
 import minimist from 'minimist'
 import { exec } from './lib/exec'
+import { WORKER_EXTERNAL_DEPS } from './lib/worker-externals'
 
 const args = minimist(process.argv.slice(2))
 const sizeLimit = Number(args['size-limit-bytes'])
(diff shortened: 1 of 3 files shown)


mimecuvalo (approved):
nice

PR #9558, opened 2026-07-10
fix(dotcom): only rewrite .schema.js when the bundled schema changes

MitjaBezensek:
In order to stop `yarn dev-app` from intermittently wedging at "Waiting for Zero...", this PR makes the schema bundler write `.schema.js` only when the bundled content actually changed.

The race: `dev.ts` bundles the schema once, then starts `bundle-schema:watch` and nodemon (`--watch ./.schema.js --exec zero-cache-dev`) at the same time. esbuild's watch mode rewrites its outfile on the initial build even though the content is identical, so nodemon would SIGINT zero-cache mid-boot. Depending on timing, zero-cache's non-zero exit landed nodemon in its crashed state ("app crashed - waiting for file changes before starting..."), where it only restarts on another `.schema.js` change that never comes. `dev.ts` monitors nodemon itself (still alive), so the orchestrator never noticed and the whole stack polled "Waiting for Zero..." forever.

The fix replaces the CLI esbuild invocations with a small script using esbuild's API (`write: false` plus a write-if-changed plugin). The initial watch build now never touches `.schema.js`, so nodemon restarts only on real schema changes. Verified: clean boot with no spurious restart, identical rebuilds skip the write, and a real `tlaSchema.ts` change still rebuilds and restarts zero-cache.

### Change type

- [x] `bugfix`

### Test plan

1. Run `yarn dev-app` and confirm zero-cache boots without a nodemon restart or crash.
2. Change `packages/dotcom-shared/src/tlaSchema.ts` and confirm `.schema.js` rebuilds and nodemon restarts zero-cache.
3. Touch `tlaSchema.ts` without a content change and confirm nodemon does not restart.

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +50 / -2   |


mimecuvalo (approved):

MitjaBezensek:
Closed in favor of #9561, which removes the schema-restart apparatus (including this file) entirely.

PR #9562, opened 2026-07-10
feat(commenting): add "only unread" sidebar filter

MitjaBezensek:
In order to let users find comments they haven't seen yet, this PR adds an "Only unread" toggle to the comments sidebar filter menu, next to "Only your threads", and makes reading a thread on canvas actually mark its comments read.

The SDK sidebar doesn't know what's read — read receipts live in dotcom's Zero (#9556), not in the document store. So the sidebar takes a new optional prop, `isCommentUnread(commentId)`, the same way it already takes `resolveName`. When the prop is present, the toggle appears and a thread counts as unread if any of its comments is unread. When it's absent, the toggle is hidden and the filter is ignored (same guard "Only your threads" uses for signed-out users).

`CanvasComments` takes the same `isCommentUnread` plus an `onCommentRead(commentId)` callback: while a thread popover is open, every unread comment on display is reported read (including replies arriving while it's open). Without this, receipts were only written from the `/comments` page toggle, so the filter would never let go of a thread you'd read on canvas.

dotcom wires both from the Zero comments query: a set of unread comment ids (others' comments with no receipt — own comments never count as unread) for `isCommentUnread`, and `app.markCommentRead` for `onCommentRead`. This works because Zero comment ids equal `TLComment` record ids verbatim. Receipts update the `comments$` signal, so the list re-filters live.

Also: when the toggles (not page scoping) empty the list, the sidebar now says "No comments match your filters." instead of "No comments on this page yet." — with "Only unread" on, an emptied list is the steady state.

Known gaps, deliberately out of scope: comments visible via the file room but not yet replicated through Zero briefly count as read, and before the first Zero sync nothing counts as unread (distinguishing "not synced yet" from "no receipts" needs Zero's `resultType` plumbed through `signalizeQuery`).

### Change type

- [x] `feature`

### Test plan

1. As user A, comment on a shared file; as user B, open the file and activate the comment tool.
2. In the sidebar filter menu, enable "Only unread" — only threads with comments B hasn't read remain.
3. Open a thread from the sidebar — it drops out of the filtered list (receipt written on view), and shows as read on B's `/comments` page.
4. With everything read, the sidebar says "No comments match your filters."
5. Signed out (or on a host without read status), the toggle is hidden and the filter has no effect.

### Release notes

- Add an "Only unread" filter to the comments sidebar, and mark comments read when their thread is viewed on canvas, via the new `isCommentUnread` and `onCommentRead` props on `CanvasCommentsSidebar` and `CanvasComments`.

### API changes

- Added optional `isCommentUnread(commentId)` to `CanvasCommentsSidebarProps` and `CanvasCommentsProps`
- Added optional `onCommentRead(commentId)` to `CanvasCommentsProps`
- New `comments.only-unread` and `comments.empty-filtered` translation keys

### Code changes

| Core code       | +80 / -7   |
| Automated files | +5 / -1    |
| Apps            | +33 / -4   |


jsscclr (approved):
LGTM 😄

mimecuvalo (approved):
seems straightforward enough 👍

PR #9564, opened 2026-07-10
refactor(dotcom): remove zero feature flags and frontend polyfill

### LINKED ISSUE #9124: Run dotcom e2e tests against Zero instead of the legacy replicator
🗣️: we should start using zero for e2e tests, we currently still use our replicator.

---

Playwright runs are currently forced onto the legacy sync stack: the client detects automated browsing and falls back to the polyfill that talks to the sync worker's replicator, and the override that could opt back in is checked too late to help. Production now runs fully on Zero, so the e2e suite exercises a sync path users no longer hit, and behavior can drift between what tests verify and what production does (mutation-rejection handling is a known example). The work is to switch the e2e suite over to Zero entirely — the zero-cache server is already part of the dev stack the test runner boots — and to re-validate the test database reset flow on that path. The legacy polyfill path does not need to keep e2e coverage. This is independent infra cleanup, not tied to the workspaces launch.

## Open questions

1. **Should the legacy polyfill path keep any e2e coverage while it remains the kill-switch fallback, or do we switch the suite over entirely?**
   Switch over entirely; no polyfill coverage needed.
2. **Is this blocking for the workspaces launch, or independent infra cleanup?**
   Completely independent.

MitjaBezensek:
In order to stop maintaining a second sync path nobody uses — production has been 100% on Zero for a while — this PR deletes the `zero_enabled`/`zero_kill_switch` feature flags and the frontend Zero polyfill (the Zero-shaped shim over the legacy replicator websocket). The client now always constructs the real Zero client. This also removes the `navigator.webdriver` fallback that kept Playwright on the legacy path, so the e2e suite finally runs against real zero-cache. Closes #9124.

Notable decisions:

- The kill switch goes with no replacement lever: the fallback path it switched to was unmaintained and drifting, so it was fake safety. A Zero outage is now mitigated by deploy/revert.
- Backend custom sync engine code (`TLUserDurableObject`, `UserDataSyncer`, the `/app/:userId/connect` endpoint) stays intact per scope; it just has no frontend consumer anymore. Only the now-dead `downgrade-client`/`upgrade-client` test routes were removed.
- The client-too-old e2e used to flip a protocol flag on the user DO — a signal that only reached the client through the polyfill. Zero owns that detection now (`onUpdateNeeded`), and there's no supported way to force a real schema mismatch from a test, so the test triggers the same `onClientTooOld` callback via a webdriver-gated hook and verifies the reload-recovery UI.
- Known wart: the file-duplicate e2e gained a 1s wait because duplication copies server-side room content, which only reflects shapes that round-tripped the room websocket (`waitForMutationResolution` covers Zero mutations, not room sync). A deterministic room-persistence wait would be better; taking suggestions.
- Local e2e state: two unrelated-looking tests flaked once each across local runs (`published snapshots update only after publishing`, `workspace creation/moving/deletion navigation`) with the whole suite otherwise green. Watching CI's run here for a clean-room verdict.

### Change type

- [x] `improvement`

### Test plan

1. `yarn e2e-dotcom` — full suite now runs against real zero-cache (28 passing locally)
2. Sign in on a fresh dev DB with any email — client must use real Zero (no polyfill path exists)

- [x] Unit tests
- [x] End to end tests

### Code changes

| Section   | LOC change  |
| --------- | ----------- |
| Core code | +1 / -1     |
| Tests     | +49 / -761  |
| Apps      | +85 / -688  |


apps/dotcom/client/e2e/fixtures/scenario-test.ts
mimecuvalo: auto-review:  `__e2e__waitForMutationResolution was only defined on the deleted zero-polyfill.ts`

mimecuvalo (approved):
thx for the cleanup, pre-approving modulo one comment

MitjaBezensek:
Will keep this open since I'm off next week, just to make sure it doesn't break something and I'm not around.

driev (approved):

apps/dotcom/client/e2e/fixtures/scenario-test.ts
MitjaBezensek: Removed in 4c2a4dd046 — the helper died with the polyfill so every call site was a silent no-op; deleted them all. The one spot that held actual mutation handles (setupWorkspaceWithMember) now awaits `.server` instead, which is a real server-ack wait on proper Zero.

PR #9565, opened 2026-07-10
feat: add clear search buttons to the dotcom and examples sidebars

### LINKED ISSUE #9452: Add a "Clear search" button at the end of search results on tldraw.com
🗣️: When searching on tldraw.com, at the end of the search results, I want a "Clear search" button

---

On tldraw.com, the file/sidebar search returns a list of results but offers no obvious way to reset the search. The request is to add a "Clear search" button at the bottom of the results, after a section gap following the last item, that clears the current search query and returns the user to the unfiltered view. The button should appear in both the has-results and zero-results states, and its only behavior is to clear the query. This is a small UX improvement to make exiting a search faster than manually deleting the query text.

## Open questions

1. **Where exactly should the button appear, and should it show in all search states?**
   At the bottom of the list in both the has-results and zero-results states, separated from the last item by a section gap space.
2. **What should the button do beyond clearing the query?**
   Nothing — it should just clear the search query.

m31-galaxy:
In order to make exiting a sidebar search faster than manually deleting the query text, this PR adds a "Clear search" action row at the end of the search results on tldraw.com — shown in both the has-results and no-results states — that clears the query and returns the sidebar to the unfiltered view. Clearing from there also resets and collapses the search input, exactly like the X button in the search box. Closes #9452.

The examples app sidebar had the same gap but worse — its search had no clear affordance at all — so this brings it to parity with the dotcom pattern: an X button in the search box, the same full-width "Clear search" row after the results, and a "No examples found" empty state. The examples page is a plain React shell that can't use tldraw UI components, so parity is achieved by porting the dotcom markup and CSS values onto the examples sidebar's existing classes (the two sidebars already share the same visual system); the X gets a native `title` tooltip instead of the styled tldraw one. This also surfaced that `<button>` elements don't inherit the page font like the sidebar's anchor rows do, so the examples action-row class now sets `font: inherit` — the counterpart of `tla-text_ui__regular` on the dotcom buttons.

One deliberate spacing detail: the gap between the last result section and "Clear search" is half the normal inter-section gap (via a negative margin on an adjacent-sibling selector), so the control reads as attached to the results. The selector intentionally doesn't match in the no-results state, where the empty-state message provides the spacing.

### Change type

- [x] `feature`

### Test plan

1. On tldraw.com, click Search in the sidebar and type a query with matches: an X appears in the box and a full-width "Clear search" row appears after the results.
2. Type a query with no matches: "No files found", then the same "Clear search" row.
3. Click "Clear search": the query clears, the full file list returns, and the search input collapses back to the Search action row (same as the X).
4. In the examples app sidebar, repeat: X in the search box, "Clear search" / "No examples found" at the end of the list; both clear buttons restore the full list and reset the `?filter=` URL param.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Add a "Clear search" button at the end of sidebar search results on tldraw.com and in the examples app sidebar.

### Code changes

| Documentation | +89 / -0   |
| Apps          | +57 / -2   |

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarRecentFiles.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarRecentFiles.tsx
@@ -158,20 +158,24 @@ export function TlaSidebarRecentFiles() {
 			) : null}
 			{isSearching ? (
 				<>
-					<div className={styles.sidebarDivider} />
-					<button
-						className={classNames(
-							styles.sidebarActionButton,
-							styles.hoverable,
-							'tla-text_ui__regular'
-						)}
-						onClick={handleClearSearch}
-						data-testid="tla-sidebar-search-clear-button"
-					>
-						<span className={styles.sidebarActionButtonLabel}>
-							<F defaultMessage="Clear search" />
-						</span>
-					</button>
+					{/* The same section wrapper as the workspace action rows, so the
+					    button gets the same full width and vertical rhythm. The gap
+					    above comes from the file sections' own bottom margin. */}
+					<div className={styles.sidebarSection}>
+						<button
+							className={classNames(
+								styles.sidebarActionButton,
+								styles.hoverable,
+								'tla-text_ui__regular'
+							)}
+							onClick={handleClearSearch}
+							data-testid="tla-sidebar-search-clear-button"
+						>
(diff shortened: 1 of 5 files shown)


mimecuvalo (approved):
nice — as long as you're touching this code, can you tweak the X button so that it's aligned better? pre-approving modulo this tiny nit

Image

mimecuvalo (approved):
ty!

PR #9566, opened 2026-07-10
feat: attribute duplicated and pasted notes to the current user

mimecuvalo:
Note text attribution (`textLastEditedBy`) follows the last person to produce the note's text. Duplicating or pasting a note that has text is a new act of authorship by the current user, so the copy should be credited to them rather than carrying over the original author's identity.

Previously `editor.duplicateShapes()` and `editor.putContentOntoCurrentPage()` (which backs paste and page duplication) spread the original note's props verbatim, so a duplicate/paste kept the original author's `textLastEditedBy`. This matters for the upcoming server-side provenance checks (#9489), which must not attribute content to a user who did not produce it — otherwise a user could duplicate someone else's note and effectively forge their identity.

This PR adds a `ShapeUtil.onDuplicate(source, duplicate)` hook, invoked by `Editor.duplicateShapes` and `Editor.putContentOntoCurrentPage`, giving a shape a last chance to modify a reproduced copy before it's created. `NoteShapeUtil` overrides it to re-stamp `textLastEditedBy` to the current user (`editor.getAttributionUserId()`) when the note has text; empty notes keep no attribution.

### API changes

- Added `ShapeUtil.onDuplicate(source, duplicate)`, an optional hook called when a shape is reproduced via duplicate or paste.

### Change type

- [ ] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

1. As user A, create a sticky note and type some text — attribution shows user A.
2. As user B, duplicate the note (Cmd/Ctrl+D) — the duplicate is attributed to user B; the original still shows user A.
3. As user B, copy the note and paste it — the pasted copy is attributed to user B.
4. Duplicate an empty note — the copy has no attribution.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Duplicating or pasting a sticky note with text now attributes the copy to the current user rather than the original author. Custom shapes can hook into this via the new `ShapeUtil.onDuplicate` callback.

### DIFF
--- a/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
+++ b/packages/editor/src/lib/editor/shapes/ShapeUtil.ts
@@ -670,6 +670,28 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
 	onBeforeCreate?(next: Shape): Shape | void
 
 	/**
+	 * A callback called when a shape is reproduced from an existing shape, either by duplicating
+	 * ({@link Editor.duplicateShapes}) or by pasting/putting content onto the page
+	 * ({@link Editor.putContentOntoCurrentPage}). This provides a last chance to modify the copy
+	 * before it's created — for example, to re-stamp attribution so the copy is credited to the
+	 * current user rather than the original author.
+	 *
+	 * @example
+	 *
+	 * ```ts
+	 * onDuplicate = (source, duplicate) => {
+	 * 	return { ...duplicate, props: { ...duplicate.props, editedBy: this.editor.getAttributionUserId() } }
+	 * }
+	 * ```
+	 *
+	 * @param source - The shape being copied from.
+	 * @param duplicate - The new copy (with its own id), before it's created.
+	 * @returns The next shape or void.
+	 * @public
+	 */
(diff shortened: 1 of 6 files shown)


packages/editor/src/lib/editor/shapes/ShapeUtil.ts
```diff
@@ -669,6 +669,28 @@ export abstract class ShapeUtil<Shape extends TLShape = TLShape> {
...
+	 * 	return { ...duplicate, props: { ...duplicate.props, editedBy: this.editor.getAttributionUserId() } }
+	 * }
+	 * ```
+	 *
+	 * @param source - The shape being copied from.
+	 * @param duplicate - The new copy (with its own id), before it's created.
+	 * @returns The next shape or void.
+	 * @public
+	 */
+	onDuplicate?(source: Shape, duplicate: Shape): Shape | void
```
max-drake: nit but should we call this `onBeforeDuplicate` to match `onBeforeUpdate` / `onBeforeCreate`? The jsdoc  here uses the same "provides a last chance to..." verbiage as those two

max-drake (approved):
looks great!

PR #9577, opened 2026-07-13
feat(collaboration): add collaboration umbrella package

mimecuvalo:
In order to give consumers a single entry point for tldraw's collaboration features, this PR adds `packages/collaboration` — an umbrella package. For now it re-exports `@tldraw/commenting` (both its root and `/canvas` entry points). Over time it will re-export the other collaboration packages as well.

The package mirrors `@tldraw/commenting`'s setup (private, node vitest preset, `tsconfig` project reference) and exposes the same two entry points: `@tldraw/collaboration` and `@tldraw/collaboration/canvas`.

> Note: based on the `commenting` branch, which introduces `@tldraw/commenting`. Retarget to `main` once that lands.

### Change type

- [x] `feature`

### Test plan

- `yarn typecheck` passes.
- `yarn workspace @tldraw/collaboration lint` passes.

### Release notes

- Add `@tldraw/collaboration`, an umbrella package that re-exports the collaboration packages (currently `@tldraw/commenting`).


packages/collaboration/package.json
```diff
@@ -0,0 +1,55 @@
...
+	},
+	"peerDependencies": {
+		"react": "^18.2.0 || ^19.2.1",
+		"react-dom": "^18.2.0 || ^19.2.1"
+	},
+	"engines": {
+		"node": ">=22.12.0"
+	},
+	"dependencies": {
+		"@tldraw/commenting": "workspace:*"
```
frolic, line 53: fly by: thoughts on naming this comments rather than commenting 😁

jsscclr: any particular reason why? I went with commenting since it's everything around enabling commenting, rather than just a dumb comment shape (for example)

jsscclr (approved):

packages/collaboration/package.json
mimecuvalo: agreed that present participle feels a bit much but ¯\\_(ツ)_/¯ 

one thing we can consider (in a different PR) is maybe we just end up having the packages/collaboration and commenting is just a sub-part of that

jsscclr: I don't feel too strongly about it so whatever works! That was my rationale but I wasn't particularly convinced either way, so if people would prefer comments, go for it imo

frolic: just find the naming a bit odd at a glance, usually don't see packages or classes or whatever named with verb -ing

defer to you all though!

PR #9581, opened 2026-07-13
fix(tldraw): prevent restoring shapes with missing parents

### LINKED ISSUE #9567: Undo may restore a shape with a deleted parent during multiplayer
In multiplayer, undo may restore a shape whose parent frame was deleted by another collaborator, leaving the shape parented to a record that no longer exists.

This likely occurs when one collaborator moves a shape inside a frame, another collaborator deletes the frame, and the first collaborator then undoes the move. Undo should reconcile with concurrent changes and preserve a valid shape hierarchy.

## Probable steps to reproduce

1. Open the same multiplayer room in two clients.
2. In client A, create a frame and a shape inside it.
3. Move the shape in client A.
4. Delete the frame in client B.
5. Undo in client A.

## Actual behavior

The shape may be restored with the deleted frame as its parent, leaving an invalid parent reference.

## Expected behavior

Undo should not restore a shape with a missing parent. The shape must retain a valid place in the hierarchy after concurrent changes are reconciled.

## Open questions

1. Can this be reproduced consistently in a live multiplayer room?
   _Deferred; this report describes a probable multiplayer manifestation that still needs confirmation._

Confidence: 74%, ready for investigation.

kaneel:
In multiplayer, undo could leave a shape parented to a record that no longer exists. If one collaborator moves or reparents a shape while another deletes the frame it lived in, undoing the first collaborator's change replays the old state and points the shape back at the deleted frame. The resulting broken shape then syncs out to everyone - and because tldraw tolerates a missing parent by falling back to an identity transform, it doesn't crash, it just silently becomes an untracked, invisible shape in everyone's document. Fixes #9567.

The store integrity check (`ensureStoreIsUsable`) already runs after undo, redo, and remote merges, and already enforces invariants like "instance points at a valid page". This adds one more: a shape's parent must exist. Any shape whose `parentId` refers to a shape that's no longer in the store is removed; removing a shape can orphan its own children, which the checker's existing re-run pass then handles.

Keeping it here rather than in editor side effects means it lives in one place next to the related invariants, and because it runs on the client that made the change, the corrected state is what propagates to collaborators - they never receive the broken shape.

The behavior is uniform on purpose: an orphaned shape is discarded, not reparented. Reparenting to the page was considered and rejected, because the deleted frame's transform is gone, so the shape would reappear at the wrong position.

### Change type

- [x] `bugfix`

### Test plan

1. Open the same multiplayer room in two clients.
2. In client A, create a frame with a shape inside it, then move the shape.
3. In client B, delete the frame.
4. Undo in client A.
5. The shape should be discarded, not left parented to the deleted frame, on both clients.

- [x] Unit tests

### Release notes

- Fix a multiplayer bug where undo could leave a shape parented to a frame another collaborator had deleted, resulting in a broken, invisible shape synced to everyone.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +16 / -0   |
| Tests     | +45 / -0   |

### DIFF
--- a/packages/tlschema/src/TLStore.ts
+++ b/packages/tlschema/src/TLStore.ts
@@ -15,6 +15,7 @@ import { PageRecordType, TLPageId } from './records/TLPage'
 import { InstancePageStateRecordType, TLInstancePageStateId } from './records/TLPageState'
 import { PointerRecordType, TLPOINTER_ID } from './records/TLPointer'
 import { TLRecord } from './records/TLRecord'
+import { isShapeId } from './records/TLShape'
 import { TLUser } from './records/TLUser'
 
 /**
@@ -462,6 +463,7 @@ function getDefaultPages() {
 export function createIntegrityChecker(store: Store<TLRecord, TLStoreProps>): () => void {
 	const $pageIds = store.query.ids('page')
 	const $pageStates = store.query.records('instance_page_state')
+	const $shapes = store.query.records('shape')
 
 	const ensureStoreIsUsable = (): void => {
 		// make sure we have exactly one document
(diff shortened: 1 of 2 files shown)


packages/tlschema/src/TLStore.ts
```diff
@@ -531,6 +533,20 @@ export function createIntegrityChecker(store: Store<TLRecord, TLStoreProps>): ()
...
+		// when a change is undone that references a shape (e.g. a frame) which has since been
+		// deleted by another collaborator: the undo would otherwise leave a shape parented to a
+		// record that doesn't exist. Such a shape has no valid place in the hierarchy, so we
+		// discard it. Removing a shape can orphan its own children, which the re-run then handles.
+		const orphanedShapeIds = $shapes
+			.get()
+			.filter((shape) => isShapeId(shape.parentId) && !store.has(shape.parentId))
+			.map((shape) => shape.id)
+		if (orphanedShapeIds.length > 0) {
+			store.remove(orphanedShapeIds)
```
mimecuvalo, line 546: mm, i know the tasks says it should remove the shape but i'm wondering if that's really the right move.
feels like it should still be restored, just not parented. this might need a #needs-design-review label

kaneel: problem for restoring is that we would not know where to restore it because we would have lost the parent so we can't figure the difference between parent coordinates and ancestor coordinates.

mimecuvalo: i see, ok then

mimecuvalo (approved):

PR #9586, opened 2026-07-14
fix(editor): guard possible non-finite delta values in the wheel events normalization

### LINKED ISSUE #9122: Panning can set the camera to NaN and crash with a validation error
🗣️: Sentry shows panning crashing the app with a camera NaN validation error (LITE-3VH, 550 events / 67 users since June 2024) — let's create an issue for it with the Sentry context.

---

Panning can crash the app with `ValidationError: At camera.x: Expected a number, got NaN`. The wheel/pan tick path computes the next camera position as `cx + (dx * panSpeed) / cz`, and under some input conditions that arithmetic produces NaN; the store's validator then rejects the camera record and the unhandled error crashes the app (`willCrashApp: true`). The event tags show the NaN is introduced in that update (`isExistingValidationIssue: false`, `storePhase: updateRecord`). Camera math should reject/clamp non-finite inputs rather than letting validation be the crash point.

## Sentry context

- Sentry: [LITE-3VH](https://tldraw.sentry.io/issues/LITE-3VH) — **550 events / 67 users**, since 2024-06-04
- Culprit: `Editor._setCamera` via `_flushEventForTick` wheel-pan branch (`packages/editor/src/lib/editor/Editor.ts:11112`)
- Sample event: Firefox 151 / Windows, idle select state, camera updated during scroll-pan

```ts
this._setCamera(new Vec(cx + (dx * panSpeed) / cz, cy + (dy * panSpeed) / cz, cz), {  // ← NaN
```

kaneel:
Fixes #9122

### The issue

Panning with the wheel can crash the app with `ValidationError: At camera.x: Expected a number, got NaN` (Sentry LITE-3VH, 550 events / 67 users since June 2024). The wheel-pan tick path computes `cx + (dx * panSpeed) / cz`; when the wheel delta arrives non-finite, that NaN is written to the camera record and rejected by the store validator, crashing the app.

### Root cause

`normalizeWheel` takes the raw `WheelEvent.deltaX`/`deltaY` and passes them through without a finiteness check. Some devices, drivers, and browsers can emit non-finite wheel deltas (the reported events are Firefox on Windows). That value propagates unchanged into the camera math.

The camera math itself is fine — the bad data enters at the device-input boundary.

### The fix

Guard the deltas in `normalizeWheel`, the single boundary every wheel event passes through. Non-finite `deltaX`/`deltaY` are coerced to `0`, and `useGestureEvents` already ignores zero-delta wheel events, so a garbage event becomes a no-op instead of a crash.

This follows the existing pattern of sanitizing untrusted input at the boundary (e.g. `updateViewportScreenBounds` floors DOM measurements) rather than massaging values inside the editor. Bad data from other sources (e.g. a consumer passing a non-finite value to `setCamera` or `panSpeed`) still fails validation by design.

### Scope

Wheel input only, matching the reported crash. The pinch/gesture boundary (`dispatchPinchEvent`) has the same shape but no reported occurrences, so it is intentionally left for a follow-up.

### Testing

Added `normalizeWheel.test.ts` covering finite pass-through and non-finite `deltaX`/`deltaY`/zoom deltas.

[x] unit tests

### Release notes

- Improved the normalization of the wheel events by guarding possible non-finite delta values

### DIFF
--- a/packages/editor/src/lib/utils/normalizeWheel.ts
+++ b/packages/editor/src/lib/utils/normalizeWheel.ts
@@ -9,6 +9,12 @@ const MAX_ZOOM_STEP = 10
  * @internal */
 export function normalizeWheel(event: WheelEvent | React.WheelEvent<HTMLElement>) {
 	let { deltaY, deltaX } = event
+
+	// Coerce non-finite deltas to no movement so they can't propagate into the camera and
+	// crash validation. Also covers deltaZ below, which is derived from deltaY.
+	if (!Number.isFinite(deltaX)) deltaX = 0
+	if (!Number.isFinite(deltaY)) deltaY = 0
+
 	let deltaZ = 0
 
 	// wheeling
--- a/packages/editor/src/lib/utils/normalizeWheel.test.ts
+++ b/packages/editor/src/lib/utils/normalizeWheel.test.ts
@@ -1,0 +1,39 @@
+import { normalizeWheel } from './normalizeWheel'
+
+function wheelEvent(props: Partial<WheelEvent>): WheelEvent {
+	return {
+		deltaX: 0,
+		deltaY: 0,
+		ctrlKey: false,
+		altKey: false,
+		metaKey: false,
+		shiftKey: false,
+		...props,
+	} as WheelEvent
+}
+
+describe('normalizeWheel', () => {
+	it('negates finite deltas for panning', () => {
+		const { x, y } = normalizeWheel(wheelEvent({ deltaX: 5, deltaY: 10 }))
+		expect(x).toBe(-5)
+		expect(y).toBe(-10)
+	})
+
+	it('coerces a non-finite deltaX to zero movement', () => {
+		const { x, y } = normalizeWheel(wheelEvent({ deltaX: NaN, deltaY: 10 }))
+		expect(x === 0).toBe(true)
+		expect(y).toBe(-10)
+	})
+
+	it('coerces a non-finite deltaY to zero movement', () => {
+		const { x, y } = normalizeWheel(wheelEvent({ deltaX: 5, deltaY: Infinity }))
+		expect(x).toBe(-5)
+		expect(y === 0).toBe(true)
+	})
+
+	it('does not produce a non-finite zoom delta when wheeling with a bad deltaY', () => {
+		const { z } = normalizeWheel(wheelEvent({ deltaY: NaN, ctrlKey: true }))
+		expect(Number.isFinite(z)).toBe(true)
+		expect(z === 0).toBe(true)
+	})
+})


steveruizok:
Do we know where/why/how those NaN values were being produced? Input devices, browser, etc?

In general we try not to sanitize values in the editor (as a design opinion) but rather at the source that produces the bad data, and then crash in validation if bad data makes it through.

kaneel:
> Do we know where/why/how those NaN values were being produced? Input devices, browser, etc?
> 
> In general we try not to sanitize values in the editor (as a design opinion) but rather at the source that produces the bad data, and then crash in validation if bad data makes it through.

Right, and it makes sense, I got carried away. It happens when panning with a wheel, which has its own loc in Editor, which mean something, somewhere, at this point gives us bad data

packages/editor/src/lib/utils/normalizeWheel.ts
```diff
@@ -9,6 +9,12 @@ const MAX_ZOOM_STEP = 10
  * @internal */
 export function normalizeWheel(event: WheelEvent | React.WheelEvent<HTMLElement>) {
 	let { deltaY, deltaX } = event
+
+	// Coerce non-finite deltas to no movement so they can't propagate into the camera and
+	// crash validation. Also covers deltaZ below, which is derived from deltaY.
+	if (!Number.isFinite(deltaX)) deltaX = 0
```
mimecuvalo, line 15: wait, how do they get to be non-finite in the first place? do you have a sense of this gets to be like that?

kaneel: Following the ticket, the trace gives us a clue about where some values may arrive wrong, but I admit it's unverifiable unless I get a wheel mouse :(

mimecuvalo (changes requested):
mm, I think this is looking at the wrong layer maybe. the original issue talks about `this._setCamera(new Vec(cx + (dx * panSpeed) / cz, cy + (dy * panSpeed) / cz, cz), {` which makes me think this needs to be some work in Editor.ts

```

--- a/packages/editor/src/lib/editor/Editor.ts
+++ b/packages/editor/src/lib/editor/Editor.ts
@@ -3392,6 +3392,14 @@ export class Editor extends EventEmitter<TLEventMap> {
              const currentCamera = this.getCamera()

              let { x, y, z = currentCamera.z } = point

+             // Camera values must be finite. Non-finite values can arrive from device input (wheel
+             // and pinch deltas), from a consumer's camera options, or from arithmetic against a
+             // zero zoom. Holding the current value on that axis makes a bad event a no-op, rather
+             // than writing NaN into the camera record, which the store validator rejects — and an
+             // unhandled validation error crashes the app.
+             if (!Number.isFinite(x)) x = currentCamera.x
+             if (!Number.isFinite(y)) y = currentCamera.y
+             if (!Number.isFinite(z)) z = currentCamera.z
+
```

and we can then drop:
```
@@ -3624,10 +3632,6 @@ export class Editor extends EventEmitter<TLEventMap> {
              const _point = Vec.Cast(point)

-             if (!Number.isFinite(_point.x)) _point.x = 0
-             if (!Number.isFinite(_point.y)) _point.y = 0
-             if (_point.z === undefined || !Number.isFinite(_point.z)) point.z = this.getZoomLevel()
```

which is buggy anyway because it does `point.z =` instead of `_point.z =` 🤦 

lemme know if that makes sense!

kaneel:
> mm, I think this is looking at the wrong layer maybe. the original issue talks about `this._setCamera(new Vec(cx + (dx * panSpeed) / cz, cy + (dy * panSpeed) / cz, cz), {` which makes me think this needs to be some work in Editor.ts
> 
> ```
> 
> --- a/packages/editor/src/lib/editor/Editor.ts
> +++ b/packages/editor/src/lib/editor/Editor.ts
> @@ -3392,6 +3392,14 @@ export class Editor extends EventEmitter<TLEventMap> {
>               const currentCamera = this.getCamera()
> 
>               let { x, y, z = currentCamera.z } = point
> 
> +             // Camera values must be finite. Non-finite values can arrive from device input (wheel
> +             // and pinch deltas), from a consumer's camera options, or from arithmetic against a
> +             // zero zoom. Holding the current value on that axis makes a bad event a no-op, rather
> +             // than writing NaN into the camera record, which the store validator rejects — and an
> +             // unhandled validation error crashes the app.
> +             if (!Number.isFinite(x)) x = currentCamera.x
> +             if (!Number.isFinite(y)) y = currentCamera.y
> +             if (!Number.isFinite(z)) z = currentCamera.z
> +
> ```
> 
> and we can then drop:
> 
> ```
> @@ -3624,10 +3632,6 @@ export class Editor extends EventEmitter<TLEventMap> {
>               const _point = Vec.Cast(point)
> 
> -             if (!Number.isFinite(_point.x)) _point.x = 0
> -             if (!Number.isFinite(_point.y)) _point.y = 0
> -             if (_point.z === undefined || !Number.isFinite(_point.z)) point.z = this.getZoomLevel()
> ```
> 
> which is buggy anyway because it does `point.z =` instead of `_point.z =` 🤦
> 
> lemme know if that makes sense!

This is where my habit of rebasing my commit is biting me in the arse, I first looked into the Editor file but @steveruizok told me we should avoid, as a rule, as much as we can, to nudge values in the editor and do it at the source!

mimecuvalo:
hm, ok, i guess this is a needs-design-review ¯\\_(ツ)_/¯

mimecuvalo (approved):
LGTM — thx!

@kaneel can you make a separate PR to address the bug that shook out of this? this line:
`if (_point.z === undefined || !Number.isFinite(_point.z)) point.z = this.getZoomLevel()`

kaneel:
Thanks! Follow-up is up in #9903 — I removed the guard rather than fixing the `point.z` typo, since this PR handles the bad data at the wheel boundary and the design opinion is to let validation reject anything that gets through. One open question in there about `Vec.Cast` forcing `z` to 1 when it's omitted.

PR #9589, opened 2026-07-14
fix(tldraw): don't render submenus if required actions have been removed

### LINKED ISSUE #9133: Deleting export actions results in empty menu option
### What happened?

Usually, when you remove a default tldraw action (as shown here: https://tldraw.dev/examples/action-overrides), it will remove the UI corresponding to that action as well.

However, if you remove all of the export actions, you are left with an empty menu option:

Image

Your only option to fix this is either to reimplement the entire edit menu and omit the export option, or write some hacky `display:none` css that tries to target the export menu option

Remove all export actions:

```
  actions(editor, actions) {
    delete actions['export-as-svg'];
    delete actions['export-all-as-svg'];
    delete actions['export-as-png'];
    delete actions['export-all-as-png'];
```

Then, open the edit menu and click export

Chrome

### Where did this happen?

kaneel:
Closes https://github.com/tldraw/tldraw/issues/9133

The issue shows that if an user of the SDK hides some actions that are rendered by our context menu items, we still display the context menu item. The problem has been fixed by adding a `requiredActions` property to the `TldrawUiMenuSubmenu` component that will early return `null` if none of the actions are available.

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

Reproduction in the issue.

- [x] Unit tests
- [ ] End to end tests

### API changes

- Add a `requiredActions` prop to `TldrawUiMenuSubmenu` (`TLUiMenuSubmenuProps`). When set, the submenu renders nothing unless at least one of the listed action ids exists, so removing its actions via `overrides` hides the submenu instead of leaving an empty trigger. Also adds a new `ExportAsMenuGroup` component (mirroring `CopyAsMenuGroup`) for the "Export as" submenu.

### Release notes

- Add a `requiredActions` prop to `TldrawUiMenuSubmenu` so a submenu hides itself when none of the actions it depends on exist.

### DIFF
--- a/packages/tldraw/src/lib/ui/components/menu-items.tsx
+++ b/packages/tldraw/src/lib/ui/components/menu-items.tsx
@@ -1,5 +1,5 @@
 import { TLPageId, useEditor, useValue } from '@tldraw/editor'
-import { supportsDownloadingOriginal } from '../context/actions'
+import { supportsDownloadingOriginal, useActions } from '../context/actions'
 import { useUiEvents } from '../context/events'
 import { useToasts } from '../context/toasts'
 import {
@@ -239,8 +239,12 @@ export function ClipboardMenuGroup() {
 /** @public @react */
 export function CopyAsMenuGroup() {
 	const editor = useEditor()
+	const actions = useActions()
 	const atLeastOneShapeOnPage = useHasShapesOnPage()
 	const isDebugMode = useValue('isDebugMode', () => editor.getInstanceState().isDebugMode, [editor])
+
+	// If a consumer has removed the copy actions via `overrides`, don't render an empty submenu.
+	if (!actions['copy-as-svg'] && !actions['copy-as-png']) return null
 
 	return (
 		<TldrawUiMenuSubmenu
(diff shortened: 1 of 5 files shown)


packages/tldraw/src/lib/ui/components/menu-items.tsx
```diff
@@ -263,6 +264,26 @@ export function CopyAsMenuGroup() {
...
 }
 
+/** @public @react */
+export function ExportAsMenuGroup() {
+	return (
+		<TldrawUiMenuSubmenu
+			id="export-as"
+			label="context-menu.export-as"
+			size="small"
+			requiredActions={['export-as-svg', 'export-as-png']}
```
mimecuvalo: `requiredActions` is a bit of code smell to me, as it will require keeping two different lists in sync. this is already showing above as `copy-as-json` is missing in the requiredActions array. also "required" is a bit confusing.

let's drop the `requiredActions` prop and do something like:
```
const actions = useActions()
if (!actions['export-as-svg'] && !actions['export-as-png']) return null
```

that still has special casing each action here but it at least matches how `TldrawUiMenuActionItem` works

kaneel: In retrospective, this is a bit smelly and this one's on me :D

mimecuvalo (approved):
👍

PR #9591, opened 2026-07-14
feat(examples): add stamp tool example 

angrycaptain19:
Shows how you can create a stamp tool which lets you click to add a fixed size png at your cursor.


mimecuvalo (commented):
nice, might be good to also make clicking and holding do multiple stamps

angrycaptain19:
> nice, might be good to also make clicking and holding do multiple stamps

Great call - will add in

mimecuvalo (approved):
🤘

PR #9593, opened 2026-07-15
refactor(commenting): namespace cmt- classes under tlui-cmt-

jsscclr:
## What this does

Namespaces the commenting feature's CSS classes under tldraw's `tlui-` UI prefix: every `cmt-*` class becomes `tlui-cmt-*`. This brings commenting in line with the rest of the SDK's UI classes (`tlui-button`, `tlui-menu`, …) so the styles share one predictable namespace.

Purely mechanical — a literal `cmt-` → `tlui-cmt-` substitution. Because the selector definitions (`.css`) and the `className` usages (`.tsx`) both hold the same token, both sides move together and stay in sync by construction.

## Scope

- **254 occurrences across 31 files** — `packages/commenting`, its `component-studio` sketchbooks, the commenting examples, and the dotcom notifications panel that consumes the classes.
- The dotcom consumer references the commenting package's global classes by literal string, so it stays matched to the renamed selectors.
- Covers the one dynamic case (`tlui-cmt-cluster-fade--${phase}` template literal) and the doc-comment references.

## Verification

- `packages/commenting`: 174/174 tests pass
- `lint-current`: clean
- No bare `cmt-` remaining; no double-prefixing

## Follow-up

Any still-open commenting branches will need the same `cmt-` → `tlui-cmt-` substitution re-applied when they rebase onto `commenting`. The rename is mechanical, so those conflicts resolve by re-running the substitution on the affected files.

### DIFF
--- a/apps/component-studio/src/sketchbooks/comments/anchored-comment.css
+++ b/apps/component-studio/src/sketchbooks/comments/anchored-comment.css
@@ -1,11 +1,11 @@
 /* an anchored comment: pin + its thread popover */
-.cmt-anchored {
+.tlui-cmt-anchored {
 	display: flex;
 	align-items: flex-start;
 	gap: 10px;
 }
 
-.cmt-thread-popover {
+.tlui-cmt-thread-popover {
 	width: 280px;
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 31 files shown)


mimecuvalo (approved):
tyty 👍

PR #9595, opened 2026-07-15
perf(tldraw): don't subscribe to color mode in getDisplayValues by default

### LINKED ISSUE #8678: Changing the theme recomputes text size for all shapes
## Description

Toggling the color mode (light/dark) or otherwise changing the theme invalidates the text-size caches for every text-bearing shape in the page, even though the dimension-affecting display values (font family, font size, line height) are typically identical between modes. The work is dominated by re-running text measurement against the DOM, so it's noticeable on canvases with many text, geo, note, and arrow-label shapes.

## Steps to reproduce

1. Open a canvas with a few hundred text-bearing shapes.
2. Toggle the editor's color mode.

Observed: every shape's text-size cache is recomputed, with measurable jank scaling with shape count.

Expected: theme/color-mode changes only invalidate caches whose output actually depends on theme.

## Why this happens

The shared `getDisplayValues` helper reads `editor.getCurrentTheme()` and `editor.getColorMode()`, so any computed cache that calls it (e.g. text size) takes a reactive dependency on theme even when the size isn't actually theme-dependent. Color is the only display value that meaningfully changes between light and dark modes.

## Proposed direction

Split dimension-relevant display values from theme-relevant display values, or have the text-size caches consume only the dimension-affecting subset, so that recoloring doesn't invalidate measurement.

mimecuvalo:
In order to stop light/dark toggles from janking canvases full of text shapes (closes #8678), this PR makes `getDisplayValues` read the color mode without creating a reactive dependency on it when no explicit `colorMode` is passed. Within a theme, light and dark only differ in colors — font size, line height, padding, and stroke width are identical — so the computed caches that call `getDisplayValues` for measurement and geometry (text size, arrow labels, `getGeometry`, arrow info) no longer invalidate and re-measure on every toggle.

This is an alternative to #8906, which kept the capturing default and added a separate `getDimensionDisplayValues` for cache-safe call sites. That approach works but leaves a footgun: every size-sensitive caller (including custom shape authors) must remember to reach for the dimension variant, and forgetting is an invisible perf cliff. This PR flips the failure mode instead: the default is now cache-safe, and callers that render colors opt into color-mode reactivity by passing `colorMode` explicitly — `useColorMode()` in components, `ctx.colorMode` in exports. Forgetting that is loud and benign (colors don't live-update on toggle) rather than silent jank. Almost every default shape component already passed `colorMode` explicitly; this updates the three that didn't (draw, arrow, and highlight's main `component()`), matching the pattern highlight's `backgroundComponent` already used. No new public API is added.

The behavior change for SDK users: custom shape components that call `getDisplayValues(this, shape)` bare will keep rendering the colors from mount time when the user toggles light/dark. The fix is one line — pass `useColorMode()` — and is documented on `getDisplayValues`. Dimension values remain correct in all cases (assuming `getCustomDisplayValues` overrides don't derive dimensions from `colorMode`, which the docs now note).

### Change type

- [x] `improvement`

### Test plan

1. Open a canvas with many text, geo, note, and arrow-label shapes.
2. Toggle light/dark mode; confirm no re-measurement jank and that shape colors still update.

- [x] Unit tests

New tests assert that a computed reading dimension values does not recompute on color-mode toggle, that the text-size cache is not invalidated (spying `textMeasure.measureHtml`), that passing `colorMode` from a reactive read does create the dependency, that theme-definition changes (font size) still invalidate, and a rendered-component test that toggles color scheme and asserts the draw shape repaints.

### API changes

- Changed `getDisplayValues` to read the current color mode without reactive capture when `colorMode` is omitted. Callers that render colors and need live updates on light/dark toggle must pass `colorMode` explicitly (`useColorMode()` in components, `ctx.colorMode` in exports). Signature is unchanged.

### Release notes

- Improve light/dark mode toggle performance: size and geometry caches are no longer invalidated by color-mode changes. If a custom shape component calls `getDisplayValues` without a `colorMode`, pass `useColorMode()` to keep its colors updating live on toggle.

### Code changes

| Section | Net LOC |
| --- | --- |
| Core code | +18 |
| Tests | +162 |

### DIFF
--- a/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/arrow/ArrowShapeUtil.tsx
@@ -42,6 +42,7 @@ import {
 	toDomPrecision,
 	toRichText,
 	track,
+	useColorMode,
 	useEditor,
 	useIsEditing,
 	useSharedSafeId,
@@ -820,7 +821,8 @@ export class ArrowShapeUtil extends ShapeUtil<TLArrowShape> {
 
 	component(shape: TLArrowShape) {
 		const { editor } = this
-		const dv = getDisplayValues(this, shape)
+		const colorMode = useColorMode()
+		const dv = getDisplayValues(this, shape, colorMode)
 
 		const isSelected = useValue(
 			'is selected',
--- a/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/draw/DrawShapeUtil.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/rules-of-hooks */
 import {
 	Box,
 	Circle2d,
@@ -18,6 +19,7 @@ import {
 	last,
 	lerp,
 	rng,
+	useColorMode,
 	useEditor,
 	useValue,
 } from '@tldraw/editor'
@@ -161,7 +163,8 @@ export class DrawShapeUtil extends ShapeUtil<TLDrawShape> {
 	}
 
 	component(shape: TLDrawShape) {
-		const dv = getDisplayValues(this, shape)
+		const colorMode = useColorMode()
+		const dv = getDisplayValues(this, shape, colorMode)
 		return (
 			<SVGContainer>
 				<DrawShapeSvg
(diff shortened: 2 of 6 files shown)


mimecuvalo:
Folding this approach into #8906 instead — closing in favor of iterating there.

PR #9597, opened 2026-07-15
chore(examples): turn on debug mode by default on the develop page

m31-galaxy:
In order to save maintainers from manually enabling debug mode every time they open `/develop` in a fresh tab, this PR turns it on by default there.

The default is applied in `onMount`, guarded by a per-tab sessionStorage flag: a fresh tab comes up with debug mode on, but because `isDebugMode` is per-tab persisted session state that the SDK treats as a sticky user preference, turning it off in the debug menu still sticks across reloads. Always forcing it on in `onMount` would have re-enabled it on every reload.

One pre-existing nuance worth knowing: `/develop` shares `persistenceKey="example"` with ~17 example pages, so debug mode enabled here also shows up on those pages in the same tab. That already happened whenever debug mode was toggled on manually; this change just makes it happen by default.

### Change type

- [x] `improvement`

### Test plan

1. Open `/develop` in a fresh browser tab — the debug panel and debug menu are visible.
2. Turn debug mode off in the debug menu and reload — it stays off.
3. Open `/develop` in a new tab — debug mode is on again.

### Code changes

| Documentation | +15 / -0   |


mimecuvalo (approved):

mimecuvalo:
this is still good to go @m31-galaxy !

PR #9598, opened 2026-07-15
fix(tldraw): validate url before treating as external content

### LINKED ISSUE #8097: Dragging SVG onto canvas in Chrome crashes app when ad blocker is active
## Description

Dragging and dropping an SVG onto the tldraw canvas in Chrome throws a `ValidationError` that crashes the app:

```
ValidationError: At asset(type = bookmark).props.src: Expected a valid url, got "about:blank#blocked" (invalid protocol)
```

Safari handles the same drag and drop correctly.

## Steps to reproduce

1. Drag an SVG from a browser tab (not file system) onto the tldraw canvas in Chrome
2. App crashes with "Something went wrong" error page

Likely requires a Chrome extension (ad blocker) that rewrites blocked URLs to `about:blank#blocked`.

## Environment

- **Broken**: Chrome 145.0.7632.109 (arm64, macOS)
- **Works**: Safari (same drag and drop method)

## Root cause

The drop handler in `useCanvasEvents.ts` (line 100-136) checks:
1. `dataTransfer.files` — if present, handles as file upload (line 116)
2. `dataTransfer.getData('url')` — if present, handles as URL → creates bookmark (line 127)

When dragging an SVG from a browser tab (not from the file system), Chrome provides no files in `dataTransfer.files`, only a URL. If an ad blocker is active, that URL becomes `about:blank#blocked`. Line 128 (`if (url)`) is truthy, so it calls `putExternalContent({ type: 'url', url: 'about:blank#blocked' })`, which creates a bookmark asset with `src: "about:blank#blocked"`. The validator rejects `about:` as an invalid protocol, throwing an unrecoverable error.

The network tab confirms this: our unfurl endpoint gets called with `unfurl?url=about%3Ablank%23blocked`, which returns 400.

Safari either doesn't include the URL data in the DataTransfer for this case, or doesn't rewrite the URL.

## Proposed fix

Validate the URL in the drop handler before treating it as external content:

```ts
// useCanvasEvents.ts line 127-128
const url = e.dataTransfer.getData('url')
if (url && isValidHttpURL(url)) {
```

This would skip invalid URLs, preventing the bookmark creation path for `about:blank#blocked`.

May also want a similar guard in the paste handler's URL path (`useClipboardEvents.ts` line 566) for the same class of issue.

kaneel:
Closes https://github.com/tldraw/tldraw/issues/8097 

This one case seems to occur when trying to drag files from one tab to the url which end up beind treated as an url which is blocked, somehow, by an adblocker. 

I could not reproduce entirely but seems plausible and @MitjaBezensek filed it after the error occured for an user and figured that it was because of an adblocker, we should anyway gracefully fail when an url is not valid, seems the right thing to do

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. Drag an SVG from a browser tab (not file system) onto the tldraw canvas in Chrome
(url and adblocker TBD)
2. App does not crashes with "Something went wrong" error page

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fixed a bug occuring when dragging blocked content from tabs to the canvas.

### DIFF
--- a/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
+++ b/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
@@ -1,6 +1,7 @@
 import {
 	AssetRecordType,
 	Editor,
+	T,
 	TLAsset,
 	TLAssetId,
 	TLBookmarkAsset,
@@ -577,6 +578,15 @@ export async function defaultHandleExternalUrlContent(
 	{ point, url }: { point?: VecLike; url: string },
 	{ toasts, msg }: TLDefaultExternalContentHandlerOpts
 ) {
+	// Ignore urls with a protocol we can't turn into a bookmark (bookmark shapes
+	// validate their `url` prop with T.linkUrl). Dropping content from a browser
+	// tab in Chrome with an ad blocker active supplies a url rewritten to
+	// `about:blank#blocked`; these are never user-intended content, so we ignore
+	// them silently rather than crashing on the validation error (#8097).
+	if (!T.linkUrl.isValid(url)) {
+		return
+	}
+
 	// try to paste as an embed first
 	const embedUtil = editor.getShapeUtil('embed') as EmbedShapeUtil | undefined
 	const embedInfo = embedUtil?.getEmbedDefinition(url)
--- a/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
+++ b/packages/tldraw/src/lib/defaultExternalContentHandlers.ts
@@ -578,12 +578,14 @@ export async function defaultHandleExternalUrlContent(
 	{ point, url }: { point?: VecLike; url: string },
 	{ toasts, msg }: TLDefaultExternalContentHandlerOpts
 ) {
-	// Ignore urls with a protocol we can't turn into a bookmark (bookmark shapes
-	// validate their `url` prop with T.linkUrl). Dropping content from a browser
-	// tab in Chrome with an ad blocker active supplies a url rewritten to
-	// `about:blank#blocked`; these are never user-intended content, so we ignore
-	// them silently rather than crashing on the validation error (#8097).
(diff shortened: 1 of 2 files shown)


packages/tldraw/src/lib/defaultExternalContentHandlers.ts
```diff
@@ -577,6 +578,17 @@ export async function defaultHandleExternalUrlContent(
...
 	{ toasts, msg }: TLDefaultExternalContentHandlerOpts
 ) {
+	// If the url isn't one we can turn into a bookmark (bookmark shapes validate
+	// their `url` prop with T.linkUrl), fall back to dropping the raw string as a
+	// text shape rather than crashing on the validation error (#8097). Chrome with
+	// an ad blocker active rewrites dragged content urls to `about:blank#blocked`,
+	// which isn't a valid bookmark protocol; the text fallback keeps the drop from
+	// throwing while leaving whatever was dropped visible.
+	if (!T.linkUrl.isValid(url)) {
+		await editor.putExternalContent({ type: 'text', text: url, point })
```
MitjaBezensek: I think I'd rather show a toast than put this on the canvas as text. See no value in having `about:blank#blocked` as text.

kaneel: good point I was probably thinking it's better than not showing anything and we get a hunch about why it's not there. I'll investigate a mixed solution where we can trigger a console.warn maybe?

MitjaBezensek: Yeah, we should consider the ux from the user's perspective. Like do they know that something went wrong? Can they do something about it?

MitjaBezensek (approved):

PR #9611, opened 2026-07-16
fix(utils): don't treat any as optional in MakeUndefinedOptional

### LINKED ISSUE #7260: MakeUndefinedOptional also types any as optional
### What happened?

The type `MakeUndefinedOptional` (added in #5055) currently also marks properties as optional that have been typed as `any`.

Example:
```ts
T.object({foo: T.any})
```

Both the type name and PR description suggest that this behavior (while _technically_ correct) is unintentional.

For context, I ran into this while creating a mapping type from Zod schemas to Validator schemas.

_No response_

_No response_

### Where did this happen?

The developer package

### Contact Details

_No response_

### Code of Conduct

kaneel:
Closes #7260

`MakeUndefinedOptional` marked `any`-typed properties as optional because `undefined extends any` is true. The mapped type now splits keys into a required half and an optional half, using an inline `0 extends 1 & T[K]` any-check, so only genuinely undefined-able properties become optional while keys that were already declared optional stay optional.

Switching to `as` key remapping also means the mapped type preserves existing `readonly` / `?` modifiers instead of stripping them.

`TLRichText` is now declared explicitly as an interface and passed to `T.object<TLRichText>()`, since an optional `any` (`attrs`) can't be inferred from the validator's value types once `any` is treated as required.

#### Behaviour

```ts
// A required `any` property stays required (previously became optional)
MakeUndefinedOptional<{ foo: any }>
// before: { foo?: any }
// after:  { foo: any }

// An already-optional `any` property stays optional
MakeUndefinedOptional<{ foo?: any }>
// before: { foo?: any }
// after:  { foo?: any }   (unchanged, but now for the right reason)

// Undefined-able properties still become optional
MakeUndefinedOptional<{ foo: string | undefined }>
// before: { foo?: string | undefined }
// after:  { foo?: string | undefined }

// Plain required properties stay required
MakeUndefinedOptional<{ foo: string }>
// before: { foo: string }
// after:  { foo: string }

// `readonly` / optional modifiers are now preserved (mapped type is homomorphic)
MakeUndefinedOptional<{ readonly id: string; name: string | undefined }>
// before: { id: string; name?: string | undefined }
// after:  { readonly id: string; name?: string | undefined }
```

#### Known limitation

An optional `any` still can't be inferred purely from a validator, because `.optional()` produces `any | undefined`, which collapses to `any`:

```ts
T.object({ foo: T.any.optional() })
// foo is still inferred as required `any`
```

If you need an optional `any` field, state the type explicitly at the boundary, as `richTextValidator` now does:

```ts
interface MyType {
	foo?: any
}
const v = T.object<MyType>({ foo: T.any.optional() })
```

Fully supporting the inferred case would mean retaining optionality metadata on the validator (a branded optional validator), which is better handled as a separate, wider validator API change.

#### Follow-on cleanup

Declaring `TLRichText` as a named interface made api-extractor emit a duplicate `TLRichText_2` in `packages/tldraw`'s API report, because the package's source imported the type from `@tldraw/editor` while the generated shape prop types resolved it from `@tldraw/tlschema`. `packages/tldraw` now imports `TLRichText` from `@tldraw/tlschema` everywhere and declares tlschema as a dependency (it was already importing from it in a few files), which collapses the report back to a single named reference.

### Change type

- [x] `bugfix`
- [ ] `improvement`
- [ ] `feature`
- [x] `api`
- [ ] `other`

### Test plan

- [x] Unit tests
- [ ] End to end tests

`packages/utils/src/lib/types.test.ts` covers the required `any`, already-optional `any`, undefined-able, and plain-required cases.

### API changes

- `@tldraw/utils`: `MakeUndefinedOptional` no longer treats `any` as optional, and now preserves `readonly` / `?` modifiers.
- `@tldraw/tlschema`: `TLRichText` is now declared as an interface (still `{ type: string; content: unknown[]; attrs?: any }` — no consumer-visible shape change).
- `@tldraw/tldraw`: `TLRichText` is re-exported from `@tldraw/tlschema` rather than `@tldraw/editor`; both resolve to the same declaration, so this is a report-only change.

### Release notes

- Fixed `MakeUndefinedOptional` incorrectly marking `any`-typed properties as optional. Validator properties defined with `T.any` are now correctly typed as required.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarNotificationsPanel.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarNotificationsPanel.tsx
@@ -12,6 +12,7 @@ import { ReactNode, useCallback, useState } from 'react'
 import { Link } from 'react-router-dom'
 import {
 	createDeepLinkString,
+	richTextValidator,
 	TldrawUiButton,
 	TldrawUiDropdownMenuContent,
 	TldrawUiDropdownMenuRoot,
@@ -21,7 +22,6 @@ import {
 	TldrawUiMenuGroup,
 	TldrawUiMenuItem,
 	TldrawUiTooltip,
-	TLRichText,
 	useValue,
 } from 'tldraw'
 import { routes } from '../../../../routeDefs'
@@ -234,7 +234,8 @@ export function TlaSidebarNotificationsPanel({ onClose }: { onClose(): void }) {
 				name: c.authorName || unknownAuthor,
 				color: c.authorColor || undefined,
 			},
-			preview: richTextToPlaintext(c.body as TLRichText),
+			// a malformed body should only cost this row its preview, not break the whole list
+			preview: richTextValidator.isValid(c.body) ? richTextToPlaintext(c.body) : '',
 			// the notified-about event: a reaction entry dates from its newest foreign reaction,
 			// not from the (older) comment it decorates
 			date: new Date(n.timestamp).toISOString(),
--- a/packages/tldraw/api-report.api.md
+++ b/packages/tldraw/api-report.api.md
@@ -139,7 +139,6 @@ import { TLPointerEventInfo } from '@tldraw/editor';
 import { TLPropsMigrations } from '@tldraw/tlschema';
 import { TLResizeInfo } from '@tldraw/editor';
 import { TLRichText } from '@tldraw/tlschema';
-import { TLRichText as TLRichText_2 } from '@tldraw/editor';
 import { TLSchema } from '@tldraw/editor';
 import { TLScribble } from '@tldraw/editor';
 import { TLSelectionHandle } from '@tldraw/editor';
@@ -3266,19 +3265,19 @@ export function removeFrame(editor: Editor, ids: TLShapeId[]): void;
 export function RemoveFrameMenuItem(): JSX.Element | null;
 
 // @public
-export function renderHtmlFromRichText(editor: Editor, richText: TLRichText_2): string;
+export function renderHtmlFromRichText(editor: Editor, richText: TLRichText): string;
 
 // @public
-export function renderHtmlFromRichTextForMeasurement(editor: Editor, richText: TLRichText_2): string;
+export function renderHtmlFromRichTextForMeasurement(editor: Editor, richText: TLRichText): string;
 
 // @public
-export function renderHtmlFromRichTextWithExtensions(richText: TLRichText_2, extensions: Extensions): string;
+export function renderHtmlFromRichTextWithExtensions(richText: TLRichText, extensions: Extensions): string;
 
 // @public
-export function renderPlaintextFromRichText(editor: Editor, richText: TLRichText_2): string;
+export function renderPlaintextFromRichText(editor: Editor, richText: TLRichText): string;
 
 // @public
-export function renderRichTextFromHTML(editor: Editor, html: string): TLRichText_2;
+export function renderRichTextFromHTML(editor: Editor, html: string): TLRichText;
 
 // @public (undocumented)
 export function ReorderMenuItems(): JSX.Element;
@@ -3318,7 +3317,7 @@ export interface RichTextLabelProps {
     // (undocumented)
     padding?: number;
     // (undocumented)
-    richText?: TLRichText_2;
+    richText?: TLRichText;
     // (undocumented)
     shapeId: TLShapeId;
     // (undocumented)
@@ -3333,7 +3332,7 @@ export interface RichTextLabelProps {
     textWidth?: number;
     // (undocumented)
     type: ExtractShapeByProps<{
-        richText: TLRichText_2;
+        richText: TLRichText;
     }>['type'];
     // (undocumented)
     verticalAlign: 'end' | 'middle' | 'start';
@@ -3359,7 +3358,7 @@ export interface RichTextSVGProps {
     // (undocumented)
     padding: number;
     // (undocumented)
-    richText: TLRichText_2;
+    richText: TLRichText;
     // (undocumented)
     showTextOutline?: boolean;
     // (undocumented)
@@ -3760,7 +3759,7 @@ export interface TextAreaProps {
     // (undocumented)
     handleChange(changeInfo: {
         plaintext?: string;
-        richText?: TLRichText_2;
+        richText?: TLRichText;
     }): void;
     // (undocumented)
     handleDoubleClick(e: any): any;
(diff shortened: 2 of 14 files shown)


packages/tldraw/api-report.api.md
```diff
@@ -135,7 +135,8 @@ import { TLParentId } from '@tldraw/tlschema';
 import { TLPointerEventInfo } from '@tldraw/editor';
 import { TLPropsMigrations } from '@tldraw/tlschema';
 import { TLResizeInfo } from '@tldraw/editor';
-import { TLRichText } from '@tldraw/editor';
+import { TLRichText } from '@tldraw/tlschema';
+import { TLRichText as TLRichText_2 } from '@tldraw/editor';
```
steveruizok: ?

kaneel: @steveruizok I'm not sure what you mean here, are you arguing the case of the `_2`? It can be found throughout the same api report with, for example `TLShapeId_2` l. 146

mimecuvalo:
AI review note: I think the `any` exception is directionally right, but the current implementation also turns an already-optional `any` property back into a required property:

```ts
type Result = MakeUndefinedOptional<{ foo?: any }>
// currently becomes { foo: any }
```

The same underlying limitation means `T.any.optional()` and `T.any` are indistinguishable once reduced to their value types, since `any | undefined` collapses to `any`. The explicit `TLRichText` interface preserves the public alias, but `T.TypeOf<typeof richTextValidator>` still changes to require `attrs`.

A low-blast-radius alternative would be to preserve syntactically optional keys before applying the `any` exception:

```ts
export type MakeUndefinedOptional<T extends object> = Expand<
	{
		[K in keyof T as {} extends Pick<T, K>
			? never
			: 0 extends 1 & T[K]
				? K
				: undefined extends T[K]
					? never
					: K]: T[K]
	} & {
		[K in keyof T as {} extends Pick<T, K>
			? K
			: 0 extends 1 & T[K]
				? never
				: undefined extends T[K]
					? K
					: never]?: T[K]
	}
>
```

That produces:

```ts
MakeUndefinedOptional<{ foo: any }>                // { foo: any }
MakeUndefinedOptional<{ foo?: any }>               // { foo?: any }
MakeUndefinedOptional<{ foo: string | undefined }> // { foo?: string | undefined }
```

Then rich text could state its intended type explicitly at the validator boundary:

```ts
export interface TLRichText {
	type: string
	content: unknown[]
	attrs?: any
}

export const richTextValidator = T.object<TLRichText>({
	type: T.string,
	content: T.arrayOf(T.unknown),
	attrs: T.any.optional(),
})
```

This also lets the `IsAny` test stay inline rather than adding a public implementation-helper type, which seems consistent with the approach discussed in #7091 and #7755.

This still cannot infer that `T.object({ foo: T.any.optional() })` has an optional key without retaining optionality metadata on the validator. Fully supporting that case would require a branded/specialized optional validator, which feels better handled as a separate, wider validator API change.

packages/tldraw/api-report.api.md
frolic: usually means things are imported from two different places within the code somewhere
can fix these types of issues if we consistently import from the same place

packages/utils/src/lib/types.ts
```diff
@@ -138,12 +138,30 @@ export type Required<T, K extends keyof T> = Expand<Omit<T, K> & { [P in K]-?: T
...
+ *
  * @public
  */
 export type MakeUndefinedOptional<T extends object> = Expand<
 	{
-		[P in { [K in keyof T]: undefined extends T[K] ? never : K }[keyof T]]: T[P]
+		// Required keys: not already optional, and either `any` or not undefined-able.
+		[K in keyof T as Record<never, never> extends Pick<T, K>
+			? never
+			: 0 extends 1 & T[K]
```
frolic, line 152: what is this doing? I know there's a bunch of TS tricks you can do to coerce types or more accurately check for an intended type, but I haven't seen this particular case before

kaneel: as discussed, this is differentiating any from the rest, it's basically a test for if the type in question is any.

frolic: since it's a bit magical unless you really know your TS, would be good to lift this out into a utility type like `IsAny<T>` or at least leave a comment that this is an "is any" check

packages/tldraw/api-report.api.md
kaneel: Fixed in 3d4a3f71c5 — you were right about the cause. `packages/tldraw` imported `TLRichText` from `@tldraw/editor`, while the generated shape prop types (geo/note/text `richText`) resolved it from `@tldraw/tlschema`, so api-extractor saw two entry points for one type and disambiguated with `_2`.

The package now imports `TLRichText` from `@tldraw/tlschema` everywhere (and declares tlschema as a dependency, which it was already using in a handful of files). The report is down to a single `import { TLRichText } from '@tldraw/tlschema'`, and as a bonus the shape prop types now read `richText: TLRichText` instead of the inline `{ attrs?: any; content: unknown[]; type: string }` expansion.

Note this only covers `TLRichText`. The other pre-existing duplicates on main (`TLShapeId_2`, `IndexKey_2`, `TLFontFace_2`, `VecModel_2`) have the same root cause but I've left them out of this PR.

apps/dotcom/client/src/tla/components/TlaSidebar/components/TlaSidebarNotificationsPanel.tsx
```diff
@@ -100,7 +100,7 @@ export function TlaSidebarNotificationsPanel({ onClose }: { onClose(): void }) {
 			name: c.authorName || unknownAuthor,
 			color: c.authorColor || undefined,
 		},
-		preview: richTextToPlaintext(c.body as TLRichText),
+		preview: richTextToPlaintext(richTextValidator.validate(c.body)),
```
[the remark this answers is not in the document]
kaneel: Good catch, fixed. The panel no longer throws on a malformed body:

```ts
preview: richTextValidator.isValid(c.body) ? richTextToPlaintext(c.body) : '',
```

A bad row now costs only its own preview instead of aborting the whole list render. (This area also moved on main in the meantime, so the change landed as part of merging main in.)

kaneel:
Went with the split required/optional mapped type from the suggestion above — an already-optional `any` stays optional, the `IsAny` check stays inline rather than becoming a public helper, and `richTextValidator` states its type explicitly via `T.object<TLRichText>()`. `packages/utils/src/lib/types.test.ts` covers the four cases. The PR description had gone stale on this (it still claimed a public `IsAny<T>`); it's been corrected.

kaneel:
I'm going to close this one for now

PR #9613, opened 2026-07-16
fix(tldraw): size Vimeo embeds to their real aspect ratio

### LINKED ISSUE #6891: [Bug]: Vimeo embed not responsively sized
### What happened?

Image

with url: https://vimeo.com/817841251 (watch this if u havn't, thank me later)

_No response_

_No response_

### Where did this happen?

The tldraw.com website

### Contact Details

_No response_

### Code of Conduct

kaneel:
Closes https://github.com/tldraw/tldraw/issues/6891

Embeds are created at a fixed 640×360 (16:9) box, so Vimeo videos with a different intrinsic aspect ratio were letterboxed (white "ears") inside the iframe.

When an embed is created or re-pointed at a new URL, we now resolve the content's real aspect ratio from the URL's OpenGraph image dimensions and resize the shape to match, keeping its center fixed. The dimensions come from the same server-side "unfurl" path bookmarks already use (`editor.getAssetForExternalContent({ type: 'url' })`), so there's no extra client-side fetch and it degrades gracefully when no metadata is available.

How it works:

- A new `sizeToContentAspectRatio` flag on embed definitions opts a provider in (enabled for Vimeo). It's generic — any provider whose page reports `og:image:width`/`og:image:height` benefits.
- Resolution is driven by a store side effect (on shape create / URL change), not by rendering, so it never runs during SVG export and doesn't fire per-view.
- `EmbedShapeUtil.resolveAspectRatio` reads `imageWidth`/`imageHeight` from the unfurled asset's `meta`, then resizes via the editor (correct under rotation, no undo-stack entry). Guards handle deleted shapes and stale/overlapping resolutions.

**Why not YouTube?** YouTube advertises the *same* 16:9 thumbnail dimensions for every video — vertical videos are padded into a 16:9 `maxresdefault` thumbnail, and its oEmbed/OG dimensions are identical for a 16:9 video and a 9:16 Short. So the unfurled dimensions can never reveal a non-16:9 ratio, and the correction would always be a no-op. (This is despite YouTube *doing* return OG tags to our unfurl UA — the limitation is the fixed thumbnail ratio, not missing metadata.) It's left out to avoid a pointless unfurl round-trip per YouTube embed.

https://github.com/user-attachments/assets/c04ab77e-1fd4-4e5a-9540-517b5ef7b58b

### Change type

- [ ] `bugfix`
- [x] `improvement`
- [ ] `feature`
- [ ] `api`
- [ ] `other`

### Test plan

1. On tldraw.com (with the unfurl service returning image dimensions), copy https://vimeo.com/817841251 onto a canvas.
2. The embed should take the video's real 3:2 ratio — no white "ears".
3. A 16:9 video (e.g. https://vimeo.com/71278954) should be unchanged.

- [x] Unit tests
- [x] End to end tests

### API changes

- **`EmbedDefinition`** gains an optional `sizeToContentAspectRatio?: boolean`. When set, the embed shape corrects its size after creation to its content's real aspect ratio, resolved from the URL's OpenGraph metadata via the editor's url asset handler. Enabled for the built-in Vimeo definition.
- **`EmbedShapeUtil.resolveAspectRatio(shape: TLEmbedShape): Promise<void>`** — public method that resolves an embed's true ratio and corrects its size. Triggered by a store side effect on create / URL change; exposed so subclasses can override or trigger it.

All additive; existing custom embed definitions are unaffected.

### Release notes

- Embeds for providers that report their dimensions (e.g. Vimeo) now size themselves to their content's real aspect ratio, so videos are no longer letterboxed.

### DIFF
--- a/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
+++ b/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
@@ -280,6 +280,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 		},
 		isAspectRatioLocked: true,
 		embedOnPaste: true,
+		sizeToContentAspectRatio: true,
 		toEmbedUrl: (url) => {
 			const urlObj = safeParseUrl(url)
 			if (!urlObj) return
@@ -539,6 +540,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 		doesResize: true,
 		isAspectRatioLocked: true,
 		embedOnPaste: true,
+		sizeToContentAspectRatio: true,
 		toEmbedUrl: (url) => {
 			const urlObj = safeParseUrl(url)
 			if (urlObj && urlObj.hostname === 'vimeo.com') {
@@ -742,6 +744,14 @@ export interface EmbedDefinition<Config = never> {
 	readonly canEditWhileLocked?: boolean
 	// TODO: FIXME this is ugly be required because some embeds have their own border radius for example spotify embeds
 	readonly overrideOutlineRadius?: number
+	/**
+	 * When true, the embed shape corrects its size after creation to match the real aspect ratio of
+	 * its content, so fixed-aspect media (like video) isn't letterboxed inside the default box. The
+	 * dimensions are resolved from the URL's OpenGraph metadata via the editor's url asset handler
+	 * (the same "unfurl" path used for bookmarks), so it degrades gracefully when no metadata is
+	 * available.
+	 */
+	readonly sizeToContentAspectRatio?: boolean
 	// eslint-disable-next-line tldraw/method-signature-style
 	readonly toEmbedUrl: (url: string, config?: Config) => string | undefined
 	// eslint-disable-next-line tldraw/method-signature-style
--- a/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
+++ b/packages/tldraw/src/lib/defaultEmbedDefinitions.ts
@@ -280,7 +280,9 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 		},
 		isAspectRatioLocked: true,
 		embedOnPaste: true,
-		sizeToContentAspectRatio: true,
+		// Not opted into `sizeToContentAspectRatio`: YouTube advertises the same 16:9 thumbnail
+		// dimensions for every video (it pads vertical videos into a 16:9 thumbnail), so unfurled
+		// dimensions can never reveal a non-16:9 ratio — the correction would always be a no-op.
 		toEmbedUrl: (url) => {
 			const urlObj = safeParseUrl(url)
 			if (!urlObj) return
--- a/packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
+++ b/packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
@@ -24,7 +24,7 @@ import {
 	embedShapePermissionDefaults,
 	unknownEmbedShapePermissionOverrides,
 } from '../../defaultEmbedDefinitions'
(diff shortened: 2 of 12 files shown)


mimecuvalo (changes requested):
we're gonna wanna make this a little more generalizable. a couple things:
- youtube "supports" oembed but badly, and youtube is much more popular than vimeo (sorry), example youtube oembed that doesn't report the height/width correctly: https://www.youtube.com/oembed?format=json&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DQqtW2q9ftu0

- we can get the height/width from the og data anyway (on vimeo + youtube)

Image

i would like us to take a different approach. we own the unfurl service (although it sounds like we don't b/c the package is called `cloudflare-workers-unfurl`). it's found here: https://github.com/tldraw/cloudflare-workers-unfurl  (btw cc @driev that we have this microservice)

we should run this fetch request thru our workers / server-side (instead of client-side).

kaneel:
@mimecuvalo thanks for the headsup regarding the endpoint, will get to this!

kaneel:
let's go @mimecuvalo https://github.com/tldraw/cloudflare-workers-unfurl/pull/2

packages/tldraw/src/lib/shapes/embed/EmbedShapeUtil.tsx
```diff
@@ -103,6 +103,76 @@ export class EmbedShapeUtil extends BaseBoxShapeUtil<TLEmbedShape> {
...
+		// Bail if the shape is gone or its url changed while we were awaiting: a later run for the
+		// new url will handle it, and we must not apply this url's ratio to a different video.
+		if (!latest || latest.props.url !== url) return
+
+		const corrected = getCorrectedEmbedSize({
+			w: latest.props.w,
+			h: latest.props.h,
+			resolvedRatio,
+		})
+		if (!corrected) return
```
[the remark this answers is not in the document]
kaneel: we don't have a way to change the url, the embed exists or does not exist

packages/tldraw/src/lib/defaultEmbedDefinitions.ts
```diff
@@ -280,6 +280,7 @@ export const DEFAULT_EMBED_DEFINITIONS = [
 		},
 		isAspectRatioLocked: true,
 		embedOnPaste: true,
+		sizeToContentAspectRatio: true,
```
mimecuvalo: the PR desc says this doesn't work for YouTube so let's disable it and leave it just for vimeo

mimecuvalo (approved):
pre-approving modulo one comment

packages/tldraw/src/lib/defaultEmbedDefinitions.ts
kaneel: Done — dropped it, Vimeo only now.

Small correction on the reasoning though: that caveat is actually from the unfurl worker PR description, not this one (this PR's description said it was enabled for both). And it turns out "doesn't work for YouTube" isn't quite the reason either — YouTube *does* return OG tags (incl. `og:image:width/height`) to our `tldraw-bot` UA.

The real reason to disable it is that YouTube advertises the **same 16:9 thumbnail dimensions for every video** — it pads vertical videos into a 16:9 `maxresdefault` thumbnail. I checked via oEmbed and a 16:9 video and two vertical Shorts all report identical dimensions. So the unfurled ratio is always ~16:9 and the correction is a guaranteed no-op — no point paying for the unfurl round-trip. Updated the PR description with this.

kaneel: sorry @mimecuvalo my LLM got wild and wrote this comment on my behalf (real story)

PR #9623, opened 2026-07-17
feat(commenting): support multi-shape comment anchors

mimecuvalo:
In order to let a comment refer to a group of shapes — "these three boxes need to line up" — rather than only a single shape, this PR generalizes the `shape` comment anchor from `shapeId: TLShapeId` to `shapeIds: TLShapeId[]`. It's a spike to pressure-test folding multi-shape anchoring into the existing `shape` anchor kind instead of adding a sibling `shapes` variant: a single shape is just the length-1 case, so every consumer keeps one code path and the public union doesn't carry two near-identical members. The deep-link union made the same call — it only ever shipped the plural form.

The pin resolves against the common page bounds of the anchored shapes that still exist (via the existing `editor.getShapesPageBounds`), so it tracks the group as any of them move; deleting some shapes keeps the anchor tracking the rest, and it hides only when none remain. `shapeIds[0]` is the primary shape — it's what the sync worker denormalizes into the single-value `comment_thread.shapeId` column and what notification deep links use. With the comment tool, clicking a shape that's part of a multi-shape selection anchors the thread to the whole selection, clicked shape first.

Data and compat notes for review:

- No data migrations: commenting has never shipped, so the anchor shape change lands directly in the validator with no migration chain. The only migration on the comment record types is the pre-existing `/1` identity guard, which rejects pre-commenting clients with CLIENT_TOO_OLD instead of sending them record types their store can't represent.
- Precise pins on multi-shape anchors are normalized within the common bounds, which reflow as shapes move relative to each other — a precise pin tracks the group approximately rather than any one shape exactly. If we'd rather make multi-shape anchors imprecise-only, that would be an argument for a separate leaner variant instead; this spike deliberately takes the folded trade.
- Drag-to-re-anchor deliberately collapses to the single dropped-on shape.

### Change type

- [x] `feature`

### Test plan

1. Open the comment-anchors example; the "both ellipses" thread's pin sits at the top-right of the pair's common bounds and follows either ellipse when moved.
2. Delete one ellipse — the pin keeps tracking the other.
3. Select several shapes, pick the comment tool, click one of the selected shapes — the posted thread anchors to the whole selection.

- [x] Unit tests

### Release notes

- Comment threads can now anchor to multiple shapes: the `shape` anchor's `shapeId` became `shapeIds: TLShapeId[]`, with the pin tracking the common bounds of the surviving shapes.

### API changes

- Breaking (pre-release): the `shape` variant of `TLCommentAnchor` now carries `shapeIds: TLShapeId[]` (non-empty) instead of `shapeId: TLShapeId`. No migration — the feature has never shipped.
- Changed `shapeAnchorAt(editor, shapeIds, page, precise)` to take an array of shape ids.

### Code changes

| Area | Net LOC |
| --- | --- |
| Core code | -1 |
| Tests | +10 |
| Automated files | +0 |
| Documentation | +1 |
| Apps | +15 |

### DIFF
--- a/apps/component-studio/src/comment-model.tsx
+++ b/apps/component-studio/src/comment-model.tsx
@@ -60,7 +60,7 @@ const PAGE_ID = 'page:studio' as TLPageId
 /** A sample thread, anchored to a shape. */
 export const sampleThread: TLCommentThread = createCommentThread({
 	pageId: PAGE_ID,
-	anchor: { type: 'shape', shapeId: createShapeId('box'), x: 1, y: 0, isPrecise: false },
+	anchor: { type: 'shape', shapeIds: [createShapeId('box')], x: 1, y: 0, isPrecise: false },
 	createdBy: 'ada',
 	now: NOW - 2 * HOUR,
 })
--- a/apps/component-studio/src/sketchbooks/anchoring/comment-anchor.sketchbook.tsx
+++ b/apps/component-studio/src/sketchbooks/anchoring/comment-anchor.sketchbook.tsx
@@ -11,7 +11,7 @@ const sketchbook: Sketchbook<CommentAnchorProps> = {
 			control: 'union',
 			discriminant: 'type',
 			variants: {
-				shape: { type: 'shape', shapeId: createShapeId('box'), x: 1, y: 0, isPrecise: false },
+				shape: { type: 'shape', shapeIds: [createShapeId('box')], x: 1, y: 0, isPrecise: false },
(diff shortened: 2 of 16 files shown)


jsscclr (approved):
API makes sense to me!

mimecuvalo:
i think i hate this behavior in general, i'll bring it up during standup but i'm gonna kill this PR, i just think the multi-shape comment doesn't work very well

mimecuvalo:
closing, doesn't feel good

PR #9624, opened 2026-07-17
feat(dotcom): enforce note text attribution on shape create

mimecuvalo:
Stacked on `commenting` (which now has `main` merged in, bringing #9566). Closes the note-attribution create loophole that #9489 had to leave open.

### Background

#9489 added `authorizeFileRecord`, which stamps/vetoes note text attribution (`props.textLastEditedBy`) from the session's authenticated identity. Its `authorizeShape` **passed all creates through untouched**: before duplication re-stamped attribution, a duplicate legitimately carried the original author's identity, so a forged create was indistinguishable from a duplicate.

#9566 added `ShapeUtil.onBeforeDuplicate`, and `NoteShapeUtil` uses it to re-stamp `textLastEditedBy` to the current user whenever a note with text is duplicated or pasted. A legitimate create can therefore only ever carry `null` or the session user's own id.

### Change

`authorizeShape` now enforces attribution on **create** the same way it already did on **update**: the attribution may only be `null` or the session's own user; anything else is vetoed. On update an *unchanged* attribution is still allowed from any user (so anyone with access can edit a note's other props), and deletes stay open for cascade cleanup.

This closes the forged-create loophole (create a note directly with someone else's `textLastEditedBy`) without affecting real duplication/paste, which now stamps the copy to the current user client-side.

### Tests

`apps/dotcom/sync-worker/src/authorizeFileRecord.test.ts`: the create passthrough cases are replaced with enforcement — a create with foreign attribution is vetoed, a create attributed to the session user or with no attribution passes, an anonymous session can't create a note with any attribution, and non-note creates pass through.

### Change type

- [x] `feature`

### Release notes

- Internal: the sync-worker record authorizer now enforces note text attribution on shape create (not just update), so a note can't be created in another user's name. Not user-facing.

### DIFF
--- a/apps/dotcom/sync-worker/src/authorizeFileRecord.test.ts
+++ b/apps/dotcom/sync-worker/src/authorizeFileRecord.test.ts
@@ -175,16 +175,31 @@ describe('authorizeFileRecord', () => {
 			return { ...note, id: createShapeId('geo1'), type: 'geo', props } as unknown as TLShape
 		}
 
-		it('passes creates through untouched, even with foreign attribution (duplication)', () => {
+		it('vetoes a create with foreign attribution (duplication now re-stamps the copy)', () => {
 			const next = makeNote('real-alice')
+			expect(
+				authorize({ session: session('real-bob'), type: 'create', prev: null, next })
+			).toBeNull()
+		})
+
+		it('allows a create attributed to the session user', () => {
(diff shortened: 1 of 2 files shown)


mimecuvalo:
@MitjaBezensek for your review async next week, submitting now though

PR #9630, opened 2026-07-20
feat(commenting): color / image comments by author

MitjaBezensek:
In order to show who a comment belongs to at a glance, this PR colors comment pins and avatars with the author's color, and lets hosts supply an avatar image instead — on tldraw.com, comments use the author's Clerk avatar.

The name resolver prop on `CanvasComments` and `CanvasCommentsSidebar` becomes an author resolver: `resolveName(id): string | undefined` → `resolveAuthor(id): CommentAuthor | undefined`, where `CommentAuthor` is `{ name, color?, image?, meta? }`. `color` accepts any CSS color and applies to `Avatar`, `CommentPin` (unresolved pins only — resolved stays grey), the composers, thread cards, and sidebar list items, falling back to the default tint when omitted. `image` shows in the avatar circles instead of the colored initial; the canvas pin keeps the initial-on-color. `meta` is user-defined metadata the toolkit never reads (like a shape's `meta`). The mention and rich-text paths keep a name-only resolver derived at the boundary.

`CommentAuthor` is the one author shape across the toolkit: `Avatar` and `Byline` take the author object, `CommentCard`/`CommentComposer`/`CommentsList` items carry it (so `meta` survives into `renderItem`/`renderComment` overrides), and `MentionMember` extends it (its `avatar` field is renamed to `image`).

dotcom resolves author info from the same sources as names: the signed-in user's preferences, the Zero comments query's author join, live presence, and the workspace roster. Avatar images come from the user's Clerk avatar (self via the Zero user row, other authors via the author join, notification rows too); presence and the roster carry no image and fall back to colored initials.

CleanShot 2026-07-21 at 09 52 51

### Change type

- [x] `improvement`

### Test plan

1. Open a shared file with two users with different cursor colors
2. Place comments as each user — pins, avatars, and sidebar rows use each author's color
3. Resolve a thread — the pin turns grey regardless of author color
4. In the commenting example, Ada has an avatar image — cards, composer, sidebar rows, and the mention picker show it; her canvas pin stays initial-on-color
5. @-mention a member — mentions still resolve to current names
6. On tldraw.com, comment cards, sidebar rows, and notifications show authors' Clerk avatars; pins stay initial-on-color

- [ ] Unit tests
- [ ] End to end tests

### API changes

- Breaking! `CanvasCommentsProps.resolveName` / `CanvasCommentsSidebarProps.resolveName` replaced with `resolveAuthor(id): CommentAuthor | undefined`
- Breaking! `CommentCardProps.author`, `CommentComposerProps.author`, `CommentListItemProps.author`, `BylineProps.author`, and `AvatarProps` are now `CommentAuthor`-based instead of separate strings
- Breaking! `MentionMember.avatar` renamed to `image`; `MentionMember` now extends `CommentAuthor`
- Added `CommentAuthor` (`{ name, color?, image?, meta? }`)
- Added `CommentPinProps.color`

### Code changes

| Core code       | +89 / -53   |
| Automated files | +21 / -16   |
| Documentation   | +31 / -15   |
| Apps            | +111 / -87  |

### Release notes

- Comment pins and avatars now use the author's color, and avatars can show an image (tldraw.com shows Clerk avatars). `CanvasComments`/`CanvasCommentsSidebar` take `resolveAuthor(id): CommentAuthor | undefined` (`{ name, color?, image?, meta? }`) in place of `resolveName`, and the presentational comment components take `CommentAuthor` objects.


packages/commenting/src/ui/comment-author.ts
```diff
@@ -0,0 +1,12 @@
+import { JsonObject } from 'tldraw'
+
+/** A comment author's display identity, resolved from an author id.
+ * @public */
+export interface CommentAuthor {
+	name: string
+	/** The author's color (any CSS color) — colors their avatar and pins.
+	 *  Omit for the default tint. */
+	color?: string
```
MitjaBezensek, line 9: Unsure if we should expose color here. Feels like a common enough use case to add it? We could also pass it via the `meta` property for dotcom and then omit it here.

I would definitely add meta though. If somebody wants to reuse our `SidebarPanel`, but implement their own `CommentsList` that shows more data they could pass that data via the meta prop.

mimecuvalo (commented):
my high-level feedback is that if we add color is that we should also add an option to be able to specify an image instead of a color.

MitjaBezensek:
> my high-level feedback is that if we add color is that we should also add an option to be able to specify an image instead of a color.

Added 👍 Also added an example of how to set it up.

apps/dotcom/client/src/tla/components/TlaEditor/CommentsOnCanvas.tsx
```diff
@@ -48,12 +58,18 @@ export function CommentsOnCanvas({ fileId }: { fileId: string }) {
...
-		const names = new Map<string, string>()
+	const commentAuthors = useMemo(() => {
+		const authors = new Map<string, CommentAuthor>()
 		for (const c of fileComments) {
-			if (c.author?.name) names.set(c.authorId, c.author.name)
+			if (c.author?.name) {
+				authors.set(c.authorId, {
+					name: c.author.name,
+					color: c.author.color,
+					image: c.author.avatar || undefined,
```
MitjaBezensek: Unsure if we should use the avatar, we don't use it for presence?

I'd probably lean against using it for now. We don't have a way for users to customize their avatars and it could cause issues for some users if we show their google avatar without the ability for them to change it. I'd add it when we add user settings.

jsscclr:
Ah, I made life a bit difficult for you with the image options with my UI fixes...! if there is an avatar, I would expect the avatar to be full size (the previous size of 24px). I made it so the colour only option is 14px.

apps/dotcom/client/src/tla/components/TlaEditor/CommentsOnCanvas.tsx
jsscclr: yeah for dotcom makes sense to me not to show avatar

packages/commenting/src/ui/comment-author.ts
jsscclr: I think it's not a crazy concept - especially since Miro for example provides a colour picker for comments as well? Hmmm... Maybe one of those things to ask customers, seeing if any could see themselves using it

MitjaBezensek:
> Ah, I made life a bit difficult for you with the image options with my UI fixes...! if there is an avatar, I would expect the avatar to be full size (the previous size of 24px). I made it so the colour only option is 14px.

No worries, will add some additional css for the avatar case.

packages/commenting/src/ui/comment-author.ts
jsscclr: @angrycaptain19

jsscclr (approved):
LGTM... sorry about the merge conflicts 😅

PR #9636, opened 2026-07-21
feat(dotcom): show empty pages named --- as dividers in the page menu

### LINKED ISSUE #9445: Convert pages named --- into dividers
Image

On tldraw.com, pages whose name is `---` should be displayed as dividers in the pages list rather than as regular, selectable pages. This is similar to how Figma and other tools treat pages named `---` as visual separators for organizing the page list.

m31-galaxy:
In order to let people organize long page lists on tldraw.com the way they do in Figma — separator rows between groups of pages — this PR renders empty pages named `---` (three or more hyphens) as divider lines in the pages dropdown. Closes #9445.

Renaming a brand-new (empty) page to `---` converts it, and that is the only path in: a page that already has content just gets the literal name `---` and stays a normal page. Dividers are half the height of a page row, show no tooltip or hover highlight, and can't be selected, renamed, or drawn on — they only support drag-to-reorder plus move up/down and delete from their ⋮ menu. They're also excluded from the context menu's "Move to page" list.

Decisions worth knowing when reviewing:

- **Divider status is derived** (name + zero shapes), not stored — no schema change, collaboration-safe, and content can never end up hidden behind an unselectable row. Undoing the rename is the only way back, by design.
- **The current page is never treated as a divider**, so the editor's automatic page fallback can't strand anyone on an unselectable page. Completing a rename that creates a divider steps back to the nearest regular page (creating a page makes it current before its inline rename starts), and deleting the current page from the menu pre-selects a non-divider neighbor.
- **The SDK is untouched.** This is a dotcom-side fork of `DefaultPageMenu` (plus its submenu and a copy of the default context menu content), wired into all dotcom editor surfaces. Regular-row behavior is preserved exactly; the fork files carry provenance headers pinned to the SDK commit so future drift-diffing stays mechanical.
- Because divider rows are half-height, row positions come from prefix-summed offsets instead of index × row height. The drag-reorder math generalizes accordingly and reduces to the SDK's exact behavior when all rows are full height.

### Change type

- [x] `feature`

### Test plan

1. In a file on tldraw.com, open the pages dropdown, create a page, and rename it to `---` — it becomes a divider line and you're returned to your previous page.
2. Click the divider (nothing happens), drag it to reorder, hover it (no highlight, no tooltip), and check its ⋮ menu only offers move and delete.
3. Draw on a page, then rename that page to `---` — it stays a normal, selectable page.
4. Right-click a shape → Move to page — dividers aren't listed.

- [x] Unit tests
- [x] End to end tests

### Release notes

- Organize the pages list on tldraw.com by renaming an empty page to `---` to turn it into a divider.

### Code changes

| Tests           | +205 / -0   |
| Apps            | +1254 / -3  |

### DIFF
--- a/apps/dotcom/client/e2e/fixtures/Editor.ts
+++ b/apps/dotcom/client/e2e/fixtures/Editor.ts
@@ -9,6 +9,8 @@ export class Editor {
 	private readonly fileName: Locator
 	private readonly shapes: Locator
 	private readonly pageMenu: Locator
+	public readonly pageMenuItems: Locator
+	public readonly pageMenuTriggerLabel: Locator
 
 	constructor(
 		public readonly page: Page,
@@ -18,6 +20,8 @@ export class Editor {
 		this.fileName = this.page.getByTestId('tla-file-name')
 		this.shapes = this.page.locator('.tl-shape')
 		this.pageMenu = this.page.getByTestId('tla-main-menu')
+		this.pageMenuItems = this.page.getByTestId('page-menu.item')
+		this.pageMenuTriggerLabel = this.page.getByTestId('page-menu.button')
 	}
 
 	async toggleSidebar() {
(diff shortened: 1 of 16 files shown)


mimecuvalo:
this isn't on you @m31-galaxy but i'm not confident on the original issue's request being the right direction here. couple things:
- it might pseudo-solve "grouping" pages together, but it doesn't do it really semantically. 
- also, this is a super-hidden power feature that seems like it'll bit-rot / not get supported long-term
- finally, it's hard to solve the resulting awkward UI (double hair-lines and overlapping kebab-menus):

Screenshot 2026-07-22 at 20 01 58

marking as #needs-design-review

m31-galaxy:
Makes sense! It felt very hacky and rough around the edges to me as well, just pushed out the PR so we have a starting point to work off of.

steveruizok:
> this isn't on you @m31-galaxy but i'm not confident on the original issue's request being the right direction here. couple things:
> 
> * it might pseudo-solve "grouping" pages together, but it doesn't do it really semantically.
> * also, this is a super-hidden power feature that seems like it'll bit-rot / not get supported long-term
> * finally, it's hard to solve the resulting awkward UI (double hair-lines and overlapping kebab-menus):
> 
> Screenshot 2026-07-22 at 20 01 58
> marking as #needs-design-review

We can fix that overlap (there should be no menu item) but this is a feature I'd like, so I may push forward with it.

mimecuvalo:
closing up some PRs that are eating up some CF usage 😬 , reopen if needed later (and rebase with `main`)

PR #9639, opened 2026-07-22
fix(commenting): make coincident comment pins reachable via a count-badge thread list

jsscclr:
Multiple imprecise comments on one shape all resolve to the same anchor point (the imprecise-anchor spot, top-right by default), so their pins render pixel-identical at every zoom. Only the top pin is reachable; the clustering badge's zoom-to-split lands on the same coincident overlap it started from, so the buried threads can never be opened.

This makes coincident pins a single count badge (the cluster badge's face) that opens the threads as a list of cards. Clicking a card expands that thread inline — reply composer, edit, resolve, delete — while the others stay collapsed; expanding one collapses the other via the existing single-open-thread state. Escape steps back: expanded thread → card list → closed.

How it's put together:

- `computePinStacks` groups threads by their resolved page-space anchor point (camera-independent, so it never recomputes on camera moves) with a small quantum to absorb float noise. Groups of 2+ render as one `ThreadStackPin`; everything else pins as before.
- The layer picks one *owner* member per group to render the stack, decided across all render paths — members can arrive via different paths (a leaf via clustering while its open sibling renders via the open slot), so ownership can't be decided per-path.
- The thread popover's interactive content moved out of `ThreadPin` into `ThreadView`, shared by the single-pin popover and the stack's expanded card. `ThreadPin` keeps the positioning/drag/region concerns.
- The list's open state is a new editor atom (`openStackId`) rather than component state: the stack pin remounts when its owning render path changes (exactly when a member opens), and the open list must survive that.
- Stack-list styling: cards flattened to a single surface (same pattern as the thread panel's card flattening), opaque `--tl-color-low` hover (the card floats over the canvas, so translucent tokens would let shapes show through), and the scroll container's clip box padded outward so card shadows aren't clipped.

Known limitation: threads can't be dragged apart while stacked — the badge has no drag affordance. `ThreadView`/`ThreadStackPin` are not exported from the package barrel yet; `computePinStacks` and `openStackId` are.

### Change type

- [x] `bugfix`

### Test plan

1. Place two comments on the same shape with default clicks (no Alt).
2. A count badge renders at the shape's anchor instead of stacked pins.
3. Click the badge → both threads listed as cards; click a card → it expands with a working reply composer; click the other card → it expands and the first collapses.
4. Escape collapses the expanded thread back to the list; Escape again closes the list.
5. Zoom out until the badge folds into clustering and back — the stack badge returns after the split.
6. A lone comment's pin, popover, drag, and region behaviour are unchanged.

- [x] Unit tests

### Release notes

- Fixed comments placed on the same shape overlapping into an unreachable stack of pins; they now show as a count badge that opens the threads as a list.

### DIFF
--- a/packages/commenting/api-report.api.md
+++ b/packages/commenting/api-report.api.md
@@ -327,6 +327,12 @@ export function commitCommentMutation<T>(editor: Editor, fn: () => T, kind?: 'dr
 // @public (undocumented)
 export function computeClusterTable(leaves: readonly LeafInput[], options: ClusterOptions): ClusterTable;
 
+// @public
+export function computePinStacks(editor: Editor, threads: readonly TLCommentThread[], impreciseShapeAnchor?: {
+    x: number;
+    y: number;
+}): Map<string, number>;
+
 // @public (undocumented)
 export function CountBadge({ count }: CountBadgeProps): JSX.Element;
 
@@ -470,6 +476,9 @@ export interface PendingComment {
 
 // @public
 export const pendingComment: EditorAtom<null | PendingComment>;
+
+// @public
+export const PIN_STACK_STEP_PX = 12;
 
 // @public
 export function putCommentRecords(editor: Editor, records: TLCommentRecord[]): void;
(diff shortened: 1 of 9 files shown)


jsscclr:
closing in favour of #9675

PR #9644, opened 2026-07-22
feat(commenting): add emoji reactions to comments

meg-an31:
Adds emoji reactions to comments — hover a comment, open the smiley picker, pick an emoji; a pill appears with a live count and your own reactions render highlighted. Reactions sync live, survive a room going cold, and name who reacted on hover. This PR also makes reactions **multi-select**, the **palette pluggable**, and the **reactor tooltip replaceable**.

## Design: a reaction is its own record

Each reaction is a `comment-reaction` record — one per `(comment, user, emoji)` — rather than a field on the comment or thread. This is the key decision and it's what makes multiplayer correct:

- **No shared field to clobber.** When you react, your client writes only your own record, so two people reacting at the same moment write different records and neither can overwrite the other.
- **Attribution is trivial.** A reaction carries its own `userId`, stamped from the session, so the sync authorizer reuses the same owner-only rule comments use — mirroring how `comment_read` models per-user facts.
- **Identity is structural.** The record id is derived from `(commentId, userId, emoji)`, and the authorizer validates that derivation on write. So a user holds one record per emoji (multi-select), re-picking an emoji toggles it, and a forged or mismatched id is rejected.

Reactions ride the sync object-store lane alongside comments and threads, persist to a `comment_reaction` Postgres table (migration 043), and rebuild on cold start — following the existing `comment`/`comment-thread` plumbing at every step.

## Multi-select

`CommentingOptions.allowMultipleReactions` (default `true`, the Slack model): each emoji toggles independently. Set it `false` for single-select, where picking a new emoji replaces your existing one. This is enforced client-side; the server accepts per-emoji records either way, verified by identity and the canonical id.

## Pluggable palette

A reaction's emoji is treated as an opaque **token** — the layer only stores, syncs, and hands it to a renderer, never assuming it's a glyph. So the whole palette is swappable via three extension points (all defaulting to the built-in emoji system, so plain emoji reactions need no config):

- `components.ReactionContent` / the `renderReaction` prop — how a token is drawn (glyph, ``, SVG, anything)
- `components.ReactionPalette` — what the add-reaction button opens
- `isAllowedReaction` — which tokens may be written

`drawing-reactions` ships a complete custom palette (draw-your-own, stored as a `data:` image token) as the worked example, and `REACTIONS.md` documents how to swap the palette. dotcom uses the defaults (plain emoji).

## Reactor tooltip

Hovering a reaction pill shows who reacted as an inline sentence — "You reacted", "You and Bo reacted", "You, Bo and Ada reacted", then "You, Bo, Ada and N others reacted" past three names. It always lists everyone who reacted, you included. It renders through the same `TldrawUiTooltip` primitive the toolbar uses (there is no separate hover-card component).

This is its own override point, independent of the palette, at two levels:

- **Reword it** by translating the `comments.reacted-*` strings — the grammar lives in the translation strings, so each locale controls word order via the `{a}`/`{b}`/`{c}`/`{count}` placeholders.
- **Replace it** via `components.ReactionTooltip`. The override receives the reactors **and the pill itself** (as `children`) and returns the whole affordance — so it owns the tooltip, its box, size, shape, and position, not just the text inside. The default splits into `DefaultReactionTooltip` (the wrapper) and `DefaultReactionTooltipContent` (the reusable sentence), so a custom tooltip can keep the default box or drop it entirely.

## Deliberate behaviours worth a reviewer's eye

- **Deleting a reaction is open by design.** Anyone with file access can hard-delete any reaction, and reactions carry no soft-delete / `isDeleted` flag. A reaction is a toggle (react / un-react), so removal is a plain record delete, and the comment/thread delete-cascade needs to sweep every reactor's records, not just the caller's own. Forging and tampering are still prevented — `userId` and the canonical id are owner-locked; only deletion is open.
- **The Zero `comment_reaction` table has no consumer yet** — it mirrors `comment_mention` so a future app-level "reacted to your comment" query can reach reactions without a schema change.
- **Palette validation is client-side** — a scripted client's junk tokens aren't hard-bounded server-side yet (the server verifies identity + id, not palette membership). Reaction content (like the comment body) is also unbounded in size; this is consistent with the existing comment model rather than new to reactions.

## Known follow-ups

- **Token in the id.** Large tokens (e.g. an inlined drawing) bloat the record id / Postgres PK; the drawing palette caps token size, but the clean fix is to key the id on a hash of the token.
- **Config fragmentation.** The three palette extension points should collapse into one `reactionPalette` object; documented as-is for now.
- **Server-side reaction cleanup** when a comment/thread is pruned by the author cascade (the direct case is handled).

## Testing

- Unit tests for the record (id derivation, injectivity, validator), the authorizer (forging, no-identity, cross-user tamper, id-slot forgery, emoji-immutable-on-update), the row round-trip and drain routing, the FK-violation predicate, the palette check, and the reaction tally.
- `yarn typecheck`, `yarn api-check`, and `yarn lint-current` clean. The one failing sync-worker test (`replicatorMigrations` snapshot) is pre-existing and unrelated.
- Not exercised in the running app with two live accounts — the multiplayer merge and cold-start round-trip are verified by tests and tracing, not a click-through.

### Change type

- [ ] `bugfix`
- [x] `feature`
- [ ] `improvement`
- [x] `api`
- [ ] `other`

### Test plan

- Unit tests across tlschema, the sync-worker, and commenting (record/authorizer/rows/palette/tally). Typecheck, api-check, lint clean; only the pre-existing `replicatorMigrations` snapshot fails.
- Manual two-account multiplayer + cold-start verification still to do.

### Release notes

- Added emoji reactions to comments, with multi-select, a pluggable reaction palette, and a replaceable reactor tooltip.

### API changes

New in `@tldraw/tlschema`:

- `TLCommentReaction` record + `TLCommentReactionId`; `createCommentReaction`, `createCommentReactionId(commentId, userId, emoji)`, `isCommentReactionId`, `commentReactionRecordConfig`. `commentSchemaRecords` includes `comment-reaction`.

New in `@tldraw/commenting`:

- `CommentReactions`, `CommentReactionPicker`, `Reactions`, `Reaction`, `ReactionPicker`, `EmojiPicker`, `DEFAULT_REACTION_EMOJI`.
- `summarizeReactions`, `toggleCommentReaction`, `useCommentReactions`, `getCommentReactions`, `isAllowedReactionEmoji`.
- Token/palette API: `RenderReaction`, `defaultRenderReaction`, and `CommentingComponents.ReactionContent` / `.ReactionPalette`; `CommentingOptions.allowMultipleReactions`, `.isAllowedReaction`.
- Reactor tooltip API: `CommentingComponents.ReactionTooltip`, `DefaultReactionTooltip`, `DefaultReactionTooltipContent`, `ReactionTooltipProps` (the tooltip override receives the pill as `children`).
- The `drawing-reactions` module: `DrawingReactionPalette`, `DrawingReactionContent`, `renderDrawingReaction`, `isDrawingReactionToken`, `exportDrawingReactionToken`, `DrawingReactionTooLargeError`.
- `TLCommentRecord` includes `TLCommentReaction`; `removeCommentRecords` accepts `TLCommentReactionId`.

The reactor list uses the existing `TldrawUiTooltip` primitive — no new `tldraw`-level component is added.


jsscclr:
would be good to have a reactions example page in the examples app!

packages/tlschema/src/records/TLComment.ts
```diff
@@ -233,6 +289,33 @@ export function createCommentId(id?: string): TLCommentId {
 	return createCustomRecordId('comment', id) as TLCommentId
 }
 
+/**
+ * The id of one user's reaction to one comment.
+ *
+ * Derived from the pair rather than random, so a user reacting twice to the same comment addresses
+ * the same record: picking a different emoji overwrites their reaction instead of adding a second
+ * one, and doing it from two tabs at once converges on one record rather than racing to create
```
jsscclr: do we only want 1 reaction per user? think I would want more!

meg-an31: ooo wait this was a conscious decision by me, can defo go back on it! i'll make it an optional flag for sdk stuff

jsscclr: sounds good! happy to do whatever for dotcom (although I am a big multi-react kind of gal), but we should support multi-reactions at least in sdk

mimecuvalo: +1 on multi-reaction being the default, a single-reaction is kinda only for chat-like apps (whatsapp, etc. — and i don't think github's a great example of single-reaction)

packages/commenting/src/canvas/options.ts
```diff
@@ -55,6 +55,12 @@ export interface CommentingOptions {
 	// ── Feature toggles ──────────────────────────────────────────────────────────────────────
 	/** Fold nearby pins into count badges as the camera zooms out. */
 	readonly enableClustering: boolean
+	/**
+	 * Whether the current user appears in the hover list of who reacted with an emoji. When false,
+	 * that list shows only other people — so a reaction only you made shows no list at all.
+	 * Defaults to true.
+	 */
+	readonly showSelfInReactionList: boolean
```
mimecuvalo: eh, i don't think we need this as an option, let's nix it.
my bias would be to reduce the amount of options as much as possible until there's maybe a need for it downstream

packages/commenting/src/ui/comments.css
```diff
@@ -375,20 +381,101 @@
...
+/* your own reactions read as filled-in: a deeper wash of the selection colour than the hover
+   state, so "I reacted with this" is distinguishable from "my pointer is over this" */
 .tlui-cmt-reaction--active {
 	border-color: var(--tl-color-selected);
-	background: color-mix(in srgb, var(--tl-color-selected) 12%, var(--tl-color-panel));
+	background: color-mix(in srgb, var(--tl-color-selected) 28%, var(--tl-color-panel));
+}
+
+.tlui-cmt-reaction--active:hover,
+.pseudo-hover .tlui-cmt-reaction--active {
```
mimecuvalo, line 376: not for this PR but we should clean up these pseudo-hover rules later cc @jsscclr

apps/dotcom/sync-worker/src/commentRows.ts
```diff
@@ -57,6 +60,24 @@ export function isCommentMentionFkViolation(error: unknown): boolean {
...
+ * 23503): the comment, thread, or reacting user's row is gone. Deleting any of them cascades the
+ * reaction rows away, so hitting this means a warm room still holds a reaction for something
+ * Postgres has already removed. Retrying can never succeed, so the caller prunes the record
+ * instead — the same contract as {@link isCommentAuthorFkViolation}.
+ */
+export function isCommentReactionFkViolation(error: unknown): boolean {
+	if (typeof error !== 'object' || error === null) return false
+	const { code, constraint } = error as { code?: unknown; constraint?: unknown }
+	return (
+		code === '23503' &&
```
mimecuvalo, line 74: fine for this PR — cc @MitjaBezensek this seems like something to cleanup? is this the best way to capture these errors for now?

apps/dotcom/zero-cache/migrations/043_add_comment_reactions.sql
```diff
@@ -0,0 +1,42 @@
...
+--
+-- `meta` and `lastChangedClock` are persistence/rehydration columns not declared in the Zero
+-- schema, so they never reach clients.
+
+CREATE TABLE comment_reaction (
+  "id" VARCHAR PRIMARY KEY,
+  "fileId" VARCHAR NOT NULL,
+  "commentId" VARCHAR NOT NULL,
+  "threadId" VARCHAR NOT NULL,
+  "pageId" VARCHAR NOT NULL,
```
mimecuvalo, line 20: mm, i think we should nix `pageId` (maybe `threadId` too?) 
if for some reason the comment is moved to a different page i wouldn't want to think updating the page in another place. is the reason we need the pageId/threadId to construct the notification for easier linking / lookups?

apps/dotcom/zero-cache/migrations/043_add_comment_reactions.sql
```diff
@@ -0,0 +1,42 @@
...
+  "userId" VARCHAR NOT NULL,
+  "emoji" VARCHAR NOT NULL,
```
mimecuvalo, line 22: in a followup PR we should think about custom emojis :)

at Paper we used Private Use Areas in Unicode: https://en.wikipedia.org/wiki/Private_Use_Areas
so, it was:
- a unicode character (privately assigned across a workspace)

in the backend that would map to a URL, canonical name, and related tags

packages/commenting/src/ui/reaction-picker.tsx
```diff
@@ -0,0 +1,83 @@
...
+			</TldrawUiDropdownMenuTrigger>
+			{/* right-aligned under the trigger, so the grid hangs down the card's right edge */}
+			<TldrawUiDropdownMenuContent side="bottom" align="end">
+				<EmojiPicker emoji={emoji} selected={selected} onSelect={onSelect} />
+			</TldrawUiDropdownMenuContent>
+		</TldrawUiDropdownMenuRoot>
+	)
+}
+
+function SmileyIcon() {
```
mimecuvalo, line 82: let's use the Radix icon here btw (that's what we use across the app) https://www.radix-ui.com/icons

packages/commenting/src/ui/reaction.tsx
```diff
@@ -1,19 +1,85 @@
...
+	if (!enableHoverList || !hasOthers) return pill
+
+	return (
+		// A short open delay so the reactor list feels responsive on hover, rather than the
+		// tooltip-length default.
+		<TldrawUiHoverCard openDelay={300}>
+			<TldrawUiHoverCardTrigger>{pill}</TldrawUiHoverCardTrigger>
+			<TldrawUiHoverCardContent side="bottom" className="tlui-cmt-reactors">
+				<div className="tlui-cmt-reactors__title">
+					{msg('comments.reacted')} {emoji}:
```
mimecuvalo: i think to make this hover a bit cleaner (i'm thinking i18n where "reacted + emoji" might not translate well): prbly let's just list the names, i think that'll be fine (names that are comma-separated, no need to state again that what emoji it is as that's a given from the hover)

packages/tldraw/src/lib/ui/components/primitives/TldrawUiHoverCard.tsx
```diff
@@ -0,0 +1,89 @@
+import { useContainer } from '@tldraw/editor'
+import classNames from 'classnames'
+import { HoverCard as _HoverCard } from 'radix-ui'
+import React from 'react'
+import { useDirection } from '../../hooks/useTranslation/useTranslation'
+
+/** @public */
+export interface TLUiHoverCardProps {
```
mimecuvalo: given the prev. comment — i think we can get away with just a tooltip in this case and not have this new UI component 👍

packages/commenting/src/canvas/comment-reactions.tsx
```diff
@@ -0,0 +1,200 @@
...
+	currentUserId?: string | null,
+	resolveName?: (userId: string) => string | undefined
+): ReactionSummary[] {
+	const groups = new Map<
+		string,
+		{ count: number; active: boolean; firstAt: number; reactors: ReactionSummary['reactors'] }
+	>()
+	for (const reaction of reactions) {
+		const mine = currentUserId != null && reaction.userId === currentUserId
+		const reactor = { name: resolveName?.(reaction.userId) ?? reaction.userId, you: mine }
```
mimecuvalo: let's not fallback to reaction.userId — ends up looking weird :P

Image

let's use "Someone" like we use elsewhere

packages/commenting/src/ui/emoji-picker.tsx
```diff
@@ -0,0 +1,53 @@
...
+	/** Called when an emoji is chosen. */
+	onSelect?(emoji: string): void
+}
+
+/**
+ * A grid of emoji to react with. Presentational and unpositioned — the host owns placement and
+ * dismissal; `ReactionPicker` renders it in a dropdown under the add-reaction button.
+ * @public @react
+ */
+export function EmojiPicker({
```
mimecuvalo, line 47: i think we're gonna want this component and related reaction ui components to be swappable via `CommentingComponents`. i very easily can imagine folks wanting to bring in their own emoji picker that has many more emojis (and virtualized lists of emojis since there are so many of them)

packages/commenting/src/canvas/options.ts
```diff
@@ -55,6 +55,12 @@ export interface CommentingOptions {
...
```
mimecuvalo, line 143: to followup on the options here, i think the option we can add here is a boolean: enableReactions

packages/commenting/src/canvas/comments-overlay.tsx
```diff
@@ -1320,46 +1338,56 @@ const ThreadPin = memo(function ThreadPin({
...
+												onClick={() => deleteComment(comment)}
+											>
+												<span>{msg('comments.delete-comment')}</span>
+											</button>
+										</TldrawUiDropdownMenuItem>
+									</TldrawUiDropdownMenuGroup>
+								</TldrawUiDropdownMenuContent>
+							</TldrawUiDropdownMenuRoot>
+						)}
+						<CommentReactionPicker comment={comment} currentUserId={currentUserId} />
```
mimecuvalo: looks like the reaction button is running into long text oops 🫠 

Image

mimecuvalo (commented):
good stuff! i love me some reacji
some comments to address before landing — very excited that this is happening ^_^

packages/commenting/src/ui/reaction.tsx
meg-an31: 🫡🫡🫡🫡

packages/commenting/src/canvas/comment-reactions.tsx
meg-an31: OOPS nice catch lol

packages/commenting/src/ui/comments.css
jsscclr: what's the intent of a 'pseudo-hover' @meg-an31?

apps/dotcom/zero-cache/migrations/043_add_comment_reactions.sql
meg-an31: oh this is super interesting !

packages/commenting/src/canvas/options.ts
meg-an31: should i keep this as the default behaviour then? or have the menu always show even if you're the only react-er?

packages/commenting/src/ui/reaction.tsx
meg-an31: i kept this as each name on a new line but lmk if you think i should swap to comma separated

packages/commenting/src/ui/comments.css
meg-an31: wait idk whether that was me that added this

packages/commenting/src/canvas/comments-overlay.tsx
meg-an31: i still dont know exactly what to do with this... cut down the length of comment that can show? extend the box? hmmm

packages/commenting/src/ui/comments.css
mimecuvalo: yeah this is some sketchbook code afaict

packages/commenting/src/canvas/options.ts
mimecuvalo: yeah, let's keep as default behaviour 👍

packages/commenting/src/canvas/comments-overlay.tsx
mimecuvalo: discussed offline: i think let's move Edit | Delete below the text of the comment, only for authors (so the kebab menu disappears / is nixed)

PR #9646, opened 2026-07-22
feat(dotcom): add asset diagnostics to the admin page

MitjaBezensek:
In order to diagnose the repeated `R2 connection queue depth` Sentry alerts (same file, same count every open = the asset-association pass never makes progress, likely because source objects are gone from the uploads bucket), this PR adds a read-only asset diagnostics tool to the admin page.

Given a file slug, `GET /app/admin/file-assets/:slug` reports per asset whether its object exists in the uploads bucket and whether it's associated with the file, plus the Postgres `asset` table cross-check and whether the file's `createSource` still exists. The admin page shows a summary grid, a problem-assets table, and raw JSON.

Read-only for now; it reads the last persisted snapshot. In local dev the bucket check always reports missing (each worker's `wrangler dev` has its own simulated R2 state).

### Change type

- [x] `other`

### Test plan

1. In `/admin`, enter a file slug under "Asset diagnostics"
2. A healthy file reports all assets associated and present; a stuck file (from the Sentry alerts) reports pending assets missing from the bucket

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change |
| ------- | ---------- |
| Apps    | +309 / -3  |

### DIFF
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -806,9 +806,8 @@ function AssetDiagnostics() {
 		<div className={styles.fileOperation}>
 			<h4 className="tla-text_ui__medium">Asset diagnostics</h4>
 			<p className="tla-text_ui__regular">
-				Checks every asset in a file&apos;s last persisted snapshot: is its object still in the
-				uploads bucket, and is it associated with the file? Reads the persisted snapshot, which can
-				lag a live room by a persist tick.
+				Checks whether each asset in the file&apos;s last persisted snapshot exists in the uploads
+				bucket and is associated with the file.
 			</p>
 			{error && <div className={styles.errorMessage}>{error}</div>}
 			<div className={styles.searchContainer}>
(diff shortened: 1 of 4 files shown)


apps/dotcom/client/src/pages/admin.tsx
```diff
@@ -8,7 +14,7 @@ import { useTldrawCurrentUser } from '../tla/hooks/useUser'
 import styles from './admin.module.css'
 
 // Helper component for structured data display.
-function StructuredDataDisplay({ data }: { data: ZStoreData }) {
+function StructuredDataDisplay({ data }: { data: object }) {
```
frolic, line 17: 🙈

mimecuvalo (approved):
as a followup, would be good to have:
- stats of a files shape types (3 images, 2 geos, etc.)
- stats of file sizes for assets

MitjaBezensek:
Added both in 48dcbaaf8c: shape counts by type and asset size stats (total + largest, from the head calls we already make) now show in the summary.

PR #9647, opened 2026-07-22
fix(dotcom): revert the thumbnail zoom floor

driev:
In order to get board thumbnails rendering again, this reverts `07f2a2a0ac` ("fit thumbnail content below the default zoom floor"), which shipped as part of #9503.

That commit lowered the render page's zoom floor from the default 0.05 to 0.001, so that boards larger than roughly 22k × 10k page units would fit the thumbnail viewport instead of rendering with their edges cropped. After it shipped, renders through Cloudflare Browser Rendering stopped completing: the page never reached `data-thumbnail-ready`, the capture spent its full 45s `THUMBNAIL_RENDER_TIMEOUT_MS` and came back as a Cloudflare 6002 selector timeout, and link unfurls for affected boards fell back to the generic tldraw image. Reverting restores rendering.

The mechanism is not confirmed. `exportThumbnailImage` takes all three of its `editor.toImage` inputs from the camera — `bounds` is the viewport in page space, `scale` is `camera.z`, and the shape list is everything not culled at that viewport — so a 50× lower floor makes the viewport cover 2500× the page area and stops the culling filter excluding anything. That would hand `toImage` considerably more than it was meant to draw, but it has not been measured from inside the fleet.

This reinstates the cropping the original commit was fixing: boards past that size render with their edges cut again. Fixing it properly needs an approach that doesn't widen what the export is handed.

### Change type

- [x] `bugfix`

### Test plan

1. Call `get_shared_board_screenshot` through `POST /api/app/mcp` for a board that was timing out, and confirm it returns a PNG rather than a render failure.
2. Confirm a board larger than ~22k × 10k page units renders cropped again — the known regression this reintroduces.

- [ ] Unit tests
- [ ] End to end tests

### DIFF
--- a/apps/dotcom/client/src/pages/thumbnail-render.tsx
+++ b/apps/dotcom/client/src/pages/thumbnail-render.tsx
@@ -250,38 +250,20 @@ function getRepresentativeContentInset(width: number, height: number) {
 	return Math.max(48, Math.min(160, width * 0.12, height * 0.18))
 }
 
-// Zoom floor for the content fit. The camera's zoom steps double as the clamp range for both
-// zoomToBounds and setCamera, and the default floor (0.05) is far too high here: at 1200x630 it
-// stops fitting at roughly 22k x 10k page units, so a bigger board quietly renders with its edges
-// cropped. Nothing on this page zooms interactively, so the floor exists only to bound the fit —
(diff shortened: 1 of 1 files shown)


mimecuvalo (approved):
makes sense to me 👍 let's give it a go

PR #9650, opened 2026-07-22
fix(commenting): open the linked thread when a notification is clicked

jsscclr:
## Problem

Clicking a notification did nothing when it pointed at the file you were already viewing, and on a cross-file click it panned to the shape but never opened the thread.

A notification row navigates to the file route with `?comment=<id>&d=<deeplink>`, but nothing consumed `?comment=` — and `d` is read exactly once, in `TlaEditor`'s `handleMount`. Same-route navigation only rewrites the query string, so the editor never remounts and neither param is ever read.

## Fix

Adds `SneakyCommentDeepLink` inside the editor (following the `Sneaky*` pattern). It watches `?comment=`, reactively waits for the linked comment and its thread records to sync into the store (on a cross-file open they arrive after mount), then `focusThread`s it — page switch, popover open, center on pin — unhiding pins if they were toggled off, and strips the consumed param with a `replace` so back/refresh doesn't re-open it. Passes the editor's configured `impreciseShapeAnchor` through, matching the comments sidebar's row-click.

### Test plan

1. Open a file, have another user (or account) mention you in a comment on that same file
2. Click the notification → the thread popover opens, camera centers on the pin
3. Repeat from a different file → navigates, then opens the thread once comments sync in
4. Hide comment pins, click a notification → pins unhide and the thread opens
5. Back/refresh after following a link → thread does not re-open
6. From a second account, delete a comment that mentioned you, then click its (stale) notification → "This comment was deleted." toast, notification marked read, no camera move

### API changes

`@tldraw/commenting`:
- New: `revealThreadRequest` — a per-editor `EditorAtom<string | null>` holding a pending request to reveal a thread (a thread or comment id). Write an id to open that thread and bring it into view; `CanvasComments` serves the request once the records have synced (zooming to the cluster split that reveals the pin when clustered, unhiding pins) and clears it.
- Behavior change: the package no longer reads `?comment=` from the URL itself. URL deep-linking is now the host app's responsibility — write the param's id into `revealThreadRequest` (tldraw.com does this in `SneakyCommentDeepLink`). Consumers relying on the built-in URL read need that one-line adapter.

### DIFF
--- a/packages/commenting/src/canvas/comments-overlay.tsx
+++ b/packages/commenting/src/canvas/comments-overlay.tsx
@@ -77,6 +77,7 @@ import {
 	openThreadId,
 	pendingComment,
 	regionDraft,
+	revealThreadRequest,
 	toggleCommentsHidden,
 	usePendingComment,
 } from './state'
@@ -233,7 +234,6 @@ function CanvasCommentsLayer(props: CanvasCommentsProps) {
 	// canvas, so any pin past its bottom/right edge inflates the root's scrollHeight — which the
 	// wheel hook's is-this-scrollable guard reads as scrollable, silently disabling pass-through.
 	usePassThroughMouseOverEvents(layerRef)
-	const deepLinkHandled = useRef(false)
 	const threads = useCommentThreads(editor)
 	const pending = usePendingComment()
 	const openId = useValue('open thread id', () => openThreadId.get(editor), [editor])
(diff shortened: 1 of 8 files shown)


packages/commenting/src/canvas/comments-overlay.tsx
```diff
@@ -401,46 +399,65 @@ function CanvasCommentsLayer(props: CanvasCommentsProps) {
...
 	}, [editor])
 
-	// Open the thread named by a deep link (?comment=<thread or comment id>). If the thread is
-	// currently inside a cluster, zoom to the first split that reveals it before opening.
+	// Seed a reveal request from a deep link (?comment=<thread or comment id>) once at mount, so a
+	// plain shared link works with no consumer wiring. Requests after mount come from writers of
+	// {@link revealThreadRequest} — the URL is not watched for changes here.
 	useEffect(() => {
-		if (deepLinkHandled.current) return
 		const id = new URLSearchParams(window.location.search).get('comment')
```
mimecuvalo: re-examining this: arguably, we shouldn't be looking at the URL here for actionable parameters — feels leaky on the part of the commenting pkg. it might be better if the SneakyCommentDeepLink handles this all and it's up to the downstream SDK user maybe to handle. thoughts?

mimecuvalo (commented):
one other idea here: if a comment is deleted, what should happen? a toast?

packages/commenting/src/canvas/comments-overlay.tsx
jsscclr: I'm happy with this - doesn't feel quite right at the moment, moving the logic to app makes the most sense

jsscclr:
> one other idea here: if a comment is deleted, what should happen? a toast?

@mimecuvalo what scenario are you thinking here? if someone on your document deletes their comment? or a thread is deleted?

mimecuvalo:
> @mimecuvalo what scenario are you thinking here? if someone on your document deletes their comment? or a thread is deleted?

yeah, if someone deletes a comment that mentioned me, i have the notification still and clicking it takes me to an empty area / empty comment. two options might be: displaying a toast or maybe having a ghost-comment saying "comment deleted"

oh actually, that reminds me: if a comment is resolved what happens at the moment if open the linked thread? should we should a ghost 👻 comment (a comment with some special treatment showing it's not actionable anymore) in it's place? or does the sidebar open and highlight the resolved comment?

mimecuvalo (approved):
LGTM!

PR #9651, opened 2026-07-22
feature(commenting): configurable shape-comment precision (preciseShapeAnchors)

jsscclr:
Following the discussion with Tim about whether SDK consumers want shape+offset commenting, shape-only, or both: this makes it a commenting option, with an example so we can align on the three behaviours.

### What

- Adds `preciseShapeAnchors` to `CommentingOptions`, set via `CommentTool.configure`:
  - `'always'` (the default) — every shape comment is precise (pinned to the exact clicked spot within the shape)
  - `'never'` — every shape comment is imprecise (addresses the shape as a whole; the pin renders at `impreciseShapeAnchor`)
  - `'alt'` — the previous behaviour: imprecise unless Alt is held while placing
- Both paths that create shape anchors resolve through the new `resolveShapeAnchorPrecision`: placing with the comment tool, and dropping a dragged pin onto a shape.
- The clicked `x`/`y` are stored either way — the mode only gates `isPrecise` — so the choice affects rendering, not the data model, and existing anchors render as stored.
- **Behaviour change**: precise is now the default. Consumers relying on the old Alt-gated placement opt back in with `'alt'`. dotcom needs no change — it inherits the default.

### Example

New `comment-shape-precision` example with a mode switcher (shared store across switches) to feel the difference between the three modes.

### Test plan

- Verified all four mode × Alt combinations live in the example, checking the created anchor records
- `packages/commenting` tests (179), typecheck, api-report regenerated

### API Changes

- New option on `CommentingOptions`: `preciseShapeAnchors: 'always' | 'never' | 'alt'`, defaulting to `'always'`. This changes default placement behaviour: shape comments are precise unless configured otherwise (the previous behaviour is `'alt'`).
- New export: `resolveShapeAnchorPrecision(editor, altKey)` — resolves whether a shape placement lands precise, per the configured mode.

### DIFF
--- a/packages/commenting/src/canvas/options.ts
+++ b/packages/commenting/src/canvas/options.ts
@@ -59,6 +59,15 @@ export interface CommentingOptions {
 	// ── Anchoring ────────────────────────────────────────────────────────────────────────────
 	/** Normalized (0–1) spot within a shape where imprecise shape pins sit. Default top-right. */
 	readonly impreciseShapeAnchor: { readonly x: number; readonly y: number }
+	/**
+	 * When placing (or drag-re-anchoring) a comment on a shape, whether the anchor is precise —
+	 * pinned to the exact clicked spot within the shape — or imprecise — pinned to the shape as a
+	 * whole, rendered at `impreciseShapeAnchor`. `'always'` (the default) and `'never'` fix it
+	 * either way and ignore the modifier; `'alt'` makes it the user's call per placement:
+	 * imprecise normally, precise while Alt is held. Governs new placements only; existing anchors
+	 * render as stored.
+	 */
+	readonly preciseShapeAnchors: 'always' | 'never' | 'alt'
 
 	// ── Clustering tuning ─────────────────────────────────────────────────────────────────────
 	/** Screen-pixel margin by which the viewport is inflated when culling cluster badges. */
@@ -81,6 +90,7 @@ export const defaultCommentingOptions = {
 	dragHistory: undefined,
 	enableClustering: true,
 	impreciseShapeAnchor: { x: 1, y: 0 },
+	preciseShapeAnchors: 'always',
 	clusterCullMargin: 120,
 	clusterSplitZoomFactor: 1.05,
 	components: {},
--- a/apps/examples/src/examples/collaboration/comment-shape-precision/CommentShapePrecisionExample.tsx
+++ b/apps/examples/src/examples/collaboration/comment-shape-precision/CommentShapePrecisionExample.tsx
@@ -1,0 +1,127 @@
+import {
+	CanvasComments,
+	CommentAuthor,
+	CommentingOptions,
+	CommentTool,
+	commentToolOverrides,
+} from '@tldraw/commenting'
+import { getLicenseKey } from '@tldraw/dotcom-shared'
+import { useMemo, useState } from 'react'
+import {
+	commentSchemaRecords,
+	createTLSchema,
+	createTLStore,
+	Editor,
+	TLComponents,
+	Tldraw,
+	TldrawUiButton,
+	TldrawUiButtonLabel,
+	toRichText,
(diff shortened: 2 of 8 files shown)


MitjaBezensek:
Looks good 👍

An alternative API could be something like 
```ts
CommentTool.configure({ shouldBePrecise: (_editor: Editor) => true })
```
This would allow complete customization depending on some dynamic data. Like use precise for frames, but not for geo shapes. We do something similar in shape utils with all our `canSomething` methods. Not sure what's better tbh 🤷‍♂️

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
```diff
@@ -63,6 +63,9 @@ import { TlaEditorWrapper } from './TlaEditorWrapper'
 import { useExtraDragIconOverrides } from './useExtraToolDragIcons'
 import { useFileEditorOverrides } from './useFileEditorOverrides'
 
+// Comments on tldraw.com pin to the exact clicked point within a shape, no modifier needed.
+const commentToolsForDotcom = [CommentTool.configure({ preciseShapeAnchors: 'always' })]
```
mimecuvalo: i forget our discussion around this, i thought dotcom would have `alt` as the default?

packages/commenting/src/canvas/options.ts
```diff
@@ -59,6 +59,14 @@ export interface CommentingOptions {
...
 	/** Normalized (0–1) spot within a shape where imprecise shape pins sit. Default top-right. */
 	readonly impreciseShapeAnchor: { readonly x: number; readonly y: number }
+	/**
+	 * When placing (or drag-re-anchoring) a comment on a shape, whether the anchor is precise —
+	 * pinned to the exact clicked spot within the shape — or imprecise — pinned to the shape as a
+	 * whole, rendered at `impreciseShapeAnchor`. `'alt'` (the default) makes it the user's call per
+	 * placement: imprecise normally, precise while Alt is held. `'always'` / `'never'` fix it either
+	 * way and ignore the modifier. Governs new placements only; existing anchors render as stored.
+	 */
+	readonly preciseShapeAnchors: 'always' | 'never' | 'alt'
```
mimecuvalo, line 70: i think this set as a trio makes a lot of sense as far just enough options to make it work for folks.
my quibble here would be `alt` is a weird naming for the middle state: maybe `allowable`/`optionable`/`switchable`? open to suggestions here

and nit: i'd put `alt` in the middle to kinda denote the spectrum, or, put `alt` first because if alt is the default it feels like it should go first to help signal that

apps/dotcom/client/src/tla/components/TlaEditor/TlaEditor.tsx
jsscclr: Ah, honestly I made an executive decision here! dotcom behaviour feels pretty strange to me as is - I expect the comment to be x/y positioned based on where I pinned it. Happy to change back to 'alt', but it feels quite unintuitive to me at the moment!

packages/commenting/src/canvas/options.ts
jsscclr: agreed! hmmm... 

`'modifier' | 'always' | 'never'`?

to indicate that precise shape anchors is triggered by the modifier key? I like modifier first if it's the default

jsscclr: or `optional`? others just have a weird mouth feel to me to say 😆

jsscclr:
I think @MitjaBezensek has the right idea - was chatting with @meg-an31 and we were thinking we might want different anchoring behaviour for different elements - e.g. for a rectangle, just on the shape is fine, whilst for a line it can be a bit weird to have the same anchoring behaviour

jsscclr:
Closing in favour of #9662

PR #9652, opened 2026-07-22
feat(commenting): sign-in prompt in place of composers for signed-out viewers

MitjaBezensek:
In order to let signed-out visitors on a shared file read comments without hitting a dead end, this PR gives them a "Sign in to comment" call to action where the composer would be. The backend already lets guests read comments, but the composer surfaces silently vanished for a null `currentUserId` — no path from "I want to reply" to signing in.

Two pieces of SDK surface, both via `CommentTool.configure`:

- `canComment` — `({ editor, currentUserId }) => boolean`, defaults to `currentUserId != null`. Gates every comment write affordance: composers, edit/delete menus, resolve/reopen/delete, and pin/region dragging (previously ungated).
- `ComposerFallback` — component slot rendered in a composer's place when `canComment` resolves false, with `context: 'pending' | 'thread'` to tell the comment tool's placement popover from the thread footer. Unset keeps those surfaces empty, so this is opt-in.

tldraw.com registers a sign-in CTA (tracked as `sign-up-clicked`, source `comments`) and points the comment tool's button and `c` shortcut at the sign-in dialog for anon viewers. Writing while signed out stays impossible server-side (authorizer + read-only object lane); this ships the read+CTA half only.

### Change type

- [x] `feature`

### Test plan

1. Open a shared file while signed out: pins render; a thread opens read-only (no resolve/edit/delete) with a "Sign in to comment" button that opens the sign-in dialog.
2. Signed out, try dragging a pin or resizing a region: nothing moves. The comment tool button and `c` open the sign-in dialog.
3. Signed in: composers, thread actions, and pin dragging are unchanged.

- [x] Unit tests

### API changes

- Added `CommentingOptions.canComment` and the `getCanComment` / `useCanComment` helpers (plain read / reactive hook)
- Added `CommentingComponents.ComposerFallback`, receiving `{ context: 'pending' | 'thread' }`
- Added `CommentThreadProps.footer?: ReactNode` — slot at the bottom of the thread, after the composer

### Code changes

| Core code       | +157 / -33 |
| Tests           | +50 / -1   |
| Automated files | +22 / -1   |
| Apps            | +74 / -5   |

### Release notes

- Show a "Sign in to comment" prompt to signed-out viewers on shared files, in place of the comment composers. SDK: gate commenting with `CommentTool.configure({ canComment })` and register a `ComposerFallback` component; `CommentThread` accepts a `footer` slot.


packages/commenting/src/canvas/comments-overlay.tsx
```diff
@@ -239,6 +240,16 @@ function CanvasCommentsLayer(props: CanvasCommentsProps) {
 	const deepLinkHandled = useRef(false)
 	const threads = useCommentThreads(editor)
 	const pending = usePendingComment()
+	const canComment = useCanComment(props.currentUserId)
+	// With composing blocked and no fallback slot there's nothing to render for a pending comment —
+	// and the dismiss handlers (Escape, click-away) live inside PendingComposer, which would never
+	// mount. Clear the atom instead of stranding it (a stale pending would pop a composer at the
+	// old click point if `canComment` later flips true).
+	const showPendingComposer =
+		pending != null && (canComment || options.components.ComposerFallback != null)
```
jsscclr: little nit! my brain melts trying to read these kinds of conditionals - could we perhaps do something like this instead?

```
const canRenderComposer = canComment || options.components.ComposerFallback != null
const showPendingComposer = pending != null && canRenderComposer
```

jsscclr (approved):
LGTM!

packages/commenting/src/canvas/comments-overlay.tsx
MitjaBezensek: Same tbh, nice shout!

MitjaBezensek: Done in 522e119862.

PR #9657, opened 2026-07-22
fix(commenting): layer pins between selection overlays and collaborator cursors

jsscclr:
Comment pins currently paint above everything drawn on the canvas — including collaborator cursors, which slide underneath pins. This PR re-layers the canvas so pins sit where they should: above shapes and selection chrome, below cursors.

Two halves, which only make sense together:

- **Pins move inside the canvas.** Collaborator cursors are 2D-drawn on the overlays canvas inside `.tl-canvas`, whose `contain: strict` flattens it into a single paint layer — a layer portaled outside it (the pins today) can only paint entirely above or below the whole canvas. The pin layer now portals into `.tl-canvas` itself at z 400 (above the shapes layer at 300). The thread popover and pending composer still portal to the container and keep their UI-layer stacking; pin pointer handling is unchanged (React events bubble through the React tree, not the DOM).
- **The overlays canvas splits into two bands.** Alone, the move would put *all* overlay drawing — selection outlines, brush, handles, snaps, not just cursors — above the pins. The overlays now render to two stacked canvases: an `under` band (z 350) beneath DOM content mounted inside the canvas and an `over` band (z 500) above it, chosen per util via a new `band` option on `OverlayUtil` (`'under' | 'over'`, default `'under'`). Collaborator cursors and their off-screen hints are `over`; everything else stays `under`.

Net stacking inside the canvas: shapes → selection/brush/indicators → comment pins → collaborator cursors. Known tradeoff: a resize handle under a pin is occluded by it, like the outline.

### Change type

- [x] `bugfix`
- [x] `api`

### Test plan

1. With a second client connected, move the other cursor over a comment pin — the cursor draws above it.
2. Create a comment on a shape's corner and select the shape — the pin covers the selection outline, not the reverse.
3. Marquee-select, scribble, snap, and handle interactions all still draw (under band); minimap unchanged.

- [x] Unit tests

### Release notes

- Comment pins now stack between selection overlays and collaborator cursors. `OverlayUtil` options gain a `band` field (`'under' | 'over'`) choosing which of the two stacked overlay canvases a util paints to.

### API changes

- `OverlayUtil` options gain an optional `band?: TLOverlayBand` field (`'under' | 'over'`, default `'under'`) choosing which of the two stacked overlay canvases a util paints to; new exported type `TLOverlayBand`. `CollaboratorCursorOverlayUtil` and `CollaboratorHintOverlayUtil` set `band: 'over'`; all other built-in utils are unchanged.

### DIFF
--- a/apps/examples/e2e/tests/test-overlay-snapshots.spec.ts
+++ b/apps/examples/e2e/tests/test-overlay-snapshots.spec.ts
@@ -7,10 +7,10 @@ declare const editor: Editor
 
 // Screenshot-based regression tests for canvas overlay appearance.
 //
-// Overlays render to a single <canvas class="tl-canvas-overlays"> element, so
-// we screenshot the `.tl-canvas` region and let the image compare cover every
-// overlay kind (selection foreground, shape handles, brush, snap indicators,
-// scribble, etc.) at once.
+// Overlays render to a pair of stacked <canvas class="tl-canvas-overlays">
+// elements (the 'under' and 'over' bands), so we screenshot the `.tl-canvas`
+// region and let the image compare cover every overlay kind (selection
+// foreground, shape handles, brush, snap indicators, scribble, etc.) at once.
 //
 // Tests use direct editor state injection (brush/zoomBrush via
 // updateInstanceState, scribbles via ScribbleManager, snap indicators via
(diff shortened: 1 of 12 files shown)


jsscclr:
Closing, superseded by #9695

PR #9661, opened 2026-07-22
feat(commenting): keep threads alive across shape deletion and page moves

jsscclr:
Comment threads anchored to a shape currently die with it in two ways: deleting the shape leaves the thread stranded invisibly (a missing shape has no bounds, so the pin stops rendering — indistinguishable from deletion), and moving the shape to another page leaves the thread's \`pageId\` behind on the old page.

This adds \`registerCommentAnchorLifecycle\` (registered by \`CanvasComments\` on mount, exported for parts-built consumers), which keeps \`shape\` and \`text-range\` anchored threads consistent through their shape's lifecycle:

- **Deleting a shape** converts its threads to \`point\` anchors at the spot each pin last occupied, so the conversation outlives the shape.
- **Moving a shape to another page** re-homes its threads — \`pageId\` on the thread and its comments' denormalized \`pageId\` — to the new page, with the anchor still riding the shape. Threads anchored to children of a moved frame follow too.
- **A restored shape re-attaches its threads**: undoing a delete (or the re-create half of a page move) restores the original shape anchor, unless the pin was manually moved after the conversion — a deliberate placement wins.
- **Undo/redo of a page move** re-homes threads through the cross-page reparent path.

### Why four cooperating handlers

\`Editor.moveShapesToPage\` is delete + re-create with preserved ids, and each store write is its own operation — so "this shape is moving" never exists as a single observable event. On top of that, the squashed history diff collapses that remove+add into an *update*, so undoing a move replays as a \`parentId\` change. The same user action wears three different guises, and each needs its own handler: \`beforeDelete\` captures pin positions while bounds still resolve (the store runs all before-delete callbacks before removing anything, so this works for children of a deleted frame), the operation-complete pass converts threads whose shape is really gone (remembering the original anchor), \`afterCreate\` re-attaches when the shape id comes back, and \`afterChange\` re-homes on cross-page reparents.

Remote changes are ignored: the client performing the operation runs the same maintenance and syncs the resulting thread updates. Known gap: a shape deleted by a client *not* running this maintenance still strands its threads invisibly.

Writes go through \`commitCommentMutation\`, so the consumer's history option governs undoability as elsewhere in the package.

### API changes

- Added `registerCommentAnchorLifecycle(editor, impreciseShapeAnchor?)`, which keeps shape-anchored comment threads consistent through their shape's deletion, restoration, and page moves, and returns a disposer. `CanvasComments` registers it on mount; only parts-built consumers call it directly.

### Change type

- [x] \`improvement\`

### Test plan

1. Create a shape, comment on it, delete the shape → the pin stays put as a point-anchored thread; undo → the thread re-attaches to the shape.
2. Comment on a shape (and on a shape inside a frame), context menu → Move to page → the threads appear on the destination page riding their shapes; undo → they come back.

- [x] Unit tests

### Release notes

- Comment threads now survive shape deletion (converting to a point anchor at the pin's last position) and follow their shape when it moves to another page.

### DIFF
--- a/packages/commenting/src/canvas/anchor-lifecycle.ts
+++ b/packages/commenting/src/canvas/anchor-lifecycle.ts
@@ -120,7 +120,11 @@ export function registerCommentAnchorLifecycle(
 	// page yet — and the conversion memory must outlive any such partial state.
 	const returnedShapeIds = new Set<TLShapeId>()
 
-	const disposeOperationComplete = editor.sideEffects.registerOperationCompleteHandler(() => {
+	const disposeOperationComplete = editor.sideEffects.registerOperationCompleteHandler((source) => {
+		// Only user-sourced handlers populate the maps and only local operations should settle
+		// them; a remote operation completing leaves any (impossible-in-practice) leftovers for
+		// the next local settle rather than judging them against remote state.
+		if (source === 'remote') return
 		if (pendingByShape.size === 0 && returnedShapeIds.size === 0) return
 		const settled = [...pendingByShape.entries()]
 		pendingByShape.clear()
--- a/packages/commenting/src/canvas/anchor-lifecycle.ts
+++ b/packages/commenting/src/canvas/anchor-lifecycle.ts
@@ -1,0 +1,182 @@
+import { Editor, TLCommentAnchor, TLCommentThread, TLPageId, TLShapeId, VecLike } from 'tldraw'
+import {
+	getCommentRecord,
+	getComments,
+	getCommentThreads,
+	putCommentRecords,
+	TLCommentRecord,
+} from './comment-store'
+import { commitCommentMutation } from './state'
+import { anchorPagePoint, DEFAULT_IMPRECISE_SHAPE_ANCHOR } from './thread-state'
+
+type ShapeAnchor = Extract<TLCommentAnchor, { type: 'shape' | 'text-range' }>
+
+function isAnchoredToShape(
+	thread: TLCommentThread,
+	shapeId: TLShapeId
+): thread is TLCommentThread & { anchor: ShapeAnchor } {
+	const anchor = thread.anchor
+	return (anchor.type === 'shape' || anchor.type === 'text-range') && anchor.shapeId === shapeId
+}
+
+/**
+ * Keep shape-anchored threads (`shape` and `text-range` anchors) alive across their shape's
+ * lifecycle:
+ *
+ * - When the shape is deleted, the thread converts to a `point` anchor at the spot its pin last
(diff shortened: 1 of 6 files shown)


packages/commenting/src/canvas/anchor-lifecycle.ts
```diff
@@ -0,0 +1,215 @@
...
+ * undoable.
+ *
+ * Registered by `CanvasComments` on mount; parts-built consumers can call this directly.
+ * Returns a cleanup function that unregisters all handlers.
+ *
+ * @public
+ */
+export function registerCommentAnchorLifecycle(
+	editor: Editor,
+	impreciseShapeAnchor: { x: number; y: number } = DEFAULT_IMPRECISE_SHAPE_ANCHOR
```
MitjaBezensek: Should we use `getCommentingOptions(editor).impreciseShapeAnchor` as default?

packages/commenting/src/canvas/anchor-lifecycle.ts
```diff
@@ -0,0 +1,215 @@
...
+		const settled = [...pendingByShape.entries()]
+		pendingByShape.clear()
+		const returned = [...returnedShapeIds]
+		returnedShapeIds.clear()
+
+		const updates: TLCommentRecord[] = []
+		for (const [shapeId, threadPoints] of settled) {
+			if (editor.getShape(shapeId)) continue
+			for (const [threadId, point] of threadPoints) {
+				if (!point) continue
```
MitjaBezensek, line 139: Should we still use some default position in this case? Otherwise the thread would disappear?

Not sure when this would happen though.

packages/commenting/src/canvas/anchor-lifecycle.ts
```diff
@@ -0,0 +1,215 @@
...
+	)
+
+	// Shape ids from `convertedByShape` re-created during the current operation. Restores are
+	// settled at operation complete rather than in `afterCreate` itself: creation order within an
+	// operation is arbitrary, so a shape can appear before its parent and not resolve an ancestor
+	// page yet — and the conversion memory must outlive any such partial state.
+	const returnedShapeIds = new Set<TLShapeId>()
+
+	const disposeOperationComplete = editor.sideEffects.registerOperationCompleteHandler(() => {
+		if (pendingByShape.size === 0 && returnedShapeIds.size === 0) return
```
MitjaBezensek, line 129: Should we do an early return here for remote changes, just like we do in some other handlers? Think we are only interested in the local ones here?

MitjaBezensek (approved):

PR #9662, opened 2026-07-22
feature(commenting): shape-comment precision via a shouldBePrecise predicate

jsscclr:
Comparison PR to #9651, per review discussion there: the same capability expressed as a predicate instead of an enum. Not meant to both land — pick one.

### What

- Adds `shouldBePrecise(editor, context)` to `CommentingOptions`, set via `CommentTool.configure`. The context carries the gesture — target `shapeId`, release `point`, and `altKey` — so the predicate can be:
  - the default: `() => true` — every shape comment is precise
  - a constant `() => false` — shape-only anchoring
  - the previous Alt behaviour: `(_editor, { altKey }) => altKey`
  - a per-shape policy: `(editor, { shapeId }) => editor.getShape(shapeId)?.type === 'note'`
- Runs at both shape-anchor creation sites: comment-tool placement and pin-drag re-anchor.
- Same data-model note as #9651: clicked `x`/`y` are stored either way; the predicate only gates `isPrecise`.
- **Behaviour change**: precise is now the default. Consumers relying on the old Alt-gated placement opt back in with the alt predicate. dotcom needs no change — it inherits the default.

### Example

Same `comment-shape-precision` example as #9651 (diff the two branches to compare), plus a fourth mode only the predicate can express: notes precise, everything else shape-level.

### One caveat found while building this

The motivating case from review — "precise for frames" — can't fire today: the comment tool's hit test uses `hitInside` without `hitFrameInside`, so releasing over an empty frame interior creates a *point* anchor, never a frame anchor (verified live). Whether frame interiors should be comment targets is a separate placement-behaviour decision; the example uses notes instead. Worth deciding alongside whichever API wins.

### Test plan

- Verified live in the example: always / never / alt modes create the expected `isPrecise`, and in notes mode the same no-modifier gesture yields `isPrecise: false` on a geo and `isPrecise: true` on a note
- `packages/commenting` tests (179), typecheck, api-report regenerated

### API Changes

- New option on `CommentingOptions`: `shouldBePrecise(editor: Editor, context: ShapeCommentPrecisionContext): boolean`, defaulting to `() => true`. This changes default placement behaviour: shape comments are precise unless configured otherwise (the previous behaviour is `(_editor, { altKey }) => altKey`).
- New export: `ShapeCommentPrecisionContext` — the gesture context (`shapeId`, `point`, `altKey`).

### DIFF
--- a/apps/examples/src/examples/collaboration/comment-shape-precision/CommentShapePrecisionExample.tsx
+++ b/apps/examples/src/examples/collaboration/comment-shape-precision/CommentShapePrecisionExample.tsx
@@ -1,0 +1,134 @@
+import {
+	CanvasComments,
+	CommentAuthor,
+	CommentTool,
+	commentToolOverrides,
+} from '@tldraw/commenting'
+import { getLicenseKey } from '@tldraw/dotcom-shared'
+import { useMemo, useState } from 'react'
+import {
+	commentSchemaRecords,
+	createTLSchema,
+	createTLStore,
+	Editor,
+	TLComponents,
+	Tldraw,
+	TldrawUiButton,
+	TldrawUiButtonLabel,
+	toRichText,
+} from 'tldraw'
+import '@tldraw/commenting/commenting.css'
+import 'tldraw/tldraw.css'
+
+// One configured comment tool per mode, built once at module level so each array keeps a stable
+// identity. `shouldBePrecise` decides what commenting on a shape produces: a precise anchor
+// (pinned to the exact clicked spot within the shape — the default) or an imprecise one (pinned
+// to the shape as a whole, rendered at its top-right). It's called with the target shape, the
+// release point, and the Alt key's state — so it can be a constant, an Alt-gated choice, or a
+// decision from the shape itself, like "precise only on notes".
+const MODE_TOOLS = {
+	always: [CommentTool],
+	never: [CommentTool.configure({ shouldBePrecise: () => false })],
+	alt: [CommentTool.configure({ shouldBePrecise: (_editor, { altKey }) => altKey })],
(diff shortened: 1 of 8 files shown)


mimecuvalo (commented):
this is a much better call / direction! (thx @MitjaBezensek!)

it's prbly worth a discussion tomorrow on the right default here. my take would be that, in a whiteboarding app the slider is closer to the right: `precision <——————|—> intent` (if those 2 can be an axis hah) i guess, what i mean is in Figma i'd desire precision but for a casual whiteboarding experience i'd maybe want it to just snap to the shape i'm intending to comment on (and hopefully out of the way of any text within the shape).

jsscclr:
> it's prbly worth a discussion tomorrow on the right default here. my take would be that, in a whiteboarding app the slider is closer to the right: precision <——————|—> intent (if those 2 can be an axis hah) i guess, what i mean is in Figma i'd desire precision but for a casual whiteboarding experience i'd maybe want it to just snap to the shape i'm intending to comment on (and hopefully out of the way of any text within the shape).

I'm making it default to precise for the moment, with the caveat that the shape-level snapping probably does make sense in many cases, but I feel like more care needs to be taken around positioning of the snapping + clustering behaviour. Can get @meg-an31 to deep dive into what sensible per shape could look like (e.g. rectangle vs ellipse vs notes, etc)

MitjaBezensek (approved):
Nice 👍

jsscclr:
Merging but feel free to add any thoughts @mimecuvalo!

mimecuvalo:
> Merging but feel free to add any thoughts @mimecuvalo!

all good 👍 thx :)

PR #9666, opened 2026-07-23
fix(dotcom): asset diagnostics subrequest budget and asset table indexes

MitjaBezensek:
Follow-up to #9646. Running the asset diagnostics report against a file with ~5.8k assets never returned a result. Two causes:

## Worker subrequest budget

A worker invocation gets ~1000 subrequests and every R2 \`head()\` costs one. The report tried to head every asset, so large files burned the budget partway through; every subsequent subrequest threw, and the \`retry\` wrapper slept 500ms before re-throwing each one — adding minutes of pure sleep before the Postgres cross-check queries failed too.

- Cap head checks at 500 per report, spending the budget on unassociated assets (the diagnostic targets) first. Skipped assets report as \`notChecked\` instead of counting as head failures.
- Stop retrying "Too many subrequests" errors — the invocation can't make progress, so failing fast beats sleeping 500ms per asset.
- Cap the \`problems\` list (200 entries, with the true total in \`problemsTotal\`) and \`warnings\` (20) so the response for a multi-thousand-asset file isn't megabytes.

## Missing indexes on the asset table

\`asset\` only has a primary key on \`objectName\`; the \`fileId\` and \`userId\` foreign keys are unindexed, so lookups by file or user seq-scan the whole table. \`EXPLAIN ANALYZE\` on production for the diagnostics query: parallel seq scan filtering 7.9M rows, 441ms warm. The same scan runs on the tldr download path, admin hard delete, user deletion, and every cascading FK check when file or user rows are deleted.

Migration \`040_asset_indexes.sql\` adds indexes on both columns. It uses \`IF NOT EXISTS\` so the indexes can be pre-built with \`CREATE INDEX CONCURRENTLY\` outside the migration transaction (migrations run inside one, where \`CONCURRENTLY\` is not allowed), making the locking migration a no-op at deploy time.

## Admin UI

Shows the new "not checked" count and labels problem rows as not checked vs check failed; notes when the problems table is truncated.

### DIFF
--- a/apps/dotcom/client/src/pages/admin.tsx
+++ b/apps/dotcom/client/src/pages/admin.tsx
@@ -867,6 +867,7 @@ function AssetDiagnostics() {
 								['Pending association', report.assets.pending],
 								['Missing in bucket', report.assets.missingInBucket],
 								['Head check failures', report.assets.headFailures],
+								['Not checked (over budget)', report.assets.notChecked],
 								['Old-format URLs', report.assets.oldFormatUrls],
 								[
 									'DB asset rows',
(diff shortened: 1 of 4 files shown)


MitjaBezensek:
Not sure if needed.

PR #9667, opened 2026-07-23
feat(dotcom): render board thumbnails ahead of the first share

driev:
Boards unfurl with the generic tldraw image the first time their link is shared. Rendering on crawler demand cannot fix that: unfurl platforms resolve a card once and cache it for days, so a render triggered by the first crawler lands long after that crawler has taken the default away. This makes the thumbnail exist *before* the first share.

The same primitive is what a folder or workspace view needs to show board previews, so the render path covers every board rather than only the publicly viewable ones.

Relates to #9502 and #9497. The design notes — the reasoning behind each decision, the approaches rejected, and the ops setup — are in `apps/dotcom/browser-run-thumbnails.md`.

### What ships

**Render before the share, not on demand.** Two triggers enqueue onto the existing queue and consumer: the `publish` effect in `TLPostgresReplicator`, right after `publishSnapshot` writes the frozen R2 snapshot; and `TLFileDurableObject` when a persist advances the document clock. The second is **debounced, not throttled** — each persist pushes the deadline out by `OG_RENDER_DEBOUNCE_MS` (30s), so a board renders once its editing settles rather than on a cadence while it is still being drawn on, with `OG_RENDER_MAX_WAIT_MS` (5 min) stopping a board that is never left alone from never rendering. The durable object's alarm carries the deadline itself rather than a lower bound on it, so an eviction loses the in-memory copy and nothing else. The arithmetic lives in `utils/ogRenderDebounce.ts` so it is testable without standing up a durable object.

**Rendering is no longer gated on public viewability — serving is.** A thumbnail is generated for every board, private ones included, so an owner-facing surface always has one to show. That is expressed as an explicit `ThumbnailBoardAccess` required at every call site: `public` for anything anonymous-facing, `render` for the generation path, which asks only that the board exists and has content. The gate is re-applied per request when serving, which it has to be, since nothing deletes a thumbnail when a board stops being public.

**A render token is now two factors, not one.** This follows directly from the above: the snapshot route serves a *private* board's full document, so a leaked `MCP_SCREENSHOT_TOKEN_SECRET` would have been enough to read any board. Every mint of a `render` job also records the token's hash in R2 and the route requires that record, so forged signatures are refused without write access to our bucket. The TTL drops from 5 minutes to 60s, sized against a capture's measured p90 rather than generously.

The access level rides inside the signed job, taken from the resolution rather than the caller, so a surface cannot render under a gate it did not resolve under. The MCP tool mints `public` and is deliberately **not** recorded: it only ever renders boards anyone could already fetch, so the record would guard nothing — and recording it would put both surfaces in one per-board key space, where each mint invalidates the other's in-flight token. Authenticating the MCP endpoints would invite minting `render` there, so the key needs namespacing by surface before that happens; that is noted where it is keyed.

**The image route is a pure read.** It answers two questions — is this board publicly viewable, and does a thumbnail exist — and `302`s to the site-wide default otherwise. The on-miss enqueue is gone (it was work whose result nobody came back for), and with it the route's per-board limit and minimum refresh age. The worker does not proxy the default's bytes: it is a static asset already cached at the edge, and on a URL that deliberately carries no `s-maxage` that would put worker egress in front of every unfurl of an unrendered board.

**Rate limiting is confined to the MCP endpoint**, the one Browser Run-spending surface an outside caller can drive directly. Per-IP, per-board and global budgets now hold different numbers, which needs three bindings rather than two — a binding carries a single `limit` applied per key. The per-board and global bindings are named `MCP_SERVER_*` rather than for screenshots, since #9774 puts cluster tools on the same budgets. Both caller-scoped budgets are temporary — authentication on the MCP server replaces them with a single per-user limiter — so the per-IP binding keeps its old name rather than churning live state on its way out, while the per-board binding, which is introduced here and has never been deployed, ships under the better one. Every `namespace_id` is unchanged, so existing counters carry across the deploy. Thumbnail rendering is deliberately uncapped: see below.

**MCP screenshots move to their own bucket**, `dotcom-mcp-data`, behind an `MCP_DATA_BUCKET` binding. `og/…` keys carry no version and overwrite in place, so they must never expire. `mcp/…` keys include the content version and accumulate forever, so they want a lifecycle rule. Separate buckets rather than a prefix-scoped rule, because a future rule added without a prefix — or with a typo'd one — would silently delete every board's live thumbnail, and R2 expiration has no undo.

**Spend and failures are measurable, and boards are not identifiable.** A failed capture still created a browser and held it, often for the full timeout, so failures now record their Browser Run duration — telemetry counts spend per delivery, Sentry counts problems per job. Browser Run answers `422` for a crashed page, an OOM render and every timer alike, so failures are classified from the response body rather than the status, which is what makes the timeout rate a real number. Separately, no board identity is written anywhere — not to telemetry, not to a log line, not to a Sentry event — since for a link-shared file the slug *is* the capability to view the board. `persist_success` is the exception that sizes all of this: it carries the durable object id (one-way) and the board's shared state.

### Why a debounce and not a throttle

This started as a 30s leading-and-trailing throttle. Measurement killed it: production runs **~555 persists/min across ~359 boards** active in a 10 minute window — a mean gap between a given board's persists of **~39s**. A throttle whose window is shorter than the gap between events suppresses almost nothing, so its leading edge fired on nearly every persist and render volume tracked persist volume rather than the interval it was nominally set to.

Ten minutes of unbroken editing (76 persists) costs 2 renders under the debounce against roughly 20 under the throttle. For a short burst — which the 39s gap says is the common case — both cost exactly one, but the debounce renders the *finished* burst rather than its first stroke.

### Not yet settled

Two numbers are still assumptions, and the telemetry above is what answers them once this is deployed:

- the link-shared fraction of actively edited boards, assumed ~30%
- `OG_RENDER_MAX_WAIT_MS`, the dominant cost term for busy boards, which nothing has constrained

Both are Analytics Engine queries against `persist_success` — deliberately not database questions, since Postgres knows which files are shared but not which are being edited, and there is no replica to ask.

Worth flagging for review: **no per-board mechanism can cap total spend**, because total is `active boards × per-board rate`. Cloudflare's rate limit binding cannot close that gap either — limits are applied per Cloudflare location, so a single global key gives `limit × locations`. (The same caveat applies to `GLOBAL_BROWSER_RUN_RATE_LIMIT` on the MCP path: it bounds a rogue caller, which is its job, but it is not an account-wide ceiling.) The real ceiling is Browser Run's account limits.

Also worth knowing: a board shared within the debounce window of its first edit still unfurls with the generic card, and platforms cache that for days. Phase 4 in the design notes (warming the image from `getSocialPreview`) is what would close it, and is not started.

### Before deploying

One-time ops setup, with exact commands in the design notes:

1. Create the `dotcom-mcp-data-preview` and `dotcom-mcp-data` R2 buckets.
2. Add a 30 day expiration rule to those two only — **not** to `thumbnails`, whose objects are the live thumbnails and must outlive any lifecycle window.
3. Purge the legacy `mcp/` prefix from `thumbnails`. Those objects are orphaned: the tool now reads and writes the new bucket, and the version in the key means nothing will hit them again.

### Change type

- [x] `feature`

### Test plan

1. Share a board via link, draw on it, and wait for editing to settle. The OG image endpoint should serve the rendered board rather than redirecting to the default.
2. Unfurl that board's link and confirm the card shows the board. Note the preview HTML is only served to user-agents matching the crawler list in `apps/dotcom/client/scripts/build.ts`; anything else gets the SPA shell and the static site-wide card, so test with a matching agent.
3. Publish a board and confirm its image renders without waiting for a crawler.
4. Edit a board that is **not** shared, and confirm a thumbnail is rendered for it but the OG image route still redirects to the default for its URL.
5. Unshare a shared board and confirm the OG route falls back to the default while the cached image is kept, so resharing is an immediate hit.
6. Unpublish a published board and confirm its published image is deleted and its file-keyed image is not.
7. Screenshot two different pages of one board through the MCP tool concurrently, and confirm both succeed.

- [x] Unit tests
- [ ] End to end tests

Unit coverage added or extended for the fallback redirect and cache decisions (`getOgImage.test.ts`), the queue consumer's resolve/version/gate behaviour and its failure classification (`ogImageQueue.test.ts`), the shared error and reporting helpers (`thumbnailShared.test.ts`), the minted-token check and the access split (`renderTokens.test.ts`, `getThumbnailSnapshot.test.ts`), the render/serve access split (`getSharedFile.test.ts`), the `unshare` effect (`ChangeCollator.test.ts`), and the render debounce including its cost and durable-write behaviour (`ogRenderDebounce.test.ts`).

### Code changes

| Core code      | +5 / -0      |
| Tests          | +1611 / -171 |
| Apps           | +1873 / -544 |
| Config/tooling | +9 / -5      |

`apps/dotcom/browser-run-thumbnails.md`, the design notes, is +443 / -21 of the Apps row.

### Release notes

- tldraw.com boards now show a real preview image the first time their link is shared, instead of the generic tldraw image.


apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
```diff
@@ -47,64 +44,41 @@ export async function getOgImage(
...
+			maxAgeSeconds: isCurrent ? FRESH_IMAGE_MAX_AGE_SECONDS : STALE_IMAGE_MAX_AGE_SECONDS,
 			version: cached.customMetadata?.version,
 		})
 	}
 
+	// Never rendered, so the board has nothing of its own to show and the request is sent to the
+	// site-wide default instead. The short max-age is what lets the real image take over as soon as a
+	// publish or an edit lands it.
 	writeScreenshotTelemetry(env, {
 		source: 'og',
```
driev, line 95: should this write the DO ID?

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
```diff
@@ -1,65 +1,71 @@
...
+// first page when none have content, which renders as a blank.
 function pickOgImagePageId(snapshot: RoomSnapshot): string | undefined {
 	const pages = enumerateBoardPages(snapshot)
 	if (pages.length === 0) return undefined
 	return (pages.find((page) => page.hasContent) ?? pages[0]).id
 }
 
-export function getOgImageCacheKey(board: Pick<ResolvedThumbnailBoard, 'kind' | 'slug'>) {
+export function getOgImageCacheKey(board: ThumbnailBoardRef) {
 	return `og/${board.kind}/${board.slug}/${DEFAULT_THUMBNAIL_WIDTH}x${DEFAULT_THUMBNAIL_HEIGHT}/light.png`
```
driev: Remove dimensions from the key.

apps/dotcom/sync-worker/wrangler.toml
```diff
@@ -352,6 +355,27 @@ bucket_name = "thumbnails-preview"
 binding = "THUMBNAILS"
 bucket_name = "thumbnails"
 
+#################### MCP screenshots R2 bucket ####################
+# cache for MCP tool screenshots (`mcp/…` keys). Separate from THUMBNAILS because the retention is the
+# opposite: this key includes the board's content version, so every edit strands the previous object.
+# These buckets carry an expiration lifecycle rule (see the ops setup in browser-run-thumbnails.md).
+[[env.dev.r2_buckets]]
+binding = "MCP_SCREENSHOTS"
+bucket_name = "mcp-screenshots-preview"
```
driev: Make this generic - `dotcom-mcp-data` or something.

apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
```diff
@@ -4,40 +4,45 @@ import { getPublicOrigin } from '../../utils/getPublicOrigin'
...
-// Stale images and redirects use short TTLs so scrapers and browsers come back for the fresh
-// render soon after the queued job completes.
+// Short, and deliberately not chosen for cost or freshness. Neither is the binding constraint here:
+// R2 reads are a rounding error next to rendering, and unfurl platforms cache a card their own side
+// for days whatever we say. What the lifetime actually decides is how long an image can be served
+// without the share gate being consulted — nothing deletes a board's image when it stops being
+// public, so this route re-checking the gate per request is the only thing keeping an unshared
+// board's thumbnail off the internet. Every second of cache lifetime is a second that check does not
+// happen, so these are minutes, and revalidation is made cheap with an etag instead.
+const FRESH_IMAGE_MAX_AGE_SECONDS = 5 * 60
```
mimecuvalo: should this go to the config.ts file?

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
```diff
@@ -1,65 +1,98 @@
...
+// limit: the consumer clears it as soon as a render lands, so on a healthy board it never reaches
+// this TTL, which exists only so a consumer dying cannot wedge a board permanently. To change how
+// often a board renders, change OG_RENDER_DEBOUNCE_MS.
+//
+// Must outlive a job's worst-case retry chain — every capture running to its timeout plus every
+// backoff delay — or the marker lapses while the job is still alive, a fresh ask enqueues a second
+// job for the same board, and the two clobber each other's per-board render token record
+// (renderTokens.ts is keyed on the single-flight this marker provides). A test pins the inequality
+// against the constants below.
+export const PENDING_MARKER_TTL_MS = 5 * 60_000
```
mimecuvalo: should these go to the config.ts file?

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
```diff
@@ -30,12 +29,68 @@ const SCREENSHOT_TOOL_NAME = 'get_shared_board_screenshot'
...
+// in browser-run-thumbnails.md.
+//
+// These constants are only the isolate-local fallback for local dev and tests. Deployed environments
+// are governed by the Cloudflare rate limit bindings in wrangler.toml, so changing a number here alone
+// changes nothing in production. Each budget has its own binding, and the pairs must move together:
+//
+//   PER_IP_RATE_LIMIT              ->  MCP_SCREENSHOT_RATE_LIMITER      (limit = 10)
+//   PER_BOARD_RATE_LIMIT           ->  MCP_SERVER_BOARD_RATE_LIMITER    (limit = 2)
+//   GLOBAL_BROWSER_RUN_RATE_LIMIT  ->  MCP_SERVER_BROWSER_RATE_LIMITER  (limit = 20)
+const PER_IP_RATE_LIMIT = 10
```
mimecuvalo: should this go to the config.ts file?

apps/dotcom/sync-worker/src/utils/renderTokens.ts
```diff
@@ -41,7 +61,17 @@ export interface ThumbnailRenderJob {
...
+ * How long a minted render token stays valid.
+ *
+ * Sized against what a capture needs rather than generously, because this token is what stands between
+ * an HMAC signature and a *private* board's full document: thumbnails are rendered for every board, not
+ * only public ones. Measured renders run 4s at p50 and 12-17s at p90, so 60s is several times p90.
+ *
+ * Queue linger and retry backoff do not eat into it — a token is minted immediately before the
+ * `quickAction` call, not at enqueue. A render slower than this fails and retries with a fresh token.
+ */
+export const THUMBNAIL_RENDER_TOKEN_TTL_MS = 60_000
```
mimecuvalo: should this go to the config.ts file?

apps/dotcom/sync-worker/src/config.ts
```diff
@@ -4,6 +4,43 @@
...
+ * costs thumbnail freshness. See "Why a debounce and not a throttle" in
+ * apps/dotcom/browser-run-thumbnails.md for the measurements and the cost table.
+ *
+ * 60s rather than 30s because replaying both values over six hours of production `persist_success`
+ * timings costs ~35.7 renders/min at 30s against ~32.3 at 60s. Note how small that is: raising this
+ * alone cannot save much, because a merged session then runs long enough to hit
+ * `OG_RENDER_MAX_WAIT_MS` and trades a settle render for a max-wait one. The two constants have to
+ * move together to matter — see "What it costs" in the doc for the grid.
+ */
+export const OG_RENDER_DEBOUNCE_MS = 60_000
```
mimecuvalo, line 31: a lot of the config values seem to be 60_000 — is there an interdependency of these values being around the same value of 60s? if you consolidate the const values to this file it feels like there will be a lot that just say 60s over and over

apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
```diff
@@ -128,34 +128,110 @@ function parseSlug(value: unknown) {
...
+ * So this fires only for `published`, and only on a total miss: no image at all, not a stale one.
+ * The pending marker dedupes it, and a board that fails to render permanently is limited to one
+ * attempt per marker TTL rather than one per crawler.
+ *
+ * This does not undo the reasoning that removed the general on-miss enqueue. That enqueue was
+ * pointless because the crawler which triggered it has already cached the default by the time the
+ * render lands, and unfurl platforms resolve a card once — both still true. The point here is not to
+ * serve *this* request; it is that without it, nothing else will ever ask.
+ */
+async function repairMissingPublishedImage(
```
mimecuvalo, line 147: 1. repairMissingPublishedImage is an anonymous, unlimited Browser Run spend path

getOgImage.ts:149 re-adds an on-miss enqueue for published boards. The PR frames rate limiting as "confined to the MCP endpoint, the one Browser Run-spending surface an outside caller can drive directly" — that's no longer accurate.

The pending marker bounds it while a job is alive, but retryOrDrop (ogImageQueue.ts:413) clears the marker when it gives up. So for a published board that reliably fails to render (a large board that times out at 45s, say), the cycle is: any unauthenticated request → 3 deliveries × up to 45s of Browser Run + 90s of backoff ≈ 3.75 min → marker cleared → next request re-arms it. That's roughly 45 captures/hour/board, driven by traffic an attacker controls, with no limiter anywhere on the path.

mimecuvalo (approved):
pre-approved modulo couple comments 🎉

apps/dotcom/sync-worker/wrangler.toml
driev: Done in c0c81f5967: the bucket is now `dotcom-mcp-data` / `dotcom-mcp-data-preview` behind an `MCP_DATA_BUCKET` binding, and the per-board and global rate limiters were renamed to `MCP_SERVER_*` in the same commit.

apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
driev: Moved in 3667c3c957 — the three cache lifetimes now live in config.ts as `OG_FRESH_IMAGE_MAX_AGE_SECONDS` / `OG_STALE_IMAGE_MAX_AGE_SECONDS` / `OG_FALLBACK_MAX_AGE_SECONDS`, with the share-gate rationale comment alongside them.

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
driev: Moved in 3667c3c957 — now `OG_PENDING_MARKER_TTL_MS`, `OG_MAX_RENDER_ATTEMPTS` and `OG_RETRY_DELAY_SECONDS` in config.ts, next to the debounce constants they interact with. The comment about the marker TTL having to outlive the retry chain moved with them, and the test still pins that inequality.

apps/dotcom/sync-worker/src/routes/tla/sharedBoardScreenshotMcp.ts
driev: Moved in 3667c3c957 — the fallback budgets are now `MCP_PER_IP_RATE_LIMIT`, `MCP_PER_BOARD_RATE_LIMIT`, `MCP_GLOBAL_BROWSER_RUN_RATE_LIMIT` and `MCP_RATE_LIMIT_WINDOW_MS` in config.ts, with the mapping table to their wrangler.toml bindings. The limiters are still applied here so a new surface on the shared render core can't pick one up by accident.

apps/dotcom/sync-worker/src/utils/renderTokens.ts
driev: Moved in 3667c3c957 — `THUMBNAIL_RENDER_TOKEN_TTL_MS` now lives in config.ts with its sizing rationale (several times the measured p90 capture time).

apps/dotcom/sync-worker/src/config.ts
driev: Good question — no, there's no interdependency between the 60s values. Each is independently derived: the debounce comes from replaying production persist-gap timings, the token TTL is sized against capture p90 (12–17s), and the rate limit window matches the fixed 60s period on Cloudflare's rate limit bindings. Any of them could move without the others.

The one genuine coupling in the pipeline is `OG_PENDING_MARKER_TTL_MS` having to outlive the worst-case retry chain (`OG_MAX_RENDER_ATTEMPTS` × capture timeout + backoff), which a test pins. With the constants now consolidated in config.ts (3667c3c957), the file header says exactly this — that the repeated 60s is coincidence, not coupling — and each constant keeps its own derivation comment so none of them reads as an unexplained "60s".

apps/dotcom/sync-worker/src/routes/tla/getOgImage.ts
driev: You're right — the marker-clear on give-up handed control of the retry cadence to crawler traffic, and the PR's "rate limiting is confined to MCP" framing didn't hold for this path. Fixed in 3667c3c957 with a per-board repair cooldown:

- When a **crawler-triggered** job for a published board burns its whole retry budget, the give-up arms a `.repair-cooldown` marker (`OG_REPAIR_COOLDOWN_MS`, 1 hour) next to the pending marker, and `repairMissingPublishedImage` checks it before enqueueing. Your failing-board scenario goes from ~45 captures/hour/board of attacker-drivable spend to one 3-delivery chain per hour.
- Only the crawler-reason give-up arms it, so a publish-triggered render that fails transiently still gets one immediate repair on the next crawl — the case the repair exists for — and the cooldown only meters attempts after that repair has itself failed.
- `enqueueOgImageRender` doesn't consult it, so a genuine republish renders straight away regardless; unpublish and hard delete remove the marker along with the image, so nothing orphans.

The "Request limits" section and the repair paragraph in browser-run-thumbnails.md now describe this bound instead of the marker-TTL claim you called out.

apps/dotcom/sync-worker/src/utils/renderTokens.ts
driev: Yep, moved them all.

apps/dotcom/sync-worker/src/config.ts
driev: They're all independent values and might need to change - let's see.

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
```diff
@@ -73,52 +119,143 @@ export async function enqueueOgImageRender(
...
+
+// Deletes the image as well as the marker and the repair cooldown. Only for `published` boards
+// losing their publication. Scoped by the `board` passed in, so calling it with `kind: 'published'`
+// cannot touch the file-keyed image of the same board.
+export async function deleteOgImage(env: Environment, board: ThumbnailBoardRef): Promise<void> {
+	if (!env.THUMBNAILS) return
+	await Promise.all([
+		env.THUMBNAILS.delete(getOgImageCacheKey(board)).catch(() => {}),
+		env.THUMBNAILS.delete(getOgImagePendingKey(board)).catch(() => {}),
+		env.THUMBNAILS.delete(getOgImageRepairCooldownKey(board)).catch(() => {}),
```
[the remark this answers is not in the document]
driev: Fixed in 55609b04a5. `enqueuePublishThumbnailRender` now deletes the repair cooldown before enqueueing — the cooldown is evidence about the snapshot that failed, and an in-place republish replaces that snapshot, so the new one gets its own repair backstop even if the publish-time render (or the enqueue itself) fails. A republished board that still cannot render re-arms the cooldown on its next crawler give-up. Covered by a test that arms the cooldown via a crawler give-up and asserts the republish clears it.

apps/dotcom/sync-worker/src/routes/tla/ogImageQueue.ts
```diff
@@ -73,52 +122,145 @@ export async function enqueueOgImageRender(
...
+	env: Environment,
+	publishedSlug: string,
+	reportProblem: (error: unknown) => void
+): Promise<void> {
+	const board: ThumbnailBoardRef = { kind: 'published', slug: publishedSlug }
+	// An in-place republish reuses the slug, so a cooldown armed against the previous snapshot's
+	// failure would otherwise outlive the snapshot it was evidence about and block the new one's
+	// repair. Cleared unconditionally and first: even if the enqueue below fails, the next crawl's
+	// repair is exactly the backstop that failure needs.
+	await env.THUMBNAILS?.delete(getOgImageRepairCooldownKey(board)).catch(() => {})
```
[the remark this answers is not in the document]
driev: False positive: optional chaining short-circuits the whole chain, not just the next access. In `env.THUMBNAILS?.delete(...).catch(...)`, an undefined `THUMBNAILS` skips both the `.delete` call and the `.catch`, and the expression evaluates to `undefined` — which awaits harmlessly. A TypeError would need the chain broken with parentheses, `(env.THUMBNAILS?.delete(...)).catch(...)`, which this is not. Verified directly: `await ({}).THUMBNAILS?.delete('k').catch(() => {})` runs without throwing. `retryOrDrop` has used the identical `THUMBNAILS?.put(...).catch(...)` shape since 3667c3c957.

PR #9584, opened 2026-07-14
feat(commenting): add holistic options via CommentTool.configure()

mimecuvalo:
## What

Adds a single, imperative configuration surface for the commenting layer, mirroring `ShapeUtil.configure()`. `CommentTool` gains a static `configure()` that returns a configured subclass carrying merged `options`. The tool reads `this.options`; React components read the same off the tool node via `editor.getStateDescendant('comment')` — no provider, no context.

```tsx
<Tldraw tools={[CommentTool.configure({ history: 'ignore', enableDelete: false })]} />
```

## Why

Configuration was scattered across piecemeal render props, hardcoded constants, a hardwired `history: 'ignore'`, and global module-level atoms. This consolidates it into one `CommentingOptions` object (modeled on `TldrawOptions`) delivered through the tool registration consumers already do — and the same effort fixes a latent multi-editor bug.

## What's configurable

- **Undo/redo** — `history` (default `'ignore'`, matching the `TLComment` schema warning), plus a separate `dragHistory` for pin re-anchors. All six mutation sites now route through a `runComment` helper.
- **Feature toggles** — resolve, delete, edit, drag-to-move, clustering, deep links.
- **Tuning** — clustering/animation constants, drag threshold, imprecise-anchor spot.
- **Shortcuts** — configurable visibility toggle (default `shift+c`) and Escape-to-close.
- **Sidebar** — tool ids that reveal it, default filters, initial visibility.
- **Component slots** — `CommentBody` / `PinContent` / `ThreadPreview`, generalizing the now-`@deprecated` `renderBody` / `renderPinContent` / `renderPreview` render props (the props still win when passed).

## Multi-editor fix

The four transient UI atoms (`openThreadId`, `pendingComment`, `commentsHidden`, `sidebarFilters`) move from global module singletons to per-editor `EditorAtom`s, so two editors on one page no longer share open-thread, visibility, and filter state.

## Notes

- Defaults preserve current behavior exactly; the dotcom adapter needed no changes.
- `@tldraw/commenting` is `private`, so the module-atom API was migrated cleanly (no deprecation shims) with in-repo consumers updated.

## Testing

- 13 new unit tests: `configure()` merge + chaining, `runComment` history threading (mutation vs drag), `getCommentingOptions` round-trip, per-editor state isolation, `matchesKbd`. Full package suite: 161 passing.
- `yarn typecheck` green across the monorepo (including the dotcom consumer); lint + format clean.

PR #9674, opened 2026-07-23
refactor(commenting): fold region options into CommentTool.configure

jsscclr:
In order to give the commenting layer a single configuration channel, this PR moves the region options from the `regionOptions` prop on `<CanvasComments>` into `CommentingOptions`, configured through `CommentTool.configure` like every other commenting option. Region commenting (#9582) landed in parallel with the `configure` surface (#9584), so its config predates the channel it now joins — and #9662 already put `shouldBePrecise` on the `configure` side.

The options flatten into `CommentingOptions`' existing shape (mirroring the flat cluster-tuning fields, and `TldrawOptions` generally): `enableRegions` joins the feature toggles, and `regionPinCorner` / `regionReveal` / `regionMove` / `regionResize` form a region section. Flat fields mean a partial call like `CommentTool.configure({ enableRegions: true })` type-checks through the existing `Partial<Options>` signature, with no extra merge machinery.

Because options are fixed at tool registration, the per-editor publish bridge (`region-options.ts` — a `WeakMap` of atoms the overlay wrote and the tool read) is deleted outright, along with the reactivity it existed to provide: every reader is now a plain `getCommentingOptions(editor)` read. `<CanvasComments>` gathers the five fields internally into the same object the region components always took, so the render path is unchanged.

```tsx
// before
<Tldraw tools={commentTools} components={components} />
// … <CanvasComments regionOptions={{ enabled: true, pinCorner: { x: 0, y: 0 } }} />

// after
<Tldraw
	tools={[CommentTool.configure({ enableRegions: true, regionPinCorner: { x: 0, y: 0 } })]}
	components={components}
/>
// … <CanvasComments />
```

The trade: region config can no longer change under a running editor — reconfiguring means remounting with a newly configured tool, the same contract as every other commenting option. (The component-studio sketchbook mounts one editor per variant, so it needed no such trick.)

A new `comment-regions` example demos the full region surface: a control panel drives all five options against a seeded region, remounting the editor with a newly configured tool on each change while a shared store carries the comments across — the same pattern as the shape-precision example.

### Change type

- [x] `api`

### Test plan

Preview: https://pr-9674-preview-deploy.tldraw.com/ — a regression check only, since dotcom doesn't enable regions.

1. In the examples app, open the commenting example: with the comment tool, drag out an area — a dashed region box draws and the composer opens at its bottom-right corner; a plain click still places a point comment.
2. In the examples app, open the new [region comments example](https://examples-git-jessica-commenting-region-configure.tldraw.club/comment-regions) (deployed from this branch): flip each control (regions on/off, reveal, move, resize, pin corner) and check the seeded region picks up the behaviour, with existing threads surviving each switch.
3. In component-studio, open Flows → Region comments: each variant (pin corner, reveal, move, resize) still exercises its dimension via a per-sketch configured tool.
4. dotcom is unchanged — it doesn't enable regions; comments there still place, reply, and resolve as before.

- [x] Unit tests

### Release notes

- Configure region comments through the comment tool — `CommentTool.configure({ enableRegions: true, ... })` — instead of the removed `regionOptions` prop on `<CanvasComments>`.

### API changes

- Added `enableRegions`, `regionPinCorner`, `regionReveal`, `regionMove`, and `regionResize` to `CommentingOptions`, set via `CommentTool.configure`
- Breaking! Removed the `regionOptions` prop from `CanvasCommentsProps` — pass the flat options to `CommentTool.configure` instead
- Breaking! Removed `RegionCommentOptions` and `DEFAULT_REGION_COMMENT_OPTIONS` from the package exports; the defaults now live in `defaultCommentingOptions`

### Code changes

| Core code       | +50 / -78  |
| Tests           | +5 / -3    |
| Automated files | +16 / -13  |
| Documentation   | +295 / -7  |
| Apps            | +27 / -29  |


packages/commenting/src/canvas/options.ts
```diff
@@ -109,6 +116,18 @@ export interface CommentingOptions {
 	 */
 	shouldBePrecise(editor: Editor, context: ShapeCommentPrecisionContext): boolean
 
+	// ── Region comments ───────────────────────────────────────────────────────────────────────
+	/** Which corner of a region its pin and composer sit on, as a normalized 0–1 offset within the
+	 *  region. Default bottom-right. */
+	readonly regionPinCorner: { readonly x: number; readonly y: number }
+	/** When a region's dashed box and resize handles reveal: while the pointer is within the
+	 *  region, while its pin is hovered, or only while its thread is open. */
+	readonly regionReveal: 'pointer' | 'pin-hover' | 'open'
```
mimecuvalo, line 125: let's land this PR but let's think about which options we can nix before shipping comments.
i'd rather have less options going out the door and then we add options as we feel out what people are asking for. the bias here would be to have the opinionated choices for these i think (regionReveal/regionMove/regionResize) — less fiddly for people downstream to think about and also, on our end, less maintanence of potentially unused options

mimecuvalo (approved):
good to go 👍 
but see comment on maybe nixing some of the options here as a followup, worth looking at our options in general before shipping

PR #9675, opened 2026-07-23
feat(commenting): preview a marker's threads on hover

meg-an31:
Builds on @jsscclr's #9639 — the coincident-pin spreading, the count-badge thread list, and the stack-list cards are all Jessica's work, included here in full and ported onto current `commenting`. That PR is the place to review those pieces; this one adds the hover preview and the coincident-stack behaviour on top.

In order to tell what's behind a comment marker without opening it, this adds a hover preview to every marker on the canvas: a single pin, a coincident stack, and a cluster badge. Today a badge showing "7" says nothing about whether any of those seven are worth your attention, and the only way to find out is to open the list and read them one at a time.

The preview imitates whatever the click would open. Hovering a pin shows the thread panel minus its header; hovering a badge shows the same card list the click produces. Both sit at the popover's own origin, so the first comment lands on exactly the pixel it would occupy once opened — clicking swaps the panel for the real thing without the text shifting under the pointer. Clicking a card in a stack or cluster preview opens that thread directly, so the preview is a way in, not just a tooltip.

### Coincident imprecise stacks open immediately

Several imprecise comments on one shape all resolve to the same corner point, so they're genuinely coincident and zoom can never separate them. Left to the clustering, they surfaced as a zoom-to-split cluster badge — you had to zoom in near `maxSplitZoom` before the cascading list appeared. This routes a cluster node that is *purely* one coincident stack (every member in a single `pin-stacking` group, nothing distinct mixed in) to the `ThreadStackPin` cascade instead, so it opens on click with no zoom. A stack still clusters normally with genuinely-distinct neighbours — the mixed node stays a cluster badge with the correct count, and only flips to the cascade once its neighbours have split off. The change is a single render branch on the clustering output plus a `stackGroupOf` helper; the clustering input, the zoom schedule, and the counts are untouched.

Two reviewer notes:

- **`thread-view.tsx` was ported onto current `commenting`.** #9639 extracted the thread rendering before author denormalization (#9630/#9631), the `@tldraw/mentions` package (#9632), and the sign-in `canComment`/`ComposerFallback` work all landed. So the extracted `ThreadView` gained the `CommentAuthor` resolver, the mentions imports, and the `canComment` gating + `ComposerFallback` footer to match what landed inline on `commenting`. #9639 needs this port regardless to land.
- **The stack badge was never wired for pointer events.** `.tlui-cmt-canvas-pin` is deliberately `pointer-events: none`; the regular pin re-enables clicks through its hit layer, but `.tlui-cmt-canvas-stack-badge` never did — it rendered but couldn't be hovered or clicked. Latent because stacks previously surfaced as the cluster badge (which does set `pointer-events: auto`); exposed once coincident stacks route to `ThreadStackPin`. Fixed with the matching one-liner.

### Change type

- [x] `feature`

### Test plan

1. Open the **Comment shape precision** example, switch to **"Shape only"** (every comment imprecise).
2. Place 2+ comments on one shape — they snap to the same corner and stack.
3. The corner shows a count badge: **hover** previews the threads, **click** opens the cascading list — no zoom.
4. Click a card to open that thread inline; place another coincident comment and confirm it joins the list with no duplicate pin.
5. Also comment on a second, distant shape; zoom out until the stack and that comment merge — one cluster badge, correct total count — then zoom in and confirm the stack peels off into its cascade.
6. Hover a single pin: its opening comment appears where the thread will open; clicking swaps without the text moving.
7. In light and dark themes, hover a collapsed card — the whole card tints, no pale outline at its edge.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Add hover previews to canvas comment markers, and open coincident imprecise-comment stacks as a cascading list on click rather than requiring a zoom.

### API changes

- Added optional `open` prop to `CountBadge`, for the selection ring when a stack's list is showing

### Code changes

| Core code       | +1472 / -444 |
| Tests           | +195 / -0    |
| Automated files | +16 / -3     |


jsscclr:
When I add multiple comments on a shape without adjusting zoom level, the comment clusters can't be interacted with. Do we need to handle the attached to shape case here, where we force recomputing the cluster tree...?

https://github.com/user-attachments/assets/07c08a80-77d5-45b8-98df-88670effc42b

https://github.com/user-attachments/assets/9029420e-0e88-48ab-8cfb-990fccd2e22c

PR #9676, opened 2026-07-24
feat(dotcom): restructure sync-worker analytics around per-domain writers

driev:
In order to answer basic questions about the sync worker in production — which room an event belongs to, why postgres connections die, how long queue messages sit before delivery, and whether room durable objects are actually hibernating — this PR gives the `MEASURE` dataset a schema, adds the datapoints that were missing, and documents both.

The dataset had accumulated four conventions and no written record of any of them. `blob3` meant seven different things depending on the event; most events carried no room or user identity at all, so the highest-volume ones couldn't be attributed to anything. `apps/dotcom/analytics-blob-layout.md` is the place to start reviewing — it records what the layout was, what it is now, and why each decision went the way it did.

**One writer per domain.** `writeDataPoint` took a positional `blobs: string[]`, so the meaning of `blob3` was decided independently at ~15 call sites with nothing to compare them against. Now a small header is global — `index1` subject, `blob1` event, `blob2` env — and everything from `blob3` on belongs to the domain that writes it: room, user, replicator, postgres, queue, screenshot. Three of those writers already existed as `logEvent` switches; this formalises them. Payloads are byte-identical, so no existing query changes and there's no cutover.

**`index1` is the subject, not the emitter.** It holds the durable object id of the room a datapoint is _about_, which is also `$workers.durableObjectId` in Workers Logs and the per-object filter on the namespace's metrics — one value joins analytics, logs and CPU/storage. Worker-context writers derive it locally via `getRoomDurableObjectId`, which shares its name derivation with `getRoomDurableObject` so the two can't drift. The screenshot surfaces previously indexed on a hash of the board slug, which joined to nothing; they now resolve the file id during board resolution and index on the room.

**Identifiers stay out of blobs.** Cloudflare's guidance is that blobs hold low-cardinality dimensions and that raw ids don't belong in them — and `index1` is the opposite case, because Analytics Engine samples _per index value_. An earlier revision of this branch added a `blob16` user id; it's gone. A blob isn't the sampling key, so filtering on one queries a sampled dataset and can come back empty for a low-volume user.

**Hibernation residency.** Room durable object duration is ~97% of the account's durable object bill, and it's billed as wall-clock while the object can't hibernate. `room_wake`, `room_cold_start` and `room_resident_gap` make that observable: hibernation is defined by wiping in-memory state, so an instance field survives a gap only if the object stayed resident and was billed for it. `sum` of the gap converts directly to GB-s and to money, attributable per room.

Two decisions worth a look, both of which removed things rather than adding them:

- **The domain isn't a column, and isn't a parameter.** An event name already determines its domain, so a column holding it would be a second copy that could disagree. `EVENT_DOMAINS` is exported for the query side instead, typed `Record<TLDataPointName, TLAnalyticsDomain>` so adding a datapoint name without a domain is a type error.
- **`room_size_mb` is now indexed** like every other room event. It used to write directly on the grounds that it feeds percentile queries rather than lookups, but an index costs a percentile query nothing and makes outliers attributable.

`apps/dotcom/access-logging.md` is a proposal only — nothing in it is implemented. It records why access logging can't live in Analytics Engine (it samples, it has no delete API, and retention caps at 90 days), with production volume and cost figures attached so the retention decision has real numbers.

### Change type

- [x] `improvement`

### Test plan

1. Confirm a Grafana panel querying older events (`enter`, `rate_limited`, `replicator`) still resolves — payload positions are unchanged by the domain split
2. Check that screenshot rows now carry a room durable object id in `index1` rather than a slug hash, and that an unresolvable board carries no index at all
3. Watch `room_start` and the new residency datapoints in production to establish whether room durable objects hibernate between events

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Section | LOC change  |
| ------- | ----------- |
| Apps    | +489 / -110 |
| Tests   | +326 / -0   |
| Docs    | +357 / -0   |


driev:
## Proposed blob layout v2, for when this gets reworked

Context from #9708, parked here because it belongs with `Metrics` rather than in one durable object. Nothing to action on this PR as it stands — this is the shape to aim at when you clean it up.

### The problem

`MEASURE` has accumulated four conventions and no written schema:

- `blob3` means three different things depending on the event — `userId` on `rate_limited`, the operation type on `r2_queue_depth`, the room elsewhere. (#9708 removes the room case and the `'unused'` placeholder that sat on `blob4`, but the rest stands.)
- The file DO puts the *specific* event in `blob1` (`enter`, `room_start`, `on_request_total`, ~28 distinct values). The replicator and user DO do the opposite: one name (`replicator`, `user_durable_object`) with the real type in `blob3`.
- The screenshot surfaces use self-describing `key:value` blobs (`source:og`, `cache:hit`); everything else is bare positional.
- Most events carry no room or user identity at all, so the highest-volume ones — the `on_request_*` timers, `r2_queue_depth`, `postgres_*` — can't be attributed to anything.

Positions are load-bearing in three places: the **Events** and **Events V2** Grafana dashboards, and `internal/scripts/fetch-screenshot-metrics.ts` (`blob3 AS source`), the only tracked consumer.

### Proposal: name the fields in code, keep positions an implementation detail

Analytics Engine has no way to name columns — the schema is fixed at `blob1…blob20` / `double1…double20` / `index1`, and naming only happens at query time via `AS` aliases. So the naming has to live in our code:

```ts
// The single source of truth for the layout. Call sites use names; nothing
// outside this module knows a position.
const BLOB_ORDER = ['env', 'doType', 'event', 'room', 'user'] as const

type EventFields = Partial<Record<(typeof BLOB_ORDER)[number], string>>

write(name: string, fields: EventFields, data?: { doubles?: number[]; index?: string })
```

Export the alias map alongside it so Grafana and `fetch-screenshot-metrics.ts` consume the same mapping instead of hand-writing `blob4 AS room` in every panel.

Resulting layout:

| Slot | Contents |
| --- | --- |
| `index1` | id of the DO this datapoint is *about* (see below) |
| `blob1` | env / worker name (`production-tldraw-multiplayer`) |
| `blob2` | DO type — `file`, `user`, `replicator`, `worker` |
| `blob3` | event type |
| `blob4` | room key — derived DO id, empty when N/A |
| `blob5` | user id — raw, empty when N/A |
| `blob6`+ | event-specific payload |

The two identity slots are deliberately asymmetric. A room key is a derived DO id because the slug is a bearer credential — `tldraw.com/f/<id>` is the access. A user id is only an identifier, and user pages are behind authz, so it goes in raw and joins straight to the `user` table, Clerk and support tickets with no derivation step. (It does make those rows personal data, and AE has no delete API, so retention is what bounds it.)

Two further things worth being deliberate about:

**Identity slots are fixed, not DO-dependent.** The tempting version is "room id or user id next, depending on which DO emitted it" — that reintroduces exactly the varying-column defect above. Fixed slots with empty strings cost nothing: the per-datapoint blob budget is 16 KB (raised from 5 KB in June 2025), so padding is free.

**Skip value prefixes.** `LIKE` and `HAVING` landed in January 2026, so `WHERE blob3 LIKE 'room:%'` does work now. But a prefix makes a value self-*describing*, not self-*locating* — you still have to know which column to read, and AE has no array or map functions to scan generically. A prefix is a seatbelt against drift, not a schema. Better to have a schema.

### Why `index1` is the DO id

Cloudflare already keys its own telemetry on it: `$workers.durableObjectId` identifies the emitting instance in Workers Logs, and a namespace's metrics can be filtered to a single object by id or name. So analytics, logs, traces and CPU/storage all join on one value — which is worth more once the log/trace export on this PR is on. It's also one-way, so unlike a raw slug an id read out of the dataset doesn't open the board. #9708 already does this for the file DO.

Note AE allows exactly one index, so this spends it. In the file DO that displaced `localClientId`, which turned over on every page load and was never queried.

**The index is the subject, not the emitter.** For a DO those coincide, but worker-context writers (the OG route, the queue consumer) are *about* a room even though no DO emitted them — and they can derive the same value, since `idFromName` is a local derivation with no network cost that does not create or wake the object:

```ts
env.TLDR_DOC.idFromName(`/${ROOM_PREFIX}/${fileId}`).toString()
```

That gives one uniform join column across every writer, and better sampling: a hot board's renders sample alongside its sync events, which is the tenant isolation the index is designed for. `blob2` still says whether a DO or a worker emitted the row. The cost is that a worker-emitted row's index won't match a `$workers.durableObjectId` log line — there isn't one — though it still lines up with that object's DO metrics.

For published boards derive from `file.id`, not `board.slug`: the latter is the published slug, and `getPublishedFileInfo` already resolves it through `SNAPSHOT_SLUG_TO_PARENT_SLUG` and returns the file id. Deriving from the published slug yields a valid-looking id for an object that doesn't exist.

### Migration

Two Grafana dashboards (**Events**, **Events V2**) plus `fetch-screenshot-metrics.ts` read positions, so a v2 layout rewrites all three. They are the only consumers, and between them they use well under a dozen distinct columns — so this is a contained sweep rather than an archaeology project, provided it happens in one PR alongside the code.

Old and new rows coexist in the dataset for the 90-day retention window, and `blob1` would mean "event name" on the old ones and "env" on the new. A hard cutover is cleanest: change everything at once, write the date down, and after 90 days only one layout exists. A version-marker blob is the alternative, buying cross-boundary queries at the cost of a permanent column.

Two things worth fixing in the same pass, since every panel gets touched anyway: several use bare `count()`/`sum()` where sampling requires `sum(_sample_interval)`, and **Events** panel 13 has no `blob2` filter at all, so it silently mixes production, staging and every PR environment.

### Open questions

- **Doubles need a rule too.** They're positional with no self-describing values and vary per event today. Simplest honest rule: doubles are always event-specific and never queried across events.
- **Events with no subject.** `postgres.ts` only sees `env`, so it has no room or user to index on. Suggest omitting `indexes` entirely there, so "no index" reads as "not object-scoped" rather than inventing a sentinel. The replicator, stats and logger DOs are singletons, so their index is a constant — harmless but inert.

mimecuvalo:
is this meant to be stacked on top of https://github.com/tldraw/tldraw/pull/9667 ?

driev:
> is this meant to be stacked on top of #9667 ?

Nope... this grew arms and legs, and became more of a research PR. I'll close this and break things down into phases.

PR #9677, opened 2026-07-24
feat(commenting): extract comment authorizers into @tldraw/sync-collaboration

MitjaBezensek:
In order to let SDK users running their own sync server enforce the same comment permissions tldraw.com does — authors edit only their own comments, deletion is a write-once `isDeleted` soft-delete flag on both comments and threads (client hard-deletes are always rejected), resolutions are self-attributed, and reactions must land at the canonical id for their (comment, user, emoji) triple — this PR moves the comment, comment-thread, and comment-reaction record authorizers out of dotcom's sync-worker into a new server-safe package, `@tldraw/sync-collaboration`. Previously this policy lived only in `apps/dotcom/sync-worker/src/authorizeFileRecord.ts`, so anyone wiring `authorizeRecord` on a `TLSocketRoom` had to reimplement it by hand.

The package is scoped to collaboration features generally rather than commenting specifically: it's the server-side home for the collaboration features' sync logic, so future server-side pieces (note text attribution, mentions, soft-delete room seeding) get a home without a new package each time. Its only exports for now are a factory and its options type, generic over the host's session meta. The host's `getUserId` callback is called exactly once per authorized write, so it can't split identity across checks even if it's impure:

```ts
interface SessionMeta {
	userId: string | null
}

// Comment records ride alongside your document records, so widen the room's record union.
type MyRecord = TLRecord | TLComment | TLCommentThread | TLCommentReaction

new TLSocketRoom<MyRecord, SessionMeta>({
	authorizeRecord: {
		...createCommentAuthorizers<SessionMeta>({ getUserId: (session) => session.meta.userId }),
	},
})
```

Dotcom now composes the factory output with its local note-shape attribution authorizer; the policy itself moved verbatim from the base branch (including the soft-delete and reaction rules — the policy tests moved with it, and dotcom keeps wiring smoke tests plus its note-shape suite).

A separate package rather than an export from `@tldraw/commenting` was deliberate: commenting's modules value-import the `tldraw` barrel, which bundlers can't tree-shake (no `sideEffects` annotations, impure top-level statements), so a server importing anything from commenting's index pulls the entire client graph — react, tiptap, the editor. We also ruled out `@tldraw/sync-core` (feature policy doesn't belong in the generic transport package) and a `./authorize` subpath export (the build/prepack/api-extractor pipeline is single-entry). `@tldraw/sync-collaboration` has only server-safe deps (`store`, `sync-core`, `tlschema`, `utils`), so worker and node bundles stay slim by construction.

### API changes

- New package `@tldraw/sync-collaboration` with `createCommentAuthorizers<SessionMeta>(opts)` and `CommentAuthorizerOptions<SessionMeta>`
- No changes to `@tldraw/commenting`'s public API

### Change type

- [x] `feature`

### Test plan

- [x] Unit tests

Policy suites (36 tests, including a once-per-write `getUserId` contract test) live in `packages/sync-collaboration/src/comment-authorizers.test.ts`; dotcom keeps 2 wiring smoke tests + the note-shape suite. `yarn typecheck` and `yarn api-check` green.

### Code changes

| Core code       | +230 / -0   |
| Tests           | +387 / -305 |
| Automated files | +25 / -0    |
| Apps            | +7 / -136   |
| Config/tooling  | +86 / -0    |

### Release notes

- Add `@tldraw/sync-collaboration`, a server-safe package exporting `createCommentAuthorizers` so custom sync servers can enforce tldraw.com's comment permissions (author-only edits, owner-only write-once soft-deletes for comments and threads, self-attributed resolutions, canonical-id reactions).


mimecuvalo (commented):
high-level comment: do we need to create another pkg just for this?

MitjaBezensek:
> high-level comment: do we need to create another pkg just for this?

Yeah, unsure. If we put it into the main commenting package then it complicates importing it on the server. Like you'd install a bunch of UI / React code, just to use a few pure functions. Maybe subpath export would work?

PR #9680, opened 2026-07-24
fix(commenting): finish the text-range sweep in anchor-lifecycle

jsscclr:
In order to make `commenting` typecheck again, this PR finishes #9679's text-range sweep in `anchor-lifecycle.ts`. #9661 added that file (with `text-range` handled as a shape-anchor kind) while #9679 (which dropped the text-range anchor) was in flight — the two merged in the same minute with no textual conflict, so the first typecheck to see the combination failed: the union no longer has `'text-range'`, making the comparisons `TS2367` no-overlap errors.

The change is the mechanical completion of the sweep: `ShapeAnchor` narrows to the `shape` variant, and the two type guards drop their `text-range` arms. No behavior change — the text-range code paths were already unreachable once the anchor kind was removed.

### Change type

- [x] `bugfix`

### Test plan

1. `yarn typecheck` passes at the branch tip (it fails on `commenting` today).
2. Package tests: 195 passing, unchanged.

- [x] Unit tests

### Release notes

- Fix a type error on the commenting branch from two crossing PRs (#9661, #9679).

### DIFF
--- a/packages/commenting/src/canvas/anchor-lifecycle.ts
+++ b/packages/commenting/src/canvas/anchor-lifecycle.ts
@@ -11,7 +11,7 @@ import { getCommentingOptions } from './options'
 import { commitCommentMutation } from './state'
 import { anchorPagePoint, impreciseShapePinInset } from './thread-state'
 
-type ShapeAnchor = Extract<TLCommentAnchor, { type: 'shape' | 'text-range' }>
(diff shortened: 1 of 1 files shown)


jsscclr:
cc: @mimecuvalo, we merged our PRs at the same time 😂

mimecuvalo:
> cc: @mimecuvalo, we merged our PRs at the same time 😂

achievement unlocked :high-five:

PR #9682, opened 2026-07-24
feat(tldraw): smooth collaborator cursors between presence updates

frolic:
In order to stop remote cursors from looking choppy in multiplayer, this smooths collaborator cursor motion on the receiver. Today presence carries the cursor position throttled to ≤30fps, and the overlay draws each sample at its raw position — so remote cursors teleport from point to point even on a fast connection.

This is a spike.

### The change

`CollaboratorCursorOverlayUtil` now measures the interval between the position updates it receives per collaborator and, on the editor tick loop, tweens the rendered cursor between the last two received positions over that interval. It's **pure interpolation** — a cursor only ever moves between points that have actually arrived — so it can't overshoot or rubber-band, at the cost of trailing ~one send interval (imperceptible at 30fps). The interval is tracked as an EMA so it adapts to tldraw's changing sync rate, and a cursor just holds in place when its sender goes idle.

The overlay canvas only repaints on reactive change, so an animation-clock atom bumped whenever a displayed position moves drives the per-frame repaints between updates.

It's on by default (`smoothing` option, default `true`), so every collaborator cursor across the SDK gets it with no opt-in.

### Approach note

An earlier revision of this branch broadcast a velocity vector in the presence/pointer schema and dead-reckoned on the receiver. It was dropped: extrapolation is the only way to beat the send-interval latency, but it rubber-bands when a sender stops or reverses mid-flick, and for a cursor that trade isn't worth it. Interpolation gives the smoothness with none of the overshoot and — importantly — **needs no schema change, no migrations, and no new wire data**. This revision is purely receiver-side and non-breaking. (History note: the branch still contains the velocity commit followed by this refactor; I'll squash to a single commit before marking ready.)

Cursors are drawn to a canvas overlay (the `OverlayUtil` system), not DOM, so the tween is done in the render loop — a CSS `transition` isn't reachable here.

### Change type

- [x] `improvement`

### API changes

- `CollaboratorCursorOverlayUtil` options gain `smoothing` (boolean, default `true`).

### Test plan

1. Open the **Cursor smoothing** example (`/cursor-smoothing`).
2. The simulated peer's cursor is only updated every 120ms; confirm it glides smoothly.
3. Untick **Smooth cursor** to see the raw, throttled stepping.
4. Tick **Stress test** (abrupt stops & direction reversals) and confirm the motion stays smooth with no overshoot/pull-back.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Smooth collaborator cursors by interpolating between throttled presence updates.

### New examples

- `collaboration/cursor-smoothing` — compares smoothed vs raw collaborator cursor motion, with a stress-test toggle for abrupt stops and reversals.

### Code changes

| Core code       | +166 / -4   |
| Automated files | +4 / -0     |
| Documentation   | +165 / -0   |

### DIFF
--- a/apps/examples/src/examples/collaboration/cursor-smoothing/CursorSmoothingExample.tsx
+++ b/apps/examples/src/examples/collaboration/cursor-smoothing/CursorSmoothingExample.tsx
@@ -1,0 +1,112 @@
+import { useRef, useState } from 'react'
+import {
+	CollaboratorCursorOverlayUtil,
+	createUserId,
+	Editor,
+	InstancePresenceRecordType,
+	TLInstancePresence,
+	Tldraw,
+} from 'tldraw'
+import 'tldraw/tldraw.css'
+
+// How often the simulated peer "broadcasts" a cursor update, in milliseconds.
+// Real collaborative sync is throttled to ~30fps (≈33ms); we use a slower rate
+// here to make the difference between raw and smoothed motion obvious.
+const BROADCAST_INTERVAL_MS = 120
+
+// A smooth wandering path (two out-of-phase sine pairs) in page space. Returns
+// the peer's "true" cursor position at a given time.
+function pathAt(ms: number) {
+	const t = ms / 1000
+	return {
+		x: 320 + 180 * Math.sin(t * 0.9) + 60 * Math.sin(t * 2.3),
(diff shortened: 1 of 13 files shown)


frolic:
I learned that we've been down this road and the trade-offs aren't great so closing this out

PR #9685, opened 2026-07-24
spike(commenting): canvas-drawn comment pins as overlay utils (alternative to the DOM pin layer)

jsscclr:
> [!NOTE]
> **Status: exploratory alternative, not the plan of record.** The commenting layering question was resolved by moving collaborator cursors to a DOM layer above the comments layer (#9695) — commenting keeps its DOM pin layer unchanged. This PR stands as the worked-out canvas-native alternative: what pins-as-overlay-utils looks like end to end (marker, badges, regions, drag, hit-testing), if pins ever need to interleave with canvas-drawn content rather than just sit above it.

An alternative to the two-canvas overlay-band split in #9657: the pin marker becomes a canvas `OverlayUtil` — a canvas-space pseudo-shape like the built-in handles and collaborator cursors — while everything richer than the marker stays DOM.

## The idea

The built-in overlays already form one ordered plane by `zIndex`:

```
ShapeIndicator 50 · SelectionForeground 100 · ShapeHandle 200 · Brush 300 ·
ZoomBrush 400 · SnapIndicator 500 · Scribble 600 · CollaboratorBrush 700 ·
CollaboratorScribble 800 · CollaboratorHint 900 · ArrowHint 1000 · CollaboratorCursor 1100
```

So "pins above selection chrome, below cursors" is a single integer: **`zIndex: 1050`**. No second canvas, no `band` enum — the pin joins the one flattened overlay canvas at its natural z.

## What's in the PR

**`CommentPinOverlayUtil`** (`packages/commenting/src/canvas/pin-overlay.ts`) draws the pin markers and cluster badges:

- Teardrop marker matched to the DOM pin's CSS (34px, `border-radius: 50% 50% 50% 3px`, theme colors read from the live CSS variables, resolved greys per color mode, open ring, hover grow via the overlay manager's `hoveredOverlayId`).
- `getGeometry` + `onPointerDown` carry hit-testing and clicks through the existing `OverlayManager` routing: a pin click toggles `openThreadId`, a badge click requests its cluster's expand zoom.
- Minimap dots via `renderMinimap`.

**Pin dragging re-anchors threads**, matching the DOM marker's gesture. A consumed overlay press leaves the select tool in `select.idle`, so the util owns the gesture with window pointer listeners — no tool changes:

- 4px threshold decides click vs drag; a click toggles the popover on release.
- The drag previews through a `commentPinDrag` atom: the util draws the dragged pin at the drag point; `CanvasComments` translates a region's box preview and keeps the open popover riding the marker.
- Drop re-anchors onto a shape (precise per `shouldBePrecise`, alt-aware), else to a point; regions translate keeping size and pin corner. Commits through `commitCommentMutation` with the drag history option. Shape hinting and the move cursor match the DOM behavior.
- A per-pin `movable` flag mirrors the commenting permission and `regionMove`; non-primary buttons keep their default routing (middle/space pans never reach the state chart).

**`CanvasComments` stays the brain.** It computes clustering, holds, orphans, and the open thread exactly as before, and mirrors the resulting pin/badge list into a `commentPinDisplay` atom the util renders. Entries carry **anchors, not points** — the util resolves each anchor's page point reactively at draw time, so pins ride shape moves and region edits without the mirror rewriting.

**The DOM half is unchanged**: thread popover, composers, region box and resize handles, sidebar. The popover was already coordinate-anchored (`pageToViewport`), so removing the DOM marker didn't touch its positioning.

**Deleted along the way**: the marker pass-through plumbing (`usePassThroughWheelEvents` on markers, hover pass-through, pan-gesture forwarding to the canvas) — all of it existed only because a DOM node sat over the canvas intercepting events. Canvas markers intercept nothing.

Cross-boundary wiring uses request/serve atoms, following the `revealThreadRequest` pattern: the util writes `clusterExpandRequest`; the React layer (which owns the cluster table) serves the zoom.

## Not carried over yet

- **Cluster fade animation** — the DOM fade machinery is deleted; canvas alpha animation would be its replacement.
- **The `PinContent` component slot** — arbitrary React content can't render to the canvas. This is the one real API conflict: consumers who set `PinContent` get the default author-initial marker instead. Open design question, not silently dropped.

## Verification

- 194/194 commenting package tests pass; full `tsc --build` clean; API report regenerated.
- Not yet run in a browser — layering and interaction claims are code-verified only.

## Relative to #9657 and #9695

Three answers to "cursors above pins, pins above selection," one per available move:

- #9657 — split the overlays canvas into bands around the DOM pins.
- **#9695 (plan of record)** — move cursors to the DOM element ladder above the comments layer; pins untouched.
- This PR — move pins into the canvas at `zIndex: 1050`, below the canvas cursors.

With #9695, the ordering problem is solved without this PR or #9657. What this PR uniquely provides is pins that interleave with *canvas-drawn* content (selection chrome, brushes) — relevant only if some future requirement puts canvas content above pins.

### API changes

**Added (`@tldraw/commenting`):**

- `CommentPinOverlayUtil` + `TLCommentPinOverlay` — the canvas pin/badge overlay (zIndex 1050).
- `CommentRegionOverlayUtil` + `TLCommentRegionOverlay` — the canvas region box/handles overlay (zIndex 1040).
- `commentPinDisplay` + `CommentPinDisplay` / `CommentPinDisplayPin` / `CommentPinDisplayBadge` — the display atom `CanvasComments` mirrors into and the utils render from.
- `commentPinDrag`, `commentRegionEdit`, `clusterExpandRequest` — the gesture-preview and request/serve atoms.
- `RegionHandle`, `REGION_CORNERS`, `REGION_EDGES`, `resizeRegion` — the region resize geometry, moved to `region-geometry.ts`.

**Behavior changes:**

- `CanvasComments` no longer renders DOM pin markers, cluster badges, or region boxes/handles — those draw on the overlay canvas. The popover, composers, and sidebar are unchanged.
- The `PinContent` component slot is currently not honored (arbitrary React content can't render to the canvas) — threads show the default author-initial marker. Open design question.
- Cluster fade animation is removed pending a canvas replacement.

**Removed:** nothing public — the deleted DOM components (`RegionBox`, `RegionResizeHandles`, marker internals) were not exported.


jsscclr:
Closing, superseded by #9695

PR #9687, opened 2026-07-26
fix(commenting): break comment ordering ties on id

steveruizok:
🙇‍♂️ Tiny one. We have a bunch of places where we sort based on some property that may be duplicated, so this adds a little deterministic backup for tie breaking (using the id of the item).

---

In order to keep a thread's replies in the same order for everyone reading it, this gives comment ordering a tie-break.

Comment ordering sorted on `createdAt` alone. `createdAt` is client-supplied wall-clock (`createComment` uses `props.now ?? Date.now()`, and nothing stamps it server-side), so ties are reachable — a seeded or imported batch, a scripted burst, a skewed clock. `Array.prototype.sort` is stable, so a tie left the records in whatever order the input array already had. For records read out of the store that's insertion order into the query index, which differs between peers and moves a record to the end when it's removed and re-added — which the anchor lifecycle does routinely, on undo of a delete and on the re-create half of a page move.

`sortByCreatedAt` in `@tldraw/utils` falls back to `id`. Ids aren't time-ordered, so the order within a tie is arbitrary — the point is that it's the same arbitrary order everywhere. It sits next to the existing `sortById` and uses relational comparisons rather than subtraction, so a non-numeric timestamp falls through to the id instead of returning `NaN` and making the comparator non-total.

Two things worth a reviewer's attention:

**The sidebar sort moved off display strings.** It sorted the mapped `CommentListItemProps`, whose `date` is a preformatted ISO string, so the tie-break would have had to key off rendered text. It now sorts the threads before mapping, against a precomputed key of `(first comment ?? thread).createdAt`. Same ordering as before — resolved last, then newest first.

**The Zero query change is for the polyfill, not the server.** Real Zero completes every query's ordering with the primary key (`addPrimaryKeys` in `complete-ordering`), so the notifications feed was already deterministic in production. The local polyfill sorts the raw AST `orderBy` and does no such completion, so ties there fell to input order — and at the `LIMIT 50` boundary that decides which rows sync at all. The explicit `.orderBy('id', 'asc')` is `asc` deliberately, to match what Zero appends rather than diverge from it.

Also fixes an adjacent bug in a line this PR was already rewriting: `useComments` sorted the store query's memoized array in place, which reordered it for every other reader of `getComments` and defeated the query's own shallow-equality check, re-rendering consumers on unchanged data.

Not taken: `TldrawApp.ts:505` and `TlaSidebarRecentFiles.tsx:137` sort files by a `date` that falls back to `0`, so every file without a visit date ties. Same hazard, different data — worth its own PR rather than widening this one.

### Change type

- [x] `bugfix`

### Test plan

1. Open a board with several comment threads and the comments sidebar.
2. Confirm thread and reply order is unchanged, and identical across two browsers on the same board.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix comment and reply ordering being inconsistent between collaborators when two comments share a timestamp.

### API changes

- Added `sortByCreatedAt()` to `@tldraw/utils` — sorts by `createdAt` oldest-first, breaking ties on `id`

### Code changes

| Core code       | +70 / -12  |
| Tests           | +81 / -1   |
| Automated files | +6 / -0    |
| Apps            | +3 / -1    |

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/commentNotifications.ts
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/commentNotifications.ts
@@ -1,4 +1,5 @@
 import { extractMentionIds } from '@tldraw/dotcom-shared'
+import { sortByCreatedAt } from '@tldraw/utils'
 
 /**
  * Why a comment shows up in a user's notifications feed:
@@ -89,5 +90,6 @@ export function categorizeCommentNotifications<T extends CommentNotificationInpu
 		notifications.push({ comment, reasons, primaryReason })
 	}
 
-	return notifications.sort((a, b) => b.comment.createdAt - a.comment.createdAt)
+	// Newest first, ties broken on comment id so the feed reads the same on every device.
+	return notifications.sort((a, b) => sortByCreatedAt(b.comment, a.comment))
 }
--- a/packages/commenting/src/canvas/comments-sidebar.tsx
+++ b/packages/commenting/src/canvas/comments-sidebar.tsx
@@ -1,9 +1,11 @@
 import { type CommentAuthor } from '@tldraw/mentions'
+import { sortByCreatedAt } from '@tldraw/utils'
 import { ReactNode, useCallback, useRef } from 'react'
 import { createPortal } from 'react-dom'
 import {
 	TLComment,
 	TLCommentId,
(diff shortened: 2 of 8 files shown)


steveruizok:
Seems too rare to matter, and wrong sort has no consequence.

PR #9683, opened 2026-07-24
chore(commenting): remove component-studio, now in tldraw-internal

jsscclr:
## What this does

Removes `apps/component-studio` from the commenting branch. The studio moved to tldraw-internal (tldraw/tldraw-internal#1054), where it lives at `apps/private/component-studio` and consumes the published `tldraw` and `@tldraw/commenting` builds (the npm `internal` dist-tag) instead of workspace packages.

Besides the app itself: the workspace entry is pruned from `yarn.lock`, and the `packages/commenting` README drops its pointer to the studio (that package is publishable, and the studio's new home isn't reachable from a public README).

## How to test

Preview: https://pr-9683-preview-deploy.tldraw.com/

Nothing user-facing changes — the studio was a private dev-only workspace nothing else imported. Expect the preview to behave identically to `commenting`; repo typecheck and the commenting package's API extractor pass locally.

PR #9688, opened 2026-07-26
feat(commenting): gate every commenting component on the license

steveruizok:
In order to make "commenting is a licensed feature" true of the code and not just the built-in layout, this gates every component `@tldraw/commenting` exports on the commenting license.

Auditing what the license actually enforced turned up a gap: only `CanvasComments`, `CanvasCommentsSidebar`, and the comment toolbar button checked it. The tool, the store writes, the anchoring and clustering logic, and every presentational component — `CommentCard`, `CommentThread`, `CommentComposer`, `CommentPin`, `CommentsList` — rendered and ran without a license. So the practical bypass wasn't "reimplement commenting", it was to import the same components and write the ~200 lines of positioning glue that `CanvasComments` provides, with the clustering algorithm and anchor lifecycle coming along for free.

This PR closes the component half of that gap. The data-layer half is still open (see below).

### Where the line now sits

Gated: all 17 exported components, via an internal `withCommentingLicense` wrapper.

Not gated, deliberately:

- `formatRelativeTime`, `renderMarkdown`, `richTextToPlaintext` — pure functions with no render to suppress.
- The `@tldraw/mentions` re-exports — that package is shared with shape rich text, so gating it would gate a core text feature.
- The store functions (`putCommentRecords` and friends), `CommentTool`, and the anchoring/clustering modules. These take an `Editor`, and `LicenseManager` currently lives only in React context — there's no path from an editor to the license. Gating them means first hanging the manager off the `Editor`, which felt like a separate change.

### Design notes

**The gate is a wrapper, not an early return.** The license flag starts `false` in production and flips when async validation resolves. An early return inside each component would change how many hooks it calls between the pending render and the licensed one, which React rejects with "rendered more hooks than during the previous render". Wrapping keeps the inner component's hooks all-or-nothing.

**`useLicenseFeatureFlag` now fails closed instead of throwing.** `LicenseContext` defaults to `{} as LicenseManager`, so a gated component rendered outside a `LicenseProvider` was calling `isFeatureEnabled` on an empty object. It now returns `false`: no provider means no license key, which means no feature.

**This changes the presentational components' contract.** They previously worked standalone; they now need a licensed `<Tldraw>` or `LicenseProvider` above them. That's the point, but it's a real behavior change for anyone composing them directly, and it's why the exported signatures move from `function X(props): JSX.Element` to `const X: (props) => ReactNode` in the API report.

### Still open

The comment record types (`TLComment`, `TLCommentThread`, `commentSchemaRecords`) live in `tlschema`, a core package every SDK licensee gets. They're opt-in rather than in the default schema, but a core licensee can reasonably say the record types came with their core license. Worth settling before ship, since it's a schema change and gets harder later.

Also worth knowing: this touches `apps/component-studio/src/harness.tsx`, which #9683 deletes wholesale. The studio's comment sketches render outside any editor, so they need something to look licensed against — this adds a stub supplying only `isFeatureEnabled`. It's a hack and commented as one, and it disappears with that PR. Expect a modify/delete conflict there; take the delete.

### Change type

- [x] `api`

### Test plan

1. Run a production build of any app embedding commenting without a license key — the comment UI should render nothing rather than crash.
2. With a key that includes commenting (or collaboration), the comment UI should appear as before.
3. Run the component studio (`yarn dev` in `apps/component-studio`) — comment sketches should still render.

- [x] Unit tests

### API changes

- Changed all 17 exported `@tldraw/commenting` components from function declarations to `const` wrappers; call sites are unaffected, but the reported signatures change from `(props) => JSX.Element` to `(props) => ReactNode`
- Added `LicenseContext` to `@tldraw/editor` (`@internal`), so a license value can be supplied without a full `LicenseProvider`
- Changed `useLicenseFeatureFlag` to return `false` rather than throw when no license manager is in context

### Code changes

| Core code       | +128 / -57 |
| Tests           | +42 / -0   |
| Automated files | +21 / -17  |
| Apps            | +31 / -7   |
| Config/tooling  | +4 / -0    |

### Release notes

- Commenting components now require a license that includes the commenting feature. Without one they render nothing.


mimecuvalo:
The audit in this description is the valuable part — mapping what the license actually enforced, and naming the real bypass as "import the components and skip `CanvasComments`" rather than "reimplement commenting," is a genuinely useful finding.

Where we landed, though, is that we don't want to gate the individual presentational components. It's a lot of checks on components that aren't much use standalone, and it puts a license read on things like `CountBadge` and `SendButton` that are a `div` and a `button`.

That leaves the gap where you already identified it in "Still open": `CommentTool`, the store writes, and the anchoring/clustering modules. That's where the clustering algorithm, anchor lifecycle, and positioning glue live — the part that's actually worth licensing — and they can't be gated today because `LicenseManager` only exists in React context.

So the prerequisite is getting a license somewhere an `Editor` can reach it. Opened #9704 to explore an imperative API for that, roughly `tldraw.setLicense({ licenseId: '...' })` as an alternative to passing a key through the React tree. Your "hanging the manager off the `Editor`" note is the same idea from the other direction.

Worth carrying over separately: the `tlschema` question at the end of the description. `TLComment`, `TLCommentThread`, and `commentSchemaRecords` living in a core package is the one item here with a real deadline, since it's a schema change that gets harder after ship.

Leaving this open for you to close — nothing regresses by not merging it, since the built-in commenting surfaces are already gated.

PR #9691, opened 2026-07-27
fix(editor): end stuck pans when the pointerup was missed

### LINKED ISSUE #9692: Pan gets stuck following the cursor after the button is released outside the window
When you start a right-click or middle-click pan and release the button outside the browser window, the browser can fail to deliver the `pointerup`. No `pointercancel` or `lostpointercapture` arrives either, even though the canvas takes pointer capture on pointerdown (see [w3c/pointerevents#407](https://github.com/w3c/pointerevents/issues/407) and [Mozilla bug 1684355](https://bugzilla.mozilla.org/show_bug.cgi?id=1684355)).

The editor still thinks the pan is active, so when the mouse re-enters the window the camera keeps following the cursor with no button held, until the user clicks to break out of it. The same applies to spacebar + left-click pans.

**Reproduction**

1. Start a right-click (or middle-click) pan and drag the cursor outside the browser window.
2. Release the button while the cursor is outside the window.
3. Move the mouse back into the window.

https://github.com/user-attachments/assets/b2e977ec-c232-41cb-8bc6-37fd0788a1c7

**Expected:** the pan ends as soon as we can tell the button is no longer down.

**Actual:** the camera keeps following the cursor until the user clicks again.

**Open questions**

1. Which browser and OS combinations was this seen on?
2. Did this come from user reports (e.g. on tldraw.com), or was it found while developing?

A fix is up in #9691, which ends the pan on the next `pointermove` whose `buttons` bitmask shows the pan button is no longer down.

MitjaBezensek:
In order to stop pans from getting stuck when the mouse button is released outside the browser window, this PR ends a pan on the next pointermove whose `buttons` bitmask shows the pan button is no longer down. Even with pointer capture set on pointerdown, browsers don't reliably deliver the `pointerup`, `pointercancel`, or `lostpointercapture` that would end the pan in that case (see [w3c/pointerevents#407](https://github.com/w3c/pointerevents/issues/407) and [Mozilla bug 1684355](https://bugzilla.mozilla.org/show_bug.cgi?id=1684355)), so the camera keeps following the cursor with no button held until the user clicks again. Closes #9692.

The recovery lives in the document-level `pointermove` listener in `useCanvasEvents`, which is the first thing to fire when the mouse re-enters the window. When a tracked pan button is missing from the [`buttons` bitmask](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons), it dispatches a synthetic `pointer_up` through the existing pointer path, so cursor restore, pan-state reset, the spacebar keep-panning case, and the right-button state-chart skip all run unchanged. One subtlety: macOS ctrl+click maps to button 2 while the physical bit stays 1 (left), so button 2 only counts as released when neither bit is set — covered by a dedicated test.

Tests drive real DOM events through the rendered component (the recovery is invisible to `TestEditor`/`Driver`, which bypass the DOM listener) and cover right-click, middle-click, spacebar+left, and darwin ctrl+click pans.

### Before

https://github.com/user-attachments/assets/379377d3-b81f-4e1b-bd94-76da773e9d6d

### After

https://github.com/user-attachments/assets/ca815644-420b-4f80-aec5-55d3ea3bc387

### Change type

- [x] `bugfix`

### Test plan

1. Start a right-click (or middle-click) pan and drag the cursor outside the browser window.
2. Release the button outside, then move the mouse back into the window.
3. The pan should end instead of the camera following the cursor.

- [x] Unit tests

### Release notes

- Fixed pans getting stuck following the cursor when the pan button was released outside the browser window.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +40 / -0   |
| Tests     | +248 / -0  |

### DIFF
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -1,5 +1,5 @@
 import { useValue } from '@tldraw/state-react'
-import React, { useEffect, useMemo } from 'react'
+import React, { useEffect, useMemo, useRef } from 'react'
 import { LEFT_MOUSE_BUTTON, MIDDLE_MOUSE_BUTTON, RIGHT_MOUSE_BUTTON } from '../constants'
 import { tlenv } from '../globals/environment'
 import {
@@ -21,9 +21,9 @@ function isButtonStillDown(button: number, buttons: number): boolean {
 		case MIDDLE_MOUSE_BUTTON:
 			return (buttons & 4) !== 0
 		case RIGHT_MOUSE_BUTTON:
-			// On darwin, ctrl+click reports button 2 while the physical bit is
-			// the left button's, so right still counts as down while either bit
-			// is set.
+			// On darwin, ctrl+click is mapped to button 2 (getPointerEventButton),
+			// but browsers disagree on the buttons bit: some report the physical
+			// left bit, others the translated right bit. Accept either.
 			return tlenv.isDarwin ? (buttons & (1 | 2)) !== 0 : (buttons & 2) !== 0
 		default:
 			// Other buttons (stylus eraser etc.) are never treated as stale.
@@ -35,15 +35,16 @@ export function useCanvasEvents() {
 	const editor = useEditor()
 	const ownerDocument = editor.getContainerDocument()
 	const currentTool = useValue('current tool', () => editor.getCurrentTool(), [editor])
+	// Shared with the document-level pointermove listener below so stale-button
+	// recovery can clear it when the pointerup it describes never arrived.
+	const isSecondaryClickPointerDown = useRef(false)
 
 	const events = useMemo(
 		function canvasEvents() {
-			let isSecondaryClickPointerDown = false
-
 			function onPointerDown(e: React.PointerEvent) {
 				if (editor.wasEventAlreadyHandled(e)) return
 				const button = getPointerEventButton(e)
-				isSecondaryClickPointerDown = button === 2
+				isSecondaryClickPointerDown.current = button === 2
 
 				// With right-click panning disabled, fire right_click on press and let the
 				// native contextmenu through so the menu opens at the pointer-down location.
@@ -76,7 +77,7 @@ export function useCanvasEvents() {
 
 			function onPointerUp(e: React.PointerEvent) {
 				if (editor.wasEventAlreadyHandled(e)) return
-				const button = isSecondaryClickPointerDown ? 2 : getPointerEventButton(e)
+				const button = isSecondaryClickPointerDown.current ? 2 : getPointerEventButton(e)
 				if (button !== 0 && button !== 1 && button !== 2 && button !== 5) return
 
 				const rightClickPanning = editor.options.rightClickPanning
@@ -108,7 +109,7 @@ export function useCanvasEvents() {
 					})
 					e.currentTarget.dispatchEvent(contextMenuEvent)
 				}
-				isSecondaryClickPointerDown = false
+				isSecondaryClickPointerDown.current = false
 			}
 
 			function onPointerEnter(e: React.PointerEvent) {
@@ -257,14 +258,21 @@ export function useCanvasEvents() {
 			// through the normal pointer_up path before dispatching this move.
 			if (
 				e.pointerType === 'mouse' &&
-				// bare Event objects lack a buttons field entirely; skip the check
-				// for those (a real PointerEvent always has a numeric buttons)
+				// polyfilled or synthetic events can lack a numeric buttons field;
+				// skip the check for those
 				typeof e.buttons === 'number' &&
 				(editor.inputs.getIsPanning() || editor.inputs.getIsRightPointing()) &&
 				editor.inputs.getIsPointing()
 			) {
 				for (const button of Array.from(editor.inputs.buttons.keys())) {
 					if (isButtonStillDown(button, e.buttons)) continue
+					if (button === RIGHT_MOUSE_BUTTON) isSecondaryClickPointerDown.current = false
+					// The missed pointerup also means the capture taken on
+					// pointerdown was never explicitly released. Browsers usually
+					// drop it implicitly, but if it's still held this event was
+					// retargeted to the captured element, so release it there.
+					const captureTarget = e.target as Element | null
+					if (captureTarget?.hasPointerCapture) releasePointerCapture(captureTarget, e)
 					editor.dispatch({
 						type: 'pointer',
 						target: 'canvas',
--- a/packages/editor/src/lib/hooks/useCanvasEvents.ts
+++ b/packages/editor/src/lib/hooks/useCanvasEvents.ts
@@ -228,14 +228,17 @@ export function useCanvasEvents() {
 			lastX = e.clientX
 			lastY = e.clientY
 
-			// A pan button released outside the window can lose its pointerup —
-			// no pointercancel or lostpointercapture fires either — leaving the
-			// pan stuck following the cursor. If the buttons bitmask says a
-			// tracked pan button is no longer down, end it through the normal
-			// pointer_up path before dispatching this move.
+			// A pan button released outside the window can lose its pointerup:
+			// even with pointer capture set on pointerdown, browsers don't
+			// reliably deliver the pointerup, pointercancel, or
+			// lostpointercapture that would end it (w3c/pointerevents#407),
+			// leaving the pan stuck following the cursor. If the buttons
+			// bitmask says a tracked pan button is no longer down, end it
+			// through the normal pointer_up path before dispatching this move.
 			if (
 				e.pointerType === 'mouse' &&
-				// synthetic/test events may omit buttons entirely; skip the check for those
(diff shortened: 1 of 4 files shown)


kaneel (approved):
good fix

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -228,6 +228,46 @@ export function useCanvasEvents() {
...
+			// lostpointercapture that would end it (w3c/pointerevents#407),
+			// leaving the pan stuck following the cursor. If the buttons
+			// bitmask says a tracked pan button is no longer down, end it
+			// through the normal pointer_up path before dispatching this move.
+			if (
+				e.pointerType === 'mouse' &&
+				// bare Event objects lack a buttons field entirely; skip the check
+				// for those (a real PointerEvent always has a numeric buttons)
+				typeof e.buttons === 'number' &&
+				(editor.inputs.getIsPanning() || editor.inputs.getIsRightPointing()) &&
```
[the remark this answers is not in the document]
mimecuvalo: can you check this one @MitjaBezensek ?

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -228,6 +228,46 @@ export function useCanvasEvents() {
...
+				editor.inputs.getIsPointing()
+			) {
+				for (const button of Array.from(editor.inputs.buttons.keys())) {
+					// Button 0 (left) → bit 1, button 1 (middle) → bit 4, button 2
+					// (right) → bit 2. On darwin, ctrl+click maps to button 2 while
```
mimecuvalo: is this meant to check Darwin?

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -228,6 +228,46 @@ export function useCanvasEvents() {
...
+					// the physical bit is 1 (left), so button 2 only counts as stale
+					// when neither bit is set.
+					const stale =
+						button === 0
+							? !(e.buttons & 1)
+							: button === 1
+								? !(e.buttons & 4)
```
mimecuvalo: this is a hell of a ternary :P

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -228,6 +228,46 @@ export function useCanvasEvents() {
...
+				editor.inputs.getIsPointing()
+			) {
+				for (const button of Array.from(editor.inputs.buttons.keys())) {
+					// Button 0 (left) → bit 1, button 1 (middle) → bit 4, button 2
+					// (right) → bit 2. On darwin, ctrl+click maps to button 2 while
+					// the physical bit is 1 (left), so button 2 only counts as stale
+					// when neither bit is set.
+					const stale =
+						button === 0
+							? !(e.buttons & 1)
```
mimecuvalo: we do have constants in editor constants.ts for mouse buttons

mimecuvalo (commented):
couple comments before landing

mimecuvalo (changes requested):
blocking just so you see my comments

packages/editor/src/lib/hooks/useCanvasEvents.ts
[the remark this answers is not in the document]
MitjaBezensek: Confirmed and fixed in d39aac0ddb. The synthesized pointer_up while right-pointing did fall through to the state chart as right_click, so recovery could change or clear the selection at the re-entry point. Editor's pointer_up now suppresses the right_click fall-through when the release lands beyond the drag threshold — a real interactive release can't do that, since moves past the threshold become a pan first. Added a regression test that a selected shape stays selected across recovery.

packages/editor/src/lib/hooks/useCanvasEvents.ts
MitjaBezensek: Yes — it kept darwin ctrl+drag panning alive (ctrl+click reports button 2 while the physical buttons bit is the left one). It's now explicitly gated on tlenv.isDarwin, so other platforms recover a stale right button even while the left button is held; added a test for that case. (d39aac0ddb)

packages/editor/src/lib/hooks/useCanvasEvents.ts
MitjaBezensek: Fair :) refactored into an isButtonStillDown helper with a switch in d39aac0ddb.

packages/editor/src/lib/hooks/useCanvasEvents.ts
MitjaBezensek: Now using LEFT/MIDDLE/RIGHT_MOUSE_BUTTON in the new helper (d39aac0ddb).

MitjaBezensek:
> blocking just so you see my comments

Addressed the comments, thanks for feedback!

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -228,6 +249,40 @@ export function useCanvasEvents() {
...
+			// through the normal pointer_up path before dispatching this move.
+			if (
+				e.pointerType === 'mouse' &&
+				// polyfilled or synthetic events can lack a numeric buttons field;
+				// skip the check for those
+				typeof e.buttons === 'number' &&
+				(editor.inputs.getIsPanning() || editor.inputs.getIsRightPointing()) &&
+				editor.inputs.getIsPointing()
+			) {
+				for (const button of Array.from(editor.inputs.buttons.keys())) {
```
mimecuvalo, line 340: ### No record that the gesture was recovered — useCanvasEvents.ts:267 (medium)
Recovery clears isPanning but leaves no marker, so a pointerup that does eventually arrive for the same pointer is processed as a fresh static right-click. Chrome delivers a pointermove with buttons: 0 just before pointerup when the pointer moves during release; then onPointerUp computes wasRightClickPanning = ... && getIsPanning() → false and fires the synthetic contextmenu — the menu pops open at the end of a right-drag pan, the exact case the code otherwise suppresses. It also pushes a second pointer_up into the state chart. Same hole at line 269 for isSecondaryClickPointerDown.current = false: on macOS, releasing ctrl before the button makes the late pointerup resolve to button 0. Fix: record the recovered pointerId/button and have onPointerUp ignore the match.

MitjaBezensek: Fixed in af8277a900. Recovery now records the synthesized pointer_up per pointer, and a matching late real pointerup is swallowed instead of being dispatched as a fresh static right-click; any document-level pointerdown clears the record so a new gesture's pointerup can't be swallowed. The record is keyed per editor rather than per hook instance — useCanvasEvents runs in both Canvas and MenuClickCapture, and recovery can fire in one while the pointerup lands in the other. On darwin the match treats buttons 0 and 2 as the same physical button, covering ctrl released before or after the late pointerup in both directions.

Two adjacent holes surfaced while testing this, fixed in the same commit: MenuClickCapture's handlePointerUp overrides the spread canvasEvents.onPointerUp, so a late pointerup landing on the overlay would have re-dispatched pointer_up(2) and reopened the menu — it now consults the same record; and its handlePointerMove kept forwarding moves after a missed pointerup, panning the camera to the re-entry point — it now stops forwarding when the tracked button is gone from the buttons bitmask, letting the document-level recovery end the gesture.

Tests cover the late pointerup on canvas and on the overlay, both darwin resolutions, and a fresh right-click after recovery. The assertions capture dispatch names at call time (Editor.dispatch mutates info.name in place, so inspecting spy args after the fact is vacuous) and were mutation-checked against the swallow being disabled.

packages/editor/src/lib/hooks/useCanvasEvents.ts
```diff
@@ -56,9 +126,16 @@ export function useCanvasEvents() {
...
 				if (editor.wasEventAlreadyHandled(e)) return
-				const button = isSecondaryClickPointerDown ? 2 : getPointerEventButton(e)
+				const button = isSecondaryClickPointerDown.current ? 2 : getPointerEventButton(e)
 				if (button !== 0 && button !== 1 && button !== 2 && button !== 5) return
 
+				// Recovery already synthesized this button's pointer_up.
+				if (consumeRecoveredPointerUp(editor, e.pointerId, button)) {
+					releasePointerCapture(e.currentTarget, e)
+					isSecondaryClickPointerDown.current = false
+					return
```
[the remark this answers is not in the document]
MitjaBezensek: Fixed in f30f40f305. Recovery now mirrors onPointerUp: when it consumes a stale button 2 while right-pointing (not panning) and the release is within the screen-space drag threshold — the same condition under which the editor delivers right_click — it fires the synthetic contextmenu on the canvas itself. Selection update and menu stay consistent, the late real pointerup stays swallowed, and beyond the threshold neither fires. Tests cover both: menu opens once for a mid-release stale move (mutation-checked against the branch being disabled) and no menu when recovery lands far from the press point.

MitjaBezensek:
Decided to drop this for now.

Issue #9638, opened 2026-07-21
Make it possible to upload .tldraw bundles to tldraw.com


a10y:
### What's the feature?

Offline tldraw allows you to save in .tldraw format (zipped SQLite) but AFAICT there is not a way to import .tldraw files into the web app, only .tldr files.

It would be really slick if one/either of 

- Import .tldraw in tldraw.com
- Upload to tldraw.com from offline desktop app

worked!

### Where would you like this feature?

tldraw.com

PR #9694, opened 2026-07-27
improvement: acknowledge .tldraw files from tldraw offline

frolic:
People are saving files out of tldraw offline and trying to open them on tldraw.com — see #9638. Those are `.tldraw` files, which are archives (a zipped SQLite database plus assets) rather than the JSON a `.tldr` file holds, so nothing here can read one. Until support lands, this recognizes the extension and says so.

Today the dead ends are silent or misleading: dropping a `.tldraw` on the dotcom canvas gives a generic "file type is not allowed" error, dropping one on the sidebar does nothing at all, and VS Code falls back to its "file is binary" placeholder. None of them mention tldraw.

This only detects the extension — it does no parsing. Reading the archive is separate work in the `tldr-archive` package.

In dotcom, a small hook drops tldraw offline files from a list, toasts once, and returns the files we can still handle, so a mixed drop still imports the rest and doesn't stack a second generic toast on top.

In VS Code this registers a **separate** read-only view type rather than adding `*.tldraw` to the existing `tldraw.tldr` selector. Going through `TldrawEditorProvider` would read the archive's bytes as UTF-8 and hand them to `loadFile()`, which returns a deliberately invalid record — so the user would get "the file you tried to open is corrupted" on a mounted canvas. The new provider never reads the file.

Two known gaps, both deliberate:

- Signed-out users on a shared room won't see the message — all four `SneakyTldrawFileDropHandler` mount sites are gated on `app`. Closing that means registering a files handler for anonymous users where none exists today, which risks their image drops for the sake of a temporary message. Signed-in paths and the local editor are covered.
- Dropping a `.tldraw` onto an already-open canvas *inside* VS Code still hits the SDK's generic toast; the view type only covers opening one from the explorer.

### Change type

- [x] `improvement`

### Test plan

1. Save a file from tldraw offline so you have a `.tldraw` file.
2. On tldraw.com, drag it onto the canvas — expect "Can't open .tldraw files yet".
3. Drag it onto the sidebar — same message (previously nothing happened).
4. Drag it in alongside a PNG — the message appears and the PNG still imports.
5. Use "Import file…" in the user menu; the `.tldraw` file is selectable and shows the message.
6. Confirm a normal `.tldr` file still imports everywhere, and that dropping an image is unchanged.
7. In VS Code, open a `.tldraw` file from the explorer — expect the message instead of the binary placeholder. Confirm `.tldr` files still open the canvas.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Explain that `.tldraw` files from tldraw offline aren't supported yet, instead of failing with a generic error.

### Code changes

| Apps            | +162 / -12 |
| Automated files | +18 / -0   |


frolic:
looks like:

image


 happy to adjust messaging

frolic:
on dotcom:

image

apps/dotcom/client/src/tla/components/dialogs/TlaTldrawOfflineFileDialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+			<TldrawUiDialogHeader>
+				<TldrawUiDialogTitle>
+					<F defaultMessage="Can’t open .tldraw files yet" />
+				</TldrawUiDialogTitle>
+				<TldrawUiDialogCloseButton />
+			</TldrawUiDialogHeader>
+			<TldrawUiDialogBody style={{ maxWidth: 350 }}>
+				<p>
+					<F
+						defaultMessage="We’re working on support for files from <a>tldraw offline</a>."
```
mimecuvalo: b/c 'tldraw offline' as an app name/brand name, let's actually prbly pull that out so it doesn't accidentally get translated

mimecuvalo (approved):

Issue #9520, opened 2026-07-07
Cursor chat gives the author no feedback when their message is truncated for collaborators


kostyafarber:
> currently the cursor chat experience is not great, the text overflows from perspective of collobaratoes but for me I don't know when this happens, meaning my message gets truncated and I'm not aware of it

When you type a cursor chat message in a multiplayer room, collaborators see your message cut off with an ellipsis once it's longer than the bubble allows — but your own bubble keeps growing and always shows the full text. Nothing on your side signals that part of the message is invisible to everyone else. Separately, the input silently stops accepting characters past a fixed length, again with no cue. The author should either see the same thing collaborators see, or get a clear signal (wrap, ellipsis, counter) when their message is being clipped.

## Steps to reproduce

1. Open a multiplayer room with two clients (e.g. tldraw.com in two browser windows with different users).
2. In client A, press `/` to start cursor chat.
3. Type a message longer than ~30–40 characters.
4. Compare what client A sees (full message in an auto-growing bubble) with what client B sees (bubble capped with `…`).
5. Keep typing past 64 characters in client A — the input silently stops accepting characters.

## Expected behavior

The author can tell when their message won't be fully visible to collaborators.

## Actual behavior

- The author's bubble grows unboundedly and always shows the full message.
- Collaborators see the message ellipsized at a fixed bubble width.
- Typing past 64 characters silently drops input.

## Open questions

1. Should long messages wrap onto multiple lines for everyone (no information loss), or is it enough for the author's bubble to truncate the same way collaborators see it?
2. Is the 64-character input cap intentional product behavior to keep, or should the limit change as part of this fix?

Confidence: 85%, ready to get started.

kostyafarber:
# Research findings

## Root cause (short version)

Cursor chat has two independent renderers that disagree on how a message is displayed:

1. **Author's own bubble** (DOM, `CursorChatBubble.tsx`) auto-grows to fit the *full* message with no width cap.
2. **Collaborator bubbles** (canvas, `CollaboratorCursorOverlayUtil.ts`) hard-truncate with an ellipsis at a fixed `chatMaxWidth: 200` screen px.

Additionally the author-side 64-char cap (`value.slice(0, 64)`) is applied *only to local `useState`*, **not** to the value that is broadcast — so the two limits (64 chars local display vs. 200px ellipsized remote display) are unrelated and neither is surfaced to the author. The author literally cannot tell what collaborators see.

[...]

PR #9695, opened 2026-07-27
editor: move collaborator cursors to a DOM layer above the canvas

### LINKED ISSUE #9693: Bring back the collaborator cursors component as a DOM layer
🗣️: separate cursors into a new (separate) layer, perhaps

🗣️: ahh lets modify this, I want to bring back the Collaborator Cursors component as it was before we moved that to the canvas element

---

Collaborator cursors used to render as React DOM elements, with an overridable component that SDK users could swap out. When the canvas overlay system landed, cursors moved into the shared overlay canvas along with the other ephemeral UI. The ask is to bring the collaborator cursors component back as it was: DOM-rendered, overridable, and living in its own layer that sits in front of the canvas element and behind the "in front of the canvas" component. What moves back is the cursor and the things attached to it — name labels, cursor chat, and similar. Scribbles, viewport-edge hint arrows, and shape indicators stay on the canvas. The canvas overlay versions of whatever moves back should be removed rather than kept alongside the DOM ones.

## Open questions

1. **Critical: What do we want back, specifically — the overridable cursor component, DOM-based rendering, or both?**
   Both. It needs to be a component that sits in front of the canvas and behind "in front of the canvas".
2. **Should this restore the old public API exactly as it was, or take a new shape?**
   The same API shape is okay.
3. **Does this cover only cursors, or also the other presence pieces that moved to the canvas — name labels, cursor chat, hint arrows, collaborator brushes, scribbles, and shape indicators?**
   The cursor itself plus everything attached to it — name labels, cursor chat, and so on. Scribbles, the viewport-edge hint arrows, and indicators stay on the canvas.
4. **Should the canvas versions of whatever moves back be removed, or should both paths coexist?**
   Remove the canvas versions.
5. **Do collaborator brushes move back to DOM with the cursor, or stay on the canvas with the scribbles and indicators?**
   _Awaiting answer._

Confidence: 84%, ready to get started.

jsscclr:
Closes #9693.

Collaborator cursors — the arrow, name tag, and chat message — leave the canvas overlay system and render as DOM elements in a dedicated layer: above all canvas content and the comments layer, below the in-front content and the UI panels. Per the issue's scoping, the viewport-edge hint arrows stay canvas-drawn with the scribbles and indicators.

## Why

The overlay-util migration put two different kinds of chrome on one canvas. Drawing chrome — brushes, scribbles, snap lines, indicators, hints — repaints with every canvas frame and belongs there. Cursors are **presence chrome**: they carry styled text (names, chat), want CSS for layout and theming, and are the thing consumers most often customize. On the canvas they could only be customized by reimplementing 2D draw code; as DOM they're components again.

This is also what makes the commenting layering work without touching commenting behavior. The comments layer is DOM portaled to the container; canvas-drawn cursors could never paint above it, no matter their overlay `zIndex`. With cursors as DOM, the stack reads straight off the element ladder:

```
canvas (shapes + drawing chrome) → comments layer (collaborators − 5, 240) →
cursors (--tl-layer-collaborators, 245) → watermark (248) → in-front (250) → panels (300) → popovers (400)
```

Cursors ride over comment pins; the watermark (a licensing element) paints over both so it can't be occluded; the rich-text/image toolbars and the local cursor-chat input paint over cursors; open thread panels cover everything. The comments layer previously borrowed `--tl-layer-canvas-in-front` — it now derives its z from the cursor slot (`calc(var(--tl-layer-collaborators) - 5)`), so the cursors-over-pins invariant is structural rather than two numbers that happen to be adjacent.

## What's in the PR

- **`.tl-collaborators` layer** at `--tl-layer-collaborators` (245), documented in the `editor.css` layer comment. An inner html-layer carries the camera transform using the same formula as the canvas layers, so cursor components position in page space. The transform is written both on camera change (reactor) and on layer (re)attach (callback ref) — the layer unmounts while no collaborators are visible.
- **`LiveCollaborators`** (not exported — internal, like `Shape`/`CanvasOverlays`) mounts the layer. Visibility (activity state, following, highlighting) comes from `editor.getVisibleCollaboratorsOnCurrentPage()`. Per-cursor movement writes `transform` directly via `useTransform`, avoiding a per-frame React render *for the transform itself*; presence changes still re-render the affected `Collaborator` component (and camera frames re-render each one, since they read zoom/viewport), bounded by collaborator count. This is parity with the pre-#8633 DOM implementation. Off-viewport collaborators are the canvas hint util's job (`isCursorInViewport`, now a shared predicate so cursor and hint stay mutually exclusive); the layer renders in-viewport cursors only.
- **`DefaultCursor` returns as a component**, with a `CollaboratorCursor` slot on `TLEditorComponents` — restoring the pre-migration customization surface. Implementation resurrected from the pre-migration tree (two resurrection bugs found and fixed: a renamed CSS variable in the hint outline, and the remount transform gap above).
- **`CollaboratorCursorOverlayUtil` removed** from the default overlay utils and the public API, with its tests. `CollaboratorHintOverlayUtil` stays, per the issue.
- **Remote chat bubbles no longer truncate** — the canvas-era 200px `chatMaxWidth` was a port necessity, not a product decision. The input caps messages at 64 characters, so bubbles are bounded, sender and receivers see the same text, and #9520's silent-truncation complaint no longer applies.
- **Minimap** draws collaborator cursor dots natively (it previously relied on the cursor util's `renderMinimap`).
- **Docs**: `cursors.mdx` documents the component slot; the overlay-utils table and `cursor-chat.mdx` updated.

## Verification

- Full `tsc --build` clean; editor, tldraw, and commenting test suites pass; API reports regenerated; lint clean.
- Added a `isCursorInViewport` unit test (the shared cursor/hint boundary); the `DefaultCursor`/`LiveCollaborators` render paths and the minimap dot loop are still uncovered.
- Verified live (examples app, synthetic collaborator): DOM cursor renders (arrow + name tag + untruncated chat); the layer is a sibling of `.tl-canvas` carrying the camera transform; the computed ladder is comments 240 < cursors 245 < watermark 248 < in-front 250; and a collaborator cursor paints **over** a comment pin (the target fix) while the watermark paints over the cursor. — cursor/chat visuals, the layering (cursor above pins, below toolbars), and hint arrows at viewport edges need an eye on the commenting + user-presence examples.

## Notes

- This partially reverses #8633 for cursors only. The overlay-util system is unchanged for everything else; hints, collaborator brushes, scribbles, and indicators stay canvas.
- One commenting file is touched, deliberately: the comments layer's `z-index` moves from borrowing `--tl-layer-canvas-in-front` to deriving from the cursor slot. That change is what reconciles the issue's "cursors behind in-front" with "cursors above comment pins" — they were contradictory while comments shared in-front's z. Pin/thread behavior is unchanged.
- Perf: the pre-migration DOM cursors updated via direct style writes outside React, preserved here. Per-cursor React re-render on presence change is bounded by collaborator count, not shape count.

### API changes

**Added (`@tldraw/editor`):**

- `LiveCollaborators` — the DOM collaborator cursor layer component.
- `DefaultCursor` + `TLCursorProps` — the default cursor component (restored from the pre-overlay-util API).
- `CollaboratorCursor` component slot on `TLEditorComponents`.
- `useTransform` — position an element by writing its transform directly (restored).
- New CSS layer variable `--tl-layer-collaborators` (245).

**Removed (`tldraw`) — breaking:**

- `CollaboratorCursorOverlayUtil` and `TLCollaboratorCursorOverlay`; the util is gone from `defaultOverlayUtils`. Consumers who subclassed it to customize cursors should move to the `CollaboratorCursor` component slot.

**Changed (`@tldraw/commenting`):**

- `.tlui-cmt-canvas-layer` z-index now derives from `--tl-layer-collaborators` instead of borrowing `--tl-layer-canvas-in-front`. No API surface change.


mimecuvalo (commented):
> 🤖 **AI-generated review** (Claude Code, requested by @mimecuvalo). Findings were checked against the PR branch, its base (`commenting`), and the pre-#8633 tree, but please treat the conclusions as a starting point rather than a verdict.

## Overview

This reverts the cursor half of #8633: the collaborator arrow, name tag, and chat move from the overlay canvas back to a DOM layer, restoring the `CollaboratorCursor` component slot, `DefaultCursor`/`TLCursorProps`, and `useTransform`. Hint arrows, brushes, scribbles, and indicators stay canvas. The minimap grows a native collaborator-dot loop to replace the removed `renderMinimap`.

The motivation is sound and the diff is a genuine net simplification (−718/+358): ~300 lines of hand-transcribed `Path2D` arrow geometry and a canvas `measureText` truncation cache with LRU-ish eviction all go away, replaced by CSS `text-overflow: ellipsis`. The layering rationale — deriving the comments layer from `calc(var(--tl-layer-collaborators) - 5)` — is well argued and correctly documented in the CSS comment.

I diffed the resurrected files against `17b19dc637^`. `DefaultCursor.tsx`, `useTransform.ts`, the `.tl-collaborators__item` / `.tl-cursor` / `.tl-collaborator__cursor` CSS, and the `isCursorInViewport` predicate are all faithful ports. Good discipline.

One piece of context worth stating explicitly, since it shapes several notes below: **before #8633 this was already DOM**, so the PR is largely a resurrection. The structural difference is *where the layer lives*. `.tl-overlays` was a child of `.tl-canvas` at `--tl-layer-canvas-overlays: 500`; since `.tl-canvas` has `contain: strict` (its own stacking context) and `z-index: auto`, everything inside painted below the container-level layers. The new `.tl-collaborators` is a **sibling** of `.tl-canvas` at 245. That relocation is what makes the commenting layering work — and it has one side effect worth a decision (item 4).

## Issues

### 1. Docs: the `CollaboratorHintOverlayUtil` row was deleted by mistake

`apps/docs/content/sdk-features/overlay-utils.mdx` drops **two** rows, but only one util was removed:

```diff
-| [CollaboratorCursorOverlayUtil](?)         | Remote users' cursors, name tags, and chat    |
 | [CollaboratorBrushOverlayUtil](?)          | Remote users' brush rectangles                |
 | [CollaboratorScribbleOverlayUtil](?)       | Remote users' scribbles                       |
-| [CollaboratorHintOverlayUtil](?)           | Remote users' viewport hints                  |
```

`CollaboratorHintOverlayUtil` is still in `defaultOverlayUtils` and still exported — the PR description says so explicitly, and `cursors.mdx` even links to it. The table now under-documents the defaults. Restoring the row should be all that's needed.

### 2. The viewport-cull predicate is duplicated across two packages

`LiveCollaborators.tsx` (editor):

```ts
const isCursorInViewport = !(
	cursor.x < viewportPageBounds.minX - 12 / zoomLevel || …
)
```

`CollaboratorHintOverlayUtil._isCursorInViewport` (tldraw) has the byte-identical expression. They are correct *and* mutually exclusive today, but only by coincidence — pre-#8633 this was a single boolean feeding one `if/else`, so exclusivity was structural. Now, changing the margin in one place yields either a dead zone (a peer shows neither cursor nor hint) or a double-render (both). Consider exporting a shared helper from `@tldraw/editor` and having the hint util call it.

Related, and pre-existing rather than a regression (it is in both the old and the new code): the max-edge checks subtract the margin (`cursor.x > maxX - 12 / zoom`) where the comment says the padding is for the top-left. That hides cursors within 12/16px of the right/bottom edge slightly earlier than intended. Fine to leave here, but it might deserve a follow-up issue.

### 3. `aria-hidden="true"` was dropped from the cursor `<svg>`

```tsx
// pre-#8633
<svg className="tl-cursor" aria-hidden="true">
// this PR
<svg className="tl-cursor">
```

The `<use>` element has no accessible name, so this is a bare unlabeled graphic in the accessibility tree. The `<defs>` svg in `LiveCollaborators` in this same PR *does* set `aria-hidden`, so it reads as an accidental drop during resurrection.

### 4. Cursors now paint over the watermark

`--tl-layer-collaborators: 245` versus `--tl-layer-watermark: 200`, both in the container stacking context. Both before *and* after #8633, cursors lived inside `.tl-canvas` (`contain: strict`, `z-index: auto`) and therefore painted **below** the watermark. This is new behavior, and the watermark is a licensing-enforcement element.

If it is deliberate, it is worth saying so in the PR description. Otherwise 245 wants to be below 200, which then conflicts with the comments-layer derivation and needs rethinking. This is the one item I would want resolved before merge.

### 5. The perf claim in the description is stronger than the code

> Per-cursor movement writes `transform` directly via `useTransform`, so pointer-frequency updates never re-render more than the moved cursor.

`Collaborator` is `track`ed and reads `editor.getZoomLevel()` and `editor.getViewportPageBounds()`, so **every camera frame re-renders every collaborator component**. And `LiveCollaborators` reads the whole `getVisibleCollaboratorsOnCurrentPage()` array, so any peer's move re-renders the parent. (Children do memo-bail on stable record identity, which holds — store records are immutable.)

This is parity with pre-#8633, so it is not a regression, but the claim as written is only true of the transform write, not of React renders. Worth softening so nobody later relies on it.

### 6. No tests added; 343 deleted

Nothing covers `LiveCollaborators`, `DefaultCursor`, the cull predicate, the `'New User' → null` name mapping, or the new minimap dot loop.

The visibility filtering the old tests exercised (idle / inactive / following / highlighted) is still covered by `CollaboratorsManager.test.ts`, so that is not lost. But the cull boundary and the name mapping now have zero coverage, and they are exactly the kind of thing that drifts — see item 2. A small `LiveCollaborators.test.tsx` asserting cursor-in / out-of-viewport and the `'New User'` case would close the gap cheaply.

### 7. `cameraLayerTransform` duplicates `DefaultCanvas`'s camera formula

The `modulate(z, [1, 8], [0.125, 0.5], true)` / `modulate(z, [0.1, 1], [-2, 0.125], true)` offset now exists verbatim in two files. The comment honestly says "the same formula as the canvas's html layers" — which is the argument for extracting it rather than restating it. A one-line `getHtmlLayerTransform(editor)` called from both removes a silent-drift class of bug.

### 8. `.tl-collaborator__cursor`'s z-index is now dead

```css
.tl-collaborator__cursor { z-index: var(--tl-layer-overlays-collaborator-cursor); } /* 130 */
```

Every cursor is a sibling inside `.tl-html-layer` with the same z-index, so this orders nothing. It is also a variable from the `--tl-layer-overlays-*` namespace — the canvas overlay stack, which cursors have now left. Faithful to the original, but in the original it sat alongside `--tl-layer-overlays-collaborator-cursor-hint` in a shared stack where it did matter. Either drop the rule or leave a note explaining why it is kept.

## Smaller notes

- **`LiveCollaborators` as `@public`**, typed `NamedExoticComponent<unknown>`. A consumer cannot do much with it — no props, must live inside the canvas, and `DefaultCanvas` already mounts it. `@internal` seems more honest, and it avoids owing stability to a component nobody should be mounting themselves.
- **The docs example contradicts its own prose.** `CustomCollaboratorCursor` in `cursors.mdx` applies `translate(...)` with no `scale(1 / zoom)`, so the 16px dot grows and shrinks with zoom — while the paragraph directly beneath says sizes stay constant on screen by scaling with `1 / zoom`. Either add the scale to the example or drop `zoom` from the sentence.
- **`DefaultCursor` is public but depends on `<CursorDef>`**, which only `LiveCollaborators` renders. Used standalone, `<use href="#…cursor">` resolves to nothing and the arrow is invisible. Worth a line in the tsdoc.
- **`Math.PI * 2` in `MinimapManager`** — the deleted `renderMinimap` used the exported `PI2`, which is already imported into that package's overlay files. Nit.
- **Breaking change, unlabeled.** `CollaboratorCursorOverlayUtil` and `TLCollaboratorCursorOverlay` leave `tldraw`'s public API. Given the base here is the `commenting` branch rather than `main`, the label and release note may be intended for the eventual merge — flagging so it does not get lost.
- **API report import style.** The report picks up a bare `import { NamedExoticComponent } from 'react'` for the two new symbols, while the existing `TldrawEditor` entry uses `React_3.NamedExoticComponent`. That is just api-extractor's import resolution in a generated file, not worth chasing. (For context: `TldrawEditor` was previously the only publicly-exported memo'd component in `@tldraw/editor` — `Shape`, `InnerShape`, `CanvasOverlays`, and `Watermark` are all memo'd but not exported from the index, which is why the type had barely appeared before.)

## Verdict

The direction is right and the execution is careful — the resurrection is faithful, the layering argument is genuinely good, and the deleted canvas code was carrying real complexity.

Items 1 and 3 are straightforward fixes. **Item 4 (watermark) needs a decision** before this lands. Items 2, 6, and 7 are the ones worth pushing hardest on for maintainability, since the description flags that browser verification has not happened yet — which is exactly the situation where a duplicated cull predicate bites.

Both API reports are regenerated and consistent with the source, and no stale references to the removed util remain anywhere except the historical `v5.0.0.mdx` release notes, which is correct.

PR #9700, opened 2026-07-27
docs: restore shape meta documentation

frolic:
In order to give the shape `meta` property a home on tldraw.dev again, this PR restores a Meta section to the shapes guide. The SDK docs restructure dropped the old "Meta information" section, leaving `meta` and `Editor#getInitialMetaForShape` undocumented while the meta examples still linked to the dead `/docs/shapes#Meta-information` anchor.

The restored section covers the `meta` field on records (shapes, pages, bindings, assets, the document record), typing it with an intersection, setting initial values with `Editor#getInitialMetaForShape`, keeping it current with a before-change side effect, and validating it at runtime with `createTLSchema` meta validators. The sdk-features shapes page gains a one-line cross-reference (its `TLBaseShape` overview omitted `meta` entirely), and the two meta example comments now point at the restored `/docs/shapes#Meta` anchor. The anchor is capitalized because the docs slugger preserves heading case.

Every code sample was verified against the current `packages/editor` and `packages/tlschema` source. Note that meta validators go through `createTLSchema`, not ShapeUtil statics: the editor's schema builder only forwards `props` and `migrations` from shape utils, so the schema route documented here is the only one that works.

### Change type

- [x] `improvement`

### Test plan

1. Run `yarn dev-docs` and open `localhost:3001/docs/shapes#Meta` — the page scrolls to the new section and all API links resolve.
2. `yarn check-links` in `apps/docs` passes (validates the new internal links and the autolinked `(?)` API references).

### Release notes

- Docs: restore documentation for the `meta` property on shapes and other records, including `Editor#getInitialMetaForShape`, side-effect updates, and meta validators.

### Code changes

| Documentation | +94 / -3   |

### DIFF
--- a/apps/docs/content/docs/shapes.mdx
+++ b/apps/docs/content/docs/shapes.mdx
@@ -9,6 +9,8 @@ keywords:
   - shapes
   - shapeutils
   - utils
+  - meta
+  - metadata
 ---
 
 In tldraw, a shape is something that can exist on the page, like an arrow, an image, or some text. This article provides an overview of shapes and how to create custom ones.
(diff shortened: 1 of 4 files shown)


mimecuvalo (approved):
thx! this is a good shout

PR #9702, opened 2026-07-27
fix(dotcom): report worker exceptions that were being dropped

driev:
In order to find out why some dotcom worker failures never showed up in Sentry, this audits the four workers' Sentry integration and fixes the paths where exceptions were being dropped.

The sync-worker queue consumer swallowed every failure, so a message that exhausted its retries arrived in the dead letter queue with no record of why. The two branches want different treatment: asset-upload retries the same failing insert on every delivery, so it reports on the final one — the delivery that lands it in the DLQ — rather than filing eleven copies of one issue. og-image reports immediately, because `handleOgImageRenderMessage` already reports the render failures it catches, so anything reaching the batch loop's guard escaped the handler entirely, and an escape that doesn't recur is acked by the handler's own budget long before any final delivery.

`TLFileDurableObject` has four entry points the runtime drops errors from. The WebSocket handlers recognised `RoomNotFoundError` and rethrew everything else, so a `getRoom()` failure while loading the room from storage broke the main sync path and left no trace; `webSocketError` discarded its `error` argument entirely. `alarm()` and the constructor's `blockConcurrencyWhile` had no handler at all.

Those last two only run on a cold start, which is why they didn't matter until #9978 let the file DO hibernate on 2026-08-13. The namespace's uncaught exceptions went from ~67/day to 7.2k that day and 11.6k the next, and are still running in the thousands.

The socket handlers report and continue, and `webSocketMessage` now also closes the socket on anything that isn't `RoomNotFoundError` — the message is lost either way, and leaving the socket up lets a client believe it is synced while its local state silently diverges. `alarm()` and the constructor report and rethrow instead: the runtime retries a rejected alarm, and an object that can't load its `documentInfo` should be reset rather than left half-built.

`room_start` now also carries how many sockets the boot resumed. Nothing else separates a hibernation wake from a cold boot, and that's the population this whole failure surface lives in.

image-resize-worker had no version metadata binding, which `createSentry` needs to tag a release, so this adds one and types its environment as a `SentryEnvironment`. The deploy script still passes it no `SENTRY_DSN`, so it reports nothing to Sentry either way — wiring one up is a config-only change, deliberately left as a follow-up.

### Notes for reviewers

- `createSentry` is unchanged. An earlier version of this PR made it return `null` instead of throwing on missing config; that's reverted after review. The throw surfaces as a worker exception, which is on the `Worker errors` panel and is alertable, and that is a better signal than the `console.error` line that replaced it. Worth knowing for anyone reading the old discussion: the crash path it would have fixed is latent rather than active — image-optimizer has served 75,411 requests over the last 30 days with zero errors, so `createSentry` is never actually reached there.
- `webSocketError` now reports every abnormal socket termination. The file DO carries a lot of connections, so this may prove noisy — it filters on the `source` extra if so.
- Merged with main, which had since added the `RoomNotFoundError` branches to the same WebSocket handlers and removed Pierre storage from the file DO. Those branches are kept exactly as they were; what changed is the `throw e` that followed them.
- #10052 was the companion, splitting Cloudflare's DO exception metric per namespace and alerting on the file room namespace. It merged but its deploy failed on the alert rule, so neither the panel nor the alert is live yet.
- Still open from the audit, not addressed here: `handleApiRequest` skips Sentry for `StatusError` (fine for 4xx, latent for a 5xx — nothing throws one today), `writeDataPoint`'s capture is commented out, `exhaustiveSwitchError` throws uncaught from the queue batch loop, and no dotcom worker has `tail_consumers` configured.

### Change type

- [x] `bugfix`

### Test plan

1. `cd packages/worker-shared && yarn test run sentry` — the existing dev/production cases, plus a new one pinning the `release` and `environment` a fully configured client is built with.
2. `cd apps/dotcom/image-resize-worker && yarn wrangler deploy --dry-run --env production` — `env.CF_VERSION_METADATA` should now appear in the bindings list. Same for `--env preview` after `setWranglerPreviewConfig` rewrites the preview table.

- [x] Unit tests
- [ ] End to end tests

### Code changes

| Core code      | +1 / -1    |
| Tests          | +18 / -2   |
| Apps           | +153 / -42 |
| Config/tooling | +18 / -0   |


packages/worker-shared/src/sentry.ts
```diff
@@ -64,17 +63,32 @@ export interface SentryEnvironment {
  *
  * @public
  */
+// Warn at most once per isolate. Durable objects call createSentry in their constructors and at
+// the top of fetch(), so warning per call would be a line per request on a misconfigured deploy.
+let hasWarnedAboutMissingConfig = false
```
MitjaBezensek: This should be moved above the tsdoc, otherwise the doc won't be connected to `createSentry` function.

apps/dotcom/sync-worker/src/worker.ts
```diff
@@ -372,10 +402,17 @@ export default class Worker extends WorkerEntrypoint<Environment> {
...
 						// handleOgImageRenderMessage settles the message itself; this guards the batch loop
 						// against an unexpected throw escaping it, so one bad message can't abort processing
 						// of the rest of the batch. Retry is a no-op if the handler already settled.
+						//
+						// Reported on every delivery rather than on the last one. The handler already
+						// reports the render failures it catches, so anything reaching here escaped it
+						// entirely — a bug in the handler, not a failing render. Those are rare, and an
+						// escape that doesn't recur never reaches a final delivery anyway: the handler's
+						// own budget acks the message at attempts >= OG_MAX_RENDER_ATTEMPTS.
+						this.reportQueueFailure(message, e)
```
MitjaBezensek, line 413: Can this throw? If it can we would skip the `message.retry` below.

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -586,29 +611,32 @@ export class TLFileDurableObject extends DurableObject {
 				ws.close(TLSyncErrorCloseEventCode, TLSyncErrorCloseEventReason.NOT_FOUND)
 				return
 			}
-			throw e
+			this.reportError(e, { source: 'webSocketMessage' })
```
MitjaBezensek, line 604: Not really a regression, but wondering if we should close the socket here and allow the connection to be retried? Not really a big deal since any error apart from R`oomNotFoundError` probably means some infra issue.

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -586,29 +611,32 @@ export class TLFileDurableObject extends DurableObject {
...
-		return this.handleWebSocketEnd(ws, 'handleSocketClose')
+		await this.handleWebSocketEnd(ws, 'handleSocketClose')
 	}
 
-	override async webSocketError(ws: WebSocket) {
-		return this.handleWebSocketEnd(ws, 'handleSocketError')
+	override async webSocketError(ws: WebSocket, error: unknown) {
+		// The socket failed rather than closing cleanly. That error was previously dropped on the
+		// floor, so a client whose connection kept breaking left no trace here at all.
+		this.reportError(error, { source: 'webSocketError' })
```
MitjaBezensek, line 622: Think this might happen quite frequently, things like people closing a tab, browser crashing. I guess we can leave this in then see how many of these we get and adjust if needed..

packages/worker-shared/src/sentry.ts
```diff
@@ -64,17 +63,32 @@ export interface SentryEnvironment {
...
+				!CF_VERSION_METADATA && 'CF_VERSION_METADATA',
+			].filter(Boolean)
+			console.error(
+				`Sentry is not configured, errors will not be reported. Missing: ${missing.join(', ')}`
+			)
+		}
 		return null
 	}
 
-	const { SENTRY_DSN, WORKER_NAME, CF_VERSION_METADATA } = requiredEnv(env, {
```
MitjaBezensek, line 72: Wasn't this loud failure better for detecting issues?

apps/dotcom/sync-worker/src/worker.ts
driev: Good catch.

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
driev: Yep. I've got a tail worker PR in the works. Maybe this kind of error would be better off shipped to Loki, where we could keep an eye on it at least, whilst keeping sentry clean?

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
driev: Yeah, good shout.

packages/worker-shared/src/sentry.ts
driev: Yeah, I'll revert this.

MitjaBezensek (approved):

PR #9703, opened 2026-07-27
feat(commenting): @-mention board viewers, not just workspace members

mimecuvalo:
In order to let people @-mention anyone who's actually working on a board — not only its formal workspace members — this expands the comment mention roster to include signed-in viewers who have opened the file. Before this, someone who opened a shared board via a link (a real, signed-in collaborator sitting right there in the document) couldn't be mentioned at all, because the roster was built purely from `group_user` workspace membership.

The notification side already handled viewers: the `comments` synced-query mention gate delivers to anyone with a `file_state` row for the file, and every viewer gets one on open. So the actual gap was only the client roster — the client can see neither other users' `file_state` rows nor their profiles (there's no global user directory). This fills that gap.

### How it works

- A new **`file_visitor`** table is a shareable, per-file projection of `file_state`, maintained by Postgres triggers (migration `044`) and carrying denormalized `userName`/`userColor` — the same denormalization pattern as `comment.authorName` and `group_user.userName`, so no private `user` row is ever joined or synced.
- A new **`fileVisitors`** synced query reads it, access-gated exactly like `fileComments` (only someone who has themselves opened the file sees the list) and bounded to the 100 most-recent viewers. `lastVisitAt` is `NOT NULL` and always stamped (falling back to `lastEditAt`/`firstVisitAt`/`now()`), so the most-recent-first window is deterministic — ZQL has no `NULLS LAST`, and Postgres and SQLite disagree on where nulls sort.
- `CommentsOnCanvas` merges those viewers into the mention roster alongside workspace members (members win on id collision) and into the author resolver, so a mentioned viewer renders with their name.

### Key design decision: a dedicated table, not columns on `file_state`

The obvious first move — denormalize `userName`/`userColor` straight onto `file_state` — is a privacy leak. Zero synced queries replicate **whole rows**, and `file_state` also holds private per-user data: `lastSessionState` (a user's camera position, current page, and selected shapes) and their visit/edit timestamps. Syncing the viewer list from `file_state` would ship all of that to every collaborator. `file_visitor` exists so that every column in the synced row is safe to expose.

### Decisions worth a look

- **Scope is signed-in viewers only.** This falls out for free: a `file_visitor` row exists only where a `file_state` row does, which only authenticated users get. Anonymous public-link visitors have none and never appear.
- **The viewer list is exposed to anyone who can access the file, on all board types — including public shared links.** This is deliberate (a mention roster has to be readable to be useful), but it does mean "who has opened this board" is no longer private to collaborators. Flagging it explicitly.
- **No automated coverage yet** for the trigger + synced-query path — it needs a live Postgres/zero-cache, so verification below is manual. The client merge and query are exercised only via the existing type/build checks.

### Deploy notes

The migration backfills `file_visitor` from `file_state` (~750k rows in production) inside the migration transaction. Zero 1.8 absorbs the newly published table in place — no replica reset, and the backfill COPY streams in chunks that yield to the main change stream, so it holds no replication slot and doesn't stall other tables. The catch: **the table stays invisible to view-syncers until each replica finishes the backfill** (expect minutes at this scale), and a client materializing `fileVisitors` in that window errors — the failure shape of the 2026-06-16 incident. So don't let clients that reference the query go live in the same window: either ship this migration one deploy ahead of the client/query changes, or rely on commenting staying feature-flagged off through the deploy. Backfill completion is visible in the view-syncer logs as `finished backfilling file_visitor`.

### Change type

- [x] `feature`

### Test plan

1. Apply migration `044` to a running Postgres/zero-cache; confirm the `file_visitor` table, its triggers, and the backfill.
2. As user A (a workspace member), open a board. As user B (signed in, **not** a member), open it via a share link.
3. In A's comment composer, type `@` — B appears in the list. Mention B and post.
4. As B, confirm the mention arrives in the sidebar notifications feed and renders A's comment with B's name.
5. Confirm an anonymous (signed-out) visitor does **not** appear in the roster.

- [ ] Unit tests
- [ ] End to end tests

### Release notes

- Comments: you can now @-mention anyone who has opened a board, not just workspace members.

### Code changes

| Section                              | LOC change |
| ------------------------------------ | ---------- |
| Core code (`packages/dotcom-shared`) | +69 / -0   |
| Apps (`apps/dotcom`)                 | +143 / -9  |


mimecuvalo:
Alternative approach opened as #9705: source the mention roster from live board **presence** (currently-present users) instead of a persisted `file_visitor` table. Lighter (one client file, no migration) and privacy-neutral, but covers only who's on the board right now, not historical viewers. The two are mutually exclusive takes on the same feature — see #9705 for the tradeoff table.

PR #9706, opened 2026-07-27
feat(editor): grandfather pre-launch licenses into collaboration

mimecuvalo:
In order to let customers who already bought tldraw use commenting without buying a new license, this grandfathers every license issued before collaboration launches (2026-08-05) into the collaboration umbrella — the same access `FLAGS.FEAT_COLLABORATION` grants, so commenting and any future sub-feature come with it.

A license carries an expiry date but no issue date, so issuance is inferred from expiry: we never issue a term longer than 370 days, so anything expiring before launch + 370 days (2027-08-10) must have been issued before the launch. Renewals that run past the cutoff, and anything issued from launch day on, need the flags.

Two calls worth a look:

- **Evaluation licenses are excluded.** Their terms are weeks, so they'd sit under the cutoff long after the launch and hand collaboration to every future trial. Trials that should include it can be minted with the flags set, which still works.
- **Grandfathering doesn't revive dead licenses.** It flows through `getEnabledFeatures`, which still requires a `licensed` or `licensed-with-watermark` state, so a license expired past its grace period gets nothing.

One gap to be aware of when minting keys: a *non-evaluation* license issued after the launch with a term shorter than ~370 days — a six-month deal, a pro-rated renewal — also lands under the cutoff and gets grandfathered by accident. Closing that would mean putting a real issue date in newly minted keys.

Relates to #9701, which gates commenting on the dotcom side; this is the SDK-license half.

### Change type

- [x] `feature`

### Test plan

1. Mint a license with `yarn generate-test-licenses` whose expiry is before 2027-08-10 and no feature flags, and load it in a production build.
2. Commenting UI is available.
3. Repeat with an expiry on or after 2027-08-10: commenting UI is hidden until the license carries `FEAT_COMMENTING` or `FEAT_COLLABORATION`.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Licenses issued before collaboration launched now include commenting and the other collaboration features.

### API changes

- Added `isGrandfathered` to `ValidLicenseKeyResult` (internal)

### Code changes

| Core code       | +35 / -2   |
| Tests           | +77 / -6   |

### DIFF
--- a/packages/editor/src/lib/license/LicenseManager.test.ts
+++ b/packages/editor/src/lib/license/LicenseManager.test.ts
@@ -32,6 +32,20 @@ const STANDARD_LICENSE_INFO = JSON.stringify([
 	expiryDate,
 ])
 
+// Licenses issued once collaboration had launched expire past the grandfathering cutoff (the
+// launch date plus the longest term we issue), so they get only the features their flags name.
+// Two years out clears the cutoff while staying in the future, so the license never expires.
+const postGrandfatheringExpiryDate = new Date(
+	Date.UTC(now.getUTCFullYear() + 2, now.getUTCMonth(), now.getUTCDate())
+).toISOString()
+
+const POST_GRANDFATHERING_LICENSE_INFO = JSON.stringify([
+	'id',
+	['www.example.com'],
+	FLAGS.ANNUAL_LICENSE,
+	postGrandfatheringExpiryDate,
+])
+
 describe('LicenseManager', () => {
 	let keyPair: { publicKey: string; privateKey: string }
 	let licenseManager: LicenseManager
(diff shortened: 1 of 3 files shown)


packages/editor/src/lib/license/LicenseManager.test.ts
```diff
@@ -567,6 +580,63 @@ describe('LicenseManager', () => {
...
+			const licenseInfo = JSON.parse(STANDARD_LICENSE_INFO)
+			licenseInfo[PROPERTIES.FLAGS] = flags
+			licenseInfo[PROPERTIES.EXPIRY_DATE] = expiry
+			const licenseKey = await generateLicenseKey(JSON.stringify(licenseInfo), keyPair)
+			return (await licenseManager.getLicenseFromKey(licenseKey)) as ValidLicenseKeyResult
+		}
+
+		it('Grants the collaboration umbrella to a license issued before the launch', async () => {
+			const result = await getResultForExpiry(expiryDate)
+			expect(result.isGrandfathered).toBe(true)
```
MitjaBezensek, line 602: Won't this start failing once we are past the grandfathering date? Should probably use a fixed date here?

mimecuvalo: oops yes 🙃

PR #9736, opened 2026-07-28
feat(dotcom): capture real board thumbnails locally

driev:
In order to iterate on the thumbnail render page without a preview deploy for every change, this adds a `--board` mode to the thumbnail capture script that renders a real published or shared board locally.

The script could only render the dev fixture page, so real board resolution and the production render page had no local coverage. The MCP screenshot tool can't stand in for that: under `yarn dev-app` the `BROWSER` binding is deliberately non-functional and `[env.dev]` pins a `compatibility_date` that predates `quickAction`, so `get_shared_board_screenshot` always fails at the Browser Run call.

`--board` points local Playwright at the production `/__thumbnail-render` page, signing its own render token. That exercises real board resolution, the real render page, and the real token gate — everything the MCP tool does except the Browser Run call itself. It can't catch Browser Run-specific behaviour (viewport defaults, fonts in Cloudflare's fleet), so a preview deploy is still the check for those.

```bash
MCP_SCREENSHOT_TOKEN_SECRET=<the [env.dev.vars] placeholder> \
yarn workspace dotcom browser-run-thumbnail --board /f/your-slug --output board.png
```

The board kind is derived from the `/p/` or `/f/` path rather than a separate flag, since it's a property of the board — a whole board link works too, and only its path is read. The secret is never hardcoded: `--board` requires `--secret` or `MCP_SCREENSHOT_TOKEN_SECRET`.

The docs changes correct two claims this work disproved: `wrangler dev --remote` does not enable real local captures (the dev compatibility date and an unreachable render origin both block it), and `POST /app/mcp` does not exercise the production path locally — `get_board_info` works, `get_shared_board_screenshot` never does.

### Change type

- [x] `other`

### Test plan

With `yarn dev-app` running and a board shared or published locally:

1. `MCP_SCREENSHOT_TOKEN_SECRET=<placeholder> yarn workspace dotcom browser-run-thumbnail --board /f/<slug> --output board.png` — writes a 1200x630 PNG of the board.
2. Repeat with `--theme dark`, and with a full `https://www.tldraw.com/f/<slug>` link.
3. Omit the secret, and pass a wrong one — the first fails with an actionable message, the second reaches the render page and fails its token check (403).
4. `--board https://www.tldraw.com/r/<slug>` is refused as an unsupported link, rather than minting a token for the wrong board kind.
5. `yarn workspace dotcom browser-run-thumbnail --fixture layer-panel` still captures the fixture page unchanged.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Apps            | +137 / -10 |

### Related issues

- Relates to #9503

### DIFF
--- a/apps/dotcom/browser-run-thumbnails.md
+++ b/apps/dotcom/browser-run-thumbnails.md
@@ -72,7 +72,7 @@ ORDER BY requests DESC
 
 The sync worker needs:
 
-- `BROWSER` binding - the Cloudflare Browser Rendering binding, declared per environment in `wrangler.toml` (`[env.<env>.browser]`). The worker calls its `quickAction` Quick Actions method (`env.BROWSER.quickAction('screenshot', …)`) directly — no `@cloudflare/puppeteer`, no API token. This requires `compatibility_date` `2026-03-24` or later, which the deployed envs use; `[env.dev]` pins an older date the local workerd supports (see Local development). The dev binding is deliberately not `remote`, so plain `wrangler dev` (and the credential-free e2e stack) boots without a `CLOUDFLARE_API_TOKEN`; the binding is then a non-functional local one and the render path fails closed. Real local captures need `wrangler dev --remote` with credentials, or a preview deploy.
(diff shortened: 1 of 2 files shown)


driev:
Superseded by #9745, which makes the worker's own screenshot path work under `yarn dev-app` — that renders through the real MCP and OG queue surfaces rather than driving the render page from a CLI, so it's closer to production for the local-board case this PR was aimed at.

The fixture page and `--mode browser-run` against a preview deploy are unaffected and still work as they did before this PR.

PR #9738, opened 2026-07-28
fix(commenting): sort the sidebar by a thread's latest comment

mimecuvalo:
In order to keep the threads people are actually talking in at the top of the comments sidebar, this sorts the list by each thread's most recent comment instead of its first one. Today the sort key is the row's displayed date, which is the thread's opening comment — so replying to a week-old thread leaves it buried a week down the list, and a busy conversation is harder to find than a stale one.

Each row now carries a `lastActivity` key alongside what it displays (the most recent live comment's `createdAt`, since deleted comments shouldn't count as activity), and the list sorts on that. Resolved threads still sink below unresolved ones. The row itself is unchanged — it still shows who started the thread and when, which is what identifies it; only the ordering follows the replies.

The ordering moved into a `sortSidebarRows` helper so it can be unit tested, matching how `sortThreadsForPreview` is factored out next door. That also picked up a thread-id tiebreak, so threads with identical timestamps no longer reshuffle between renders.

### Change type

- [x] `bugfix`

### Test plan

1. Open a board with several comment threads and open the comments sidebar.
2. Reply to the oldest thread — it should jump to the top of the list.
3. Resolve that thread — it should drop below the unresolved ones, still ordered by recency among the resolved ones.

- [x] Unit tests

### Release notes

- Fix the comments sidebar ordering threads by when they were started rather than by their latest reply.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +43 / -19  |
| Tests     | +45 / -0   |

### DIFF
--- a/packages/commenting/src/canvas/comments-sidebar.test.ts
+++ b/packages/commenting/src/canvas/comments-sidebar.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest'
+import type { CommentListItemProps } from '../ui/comments-list'
+import { sortSidebarRows } from './comments-sidebar'
+
+function row(id: string, lastActivity: number, resolved = false) {
+	return {
+		item: {
+			id,
+			resolved,
+			author: { id: 'user:1', name: 'A' },
+			preview: '',
+			date: '',
+		} as unknown as CommentListItemProps,
+		lastActivity,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


mimecuvalo:
@MitjaBezensek was noticing this, and it felt odd

@jsscclr i'm pushing this in but feel free to push back on this decision after the fact

PR #9741, opened 2026-07-28
fix(commenting): don't let view-only guests comment

mimecuvalo:
In order to stop a guest holding a view-only share link from commenting on a board they can't edit, this PR decides comment access from the file's share tier instead of from "is this session signed in".

Comments sync over the file room's object lane, gated per session by `objectAccess` independently of the canvas `isReadonly` flag. That flag was set to `write` for any authenticated session, so a signed-in viewer could post, reply, and resolve on a `view` file.

Now a small pure helper maps `{ sharedLinkType, hasOwnerAccess, isAuthenticated }` onto both lanes at connect time: owners and workspace members get full access, `edit` guests keep canvas write and can comment when signed in, and everyone else is read-only on both. The reconnect-on-share-change path compares both lanes per session, so changing a file's tier reconnects the guests it affects — including sessions still holding the old blanket comment access. On the client, `useSync` surfaces the server-granted access so `TlaEditor` can drop the comment tool, its overrides, and the composer. A viewer still sees existing threads; the server rejects stray writes as a backstop.

Two things worth a reviewer's eye:

- **Non-`edit` tiers now fail closed on the canvas too.** Previously only `sharedLinkType === 'view'` forced read-only, so a shared file carrying any other value (there are legacy ones in the column, e.g. `private`) granted guests edit access. This matches what `hasWriteAccessToFile` in `getAuth.ts` already does for HTTP requests.
- **`canComment` feeds the `tools` prop**, and changing that prop recreates the editor. It's resolved before the editor mounts (the store is synced by then), so this only bites on a mid-session tier change, where the server is already reconnecting the guest.

This is #9604 with the feature removed. That PR added a third "Commenter" tier — read-only canvas, comments allowed — and built this gating as its foundation; we decided not to ship comment-only mode, so the tier is commented out at four sites (`computeFileAccess`, the reconnect check, and the share menu's option and label), each pointing at the others.

### Change type

- [x] `bugfix`

### API changes

- Changed `RemoteTLStoreWithStatus` (`@tldraw/sync`): the `synced-remote` status now carries a `readonly objectAccess: TLObjectStoreAccess` field — the object-lane write access the server granted this session. Defaults to `'write'`, so servers that don't send it are unaffected.

### Test plan

1. As the owner, share a file as **Editor** and open the link as a different signed-in user: the comment tool works, comments post/reply/resolve.
2. Switch the file to **Viewer**: the guest reconnects, the comment tool and composer disappear, existing threads stay visible and readable in the sidebar, and the canvas is read-only.
3. Switch back to **Editor**: the guest regains both.
4. Confirm the owner and workspace members can always comment, whatever the tier says.
5. Confirm signed-out viewers still get the sign-in prompt rather than a broken composer.

- [x] Unit tests
- [ ] End to end tests

### Release notes

- Fix commenting being available to anyone signed in on a view-only board — commenting now follows the file's share tier.

### Code changes

| Core code       | +24 / -6   |
| Tests           | +55 / -0   |
| Automated files | +8 / -1    |
| Apps            | +131 / -23 |

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaEditor/CommentsOnCanvas.tsx
+++ b/apps/dotcom/client/src/tla/components/TlaEditor/CommentsOnCanvas.tsx
@@ -30,10 +30,22 @@ type FileVisitors = QueryResultType<typeof queries.fileVisitors>
  * notifications feed, which is bounded to recent comments), so every unread pin resolves however
  * old the comment is.
  */
-export function CommentsOnCanvas({ fileId }: { fileId: string }) {
+export function CommentsOnCanvas({
+	fileId,
+	canComment = true,
+}: {
+	fileId: string
+	/** Whether this session may write comments. When false the layer is read-only: existing
+	 *  comments show, but there's no composer (the server also rejects writes). */
+	canComment?: boolean
+}) {
 	const editor = useEditor()
 	const app = useMaybeApp()
 	const currentUserId = app?.userId ?? null
+	// The compose identity: a signed-in user who's allowed to comment. Null makes CanvasComments a
+	// read-only viewer (no composer/reply/resolve). Everything read-only — read status, name
+	// resolution, the sidebar's "only mine" filter — still uses the real currentUserId.
+	const composeUserId = canComment ? currentUserId : null
 
 	const currentUser = useValue(
 		'current user',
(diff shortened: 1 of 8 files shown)


MitjaBezensek:
I'm wondering if this is the right approach. Here's my thinking:
1. This prevents all object writes for view links. What if somebody would want to prevent (possible future) collections, but allow comments that might ride the same object lane?
2. We already have authorizers that can achieve the same thing and can distinguish between different types that ride the object lane. We already use that mechanism to prevent anon writes, so why use another way of expressing a similar thing? To allow that we could add room open status to session meta or fully add support for this in the sdk.
3. This would allow us to skip the `useSync` changes which feel a bit forced. Client already gets readonly status, we can use that to gate the UI on the client side for dotcom.

The fail-closed handling of legacy sharedLinkType values looks right regardless, we should keep that 👍

MitjaBezensek:
This also seems to remove the `Sign in to comment` CTA for anon users. Did we decide not to show that?

Issue #9720, opened 2026-07-28
Comments: can't leave a comment on an arrow


max-drake:
🗣️: you cant leave comments on arrow

---

**What happened?**

In the comments feature, a comment can be attached to a shape, but not to an arrow. Arrows are ordinary shapes, so leaving a comment on one should work the same way it does for any other shape: the pin should anchor to the arrow and follow it as the arrow moves, rather than being rejected or dropped onto a bare canvas position.

**How can we reproduce the bug?**

1. Open a canvas with the comments feature enabled.
2. Draw an arrow.
3. Try to attach a comment to the arrow.

Expected: a pin anchors to the arrow and follows it when the arrow is moved.
Actual: the comment can't be attached to the arrow.

**Where did this happen?**

tldraw SDK (@tldraw/tldraw npm package)

**Open questions**

1. **What happens when you try?** Does the comment get refused outright, or does it land as a loose canvas comment that doesn't follow the arrow?
   _Awaiting answer._
2. **Where should the pin sit on an arrow?** An arrow is a thin line with no real interior, so the pin needs somewhere sensible to live — the arrow's midpoint, the exact point you clicked along the line, or somewhere else?
   _Awaiting answer._
3. **Should the pin track a bound arrow?** When the arrow is bound to two shapes and those shapes move, so the arrow reshapes and rebends, should the comment pin follow along?
   _Awaiting answer._
4. **Is this arrows only, or all line-like shapes?** Do you want the same fix to cover lines, draw, and highlighter shapes, or is arrows the scope you care about right now?
   _Awaiting answer._

**Notes**

Part of the comments feature work. Related: #9715 (comments on groups), #9718 and #9719 (comment pins not stable when a shape is rotated).

Confidence: 45%, still need more information.

Issue #9719, opened 2026-07-28
Comments: comment pins aren't stable when a shape is rotated


jsscclr:
### What happened?

Comments attached to a shape aren't stable when that shape is rotated. Instead of holding their position relative to the shape, the comment pin moves to the wrong place, so the comment no longer marks the part of the shape it was attached to.

Reported from the comments feature work with no further detail, so the exact behaviour still needs to be pinned down: whether the pin drifts continuously while rotating, only ends up wrong once the rotation is committed, or both.

Closely related to #9718 (rotating a shape doesn't normalize the comment into the correct position) — these may turn out to be the same underlying issue.

### How can we reproduce the bug?

1. Attach a comment to a shape.
2. Rotate the shape.
3. The comment pin isn't where it should be relative to the shape.

### Where did this happen?

tldraw SDK (@tldraw/tldraw npm package)

### Notes

Part of the comments feature work. Related: #9712, #9714, #9715, #9716, #9718.

Issue #9718, opened 2026-07-28
Comments: rotating a shape doesn't move its comment pin to the correct position


mimecuvalo:
### What happened?

In the comments feature, a comment pin attached to a shape does not stay in the right place when the shape is rotated. The pin keeps its original position instead of being transformed along with the shape, so after rotating it no longer points at the part of the shape it was placed on.

A pin position should be stored relative to the shape and re-normalized through the transform of that shape, so that rotating (and presumably resizing) the shape keeps the pin anchored to the same point on it.

### How can we reproduce the bug?

1. Open a canvas with the comments feature enabled.
2. Add a comment to a shape so a pin appears on it.
3. Rotate the shape.

Expected: the pin follows the rotation and stays anchored to the same point on the shape.
Actual: the pin stays where it was and no longer lines up with the shape.

### Where did this happen?

tldraw SDK (@tldraw/tldraw npm package)

### Notes

Related: #9719 (comments not stable on rotation) looks like the same underlying problem reported separately.

Issue #9715, opened 2026-07-28
Comments: allow comments to be attached to groups


jsscclr:
### What feature would you like to see?

In the comments feature, a comment can be attached to a shape, but not to a group. Attaching a comment to a group should work the same way it does for any other shape, so a comment can be anchored to a grouped set of shapes rather than to a single member of the group or to a bare canvas position.

### Open questions

- Does the comment attach to the group shape itself, or to the current selection when a group is selected?
- What should happen to the comment when the group is ungrouped, or when a member shape is removed from the group?
- Should the comment follow the group when the group moves or resizes?

### Where would this happen?

tldraw SDK (@tldraw/tldraw npm package)

### Notes

Part of the comments feature work. Related: #9712, #9714, #9716.

PR #9743, opened 2026-07-28
fix(commenting): give the comment hit-test a margin so arrows are commentable

mimecuvalo:
In order to let people comment on arrows — reported in #9720 — this gives the comment placement hit-test the same margin that select and hover already use. Today the comment lands as a loose page point instead of anchoring to the arrow.

Placement hit-tested with `getShapeAtPoint(point, { hitInside: true })`, leaving `margin` at its default of 0. An arrow's geometry is a `Group2d`, and `Group2d` is hard-coded to `isClosed: true, isFilled: false`, so arrows take the closed-but-hollow branch and need `distance <= margin` to register. The distance from a point to an arrow's line is positive everywhere off the stroke, so a zero margin only matched a pixel-perfect click on the line itself. `hitInside` doesn't rescue it — that flag only admits negative distances, i.e. points inside a closed region, which an arrow has none of.

All four placement hit-tests (tool hover hint, tool placement, pin-drag hover hint, pin-drag commit) now go through one `commentTargetShapeAt` helper that passes `editor.getHitTestMargin()`. They were already four identical copies — one carrying a comment promising it matched another — so centralizing keeps them honest, and commenting picks up whatever slack an app configures.

### Answers to the open questions on #9720

1. **What happens when you try?** Not refused — the hit-test returns nothing and placement falls through to `{ type: 'point' }`, so you get a loose canvas comment that doesn't follow the arrow.
2. **Where should the pin sit?** Where you clicked. `shapeAnchorAt` stores the click as a normalized fraction of the arrow's page bounds, precise by default — no new behavior needed here.
3. **Does the pin track a bound arrow?** It follows, with a caveat. `anchorPagePoint` re-resolves the stored fraction against `getShapePageBounds` on every render, so the pin moves as the arrow moves and reshapes. But the fraction is relative to the axis-aligned bounding box, so a rebending arrow slides the pin relative to the actual stroke — the same drift class as #9718 / #9719, not fixed here.
4. **Arrows only, or all line-like shapes?** All of them. The cause is geometric, not arrow-specific, so lines, draw, and highlighter shapes are fixed by the same change, as are arrows nested inside groups.

Related: #9715 (comments on groups) is untouched — `getShapeAtPoint` filters out `group` shapes entirely, so a group can't be a comment target at any margin. Clicking a grouped shape anchors to the child under the cursor.

### Overlap with #9711

#9711 refactors the two comment-tool call sites into `COMMENT_HIT_OPTS = { hitInside: true, hitFrameInside: true }`, so it conflicts textually with this. The fixes are orthogonal and combine cleanly — whoever merges second folds the margin in:

```ts
const COMMENT_HIT_OPTS = { hitInside: true, hitFrameInside: true, margin: editor.getHitTestMargin() }
```

Note that #9711 leaves the two `comments-overlay.tsx` pin-drag hit-tests on the old options; this PR routes those through the helper too.

### Change type

- [x] `bugfix`

### Test plan

1. Enable commenting, draw an arrow, and place a comment on it — the pin anchors to the arrow and follows it when the arrow moves.
2. Hover the arrow with the comment tool — the hint outlines the arrow before you click.
3. Drag an existing pin onto an arrow — it re-anchors to the arrow.
4. Same for a line, a draw stroke, and an arrow inside a group.

- [x] Unit tests

### Release notes

- Fix comments not attaching to arrows, lines, and draw strokes. Placing a comment on one dropped a loose canvas comment instead of anchoring to the shape.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +29 / -6   |
| Tests     | +81 / -0   |


jsscclr (approved):
LGTM!

PR #9745, opened 2026-07-28
feat(dotcom): render thumbnails locally through the client dev server

driev:
In order to work on board thumbnails without a preview deploy for every change, this makes the worker's screenshot surfaces run under `yarn dev-app`. `POST /app/mcp` now returns a real PNG locally, and the OG image route renders through its queue consumer — neither of which was previously possible on a dev machine.

Both surfaces need a browser to visit the render page, and local dev has no route to Browser Run. `[env.dev.browser]` is deliberately not `remote` so the credential-free process-compose stack can boot without a `CLOUDFLARE_API_TOKEN`, and `[env.dev]` pins a `compatibility_date` older than `quickAction` needs, because the newer date is ahead of the workerd bundled with our pinned wrangler. That second reason applies to `wrangler dev --remote` too, since it deploys with the same config, and dev's render origin is unreachable from Cloudflare regardless. So both surfaces failed at the browser call and a preview deploy was the only way to exercise them.

The worker isolate can't drive a browser, but the client's dev server is a Node process that already depends on Playwright. This adds a dev-only vite plugin serving `POST /__screenshot`, and the worker uses it when `LOCAL_SCREENSHOT_SERVICE_URL` is set.

### Design notes

**Selection is config-driven, not environment-named.** Only an environment that sets the var can take the local path; deployed environments leave it unset and stay on the `BROWSER` binding. There's no top-level `[vars]` block, so the var appears exactly once in `wrangler.toml`, under `[env.dev.vars]`.

**The render contract isn't duplicated.** The worker POSTs the service the same body it would send the Quick Action (`getThumbnailScreenshotRequestBody`), and the plugin reads the wait strategy, capture selector, timeout and viewport out of it. One builder, two consumers, so they can't drift. This also sidesteps a real constraint: `vite.config.ts` is loaded by Node, which can't resolve `@tldraw/dotcom-shared`'s extensionless re-exports, so the plugin has no workspace imports at all.

**The plugin can't be pointed at another URL.** It discards the caller's host and fixes the path, so a body naming another origin still renders the local render page. The path allowlist is asserted in the plugin rather than taken from the request.

**It is not Browser Run** — different Chrome build, different flags, no billing headers. It makes the local path work; it isn't evidence a render will succeed in production, and a preview deploy is still the check for anything Browser Run-specific such as font availability in Cloudflare's fleet. The naming and comments say so deliberately.

### Production impact

The Browser Run path is untouched — the worker diff is purely additive, with no lines removed. Production gains one falsy check per render and ~45 lines that never execute. `apply: 'serve'` keeps the plugin out of built clients: `yarn build-app` succeeds and `dist` contains no trace of it.

### Change type

- [x] `other`

### Test plan

With `yarn dev-app` running and a board shared or published locally:

1. `POST /app/mcp` with `tools/call` for `get_shared_board_screenshot` returns a text item and an `image/png` — verified as a real 1200x630 PNG.
2. Repeat the call — served from the R2 cache in ~17ms instead of rendering.
3. Request it with `"theme":"dark"` — a separate render under its own cache key.
4. `GET /app/social-preview/f/:slug/image` — first hit 302s to the default image and enqueues; after the queue consumer runs, it serves the board PNG.
5. `POST /__screenshot` directly: `GET` gives 405; a body naming another origin still renders the local page; a body with a path other than the render page is refused; a bad token surfaces the render page's own error rather than timing out.

- [ ] Unit tests
- [ ] End to end tests

### Code changes

| Apps            | +210 / -3  |

### Related issues

- Relates to #9503


mimecuvalo (approved):
👍

PR #9711, opened 2026-07-28
fix(commenting): polish the comment UI from wg-comments feedback

jsscclr:
A polish pass over the commenting UI addressing feedback from #wg-comments and a working session. Small, self-contained commits grouped by area:

**Pins & markers** — 34→28px, author colors on posted pins (draft stays white), softened resting/hover/open-ring shadows to match the context menu, no pointer cursor on the draft pin.

**Composer** — up-arrow send button (round, centred on the input line), vertically-centred text/placeholder, tighter left padding, whole field focuses the input, compact height cap + wheel-scroll for long pastes, lighter field fill to match the dialog input.

**Thread & cards** — header actions right-aligned at the panel's 8px inset, card actions aligned under them, edit/delete no longer recolor on hover, unified surface shadows, tighter card padding.

**Hover preview** — surfaces matched to the thread's shadow.

**Icons** — radix comment tool icon, enlarged toolbar icon, rounded emoji-picker hover, filled notification bell.

**Mentions** — dropped the @-picker double outline, loosened list spacing, smaller names.

**Comment tool** — anchor-target hinting while placing (incl. frames), with test coverage.

**dotcom** — comments sidebar button (canvas bg + open state), share/comment mutual exclusivity, Import file moved into the File submenu.

### API changes

No public surface change. `SendButtonProps.label` gained a doc comment (api-report no longer marks it undocumented).

### Preview & how to test

Preview: https://pr-9711-preview-deploy.tldraw.com/

1. Sign in and open or create a file.
2. Use the comment tool to place a comment on a shape and on a frame.
3. Open the thread; check the header ✓/kebab alignment, hover a card for edit/delete/react, add a reaction, `@`-mention someone.
4. Hover a pin to see the preview card; check pin/hover shadows.

PR #9747, opened 2026-07-28
fix(commenting): re-derive the comment UI metrics the pin resize left behind

mimecuvalo:
In order to make the pins and pills from #9711 behave the way they look, this fixes the measurements that were derived from the sizes that PR changed. The pin shrank 34→28px, the thread actions grew 22→26px, and the header moved out of the panel into a floating pill — but several numbers computed from the old values stayed put, so a few things are now quietly wrong:

- **Hovering a pin flickers.** The hover preview has an invisible bridge spanning the gap between marker and panel, so moving the pointer across doesn't cross bare canvas and retract the preview. It's sized as `offset − marker-size`, and it was still using the 34px badge: 14px of bridge over a 20px gap. The last 6px is exactly the bare canvas the bridge exists to cover.
- **The right edge of "edit" deletes your comment.** The new enlarged tap targets are `inset: -6px` pseudo-elements on buttons laid out with a 2px gap, so each one reaches 4px into its neighbour's *visible* area. They're siblings with no `z-index`, so the later one wins — and in a card's action row, delete comes after edit.
- A single pin's popover re-bases itself by half a marker to line up with a stack preview's top card, still using the badge's half rather than the pin's.
- The byline's reserved width still assumed 22px action buttons, so a long name can now run under them — the thing that reserve exists to prevent.

The fix is mostly to stop hardcoding these: the pin's size becomes `--tlui-cmt-pin-size` (mirrored by `PIN_SIZE` in `thread-view.tsx`, which the comments now cross-reference), and the byline's reserve derives from `--tlui-cmt-thread-action-size`. The composer offsets and the preview geometry compute to the same pixels as before — verified, not assumed.

One deliberate visual change while we were in here: the card's action pill now sits at `top: -2px`, just above the card's top edge, rather than straddling it at half its own height.

Three smaller things in the same area:

- Inside a stack, the expanded thread's pills float ~15px above its panel but the list's gap is 10px, so they land on the card above and clip against the scroll box. That one entry gets the room it needs.
- `CommentThread`'s `header` prop is `@public` and documented, but the pill it now renders into is right-anchored and only as wide as its buttons — a title would grow leftwards off the panel edge. A title now takes the header back in flow. tldraw itself passes no title, so nothing changes for us.
- `--tlui-cmt-pill-shadow`'s hairline ring is the only thing separating a pill from the panel it overlaps, and it was a literal near-black — invisible on a dark panel. It's a theme token now.

Worth a look on the preview deploy: the stack spacing and the dark-mode pill edge are the two changes I couldn't verify locally (there's no commenting example in `apps/examples`, and these need real threads).

### Change type

- [x] `bugfix`

### Test plan

1. Hover a comment pin and move the pointer right onto the preview card — it should stay open the whole way across.
2. Hover your own comment in an open thread and click the right edge of the edit button — it should edit, not delete.
3. Open a cluster/stack popover and expand a thread that isn't the first card — its action pills shouldn't touch the card above, and shouldn't clip while scrolling.
4. Give a comment a long author name and check it doesn't run under the hover actions.
5. Switch to dark mode and check the thread header pill and card action pills still have a visible edge against the panel.

- [ ] Unit tests
- [ ] End to end tests

### API changes

None. `CommentThreadProps.header` keeps its type and documented meaning; only its layout is repaired.

### Code changes

| Section   | LOC change |
| --------- | ---------- |
| Core code | +100 / -40 |

Follow-up to #9711.

### DIFF
--- a/packages/commenting/src/canvas/canvas.css
+++ b/packages/commenting/src/canvas/canvas.css
@@ -96,6 +96,14 @@
 	margin: -16px;
 }
 
+/* The expanded thread is the only entry with pills floating above its panel — its header actions,
+   and its first comment's hover actions. Those reach ~15px up, past the list's 10px gap, so give
+   this one entry extra room: without it the pills land on the card above, and against the scroll
+   box's clip edge. Collapsed cards keep their own actions inside, so they need none of this. */
+.tlui-cmt-stack-list__thread {
+	margin-top: 8px;
+}
+
 /* Collapsed entries match the expanded thread's surface (same width, panel, elevation) so the
    list reads as one column of cards with one of them opened up. */
 .tlui-cmt-stack-list__card {
@@ -156,7 +164,7 @@
    The popover and preview share a first-comment position now that the actions float outside the
    panel rather than reserving a header row, so no vertical shift is needed. */
 .tlui-cmt-canvas-preview--thread {
-	--tlui-cmt-preview-bridge: calc(var(--tlui-cmt-preview-offset) - var(--tlui-cmt-marker-size));
+	--tlui-cmt-preview-bridge: calc(var(--tlui-cmt-preview-offset) - var(--tlui-cmt-pin-size));
 }
 
 /* A badge is centred on its anchor, so only half of it does. */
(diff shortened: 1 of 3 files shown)


mimecuvalo:
@jsscclr for you to review after the fact please 🙏

PR #9748, opened 2026-07-28
fix(commenting): address review findings across clustering, sync, and dotcom

mimecuvalo:
Fixes gathered from a full review pass over the commenting branch (#9471), coalescing findings from two independent reviews. Accessibility is deliberately **not** in scope here — the canvas comment UI is pointer-only (pins, stacks, previews and the emoji grid are `div`s with `onClick`, and no focus moves into the thread popover), which is a large enough change to deserve its own PR.

Every fix below is covered by a test that fails without it.

### Highest-impact

**Clustering partition corruption.** `onCamera` healed suppressed merge events *after* the merge walk. When one discontinuous zoom-out crossed both a suppressed event and an event consuming its result, the consumer was applied first — its `delete` of the never-applied producer was a no-op — and the producer was then re-applied on top, leaving both results visible. Reproduced with the module's own documented micro-trace table: 7 member slots over 4 leaves, i.e. duplicate cluster badges on screen that persist through further zooming. Reachable in normal use, since `seedFrom` runs on every rebuild adoption and routinely produces suppressed events, and any zoom-to-fit, deep link or keyboard zoom can then trigger it. Healing now runs first; threshold monotonicity guarantees a suppressed producer's `zMerge` is >= its consumer's, so the walk always finds what it needs.

The reason no existing test caught it: the (otherwise excellent) conservation property walks all start from `seed()`, never `seedFrom()`, so suppression was never exercised under a random zoom walk.

**Timestamps could crash every viewer's comments UI.** `createdAt`/`editedAt`/`resolved.at` accepted any finite number. `9e15` fits a Postgres `bigint` but is outside `Date`'s representable range, so it validated, persisted, synced — and then threw `RangeError: Invalid time value` in every client that formatted it, again after reload. Timestamps are now bounded to `Date`'s range at the schema level, so the value can't enter the store.

**Legacy rooms poisoned the Postgres outbox permanently.** `objectAccess` was granted to any signed-in user with no `isApp` gate, while `objectTypes` is wired into every room and the drain uses the room slug as `fileId`. A forged client could push comments into a legacy room (which has no comment UI, so no moderation path); those rows then failed `comment_thread_file_id_fkey` on every drain forever, because thread upserts were the one path with no prune matcher. Now gated on `isApp`, with prune matchers added for thread and comment parent/file FK violations — a comment whose parent thread was vetoed or never existed no longer retries for the life of the room.

### Authorization

- Only the reactor may delete their own reaction. Cascades still sweep every reactor's rows, because server-initiated writes carry no session and skip authorizers entirely — open client deletion was never what made the sweep work.
- A comment's `threadId` and `createdAt` are immutable. Re-parenting moved a comment between conversations (and, where threads span files, between files); a mutable `createdAt` let a comment be re-sorted or pinned to the top of the notification feed after the fact. `pageId` stays mutable, since moving an anchored thread between pages rewrites it on every comment.
- The typeName-swap guard on puts only ran when `authorizeRecord` was configured. The write gate keys off the *incoming* typeName, so in a room without authorizers a session denied on one lane could relabel an existing record and write through the other. The guard is now unconditional.
- Bounded the reaction `emoji` token — reaction ids embed it verbatim, so an unbounded token meant an unbounded record id.

### Correctness and performance

- The notifications query, the most expensive in the schema (nested `EXISTS` over comment/thread/file/group), was subscribed for every signed-in user regardless of the commenting flag — so a closed flag didn't actually keep commenting off the wire during rollout.
- Comments are grouped by thread once in a shared computed instead of every mounted thread scanning and sorting the whole comment set, which made one comment mutation do N full passes for N open threads.
- Region resize/move committed via `editor.run(..., { history: 'ignore' })` directly, so a consumer configuring `history: 'record'` got undoable pin drags but silently non-undoable region resizes. It now goes through `commitCommentMutation` like every other write.
- `renderMarkdown` emitted `<a href>` straight from comment text with no protocol check. React does not block `javascript:` hrefs, and the renderer is public API, so any host rendering another user's comment body had click-to-execute XSS. Targets are now restricted to http, https, mailto, tel and relative urls, and an unsafe target degrades to plain text. (tldraw.com was never exposed — it renders through TipTap, which sanitizes hrefs itself.)

### dotcom polish

- Notification previews used a local `richTextToPlaintext` that collected only text nodes, silently dropping `@`-mentions from the one notification category guaranteed to contain them. Replaced with the package export that renders them; the local copy is deleted.
- Notification bylines fell back to the raw `authorId` uuid, and rows to the raw `fileId`. Now "Someone" and "Untitled file", matching the rest of the commenting UI.
- The comments sidebar button rendered in legacy and published-file contexts, where no sidebar is mounted — clicking it toggled state nothing reads.
- Corrected the reaction contract docs, which claimed one reaction per user per comment while the id derivation, the unique constraint and the UI all allow one per emoji.
- Fixed mixed tabs/spaces in `REACTIONS.md` that made `git diff --check` fail.

### Deliberately not fixed here

- **Accessibility** — separate PR, as noted above.
- **Moving an anchored thread between pages** leaves other authors' comments with a stale `pageId`, since only the mover's own comment updates pass the author check. A correct fix needs a trusted server-side rehome or a narrowly validated page-only update path — the authorizer can't see other records to validate against.
- **SQLite/InMemory storage divergence**: SQLite routes `get`/`delete` by id prefix against the *current* `objectTypes`, so removing a type on reopen strands rows that are still hydrated to clients but can never be patched or deleted. Needs a storage-layer decision (check both tables, or a reverse sweep).
- **Unbounded comment body size** — `richTextValidator` has no length or depth cap. Shared with text shapes, so it wants a considered fix rather than a comment-only cap.

### API changes

Removes three unused `@public` exports from `@tldraw/commenting`:

- `renderMarkdown(text: string)`
- `CommentText` and `CommentTextProps`

They had no callers anywhere in the repo and their emitted class names (`md-p`, `md-list`, `md-code`) are styled by no CSS — they date from a prototype where comment bodies were markdown strings rather than `TLRichText`. Comment bodies render through `CommentBody` / `renderCommentHtml`, which mirrors the editor's `renderHtmlFromRichText`. `CommentCardProps.body` still accepts any `ReactNode`, so a consumer with markdown can supply their own renderer. Removed now rather than after the package publishes, when it would be a breaking change.

No other public API changes: the sync-core and tlschema edits in this PR are all internal behavior.

### Change type

- [x] `bugfix`

### Test plan

1. Place several comments close together, zoom out until they merge into a badge, then zoom out again in one jump (zoom-to-fit or a deep link) — no duplicate or overlapping badges.
2. React to a comment as two different users; confirm neither can remove the other's reaction, and that deleting a comment still clears every reactor's.
3. With the commenting flag off, confirm the notifications query is never subscribed.
4. Open a legacy `/r/` room while signed in and confirm comment writes are refused.
5. Post a comment containing `[click me](javascript:alert(1))` and confirm it renders as text, not a link.
6. Check a mention notification preview still shows the `@name`.

- [x] Unit tests

### Release notes

- Fixed a bug where comment pins could render duplicate cluster badges after zooming out in one jump.
- Fixed a bug where a comment with an out-of-range timestamp could break the comments UI.


packages/commenting/src/clustering/runtime.ts
```diff
@@ -141,6 +141,24 @@ class ClusterRuntimeImpl implements ClusterRuntime {
...
+		// Heal suppressed events at their own merge threshold BEFORE the merge walk: a zoom-out
+		// past zMerge merges a held-out band event exactly as if it had still been ahead of the
+		// cursor. Healing must precede the walk because a single zoom jump can cross both a
+		// suppressed event and an event that consumes its result; applying the consumer first
+		// would leave the unapplied producer's children in `visible` (its delete is a no-op) and
+		// then re-add the producer on top, double-counting those leaves. Threshold monotonicity
+		// guarantees a suppressed producer's zMerge is >= its consumer's, so anything the walk
+		// needs has already healed. Set iteration is insertion order (ascending index), so a
+		// healed event's suppressed children (larger zMerge, smaller index) heal before it.
+		if (this.suppressed.size > 0) {
```
mimecuvalo, line 153: cc @meg-an31 lemme know if this isn't the right approach

apps/dotcom/sync-worker/src/TLFileDurableObject.ts
```diff
@@ -1921,7 +1924,12 @@ export class TLFileDurableObject extends DurableObject {
 				const runBatchWithFallback = async <R extends { id: string }>(
 					rows: R[],
 					insert: (rows: R[]) => Promise<unknown>,
-					shouldPrune?: (error: unknown) => boolean
+					// Called with the offending row on the row-by-row pass, and without one for the
```
mimecuvalo, line 1927: @MitjaBezensek pls take a look at this bit of logic when you get the chance

MitjaBezensek: This part looks good 👍

PR #9757, opened 2026-07-29
fix(commenting): tolerate clock skew in the comment join-time gate

mimecuvalo:
Follow-up to #9753. That PR gated the reply notification reason on the comment postdating when you joined the thread, which is the right fix — but the two sides of that comparison come from different machines' clocks. `createComment` stamps `createdAt` with the authoring client's `Date.now()` (`packages/tlschema/src/records/TLComment.ts`), and neither dotcom call site passes an explicit `now`.

So if your clock runs a minute fast, your own join comment is stamped a minute late, and a reply that genuinely followed it reads as thread history and is dropped — silently, with nothing in the feed to hint at what's missing. The previous set-based membership test was immune to clock skew; making the gate time-based is what exposes it.

This allows a comment to predate the join time by up to a minute and still count as a reply. The failure directions aren't symmetric: too small a tolerance loses real notifications, too large readmits comments the user already saw when they joined. A minute is well past the drift of a roughly-synced clock and far short of what it would take to readmit a thread's history, which is what the gate exists to keep out.

Server-side timestamping would fix this at the root, but it needs a schema migration and changes how the commenting SDK creates records — out of scope here.

### Change type

- [x] `bugfix`

### Test plan

1. As user A, start a thread and comment in it
2. As user B (clock set a minute ahead), reply to the thread
3. As user A, reply again
4. B's notifications: A's reply from step 3 appears

- [x] Unit tests

### Release notes

- Fix comment notifications being dropped when collaborators' device clocks disagree.

### DIFF
--- a/apps/dotcom/client/src/tla/components/TlaSidebar/components/commentNotifications.test.ts
+++ b/apps/dotcom/client/src/tla/components/TlaSidebar/components/commentNotifications.test.ts
@@ -6,6 +6,13 @@ const ME = 'user_me'
 const OTHER = 'user_other'
 const THIRD = 'user_third'
 
+/**
+ * Event times, in minutes from an arbitrary epoch. The join gate tolerates a minute of clock skew
+ * between authors, so tests that turn on it space their events well past that — timestamps a few
+ * milliseconds apart would all land inside the tolerance and say nothing about the gate.
+ */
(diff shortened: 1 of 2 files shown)


MitjaBezensek:
Thanks!

PR #9760, opened 2026-07-29
fix(commenting): gate comment writes in the record authorizers instead of the object lane

MitjaBezensek:
In order to stop view-only guests from commenting without closing the whole object lane to them, this PR replaces #9741's blanket per-session `objectAccess` gating with per-type authorizer enforcement, and derives the client's comment UI state from the readonly signal it already has.

#9741 solved the bug by flipping the coarse lane flag from the file's share tier. That worked, but it gated every object-lane record type at once (a future type riding the lane would inherit comment policy), duplicated the authorizer mechanism we already use for anon writes, and pushed a new `objectAccess` field through `useSync` that dotcom didn't really need — losing the anon "sign in to comment" CTA along the way.

The rework, layer by layer:

- `@tldraw/sync-core` exposes the session's existing `isReadonly` on the record-authorizer session argument — surfacing state the room already tracks, no policy imposed.
- `createCommentAuthorizers` gains a `canComment(session)` option, defaulting to `({ isReadonly }) => !isReadonly`: comment writes follow canvas access unless the host says otherwise. Comment-only setups opt out with `canComment: () => true`; custom criteria (roles, bans) fit the same hook. The default runs before every per-type rule on create, update, and delete of all three comment record types.
- The sync worker drops `computeFileAccess` and the tier-driven lane flag: `objectAccess` reverts to the anon-only rule (comment authors need a user row in Postgres), tier gating rides the authorizers via the session's readonly state, and the reconnect check collapses back to the single readonly comparison — a tier change flips `isReadonly`, and the reconnect refreshes what the authorizers see. The fail-closed `sharedLinkType !== 'edit'` handling from #9741 is kept at both sites. `OBJECT_TYPES` now carries a `satisfies` clause against `authorizeFileRecord`'s keys, so adding an object type without an authorizer is a compile error rather than an open lane (verified by negative test during review: a bogus entry fails typecheck).
- The dotcom client stops consuming `store.objectAccess`. The commenting plugin's `canComment` option does the gating (`signedIn && !editor.getIsReadonly()`, reactive through the plugin's signal tracking), the comment tool stays registered whenever the flag is on (the toolbar button and `c` shortcut already hide on readonly canvases via the UI's `readonlyOk` filtering, so no editor-recreating `tools` prop juggling), and `SignInToComment` returns to anon users on editable canvases — it now guards itself against readonly canvases and signed-in sessions, where the CTA would lie.

One deliberate keep: the `RemoteTLStoreWithStatus.objectAccess` surface from #9741 stays in `@tldraw/sync`. dotcom no longer uses it, but it's the in-band signal SDK consumers need to feed a server-decided comment grant into the plugin's `canComment`.

Follow-ups noted during review, deliberately not in this PR: pinning `objectAccess: 'read'` for legacy (non-app) rooms as belt-and-braces (currently inert — legacy rooms are always readonly, so the `canComment` default blocks), and a DO-level test for the inline fail-closed tier handling.

### Change type

- [x] `improvement`

### Test plan

1. As the owner, share a file as Editor and open it as a different signed-in user: comment tool works, post/reply/resolve all function.
2. Switch the file to Viewer: the guest reconnects, the canvas goes read-only, the comment tool and composers disappear, existing threads stay readable.
3. Switch back to Editor: the guest regains both.
4. As a signed-in guest on a Viewer link, forge a comment write from the console: the server rejects it and the client self-corrects.
5. As a signed-out visitor on an Editor link: the comment button opens the sign-in dialog, and opening a thread shows the "sign in to comment" prompt.
6. As a signed-out visitor on a Viewer link: no comment button, no prompt, threads readable.

- [x] Unit tests
- [ ] End to end tests

### API changes

- Changed `TLRecordAuthorizer` (`@tldraw/sync-core`): the session argument now carries `isReadonly: boolean` — the canvas-lane readonly state the room already tracks per session.
- Added `CommentAuthorizerOptions.canComment` (`@tldraw/sync-collaboration`): per-session comment write policy, defaulting to blocking canvas-readonly sessions.

### Release notes

- Fix view-only guests being able to comment; comment write policy now lives in the record authorizers and follows canvas access by default.

### Code changes

| Core code       | +46 / -10  |
| Tests           | +214 / -60 |
| Automated files | +7 / -0    |
| Apps            | +41 / -132 |


mimecuvalo (approved):
one issue before landing, pre-approving though:
 View-only guests get clickable reaction pills that the server rejects — packages/commenting/src/canvas/thread-view.tsx:385

This is the real fallout of dropping composeUserId. CommentsOnCanvas now passes the real currentUserId down unconditionally, and the reaction tally row is gated on identity, not permission:

// thread-view.tsx:385
<CommentReactions comment={comment} currentUserId={currentUserId} ... />
// comment-reactions.tsx:187
canReact={currentUserId != null}

A signed-in view-only guest opens a thread that already has reactions → the pills are live. Clicking one calls toggleCommentReaction, writes a comment-reaction record optimistically, the authorizer vetoes it, and the client snaps back. Every other write affordance in that file is behind canComment (canReply at :185, actions at :392, headerActions/CommentReactionPicker at :439/:451) — the tally row is the one that escaped, and it contradicts the option's own contract ("the action affordances are hidden").

Fix in the toolkit, not dotcom, so all consumers get it:

<CommentReactions
      comment={comment}
      currentUserId={canComment ? currentUserId : null}
      resolveName={resolveName}
/>

Same check is worth applying wherever CommentReactions renders outside the thread view.

MitjaBezensek:
Fixed in 9356fe92c0 — the thread footer now passes `currentUserId={canComment ? currentUserId : null}` to `CommentReactions`, so the tally renders read-only for view-only sessions (no live pills, no optimistic write to snap back). Checked the other render paths: `CommentReactionPicker` at thread-view.tsx:409 already sits inside the `canComment &&` actions block, and there are no `CommentReactions` usages outside the thread view — the footer row was the only leak. Side effect of nulling identity there: view-only guests lose the own-reaction highlight and "you" naming on pills, which matches the pre-existing `composeUserId` behavior this branch removed.

